PR #1863, opened 2017-07-18
update reference and how to cite

tjhei:
I am updating homepage and deal.II website too...

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -298,16 +298,19 @@ relevance.
 In addition, many have contributed to the development of \aspect{} and deserve credit
 for their work.
 To this end, we ask that if you publish results that were obtained
-to some part using \aspect{}, cite the following, canonical reference for
+to some part using \aspect{}, cite the following, canonical reference \cite{heister_aspect_methods2} for
 this software:
 \begin{lstlisting}[frame=single,language=tex]
-@Article{KHB12,
-  author =       {M. Kronbichler and T. Heister and W. Bangerth},
-  title =        {High Accuracy Mantle Convection Simulation through Modern Numerical Methods},
-  journal =      {Geophysics Journal International},
-  year =         2012,
-  volume =       191,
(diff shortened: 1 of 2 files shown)


doc/manual/manual.tex
```diff
@@ -298,16 +298,19 @@ \subsection{Referencing \aspect{}}
 In addition, many have contributed to the development of \aspect{} and deserve credit
 for their work.
 To this end, we ask that if you publish results that were obtained
-to some part using \aspect{}, cite the following, canonical reference for
+to some part using \aspect{}, cite the following, canonical reference (\cite{heister_aspect_methods2}) for
```
bangerth: Do we need the extra parentheses?

doc/manual/manual.tex
```diff
@@ -298,16 +298,19 @@ \subsection{Referencing \aspect{}}
...
 \begin{lstlisting}[frame=single,language=tex]
-@Article{KHB12,
-  author =       {M. Kronbichler and T. Heister and W. Bangerth},
-  title =        {High Accuracy Mantle Convection Simulation through Modern Numerical Methods},
-  journal =      {Geophysics Journal International},
-  year =         2012,
-  volume =       191,
-  pages =        {12--29}
+@Article{heister_aspect_methods2,
+  Title                    = {{High Accuracy Mantle Convection Simulation through Modern Numerical Methods. II: Realistic Models and Problems}},
```
bangerth: Cut the extra braces and put them only around `{II}` and `{R}ealistic`

tjhei: Then all words except "High" and "Realistic" are lower case. Is that what you intend to have (see title before)?

bangerth: Correct. That's what the typical styles all suggest to happen for article titles. (They differ on how to treat book titles.) That's why bibtex does it automatically.

tjhei:
fixed.

PR #1864, opened 2017-07-18
[WIP] Vankeken et al. 2008 subduction benchmark

maxrudolph:
This is an attempt to reproduce the benchmark Case 1a from Van Keken et al. 2008. The code is contributed by @maxrudolph and @jperryhouts. The comparison with the Van Keken benchmark values is ongoing.


source/simulator/core.cc
```diff
@@ -1043,6 +1046,7 @@ namespace aspect
             {
               coupling[x.velocities[d]][x.pressure] = DoFTools::always;
               coupling[x.pressure][x.velocities[d]] = DoFTools::always;
+              coupling[x.pressure][x.pressure] = DoFTools::always;
```
bangerth: I think this ought to go into a separate pull request because it is logically independent. But I'm doubtful that that is correct -- the bottom right block of the Stokes system is empty, so why do you need this entry?

cookbooks/prescribed_velocity/CMakeLists.txt
```diff
@@ -35,5 +35,5 @@ DEAL_II_INITIALIZE_CACHED_VARIABLES()
 SET(TARGET "prescribed_velocity")
 PROJECT(${TARGET})
 
-ADD_LIBRARY(${TARGET} SHARED prescribed_velocity.cc)
+ADD_LIBRARY(${TARGET} SHARED prescribed_velocity_pressure.cc)
```
bangerth: is this a mistake?

bangerth (commented):
I've marked up two things that look odd and should probably go into a separate PR.

As for the benchmark itself -- don't we already have that somewhere in the manual?

jdannberg:
> As for the benchmark itself -- don't we already have that somewhere in the manual?

I think this is a different van Keken Benchmark (we have the 1997 one with thermo-chemical convection, this is a subduction benchmark).

jperryhouts:
>> As for the benchmark itself -- don't we already have that somewhere in the manual?

> I think this is a different van Keken Benchmark (we have the 1997 one with thermo-chemical convection, this is a subduction benchmark).

Yep, this is different than the convection model from van Keken 1997. It's a corner flow benchmark. The flow field looks like:
screenshot from 2017-07-12 21-47-37
screenshot from 2017-07-12 21-58-48
I'm planning on writing a manual section about it soon.

source/simulator/core.cc
maxrudolph: So, we originally tried to prescribe pressures without doing this and when running in debug mode produced a trilinos error. I will try to reproduce again in a minute. I presumed that in order to constrain a pressure DOF, a 1 is inserted on the diagonal of the lower left block of the stokes system, and this entry is typically not allocated. After making this change, I no longer encounter the error. There may be an alternative solution?

maxrudolph:
The benchmark itself also involves a comparison of the wedge thermal structure, the temperature along the slab-wedge interface, and the temperature at the tip of the wedge. I will update the PR to include this comparison shortly and @jperryhouts can include this in the manual.

source/simulator/core.cc
maxrudolph: `max@hood:/opt/aspect/benchmarks/vankeken_subduction$ ./aspect vankeken_subduction.prm 
--     . version 2.0.0-pre
--     . using Trilinos

Loading shared library <./libprescribed_velocity_pressure.so>

Number of active cells: 7,040 (on 4 levels)
Number of degrees of freedom: 92,700 (56,994+7,209+28,497)

*** Timestep 0:  t=0 years
   Solving temperature system... 0 iterations.
------------------------------------------
Got error 2 in row 2 of proc 0 when trying to add the columns:
2 

Matrix row  has the following indices:



An error occurred in line <1717> of file </opt/aspect-stack/tmp/unpack/deal.II-master/source/lac/trilinos_sparse_matrix.cc> in function
    void dealii::TrilinosWrappers::SparseMatrix::add(dealii::TrilinosWrappers::SparseMatrix::size_type, dealii::TrilinosWrappers::SparseMatrix::size_type, const size_type*, const TrilinosScalar*, bool, bool)
The violated condition was: 
    ierr <= 0
Additional information: 
    You tried to access element (2/2) of a sparse matrix, but it appears to not exist in the Trilinos sparsity pattern.

Stacktrace:
-----------
#0  /opt/aspect-stack/deal.II-master/lib/libdeal_II.g.so.9.0.0-pre: dealii::TrilinosWrappers::SparseMatrix::add(unsigned int, unsigned int, unsigned int const*, double const*, bool, bool)
#1  ./aspect: dealii::BlockMatrixBase<dealii::TrilinosWrappers::SparseMatrix>::add(unsigned int, unsigned int, double)
#2  /opt/aspect-stack/deal.II-master/lib/libdeal_II.g.so.9.0.0-pre: void dealii::ConstraintMatrix::distribute_local_to_global<dealii::TrilinosWrappers::BlockSparseMatrix, dealii::TrilinosWrappers::MPI::BlockVector>(dealii::FullMatrix<dealii::TrilinosWrappers::BlockSparseMatrix::value_type> const&, dealii::Vector<dealii::TrilinosWrappers::MPI::BlockVector::value_type> const&, std::vector<unsigned int, std::allocator<unsigned int> > const&, dealii::TrilinosWrappers::BlockSparseMatrix&, dealii::TrilinosWrappers::MPI::BlockVector&, bool, dealii::internal::bool2type<true>) const
#3  ./aspect: aspect::Simulator<2>::copy_local_to_global_stokes_system(aspect::internal::Assembly::CopyData::StokesSystem<2> const&)
#4  ./aspect: void dealii::WorkStream::run<std::_Bind<std::_Mem_fn<void (aspect::Simulator<2>::*)(dealii::TriaActiveIterator<dealii::DoFCellAccessor<dealii::DoFHandler<2, 2>, false> > const&, aspect::internal::Assembly::Scratch::StokesSystem<2>&, aspect::internal::Assembly::CopyData::StokesSystem<2>&)> (aspect::Simulator<2>*, std::_Placeholder<1>, std::_Placeholder<2>, std::_Placeholder<3>)>, std::_Bind<std::_Mem_fn<void (aspect::Simulator<2>::*)(aspect::internal::Assembly::CopyData::StokesSystem<2> const&)> (aspect::Simulator<2>*, std::_Placeholder<1>)>, dealii::FilteredIterator<dealii::TriaActiveIterator<dealii::DoFCellAccessor<dealii::DoFHandler<2, 2>, false> > >, aspect::internal::Assembly::Scratch::StokesSystem<2>, aspect::internal::Assembly::CopyData::StokesSystem<2> >(dealii::FilteredIterator<dealii::TriaActiveIterator<dealii::DoFCellAccessor<dealii::DoFHandler<2, 2>, false> > > const&, dealii::identity<dealii::FilteredIterator<dealii::TriaActiveIterator<dealii::DoFCellAccessor<dealii::DoFHandler<2, 2>, false> > > >::type const&, std::_Bind<std::_Mem_fn<void (aspect::Simulator<2>::*)(dealii::TriaActiveIterator<dealii::DoFCellAccessor<dealii::DoFHandler<2, 2>, false> > const&, aspect::internal::Assembly::Scratch::StokesSystem<2>&, aspect::internal::Assembly::CopyData::StokesSystem<2>&)> (aspect::Simulator<2>*, std::_Placeholder<1>, std::_Placeholder<2>, std::_Placeholder<3>)>, std::_Bind<std::_Mem_fn<void (aspect::Simulator<2>::*)(aspect::internal::Assembly::CopyData::StokesSystem<2> const&)> (aspect::Simulator<2>*, std::_Placeholder<1>)>, aspect::internal::Assembly::Scratch::StokesSystem<2> const&, aspect::internal::Assembly::CopyData::StokesSystem<2> const&, unsigned int, unsigned int)
#5  ./aspect: aspect::Simulator<2>::assemble_stokes_system()
#6  ./aspect: aspect::Simulator<2>::solve_timestep()
#7  ./aspect: aspect::Simulator<2>::run()
#8  ./aspect: main

[hood:28671] *** Process received signal ***
[hood:28671] Signal: Aborted (6)
[hood:28671] Signal code:  (-6)
[hood:28671] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x11390)[0x7f046f414390]
[hood:28671] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x38)[0x7f046e9e3428]
[hood:28671] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x16a)[0x7f046e9e502a]
[hood:28671] [ 3] /opt/aspect-stack/deal.II-master/lib/libdeal_II.g.so.9.0.0-pre(_ZN6dealii18deal_II_exceptions9internals5abortERKNS_13ExceptionBaseEb+0x11f)[0x7f047669fc51]
[hood:28671] [ 4] /opt/aspect-stack/deal.II-master/lib/libdeal_II.g.so.9.0.0-pre(_ZN6dealii18deal_II_exceptions9internals11issue_errorINS_16TrilinosWrappers12SparseMatrix28ExcAccessToNonPresentElementEEEvNS1_17ExceptionHandlingEPKciS8_S8_S8_T_+0x36)[0x7f04765b39be]
[hood:28671] [ 5] /opt/aspect-stack/deal.II-master/lib/libdeal_II.g.so.9.0.0-pre(_ZN6dealii16TrilinosWrappers12SparseMatrix3addEjjPKjPKdbb+0xb7f)[0x7f04765a704b]
[hood:28671] [ 6] ./aspect(_ZN6dealii15BlockMatrixBaseINS_16TrilinosWrappers12SparseMatrixEE3addEjjd+0x32e)[0xe21680]
[hood:28671] [ 7] /opt/aspect-stack/deal.II-master/lib/libdeal_II.g.so.9.0.0-pre(_ZNK6dealii16ConstraintMatrix26distribute_local_to_globalINS_16TrilinosWrappers17BlockSparseMatrixENS2_3MPI11BlockVectorEEEvRKNS_10FullMatrixINT_10value_typeEEERKNS_6VectorINT0_10value_typeEEERKSt6vectorIjSaIjEERS7_RSD_bNS_8internal9bool2typeILb1EEE+0xda7)[0x7f04760cf2a5]
[hood:28671] [ 8] ./aspect(_ZN6aspect9SimulatorILi2EE34copy_local_to_global_stokes_systemERKNS_8internal8Assembly8CopyData12StokesSystemILi2EEE+0x42)[0xe26c7a]
[hood:28671] [ 9] ./aspect(_ZN6dealii10WorkStream3runISt5_BindIFSt7_Mem_fnIMN6aspect9SimulatorILi2EEEFvRKNS_18TriaActiveIteratorINS_15DoFCellAccessorINS_10DoFHandlerILi2ELi2EEELb0EEEEERNS4_8internal8Assembly7Scratch12StokesSystemILi2EEERNSG_8CopyData12StokesSystemILi2EEEEEPS6_St12_PlaceholderILi1EEST_ILi2EEST_ILi3EEEES2_IFS3_IMS6_FvRKSN_EESS_SU_EENS_16FilteredIteratorISC_EESJ_SN_EEvRKT1_RKNS_8identityIS18_E4typeET_T0_RKT2_RKT3_jj+0x345)[0xe76817]
[hood:28671] [10] ./aspect(_ZN6aspect9SimulatorILi2EE22assemble_stokes_systemEv+0x44a)[0xe777d4]
[hood:28671] [11] ./aspect(_ZN6aspect9SimulatorILi2EE14solve_timestepEv+0x326)[0xf03ddc]
[hood:28671] [12] ./aspect(_ZN6aspect9SimulatorILi2EE3runEv+0x10c3)[0xf06639]
[hood:28671] [13] ./aspect(main+0x7b4)[0xf74f22]
[hood:28671] [14] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0)[0x7f046e9ce830]
[hood:28671] [15] ./aspect(_start+0x29)[0x8bb249]
[hood:28671] *** End of error message ***
Aborted
max@hood:/opt/aspect/benchmarks/vankeken_subduction$ 
`

cookbooks/prescribed_velocity/CMakeLists.txt
maxrudolph: Yes

bangerth:
Oh, I see. Thanks for pointing this out -- this is one of the first benchmarks I attempted when I learned about many of these things on a visit with Scott King many years ago. I failed at the time because we couldn't prescribe a velocity field, and this benchmark crucially depends on that ability.

So is the plan to keep this pull request open until the section in the manual is written, or do you want us to consider it independently?

source/simulator/core.cc
bangerth: I see. I think that's still an independent issue that would be easiest to discuss in a separate pull request. Can you separate it into one, and copy paste your explanation and the error message? I don't think the issue is trivial.

jperryhouts:
> So is the plan to keep this pull request open until the section in the manual is written, or do you want us to consider it independently?

I think either way is fine. In the words of James Comey, the two are "touching each other but separate." You can wait for the manual section if they will be easier to review together.

bangerth:
Let's keep things together and review it all together, if you don't mind.

maxrudolph:
I think that keeping it together makes most sense.

source/simulator/core.cc
tjhei: > tried to prescribe pressures

Interesting. I don't like having to add coupling for the pressures but I think this should work without. Do you really need to prescribe pressures? This is not something typically done for the Stokes problem.

maxrudolph: It is necessary either to prescribe pressures or ignore the residual in the kinematically prescribed regions. Because there is a discontinuity in the prescribed velocity along the subduction interface, very large spurious pressures develop here and the solver will fail to converge. After prescribing pressure within the kinematically prescribed regions, we have no convergence issues. This insight was based on my past experiences with my finite volume code.

tjhei: I think the reason is that your signal only modifies ``current_constraints`` but you also need to modify ``constraints``, which is the ``ConstraintMatrix`` used for construction the sparsity pattern of the matrix. i think ``signals.pre_compute_no_normal_flux_constraints`` should work.

maxrudolph: I disagree that it makes no sense to prescribe pressure.
If you have a velocity field that contains piecewise rigid body motions everywhere in the domain, the correct solution is that the (dynamic) pressure is zero everywhere in the domain. However, the finite element space cannot represent discontinuous velocity fields, so along the interface between rigid body regions (i.e. the subduction interface in this problem) there will be very large artificial pressure oscillations in order to satisfy momentum balance.

maxrudolph: @tjhei I looked into `signals.pre_compute_no_normal_flux_constraints` and that signal only takes a Triangulation as argument. Can I access the ConstraintMatrix `constraints` through this object?

tjhei: Oh, that is true. No, there is no way to access them currently. We could add this to the signal. A better option might be to also compute the ``current_constraints`` before constructing the matrices. That way you don't need to change the signals.

maxrudolph: Or make a new signal that takes the same arguments as post_constraints_creation? I noticed that the pressure coupling IS already present in the preconditioner matrix, presumably because for a cheap preconditioner something like 1/eta is put on the diagonal of the pressure block?

tjhei: I prefer fixing it without a new signal as described above. We use the 1,1 block to store the Schur complement in the preconditioner.

maxrudolph: What about just calling `signals.post_constraints_creation(*this, constraints)`
after line 1397 in core.cc (i.e. within setup_dofs, before `constraints.close()` is called?

bangerth: @maxrudolph: Regarding "If you have a velocity field that contains piecewise rigid body motions everywhere in the domain, the correct solution is that the (dynamic) pressure is zero everywhere in the domain." -- that's not actually correct. Rather, the pressure needs to equal the gravitational potential plus an arbitrary constant, so that it counters the right hand side forces.

But you don't need to prescribe the pressure. If that's what the velocity field is, shouldn't the rest of the equations just happen to yield that pressure?

maxrudolph: I agree that the total pressure should include a contribution from gravitational potential, but the *dynamic* pressure does not. In this benchmark problem, we turned off gravity so the total pressure is the dynamic pressure. It is a purely kinematic flow benchmark in which buoyancy forces do not matter.
I think that if the discontinuity was represented correctly by the finite element, the equations should just yield the correct pressure, but in our case, there is a step discontinuity in velocity that couples into the pressure field because the laplacian of the velocity field is nonzero in the finite element approximation. I've produced as simple an example as I can that demonstrates this effect. Note that I am using initial global refinement level 1. For any higher refinement level, I see convergence problems. Note that I only prescribe the velocity in the upper left quadrant of the domain - I found that if velocity is prescribed everywhere I get division by zero errors during the stokes solve.
[pressure_for_wolfgang.prm.zip](https://github.com/geodynamics/aspect/files/1163171/pressure_for_wolfgang.prm.zip)



image

bangerth: You discover one of the reasons why I think that that benchmark is so poorly posed :-( It just doesn't fit into the typical finite element method unless you have a specially constructed code and a particular mesh that resolves the interface. I don't think the problem is mathematically well posed.

In this benchmark, the velocity needs to be continuous -- you prescribe it in parts of the domain and that forms velocity boundary conditions for the rest of the domain. But in general, the pressure will be discontinuous for this setup: if there is no gravity, then in the "special" part of the domain where you have rigid body motion, the pressure will be constant, whereas in general it will of course no be constant along the boundary of the special domain and may in fact be singular (in corners). 

I think this benchmark is poorly designed and mathematically not well posed. I've refused to touch it for the past decade since learning about it from Scott, but I recognize that others may nevertheless want to reproduce it. (I'm going to go on the record, though, saying that you can't expect good convergence from ASPECT unless you align the mesh with the interface.) The question is just how we can make this work in ASPECT without distorting ourselves and putting too many things into the code that are only applicable to this kind of benchmark. If there is a signal we can attach code to in a plugin that's specific to this benchmark, that's ok with me. But I'd rather not us change a core piece of the code just to accommodate this one benchmark because that's a slope I don't want us to get on.

maxrudolph: Yes, I agree that the problem is not well-posed for the reasons that you stated!

I think that the solution proposed by @tjhei in #1865 is probably the way to go. I think that if you allow that pull request (once the errors are resolved), this benchmark will be possible without further modification of the core routines. Attaching an additional signal is also fine with me.

I have seen some *interesting* convergence behavior with this benchmark. I am still working on the correct refinement criteria to deliver an acceptable match to Van Keken's values. I did align the mesh to the slab interface. One of the numbers required for the benchmark is the temperature at a specific point (tip of the wedge) and it is probably not possible to get good agreement without this point aligning with a node.

bangerth: I'm not surprised by the interesting convergence behavior. I do find these cases where something doesn't happen "as one would expect" often make for good teaching situations because one can explain what is going wrong and why, and then how to fix it. That's what I did in the section in the manual on the other (thermochemical) van Keken benchmark. I typically find it more instructive to also show the situations that don't work, before showing how to make it work.

I have no objections to #1865 or a variation thereof.

maxrudolph:
benchmark_quantities
This is the somewhat bizarre convergence behavior that I am seeing so far with this benchmark. I do not quite understand what is going on here but it appears that there is not monotonic convergence towards the benchmark values. In each plot, the dashed red and blue lines indicate the highest and lowest values reported in Table 2 of Van Keken et al. 2008. Changes in behavior are associated with refinement at specified additional refinement times. I am now re-running with an additional minimum refinement function criterion applied to the tip of the wedge to see whether this is significant. @jperryhouts @bangerth

bangerth:
I bet every improvement is associated with refining the cells in the vicinity where it matters for whatever we compute, and every worsening with coarsening some cells. Alternatively, with having a regular mesh whereas one with hanging nodes in the vicinity of where it matters.

maxrudolph:
benchmark_quantities

This image shows the convergence behavior for a run with slightly different alignment of the mesh and definition of the prescribed kinematic region. It turns out that using ASPECT, it really matters whether the kinematic region is defined as x<(H-y) or x<=(H-y). This may well be because the pressure nodes are included or excluded from the kinematic region accordingly. I have one more idea that I want to explore before accepting this as the best that I can achieve with ASPECT.

maxrudolph:
@bangerth @jperryhouts I have one additional question. I tried to use the locally conservative discretization with the prescribed velocity plugin and the plugin appears to have no effect. Do you understand why this is the case, and would it be difficult to fix this?

tjhei:
> I tried to use the locally conservative discretization with the prescribed velocity plugin and the plugin appears to have no effect.

Are you talking about the ``constrain_internal_velocities`` in this PR or something else? What does "no effect" mean?

maxrudolph:
@tjhei I tested it with the `prescribed_velocity_pressure.cc` in this PR, which is derivative from the `cookbooks/prescribed_velocity/prescribed_velocity.cc`. I can produce some plots if that would help.

tjhei:
> I tested it with the prescribed_velocity_pressure.cc in this PR

The DG element of the pressure has no support points so you won't be able to prescribe pressure values. From a quick look at the code, I think prescribing velocity values should work.

maxrudolph:
aspect_vankeken_difference

This is the difference between my values and the van keken benchmark values for vx, vy, and T.

tjhei:
> This is the difference between my values and the van keken benchmark values for vx, vy, and T.

I am not sure what that means. Are no values prescribed at all or can this be explained by the fact that pressure is not prescribed?

maxrudolph:
>Are no values prescribed at all or can this be explained by the fact that pressure is not prescribed?

No values are prescribed at all (except for the BCs).

maxrudolph:
@jperryhouts @bangerth @tjhei @jdannberg I wrote a manual section for the van Keken benchmark and squashed this PR into a single commit. This is basically as far as I have time to take this benchmarking effort right now. I am a bit disappointed that I wasn't able to achieve better agreement between the ASPECT results and other codes, despite using very fine mesh resolution and taking extremely small time steps for the highest-resolution part of the run. I am wondering whether using the 'locally conservative discretization' might improve accuracy, but this does not seem to be compatible with the prescribed velocity plugin. Running the benchmark requires the ability to modify the sparsity pattern within the plugin ( #1865 ). If you have ideas about how the convergence towards benchmark values might be improved, or how I might change the mesh refinement criteria to achieve this with less computer time, this would be most welcome. The run shown in the manual figure took 3 days on 100 cores (mostly at the finest mesh resolution where time steps become tiny).

doc/manual/manual.tex
```diff
@@ -8182,14 +8182,77 @@ \subsubsection{Onset of convection benchmark}
...
+
+This benchmark involves a solution to a kinematic corner flow problem that approximates some aspects
+of flow in the mantle wedge. The specific case considered here is `Case 1C' from van Keken et al. \cite{VK08}.
+The flow in this model is driven by a kinematically-prescribed downgoing plate and a kinematically fixed over-riding
+plate. The dip angle of the downgoing plate is 45$^\circ$ and the rate of convergence is 5 cm/yr. The thermal structure of
+the subducing plate is initially described by a halfspace-cooling solution for a plate age of 50 Myr. The over-riding plate is
+50 km thick and has a conductive temperature profile along the right boundary. The top boundary is isothermal. The right boundary and the bottom-right
+corner of the bottom boundary have the natural (stress-free) mechanical boundary condition. Incoming material on the right boundary is isothermal.
+
+In order to treat the kinematically-prescribed regions within the domain, we use a modified version of the `prescribed-velocity'
```
jperryhouts: Maybe add reference to the section in the manual which describes the prescribed velocity plugin (\ref{sec:prescribed-velocities}). It isn't included in the code; it only exists in the manual right now.

jperryhouts:
Thanks for doing that @maxrudolph ! Sorry I didn't get to the manual section. I went straight from CIDER to field camp, and turns out I don't get as much computer access  here as I expected I would. At any rate, the manual section looks really good. I made one small comment on it, but otherwise looks great.

maxrudolph:
@tjhei I looked into the issue with the locally-conservative discretization a bit more.
After doing some more calculations, I plotted streamlines in paraview (and other software) and noticed that the behavior of the streamlines near the prescribed regions is non-physical - there is clearly a velocity component perpendicular to the prescribed region's boundary when in fact this should be a no-slip interface. So, my working hypothesis is that the lack of local mass conservation may in fact be contributing to an inaccurate solution. Other FE codes participated in the van Keken 2008 benchmark study, including Scott King's code which used Q1-P0 quadrilaterals and produced results more consistent with the other codes at much coarser mesh resolutions than I used with ASPECT. So, before giving up on ASPECT and using my FV code, I would like to give this one more shot.

I did a simple test to attempt to isolate the issue associated with prescribed velocities, using the `cookbooks/prescribed_velocity` plugin. I modified `cookbooks/prescribed_velocity/corner_flow.prm` to use the locally conservative discretization, and in debug mode there is an error message.


```An error occurred in line <343> of file </opt/aspect/source/main.cc> in function
    void parse_parameters(const string&, dealii::ParameterHandler&)
The violated condition was: 
    false
Additional information: 
    Invalid input parameter file.

Aborting!
[Thread 0x7fffdf1dc700 (LWP 24404) exited]
[Inferior 1 (process 24400) exited with code 01]
(gdb) run corner_flow.prm
Starting program: /opt/aspect/cookbooks/prescribed_velocity/aspect corner_flow.prm
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffdf1dc700 (LWP 24419)]
--     . version 2.0.0-pre
--     . using Trilinos

Loading shared library <./libprescribed_velocity.so>

Number of active cells: 16,384 (on 8 levels)
Number of degrees of freedom: 115,715 (33,282+16,384+66,049)


An error occurred in line <2269> of file </opt/aspect-stack/tmp/unpack/deal.II-v8.5.0/source/fe/fe_values.cc> in function
    dealii::FEValuesBase<dim, spacedim>::FEValuesBase(unsigned int, unsigned int, dealii::UpdateFlags, const dealii::Mapping<dim, spacedim>&, const dealii::FiniteElement<dim, spacedim>&) [with int dim = 2; int spacedim = 2]
The violated condition was: 
    n_q_points > 0
Additional information: 
    There is nothing useful you can do with an FEValues object when using a quadrature formula with zero quadrature points!

Stacktrace:
-----------
#0  /opt/aspect-stack/deal.II-v8.5.0/lib/libdeal_II.g.so.8.5.0: dealii::FEValuesBase<2, 2>::FEValuesBase(unsigned int, unsigned int, dealii::UpdateFlags, dealii::Mapping<2, 2> const&, dealii::FiniteElement<2, 2> const&)
#1  /opt/aspect-stack/deal.II-v8.5.0/lib/libdeal_II.g.so.8.5.0: dealii::FEValues<2, 2>::FEValues(dealii::FiniteElement<2, 2> const&, dealii::Quadrature<2> const&, dealii::UpdateFlags)
#2  ./libprescribed_velocity.so: void aspect::constrain_internal_velocities<2>(aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&)
#3  ./libprescribed_velocity.so: boost::detail::function::void_function_invoker2<void (*)(aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&), void, aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&>::invoke(boost::detail::function::function_buffer&, aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&)
#4  /opt/aspect/cookbooks/prescribed_velocity/aspect: boost::signals2::detail::signal_impl<void (aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&), boost::signals2::optional_last_value<void>, int, std::less<int>, boost::function<void (aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&)>, boost::function<void (boost::signals2::connection const&, aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&)>, boost::signals2::mutex>::operator()(aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&)
#5  /opt/aspect/cookbooks/prescribed_velocity/aspect: aspect::Simulator<2>::compute_current_constraints()
#6  /opt/aspect/cookbooks/prescribed_velocity/aspect: aspect::Simulator<2>::set_initial_temperature_and_compositional_fields()
#7  /opt/aspect/cookbooks/prescribed_velocity/aspect: aspect::Simulator<2>::run()
#8  /opt/aspect/cookbooks/prescribed_velocity/aspect: main


Thread 1 "aspect" received signal SIGABRT, Aborted.
0x00007fffecf9d428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
```

So, I did a backtrace to look for the potential problem (below) and interrogated the Quadrature object. The quadrature size is indeed 0. Note that `quadrature` here is returned by `get_unit_support_points()`. I was somewhat surprised by this as I thought that for this element, there should be unit support points at least for the velocity components (I understand that there is no unit support point for pressure). Without the unit support points, I don't see a way to prescribe velocities, unless there is another way to access the nodal coordinates associated with the cell's velocity DOFs. Any idea about how to proceed here? Thanks for your help.

```(gdb) bt
...
(gdb) up 6
#6  0x00007fffd4a21bc5 in aspect::constrain_internal_velocities<2> (simulator_access=..., 
    current_constraints=...) at /opt/aspect/cookbooks/prescribed_velocity/prescribed_velocity.cc:124
124	        FEValues<dim> fe_values (simulator_access.get_fe(), quadrature, update_q_points);

print quadrature.size()
$1 = 0
```

tjhei:
> . I was somewhat surprised by this as I thought that for this element, there should be unit support points at least for the velocity components

Right. An FE_System will return no support points if one of the contained elements has no support points.  Can you try
``simulator_access.get_fe().base_element(0).get_unit_support_points()`` instead?

maxrudolph:
@tjhei Thanks for your help. That does indeed enable us to prescribe velocities when using the locally conservative discretization with `set Stokes velocity polynomial degree = 1`. Is there a way to also constrain the pressure DOF for the cell? In `prescribe_velocity_pressure.cc`, we iterate:
``` // Loop over all cells                                                     
        for (cell = simulator_access.get_dof_handler().begin_active();
             cell != simulator_access.get_dof_handler().end();
             ++cell)
          if (! cell->is_artificial())
            {
              fe_values.reinit (cell);
              std::vector<unsigned int> local_dof_indices(simulator_access.get_fe(\
).dofs_per_cell);
              cell->get_dof_indices (local_dof_indices);

              for (unsigned int q=0; q<quadrature.size(); q++)
                // If it's okay to constrain this DOF                              
                if (prescribe_internal_velocities &&
                    current_constraints.can_store_line(local_dof_indices[q]) &&
                    !current_constraints.is_constrained(local_dof_indices[q]))
                  {
                    // Get the velocity component index                            
                    const unsigned int c_idx =
                      simulator_access.get_fe().system_to_component_index(q).first\
;  ```

For the DGP element, there is no quadrature point, but if I modify the loop to extend to `quadrature.size()+1`, would I access the degree of freedom corresponding to the pressure degree of freedom for this cell? I guess that I could then constrain the pressure DOF based on the cell center location, or just keep track of whether all of the velocity DOFs were constrained and if so, also constrain pressure.

**Comment by tjhei** [comment]
> Is there a way to also constrain the pressure DOF for the cell?

For generic DGP(k) elements this is difficult, but if you only need DGP(0), then yes:
``get_dof_indices()`` should return a single dof index and you can constrain it. I am not sure if the value you set is the mean value or the integral over the cell but you should be able to figure that out easily.

**Comment by bangerth** [comment]
Regarding the code
```
simulator_access.get_fe().base_element(0).get_unit_support_points() 
```
don't you have to repeat these points `dim` times to get the equivalent of what we have?

@maxrudolph -- this is really an independent issue. Can you open a new github issue in which you attach the input file you used to get the error and copy-paste the relevant part of the discussion above? It's easier to keep track of one issue at a time, and to write patches for one issue at a time.

gassmoeller:
@maxrudolph could you give me a quick heads up on the status of this PR? Is this essentially as far as we can bring this benchmark for now?

maxrudolph:
@gassmoeller I am not really satisfied with the results of this benchmark and would prefer not to merge the PR at this time. The pressure solution for the q1-p0 case is completely wrong and I do not understand why. This is somewhat surprising since one of the codes (ConMan) that participated in the van Keken et al. 2008 benchmark paper used q1p0 elements and was able to obtain a solution that appeared more consistent with the other codes. Is this something that you would be interested in looking at in a bit more detail with me, offline?

gassmoeller:
I was mostly concerned that you might be waiting for something to happen on our side. Lets wait with merging this PR then. I am pretty busy for the next two weeks, but afterwards I might have some time. At least I will have time to run the model, and see if there is anything I can think of how to improve the solution.

maxrudolph:
I may have some time later this week to prepare a more detailed summary of what I did and what problems I encountered. I might also try to discuss this with Scott King, who contributed to the van Keken benchmark paper to see if he has any ideas about how to approach this. I also have very limited time to invest in this right now but will try to make some as I think that even if we agree that this might not be a perfect benchmark, it is still important to understand how one might  best go about setting up a corner flow problem with ASPECT.

gassmoeller:
Sounds good. I will wait for your summary and we can have a discussion about possible improvements in two weeks or so.

gassmoeller:
I was looking at this problem for a while today (thanks @maxrudolph for the description pdf). Am I correct in stating that:
- Q1P0 does not work as well in ASPECT as in Conman since it does not stabilize the Q1P0 element in any way. It was not intended to be used with unstable elements, therefore the checkerboard pattern in pressure.
- Q2Q1 causes trouble with the velocities because it has a huge pressure gradient (in the bilinear continuous Q1 element) from prescribed velocity to wedge region, independent of prescribing pressure as 0 or not. This leads to velocity divergences, or streamlines *disappearing* in the overriding plate or slab regions (I verified that with a postprocessor, which I had made for another project, that simply outputs the velocity divergence in each cell).
- Q2P-1 does not work, because we do not know where support points of the P-1 element are.

I would argue that Q2P-1 would still be the best way to go. We can still prescribe values to the P-1 pressure element if my following assumption is correct: The shape function of the P-1 element are in essence (1,x,y), and we know the pressure for the 1 shape function (we want it to be zero), and we expect the change across one element in the region of prescribed velocity to be zero as well, so the dofs for shape function x and y are zero as well. Therefore a Q2P-1 element should be able to handle the problem except we simplify a bit, because each of our pressure elements is either completely in the prescribed domain, or completely outside, but that would be the same for a P0 element. 
Unfortunately at the tip of the wedge there is a singularity in the pressure field, and while the P0 element would likely ignore / do not see this singularity, the P-1 element might well be affected. So we might focus on using Q2P-1 and taking a look at what to do with the element at the tip of the wedge. I attach the pressure field of a computation with Q2P-1, and prescribed pressure (using the same region as the velocity dofs, not just the cells with fully prescribed velocity dofs). The streamlines look much better than the Q2Q1 model, and there is no checkerboard pattern in the pressure as in Q1P0. Note however, that I limited the color scale of the pressure. The pressure at the tip of the wedge was -1.1GPa, compared to around 10-30 Mpa in the rest of the domain. @maxrudolph I attached the modified plugin and parameter file, do you think it is worth running this model and look at the deviation to the benchmark results? Did you maybe do something like this already? We could also test first if we can improve things by maybe prescribing pressure in the element in the wedge (@bangerth: could we prescribe the dofs for shape functions x and y, and therefore force the pressure to be constant in that single element, or would that be a major mathematical crime? Would it help?)

visit0000

[vankeken_subduction_case8.prm.txt](https://github.com/geodynamics/aspect/files/1389309/vankeken_subduction_case8.prm.txt)

[prescribed_velocity_pressure.cc.txt](https://github.com/geodynamics/aspect/files/1389311/prescribed_velocity_pressure.cc.txt)

bangerth:
On 10/16/2017 05:24 PM, Rene Gassmoeller wrote:
>   *
>   * Q2P-1 does not work, because we do not know where support points of
>     the P-1 element are.
> 
> I would argue that Q2P-1 would still be the best way to go. We can still 
> prescribe values to the P-1 pressure element if my following assumption 
> is correct: The shape function of the P-1 element are in essence 
> (1,x,y), and we know the pressure for the 1 shape function (we want it 
> to be zero), and we expect the change across one element in the region 
> of prescribed velocity to be zero as well, so the dofs for shape 
> function x and y are zero as well. Therefore a Q2P-1 element should be 
> able to handle the problem except we simplify a bit, because each of our 
> pressure elements is either completely in the prescribed domain, or 
> completely outside, but that would be the same for a P0 element.

The right strategy would probably be to *project* onto that element, 
instead of interpolate. This way, you're not relying on how exactly the 
shape functions are defined.

bangerth:
> We could also test first if we can improve things by maybe prescribing 
> pressure in the element in the wedge (@bangerth 
> <https://github.com/bangerth>: could we prescribe the dofs for shape 
> functions x and y, and therefore force the pressure to be constant in 
> that single element, or would that be a major mathematical crime? Would 
> it help?)

Yes, well, as already discussed elsewhere, prescribing the pressure is a 
major crime anyway because you can then no longer guarantee that the 
velocity remains divergence free. But if you insist on it (and this 
benchmark calls for that), then I think projecting onto the finite 
element space is the right thing to do.

I don't recall how exactly the shape functions are defined. It is 
probably (hopefully) something like `{1,(x-x0),(y-y0)}`, where `x0,y0` 
is the center of the element. But it could also be a linear combination 
of these functions. In any case, you don't want to rely on how exactly 
these shape functions are defined, and you don't need to if you compute 
the projection of the function you're looking at onto the finite element 
space -- that's a local operation that's relatively easy to compute.

gassmoeller:
I noticed two more things today:
- The adaptive refinement criterion is currently only based on temperature. This leads to coarsening at the tip of the wedge later during the model run, and consequently to visible jumps in the solution. I can not quantify the influence yet, since I am lacking matplotlib here, but will try to come back to this later this week. Adding `temperature` and `velocity` to the refinement strategy stabilizes the solution (as does using a uniform grid without refinement/coarsening for the whole time).
- The hdf5 output has a bug in cases of discontinuous elements and since the hdf5 output is used for this benchmark it might affect the accuracy of the output. This will only affect the *output* of the solution, not the solution itself. I will open a separate issue for that, with a reference to this discussion.

In general, the attached parameter file is my best guess for an accurate solution right now (when the changes to the hdf5 output I will open separately are included). Note that I have replaced the somewhat unreliable `x <= H` functions with something like `x < H+eps`. For a convergence study we can let `eps` go to zero with increased resolution.

[vankeken_subduction_case8.prm.txt](https://github.com/geodynamics/aspect/files/1392479/vankeken_subduction_case8.prm.txt)

bangerth:
> Note that I have replaced the somewhat unreliable |x <= H| functions 
> with something like |x < H+eps|. For a convergence study we can let 
> |eps| go to zero with increased resolution.

You're almost certainly on the safe side with eps=1e-8*diameter of the 
domain, and you can keep that fixed since you're never going to have a 
mesh size that's smaller than that.

maxrudolph:
@gassmoeller Thank you for your help with this benchmark effort. I agree that the Q_2P_{-1} element produced the most reasonable result of the possibilities that we have tried. Thank you also for your updates to the prescribed_velocity_pressure plugin. 

Unfortunately, we are still not obtaining a reasonable velocity field near the corner of the mantle wedge. In the revised pdf in 504f1dc that I just pushed, I have re-generated all of the plots of vx and vy mismatch between our solutions and what PvK provided. There is a really obvious band of large velocity errors (both vx and vy) that extend away from the corner in all calculations. These plots also make the checkerboarding admitted by the discontinuous pressure elements more obvious (e.g. fig 5). Finally, I ran a case (based on Rene's benchmark setup) to nearly thermal steady state at global refinement level 4. I show the streamlines from this calculation and from the van keken benchmark in Figure 7. In the ASPECT solution, the behavior of the streamlines right near the corner does not appear consistent with van Keken's solution. The temperature benchmark values (Figure 8) are still much too cold. 

I am wondering also whether the boundary conditions on the right side of the domain are enforced correctly. Can the prescribed velocity plugin supercede the imposed boundary conditions? The right boundary is stress free, but the over-riding plate has prescribed zero velocity. I'm actually not sure which constraint takes precedence here and it has the potential to affect the entire velocity field.

bangerth:
I'm reading through open and unmerged pull requests, coming also to this one. This has been open for 7 years, but @maxrudolph has pushed some new commits about 3 years ago :-)

What do we want to do with this? My take is that this benchmark is a particularly poorly suited benchmark for finite element methods. It does not fit well into general codes in which it is hard to "hack in" certain conditions that make no physical sense, and as a consequence it has always been quite difficult to reproduce this benchmark for everyone who has tried. I'm not entirely sure it is useful to our users to explain in great detail how one could try and reproduce this benchmark with ASPECT, and I'm also not entirely convinced that ensuring that ASPECT can reproduce it will help us learn anything new about ASPECT (such as convince ourselves that it is correct -- or that perhaps it is *not* correct in certain regards).

As a consequence, my suggestion would be to close this PR and just admit that while the benchmark had some historical relevance, it cannot easily be reproduced with ASPECT because of its specific set-up.

danieldouglas92:
@bangerth At the last hackathon we implemented the van Keken model setup (#5259) with a special dealii mesh that improved convergence behaviour, and we started with the input file that @maxrudolph used in this PR. However, we didn't do the benchmarking that @maxrudolph did here. Something to think about before deciding whether this PR gets closed or not!

maxrudolph:
@bangerth wow, time flies! I agree that the benchmark problem is not well suited to FEM. However, maybe it's worth trying briefly to reproduce the benchmark values using @danieldouglas92 mesh before we abandon this altogether and close the PR? These kinematic corner flow models do get used quite a bit still, so I think that there is value in either providing a cookbook example for how one might approach this type of model or to document that this problem is not well-suited for solution using ASPECT (so that some poor future graduate student does not waste their time or become mislead by incorrect results).

danieldouglas92:
@maxrudolph Yes as a poor graduate student who tried getting this setup working for an AGU abstract last year I think it would be great to see how well the ASPECT models compare to the benchmark before continuing down this rabbit hole ;) I would be happy to help in any way I can in this effort.

bangerth:
That would find my favor!

danieldouglas92:
@maxrudolph I talked to Wolfgang today about this PR and if you aren't opposed I would be happy to branch off of your branch and try to get your code merged (this way you still get credit)

maxrudolph:
I'm sorry to be very slow to respond. Yes, that sounds great to me. Please
let me know if I can do anything to help. My academic term is just ending
and I have no time this week but next week things get better.

bobmyhill:
@danieldouglas92 can this be closed now?

gassmoeller:
This has been merged in #5259 and the documentation is available here: https://aspect-documentation.readthedocs.io/en/latest/user/cookbooks/cookbooks/vankeken_subduction/doc/vankeken_subduction.html

Lets close this PR, if we need a part of the discussion it will still be available.

PR #1049, opened 2016-06-29
Specifying multiple particle output formats

hlokavarapu:
Using the container vector of output::interfaces, can now generate the output of multiple formats that are specified by the parameter "Data output format."

As per edge cases, if user specifies:
  set Data output format = ascii, hdf5, none
then, none is honored.

If user specifies:
  set Data output format = ascii, ascii, vtu, vtu
then, ascii and vtu data are generated only once.

PR #1869, opened 2017-07-26
Rewrite particle output to use deal.II functionality

gassmoeller:
This PR uses the recently introduced funtionality of deal.II (dealii/dealii#308) to output zero-dimensional output features (like particles). It essentially offers the same functionality that `Visualization` already implements, but for the particles. With this PR the `particle/output/` plugins will no longer be necessary once we require deal.II 9.0. @hlokavarapu: This will conflict with #1049. Are you still interested in that functionality? Then we should think about integrating that behavior into this PR.

Things still to do:
- [x] make sure existing tests are not broken with deal.II 8.5
- [x] write a changelog entry


include/aspect/postprocess/particles.h
```diff
@@ -27,10 +27,79 @@
...
+    {
+      /**
+       * This class is responsible for writing the particle data into a format that can
+       * be written by deal.II, in particular a list of 'patches' that contain one
+       * particle per patch. The base class DataOutInterface is templated with a
+       * dimension of zero (the dimension of the particle / point), and a space dimension
+       * of dim (the dimension in which this zero-dimensional particle lives).
+       */
+      template<int dim>
+      class ParticleOutput : public dealii::DataOutInterface<0,dim>
```
bangerth: Don't you have to guard this class with `#if DEAL_II_VERSION_GTE(9,0)` as well?

source/postprocess/particles.cc
```diff
@@ -31,21 +32,119 @@ namespace aspect
 {
   namespace Postprocess
   {
+    namespace internal
+    {
+      template<int dim>
+      ParticleOutput<dim>::~ParticleOutput()
+      {}
```
bangerth: do you need this destructor?

source/postprocess/particles.cc
```diff
@@ -31,21 +32,119 @@ namespace aspect
...
+          {
+            const unsigned n_components = property_information.get_components_by_field_index(field_index);
+            const std::string field_name = property_information.get_field_name_by_index(field_index);
+            // If it is a 1D element, or a vector, print just the name, otherwise use []
+            if ((n_components == 1) || (n_components == dim))
+              for (unsigned int component_index=0; component_index<n_components; ++component_index)
+                dataset_names.push_back(field_name);
+            else
+              for (unsigned int component_index=0; component_index<n_components; ++component_index)
+                dataset_names.push_back(field_name + "[" + Utilities::to_string(component_index) + "]");
```
bangerth: I don't think `[` and `]` are valid names for data set names. You may have to just append the index with an underscore.

source/postprocess/particles.cc
```diff
@@ -31,21 +32,119 @@ namespace aspect
...
+            patches[i].data.reinit(property_information.n_components()+1,1);
+
+            patches[i].data(0,0) = particle->second.get_id();
+
+            const ArrayView<const double> properties = particle->second.get_properties();
+
+            for (unsigned int property_index = 0; property_index < properties.size(); ++property_index)
+              patches[i].data(property_index+1,0) = properties[property_index];
+
+          }
```
bangerth: no need for the empty line

gassmoeller:
Yes, I did not check the old functionality. This is fixed now (should work with 8.5.0 and 9.0.pre). I also removed the empty destructor and updated the test results.

hlokavarapu:
Woah, I have been waiting for this day; compressed particle visualization data, oh man pretty excited :). I don;t want to slow things down but I am still interested in supporting writing out multiple formats for particles. I will export the relevant code of PR #1049 to this code branch. This will take me a day or two.

gassmoeller:
Sounds good. From a first glance I would think it should fit neatly into the new structure, if you make output_format a list/vector of unique formats, and then loop over the if/else if branches in particles.cc:357-492. I will check in the meantime what happens with the test failures.

gassmoeller:
This should be good to go. Supersedes #1049.

source/postprocess/particles.cc
```diff
@@ -31,21 +32,119 @@ namespace aspect
...
+        typename std::multimap<aspect::Particle::types::LevelInd, aspect::Particle::Particle<dim> >::const_iterator particle = particles.begin();
+
+        for (unsigned int i=0; particle != particles.end(); ++particle, ++i)
+          {
+            patches[i].vertices[0] = particle->second.get_location();
+            patches[i].patch_index = i;
+            patches[i].n_subdivisions = 1;
+            patches[i].data.reinit(property_information.n_components()+1,1);
+
+            patches[i].data(0,0) = particle->second.get_id();
```
bangerth: We must be building the worst-suited data structure to represent particle data here :-) But that's what it is, nothing that can be done about it ;-)

bangerth: I think it might be worthwhile specializing `Patch<0,spacedim>` to avoid some of the member variables. But that can wait.

bangerth: See https://github.com/dealii/dealii/issues/4695

gassmoeller:
I addressed the comments. Should be ready to go.

Issue #1870, opened 2017-07-26
Why were the two same parameters set different values in pages 368-369 of the manual?


Lydia823:
In page 368, the parameter 'Density below' was set to 9900 km/m^3. However, in next page, the parameter 'Density below' was set to 8000. Why is there a difference?

jaustermann:
Hi Ronghua, 

The density below is set separately for the calculation of the bottom dynamic topography and the geoid. In a consistent model these should be the same though, good catch! I submitted a pull request to make those two default values the same, it might take some time until this update reaches the manual (not sure how often that is recompiled). 
While there will be a default value (which I set to 9900 km/m^3 for both postprocessors now) you might want to think about what value is most appropriate for you model (this depends on what you assume for the mantle temperature, whether the model is compressible etc.). When you manually set the ‘density below’ value in the parameter file (i.e. overwrite the default value), make sure to set this to the same value in both the dynamic topography and the geoid parameter section.

Best,
Jacky

bangerth:
I think the online manual is updated every night. So it should show the new values soon after the patch is merged.

Lydia823:
OK, i haven't downloaded the latest manual,thanks @jaustermann @bangerth

bangerth:
Fixed by #1871.

PR #1871, opened 2017-07-26
Make default value for 'density below' consistent across postprocessors

jaustermann:
The default value for 'density below' was different in the dynamic topography postprocessor and the  geoid postprocessor but for consistency this should probably be the same value. I adapted the default density below in the geoid postprocessor to be the same as in the dynamic topography postprocessor.


bangerth:
Looks good to me. Let's see what the tester says. 

/run-tests

bangerth:
Fixes #1870.

jaustermann:
All the tests pass because the geoid tests (geoid and geoid_visualization) prescribe / overwrite the default 'density below'.

PR #1873, opened 2017-07-27
Add more instructions on special installation cases

gassmoeller:
I just guided someone through an installation of ASPECT on a nearly unsupported cluster, and thought I might take some notes on common (and uncommon) problems that might appear during the installation. 

Does the installation section of the manual get a bit long if we include every corner-case? But currently there is no much better place either (I do not like the doc/install directory much either).

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -2641,6 +2641,7 @@ non-default compilers or libraries (e.g. Intel's MKL instead of LAPACK) by chang
 in the configuration file \texttt{candi.cfg}, or by providing platform specific installation files.
 
 \subsubsection{System prerequisites}
+\label{subsubsec:system_prerequisites}
 
 \texttt{candi} will show system specific instructions on startup, but its prerequisites
 are relatively widely used and packaged
@@ -2675,6 +2676,7 @@ sudo apt-get install build-essential \
 \end{verbatim}
 
 \subsubsection{Using candi to compile dependencies}
+\label{subsubsec:candi_dependencies}
 
 In its default configuration \texttt{candi} downloads and
 compiles a \dealii{} configuration that is able to run \aspect, but it
@@ -2684,7 +2686,7 @@ installation). We require at least the packages \pfrst{}, \trilinos{}
 (or as an experimental alternative \petsc{}), and finally \dealii{}.
  
 At the time of this writing \texttt{candi} will install \pfrst{} 2.0,
-\trilinos{} 12.10.1,  \petsc{} 3.6.4, and \dealii{} 8.5.0. 
+\trilinos{} 12.10.1,  \petsc{} 3.7.6, and \dealii{} 8.5.0. 
 We strive to keep the development version of \aspect{} compatible with 
 the latest release of \dealii{} and the current \dealii{} development 
 version at any time, and we usually support several older versions of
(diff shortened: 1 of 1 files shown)


tjhei:
Have you considered adding an Installation FAQ to https://github.com/geodynamics/aspect/wiki for specific things instead?

doc/manual/manual.tex
```diff
@@ -2771,20 +2773,126 @@ \subsubsection{Compiling \aspect{} and generating documentation}
...
+
+\texttt{candi} relies on the fact that it can download packages from known
+internet locations. Some cluster systems might prevent internet access or
+disallow certain SSL certificates. In these cases please install \texttt{candi}
+and \dealii{} on a different system (e.g. a desktop system) and then copy the
+downloaded packages to your cluster. \texttt{candi} will recognize the
+existence of the packages and skip the download phase. The packages should be
+placed in the following folder: 
+
+\begin{verbatim}
```
bangerth: no empty line here (and elsewhere) -- an empty line starts a new paragraph (and indents the text), but here and below you just want the current sentence or thought to continue.

doc/manual/manual.tex
```diff
@@ -2771,20 +2773,126 @@ \subsubsection{Compiling \aspect{} and generating documentation}
...
+(\texttt{CFLAGS,CPPFLAGS,LDFLAGS}) in your terminal, or include the
+corresponding commands into the \texttt{candi.cfg} file in the \texttt{candi}
+directory to make sure they are set whenever \texttt{candi} is run.
+
+\paragraph{Installation using Intel's MKL as BLAS/LAPACK:}
+
+A common obstacle during installation is that many cluster systems utilize
+Intel's MKL library as an implementation of the BLAS/LAPACK routines.
+\texttt{candi} can automatically configure the packages that depend on this
+implementation, if it is told to do so. To this purpose modify the
```
bangerth: purpose -> end,

doc/manual/manual.tex
```diff
@@ -2771,20 +2773,126 @@ \subsubsection{Compiling \aspect{} and generating documentation}
...
+GCC-4.8.0, the first GCC version considered to have full C++11 support, was
+released in 2013, other compatible compilers are Clang 3.3, and Intel 15 or
+newer).
+
+We are aware that some clusters without regular support might not provide C++11
+support and have therefore kept \aspect{} compatible to older compilers in line
+with the implementations of deal.II (see also
+\url{https://github.com/dealii/dealii/wiki/Frequently-Asked-Questions#does-dealii-require-c11-support}).
+Note however, that \trilinos{} and \dealii{} already dropped support for
+non-C++11 compilers in their newest releases and development version
```
bangerth: mention specific releases, or this paragraph will become outdated pretty quickly.

doc/manual/manual.tex
```diff
@@ -2771,20 +2773,126 @@ \subsubsection{Compiling \aspect{} and generating documentation}
...
+respectively. Therefore, you will have to use specific versions of \trilinos{}
+and \dealii{} to compile \aspect{} with a non-C++11 conform compiler. We
+recommend using \trilinos{} 11.10.1, and \dealii 8.5.0 that both support older
```
bangerth: I think you need braces after `\dealii` or the space will be eaten

gassmoeller:
> Have you considered adding an Installation FAQ to https://github.com/geodynamics/aspect/wiki for specific things instead? It seems to me that this is easier to update.

Actually, I did, but then decided to not open another topic :smile:. But the more I think about it the more it seems like the better solution. This would also allow others to add their installation experiences without creating pull requests, and we could unify the information from different places there (i.e. move the Mac description, and the files in doc/install there as well). I will be travelling next week, but afterwards I can move this over to the wiki.

gassmoeller:
I moved the content of this PR into the wiki (https://github.com/geodynamics/aspect/wiki/Installation-FAQ). I will open a new PR to move the existing content of the manual and the doc/install directory to the wiki.

PR #1875, opened 2017-08-01
use current_constraints for SparsityPattern

tjhei:
- use current_constraints instead of constraints for SparsityPatterns
- to do that, delay construction of SP/matrices until first timestep
- also expose pressure_scaling in SimulatorAccess
- add a test

supersedes #1865
needed for #1864


tjhei:
waiting for the tester to change test output: I assume solver iterations and number of matrix entries will change.

tjhei:
We now no longer have this code inside the "Setup dof systems" computing timer. Any thoughts on this? Should I just add a new entry like "Setup matrices"?

doc/modules/changes/20170801_tjhei
```diff
@@ -0,0 +1,5 @@
+New: Sparsity patterns of the system matrix now uses the current_constraints
```
bangerth: uses -> use

tjhei:
> Does that mean that we now rebuild the sparsity pattern in every time step?

No. ``rebuild_sparsity_and_matrices`` only gets set to true once (in setup_dofs()).

> Before we commit this, may I ask you to try one thing? Namely, put constraints back into place where we build the sparsity pattern, instead of current_constraints.

I just created #1877 that should show exactly that.

tjhei:
>> Does that mean that we now rebuild the sparsity pattern in every time step?
> No. rebuild_sparsity_and_matrices only gets set to true once (in setup_dofs()).

But, thinking about it, we could allow changing that in the future if we ever need it. Say, you can signal "from now on the type of boundary conditions is different so please give me a new sparsity pattern".

bangerth:
OK, #1877 works -- thanks for trying that out! Wait for the tester, then fixup commit 4 into 1 and 3 into 2. Then merge yourself.

Also correct that we can make this more flexible from time step to time step.

PR #1877, opened 2017-08-04
[DO NOT MERGE] use current_constraints for SparsityPattern

tjhei:
this is the proof that the patch in #1875 is harmless. The tester should pass like this.


bangerth:
Yes, succeeds -- thank you!

Issue #1876, opened 2017-08-02
Allow prescribed velocities with locally conservative discretization


maxrudolph:
Currently, the `cookbooks/prescribed_velocity` plugin does not correctly enforce internal constraints when `set Use locally conservative discretization = true` is set in the parameter file.

Screen output:
```./aspect corner_flow.prm 
-----------------------------------------------------------------------------
-- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.
--     . version 2.0.0-pre
--     . running in DEBUG mode
--     . running with 1 MPI process
--     . using Trilinos
-----------------------------------------------------------------------------

Loading shared library <./libprescribed_velocity.so>

Number of active cells: 16,384 (on 8 levels)
Number of degrees of freedom: 247,299 (132,098+49,152+66,049)


--------------------------------------------------------
An error occurred in line <2269> of file </opt/aspect-stack/tmp/unpack/deal.II-v8.5.0/source/fe/fe_values.cc> in function
    dealii::FEValuesBase<dim, spacedim>::FEValuesBase(unsigned int, unsigned int, dealii::UpdateFlags, const dealii::Mapping<dim, spacedim>&, const dealii::FiniteElement<dim, spacedim>&) [with int dim = 2; int spacedim = 2]
The violated condition was: 
    n_q_points > 0
Additional information: 
    There is nothing useful you can do with an FEValues object when using a quadrature formula with zero quadrature points!

Stacktrace:
-----------
#0  /opt/aspect-stack/deal.II-v8.5.0/lib/libdeal_II.g.so.8.5.0: dealii::FEValuesBase<2, 2>::FEValuesBase(unsigned int, unsigned int, dealii::UpdateFlags, dealii::Mapping<2, 2> const&, dealii::FiniteElement<2, 2> const&)
#1  /opt/aspect-stack/deal.II-v8.5.0/lib/libdeal_II.g.so.8.5.0: dealii::FEValues<2, 2>::FEValues(dealii::FiniteElement<2, 2> const&, dealii::Quadrature<2> const&, dealii::UpdateFlags)
#2  ./libprescribed_velocity.so: void aspect::constrain_internal_velocities<2>(aspect::SimulatorAccess<2> const&, dealii::ConstraintMatrix&)
[...]

PR #1878, opened 2017-08-05
Allow prescribed velocities with locally conservative discretization.

maxrudolph:
This implements the fix to the prescribed internal velocities plugin as suggested by @bangerth , in response to #1876 .


cookbooks/prescribed_velocity/corner_flow.prm
```diff
@@ -5,6 +5,11 @@ set End time    = 0
 # Load the signal library.
 set Additional shared libraries = ./libprescribed_velocity.so
 
+subsection Discretization
```
bangerth: Instead of modifying an existing test, can you just clone it? This way, we test both the existing code path as well as the new one of this patch.

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -110,6 +110,51 @@ namespace aspect
   }
 
   /**
+   * This function retrieves the unit support points (in the unit cell) for the current element
```
bangerth: period at the end of the sentence.

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -110,6 +110,51 @@ namespace aspect
...
+   * The DGP elements used when 'set Use locally conservative discretization = true' is enabled do
```
bangerth: "is enabled do" should probably be "is enabled to", but that still sounds funny. How about "...does not have support points"

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -110,6 +110,51 @@ namespace aspect
...
+   * not have support points. If these elements are in use, a ficticious support point at the cell
```
bangerth: -> fictitious

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -110,6 +110,51 @@ namespace aspect
...
+   * center is returned. This is necessary for the internal velocities to be constrained as desired.
```
bangerth: -> is returned for each shape function that corresponds to the pressure variable, whereas the support points for the velocity are correct; the fictitious points don't matter because we only use this function when interpolating the velocity variable, and ignore the evaluation at the pressure support points.

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -110,6 +110,51 @@ namespace aspect
...
+   */
+  template <int dim>
+  std::vector< Point<dim> >
+  get_unit_support_points_for_velocity(const SimulatorAccess<dim> &simulator_access,
+                                       const bool use_locally_conservative_discretization)
+  {
+    std::vector< Point<dim> > unit_support_points;
+    if ( !use_locally_conservative_discretization )
+      {
+        unit_support_points = simulator_access.get_fe().get_unit_support_points();
```
bangerth: or, more easily,
```
  return simulator_access.get_fe().get_unit_support_points();
```
Then you can move the declaration of the `unit_support_points` variable into the `else` block, along with the `return` statement at the bottom of the function.

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -110,6 +110,51 @@ namespace aspect
...
+      }
+    else
+      {
+        //special case for discontinuous pressure elements, which lack unit support points
+        const unsigned int dofs_per_cell = simulator_access.get_fe().dofs_per_cell;
+        for (unsigned int dof=0; dof < dofs_per_cell; dof++)
```
bangerth: for consistency with the rest of the code base, use `++dof`.

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -110,6 +110,51 @@ namespace aspect
...
+            base_index = simulator_access.get_fe().system_to_base_index(dof).second;
+            // get the unit support points for the relevant element
+            std::vector< Point<dim> > my_support_points = simulator_access.get_fe().base_element(base).get_unit_support_points();
+            if ( my_support_points.size() == 0 )
+              {
+                //manufacture a support point, arbitrarily at cell center
+                if ( dim==2 )
+                  unit_support_points.push_back( Point<dim> (0.5,0.5) );
+                if ( dim==3 )
+                  unit_support_points.push_back( Point<dim> (0.5,0.5,0.5) );
```
bangerth: on second thought, rather than making the distinction between dimensions, just do
```
  // manufacture a support point, arbitrarily at cell's origin
  unit_support_points.push_back (Point<dim>());
```
This works independent of the dimension and is shorter.

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -120,7 +165,11 @@ namespace aspect
   {
     if (prescribe_internal_velocities)
       {
-        Quadrature<dim> quadrature (simulator_access.get_fe().get_unit_support_points());
+        const Parameters<dim> parameters = simulator_access.get_parameters();
+        const bool use_locally_conservative_discretization = parameters.use_locally_conservative_discretization;
+
+        std::vector< Point<dim> > points = get_unit_support_points_for_velocity<dim>(simulator_access,use_locally_conservative_discretization);
```
bangerth: * no need to explicitly say `<dim>` in the function call.

* mark the variable as `const`

* also, can you break at the comma between function arguments? The line is long enough.

cookbooks/prescribed_velocity/prescribed_velocity.cc
```diff
@@ -120,7 +165,11 @@ namespace aspect
...
+        Quadrature<dim> quadrature (points);
```
bangerth: while you're there, make this `const` as well.

cookbooks/prescribed_velocity/prescribed_velocity.cc
maxrudolph: Will the default constructor for point place the point at 0.5,0.5 or at some other value like null? There is actually a reason why I wrote the code this way. I want to place the manufactured point at the cell center so that in my `prescribed_velocity_pressure.cc` plugin ( #1864 ) which reuses this code, the pressure indicator function is evaluated at a meaningful location.

bangerth: `Point<dim>()` creates (0,0) or (0,0,0). If you like the center of the cell better, then leave it as is.

cookbooks/prescribed_velocity/corner_flow.prm
maxrudolph: This is just the cookbook input file, not a test. I am happy to add a test for the DGP case if that would be useful.

maxrudolph:
@bangerth This commit addresses your comments above and adds a test for the DGP case. I also updated the test for the default element to use the new prescribed velocity plugin code.

cookbooks/prescribed_velocity/corner_flow.prm
bangerth: Ah, I see. But why change the cookbook?

As for test: Yes, that would be very useful indeed -- we usually try to exercise all code paths in the testsuite somehow.

maxrudolph:
@bangerth I think that this is ready for a final review and merge.

gassmoeller:
Makes total sense. Ready to merge when the tester is happy. Sorry for coming back so late to this PR.

PR #1825, opened 2017-06-23
Operator splitting

jdannberg:
This pull request adds the option to use operator splitting as a solver scheme to solve advection and reactions between compositional fields (and temperature) separately. For now, there is only one solver scheme that uses this, and I've only implemented a simple first-order operator split that uses a forward Euler as a solver. These things could be improved in the future. 
However, I've also added a benchmark for this to the manual, and for the cases I tested, the error caused by the operator split was usually small, and the time used to compute the reactions was also small (compared to the advection). 

Everything seems to work, but I still have to work on the documentation, so I've marked it as work in progress.

PR #1881, opened 2017-08-10
test commit

gassmoeller:
This is a test PR to see if a test on master is broken.


gassmoeller:
It seems we have a single broken test on master. @jdannberg and I looked at the test and it was likely an oversight in the recently merged #1825. I am simply going to merge this as a correction to the test.

PR #1882, opened 2017-08-11
A new boundary temperature plugin

SiqiZhang:
The pull request add a new boundary temperature plugin that the core mantle boundary temperature evolves through time by calculating the heat flux through core mantle boundary and solving the core energy balance. And also added a postprocess of the core evolution. It is used in our paper Zhang et al. [2016] and it could be useful for people interested in long term planetary evolution.


SiqiZhang:
@gassmoeller : Thanks for those comments. I will make those changes. 
By the way, I have to rely on the postprocess to store the restart data, which makes things a bit complicated. Is it possible to store it in boundary_temperature object?

gassmoeller:
Currently there is no way to store data for serialization in all plugins that are not postprocessors, so you actually found a nice way to work around that.
 This might be something we need to change, but originally every plugin was thought to have no state that needs to be stored. Obviously, this is now different, but a change would be a larger operation, not necessarily incompatible, but a bit more work.
 Lets finish this PR first and then think about how to do that. If you would like to help implementing that please open a separate issue and we can discuss what would need to be implemented.

source/boundary_temperature/dynamic_core.cc
```diff
@@ -0,0 +1,991 @@
...
+      const Postprocess::CoreStatistics<dim> * core_statistics
+        = this->template find_postprocessor<const Postprocess::CoreStatistics<dim> >();
+      AssertThrow(core_statistics!=NULL,
+          ExcMessage ("Dynamic core boundary condition has to work with dynamic core statistics postprocessor."));
+      core_data.dt    = this->get_timestep();
+      core_data.H     = get_radioheating_rate();
+        
+      // The restart data is stored in 'core statistics' postprocessor.
+      // If restart from checkpoint, extract data from there.
+      if(!core_data.is_initialized)
```
SiqiZhang: As you suggested, I tried to use the initialize() function. However, it seems there is a problem. In the initialize() function of the DynamicCore plugin, it needs to read data from the postprocess (stored restart data). But the postprocess seem initialized after boundary temperature, so it is not available at the time initialize() function of boundary temperature plugin is called. Maybe I have to keep the current approach for now. Any further suggestion is welcome.

bangerth: Yes, that is indeed the case: the `postprocess_manager` is initialized after the boundary temperature manager (lines 557 and line 486 in `core.cc`, respectively). I don't think that there is a general solution to this problem -- one can always concoct ways in which things depend on each other. Your approach, as far as I understand it, does not depend on the order in which things are initialized, and this is probably the right way to do it.

bangerth:
I'd like to echo @gassmoeller 's point: it's great to see you back in action, @SiqiZhang ! We were a bit distracted for the past two weeks with a deal.II hackathon here in Colorado, but let's see that we can figure this PR out and get it merged. Your approach of using a postprocessor is of course a bit backward but it nicely works around a problem you're not responsible for :-)

include/aspect/boundary_temperature/dynamic_core.h
```diff
@@ -0,0 +1,521 @@
...
+        /**
+         * This function update the core-mantle boundary (CMB) temperature by
+         * the core energy balance solver using the core-mantle boundary heat flux.
+         */
+        virtual void update();
+
+        /**
+         * Pass core data to other modules
+         */
+        struct internal::CoreData get_core_data() const;
```
bangerth: C++ doesn't actually require you to write `struct` in front of the name.

But more importantly: By returning an object, the compiler will create a copy. Would it be enough to just return a `const` reference? Presumably, the calling sites have no interest in modifying the returned object (and if they do, that's likely a bug!).

include/aspect/boundary_temperature/dynamic_core.h
```diff
@@ -0,0 +1,521 @@
...
+        bool is_OES_used() const;
+
+
+      private:
+
+        /**
+         * Data for core energy balance
+         * it get updated each time step.
+         */
+        struct internal::CoreData core_data;
```
bangerth: not wrong, but also not necessary to write `struct`

include/aspect/boundary_temperature/dynamic_core.h
```diff
@@ -0,0 +1,521 @@
...
+
+        /**
+         * Temperatures at the inner boundaries.
```
bangerth: -> Temperature
(same below)

include/aspect/postprocess/core_statistics.h
```diff
@@ -0,0 +1,107 @@
...
+        std::pair<std::string,std::string>
+        execute (TableHandler &statistics);
+
+        /**
+         * Export core data stored in this object. Doing this only because the boundary
+         * temperature doesn't allowed to store restart data there. So we store the data needed
+         * for restart here and exported to boundary temperature object if required.
+         */
+        struct  BoundaryTemperature::internal::CoreData
+      get_core_data() const;
```
bangerth: the same applies as above -- I think you probably want to return a `const` reference

source/boundary_temperature/dynamic_core.cc
```diff
@@ -0,0 +1,993 @@
...
+        }
+      if (data_OES.size()!=0)
+        this->get_pcout()<<"Other energy source is in use ( "<<data_OES.size()<<" data points is read)."<<std::endl;
+    }
+
+    template <int dim>
+    double
+    DynamicCore<dim>::get_OES(double t) const
+    {
+      t/=1.e9*year_in_seconds;
```
bangerth: I think you ought to guard this with the flag that determines whether we compute in years or in seconds. Take a look at the other postprocessors that call `this->convert_output_to_years()`.

source/boundary_temperature/dynamic_core.cc
```diff
@@ -0,0 +1,993 @@
...
+
+      // It's a bit tricky here.
+      // Didn't use the initialize() function instead because the postprocess is initialized after boundary temperature.
+      // It is not available at the time initialize() function of boundary temperature is called.
+      if (is_first_call==true)
+        {
+          const Postprocess::CoreStatistics<dim> *core_statistics
+            = this->template find_postprocessor<const Postprocess::CoreStatistics<dim> >();
+          AssertThrow(core_statistics!=NULL,
+                      ExcMessage ("Dynamic core boundary condition has to work with dynamic core statistics postprocessor."));
```
bangerth: nice job figuring out how to get at the postprocessors! nice job also adding the assertion to make sure we get an error if things aren't set up to work!

tests/dynamic-core.prm
```diff
@@ -0,0 +1,104 @@
+# A simple setup for convection in a quarter of a 2d shell. See the
+# manual for more information.
```
bangerth: can you update the comment here to say that that the specific test is for the dynamic core temperature plugin?

For the test to actually run, you will also have to add at least one of the output files. Here, I think the statistics and screen output file should be sufficient. Do you know how to do that, or do you need help?

include/aspect/boundary_temperature/dynamic_core.h
SiqiZhang: @bangerth Thanks for your comments. The core data is passed from boundary temperature to postprocess once each time step, and it passed back from postprocess to boundary temperature only at the restart. And if in future, we move the heat flux calculation into postprocess as @gassmoeller proposed, we will need pass data from postprocess to boundary temperature every time step as well.
So it's not a single way data passing. I know you guys don't like keep multiple copy of the same data, it's the only solution I can figure out at the moment. I don't think const reference will work for this case. Even we pass a reference, we still need to manually copy the data over.

source/boundary_temperature/dynamic_core.cc
SiqiZhang: @bangerth : The unit here is billion years. The core evolves quite slow, that's why const CMB temperature is normally used. If someone want to include core evolution in his model, the model time length should be at least a few hundreds to a few billion years. So I don't belive anyone will have a data in the unit of seconds for this. I would prefer it stays the current way.

include/aspect/boundary_temperature/dynamic_core.h
bangerth: Right. I didn't mean that one class or the other stores a reference, just that the functions that return these objects return them by const reference. In other words, you can still do
```
  this->core_data = other_object.get_core_data();
```
and the data will be *copied* from `other_object` to `this`. It's just that the way you write the code right now, it will be copied *twice*: once from `other_object.core_data` to the object that `get_core_data()` returns, and then a second time from that object to `this->core_data`.

source/boundary_temperature/dynamic_core.cc
bangerth: Ah, I see. So if I understand you right, then you do this because all computations and constants for the core are usually done in units of billions of years because dynamics are just so slow? That makes sense. Would you mind adding a one-line comment to this effect?

include/aspect/boundary_temperature/dynamic_core.h
SiqiZhang: OK then, I will change them to use const reference.

gassmoeller:
Hi Siqi,
quick question: Did you address most of our comments, or are there things left you want to do?

SiqiZhang:
@gassmoeller 
All the previous comments has been addressed. I don't have anything left to do at the moment. This pull-request is ready to be merged if you guys happy with it.

gassmoeller:
Awesome, then this is ready to go except one small thing: Please update your test results with the diff that the reference tester posts under 'Details'. I had a look at the difference in output, and it looks like the usual minor differences caused by different library versions, operating systems and so on. After that this is ready to merge.

For the future: Please leave a little comment, if you want us to look at something again. Otherwise we do not realize you finished the things you wanted to work on.
Thanks for putting this back into the main version!

SiqiZhang:
@gassmoeller 
The failed test has already been fixed by you with a later pull request. I rebased mime to current master. It should fix this problem. While running tests on my local machine, I have lots of failed tests (The intel compiler I am using might causing this problem). I hope the new test I added for my plugin works OK with the tester machine and others.
Thanks for following up on this pull request.

gassmoeller:
Ok great then this is ready. Do not worry about the tests on your machine too much. Likely the difference is very minor. Thanks for contributing :+1:.

PR #1883, opened 2017-08-11
Correct pressure term factor

anne-glerum:
As the yield criterion is specified in terms of pressure (average normal stress) instead of the first invariant of stress (sum normal stresses), we need to incorporate a factor 3.


gassmoeller:
Ouch. I guess you checked that the new behavior is correct? It seems we are in desperate need of a good plasticity benchmark. 
Anyway, we should document this fix in the changelog.
This should break some tests, but we also had an unrelated failing test on master for the last few days. Please rebase and use the tester results after rebasing for updating the test.

jdannberg:
So what about the 2D case? Is this not an issue there?

gassmoeller:
Now this is curious. This PR only fails the one broken test on master that is unrelated to this change. Do we have no tests for the plasticity option of these material models?

naliboff:
Good catch Anne and ouch indeed. It is quite common in the geodynamics literature to write the yield criterion in terms of the pressure and not the first invariant. I wonder how many different forms of the 3-D equations (actual implementations in the code) are out there?

@jdannberg - I'm also not clear know as to why this does not affect the 2-D equations, unless the 2-D form does in fact use the pressure instead of the first stress invariant. Should we re-derive these equations from a common source and make sure we are all in agreement?

@gassmoeller - There are tests that use plasticity for the `drucker_prager` and` visco_plastic` material models, but they are all 2-D. Unfortunately, the brick shear band experiment is the de facto plasticity benchmark in the geodynamics community, but perhaps there are better tests outlined in the engineering literature that can be applied here.

gassmoeller:
How should we proceed with this PR? Can we copy one of the existing plasticity tests and make it 3D? Only so that we have a test that tracks changes in the future? And have you verified the new behavior is the correct one (as in: as correct as our current approach to plasticity can be)?

tjhei:
> ow should we proceed with this PR? Can we copy one of the existing plasticity tests and make it 3D? Only so that we have a test that tracks changes in the future?

Yes, please. Even if we are not sure it is correct, I hate having untested code.

anne-glerum:
Hi, yes, correct as can be. I'll update and add a test.

gassmoeller:
Can we finish this within the next week or so? I would like to make sure to at least fix the bugs we know about before moving forward to the release of 2.0. @anne-glerum: If you do not have the time, let me know and I can finish the rest.

anne-glerum:
@gassmoeller I've added the test, pr should be done now.

tests/crustal_model_3D.prm
```diff
@@ -0,0 +1,131 @@
...
+
+set Dimension                              = 3
+set Start time                             = 0
+set End time                               = 0
+set Use years in output instead of seconds = true
+set CFL number                             = .5
+set Linear solver tolerance                = 1e-9
+set Nonlinear solver scheme                = iterated Stokes
+set Nonlinear solver tolerance             = 1e-5
+set Max nonlinear iterations               = 3
```
naliboff: The only issue I can envision is if a later change allows the model to converge in less than 3 iterations. Is there a reason to do more than one non-linear iteration? If so, perhaps set the Nonlinear solver tolerance higher so that it would never reach 1e-5? However, this is very likely a non-issue and an easy fix later on.

tests/crustal_model_3D.prm
```diff
@@ -0,0 +1,131 @@
...
+# The velocity along the top boundary models a spreading
+# center that is moving left and right:
+subsection Boundary velocity model
+  subsection Function
+    set Variable names      = x,y,z
+    set Function constants  = cm=0.01, year=1 
+    set Function expression =  if (x<56e3 && y<=48e3 | x<72e3 && y>48e3, -1*cm/year, 1*cm/year) ; 0 ; 0
+  end
+end
+
```
naliboff: One more comment - the velocity is applied to the to left and right walls in this case. In your comment above are you referring to the velocity that arises along the top boundary due to the applied velocities along the left and right walls?

naliboff:
@anne-glerum - Looks good, thanks for the changes!

PR #1700, opened 2017-05-16
Removed interface compatibility.

alarshi:
In benchmarks/solkz

PR #1539, opened 2017-05-09
Refactor SolKz benchmark, add compositional field benchmark, and add active particle benchmark

hlokavarapu:
**The log output of SolKz benchmark at a global refinement level of 5:**
```
harsha_lv@godavari:~/ClionProjects/aspect/benchmarks/solkz$ mpirun -np 1 aspect solkz.prm 
-----------------------------------------------------------------------------
-- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.
--     . version 2.0.0-pre
--     . running in DEBUG mode
--     . running with 1 MPI process
--     . using Trilinos
-----------------------------------------------------------------------------

Loading shared library <./libsolkz.so>

Number of active cells: 1,024 (on 6 levels)
Number of degrees of freedom: 13,764 (8,450+1,089+4,225)

*** Timestep 0:  t=0 years
   Rebuilding Stokes preconditioner...
   Solving Stokes system... 47+0 iterations.
      Relative Stokes residual after nonlinear iteration 1: 1
   Solving Stokes system... 0+0 iterations.
      Relative Stokes residual after nonlinear iteration 2: 9.61484e-08

   Postprocessing:
     Errors u_L1, p_L1, u_L2, p_L2: 1.255365e-08, 1.599227e-04, 3.451232e-08, 2.647014e-04
     Writing graphical output:      output/solution/solution-00000

Termination requested by criterion: end time
```
**The log output of SolKz benchmark using compositional fields at a global refinement level of 5:**
```
harsha_lv@godavari:~/ClionProjects/aspect/benchmarks/solkz/compositional_fields$ mpirun -np 1 aspect solkz_compositional_fields.prm 
-----------------------------------------------------------------------------
-- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.
--     . version 2.0.0-pre
--     . running in DEBUG mode
--     . running with 1 MPI process
--     . using Trilinos
-----------------------------------------------------------------------------

Loading shared library <./libsolkz_compositional_fields.so>

Number of active cells: 1,024 (on 6 levels)
Number of degrees of freedom: 34,179 (8,450+3,072+4,225+9,216+9,216)

*** Timestep 0:  t=0 years
   Rebuilding Stokes preconditioner...
   Solving Stokes system... 48+0 iterations.
      Relative Stokes residual after nonlinear iteration 1: 1
   Solving Stokes system... 0+0 iterations.
      Relative Stokes residual after nonlinear iteration 2: 7.95719e-08

[...]

PR #1884, opened 2017-08-12
Fixed bug preventing particles from working in Solcx benchmark

HarryRLevinson:
The density and viscosity values at each point were not being set to the values on their respective compositional fields in solcx_compositional_fields.h.  This is because SolCxCompostionalMaterial was outdated and needed to be updated.  As a result, we would always obtain the convergence results for the direct method regardless of the method being used in the Solcx parameter file.  

Currently, if a compositional field carries viscosity, then the convergence results are less accurate then that of the direct methods.  However, the convergence results for Solcx are exactly the same as the direct method if only density is carried on a compositional field.

### DIFF
--- a/benchmarks/solcx/compositional_fields/solcx_compositional_fields.h
+++ b/benchmarks/solcx/compositional_fields/solcx_compositional_fields.h
@@ -15,21 +15,19 @@ namespace aspect
     class SolCxCompositionalMaterial : public SolCxMaterial<dim>
     {
       public:
-        double density(const double temperature,
-                       const double pressure,
-                       const std::vector<double> &composition,
-                       const Point <dim> &position) const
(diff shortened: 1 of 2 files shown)


gassmoeller:
Thanks for fixing this! For reference: This was likely caused by an overlap of #1539 and #1700, and never discovered afterwards. Lets see if it affects the tester.

bangerth:
There is one failing test. Does anyone have an idea what the reason is?

gassmoeller:
Yes, this test was broken on master (fixed in #1881 earlier today). Since this PR is completely unrelated to the broken test I am in favour of just merging it. A rebase would simply be unnecessary work here.

bangerth:
I agree.

Issue #1879, opened 2017-08-08
Move special installation instructions to wiki


gassmoeller:
In #1873 we decided that it would be easiest to collect special installation instruction in a central place, e.g. the wiki (https://github.com/geodynamics/aspect/wiki/Installation-FAQ). Move all the remaining content of doc/install and the manual to that place.

PR #1891, opened 2017-08-16
Move special installation instructions

gassmoeller:
One of two steps towards #1879. These paragraphs have been moved to the wiki. This way it is easier for anybody to update them, and it is easier to link to them (and the manual does not get filled up with special case instructions).
In a follow-up PR I will move over the content of the `doc/install` directory (although that it pretty outdated).

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,14 +16,23 @@ The project is supported by CIG (http://geodynamics.org).
 Installation instructions
 -------------------------
 
-The steps to install the necessary dependencies and ASPECT itself are described in the Installation instructions section of the ASPECT [manual](http://www.math.clemson.edu/~heister/manual.pdf).
+The steps to install the necessary dependencies and ASPECT itself are described
+in the Installation instructions section of the ASPECT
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -16,14 +16,23 @@ The project is supported by CIG (http://geodynamics.org).
 Installation instructions
 -------------------------
 
-The steps to install the necessary dependencies and ASPECT itself are described in the Installation instructions section of the ASPECT [manual](http://www.math.clemson.edu/~heister/manual.pdf).
+The steps to install the necessary dependencies and ASPECT itself are described
+in the Installation instructions section of the ASPECT
+[manual](http://www.math.clemson.edu/~heister/manual.pdf). If you encounter
```
bangerth: hm, but didn't you just remove it from there?

gassmoeller:
I removed the special cases "static executable" and "MacOS with deal.II package". The default candi install instructions are still in the manual.

bangerth:
Ah, yes, of course. Thanks for pointing out the obvious.

PR #1898, opened 2017-08-17
fix and cleanup compute_initial_pressure_field()

tjhei:
- cleanup melt variable handling
- only interpolate the pressure component


gassmoeller:
Good to go.

PR #1899, opened 2017-08-17
add a postprocessor for the volumetric strain rate

jdannberg:
This adds a visualization postprocessor for the volumetric strain rate, which is in particular useful in compressible models or models with melt transport.


source/postprocess/visualization/volumetric_strain_rate.cc
```diff
@@ -0,0 +1,88 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#include "aspect/postprocess/visualization/volumetric_strain_rate.h"
```
tjhei: make this a system include (using ``<...>``)

source/postprocess/visualization/volumetric_strain_rate.cc
```diff
@@ -0,0 +1,88 @@
...
+  {
+    namespace VisualizationPostprocessors
+    {
+      ASPECT_REGISTER_VISUALIZATION_POSTPROCESSOR(VolumetricStrainRate,
+                                                  "volumetric strain rate",
+                                                  "A visualization output object that generates output "
+                                                  "for the volumetric strain rate, i.e., for the quantity "
+                                                  "$\\nabla\\mathbf u$. "
+                                                  "This should be zero (up to the solver tolerance) in incompressible "
+                                                  "convection models, but can be nonnzero in compressible models and "
```
tjhei: typo: nonnzero

jdannberg:
Fixed.

source/postprocess/visualization/volumetric_strain_rate.cc
```diff
@@ -0,0 +1,88 @@
...
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      ASPECT_REGISTER_VISUALIZATION_POSTPROCESSOR(VolumetricStrainRate,
+                                                  "volumetric strain rate",
+                                                  "A visualization output object that generates output "
+                                                  "for the volumetric strain rate, i.e., for the quantity "
+                                                  "$\\nabla\\mathbf u$. "
```
bangerth: are you missing a cdot? maybe write this as `\\nabla\\cdot\\mathbf u = \\textrm{div}\; \\mathbf u = \textrm{trace}\; \\varepsilon(\\mathbf u)`

include/aspect/postprocess/visualization/volumetric_strain_rate.h
```diff
@@ -0,0 +1,62 @@
...
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      /**
+       * A class derived from DataPostprocessor that takes an output vector
+       * and computes a variable that represents the volumetric strain rate at every
+       * point. The volumetric strain rate is defined as $\nabla \mathbf u$
+       * (or the trace of the strain rate tensor).
```
bangerth: same comment as I leave below

source/postprocess/visualization/volumetric_strain_rate.cc
```diff
@@ -0,0 +1,88 @@
...
+                                                  "This should be zero (up to the solver tolerance) in incompressible "
+                                                  "convection models, but can be non-zero in compressible models and "
+                                                  "models with melt transport.")
```
bangerth: that's not actually true. it is zero for the *exact* solution, but for the numerical solution it is not in general. All we can say is that the divergence is zero in some average sense (namely when multiplied by a test function and integrated over) but it is not zero at every point.

tests/compressibility_iterated_stokes.prm
```diff
@@ -108,11 +108,11 @@ subsection Boundary velocity model
 end
 
 subsection Postprocess
-  set List of postprocessors = compressibility
+  set List of postprocessors = compressibility, visualization
 
   subsection Visualization
 
-    set List of output variables      = density
+    set List of output variables      = density, volumetric strain rate
```
bangerth: just copy/clone the test. we like having tests that test one particular aspect, rather than many things at the same time

jdannberg:
I think I addressed all of your comments. For the work where we compare the different formulations, we need rho * div u, which is a slightly different postprocessor (as that requires evaluating the material model), which probably also does not have that many uses, so we only implemented that as a plugin (not as one of the regular postprocessors).

source/postprocess/visualization/volumetric_strain_rate.cc
```diff
@@ -0,0 +1,89 @@
...
+
+        for (unsigned int q=0; q<n_quadrature_points; ++q)
+          {
+            // extract the velocity gradients
+            double div_u = 0.0;
+            Tensor<2,dim> grad_u;
+            for (unsigned int d=0; d<dim; ++d)
+              {
+                grad_u[d] = input_data.solution_gradients[q][d];
+                div_u += grad_u[d][d];
```
bangerth: I missed this before -- you're assigning `grad_u[d]` once, and then you use it once. It is, in principle, not necessary to even put this into a variable. I think I'd just get rid of the `grad_u` variable altogether.

jdannberg: I can do that, I just thought if I write it like this, the code is easier to understand.

bangerth: Sure, but in that case I'd first fill `grad_u`, and then compute the result by saying `trace(grad_u)`.

jdannberg: I like that, I didn't think of writing it that way. I changed it.

Issue #1896, opened 2017-08-17
Assembly error when using compositional fields with the latent heat material model.


naliboff:
An ASPECT user notified me of an issue with the `latent heat` material model. When one includes compositional fields the model exits with an error (`Arithmetic exception`) during the assembly of the compositional field.

The exact error is at `advection.cc lines 160:168`:
```
              data.local_rhs(i)
              += (field_term_for_rhs * scratch.phi_field[i]
                  + time_step *
                  scratch.phi_field[i]
                  * gamma
                  + scratch.phi_field[i]
                  * reaction_term)
                 *
                 JxW; 
```

Before diving into this and a rewriting a bit of the documentation in the `latent heat` material model, I wanted to check if anyone has used this model with compositional fields at some point in the recent past. Perhaps all of the models we tried so far simply set things up in an un-physical manner? Unless I missed one, none of the `latent heat` test cases or cookbooks include compositional fields. I'll add at least one new test to go with any revisions to the material model.

jdannberg:
Now that I think about it, I have used this material model several times, but never with compositional fields. I looked into the file, and we never set the reaction terms anywhere. So if you can reproduce the issue, maybe it's just enough to add 
`for (unsigned int c=0; c<in.composition[i].size(); ++c)`
`  out.reaction_terms[i][c] = 0.0;`
to the evaluate function. 
Do we have an input file that reproduces the issue? That could be a good starting point for a test.

naliboff:
@jdannberg - that fixed it :+1:  Thanks! I'll put in a pull request this afternoon or tomorrow with the fix and a new test case using a compositional field. I may also revise some of the wording in the material model documentation, but I've leave that discussion for the pull request.

jdannberg:
Great! :smile:

bangerth:
Yay signaling NaNs -- doing exactly what we wanted them to do!

naliboff:
This issue was resolved by #1902.

PR #1902, opened 2017-08-18
Specify reaction terms in latent heat material model

naliboff:
This pull request addresses issue #1896, which was originally pointed out by Derek Neuharth (University of Idaho). Specifying the reaction terms in the latent heat material model fixes the assembly error when using compositional fields.

### DIFF
--- a/source/material_model/latent_heat.cc
+++ b/source/material_model/latent_heat.cc
@@ -305,6 +305,11 @@ namespace aspect
             out.entropy_derivative_pressure[i] = entropy_gradient_pressure;
             out.entropy_derivative_temperature[i] = entropy_gradient_temperature;
           }
+
+         // Assign reaction terms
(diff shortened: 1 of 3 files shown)


gassmoeller:
Makes sense. For some reason astyle is not happy, can you take a look?

naliboff:
@gassmoeller - Thanks for taking a look. Odd, I thought I ran astyle the first time around but perhaps not.

naliboff:
Before merging, there are a number of small edits I would like to make in the material model documentation. Should I add those in another commit here, or put them in a separate pull request?

tjhei:
> Before merging, there are a number of small edits I would like to make in the material model documentation. Should I add those in another commit here, or put them in a separate pull request?

Just make a separate PR. This makes things easier to review.

bangerth:
Yes, indeed. If things are logically separate, then they should go into separate pull requests.

PR #1904, opened 2017-08-21
allow visualization postprocessors to use the current cell

jdannberg:
This pull request adds an active cell iterator (instead of a reference to the active cell iterator, which was used before, and is now marked as deprecated) to the material model inputs structure, which allows the cell to be filled in the constructor of material model inputs that is used for the visualization postprocessors (using `get_cell<DoFHandler<dim>`). 

This is useful in cases where the material model uses the cell information to extract, for example, the old solution, or some additional solution variables that are not in the material model inputs, like in the case of melt transport, and the computes the material model outputs using this information. With this new version, visualization postprocessors will now also be able to output these quantities from material model outputs correctly, which was not possible before.

### DIFF
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -172,124 +172,153 @@ namespace aspect
     template <int dim>
     struct MaterialModelInputs
     {
-      /**
-       * Constructor. Initialize the various arrays of this structure with the
-       * given number of quadrature points and (finite element) components.
-       *
-       * @param n_points The number of quadrature points for which input
-       * quantities will be provided.
-       * @param n_comp The number of vector quantities (in the order in which
-       * the Introspection class reports them) for which input will be
-       * provided.
-       */
-      MaterialModelInputs(const unsigned int n_points,
-                          const unsigned int n_comp);
+        /**
+         * Constructor. Initialize the various arrays of this structure with the
+         * given number of quadrature points and (finite element) components.
+         *
+         * @param n_points The number of quadrature points for which input
+         * quantities will be provided.
+         * @param n_comp The number of vector quantities (in the order in which
+         * the Introspection class reports them) for which input will be
+         * provided.
+         */
+        MaterialModelInputs(const unsigned int n_points,
+                            const unsigned int n_comp);
 
-      /**
-       * Constructor. Initialize the arrays of the structure with the number
-       * of points in the `input_data` structure, and fills them appropriately.
-       *
-       * @param input_data The data used to populate the material model input quantities.
-       * @param introspection A reference to the simulator introspection object.
-       * @param use_strain_rate Whether to compute the strain rates.
-       */
-      MaterialModelInputs(const DataPostprocessorInputs::Vector<dim> &input_data,
-                          const Introspection<dim> &introspection,
-                          const bool use_strain_rate = true);
-
-
-      /**
-       * Constructor. Initializes the various arrays of this
-       * structure with the FEValues and introspection objects and
-       * the solution_vector. This constructor calls the function
-       * reinit to populate the newly created arrays.
-       *
-       * @param fe_values An FEValuesBase object used to evaluate the finite elements.
-       * @param cell The currently active cell for the fe_values object.
-       * @param introspection A reference to the simulator introspection object.
-       * @param solution_vector The finite element vector from which to construct the inputs.
-       * @param use_strain_rates Whether to compute the strain rates.
-       */
-      MaterialModelInputs(const FEValuesBase<dim,dim> &fe_values,
-                          const typename DoFHandler<dim>::active_cell_iterator *cell,
-                          const Introspection<dim> &introspection,
-                          const LinearAlgebra::BlockVector &solution_vector,
-                          const bool use_strain_rates = true);
-
-      /**
-       * Function to re-initialize and populate the pre-existing arrays
-       * created by the constructor MaterialModelInputs.
-       */
-      void reinit(const FEValuesBase<dim,dim> &fe_values,
-                  const typename DoFHandler<dim>::active_cell_iterator *cell,
-                  const Introspection<dim> &introspection,
-                  const LinearAlgebra::BlockVector &solution_vector,
-                  const bool use_strain_rates = true);
+        /**
+         * Constructor. Initialize the arrays of the structure with the number
+         * of points in the `input_data` structure, and fills them appropriately.
+         *
+         * @param input_data The data used to populate the material model input quantities.
+         * @param introspection A reference to the simulator introspection object.
+         * @param use_strain_rate Whether to compute the strain rates.
+         */
+        MaterialModelInputs(const DataPostprocessorInputs::Vector<dim> &input_data,
+                            const Introspection<dim> &introspection,
+                            const bool use_strain_rate = true);
 
 
-      /**
-       * Vector with global positions where the material has to be evaluated
-       * in evaluate().
-       */
-      std::vector<Point<dim> > position;
+        /**
+         * Constructor. Initializes the various arrays of this
+         * structure with the FEValues and introspection objects and
+         * the solution_vector. This constructor calls the function
+         * reinit to populate the newly created arrays.
+         *
+         * @param fe_values An FEValuesBase object used to evaluate the finite elements.
+         * @param cell The currently active cell for the fe_values object.
+         * @param introspection A reference to the simulator introspection object.
+         * @param solution_vector The finite element vector from which to construct the inputs.
+         * @param use_strain_rates Whether to compute the strain rates.
+         */
+        MaterialModelInputs(const FEValuesBase<dim,dim> &fe_values,
+                            const typename DoFHandler<dim>::active_cell_iterator *cell,
+                            const Introspection<dim> &introspection,
+                            const LinearAlgebra::BlockVector &solution_vector,
+                            const bool use_strain_rates = true);
 
-      /**
-       * Temperature values at the points given in the #position vector.
-       */
-      std::vector<double> temperature;
+        /**
+         * Copy Constructor.
+         */
+        MaterialModelInputs(const MaterialModelInputs<dim> &material);
 
-      /**
-       * Pressure values at the points given in the #position vector.
-       */
-      std::vector<double> pressure;
+        /**
+         * Function to re-initialize and populate the pre-existing arrays
+         * created by the constructor MaterialModelInputs.
+         */
+        void reinit(const FEValuesBase<dim,dim> &fe_values,
+                    const typename DoFHandler<dim>::active_cell_iterator *cell,
+                    const Introspection<dim> &introspection,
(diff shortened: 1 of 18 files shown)


jdannberg:
I had to update some of the tests, but the only one where I had to update the output was the cell_reference test, which now outputs the cell reference in more places, so I think that makes sense. But @gassmoeller, I think you wrote the test, maybe you could have a look?

Also, I am not sure why the copy constructor of MaterialModelInputs works if it is called as `MaterialModelInputs<dim> in (base_inputs)`, but not when we call `MaterialModelInputs<dim> in(n_points, n_comp); in = base_inputs` as in the Newton solver test cases. So it would be great if I could get some help with that, as I think the old version should also continue to work (as for now I just changed how we call the constructor in those test files).

source/material_model/interface.cc
```diff
@@ -236,7 +236,7 @@ namespace aspect
     }
 
 
-
+    DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
```
bangerth: can you say why this is necessary?

source/material_model/interface.cc
```diff
@@ -288,7 +290,6 @@ namespace aspect
         }
     }
 
-
     template <int dim>
     MaterialModelInputs<dim>::MaterialModelInputs(const FEValuesBase<dim,dim> &fe_values,
                                                   const typename DoFHandler<dim>::active_cell_iterator *cell_x,
```
bangerth: It would of course be nice to also make this into a reference rather than a pointer. how many places would this affect?

In the interest of keeping review simple, maybe keep that for a follow-up patch if you want to do it. Or open an issue about it.

source/material_model/interface.cc
```diff
@@ -303,12 +304,28 @@ namespace aspect
       velocity(fe_values.n_quadrature_points, numbers::signaling_nan<Tensor<1,dim> >()),
       composition(fe_values.n_quadrature_points, std::vector<double>(introspection.n_compositional_fields, numbers::signaling_nan<double>())),
       strain_rate(fe_values.n_quadrature_points, numbers::signaling_nan<SymmetricTensor<2,dim> >()),
-      cell(cell_x)
+      cell(cell_x),
+      current_cell(*cell_x)
```
bangerth: What happens if `cell_x` is a `NULL` pointer? Don't you have to do something like
```
  current_cell (cell_x == NULL ? typename DoFHandler<dim>::active_cell_iterator() : *cell_x)
```
?

I think it would probably be good to also do
```
  cell (cell_x == NULL ? NULL : &current_cell)
```
This way, the calling site definitely no longer needs to keep the pointed to object around, and this avoids a whole class of possible errors.

source/material_model/interface.cc
```diff
@@ -342,7 +359,11 @@ namespace aspect
             this->composition[i][c] = composition_values[c][i];
         }
 
+      DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
       this->cell = cell_x;
+      DEAL_II_ENABLE_EXTRA_DIAGNOSTICS
+
+      this->current_cell = *cell_x;
```
bangerth: same as above, you'll have to be careful if `cell_x==NULL`

source/postprocess/visualization/melt.cc
```diff
@@ -104,9 +104,9 @@ namespace aspect
         Assert (computed_quantities.size() == n_quadrature_points,    ExcInternalError());
         Assert (input_data.solution_values[0].size() == this->introspection().n_components,   ExcInternalError());
 
-        // Set use_strain_rates to false since we have no need for viscosity.
+        // Set use_strain_rates to true since the compaction viscosity might also depend on the strain rate.
         MaterialModel::MaterialModelInputs<dim> in(input_data,
-                                                   this->introspection(), false);
+                                                   this->introspection(), true);
```
bangerth: This change is independent, right? Can you break it into a separate PR?

bangerth:
What was the error message you got with the old-style
```
MaterialModelInputs<dim> in(n_points, n_comp); 
in = base_inputs;
```
call? I don't think the class has a copy constructor so you may have to initialize everything during construction time now. But the error message would tell :-)

jdannberg:
I fixed it with help from @gassmoeller. I had to add an assignment operator to MaterialModelInputs, because with the default one, the value that was copied for the cell was not a valid iterator state, and that was not allowed.

source/material_model/interface.cc
jdannberg: Otherwise we get a warning that we use cell, which is deprecated (and I can not put the block just around the single line, because it's in the constructor). 
Or did you mean I should add a comment why this is necessary?

bangerth: comments never hurt, but I primarily wanted to understand it from the perspective of a reviewer :-)

include/aspect/material_model/interface.h
```diff
@@ -217,6 +217,16 @@ namespace aspect
...
 
       /**
+       * Copy Constructor.
+       */
+      MaterialModelInputs(const MaterialModelInputs<dim> &material);
+
+      /**
+       * Assignment operator.
+       */
+      MaterialModelInputs &operator=(const MaterialModelInputs &material);
```
bangerth: I'm not actually a fan of this. These objects are big, and copying them is expensive. I'd rather just outlaw it (as it was before, implicitly by just not providing this constructor).

jdannberg: Of course I can just revert this. But I was just thinking of @MFraters. Those tests where the assignment was used were all related to the Newton solver, and I was thinking he might have a number of his own plugins and test cases that might use a similar logic.

bangerth: Hm, but then I'm curious again how this worked before. If you remove the operator, what are the errors you get from the existing code?

I'm not saying that you should revert the patch, I'm just curious to see what changed that required you to add the copy operator. I must be missing something that it's not clear to me why you need it.

gassmoeller: Removing this operator and the constructor would not help, because they are implicitly created when called, and even worse you then get a cryptic error message that you can not assign a DoFAccessor without a particle handler (apparently `active_cell_iterator a = active_cell_iterator()` is not allowed). That is why we created the constructor and operator. Should we instead mark them as deleted in the header file like this?
```
MaterialModelInputs (...) = delete;
... operator= (...) = delete;
```

bangerth: Or are you saying that this leads to an error:
```
  active_cell_iterator cell;  // invalid state
  active_cell_iterator cell2;
  cell2 = cell;    // copy of invalid state
```

bangerth: Ah, wires crossed. The `=delete` approach doesn't work because that's C++11. We need to come up with something else.

jdannberg: The error message is:
`An error occurred in line <115> of file </home/dannberg/software/dealii-8.5-install/include/deal.II/dofs/dof_accessor.templates.h> in function
    void dealii::DoFAccessor<structdim, DoFHandlerType, lda>::set_dof_handler(DoFHandlerType*) [with int structdim = 2; DoFHandlerType = dealii::DoFHandler<2>; bool level_dof_access = false]
The violated condition was: 
    dh != 0
Additional information: 
    (none)`

Basically, I think it tries to set a dof_handler, but does not have one.

jdannberg: > 
> 
> Or are you saying that this leads to an error:
> ```
>   active_cell_iterator cell;  // invalid state
>   active_cell_iterator cell2;
>   cell2 = cell;    // copy of invalid state
> ```
> 

Yes, exactly.

source/material_model/interface.cc
jdannberg: I opened an issue for now (#1907) so that we don't forget about it, and I can do it after this pull request is merged.

include/aspect/material_model/interface.h
MFraters: @jdannberg, I am fine with changing it to the `MaterialModelInputs<dim> in (base_inputs)` style in the Newton tests. I indeed have some more tests which are not part of the repository  yet, but changing it should not be much work.

jdannberg:
@bangerth, I think I addressed all of your comments except for the issue with the assignment operator. As @MFraters said it would be fine to change the constructor to the `MaterialModelInputs<dim> in (base_inputs)` style, I'd be totally fine with that, in this case we just have to figure out what to do about the cryptic error message people will get if they try to use the assignment operator anywhere without having the cell object. Do you have a good idea that works without marking the assignment operator as deleted? Should I just change the implementation to throw an assert that says you can not assign a material model inputs object without having a cell reference? Or is there a better way?

tjhei:
> Or is there a better way?

You can also declare it but make it private and don't provide an implementation. That is the pre-c++11 way of deleting it.

source/material_model/interface.cc
```diff
@@ -342,7 +385,12 @@ namespace aspect
             this->composition[i][c] = composition_values[c][i];
         }
 
+      DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
       this->cell = cell_x;
+      DEAL_II_ENABLE_EXTRA_DIAGNOSTICS
+
+      if (cell_x != NULL)
+        this->current_cell = *cell_x;
```
tjhei: don't you have to reset ``current_cell`` if ``cell_x==NULL``?

tests/cell_reference/screen-output
```diff
@@ -34,6 +34,14 @@ Level: 1 Index: 2
 Level: 1 Index: 2
 Level: 1 Index: 3
 Level: 1 Index: 3
+Level: 1 Index: 0
```
tjhei: what happened here?

jdannberg:
Okay, thanks, sounds good to me!
@bangerth what do you think, would that be a good solution?

tests/cell_reference/screen-output
jdannberg: We now have the cell object in more places than before (because one of the constructors of MaterialModelInputs was changed to use the get_cell function). So I thought it made sense that there is more output now (the evaluate function outputs the cell if it gets it).

source/material_model/interface.cc
jdannberg: Oh, yes, I didn't think about that case. Thanks! Should be fixed now.

bangerth: Oh, great catch. I missed that one as well!

include/aspect/material_model/interface.h
bangerth: Thanks, @MFraters .

@jdannberg -- what an annoying issue. I'm fixing this in deal.II, but this is of course too late for your current patch.

bangerth:
Yes, making it `private` is the correct approach. This leads to an error message that you may want to explain in the documentation of the function, namely that copying such objects is forbidden because it is too expensive.

bangerth:
I've gone through it one more time and it looks good.

Please move the copy operator into the `private` section of the class, squash the commits into one, and merge yourself. Thanks for your patience!

PR #1906, opened 2017-08-22
Change from std_cxx11::array<double,dim> to Tensor<1,dim> for cartesi…

MFraters:
Change from std_cxx11::array<double,dim> to Tensor<1,dim> for cartesian_to_natural_coordinates and natural_to_cartesian_coordinates.

If I remember correctly we made this an array instead of a Point<dim> because the point functions are designed for Cartesian coordinates. The problem I am facing is that the std::array is completely devoid of functions like adding and substracting. Is it oke to make it into a Tensor? 

This would safe me quite a bit of work in using this function. This function is quite new (last hackaton) and as far as I am aware, no one is using it yet since non of the geometry models have implemented this.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -974,6 +974,18 @@ namespace aspect
                               const double safety_factor);
 
     /**
+     * Converts an array of size dim to a Point of size dim.
+     */
+    template <int dim>
+    Point<dim> convert_array_to_point(const std_cxx11::array<double,dim> &array);
+
+    /**
+     * Converts a Point of size dim to an array of size dim.
+     */
+    template <int dim>
+    std_cxx11::array<double,dim> convert_point_to_array(const Point<dim> &point);
+
+    /**
      * A class that represents a binary operator between two doubles. The type of
      * operation is specified on construction time, and can be checked later
      * by using the operator ==. The operator () executes the operation on two
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -2411,6 +2411,30 @@ namespace aspect
 
 
 
+    template <int dim>
+    Point<dim> convert_array_to_point(const std_cxx11::array<double,dim> &array)
+    {
+      Point<dim> point;
+      for (unsigned int i = 0; i < dim; i++)
+        point[i] = array[i];
+
+      return point;
+    }
+
+
+
+    template <int dim>
+    std_cxx11::array<double,dim> convert_point_to_array(const Point<dim> &point)
+    {
+      std_cxx11::array<double,dim> array;
+      for (unsigned int i = 0; i < dim; i++)
+        array[i] = point[i];
+
+      return array;
+    }
+
+
+
     Operator::Operator()
       :
       op(uninitialized)


tjhei:
> If I remember correctly we made this an array instead of a Point because the point functions are designed for Cartesian coordinates.

Correct.

I am okay with this change, but I think we should let @bangerth comment because I now he feels strongly about it.

bangerth:
I think I forgot what the "natural coordinate system" is? Is that defined somewhere?

The reason why we wanted to use `std::array<double,dim>` is precisely because it doesn't have `operator+`, etc defined. Think about it this way: If you have two points in space, p1 and p2, then if you do `p1 - p2`, you expect to get a vector `x` that points from one to the other.

In Cartesian coordinates, the vector `x` is simply the difference of coordinates between `p1` and `p2`. But if you work in spherical coordinate systems, `x.radius` is not equal to `p1.radius - p2.radius` not is `x.angle` equal to `p1.angle - p2.angle`. In other words, you *specifically* do not want to use a data type like `Point<dim>` or `Tensor<1,dim>` to represent points in coordinate systems in which taking the difference of points (or adding them, or multiplying them with a scalar) works differently than in Cartesian coordinate systems.

The fact that you want to use `operator+`, `operator*`, etc, to me means that you are about to introduce a bug. Namely, if you used `operator+` somewhere to add together points in their "natural coordinate system" will be correct *only if that natural coordinate system is Cartesian*. The result will be wrong (without a compiler error or warning) if you are working in any kind of coordinate system that is not specifically Cartesian.

Does this make sense?

MFraters:
The natural coordinate system is the system which is logical to use for the geometry model. So in the case of the box model it is just cartesian, in the case of the spherical model it is spherical coordinates and in the case of an ellipse, it is ellipsoidal coordinates. 

It makes some sense to me, but let me explain what I am doing with them, and why I still think I need them. 

I am using this to generate initial conditions and I want that the system is independent of the geometry model which is used. That is why we added this function. Let me take as an example a temperature initial condition in ellipsoidal coordinates. The temperature function receives a position in Cartesian coordinates (x,y,z). The first thing I do is get access to the geometry model interface and transform the position to the 'natrual' coordinates, in this case longitude, latitude and depth. Futhermore I get the depth with the cartesian coordinates through the geometry model depth function.

I now treat the natural coordinates as a box with x range = longitude, y range latitude and z range = depth. Since the users have provided their coordinates in the 'natural' coordinates, I think I can at least use addition, substraction etc. to for example compute whether an ellipsoidal coordinate (2d, long:lat) is inside or outside a polygon made of ellipsoidal coordinates (2d, long:lat) using for example the utilities polygon_contains_point function. This works very well for making for example simple continents where the equation to compute the temperature only needs to know if it is within or outside the polygon and at what depth it is (which we got with the Cartesian coordinates through the depth function). 

Now you are certainly right that once you need distances (like for example an oceanic plate with a plate model temperature, where the temperature profile changes with distance from the ridge) things get a bit more complicated, because you have to compute the arc lengths. The way I do this is a bit more involved (am am actually approximating the earth to be a sphere here, even in the ellipsoidal case), but I actually make heavy use of addition and subtraction of points/tensors to achieve this. 

The hardest part is when I want to make subducting slabs. Although the new things do not change the discussion here (because the natural_to_cartesian_coordinates already returns a point, although it uses every trick mentioned before), just for the sake of completeness I will describe what I do shortly. What I basically need to do is find the distance to one (or many, potentially curved) planes. What I do is that I turn all my ellipsoidal coordinates (including the ones handed to me by the input file) back into Cartesian coordinates with the help of the natural_to_cartesian_coordinates function. Because the Cartesian points in this case are vectors from the centre of the earth (and are by definition normal to the surface) I can do all kinds of rotation and computations to solve this problem.

What I do with the slabs, might actually be an approximation, but for the rest I think I am using the ellipsoidal/spherical coordinates correctly. Also I have tested the world generator extensively in the ellipsoidal chunk and it seems to work correctly. So I think that at least the `operator+` and `operator-` can be used.

tjhei:
One compromise would be to write little helper functions to compute between an ``array`` and a ``Tensor<1,dim>`` and back that you could use in those specific places. It sounds a little awkward, but on the plus side forces you to think about whether the operations you are doing make sense.

MFraters:
That could be a solution indeed, but I am a bit afraid it will become quite expensive, because this part of the code will be called many times during the initialisation, and I already have to transform it from points to arrays back. This is because pull and push functions require and return Point<dim>. Is this fear ungrounded? And should the push and pull function in the manifolds also not be changed, because they are not really points in a Cartesian sense?

bangerth:
I don't think you need to be worried about converting 3-element data objects to other 3-element data objects. Compilers decompose these into their scalar components these days and it will all be held in registers -- you're not going to be able to measure these conversions.

As for the manifolds: A manifold (by mathematical definition) is something that can locally be pulled back to a Cartesian coordinate system in an Euclidean space: https://en.wikipedia.org/wiki/Manifold . We also embed it into a Cartesian space. So the choice of data type makes sense there. (It does get a bit complicated because the pull-back only has to work locally, and so the pull-back coordinate system can, for example, be periodic.)

MFraters:
Ok, I guess the best way to go would be to add a utilities function to make this conversion?

Just out of curiosity, I am not sure I understand the difference. In both cases (using the push and pull and using the natural_to_cartesian_coordinates() and cartesian_to_natural_coordinates() functions) Cartesian coordinates are converted to for example spherical coordinates and back. The implementation of the natural_to_cartesian_coordinates() function and the reverse is in this case is, besides converting from and to arrays, literally just calling the manifold push/pull function. 

As you (and the Wikipedia article) say, manifolds only locally resemble the Euclidean space. But since these functions are just a shell for the manifold in the case where there is a manifold, I would think this would yield in both cases exactly the same problems when you use the plus/minus/cross-product/distance/etc operators of Point.

bangerth:
@MFraters -- I've thought about what you want to do some more, but I think I would need to see an example of where you get in trouble with the existing interface. Let me give you an example why I don't want `Point<dim>` for the natural coordinate system. Think, for example, of two points in your domain, and that you want to know where their midpoint is. In Cartesian coordinates, it is simply `(p1+p2)/2`. But if `p1` and `p2` are expressed in spherical coordinates -- let's say that they are `(r,phi,theta)=(1,170,30)` and `(1,-170,30)`, then `(p1+p2)/2` would yield `(1,0,30)` when it should really have been `(1,180,30)`. That's because the plus and division operations do not take into account the periodicity of the longitude -- the periodicity being the difference between a locally Cartesian coordinate system and a globally Cartesian coordinate system. (Of course, if you ask where the midpoint in Cartesian coordinates would be, then you'd find that it has radius less than one; if you define the midpoint as the one on a great circle, it would have a latitude of more than 30 degrees.)

What the example shows is that operations such as plus or minus are, at least, fraught with problems. They may give you the answer that makes sense in a particular context, but it's at the very least easy to make mistakes that will be very difficult to find if you don't force the author of a piece of code to actually think about what they are doing. The reason for not wanting to use `Point<dim>` is in essence to *require the author to think*.

Now, in your case it may very well be that you've thought about things and are only using operations that are actually valid and make sense. In fact, I'm sure you do, from your description above. So my suggestion is to have tool functions that convert an array to a point and the other way around (say, `express_natural_coordinates_as_point()` and `express_point_as_natural_coordinates()`, or some other more suitable name). You'd call it in the function where you need it and have come to the conclusion that it makes sense there, but leave all other uses alone. Would that work?

MFraters:
I still have some questions, but I guess that it's better to discuss them when we meet again at the next hackaton or so.

I named the functions in utilities after what they directly do: convert array to point and the other way around. Is this fine, or do you also want in the naming of the function that it is only meant for this purpose?

include/aspect/utilities.h
```diff
@@ -974,6 +974,18 @@ namespace aspect
                               const double safety_factor);
 
     /**
+     * Converts an array of size dim to an Point of size dim.
```
bangerth: -> a Point

MFraters:
addressed comments

PR #1908, opened 2017-08-22
Split cartesean natrual coordinate dimension templates

MFraters:
This is a follow up to #1906. Like #1906, this is meant as a proposition to start a discussion. 

The problem I try to solve here is that I have functions which always require 3d points, no matter if you are in 2d or 3d). To accommodate that I have split these functions in explicit 2d and 3d input and output. Are there any objections against doing this?

### DIFF
--- a/include/aspect/geometry_model/interface.h
+++ b/include/aspect/geometry_model/interface.h
@@ -138,7 +138,7 @@ namespace aspect
          * latitude) in 3d.
          */
         virtual
-        std_cxx11::array<double,dim> cartesian_to_natural_coordinates(const Point<dim> &position) const;
+        Tensor<1,dim> cartesian_to_natural_coordinates(const Point<dim> &position) const;
 
         /**
          * Undoes the action of cartesian_to_natural_coordinates, and turns the
@@ -146,7 +146,7 @@ namespace aspect
          * Cartesian coordinates.
          */
         virtual
-        Point<dim> natural_to_cartesian_coordinates(const std_cxx11::array<double,dim> &position) const;
+        Point<dim> natural_to_cartesian_coordinates(const Tensor<1,dim> &position) const;
 
         /**
          * Returns a representative point for a given depth. Such a point must
--- a/include/aspect/geometry_model/interface.h
+++ b/include/aspect/geometry_model/interface.h
@@ -138,7 +138,10 @@ namespace aspect
          * latitude) in 3d.
          */
         virtual
-        Tensor<1,dim> cartesian_to_natural_coordinates(const Point<dim> &position) const;
+        Tensor<1,2> cartesian_to_natural_coordinates(const Point<2> &position) const;
+
+        virtual
+        Tensor<1,3> cartesian_to_natural_coordinates(const Point<3> &position) const;
 
         /**
          * Undoes the action of cartesian_to_natural_coordinates, and turns the
@@ -146,7 +149,10 @@ namespace aspect
          * Cartesian coordinates.
          */
         virtual
-        Point<dim> natural_to_cartesian_coordinates(const Tensor<1,dim> &position) const;
+        Point<2> natural_to_cartesian_coordinates(const Tensor<1,2> &position) const;
+
+        virtual
+        Point<3> natural_to_cartesian_coordinates(const Tensor<1,3> &position) const;
 
         /**
          * Returns a representative point for a given depth. Such a point must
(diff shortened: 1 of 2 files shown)


tjhei:
Please explain this a bit better. What is a geometry model like a 3d spherical shell supposed to do if you give it 2d coordinates or the other way around?

MFraters:
As long as there is no difference between dim and spacedim in ASPECT, they cannot be implemented and the assert will be thrown. 

So in theory this wouldn't be necessary, but when I compile the code, and I do something like: `dealii::Tensor<1,3> check_point_pushed = geometry_model.natural_to_cartesian_coordinates(check_point);`, I get a compiler error that it can't find the 2d version of this function with tensor<1,3> in and output, even when it is enclosed in a if(dim == 3).

Is there maybe an other way to solve this problem?

bangerth:
I think this is starting at the wrong end. As I just wrote in #1906, it is not correct to represent points in a "natural coordinate system" as `Tensor<1,dim>` or as `Point<dim>`. We will have to find a different way to do what you want to do.

In your example, what data type does `check_point` have? Is it always a `Tensor<1,3>`? If so, yes, that can't work in 2d because `natural_to_cartesian_coordinates()` takes something with 2 components. So the question would be where does `check_point` come from?

MFraters:
`check_point` is currently a `dealii::Point<3>`, but it is enclosed in a `if(dim == 3)`. 

> If so, yes, that can't work in 2d because natural_to_cartesian_coordinates() takes something with 2 components.

I am not sure if I understand what you are saying here, because what is pull request is changing is that it will be possible in 2d to give 3d points to the function. As said before, this is not useful in ASPECT because dim = spacedim. But I need it because the compiler somehow doesn't recognize that the whole block will never be used in 2d, because it is enclosed in a `if(dim == 3)` statement. 

The check point comes from the fact that the code I have written is intrinsically 3d. I would find it interesting to maybe see in the future that aspect can make 2d cross sections through it, or I could rewrite a whole lot of code to explicitly support 2d initial conditions.

Again, if there is an other way to solve the compiler issue, I am also happy with that.

tjhei:
> Again, if there is an other way to solve the compiler issue, I am also happy with that.

Yes, I understand now. There is no way to write code that depends on the dimension in an if block like this because it needs to compile. I guess there are three options:
1. Use template spezialization and write a separate 2d and 3d function instead of an ``if (dim==2)`` block. If necessary, you can move individual parts into a small function that is templated on dimension.
2. Write dimension independent code (use ``Point<dim>`` and loops over ``dim``, etc.)
3. We add the functions like propose here with a default implementation that throws if the dimensions don't match.

bangerth:
> I am not sure if I understand what you are saying here, because what is 
> pull request is changing is that it will be possible in 2d to give 3d 
> points to the function. As said before, this is not useful in ASPECT 
> because dim = spacedim. But I need it because the compiler somehow 
> doesn't recognize that the whole block will never be used in 2d, because 
> it is enclosed in a |if(dim == 3)| statement.
That doesn't matter -- C++ says that even if there is code in an 'if' or 'else' block that is dead, it still needs to be syntactically and semantically correct. That's primarily because the decision of whether the 'if' or 'else' branch is dead happens *after* the compiler has to understand the code in these branches. (I'll add that C++17 has a construct that makes possible what you want to do, but we can't use that yet.)

The way to work around this is as follows: If you have code such as
```
  template <int dim>
  class X{
    void some_function (Point<dim>);

    void some_other_function () {
       Point<3>  p;
       if (dim == 3)
         some_function (p);      // doesn't compile in 2d, because some_function wants a Point<2>
   }
};
```
then you can do something like this:
```
  template <int dim>
  class X{
    void some_function (Point<dim>);

    void some_other_function () {
       Point<3>  p;
       if (dim == 3)
         some_function (reinterpret_cast<Point<dim>&>(p));
   }
};
```
That's because in 3d, the `reinterpret_cast` is the identity operation and consequently does nothing. Whereas in 2d, the cast would be wrong, but is never executed.

We typically add a comment to such code to explain that the cast will not be executed unless it is safe.

MFraters:
Thanks for the tips and explanation. 

Wolfgangs trick works, although I get some warnings about that dereferencing type-punned pointers breaks strict-aliasing rules. Since this solve the problem I had, without changing the interface, I will close this pull request. 

Thanks for the help! :)

Issue #1907, opened 2017-08-22
make the cell object input argument in the constructor of material models inputs a reference instead of a pointer


jdannberg:
This would be a follow-up to #1904, where we change the material model inputs structure to have the cell object (instead of a pointer to the cell) as a variable, so that it can be used in more places in the code.

bangerth:
For the place where this came up, see https://github.com/geodynamics/aspect/pull/1904/files#r134359434

gassmoeller:
Was this closed by #1910?

jdannberg:
Yes.

PR #1910, opened 2017-08-23
make the cell object input argument in the constructor of material mo…

jdannberg:
…del inputs a reference instead of a pointer

This addresses #1907. 
However, it only adds the new functions. What do we do with the old versions? Keep them around and mark them as deprecated too? Otherwise postprocessors people have written might not work anymore.

### DIFF
--- a/source/material_model/interface.cc
+++ b/source/material_model/interface.cc
@@ -294,7 +294,7 @@ namespace aspect
 
     template <int dim>
     MaterialModelInputs<dim>::MaterialModelInputs(const FEValuesBase<dim,dim> &fe_values,
-                                                  const typename DoFHandler<dim>::active_cell_iterator *cell_x,
+                                                  const typename DoFHandler<dim>::active_cell_iterator &cell_x,
                                                   const Introspection<dim> &introspection,
                                                   const LinearAlgebra::BlockVector &solution_vector,
                                                   const bool use_strain_rate)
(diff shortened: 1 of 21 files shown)


tjhei:
Looks good. I would be okay with leaving things as is. The constructor for the inputs is very new after all. I'll let others comment on it first though.

source/material_model/interface.cc
```diff
@@ -306,11 +306,11 @@ namespace aspect
       velocity(fe_values.n_quadrature_points, numbers::signaling_nan<Tensor<1,dim> >()),
       composition(fe_values.n_quadrature_points, std::vector<double>(introspection.n_compositional_fields, numbers::signaling_nan<double>())),
       strain_rate(fe_values.n_quadrature_points, numbers::signaling_nan<SymmetricTensor<2,dim> >()),
-      cell(cell_x == NULL ? NULL : &current_cell),
-      current_cell(cell_x == NULL ? typename DoFHandler<dim>::active_cell_iterator() : *cell_x)
+      cell(cell_x.state() == IteratorState::valid ? &current_cell : NULL),
+      current_cell(cell_x.state() == IteratorState::valid ? cell_x : typename DoFHandler<dim>::active_cell_iterator())
```
bangerth: you can safely just copy `cell_x` here now.

source/material_model/interface.cc
```diff
@@ -362,10 +362,10 @@ namespace aspect
         }
 
       DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
-      this->cell = cell_x;
+      this->cell = cell_x.state() == IteratorState::valid ? &cell_x : NULL;
       DEAL_II_ENABLE_EXTRA_DIAGNOSTICS
 
-      this->current_cell = (cell_x == NULL ? typename DoFHandler<dim>::active_cell_iterator() : *cell_x);
+      this->current_cell = (cell_x.state() == IteratorState::valid ? cell_x : typename DoFHandler<dim>::active_cell_iterator());
```
bangerth: same here

bangerth:
I'm also ok with just making the change. I don't think we had a release during which the old constructor was available.

source/material_model/interface.cc
jdannberg: For both cell and current_cell or only one of them? I don't really understand why I can just copy it now...

bangerth: Only for the second one.

It's not obvious maybe, but
```
cell_x.state() == IteratorState::valid ? cell_x : typename DoFHandler<dim>::active_cell_iterator()
```
is equivalent to
```
cell_x
```
because
```
 ! (cell_x.state() == IteratorState::valid)
```
implies
```
  cell_x == typename DoFHandler<dim>::active_cell_iterator()
```

gassmoeller: But we can not copy `active_cell_iterator()` because its copy constructor throws an assertion. That is what confused @jdannberg. This may be fixed in deal.II development, but not 8.5.

bangerth: Oh, doh, yes. That's the annoying difference between theory and practice.

Someone is going to come across this at one point and wonder. Can you write this as
```
#if DEAL_II_VERSION_GTE(9,0)
  current_cell (cell_x)
#else
  current_cell(cell_x.state() == IteratorState::valid ? cell_x : typename DoFHandler<dim>::active_cell_iterator())
#endif
```
to remind ourselves that we can simplify this at a later time?

tjhei: I think it needs to be ``DEAL_II_VERSION_GTE(9,0,0)``

jdannberg:
Okay, should be fixed now, and I also looked at the last release and you are right, it does not have the constructor for the material model.

Issue #1897, opened 2017-08-17
Particle rewrite


bangerth:
@gassmoeller is currently rewriting the particle system so that it can move into deal.II (or at least the generic parts). This is the meta-issue that tracks all of the relevant PRs.

bangerth:
@gassmoeller -- is there anything that's left here, or can we close this?

gassmoeller:
I would like to keep this open until I finish #1957, currently we still use the ASPECT version of the code instead of the one in deal.II.

tjhei:
:clinking_glasses:

gassmoeller:
Yay :tada:

PR #1913, opened 2017-08-29
Improve particle handler

gassmoeller:
This is work in progress and related to #1897. I moved most of the generic particle functionality into the particle handler class, but there is still a bug inside the serialization / transfer / deserialization of particles. I would already appreciate a review, since most stuff should be in its final state, I just need to find the problem.

### DIFF
--- a/include/aspect/particle/particle_handler.h
+++ b/include/aspect/particle/particle_handler.h
@@ -31,9 +31,12 @@
 #include <deal.II/base/subscriptor.h>
 #include <deal.II/base/array_view.h>
 #include <deal.II/base/smartpointer.h>
+#include <deal.II/fe/mapping.h>
 
 #include <boost/serialization/map.hpp>
 #include <boost/range/iterator_range.hpp>
+
+#include <deal.II/base/std_cxx11/function.h>
 
 namespace aspect
 {
@@ -61,6 +64,11 @@ namespace aspect
         typedef ParticleIterator<dim,spacedim> particle_iterator;
 
         /**
+         * A type that represents a range of particles.
+         */
+        typedef boost::iterator_range<particle_iterator> particle_iterator_range;
+
+        /**
          * Default constructor.
          */
         ParticleHandler();
@@ -72,7 +80,9 @@ namespace aspect
          *
          */
         ParticleHandler(const parallel::distributed::Triangulation<dim,spacedim> &tria,
-                        const MPI_Comm mpi_communicator);
+                        const Mapping<dim,spacedim> &mapping,
+                        const MPI_Comm mpi_communicator,
+                        const unsigned int n_properties = 0);
 
         /**
          * Destructor.
@@ -86,7 +96,9 @@ namespace aspect
          * serialization of a particle handler.
          */
         void initialize(const parallel::distributed::Triangulation<dim,spacedim> &tria,
-                        const MPI_Comm mpi_communicator);
+                        const Mapping<dim,spacedim> &mapping,
+                        const MPI_Comm mpi_communicator,
+                        const unsigned int n_properties = 0);
 
         /**
          * Clear all particle related data.
@@ -118,14 +130,57 @@ namespace aspect
          * the particles in a particular cell. The last iterator is the first
          * particle that is no longer in the cell.
          */
-        boost::iterator_range<particle_iterator>
-        particle_range_in_cell(const typename parallel::distributed::Triangulation<dim>::active_cell_iterator &cell);
+        particle_iterator_range
+        particle_range_in_cell(const typename parallel::distributed::Triangulation<dim,spacedim>::active_cell_iterator &cell);
 
         /**
          * Remove a particle pointed to by the iterator.
          */
         void
         remove_particle(const particle_iterator &particle);
+
+        /**
+         * Insert a particle into the collection of particles. Return an iterator
+         * to the new position of the particle. This function involves a copy of
+         * the particle and its properties. Note that this function is of NlogN
+         * complexity for N particles.
+         */
+        particle_iterator
+        insert_particle(const Particle<dim,spacedim> &particle,
+                        const typename parallel::distributed::Triangulation<dim>::active_cell_iterator &cell);
+
+        /**
+         * This function allows to register three additional functions that are
+         * called every time a particle is transferred to another process
+         * (i.e. during sorting into cells, during ghost particle transfer, or
+         * during serialization of all particles).
+         *
+         * @param size_callback A function that is called when serializing
+         * particle data. The function gets no arguments and is expected to
+         * return the size of the additional data that is serialized per
+         * particle. Note that this currently implies the data size has to be
+         * the same for every particle.
+         * @param store_callback A function that is called once per particle
+         * when serializing particle data. Arguments to the function are a
+         * particle iterator that identifies the current particle and a void
+         * pointer that points to a data block of size size_callback() in which
+         * the function can store additional data. The function is expected to
+         * return a void pointer pointing to a position right after its data
+         * block.
+         * @param load_callback A function that is called once per particle
+         * when deserializing particle data. Arguments to the function are a
+         * particle iterator that identifies the current particle and a void
+         * pointer that points to a data block of size size_callback() in which
+         * additional data was stored by the store_callback function. The
+         * function is expected to return a void pointer pointing to a position
+         * right after its data block.
+         */
+        void
+        register_additional_store_load_functions(const std_cxx11::function<std::size_t ()> &size_callback,
+                                                 const std_cxx11::function<void *(const particle_iterator &,
+                                                                                  void *)> &store_callback,
+                                                 const std_cxx11::function<const void *(const particle_iterator &,
+                                                                                        const void *)> &load_callback);
 
         /**
          * Return the total number of particles that were managed by this class
@@ -144,9 +199,50 @@ namespace aspect
         types::particle_index n_locally_owned_particles() const;
 
         /**
+         * Return a reference to the property pool that owns all particle
+         * properties, and organizes them physically.
+         */
+        PropertyPool &
+        get_property_pool() const;
+
+        /**
          * Return the number of particles in the given cell.
          */
-        unsigned int n_particles_in_cell(const typename Triangulation<dim>::active_cell_iterator &cell) const;
+        unsigned int
+        n_particles_in_cell(const typename Triangulation<dim,spacedim>::active_cell_iterator &cell) const;
+
+        /**
+         * Returns a vector that contains a tensor for every vertex-cell
+         * combination of the output of dealii::GridTools::vertex_to_cell_map()
+         * (which is expected as input parameter for this function).
+         * Each tensor represents a geometric vector from the vertex to the
+         * respective cell center.
+         */
+        std::vector<std::vector<Tensor<1,spacedim> > >
+        vertex_to_cell_centers_directions(const std::vector<std::set<typename parallel::distributed::Triangulation<dim,spacedim>::active_cell_iterator> > &vertex_to_cells) const;
+
+        /**
+         * Finds the cells containing each particle for all locally owned
+         * particles. If particles moved out of the local subdomain
+         * they will be sent to their new process and inserted there.
+         * After this function call every particle is either on its current
+         * process and in its current cell, or deleted (if it could not find
+         * its new process or cell).
+         *
+         * TODO: Extend this to allow keeping particles on other processes
+         * around (with an invalid cell).
+         */
+        void
+        sort_particles_into_subdomains_and_cells();
+
+
+        /**
+         * Exchanges all particles that live in cells that are ghost cells to
+         * other processes. Clears and re-populates the ghost_neighbors
+         * member variable.
+         */
+        void
+        exchange_ghost_particles();
 
         /**
          * Serialize the contents of this class.
@@ -171,11 +267,30 @@ namespace aspect
         const std::multimap<types::LevelInd, Particle<dim,spacedim> > &
         get_particles() const;
 
+        /**
+         * Const access to particles in this world.
+         * TODO: This function needs to go to not expose internal structure.
+         * This can only be done once World no longer uses this function.
+         */
+        const std::multimap<types::LevelInd, Particle<dim,spacedim> > &
+        get_ghost_particles() const;
+
       private:
+        /**
+         * A private typedef for cell iterator that makes the code of this class
+         * easier to read.
+         */
+        typedef typename parallel::distributed::Triangulation<dim,spacedim>::active_cell_iterator active_cell_it;
+
         /**
          * Address of the triangulation to work on.
          */
         SmartPointer<const parallel::distributed::Triangulation<dim,spacedim>,ParticleHandler<dim,spacedim> > triangulation;
+
+        /**
+         * Address of the mapping to work on.
+         */
+        SmartPointer<const Mapping<dim,spacedim>,ParticleHandler<dim,spacedim> > mapping;
 
         /**
          * MPI communicator.
@@ -187,6 +302,13 @@ namespace aspect
          * the level/index of the cell they are in.
          */
         std::multimap<types::LevelInd, Particle<dim,spacedim> > particles;
+
+        /**
+         * Set of particles currently in the ghost cells of the local domain,
+         * organized by the level/index of the cell they are in. These
+         * particles are marked read-only.
+         */
+        std::multimap<types::LevelInd, Particle<dim,spacedim> > ghost_particles;
 
         /**
          * This variable stores how many particles are stored globally. It is
@@ -209,6 +331,51 @@ namespace aspect
          * globally in case new particles need to be generated.
          */
         types::particle_index next_free_particle_index;
+
+        /**
+         * This object owns and organizes the memory for all particle
+         * properties.
+         */
+        std_cxx11::unique_ptr<PropertyPool> property_pool;
+
+        /**
+         * A function that can be registered by calling
+         * register_additional_store_load_functions. It is called when serializing
+         * particle data. The function gets no arguments and is expected to
+         * return the size of the additional data that is serialized per
+         * particle. Note that this currently implies the data size has to be
+         * the same for every particle, but it does not have to be the same for
+         * every serialization process (e.g. a serialization during particle
+         * movement might include temporary data, while a serialization after
+         * movement was finished does not need to transfer this data).
+         */
+        std_cxx11::function<std::size_t ()> size_callback;
+
+        /**
+         * A function that can be registered by calling
+         * register_additional_store_load_functions. It is called once per
+         * particle when serializing particle data. Arguments to the function
+         * are a particle iterator that identifies the current particle and a void
+         * pointer that points to a data block of size size_callback() in which
+         * the function can store additional data. The function is expected to
+         * return a void pointer pointing to a position right after its data
+         * block.
+         */
+        std_cxx11::function<void *(const particle_iterator &,
+                                   void *)> store_callback;
+
+        /**
+         * A function that is called once per particle
+         * when deserializing particle data. Arguments to the function are a
+         * particle iterator that identifies the current particle and a void
+         * pointer that points to a data block of size size_callback() from
+         * which the function can load additional data. This block was filled
+         * by the store_callback function during serialization. This function
+         * is expected to return a void pointer pointing to a position right
+         * after its data block.
+         */
+        std_cxx11::function<const void *(const particle_iterator &,
+                                         const void *)> load_callback;
 
         /**
          * Calculates the number of particles in the global model domain.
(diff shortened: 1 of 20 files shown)


include/aspect/particle/particle_handler.h
```diff
@@ -128,6 +139,50 @@ namespace aspect
...
+         */
+        particle_iterator
+        insert_particle(const Particle<dim,spacedim> &particle,
+                        const typename parallel::distributed::Triangulation<dim>::active_cell_iterator &cell);
+
+        /**
+         * This function allows to register three additional functions that are
+         * called every time a particle is transferred to another process
+         * (i.e. during sorting into cells, during ghost particle transfer, or
+         * during serialization of all particles.
```
bangerth: missing `)`

include/aspect/particle/particle_handler.h
```diff
@@ -128,6 +139,50 @@ namespace aspect
...
+         * @param load_callback A function that is called once per particle
+         * when deserializing particle data. Arguments to the function are a
+         * particle iterator that identifies the current particle and a void
+         * pointer that points to a data block of size size_callback() in which
+         * additional data was stored by the store_callback function. The
+         * function is expected to return a void pointer pointing to a position
+         * right after its data block.
+         */
+        void
+        register_additional_store_load_functions(const std_cxx11::function<size_t ()> &size_callback,
```
bangerth: std::size_t to be consistent with the other functions

include/aspect/particle/particle_handler.h
```diff
@@ -128,6 +139,50 @@ namespace aspect
...
+         * during serialization of all particles.
+         *
+         * @param size_callback A function that is called when serializing
+         * particle data. The function gets no arguments and is expected to
+         * return the size of the additional data that is serialized per
+         * particle. Note that this currently implies the data size has to be
+         * the same for every particle, but it does not have to be the same for
+         * every serialization process (e.g. a serialization during particle
+         * movement might include temporary data, while a serialization after
+         * movement was finished does not need to transfer this data).
```
bangerth: Out of curiosity, how is the callback supposed to know what is happening at the time it is called?

include/aspect/particle/particle_handler.h
```diff
@@ -128,6 +139,50 @@ namespace aspect
...
+         * when deserializing particle data. Arguments to the function are a
+         * particle iterator that identifies the current particle and a void
+         * pointer that points to a data block of size size_callback() in which
+         * additional data was stored by the store_callback function. The
+         * function is expected to return a void pointer pointing to a position
+         * right after its data block.
+         */
+        void
+        register_additional_store_load_functions(const std_cxx11::function<size_t ()> &size_callback,
+                                                 const std_cxx11::function<void *(const particle_iterator,
```
bangerth: maybe make the iterator argument a reference? same for the third callback

include/aspect/particle/particle_handler.h
```diff
@@ -128,6 +139,50 @@ namespace aspect
...
+                                                                                  void *)> &store_callback,
+                                                 const std_cxx11::function<const void *(const particle_iterator,
+                                                                                        const void *)> &load_callback);
```
bangerth: Here's some food for thought, not necessarily meant as a request to change things but for you to think about: In the `hp::DoFHandler` work, I've made these things more flexible by just letting the store functions return a `std::vector<char>` (could also have been a `std::string`) that the store function can size and fill as it pleases. I then let the place that calls all of these callbacks deal with the question of how to pack this data into buffers. That was easy in that case because it did not have to squeeze things into the p4est framework of attaching data to cells, but I found it kind of liberating to do it this way around because the store and load functions become so much simpler.  It required a little bit more logic in one central place, where I call these functions, but I thought that was worth it. Could something like this be done here as well?

include/aspect/particle/particle_handler.h
```diff
@@ -144,9 +199,60 @@ namespace aspect
...
          */
-        unsigned int n_particles_in_cell(const typename Triangulation<dim>::active_cell_iterator &cell) const;
+        unsigned int
+        n_particles_in_cell(const typename Triangulation<dim,spacedim>::active_cell_iterator &cell) const;
+
+        /**
+         * Get a map between subdomain id and the neighbor index. In other words
+         * the returned map answers the question: Given a subdomain id, which
+         * neighbor of the current processor's domain (in terms of a contiguous
+         * number from 0 to n_neighbors) owns this subdomain?
```
bangerth: so is the value of the map then a number between zero and `n_mpi_processes(...)`? (it's not an MPI rank, I think.) maybe say this -- I had to read this sentence a few times before I understood what the function does.

By the way do you need this function to be public?

include/aspect/particle/particle_handler.h
```diff
@@ -144,9 +199,60 @@ namespace aspect
...
+         * After this function call every particle is either on its current
+         * process and in its current cell, or deleted (if it could not find
+         * its new process or cell).
+         *
+         * @param [in] particles_to_sort Vector containing all pairs of
+         * particles and their old cells that will be sorted into the
+         * 'particles' member variable in this function.
+         */
+        void
+        sort_particles_in_subdomains_and_cells();
```
bangerth: the documentation here does not match the signature any more, as it talks about `particles_to_sort`.

include/aspect/particle/particle_handler.h
```diff
@@ -144,9 +199,60 @@ namespace aspect
...
+
+
+        /**
+         * Exchanges all particles that live in cells adjacent to ghost cells
```
bangerth: that live in cells that are ghost cells on other processors?

include/aspect/particle/particle_handler.h
```diff
@@ -171,13 +277,33 @@ namespace aspect
         const std::multimap<types::LevelInd, Particle<dim,spacedim> > &
         get_particles() const;
 
+        /**
+         * Const access to particles in this world.
+         * TODO: This function needs to go to not expose internal structure.
+         * This can only be done once World no longer uses this function.
```
bangerth: Are you worried about exposing this function because it shows the underlying data structure when you carefully hide the data structure for the locally owned particles?

include/aspect/particle/particle_handler.h
```diff
@@ -171,13 +277,33 @@ namespace aspect
...
+        typedef typename parallel::distributed::Triangulation<dim,spacedim>::active_cell_iterator active_cell_it;
+
+
+        /**
          * Address of the triangulation to work on.
          */
         SmartPointer<const parallel::distributed::Triangulation<dim,spacedim>,ParticleHandler<dim,spacedim> > triangulation;
 
         /**
+         * Address of the triangulation to work on.
```
bangerth: -> mapping

include/aspect/particle/particle_handler.h
```diff
@@ -211,6 +344,58 @@ namespace aspect
...
+         * pointer that points to a data block of size size_callback() in which
+         * the function can store additional data. The function is expected to
+         * return a void pointer pointing to a position right after its data
+         * block.
+         */
+        std_cxx11::function<void *(const particle_iterator,
+                                   void *)> store_callback;
+
+        /**
+         * @param size_callback A function that is called when serializing
```
bangerth: remove the `@param`

include/aspect/particle/particle_handler.h
```diff
@@ -237,6 +422,50 @@ namespace aspect
...
+         *
+         * @param [in,out] received_particles Vector that stores all received
+         * particles. Note that it is not required nor checked that the list
+         * is empty, received particles are simply attached to the end of
+         * the vector.
+         */
+        void
+        send_recv_particles(const std::vector<std::vector<particle_iterator> > &sent_particles,
+                            std::multimap<types::LevelInd,Particle <dim> >     &received_particles,
+                            const std::vector<std::vector<active_cell_it> > &sent_cells = std::vector<std::vector<active_cell_it> > ());
```
bangerth: can you also document the last argument?

include/aspect/particle/particle_handler.h
```diff
@@ -237,6 +422,50 @@ namespace aspect
         update_next_free_particle_index();
 
         /**
+         * Transfer particles that have crossed subdomain boundaries to other
+         * processors. The transfer occurs in two steps. As a first step all
```
bangerth: I'd drop the explanation of *how* the function works from the documentation. What's relevant here is *what* the function does. I say that because I think I'll have comments on how this is implemented once I get to this part of the patch :-)

include/aspect/particle/particle_handler.h
```diff
@@ -237,6 +422,50 @@ namespace aspect
...
+         * processes notify their neighbor processes how many particles will
+         * be sent to them. Because neighbor processes are defined as the owner
+         * of ghost cells of the current process, this also handles
+         * periodic boundaries correctly. Afterwards the transfer is done in the
+         * same way as local communication between neighbor processes.
+         * All received particles and their new cells will be appended to the
+         * @p received_particles vector.
+         *
+         * @param [in] sent_particles All particles that should be sent and
+         * their new subdomain_ids are in this map.
```
bangerth: maybe call it `particles_to_send`? how are the nested vectors organized, i.e., what do the indices mean?

include/aspect/particle/particle_handler.h
```diff
@@ -237,6 +422,50 @@ namespace aspect
...
+        void
+        send_recv_particles(const std::vector<std::vector<particle_iterator> > &sent_particles,
+                            std::multimap<types::LevelInd,Particle <dim> >     &received_particles,
+                            const std::vector<std::vector<active_cell_it> > &sent_cells = std::vector<std::vector<active_cell_it> > ());
+
+
+        /**
+         * Called by listener functions from Triangulation for every cell
+         * before a refinement step. All particles have to be attached to their
+         * element to be sent around to the new cell/processes.
```
bangerth: element -> cell

include/aspect/simulator.h
```diff
@@ -1391,6 +1390,8 @@ namespace aspect
 
       DoFHandler<dim>                                           dof_handler;
 
+      Postprocess::Manager<dim>                                 postprocess_manager;
+
```
bangerth: any particular reason to move the place where we declare this manager (rather than sticking with all of the other managers)? is it because you need to have access to the `dof_handler` when you initialize it?

source/particle/particle_handler.cc
```diff
@@ -222,6 +261,661 @@ namespace aspect
...
+        remove_particle(particles_out_of_cell[i]);
+
+      particles.insert(sorted_particles_map.begin(),sorted_particles_map.end());
+    }
+
+
+
+    template <int dim, int spacedim>
+    void
+    ParticleHandler<dim,spacedim>::exchange_ghost_particles()
```
bangerth: instead of reading through this function -- could you instead use the new `parallel::GridTools::exchange_...` function Timo wrote at the hackathon for what you're doing here? that would seem simpler. Since you're not yet working with deal.II dev here, you may have to copy that function from deal.II but that shouldn't be terribly difficult.

source/particle/particle_handler.cc
```diff
@@ -222,6 +261,661 @@ namespace aspect
...
+
+    template <int dim, int spacedim>
+    void
+    ParticleHandler<dim,spacedim>::send_recv_particles(const std::vector<std::vector<particle_iterator> > &send_particles,
+                                                       std::multimap<types::LevelInd,Particle <dim> >     &received_particles,
+                                                       const std::vector<std::vector<active_cell_it> >    &send_cells)
+    {
+      // Determine the communication pattern
+      const std::vector<types::subdomain_id> neighbors (triangulation->ghost_owners().begin(),
+                                                        triangulation->ghost_owners().end());
```
bangerth: that's dangerous -- it only works if `ghost_owners()` returns a reference to whatever it is storing internally. If it is returning a *copy* of the object, then taking `begin` and `end` will yield pointers that point to separate objects.

It may be safer to copy what that function returns.

source/particle/particle_handler.cc
```diff
@@ -222,6 +261,661 @@ namespace aspect
...
+            }
+          n_send_data[neighbor_id] = reinterpret_cast<std::size_t> (data) - send_offsets[neighbor_id] - reinterpret_cast<std::size_t> (&send_data.front());
+        }
+
+      // Notify other processors how many particles we will send
+      std::vector<MPI_Request> n_requests(2*n_neighbors);
+      for (unsigned int i=0; i<n_neighbors; ++i)
+        MPI_Irecv(&(n_recv_data[i]), 1, MPI_INT, neighbors[i], 0, mpi_communicator, &(n_requests[2*i]));
+      for (unsigned int i=0; i<n_neighbors; ++i)
+        MPI_Isend(&(n_send_data[i]), 1, MPI_INT, neighbors[i], 0, mpi_communicator, &(n_requests[2*i+1]));
```
bangerth: do the sends first then the receives

source/particle/particle_handler.cc
```diff
@@ -222,6 +261,661 @@ namespace aspect
...
+      MPI_Waitall(2*n_neighbors,&n_requests[0],MPI_STATUSES_IGNORE);
```
bangerth: if you enclose lines 689-695 into braces, you can let the `n_requests` variable die right here where it is used last

source/particle/particle_handler.cc
```diff
@@ -222,6 +261,661 @@ namespace aspect
...
+        if (n_recv_data[i] > 0)
+          {
+            MPI_Irecv(&(recv_data[recv_offsets[i]]), n_recv_data[i], MPI_CHAR, neighbors[i], 1, mpi_communicator,&(requests[send_ops]));
+            send_ops++;
+          }
+
+      for (unsigned int i=0; i<n_neighbors; ++i)
+        if (n_send_data[i] > 0)
+          {
+            MPI_Isend(&(send_data[send_offsets[i]]), n_send_data[i], MPI_CHAR, neighbors[i], 1, mpi_communicator,&(requests[send_ops+recv_ops]));
```
bangerth: same here -- do the sends before the receives

source/particle/particle_handler.cc
```diff
@@ -222,6 +261,661 @@ namespace aspect
...
+            recv_ops++;
+          }
+      MPI_Waitall(send_ops+recv_ops,&requests[0],MPI_STATUSES_IGNORE);
+
+      // Put the received particles into the domain if they are in the triangulation
+      const void *recv_data_it = static_cast<const void *> (&recv_data.front());
+
+      while (reinterpret_cast<std::size_t> (recv_data_it) - reinterpret_cast<std::size_t> (&recv_data.front()) < total_recv_data)
```
bangerth: could you avoid the `reinterpret_cast` by instead doing something like
```
  while (rec_data_it < static_cast<void*>(&recv_data.back()))
```
or similar?

source/particle/particle_handler.cc
```diff
@@ -222,6 +261,661 @@ namespace aspect
...
+              // If no target cells are given, use the iterator information
+              active_cell_it cell;
+              if (send_cells.size() == 0)
+                cell = send_particles[neighbor_id][i]->get_surrounding_cell(*triangulation);
+              else
+                cell = send_cells[neighbor_id][i];
+
+              const CellId::binary_type cellid = cell->id().template to_binary<dim>();
+              memcpy(data, &cellid, cellid_size);
+              data = static_cast<char *>(data) + cellid_size;
```
bangerth: it would be nicer to use the boost serialization facilities here, but that can wait for another day.

source/particle/world.cc
```diff
@@ -165,7 +112,7 @@ namespace aspect
                   ExcMessage("A part of the model has requested access to the ghost "
                              "particles, but the parameter 'Update ghost particles' has not "
                              "been set, therefore no ghost particles are available."));
-      return ghost_particles;
+      return particle_handler->get_ghost_particles();
```
bangerth: we should eventually convert the places that use these ghost particles to call something like `particle_handler.begin_ghost()` and `particle_handler.end_ghost()`, and then use iterators as well. This would allow you to hide the underlying data structure the same way you do for the locally owned particles. 

Not sure whether that's easy to do at the moment, though.

gassmoeller:
I think I addressed most comments, lets see what the tester says. If the tests pass I will rebase and squash. To your questions:

> Out of curiosity, how is the callback supposed to know what is happening at the time it is called?

I removed the confusing comment for now. At the moment the function is only called in one place anyway, and we can think about a better solution when it is used in more places.

> Here's some food for thought, not necessarily meant as a request to change things but for you to think about: In the hp::DoFHandler work, I've made these things more flexible by just letting the store functions return a std::vector<char> (could also have been a std::string) that the store function can size and fill as it pleases. I then let the place that calls all of these callbacks deal with the question of how to pack this data into buffers. That was easy in that case because it did not have to squeeze things into the p4est framework of attaching data to cells, but I found it kind of liberating to do it this way around because the store and load functions become so much simpler. It required a little bit more logic in one central place, where I call these functions, but I thought that was worth it. Could something like this be done here as well?

I am a bit worried about the speed of your suggestion. Usually this function is called once per transferred particle (= for every particle during mesh refinement), and every call to this function serializes a very small amount of data (often <= 50 bytes per particle). Copying 50 bytes is likely much faster than allocating an std::vector and copying the data. Of course one could try if it matter at all compared to the rest of the algorithm, but for now I would like to keep the interface constant.

> Are you worried about exposing this function because it shows the underlying data structure when you carefully hide the data structure for the locally owned particles?

Exactly, that is what the TODO was supposed to say. I will remove all usage of this function in follow-up PRs, and then it can be removed.

> any particular reason to move the place where we declare this manager (rather than sticking with all of the other managers)? is it because you need to have access to the dof_handler when you initialize it?

I need access to the mapping (or precisely: I need the mapping to be constructed first and destructed last so that the subscriptor pointer is correctly constructed and destructed). Alternatively I could leave the manager in its old position and move the mapping further up, would you prefer that?

> I don't recall, but do you say in the documentation that you are also initializing the particle in this function? Out of curiosity, is there a function that inserts a whole bunch of particles at the same time? Wasn't that one of the things that's important to do in bulk?

I now copy the particle properties, which is what a user would expect, but I also mention it in the documentation. 

Currently there is no function that enters many particles at once, but I do not use this function yet anyway (the particle generators still access the data structure directly). I will fix that in a follow-up PR that deals with all of our interface changes at once.

> instead of reading through this function -- could you instead use the new parallel::GridTools::exchange_... function Timo wrote at the hackathon for what you're doing here? that would seem simpler. Since you're not yet working with deal.II dev here, you may have to copy that function from deal.II but that shouldn't be terribly difficult.

I would prefer to keep that for a follow-up PR. There are several functions that can be replaced by deal.II functionality, but if I start mixing up moving functionality and replacing functionality that is simply a recipe for complete chaos.

> do the sends first then the receives

Do you know this is better, or is this just your gut feeling? All examples of Isend and Irecv I found are receives first then sends. And http://www.cas.mcmaster.ca/~nedialk/COURSES/mpi/Lectures/lec2_1.pdf claims that `A communication will generally have lower overhead if a receive buffer is already posted when a sender initiates communication`. I have not measured that though.

> could you avoid the reinterpret_cast by instead doing something like `while (rec_data_it < static_cast<void*>(&recv_data.back()))` or similar?

Maybe, but I remember that I played around with this casts already quite a bit (they changed several times over the iterations of the original merge), and I do not think this PR is the right time to try this.

> we should eventually convert the places that use these ghost particles to call something like particle_handler.begin_ghost() and particle_handler.end_ghost(), and then use iterators as well. This would allow you to hide the underlying data structure the same way you do for the locally owned particles.

I completely agree. That is something I will address in a follow-up PR.

bangerth:
>     Here's some food for thought, not necessarily meant as a request to
>     change things but for you to think about: In the hp::DoFHandler
>     work, I've made these things more flexible by just letting the store
>     functions return a std::vector (could also have been a std::string)
>     that the store function can size and fill as it pleases. I then let
>     the place that calls all of these callbacks deal with the question
>     of how to pack this data into buffers. That was easy in that case
>     because it did not have to squeeze things into the p4est framework
>     of attaching data to cells, but I found it kind of liberating to do
>     it this way around because the store and load functions become so
>     much simpler. It required a little bit more logic in one central
>     place, where I call these functions, but I thought that was worth
>     it. Could something like this be done here as well?
> 
> I am a bit worried about the speed of your suggestion. Usually this 
> function is called once per transferred particle (= for every particle 
> during mesh refinement), and every call to this function serializes a 
> very small amount of data (often <= 50 bytes per particle). Copying 50 
> bytes is likely much faster than allocating an std::vector and copying 
> the data. Of course one could try if it matter at all compared to the 
> rest of the algorithm, but for now I would like to keep the interface 
> constant.

Yes, that makes sense.


>     any particular reason to move the place where we declare this
>     manager (rather than sticking with all of the other managers)? is it
>     because you need to have access to the dof_handler when you
>     initialize it?
> 
> I need access to the mapping (or precisely: I need the mapping to be 
> constructed first and destructed last so that the subscriptor pointer is 
> correctly constructed and destructed). Alternatively I could leave the 
> manager in its old position and move the mapping further up, would you 
> prefer that?
> 
>     I don't recall, but do you say in the documentation that you are
>     also initializing the particle in this function? Out of curiosity,
>     is there a function that inserts a whole bunch of particles at the
>     same time? Wasn't that one of the things that's important to do in bulk?
> 
> I now copy the particle properties, which is what a user would expect, 
> but I also mention it in the documentation.
> 
> Currently there is no function that enters many particles at once, but I 
> do not use this function yet anyway (the particle generators still 
> access the data structure directly). I will fix that in a follow-up PR 
> that deals with all of our interface changes at once.
> 
>     instead of reading through this function -- could you instead use
>     the new parallel::GridTools::exchange_... function Timo wrote at the
>     hackathon for what you're doing here? that would seem simpler. Since
>     you're not yet working with deal.II dev here, you may have to copy
>     that function from deal.II but that shouldn't be terribly difficult.
> 
> I would prefer to keep that for a follow-up PR. There are several 
> functions that can be replaced by deal.II functionality, but if I start 
> mixing up moving functionality and replacing functionality that is 
> simply a recipe for complete chaos.

Yes, indeed!


>     do the sends first then the receives
> 
> Do you know this is better, or is this just your gut feeling? All 
> examples of Isend and Irecv I found are receives first then sends. And 
> http://www.cas.mcmaster.ca/~nedialk/COURSES/mpi/Lectures/lec2_1.pdf 
> <http://www.cas.mcmaster.ca/%7Enedialk/COURSES/mpi/Lectures/lec2_1.pdf> 
> claims that |A communication will generally have lower overhead if a 
> receive buffer is already posted when a sender initiates communication|. 
> I have not measured that though.

You're completely correct. I don't recall the context of my request, but 
your suggestion is entirely reasonable.

bangerth:
Thanks for working through this! I look forward to the next steps!

Issue #1912, opened 2017-08-25
ASPECT can't access all of solver history data from solver_control.cc in dealII?


naliboff:
I just started trying to track the origin of this issue, but I wanted to put it up quickly in case anyone is thinking updating deal.II.

After updating the master branch of deal.II yesterday all of the ASPECT tests I ran failed. The tests are failing due to an incorrect number of stokes solver steps. For example, the output from the test `quick_mpi` is

```
Number of active cells: 16 (on 3 levels)
Number of degrees of freedom: 268 (162+25+81)

*** Timestep 0:  t=0 years
   Rebuilding Stokes preconditioner...
   Solving Stokes system... 17+4294967295 iterations.
      Relative Stokes residual after nonlinear iteration 1: 1
   Solving Stokes system... 0+4294967295 iterations.
      Relative Stokes residual after nonlinear iteration 2: 6.60395e-08

   Postprocessing:

Termination requested by criterion: end time
```

The issue appears to be related to `solver_control.cc` in deal.II or how ASPECT accesses that data. The test `stokes_residual` gives the error

```
Number of active cells: 256 (on 5 levels)
Number of degrees of freedom: 3,556 (2,178+289+1,089)

*** Timestep 0:  t=0 seconds
   Solving temperature system... 0 iterations.
   Rebuilding Stokes preconditioner...
   Solving Stokes system...
--------------------------------------------------------
An error occurred in line <166> of file </data1/Software/dealii/master/dealii/source/lac/solver_control.cc> in function
    const std::vector<double>& dealii::SolverControl::get_history_data() const
The violated condition was:
    history_data.size() > 0
Additional information:
    The SolverControl object was asked for the solver history data, but there is no data. Possibly you requested the data before the solver was run.

Stacktrace:
-----------
#0  /data1/Software/dealii/master/install/lib/libdeal_II.g.so.9.0.0-pre: dealii::SolverControl::get_history_data() const
#1  /data1/Software/aspect/master/aspect/aspect: aspect::Postprocess::StokesResidual<2>::stokes_solver_callback(dealii::SolverControl const&, dealii::SolverControl const&)
[...]

PR #1914, opened 2017-08-29
Print 0 if solver_control returns an invalid unsigned integer for last_step

naliboff:
This addresses one of the issues in #1912. 

Through patch dealii/dealii/pull/4870, `SolverControl ` now initializes the number of iterations as `numbers::invalid_unsigned_int`. This broke most (all?) of the tests as the number of printed cheap or expensive iterations was incorrect.

This is fixed by printing `0` if the number of stokes iterations is `numbers::invalid_unsigned_int`.

### DIFF
--- a/source/simulator/solver.cc
+++ b/source/simulator/solver.cc
@@ -855,8 +855,14 @@ namespace aspect
         solution.block(block_p) = distributed_stokes_solution.block(block_p);
 
         // print the number of iterations to screen
-        pcout << solver_control_cheap.last_step() << '+'
-              << solver_control_expensive.last_step() << " iterations.";
+        pcout << (solver_control_cheap.last_step() != numbers::invalid_unsigned_int ?
+                  solver_control_cheap.last_step():
+                  0)
+              << '+'
+              << (solver_control_expensive.last_step() != numbers::invalid_unsigned_int ?
+                  solver_control_expensive.last_step():
+                  0)
+              << " iterations.";
         pcout << std::endl;
       }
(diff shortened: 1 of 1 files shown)


source/simulator/solver.cc
```diff
@@ -855,8 +855,21 @@ namespace aspect
         solution.block(block_p) = distributed_stokes_solution.block(block_p);
 
         // print the number of iterations to screen
-        pcout << solver_control_cheap.last_step() << '+'
-              << solver_control_expensive.last_step() << " iterations.";
+        if (solver_control_cheap.last_step() == numbers::invalid_unsigned_int)
+          {
+            pcout << "0" << '+'
```
bangerth: Instead of this sequence of `if`-`else` statements, I think you can do this more concisely as
```
pcout << (solver_control_cheap.last_step() != numbers::invalid_unsigned_int ? 
                solver_control_cheap.last_step()  : 
                0)
          << '+'
          << (solver_control_expensive.last_step() != numbers::invalid_unsigned_int ? 
                solver_control_expensive.last_step()  : 
                0)
          << " iterations.";
```
This makes it easier to understand that you're not missing a case. (You only have 3 cases for the potentially four situations; the fourth can't happen because either the cheap or the expensive solver is always run, but that requires thought to determine when reading the code.)

naliboff: Definitely cleaner, thanks :+1:

bangerth:

/run-tests

PR #1916, opened 2017-09-06
fix for porosity initial composition plugin

jdannberg:
This pull request fixes a problem that shows up if the `porosity` initial composition, the `adiabatic` initial temperature, and the `initial profile` adiabatic conditions are used together. They all need data from each other, so they called each other in a loop, which lead to a crash of ASPECT with an error message that was difficult to understand. This pull request fixes this particular problem in the way that the reference composition profile that is used to compute the adiabatic initial profile does not take into account boundary layers that might exist in the adiabatic initial temperature (so this is a very specific case, and it is only a problem if the initial composition depends on the initial temperature).


gassmoeller:
Makes sense.

PR #1918, opened 2017-09-08
Fix Morency-Doin description and formulas.

bangerth:


source/material_model/morency_doin.cc
```diff
@@ -313,33 +313,30 @@ namespace aspect
...
+                                   "An implementation of the visco-plastic rheology described by Morency "
+                                   "and Doin in \\cite{MD04}. Compositional fields can each be assigned individual "
                                    "activation energies, reference densities, thermal expansivities, and "
                                    "stress exponents. The effective viscosity is defined as "
-                                   "\\[v_{eff} = \\left(\\frac{1}{v_{eff}^v}+\\frac{1}{v_{eff}^p}\\right)^{-1}\\] "
+                                   "\\[v_\text{eff} = \\left(\\frac{1}{v_\text{eff}^v}+\\frac{1}{v_\text{eff}^p}\\right)^{-1}\\] "
                                    "where "
-                                   "\\[v_{eff}^v = B \\left(\\frac{\\dot{\\epsilon}}{\\dot{\\epsilon}_{ref}} "
+                                   "\\[v_\text{eff}^v = B \\left(\\frac{\\dot{\\epsilon}}{\\dot{\\epsilon}_\\text{ref}} "
                                    "\\right)^{-1+1/n_v} exp\\left(\\frac{E_a +V_a \\rho_m g z}{n_v R T}\\right) \\] "
```
naliboff: General question for future documentation: what is the rule of thumb for using `\` verse `\\` with `text{}` (e.g., `\text` verse `\\text`)?

bangerth: It should always be `\\text`. It's just plain wrong here. Thanks for catching this, I fixed the patch.

Issue #1917, opened 2017-09-07
how to set temperature pre-factor for the viscosity formula in 'simple' Material model


Lydia823:
In the part A.78 Parameters in section Material model of manual, there is a model 'simple' in page 314.
What is value 'x' mean in H(x) of the temperature pre-factor for the viscosity formula ? And how to set x in 'Simple model'?

jdannberg:
Dear @Lydia823, 
if you have a general question such as this one, please ask it on the ASPECT mailing list (see [this webpage](http://lists.geodynamics.org/cgi-bin/mailman/listinfo/aspect-devel)). We use the github issues to track bugs/problems in the code, so I will close this issue for now. 

To answer your question, the x is just the argument that is used to explain what the function H does, so you don't have to set it in the input file. The equation you are referring to just says that if the argument of the function H (which is the term on the left part of the page, `exp(-beta(T-T0)/T0)`) is smaller than a minimum value tau_min, or larger than a maximum value tau_max, it will be cut off at these values, and otherwise it will be used in this form to compute the temperature dependence of the viscosity. You can specify tau_min, tau_max, beta and T0 in the input file as described in the paragraph following the equation.

Lydia823:
Thanks a lot.@jdannberg

PR #1921, opened 2017-09-08
Fix simple material model formulas.

bangerth:


source/material_model/simple.cc
```diff
@@ -246,15 +246,15 @@ namespace aspect
                                    "\\begin{align}"
                                    "  \\tau(T) &= H\\left(e^{-\\beta (T-T_0)/T_0}\\right),"
                                    "  \\qquad\\qquad H(x) = \\begin{cases}"
-                                   "                            \\tau_{min} & \\text{if}\\; x<\\tau_{min}, \\\\"
+                                   "                            \\tau_\\text{min} & \\text{if}\\; x<\\tau_\\text{min}, \\\\"
                                    "                            x & \\text{if}\\; 10^{-2}\\le x \\le 10^2, \\\\"
-                                   "                            \\tau_{max} & \\text{if}\\; x>\\tau_{max}, \\\\"
+                                   "                            \\tau_\\text{max} & \\text{if}\\; x>\\tau_\\text{max}, \\\\"
                                    "                         \\end{cases}"
                                    "\\end{align} "
```
naliboff: I ran this through a latex equation editor and the formatting looks correct. Only one comment - would it be worth replacing `x` in ` H(x)` to another variable so that it is not confused with the composition prefactor `xi`?

bangerth: You mean in reference to #1917? Let's make this a follow-up to the current patch. I like an incremental approach where each patch fixes one issue.

naliboff: I had not looked at #1917 yet, but that was indeed what the issue I had in mind. This aside, looks good.

bangerth: Great.

You know that you're allowed to merge patches now, right? ;-)

bangerth: See #1923.

PR #1922, opened 2017-09-08
Fix visco plastic material model formulas and description.

bangerth:


naliboff:
Looks good to me. Thanks for cleaning the documentation up!

PR #1923, opened 2017-09-08
Improve wording of explanation in simple material model.

bangerth:
See also #1917.


jdannberg:
Thanks, I wanted to do this myself after the question on the mailing list, but then didn't get to it.

PR #1927, opened 2017-09-13
make melt and boundary traction work together

jdannberg:
This pull request allows models with melt migration to also use traction boundary conditions, which was not possible before.


tjhei:
:+1:

PR #1928, opened 2017-09-13
Change nullspace removal Assert to AssertThrow

jperryhouts:
Avoids accidentally using a direct solver with nullspace removal

### DIFF
--- a/source/simulator/parameters.cc
+++ b/source/simulator/parameters.cc
@@ -1560,6 +1560,13 @@ namespace aspect
     // plugin mechanism, declare their parameters if they have subscribed
     // to the relevant signals
     SimulatorSignals<dim>::parse_additional_parameters (*this, prm);
+
+    AssertThrow((!use_direct_stokes_solver) || (nullspace_removal == NullspaceRemoval::none),
+                ExcMessage("Because of the difference in system partitioning, nullspace removal is "
+                           "currently not compatible with the direct solver. "
+                           "Please turn off one or both of the options 'Model settings/Remove nullspace', "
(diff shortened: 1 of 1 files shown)


bangerth:
I would prefer if that remained a regular `Assert`, but that we have an `AssertThrow` somewhere earlier on (at the end of the function that reads parameters in `parameters.cc`?) that verifies that these two options can not be used together. There, you have the chance of providing a more useful error message.

jperryhouts:
How's this look?

source/simulator/parameters.cc
```diff
@@ -1560,6 +1560,13 @@ namespace aspect
     // plugin mechanism, declare their parameters if they have subscribed
     // to the relevant signals
     SimulatorSignals<dim>::parse_additional_parameters (*this, prm);
+
+    AssertThrow((!use_direct_stokes_solver) || (nullspace_removal == NullspaceRemoval::none),
+                ExcMessage("Because of the difference in system partitioning, nullspace removal is "
+                           "currently not compatible with the direct solver. "
+                           "Please turn off one or both of the options 'Model settings/Remove nullspace', "
+                           "or 'Use direct solver for Stokes system', or contribute code to enable "
+                           "this feature combination."));
```
bangerth: I like the last sentence :-)

bangerth:

/run-tests

source/simulator/parameters.cc
jperryhouts: Thanks, I wasn't sure if that would sound too sassy, but decided it was useful :)

PR #1929, opened 2017-09-20
Adjust particle interfaces

gassmoeller:
This is the next step towards #1897. I adjust all ASPECT interfaces to only use functionality that does not expose the internal data structure of `ParticleHandler`. After this PR the new particle classes should be ready to move to deal.II.

### DIFF
--- a/include/aspect/particle/particle_handler.h
+++ b/include/aspect/particle/particle_handler.h
@@ -106,6 +106,13 @@ namespace aspect
         void clear();
 
         /**
+         * Only clear particle data, but keep cache information about number
+         * of particles. This is useful during reorganization of particle data
+         * between processes.
+         */
+        void clear_particles();
+
+        /**
          * Return an iterator to the first particle.
          */
         ParticleHandler<dim,spacedim>::particle_iterator begin() const;
@@ -133,6 +140,15 @@ namespace aspect
         particle_iterator_range
         particle_range_in_cell(const typename parallel::distributed::Triangulation<dim,spacedim>::active_cell_iterator &cell);
 
+
+        /**
+         * Return a pair of particle iterators that mark the begin and end of
+         * the particles in a particular cell. The last iterator is the first
+         * particle that is no longer in the cell.
+         */
+        particle_iterator_range
+        particle_range_in_cell(const typename parallel::distributed::Triangulation<dim,spacedim>::active_cell_iterator &cell) const;
+
         /**
          * Remove a particle pointed to by the iterator.
          */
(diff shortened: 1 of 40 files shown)


include/aspect/particle/particle_handler.h
```diff
@@ -133,6 +140,15 @@ namespace aspect
...
         particle_range_in_cell(const typename parallel::distributed::Triangulation<dim,spacedim>::active_cell_iterator &cell);
 
+
+        /**
+         * Return a pair of particle iterators that mark the begin and end of
+         * the particles in a particular cell. The last iterator is the first
+         * particle that is no longer in the cell.
+         */
+        particle_iterator_range
+        particle_range_in_cell(const typename parallel::distributed::Triangulation<dim,spacedim>::active_cell_iterator &cell) const;
```
bangerth: That name caught me off guard a bit because I thought that the "range is in the cell". How about `particles_in_cell`? Or `particles_in_cell_range`?

include/aspect/particle/particle_handler.h
```diff
@@ -142,14 +158,22 @@ namespace aspect
         /**
          * Insert a particle into the collection of particles. Return an iterator
          * to the new position of the particle. This function involves a copy of
-         * the particle and its properties. Note that this function is of NlogN
+         * the particle and its properties. Note that this function is of O(NlogN)
```
bangerth: while you're there, make this `O(N \log N)`.

include/aspect/particle/particle_handler.h
```diff
@@ -142,14 +158,22 @@ namespace aspect
...
          * complexity for N particles.
          */
         particle_iterator
         insert_particle(const Particle<dim,spacedim> &particle,
                         const typename parallel::distributed::Triangulation<dim>::active_cell_iterator &cell);
 
         /**
+         * Insert a number of particle into the collection of particles.
+         * This function involves a copy of the particles and their properties.
+         * Note that this function is of O(N) complexity.
```
bangerth: What's N here? The number of existing particles? Or the number of new particles?

include/aspect/particle/particle_handler.h
```diff
@@ -250,31 +279,6 @@ namespace aspect
...
         void serialize (Archive &ar, const unsigned int version);
 
-      protected:
-        /**
-         * Access to particles in this handler.
-         * TODO: This function eventually needs to go, to not expose internal structure.
-         * This can only be done once World no longer uses this function.
-         */
-        std::multimap<types::LevelInd, Particle<dim,spacedim> > &
-        get_particles();
```
bangerth: Yay!

source/particle/particle_handler.cc
```diff
@@ -229,7 +243,11 @@ namespace aspect
     ParticleHandler<dim,spacedim>::n_particles_in_cell(const typename Triangulation<dim,spacedim>::active_cell_iterator &cell) const
     {
       const types::LevelInd found_cell = std::make_pair<int, int> (cell->level(),cell->index());
-      return particles.count(found_cell);
+
+      if (!cell->is_ghost())
+        return particles.count(found_cell);
+      else
+        return ghost_particles.count(found_cell);
```
bangerth: I think you ought to check that the cell is not artificial.

Issue #1924, opened 2017-09-09
update convection box cookbook


jdannberg:
The parameters in the input file of the convection box cookbook and its description in the manual are not the same. For example, for the gravity model
    `set Magnitude = 1e4`  in the .prm file and
    `set Magnitude = 1e14` in the manual description (same for the thermal expansion coefficient). 
We may have updated the input file after implementing the Boussinesq formulation, but we should make it consistent with the documentation.

jdannberg:
Fixed by #1930.

PR #1930, opened 2017-09-21
update convection box cookbook

jdannberg:
This pull request updates the text of the convection box cookbook in the manual so that it fits what we do in the input file. Before, the 'playtime' sections in the manual didn't work any more, because they included wrong values for the Gravity that made the cookbook crash. In addition, the cookbook now correctly uses the Boussinesq formulation (which it should have done before already, as it has a value of 1 instead of 1e-10 for the thermal expansivity). So this should really be considered a bugfix. It addresses #1924.

### DIFF
--- a/cookbooks/convection-box.prm
+++ b/cookbooks/convection-box.prm
@@ -108,13 +108,13 @@ subsection Gravity model
   set Model name = vertical
 
   subsection Vertical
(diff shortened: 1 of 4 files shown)


gassmoeller:
Many thanks for updating the cookbook :+1:. A boring task that still needs to be done every now and then.

bangerth:
:+1:

Issue #1931, opened 2017-09-22
prescribed velocity tests use copies of cookbook plugins instead of including them


gassmoeller:
this should be changed so that the tests always test the current version of the cookbook plugins.

jdannberg:
This was fixed by #1932.

PR #1932, opened 2017-09-22
include plugin code in tests instead of copying it

jdannberg:
This addresses #1931, and also fixes the issue for the king and tangurnis tests.


bangerth:
Thanks to @gassmoeller for noticing this, and @jdannberg for fixing it!

PR #1933, opened 2017-09-22
Fix for deal 90 pre

gassmoeller:
I did not test #1929 for deal.II 9.0.pre. Turns out some code that is only included for 9.0pre does not compile :worried:. This PR fixes that.

I would appreciate if somebody could soon merge this, before too many people notice :smile:.


gassmoeller:
Thanks!

PR #1771, opened 2017-05-24
Boundary condition map

gassmoeller:
See changelog entry. This change allows the combination of plugins for boundary temperatures that is already possible for initial temperatures (I wish I had had this functionality for several of my papers). After review, I will apply the update_prm script on all included .prm files. For now I keep it out, because it makes the same one-line change in 500 files, which would make the review very tedious.

I plan to do a similar change for boundary_velocity and boundary_composition as well, but I would like to make sure I get the format right, before having to make the same edits to all of them. Boundary_fluid_pressure and Boundary_traction do not really have enough plugins as that a combination would make sense, but in some future we can think of that as well. Luckily the change is backwards compatible, as long as we keep the old single plugin input parameter around.

PR #1937, opened 2017-09-26
Fix some warnings and a bug

gassmoeller:
This PR fixes some compiler and deprecation warnings that were introduced in #1882. These fixes reveal a bug in `BoundaryTemperature::Manager::find_boundary_temperature_model` that was introduced by me in #1771. The bug was never discovered, because it was in a template function that was not instantiated until now. Fixed now.


bangerth:
Yes, good work -- thanks!

PR #1938, opened 2017-09-27
optimalisation of compute current constraints in the Newton solver.

MFraters:
I noticed that I forgot to put this optimalisation into the newton solver. This saves a lot of runtime on bigger models on many mpi processes.

In contrast with the Picard type solver, which only needs to compute the constraints once, in the Newton solver, the first iteration and second iteration have different values because of the velocity boundary conditions. After the second iteration they stay the same, so we don't need to recompute.


gassmoeller:
Ready to merge if the tester is happy.

PR #1939, opened 2017-09-28
Fix Newton SPD debug test.

MFraters:
I was running some more tests on the Newton solver and noticed that the problem with the SPD debug test was still alive. After re-reading MFraters#15, I might have misread what Rene said about velocity and pressure. Since the code here has 22 indexes it actually includes the pressure block, which shouldn't happen and this explains why the SPD test fails. 

The problem I have is that I have a hard time figuring out how to do this the correct way in deal.ii. When I hard-code that (for 2d) only the first eighteen values should be filled with random numbers and that the rest should be zero the SPD test doesn't fail anymore. But now I need to find a way to do this correctly.

Futhermore, I do not really understand why we need the line with `if (scratch.dof_component_indices[i] < dim)`, since when I look at the values in there, they are for all i zero.


gassmoeller:
> Futhermore, I do not really understand why we need the line with if (scratch.dof_component_indices[i] < dim), since when I look at the values in there, they are for all i zero.

The last part would be the real problem then. `scratch.dof_component_indices` as it is filled in line 159 should contain the component (i.e. which of the velocity component 0,1,[2], or the pressure,3),  for every degree of freedom. That is why `if (scratch.dof_component_indices[i] < dim)` should only fill the entries in the test vector that correspond to velocity components. I am not sure what goes wrong here.

MFraters:
That make sense. So I think that I have found the problem. the `#if debug` should be in capital.

gassmoeller:
Good point, yes that is the problem.

MFraters:
Yes, this seem to fix the problem this problem. But doing some tests I noticed that some results for the parameter search I did half august for the Spiegelman benchmark are now considerably worse and do not follow the pattern I found then. Luckily I found an old development branch where it still gives me the same results as before, so I have to see if I can spot the difference.

MFraters:
I have been able to reproduce the results I had before by making the matrix symmetric again (not just the preconditioner). I remembered having played with this a while ago, but apparently it ended up without symmetrisation in the pull request. Seeing now that it can make a difference in difficult cases, I guess this is the right way to do it. I don't really understand why the SPD tester wasn't triggered, because if symmetrisation  makes a difference, it was apparently not symmetric before.

I found an other thing, namely in the preconditioner, I think that because we are now adding the derivatives, we need `2.0 * eta`, instead of just `eta`. What do you think?

gassmoeller:
> I don't really understand why the SPD tester wasn't triggered, because if symmetrisation makes a difference, it was apparently not symmetric before.

I am not sure either, but if the new form gives better results, lets keep it.

> I found an other thing, namely in the preconditioner, I think that because we are now adding the derivatives, we need 2.0 * eta, instead of just eta. What do you think?

Can you explain why you think you need the 2.0? Is this a difference of the newton matrix compared to the normal one? Probably @bangerth is more qualified to answer this.

MFraters:
> Can you explain why you think you need the 2.0? Is this a difference of the newton matrix compared to the normal one? 

Yes and no. As I understand it, the equation has always been 2.0 * eta. But since in the preconditioner of the stokes matrix the 2.0 is not relevant, because the 2.0 over the whole A submatrix is just multiplied and divided later, it is omitted. But for the newton A submatrix, the exact size of the matters because we subtract something from the 2.0*eta, so I think the 2.0 can't be omitted any more.

I hope I explained this correctly, and I am very curious what @bangerth and @tjhei think about this :)

tjhei:
> As I understand it, the equation has always been 2.0 * eta. But since in the preconditioner of the stokes matrix the 2.0 is not relevant, because the 2.0 over the whole A submatrix is just multiplied and divided later, it is omitted.

I was assuming that we would assemble the correct velocity block with 2*eta and I am surprised that we don't do so. We state in the aspect1 paper that we do assemble with 2*eta. @bangerth ??

MFraters:
So what do we do with this pull request? I would really like this pull request to be merged in some form, because it contains some important fixes and then I can make the next pull request for the Newton solver with functionality improvements.

gassmoeller:
I will try to poke @bangerth this afternoon about the factor 2.0 in the preconditioner :smile: ... but the other issues in this PR are essentially unrelated, right? I am happy to merge the fix to the SPD tester, and the symmetrisation of the matrix (if that gives you better results) in this PR and you could move the preconditioner issue to a separate PR? I am curious about which of the changes in the iterations numbers in your tests are related to the changes to the matrix and which are caused by the preconditioner anyway.

MFraters:
haha :)

If I remove the 2.0 factor from the preconditioner, the two tests trigger the SPD test (and therefore it fails) after a few iterations.

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -236,7 +295,8 @@ namespace aspect
                   for (unsigned int i=0; i<stokes_dofs_per_cell; ++i)
                     for (unsigned int j=0; j<stokes_dofs_per_cell; ++j)
                       {
-                        data.local_matrix(i,j) += ( derivative_scaling_factor * alpha * 2.0 * (scratch.grads_phi_u[i] * (viscosity_derivative_wrt_strain_rate * scratch.grads_phi_u[j]) * strain_rate)
+                        data.local_matrix(i,j) += ( derivative_scaling_factor * alpha * (scratch.grads_phi_u[i] * (viscosity_derivative_wrt_strain_rate * scratch.grads_phi_u[j]) * strain_rate
```
bangerth: why do you get rid of the factor of 2 here? Don't we need it?

tests/nonlinear_channel_flow_tractions_Newton_Stokes/screen-output
```diff
@@ -9,154 +11,199 @@ Number of degrees of freedom: 3,556 (2,178+289+1,089)
...
-      Total relative residual after nonlinear iteration 15: 0.00024161, norm of the rhs: 6.291e+07, newton_derivative_scaling_factor: 1
+   Solving Stokes system... 40+0 iterations.
+      Total relative residual after nonlinear iteration 15: 0.00207423, norm of the rhs: 2.76824e+09, newton_derivative_scaling_factor: 1
 
    Skipping temperature solve because RHS is zero.
    The linear solver tolerance is set to 0.1
    Rebuilding Stokes preconditioner...
-   Solving Stokes system... 147+0 iterations.
-      Total relative residual after nonlinear iteration 16: 3.39751e-05, norm of the rhs: 1.89797e+07, newton_derivative_scaling_factor: 1
+   Solving Stokes system... 36+0 iterations.
```
bangerth: this is fantastic?

bangerth: not `?`, but `!` !!

bangerth:
Ah, actually I understand now -- you do the symmetrization, and then you don't need the factor of 2 any more. So this looks alright to me -- thanks for working this out!

PR #1940, opened 2017-09-28
Fix for the Newton switch tolerance

MFraters:
Fix for the Newton switch tolerance, which wasn't working properly before.


bangerth:

/run-tests

PR #1941, opened 2017-09-28
Add support for natural coordinates to geometry models

MFraters:
This implements the cartesian_to_natural_coordinates and natural_to_cartesian_coordinates functions for the box, two_merged_boxes, chunk and ellipsoidal chunk geometry models. These where the easy ones since they where or just the original input, or I could use their manifold of the geometry model itself. 

I will add a test for these functions later.

### DIFF
--- a/include/aspect/geometry_model/interface.h
+++ b/include/aspect/geometry_model/interface.h
@@ -26,6 +26,8 @@
 #include <deal.II/base/parameter_handler.h>
 #include <deal.II/distributed/tria.h>
 #include <deal.II/base/std_cxx11/array.h>
+#include <aspect/utilities.h>
+#include <aspect/coordinate_systems.h>
 
 #include <set>
 
@@ -141,6 +143,13 @@ namespace aspect
         */
         virtual
         double height_above_reference_surface(const Point<dim> &position) const = 0;
+
+
+        /**
+         * Returns what the natural coordinate system for this geometry model is.
+         */
+        virtual
+        aspect::Utilities::Coordinates::CoordinateSystem natural_coordinate_system() const = 0;
 
         /**
          * Takes the Cartesian points (x,z or x,y,z) and returns standardized
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -33,12 +33,18 @@
 #include <deal.II/fe/component_mask.h>
 
 #include <aspect/geometry_model/interface.h>
+#include <aspect/coordinate_systems.h>
 
 
 
 namespace aspect
 {
   template <int dim> class SimulatorAccess;
+
+  namespace GeometryModel
+  {
+    template <int dim> class Interface;
+  }
 
   /**
    * A namespace for utility functions that might be used in many different
@@ -160,24 +166,6 @@ namespace aspect
       ellipsoidal_to_cartesian_coordinates(const std_cxx11::array<double,3> &phi_theta_d,
                                            const double semi_major_axis_a,
                                            const double eccentricity);
-
-      /**
-       * This enum lists available coordinate systems that can be used for
-       * the function variables. Allowed values are 'cartesian',
-       * 'spherical', and 'depth'. 'spherical' coordinates follow: r, phi
-       * (2D) or r, phi, theta (3D); where r is radius, phi is longitude,
-       * and theta is the polar angle (colatitude). The 'depth' is a
-       * one-dimensional coordinate system in which only the distance
-       * below the 'top' surface (depth) as defined by each geometry model,
-       * is used.
-       */
-      enum CoordinateSystem
-      {
-        depth,
-        cartesian,
-        spherical,
-        invalid
-      };
 
 
       /**
@@ -1069,6 +1057,45 @@ namespace aspect
     template <int dim>
     SymmetricTensor<2,dim> nth_basis_for_symmetric_tensors (const unsigned int k);
 
+    /*
+    * A class that represents a point in a chosen coordinate system.
+    */
+    template <int dim>
+    class NaturalCoordinate
+    {
+      public:
+        /**
+         * Constructor based on providing the geometry model as a pointer
+         */
+        NaturalCoordinate(Point<dim> &position,
+                          const GeometryModel::Interface<dim> &geometry_model);
+
+        /**
+        * Returns the coordinates in the given coordinate system, which may not be Cartesian.
+        */
+        std_cxx11::array<double,dim> &get_coordinates();
+
+        /**
+        * The coordinate that represents the 'surface' directions in the chosen coordinate system.
+        **/
+        std_cxx11::array<double,dim-1> get_surface_coordinates() const;
+
+        /**
+        * The coordinate that represents the 'depth' direction in the chosen coordinate system.
+        **/
+        double get_depth_coordinate() const;
+
+      private:
+        /**
+         * An enum which stores the the coordinate system of this natural point
+         */
+        Utilities::Coordinates::CoordinateSystem coordinate_system;
+
+        /**
+         * An array which stores the coordinates in the coordinates system
+         */
+        std::array<double,dim> coordinates;
+    };
   }
 }
 
--- a/source/geometry_model/ellipsoidal_chunk.cc
+++ b/source/geometry_model/ellipsoidal_chunk.cc
@@ -50,7 +50,8 @@ namespace aspect
       semi_major_axis_a (-1),
       eccentricity (-1),
       semi_minor_axis_b (-1),
-      bottom_depth (-1)
+      bottom_depth (-1),
+      topography(NULL)
     {}
 
     // Copy constructor
@@ -103,7 +104,6 @@ namespace aspect
     EllipsoidalChunk<dim>::EllipsoidalChunkGeometry::pull_back_ellipsoid(const Point<3> &x, const double semi_major_axis_a, const double eccentricity) const
     {
       AssertThrow (dim == 3,ExcMessage ("This can currently only be used in 3d."));
-
       const double R    = semi_major_axis_a;
       const double b      = std::sqrt(R * R * (1 - eccentricity * eccentricity));
       const double ep     = std::sqrt((R * R - b * b) / (b * b));
@@ -128,13 +128,12 @@ namespace aspect
     EllipsoidalChunk<dim>::EllipsoidalChunkGeometry::push_forward_topography(const Point<3> &phi_theta_d_hat) const
     {
       AssertThrow (dim == 3,ExcMessage ("This can currently only be used in 3d."));
-
       const double d_hat = phi_theta_d_hat[2]; // long, lat, depth
       Point<dim-1> phi_theta;
       const double rad_to_degree = 180/numbers::PI;
       if (dim == 3)
         phi_theta = Point<dim-1>(phi_theta_d_hat[0] * rad_to_degree,phi_theta_d_hat[1] * rad_to_degree);
-      const double h = topography->value(phi_theta);
+      const double h = topography != NULL ? topography->value(phi_theta) : 0;
       const double d = d_hat + (d_hat + bottom_depth)/bottom_depth*h;
       const Point<3> phi_theta_d (phi_theta_d_hat[0],
                                   phi_theta_d_hat[1],
@@ -147,13 +146,12 @@ namespace aspect
     EllipsoidalChunk<dim>::EllipsoidalChunkGeometry::pull_back_topography(const Point<3> &phi_theta_d) const
     {
       AssertThrow (dim == 3,ExcMessage ("This can currently only be used in 3d."));
-
       const double d = phi_theta_d[2];
       const double rad_to_degree = 180/numbers::PI;
       Point<dim-1> phi_theta;
       if (dim == 3)
         phi_theta = Point<dim-1>(phi_theta_d[0] * rad_to_degree,phi_theta_d[1] * rad_to_degree);
-      const double h = topography->value(phi_theta);
+      const double h = topography != NULL ? topography->value(phi_theta) : 0;
       const double d_hat = bottom_depth * (d-h)/(bottom_depth+h);
       const Point<3> phi_theta_d_hat (phi_theta_d[0],
                                       phi_theta_d[1],
@@ -733,6 +731,57 @@ namespace aspect
 
       return true;
     }
+
+    template <int dim>
+    std_cxx11::array<double,dim>
+    EllipsoidalChunk<dim>::cartesian_to_natural_coordinates(const Point<dim> &position_point) const
+    {
+      Assert(dim == 3,ExcMessage("This geometry model doesn't support 2d."));
+      // the chunk manifold works internally with a vector with longitude, latitude, depth.
+      // We need to output radius, longitude, latitude to be consistent.
+
+      Point<dim> transformed_point = manifold.pull_back(position_point);
+
+      const double radius = get_radius(position_point);
+      std::array<double,dim> position_array;
+      position_array[0] = radius + transformed_point(2);
+      position_array[1] = transformed_point(1);
+      position_array[2] = transformed_point(0);
+
+      return position_array;
+    }
+
+
+    template <int dim>
+    aspect::Utilities::Coordinates::CoordinateSystem
+    EllipsoidalChunk<dim>::natural_coordinate_system() const
+    {
+      return aspect::Utilities::Coordinates::CoordinateSystem::ellipsoidal;
+    }
+
+
+
+    template <int dim>
+    Point<dim>
+    EllipsoidalChunk<dim>::natural_to_cartesian_coordinates(const std_cxx11::array<double,dim> &position_tensor) const
+    {
+      Assert(dim == 3,ExcMessage("This geometry model doesn't support 2d."));
+
+      // We receive radius, longitude, latitude and we need to turn it first back into
+      // longitude, latitude, depth for internal use, and push_forward to cartesian coordiantes.
+      Point<3> position_point;
+      position_point(0) = position_tensor[2];
+      position_point(1) = position_tensor[1];
+
+      const double radius = semi_major_axis_a / (std::sqrt(1 - eccentricity * eccentricity * std::sin(position_point(1)) * std::sin(position_point(1))));
+      position_point(2) = position_tensor[0] - radius;
+
+      Point<3> transformed_point = manifold.push_forward(position_point);
+      Point<3> double_transformed_point = manifold.pull_back(transformed_point);
+
+      return reinterpret_cast<Point<dim>&>(transformed_point);
+    }
+
   }
 }
 
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -2532,6 +2532,7 @@ namespace aspect
     }
 
 
+
     template <int dim>
     SymmetricTensor<2,dim>
     nth_basis_for_symmetric_tensors (const unsigned int k)
@@ -2545,6 +2546,98 @@ namespace aspect
       result[indices_ij] = 1;
 
       return symmetrize(result);
+    }
+
+    template <int dim>
+    NaturalCoordinate<dim>::NaturalCoordinate(Point<dim> &position,
+                                              const GeometryModel::Interface<dim> &geometry_model)
+    {
+      coordinate_system = geometry_model.natural_coordinate_system();
+      coordinates = geometry_model.cartesian_to_natural_coordinates(position);
+    }
+
+    template <int dim>
+    std_cxx11::array<double,dim> &NaturalCoordinate<dim>::get_coordinates()
+    {
+      return coordinates;
+    }
+
+    template <>
+    std_cxx11::array<double,1> NaturalCoordinate<2>::get_surface_coordinates() const
+    {
+      std_cxx11::array<double,1> coordinate;
+
+      switch (coordinate_system)
+        {
+          case Coordinates::CoordinateSystem::cartesian:
+            coordinate[0] = coordinates[0];
+            break;
+
+          case Coordinates::CoordinateSystem::spherical:
+            coordinate[0] = coordinates[1];
+            break;
+
+          case Coordinates::CoordinateSystem::ellipsoidal:
+            coordinate[0] = coordinates[1];
+            break;
+
+          default:
+            coordinate[0] = 0;
+            Assert (false, ExcNotImplemented());
+            break;
+        }
+
+      return coordinate;
+    }
+
+    template <>
+    std_cxx11::array<double,2> NaturalCoordinate<3>::get_surface_coordinates() const
+    {
+      std_cxx11::array<double,2> coordinate;
+
+      switch (coordinate_system)
+        {
+          case Coordinates::CoordinateSystem::cartesian:
+            coordinate[0] = coordinates[0];
+            coordinate[1] = coordinates[1];
+            break;
+
+          case Coordinates::CoordinateSystem::spherical:
+            coordinate[0] = coordinates[1];
+            coordinate[1] = coordinates[2];
+            break;
+
+          case Coordinates::CoordinateSystem::ellipsoidal:
+            coordinate[0] = coordinates[1];
+            coordinate[1] = coordinates[2];
+            break;
+
+          default:
+            Assert (false, ExcNotImplemented());
+        }
+
+      return coordinate;
+    }
+
+    template <int dim>
+    double NaturalCoordinate<dim>::get_depth_coordinate() const
+    {
+      switch (coordinate_system)
+        {
+          case Coordinates::CoordinateSystem::cartesian:
+            return coordinates[dim-1];
+
+          case Coordinates::CoordinateSystem::spherical:
+            return coordinates[0];
+
+          case Coordinates::CoordinateSystem::ellipsoidal:
+            return coordinates[0];
+
+          default:
+            Assert (false, ExcNotImplemented());
+        }
+
+      return 0;
     }
 
 
@@ -2635,5 +2728,7 @@ namespace aspect
     template SymmetricTensor<2,2> nth_basis_for_symmetric_tensors (const unsigned int k);
     template SymmetricTensor<2,3> nth_basis_for_symmetric_tensors (const unsigned int k);
 
+    template class NaturalCoordinate<2>;
+    template class NaturalCoordinate<3>;
   }
 }
--- a/include/aspect/coordinate_systems.h
+++ b/include/aspect/coordinate_systems.h
@@ -1,0 +1,53 @@
+/*
+  Copyright (C) 2018 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _aspect_coordinate_sytems_h
+#define _aspect_coordinate_sytems_h
+
+namespace aspect
+{
+  namespace Utilities
+  {
+    namespace Coordinates
+    {
+      /**
+       * This enum lists available coordinate systems that can be used for
+       * the function variables. Allowed values are 'cartesian',
+       * 'spherical', and 'depth'. 'spherical' coordinates follow: r, phi
+       * (2D) or r, phi, theta (3D); where r is radius, phi is longitude,
+       * and theta is the polar angle (colatitude). The 'depth' is a
+       * one-dimensional coordinate system in which only the distance
+       * below the 'top' surface (depth) as defined by each geometry model,
+       * is used.
+       */
+      enum CoordinateSystem
+      {
+        depth,
+        cartesian,
+        spherical,
+        ellipsoidal,
+        invalid
+      };
+    }
+  }
+}
+
+#endif
+
--- a/include/aspect/geometry_model/box.h
+++ b/include/aspect/geometry_model/box.h
@@ -176,6 +176,29 @@ namespace aspect
         bool
         point_is_in_domain(const Point<dim> &p) const;
 
+        /*
+         * Returns what the natural coordinate system for this geometry model is,
+         * which for a box is Cartesian.
+         */
+        virtual
+        aspect::Utilities::Coordinates::CoordinateSystem natural_coordinate_system() const;
+
+        /**
+         * Takes the Cartesian points (x,z or x,y,z) and returns standardized
+         * coordinates which are most 'natural' to the geometry model. For a box
+         * the results is unchanged and is (x,z) in 2d or (x,y,z) in 3d.
+         */
+        virtual
+        std_cxx11::array<double,dim> cartesian_to_natural_coordinates(const Point<dim> &position) const;
+
+        /**
+         * Undoes the action of cartesian_to_natural_coordinates, and turns the
+         * coordinate system which is most 'natural' to the geometry model into
+         * Cartesian coordinates.
+         */
+        virtual
+        Point<dim> natural_to_cartesian_coordinates(const std_cxx11::array<double,dim> &position) const;
+
         /**
          * Declare the parameters this class takes through input files.
          */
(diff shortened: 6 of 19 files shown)


source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -706,6 +706,49 @@ namespace aspect
...
+    {
+      Assert(dim == 3,ExcMessage("This geometry model doesn't support 2d."));
+
+      Point<3> position_point;
+      position_point(0) = position_tensor[2];
+      position_point(1) = position_tensor[1];
+
+      const double radius = get_radius(reinterpret_cast<Point<dim>&>(position_point));
+
+      position_point(2) = radius - position_tensor[2];
```
gassmoeller: should this be `position_tensor[0]`? Above you have:
```
position_array[2] = transformed_point(0); 
```

source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -706,6 +706,49 @@ namespace aspect
...
+
+      Point<3> transformed_point;
+
+      transformed_point = manifold.push_forward(position_point);
```
gassmoeller: you can combine lines 745 and 747 into:
```
Point<3> transformed_point = manifold.push_forward(position_point);
```

MFraters:
No, they are not used yet. I am doubting what kind of testing I should do. 

There are two things I could do, first I could just check whether the two functions actually create the inverse of each other, and secondly I could come up with some setting for to visually test the results, but then I need to write for example some initial temperature plugin which uses this.

source/geometry_model/ellipsoidal_chunk.cc
MFraters: you are right, that does make more sense :)

gassmoeller:
I guess you could extent one of the function plugins for initial or boundary conditions with a `natural` coordinate system option and test that (similar to #1943). Or what is it you intend to do with these functions? Could you test that already?

MFraters:
Yes, these functions are intended to be used to make initial conditions. 

I have now added a test to at least check internal consistency, which turned out to be good to test as discussed in #1950. I could add for every geometry model a test where these functions are actually used to create initial conditions and then check visually if they are located at the expected location, but I do not think this would in practice add any new information compared to the current test.

So currently this pull request is dependent on how #1950 is resolved, I could put the ellipsoidal chunk in a different pull request, so at least the functions will be available for the other geometry models.

MFraters:
I have changed the interface a bit so that the depth/radius coordinate is always the last coordinate, like is the case with the Cartesian box. I don't think there was a particular reason to do it the way it was done except that it was the standard notation of spherical coordinates. Or am I forgetting something? @anne-glerum

anne-glerum:
r,phi,theta is consistent with the cartesian_to_spherical_coordinates function in Utilities. This order allows for simple loops over dim for both 2d and 3.

MFraters:
Yes, but it is not consistent with the ordering of the coordinates of the Box type geometry models. The main point of this function for me is to abstract away as many differences between the geometry models as possible. I also think that the loops should not become much more complex by specifying that the depth should always be at the end instead of the beginning.

gassmoeller:
I agree with @anne-glerum, I would also prefer keeping the order that already exists. The reason the order is currently `r,phi,theta` is that it is closest to the *conventional* spherical coordinates (r,theta,phi), while allowing to loop over the coordinates in a dimension independent way (for (i=0; i<dim;++i); coordinates[i] = ...)`. 
Can you give an example where this ordering would be a problem for your intended application? I feel like I do not quite understand that yet.

MFraters:
an example is that I now know that the surface is always in the first to variables, so I can write:

```
internal_position = Utilities::convert_array_to_point<dim>(geometry_model->cartesian_to_natural_coordinates(position));
Point<2> internal_position_surface(internal_position[0], internal_position[1]);
```
in stead of
```
spherical_coords = dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0 || 
                   dynamic_cast<const GeometryModel::TwoMergedBoxes<dim>*> (&this->get_geometry_model()) == 0 ? false : true;
internal_position = Utilities::convert_array_to_point<dim>(geometry_model->cartesian_to_natural_coordinates(position));
Point<2> internal_position_surface = spherical_coords ? Point<2>(internal_position[1], internal_position[2]) : Point<2>(internal_position[0], internal_position[1]);
```

It is something I could live with, but I think it makes code just easier to write and understand, because you don't have to make different cases for the geometry models.

bangerth:
Btw, the condition
```
spherical_coords 
= dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0 || 
   dynamic_cast<const GeometryModel::TwoMergedBoxes<dim>*> (&this->get_geometry_model()) == 0 
  ? 
  false : true;
```
does not really make sense to me. It returns `true` only if `not (geometry is not a box || geometry is not two boxes)`, which is equivalent to `geometry is a box && geometry is two boxes`, which I think cannot ever be true unless one of the two geometry models is derived from the other. (And if the latter is the case, then it's enough to test whether the model is of the derived kind.)

bangerth:
As far as the design is concerned -- it's too late for this, but questions such as "which of the natural coordinates correspond to depth and surface" (if any, actually) suggest that expressing "natural" coordinates as just an array of numbers is dubious. Programming languages like to use *structured types* not only because they allow to group and store data that belongs together, but also because they denote *semantic meaning*. For example, `Point<dim>` is defined as representing the coordinates of a point in a Cartesian coordinate system. An array `double[dim]` or `std::array<double,3>` does not have this connotation and can hold any kind of data. We would probably have been better off defining a type of the form `SphericalPointCoordinates` (or data types for each of the geometry models that need to describe their "natural coordinates") that is defined to represent depth or surface in specific components. It may even have a `depth()` function, for example.

It's too late to do this now. I suspect that we're going to run into the kinds of problems you mention above in a number of places in the future. And we're going to figure out in each case how to deal with it, if necessary through documentation. I'm not particularly worried about it, but I did want to bring up the issue as a learning opportunities where we can all learn a lesson on what works and what doesn't.

bangerth:
All that said, the thread above isn't quite clear where we stand with this PR. Is it ready for another review, or are there outstanding questions?

MFraters:
>Btw, the condition
>
>spherical_coords 
= dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0 || 
   dynamic_cast<const GeometryModel::TwoMergedBoxes<dim>*> (&this->get_geometry_model()) == 0 
  ? 
  false : true;
>
>does not really make sense to me. It returns true only if not (geometry is not a box || geometry is not two boxes), which is equivalent to geometry is a box && geometry is two boxes, which I think cannot ever be true unless one of the two geometry models is derived from the other. (And if the latter is the case, then it's enough to test whether the model is of the derived kind.)

You are correct, this should have been:

```
spherical_coords = dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0 &&
    	           dynamic_cast<const GeometryModel::TwoMergedBoxes<dim>*> (&this->get_geometry_model()) == 0 ? true : false;
```

MFraters:
>All that said, the thread above isn't quite clear where we stand with this PR. Is it ready for another review, or are there outstanding questions?

I think that the only outstanding question is if we want to use the same coordinate order as spherical, or if we want the same coordinate order as the box type geometry models. @anne-glerum and @gassmoeller seem to be in favour of the spherical order, but @gassmoeller asked me for an example of why the the other order can be advantages. I have given this above and am curious whether this might convince them or that they still think that the spherical order should be used for this function. I didn't get your opinion on this matter from what you have written, could you clarify that?

As noted above I have a preference for the unification of the order between geometry models, but if others have a believe that it is better to use the spherical order for this I will change it to that. 

As soon as this discussion is resolved, I will make the required changes and I am thinking of adding some tests actually using these functions to create initial conditions.

bangerth:
As for the condition: You can write
```
spherical_coords = dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0 &&
    	           dynamic_cast<const GeometryModel::TwoMergedBoxes<dim>*> (&this->get_geometry_model()) == 0 
  ? true : false;
```
equally as
```
spherical_coords = dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0 &&
    	           dynamic_cast<const GeometryModel::TwoMergedBoxes<dim>*> (&this->get_geometry_model()) == 0;
```
It reads just as well: `spherical_coords` equals "(something equals zero) and (something else equals zero)" :-)

bangerth:
As for the content, I'll gladly defer to @anne-glerum and @gassmoeller.

gassmoeller:
> internal_position = Utilities::convert_array_to_point<dim>(geometry_model->cartesian_to_natural_coordinates(position));
Point<2> internal_position_surface(internal_position[0], internal_position[1]);

> in stead of

> spherical_coords = dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0 || 
                   dynamic_cast<const GeometryModel::TwoMergedBoxes<dim>*> (&this->get_geometry_model()) == 0 ? false : true;
internal_position = Utilities::convert_array_to_point<dim>(geometry_model->cartesian_to_natural_coordinates(position));
Point<2> internal_position_surface = spherical_coords ? Point<2>(internal_position[1], internal_position[2]) : Point<2>(internal_position[0], internal_position[1]);


I see your point now. I still have the feeling that 'arbitrarily' reordering coordinates is a dangerous path. I know it is not arbitrarily, but it is done for your particular purpose right now, and we ordered it for our particular purpose earlier ... what if somebody else needs another ordering?. Would it be better, if we had the following functions and classes available?

```
Utilities::Coordinates::CoordinateSystem 
GeometryModel::Interface::natural_coordinate_system() const = 0;

namespace Utilities
{
/**
* A class that represents a point in a chosen coordinate system.
*/
template <int dim>
class GeneralizedPoint
{
public:
GeneralizedPoint(Point<dim> position, Utilities::Coordinates::CoordinateSystem coordinate_system);

/**
* Returns the cartesian position of this point.
*/
Point<dim>& position();

/**
* Returns the coordinates in the chosen coordinate system
*/
std::array<double,dim>& coordinates();

/**
* The coordinate that represents the 'surface' directions in the chosen coordinate system.
**/ 
std::array<double,dim-1> surface_coordinates() const;

/**
* The coordinate that represents the 'depth' direction in the chosen coordinate system.
**/ 
double depth_coordinate() const;

private:
Point<dim> position;
std::array<double,dim> coordinates;
}
}
```

The first function could replace a lot of `dynamic_cast` checks, which geometry model we are in, and the `GeneralizedPoint` class could shorten your example to:

```
GeneralizedPoint<dim> natural_position (position, this->get_geometry_model().natural_coordinate_system())
std::array<double,dim-1> surface_coordinates = natural_position.surface_coordinates();
```

Would this work for geometry models with complicated manifolds, too? You have more experience in that than me.

bangerth:
I like @gassmoeller 's suggestion. Would all of the magic happen in the constructor that gets a point and then needs to convert it to one of the chosen coordinate systems to store in the `coordinates` variable? Since that variable is internal, I assume that you can order the components of that variable in such a way that, for example, the depth is always at a particular index, right?

MFraters:
I like this as well. I will see if I can implement it.

MFraters:
I have implemented the GeneralizedPoint idea, and the code is ready for a new review round :)

include/aspect/utilities.h
```diff
@@ -1060,6 +1066,41 @@ namespace aspect
...
+    template <int dim>
+    class GeneralizedPoint
+    {
+      public:
+        GeneralizedPoint(Point<dim> position, const GeometryModel::Interface<dim> *geometry_model);
+
+        /**
+        * Returns the cartesian position of this point.
+        */
+        Point<dim> &get_position();
```
tjhei: Is this very useful to have? If you decide not to store the cartesian position and only use it to construct this instance, it would only contain the natural coordinates. Then we could call the class ``NaturalCoordinate`` or something and actually return it from the geometry models.

include/aspect/utilities.h
```diff
@@ -1060,6 +1066,41 @@ namespace aspect
...
     std::vector<Operator> create_model_operator_list(const std::vector<std::string> &operator_names);
+
+    /**
+    * A class that represents a point in a chosen coordinate system.
+    */
+    template <int dim>
+    class GeneralizedPoint
+    {
+      public:
+        GeneralizedPoint(Point<dim> position, const GeometryModel::Interface<dim> *geometry_model);
```
tjhei: pass position as ``const &``

include/aspect/utilities.h
```diff
@@ -1060,6 +1066,41 @@ namespace aspect
      * entry in the list must match one of the allowed operations.
      */
     std::vector<Operator> create_model_operator_list(const std::vector<std::string> &operator_names);
+
+    /**
+    * A class that represents a point in a chosen coordinate system.
```
tjhei: you are not choosing the coordinate system here

include/aspect/utilities.h
```diff
@@ -1060,6 +1066,41 @@ namespace aspect
...
+    */
+    template <int dim>
+    class GeneralizedPoint
+    {
+      public:
+        GeneralizedPoint(Point<dim> position, const GeometryModel::Interface<dim> *geometry_model);
+
+        /**
```
tjhei: a second constructor with std::array and the enum might be useful (not possible if you also store the Point of course)

source/utilities.cc
```diff
@@ -2535,7 +2535,67 @@ namespace aspect
...
+
+    template <int dim>
+    std_cxx11::array<double,dim-1> GeneralizedPoint<dim>::get_surface_coordinates() const
+    {
+      if (coordinate_system == Coordinates::CoordinateSystem::cartesian)
+        {
+          if (dim == 2)
+            {
+              std_cxx11::array<double,1> a {coordinates[0]};
+              return reinterpret_cast<std_cxx11::array<double,dim-1>& >(a);
```
tjhei: this is a bit ugly but I assume ``return a;`` doesn't compile? I think it would make sense to use template specialization and implement this function for each dimension separately.

include/aspect/utilities.h
MFraters: hmm, that is a very interesting idea. I can not currently think of a situation where I would actually need the original position for my current use cases of this function. 

As a side note, we can always get the position back anyway by using the natural_to_cartesian_coordinates function. 

So the question is then do we want to 
1) remove the whole function and not store the position, 
2) leave the function but compute the position on the fly when needed, or 
3) leave the function as it is and store both the position and coordinate.

include/aspect/geometry_model/interface.h
```diff
@@ -26,12 +26,20 @@
...
 #include <set>
 
 
 namespace aspect
 {
+  namespace Utilities
+  {
+    namespace Coordinates
+    {
+      enum CoordinateSystem : unsigned short;
```
bangerth: Do you need this in addition to `#include`ing `aspect/utilities.h`? If not, remove the duplicate declaration.

include/aspect/utilities.h
```diff
@@ -164,18 +169,19 @@ namespace aspect
...
-       * below the 'top' surface (depth) as defined by each geometry model,
+       * 'spherical', 'ellipsoidal', and 'depth'. 'spherical' and 'ellipsoidal'
+       * coordinates follow: r, phi (2D) or r, phi, theta (3D); where r
+       * is radius, phi is longitude, and theta is the polar angle (colatitude).
+       * The 'depth' is a one-dimensional coordinate system in which only the
+       * distance below the 'top' surface (depth) as defined by each geometry model,
        * is used.
        */
-      enum CoordinateSystem
+      enum CoordinateSystem : unsigned short
```
bangerth: is it important for anything in this patch to declare that the width of the `enum` is `unsigned short`? If not, just let the compiler figure out the appropriate integer type.

include/aspect/utilities.h
bangerth: I think what @tjhei wants to say that you should use the wording "in a given coordinate system that may or may not be Cartesian".

include/aspect/utilities.h
bangerth: and break the arguments onto separate lines.

bangerth: I think the constructor could also use a doc string :-)

include/aspect/utilities.h
bangerth: I think it's ok to store both representations. (We probably won't store massive numbers of these objects anywhere, and even if we did, it would at worst triple the amount of memory that we use in `class Triangulation` to store vertex locations, for example.) So option 3.

include/aspect/utilities.h
```diff
@@ -1060,6 +1066,41 @@ namespace aspect
...
+
+        /**
+        * The coordinate that represents the 'depth' direction in the chosen coordinate system.
+        **/
+        double get_depth_coordinate() const;
+
+      private:
+        Utilities::Coordinates::CoordinateSystem coordinate_system;
+        Point<dim> cartesian_position;
+        std::array<double,dim> coordinates;
```
bangerth: Can you add comments?

include/aspect/utilities.h
bangerth: By the way, I tend to think that it would be nicer from a conceptual point of view if the class did not know about geometry models. You are here introducing a circular dependency, because this class needs to know about `GeometryModel::Interface`, while that class also needs to know about `GeneralizedPoint`.

You can avoid this if the class only takes as argument a `CoordinateSystem` argument as second argument. It would *typically* be filled by asking the geometry model what the appropriate coordinate system is, but you'd break the circular dependency.

source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -74,7 +75,6 @@ namespace aspect
     EllipsoidalChunk<dim>::EllipsoidalChunkGeometry::push_forward_ellipsoid(const Point<3> &phi_theta_d, const double semi_major_axis_a, const double eccentricity) const
     {
       AssertThrow (dim == 3,ExcMessage ("This can currently only be used in 3d."));
-
```
bangerth: why remove all of these empty lines? I think they help readability of the code.

source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -117,13 +116,12 @@ namespace aspect
...
       const double d_hat = phi_theta_d_hat[2]; // long, lat, depth
       Point<dim-1> phi_theta;
       const double rad_to_degree = 180/numbers::PI;
       if (dim == 3)
         phi_theta = Point<dim-1>(phi_theta_d_hat[0] * rad_to_degree,phi_theta_d_hat[1] * rad_to_degree);
-      const double h = topography->value(phi_theta);
+      const double h = topography != NULL ? topography->value(phi_theta) : 0;
```
bangerth: This seems like an unrelated bug fix, right? That's ok, but it's often simpler to put these things into a separate pull request because two smaller PRs are usually reviewer quicker than one larger one :-)

source/utilities.cc
bangerth: Yes, I agree. I think that would make things simpler.

source/utilities.cc
```diff
@@ -2535,7 +2535,67 @@ namespace aspect
...
+    }
+
+    template <int dim>
+    std_cxx11::array<double,dim-1> GeneralizedPoint<dim>::get_surface_coordinates() const
+    {
+      if (coordinate_system == Coordinates::CoordinateSystem::cartesian)
+        {
+          if (dim == 2)
+            {
+              std_cxx11::array<double,1> a {coordinates[0]};
```
bangerth: This only works in C++11, which we aren't allowed to use yet :-( I'm afraid you'll have to initialize components individually.

source/utilities.cc
```diff
@@ -2535,7 +2535,67 @@ namespace aspect
...
+    template <int dim>
+    std_cxx11::array<double,dim> &GeneralizedPoint<dim>::get_coordinates()
+    {
+      return coordinates;
+    }
+
+    template <int dim>
+    std_cxx11::array<double,dim-1> GeneralizedPoint<dim>::get_surface_coordinates() const
+    {
+      if (coordinate_system == Coordinates::CoordinateSystem::cartesian)
```
bangerth: Since you chain `if-else if` statements, I think it would be easier to read if you made this a `switch(coordinate_system) { ...` thingy. Then you could also add a `default: Assert (false, ExcNotImplemented());` statement at the end, where you are currently just falling off the end of the function.

source/utilities.cc
```diff
@@ -2535,7 +2535,67 @@ namespace aspect
...
+    }
 
+    template <int dim>
+    double GeneralizedPoint<dim>::get_depth_coordinate() const
+    {
+      if (coordinate_system == Coordinates::CoordinateSystem::cartesian)
+        return coordinates[dim-1];
+      else if (coordinate_system == Coordinates::CoordinateSystem::spherical || coordinate_system == Coordinates::CoordinateSystem::ellipsoidal)
+        return coordinates[0];
+      return 0;
```
bangerth: same here -- with a `switch` statement, things become a bit simpler, and you can also catch the case that the `coordinate_system` has a value that has not been checked before.

include/aspect/utilities.h
MFraters: I agree that it would be nicer to use the enum CoordinateSystem as an argument and I see the circular dependency problem (that is why I need to specifically declare the type of the enum), but I am not sure if I understand what you propose as a solution.

The problem I face with making the CoordinateSystem as an input is twofold. The first problem is that is that there is no direct linkage between the enum CoordinateSystem and the geometrymodels. Multiple geometrymodels belong to one CoordinateSystem. 

And secondly, related to that, access to the geometry model will be needed anyway, and from just a enum CoordinateSystem I can't start a new geometry model to do the transformation, because for example the ellipsoidal geometrymodel needs to fill it's manifold with parameters like eccentricity etc. (It would also seem wasteful to me to just start a new geometry model to transform points.).

include/aspect/utilities.h
MFraters: Yes, because it is not possible to forward declare an enum (which I currently need ininclude/aspect/geometry_model/interface.h) , except if you septically assign a type to it.

source/geometry_model/ellipsoidal_chunk.cc
MFraters: This is needed to make the new test work :)

include/aspect/utilities.h
tjhei: I did not suggest removing it because of memory concerns. The disadvantage is that if you want it to contain both representations, you always have to supply both of them (or cartesian + geometry model) in the constructor. I can see several situations where we want to use this in the future where it would be nice if I only need to supply the natural coordinates. Consider the case where I get some coordinates from the geometry, do some math with them, and then want to give them to a function that expects natural coordinates. How would I construct these with the current design?

MFraters: having one representation means you can compute the other in the constructor. I could add a constructor in which you give the natural coordinate, and it computes and stores the Cartesian coordinate internally for you.

bangerth:
@MFraters -- I think the way I had envisioned this class is that it is essentially self-contained: it stores a point in Cartesian coordinates, and some other coordinate system, and then an `enum` that indicates what this other coordinate system is. That's fundamentally already what you do, except that you pass the geometry model as an argument when I would have just passed the `std::array` instead. Of course, in practice both the `std::array` and the `enum` will typically be computed by a geometry model, but where they come from is not of concern to the `GeneralizedPoint` class, and in fact I can see uses of this class in other parts of the code that are not immediately related to the geometry model.

As far as design is concerned, here's a second consideration we could think about once we found consensus about the one above: why not let the `GeometryModels::Interpolate::cartesian_to_natural_coordinates()` function return a `GeneralizedPoint` right away? Because `GeneralizedPoint` stores both the natural and the Cartesian coordinates, there isn't really any need for `natural_to_cartesian_coordinates()` in that case, because you can get the Cartesian coordinates right from the `GeneralizedPoint` object. Or do I misunderstand something?

MFraters:
@bangerth So do I understand correctly that the constructor would then just store the given values, like this? 
```
GeneralizedPoint(Point<dim> position_, const std_cxx11::array<double,dim> coordinates_, Utilities::Coordinates::CoordinateSystem coordinate_system_)
:
position(position_),
coordinates(coordinates_),
coordinate_system(coordinate_system_)
{};
```
I guess that could work as well. The creator of the function then has the responsibility of making sure that the info in the point is actually correct and self consistent, although that shouldn't be too much of a problem I think.

>why not let the GeometryModels::Interpolate::cartesian_to_natural_coordinates() function return a GeneralizedPoint right away?

Yes, I like that, but l think it should be done in a follow up pull request.

bangerth:
On 11/15/2017 08:13 AM, MFraters wrote:
> @bangerth <https://github.com/bangerth> So do I understand correctly that the 
> constructor would then just store the given values, like this?
> 
> |GeneralizedPoint(Point<dim> position_, const std_cxx11::array<double,dim> 
> coordinates_, Utilities::Coordinates::CoordinateSystem coordinate_system_) : 
> position(position_), coordinates(coordinates_), 
> coordinate_system(coordinate_system_) {}; |
> 
> I guess that could work as well. The creator of the function then has the 
> responsibility of making sure that the info in the point is actually correct 
> and self consistent, although that shouldn't be too much of a problem I think.

Yes, that's exactly how I think the code should look. Leaving the change to 
`cartesian_to_natural_coordinates()` for later is totally fine by me.

MFraters:
@tjhei 
>Consider the case where I get some coordinates from the geometry, do some math with them, and then want to give them to a function that expects natural coordinates. How would I construct these with the current design?

If we would go with Wolfgangs proposal, it should be possible to do it this way: `GeneralizedPoint(geometry_model->natural_to_cartesian_coordiante(point),point,spherical)`, or for the other way around: `GeneralizedPoint(point,geometry_model->cartesian_to_natural_coordiante(point),spherical)`. 

Does this take away this concern, or am I misunderstanding your point? 

I would agree that if you don't need the Cartesian point that it involves computing something which you do not need. You could off course cheat and just pass it a default `Point<dim>()`, but having a NaturalPoint class for this case would of course be cleaner in that case. 

For my current direct needs, a NaturalPoint class would be enough, but I havn't really looked into if the GeneralizedPoint can be used to optimise/clean up parts of my code.

bangerth:
I see no reason why the class can't store both representations. It's just an extra 24 bytes. I'm totally unconcerned :-)

tjhei:
> I see no reason why the class can't store both representations. It's just an extra 24 bytes. I'm totally unconcerned :-)

As I said before "I did not suggest removing it because of memory concerns.". I think introducing a class ``NaturalCoordinate`` is much cleaner than having a class ``GenericCoordinate`` that contains both.

MFraters:
I could do both, in the way that the generic point has instead of a `std::array coordinate` as a member variable, a `NaturalCoordinate natural_coordinate`.

MFraters:
The `NaturalCoordinate` would then contain the `std::array` and the `enum`, and we could then actually support computations for the `NaturalCoordinate` based on enum without having to worry about updating the Cartesian point as well.

bangerth:
That seems like a nice abstraction indeed. @tjhei -- what is your opinion on this?

tjhei:
> That seems like a nice abstraction indeed. @tjhei -- what is your opinion on this?

I don't understand the need for a ``GenericCoordinate``. I would rather have ``Point<dim>`` (Cartesian coordinate) and ``NaturalCoordinate`` (std::array+enum wrapper with some logic for surface coordinates and depth). We can then have helpers to convert between the two (for example different constructors of ``NaturalCoordinate``). Different classes like ``GeometryModel`` can have functionality to return/take one or the other.
Maybe I am wrong, but I don't see an advantage for having both in the same class. Disadvantages:
- their members have to be read-only (or you have complicated logic to update the other format if you set one of them, urgh).
- less clear logic when used (what kind of coordinates are you going to work with?)
- redundant computations (you always pay for conversion even if you don't need it)
- can be inconsistent (if you construct a GenericCoordinate with incorrect values for one of them, maybe because you "don't care" in some situation)

MFraters:
The more I think about it, the more I agree with Timo on this point. I currently see no need to have a GenericPoint if we have a well functioning NaturalPoint. And if we need it later, we can always implement it later in the way described above. @bangerth, is this oke with you?

bangerth:
@MFraters -- what is the current status of this? You still like it? Should I take another look?

include/aspect/utilities.h
MFraters: Although I would like to do this, I do not think it is possible, because an enum is not enough to find or make a geometry model from it, which is needed for the conversion.

include/aspect/geometry_model/interface.h
MFraters: I can't remove it somehow, even though it is defined in `aspect/utilities.h` which is included here at the top.

MFraters:
I think that I have addressed all the the comments on this pull request :)

include/aspect/geometry_model/interface.h
bangerth: If you remove it, what is the error message?

include/aspect/utilities.h
bangerth: But this is a C++11 feature to derive an enum from an integer type. I don't think you can use that yet.

MFraters: hmm, you are right. That is slightly problematic, because what I read is that forward declarations of enums are not possible before c++11. So I will have to rethink this solution.

bangerth:
So you have a cyclic dependency between these headers. You can break this by putting the new `enum` into a separate header file, say `coordinate_systems.h` in which you declare the things you want, and that is `#include`d from both `utilities.h` and `interface.h`.

MFraters:
@bangerth, yes, that seems to be the best way to do it, thanks! 

I think that I have adressed all comments.

include/aspect/utilities.h
```diff
@@ -1069,6 +1057,45 @@ namespace aspect
     template <int dim>
     SymmetricTensor<2,dim> nth_basis_for_symmetric_tensors (const unsigned int k);
 
+    /*
+    * A class that represents a point in a chosen coordinate system.
+    */
+    template <int dim>
+    class NaturalCoordinate
```
bangerth: Would it make sense to move this class into your new header file as well?

bangerth: Or do you not want to do this because you use `GeometryModel::Interface`?

include/aspect/utilities.h
```diff
@@ -1069,6 +1057,45 @@ namespace aspect
...
+    {
+      public:
+        /**
+         * constructor based on providing the geometry model as a pointer
```
bangerth: capitalize `constructor` here

include/aspect/utilities.h
```diff
@@ -1069,6 +1057,45 @@ namespace aspect
...
+         */
+        NaturalCoordinate(Point<dim> &position,
+                          const GeometryModel::Interface<dim> *geometry_model);
```
bangerth: I think we generally use reference instead of pointers.

source/geometry_model/ellipsoidal_chunk.cc
bangerth: I think this comment still applies

include/aspect/utilities.h
MFraters: I havn't tested if I could put it there, but besides that I think it is a matter of how you view this header. I personally think it would be good to keep this new header small. This is because I see it as a work around header which might also have been called `enums.h`, which is only needed until we switch to c++11. Do you view it differently, and why would you want to put it there?

source/geometry_model/ellipsoidal_chunk.cc
MFraters: ah, I missed this one. Fixed it :)

MFraters:
> I'm slightly confused why the conversion functions exist as virtual functions in some material models, but not in the base class Interface. virtual functions are only really useful if they are declared in the common base class.

I see the inconsistency that they are not pure (but are all virtual as far as I can see), but also do not provide a default implementation. Is this what you mean? If so, which way would have your preference?

include/aspect/utilities.h
bangerth: No, that's fine. I was just wondering whether you had thought about it.

bangerth:
Regarding the virtual functions that are only declared in the derived classes: The only way you can use these functions is if you do a `dynamic_cast` to the derived class because these functions are not declared as virtual in the base class. Furthermore, if you already do a `dynamic_cast`, then there is no longer a need to mark these functions as `virtual` in the derived class.

Ultimately, what it comes down to is how you want to use these functions. I suspect that since there is a function in the `Interface` base class that allows querying *what* the natural coordinate system is, that you probably also want the ability to actually *query* the natural coordinates. But you can't currently do this because there is no such function in the base class that you can call.

There is not currently any code in this patch that shows how one would call this function from anywhere, so I can't judge what use cases you have in mind. Could you show us an example?

MFraters:
No problem! Squashed it down into logical blocks again :)

PR #1942, opened 2017-10-01
Fix particle serialization

gassmoeller:
The particle serialization was broken by #1913. This PR fixes this (the important part is the call to `particle_handler->initialize` that needed to be moved from `World::serialize` into `World::load`). It also makes some other places related to serialization easier to understand and finally adds a test for particle serialization.


bangerth:
Looks great, thank you!

PR #1703, opened 2017-05-16
New function for boundary composition

mbweller:
Function added to boundary composition to allow for variable compositions to placed with user defined coordinate systems. Boundary composition is now consistent with Boundary: Temperature, Velocity, and traction.

PR #1943, opened 2017-10-01
Boundary composition function

gassmoeller:
Supersedes #1703. See there for an explanation, I only addressed all but one comment and updated the tests.
The one remaining comment is that we declare parameters for a function with one component, although we parse parameters for `n_compositional_field` components. I agree with @mbweller that I do not quite see why it works, but we already use the same approach for `initial_composition/function` and I also do not see an alternative. `declare_parameters` is called in a static function that is called before we know how many components there will be. Thus I would vote for simply keeping this approach.


doc/modules/changes/20171001_mbweller
```diff
@@ -0,0 +1,5 @@
+New: The 'function' boundary composition plugin now allows user defined
+Coordinate Systems [cartesian; spherical; depth].
```
bangerth: Since that 'function' plugin didn't previously exist, I think this is misleading.

include/aspect/boundary_composition/function.h
```diff
@@ -0,0 +1,126 @@
...
+        /**
+         * A function object representing the compositional fields.
+         */
+        std_cxx11::unique_ptr<Functions::ParsedFunction<dim> > function;
+
+        /**
+         * Composition at the inner and outer boundaries.
+         */
+        double min_composition;
+        double max_composition;
```
bangerth: do we even need this? how would that deal with the situation of multiple compositional fields?

include/aspect/boundary_composition/function.h
```diff
@@ -0,0 +1,126 @@
...
+        update ();
+
+        /**
+         * Return the minimal composition on that part of the boundary on
+         * which Dirichlet conditions are posed. Note: This is left over from
+         * temperature formulations, and may need to be removed later if its
+         * efficacy is not proven.
+         */
+        virtual
+        double minimal_composition (const std::set<types::boundary_id> &fixed_boundary_ids) const;
```
bangerth: This and the next function are not actually part of the `Interface` class. Could we delete these two functions?

source/boundary_composition/function.cc
```diff
@@ -0,0 +1,210 @@
...
+                             "'spherical' coordinates are interpreted as r,phi "
+                             "or r,phi,theta in 2D/3D respectively with theta "
+                             "being the polar angle. 'depth' will create a "
+                             "function, in which only the first parameter is "
+                             "non-zero, which is interpreted to be the depth of "
+                             "the point.");
+
+          Functions::ParsedFunction<dim>::declare_parameters (prm, 1);
+
+          prm.declare_entry ("Minimal composition", "0",
```
bangerth: we may not need this parameter.

source/boundary_composition/function.cc
```diff
@@ -0,0 +1,210 @@
...
+              std::cerr << "ERROR: FunctionParser failed to parse\n"
+                        << "\t'Boundary composition model.Function'\n"
+                        << "with expression\n"
+                        << "\t'" << prm.get("Function expression") << "'"
+                        << "More information about the cause of the parse error \n"
+                        << "is shown below.\n";
+              throw;
+            }
+          min_composition = prm.get_double ("Minimal composition");
+          max_composition = prm.get_double ("Maximal composition");
```
bangerth: and here

source/boundary_composition/function.cc
```diff
@@ -0,0 +1,210 @@
...
+                                               "composition, it is interpreted as having units "
+                                               "seconds unless the global input parameter ``Use "
+                                               "years in output instead of seconds'' is set, in "
+                                               "which case we interpret the formula expressions "
+                                               "as having units year."
+                                               "\n\n"
+                                               "Because this class simply takes what the "
+                                               "function calculates, this class can not "
+                                               "know certain pieces of information such as the "
+                                               "minimal and maximal composition on the boundary. "
```
bangerth: I think this is a copy-paste issue from the temperature boundary values, right?

gassmoeller:
I addressed all comments. Ready from my side.

MFraters:
Hey Rene and Matt,
Looks like a good feature. I am just wondering why both Cartesian and spherical have a test, but depth doesn't have one.

gassmoeller:
A valid question. I added a test for the depth function :smile:

MFraters:
👍

gassmoeller:
Can I understand that as a "merge yourself if the tester is happy"? You could also right now merge your first pull request.

gassmoeller:
As far as I know we do not have an official policy for that. It depends a bit on the complexity of the pull request. In this case Wolfgang only brought up formal comments (remove those unnecessary things), and you can check that I addressed them, so if I were you, I would probably go ahead and merge it without waiting for his comment. If one reviewer requests changes, and you are not quite sure if they were addressed (or you think you do not understand the issue well enough) then I would wait for the original reviewer to answer. 
Also @bangerth: Is it worth collecting some policies about this somewhere to give our new developers more guidance on when to push the merge button?

MFraters:
Ok, then I will merge it :)

bangerth:
Yes, I think you did the right thing. We generally trust each other. In this case, since the comments I had were not really controversial, if @MFraters sees that they are addressed, then for him to merge the button is the right thing. On the other hand, if I had had comments that showed that I was critical of the general direction of the patch, then a response by @MFraters of the form "This looks good to me, but @bangerth had some bigger concerns and so I let him take another look before merging" would have been the right way to go about it. Makes sense?

@jaustermann , @anne-glerum , @naliboff -- FYI.

PR #1945, opened 2017-10-02
particle doxygen fixes

tjhei:


gassmoeller:
Yes makes sense. Those were my faults 😞

PR #1949, opened 2017-10-05
Add harmonic average particle interpolation scheme

hlokavarapu:
I have implemented a harmonic averaging scheme and added a test case based on SolKz were I am interpolating both density and viscosity.  The code is ready for an initial review @gassmoeller .

### DIFF
--- a/source/particle/interpolator/harmonic_average.cc
+++ b/source/particle/interpolator/harmonic_average.cc
@@ -1,0 +1,151 @@
+/*
+  Copyright (C) 2017 by the authors of the ASPECT code.
+
+ This file is part of ASPECT.
+
+ ASPECT is free software; you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


source/particle/interpolator/harmonic_average.cc
```diff
@@ -0,0 +1,150 @@
...
+namespace aspect
+{
+  namespace Particle
+  {
+    namespace Interpolator
+    {
+      ASPECT_REGISTER_PARTICLE_INTERPOLATOR(HarmonicAverage,
+                                            "harmonic average",
+                                            "Return the harmonic average of all particle properties in the "
+                                            "given cell.")
```
gassmoeller: Please add something like the following: `If the cell contains no particles, return the harmonic average of the properties in the neighboring cells.`
This is just to make clear that there are two differences compared to `cell_average`, the averaging of particles, and the averaging in case the cell is empty.

gassmoeller:
Thanks!

PR #1951, opened 2017-10-08
add explicit instantiation of convert_point_to_array and convert_array_to_point

MFraters:


bangerth:
OK if it passes the tester.

PR #1705, opened 2017-05-16
Chunk initial topography

anne-glerum:
@gassmoeller 
Hi Rene,
do you think you could give this geometry model a try? I've added a test that you could run. Also, right now, that test fails on my laptop because I put in an AssertThrow for older deal.II versions that try to add non-zero initial topography. What would be the best way to go around this?
I'd like a more compact way to put in the derivative gradient, but that's for tomorrow :)
Thank you!

PR #1952, opened 2017-10-10
Fix depth functions for initial topography

MFraters:
This pull request intents to 'fix' the depth functions of the geometry models. Currently the depth functions functions differently, where only the ellipsoidal chunk geometry model has the initial topography model implemented and accounts for this initial topography in the depth function. 

This pull request adds the adjustment of the depth (and max depth) for the box model. Before I work on others, I would first like to know if this is oke, or that there are reasons for not doing this? 

The test box_initial_topography_ascii_data will now fail because it uses very large initial topography together with surface pressure normalisation. The surface pressure normalisation uses the depth function together with face information to look if it is near the surface (helper_functions.cc:776). I think the test fails because the cells become too deformed.


bangerth:
I thought about this on some occasion before. Do you know where we currently use the depth function?

MFraters:
We use the depth function it in the following places (in source), and I am planning to use it extensively in the initial conditions generator:

```
simulator/lateral_averaging.cc
simulator/helper_functions.cc (in normalize_pressure)

adiabatic_conditions (all the plugins)
boundary_composition/function.cc
boundary_temperature/function.cc
boundary_traction/function.cc
boundary_traction/initial_lithostatic_pressure.cc
boundary_velocity/function.cc
boundary_velocity/gplates.cc
gravity_model/ascii_data.cc
gravity_model/radial.cc
heating_model/radioactive_decay.cc
initial_composition/function.cc
initial_temperature/adiabatic.cc
initial_temperature/function.cc
initial_temperature/harmonic_perturbation.cc
initial_temperature/S40RTS_perturbation.cc
initial_temperature/S40RTS_perturbation.cc
initial_temperature/SAVANI_perturbation.cc
initial_temperature/solidus.cc
initial_temperature/spherical_shell.cc
material_model/ascii_reference_profile.cc
material_model/composition_reaction.cc
material_model/depth_dependent.cc
material_model/latent_heat.cc
material_model/morency_doin.cc
material_model/steinberger.cc
mesh_refinement/maximum_refinement_function.cc
mesh_refinement/minimum_refinement_function.cc
mesh_refinement/topography.cc
material_model/nondimensional.cc
postprocess/boundary_densities.cc
postprocess/boundary_pressures.cc
postprocess/dynamic_topography.cc
postprocess/depth_average.cc -- only uses maximal_depth
postprocess/geoid.cc
postprocess/visualization/depth.cc
postprocess/visualization/geoid.cc
postprocess/visualization/seismic_anomalies.cc
```

source/simulator/core.cc
```diff
@@ -2569,6 +2569,10 @@ namespace aspect
       }
     else
       {
+        time                      = parameters.start_time;
+        timestep_number           = 0;
+        time_step = old_time_step = 0;
+
```
gassmoeller: Is this move important or just by accident?

include/aspect/utilities.h
```diff
@@ -533,6 +533,11 @@ namespace aspect
         std::string
         get_column_name_from_index(const unsigned int column_index) const;
 
+        /**
+         * Return the maximum value of the coordinate values.
+         */
+        double get_maximum_coordinate_value() const;
```
gassmoeller: I do not like the name of this function. In particular, I do not quite remember how this class works, and the name of the function does not help me understand it either. 
For example in the ascii data topography case: Does the table of `AsciiDataLookup` have `dim` columns or `dim-1` coordinates and one data component, with the first `dim-1` being the coordinates and the last column being the topography? Then what about using the following declaration instead: `get_maximum_data_value(unsigned int component_index) const`? This way we would not already assume what is in the columns (because for boundary composition or boundary temperature plugins that can be different). Additionally it helps to understand what this function is actually returning, and can also be used when there is more than one data column.

gassmoeller: Or does this function actually return the maximum `coordinates`? But then I do not understand. The coordinates in a 2D model would be just the 1D x-coordinates, but you are interested in the maximum y-coordinate / topography, which is in the data column, not the coordinate column.

source/geometry_model/box.cc
```diff
@@ -207,8 +207,15 @@ namespace aspect
     double
     Box<dim>::depth(const Point<dim> &position) const
     {
-      // this depth does not take topography into consideration
-      const double d = maximal_depth()-(position(dim-1)-box_origin[dim-1]);
+      // Get the surface x (,y) point
+      Point<dim-1> surface_point;
+      for (unsigned int d=0; d<dim-1; d++)
```
gassmoeller: to be consistent with the rest of our code make it `++d`

source/simulator/core.cc
MFraters: This is important, because time needs to be initialized before the refinement happens.

bangerth:
I think my question with where this is used was more of the kind "in the contexts where this function is used, what would one expect the function to do?

I don't actually know the answer. I *think* that it makes sense to make this change, but one could probably also argue that the depth is a property of some kind of "reference domain". It is true, though, that the reference domain is probably best seen as whatever the geometry gives us -- i.e., base domain plus initial topography. 

What do others think?

source/simulator/core.cc
gassmoeller: But none of the initial topography models should be time dependent. If they are: Is this currently a bug, or has something changed in this PR that makes this necessary?

MFraters: If I remember correctly, it has to do with the AsciiDataBoundary. The refinement function calls this because it needs to know the position and it ends up in utilities:1859
```
    template <int dim>
    double
    AsciiDataBoundary<dim>::
    get_data_component (const types::boundary_id             boundary_indicator,
                        const Point<dim>                    &position,
                        const unsigned int                   component) const
    {
      if (this->get_time() - first_data_file_model_time >= 0.0)
```
If the time is not set, the default value of numbers::signaling_nan<double>() will be triggered. This is because the ascii data was originally made for time dependent boundary conditions (or that is what I can make from reading the code) and therefore needs the current time to decide what the current boundary conditions are. Now that that code is also used for initial topography, it is getting called earlier then it was designed for.

include/aspect/utilities.h
MFraters: Yes, you are right that it is probably better to do this more general way. I also noticed I put something in the wrong place, that should now be fixed. I hope it is a bit more clear now.

source/simulator/core.cc
gassmoeller: I see. Then this makes sense. Thanks for the explanation.

include/aspect/utilities.h
```diff
@@ -533,6 +533,11 @@ namespace aspect
         std::string
         get_column_name_from_index(const unsigned int column_index) const;
 
+        /**
+         * Return the maximum value of the coordinate values.
```
gassmoeller: The documentation needs an update.

include/aspect/utilities.h
```diff
@@ -559,6 +564,11 @@ namespace aspect
         std_cxx11::array<std::vector<double>,dim> coordinate_values;
 
         /**
+         * The maximum value of the coordinate values
```
gassmoeller: here as well

include/aspect/utilities.h
```diff
@@ -672,6 +682,13 @@ namespace aspect
                             const unsigned int                   component) const;
 
         /**
+         * Returns the data component at the given position.
```
gassmoeller: same here

MFraters:
Ok, then I will also update the test solutions and add the chunk geometry model. Note that I had to scale down the maximum topography quite a bit in the ascii initial topography to not get into trouble with the surface pressure normalisation.

MFraters:
Since the initial topography hasn't been added to chunk yet as far as I can see, I do not have to fix the depth there. That can be done directly when it is added there in an other pull request.

anne-glerum:
@MFraters ah yes, #1705 adds the chunk initial topo for dealII 9, I'll try to get back to that soon.

MFraters:
I think all the comments have been addressed, and I havn't seen any opposition against this change, so I think it is ready for a final review :)

gassmoeller:
Please add an entry to the changelog, and then merge yourself.

The entry could be: `Changed: The 'depth' function of geometry models now uses any existing initial topography instead of the undeformed geometry as reference level to compute the depth.'

source/geometry_model/box.cc
```diff
@@ -207,8 +207,15 @@ namespace aspect
...
-      const double d = maximal_depth()-(position(dim-1)-box_origin[dim-1]);
+      // Get the surface x (,y) point
+      Point<dim-1> surface_point;
+      for (unsigned int d=0; d<dim-1; ++d)
+        surface_point[d] = position[d];
+
+      // Get the surface topography at this point
+      const double topo = topo_model->value(surface_point);
+
+      const double d = maximal_depth()-(position(dim-1)-box_origin[dim-1])+topo;
```
anne-glerum: Are you sure this is correct?` maximal_depth` includes the maximum topography to which you add the local topography as well. Shouldn't it be: 
`d = extents[dim-1] + topo - (position(dim-1)-box_origin[dim-1]);`

MFraters: Good catch! You are correct :) Although I checked the results visually, I didn't notice that the scale didn't start at depth zero :(

MFraters:
@anne-glerum Could you let me know if you are fine now with it being merged (or merge it directly)?

source/geometry_model/initial_topography_model/prm_polygon.cc
```diff
@@ -100,6 +110,7 @@ namespace aspect
              * we need to fill the point lists and topography values. They
              * are stored in the Topography subsection in the Topography parameter.
              */
+            maximum_topography = 0;
```
anne-glerum: In the hypothetical case that all prescribed topography is negative, taking the maximum with zero won't see the topography. Not sure how that affects anything down the line, but I guess it would be cleanest to use -std::numeric_limits<double>::max() for example.

source/utilities.cc
```diff
@@ -1338,7 +1348,10 @@ namespace aspect
           const unsigned int column_num = field_index%(components+dim);
 
           if (column_num >= dim)
-            temp_data *= scale_factor;
+            {
+              temp_data *= scale_factor;
+              maximum_component_value[column_num-dim] = std::max(maximum_component_value[column_num-dim], temp_data);
```
anne-glerum: Same here, a max between 0 and the data.

source/geometry_model/initial_topography_model/prm_polygon.cc
anne-glerum: In the hypothetical case that all the prescribed topography is negative, this won't be seen by taking the maximum with zero. Not sure how that affects stuff down the line, but I guess it would be neatest to use something like -std::numeric_limits<double>::max().

anne-glerum:
Just one comment, then please merge yourself :)

source/geometry_model/initial_topography_model/prm_polygon.cc
MFraters: good point, I changed it.

source/utilities.cc
MFraters: I added it to the resize function.

PR #1953, opened 2017-10-11
Assert radial gravity model for box geometry models.

MFraters:
I noticed that when I was switching from an ellipsoidal geometry model to a box geometry model that I forgot to change the gravity model, and some assert deep in the gravity model was triggered (division by zero, because the box has a (0,0,0) is in the box domain). I thought it might be better to just forbid the use of the radial gravity models in box geometries.


gassmoeller:
I like the idea, but agree with Timo's comments. Please try to cast pointers instead of checking the parameters. @tjhei: The current formulation should be fine for "sphere" geometries, because @mfraters only disallows box geometries for radial gravity profiles (i.e. it is a blacklist, not a whitelist).
@mfraters: Feel free to do a follow-up PR for the `vertical` gravity model, and disallow all spherical geometries (or add it here if you have a free minute). 

As a side note I would suggest changing the error message to the following:
"Gravity model NAME should not be used with geometry model NAME. If you are sure you want to proceed please remove this error check." 
Replace NAME with the names of the plugins. This makes it easier for users to find the error, and also tells them how to work around the check in case they truly know what they are doing.

tjhei:
> The current formulation should be fine for "sphere" geometries, because @MFraters only disallows box geometries for radial gravity profiles

My point is that a sphere contains the origin, so it should also be blacklisted, or not?

MFraters:
1. good point. 
2. If the sphere/hyper_ball grid has a point in the centre itself, then I guess that is then also a problem. But I haven't tested that yet. But the point of the radial is that it should work for sphere's and such. We might want to fix the gravity model so that for example in the centre point gets a fixed gravity.

@gassmoeller sure, I will add an assert for the vertical and adjust the message.

bangerth:
Actually, I'm not a fan of providing a hint about how to work around the problem by telling people to remove a check. They may think that that's a general strategy, and it's something not all users know how to do anyway because it requires re-compiling everything. (Think of people who got ASPECT from a virtual machine, for example, or whose installation is on a shared file system.)

If we cannot see a good reason to use these models together, then we should just forbid that. If we can see good reasons to use these models together, then we shouldn't add the assertion. I believe that the former is the case.

gassmoeller:
> Actually, I'm not a fan of providing a hint about how to work around the problem by telling people to remove a check.
I do not feel strongly about that. If I ever need to have a box geometry with a radial gravity I know how to work around it :smile:. I think it is more important to let people know *which* plugin does not work with *which* plugin that was the original reason I proposed the different error message. Maybe just scratch the second sentence.

> My point is that a sphere contains the origin, so it should also be blacklisted, or not?

I now see what you mean. Yes `RadialEarthLike` should not be used with `Sphere` (it should contain an Assert), because it is a fit to a profile for the mantle only. `RadialLinear` already contains a fix for the center, and `RadialConstant` would simply need the same fix. @MFraters would you mind fixing that while you are at it?

MFraters:
sure

MFraters:
I think I have addressed all the comments.

bangerth:
On 10/11/2017 04:39 PM, Rene Gassmoeller wrote:
> I do not feel strongly about that. If I ever need to have a box geometry with 
> a radial gravity I know how to work around it 😄. I think it is more important 
> to let people know /which/ plugin does not work with /which/ plugin that was 
> the original reason I proposed the different error message

Yes, I agree with that!

source/gravity_model/radial.cc
```diff
@@ -83,6 +94,21 @@ namespace aspect
...
+    void
+    RadialEarthLike<dim>::parse_parameters (ParameterHandler &prm)
+    {
+      Assert (dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0,
+              ExcMessage ("Gravity model radial earth-like constant should not be used with geometry model Box."));
+      Assert (dynamic_cast<const GeometryModel::TwoMergedBoxes<dim>*> (&this->get_geometry_model()) == 0,
+              ExcMessage ("Gravity model radial earth-like constant should not be used with geometry model box "
+                          "with lithosphere boundary indicators."));
+      Assert (dynamic_cast<const GeometryModel::Sphere<dim>*> (&this->get_geometry_model()) == 0,
+              ExcMessage ("Gravity model radial earth-like constant should not be used with geometry model box "
```
gassmoeller: box with lithosphere boundary indicators -> sphere

include/aspect/gravity_model/radial.h
```diff
@@ -38,7 +38,7 @@ namespace aspect
      * @ingroup GravityModels
      */
     template <int dim>
-    class RadialConstant : public Interface<dim>
+    class RadialConstant : public Interface<dim>, public virtual SimulatorAccess<dim>
```
gassmoeller: Is there a reason to use virtual inheritance instead of simply `public SimulatorAccess<dim>`? If not, please remove the `virtual` (also in the two other classes).

MFraters:
It seems that there are actually 10 tests which use wrong combinations 😯

tjhei:
awesome. ;-)

MFraters:
I have updated all the tests. I have chosen for all test to only update the gravity model to match the geometry model (and not the other way around). Furthermore, for the spherical case I almost always have chosen radial constant, expect for when there was a spherical shell which had the dimensions of the earth. For quite a few test there where not results to check them against (which makes them less useful), so I added the screen_ouput and the gnuplot solutions for those ones (the commit name tells you when that happened.)

source/gravity_model/radial.cc
```diff
@@ -71,6 +77,11 @@ namespace aspect
         prm.leave_subsection ();
       }
       prm.leave_subsection ();
+      Assert (dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model()) == 0,
+              ExcMessage ("Gravity model Radial constant should not be used with geometry model Box."));
```
tjhei: Can you please add quotes around the names? Here and in other places, I don't care if they are single or double quotes.

tjhei:
Looks good otherwise. Thanks for going through the tests.

include/aspect/gravity_model/radial.h
```diff
@@ -101,7 +108,7 @@ namespace aspect
      * @ingroup GravityModels
      */
     template <int dim>
-    class RadialLinear : public Interface<dim>, public virtual SimulatorAccess<dim>
+    class RadialLinear : public Interface<dim>, public SimulatorAccess<dim>
```
bangerth: oh, good catch -- that made no sense

include/aspect/gravity_model/vertical.h
```diff
@@ -37,7 +37,7 @@ namespace aspect
      * @ingroup GravityModels
      */
     template <int dim>
-    class Vertical : public Interface<dim>
+    class Vertical : public Interface<dim>, public virtual SimulatorAccess<dim>
```
bangerth: why does the derivation have to be `virtual` here?

MFraters: You are right, I copied that and forgot to remove it.

bangerth:
Please merge yourself once you have addressed this one issue. I see that the indentation also fails -- please fix that as well.

MFraters:
That is very weird, because it says a file is not indented correctly which I did not change (the tests/particle_property_multiple_functions/particles/particles-00001.0000.vtu file is not changed by this pull request). And if I do make indent on my machine, nothing is changed. I could try to patch but the diff just says the binary files differ :s 

What should I do with this?

```
diff --git a/tests/particle_property_multiple_functions/particles/particles-00001.0000.vtu b/tests/particle_property_multiple_functions/particles/particles-00001.0000.vtu
index 8fb29b9..1dca879 100644
Binary files a/tests/particle_property_multiple_functions/particles/particles-00001.0000.vtu and b/tests/particle_property_multiple_functions/particles/particles-00001.0000.vtu differ
```

MFraters:
ah, a rebase to the newest master seems to have fixed the problem :)

bangerth:
Yes, that often helps. I'm glad that worked out!

PR #1957, opened 2017-10-16
Use deal.II particle functionality

gassmoeller:
This PR belongs to #1897. It uses the deal.II particle functionality of dealii/dealii#5125 and can only be merged after that one is finished. Thus it will require the newest deal.II-dev version after it is merged or v.8.5.0, then it uses the old code. All deal.II-dev versions between 8.5.0 and dealii/dealii#5125 will not longer work with ASPECT when this PR is merged.
This is not ready for review or merge yet, it is mostly here for @hlokavarapu and @egpuckett to see the possible future changes to ASPECT's particle code, and also for me to see the tester results.


gassmoeller:
@tjhei: I have a question about aspect-dev-gcc-hd. Is this the latest deal.II master, or some older version of 9.0.pre? I suppose it stays on a constant version, because otherwise you would need to recompile deal.II for every tester run. Is it much work to bring deal.II up to a more recent version? Otherwise we will need to accept the test failures for aspect-dev, since this PR uses functionality that got only introduced in a later version of deal.II. I will take a closer look at the test results anyway, but it would be nice to see it working on aspect-dev as well.

gassmoeller:
I think I fixed all remaining issues with using the particle functionality of deal.II. I have tested for correctness of all tests using the ascii output. The vtu output tests are harder, because they are now compressed but everything runs and looks ok. This also fixes #2602.
This is ready for a review.

include/aspect/particle/world.h
```diff
@@ -165,11 +178,13 @@ namespace aspect
          */
         void update_particles();
 
+#if !DEAL_II_VERSION_GTE(9,0,0)
         /**
```
tjhei: why are these inside an ifdef?

source/particle/generator/quadrature_points.cc
```diff
@@ -43,8 +43,13 @@ namespace aspect
                                 quadrature_formula,
                                 update_quadrature_points);
 
-
+#if DEAL_II_VERSION_GTE(9,1,0)
+        MPI_Scan(&n_particles_to_generate, &prefix_sum, 1, DEAL_II_PARTICLE_INDEX_MPI_TYPE, MPI_SUM, this->get_mpi_communicator());
+#elif DEAL_II_VERSION_GTE(9,0,0)
+        MPI_Scan(&n_particles_to_generate, &prefix_sum, 1, PARTICLE_INDEX_MPI_TYPE, MPI_SUM, this->get_mpi_communicator());
+#else
         MPI_Scan(&n_particles_to_generate, &prefix_sum, 1, ASPECT_PARTICLE_INDEX_MPI_TYPE, MPI_SUM, this->get_mpi_communicator());
```
tjhei: how about you define ASPECT_PARTICLE_INDEX_MPI_TYPE to be the correct thing so that you can remove all these ifdefs here?

source/particle/particle_handler.cc
```diff
@@ -242,11 +256,42 @@ namespace aspect
...
 
+    template <int dim,int spacedim>
+    void
+    ParticleHandler<dim,spacedim>::update_cached_numbers()
+    {
+      update_n_global_particles();
+      update_next_free_particle_index();
+      update_global_max_particles_per_cell();
+
+      return;
```
tjhei: remove

include/aspect/particle/world.h
gassmoeller: Because all of the output functionality is not longer used in deal.II 9.0. We then use the `DataOut` class instead. I guess theoretically, I could leave out the ifdefs around the output functionality, and simply not use it anymore, but leaving them in reminds us that this is dead code and we can remove it once we require 9.0.

source/particle/generator/quadrature_points.cc
gassmoeller: Hmm, currently I only use this type in two locations, and if I define ASPECT_PARTICLE_INDEX_MPI_TYPE I need an ifdef there anyway (two distinguish between PARTICLE_INDEX_MPI_TYPE for deal 9.0 and DEAL_II_PARTICLE_INDEX_MPI_TYPE for 9.1). Plus then I need to keep the definition of ASPECT_PARTICLE_INDEX_MPI_TYPE, while this ifdef can go away once we require 9.1. So I would prefer to keep it in the current form, is that ok with you?

tjhei: alright, makes sense.

gassmoeller:
Oh great. This should close #2602. Let me go ahead and try the current master.

include/aspect/particle/generator/interface.h
```diff
@@ -100,15 +112,15 @@ namespace aspect
            */
           virtual
           void
-          generate_particles(std::multimap<types::LevelInd, Particle<dim> > &particles) = 0;
+          generate_particles(std::multimap<Particles::internal::LevelInd, Particle<dim> > &particles) = 0;
```
bangerth: This suggests a design flaw in the deal.II classes -- if other codes are using internal data structures, we've done something wrong :-(

include/aspect/particle/particle.h
```diff
@@ -30,8 +31,22 @@
...
 {
+  namespace Particles
+  {
+    namespace internal
+    {
+      /**
+       * Typedef of cell level/index pair. TODO: replace this by the
+       * active_cell_index from deal.II 8.3 onwards.
+       */
+      typedef std::pair<int, int> LevelInd;
```
bangerth: We are already requiring 8.5, so the comment seems odd.

include/aspect/particle/particle.h
```diff
@@ -92,6 +101,15 @@ namespace aspect
 #endif
     }
 
+    namespace internal
+    {
+      /**
+       * Typedef of cell level/index pair. TODO: replace this by the
+       * active_cell_index from deal.II 8.3 onwards.
+       */
+      typedef std::pair<int, int> LevelInd;
```
bangerth: same here

PR #1958, opened 2017-10-17
Only filter hdf5 output for continuous elements

gassmoeller:
Something I noticed while working on #1864. 
The hdf5 filter tries to merge duplicate vertices at the same location to reduce the necessary disk space. However, for discontinuous elements this means averaging the values at the element boundary, something that is often not intended. Even worse: The merge does not always work, because the check in deal II does not take into account floating point imprecision in the vertex position, I am opening a separate issue for that. The consequence is that during mesh refinement, even if the mesh does not change and the solution stays the same, the internal position of the vertices can be changed slighty, and the filter will create a different output field, suggesting a change that does not exist. This behavior is not a problem for continuous elements (whether we average the constant values at adjacent vertices or not does not matter), but it is a problem for discontinuous solutions. Thus, this PR disables the filtering for discontinuous solutions.

### DIFF
--- a/source/postprocess/visualization.cc
+++ b/source/postprocess/visualization.cc
@@ -481,8 +481,13 @@ namespace aspect
           const std::string h5_solution_file_name = "solution/" + solution_file_prefix + ".h5";
           const std::string xdmf_filename = "solution.xdmf";
 
-          // Filter redundant values
-          DataOutBase::DataOutFilter data_filter(DataOutBase::DataOutFilterFlags(true, true));
+          // Filter redundant values if all elements are continuous, for discontinuous
+          // elements the averaging at cell vertices distorts the solution
+          const bool filter_duplicate_vertices = ! (this->get_parameters().use_locally_conservative_discretization
+                                                    || this->get_parameters().use_discontinuous_temperature_discretization
+                                                    || this->get_parameters().use_discontinuous_composition_discretization);
+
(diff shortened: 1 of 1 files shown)


tjhei:
(sorry, late to the party) How does the filtering behave if you output other discontinuous data (stress, errors, topography, DG pressure, etc.)?

gassmoeller:
Same problem. Should we disable it unconditionally? Or make it an input parameter (for people who want to save disk space)?

bangerth:
I kept a tab open in my eclipse to remind myself that I wanted to add a warning that users should only do the filtering if their data is continuous. Anything else clearly will lead to incorrect output.

We never did the filtering for VTU precisely because of the risk of doing something stupid, and because it's so difficult to find bugs if your visual output does not match the actual solution. I wouldn't mind if we ended up removing the filtering altogether -- who cares about a few extra MB or GB of disk space these days.

bangerth:
(But if we end up doing that, I'd still be interested in getting this PR merged so we can go back to it if necessary.)

gassmoeller:
I would like to keep the filtering as an option. Because we might not be talking about a few MB or GB, but several TB for some of us (although I have to admit for all my large scale models I used vtu so far, but I am also carrying around 10 TB of data right now, so I might need to start using this functionality :smile: ).

bangerth:
Fair enough :-)

tjhei:
I think an option with default = false (independent of FE choice) is the best approach if you care about it.

gassmoeller:
I just read your comment @tjhei :smile:, see #1959

PR #1959, opened 2017-10-19
Make filtering hdf5 output an input parameter

gassmoeller:
Follow up to #1958. Also adds a changelog entry to let people, who used the hdf5 output before, know.

### DIFF
--- a/source/postprocess/visualization.cc
+++ b/source/postprocess/visualization.cc
@@ -481,13 +481,8 @@ namespace aspect
           const std::string h5_solution_file_name = "solution/" + solution_file_prefix + ".h5";
           const std::string xdmf_filename = "solution.xdmf";
 
-          // Filter redundant values if all elements are continuous, for discontinuous
-          // elements the averaging at cell vertices distorts the solution
-          const bool filter_duplicate_vertices = ! (this->get_parameters().use_locally_conservative_discretization
-                                                    || this->get_parameters().use_discontinuous_temperature_discretization
-                                                    || this->get_parameters().use_discontinuous_composition_discretization);
-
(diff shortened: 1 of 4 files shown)


source/postprocess/visualization.cc
```diff
@@ -802,6 +797,14 @@ namespace aspect
                              "and a factor of 8 in 3d, when using quadratic elements for the velocity, "
                              "and correspondingly more for even higher order elements.");
 
+          prm.declare_entry ("Filter output", "false",
+                             Patterns::Bool(),
+                             " deal.II offers the possibility to filter duplicate vertices for hdf5 "
+                             " output files. This merges the vertices of adjacent cells and "
+                             " therefore saves disk space, but misrepresents discontinuous "
```
bangerth: You can omit the spaces at the beginning of each line since you already have them at the end.

source/postprocess/visualization.cc
```diff
@@ -802,6 +797,14 @@ namespace aspect
...
+                             " output properties. Activating this function reduces the disk space "
+                             " by about a factor of $2^{dim}$ for hdf5 output.");
```
bangerth: Does this flag have an effect for anything other than HDF5/XDMF output? If not, maybe mention that this.

In any case, I would put a prominent warning into this text:
```
  \\note{\\textbf{Warning:} Setting this flag to true will result in visualization output that does not accurately
  represent discontinuous fields. This may be because you are using a discontinuous finite element for the
  pressure, temperature, or compositional variables, or because you use a visualization postprocessor that
  outputs quantities as discontinuous fields (e.g., the strain rate, viscosity, etc.). These will then all be visualized
  as \\textit{continuous} quantities even though, internally, \\aspect{} considers them as discontinuous fields.}
```

PR #1960, opened 2017-10-20
Average stress limiter and pre yield viscosity

anne-glerum:
@naliboff Maybe I misunderstand, but at the moment, the stress_limiter option does not seem to take into account the pre_yield_viscosity. From the papers cited in the plugin description, I gathered, usually the viscosity_pre_yield and the viscosity_limiter are averaged, so I added that in. Or would you rather keep the min(viscosity_pre_yield, viscosity_limiter) that was commented out?

### DIFF
--- a/include/aspect/material_model/visco_plastic.h
+++ b/include/aspect/material_model/visco_plastic.h
@@ -235,7 +235,6 @@ namespace aspect
         std::vector<double> friction_strain_weakening_factors;
 
         std::vector<double> prefactors_diffusion;
-        std::vector<double> stress_exponents_diffusion;
         std::vector<double> grain_size_exponents_diffusion;
         std::vector<double> activation_energies_diffusion;
         std::vector<double> activation_volumes_diffusion;
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -201,9 +201,9 @@ namespace aspect
           // Note: values of A, d, m, E, V and n are distinct for diffusion & dislocation creep
 
           // Diffusion creep: viscosity is grain size dependent (m!=0) and strain-rate independent (n=1)
-          double viscosity_diffusion = 0.5 * std::pow(prefactors_diffusion[j],-1/stress_exponents_diffusion[j]) *
+          double viscosity_diffusion = 0.5 / prefactors_diffusion[j] *
                                        std::exp((activation_energies_diffusion[j] + pressure*activation_volumes_diffusion[j])/
-                                                (constants::gas_constant*temperature*stress_exponents_diffusion[j])) *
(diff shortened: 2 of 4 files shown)


naliboff:
@anne-glerum - Good question. I don't recall specifically why we ended up only using the pre_yield viscosity and not some average. @cedrict and I tested the stress limiter rheology at the Tahoe Hackathon and found fairly resolution independent results with the current implementation. 

Perhaps we should have a flag to use either option (average or only pre_yield viscosity)? Using only the pre_yield viscosity is more consistent with the Drucker Prager approach, but if people would like to benchmark the older papers the averaging option would be useful. If I forced to choose I would use the current implementation, but I don't really have a strong opinion either way.

anne-glerum:
@naliboff 

1) My worry was actually that the stress limiter rheology at the moment is _not_ using the pre_yield_viscosity at all. The min(viscosity_pre_yield, viscosity_limiter) line is commented out, and the final viscosity is set to viscosity_limiter. Do you agree that this should be changed?

2) Yes, especially since the strain rate term was removed from the diffusion viscosity calculation, but not other terms with the diffusion stress exponent, which is inconsistent in case someone would set it to something else but 1.

naliboff:
@anne-glerum 

I think the reason the stress limiter rheology does not depend on the pre_yield_viscosity is that the tests we did ('brick' benchmarks) were in comparison to the `drucker_prager` model, which also does not use the pre_yield_viscosity. Looking back, I believe we actually started out with a separate stress_limiter material model and then imported that functionality in `visco_plastic`.

As those tests were only for the purposes of comparison to a Drucker Prager model, I think your proposed changes are good ones and more likely to be used down the line. Easy enough to add in previous or new stress limiter options if someone requests them down the line.

Would you like to make additional changes? If not, I think this is ready to merge. Thanks again for simplifying the diffusion creep flow laws!

anne-glerum:
Ah I see. But for the stress limiter functionality to be in the visco_plastic model I still think (and agree with you) that it should include the viscous pre_yield_viscosity part. 

There don't seem to be tests using the limiter, should I change one of the viscoplastic tests to incorporate it?

naliboff:
@anne-glerum Yes, I think it would be a good idea to add a test with the stress_limiter model. If you would like to move on from this pull request, I would be happy to add one.

gassmoeller:
Sorry for letting this lie around for so long. What is the status of the changes? Are they already in the main version? If not: There seems to be a conflict with the current version, could you resolve this and rebase?

anne-glerum:
I've added the use of the limiter to an existing test and rebased to master. Ready for final review.

naliboff:
@anne-glerum - Thanks for adding a test for the stress limiter and this looks good to go.

PR #1850, opened 2017-07-11
Refactor solver schemes

gassmoeller:
This is an early stab at #972 and #1279, and is blocked until #1707 is merged. It is not ready for review, I open it mostly to take a look at the test results.

PR #1962, opened 2017-10-20
Unify rebuild of stokes matrix

gassmoeller:
Follow up to #1850 that unifies the condition under which we rebuild the Stokes matrix and preconditioner. I have a technical question here:
 The current solver schemes had three different conditions for when to rebuild:

1. Only rebuild in every nonlinear iteration if stokes matrix depends on solution
2. Rebuild if stokes matrix depends on solution or there are prescribed velocity BCs
3. Rebuild if stokes matrix depends on solution or (we are in the first iteration and there are prescribed velocity BCs)

Which of these is correct? My understanding was that the prescribed velocity BCs end up in the constraints matrix, which is applied after the matrix is assembled. Thus, if neither matrix nor constraints change in subsequent iterations there should be no need to reassemble matrix or preconditioner, is that correct?

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -1236,15 +1236,17 @@ namespace aspect
   bool
   Simulator<dim>::stokes_matrix_depends_on_solution() const
   {
-    // currently, the only coefficient that really appears on the
-    // left hand side of the Stokes equation is the viscosity. note
-    // that our implementation of compressible materials makes sure
-    // that the density does not appear on the LHS.
-    // if melt transport is included in the simulation, we have an
+    // Currently, the only coefficient that really appears on the
+    // left hand side of the Stokes equation is the viscosity and possibly
+    // the density in the case of the implicit reference density profile
+    // approximation.
+    // If melt transport is included in the simulation, we have an
     // additional equation with more coefficients on the left hand
     // side.
 
     return (material_model->get_model_dependence().viscosity != MaterialModel::NonlinearDependence::none)
+           || (parameters.formulation_mass_conservation ==
+               Parameters<dim>::Formulation::MassConservation::implicit_reference_density_profile)
            || parameters.include_melt_transport;
   }
 
--- a/source/simulator/solver_schemes.cc
+++ b/source/simulator/solver_schemes.cc
@@ -215,6 +215,14 @@ namespace aspect
   double Simulator<dim>::assemble_and_solve_stokes (const bool compute_initial_residual,
                                                     double *initial_residual)
   {
+    // If the Stokes matrix depends on the solution, or the boundary conditions
+    // for the Stokes system have changed rebuild the matrix and preconditioner
+    // before solving.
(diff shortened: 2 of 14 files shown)


bangerth:
This fails a bunch of tests, but they all seem like pretty reasonable changes. Can you update the output?

source/simulator/helper_functions.cc
```diff
@@ -1236,15 +1236,17 @@ namespace aspect
...
+    // left hand side of the Stokes equation is the viscosity and possibly
+    // the density in the case of the implicit reference density profile
+    // approximation.
+    // If melt transport is included in the simulation, we have an
     // additional equation with more coefficients on the left hand
     // side.
 
     return (material_model->get_model_dependence().viscosity != MaterialModel::NonlinearDependence::none)
+           || (parameters.formulation_mass_conservation ==
+               Parameters<dim>::Formulation::MassConservation::implicit_reference_density_profile)
```
bangerth: Hm, interesting catch. Did we just forget this before?

Is it easy to understand why the implicit reference density affects the viscosity? The density doesn't appear in the lhs of the Stokes equation, or does it?

source/simulator/solver_schemes.cc
```diff
@@ -204,6 +204,14 @@ namespace aspect
   double Simulator<dim>::assemble_and_solve_stokes (const bool compute_initial_residual,
                                                     double *initial_residual)
   {
+    // If the Stokes matrix depends on the solution, or the boundary conditions
+    // for the Stokes system have changed rebuild the matrix and preconditioner
+    // before solving.
+    if (stokes_matrix_depends_on_solution()
+        ||
+        (parameters.prescribed_velocity_boundary_indicators.size() > 0))
+      rebuild_stokes_matrix = rebuild_stokes_preconditioner = true;
```
bangerth: yes, this is a better place for this check!

source/simulator/helper_functions.cc
gassmoeller: In this formulation the density appears on the left-hand side. We simply move the explicit reference density profile term to the left-hand side, and hope that the resulting asymmetry of the matrix does not matter too much for the preconditioner (see source/simulator/assembly.cc:316-360). @tjhei created a table in our 2017 paper that shows that the change increases the number of iterations compared to the explicit case with only one nonlinear iteration, but it needs less total iterations than iterating out the explicit version (at least for that example case). 
I think it did not matter that we forgot to add this condition, because we rebuild the matrix in essentially any case anyway, but this makes the condition explicit.

gassmoeller:
Since the tests only show a few models that now rebuild the preconditioner more often than before (without changing the results more than noise), I will go ahead with the merge as you suggested.

source/simulator/helper_functions.cc
bangerth: Ah, this makes sense now. I had somehow thought that `implicit_reference_density` has to do with the material model, not the formulation. Should have read the namespace name as well! :-)

PR #1965, opened 2017-10-27
Fix missing factor of two in preconditioner.

bangerth:
#1939 mentions a missing factor of 2 in assembling the A term in the preconditioner. I think @MFraters is correct in suggesting that it shouldn't matter since whether we solve `P A = P b` or `2P A = 2P b` doesn't matter. But it looks wrong.

I'm an incrementalist, so let me try one thing at a time -- here just adding this factor to the regular Stokes preconditioner. I'm curious to see what effect this has on the tests, and if it works, we can deal with the same missing factor in other places as well.

I'll also note that while the A block of the Stokes matrix includes the compressible effects, the preconditioner does not. This *also* looks like a bug/missed opportunity to get a better preconditioner.


bangerth:
Oh, this is interesting: If you search for "Solving Stokes" in the diffs this patch produces,
http://www.math.clemson.edu/~heister/cib/aspect-8.5-gcc-hd/34c4db57e8e20c0b8c91fef1b4bfb24f7da6ecfd/changes-gcc.diff
you'll see that the number of iterations is occasionally slightly worse, and frequently *substantially* better!

On second thought, it makes sense that the factor actually does affect something. That's because, unlike  what I said above, the preconditioner is not actually just P. That is true when using the "expensive" solver option, where the preconditioner we use for the whole Stokes system is of the form
```
  [ A^{-1}  B         ]
  [ 0          M^{-1} ]
```
where `A^{-1}` is computed by actually inverting the `A` matrix iteratively -- this is where we use the preconditioner for `A` that's in question here, and for that the missing factor of 2 should not actually matter.

But for the cheap solver, we use
```
  [P          B         ]
  [ 0          M^{-1} ]
```
as a preconditioner in GMRES, and here the missing factor of 2 clearly matters!

So I'm inclined to suggest that we merge this patch here. If you all agree, then I'll pull the diffs in the output files and apply them to my patch.

bangerth:
Thanks, @MFraters, for finding this, by the way!

If you don't mind too much, can we deal with this separately from the question about SPD-ness of the matrix? I think it's a separate issue that is probably best dealt with separately. (I like treating separate issues separately, because that makes reviewing and evaluating their correctness so much easier.)

bangerth:
By the way, I've also looked at the diffs in test output due to this change that aren't in the number of iterations, and they all seem to be more or less in the noise.

gassmoeller:
Very nice! Looks like a clear improvement to me. Thanks for pointing this out @MFraters and for testing @bangerth. Should we make it a `2.0` instead of a `2` to avoid confusion? It will not matter for the result of course.
Should we add an entry to the changelog? Technically nothing has changed about the result, but it is always nice to include a sentence in the release notes that mentions that the code is now faster than before :smile:.
Can you open a separate issue about the compressible terms in the preconditioner?

bangerth:
Yes, if you agree that we should merge this (or something like this), then I can make that `2.0` and update the test output. And add a changelog entry :-)

gassmoeller:
I have looked through the results and agree we should merge it. Less Stokes iterations, more better.

bangerth:
I pushed the updated pull request. Let's see if the tester agrees (as it should).

gassmoeller:
The tester is happy. I forgot one thing in my earlier review: Could you do the same to the melt preconditioner assembler (melt.cc:190/192) and update the melt tests as well? After that and if the tester is happy, feel free to merge yourself.

MFraters:
@bangerth Thanks for explaining why and when it does and doesn't matter :)

> Less Stokes iterations, more better.

haha, well put ;)

bangerth:
@gassmoeller -- let me do this in a separate PR so we can keep the effects separate.

gassmoeller:
Ok, I am fine with this then.

tjhei:
Note that the usertime to run the testsuite improved quite a bit:
```
old: user	107m39.100s
new: user	117m58.070s
```

Awesome!

bangerth:
Oh, this is nice indeed! Best use of an extra 12 characters in the source code ever!

gassmoeller:
@tjhei Did you mix up old and new? Because this looks like a worse runtime to me.

tjhei:
Yes, mixed up old and new because I was too excited. ;-)

This reduces the total time of the tester from 27 to 25 minutes "real" time.

gassmoeller:
Nice!

PR #1968, opened 2017-10-27
Simplify error reporting.

bangerth:
In general, writing large amounts of data directly rarely leads to good
things, in particular if the program is running parallel. Instead, just
abort the program -- one has to start a debugger anyway.

This is a follow-up to #1939. @mfraters?


MFraters:
Yes, I agree. I was also thinking about simplifying this, because although the extra information looked interesting during the development, it turned out to be not very useful in practice.

bangerth:
Yes. The matrices that the old code printed were huge. There's little practical value in seeing all of these values.

PR #1970, opened 2017-10-27
Also fix the preconditioner for the melt case.

bangerth:
See the comment in #1965. Also the discussion in #1939.


bangerth:
This is already on top of #1965, so we should only see the test changes in melt-related tests.

gassmoeller:
Interesting, melt results are at best equal, in some cases significantly worse than before. And the iteration numbers for the expensive solver change as well (even in models that do not do cheap iterations before the expensive ones). Does anyone have an idea why?

tjhei:
I wonder if this is related to the Schur complement approximation. I will play with this for a bit...

tjhei:
See #1971 for an experiment where I use 2eta also for the Schur complement terms.

gassmoeller:
This was superseded by #1976.

PR #1971, opened 2017-10-27
2.0 * eta for melt preconditioner

tjhei:
See the comment in #1965. Also the discussion in #1939. Alternative to
#1970.

### DIFF
--- a/source/simulator/melt.cc
+++ b/source/simulator/melt.cc
@@ -198,7 +198,7 @@ namespace aspect
                                             :
                                             eta_two_thirds * (scratch.div_phi_u[i] * scratch.div_phi_u[j]))
                                            +
-                                           (one_over_eta *
+                                           (one_over_eta/2.0 *
                                             pressure_scaling *
                                             pressure_scaling)
                                            * scratch.phi_p[i] * scratch.phi_p[j]
@@ -209,7 +209,7 @@ namespace aspect
                                            scratch.grad_phi_p[i] *
                                            scratch.grad_phi_p[j]
                                            +
-                                           (1./eta + 1./viscosity_c) *
+                                           (1./(2.0*eta) + 1./viscosity_c) *
                                            pressure_scaling *
                                            pressure_scaling *
                                            (scratch.phi_p_c[i] * scratch.phi_p_c[j])


tjhei:
Let's see what the tester thinks.

bangerth:
This seems wrong (and not only from a stylistic point, to call `eta` the quantity that is `2 * viscosity` :-)

In the derivation of the Schur complement, the operator `-div (2*eta*eps(u))` gives rise to the mass matrix with a factor of `1/eta`, not `1/(2*eta)`.

tjhei:
> In the derivation of the Schur complement, the operator -div (2*eta*eps(u)) gives rise to the mass matrix with a factor of 1/eta, not 1/(2*eta).

Are you sure? The Schur complement approximation assumes: A \approx - \eta \triangle

tjhei:
One data point (adiabatic_heating_with_melt test with expensive solver):
```
your approach: 29 224 31
always 2: 27 207 29
old: 27 210 29 
```
where the numbers are outer and inner A and inner S.

tjhei:
melt_transport_compressible_iterative:
```
old: 0+277 iterations.
your approach: 0+816 iterations.
my approach: 0+193 iterations.
```

bangerth:
I'm not saying that *empirically*, choosing 1/(2*eta) might not lead to a better preconditioner. That is quite possible. All I'm saying is that the strict derivation of the preconditioner yields a factor of 1/eta.

That's because the computation goes like this: The Schur complement is 
```
  S = B^T (A^-1) B
```
which corresponds to the operator
```
  S = (-div)  [-div (2*eta*eps(.))]^{-1} grad
```
Let's assume for a moment that eta is constant so that we can pull it out. We then get
```
  S = 1/(2*eta) * (-div)  [-div (eps(.))]^{-1} grad
```
Next, if you deal with divergence free fields, remember that `div u = 0`, so
```
  eps(u) = 1/2 (grad u + grad u^T)
  div (eps(u)) = 1/2 (Delta u + div(grad u^T))
```
but `div(grad u^T) = grad div u = 0`, and so for divergence free fields,
```
  div (eps(u)) = 1/2 (Delta u)
```
and consequently
```
  S = 1/(2*eta) * (-div)  [-div (eps(.))]^{-1} grad
     = 1/(2*eta) * (-div)  [-1/2 Delta]^{-1} grad
     = 1/eta * (-div)  [-Delta]^{-1} grad
```
By the commutator assumption, we then get that `S` is spectrally equivalent to the operator `1/eta`, which is consequently the factor that we put into the mass matrix.

The message here is that for divergence free functions, the operator `-div(2*eta*eps(u))` corresponds to `-div (eta grad u)` without the factor of 2.

tjhei:
Thanks for spelling it out. You assume two things:
1. u is smooth enough that I can swap the order of taking derivatives.
2. div u = 0.

At least for melt transport problems 2. is not true for regions with melt. We can argue about 1. or whether it is more appropriate to assume that (eps(u),eps(v)) is spectrally equivalent to (grad(u),grad(v)) (my assumption).

So, what do we do? Your suggestion certainly causes some regressions as written above:

> melt_transport_compressible_iterative:
> 
> old: 0+277 iterations.
> your approach: 0+816 iterations.
> my approach: 0+193 iterations.

tjhei:
Of course, it is obvious that #1972 doesn't make sense by looking at the results.

bangerth:
You miss the point. Compressibility is a continuous property whereas the question of whether 1/eta or 1/(2*eta) is correct is a discrete property. We agree that 1/eta is correct for the incompressible case. But then 1/(2*eta) can not be correct for the compressible case because then there needs to be a critical compressibility below which we switch to 1/eta. 

In other words, if 1/eta were not the correct choice for compressible models, then the correct choice has to be something like 1/(f(kappa) * eta) where f(kappa) is a function of the compressibility, and f(0)=1.

bangerth:
As an unimportant side remark, I don't think that the regularity of the solution is an important factor in whether or not the commutator relationship holds. That would be the case if the derivatives in `-div ( [-Delta]^{-1} grad)` all applied to the same function, but they don't: the `grad` is applied to a pressure, the other operations to a velocity. We cannot really reorder these (so the word "commutator" is actually poorly chosen) because if we do, we end up with `[-Delta]^{-1} (-div)(grad)` where now `-Delta^{-1}` is applied to a pressure -- in other words, it is a completely different operator (no longer vector-valued, but scalar).

bangerth:
What to do: The current patch is more or less neutral as far as the number of iterations are concerned (some better, some worse), with the exception of the following ones:
* tests/melt_transport_compressible_iterative: substantial improvements for the first few, difficult steps
* tests/compression_heating: substantial deterioration

Overall, it seems like a rather modest win. I'd be ok to merge the patch if you added a detailed comment that explains that there is no *theoretical* basis for the factor, but that *empirically* it seems to help a small amount in many cases.

gassmoeller:
As discussed on Friday over lunch: Should we not first check if the inclusion of the compressible strain rate terms in the preconditioner (that are always present in the matrix for the melt models) make a difference? Before guessing why this PR sometimes improves, sometimes deteriorates the solution we should probably fix the things we know (or suspect) to always improve the solution. Afterwards we can check again, if this PR still improves the convergence.

gassmoeller:
Have we decided on how to proceed with this? We could rebase this PR to the current master and see what happens to the iterations. Or did we decide Wolfgang's argument in #1976 is the way forward?

bangerth:
Yes, now that the other preconditioner changes have all been merged (correct?), let's rebase this one and see what happens.

tjhei:
I am closing it, because you convinced me that this is not useful. I rebased the commits to the current master and while we save an iteration here or there, some iteration numbers now increase. The main improvement came through #1976, I think.

bangerth:
Great, thanks for confirming.I'm glad to hear that the math predicts what works and what doesn't after all :-)

PR #1972, opened 2017-10-27
use 2 eta in Schur complement

tjhei:
We currently assemble the Schur complement approximation as
```
B^T A^{-1} B  \approx  \int 1/\eta  phi_i phi_j
```
which comes from the fact that
```
A \approx \eta \triangle
```
But this is not correct, as our velocity system contains 2 eta. Let's
see what happens if we do this change.

### DIFF
--- a/source/simulator/assemblers/stokes.cc
+++ b/source/simulator/assemblers/stokes.cc
@@ -92,7 +92,7 @@ namespace aspect
             }
 
           const double eta = scratch.material_model_outputs.viscosities[q];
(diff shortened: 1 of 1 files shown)


bangerth:
This is just as wrong as in #1971 :-)

bangerth:
That's because `-div(2*eta*eps(u))` corresponds to `-div(eta*grad(u))`.

tjhei:
Iteration numbers increase with this, so I am convinced. See discussion in #1971.

PR #1973, opened 2017-10-27
implement simplified tests

gassmoeller:
This adds an cmake option to disable the test for correctness of our testsuite. It will still run all tests if `ASPECT_RUN_ALL_TESTS` is set, and crash if one of them crashes, but it will no longer check their results. This can be helpful for:
- an experimental tester with deal.II-dev to prevent us from implementing something that breaks for that (which we did several times in the past)
- people who know that their machines produce different results than the reference tester, but want to see locally if they broke something (without opening a pull request)
- maybe early users who want to check if they installed ASPECT correctly? Currently, we only check quick_mpi for them, and that might pass, but does not give great confidence that all possible scenarios work?

We should discuss if this should be the default option for the testuite. I am for it. I could also imagine removing the `ASPECT_RUN_ALL_TESTS` option altogether, because if the here implemented simple testsuite is the best option for everyone, than only running `quick_mpi` has the single benefit of being faster (both in running cmake, and running ctest).


CMakeLists.txt
```diff
@@ -133,7 +133,8 @@ CONFIGURE_FILE(
 # Next, set up the testsuite
 
 SET(ASPECT_RUN_ALL_TESTS OFF CACHE BOOL "Set up complete test suite to run.")
-
+SET(ASPECT_COMPARE_TEST_RESULTS OFF CACHE BOOL "Compare test results with high accuracy.")
```
tjhei: while I wouldn't mind having this as the default, it means I have to update the tester before we can merge this PR.

tests/CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ ENDIF()
 
 SET(ASPECT_BINARY "${ASPECT_BINARY}" CACHE STRING "" FORCE)
 SET(ASPECT_DIR "${ASPECT_DIR}" CACHE STRING "" FORCE)
+SET(ASPECT_COMPARE_TEST_RESULTS "${ASPECT_COMPARE_TEST_RESULTS}" CACHE STRING "" FORCE)
```
tjhei: this should be type BOOL, right?

tests/cmake/diff_test.sh
```diff
@@ -20,6 +20,8 @@ CMAKE_CURRENT_SOURCE_DIR=$3
 # 4. absolute cmake binary directory
 CMAKE_CURRENT_BINARY_DIR=$4
 
+# 5. whether to compare the produced results, or just run the test
```
tjhei: please comment that it needs to be "ON".

tjhei: also update the text in line 9

gassmoeller:
I addressed your comments and added a changelog entry. For now I kept the default values, we can always decide to change them at a later time. Ready from my side, lets see what the tester says.

tests/cmake/diff_test.sh
```diff
@@ -20,6 +20,8 @@ CMAKE_CURRENT_SOURCE_DIR=$3
 # 4. absolute cmake binary directory
 CMAKE_CURRENT_BINARY_DIR=$4
 
+# 5. when "OFF" just run the tests, otherwise also compare the results
```
tjhei: Nitpicking, but this is not quite true. It compares if set to "ON" and not otherwise.

tests/cmake/diff_test.sh
```diff
@@ -57,15 +59,19 @@ ORIGINAL_GEN_FULL_PATH=${CMAKE_CURRENT_BINARY_DIR}/output-${PRETTY_TEST_AND_FILE
...
+	${REF_FILE} ${GEN_FILE} > ${DIFF_OUTPUT}.tmp
 	;;
-    *)
+      *)
 	"${DIFF_EXE}" \
 	    ${REF_FILE} ${GEN_FILE} > ${DIFF_OUTPUT}.tmp
-esac
+  esac
+else
+  truncate -s 0 ${DIFF_OUTPUT}.tmp
```
tjhei: this won't work on a mac:
```
$ truncate -s 0 hello
-bash: truncate: command not found
```
Can we use
```
rm -f ${DIFF_OUTPUT}.tmp
touch ${DIFF_OUTPUT}.tmp
```
instead?

tjhei:
Please address the two comments and merge yourself.

With this I can create an experimental tester with deal-dev without too much trouble. I think I have everything ready to go already.

PR #1974, opened 2017-10-30
WIP: fix bug in Newton solver which increases velocities

MFraters:
Let me start with that this is not a final fix. 

The problem I found is that when using the Newton solver for several timesteps with a velocity boundary conditions is that the boundary velocity is added every time step. This means that the boundary velocity keep increasing (went in my case from 1 cm/year to 2 meter per year in about 30 time steps). 

The fix here was my first attempt to see if I correctly asses the problem, and the model with the constant boundary condition is now fine. The problem with this fix is that if there is a time dependent boundary condition, it will just apply the boundary from the first time step. 

The conceptual solution to this problem is I think  replacing `vel` by  something like `boundary_velocity_new-boundary_velocity_old`. But subtracting Functions from each other to create a new function is currently not possible. 

Before I start out coding a lot (and for example try to add this functionality to the deal.ii function), I would like to know what the best solution for this problem is.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -907,7 +907,7 @@ namespace aspect
                 mask[i]=introspection.component_masks.velocities[i];
             }
 
-          if (!assemble_newton_stokes_system || (assemble_newton_stokes_system && nonlinear_iteration == 0))
+          if (!assemble_newton_stokes_system || (assemble_newton_stokes_system && nonlinear_iteration == 0 && timestep_number == 0))
             {
               VectorTools::interpolate_boundary_values (*mapping,
                                                         dof_handler,


MFraters:
Hmm, apparently it didn't make the branch on my fork when I made this in github. I think that I should have clicked edited on my fork. Sorry. Should I delete it and reopen it through my branch?

gassmoeller:
It seems like you accidentally pushed to geodynamics/aspect instead of mfraters/aspect. Many people use the convention of naming your fork (mfraters/aspect) 'origin', while naming the main repository (geodynamics/aspect) 'upstream'. That helps to avoid this problem (maybe you have it already that way?). Anyway, yes please delete this branch, close this pull request, push to your fork and reopen it from your fork. Dont worry, I think I did the same thing a while ago :smile:.

bangerth:
The usual solution is to ensure the correct boundary values in the first Newton iteration and then using zero boundary values (in all of those places where you prescribe non-zero Dirichlet values for the velocity) for all following Newton iterations. step-15 has a discussion of this issue.

MFraters:
@gassmoeller I am quite sure that when I didn't have editing rights I have an option in github to make a branch on my fork en do a pull request from there. Now there was just one option and I assumed it was the same as before. But appareantly it wasn't. I will close this one now and delete the branch.
@bangerth let's continue in the new pull request.

PR #1975, opened 2017-10-30
fix bug in Newton solver which increases velocities and some other bugs and stuff

MFraters:
Let me start with that this is not a final fix.

The problem I found is that when using the Newton solver for several timesteps with a velocity boundary conditions is that the boundary velocity is added every time step. This means that the boundary velocity keep increasing (went in my case from 1 cm/year to 2 meter per year in about 30 time steps).

The fix here was my first attempt to see if I correctly asses the problem, and the model with the constant boundary condition is now fine. The problem with this fix is that if there is a time dependent boundary condition, it will just apply the boundary from the first time step.

The conceptual solution to this problem is I think replacing `vel` by something like `boundary_velocity_new-boundary_velocity_old`. But subtracting Functions from each other to create a new function is currently not possible.

Before I start out coding a lot (and for example try to add this functionality to the deal.ii function), I would like to know what the best solution for this problem is.


MFraters:
@bangerth 
>The usual solution is to ensure the correct boundary values in the first Newton iteration and then using zero boundary values (in all of those places where you prescribe non-zero Dirichlet values for the velocity) for all following Newton iterations. step-15 has a discussion of this issue.

This is what I do for the nonlinear iterations. But the problem is now between the time steps.

bangerth:
The solution is the same: in the first iteration, you should solve for `u` (not `delta u`) and then you ensure the *correct* boundary values in the first iteration. In subsequent iterations you solve for `delta u`, which then has zero boundary values.

MFraters:
@bangerth Something like this?

This will break the tests, because I now use the correct residual to compute the relative residual.

source/simulator/solver_schemes.cc
```diff
@@ -459,8 +459,11 @@ namespace aspect
 
         assemble_newton_stokes_system = true;
 
-        solution.block(introspection.block_indices.pressure) = 0;
-        solution.block(introspection.block_indices.velocities) = 0;
+        if (nonlinear_iteration != 0)
+          {
+            solution.block(introspection.block_indices.pressure) = 0;
+            solution.block(introspection.block_indices.velocities) = 0;
```
bangerth: can you add a comment here that explains what you are doing? I suppose what you are doing is to set the iterate delta u to zero whenever you're not on the first Newton step, but leave it as is if you *are* on the first step?

source/simulator/solver_schemes.cc
```diff
@@ -563,83 +566,92 @@ namespace aspect
...
+            pcout << "      Total relative residual after nonlinear iteration " << nonlinear_iteration+1
+                  << ": " << residual/initial_residual << ", norm of the rhs: " << residual
+                  << ", newton_derivative_scaling_factor: " << newton_handler->get_newton_derivative_scaling_factor() << std::endl;
+          }
+        else
+          {
+            /**
+             * We may need to do a line search if the solution update doesn't decrease the norm of the rhs enough.
+             * This is done by adding the solution update to the current linearization point and then assembling
+             * the Newton right hand side. If the Newton residual has decreased enough by using the this update,
```
bangerth: the this -> this

MFraters:
The original problem of this pull request is now fixed. Along the way I found some other problems, like that the velocity nonlinear channel flow test did produce the correct awnser because of a bracket error in the input file.

Furthermore I have tried to fix an other problem which has been bugging me, namely that when taking time step, it didn't use the previous time step. The current solution works wonderfully for the velocity nonlinear channel flow test:

```
...
   Skipping temperature solve because RHS is zero.
   The linear solver tolerance is set to 0.1
   Solving Stokes system... 47+0 iterations.
      Total relative residual after nonlinear iteration 28: 3.45932e-11, norm of the rhs: 2.52019e+07, newton derivative scaling factor: 1


   Postprocessing:
     RMS, max velocity:                  2.57e-08 m/s, 3.05e-08 m/s
     Pressure min/avg/max:               -1.203e+07 Pa, 5.008e+08 Pa, 1.014e+09 Pa
     Mass fluxes through boundary parts: 0 kg/s, 0 kg/s, -0.8139 kg/s, 0.8139 kg/s
     Writing graphical output:           output-nonlinear_channel_flow_velocities_Newton_Stokes/solution/solution-00000

*** Timestep 1:  t=1 seconds
   Skipping temperature solve because RHS is zero.
   Initial Newton Stokes residual = 7.28522e+17, v = 7.28522e+17, p = 0.0770426

   Solving Stokes system... 0+0 iterations.
      Total relative residual after nonlinear iteration 1: 3.45932e-11, norm of the rhs: 2.52019e+07, newton derivative scaling factor: 0


   Postprocessing:
     RMS, max velocity:                  2.57e-08 m/s, 3.05e-08 m/s
     Pressure min/avg/max:               -1.203e+07 Pa, 5.008e+08 Pa, 1.014e+09 Pa
     Mass fluxes through boundary parts: 0 kg/s, 0 kg/s, -0.8139 kg/s, 0.8139 kg/s

*** Timestep 2:  t=2 seconds
   Skipping temperature solve because RHS is zero.
   Initial Newton Stokes residual = 7.28522e+17, v = 7.28522e+17, p = 0.0770426

   Solving Stokes system... 0+0 iterations.
      Total relative residual after nonlinear iteration 1: 3.45932e-11, norm of the rhs: 2.52019e+07, newton derivative scaling factor: 0


   Postprocessing:
     RMS, max velocity:                  2.57e-08 m/s, 3.05e-08 m/s
     Pressure min/avg/max:               -1.203e+07 Pa, 5.008e+08 Pa, 1.014e+09 Pa
     Mass fluxes through boundary parts: 0 kg/s, 0 kg/s, -0.8139 kg/s, 0.8139 kg/s
     Writing graphical output:           output-nonlinear_channel_flow_velocities_Newton_Stokes/solution/solution-00001

Termination requested by criterion: end time
```

But it seems to work counter productive for the traction case:
```
...
   Skipping temperature solve because RHS is zero.
   The linear solver tolerance is set to 0.1
   Solving Stokes system... 97+0 iterations.
      Total relative residual after nonlinear iteration 33: 7.05902e-11, norm of the rhs: 130.712, newton derivative scaling factor: 1


   Postprocessing:
     RMS, max velocity:                  2.57e-08 m/s, 3.05e-08 m/s
     Pressure min/avg/max:               -0.003074 Pa, 5e+08 Pa, 1e+09 Pa
     Mass fluxes through boundary parts: 0 kg/s, 0 kg/s, -0.8139 kg/s, 0.8139 kg/s
     Writing graphical output:           output-nonlinear_channel_flow_tractions_Newton_Stokes/solution/solution-00000

*** Timestep 1:  t=1 seconds
   Skipping temperature solve because RHS is zero.
   Initial Newton Stokes residual = 3.43965e+11, v = 3.43965e+11, p = 0

   Solving Stokes system... 24+0 iterations.
      Total relative residual after nonlinear iteration 1: 5.53247, norm of the rhs: 1.90298e+12, newton derivative scaling factor: 0

   Skipping temperature solve because RHS is zero.
   Solving Stokes system... 13+0 iterations.
      Total relative residual after nonlinear iteration 2: 1.39863, norm of the rhs: 4.8108e+11, newton derivative scaling factor: 0
...
```
While if I I do it the old way of `current_linearion_point += solution` instead of `current_linearion_point = solution` for the first nonlinear iteration, the result is way better for the traction case. But then the velocity is increasing again with every timestep...
Any idea's what might be happening here?

gassmoeller:
I see what happened. You opened this pull request against the branch that you accidentally pushed to `geodynamics/aspect:fix_newton_boundary_conditions` earlier and not to `geodynamics/aspect:master`. I can also not delete that branch at the moment, because the PR depends on it. Could you close this PR and reopen it compared to geodynamics/master? You can select which branch to compare against, when you initially open the pull request (at the top above the field where you write the initial comment). Alternatively, you can close this branch, I will delete the other branch, and when you reopen afterwards it might compare to master instead.

MFraters:
Ah, that makes sense. I already didn't understand why I couldn't delete the branch. Everything should be clean again. Thanks!

Issue #1967, opened 2017-10-27
Add missing compressibility correction to the Stokes preconditioner


bangerth:
As #1939 and #1965 show, small forgotten terms can lead to suboptimal preconditioners. #1965 adds a missing factor of 2 to the Stokes preconditioner, but the resulting preconditioner is still missing the compressibility correct `-1/3*div(u)I`. Fix this to see whether it makes a difference.

gassmoeller:
#1976 implements this for the compressible case, and the melt case. Still need to fix this for the Newton assembly.

gassmoeller:
Also the Schur complement might need to be adjusted as Wolfgang and Timo discussed in #1976.

PR #1976, opened 2017-10-30
Improve compressible preconditioner

gassmoeller:
This is the improvement mentioned in #1971. In single tests it seems to improve convergence in compressible models on my machine, but I would like to see the tester results. I will work on making the implementation better when we see the test results (currently it simply adds the term to all models, but in incompressible models it will be 0 and we should just not assemble it).


gassmoeller:
I would argue that these results look better than either #1970 or #1971. In particular nearly all compressible models (except `steinberger_viscosity_adiabatic`), and nearly all melt models (except `compression_heating`) see less Stokes iterations by 10-20%. Some melt models benefit dramatically, see `melt_transport_compressible_iterative`, which now need 365 instead of 616 expensive Stokes iterations for the first timestep. To compare with the numbers from https://github.com/geodynamics/aspect/pull/1971#issuecomment-340073378:

```
old (master): 0+277 iterations.
your approach (#1970): 0+816 iterations.
my approach (#1971): 0+193 iterations.
new approach (this PR): 0+130 iterations.
```
The comparison is a bit unfair, because for this model the first nonlinear iteration benefits particularly strongly, the second nonlinear iteration still needs 176 instead of 226 iterations before the change, but that is still a reduction by 25%.

There are some tests that show slightly different results, in particular some melt fractions (they are probably very sensitive to the Stokes solution), and some models with traction boundary conditions (same, I suppose?). Would you agree going forward with this PR?
And credit where credit is due: The original idea to fix the preconditioner for compressible models came from @bangerth, and @jdannberg and @tjhei pointed out this is always necessary for melt models. I just implemented the stuff here :smile:.

gassmoeller:
Oh and @MFraters we should look into how this change translates into the Newton preconditioner. This might be important for compressible models with that as well.

bangerth:
And @MFraters's comment about the missing factor in the Newton solver made me look at the other places.

bangerth:
This looks great. Please apply the diffs and merge! (You can also update my changelog entry and add yourself there.)

gassmoeller:
I applied the diffs, but will keep it open for now so that @tjhei has a chance of taking a look. Will merge tomorrow morning.

MFraters:
@gassmoeller Yes, we should look into that. Could you make an issue for that, or do you want to look at it now? 
As a side note: currently the compressible Newton is not yet stabilized and therefore hardly tested.

tjhei:
Two comments:
1. I was worried that this might make the innerA solver worse (basically deteriorating the convergence of the AMG because it is less "laplace-like") but the opposite seems to be the case when looking at the numbers. 
2. "currently it simply adds the term to all models, but in incompressible models it will be 0 and we should just not assemble it". @gassmoeller you changed the implementation since then to only do it in the compressible case. What happens if you do it also in incompressible models? No difference in iteration numbers? I am asking because discretely and especially during the iteration div u_h != 0, right?

bangerth:
I think that the code as written is correct: you want the preconditioner to match the matrix, so whether or not the discrete solution is pointwise divergence free is not important.

There *is*, however, a question of how one should construct the Schur complement if one had a compressible model. Maybe the mass matrix is no longer the best choice in that case. But that is a separate question, I suppose.

tjhei:
> I think that the code as written is correct: you want the preconditioner to match the matrix, so whether or not the discrete solution is pointwise divergence free is not important.

Of course. I am just curious what happens. 

> There is, however, a question of how one should construct the Schur complement if one had a compressible model. Maybe the mass matrix is no longer the best choice in that case. But that is a separate question, I suppose.

Good point. The correct Schur complement should be ``((2-2/3)\eta p, q)`` (see my Phd thesis). Rene, can you try that?

gassmoeller:
Sure, I will look into it later today or tomorrow. I am currently optimizing something in the melt assembly I noticed during this PR. Will open a pull request soon for that.

bangerth:
> Good point. The correct Schur complement should be |((2-2/3)\eta p, q)| 
> (see my Phd thesis). Rene, can you try that?
Oh, so many pitfalls :-)

I suspect you really meant `(1/[(1-1/3)*eta] p, q)` (i.e., dropping the factor of 2, and using 1/eta instead of eta).

But that is not correct either. The reasoning behind such a choice would be that we would like to reorder the operations in
```
  -div ( eta * (2 div(eps) - 2/3 grad div) )^{-1} grad
```
to
```
  -div ( eta * (Delta - 1/3 Delta) )^{-1} grad
```
to
```
  -div ( (1-1/3) eta * Delta )^{-1} grad
```
to
```
(***) S = 1/[(1-1/3) eta] I
```

But this is not correct. It cannot be correct for a similar argument as made earlier today because we know that for the incompressible case, `div(u)=0`, and so
```
  -div ( eta * (2 div(eps) ) )^{-1} grad
```
the Schur complement is going to be
```
(***) S = 1/[eta] I
```
But the Schur complement cannot be simultaneously be both of the forms marked with `(***)`.

The problem is that in the expression `eta * (2 div(eps) - 2/3 grad div)`, the second term is of size `O(kappa)` compared to the former where `kappa` is the compressibility. It cannot be reordered with the former by just looking at the number of derivatives. In other words, `grad div` is fundamentally of different order of magnitude than `div grad` -- trivially so for incompressible models, but also for compressible models.

So, I think that the correct formula for the preconditioner would be something along the lines of 
```
  S = (1 / (1-k)*eta p, q)
```
where `k` is proportional to `kappa` in such a way that it has no physical units -- i.e., it must contains some scaling factors such as size of domain, time scale, etc, to make it dimensionless.

gassmoeller:
I will merge this for now, lets discuss the other arguments in pull request #1978 and issue #1967. Thanks Wolfgang for discussing Timo's suggestion in more detail, I was about to ask about the inverse. I have to admit that the rest of the argument goes a bit beyond my everyday math (i.e. without spending some time thinking about it).
About the scaling factor `k`: Would a natural scaling factor be something like the pressure difference across the model (i.e. the maximum pressure at the CMB for our case)? Then the factor k would essentially answer the question: By how much does the density change due to compressibility over the model domain? Or would it be more useful to scale it with the pressure difference across one typical cell?

bangerth:
Good question. I don't know.

But I would first try something like "typical pressure changes over typical lengthscales of features of the domain". We already do this when doing the pressure scaling where we use as "lengthscale" 1% of the domain's diameter.

PR #1977, opened 2017-10-30
Optimize compute melt variables

gassmoeller:
While working on #1976 I noticed that `MeltHandler<dim>::compute_melt_variables` took a significant amount of the compute time of the "Solve Stokes" timer section. In fact for `melt_transport_compressible_iterative` it takes about as much time as the Stokes solver itself, although it is simply a postprocessing step to compute the other variables out of the solution. When I looked into the function I found a number of things to optimize. The new version of the function reduces expensive calls to the fe_values object and the gravity model (the function object sometimes used inside is slow). The most intrusive change is that the `u_f` system is now only assembled and distributed for the `u_f` degrees of freedom instead of the whole system. We do this already for all the other equations, and it turns out to be particularly important for the melt models, because their `FESystem` is so large (many more variables than usual). Compared to #1976 the new version of the function is about 5 times faster (-80% runtime), which for the `melt_transport_compressible_iterative` test means the Stokes solver needs 35% less time, and the whole model is now 23% faster than in #1976 (45% compared to master).

Lets see if I broke all melt models on the tester.

### DIFF
--- a/source/simulator/melt.cc
+++ b/source/simulator/melt.cc
@@ -1036,30 +1036,40 @@ namespace aspect
             Assert(melt_outputs != NULL, ExcMessage("Need MeltOutputs from the material model for computing the melt variables."));
             const FEValuesExtractors::Vector fluid_velocity_extractor = this->introspection().variable("fluid velocity").extractor_vector();
 
+            std::vector<Tensor<1,dim> > phi_u_f(dofs_per_cell);
+
             for (unsigned int q=0; q<n_q_points; ++q)
-              for (unsigned int i=0; i<dofs_per_cell; ++i)
-                {
-                  for (unsigned int j=0; j<dofs_per_cell; ++j)
-                    cell_matrix(i,j) += fe_values[fluid_velocity_extractor].value(j,q) *
-                                        fe_values[fluid_velocity_extractor].value(i,q) *
-                                        fe_values.JxW(q);
+              {
+                for (unsigned int i=0; i<dofs_per_cell; ++i)
+                  {
+                    phi_u_f[i] = fe_values[fluid_velocity_extractor].value(i,q);
+                  }
 
-                  const double phi = std::max(0.0, porosity_values[q]);
+                const Tensor<1,dim>  gravity = this->get_gravity_model().gravity_vector(in.position[q]);
+                const double JxW = fe_values.JxW(q);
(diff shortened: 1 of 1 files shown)


source/simulator/melt.cc
```diff
@@ -1036,30 +1036,40 @@ namespace aspect
...
-                      const double K_D = melt_outputs->permeabilities[q] / melt_outputs->fluid_viscosities[q];
-                      const Tensor<1,dim>  gravity = this->get_gravity_model().gravity_vector(in.position[q]);
-                      cell_vector(i) += (u_s_values[q] - K_D * (grad_p_f_values[q] - melt_outputs->fluid_densities[q]*gravity) / phi)
-                                        * fe_values[fluid_velocity_extractor].value(i,q)
-                                        * fe_values.JxW(q);
-                    }
+                for (unsigned int i=0; i<dofs_per_cell; ++i)
+                  {
+                    for (unsigned int j=0; j<dofs_per_cell; ++j)
+                      cell_matrix(i,j) += phi_u_f[j] * phi_u_f[i] *
```
bangerth: This is pre-existing, and not important in this case, but maybe reorder the indices i and j on the rhs here?

source/simulator/melt.cc
```diff
@@ -1036,30 +1036,40 @@ namespace aspect
...
-              for (unsigned int i=0; i<dofs_per_cell; ++i)
-                {
-                  for (unsigned int j=0; j<dofs_per_cell; ++j)
-                    cell_matrix(i,j) += fe_values[fluid_velocity_extractor].value(j,q) *
-                                        fe_values[fluid_velocity_extractor].value(i,q) *
-                                        fe_values.JxW(q);
+              {
+                for (unsigned int i=0; i<dofs_per_cell; ++i)
+                  {
+                    phi_u_f[i] = fe_values[fluid_velocity_extractor].value(i,q);
```
bangerth: if you wanted to optimize this further, then you'd initialize the vector `phi_u_f` above by zeros and added an `if (i is a u_f DoF) to this loop. In fact, a similar statement could be made in the i-j double loop below.

bangerth:
In sum, looks good to me!

jdannberg:
And it's very cool that you could optimize this! I always noticed that this part seemed to take quite some time in my melt models, but never thought much about it. 
Thanks 👍

PR #1978, opened 2017-10-30
[WIP] Always apply compressible preconditioner

gassmoeller:
This is the test for the first comment of Timo on top of #1976. Lets see what the tester says.

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -673,11 +673,10 @@ namespace aspect
                                   std_cxx11::cref (*stokes_assembler),
                                   std_cxx11::_1, std_cxx11::_2, std_cxx11::_3));
 
-        if (material_model->is_compressible())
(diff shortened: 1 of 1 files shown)


gassmoeller:
Well this certainly changes things, but some tests benefit while others need more iterations. How to decide?

bangerth:
Can you rebase on top of all of the other patches we merged?

gassmoeller:
From what I see there are at least as many increases in iteration numbers as decreases. Therefore I will close this for now, until we have more time for testing or a better theoretical understanding for how to proceed.

bangerth:
OK, just to make sure I understand -- we now use the compressible terms only if the model is in fact compressible? This makes sense to me.

gassmoeller:
Yes, we now only subtract the compressible strain-rate terms in compressible models.

Issue #1270, opened 2016-11-10
Write a postprocessor that outputs the elevation from the dynamic topography/mesh movement scheme


bangerth:
From an email by @gassmoeller to Payman Janbakhsh on the mailing list:

surprisingly (also to me) ASPECT currently does not have a postprocessor that outputs a map of the topography of a deformed mesh. The closest thing is the visualization/depth postprocessor that will output the depth of every point of the domain. However, the `depth()` function of the geometry model is bounded to (0,maximal_depth), so it will not output positive topography correctly. If you would be willing to spend some time helping create a postprocessor for topography that would certainly be a useful addition. Let us know if you would have the time and willingness to do so, we could then discuss the details of the implementation (always easier to discuss that first, before writing something and then reworking it many times).

Depending on your intended purpose you would either need to create a general postprocessor that outputs the topography into a separate file (as for example postprocess/heat_flux_map), or a visualization postprocessor that adds another field to the .vtu output of the solution (as for example postprocess/visualization/depth). However, there might be some challenges to make it work for arbitrary model geometries.

ian-r-rose:
This is something that used to exist (it was very similar to
heat_flux_map), but I think got left behind in the big merge of the free
surface implementation. This was mostly an oversight on my part.

The way topography is calculated in the topography postprocessor (which has
always been an unfortunate name due to confusion with dynamic_topography)
is that it references the topography to some datum that depends on the
geometry (i.e., outer_radius for spherical shells, height for boxes). If
the geometry model is not something that the topography postprocessor knows
how to query for the reference surface, then it will fail.

All of this is to say that a merger of the code from the topography
heat_flux_map postprocessors should to the trick.

On Thu, Nov 10, 2016 at 8:57 AM, Wolfgang Bangerth <notifications@github.com

gassmoeller:
I agree with @ian-r-rose that a merged version of topography and heat_flux_map is the easiest  way to implement this.
However, I wonder if it would be better to implement a new function for geometry models that handles the computation itself, because it is a property of the geometry model if something is below or above this datum. This way the postprocessor does not need to know all geometry models. It is a bit unfortunate that `GeometryModel::Interface::depth()` is bounded to `0,max_depth`. Should we introduce a new function `GeometryModel::Interface::unbounded_depth` or `distance_to_surface` (although it is not a distance, I am lacking a good name here)?

bangerth:
[...]

PR #705, opened 2016-01-11
Allow free surfaces with positive topography in the box geometry model

gassmoeller:
This addresses #702 and unifies the depth() function boundary handling of the box geometries with the other geometry models that also do a truncation instead of throwing an exception. 
I do not think this will change any behaviour, because previously the cases that are now truncated were throwing exceptions, so there are no models that now behave differently.

PR #1980, opened 2017-10-31
Topography output to file postprocessor

lkaratun:
Added height_above_original_surface method to material models. Implemented a postprocessor writing topography to a text file. Addresses #1270


ian-r-rose:
This overall looks good to me. I like the API extension for each geometry model computing its own height from the reference surface (that's a lot better than what is done in the `topography` postprocessor).

Your postprocessor does a lot more than the `topography` postprocessor, and I wonder whether it would be better to merge the two into a single postprocessor, which computes the min/max topography for the `statistics` file, and also (optionally) outputs the full topography solution to a file. This is similar, e.g., to the `heat_flux` postprocessor.

lkaratun:
Hi Ian,
thank you for the suggestions. I merged the functionality of the plug-ins. If it looks good to everyone, I'll go ahead and replace the current topography plug-in with the new one.

source/geometry_model/box.cc
```diff
@@ -221,6 +221,16 @@ namespace aspect
 
 
     template <int dim>
+    double
+    Box<dim>::height_above_original_surface(const Point<dim> &position) const
+    {
+      const double d = (position(dim-1)-box_origin[dim-1]) - extents[dim-1] ;
+      return std::min (std::max (d, -maximal_depth()), maximal_depth());
```
MFraters: what is your reason for limiting the return of height above the surface to the maximal depth? 

Although I can imagine that the mesh will be probably horribly deformed, the purpose of this function seem to be just for visualisation. I therefore think that  it should just return what the height above the surface is, so that it can output the true height. Currently it will show caped off values in that file, which might not be expected by the user. 

The same argumentation can be used for also not limiting it to the -minus depth. The reason why we limit the depth is that we know depth should not be smaller then zero and larger then the maximum depth. I do not think that is the case for the height above a original/reference surface.

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+      //Calculate min/max topography across all processes
+      const double max_topography = Utilities::MPI::max(local_max_height, this->get_mpi_communicator());
+      const double min_topography = Utilities::MPI::min(local_min_height, this->get_mpi_communicator());
+
+      //Write results to statistics file
+      statistics.add_value ("Minimum topography (m)",
+                            min_topography);
+      statistics.add_value ("Maximum topography (m)",
+                            max_topography);
+      //{
```
MFraters: remove this if it isn't used. Same for the \\} below

source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -600,6 +600,16 @@ namespace aspect
 
     template <int dim>
     double
+    EllipsoidalChunk<dim>::height_above_original_surface(const Point<dim> &position) const
+    {
+      AssertThrow(false, ExcMessage("The topography postprocessor does not recognize the geometry model. "
```
MFraters: You can also support this by using the topo plugin and to get the topo height and adding it to the manifold.pull_back(position)[2] (this is the height above the reference surface - topo).

include/aspect/geometry_model/box.h
```diff
@@ -109,6 +109,9 @@ namespace aspect
         double depth(const Point<dim> &position) const;
 
         virtual
+        double height_above_original_surface(const Point<dim> &position) const;
```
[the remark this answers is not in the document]
lkaratun: I used the name that Wolfgang suggested in #1270. I don't personally care what name to use, but would rather not change back and forth before the decision is made. Is there a person in charge of naming stuff? =)

include/aspect/postprocess/topography_to_file.h
```diff
@@ -0,0 +1,103 @@
...
+         * Whether or not to produce text files with topography values
+         */
+        bool write_to_file;
+
+        /**
+        * Whether or not to output topography statistics to screen and statistics file
+        */
+        bool write_statistics;
+
+
```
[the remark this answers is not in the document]
lkaratun: Sure. I thought they would be removed by astyle, guess it's not the case.

include/aspect/geometry_model/box.h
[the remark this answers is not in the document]
ian-r-rose: I agree with Menno here.

source/geometry_model/box.cc
lkaratun: Sure, I'll take out the max height limit. As for the low limit, I think it can't be less than -extent[dim-1] because we're calculating topography from "sea level", not from original topography.

source/postprocess/topography_to_file.cc
lkaratun: Sure. I copied this part from topography.cc and wasn't sure if it was just me or the curly braces were not needed.

source/geometry_model/ellipsoidal_chunk.cc
lkaratun: Sorry I didn't quite get what you mean. I'm not really familiar with the ellipsoidal chunk model, should the function return just -manifold.pull_back(position)[2] ?

lkaratun:
Hi Menno, 

thank you for the comments, I made some modifications.  
I tested the plug-in with box geometry, it looks fine. Is this enough for tests or do I need to run a simulation for each of the geometry models? Also, after I run the model, do I just eyeball the result, or do I need to compare it to something?

source/geometry_model/box.cc
MFraters: Good point. If you know for sure it shouldn't every be less then -extent[dim-1], I think you  should use an assert (see #705 on how this was done before for the depth function).

source/geometry_model/ellipsoidal_chunk.cc
MFraters: I think you should be able to get it working with the following code (although I haven't tested it):
```
const Point<dim> phi_theta_h = manifold.pull_back(position);
Point<dim-1> phi_theta;
const double rad_to_degree = 180/numbers::PI;
if (dim == 3)
  phi_theta = Point<dim-1>(phi_theta_h[0] * rad_to_degree,phi_theta_h[1] * rad_to_degree);
return phi_theta_h[2] + topography->value(phi_theta).
```

If it doesn't work, you can leave it as it is if you want. It can be added in an other pull request.

MFraters:
I think in this case a eyball norm is fine, but make sure they all work ;)

As for testing, what you could do is test all the geometry models with just one test by using global functions like in tests/prm_polygon_in_out.cc. You could initialize every topography, displace it with the initial topography and then see if the output is what you expect. Would that work?

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+              {
+                if ( cell->face(face_no)->boundary_id() != relevant_boundary)
+                  continue;
+
+                face_vals.reinit( cell, face_no);
+
+                for (unsigned int corner = 0; corner < face_corners.size(); ++corner)
+                  {
+                    Point<dim> vertex = face_vals.quadrature_point(corner);
+                    double elevation = this->get_geometry_model().height_above_original_surface(vertex);
```
gassmoeller: make this `const`

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+            if (cell->face(face_no)->at_boundary())
+              {
+                if ( cell->face(face_no)->boundary_id() != relevant_boundary)
+                  continue;
+
+                face_vals.reinit( cell, face_no);
+
+                for (unsigned int corner = 0; corner < face_corners.size(); ++corner)
+                  {
+                    Point<dim> vertex = face_vals.quadrature_point(corner);
```
gassmoeller: make this `const Point<dim>`

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+      // always produce data during the first time step
+      if (std::isnan(last_output_time))
+        {
+          last_output_time = this->get_time() - output_interval;
+        }
+
+      // return if graphical output is not requested at this time
+      if ((this->get_time() < last_output_time + output_interval)
+          && (this->get_timestep_number() != 0))
+        return std::pair<std::string,std::string>();
```
gassmoeller: I think this is wrong. You still want to return the statistics, even if you do not write the file output. Just copy the return statement at the end to here.

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+
+
+
```
gassmoeller: Just a style question, but please remove some of these empty lines. In general we use 3 empty lines between function implementations, 1 empty line between function or variable declarations (in header files), and 1 empty line to separate logical blocks of statements inside of functions. If a bigger separation is necessary you can add a comment line describing what is the next logical block. Unfortunately, it is not possible to set up astyle to do this for us, but we try to follow it manually.

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+
+      for (unsigned int i=0; i<stored_values.size(); ++i)
+        {
+          output_file << stored_values[i].first
+                      << ' '
+                      << stored_values[i].second
+                      << std::endl;
```
gassmoeller: Since you just write them in the order you put them into the vector anyway, would it make sense to immediately write the pair into the ostringstream output_file in line 82? That way you would save the vector of pairs.

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+    {
+      prm.enter_subsection("Postprocess");
+      {
+        prm.enter_subsection("Topography");
+        {
+          prm.declare_entry ("Output to file", "false",
+                             Patterns::List(Patterns::Bool()),
+                             "Whether or not to write topography to a text file named named"
+                             "'topography.NNNNN' in the output directory");
+
```
gassmoeller: remove this empty line

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(TopographyToFile,
+                                  "topography to file",
+                                  "A postprocessor that outputs topography into text "
+                                  "files named `topography.NNNNN' in the output directory, "
+                                  "where NNNNN is the number of the time step."
+                                  "\n\n"
+                                  "The file format then consists of lines with Euclidean coordinates "
+                                  "followed by the corresponding topography value.")
```
gassmoeller: Please update / merge with the description of the topography postprocessor.

source/postprocess/topography_to_file.cc
```diff
@@ -0,0 +1,290 @@
...
+    template <int dim>
+    void
+    TopographyToFile<dim>::parse_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Postprocess");
+      {
+        prm.enter_subsection("Topography");
+        {
+          write_to_file        = prm.get_bool ("Output to file");
+          write_statistics       = prm.get_bool ("Output statistics");
```
gassmoeller: I think this parameter is currently not used. Either implement it, or remove the parameter.

bangerth:
If @gassmoeller says that a name is ok with him, then that shall be good enough for me as well! :-)

lkaratun:
Hi Menno,
I was wondering if there is any documentation for prm_polygon_in_out.cc or a similar script, of just for the approach used? I can't figure out how it works =(

MFraters:
@lkaratun As far as I know there is not documentation about it. What happens is the following. When ASPECT starts, one of the first things it does is load shared libraries. When this shared library is loaded by ASPECT you can take over by decaring a global variable (in this case a int i) and assign it a values based on a function f(). Now the code of f() will be executed to determine the value of i. From this point onward we can do whatever we want in f() and in this case I initialize the initial topography and pass some values to it so see if returns the solution I expect. 

In your case you would have to initialize all the geometry models and the initial topography of those geometry models. I took a look at the geometry model code and at least in the case of the box I now see that this approach will be very difficult because the box model tries to access the simulator to get the initial topography model. But since the simulator doesn't exists yet in this case, I would suggest we should let this option go.

So I guess you will have to make a test for each geometry model for which you implemented this.

lkaratun:
Hi everyone,
I was trying to make a test for the sphere geometry model, but couldn't figure out how to do it. Initial topography is not supported for it, and when I tried to make a model with a small dense rigid block near the north pole, it failed to converge. I checked the cookbooks and it seems there aren't any that feature free surface with a non-box geometry model. What can I do in this case?

bangerth:
Have you looked into the `tests/` directory? If you do something like
```
  grep sphere tests/*prm
```
you'll get a list of tests in which the name `sphere` appears. With a bit more filtering, you should be able to find out which of these use the sphere as the geometry, and then among those you can look for some that may use the free surface. Or you first look for free surface tests.

It's quite possible that you can't find anything that matches what you need. But let us know what you find (or don't find).

lkaratun:
Hi Wolfgang,
there aren't any tests that use Free surface with a sphere geometry unfortunately. Should I just drop the sphere geometry support from the postprocessor?

bangerth:
Yes, I think that's a reasonable choice. I don't think that it makes much of a difference whether we use a sphere, or a spherical shell. The latter is definitely tested. But if you're not sure it works with the sphere, then just drop it.

ian-r-rose:
To comment briefly on the sphere case:

The free surface itself has no problem with the spherical geometry, but there is a big caveat: once you have a sphere with a free surface, there is a big nullspace in the Stokes system. Specifically, the solution is invariant to translations and rotations, and is thus highly nonunique. 

Even worse, there is a nasty compatibility condition for the right-hand-side: it cannot have any net forces or torques. For a purely self-gravitating sphere there should be no torque or force, but with an imposed gravity model it is easy to accidentally have them.

So in order to run a model with a sphere you need to project forces and torques out of the RHS, and remove the nullspace from the ultimate solution. This is possible. but it is *extremely* fiddly.

gassmoeller:
Then let's just disallow the sphere case for now. There is no point in implementing very complicated features until somebody actually needs them. Thanks for the explanation @ian-r-rose.

MFraters:
@lkaratun Looks good, thanks for doing all this! 
/run-tests

lkaratun:
Thanks everyone for the answers! I made the modifications. 
Not sure about how to fix the checker errors -- it seems it's expecting a blank line instead of a blank line?

bangerth:
The diffs are here for the `statistics` file -- you can download them and apply them by hand using `patch -p1 < changes-gcc.diffs`.

For the `box.cc` file, I bet you have two lines that are empty but have spaces, whereas the indentation checker wants the lines to be completely empty. Can you update that as well?

source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -600,6 +600,17 @@ namespace aspect
 
     template <int dim>
     double
+    EllipsoidalChunk<dim>::height_above_reference_surface(const Point<dim> &position) const
+    {
+      AssertThrow(false, ExcMessage("Function height_above_reference_surface is not yet implemented"
+                                    "for the ellipsoidal chunk geometry model."
+                                    "Consider using a box, spherical shell, or chunk.") );
```
bangerth: C++ takes successive `"quote enclose" "d strings"` and just concatenates them. Here, this means that you'll get an error message that doesn't have a space after `implemented` and `for the ellipsoidal`, and the same between the second and third line. Can you just add a space at the end of the first and second lines?

source/postprocess/topography.cc
```diff
@@ -40,55 +43,34 @@ namespace aspect
...
-          reference_height = gm->outer_radius();
-          vertical_gravity = false;
-        }
-      else
-        {
-          AssertThrow(false, ExcMessage("The topography postprocessor does not recognize the geometry model. "
-                                        "Consider using a box, spherical shell, sphere, or chunk.") );
+          AssertThrow(false, ExcMessage("Topography postprocessor is not yet implemented "
+                                        "for the sphere geometry model. "
+                                        "Consider using a box, spherical shell, or chunk.") );
```
bangerth: you got it right here!

source/postprocess/topography.cc
```diff
@@ -101,57 +83,196 @@ namespace aspect
...
+              // by a \0 character. write only this part by outputting it as a
+              // C string object, rather than as a std::string
+              file << tmp.c_str();
+            }
+        }
+      else
+        // on other processors, send the data to processor zero. include the \0
+        // character at the end of the string
+        {
+          MPI_Send (&output_file.str()[0], output_file.str().size()+1, MPI_CHAR, 0, mpi_tag,
```
bangerth: you will get an out-of-bounds error here because you assume that `output_file.str()` points to a string that is terminated by a `\0` character that `str().size()` does not actually count. This may be the case, or maybe not -- I don't think that you can count on that. So I think you really should only transmit `str().size()` characters. On the receiving side, you will then have to make sure you deal with the fact that no `\0` was transmitted -- or you append a `\0` character before sending and then you don't have to deal with it at the receiving end.

source/postprocess/topography.cc
```diff
@@ -101,57 +83,196 @@ namespace aspect
...
-          {
-            statistics.set_precision (columns[i], 8);
-            statistics.set_scientific (columns[i], true);
-          }
+        prm.enter_subsection("Topography");
+        {
+          prm.declare_entry ("Output to file", "false",
+                             Patterns::List(Patterns::Bool()),
+                             "Whether or not to write topography to a text file named named"
+                             "'topography.NNNNN' in the output directory");
```
bangerth: missing space at the end of the first line.

source/postprocess/topography.cc
```diff
@@ -101,57 +83,196 @@ namespace aspect
...
   {
     ASPECT_REGISTER_POSTPROCESSOR(Topography,
                                   "topography",
-                                  "A postprocessor intended for use with a free surface.  After every step, "
+                                  "A postprocessor intended for use with a free surface.  After every step "
                                   "it loops over all the vertices on the top surface and determines the "
                                   "maximum and minimum topography relative to a reference datum (initial "
                                   "box height for a box geometry model or initial radius for a sphere/"
-                                  "spherical shell geometry model).  Outputs topography in meters")
+                                  "If 'Topography.Output to file' is set to true, also outputs topography "
```
bangerth: This doesn't match any more -- there ought to be something after the `/` and a closing parenthesis.

MFraters:
Could you rebase this on to the newest master (you will probably have to fixes your test again, because the output of all solvers changed because #1983 was merged yesterday), and then squash your commits into a few self contained commits (I would propose one commit for each geometry model and one for each test)?

MFraters:
Ah, I now see that you are using IMPES. Then it indeed didn't change. You also won't need the line `set Max nonlinear iterations = 50 ` in your input file then. Other then that, I am fine with it being merged :)

lkaratun:
Sure, fixed.

PR #1981, opened 2017-10-31
Fix deprecation warning

gassmoeller:
This primarily fixes a deprecation warning with deal.II 9.0.pre.

Note that I do not set the manifold ids for all cells yet. The reason is that the resulting mesh looks strange, and I wanted an opinion on that. The figure below shows from left to right the mesh with 8.5.0, the mesh with 9.0.pre with this PR (no difference), and the mesh with 9.0.pre with the spherical manifold attached to all cells. The mesh on the right has some pretty distorted cells. Is there some best practice for how to prescribe manifolds to this case?

mesh

### DIFF
--- a/source/geometry_model/sphere.cc
+++ b/source/geometry_model/sphere.cc
@@ -38,8 +38,14 @@ namespace aspect
       GridGenerator::hyper_ball (coarse_grid,
                                  Point<dim>(),
                                  R);
+
+#if DEAL_II_VERSION_GTE(9,0,0)
+      coarse_grid.set_manifold(0,spherical_manifold);
+      coarse_grid.set_all_manifold_ids_on_boundary(0);
+#else
       static const HyperBallBoundary<dim> boundary_ball(Point<dim>(), R);
       coarse_grid.set_boundary (0, boundary_ball);
+#endif
     }
 
 
--- a/include/aspect/geometry_model/sphere.h
+++ b/include/aspect/geometry_model/sphere.h
@@ -129,6 +129,12 @@ namespace aspect
          */
         double R;
 
+#if DEAL_II_VERSION_GTE(9,0,0)
+        /**
+         * The manifold that describes the geometry.
+         */
+        const SphericalManifold<dim> spherical_manifold;
+#endif
     };
   }
 }


bangerth:
For the full sphere, you definitely can't use a spherical manifold everywhere. I think the results section of step 6 has a discussion of how to create a good mesh for the sphere.

gassmoeller:
Thanks for the hint. I looked into the discussion in step-6, but currently I do not think it is worth complicating the geometry model in order to achieve a better mesh. Lets keep that for a time somebody actually needs it. Would you be ok with this PR as it is?

tjhei:
1. You could put the manifold into the .cc as a static member like we do for the boundary.
2. You can not use the spherical manifold on the whole mesh. An easy fix would be to only use the manifold on the boundary faces.

source/geometry_model/sphere.cc
```diff
@@ -38,8 +38,13 @@ namespace aspect
       GridGenerator::hyper_ball (coarse_grid,
                                  Point<dim>(),
                                  R);
+
+#if DEAL_II_VERSION_GTE(9,0,0)
+      coarse_grid.set_manifold(0,spherical_manifold);
```
bangerth: But doesn't this imply setting it for all objects? Wouldn't you have to give a different manifold id to only the faces at the boundary, and then attach the manifold to the triangulation for that specific manifold id?

I have to admit that I don't quite recall what the default manifold id of objects is. Your code seems to suggest that it is something different than zero, and that for reasons I don't recall, `GridGenerator::hyper_ball()` only sets the manifold id of boundary faces to zero. Is this correct?

tjhei: > But doesn't this imply setting it for all objects? 

Yes. The default manifold id is 0 for every object and currently no ``GridGenerator`` function touches the manifold id.

gassmoeller:
Sorry for the confusion, I had tested different versions and uploaded the wrong one. The current one should work.
As for your concerns:
- The default manifold id is `invalid_manifold_id` (i.e. (unsigned int) -1), so the previous version did create the manifold, but did not assign it to anything. The current version only assigns it to the faces at the boundary, which is exactly what the old code did.
- Is it necessary to create the manifold as a static member? I somehow dislike that use (it makes no sense to me). The order of construction / destruction should ensure that the manifold is still around when the triangulation is destroyed.

tjhei:
> Is it necessary to create the manifold as a static member? I somehow dislike that use (it makes no sense to me).

No and I am fine with it as is. My suggestion was mostly to keep both version as similar as possible.

gassmoeller:
Is this ready to merge then?

tjhei:
Yeah, I guess so. ;-)

bangerth:
I had to double check, but indeed the geometry model lives longer than the triangulation -- so using non-`static` member variable is definitely allowed.

PR #1982, opened 2017-11-01
parameter names without underscores

tjhei:
prm names with underscores are not our style and break the manual (names
need to be escaped, but not in the index keys).
Also some other minor changes.

### DIFF
--- a/source/boundary_temperature/dynamic_core.cc
+++ b/source/boundary_temperature/dynamic_core.cc
@@ -101,21 +101,21 @@ namespace aspect
                              Patterns::Double (),
                              "Temperature at the inner boundary (core mantle boundary) at the "
                              "beginning. Units: K.");
-          prm.declare_entry ("dT_dt", "0",
+          prm.declare_entry ("dT over dt", "0",
                              Patterns::Double (),
                              "Initial CMB temperature changing rate. Units: K/year");
-          prm.declare_entry ("dR_dt", "0",
+          prm.declare_entry ("dR over dt", "0",
                              Patterns::Double (),
                              "Initial inner core radius changing rate. Units: km/year");
-          prm.declare_entry ("dX_dt", "0",
+          prm.declare_entry ("dX over dt", "0",
                              Patterns::Double (),
                              "Initial light composition changing rate. Units: 1/year");
           prm.declare_entry ("Core density", "12.5e3",
                              Patterns::Double (),
-                             "Density of the core. Units: kg/m^3");
+                             "Density of the core. Units: $kg/m^3$");
           prm.declare_entry ("Gravity acceleration", "9.8",
                              Patterns::Double (),
-                             "Gravitation acceleration at CMB. Units: m/s^2.");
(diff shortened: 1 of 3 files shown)


tjhei:
What is our policy for putting the units into math-mode? We currently have a mix of both throughout. It is necessary for ``^3``, but looks ugly for variables with more than one character unless we force them back to text mode, which in turn looks ugly in the .prm. Thoughts?

bangerth:
These are all parameters in new models, right?

As for style, I had to look up something in the .prm file that's written into the output directory. It's thousands of lines long. I think that at this point, everyone just looks stuff up in the manual, so let's make that the default.

bangerth:
I think this was @SiqiZhang 's patch, right? I'm CC:ing him here to make sure he hears about it.

jdannberg:
I have to say that I still use the parameters.prm file rather than the manual, and we should also make sure that this will look okay in the GUI (as that is what I would probably use in tutorials), and that also uses the style from the .prm file. I agree that the math-mode units don't look great in the pdf, but I think it's more important to have something that's readable in the parameter file than to have the perfect style in the manual.

source/boundary_temperature/dynamic_core.cc
```diff
@@ -128,14 +128,14 @@ namespace aspect
                              "The max iterations for nonliner core energy solver.");
           prm.declare_entry ("Core heat capacity", "840",
                              Patterns::Double (0),
-                             "Heat capacity of the core. Unit: J/kg/K");
+                             "Heat capacity of the core. Units: $J/kg/K$");
           prm.declare_entry ("K0", "4.111e11",
                              Patterns::Double (0),
                              "Core compressibility at zero pressure. "
                              "Referring to Nimmo et al. (2004) for more detials.");
```
jdannberg: detials --> details.

jdannberg:
I think you forgot the parameters in section "Geotherm parameters" and "Radioactive heat source". At least the parameter "Tm2" should also have units in math mode (1/TPa^2). 
Looks good other wise, thanks!

@SiqiZhang: It seems there are a number of parameters without units in the declaration section (for example, for K0, Beta_c and Delta). It would be great if you could add them if you find the time

tjhei:
> At least the parameter "Tm2" should also have units in math mode (1/TPa^2).

Agreed. Let's leave the other edits (parameter names, units, formatting) to a follow-up PR. I want to be able to compile the manual again first.

jdannberg:
Sure! If you've tested that that manual compiles now with the changes you made, merge yourself.

SiqiZhang:
Thanks guys for the cleaning up work. I didn't know underscores will cause such trouble.
I can do a following up PR to deal with other issue raised.

tjhei:
> I didn't know underscores will cause such trouble.

No problem. It is actually a combination of two things:
1. We prefer to spell out parameter names in ASPECT. This makes parameter files more readable too (compare "k_c" with "core conductivity" for example).
2. Our parameter.tex exporting is not robust enough to correctly escape ``_`` (this could be fixed in the next release)

Issue #1964, opened 2017-10-26
Unify solver output


MFraters:
See #1850 and #1433:

I think we should also unify these outputs, so that it is easier to grep them. Some for example start with 'Total relative ... 'and others with 'Relative ... '. This could then for example become something like 'Nonlinear residual (Stokes) for iteration ... ' or 'Nonlinear residual (Total) for iteration ...' etc. And then always for every timestep end with Final residual for timestep ... . 

But as stated in #1850 this might be more complex then it looks and it will be a lot of work to update all the tests, so I think it is a good idea to first discuss here.

jdannberg:
I think if we just want to change the output for the log file, that should not be too difficult. The main problem I had with unifying the output of the different solver schemes (after #1433) was the output for the statistics file. So I can have a look at this in the next few days. 

But of course it would be great if we could agree on a scheme for the output before. I like @MFraters suggestion 'Nonlinear residual (Total/Stokes/etc.) for iteration ...'. We could also leave in the 'relative', making it 'Relative nonlinear residual (Total/Stokes/etc.) for iteration ...', if we think that's important.

gassmoeller:
Was this resolved by #1983?

gassmoeller:
I think we resolved this with all the changes to the nonlinear solver schemes of the last months. Feel free to reopen if the problem persists.

PR #1983, opened 2017-11-01
Unify the screen output of the different solver schemes

jdannberg:
This addresses #1964. The line of each nonlinear solver scheme that outputs the nonlinear residual now starts with 'Nonlinear residual ...', and then specifies the system that is being solved, following @MFraters suggestion. This should make it easier to grep that line from the screen output of different models. 
In addition, there is now always one empty line between one nonliear iteration and the next, and two empty lines between the last iteration and the Postprocessing output.

### DIFF
--- a/source/simulator/solver_schemes.cc
+++ b/source/simulator/solver_schemes.cc
@@ -291,8 +291,9 @@ namespace aspect
         const double relative_stokes_residual =
           assemble_and_solve_stokes(nonlinear_iteration == 0, &initial_stokes_residual);
 
-        pcout << "      Relative Stokes residual after nonlinear iteration " << nonlinear_iteration+1
+        pcout << "      Relative nonlinear residual (Stokes system) after nonlinear iteration " << nonlinear_iteration+1
               << ": " << relative_stokes_residual
+              << std::endl
               << std::endl;
 
         if (parameters.run_postprocessors_on_nonlinear_iterations)
@@ -337,7 +338,7 @@ namespace aspect
           assemble_and_solve_stokes(nonlinear_iteration == 0, &initial_stokes_residual);
 
         // write the residual output in the same order as the solutions
-        pcout << "      Relative nonlinear residuals: " << relative_temperature_residual;
+        pcout << "      Relative nonlinear residuals (temperature, compositional fields, Stokes system): " << relative_temperature_residual;
         for (unsigned int c=0; c<introspection.n_compositional_fields; ++c)
           pcout << ", " << relative_composition_residual[c];
         pcout << ", " << relative_stokes_residual;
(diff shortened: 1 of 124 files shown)


source/simulator/solver_schemes.cc
```diff
@@ -291,8 +291,9 @@ namespace aspect
         const double relative_stokes_residual =
           assemble_and_solve_stokes(nonlinear_iteration == 0, &initial_stokes_residual);
 
-        pcout << "      Relative Stokes residual after nonlinear iteration " << nonlinear_iteration+1
+        pcout << "      Nonlinear residual (Stokes system only) after nonlinear iteration " << nonlinear_iteration+1
```
MFraters: I think it is better to name them all `Relative nonlinear residual`, like you suggested in #1964.

jdannberg:
Oh, you're right, I didn't think about a way to only grep the final residual of each time step. I don't think there's a way to do that at the moment. 

What exactly would you want to do with the residual? Something like plotting it in gnuplot to make sure there was convergence for each time step? I guess what I am asking is: Would it make more sense to have that info in the statistics file rather than the screen output? We already have the number of iterations in the statistics file, so maybe that would be a good place to put it, and we could make that output optional? 

As it would be a new functionality anyway, I guess it would also make sense to put it into a separate pull request.

MFraters:
Yes, I guess that you are right that it is better to put it into the statistics file since the information might not be that interesting for everyone. If you add the Relative thing, I am fine with it being merged.

source/simulator/solver_schemes.cc
```diff
@@ -291,8 +291,9 @@ namespace aspect
         const double relative_stokes_residual =
           assemble_and_solve_stokes(nonlinear_iteration == 0, &initial_stokes_residual);
 
-        pcout << "      Relative Stokes residual after nonlinear iteration " << nonlinear_iteration+1
+        pcout << "      Relative  nonlinear residual (Stokes system only) after nonlinear iteration " << nonlinear_iteration+1
```
gassmoeller: Is it important to differentiate between `Stokes system` and `Stokes system only`? The residual is calculated the same way after all.

gassmoeller:
Apart from my one question this looks good. I agree with putting the residual into the statistics file as a separate issue.

source/simulator/solver_schemes.cc
jdannberg: No, I guess it's not. I thought it would be a nice way to remind the user of the solver scheme used, but I guess that's easy to see anyway due to the equations that are solved. So should I just call it Stokes system in both cases?

jdannberg:
I opened an issue about putting the residual into the statistics file: #1986.

source/simulator/solver_schemes.cc
gassmoeller: Yes that seems more consistent.

jdannberg:
Okay, I think I addressed all of your comments.

source/simulator/solver_schemes.cc
```diff
@@ -291,8 +291,9 @@ namespace aspect
         const double relative_stokes_residual =
           assemble_and_solve_stokes(nonlinear_iteration == 0, &initial_stokes_residual);
 
-        pcout << "      Relative Stokes residual after nonlinear iteration " << nonlinear_iteration+1
+        pcout << "      Relative  nonlinear residual (Stokes system) after nonlinear iteration " << nonlinear_iteration+1
```
MFraters: one space too much between Relative and nonlinear

jdannberg: Thanks for catching that!

PR #1988, opened 2017-11-03
Fix solver history for deal.II-dev

gassmoeller:
This closes the remaining issue in #1912. The solver_history test should now work correctly with deal.II 9.0.pre.

### DIFF
--- a/source/postprocess/stokes_residual.cc
+++ b/source/postprocess/stokes_residual.cc
@@ -107,7 +107,11 @@ namespace aspect
 #endif
 
       // If there were expensive iterations add them.
-      if (solver_control_expensive.last_step() > 0)
+#if DEAL_II_VERSION_GTE(9,0,0)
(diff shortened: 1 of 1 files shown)


source/postprocess/stokes_residual.cc
```diff
@@ -107,7 +107,11 @@ namespace aspect
 #endif
 
       // If there were expensive iterations add them.
-      if (solver_control_expensive.last_step() > 0)
+#if !DEAL_II_VERSION_GTE(9,0,0)
+      if (solver_control_expensive.last_step() != 0)
+#else
+      if (solver_control_expensive.last_step() != numbers::invalid_unsigned_int)
```
bangerth: I find this `! DEAL_II_VERSION_GTE(9,0,0)` hard to read. Can you just drop the `!` and switch the two branches?

OK to merge yourself then.

PR #1992, opened 2017-11-03
fix tests/viz_plugin_dependency

### LINKED ISSUE #1991: [deal 9.0.0] tests/viz_plugin_dependency.prm fails
fails with
```
An error occurred in line <41> of file <../source/numerics/data_postprocessor.cc> in function
    void dealii::DataPostprocessor<dim>::evaluate_vector_field(const dealii::DataPostprocessorInputs::Vector<dim>&, std::vector<dealii::Vector<double> >&) const [with int dim = 2]
The violated condition was: 
    false
```
(we need to replace the evaluate_vector_field() function, working on a fix)

tjhei:
with deal.II master, closes #1991

### DIFF
--- a/tests/viz_plugin_dependency.cc
+++ b/tests/viz_plugin_dependency.cc
@@ -28,12 +28,8 @@ namespace aspect
 
           virtual
           void
-          compute_derived_quantities_vector (const std::vector<Vector<double> > &,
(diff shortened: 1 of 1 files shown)


bangerth:
That seems obvious. 

Issue #1989, opened 2017-11-03
[deal 9.0.0] tests/stokes_residual.prm


tjhei:
on deal.II master stokes_residual.prm fails with
```
An error occurred in line <161> of file <../source/lac/solver_control.cc> in function
    const std::vector<double>& dealii::SolverControl::get_history_data() const
The violated condition was: 
    history_data.size() > 0
Additional information: 
    The SolverControl object was asked for the solver history data, but there is no data. Possibly you requested the data before the solver was run.
```

gassmoeller:
This should be solved by #1988

gassmoeller:
Reopening, because now stokes_residual_cheap fails.

gassmoeller:
Fixed by #1993.

PR #1993, opened 2017-11-04
Fix cheap solver history

gassmoeller:
This should resolve the remaining issues of #1989.


bangerth:

PR #1707, opened 2017-05-16
Newton merge 5: adding the keystone to get access to the Newton pipeline

MFraters:
This pull request can be concidered to be the keystone of the Newton solver in ASPECT. It adds a nonlinear solver to the core and solver so that the Newton solver can actually be used. It is still in a WIP state (I currently only know that it compiles), but the structure should not change too much any more. 

Will look further into this tomorrow.

Issue #302, opened 2015-04-15
Prevent a non-ending loop of nonlinear iterations


gassmoeller:
Apparently in an ASPECT version from September it is possible to choose the solver tolerance in a way, that lets the linear stokes solver skip any calculation (it thinks it has converged), but the nonlinear solver still triggers new iterations (because it thinks it is not converged yet), which leads to an endless loop of (not)iterations. We should investigate if this is still possible after the changes to the residual a few weeks back.

jdannberg:
As I understood it, this was only a problem in the iterated_Stokes solver scheme, and after having a look at it I found that for that scheme we still use the old method: core.cc:1627 `initial_stokes_residual = stokes_residual`.
Should we change this to make it consistent with the iterated IMPES and use `initial_stokes_residual = compute_initial_stokes_residual()` instead?

bangerth:
Can anyone share an input file that demonstrates the issue?

jdannberg:
@MFraters, maybe you could share an input file?
And show us what changes to the code you made, if you do not use the original version of the iterated Stokes?

gassmoeller:
I did not observe this behavior in any model so far, and we do not have a way to reproduce it. I will close this issue. Feel free to reopen if anyone has an example case for this.

gassmoeller:
This appears again in #1996. Reopen the issue.

PR #1996, opened 2017-11-07
fix bug in Newton solver which increases velocities and some other bugs and stuff

MFraters:
This is a continuation of #1975 and #1974. 

This fixes several bugs, makes optimization and adds documentation, all mostly related to time stepping. The fixes/optimizations include fixing the velocity boundary conditions when doing time steps, resetting the linear tolerance after the time step and using the previous solution as an initial guess.

The tests for the Newton solve have been expended so that they are are now also very sensitive to changes relate to time stepping.


source/simulator/helper_functions.cc
```diff
@@ -1805,27 +1805,36 @@ namespace aspect
   double
   Simulator<dim>::compute_initial_newton_residual(LinearAlgebra::BlockVector &linearized_stokes_initial_guess)
   {
+    // Store the values of the current_linearization_point and linearized_stokes_initial_guess so we can reset them again.
     LinearAlgebra::BlockVector temp_linearization_point = current_linearization_point;
+    LinearAlgebra::BlockVector temp_linearized_stokes_initial_guess = linearized_stokes_initial_guess;
```
gassmoeller: I am confused about two things in this function:
- `linearized_stokes_initial_guess` after the function will be the same as before the function (you reset it at the end). Can it be `const` and you hand over `temp_linearized_stokes_initial_guess` to `denormalize_pressur` instead?
- You are only interested in the `system_rhs` at least that is the only result from the call to `assemble_stokes_system` that seems to be used. Why do we need to assemble the matrix for that?

source/simulator/solver_schemes.cc
```diff
@@ -563,86 +581,110 @@ namespace aspect
...
+        double test_residual = residual;
+        if (nonlinear_iteration == 0)
           {
-            current_linearization_point = backup_linearization_point;
-
-            LinearAlgebra::BlockVector search_direction = solution;
-
-            search_direction *= lambda;
+            /**
+             * The first nonlinear iteration we are computing the whole system in a non-defect corected Picard way,
```
gassmoeller: corected -> corrected

source/simulator/solver_schemes.cc
```diff
@@ -429,9 +429,20 @@ namespace aspect
   template <int dim>
   void Simulator<dim>::solve_newton_stokes ()
   {
+    // The values of the pressure in current_linearization_point end up being
+    // different then what they where at the end of this loop, and different
+    // from solution in a bad way for the models I tested. Therefore I first
+    // replace current_linearization_point by the solution.
```
gassmoeller: Before starting the nonlinear loop the current linearization point will be the extrapolation of the two available old solutions to the current time (see core.cc:1724-1738). This is part of the BDF2 time stepping scheme. I am not sure why this does not work for your test problem, but I am pretty sure that the convergence order of the scheme depends on this fact. Can you specify `a bad way`?

source/simulator/solver_schemes.cc
```diff
@@ -515,12 +519,27 @@ namespace aspect
...
+          {
+            denormalize_pressure (last_pressure_normalization_adjustment,
+                                  linearized_stokes_initial_guess,
+                                  current_linearization_point);
+          }
 
         if (nonlinear_iteration <= 1)
-          compute_current_constraints ();
+          {
+            // So this should only happend on nonlinear iteration 0 and 1.
```
gassmoeller: happend -> happen

source/simulator/helper_functions.cc
```diff
@@ -1805,27 +1805,36 @@ namespace aspect
...
     compute_current_constraints ();
+
     assemble_stokes_system();
 
-    last_pressure_normalization_adjustment = normalize_pressure(temp_linearization_point);
+    last_pressure_normalization_adjustment = normalize_pressure(current_linearization_point);
 
     const double initial_newton_residual_velo = system_rhs.block(introspection.block_indices.velocities).l2_norm();
     const double initial_newton_residual_pres = system_rhs.block(introspection.block_indices.pressure).l2_norm();
     const double initial_newton_residual = std::sqrt(initial_newton_residual_velo * initial_newton_residual_velo + initial_newton_residual_pres * initial_newton_residual_pres);
```
gassmoeller: I know you did not change these lines, but in all other places we use `vel` and `p` as abbreviations for velocity and pressure. Could you do the same here? velo -> vel, and pres -> p

source/simulator/solver_schemes.cc
```diff
@@ -563,86 +581,110 @@ namespace aspect
...
+            // The while condition should actually never be false, because the break statement above should have caught it.
           }
-        while (line_search_iteration <= parameters.max_newton_line_search_iterations);
-        // The while condition should actually never be false, because the break statement above should have caught it.
 
 
         if (use_picard == true)
           {
+            // When we are using (defect corrected) Picard, keep the
+            // newton_derivative_scaling_factor at zero.
```
gassmoeller: I do not see a zero in the lines below. Am I misunderstanding something? If so, please document what I am missing.

tests/nonlinear_channel_flow_velocities_iterated_IMPES/screen-output
```diff
@@ -7,268 +9,1600 @@ Number of degrees of freedom: 3,556 (2,178+289+1,089)
...
+   Skipping temperature solve because RHS is zero.
+   Rebuilding Stokes preconditioner...
+   Solving Stokes system... 0+0 iterations.
+      Relative nonlinear residuals: 0, 9.92785e-09
+      Total relative residual after nonlinear iteration 15: 9.92785e-09
+
+
+   Skipping temperature solve because RHS is zero.
+   Rebuilding Stokes preconditioner...
+   Solving Stokes system... 0+0 iterations.
```
gassmoeller: Is there anything we can do about this case? The linear solver does not solve anything, because it thinks it has converged, but the nonlinear residual is not yet reached, so it keeps iterating.

source/simulator/solver.cc
```diff
@@ -661,26 +661,29 @@ namespace aspect
         // linearized_stokes_variables has a different
         // layout than current_linearization_point, which also contains all the
         // other solution variables.
-        if (parameters.nonlinear_solver != NonlinearSolver::Newton_Stokes)
+
+        if (!assemble_newton_stokes_system)
```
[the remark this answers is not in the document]
MFraters: Thanks :)
It turned out to be crucial to be able so easily switch between the Stokes and the Newton stokes solver, which I now again always do in the first nonlinear iteration.

source/simulator/solver.cc
```diff
@@ -661,26 +661,29 @@ namespace aspect
...
         else
           {
             /**
-             * The Newton solver solves for updates to variables, for which we have no good initial guesses.
-             * Therefore, initialize the solution vector for the solver to zero.
-             * TODO: I think that for the first iteration we could actually use a non-zero inital guess.
+             * The Newton solver solves for updates to variables, for which our best guess is zero when
+             * the it isn't the first nonlinear iteration. When it is the first nonlinear iteration, we
+             * have to assemble the full (non-defect correction) Picard, to get the boundary conditions
+             * right in combination with being able to use the initial guess optimally.
```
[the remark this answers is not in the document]
MFraters: Hmm, I guess I should rephrase it then. What I try to explain here is that for the Newton solver the initial guess has to be always zero (which is done in the lines below here). This is something you might not expect, because normally you set in the first nonlinear iteration of the newton solver the initial guess to the expected solution (current liniasation point for example). We don't do that here because we always use a full Picard iteration for that nonlinear iteration. Does this make more sense?

source/simulator/solver_schemes.cc
```diff
@@ -501,10 +508,7 @@ namespace aspect
         // create a completely distributed vector that will be used for
         // the scaled and denormalized solution and later used as a
         // starting guess for the linear solver
-        LinearAlgebra::BlockVector linearized_stokes_initial_guess (introspection.index_sets.stokes_partitioning, mpi_communicator);
-
-        linearized_stokes_initial_guess.block (block_vel) = current_linearization_point.block (block_vel);
-        linearized_stokes_initial_guess.block (block_p) = current_linearization_point.block (block_p);
+        LinearAlgebra::BlockVector linearized_stokes_initial_guess = current_linearization_point;
```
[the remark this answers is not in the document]
MFraters: I guess that since they are not used in the solver, it will be better to reverse this change. Thanks for catching this :)

source/simulator/solver_schemes.cc
MFraters: Ok, I will add to the documentation that the newton_derivative_scaling_factor is actually set above with the help of these two values. I now see that the naming of the variables below is suboptimal...

tests/nonlinear_channel_flow_velocities_iterated_IMPES/screen-output
MFraters: ah, I havn't seen this in a long time (didn't pay much attention to the convergence of the impes, just made sure the result was the same a the analytical result). This goes back to #302, where back then I couldn't find a case any more to reproduce the problem.

MFraters:
@gassmoeller Yes, I also slowly feel more and more comfortable with the code. I don't think it is production ready yet, but I start to think it isn't far off any more :)

I think that I have addressed all your comments. Since the comment about the zero iterations is not directly related to the Newton solver, we might want to open up #302 again, and deal with this problem there.

source/simulator/solver_schemes.cc
```diff
@@ -516,12 +515,26 @@ namespace aspect
...
+          {
+            denormalize_pressure (last_pressure_normalization_adjustment,
+                                  linearized_stokes_initial_guess,
+                                  current_linearization_point);
+          }
 
         if (nonlinear_iteration <= 1)
-          compute_current_constraints ();
+          {
+            // So this should only happen on nonlinear iteration 0 and 1.
```
tjhei: That comment is not very useful as it just restates the if above.

I also don't understand why you need this block because you do the same thing in line 521 above if nonlinear_iteration == 0

source/simulator/solver_schemes.cc
```diff
@@ -564,86 +576,113 @@ namespace aspect
...
+          {
+            /**
+             * We may need to do a line search if the solution update doesn't decrease the norm of the rhs enough.
+             * This is done by adding the solution update to the current linearization point and then assembling
+             * the Newton right hand side. If the Newton residual has decreased enough by using this update,
+             * then we continue, otherwise we reset the current linearization point with the help of the backup and
+             * add each iteration an increasingly smaller solution update until the decreasing residual condition
+             * is met, or the line search iteration limit is reached.
+             */
+            LinearAlgebra::BlockVector backup_linearization_point = current_linearization_point;
```
tjhei: this assignment is not necessary

source/simulator/solver_schemes.cc
```diff
@@ -564,86 +576,113 @@ namespace aspect
...
-                 */
-                lambda *= (2.0/3.0);// TODO: make a parameter out of this.
+                // Reset the current linearization point and the search direction
+                current_linearization_point = backup_linearization_point;
+                LinearAlgebra::BlockVector search_direction = solution;
+
+                // Scale the search direction down and add it to the current linearization point.
+                search_direction *= lambda;
+
+                current_linearization_point.block(introspection.block_indices.pressure) += search_direction.block(introspection.block_indices.pressure);
```
tjhei: you can use ``current_linearization_point.block(..).add(lambda, solution.block(..))`` and remove ``search_direction``

source/simulator/solver_schemes.cc
```diff
@@ -564,86 +576,113 @@ namespace aspect
...
+                                   "above should have caught it."));
               }
-
-            line_search_iteration++;
-            Assert(line_search_iteration <= parameters.max_newton_line_search_iterations,
-                   ExcMessage ("This tests the while condition. This condition should "
-                               "actually never be false, because the break statement "
-                               "above should have caught it."));
+            while (line_search_iteration <= parameters.max_newton_line_search_iterations);
+            // The while condition should actually never be false, because the break statement above should have caught it.
```
tjhei: then change it to ``while(true)`` and remove this comment

MFraters: hmm, this was a discussion in #1707 (@gassmoeller review of 16 May) , @gassmoeller and @bangerth where in favour of making this explicit. I am personally fine with changing it back, if that is the consensus.

source/simulator/solver_schemes.cc
MFraters: I agree with you that the assignemnet should not be nessesary. The problem I face is that need to get the structure of the vector right. I tried `LinearAlgebra::BlockVector backup_linearization_point(introspection.index_sets.stokes_partitioning, mpi_communicator);`, but this gives me an error. How should I do this?
```
An error occurred in line <64> of file </home/fraters/deal.ii/deal.ii_8.5/dealii/source/lac/trilinos_block_vector.cc> in function
292:     dealii::TrilinosWrappers::MPI::BlockVector& dealii::TrilinosWrappers::MPI::BlockVector::operator=(const dealii::TrilinosWrappers::MPI::BlockVector&)
The violated condition was: 
    n_blocks() == 0 || n_blocks() == v.n_blocks()
Additional information: 
    Dimension 3 not equal to 2.
```

source/simulator/solver_schemes.cc
MFraters: good point. I will remove the lines 
``` 
assemblers.reset (new internal::Assembly::AssemblerLists<dim>());
set_assemblers();
```
at `if (nonlinear_iteration == 0)`, and the unnecessary comment.

source/simulator/solver_schemes.cc
MFraters: this is indeed a lot cleaner, thanks!

source/simulator/solver_schemes.cc
tjhei: if that is what they prefer, then leave it as is.

source/simulator/solver_schemes.cc
tjhei: This highlights another fact: your backup_linearization_point currently contains not only the stokes part but also temperature and all compositional fields. If you create it just for the Stokes part (like you wrote). You have to copy the individual blocks of velocity and pressure.

MFraters: hmm, you are right. But I noticed that I need an assignment anyway, so that the first current_linearisation_point is not set to zero every time. So the code now reads now.
```
LinearAlgebra::BlockVector backup_linearization_point(introspection.index_sets.stokes_partitioning, mpi_communicator);
backup_linearization_point.block(introspection.block_indices.pressure) = current_linearization_point.block(introspection.block_indices.pressure);
backup_linearization_point.block(introspection.block_indices.velocities) = current_linearization_point.block(introspection.block_indices.velocities);

```

tjhei: > But I noticed that I need an assignment anyway

Really? Inside the ``do`` you will be assigning it again and your are not reading it in-between or am I reading the diff incorrectly?

MFraters: The relevant code in that part now reads:
```
            LinearAlgebra::BlockVector backup_linearization_point(introspection.index_sets.stokes_partitioning, mpi_communicator);
            backup_linearization_point.block(introspection.block_indices.pressure) = current_linearization_point.block(introspection.block_indices.pressure);
            backup_linearization_point.block(introspection.block_indices.velocities) = current_linearization_point.block(introspection.block_indices.velocities);

...

            do
              {
                // Reset the current linearization point and the search direction
                current_linearization_point.block(introspection.block_indices.pressure) = backup_linearization_point.block(introspection.block_indices.pressure);
                current_linearization_point.block(introspection.block_indices.velocities) = backup_linearization_point.block(introspection.block_indices.velocities);
...
```
So for the first loop, the backup_linearization_point needs to have the value of current_linearization_point. After this point the backup_linearization_point is never reassigned.

source/simulator/solver_schemes.cc
MFraters: on second thought, it doesn't work in debug mode because of the ghost vectors of the current_linearization_point. I get an assert: `You are trying an operation on a vector that is only allowed if the vector has no ghost elements, but the vector you are operating on does have ghost elements. Specifically, vectors with ghost elements are read-only and cannot appear in operations that write into these vectors.`

MFraters: @tjhei is there a way around this or should I just revert to what it was?

MFraters:
I think that I have addressed all comments.

gassmoeller:
This looks much better, and is easier to read. Thanks!

PR #1997, opened 2017-11-07
Do not look at output if ASPECT_COMPARE_TEST_RESULTS=OFF

gassmoeller:
Only perform normalization of output files if we actually look at the results. This fixes the last remaining test failures for deal.II-dev with ASPECT_COMPARE_TEST_RESULTS=OFF. 
This PR makes some of the earlier changes in #1973 unnecessary, because we now only call `tests/cmake/diff_test.sh` if we want to compare the results. Should I revert the now unnecessary changes to that file?

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -292,56 +292,58 @@ FOREACH(_test ${_tests})
 
     GET_FILENAME_COMPONENT(ASPECT_BASE_DIR ${CMAKE_CURRENT_SOURCE_DIR} PATH)
 
-    # Commands to normalize each output. Note that we depend on screen-output
-    # instead of ${_output_name} here, to avoid having to specify BYPRODUCTS
-    # for ninja.
-    FOREACH(_output_name ${_output_files})
-	ADD_CUSTOM_COMMAND(OUTPUT ${_output_name}.notime
-	  COMMAND ${PERL_EXECUTABLE}
-		  ${CMAKE_CURRENT_SOURCE_DIR}/normalize.pl
-		  ${_output_name} ${ASPECT_BASE_DIR}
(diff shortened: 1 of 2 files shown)


tjhei:
I am not sure I understand. Why is it incorrect to normalize the output files if we don't compare them?

tjhei:
> Should I revert the now unnecessary changes to that file?

All changes to diff_test.sh are unnecessary, right? I would think it makes sense to go back to the original version if the new stuff is unused.

gassmoeller:
It is not incorrect, but it is unnecessary. The reason I propose this is because the names of the particle output files change when we use 9.0.pre compared to 8.5.0 (to unify the naming conventions with the names of the solution files). We then try to normalize the output files of the test results, using the file names from the reference results, and this will fail because there are no such files. This PR is just a workaround for that problem. If we do not care about the results in those files, we should not touch them in any way.

> All changes to diff_test.sh are unnecessary, right? I would think it makes sense to go back to the original version if the new stuff is unused.

If you agree this PR is the right approach, then I will revert them. Will wait for your answer to my above comment before doing that.

tjhei:
> If you agree this PR is the right approach, then I will revert them. Will wait for your answer to my above comment before doing that.

Yes. I agree with this PR. I tried to convince myself that the dependencies of all the custom cmake targets is still correct and I think it is the case. :+1:

tjhei:
> This fixes the last remaining test failures for deal.II-dev

agreed, these are the last bugs: http://www.math.clemson.edu/~heister/cib/aspect-dev-gcc-hd/fce0b3e359d4ca810cd8054e6f670c6e96ee19c2/

gassmoeller:
I reverted the earlier change to the diff script. This should be ready to merge.

PR #1998, opened 2017-11-07
make boundary filename convention warning print once per file

bartniday:
The "WARNING: Filename convention concerning geometry boundary names changed..." message was printing three times on every processor. Now it should only print once for each offending filename.


bangerth:
Yes, this sounds completely reasonable! Thanks for the patch -- let's see what the tester says!

PR #2005, opened 2017-11-15
Move morency doin material model to cookbook

gassmoeller:
Something I wanted to do for a long time. Move the Morency & Doin material model into a separate folder in the cookbooks, and remove it from the other material models. Also shorten the manual section by removing unnecessary parameters.


bangerth:
Hm, the `morency_doin` test fails. Can you take a look?

As for the actual content of the patch, could someone with more knowledge of the material models take a look?

gassmoeller:
Yes, the test now has a line about loading the shared library.

 I think @jperryhouts was the only one working with this material model. Jonathan, could you have a look at this? It is really just moving two files and some lines around (and shortening the manual somewhat).

jperryhouts:
I am all for this! I implemented this to reproduce a paper once, and haven't found it useful since. I even suggested removing it all together a while ago.

Also, does anyone know of a way to get more obvious notifications of @ mentions? As it is I'm 'watching' this repository so I get emails about all these comments anyway. Who knows how many threads I've accidentally ignored because of that!

jdannberg:
@jperryhouts 
I was curious about the @ mentions to, and I found [this](https://help.github.com/articles/about-email-notifications/): If you were mentioned in a post, in the cc of the email `mention@noreply.github.com` will appear. So you can filter your emails using that.

gassmoeller:
I fixed the copyright years. The different column is a consequence of our changes to the statistics file over time (likely the addition of the time step length, possibly the nonlinear iteration count, and or Stokes inner iterations). I checked that the figure now produces the same result as is included in the manual.

jperryhouts:
@jdannberg Thanks! That's super helpful. It looks like there's also an undocumented header,
"X-GitHub-Reason: mention", which can be filtered too.

PR #2006, opened 2017-11-17
Split newton stabilisation as options and add a failsafe

MFraters:
This is something which I have been working on and with for a while now, and since #1996 is (hopefully) nearly merged, it seems me like a good time to discuss this now. This pull request actually add two features at the same time, because they used to be intertwined, but I untangled them just before this pull request. If you want I can make two pull requests from this one.

The first feature is that it will  now be possible to chose whether to use the stabilisation or not (and which of the two) for both the preconditioner and the A block. 

My tests with the Spiegelman benchmark (I only tested with preconditioner being `SPD`) seem to indicate that if the solver doesn't fail, the convergence rate is by far the best if you only symmetrise the the A block (`symmetric`), then much closer together `SPD`, `none`, `PD`. But both the `symmetric` and the `none` fail when problems get tougher. This is why I introduce the second feature.

The second feature is that when the solver doesn't converge, we try again, but set everything to `SPD`. This way I can get the best of both world (being `symmetric` and `SPD`).

If you all agree with this approach, then will actually clean up the code, add more documentation and for example change the strings of `"SPD"` and `"none"` into enums.

### DIFF
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -62,19 +62,36 @@ namespace aspect
    * material model outputs for the Newton solver.
    */
 
-  enum class PreconditionerStabilization
-  {
-    SPD, PD, symmetric, none
-  };
-  enum class VelocityBlockStabilization
-  {
-    SPD, PD, symmetric, none
-  };
-
   template <int dim>
   class NewtonHandler: public SimulatorAccess<dim>
   {
     public:
+
+      /**
+       * This enum describes the type of stabilization is used
+       * for the Newton solver. None represents no stabilization,
+       * SPD represent that the resulting matrix is made Symmetric
+       * Positive Definite, symmetric represents that the matrix is
+       * only symmetrized, and PD represents that we do the same as
+       * what we do for SPD, but without the symmetrization.
+       */
+      enum NewtonStabilization
+      {
+        none = 0,
+        symmetric = 1,
+        PD = 2,
+        SPD = symmetric | PD
+      };
+
+      friend
+      NewtonStabilization
+      operator| (const NewtonStabilization a,
+                 const NewtonStabilization b)
+      {
+        return static_cast<NewtonStabilization>(
+                 static_cast<int>(a) | static_cast<int>(b));
+      }
+
       /**
        * Determine, based on the run-time parameters of the current simulation,
        * which functions need to be called in order to assemble linear systems,
@@ -118,22 +135,22 @@ namespace aspect
       /**
        * Get the stabilization type used in the preconditioner.
        */
-      PreconditionerStabilization get_preconditioner_stabilization() const;
+      NewtonStabilization get_preconditioner_stabilization() const;
 
       /**
        * Set the stabilization type used in the preconditioner.
        */
-      void set_preconditioner_stabilization(const PreconditionerStabilization preconditioner_stabilization);
+      void set_preconditioner_stabilization(const NewtonStabilization preconditioner_stabilization);
 
       /**
        * Get the stabilization type used in the velocity block.
        */
-      VelocityBlockStabilization get_velocity_block_stabilization() const;
+      NewtonStabilization get_velocity_block_stabilization() const;
 
       /**
        * Sets the stabilization type used in the velocity block.
        */
-      void set_velocity_block_stabilization(const VelocityBlockStabilization velocity_block_stabilization);
+      void set_velocity_block_stabilization(const NewtonStabilization velocity_block_stabilization);
 
       /**
        * Get whether to use the Newton failsafe. If the failsafe is used, a failure
@@ -172,13 +189,7 @@ namespace aspect
       /**
        * get a std::string describing the stabilization type used for the preconditioner.
        */
-      std::string get_preconditioner_stabilization_string(PreconditionerStabilization preconditioner_stabilization);
-
-      /**
-       * get a std::string describing the stabilization type used for the velocity block.
-       */
-      std::string get_velocity_block_stabilization_string(VelocityBlockStabilization velocity_block_stabilization);
-
+      std::string get_newton_stabilization_string(const NewtonStabilization preconditioner_stabilization) const;
 
       /**
        * Declare additional parameters that are needed for the Newton.
@@ -200,15 +211,15 @@ namespace aspect
        * See the get_newton_derivative_scaling_factor() function for an
        * explanation of the purpose of this factor.
        */
-      double                      newton_derivative_scaling_factor;
-      PreconditionerStabilization preconditioner_stabilization;
-      VelocityBlockStabilization  velocity_block_stabilization;
-      bool                        use_Newton_failsafe;
-      double                      nonlinear_switch_tolerance;
-      unsigned int                max_pre_newton_nonlinear_iterations;
-      unsigned int                max_newton_line_search_iterations;
-      bool                        use_newton_residual_scaling_method;
-      double                      maximum_linear_stokes_solver_tolerance;
+      double              newton_derivative_scaling_factor;
+      NewtonStabilization preconditioner_stabilization;
+      NewtonStabilization velocity_block_stabilization;
+      bool                use_Newton_failsafe;
+      double              nonlinear_switch_tolerance;
+      unsigned int        max_pre_newton_nonlinear_iterations;
+      unsigned int        max_newton_line_search_iterations;
+      bool                use_newton_residual_scaling_method;
+      double              maximum_linear_stokes_solver_tolerance;
   };
 
   namespace Assemblers
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -109,13 +109,13 @@ namespace aspect
        * Gets the Newton derivative scaling factor used for scaling the
        * derivative part of the Newton Stokes solver in the assembly.
        */
-      bool get_use_spd_factor() const;
+      std::pair<std::string,std::string> get_Newton_stabilisation() const;
 
       /**
        * Sets the Newton derivative scaling factor used for scaling the
        * derivative part of the Newton Stokes solver in the assembly.
        */
-      void set_use_spd_factor(const bool set_spd_factor);
+      void set_Newton_stabilisation(const std::string use_Newton_stabilisation_preconditioner,const std::string use_Newton_stabilisation_A_block);
 
     private:
       /**
@@ -126,7 +126,7 @@ namespace aspect
        * explanation of the purpose of this factor.
        */
       double newton_derivative_scaling_factor;
-      bool use_spd_factor;
+      std::pair<std::string,std::string> Newton_stabilisation;
   };
 
   namespace Assemblers
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -105,6 +105,17 @@ namespace aspect
        */
       void set_newton_derivative_scaling_factor(const double newton_derivative_scaling_factor);
 
+      /**
+       * Gets the Newton derivative scaling factor used for scaling the
+       * derivative part of the Newton Stokes solver in the assembly.
+       */
+      bool get_use_spd_factor() const;
+
+      /**
+       * Sets the Newton derivative scaling factor used for scaling the
+       * derivative part of the Newton Stokes solver in the assembly.
+       */
+      void set_use_spd_factor(const bool set_spd_factor);
 
     private:
       /**
@@ -115,6 +126,7 @@ namespace aspect
        * explanation of the purpose of this factor.
        */
       double newton_derivative_scaling_factor;
+      bool use_spd_factor;
   };
 
   namespace Assemblers
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -61,6 +61,16 @@ namespace aspect
    * A Class which can declare and parse parameters and creates
    * material model outputs for the Newton solver.
    */
+
+  enum class PreconditionerStabilization
+  {
+    SPD, PD, symmetric, none
+  };
+  enum class VelocityBlockStabilization
+  {
+    SPD, PD, symmetric, none
+  };
+
   template <int dim>
   class NewtonHandler: public SimulatorAccess<dim>
   {
@@ -108,63 +118,76 @@ namespace aspect
       /**
        * Get the stabilization type used in the preconditioner.
        */
-      std::string get_preconditioner_stabilization() const;
+      PreconditionerStabilization get_preconditioner_stabilization() const;
 
       /**
        * Set the stabilization type used in the preconditioner.
        */
-      void set_preconditioner_stabilization(const std::string preconditioner_stabilization);
+      void set_preconditioner_stabilization(const PreconditionerStabilization preconditioner_stabilization);
 
       /**
        * Get the stabilization type used in the velocity block.
        */
-      std::string get_velocity_block_stabilization() const;
+      VelocityBlockStabilization get_velocity_block_stabilization() const;
 
       /**
        * Sets the stabilization type used in the velocity block.
        */
-      void set_velocity_block_stabilization(const std::string velocity_block_stabilization);
+      void set_velocity_block_stabilization(const VelocityBlockStabilization velocity_block_stabilization);
 
       /**
-       * Get whether to use the Newton failsafe
+       * Get whether to use the Newton failsafe. If the failsafe is used, a failure
+       * of the linear solver is catched and we try to solve it again with both the
+       * preconditioner and the velocity block being stabilized with the SPD stabilization.
        */
       bool get_use_Newton_failsafe();
 
       /**
        * Get the nonlinear tolerance at which to switch the
-       * nonlinear solver from defect corrected Piccard to
+       * nonlinear solver from defect correction Picard to
        * Newton.
        */
       double get_nonlinear_switch_tolerance();
 
       /**
-       * Get the maximum amount of pre-Newton nonlinear iterations
+       * Get the maximum number of pre-Newton nonlinear iterations.
        */
       unsigned int get_max_pre_newton_nonlinear_iterations();
 
       /**
-       * Get the maximum amount of line search iterations
+       * Get the maximum number of line search iterations.
        */
       unsigned int get_max_newton_line_search_iterations();
 
       /**
-       * Get whether to use the residual scaling method
+       * Get whether to use the residual scaling method.
        */
       bool get_use_newton_residual_scaling_method();
 
       /**
-       * Get the maximum linear Stokes solver tolerance
+       * Get the maximum linear Stokes solver tolerance.
        */
       double get_maximum_linear_stokes_solver_tolerance();
 
       /**
-       * Declare additional parameters that are needed for the Newton
+       * get a std::string describing the stabilization type used for the preconditioner.
+       */
+      std::string get_preconditioner_stabilization_string(PreconditionerStabilization preconditioner_stabilization);
+
+      /**
+       * get a std::string describing the stabilization type used for the velocity block.
+       */
+      std::string get_velocity_block_stabilization_string(VelocityBlockStabilization velocity_block_stabilization);
+
+
+      /**
+       * Declare additional parameters that are needed for the Newton.
        * solver.
        */
       static void declare_parameters (ParameterHandler &prm);
 
       /**
-       * Parse additional parameters that are needed for the Newton
+       * Parse additional parameters that are needed for the Newton.
        * solver.
        */
       void parse_parameters (ParameterHandler &prm);
@@ -177,15 +200,15 @@ namespace aspect
        * See the get_newton_derivative_scaling_factor() function for an
        * explanation of the purpose of this factor.
        */
-      double       newton_derivative_scaling_factor;
-      std::string  preconditioner_stabilization;
-      std::string  velocity_block_stabilization;
-      bool         use_Newton_failsafe;
-      double       nonlinear_switch_tolerance;
-      unsigned int max_pre_newton_nonlinear_iterations;
-      unsigned int max_newton_line_search_iterations;
-      bool         use_newton_residual_scaling_method;
-      double       maximum_linear_stokes_solver_tolerance;
+      double                      newton_derivative_scaling_factor;
+      PreconditionerStabilization preconditioner_stabilization;
+      VelocityBlockStabilization  velocity_block_stabilization;
+      bool                        use_Newton_failsafe;
+      double                      nonlinear_switch_tolerance;
+      unsigned int                max_pre_newton_nonlinear_iterations;
+      unsigned int                max_newton_line_search_iterations;
+      bool                        use_newton_residual_scaling_method;
+      double                      maximum_linear_stokes_solver_tolerance;
   };
 
   namespace Assemblers
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -106,16 +106,68 @@ namespace aspect
       void set_newton_derivative_scaling_factor(const double newton_derivative_scaling_factor);
 
       /**
-       * Gets the Newton derivative scaling factor used for scaling the
-       * derivative part of the Newton Stokes solver in the assembly.
+       * Get the stabilization type used in the preconditioner.
        */
-      std::pair<std::string,std::string> get_Newton_stabilisation() const;
+      std::string get_preconditioner_stabilization() const;
 
       /**
-       * Sets the Newton derivative scaling factor used for scaling the
-       * derivative part of the Newton Stokes solver in the assembly.
+       * Set the stabilization type used in the preconditioner.
        */
-      void set_Newton_stabilisation(const std::string use_Newton_stabilisation_preconditioner,const std::string use_Newton_stabilisation_A_block);
+      void set_preconditioner_stabilization(const std::string preconditioner_stabilization);
+
+      /**
+       * Get the stabilization type used in the velocity block.
+       */
+      std::string get_velocity_block_stabilization() const;
+
+      /**
+       * Sets the stabilization type used in the velocity block.
+       */
+      void set_velocity_block_stabilization(const std::string velocity_block_stabilization);
+
+      /**
+       * Get whether to use the Newton failsafe
+       */
+      bool get_use_Newton_failsafe();
+
+      /**
+       * Get the nonlinear tolerance at which to switch the
+       * nonlinear solver from defect corrected Piccard to
+       * Newton.
+       */
+      double get_nonlinear_switch_tolerance();
+
+      /**
+       * Get the maximum amount of pre-Newton nonlinear iterations
+       */
+      unsigned int get_max_pre_newton_nonlinear_iterations();
+
+      /**
+       * Get the maximum amount of line search iterations
+       */
+      unsigned int get_max_newton_line_search_iterations();
+
+      /**
+       * Get whether to use the residual scaling method
+       */
+      bool get_use_newton_residual_scaling_method();
+
+      /**
+       * Get the maximum linear Stokes solver tolerance
+       */
+      double get_maximum_linear_stokes_solver_tolerance();
+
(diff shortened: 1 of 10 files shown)


include/aspect/parameters.h
```diff
@@ -308,6 +308,9 @@ namespace aspect
 
     double                         nonlinear_tolerance;
     double                         nonlinear_switch_tolerance;
+    std::string                    use_Newton_stabilisation_preconditioner;
```
gassmoeller: I feel like it is time to give the Newton handler its own `declare_parameters` and `parse_parameters` functions. This way we do not clutter the main parameters with things that are specific to the Newton handler. The melt handler already does this, and probably you can copy the same approach (see parameters.cc:1728, core.cc:544, melt.cc:1367). Also as you mentioned these can be `enum`

include/aspect/newton.h
```diff
@@ -115,6 +126,7 @@ namespace aspect
        * explanation of the purpose of this factor.
        */
       double newton_derivative_scaling_factor;
+      std::pair<std::string,std::string> Newton_stabilisation;
```
gassmoeller: Keep this as two separate variables and document what they do. You can return them as a pair in `get_Newton_stabilization`, but set them as two separate input arguments in `set_Newton_stabilization`

include/aspect/newton.h
```diff
@@ -105,6 +105,17 @@ namespace aspect
        */
       void set_newton_derivative_scaling_factor(const double newton_derivative_scaling_factor);
 
+      /**
+       * Gets the Newton derivative scaling factor used for scaling the
+       * derivative part of the Newton Stokes solver in the assembly.
+       */
+      std::pair<std::string,std::string> get_Newton_stabilisation() const;
```
gassmoeller: `stabilization` please. Otherwise we end up in an American vs. British English war :smile:.

include/aspect/parameters.h
```diff
@@ -308,6 +308,9 @@ namespace aspect
...
+    std::string                    use_Newton_stabilisation_A_block;
+    bool                           use_Newton_failsafe;
```
gassmoeller: If we decide this is a good way, we might even omit this variable and always fall back to the failsafe.

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -96,6 +96,7 @@ namespace aspect
 
           const double JxW = scratch.finite_element_values.JxW(q);
 
+
```
gassmoeller: Do you need this line?

include/aspect/newton.h
gassmoeller: After reading through the rest of the PR, I think it would be better to split both functions into separate functions that could be named `set_preconditioner_stabilization` and `set_velocity_block_stabilization` respectively (and the same for the `get_...` functions. This is more intuitive than returning or setting a pair of parameters. Otherwise it would get real complicated if somebody added a third stabilization option.

bangerth:
@MFraters -- can you address @gassmoeller 's comments first? Once you're done with those, let me know and I'll take another look.

I'd say make this a priority today, if you can!

MFraters:
I addressed all comments except the enum one. If I make it an enum, is there an easy way to output the enum name to the screen without a bunch of if statements?

include/aspect/newton.h
```diff
@@ -105,6 +105,69 @@ namespace aspect
...
 
+      /**
+       * Get the stabilization type used in the preconditioner.
+       */
+      std::string get_preconditioner_stabilization() const;
+
+      /**
+       * Set the stabilization type used in the preconditioner.
+       */
+      void set_preconditioner_stabilization(const std::string preconditioner_stabilization);
```
bangerth: I see what you mean by using these strings. I think it would indeed be cleaner if you defined `enum`s here and in similar places that these functions could take/return

include/aspect/newton.h
```diff
@@ -105,6 +105,69 @@ namespace aspect
...
+
+      /**
+       * Get the stabilization type used in the velocity block.
+       */
+      std::string get_velocity_block_stabilization() const;
+
+      /**
+       * Sets the stabilization type used in the velocity block.
+       */
+      void set_velocity_block_stabilization(const std::string velocity_block_stabilization);
```
bangerth: same here

include/aspect/newton.h
```diff
@@ -105,6 +105,69 @@ namespace aspect
...
+
+      /**
+       * Get whether to use the Newton failsafe
+       */
+      bool get_use_Newton_failsafe();
```
bangerth: can you add another sentence to the documentation that explains what the "failsafe" is?

Also end the sentence with a period.

include/aspect/newton.h
```diff
@@ -105,6 +105,69 @@ namespace aspect
...
+
+      /**
+       * Get the nonlinear tolerance at which to switch the
+       * nonlinear solver from defect corrected Piccard to
```
bangerth: -> defect correction
-> Picard (one 'c')

include/aspect/newton.h
```diff
@@ -105,6 +105,69 @@ namespace aspect
...
+       * Newton.
+       */
+      double get_nonlinear_switch_tolerance();
+
+      /**
+       * Get the maximum amount of pre-Newton nonlinear iterations
```
bangerth: amount -> number
(maybe also in other places; "amount" is used for uncountable things, such as milk, flour, love; "number" is used for countable things, such as iterations, words per doc string, software bugs, etc.)

include/aspect/newton.h
```diff
@@ -114,7 +177,15 @@ namespace aspect
        * See the get_newton_derivative_scaling_factor() function for an
        * explanation of the purpose of this factor.
        */
-      double newton_derivative_scaling_factor;
+      double       newton_derivative_scaling_factor;
+      std::string  preconditioner_stabilization;
+      std::string  velocity_block_stabilization;
```
bangerth: These two variables would simply become `enum`s then, possibly of different `enum` types. Of course you'd have to have a way to translate from a string to an enum and possibly back, but that should be easy to write in the `parse_parameters()` function.

The advantage is that `enum`s have only finitely many possible values, and so you can easily cover them in a `switch` statement without much trouble. For `std::string` objects, there are infinitely many possibilities that, at least in principle, you'd have to check at one point or other. You also can't use them in a `switch` statement.

include/aspect/newton.h
```diff
@@ -61,6 +61,16 @@ namespace aspect
    * A Class which can declare and parse parameters and creates
    * material model outputs for the Newton solver.
    */
+
+  enum class PreconditionerStabilization
```
bangerth: The comment above this enum is still about the `NewtonHandler` class.

I'd say move these `enum`s *into* the `NewtonHandler` class. Also add documentation for both.

include/aspect/newton.h
```diff
@@ -61,6 +61,16 @@ namespace aspect
...
+  {
+    SPD, PD, symmetric, none
```
bangerth: I think there is an easier way: declare it like this:
```
  none = 0,
  symmetric = 1;
  PD = 2;
  SPD = symmetric | PD;
```
You can then in the appropriate place ask whether a particular bit is set.

Same for the enum below.

include/aspect/newton.h
```diff
@@ -61,6 +61,16 @@ namespace aspect
...
+  };
+  enum class VelocityBlockStabilization
+  {
+    SPD, PD, symmetric, none
```
bangerth: Thinking about it, since the two are the same, could you not use the same `enum` for both purposes, and simply have two variables of this enum type?

source/simulator/core.cc
```diff
@@ -646,6 +646,7 @@ namespace aspect
       {
         assemble_newton_stokes_system = true;
         newton_handler->initialize_simulator(*this);
+        newton_handler->parse_parameters(prm);
```
bangerth: Could you investigate whether this whole `if` block could be moved to the same location where we currently initialize the melt handler and the free surface handler (around line 530)?

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+        case PreconditionerStabilization::none:
+          return "none";
+          break;
+      }
+  }
+
+  template <int dim>
+  std::string
+  NewtonHandler<dim>::
+  get_velocity_block_stabilization_string(VelocityBlockStabilization velocity_block_stabilization)
```
bangerth: make this argument `const`. same in the declaration of the function if it's not yet there. same also for the previous function.

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+          return "PD";
+          break;
+        case PreconditionerStabilization::symmetric:
+          return "symmetric";
+          break;
+        case PreconditionerStabilization::none:
+          return "none";
+          break;
+      }
+  }
```
bangerth: You'll want to put a
```
  default:
    Assert (false, ExcNotImplemented());
    return "";
```
case into the `switch` to make sure you return *something* and that you catch potential mistakes.

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+  template <int dim>
+  std::string
+  NewtonHandler<dim>::
+  get_preconditioner_stabilization_string(PreconditionerStabilization preconditioner_stabilization)
+  {
+    switch (preconditioner_stabilization)
+      {
+        case PreconditionerStabilization::SPD:
+          return "SPD";
+          break;
```
bangerth: no need for the `break` after the `return`. Same for the other cases below and the next function.

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+  {
+    prm.enter_subsection ("Solver parameters");
+    {
+      prm.enter_subsection ("Newton solver parameters");
+      {
+        prm.declare_entry ("Nonlinear Newton solver switch tolerance", "1e-5",
+                           Patterns::Double(0,1),
+                           "A relative tolerance with respect to the residual of the first "
+                           "iteration, up to which the nonlinear Picard solver will iterate, "
+                           "before changing to the newton solver.");
```
bangerth: -> Newton

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+
+        prm.declare_entry ("Max pre-Newton nonlinear iterations", "10",
+                           Patterns::Integer (0),
+                           "The maximum number of Picard nonlinear iterations to be performed "
+                           "before switching to Newton iterations.");
```
bangerth: Maybe say "If the ``Nonlinear Newton solver switch tolerance'' is reached before the maximal number of Picard iteration, then the solver switches to Newton solves anyway.", just to make the connection between the two parameters.

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+
+        prm.declare_entry ("Max Newton line search iterations", "5",
+                           Patterns::Integer (0),
+                           "The maximum number of line search iterations allowed. If the "
+                           "criterion is not reached after this iteration, we apply the scaled "
```
bangerth: -> this number of iterations

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+                           "increment to the solution and continue.");
```
bangerth: -> apply the scaled increment even though it does not satisfy the necessary criteria and simply continue with the next Newton iteration

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+
+        prm.declare_entry ("Use Newton residual scaling method", "false",
+                           Patterns::Bool (),
+                           "This method allows to slowly introduce the derivatives based on the improvement "
+                           "of the residual. If set to false, the scaling factor for the Newton derivatives "
+                           "is set to one immediately when switching on the Newton solver.");
```
bangerth: Can you provide a one- or two-sentence summary of what will happen if set to `true`?

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+
+        prm.declare_entry ("Maximum linear Stokes solver tolerance", "0.9",
+                           Patterns::Double (0,1),
+                           "When the linear Stokes solver tolerance is dynamically chosen, this defines "
+                           "the most loose tolerance allowed.");
```
bangerth: I don't know what this parameter does, either from the parameter's name nor from the description. Can you elaborate?

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+
+        prm.declare_entry ("Stabilization preconditioner", "SPD",
+                           Patterns::Selection ("SPD|PD|symmetric|none"),
+                           "This parameters allows for the stabilisation of the preconditioner. By default, the "
+                           "matrix created for the preconditioning is not necessarily Symmetric Positive Definite. "
+                           "This is problematic (see Fraters et al, in prep). When none is chosen, the perconitioner "
```
bangerth: -> surround `none` with backtick...forwardtick
perconitioner -> preconditioner

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+                           "is set to one immediately when switching on the Newton solver.");
+
+        prm.declare_entry ("Maximum linear Stokes solver tolerance", "0.9",
+                           Patterns::Double (0,1),
+                           "When the linear Stokes solver tolerance is dynamically chosen, this defines "
+                           "the most loose tolerance allowed.");
+
+        prm.declare_entry ("Stabilization preconditioner", "SPD",
+                           Patterns::Selection ("SPD|PD|symmetric|none"),
+                           "This parameters allows for the stabilisation of the preconditioner. By default, the "
```
bangerth: The "default" here is actually SPD, but that's not what you mean when you use the word.

Maybe say: "If one derives the Newton method without any modifications, the ..."

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+                           "matrix created for the preconditioning is not necessarily Symmetric Positive Definite. "
+                           "This is problematic (see Fraters et al, in prep). When none is chosen, the perconitioner "
+                           "is not stabilized. The Symmetric parameters symmetrizes the matrix, and PD makes the matrix "
```
bangerth: same here -- surround all possible values for the parameter by back...forwardticks.

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+                           "Positive Definite. SPD is the full stabilization, where the matrix is garanteed Symmetric "
```
bangerth: -> guaranteed

source/simulator/newton.cc
bangerth: -> et al. (not the period after "al")

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+                           "Positive Definite.");
+        prm.declare_entry ("Stabilization velocity block", "SPD",
+                           Patterns::Selection ("SPD|PD|symmetric|none"),
+                           "This parameters allows for the stabilisation of the velocity block. By default, the "
```
bangerth: all the same issues here. You may also just reference the preconditioner parameter (or the other way around) and just say something like "Stabilization for the velocity block of the system matrix used in the Newton method. Possible values for this parameter are the same as for ..."

source/simulator/newton.cc
```diff
@@ -161,6 +161,239 @@ namespace aspect
...
+                           "matrix created for the velocity block is not necessarily Symmetric Positive Definite. "
+                           "This is problematic (see Fraters et al, in prep). When none is chosen, the velocity block "
+                           "is not stabilized. The Symmetric parameters symmetrizes the matrix, and PD makes the matrix "
+                           "Positive Definite. SPD is the full stabilization, where the matrix is garanteed Symmetric "
+                           "Positive Definite.");
+        prm.declare_entry ("Use Newton failsafe", "false",
+                           Patterns::Bool (),
+                           "Switches on SPD stabilization when solver fails.");
```
bangerth: Can you elaborate on this in a couple more sentences? I.e., *what* solver needs to fail, and maybe reference what other parameter values are then changed? And for how long -- all future time steps, or just the current time step?

source/simulator/solver_schemes.cc
```diff
@@ -573,19 +573,67 @@ namespace aspect
...
+                                                            newton_handler->get_maximum_linear_stokes_solver_tolerance(),
                                                             parameters.linear_stokes_solver_tolerance,
                                                             stokes_residual,
                                                             residual,
                                                             residual_old);
 
-                pcout << "   The linear solver tolerance is set to " << parameters.linear_stokes_solver_tolerance << std::endl;
+                const std::string preconditioner_stabilization = newton_handler->get_preconditioner_stabilization_string(newton_handler->get_preconditioner_stabilization());
+                const std::string velocity_block_stabilization = newton_handler->get_velocity_block_stabilization_string(newton_handler->get_velocity_block_stabilization());
+                pcout << "   The linear solver tolerance is set to " << parameters.linear_stokes_solver_tolerance << ". Stabilisation Preconditioner is " << preconditioner_stabilization << " and A block is " << velocity_block_stabilization << "." << std::endl;
```
bangerth: -> stabilization (with a z)

I'd also show the value for the *matrix* first, before the value for the preconditioner

source/simulator/solver_schemes.cc
```diff
@@ -573,19 +573,67 @@ namespace aspect
...
+          }
+        else
+          {
+            try
+              {
+                stokes_residual = solve_stokes();
+              }
+            catch (...)
+              {
+                computing_timer.exit_section();
```
bangerth: leave a comment here of the form
```
// start the solve over again and try with a stabilized version
```

bangerth: About line 602: you are exiting a timer block here, but what timer block does the remainder of this block correspond to? Don't you have to start a timer block again somewhere?

source/simulator/solver_schemes.cc
MFraters: Why do you think that that order is more logical? In my mind the preconditioner comes first.

source/simulator/solver_schemes.cc
MFraters: So this exit solver.cc:525 `computing_timer.enter_section ("   Solve Stokes system");` (at the beginning of `solve_stokes()`. When the solver fails the `computing_timer.exit_section();` on line 921 is never executed, so I do it here.

source/simulator/solver_schemes.cc
bangerth: I don't know -- I think as preconditioners as an ancillary tool to solving the linear system.

I don't feel strongly about the issue if you do :-)

source/simulator/solver_schemes.cc
bangerth: I don't think I understand this.

You have code like this:
```
  timer.enter_section (...);
  try {
    ...do something...
  } catch (...) {
    timer.leave_section();
    ...code now not associated with any timer section...
  }

  ...code that is now under the timer section again, but only if the try
     succeeded; if the try failed, this whole code block is not timed...
```
(I can't compare with the current code any more, because the line numbers you cite don't seem to match any more.)

Or are you saying that the `solve_stokes()` function enters a section at the top but the leaves through a `throw` statement with leaving the section? That would be a bug in that function -- it needs to leave the timer section before throwing the exception. If you don't right away see how to do this, show me the place tomorrow.

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -119,54 +119,82 @@ namespace aspect
               const SymmetricTensor<2,dim> viscosity_derivative_wrt_strain_rate = derivatives->viscosity_derivative_wrt_strain_rate[q];
               const SymmetricTensor<2,dim> strain_rate = scratch.material_model_inputs.strain_rate[q];
 
+              const typename NewtonHandler<dim>::NewtonStabilization preconditioner_stabilization = this->get_newton_handler().get_preconditioner_stabilization();
               // todo: make this 0.9 into a global input parameter
-              double alpha = Utilities::compute_spd_factor<dim>(eta, strain_rate, viscosity_derivative_wrt_strain_rate, 0.9);
+              const double alpha = preconditioner_stabilization == NewtonHandler<dim>::NewtonStabilization::SPD || preconditioner_stabilization == NewtonHandler<dim>::NewtonStabilization::PD ?
```
bangerth: Break long lines.

If you do the trick with the bit masks for the `enum`, you can just write this as
```
  const double alpha = (preconditioner_stabilization | NewtonHandler<dim>::NewtonStabilization::SPD != 0 ?
                                      ...);
```

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -119,54 +119,82 @@ namespace aspect
...
+                                   Utilities::compute_spd_factor<dim>(eta, strain_rate, viscosity_derivative_wrt_strain_rate, 0.9)
+                                   :
+                                   1;
+
+              if (preconditioner_stabilization == NewtonHandler<dim>::NewtonStabilization::SPD || preconditioner_stabilization == NewtonHandler<dim>::NewtonStabilization::symmetric)
```
bangerth: same in all other places where you have to make this sort of comparison

i.e., here you'll have to test `preconditioner_stabilization | ...::symmetric`

source/simulator/solver_schemes.cc
MFraters: >Or are you saying that the solve_stokes() function enters a section at the top but the leaves through a throw statement with leaving the section? That would be a bug in that function -- it needs to leave the timer section before throwing the exception. If you don't right away see how to do this, show me the place tomorrow.

Yes. I opened a separate pull request for this #2084 and removed it from this one.

include/aspect/newton.h
```diff
@@ -61,10 +61,28 @@ namespace aspect
...
+
+      /**
+       * This enum describes the type of stabilization is used
+       * for the Newton solver. None represents no stabilization,
+       * SPD represent that the resulting matrix is make Symmetric
+       * Positive Definite, symmetric represents that the matrix is
+       * only symmetrized, and PD represents that we do the same as
+       * what we do for SPD, but without the symmetrization.
+       */
+      enum NewtonStabilization
```
MFraters: I had change this form `enum class NewtonStabilization` to enum NewtonStabilization` to make sure I could use the operator|. The downside of this is that is isn't strongly typed anymore. An other option I tried was to add a below this the function:

```
inline NewtonStabilization operator|(NewtonStabilization a, NewtonStabilization b)
{return static_cast<NewtonStabilization>(static_cast<int>(a) | static_cast<int>(b));}
```
but then I got the error:
```
error: ‘aspect::NewtonHandler<dim>::NewtonStabilization aspect::NewtonHandler<dim>::operator|(aspect::NewtonHandler<dim>::NewtonStabilization, aspect::NewtonHandler<dim>::NewtonStabilization)’ must take exactly one argument
       inline NewtonStabilization operator|(NewtonStabilization a, NewtonStabilization b)
```

bangerth: You can't make this function a member of the surrounding `NewtonHandler` class. It needs to be a function at namespace scope. Just declare it below the end of the `NewtonHandler` class as
```
inline
typename NewtonHandler<dim>::NewtonStabilization 
operator|(typename NewtonHandler<dim>::NewtonStabilization a, 
               typename NewtonHandler<dim>::NewtonStabilization b)
{
  return static_cast<typename NewtonHandler<dim>::NewtonStabilization>(
    static_cast<int>(a) | static_cast<int>(b));
}
```

MFraters: hmm, this somehow doesn't work. It doesn't seem to recognise that the inline opertor is pressent, but it compiles newton.h fine. So when I put:
```
  template <int dim>
  inline
  typename NewtonHandler<dim>::NewtonStabilization
  operator|(typename NewtonHandler<dim>::NewtonStabilization a,
                 typename NewtonHandler<dim>::NewtonStabilization b)
  {
    return static_cast<typename NewtonHandler<dim>::NewtonStabilization>(
      static_cast<int>(a) | static_cast<int>(b));
  }
```
I get as error:
```
error: no match for ‘operator|’ (operand types are ‘aspect::NewtonHandler<2>::NewtonStabilization’ and ‘aspect::NewtonHandler<2>::NewtonStabilization’)
               const double alpha = preconditioner_stabilization | NewtonHandler<dim>::NewtonStabilization::PD ?
```

MFraters: So I have tried to add `  NewtonHandler<dim>::NewtonStabilization operator|(typename NewtonHandler<dim>::NewtonStabilization a, typename NewtonHandler<dim>::NewtonStabilization b); \` to the bottom of newton.cc, but that doesn't help. I also tried to just explicity instanciate them:
```
  typename NewtonHandler<2>::NewtonStabilization
  operator|(typename NewtonHandler<2>::NewtonStabilization a,
                 typename NewtonHandler<2>::NewtonStabilization b)
  {
    return static_cast<typename NewtonHandler<2>::NewtonStabilization>(
      static_cast<int>(a) | static_cast<int>(b));
  }

  typename NewtonHandler<3>::NewtonStabilization
  operator|(typename NewtonHandler<3>::NewtonStabilization a,
                 typename NewtonHandler<3>::NewtonStabilization b)
  {
    return static_cast<typename NewtonHandler<3>::NewtonStabilization>(
      static_cast<int>(a) | static_cast<int>(b));
  }
```
This works, but then I will have to specifically convert it into an int or bool again because it complains: 
```
error: could not convert ‘aspect::operator|(velocity_block_stabilization, (aspect::NewtonHandler<2>::NewtonStabilization)1)’ from ‘aspect::NewtonHandler<2>::NewtonStabilization’ to ‘bool’
```
So if we really want this approach, we would also need to implement a way to covert this class to a bool.

bangerth: I think I understand why it doesn't work. 

Providing two specializations is ok with me. If you don't like this approach: What happens if you put the operator back into the `NewtonHandler` class as you had it initially, but prefix the whole declaration with the `friend` keyword? I.e., 
```
friend
NewtonStabilization 
operator| (const NewtonStabilization a, 
                const NewtonStabilization b)
{
  return static_cast<NewtonStabilization>(
    static_cast<int>(a) | static_cast<int>(b));
}
```
(But if you like the explicit specialization you show, then no need to try this approach.)

As for the conversion to `bool`: Just do the comparison via
```
  if ((value | ...::SPD) != ...::none)
```
`...::none` has numerical value zero, but has type equal to the `enum`, so you can compare with it.

MFraters: ok, I got it working now with the class enum. I do not have a strong preference for either way, but I guess it is nicer to not have explicit specializations.

include/aspect/newton.h
```diff
@@ -61,10 +61,37 @@ namespace aspect
...
+
   template <int dim>
   class NewtonHandler: public SimulatorAccess<dim>
   {
     public:
+
+      /**
+       * This enum describes the type of stabilization is used
+       * for the Newton solver. None represents no stabilization,
+       * SPD represent that the resulting matrix is make Symmetric
```
bangerth: make -> made

include/aspect/newton.h
```diff
@@ -61,10 +61,37 @@ namespace aspect
...
+       * Positive Definite, symmetric represents that the matrix is
+       * only symmetrized, and PD represents that we do the same as
+       * what we do for SPD, but without the symmetrization.
+       */
+      enum class NewtonStabilization
```
bangerth: I blanked -- `enum class` is only C++11, but we can't presume that yet for ASPECT. Can you just make it `enum`? I think everything should continue to work anyway without further modifications.

PR #1785, opened 2017-05-26
Unify assembler structure

gassmoeller:
Our current assembler system is somewhat messy, because the structure for how an assembler has to look like is not well defined. Essentially any function can be attached to the available signals, and we do not know which functions are already attached (because we can not check which type each connection has), or which quantities an individual assembler needs to work correctly. This has lead to the extremely confusing logic inside of `set_assemblers`, and to the fact that for every new assembler we need to modify three objects: Connect to the associated signal. Add the parent object to `assembler_objects`. Modify the correct `assembler_properties` object to compute the correct quantities before calling the assembler.
If instead we had an interface-plugin system as for the other parts of ASPECT, all this would be handled in one object, the individual assembler plugin. It would have a defined function that is called by Simulator (which replaces the signal), and could (optionally) implement another function that specifies which quantities need to be provided (which replaces the properties object). We could still keep the `set_assemblers` signal that allows every part of the code to modify the list of existing assembler plugins with the added benefit that we can then remove specific plugins, instead of wiping them all and starting from scratch, which is our only option right now.
This is a very early draft on which changes we would need to implement in order to move in such a direction, but I wanted to start a discussion about it first, before spending more time with it. The important bits can be seen in `assembly.h` below. For now I moved the `StokesPreconditioner` assembly out of the `StokesAssembler` class and made it a separate fully-functional assembler plugin. It still connects its `execute` function to the old signal, but we could as well call it via the pointer to the base class stored in `assembler_objects`. The one disadvantage of the new system is that we have less flexibility of providing plugin-specific information into the assembler (see my workaround for `pressure_scaling` below). However, I would argue that we can provide all necessary information in two ways: Either add it to the `scratch` object that is already designed to store all necessary information or provide it via the `SimulatorAccess` class.

I would like to ask for opinion from @tjhei and @bangerth (and all other interested parties). I think the new way would keep the code cleaner and will be easier to understand. Everything else (performance, flexibility) should be similar to the current approach with signals.

PR #2007, opened 2017-11-20
Refactor set_assemblers into separate functions

gassmoeller:
This is on top of #1785 and simplifies `set_assemblers` by outsorcing many parts into the melt handler and newton handler classes.

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -21,6 +21,7 @@
 
 #include <aspect/simulator.h>
 #include <aspect/utilities.h>
+#include <aspect/compat.h>
 #include <aspect/simulator_access.h>
 
 #include <aspect/simulator/assemblers/interface.h>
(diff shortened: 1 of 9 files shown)


gassmoeller:
Rebased on top of #2014, and changed according to our conversation @tjhei. Review is simpler in split view mode.

source/simulator/melt.cc
```diff
@@ -32,7 +32,9 @@
 #include <deal.II/fe/fe_dgp.h>
 #include <deal.II/fe/fe_values.h>
 
-
+#if DEAL_II_VERSION_GTE(9,0,0)
+#include <deal.II/base/std_cxx14/memory.h>
```
tjhei: put this into utilities instead

source/simulator/assembly.cc
```diff
@@ -121,251 +126,119 @@ namespace aspect
...
       {
-        aspect::Assemblers::AdvectionSystem<dim> *advection_system
-          = new aspect::Assemblers::AdvectionSystem<dim>();
-
-        assemblers->advection_system.push_back(
-          std_cxx11::unique_ptr<aspect::Assemblers::AdvectionSystem<dim> > (advection_system));
+        newton_handler->set_assemblers(*assemblers);
 
+        // Let the free surface add its assembler:
+        if (parameters.free_surface_enabled)
```
tjhei: I think I would put this blob (or rather the assert for now) also in the block above (after melt_handler->set_assemblers()) instead of asserting() inside. Or maybe both?

source/simulator/assembly.cc
```diff
@@ -121,251 +126,119 @@ namespace aspect
...
+    // first let the manager delete all existing assemblers:
+    assemblers->reset();
 
-        assemblers->advection_system.push_back(
-          std_cxx11::unique_ptr<aspect::Assemblers::MeltAdvectionSystem<dim> > (melt_advection_system));
+    assemblers->advection_system_assembler_properties.resize(1+introspection.n_compositional_fields);
+    assemblers->advection_system_assembler_on_face_properties.resize(1+introspection.n_compositional_fields);
 
+    // Set up the set of assemblers depending on the mode of equations
+    if (parameters.include_melt_transport)
```
tjhei: where do you currently handle the error if somebody uses melt and newton? Maybe write this as ``if(melt && !newton) ... else if (newton) ... else if (!newton && !melt) ... else error``?

gassmoeller:
Done

source/simulator/newton.cc
```diff
@@ -20,8 +20,14 @@
 
 
 #include <aspect/newton.h>
+#include <aspect/simulator/assemblers/advection.h>
+#include <aspect/simulator/assemblers/stokes.h>
+
 #include <aspect/simulator.h>
 
+#if DEAL_II_VERSION_GTE(9,0,0)
+#include <deal.II/base/std_cxx14/memory.h>
```
tjhei: I think this can be removed, right?

PR #2008, opened 2017-11-21
make prescribed stokes solution work in models with melt transport

jdannberg:
This change makes it possible to use the advection only scheme together with melt transport models. This is made possible by adding support for melt transport in two out of the three prescribed Stokes solution plugins (and adding an assertion in the third one). 

The added tests are copies of existing tests (just with melt transport added), and they reproduce the results of the original tests.


tests/shear_heating_compressible_melt.prm
```diff
@@ -0,0 +1,148 @@
+# This is a copy of the shear_heating_compressible
+# test, exxcept that it uses melt transport.
```
gassmoeller: except

PR #2014, opened 2017-11-22
Implement make_unique

gassmoeller:
This will be useful in a follow up PR, and also in other places.

### DIFF
--- a/include/aspect/compat.h
+++ b/include/aspect/compat.h
@@ -146,5 +146,31 @@ namespace aspect
 }
 #endif
 
+#if !DEAL_II_VERSION_GTE(9,0,0)
+namespace dealii
+{
+  namespace std_cxx14
+  {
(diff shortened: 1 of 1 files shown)


include/aspect/compat.h
```diff
@@ -146,5 +146,25 @@ namespace aspect
 }
 #endif
 
+#if !DEAL_II_VERSION_GTE(9,0,0)
```
tjhei: I think you need to ``#include <std_cxx14/memory.h>`` in an #else block.

tjhei:
Can you also leave a comment in that function that this is just a hack and doesn't allow for constructors with parameters? Otherwise people might be surprised later.

gassmoeller:
Done

bangerth:
What a shame that we can't use C++11 in ASPECT yet :-(

PR #2015, opened 2017-11-23
new static composition field type

tjhei:


source/simulator/core.cc
```diff
@@ -1081,6 +1081,7 @@ namespace aspect
                   = DoFTools::always;
                 break;
               // Non-coupling cases
+              case Parameters<dim>::AdvectionFieldMethod::static_field:
```
gassmoeller: break?

source/simulator/core.cc
```diff
@@ -1212,6 +1213,7 @@ namespace aspect
                 = DoFTools::always;
               break;
             // Non-coupling cases
+            case Parameters<dim>::AdvectionFieldMethod::static_field:
```
gassmoeller: break?

PR #2017, opened 2017-11-28
some formatting and typo fixes

bartniday:


bangerth:
Ah, most excellent! Nice finds!

While you're there, could I ask you for a favor: Would you also change the `n_{dislocation}` to `n_\\text{dislocation}` to make "dislocation" be printed as a word (rather than as a product of a lot of single-letter, italic variable names)? Same for "diffusion".

bartniday:
That should fix it, at least in those particular files

bangerth:
Most appreciated -- thank you very much!

PR #2018, opened 2017-11-28
avoid deadlock when exception is thrown

tjhei:
If an exception is thrown, we will call the TimerOutput destructor,
which will exit subsections and in turn call Timer functions that
trigger MPI communication. This can a) hang if only some processors
throw, or b) report incorrect messages (about Trilinos compress()).

Avoid this problem (at least in c++11 or newer), by reporting the
exception and then aborting before the destructor is executed.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -686,6 +686,19 @@ namespace aspect
     // wait if there is a thread that's still writing the statistics
     // object (set from the output_statistics() function)
     output_statistics_thread.join();
+
+    // Detect if we are being destroyed because an uncaught exception has
+    // been triggered. If so, reset() the TimerOutput class. This will clear
+    // the currently active timing sections. Otherwise, its destructor will try to
+    // do MPI communication when leaving the open sections, which can cause one of the
+    // following problems:
+    // 1. deadlocks or hangs in the MPI commands synchronizing the timers.
+    // 2. Incorrect error reporting about mismatched Trilinos compress() calls.
+    //
+    // In either case this might not show the real message of the exception
+    // that was being triggered.
+    if (std::uncaught_exception())
+      computing_timer.reset();
   }
 
 
--- a/source/main.cc
+++ b/source/main.cc
@@ -441,11 +441,6 @@ int main (int argc, char *argv[])
 {
   using namespace dealii;
 
-  // disable the use of thread. if that is not what you want,
-  // use numbers::invalid_unsigned_int instead of 1 to use as many threads
-  // as deemed useful by TBB
-  Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, /*n_threads =*/ 1);
-
 #ifdef DEBUG
 #ifdef ASPECT_USE_FP_EXCEPTIONS
   // Some implementations seem to not initialize the floating point exception
@@ -459,6 +454,16 @@ int main (int argc, char *argv[])
 
   try
     {
+      // Disable the use of threads. If that is not what you want,
+      // use numbers::invalid_unsigned_int instead of 1 to use as many threads
+      // as deemed useful by TBB.
+      //
+      // Note: we initialize this class inside the try/catch block and not
+      // before, so that the destructor of this instance can react if we are
+      // currently unwinding the stack if an unhandled exception is being
+      // thrown to avoid MPI deadlocks.
+      Utilities::MPI::MPI_InitFinalize mpi_initialization(argc, argv, /*n_threads =*/ 1);
+
       deallog.depth_console(0);
 
       int current_idx = 1;
@@ -712,7 +717,6 @@ int main (int argc, char *argv[])
                 << "Aborting!" << std::endl
                 << "----------------------------------------------------"
                 << std::endl;
-
       return 1;
     }
   catch (std::exception &exc)
@@ -727,7 +731,6 @@ int main (int argc, char *argv[])
                 << "Aborting!" << std::endl
                 << "----------------------------------------------------"
                 << std::endl;
-
       return 1;
     }
   catch (aspect::QuietException &)
--- a/tests/fail/screen-output
+++ b/tests/fail/screen-output
@@ -1,4 +1,5 @@
 
+ERROR: Uncaught exception in MPI_InitFinalize on proc 0. Skipping MPI_Finalize() to avoid a deadlock.
 
 
 Exception 'ExcMessage ("ASPECT can only be run in 2d and 3d but a " "different space dimension is given in the parameter file.")' on rank 0 on processing: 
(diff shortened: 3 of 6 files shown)


tjhei:
The problem with this code is that we now no longer trigger the fancy print-out from main.cc (with rank, exc_name, etc.). Way around that:
1. change deal.II to not do MPI stuff in ``~TimerOutput``
2. make the new code blob above more verbose.
3. somehow avoid calling ``~TimerOutput`` (leaking pointer?)
4. only throw on processor != 0?

source/simulator/core.cc
```diff
@@ -686,6 +686,31 @@ namespace aspect
...
+          {
+            auto eptr = std::current_exception();
+            if (eptr)
+              std::rethrow_exception(eptr);
+          }
+        catch (const std::exception &e)
+          {
+            std::cout << e.what() << std::endl;
+          }
+        throw; // terminate to avoid deadlock in ~TimerOutput
```
bangerth: Hm, you are throwing an exception in a destructor. This isn't allowed.

Also, which exception is this exactly? (I think I don't understand C++ well enough :-) )

source/simulator/core.cc
```diff
@@ -686,6 +686,31 @@ namespace aspect
     // wait if there is a thread that's still writing the statistics
     // object (set from the output_statistics() function)
     output_statistics_thread.join();
+
+    // Detect if we are being destructed because an uncaught exception has been triggered.
+    if (std::uncaught_exception())
+      {
+#ifdef DEAL_II_WITH_CXX11
+        std::cout << "ERROR: uncaught exception detected:" << std::endl;
```
bangerth: maybe say which processor this is on?

tjhei:
> I don't think I quite understand what you are doing here -- and I don't understand why that would lead to the Trilinos error. Can you explain?

Did you see the description of this PR? I am not sure I can explain it much better.

This PR helps with Gerry's crash. I assume that the MPI commands that are sent inside ``~TimerOutput`` interact with the ``compress()`` on the other processors. It doesn't make any sense trying to compute timing results while an exception is being thrown anyways, so I want to abort instead.

source/simulator/core.cc
tjhei: Yes. I want to terminate.

tjhei:
update: I found a much cleaner solution that also works pre-c++11.

source/simulator/core.cc
```diff
@@ -686,6 +686,13 @@ namespace aspect
     // wait if there is a thread that's still writing the statistics
     // object (set from the output_statistics() function)
     output_statistics_thread.join();
+
+    // Detect if we are being destructed because an uncaught exception has
```
bangerth: -> destroyed

source/simulator/core.cc
```diff
@@ -686,6 +686,13 @@ namespace aspect
...
+    // been triggered. If so, reset the TimerOutput class, because it tries to
+    // do MPI communication in its destructor otherwise, which can cause weird
+    // MPI errors or hangs.
+    if (std::uncaught_exception())
+      computing_timer.reset();
```
bangerth: `computing_timer` is a smart pointer, right? Doesn't resetting it cause the destructor of the object pointed to to run? Then what is the difference to before? The smart pointer would have gone out of scope at the end of this function anyway, so would have done the exact same operation.

What you would want to do (maybe?) is to actually leak the object, no?

tjhei: No, the type is ``TimerOutput`` and ``reset()`` does the right thing: https://github.com/dealii/dealii/blob/master/source/base/timer.cc#L681

It just empties the currently active sections, meaning the destructor will not do MPI communication. ;-)

tjhei:
By the way, the whole point is to avoid the MPI communication in ``leave_subsection`` called by the destructor here: https://github.com/dealii/dealii/blob/master/source/base/timer.cc#L350

source/simulator/core.cc
bangerth: Ah, I see - not the kind of `reset()` I had had in mind.

bangerth: That now makes more sense to me. It will avoid *this processor* from initiating communication, though the others will then of course deadlock.

Would it not be nicer to make sure that exceptions can't propagate this far? We know where this particular exception comes from (and could just convert it to an `abort()` or `AssertNothrow`), though I agree that in the general case we have to deal with the situation where *any* kind of exception happens. Your solution looks reasonable in that case, though the deadlock is concerning.

tjhei: > Would it not be nicer to make sure that exceptions can't propagate this far? We know where this particular exception comes from (and could just convert it to an abort() or AssertNothrow),

Sure, we could change this one place, but any ``AssertThrow`` that triggers while we are inside a timing section will cause a similar issue (no reporting and a deadlock or reporting a confusing ``compress`` error that is unrelated). For that reason I prefer the current proposed solution.

At least on my machine (with this patch!) we will cleanly report the error and abort. Because the offending process is killed, MPI notices this and also kills the other processes (see below). This is not true without the patch.

I now get:
```
An error occurred in line <120> of file </ssd/aspect-git/source/particle/interpolator/cell_average.cc> in function (...)

The violated condition was: 
    non_empty_neighbors != 0
Additional information: 
    A cell and all of its neighbors do not contain any particles. The `cell average' interpolation scheme does not support this case.

Aborting!
-------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code.. Per user-direction, the job has been aborted.
-------------------------------------------------------
MPI_ABORT was invoked on rank 6 in communicator MPI_COMM_WORLD 
with errorcode 59.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
```

bangerth: OK, thanks for explaining.

Can you add a comment to this line that explains what you are doing and why, in particular why the `.reset()` call does and why that solves the problem? In essence a summary of the discussion above.

OK to merge if you add this!

tjhei:
updated. I hope it is understandable now.

bangerth:
Thanks for explaining it!

Would it be useful to do a change in `TimerOutput::~TimerOutput` that avoids the communication if there is a pending exception? This does not help ASPECT right away, but may be an issue for other codes that have similar issues.

tjhei:
> Would it be useful to do a change in TimerOutput::~TimerOutput that avoids the communication if there is a pending exception? This does not help ASPECT right away, but may be an issue for other codes that have similar issues.

Probably, yes. I am not sure if there is ever a legitimate case where you want the current behavior...

bangerth:
See https://github.com/dealii/dealii/issues/5557.

Issue #2019, opened 2017-11-28
particle serialization without output crashes


tjhei:
The particle serialization will dereference the nullptr ``output`` in the following code:
```
    World<dim>::save (std::ostringstream &os) const
    {
      aspect::oarchive oa (os);
      oa << (*this);
#if !DEAL_II_VERSION_GTE(9,0,0)
      output->save(os);
#endif
    }
```

bangerth:
Fixed by #2021.

PR #2021, opened 2017-11-30
Fix serialization with deal.II 8.5 and no particle output

gassmoeller:
This fixes an issue that emerged during a discussion with @egpuckett on the mailing list. The serialization of the output plugin of the particle world tried to dereference an uninitialized unique pointer if no plugin was selected. Now we simply ignore the null pointer (it is not used anyway, and in the load function it is not initialized either).

This is not an issue for deal.II 9.0.pre, because we no longer use the plugin in that case. The new test ensures this behavior is tested.


gassmoeller:
Fix #2019

tjhei:
Looks good. Can you please fix the test output?

gassmoeller:
Done

Issue #2024, opened 2017-12-01
std::isnan() and pre C++11


tjhei:
We currently don't require c++11 but we are using ``std::isnan`` in several places which got added in c++11: http://en.cppreference.com/w/cpp/numeric/math/isnan
We also have ``dealii::numbers::is_nan()`` but that is now deprecated in favor of ``std::isnan``.

Before I go change things, does anybody have a better idea than creating ``Utilities::is_nan()`` that uses ``std::is_nan`` or ``dealii::numbers::is_nan``, respectively?

bangerth:
Do you have evidence of compilers that don't actually understand `std::is_nan()`? It may be that all compilers already supported this as part of including C99 functions, even though C++03 didn't require it.

Are these recently introduced places?

tjhei:
> Are these recently introduced places?

No, pretty old. 

What compilers do we support that do not support c++11?

I played with compiler explorer ( https://godbolt.org/g/XnfNNV ) and got the following result:
It seems we are okay with clang >=3.4 (we say deal.II supports clang 3.3), MSVC, gcc with c++11 (>=4.4), recent ICC, recent MSVC

bangerth:
If we had this code for a long time, then I would say let's just not worry about it.

C99 added a bunch of math functions that were not part of C++98 or C++03 (though were finally formally added in C++11), but that all compilers still had in their `<cmath>` header file. I bet `isnan` is one of them. That would explain why we used a function successfully that we strictly speaking could not use, and why nobody ever complained about it.

Any change we could make now is likely unnecessary in view of this, and will be obsolete after 2.0 anyway. So my instinct would be to just ignore things.

tjhei:
agreed.

PR #2025, opened 2017-12-01
Validate MaterialModelOutputs

tjhei:
It is easy to forget filling all values inside ``MaterialModel::evaluate()`` and we actually did for at least two. This patch checks that all NaNs in the outputs are correctly overwritten. I am only doing this on the first cell of the advection assembler and only in debug mode.
Note that ``std::isnan`` should not trigger an exception checking a signaling nan, but it does on my machine, so I have to disable the FPE checks first.

relates: #2024

### DIFF
--- a/source/material_model/composition_reaction.cc
+++ b/source/material_model/composition_reaction.cc
@@ -111,6 +111,9 @@ namespace aspect
           out.thermal_conductivities[i] = k_value;
           out.thermal_expansion_coefficients[i] = thermal_alpha;
           out.compressibilities[i] = 0.0;
+
+          out.entropy_derivative_pressure[i] = 0.0;
+          out.entropy_derivative_temperature[i] = 0.0;
         }
     }
 
--- a/source/material_model/interface.cc
+++ b/source/material_model/interface.cc
@@ -32,6 +32,12 @@
 
 #include <list>
 
+#ifdef DEBUG
+#ifdef ASPECT_USE_FP_EXCEPTIONS
+#include <fenv.h>
+#endif
+#endif
+
 
 namespace aspect
 {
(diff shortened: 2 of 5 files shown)


tjhei:
It looks like many tests and benchmarks are missing the ``entropy_derivative_*``. ;-( Should we go through all of them and set them to zero?
You might argue that it is okay not to set them unless they are used, but I was getting FPEs if you turn on averaging...

bangerth:
Hm, I see the issue with averaging. Other than that, I don't see the point in doing this test: a signaling NaN, by definition, cannot be used for anything. So checking that a number is not a NaN seems superfluous: you're going to hear about it anyway, as soon as something is done with it.

What if you just disabled averaging for quantities that are sNaNs?

tjhei:
> Other than that, I don't see the point in doing this test: a signaling NaN, by definition, cannot be used for anything

Agreed. The only exception is somebody with a setup where FPEs are not supported. One data point: the file Lev posted to the mailing list was not usable in debug mode on my machine because it triggered an FPE.

> What if you just disabled averaging for quantities that are sNaNs?

How would I do that? It seems to me that I can not use ``std::is_nan`` without disabling FPE exceptions. This means I would have to disable these exceptions first... ;-(

bangerth:
Yes, correct. But you'd disable them where you know that you might read signaling_NaNs *on purpose*, namely in the averaging code.

gassmoeller:
This looks nearly ready to merge. Timo can you rebase to a current version?

source/material_model/interface.cc
```diff
@@ -393,6 +399,85 @@ namespace aspect
...
+    {
+      template <int dim>
+      void check(const std::vector<double> &quantity,
+                 const char *name,
+                 const unsigned int n_points)
+      {
+        Assert(quantity.size() == n_points,
+               ExcMessage("Quantity '"+std::string(name)+"' has an incorrect size!"));
+        for (unsigned int i=0; i<n_points; ++i)
+          Assert(!std::isnan(quantity[i]),
```
bangerth: This and the following 4 versions of the function only differ in how the check `isnan(...)`. I think it would easier if you just overload the `isnan` functions and make the `check` function generic for type `T`.

deal.II has `numbers::is_nan()`, but that exists only for `double`. Maybe extend that in the ASPECT utilities namespace to other data types?

source/material_model/composition_reaction.cc
```diff
@@ -111,6 +111,9 @@ namespace aspect
           out.thermal_conductivities[i] = k_value;
           out.thermal_expansion_coefficients[i] = thermal_alpha;
           out.compressibilities[i] = 0.0;
+
+          out.entropy_derivative_pressure[i] = 0.0;
+          out.entropy_derivative_temperature[i] = 0.0;
```
bangerth: Was this a bug before?

source/material_model/interface.cc
```diff
@@ -393,6 +399,85 @@ namespace aspect
...
+    }
+
+    template <int dim>
+    void
+    MaterialModelOutputs<dim>::validate() const
+    {
+#ifdef DEBUG
+#ifdef ASPECT_USE_FP_EXCEPTIONS
+      // disable floating point exceptions so we can report nicer errors:
+      fedisableexcept(FE_DIVBYZERO|FE_INVALID);
```
bangerth: You need this because you want to deal with signaling NaNs, right? Maybe say so in the comment.

source/material_model/interface.cc
```diff
@@ -393,6 +399,85 @@ namespace aspect
       reaction_terms(n_points, std::vector<double>(n_comp, numbers::signaling_nan<double>()))
     {}
 
+    namespace internal
```
bangerth: make this namespace anonymous

source/simulator/assembly.cc
```diff
@@ -822,6 +822,21 @@ namespace aspect
 
     material_model->evaluate(scratch.material_model_inputs,
                              scratch.material_model_outputs);
+
+#ifdef DEBUG
+    {
+      // Check that the user filled all entries inside the material model and
+      // hasn't messed with the size of the objects:
+      static bool first = true;
+      if (first)
```
bangerth: Maybe add a comment to this to the effect that this really only checks the first time a matrix is assembled, but that that already catches the majority of cases?

source/material_model/composition_reaction.cc
bangerth: I thought this is a bug, but the array is filled with sNaNs in the current state.

bangerth:
We didn't have much of a consensus here, and it's quite heavy machinery for what I'm not sure is a substantial gain. Let me close this.

PR #2026, opened 2017-12-05
inclusion benchmarks: use seconds not years

tjhei:
The graphical output for the inclusion benchmark were set to "Use years
in output instead of seconds" which doesn't make much sense for a
problem on a unit square. The computed errors are independent of this
change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,8 @@
 /benchmarks/**/cmake_install.cmake
 /benchmarks/**/lib*.so
 /benchmarks/**/output/
+/benchmarks/**/aspect
+/benchmarks/**/temp.prm
 /benchmarks/tangurnis/*/*csv
 /benchmarks/tangurnis/*/solution*
 /build*/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -17,6 +17,8 @@
 /benchmarks/**/cmake_install.cmake
 /benchmarks/**/lib*.so
 /benchmarks/**/output/
+/benchmarks/**/aspect
+/benchmarks/**/temp.prm
```
bangerth: Not wrong but unrelated

tjhei:
> What to do?

Notify @tjhei so he can restart the virtual machine? ;-) Should be working again soon.

tjhei:
Oh, and I wouldn't be surprised if tests break.

bangerth:
Then it's good that I didn't just merge :-)

Be that as it may, the tester seems happy now, so I'll merge.

tjhei:
Excellent. That means we did not check the graphical output of the inclusion benchmark.

PR #2028, opened 2017-12-12
[WIP] Add coastlines dataset

gassmoeller:
For many applications it would be useful to have a coastlines visualization for spatial reference. Add one.


MFraters:
I could not find a reference in the file where it is from. Could you add it somewhere? Could be as a comment in the file or as a separate readme.

bangerth:
Out of curiosity, how does one use such a file? Just load it in visit our paraview?

bobmyhill:
I like this, with the changes suggested by @MFraters and @bangerth
In case a range of resolutions/different geographical types are needed, GSHHG is licensed under LGPL http://www.soest.hawaii.edu/wessel/gshhg/

bangerth:
I didn't know I had any comments -- but I wonder what the `.vtp` ending is for. Isn't this just a regular `.vtu` file?

gassmoeller:
The problem with this PR is that I can not find the reference I created this file from (not even the script I used to create it). Thus I would not like to merge it in the current state. I think @jaustermann mentioned she has an alternative dataset during the hackathon, or do I misremember that Jacky?

@bangerth: .vtu is for unstructured grid data, while .vtp is the canonical ending for Polygon data (there is also .vts for structured grids). Not sure if it actually makes a difference while reading, I just followed the convections.

jaustermann:
Bob already linked a good option: https://www.ngdc.noaa.gov/mgg/shorelines/
Don't know if there's an easy conversion to vtk. Matlab has tools to read this in: https://www.mathworks.com/help/map/examples/converting-coastline-data-gshhs-to-shapefile-format.html
There reference is Wessel, P., and W. H. F. Smith, 1996, A global self-consistent, hierarchical, high-resolution shoreline database, Journal of Geophysical Research, Vol. 101, pp. 8741-8743.
In matlab you can read in the bin files that come with matlab or download the ones from noaa, which might be a bit more up-to-date (I don't think it matters though). I'm not sure how to write from matlab to vtu or vtp, but that shouldn't be too difficult.

cedrict:
so I have all those data in vtk format ( 
http://cedricthieulot.net/earth3d.html, at the bottom ), but I don't 
really know where they come from.

I could probably make vtu files if you tell me which files you wish to 
convert.

Ce/

gassmoeller:
We are just looking for a low-medium resolution file for coastlines to put into the data/ subdirectory so that users can easily visualize 3D spherical models. Could you create a .vtu for coastlines just like shown in the GPlates velocities figures of your linked website and either add it as a PR, or send them to me?

gassmoeller:
Since there exist good links (posted above), and we do not really want to store more data than necessary in the repo, lets close this.

PR #2029, opened 2017-12-13
Include more version information in header

gassmoeller:
As mentioned in the changelog entry. This change will only require a rerun of cmake and a recompilation of global.cc if the git HEAD changes, no full recompilation required. I several times wished I had this output for old models.
+1 for reproducibility of old models.

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -2838,8 +2838,8 @@ something like this (numbers will all be different, of course):
 \begin{lstlisting}[frame=single,language=ksh]
 -----------------------------------------------------------------------------
 -- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.
---     . version 2.0.0-pre (git revision c620893)
---     . using deal.II 9.0.0-pre (git revision 952baa0)
+--     . version 2.0.0-pre (include_dealii_version, c20eba0)
+--     . using deal.II 9.0.0-pre (master, 952baa0)
 --     . using Trilinos 12.10.1
 --     . using p4est 2.0.0
 --     . running in DEBUG mode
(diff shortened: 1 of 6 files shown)


MFraters:
I really like this change! This will help reproducibility and tracking problems a lot. I just have one request, you are defining `ASPECT_GIT_BRANCH`, but you are not using it. I think that it would be quite useful to also include that in the header.

gassmoeller:
I tested including the branch name, but it does not look nice any more if there is too much information in those lines. Also you can find out the branch name by using `git branch --contains COMMIT`, is that sufficient? I would keep the branch information in the header for now (it might be useful to find out which branch was used to compile a build folder), but I would not include it in the header.

doc/manual/manual.tex
```diff
@@ -2886,9 +2896,21 @@ \subsection{Overview}
 ...
 \end{lstlisting}
 
-This output was produced by a parameter file that, among other settings,
-contained the following values (we will discuss many such input files in
-Section~\ref{sec:cookbooks}:
+The output starts with a header that lists the used \aspect{}, \dealii{}, and
```
MFraters: and trilinos

doc/manual/manual.tex
```diff
@@ -2886,9 +2896,21 @@ \subsection{Overview}
...
+\pfrst{} versions as well as compiler settings and the number of parallel
```
MFraters: With compiler settings you mean whether it is debug or optimized? If it is just that, then you might just want to say that.

gassmoeller:
Thanks, good points. Actually I changed my mind about the branch. On development branches I often amend or squash commits, and in these cases the commit hash is useless. I will try to include the branch name in a nice way.

tjhei:
Does this mean that any commit will now trigger a rebuild of global.cc and a relink? Or if the revision.h only updated when you run cmake manually? Either way, I think we should not output the ASPECT sha1 for that reason.

gassmoeller:
@MFraters I addressed your comments.

@tjhei: Yes every commit will trigger a run of cmake, a compilation of global.cc and a relink. I would argue though that if you made a commit you made most of this anyway (so it is not much more time), and the gain in terms of reproducibility is worth it (compared to our current state where models are essentially not reproducible if they are not done with a release). I would have needed the commit hash of old models on several occasions.

gassmoeller:
I think @tjhei and @bangerth agreed to proceed with this Pr for now, and take a look if it affects workflows too much, right?

tjhei:
Let's see how this plays out.

PR #2030, opened 2017-12-13
Add derivatives to visco plastic material model

MFraters:
Adding finite difference derivatives to the visco-plastic material model, so it can be used with the Newton solver. With this finite difference model, it should be relatively easy to add derivatives to other material models. 

The finite difference could actually be done somewhere higher up or abstracted into a single function.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1062,10 +1062,12 @@ namespace aspect
     std::vector<Operator> create_model_operator_list(const std::vector<std::string> &operator_names);
 
     /**
-     * Create matrix with unit at independent indices
+     * A function that returns a SymmetricTensor, whose entries are zero, except for
+     * the k'th component, which is set to one. If k is not on the main diagonal the
+     * resulting tensor is symmetrized.
      */
     template <int dim>
-    SymmetricTensor<2,dim> symmetric_independent_component_matrix (const unsigned int k);
+    SymmetricTensor<2,dim> nth_basis_for_symmetric_tensors (const unsigned int k);
 
   }
 }
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1060,6 +1060,13 @@ namespace aspect
      * entry in the list must match one of the allowed operations.
      */
     std::vector<Operator> create_model_operator_list(const std::vector<std::string> &operator_names);
+
+    /**
+     * Create matrix with unit at independent indices
+     */
+    template <int dim>
+    SymmetricTensor<2,dim> symmetric_independent_component_matrix (const unsigned int k);
+
   }
 }
 
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -358,12 +358,6 @@ namespace aspect
       MaterialModel::MaterialModelDerivatives<dim> *derivatives;
       derivatives = out.template get_additional_output<MaterialModel::MaterialModelDerivatives<dim> >();
 
-      // If we use the Newton solver, get the derivative scaling factor,
-      // otherwise set it to zero.
-      double derivative_scaling_factor = 0.0;
-      if (this->get_parameters().nonlinear_solver == Parameters<dim>::NonlinearSolver::Newton_Stokes)
-        derivative_scaling_factor = this->get_newton_handler().get_newton_derivative_scaling_factor();
-
       // Loop through points
       for (unsigned int i=0; i < in.temperature.size(); ++i)
         {
@@ -408,7 +402,10 @@ namespace aspect
               // TODO: This is only consistent with viscosity averaging if the arithmetic averaging
               // scheme is chosen. It would be useful to have a function to calculate isostress viscosities.
               const std::vector<double> composition_viscosities =
-                calculate_isostrain_viscosities(volume_fractions, pressure, temperature, composition, strain_rate,viscous_flow_law,yield_mechanism);
+                calculate_isostrain_viscosities(volume_fractions, pressure,
+                                                temperature, composition,
+                                                strain_rate, viscous_flow_law,
+                                                yield_mechanism);
 
               std::vector<SymmetricTensor<2,dim> > composition_viscosities_derivatives(volume_fractions.size());
               std::vector<double> composition_dviscosities_dpressure(volume_fractions.size());
@@ -420,55 +417,66 @@ namespace aspect
               out.viscosities[i] = average_value(composition, composition_viscosities, viscosity_averaging);
 
               // compute derivatives if nessesary
-              if (derivative_scaling_factor != 0 && derivatives != NULL)
+              if (derivatives != NULL)
                 {
                   const double finite_difference_accuracy = 1e-7;
 
                   // For each independent component, compute the derivative.
-                  for (unsigned int independent_component_k = 0; independent_component_k < SymmetricTensor<2,dim>::n_independent_components; independent_component_k++)
+                  for (unsigned int component = 0; component < SymmetricTensor<2,dim>::n_independent_components; ++component)
                     {
-                      const TableIndices<2> indices_ij = SymmetricTensor<2,dim>::unrolled_to_component_indices (independent_component_k);
-                      SymmetricTensor<2,dim> strain_rate_component = strain_rate + std::max(std::fabs(strain_rate[indices_ij]), min_strain_rate) * finite_difference_accuracy
-                                                                     * Utilities::symmetric_independent_component_matrix<dim>(independent_component_k);
-                      std::vector<double> eta_component = calculate_isostrain_viscosities(volume_fractions, pressure, temperature, composition, strain_rate_component,viscous_flow_law,yield_mechanism);
+                      const TableIndices<2> strain_rate_indices = SymmetricTensor<2,dim>::unrolled_to_component_indices (component);
+
+                      const SymmetricTensor<2,dim> strain_rate_difference = strain_rate
+                                                                            + std::max(std::fabs(strain_rate[strain_rate_indices]), min_strain_rate)
+                                                                            * finite_difference_accuracy
+                                                                            * Utilities::nth_basis_for_symmetric_tensors<dim>(component);
+                      std::vector<double> eta_component =
+                        calculate_isostrain_viscosities(volume_fractions, pressure,
+                                                        temperature, composition,
+                                                        strain_rate_difference,
+                                                        viscous_flow_law,yield_mechanism);
 
                       // For each composition of the independent component, compute the derivative.
-                      for (unsigned int composition_i = 0; composition_i < eta_component.size(); composition_i++)
+                      for (unsigned int composition_index = 0; composition_index < eta_component.size(); ++composition_index)
                         {
                           // compute the difference between the viscosity with and without the strain-rate difference.
-                          double derivative_component = eta_component[composition_i] - composition_viscosities[composition_i];
-                          if (derivative_component != 0)
+                          double viscosity_derivative = eta_component[composition_index] - composition_viscosities[composition_index];
+                          if (viscosity_derivative != 0)
                             {
                               // when the difference is non-zero, divide by the difference.
-                              derivative_component /= std::max(std::fabs(strain_rate_component[indices_ij]), min_strain_rate) * finite_difference_accuracy;
+                              viscosity_derivative /= std::max(std::fabs(strain_rate_difference[strain_rate_indices]), min_strain_rate)
+                                                      * finite_difference_accuracy;
                             }
-                          composition_viscosities_derivatives[composition_i][indices_ij] = derivative_component;
+                          composition_viscosities_derivatives[composition_index][strain_rate_indices] = viscosity_derivative;
                         }
                     }
 
                   /**
                    * Now compute the derivative of the viscosity to the pressure
                    */
-                  double pressure_difference = in.pressure[i] + (std::fabs(in.pressure[i]) * finite_difference_accuracy);
+                  const double pressure_difference = in.pressure[i] + (std::fabs(in.pressure[i]) * finite_difference_accuracy);
 
-                  std::vector<double> pressure_difference_eta = calculate_isostrain_viscosities(volume_fractions, pressure_difference, temperature, composition, strain_rate,viscous_flow_law,yield_mechanism);
+                  const std::vector<double> viscosity_difference =
+                    calculate_isostrain_viscosities(volume_fractions, pressure_difference,
+                                                    temperature, composition, strain_rate,
+                                                    viscous_flow_law, yield_mechanism);
 
 
-                  for (unsigned int composition_i = 0; composition_i < pressure_difference_eta.size(); composition_i++)
+                  for (unsigned int composition_index = 0; composition_index < viscosity_difference.size(); ++composition_index)
                     {
-                      double deriv_pressure = pressure_difference_eta[composition_i] - composition_viscosities[composition_i];
-                      if (pressure_difference_eta[composition_i] != 0)
+                      double viscosity_derivative = viscosity_difference[composition_index] - composition_viscosities[composition_index];
+                      if (viscosity_difference[composition_index] != 0)
                         {
                           if (in.pressure[i] != 0)
                             {
-                              deriv_pressure /= std::fabs(in.pressure[i]) * finite_difference_accuracy;
+                              viscosity_derivative /= std::fabs(in.pressure[i]) * finite_difference_accuracy;
                             }
                           else
                             {
-                              deriv_pressure = 0;
+                              viscosity_derivative = 0;
                             }
                         }
-                      composition_dviscosities_dpressure[composition_i] = deriv_pressure;
+                      composition_dviscosities_dpressure[composition_index] = viscosity_derivative;
                     }
 
                   double viscosity_averaging_p = 0; // Geometric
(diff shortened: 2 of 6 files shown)


naliboff:
@MFraters - Great! Given that there will likely be a few more material models added that use the Newton solver, perhaps it would be worth moving the finite difference section higher up as you suggested? Regardless, I'll take go through this today or tomorrow.

source/material_model/visco_plastic.cc
```diff
@@ -397,14 +406,127 @@ namespace aspect
...
               // The isostrain condition implies that the viscosity averaging should be arithmetic (see above).
               // We have given the user freedom to apply alternative bounds, because in diffusion-dominated
               // creep (where n_diff=1) viscosities are stress and strain-rate independent, so the calculation
               // of compositional field viscosities is consistent with any averaging scheme.
               out.viscosities[i] = average_value(composition, composition_viscosities, viscosity_averaging);
 
+              // compute derivatives if nessesary
+              if (derivative_scaling_factor != 0 && derivatives != NULL)
+                {
+                  const double finite_difference_accuracy = 1e-7;
```
gassmoeller: This parts needs significantly more explanation. Why do you split up the components of the strain rate? I would prefer if we can move a lot of this into a separate function (either private in this material model or in `aspect::Utilities`). Is this currently only implemented for 2D (you only access elements 0 or 1)? What happens if strain_rate is close to zero?

source/material_model/visco_plastic.cc
```diff
@@ -544,6 +666,14 @@ namespace aspect
     void
     ViscoPlastic<dim>::declare_parameters (ParameterHandler &prm)
     {
+      prm.enter_subsection("Compositional fields");
```
gassmoeller: You do not need to declare a parameter again, this just needs to be done by the plugin owning this parameter.

source/material_model/visco_plastic.cc
```diff
@@ -730,7 +860,13 @@ namespace aspect
     ViscoPlastic<dim>::parse_parameters (ParameterHandler &prm)
     {
       // increment by one for background:
-      const unsigned int n_fields = this->n_compositional_fields() + 1;
+      unsigned int n_fields = 0;
```
gassmoeller: Why does the above not work? In theory you should have access to that variable already.

source/material_model/visco_plastic.cc
```diff
@@ -182,7 +183,7 @@ namespace aspect
       // The first time this function is called (first iteration of first time step)
       // a specified "reference" strain rate is used as the returned value would
       // otherwise be zero.
-      const double edot_ii = ( (this->get_timestep_number() == 0 && strain_rate.norm() <= std::numeric_limits<double>::min())
+      const double edot_ii = ( (&(this->get_simulator()) != nullptr && this->get_timestep_number() == 0 && strain_rate.norm() <= std::numeric_limits<double>::min())
```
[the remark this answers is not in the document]
MFraters: Yes, during the test of this material model, there is no simulator. So this is to make sure that the test actually works.

source/material_model/visco_plastic.cc
```diff
@@ -354,13 +355,21 @@ namespace aspect
     evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
              MaterialModel::MaterialModelOutputs<dim> &out) const
     {
+      // set up additional output for the derivatives
+      MaterialModel::MaterialModelDerivatives<dim> *derivatives;
+      derivatives = out.template get_additional_output<MaterialModel::MaterialModelDerivatives<dim> >();
+
+      const double derivative_scaling_factor = &(this->get_simulator()) != nullptr ? (this->get_parameters().nonlinear_solver == Parameters<dim>::NonlinearSolver::Newton_Stokes ? this->get_newton_handler().get_newton_derivative_scaling_factor() : 0) : 1;
```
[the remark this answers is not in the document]
MFraters: If I remove the const I can spell it out in if-else statements. I could also add a comment explaining the line. Something like: If there is no simulator, it is a test environment and we want to enable the derivatives, so we set this variable to one. If we do not use the Newton solver, we do not need the derivatives, so we set this variable to zero. Otherwise, we use the value from the newton handler.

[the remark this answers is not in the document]
bangerth: If having the `const` leads to code that is hard to read, then get rid of it. The comment sounds good to me!

source/material_model/visco_plastic.cc
MFraters: So this, and the next comment are the same problem as above, the `this->n_compositional_fields()` is not available in when it is being tested.

source/material_model/visco_plastic.cc
MFraters: Good point. I have moved the splitting of the components into a separate function, which allows for much less code :) It works now both in 2d and 3d. I also added more comments in the code.

MFraters:
It will be difficult to check the things I would like to check in a way where there is a simulator. We can of course discuss to see if there is a way to do it.

MFraters:
@gassmoeller I guess I will have to make different tests for them after all, because I can't change the material model member variable `Viscosity averaging scheme` through creating a new parameter object and use `simulator_access.get_material_model().parse_parameters(prm);`, because get_material_model() returns a constant reference, so I get as an error:
```
error: passing ‘const aspect::MaterialModel::Interface<2>’ as ‘this’ argument discards qualifiers [-fpermissive]
```
Or do you know a way around this?

gassmoeller:
Check by my office some time today, I am sure we can find a solution to this.

MFraters:
@gassmoeller I think that I have addressed all your comments.

bangerth:
@gassmoeller -- are you working this through with @MFraters or do you want me to also take a look?

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,82 @@ namespace aspect
...
+              // compute derivatives if nessesary
+              if (derivative_scaling_factor != 0 && derivatives != NULL)
+                {
+                  const double finite_difference_accuracy = 1e-7;
+
+                  // For each independent component, compute the derivative.
+                  for (unsigned int independent_component_k = 0; independent_component_k < SymmetricTensor<2,dim>::n_independent_components; independent_component_k++)
+                    {
+                      const TableIndices<2> indices_ij = SymmetricTensor<2,dim>::unrolled_to_component_indices (independent_component_k);
+                      SymmetricTensor<2,dim> strain_rate_component = strain_rate + std::max(std::fabs(strain_rate[indices_ij]), min_strain_rate) * finite_difference_accuracy
```
gassmoeller: lets call this `strain_rate_difference` to stay consistent with the `pressure_difference` below.

gassmoeller: also can this be `const`?

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,82 @@ namespace aspect
...
+                                                                     * Utilities::symmetric_independent_component_matrix<dim>(independent_component_k);
+                      std::vector<double> eta_component = calculate_isostrain_viscosities(volume_fractions, pressure, temperature, composition, strain_rate_component,viscous_flow_law,yield_mechanism);
+
+                      // For each composition of the independent component, compute the derivative.
+                      for (unsigned int composition_i = 0; composition_i < eta_component.size(); composition_i++)
```
gassmoeller: call this `composition_index` and do `++composition_index`

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,82 @@ namespace aspect
...
+                        {
+                          // compute the difference between the viscosity with and without the strain-rate difference.
+                          double derivative_component = eta_component[composition_i] - composition_viscosities[composition_i];
+                          if (derivative_component != 0)
+                            {
+                              // when the difference is non-zero, divide by the difference.
+                              derivative_component /= std::max(std::fabs(strain_rate_component[indices_ij]), min_strain_rate) * finite_difference_accuracy;
```
gassmoeller: How is the term on the right hand side different from what you add in line 431? If it is the same: Give it a name and define it in line 430.

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,82 @@ namespace aspect
...
+                      const TableIndices<2> indices_ij = SymmetricTensor<2,dim>::unrolled_to_component_indices (independent_component_k);
+                      SymmetricTensor<2,dim> strain_rate_component = strain_rate + std::max(std::fabs(strain_rate[indices_ij]), min_strain_rate) * finite_difference_accuracy
+                                                                     * Utilities::symmetric_independent_component_matrix<dim>(independent_component_k);
+                      std::vector<double> eta_component = calculate_isostrain_viscosities(volume_fractions, pressure, temperature, composition, strain_rate_component,viscous_flow_law,yield_mechanism);
+
+                      // For each composition of the independent component, compute the derivative.
+                      for (unsigned int composition_i = 0; composition_i < eta_component.size(); composition_i++)
+                        {
+                          // compute the difference between the viscosity with and without the strain-rate difference.
+                          double derivative_component = eta_component[composition_i] - composition_viscosities[composition_i];
```
gassmoeller: `viscosity_derivative`

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,82 @@ namespace aspect
...
 
+              // compute derivatives if nessesary
+              if (derivative_scaling_factor != 0 && derivatives != NULL)
+                {
+                  const double finite_difference_accuracy = 1e-7;
+
+                  // For each independent component, compute the derivative.
+                  for (unsigned int independent_component_k = 0; independent_component_k < SymmetricTensor<2,dim>::n_independent_components; independent_component_k++)
+                    {
+                      const TableIndices<2> indices_ij = SymmetricTensor<2,dim>::unrolled_to_component_indices (independent_component_k);
```
gassmoeller: rename this to `strain_rate_indices`

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,82 @@ namespace aspect
...
+                              derivative_component /= std::max(std::fabs(strain_rate_component[indices_ij]), min_strain_rate) * finite_difference_accuracy;
+                            }
+                          composition_viscosities_derivatives[composition_i][indices_ij] = derivative_component;
+                        }
+                    }
+
+                  /**
+                   * Now compute the derivative of the viscosity to the pressure
+                   */
+                  double pressure_difference = in.pressure[i] + (std::fabs(in.pressure[i]) * finite_difference_accuracy);
```
gassmoeller: `const`

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,82 @@ namespace aspect
...
+
+                  std::vector<double> pressure_difference_eta = calculate_isostrain_viscosities(volume_fractions, pressure_difference, temperature, composition, strain_rate,viscous_flow_law,yield_mechanism);
```
gassmoeller: `viscosity_difference` and `const`?

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,82 @@ namespace aspect
...
+
+
+                  for (unsigned int composition_i = 0; composition_i < pressure_difference_eta.size(); composition_i++)
+                    {
+                      double deriv_pressure = pressure_difference_eta[composition_i] - composition_viscosities[composition_i];
```
gassmoeller: rename to `viscosity_derivative`

source/utilities.cc
```diff
@@ -2532,6 +2532,21 @@ namespace aspect
     }
 
 
+    template <int dim>
+    SymmetricTensor<2,dim>
+    symmetric_independent_component_matrix (const unsigned int k)
+    {
+      // somehow I need this, because otherwise it complains that I am passing 3 arguments and it only takes 2...
+      const bool boolian = k < SymmetricTensor<2,dim>::n_independent_components;
```
gassmoeller: @bangerth: I have tested this and found the same strange behavior. The following code does not work with deal.II 8.5, although it should:
```
const unsigned int k=0;
Assert(k< SymmetricTensor<2,dim>::n_independent_components,ExcMessage(""))
```

The compiler error message is:
```
/home/rengas/Software/Aspect-Versionen/aspect/source/material_model/visco_plastic.cc:815:80: error: macro "Assert" passed 3 arguments, but takes just 2
       Assert(k< SymmetricTensor<2,dim>::n_independent_components,ExcMessage(""));
                                                                                ^
```

include/aspect/utilities.h
```diff
@@ -1060,6 +1060,13 @@ namespace aspect
      * entry in the list must match one of the allowed operations.
      */
     std::vector<Operator> create_model_operator_list(const std::vector<std::string> &operator_names);
+
+    /**
+     * Create matrix with unit at independent indices
+     */
+    template <int dim>
+    SymmetricTensor<2,dim> symmetric_independent_component_matrix (const unsigned int k);
```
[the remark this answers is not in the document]
bangerth: For the text, it should be "...WHOSE entries are zero..."

Fundamentally, this function returns the *elements of a basis* for the set of symmetric tensors since every tensor can be written as a linear combination of the tensors returned by this function. So maybe something like `nth_basis_for_symmetric_tensors()`?

source/utilities.cc
bangerth: As discussed on the way to lunch, you'll have to put a set of (parentheses) around the condition or the preprocessor is going to interpret the comma in `2,dim` as one that separates macro arguments.

source/material_model/visco_plastic.cc
MFraters: one uses the strain-rate and the other one uses the strain-rate plus the finite difference of the strain-rate.

MFraters:
It is ready for an new round of reviews :)

source/material_model/visco_plastic.cc
```diff
@@ -353,13 +354,24 @@ namespace aspect
...
     {
+      // set up additional output for the derivatives
+      MaterialModel::MaterialModelDerivatives<dim> *derivatives;
+      derivatives = out.template get_additional_output<MaterialModel::MaterialModelDerivatives<dim> >();
+
+      // If we use the Newton solver, get the derivative scaling factor,
+      // otherwise set it to zero.
+      double derivative_scaling_factor = 0.0;
+      if (this->get_parameters().nonlinear_solver == Parameters<dim>::NonlinearSolver::Newton_Stokes)
+        derivative_scaling_factor = this->get_newton_handler().get_newton_derivative_scaling_factor();
```
bangerth: Hm, why do you actually need this in the material model? Shouldn't the scaling factor be applied only at the one or two places in the assembly where we compute the Newton matrix? If that were the case, you wouldn't have to deal with it in every single material model.

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,83 @@ namespace aspect
               // TODO: This is only consistent with viscosity averaging if the arithmetic averaging
               // scheme is chosen. It would be useful to have a function to calculate isostress viscosities.
               const std::vector<double> composition_viscosities =
-                calculate_isostrain_viscosities(volume_fractions, pressure, temperature, composition, in.strain_rate[i],viscous_flow_law,yield_mechanism);
+                calculate_isostrain_viscosities(volume_fractions, pressure, temperature, composition, strain_rate,viscous_flow_law,yield_mechanism);
```
bangerth: while you're here, break the line into shorter ones around 80 characters of length, and put a space after every comma separating the arguments

source/material_model/visco_plastic.cc
```diff
@@ -396,14 +408,83 @@ namespace aspect
...
+                              viscosity_derivative /= std::max(std::fabs(strain_rate_difference[strain_rate_indices]), min_strain_rate) * finite_difference_accuracy;
+                            }
+                          composition_viscosities_derivatives[composition_index][strain_rate_indices] = viscosity_derivative;
+                        }
+                    }
+
+                  /**
+                   * Now compute the derivative of the viscosity to the pressure
+                   */
+                  const double pressure_difference = in.pressure[i] + (std::fabs(in.pressure[i]) * finite_difference_accuracy);
```
bangerth: there are several long lines here. can you break them as well?

source/utilities.cc
```diff
@@ -2532,6 +2532,21 @@ namespace aspect
     }
 
 
+    template <int dim>
+    SymmetricTensor<2,dim>
+    nth_basis_for_symmetric_tensors (const unsigned int k)
+    {
+      // somehow I need this, because otherwise it complains that I am passing 3 arguments and it only takes 2...
+      const bool boolian = k < SymmetricTensor<2,dim>::n_independent_components;
+      Assert(boolian, ExcMessage("The component is larger then the amount of independent components in the matrix.") );
```
bangerth: You can write this as
```
  Assert((k < SymmetricTensor<2,dim>::n_independent_components), 
             ExcMessage("The component is larger then the amount of independent components in the matrix.") );
```
This has the advantage that the violated condition appears in the screen output.

(Unrelated: The correct spelling is "boolean" because it is named after Mr. Boole.)

source/material_model/visco_plastic.cc
MFraters: So the factor itself is actually not used to compute something here. It is only used to determine wheter the derivatives should be computed. Because when this scaling factor is zero, which may happen even when `derivatives != NULL`, there is no use in computing the derivatives.

bangerth:
About the derivative scaling factor, as I mentioned yesterday my preference would be if the assembly loop simply didn't ask about information it doesn't need, rather than every material model separately trying to figure out whether the information it is being asked to deliver is actually used.

bangerth:
We convinced ourselves that the code is correct as is. So I'll merge.

PR #2031, opened 2017-12-14
add zenodo json file

tjhei:
As discussed with @ljhwang . This will fill the author information in zenodo whenever we do a release.


ljhwang:
The .json file will override zenodo defaults such that the entire commit list of contributors is not consider as code authors in the citation.

gassmoeller:
Looks good. Thanks!

PR #2032, opened 2017-12-16
reuse compositional field matrix memory

tjhei:
This patch lets all compositional fields share the same system matrix.
As we assemble the matrices before each solve anyways, this doesn't cost
any extra time. Sadly, handling of constraints is more complicated as
the value of the boundary conditions can be different for each field.
For this, we now have separate constraint matrices for each field.


tjhei:
WIP obviously. Code is still somewhat messy, but I want to see if any tests fail with this change.

bangerth:
Without having looked at the patch itself, can you explain why you want to do this? It seems like a substantial complification for gains that are not clear to me.

(To wit, the compositional fields never couple with anything else. So these matrix blocks are actually quite empty, compared to the velocity blocks, for example. I doubt that they take up any noticeable amount of memory at all, compared to everything else for which we store matrices.)

tjhei:
> Without having looked at the patch itself, can you explain why you want to do this?

@naliboff needs 6+ compositional fields and is running into the problem that he needs a large amount of memory.

Here is an example of a 3d shell with 8 compositional fields (running on 20 cores):
```
Number of degrees of freedom: 1,272,170 (314,409+14,534+104,803+104,803+104,803+104,803+104,803+104,803+104,803+104,803+104,803)
Total system matrix memory consumption: 1099.99 MB.
Total system matrix nnz: 91,220,739
system matrix nnz by block:
  40,306,039   3,346,223           1           1           1           1           1           1           1           1           1
   3,346,223       4,613           1           1           1           1           1           1           1           1           1
           1           1   4,029,589           1           1           1           1           1           1           1           1
           1           1           1   5,023,493           1           1           1           1           1           1           1
           1           1           1           1   5,023,493           1           1           1           1           1           1
           1           1           1           1           1   5,023,493           1           1           1           1           1
           1           1           1           1           1           1   5,023,493           1           1           1           1
           1           1           1           1           1           1           1   5,023,493           1           1           1
           1           1           1           1           1           1           1           1   5,023,493           1           1
           1           1           1           1           1           1           1           1           1   5,023,493           1
           1           1           1           1           1           1           1           1           1           1   5,023,493

Total system preconditioner matrix memory consumption: 737.59 MB.
Total system preconditioner matrix nnz: 59,553,838
  15,085,359           1           1           1           1           1           1           1           1           1           1
           1     250,836           1           1           1           1           1           1           1           1           1
           1           1   4,029,589           1           1           1           1           1           1           1           1
           1           1           1   5,023,493           1           1           1           1           1           1           1
           1           1           1           1   5,023,493           1           1           1           1           1           1
           1           1           1           1           1   5,023,493           1           1           1           1           1
           1           1           1           1           1           1   5,023,493           1           1           1           1
           1           1           1           1           1           1           1   5,023,493           1           1           1
           1           1           1           1           1           1           1           1   5,023,493           1           1
           1           1           1           1           1           1           1           1           1   5,023,493           1
           1           1           1           1           1           1           1           1           1           1   5,023,493
```
new:
```
Total system matrix memory consumption: 699.39 MB.
Total system matrix nnz: 56,215,741
system matrix nnz by block: 
  40,306,039   3,346,223           1           1           1           1           1           1           1           1           1
   3,346,223       4,613           1           1           1           1           1           1           1           1           1
           1           1   4,029,589           1           1           1           1           1           1           1           1
           1           1           1   5,023,493           1           1           1           1           1           1           1
           1           1           1           1      22,779           1           1           1           1           1           1
           1           1           1           1           1      22,779           1           1           1           1           1
           1           1           1           1           1           1      22,779           1           1           1           1
           1           1           1           1           1           1           1      22,779           1           1           1
           1           1           1           1           1           1           1           1      22,779           1           1
           1           1           1           1           1           1           1           1           1      22,779           1
           1           1           1           1           1           1           1           1           1           1      22,779

Total system preconditioner matrix memory consumption: 336.99 MB.
Total system preconditioner matrix nnz: 24,548,840
  15,085,359           1           1           1           1           1           1           1           1           1           1
           1     250,836           1           1           1           1           1           1           1           1           1
           1           1   4,029,589           1           1           1           1           1           1           1           1
           1           1           1   5,023,493           1           1           1           1           1           1           1
           1           1           1           1      22,779           1           1           1           1           1           1
           1           1           1           1           1      22,779           1           1           1           1           1
           1           1           1           1           1           1      22,779           1           1           1           1
           1           1           1           1           1           1           1      22,779           1           1           1
           1           1           1           1           1           1           1           1      22,779           1           1
           1           1           1           1           1           1           1           1           1      22,779           1
           1           1           1           1           1           1           1           1           1           1      22,779

```
So 1836MB vs 1035MB, which is quite substantial or not?

p.s.: I wonder if the 22k unnecessary DoFs are because of hanging nodes, boundary conditions, or both.. Weird.

naliboff:
@tjhei Thanks for following up on this so quickly and glad to see the change is making a noticeable dent in the memory consumption!

@bangerth -  Many lithospheric problems could end up using 10+ or even 20+ compositional fields. This would reflect different lithologic layers, tracking tensor fields (e.g., elasticity) and other properties (finite strain, grain size). At this stage I think the large memory consumption will require not using all of the available cores on a node in order to preserve decent scaling. However, at this point is it likely more efficient to use tracers?

bangerth:
Hm, I wished we had discussed the design before you spent the time implementing this :-( I'm not happy with the added complexity of mapping things from one index space to another.

How about this approach:
* We only build the sparsity pattern of the diagonal block that corresponds to the first compositional field
* When we want to solve for compositional field `c`, we copy the sparsity pattern of the first block to block `(c,c)`, reinit that matrix block with it, assemble that matrix block, and solve the corresponding linear system
* When we are done with solving for compositional field `c`, we reset the `(c,c)` block of the matrix and sparsity pattern to something empty again. 

This way, you also get the memory savings, but you save yourself the complexity of remapping indices.  I would really like to avoid this because it would also make it so much more complicated to later on implement a Newton scheme on the global system, and it just looks error prone to me.

tjhei:
> Hm, I wished we had discussed the design before you spent the time implementing this :-(

This was just an experiment that I was able to do quite quickly. I agree that it is not elegant, but it seems to work...

> How about this approach:

I don't think there is a way to get the SparsityPattern out of a Trilinos::SparseMatrix, but I think we can store the SparsityPattern separately after constructing it. I can try it out.

bangerth:
I'd appreciate that. Save your current effort somewhere in case we can't anything else to work out.

You're right that you could just store the compositional field sparsity pattern somewhere else, and then let the sparse matrix block point to it, rather than to the block of the block sparsity pattern. That's not super elegant either, but should work nonetheless.

bangerth:
I'll add that I'm partially uneasy about your remapping approach because we now have multiple different ways of solving for the current values of a compositional field, and it's not entirely clear to me that they should all use the same sparsity pattern. It's even less clear to me that that's what we want going forward.

bangerth:
By the way, do you know why all of the unused off-diagonal blocks report one nonzero entry? Shouldn't it be zero?

tjhei:
> By the way, do you know why all of the unused off-diagonal blocks report one nonzero entry? Shouldn't it be zero?

I agree. We are reporting ``Epetra_CrsMatrix::NumGlobalNonzeros()``. No idea why this is not equal to 0.

bangerth:
I'm trying to reproduce the issue with the ones instead of zeros, but if I add the `matrix statistics` postprocessor to a random test, it actually shows zero for me. How did you generate the table that has the ones?

tjhei:
> I'm trying to reproduce the issue with the ones instead of zeros, but if I add the matrix statistics postprocessor to a random test, it actually shows zero for me. How did you generate the table that has the ones?

Here is the file in question: https://gist.github.com/tjhei/339d5181b0862c52947099854e68b56c
I noticed that the "1" is only there if the number of ranks >=8. I have no idea why that is.

tjhei:
There is no easy way to copy the SparsityPattern out (missing functionality in deal.II 8.5). While https://github.com/dealii/dealii/pull/5665 should work in the future, I decided to always put a matrix into composition 0 and ``reinit()`` the matrices in the other slots as needed. The patch also contains some logic to delete the diagonal blocks from sparsity pattern of the matrices that contained some entries (constrained ones).

source/simulator/assembly.cc
```diff
@@ -1010,8 +1010,18 @@ namespace aspect
       computing_timer.enter_section ("   Assemble composition system");
 
     const unsigned int block_idx = advection_field.block_index(introspection);
+
+    if (!advection_field.is_temperature() && advection_field.compositional_variable!=0)
+      {
+        // allocate the system matrix for the current compositional field by taking the matrix
+        // from composition 0 (this is the place we allocate the matrix at)
+        const unsigned int block0_idx = AdvectionField::composition(0).block_index(introspection);
+        system_matrix.block(block_idx, block_idx).reinit(system_matrix.block(block0_idx, block0_idx));
```
bangerth: This creates a new matrix, but uses the same sparsity pattern as the copied-from block. can you make this clear in the comment above?

source/simulator/assembly.cc
```diff
@@ -1010,8 +1010,18 @@ namespace aspect
...
+      }
+
     system_matrix.block(block_idx, block_idx) = 0;
-    system_rhs = 0;
+    system_rhs.block(block_idx) = 0;
```
bangerth: nice micro-optimization :-)

source/simulator/core.cc
```diff
@@ -1083,28 +1094,11 @@ namespace aspect
...
-                break;
 
-              default:
-                Assert(false,ExcNotImplemented());
-            }
-        }
+      // if we have at least one compositional field that is a FEM field, we create a matrix block in the first
+      // compositional block
+      if (have_fem_compositional_field)
+        coupling[x.compositional_fields[0]][x.compositional_fields[0]] = DoFTools::always;
```
bangerth: can you add to the comment that the sparsity patterns for the remaining compositional (diagonal) blocks will remain empty this way, but that the corresponding matrix blocks will then just re-direct to the one block created here? otherwise, the purpose of this piece of code will remain a bit obscure and require digging up code elsewhere.

source/simulator/core.cc
```diff
@@ -1128,29 +1122,8 @@ namespace aspect
...
-                      face_coupling[x.compositional_fields[c]][x.compositional_fields[c]]
-                        = DoFTools::none;
-                      break;
-                    default:
-                      Assert(false,ExcNotImplemented());
-                  }
-              }
-          }
+        if (parameters.use_discontinuous_composition_discretization && have_fem_compositional_field)
+          face_coupling[x.compositional_fields[0]][x.compositional_fields[0]] = DoFTools::always;
```
bangerth: add a comment here that says that we do the same we did for the non-DG case above

source/simulator/core.cc
```diff
@@ -1178,6 +1151,16 @@ namespace aspect
 #else
     sp.compress();
 
+    // Deleting other matrix blocks (no need to have entries for
+    // constrained/hanging nodes). Note that no other matrices than c=0 are
```
bangerth: I don't understand the parenthetical remark -- why do we not need these?

source/simulator/core.cc
```diff
@@ -1178,6 +1151,16 @@ namespace aspect
...
+    // fully created anyways.
```
bangerth: this is not true -- they will be fully formed, just not at the current time.

source/simulator/core.cc
```diff
@@ -1178,6 +1151,16 @@ namespace aspect
...
+    for (unsigned int c=1; c<introspection.n_compositional_fields; ++c)
+      {
+        const unsigned int block_idx = introspection.block_indices.compositional_fields[c];
+        sp.block(block_idx, block_idx).reinit(sp.block(block_idx, block_idx).locally_owned_range_indices(),sp.block(block_idx, block_idx).locally_owned_domain_indices()); // TODO: clearing it would be nice but clear() also resets the size
+        sp.block(block_idx, block_idx).compress();
```
bangerth: It took me a bit to understand the code. I think what you're doing is not to *delete* the block, but you're initializing it to a block of the correct size that simply has not entries within it? If so, maybe update the comment above.

Extra points if you wrap line 1160 :-)

source/simulator/core.cc
```diff
@@ -1178,6 +1151,16 @@ namespace aspect
 #else
     sp.compress();
 
+    // Deleting other matrix blocks (no need to have entries for
```
bangerth: we usually use infinitive form ("delete") for comments. same in line 1241 below

source/simulator/core.cc
```diff
@@ -1256,6 +1238,21 @@ namespace aspect
 #else
     sp.compress();
 
+    // Deleting temperature and composition matrix blocks:
```
bangerth: same comments about this here

tjhei:
updated. Let me know if the comments are still unclear.

tjhei:
I guess this was "okay-ed" a while ago, oops.

PR #2033, opened 2017-12-17
Use Unix line endings.

bangerth:
This just converts the files that are used to generate the publications list on our website to Unix line endings. No changes otherwise.


tjhei:
Those were generated by the tool running under linux. I would have to test if this still works with these changes. Or did you do that already?

bangerth:
I wished I had a way to test all of this, but there is some black magic that sits between these scripts and what later appears on http://aspect.geodynamics.org/publications.html . I have no idea how this file is actually generated. I know that these scripts play a role, but the driver script that generates `publications.html` is apparently not in the repository.

tjhei:
> I wished I had a way to test all of this, but there is some black magic that sits between these scripts and what later appears on

I kept all this separate from the main repo but it is available at https://github.com/tjhei/aspect-www-admin

tjhei:
I tried it out locally and it looks like it works correctly. So should we merge and see if it works?

bangerth:
Thanks. Follow-up coming soon.

PR #2034, opened 2017-12-21
update citing documents

tjhei:


doc/manual/citing_aspect.bib
```diff
@@ -158,7 +173,7 @@ @article{He2017
 }
 @phdthesis{Blom2016,
 author = {Blom, CAH},
-file = {:home/heister/.local/share/data/Mendeley Ltd./Mendeley Desktop/Downloaded/Blom - 2016 - State of the art numerical subduction modelling with ASPECT thermo-mechanically coupled viscoplastic compressible rheol(2).pdf:pdf},
+file = {::},
```
bangerth: I don't know what this line does, but shouldn't we just remove it altogether if there is no useful information in it?

tjhei: This file is auto-generated from Mendeley. I don't think editing it by hand after each export is a useful thing to do.

PR #2036, opened 2017-12-21
Fix: Do not allocate unnecessary preconditioner matrix blocks

tjhei:
The system_preconditioner_matrix is only used for the Stokes system, so
there is no need to allocate entries for temperature/composition. Fix
this.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -1190,10 +1190,12 @@ namespace aspect
   {
     Amg_preconditioner.reset ();
     Mp_preconditioner.reset ();
-    T_preconditioner.reset ();
-    C_preconditioner.reset ();
-
     system_preconditioner_matrix.clear ();
+
+    // The preconditioner matrix is only used for the Stokes block (velocity and Schur complement) and is of course not
+    // used if we use a direct solver.
+    if (parameters.use_direct_stokes_solver)
+      return;
 
     Table<2,DoFTools::Coupling> coupling (introspection.n_components,
                                           introspection.n_components);
@@ -1202,9 +1204,11 @@ namespace aspect
     const typename Introspection<dim>::ComponentIndices &x
       = introspection.component_indices;
 
+    // velocity-velocity block (only block diagonal):
     for (unsigned int d=0; d<dim; ++d)
       coupling[x.velocities[d]][x.velocities[d]] = DoFTools::always;
 
+    // Schur complement block (pressure - pressure):
     if (parameters.include_melt_transport)
       {
         coupling[introspection.variable("fluid pressure").first_component_index]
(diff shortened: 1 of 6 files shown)


source/simulator/core.cc
```diff
@@ -1250,57 +1230,11 @@ namespace aspect
...
-                      face_coupling[x.compositional_fields[c]][x.compositional_fields[c]] = DoFTools::none;
-                      break;
-
-                    default:
-                      Assert(false,ExcNotImplemented());
-                  }
-              }
-          }
-
-        DoFTools::make_flux_sparsity_pattern (dof_handler,
```
bangerth: How come we don't need to call this function any more? Aren't there cases where we need these entries? Or is this just in the system matrix, not the preconditioner? And what do we use for a preconditioner when we solve temperature/composition?

source/simulator/core.cc
```diff
@@ -1190,21 +1190,25 @@ namespace aspect
   {
     Amg_preconditioner.reset ();
     Mp_preconditioner.reset ();
-    T_preconditioner.reset ();
-    C_preconditioner.reset ();
```
bangerth: Wouldn't you want to release the memory here?

tjhei:
> How come we don't need to call this function any more?

The ``system_preconditioner_matrix`` is only used for the Stokes system (A and S block), so
there is no need to allocate entries for temperature/composition in it.

source/simulator/core.cc
tjhei: This seems completely unrelated. Why do I need to reset the temperature preconditioner if I reassemble the Stokes preconditioner matrix?

tjhei:
old:
```
Total system preconditioner matrix memory consumption: 737.59 MB.
Total system preconditioner matrix nnz: 59,553,838
  15,085,359           1           1           1           1           1           1           1           1           1           1
           1     250,836           1           1           1           1           1           1           1           1           1
           1           1   4,029,589           1           1           1           1           1           1           1           1
           1           1           1   5,023,493           1           1           1           1           1           1           1
           1           1           1           1   5,023,493           1           1           1           1           1           1
           1           1           1           1           1   5,023,493           1           1           1           1           1
           1           1           1           1           1           1   5,023,493           1           1           1           1
           1           1           1           1           1           1           1   5,023,493           1           1           1
           1           1           1           1           1           1           1           1   5,023,493           1           1
           1           1           1           1           1           1           1           1           1   5,023,493           1
           1           1           1           1           1           1           1           1           1           1   5,023,493
```
new:
```
Total system preconditioner matrix memory consumption: 234.05 MB.
Total system preconditioner matrix nnz: 15,553,512
  15,085,359           1           1           1           1           1           1           1           1           1           1
           1     250,836           1           1           1           1           1           1           1           1           1
           1           1      34,975           1           1           1           1           1           1           1           1
           1           1           1      22,779           1           1           1           1           1           1           1
           1           1           1           1      22,779           1           1           1           1           1           1
           1           1           1           1           1      22,779           1           1           1           1           1
           1           1           1           1           1           1      22,779           1           1           1           1
           1           1           1           1           1           1           1      22,779           1           1           1
           1           1           1           1           1           1           1           1      22,779           1           1
           1           1           1           1           1           1           1           1           1      22,779           1
           1           1           1           1           1           1           1           1           1           1      22,779
```
so overall, in this 3d example with 8 compositional fields we go from 1800mb to 1300mb in total matrix storage. I would call this a victory!

source/simulator/core.cc
bangerth: Fair enough.

bangerth:
Nice observation. Thanks!

PR #2037, opened 2017-12-21
Try something with the format for references.

bangerth:
The patch disentangles the formatting used in the file (without any actual change other than making it more readable what is happening here).

The only change is adding a `<br>` at the end of each entry. That's because https://aspect.geodynamics.org/publications.html just looks a bit unkempt between entries -- the text would be easier to read if there was more space between entries.

I don't know how this new formatting is going to work out, but thought I'd give it a shot. I may tweak it some further later on.


tjhei:
Looks good.

PR #2038, opened 2017-12-22
Simplify advection preconditioner lifetime

tjhei:
The advection (temperature and composition) preconditioner is always re-
initialized before using it, so keeping them around afterwards is not
necessary. This should also save some memory.


tjhei:
Currently based on #2036, so only look at the last commit.

bangerth:
Nice. Please rebase and merge yourself.

PR #2039, opened 2017-12-23
Update the description of the system preconditioner matrix.

bangerth:
This is an update to #2036.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -1219,6 +1219,15 @@ namespace aspect
     else
       coupling[x.pressure][x.pressure] = DoFTools::always;
 
+    // the system_preconditioner matrix is only used for the Stokes
+    // system as there we use a separate matrix for preconditioning
+    // than the system_matrix object). for temperature and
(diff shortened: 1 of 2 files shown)


source/simulator/core.cc
```diff
@@ -1219,6 +1219,15 @@ namespace aspect
     else
       coupling[x.pressure][x.pressure] = DoFTools::always;
 
+    // the system_preconditioner matrix is only used for the Stokes
```
tjhei: would it make more sense to have this comment in line 1195 instead?

tjhei:
Looks good otherwise.

source/simulator/core.cc
bangerth: I put it here because I would have expected the temperature/composition block description to come right here, after the Stokes description. The comment was meant to explain why there is not in fact anything right where the comment is.

But I obviously don't feel strongly. You decide.

tjhei: same here. ;-)

PR #2041, opened 2017-12-31
Fix formatting

bangerth:
In particular, avoid white space in the details of a publication. For examples, take a look at each of the ones here:
https://aspect.geodynamics.org/publications.html


gassmoeller:
There seemed to be a problem with the tester, or is this caused by the changes? I know too little about the jabref format to say anything about this, except that if you checked the new format looks ok, I am fine with it.

bangerth:
The tester only works on branches that fork off from `master`. These branches, for example, have the indentation script within the directory tree. But the `www` branch is not rooted in `master` but freestanding. As a consequence, there is no indentation script (and indeed none of the other files you would always see, such as the `source/` and `include/` directories). The continuous integration scripts are not prepared for this and fail.

It is safe to ignore this -- nothing can be tested on the `www` branch anyway.

gassmoeller:
Ah, I did not notice that this goes into the `www` branch.

PR #2042, opened 2018-01-02
Add grain size material model

gassmoeller:
This is the material model that was used in Dannberg, J., Eilon, Z., Faul, U., Gassmöller, R., Moulik, P., & Myhill, R. (2017). The importance of grain size to mantle dynamics and seismological observations. Geochemistry, Geophysics, Geosystems.

It is one of the most complicated material models we have yet, but it also has a pretty unique feature compared to most other geodynamic codes (the dynamically evolving grain size). It also makes the lookup data functionality of the Steinberger material model more widely available (and usable with other software than just Perplex). 

I went over the files and adjusted them for the current ASPECT version, but let me know if there is something that still needs to be updated.


include/aspect/material_model/steinberger.h
```diff
@@ -219,7 +249,7 @@ namespace aspect
          * List of pointers to objects that read and process data we get from
          * Perplex files.
          */
-        std::vector<std_cxx11::shared_ptr<internal::MaterialLookup> > material_lookup;
+        std::vector<std_cxx11::shared_ptr<Lookup::PerplexReader> > material_lookup;
```
tjhei: why is the type PerplexReader and not MaterialLookup?

include/aspect/material_model/damage_rheology.h
```diff
@@ -0,0 +1,495 @@
...
+  {
+    using namespace dealii;
+
+    /**
+     * Additional output fields for the dislocation viscosity parameters
+     * to be added to the MaterialModel::MaterialModelOutputs structure
+     * and filled in the MaterialModel::DamageRheology::evaluate() function.
+     */
+    template <int dim>
+    class DislocationViscosityOutputs : public NamedAdditionalMaterialOutputs<dim>
```
tjhei: This class (naming of variables) is very specific to the material model you implement. It is realistic that we will have other cases where we want to reduce the ``heating_work_fraction``? If yes, then this class could be made more generic, no?

gassmoeller:
Tester is happy.
About your questions:
1. You are right it is pretty specific right now and it could be made more generic (i.e. two different `AdditionalMaterialOutput` objects). On the other hand, I have never seen something comparable in a model (reducing the work fraction) and it might be the only application we will ever need it, which is why I did not put much effort into generalizing it. I removed the new shear heating plugin though and merged it with the existing one. This way we also cover the case with anisotropic viscosity (and one plugin less to compile).
2. The type is `PerplexReader`, because that is all the `Steinberger` material model could ever do. We added the new functionality to read Hefesto files for the damage material model. In theory the Steinberger model should work as well with the Hefesto files, but I never tested it, and decided to just stick to the old functionality.

tjhei:
okay, makes sense. I think this is good to be merged, but I'll wait if somebody else has a comment.

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+              break;
+            }
+        }
+      while (time < timestep);
+
+      // reduce grain size to recrystallized_grain_size when crossing phase transitions
+      // if the distance in radial direction a grain moved compared to the last time step
+      // is crossing a phase transition, reduce grain size
+
+      // TODO: recrystallize first, and then do grain size growth/reduction for grains that crossed the transition
```
jdannberg: Can you add that to the issue about this material model?

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+      double grain_size = original_grain_size;
+      double grain_size_change = 0.0;
+      const double timestep = this->get_timestep();
+
+      // use a sub timestep of 500 yrs, currently fixed timestep
+      double grain_growth_timestep = 500 * 3600 * 24 * 365.25;
+      double time = 0;
+
+      // find out in which phase we are
+      const unsigned int ol_index = get_phase_index(position, temperature, pressure);
```
jdannberg: Shouldn't that name be something like phase_index?

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+      if (this->introspection().name_for_compositional_index(field_index) == "olivine_grain_size"
+          &&
+          this->get_timestep_number() > 0)
+        {
+          // check if material has crossed any phase transition, if yes, reset grain size
+          if (crossed_transition != -1)
+            if (recrystallized_grain_size[crossed_transition] > 0.0)
+              phase_grain_size_reduction = grain_size - recrystallized_grain_size[crossed_transition];
+        }
+      else if (this->introspection().name_for_compositional_index(field_index) == "pyroxene_grain_size")
```
jdannberg: Does that even work with the rest of the code? Maybe just remove that part?

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+                         const double                  pressure,
+                         const std::vector<double>    &composition,
+                         const SymmetricTensor<2,dim> &strain_rate,
+                         const Point<dim>             &position) const
+    {
+      const SymmetricTensor<2,dim> shear_strain_rate = strain_rate - 1./dim * trace(strain_rate) * unit_symmetric_tensor<dim>();
+      const double second_strain_rate_invariant = std::sqrt(std::abs(second_invariant(shear_strain_rate)));
+
+      // TODO: make this more general, for more phases we have to average grain size somehow
+      // TODO: default when field is not given & warning
```
jdannberg: We should decide what we want to do if there is no grain size field. There are a number of places that ask for the name of the field explicitly, and if it's not there, I would just crash with a Assert.

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+      // limit the grain size to a global minimum
+      const std::string field_name = "olivine_grain_size";
+      const double grain_size = this->introspection().compositional_name_exists(field_name)
```
jdannberg: Assert that the field exists, and update name.

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+
+      // Currently this will never be called without adiabatic_conditions initialized, but just in case
+      const double adiabatic_pressure = this->get_adiabatic_conditions().is_initialized()
+                                        ?
+                                        this->get_adiabatic_conditions().pressure(position)
+                                        :
+                                        pressure;
+
+      // find out in which phase we are
+      const unsigned int ol_index = get_phase_index(position, temperature, adiabatic_pressure);
```
jdannberg: Same here, should be phase index?

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+
+      // TODO: we use the prefactors from Behn et al., 2009 as default values, but their laws use the strain rate
+      // and we use the second invariant --> check if the prefactors should be changed
```
jdannberg: Is this still relevant? I thought we were using Hirth and Kohlstedt values?

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+      double energy_term = exp((diffusion_activation_energy[ol_index] + diffusion_activation_volume[ol_index] * adiabatic_pressure)
+                               / (diffusion_creep_exponent[ol_index] * constants::gas_constant * temperature));
+      if (this->get_adiabatic_conditions().is_initialized())
```
jdannberg: Can you write a comment that this is to limit the temperature dependence of viscosity?

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+    dislocation_viscosity (const double      temperature,
+                           const double      pressure,
+                           const std::vector<double> &composition,
+                           const SymmetricTensor<2,dim> &strain_rate,
+                           const Point<dim> &position,
+                           const double viscosity_guess) const
+    {
+      const double diff_viscosity = diffusion_viscosity(temperature,pressure,composition,strain_rate,position) ;
+
+      // Start the iteration with the full strain rate
```
jdannberg: At some point, do we want to do the Newton iteration here that's implemented in the diffusion dislocation model? Could be more effective... But then it might not be that important, given that we don't have to compute the grain size change for this.

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+
+      return;
+    }
+
+
+
+    template <int dim>
+    double
+    DamageRheology<dim>::
+    grain_size_growth_rate (const double                  temperature,
```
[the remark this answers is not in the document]
jdannberg: For the future: Can we write sown somewhere that the material model should only compute this if the reaction terms are needed?

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+      double dis_viscosity;
+      if (viscosity_guess == 0)
+        dis_viscosity = dislocation_viscosity_fixed_strain_rate(temperature,pressure,std::vector<double>(),strain_rate,position);
+      else
+        dis_viscosity = viscosity_guess;
+
+      double dis_viscosity_old = 0;
+      unsigned int i = 0;
+      while ((std::abs((dis_viscosity-dis_viscosity_old) / dis_viscosity) > dislocation_viscosity_iteration_threshold)
+             && (i < dislocation_viscosity_iteration_number))
```
jdannberg: What if this does not converge? Shouldn't it crash? Or at least give a warning?

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+
+      // Currently this will never be called without adiabatic_conditions initialized, but just in case
+      const double adiabatic_pressure = this->get_adiabatic_conditions().is_initialized()
+                                        ?
+                                        this->get_adiabatic_conditions().pressure(position)
+                                        :
+                                        pressure;
+
+      // find out in which phase we are
+      const unsigned int ol_index = get_phase_index(position, temperature, adiabatic_pressure);
```
jdannberg: Same here.

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+
+      double energy_term = exp((dislocation_activation_energy[ol_index] + dislocation_activation_volume[ol_index] * adiabatic_pressure)
+                               / (dislocation_creep_exponent[ol_index] * constants::gas_constant * temperature));
+      if (this->get_adiabatic_conditions().is_initialized())
```
jdannberg: Same here about the comment.

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+    density (const double temperature,
+             const double pressure,
+             const std::vector<double> &compositional_fields, /*composition*/
+             const Point<dim> &) const
+    {
+      if (!use_table_properties)
+        {
+          const double composition_dependence = compositional_fields.size()>0
+                                                ?
+                                                compositional_delta_rho * compositional_fields[0]
```
jdannberg: Same here. What if the first field is the grain size?!

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+                             Patterns::Double (0),
+                             "The value of the thermal conductivity $k$. "
+                             "Units: $W/m/K$.");
+          prm.declare_entry ("Reference specific heat", "1250",
+                             Patterns::Double (0),
+                             "The value of the specific heat $cp$. "
+                             "Units: $J/kg/K$.");
+          prm.declare_entry ("Thermal expansion coefficient", "2e-5",
+                             Patterns::Double (0),
+                             "The value
```
jdannberg: Is that even used anywhere? I couldn't find it...

jdannberg: V_g

jdannberg: That comment does not make sense. Remove second part?

jdannberg: Same here.

jdannberg: wrong comment.

jdannberg: Same here.

jdannberg: And here.

jdannberg: And here.

jdannberg: Remove space.

jdannberg: space missing

tests/damage_rheology_friction_heating.prm
```diff
@@ -0,0 +1,135 @@
+# A test for the damage rheology material model similar to damage-rheology-
+# strain. The only difference is the activation of shear heating, which
+# is tested to be reduced by the work fraction that is used to change
+# the boundary area of grains.
```
jdannberg: Did you check that the output is correct?

tests/damage_rheology_growth.prm
```diff
@@ -0,0 +1,150 @@
...
+#
+# d(x) = (d_0 + A * x)^(1/p)
+#
+# with
+# d_0 = 8e-5 m         (initial grain size)
+# p   = 10             (grain size exponent)
+# A   = 2.0294e-46 1/m
+#
+# leading to a grain size of d=8.8957E-5 m at the
+# right boundary of the model.
```
jdannberg: Did you check that? The screen output says 8.072e-05, but I guess you would have to run it for a longer time?

tests/damage_rheology_growth_log.prm
```diff
@@ -0,0 +1,149 @@
...
+#
+# d(x) = (d_0 + A * x)^(1/p)
+#
+# with
+# d_0 = 8e-5 m         (initial grain size)
+# p   = 10             (grain size exponent)
+# A   = 2.0294e-46 1/m
+#
+# leading to a grain size of d=8.8957E-5 m at the
+# right boundary of the model.
```
jdannberg: Same here.

tests/damage_rheology_latent_heat.cc
```diff
@@ -0,0 +1,246 @@
...
+  namespace MaterialModel
+  {
+    ASPECT_REGISTER_MATERIAL_MODEL(DamageRheologyLatentHeat,
+                                   "damage rheology latent heat",
+                                   "A material model that behaves in the same way as "
+                                   "the damage rheology model, but is modified to "
+                                   "resemble the latent heat benchmark. Due to the "
+                                   "nature of the benchmark the model needs to be "
+                                   "incompressible despite a material table and "
+                                   "use an constant density for the calculation of "
```
jdannberg: a constant

tests/damage_rheology_strain.prm
```diff
@@ -0,0 +1,134 @@
+# A test for the damage rheology material model. The test produces nearly 
+# constant temperatures, pressures, strain rates and uses experimentally 
+# derived values for olivine. Therefore the grain size in the test approaches
+# the equilibrium grain size for t -> infinity
```
jdannberg: Can we write down what that value should be?

tests/latent_heat_enthalpy.prm
```diff
@@ -0,0 +1,117 @@
...
+# This is a model setup to test the latent heat generation. 
+# Material flows in from the top and crosses a phase 
+# transition, where energy is released and the temperature
+# increases. The energy release in this model is calculated
+# from a lookup table for enthalpy values and the enthalpy
+# is evaluated at the edges of each cell instead of the 
+# quadrature point to catch even very sharp transitions.
+# The steady-state result is around 10 % off the original 
+# test, because the data table was created as small as 
+# possible.
```
jdannberg: Hm, so we don't know what the correct solution is?

source/material_model/damage_rheology.cc
gassmoeller: I would prefer we leave this for an update. Who knows what that would do to the stability of the iteration.

source/material_model/damage_rheology.cc
gassmoeller: Yes, it is stored in the variable `thermal_alpha` and used if `use_table_properties` is `false`.

jdannberg: Somehow github messed this up. The comment is about line 1551.

jdannberg: prm.declare_entry ("Corresponding phase for transition", "",

jdannberg: Grain growth activation volume

jdannberg: This should be m.

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2028 @@
...
+          if (temperature_dependence > max_temperature_dependence_of_eta)
+            energy_term = adiabatic_energy_term * max_temperature_dependence_of_eta;
+          if (temperature_dependence < 1.0 / max_temperature_dependence_of_eta)
+            energy_term = adiabatic_energy_term / max_temperature_dependence_of_eta;
+        }
+
+      const double strain_rate_dependence = (1.0 - dislocation_creep_exponent[phase_index]) / dislocation_creep_exponent[phase_index];
+
+      return pow(dislocation_creep_prefactor[phase_index],-1.0/dislocation_creep_exponent[phase_index])
+             * std::pow(second_strain_rate_invariant,strain_rate_dependence)
```
bangerth: use `std::pow` in both places

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2028 @@
...
+      double effective_viscosity;
+      if (std::abs(second_strain_rate_invariant) > 1e-30)
+        {
+          const double disl_viscosity = dislocation_viscosity(temperature, pressure, composition, strain_rate, position);
+          effective_viscosity = disl_viscosity * diff_viscosity / (disl_viscosity + diff_viscosity);
+        }
+      else
+        effective_viscosity = diff_viscosity;
+
+      return effective_viscosity;
```
bangerth: you could just push the `return` into the `if` and `else` branches and get rid of the `effective_viscosity` variable altogether.

bangerth: same below if you choose to do this (though I can also see how keeping the variable makes it clear what it is you are computing here).

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2028 @@
...
+              const double      pressure,
+              const std::vector<double> &compositional_fields,
+              const Point<dim> &/*position*/) const
+    {
+      AssertThrow ((reference_compressibility != 0.0) || use_table_properties,
+                   ExcMessage("Currently only compressible models are supported."));
+
+      double enthalpy = 0.0;
+      if (n_material_data == 1)
+        enthalpy += material_lookup[0]->enthalpy(temperature,pressure);
```
bangerth: could as well be `enthalpy = ...`. same for the following functions.

(of course, only for the `if` branch -- in the `else` branch you have a loop)

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2028 @@
...
+                             Patterns::Double (0),
+                             "The value of the specific heat $cp$. "
+                             "Units: $J/kg/K$.");
+          prm.declare_entry ("Thermal expansion coefficient", "2e-5",
+                             Patterns::Double (0),
+                             "The value of the thermal expansion coefficient $\\alpha$. "
+                             "Units: $1/K$.");
+          prm.declare_entry ("Reference compressibility", "4e-12",
+                             Patterns::Double (0),
+
```
bangerth: same comments as for the class doc

source/postprocess/viscous_dissipation_statistics.cc
```diff
@@ -78,11 +80,19 @@ namespace aspect
             // calculate the local viscous dissipation integral
             for (unsigned int q = 0; q < n_q_points; ++q)
               {
+                // In some material models a part of the total work done is
+                // converted into other energy types than heat, e.g. surface energy.
+                // If we use such a material model, reduce the work fraction appropriately.
+                double heating_work_fraction = 1.0;
+                if (const MaterialModel::DislocationViscosityOutputs<dim> *disl_viscosities_out =
+                      out.template get_additional_output<MaterialModel::DislocationViscosityOutputs<dim> >())
+                  heating_work_fraction -= disl_viscosities_out->boundary_area_change_work_fractions[q];
```
bangerth: This modification is in one of the output postprocessors. Would we have to do similar modifications in any other parts of the code? It's a bit unfortunate that we would need to adjust *every single place* of the code base with knowledge about one particular additional heating model thingy...

tests/damage_rheology_friction_heating.prm
```diff
@@ -0,0 +1,135 @@
+# A test for the damage rheology material model similar to damage-rheology-
+# strain. The only difference is the activation of shear heating, which
```
bangerth: put the name of the other .prm on one line so that it's easier to `grep` for it

source/material_model/damage_rheology.cc
```diff
@@ -0,0 +1,2027 @@
...
+        {
+          // Use the adiabatic pressure instead of the real one, because of oscillations
+          const double pressure = (this->get_adiabatic_conditions().is_initialized())
+                                  ?
+                                  this->get_adiabatic_conditions().pressure(in.position[i])
+                                  :
+                                  in.pressure[i];
+
+          // convert the grain size from log to normal
+          std::vector<double> composition (in.composition[i]);
```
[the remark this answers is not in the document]
gassmoeller: Neither, this is a copy of all compositional fields (grain size or not), and only the field with the grain size index is modified.

include/aspect/material_model/damage_rheology.h
```diff
@@ -0,0 +1,527 @@
...
+     * named 'grain_size' is present. In the diffusion
+     * creep regime, the viscosity depends on this grain size. We use the grain
+     * size evolution laws described in Behn et al., 2009. Implications of grain
+     * size evolution on the seismic structure of the oceanic upper mantle, Earth
+     * Planet. Sci. Letters, 282, 178–189. Other material parameters are either
+     * prescribed similar to the 'simple' material model, or read from data files
+     * that were generated by the Perplex or Hefesto software. The material model
+     * is described in more detail in "Dannberg et al., 2017. The importance of
+     * grain size to mantle dynamics and seismological observations. Geochemistry,
+     * Geophysics, Geosystems.", which is the canonical reference for this
```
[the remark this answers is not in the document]
jdannberg: Dannberg, J., Z. Eilon, U. Faul, R. Gassmöller, P. Moulik, and R. Myhill (2017), The importance of grain size to mantle dynamics and seismological observations, Geochem. Geophys. Geosyst., 18, 3034–3061, doi:10.1002/2017GC006944.

include/aspect/material_model/damage_rheology.h
```diff
@@ -0,0 +1,495 @@
...
+                                                      const std::vector<double> &compositional_fields,
+                                                      const Point<dim> &position) const;
+        /**
+         * Rate of grain size growth (Ostwald ripening) or reduction
+         * (due to phase transformations) in dependence on temperature
+         * pressure, strain rate, mineral phase and creep regime.
+         * We use the grain size evolution laws described in Solomatov
+         * and Reese, 2008. Grain size variations in the Earth’s mantle
+         * and the evolution of primordial chemical heterogeneities,
+         * J. Geophys. Res., 113, B07408.
```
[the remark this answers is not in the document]
jdannberg: I think you can say that we use the grain size laws as described for example in Behn et al. 2009 (in agreement with all the other comments): Behn, M. D., Hirth, G., & Elsenbeck, J. R. (2009). Implications of grain size evolution on the seismic structure of the oceanic upper mantle. Earth and Planetary Science Letters, 282(1), 178-189.

We should also change the description: It is the rate of grain size reduction due to dynamic crystallization and phase transformations, and then add that for dynamic recrystallization there's the choice between the paleowattmeter (Austins and Evans, 2007) and the paleopiezometer (Hall and Parmentier, 2003).

include/aspect/material_model/damage_rheology.h
```diff
@@ -0,0 +1,527 @@
...
+        std::vector<double> grain_growth_rate_constant;
+        std::vector<double> grain_growth_exponent;
+        std::vector<double> reciprocal_required_strain;
+        std::vector<double> recrystallized_grain_size;
+
+        /**
+         * Parameters controlling the dynamic grain recrystallization
+         * (following paleowattmeter).
+         */
+        bool use_paleowattmeter;
```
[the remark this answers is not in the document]
jdannberg: We added the reference for that, but just to explain it, it refers to how the way grain size is reduced when the material deforms, and in particular the relation between the rate of grain size reduction and the change of internal energy (or the energy that has to be put in to decrease the grain size). According to the idea of the paleowattmeter, the rate of change of internal energy is proportional to the rate of change of grain boundary area (Austin and Evans, 2007). There's also a paleopiezometer (which is also implemented in the material model, and you can choose which one to use), where this relation is different.

tests/damage_rheology_growth_scaled.prm
```diff
@@ -0,0 +1,150 @@
...
+#
+# d(x) = (d_0 + A * x)^(1/p)
+#
+# with
+# d_0 = 8e-5 m         (initial grain size)
+# p   = 10             (grain size exponent)
+# A   = 2.0294e-46 1/m
+#
+# leading to a grain size of d=8.8957E-5 m at the
+# right boundary of the model.
```
jdannberg: I checked this test, and the output it produces is the same as the one of the damage_rheology_growth test, only scaled by a factor of ten (which is the internal scaling in the test). 

This means that this scaling option is a bit misleading: Internally, the code uses a different grain size to compute grain size evolution than what is given in the output. So I think the test is working as intended, we just have to think about this option again (`Lower mantle grain size scaling`), because I think what it should do is use a scaled grain size internally, but output the correct grain size. On the other hand, that might be difficult to do, as this is based on compositional fields, and we can not just output a different field value from the one the simulation uses.

tests/damage_rheology_latent_heat.prm
```diff
@@ -0,0 +1,138 @@
...
+set Use years in output instead of seconds = false
+set Adiabatic surface temperature          = 1000
+
+subsection Geometry model
+  set Model name = box
+
+  subsection Box
+    set X extent = 50000
+    set Y extent = 1000000
+    set Y repetitions = 20
```
jdannberg: This test has a different geometry (and refinement level) than the `latent_heat_enthalpy` test, which is the only thing we have to compare it with. They should be made consistent, so that we can then see if their results are identical (which they should be, I think).

tests/damage_rheology_growth.prm
jdannberg: Okay, so if you let this model run until and end time of 1.5e6 years, it reaches a steady state of 8.893e-05 (for max composition, which is the grain size in m, and 8.8942e-05 if I increase the global refinements by one). Is that good enough? 
We should probably also mention the end time needed to reach that steady state in the input file, so that future users can reproduce that result more easily.

tests/latent_heat_enthalpy/statistics
```diff
@@ -0,0 +1,15 @@
...
+# 6: Number of temperature degrees of freedom
+# 7: Iterations for temperature solver
+# 8: Iterations for Stokes solver
+# 9: Velocity iterations in Stokes preconditioner
+# 10: Schur complement iterations in Stokes preconditioner
+# 11: Minimal temperature (K)
+# 12: Average temperature (K)
+# 13: Maximal temperature (K)
+0 0.000000000000e+00 0.000000000000e+00 256 2467 1089  0 31 32 32 9.99000000e+02 9.99500000e+02 1.00000000e+03 
+1 1.000000000000e+15 1.000000000000e+15 256 2467 1089 17 28 29 29 9.98956206e+02 1.00253772e+03 1.01739649e+03
```
jdannberg: The original test output was 9.99032569e+02 1.00157844e+03 1.01243949e+03 (for the temperature at the end time, from your old branch). So we should check why this is so different, and what's going on.

tests/damage_rheology_growth_log.prm
jdannberg: This test's output is completely wrong. It should be around 9.424524182 for the test as it is, and for an end time of t=1.5e6 (steady state) it should be 9.327357451. 
However, it just stays at its initial value (9.4334). So that means that the functionality with the log scaling does not work anymore, and we should find out why.

tests/damage_rheology_strain/statistics
```diff
@@ -0,0 +1,24 @@
...
+# 15: Global mass for composition olivine_grain_size
+# 16: Minimal temperature (K)
+# 17: Average temperature (K)
+# 18: Maximal temperature (K)
+# 19: Average nondimensional temperature (K)
+# 20: RMS velocity (m/year)
+# 21: Max. velocity (m/year)
+0 0.000000000000e+00 0.000000000000e+00 256 2467 1089 1089 0  0 30 31 92 1.00000000e-03 1.00000010e-03 1.00000005e+07 1.60000000e+03 1.60000000e+03 1.60000000e+03 3.73686192e-15 5.77350322e-01 9.92956151e-01 
+1 3.125000000000e+03 3.125000000000e+03 256 2467 1089 1089 0 14  0  0  0 9.98808671e-04 9.98808791e-04 9.98808733e+06 1.60000000e+03 1.60000000e+03 1.60000000e+03 3.73686192e-15 5.77350322e-01 9.92956151e-01 
+2 5.000000000000e+03 1.875000000000e+03 256 2467 1089 1089 0  6  0  0  0 9.98106927e-04 9.98107057e-04 9.98106995e+06 1.60000000e+03 1.60000000e+03 1.60000000e+03 3.73686192e-15 5.77350322e-01 9.92956151e-01
```
jdannberg: The original results for the end time are: 9.98106935e-04 9.98107043e-04 9.98106992e+06 (for min, max, average grainsize). So those are very similar, I would say this test is working.

tests/damage_rheology_friction_heating/statistics
```diff
@@ -0,0 +1,29 @@
...
+1 3.125000000000e+03 3.125000000000e+03 256 2467 1089 1089 9 14  2  3  9 9.98808684e-04 9.98808927e-04 9.98808867e+06 1.60000000e+03 1.60000132e+03 1.60000152e+03 5.54208165e-07 5.77350417e-01 9.92956243e-01 1.63845356e+10 1.69503369e-11 5.76328384e+02 
+2 6.250000000000e+03 3.125000000000e+03 256 2467 1089 1089 6  8  1  2  6 9.97659878e-04 9.97660195e-04 9.97660129e+06 1.60000000e+03 1.60000261e+03 1.60000300e+03 1.10115573e-06 5.77350604e-01 9.92956364e-01 1.63295237e+10 1.68936134e-11 5.74399709e+02 
+3 9.375000000000e+03 3.125000000000e+03 256 2467 1089 1089 6  8  1  2  6 9.96544556e-04 9.96544921e-04 9.96544857e+06 1.60000000e+03 1.60000390e+03 1.60000445e+03 1.64236848e-06 5.77350845e-01 9.92956492e-01 1.62762052e+10 1.68386333e-11 5.72530311e+02 
+4 1.000000000000e+04 6.250000000000e+02 256 2467 1089 1089 3  6  0  0  0 9.96306624e-04 9.96307004e-04 9.96306933e+06 1.60000000e+03 1.60000415e+03 1.60000473e+03 1.74996623e-06 5.77350893e-01 9.92956518e-01 1.62648483e+10 1.68269221e-11 5.72132118e+02
```
jdannberg: The original results are 9.96306619e-04 9.96306991e-04 9.96306927e+06 (for the min, max, mass of grain size), again very similar to the new results, so this test is fine.

tests/damage_rheology_strain.prm
jdannberg: Just for the record, I run the test until t = 4e6 years, and got a maximum grain size of 9.35850647e-04 (which was the equilibrium value). Could you add those number to the .prm file? Otherwise I think this test is fine.

tests/damage_rheology_latent_heat/statistics
```diff
@@ -0,0 +1,23 @@
...
+# 14: Average temperature (K)
+# 15: Maximal temperature (K)
+# 16: Minimal value for composition olivine_grain_size
+# 17: Maximal value for composition olivine_grain_size
+# 18: Global mass for composition olivine_grain_size
+0 0.000000000000e+00 0.000000000000e+00 320 3303 1449 1449  0  0 25 26 26 9.99000000e+02 9.99500000e+02 1.00000000e+03 8.00000000e-05 8.00000000e-05 4.00000000e+06 
+1 2.916757647458e+14 2.916757647458e+14 320 3303 1449 1449 34 22 20 21 21 9.99017651e+02 1.00006385e+03 1.00484664e+03 8.00000000e-05 1.33641962e-03 9.12582750e+06 
+2 5.833396885589e+14 2.916639238131e+14 320 3303 1449 1449 25 17 18 19 19 9.99028974e+02 1.00063223e+03 1.00987049e+03 8.00000000e-05 1.82413197e-03 1.15262122e+07 
+3 8.750015227292e+14 2.916618341703e+14 320 3303 1449 1449 25 16 16 17 17 9.99038070e+02 1.00121282e+03 1.01454952e+03 8.00000000e-05 2.07596557e-03 1.30863246e+07 
+4 1.000000000000e+15 1.249984772708e+14 320 3303 1449 1449 15 10 14 15 15 9.99041607e+02 1.00146560e+03 1.01643772e+03 8.00000000e-05 2.15063808e-03 1.36339991e+07
```
jdannberg: The original values (from your old branch) are 9.99041621e+02 1.00146708e+03 1.01643824e+03 (for min/max/mass of temperature). So this is very similar to the original test results. 
This is surprising to me, as the latent_heat_enthalpy test was very different from the old branch. So maybe for the latent_heat_enthalpy the resolution is just too coarse.

tests/latent_heat_enthalpy/statistics
jdannberg: As the `damage_rheology_latent_heat` works, could you try changing the geometry and refinement level to the same numbers used in `damage_rheology_latent_heat`, and compare the results. Based on the description of the test, I think they should be the same.

jdannberg:
Okay, I looked through all of the tests. I post a summary below:
`damage_rheology_friction_heating`: works
`damage_rheology_growth`: works (are the results I posted accurate enough?)
`damage_rheology_growth_log`: fails (grain size does not change)
`damage_rheology_growth_scaled`: works
`damage_rheology_latent_heat`: works
`damage_rheology_strain`: works
`latent_heat_enthalpy`: different than original (change geometry to the one used in damage_rheology_latent_heat?)

include/aspect/material_model/damage_rheology.h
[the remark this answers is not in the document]
bangerth: I meant to put this into the documentation :-) But I guess @gassmoeller will do this.

include/aspect/material_model/damage_rheology.h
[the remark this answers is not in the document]
bangerth: Huh, what an oddly chosen name then. Thanks for the explanation!

include/aspect/material_model/damage_rheology.h
```diff
@@ -0,0 +1,495 @@
...
+         * vector that corresponds to the grain size to its
+         * logarithms and back and limits the grain size to
+         * a global minimum.
+         * @in normal_to_log: if true, convert from the grain
+         * size to its logarithm, otherwise from log to grain
+         * size
+         */
+        virtual
+        void
+        convert_log_grain_size (const bool normal_to_log,
```
jdannberg: This function is never called with the argument `normal_to_log = true`, as far as I can see. We just need it to find out what the real grain size is, based on the compositional field. So you can just remove the argument, and this case from the implementation.

tests/damage_rheology_growth_log.prm
jdannberg: Sorry, I looked at the max instead of the min grain size (it's switched in this case, because we convect -log grain size). So I believe this test produces the correct results too.

jdannberg:
We also went through all of the tests, and they all produce the output we expect.

gassmoeller:
Finally ready :smile:

jdannberg:
This is ready to go from my side.

gassmoeller:
Since @tjhei and @jdannberg approved, and I addressed all of Wolfgang's comments I am going to merge this (I have some PRs waiting that need this change).

bangerth:
Yes, great. Thanks!

PR #2043, opened 2018-01-04
Add BoundaryComposition::Manager class

gassmoeller:
Same change as #1771 but for boundary compositions instead of temperatures. One step further towards 2.0.


bangerth:
I haven't looked at every line (assuming it's mostly a copy paste of existing code), but what I've seen looks good. Thanks for also fixing typos in other places!

gassmoeller:
Correct, it is mostly copy-paste with some substitutions. Next step will be to allow different plugins on different boundaries (as @anne-glerum requested).

Tests pass, so I am going ahead and merge.

PR #2044, opened 2018-01-05
Create boundary velocity manager

gassmoeller:
Same as #2043, but for velocity plugins. This one is much more complicated, because we already allow different velocity plugins for different boundaries, and there is the additional complication of component selectors. I need to clean the code one more time before a proper review, but there are a few design decisions I would like to hear an opinion about:

1. Long term I would like to move all input parameters like "Prescribed velocity boundary indicators" from the generic "Model settings" subsection into the more appropriate "Boundary velocity model" subsection (same for temperature, composition, and so on). This is why I created a new "List of model names" parameter in the boundary velocity subsection. I feel this is easier to understand, also this would logically mean the Simulator should not longer know which boundaries have which plugins, this is the responsibility of the manager (and Simulator can ask for it). Is this appropriate?

2. Currently I enforce to have the same component selector for all plugins on a certain boundary. This seems reasonable for velocities, because one either wants to prescribe a component (with all selected plugins), or leave it unconstrained. For the composition however, I can easily imagine cases in which certain plugins are only responsible for certain compositions (=components). So the question is: Should I allow different component selectors for velocity also, to be consistent, or is this overdesigning things?

3. Currently I create multiple plugin objects if the names of plugins appear for different boundaries. This is consistent with the previous behavior, but feels a bit weird (it means a certain boundary object is only called for one boundary, and the boundary_id input parameter to `boundary_velocity` is more or less redundant). I could reuse objects by simply copying shared pointers into the lists of other boundaries upon construction, but I do not know if some user plugins rely on the fact that a plugin only serves a single boundary (for example the ascii data plugin does rely on this fact). Should I just keep things as before for now, and we only change it if it becomes a problem in the future?

### DIFF
--- a/include/aspect/boundary_velocity/interface.h
+++ b/include/aspect/boundary_velocity/interface.h
@@ -138,22 +138,6 @@ namespace aspect
         update ();
 
         /**
-         * Declare the parameters of all known boundary velocity plugins, as
-         * well as the ones this class has itself.
-         */
-        static
-        void
-        declare_parameters (ParameterHandler &prm);
-
-        /**
-         * Read the parameters this class declares from the parameter file.
-         * This determines which boundary velocity objects will be created;
-         * then let these objects read their parameters as well.
-         */
-        void
-        parse_parameters (ParameterHandler &prm);
-
-        /**
          * A function that calls the boundary_velocity functions of all the
          * individual boundary velocity objects and uses the stored operators
          * to combine them.
@@ -214,6 +198,22 @@ namespace aspect
          */
         const std::set<types::boundary_id> &
         get_tangential_boundary_velocity_indicators () const;
+
+        /**
+         * Declare the parameters of all known boundary velocity plugins, as
+         * well as the ones this class has itself.
+         */
+        static
+        void
+        declare_parameters (ParameterHandler &prm);
+
+        /**
+         * Read the parameters this class declares from the parameter file.
+         * This determines which boundary velocity objects will be created;
+         * then let these objects read their parameters as well.
+         */
+        void
+        parse_parameters (ParameterHandler &prm);
 
         /**
          * Go through the list of all boundary velocity models that have been selected in
--- a/include/aspect/boundary_velocity/interface.h
+++ b/include/aspect/boundary_velocity/interface.h
@@ -172,28 +172,41 @@ namespace aspect
 
 
         /**
-         * Return a list of names of all boundary velocity models currently
-         * used in the computation, as specified in the input file.
+         * Return the names of all prescribed boundary velocity models currently
+         * used in the computation as specified in the input file. The function
+         * returns a map between a boundary identifier and a pair. The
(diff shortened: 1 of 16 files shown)


bangerth:
1. Yes, this seems like a reasonable design.
2. No. That's overdesigned. Let each plugin deal with all selected components, not a subset.
3. Do we have plugins for the velocity that are *really large* in terms of memory consumption? I guess if that were the case, then that would be an issue. On the other hand, if we've never had a problem so far, then maybe it's not worth worrying about. 

Let me know once you want me to take a look at the actual code.

gassmoeller:
I will postpone 1 and do the change together with the temperature and composition parameters (when I have finished their PRs).
Regarding 3: We have some plugins that for example read in tables, but they need to read in different tables for different boundaries anyway, so there would be no reduction in memory (currently x plugins read 1 table, then 1 plugin would read x tables). So lets keep it in the current form for now.

This is ready for a review.

include/aspect/boundary_velocity/interface.h
```diff
@@ -119,98 +109,223 @@ namespace aspect
...
+                                    void (*declare_parameters_function) (ParameterHandler &),
+                                    Interface<dim> *(*factory_function) ());
+
+
+        /**
+         * Return a list of names of all boundary velocity models currently
+         * used in the computation, as specified in the input file.
+         */
+        const std::map<types::boundary_id, std::pair<std::string,std::vector<std::string> > > &
+        get_active_boundary_velocity_names () const;
```
bangerth: Can you describe what the return type of this function is? It looks complicated and could use some explanation :-)

include/aspect/boundary_velocity/interface.h
```diff
@@ -119,98 +109,223 @@ namespace aspect
...
+
+        /**
+         * Return a list of pointers to all boundary velocity models
+         * currently used in the computation, as specified in the input file.
+         */
+        const std::map<types::boundary_id,std::vector<std_cxx11::shared_ptr<BoundaryVelocity::Interface<dim> > > > &
+        get_active_boundary_velocity_conditions () const;
```
bangerth: same here

include/aspect/boundary_velocity/interface.h
```diff
@@ -119,98 +109,223 @@ namespace aspect
...
+
+        /**
+         * Return a list of boundary ids, on which the velocity is prescribed
```
bangerth: no need for the comma

include/aspect/boundary_velocity/interface.h
```diff
@@ -119,98 +109,223 @@ namespace aspect
...
+         * to be zero (no-slip).
+         */
+        const std::set<types::boundary_id> &
+        get_zero_boundary_velocity_indicators () const;
+
+        /**
+         * Return a list of boundary ids, on which the velocity is prescribed
```
bangerth: same here

PR #2047, opened 2018-01-07
Fix version if using deal.II release

gassmoeller:
A deal.II release has a commit ID, but no branch information in revision.h. In this case currently we output a strange string `(, commitID)`. 
Do not output anything about ID or branch in this case, because a release version is unique already (I could have fixed if with more complicated logic about which of the two is present, but this seems easier, and provides the same information.

One problem I just noticed while testing this patch is that we use a deal.II cmake macro to generate Aspect's version information. That means if we use a deal.II installation (not a build folder) we can not use that script (cmake scripts are not installed), and even if git is available we have no version information for aspect. Is it possible to find the source or build folder for an installation (if those still exist)? Or would it be ok to copy that macro (macro_deal_ii_query_git_information.cmake) into aspect/cmake?

### DIFF
--- a/source/global.cc
+++ b/source/global.cc
@@ -118,13 +118,13 @@ void print_aspect_header(Stream &stream)
   stream << "-----------------------------------------------------------------------------\n"
          << "-- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.\n"
          << "--     . version " << ASPECT_PACKAGE_VERSION;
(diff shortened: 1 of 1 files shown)


bangerth:
I think you probably want to just copy the macro. How many lines does it have?

gassmoeller:
About 100 lines. See #2048.

PR #2048, opened 2018-01-08
Copy git version macro from deal.II

gassmoeller:
See #2047. This script allows us to query the git information of ASPECT, even if we use a deal.II installation that does not contain this script. I checked that the script works as intended with a deal.II installation.


gassmoeller:
Also added the installation rule for the version information, so that installations know about their version.

tjhei:
excellent!

PR #2051, opened 2018-01-09
[WIP] Allow different boundary temperature plugins on different boundaries

gassmoeller:
This sits on top of #2044 and does the same thing for boundary temperature plugins. Now, one can select different boundary temperature plugins for different boundaries in the same style and input format as for velocities. I still want to make the parsing a bit more intuitive by allowing to only specify plugin names (that are then used for all boundaries) and not only pairs of boundaries and plugins, but this gets a bit complicated due to backwards compatibility to the current system, where a single entry is interpreted as a boundary id (not a plugin name). I need to do some more work on the parsing and documentation before this is finished.


bangerth:
Let me know once you have this rebased after merging #2044.

gassmoeller:
This should be ready for review. I recommend reviewing in 'split' mode, as much of interface.cc has changed. On the upside I was able to remove several deprecated things, so that the new structure looks actually simpler than before. The only complicated part is the `Manager::parse_parameters` function which figures out which plugins to use on which boundaries.

source/boundary_temperature/interface.cc
```diff
@@ -96,102 +85,48 @@ namespace aspect
...
+      return boundary_temperature_objects;
     }
 
 
 
     template <int dim>
     double
-    Manager<dim>::boundary_temperature (const types::boundary_id boundary_indicator,
-                                        const Point<dim> &position) const
+    Manager<dim>::minimal_temperature (const std::set<types::boundary_id> &/*fixed_boundary_ids*/) const
```
bangerth: oh, this negates my comment on the documentation of the function -- the function apparently ignores its argument. is this intended? If so, this is not a virtual function, so it's not possible to overload it, and if the only existing implementation ignores the argument, why not drop it altogether?

source/boundary_temperature/interface.cc
```diff
@@ -200,13 +135,15 @@ namespace aspect
 
     template <int dim>
     double
-    Manager<dim>::minimal_temperature (const std::set<types::boundary_id> &fixed_boundary_ids) const
+    Manager<dim>::maximal_temperature (const std::set<types::boundary_id> &/*fixed_boundary_ids*/) const
```
bangerth: same here, of course

source/utilities.cc
```diff
@@ -2468,6 +2468,26 @@ namespace aspect
...
+        op = Operator::subtract;
+      else if (operation == "minimum")
+        op = Operator::minimum;
+      else if (operation == "maximum")
+        op = Operator::maximum;
+      else
+        AssertThrow(false,
+                    ExcMessage ("ASPECT only accepts the following operators: "
+                                "add, subtract, minimum and maximum. But your parameter file "
+                                "contains: " + operation + ". Please check your parameter file.") );
```
bangerth: just write this as
```
  "contains <" + operation + ">. Please check..."
```
for consistency with other places.

In some places, we pass the name of the parameter where this happened to similar functions (e..g., to `possibly_extend_1_to_N()`) so that it's easier to find the location where this happened in the input file. Could we do the same here?

tests/edit_parameters.cc
```diff
@@ -39,7 +40,9 @@ namespace aspect
     if ( simulator_access.get_timestep_number() >= switch_step && !switched )
       {
         simulator_access.get_pcout()<<"Switching boundary condition!"<<std::endl;
-        parameters.fixed_temperature_boundary_indicators.erase(bottom);
+        std::set<types::boundary_id> &temperature_boundary_ids =
+          const_cast<std::set<types::boundary_id> &> (simulator_access.get_boundary_temperature_manager().get_fixed_temperature_boundary_indicators());
+        temperature_boundary_ids.erase(bottom);
```
bangerth: oh, if this was anywhere other than in a test, I'd be quite unhappy about such a trick :-)

gassmoeller:
I addressed most issues. About your comments:

First comment: I removed the part about the purpose, because it is not happening in this function. The `Manager::update` function only calls all the `update` functions of the plugins, and I did not want to duplicate the documentation too much. Would you prefer to keep the paragraph?

I updated the documentation and implementation of the `minimal_temperature` and `maximal_temperature` functions to do what they were meant to do. I do not like those functions though, and would vote for deprecating and at some point removing them (maybe replacing them by a `representative_surface_temperature` and `representative_bottom_temperature` function in the manager class). I will open a separate issue about that.

I added a parameter to the `create_model_operator_list` function to indicate the wrong input parameter.

Last comment: I am also not happy about this. Problem is that the test tested functionality that is now not longer easily available. The parameter to change is not longer a member of `Parameters`, but only `Parameters` is handed over as writable reference. I am not sure what to do here. It is a pretty exotic case though (removing a boundary condition during the model run).

source/utilities.cc
```diff
@@ -2507,25 +2525,25 @@ namespace aspect
...
-                                    "contains: " + operator_names[i] + ". Please check your parameter file.") );
+          try
+            {
+              operator_list[i] = Operator(operator_names[i]);
+            }
+          catch (Operator::ExcOperatorUnknown &exc)
+            {
+              AssertThrow(false,
+                          ExcMessage ("ASPECT only accepts the following operators: "
+                                      "add, subtract, minimum and maximum. But your parameter file "
```
jdannberg: Is there a way to get the operators we allow from the Operator object instead of spelling them out by hand? This would allow adding operators in the future without having to change the error message by hand (which we might forget).

source/boundary_temperature/interface.cc
```diff
@@ -272,17 +275,6 @@ namespace aspect
                           "will be used to append the listed temperature models onto "
                           "the previous models. If only one operator is given, "
                           "the same operator is applied to all models.");
```
jdannberg: I know this doesn't have anything to do with your pull request, but can you please add a sentence here (and in other places where we declare these operators), that this list of operators has to be exactly as long as the list of plugins (if it doesn't have just one entry), and that the order has to be the same as in the 'List of model names' because each operator will be applied to a plugin according to the order they're given in.

source/boundary_temperature/interface.cc
```diff
@@ -215,39 +162,93 @@ namespace aspect
...
     {
+      prm.enter_subsection ("Model settings");
+      {
+        prm.declare_entry ("Fixed temperature boundary indicators", "",
+                           Patterns::List (Patterns::Anything()),
+                           "A comma separated list of names denoting those boundaries "
+                           "on which the temperature is fixed and described by the "
+                           "boundary temperature objects selected in the "
+                           "<Boundary temperature model/List of model names> parameter "
+                           "of this input file. All boundary indicators used by the geometry "
```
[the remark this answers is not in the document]
jdannberg: Update: I see, you need it to apply the operators. So do we have plans for the future to allow several plugins for the velocity boundary conditions too? Because then I guess it would make sense to think about a format that would work for all three (temperature, composition, velocity).

source/boundary_temperature/interface.cc
```diff
@@ -215,39 +162,93 @@ namespace aspect
...
+                           "identifier as described above, or a pattern of the type "
+                           "'boundary identifier : plugin name', where plugin name is "
+                           "one of the names given in the "
+                           "<Boundary temperature model/List of model names> parameter. "
+                           "If only a boundary identifier is given, the boundary temperature "
+                           "of all active boundary temperature plugins will be combined to "
+                           "compute the boundary temperature. If only one plugin is given, "
+                           "only this plugin is used. The same boundary identifier can "
+                           "appear multiple times with different plugin names, in which "
+                           "case all given plugins will be combined to compute the boundary "
```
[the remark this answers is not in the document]
jdannberg: After thinking about this for a while, I guess the most reasonable approach I see would be to change this format to allow something like:

set Fixed temperature boundary indicators   = top : box & function & constant, left : box & initial temperature, bottom

If there is no colon, all plugins are applied (as it is right now, so it's backwards compatible). I guess that means duplicating the list of plugins in the 'List of model names' parameter, but I don't really see any other way to do this, and it would only come up in complex models. 

So I guess the only other question we have to think about here is: Do we ever want to allow having the same kind of plugin, but with different parameters, for different boundaries (like 2 different functions)? Because I don't really see how we would put that into this framework, so we would essentially say that this is someting we do not want to change in the near future.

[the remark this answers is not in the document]
bangerth: I haven't looked at this since my last review, but just about the format: instead of `&` we've generally used `;` as a top-level separator between entries I believe.

gassmoeller:
Same as #2099, lets postpone this after 2.0.

bangerth:
@gassmoeller: What's the status here? Can you rebase and should I then take a look?

PR #2052, opened 2018-01-12
Fix heat flux map for boxes

gassmoeller:
A small fix that avoids computing the heat flux map at side boundaries of geometries. This previously lead to averaging over the top/bottom surface (normal heatflux) and side boundaries (either none, or some random heatflux if prescribed temperatures are present), which looks strange and makes no sense unless one knows about the internals of the code. Since it is unlikely somebody wants to compute the heatflux map through a boundary, I disable that for now.
It is really just a two line change, but it looks bigger for the normal postprocessor, because of a different loop structure, and a changed indentation.


bangerth:
This looks reasonable. Can you add a changelog entry and fix the failing test? OK to merge after that.

PR #2053, opened 2018-01-14
Add limiting timestep increase function.

MFraters:
This pull requests adds the possiblity to set a percentage with which the the time step is limited to increase. Generally the time step based on the CFL number should be sufficient, but for complicated models which may suddenly drastically change behavoir, it may be usefull to limit the increase in the timestep, without limiting the time step size of the whole simulation to a particular number.

### DIFF
--- a/include/aspect/parameters.h
+++ b/include/aspect/parameters.h
@@ -312,6 +312,7 @@ namespace aspect
     double                         start_time;
     double                         CFL_number;
     double                         maximum_time_step;
+    double                         maximum_relative_increase_time_step;
     double                         reaction_time_step;
     unsigned int                   reaction_steps_per_advection_step;
     bool                           use_artificial_viscosity_smoothing;
--- a/source/simulator/parameters.cc
+++ b/source/simulator/parameters.cc
@@ -142,6 +142,15 @@ namespace aspect
                        "Units: Years or seconds, depending on the ``Use years "
                        "in output instead of seconds'' parameter.");
 
+    prm.declare_entry ("Maximum relative increase in time step", boost::lexical_cast<std::string>(std::numeric_limits<int>::max()),
+                       Patterns::Double (0),
+                       "Set a percentage with which the the time step is limited to increase. Generally the "
+                       "time step based on the CFL number should be sufficient, but for complicated models "
+                       "which may suddenly drastically change behavoir, it may be usefull to limit the increase "
+                       "in the timestep, without limiting the time step size of the whole simulation to a "
+                       "particular number if this value is $50$, then that means that the time step can at most "
+                       "increase by 50\\% from one time step to the next, or by a factor of 1.5.");
+
     prm.declare_entry ("Use conduction timestep", "false",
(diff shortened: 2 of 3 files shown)


source/simulator/parameters.cc
```diff
@@ -142,6 +142,14 @@ namespace aspect
                        "Units: Years or seconds, depending on the ``Use years "
                        "in output instead of seconds'' parameter.");
 
+    prm.declare_entry ("Maximum increase in time step", boost::lexical_cast<std::string>(std::numeric_limits<int>::max()),
+                       Patterns::Double (0),
+                       "Set a percentage with which the the time step is limited to increase. Generally the "
+                       "time step based on the CFL number should be sufficient, but for complicated models "
+                       "which may suddenly drastically change behavoir, it may be usefull to limit the increase "
+                       "in the timestep, without limiting the time step size of the whole simulation to a "
+                       "particular number.");
```
bangerth: Can you maybe add an example? Such as "For example, if this value is $50$, then that means that the time step can only increase by 50\\% at most from one time step to the next, or by a factor of 1.5."

include/aspect/parameters.h
```diff
@@ -312,6 +312,7 @@ namespace aspect
     double                         start_time;
     double                         CFL_number;
     double                         maximum_time_step;
+    double                         maximum_increase_time_step;
```
bangerth: I think a better name for the variable would be `relative_maximum_time_step_increase`.

MFraters: Yes, the relative should be there, but should it then be `maximum_relative_time_step_increase`?

bangerth: Sure, either is fine -- choose whatever you think is better.

MFraters:
I have addressed the comments :)

source/simulator/parameters.cc
```diff
@@ -142,6 +142,15 @@ namespace aspect
...
                        "in output instead of seconds'' parameter.");
 
+    prm.declare_entry ("Maximum relative increase in time step", boost::lexical_cast<std::string>(std::numeric_limits<int>::max()),
+                       Patterns::Double (0),
+                       "Set a percentage with which the the time step is limited to increase. Generally the "
+                       "time step based on the CFL number should be sufficient, but for complicated models "
+                       "which may suddenly drastically change behavoir, it may be usefull to limit the increase "
+                       "in the timestep, without limiting the time step size of the whole simulation to a "
+                       "particular number if this value is $50$, then that means that the time step can at most "
+                       "increase by 50\% from one time step to the next, or by a factor of 1.5. ");
```
bangerth: You'll need two backslashes here. The C++ preprocessor is going to convert the two into one in the string representation that is then given to LaTeX, and LaTeX needs one before the percent sign to escape it.

You also don't need the space at the end of the last line here, though that's a minor issue.

MFraters: you mean like what I did now?

bangerth:
Yes, exactly. Thanks!

PR #2054, opened 2018-01-15
Remove superfluous postprocessors

gassmoeller:
This removes a part of the work of #2042. These two plugins are not longer necessary, because their functionality is completely included in the more general 'heating statistics' and 'heating' plugins.
I also ran the update script to remove them in parameter files, and this resulted in a lot of other updates of deprecated input parameters.


bangerth:
1/ So does the 'heating statistics' just output statistics about *all* enabled heating terms?

2/ There is a tester failure. Can you check?

3/ The removed plugins have detailed descriptions of the formulas that describe the quantities that are output. Do we have corresponding explanations somewhere else?

gassmoeller:
1) 'heating statistics' writes the sum of all heating plugins to the screen, and information about the average and sum of *each single* plugin to the statistics file. 'viscous dissipation statistics' did the same for screen and statistics, but only for a hand-written formula (that it claimed to be the viscous dissipation, but it was written differently than in the shear heating plugin, and it produced different results in the test).

2) I updated the results, it were simply the additional lines of the old plugins that are now missing (all of them already used the new plugins).

3) Good point. I simply assumed it is documented in the shear heating plugin, but it was only in the class documentation in the header file, not in the parameter  description. I added that, but took the version of the documentation from the shear heating plugin, because it is simpler to read, and likely more accurate than the outdated description in the removed plugins.

If the tester is happy this is ready from my side.

cookbooks/global_melt.prm
```diff
@@ -117,7 +117,7 @@ subsection Discretization
   end
 end
 
-##################### Initial conditions ########################
+##################### Initial temperature model ########################
```
tjhei: I think this old heading is more appropriate (we handle temperature and composition)

bangerth:
OK from my side if you address the issue @tjhei had.

In an ideal world, you'd split all of the "Set model name" -> "Set List of model names" changes into a separate PR. I think this is unrelated to the @SUBJECT of this PR, right?

tjhei:
> In an ideal world

Thanks to Rene we have at least an ideal corner of the world. ;-)

bangerth:
This is so good to know! I was already despairing.

gassmoeller:
In an ideal world the tester would approve of my pull request :smile:

tjhei:
> In an ideal world the tester would approve of my pull request

hard disk was full. I hope the machine is back to running now.

gassmoeller:
Thanks!

PR #2057, opened 2018-01-16
Remove arbitrary limits in Steinberger material model

gassmoeller:
These limits were introduced at a time when I wanted to stabilize things, but they fooled us for a day thinking our model was wrong, and they are bad for at least two reasons:
1. They give wrong results instead of crashing when things go wrong.
2. Their numbers are arbitrary, at least they would need to be input parameters.

I think it is best to remove them, bad things should happen when these get unreasonable, and it is not possible to provide bounds for what "unreasonable" means.


bangerth:
OK if you add a changelog entry that explains the difference.

PR #2058, opened 2018-01-17
Update parameter files

gassmoeller:
This is the part of #2054 that is independent of the removal of the postprocessors. Since #2054 was already approved, I will just wait for the tester and then merge.


gassmoeller:
Timo could you check the tester? It seems to be asleep :smile:.

PR #2059, opened 2018-01-17
Allow for a variable number of points in the initial profile adiabati…

jdannberg:
…c conditions
This is useful if the profile has to be very accurate, for example because the thermal expansivity/density/specific heat have narrow spikes or jumps, such as caused by a phase transition.


source/adiabatic_conditions/initial_profile.cc
```diff
@@ -281,6 +278,12 @@ namespace aspect
                             "is computed. This profile is used to evaluate the "
                             "material model, when computing the pressure and "
                             "temperature profile.");
+          prm.declare_entry ("Number of points", "1000",
+                             Patterns::Integer (0),
```
tjhei: I think the min value should be at least 2 if not even something more sensible.

PR #2062, opened 2018-01-17
Add enthalpy derivatives averaging for Steinberger material model

gassmoeller:
The latent heat calculation in the Steinberger material so far was a very unstable algorithm, because it relied on derivatives of a quantity (enthalpy) that essentially jumps across phase transitions. This new algorithm (the same as in the grain size material model) makes things much more stable by calculating a finite difference approximation across the current cell (instead of the resolution of the material data table). I am currently working with @jdannberg and Carolina Lithgow-Bertelloni on creating a cookbook out of this, and will update the pull request and test later.


gassmoeller:
I added the functionality and the test and created an issue for myself to create the cookbook. This is ready for a review.

tests/steinberger_latent_heat.prm
```diff
@@ -0,0 +1,95 @@
+# This is a copy of the steinberger compressible test. However, it activates
+# latent heat by computing the thermal expansivity and specific heat as
+# derivatives of the enthalpy, which leads to temperature jumps across phase
+# transitions.
```
jdannberg: Have you verified that the temperature jumps across the phase transitions make sense? I would have taken the test we have for the grain size material model and copied that.

source/material_model/steinberger.cc
```diff
@@ -489,6 +529,53 @@ namespace aspect
               seismic_out->vs[i] = seismic_Vs(in.temperature[i], in.pressure[i], in.composition[i], in.position[i]);
             }
         }
+
+      /* We separate the calculation of specific heat and thermal expansivity,
+        * because they may depend on cell-wise averaged values that are only
+        * available here.
+        */
```
jdannberg: You only need any of the things you do below if latent_heat = true, correct? Why not put the `if (latent_heat)` statement around everything or return if latent_heat = false?

include/aspect/material_model/steinberger.h
```diff
@@ -252,6 +252,13 @@ namespace aspect
 
 
       private:
+        /**
+         * Returns the enthalpy derivatives for the evaluate function and
+         * postprocessors.
+         */
```
jdannberg: Can you add an explanation somewhere which one is the temperature and which one is the pressure derivative?

source/material_model/steinberger.cc
jdannberg: Oh, and could you add the equations how alpha and cp are computed based on H? Just in the comment above, like `c_p = dH/dT`or something like that...

source/material_model/steinberger.cc
```diff
@@ -461,6 +461,46 @@ namespace aspect
...
+      fe_values[this->introspection().extractors.temperature]
+      .get_function_values (this->get_current_linearization_point(), temperatures);
+      fe_values[this->introspection().extractors.pressure]
+      .get_function_values (this->get_current_linearization_point(), pressures);
+
+      // compute the averaged enthalpy derivatives for all temperatures and
+      // pressures in this cell
+      derivative = material_lookup[0]->enthalpy_derivatives(temperatures,
+                                                            pressures,
+                                                            1);
```
jdannberg: Can you document somewhere that the 1 means you do one substep?

jdannberg: And it would be great if you could add what the second output of that function is (the integers), because I don't think that's documented anywhere (so you might want to add that in the grain size model, and in this file just add a short note below where it is used).

source/material_model/steinberger.cc
```diff
@@ -489,6 +529,53 @@ namespace aspect
...
+      for (unsigned int i = 0; i < in.position.size(); ++i)
+        {
+          if (latent_heat)
+            {
+              // Use the adiabatic pressure instead of the real one, because of oscillations
+              const double pressure = this->get_adiabatic_conditions().pressure(in.position[i]);
+
+              if ((in.current_cell.state() == IteratorState::valid)
+                  && (dH[0].second > 0)
+                  && (dH[1].second > 0))
```
jdannberg: Can you add one line here that this means that at least one derivative across the cell was computed successfully?

source/material_model/steinberger.cc
```diff
@@ -489,6 +529,53 @@ namespace aspect
...
+                {
+                  out.thermal_expansion_coefficients[i] = (1 - average_density * dH[1].first) / average_temperature;
+                  out.specific_heat[i] = dH[0].first;
+                }
+              else
+                {
+                  if (material_lookup.size() == 1)
```
jdannberg: Is the case above implemented for `material_lookup.size() > 1`? If not, it might be easier to just make an Assert out of that if statement and put it somewhere above.

gassmoeller:
I addressed all comments, and verified that the `steinberger_latent_heat_2` test gives the same result as the `grain_size_latent_heat` test, if I do the same modifications in the material model (like fixing the density to the analytic density).

source/material_model/grain_size.cc
```diff
@@ -176,7 +176,7 @@ namespace aspect
                         ++n_T;
                       }
                   }
-                if (std::fabs(pressures[q] - pressures[p]) > get_pT_steps()[0] / 4.0)
+                if (std::fabs(pressures[q] - pressures[p]) > 1)
```
jdannberg: Do you want to scale this with the pressure too, in case people do exoplanets or some other weird things where the pressure is a different order of magnitude?

gassmoeller: No, that would not work, and was actually the bug I was looking for. Pressure might be zero, so the equation `something_small - the_same_small > 0.0` can be true due to floating point math. I could try to come up with something complicated like `... > pressures[q] * 1e-12 + epsilon`, would you prefer that?

jdannberg: I am not sure. I'll let you decide what makes more sense.

jdannberg:
Perfect! Address the one comment if you want to, and whatever the tester complains about, and then merge yourself.

gassmoeller:
I have checked that the tests pass with a rebased version of this branch after #2065. Since we have not much time left, I convinced @jdannberg to merge this without rebasing and waiting again for the tester. In case anything breaks, I will take the blame :smile:.

PR #2064, opened 2018-01-19
Add prem gravity profile

gassmoeller:
This is probably the most earth-like gravity profile currently available in ASPECT. At some point I will prepare a pull request that removes the current 'earth-like' gravity profile, and replace it with this one.
Ready for review.


data/gravity-model/prem.txt
```diff
@@ -0,0 +1,50 @@
+# A gravity profile that follows the description of
+# the preliminary reference Earth model (PREM) by
+# Dziewonski & Anderson (1981)
```
jdannberg: Can you add a doi here? 
All Earth scientists will know what you mean anyway, but @bangerth will complain about it ;-)

data/gravity-model/prem.txt
```diff
@@ -0,0 +1,50 @@
...
+# POINTS: 44
```
jdannberg: Is there any case where you would ever need a higher resolution compared to what you have here? As this includes the whole Earth, you have ~25 points for the mantle, so do you think that covers the jumps well enough? Or is the gravity of PREM stepwise linear anyway?

gassmoeller:
I added the doi. The data table in the paper is also given at discrete points like here, but there are about twice as many. But since gravity is pretty smooth, and the only drastic changes happen at boundary layers (which have support points in this table), I can not imagine somebody needing a higher resolved version of this table.

The one failing test is unrelated, and fixed in #2065.

jdannberg:
Okay, then this should be ready to merge once the tester is happy.

bangerth:
Yes, the gravity profile must be (and is) pretty smooth with the exception of kinks at radii where the density is discontinuous. That's because the gravity field is the gradient of the gravity potential, and the gravity potential satisfies a Laplace equation with the density on the right hand side. So you have something like (not correct, because it's really in spherical coordinates; but the argument still works):
```
  -phi''(r) = rho(r)
```
So a discontinuous but bounded rho yields a phi that is twice differentiable at all points where rho is continuous. Then 
```
  g(r) = -phi'(r)
```
is at least once differentiable at all points where rho is continuous; it may have kinks at places where rho is discontinuous, but g(r) cannot be discontinuous.

In other words, if you put interpolation points at the places where there is a kink, you're going to get a really good approximation because between these points, g(r) is really smooth.

PR #2068, opened 2018-01-23
add a phase transition to the inner core material model

jdannberg:
This is the first one of (hopefully) a series of pull requests that will extend the inner core cookbook, adding a number of new features to it. Once we have all of the new functionality, we will also modify the cookbook text and add an input file that demonstrates how to use these new features. 

This pull request adds the option to include a phase transition with variable depth/width/Clapyeron slope/etc. in the model, similarly to the latent heat material model, just much simpler.

### DIFF
--- a/cookbooks/inner_core_convection/inner_core_convection.cc
+++ b/cookbooks/inner_core_convection/inner_core_convection.cc
@@ -51,6 +51,25 @@ namespace aspect
          * the model time).
          */
         Functions::ParsedFunction<dim> resistance_to_phase_change;
+
+      private:
+        /**
+         * Parameters related to the phase transition.
+         */
+        double transition_depth;
+        double transition_width;
+        double transition_temperature;
+        double transition_clapeyron_slope;
+        double transition_density_change;
+
+        /**
+         * Percentage of material that has already undergone the phase
+         * transition to the higher-pressure material.
+         */
+        virtual
+        double
+        phase_function (const Point<dim> &position,
+                        const double temperature) const;
     };
 
   }
@@ -60,6 +79,32 @@ namespace aspect
 {
   namespace MaterialModel
   {
+    template <int dim>
+    double
+    InnerCore<dim>::
+    phase_function (const Point<dim> &position,
+                    const double temperature) const
+    {
+      // We need to convert the depth to pressure,
+      // keeping in mind that for this model, the pressure is only the dynamic pressure,
+      // so we have to compute the hydrostatic pressure explicitly as rho * g.
+      const double depth = this->get_geometry_model().depth(position);
+      const double reference_rho = 1.0; // formulation is nondimensionalized
+      const double pressure_gradient = this->get_gravity_model().gravity_vector(position).norm() * reference_rho;
+      double depth_deviation = depth - transition_depth;
+
+      if (std::abs(pressure_gradient) > 0)
+        depth_deviation -= transition_clapeyron_slope * (temperature - transition_temperature) / pressure_gradient;
+
+      double phase_func;
+      // use delta function for width = 0
+      if (transition_width == 0)
+        (depth_deviation > 0) ? phase_func = 1 : phase_func = 0;
+      else
+        phase_func = 0.5*(1.0 + std::tanh(depth_deviation / transition_width));
+      return phase_func;
+    }
+
 
     template <int dim>
     void
@@ -76,7 +121,8 @@ namespace aspect
       // density * cp to be 1
       for (unsigned int q=0; q < in.position.size(); ++q)
         {
-          out.densities[q] = - out.thermal_expansion_coefficients[q] * in.temperature[q];
+          out.densities[q] = - out.thermal_expansion_coefficients[q] * in.temperature[q]
+                             + phase_function (in.position[q], in.temperature[q]) * transition_density_change;
           if (std::abs(out.densities[q]) > 0.0)
             out.specific_heat[q] /= out.densities[q];
         }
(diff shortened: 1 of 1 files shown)


jdannberg:
I just noticed this is not going to work in the innermost region of the inner core model (where the gravity and hence the pressure gradient is zero), so I have to think about a fix for this.

cookbooks/inner_core_convection/inner_core_convection.cc
```diff
@@ -60,6 +79,32 @@ namespace aspect
...
+      const double pressure_gradient = this->get_gravity_model().gravity_vector(position).norm() * reference_rho;
+      double depth_deviation = depth - transition_depth;
+
+      if (std::abs(pressure_gradient) > 0)
+        depth_deviation -= transition_clapeyron_slope * (temperature - transition_temperature) / pressure_gradient;
+
+      double phase_func;
+      // use delta function for width = 0
+      if (transition_width == 0)
+        (depth_deviation > 0) ? phase_func = 1 : phase_func = 0;
```
tjhei: this is missing an assignment ``phase_func = ``

bangerth: Yes, I would agree -- legal, but hard to read :-)

cookbooks/inner_core_convection/inner_core_convection.cc
```diff
@@ -111,6 +157,39 @@ namespace aspect
             Functions::ParsedFunction<dim>::declare_parameters (prm, 1);
           }
           prm.leave_subsection();
+
+          prm.declare_entry ("Phase transition depth", "0.0",
```
bangerth: do the inner core people state locations in terms of depth from the surface, or in terms of distance from the center?

cookbooks/inner_core_convection/inner_core_convection.cc
jdannberg: Oh, right. I just copied that from another material model. I'll make a separate pull request to fix it there too.

cookbooks/inner_core_convection/inner_core_convection.cc
jdannberg: I asked, and it looks like the modellers state locations in terms of distance from the centre, and the seismologists in terms of distance from the inner core boundary. I changed everything to use the radius instead of depth, which is useful anyway, as eventually we want to let the core grow over time...

jdannberg:
Okay, I addressed your comments and fixed the part about the pressure gradient at the centre of the model. As this will only be used in a cookbook anyway, I would prefer to only add a test for the final new cookbook setup that includes all of the new code once we've finished that.

cookbooks/inner_core_convection/inner_core_convection.cc
```diff
@@ -60,6 +103,64 @@ namespace aspect
...
+    template <int dim>
+    double
+    InnerCore<dim>::
+    phase_function (const Point<dim> &position,
+                    const double temperature) const
+    {
+      // We need to convert the depth to pressure,
+      // keeping in mind that for this model, the pressure is only the dynamic pressure,
+      // so we have to get the hydrostatic pressure explicitly as an input.
+      const double radius = this->get_geometry_model().maximal_depth() - this->get_geometry_model().depth(position);
```
bangerth: Can you add an assertion that the radius is `>=0`?

cookbooks/inner_core_convection/inner_core_convection.cc
```diff
@@ -60,6 +103,64 @@ namespace aspect
...
+      const double pressure_deviation = hydrostatic_pressure(radius) - hydrostatic_pressure(transition_radius);
+
+      double depth_deviation = transition_radius - radius;
+      if (depth_deviation != 0.0)
```
bangerth: In floating point arithmetic, this condition is always true. You will want to compare against something that is small.

jdannberg:

cookbooks/inner_core_convection/inner_core_convection.cc
```diff
@@ -60,6 +103,68 @@ namespace aspect
...
+      // so we have to get the hydrostatic pressure explicitly as an input.
+      const double radius = this->get_geometry_model().maximal_depth() - this->get_geometry_model().depth(position);
+      Assert(radius >= 0,
+             ExcMessage("There is a point in the model where the depth is larger "
+                        "than the maximal depth of this geometry."));
+
+      const double pressure_deviation = hydrostatic_pressure(radius) - hydrostatic_pressure(transition_radius);
+
+      double depth_deviation = transition_radius - radius;
+      if (std::fabs(pressure_deviation) > 100.0 * std::numeric_limits<double>::epsilon())
```
bangerth: I think I should have explained better. If your pressure is, say, 100GPa, i.e., 1e11, then the smallest difference you can compute in line 127 is 1e-16*1e11=1e-5. This is *always* bigger than 100*eps.

To make this `if` statement meaningful, you need to make the tolerance *relative* to the size of quantities you are comparing. So maybe something like
```
if (std::fabs(pressure_deviation) 
   > 
   100.0 * std::numeric_limits<double>::epsilon() * 
   (hydrostatic_pressure(radius) + hydrostatic_pressure(transition_radius))/2)
```
i.e., 100*eps times the *average* of the two pressures.

jdannberg: I see your point, and I've changed the line accordingly. 
However, keep in mind that this model is nondimensionalized, and that the pressure is only the dynamic pressure. So the pressures are all very small, and might as well become negative, and the only thing I want to make sure is that we do not divide by zero in the next line.

bangerth: Ah, I hadn't realized that it's non-dimensional. Either way, even in non-dimensional models, quantities may be on the order of 1e-6 or 1e6, and you could run into the same problem.

But the new version looks good.

PR #2070, opened 2018-01-24
Change newton initial residual and linear tolerance computation

MFraters:
This change makes sure that the Eisenstat Walker method (https://doi.org/10.1137/0917003) for computing the linear tolerance more correctly follow the paper, which states for the gmres iteration: `|| F( x_k ) + F'( x_k ) s_k || < eta_k || F( x_k ) ||`, and then computes `eta_k` through `eta_k = (abs( || F( x_k )|| - || F( x_(k-1) ) + F'( x_(k-1) ) s_(k-1) || ) / F( x_(k-1) )`

So the two thing which are corrected in this patch is that the Newton rhs is used now ( `|| F( x_k ) ||` ) and that the residual is computed directly after the solve, so it can be used for the next iteration `|| F( x_(k-1) ) + F'( x_(k-1) ) s_(k-1) ||`

My initial tests show that the convergence of the nonlinear iteration is greatly improved when very lose/high `Maximum linear Stokes solver tolerance` are allowed.


source/simulator/solver.cc
```diff
@@ -853,6 +863,12 @@ namespace aspect
                                    solver_control_cheap,
                                    solver_control_expensive);
 
+        // We do not need the linearized_stokes_initial_guess vectory anymore, so
+        // (ab)use it for computing the initial residual when the Newton solver is used
```
gassmoeller: Can you extend this comment a bit. Specifically, I do not quite understand why we compute the initial residual before the solution for the fixed point iterations, but for the Newton iterations we compute it after the solution.

source/simulator/solver.cc
```diff
@@ -682,31 +682,41 @@ namespace aspect
...
+            // are only interested in the part of the rhs not balanced by the static
+            // pressure (the current pressure is a good approximation for the static
+            // pressure).
+            const double residual_u = system_matrix.block(0,1).residual (distributed_stokes_solution.block(0),
+                                                                         linearized_stokes_initial_guess.block(1),
+                                                                         system_rhs.block(0));
+            const double residual_p = system_rhs.block(1).l2_norm();
+
+            solver_tolerance = parameters.linear_stokes_solver_tolerance *
+                               sqrt(residual_u*residual_u+residual_p*residual_p);
```
bangerth: std::sqrt

source/simulator/solver.cc
MFraters: I put more comments for both the newton and non-Newton residual computation. I hope this helps a bit.

source/simulator/solver.cc
```diff
@@ -682,31 +682,45 @@ namespace aspect
...
-        const double residual_u = system_matrix.block(0,1).residual (distributed_stokes_solution.block(0),
-                                                                     linearized_stokes_initial_guess.block(1),
-                                                                     system_rhs.block(0));
-        const double residual_p = system_rhs.block(1).l2_norm();
-
-        const double solver_tolerance = parameters.linear_stokes_solver_tolerance *
-                                        sqrt(residual_u*residual_u+residual_p*residual_p);
-
+        double solver_tolerance = 0;
+        if (!assemble_newton_stokes_system)
```
bangerth: we usually like to spell this out: `assemble_newton_stokes_system == false`.

source/simulator/solver.cc
```diff
@@ -682,31 +682,45 @@ namespace aspect
...
+                                                                         system_rhs.block(0));
+            const double residual_p = system_rhs.block(1).l2_norm();
+
+            solver_tolerance = parameters.linear_stokes_solver_tolerance *
+                               std::sqrt(residual_u*residual_u+residual_p*residual_p);
+          }
+        else
+          {
+            const double residual_u = system_rhs.block(0).l2_norm();
+            const double residual_p = system_rhs.block(1).l2_norm();
```
bangerth: add a comment in front of these two lines of the form
```
// if we are solving for the Newton update, then the initial guess of the solution
// vector is the zero vector, and the starting (nonlinear) residual is simply
// the norm of the right hand side vector
```

source/simulator/solver.cc
```diff
@@ -682,31 +682,45 @@ namespace aspect
...
+            solver_tolerance = parameters.linear_stokes_solver_tolerance *
+                               sqrt(residual_u*residual_u+residual_p*residual_p);
```
bangerth: make this `std::sqrt`, to be consistent with the above.

source/simulator/solver.cc
```diff
@@ -853,6 +867,15 @@ namespace aspect
                                    solver_control_cheap,
                                    solver_control_expensive);
 
+        // We do not need the linearized_stokes_initial_guess vectory anymore, so
```
bangerth: vectory -> vector

source/simulator/solver.cc
```diff
@@ -853,6 +867,15 @@ namespace aspect
...
+        // (ab)use it for computing the initial residual when the Newton solver is used
+        // For the Newton solver we are interested in || F( x^(k-1) ) + F'( x^(k-1) )
+        // \partial x^(k-1) || to determine the correct linear tolerance, which is what
```
bangerth: \partial -> \delta

And in both lines, replace `x^(k-1)` by `x^{(k-1)}` for proper markup (3 times)

source/simulator/solver.cc
```diff
@@ -853,6 +867,15 @@ namespace aspect
...
+        // we compute here.
+        if (assemble_newton_stokes_system)
+          initial_residual = stokes_block.residual (linearized_stokes_initial_guess,
+                                                    distributed_stokes_solution,
+                                                    system_rhs);
```
bangerth: This is the only part of the patch that I'm not actually clear about. You are computing the linear residual left after solving the linear system. If you were solving the linear system exactly, shouldn't this be zero? Or what am I missing?

source/simulator/solver.cc
MFraters: should I also surround the equation with $?

source/simulator/solver.cc
MFraters: With the Eisenstat Walker method of determining the linear tolerance, the linear residual may actually be still relatively large. To compute the new linear tolerance we need to compute `(abs( || F( x_k )|| - || F( x_(k-1) ) + F'( x_(k-1) ) s_(k-1) || ) /|| F( x_(k-1) ) ||`, where `|| F( x_k )||` and `|| F( x_(k-1) ) ||` are the Newton rhs for this and the previous time step. So we only need the term: ` || F( x_(k-1) ) + F'( x_(k-1) ) s_(k-1) ||`, which we compute here.

source/simulator/solver.cc
bangerth: Oh, I didn't realize that we're in a .cc file. Forget about it in that case -- it's not going to be run through doxygen anyway, so markup doesn't matter. 

But do do the delta thing.

source/simulator/solver.cc
bangerth: And s_(k-1) is the solution of the previous linear system? So the quantity you compute here should then be the last residual of the previous GMRES solve, right? If that's the case, I think we can get that out of the SolverControl object, but I first wanted to confirm.

MFraters: hmm, yes, that would then be better. Is it the last_value() which I need for that?

bangerth: Yes. I think you want to double-check first that these really give the same result plus or minus a few epsilon. But that should be it.

MFraters: So, I tested it and the problem is that there are two solver controllers: the cheap and the expensive one. If you use the wrong one (I tried outputing them with pcout) then you get `Floating point exception (core dumped)`. When the correct one is used, the correct controller is returned.

bangerth: Yes, you'll have to get the value from the correct `SolverControl` object. In particular, if the cheap solver succeeds, then you'll have to get that value. If the cheap solver fails, you'll have to get the value from the expensive solver.

You'll also have to be prepared to compute the value the way you do in the (original) patch if we happened to use the direct solver -- though, in that case, we could also just set the value to zero because we solved the system exactly.

MFraters:
I think that I addressed all comments :)

bangerth:
Yes, nice and clean. Thanks!

Issue #2067, opened 2018-01-23
Allow more flexible surface movements


gassmoeller:
We have several users that want to create models with more flexible surface deformation algorithms than just 'free surface'. Examples would be surface evolution models (erosion/deposition), or inner core models with a growing domain over time. For these models one would prescribe velocities for the mesh at the surface, but have some other type of boundary (tangential/traction) for the solution velocity. I have been thinking about how to properly incorporate that into our current scheme, and think our current `FreeSurfaceHandler` mixes functionalities that can be logically separated:

1. Calculate the mesh velocity of the boundary nodes.
2. Adjust the overall mesh according to the boundary mesh velocities.
3. Correct advection velocity according to mesh velocity.

Steps 2 and 3 are the same for all of these approaches, but step 1 can use different velocities (e.g. the solution velocity in the free surface case, a modified solution velocity in the erosion case, or a prescribed velocity in the inner core case). I would therefore go ahead and extend the `FreeSurfaceHandler` to handle different boundary velocities. Optionally, we could think about renaming `FreeSurfaceHandler` to `MovingSurfaceHandler` or something similar.

I mostly open this issue to collect opinions, but I can also make a WIP pull request and ask for comments there.

ian-r-rose:
I think this makes good sense @gassmoeller. I could envision a new plugin structure where people can enable various "rules" for how to move the free surface. Some questions that would need to be answered:
1. Should only one rule be enabled at a time, or can people activate different rules (for instance, combining dynamic topography with erosion)?
2. If more than one rule can be active at the same time, how do we combine them? The simplest would be to do an operator-splitting approach, where we just apply the various rules one-by-one. The downside of this is that it probably would have unknown error and stability characteristics.
3. If there is an operator splitting approach, in what order would one apply the rules?
4. How would this interact with the free surface stabilization algorithm? The dynamic topography rule is simple enough that it can be made "quasi-implicit", allowing a boundary stress to be assembled on the LHS for stabilization purposes. Other rules are probably not as straightforward.

This is only tangentially (no pun intended) related, but I think that the step 3 (ALE velocity correction) could perhaps use some revisiting as well. When the motion of the free (or moving) surface is relatively large, this velocity correction can cause some difficulties for the advection solver. I think this is the cause of some instabilities in the advection solver that have been reported recently. There may some assumption in the preconditioner or extrapolation that is being violated by this ALE correction step.

bangerth:
[...]

PR #2071, opened 2018-01-25
Free surface extension to arbitrary deformation

gassmoeller:
This is a suggestion for an implementation of #2067 (some documentation still missing). I have verified that it works as intended, but I would like to ask for opinions about the interface design. Currently each new mesh deformation plugin can add constraints to the constraint matrix that is used to solve the Laplace problem of distributing the mesh vertices. This gives plugins great flexibility, but also feels a bit weird, because users need to know a lot about the internals to be able to write a new plugin. I have considered two alternative approaches, but they have their own problems:
1. an interface of the form `Tensor<1,dim> vertex_velocity (Point<dim> position)` is very intuitive, but for plugins that  internally solve some kind of other problem (e.g. a projection in `FreeSurface`) it is actually not simple to compute this kind of information. These plugins by design handle deformation-velocity as a vector for all points, not a point-wise quantity.
2. an interface of the form `void mesh_deformation(Vector &mesh_velocity)` would allow even more flexibility. In particular if a plugin wants to prescribe every vertex, it does not even need to solve the Laplace problem. In my above proposed interface one could maybe improvise this behavior by constraining every DoF, but that might lead to solver problems, or might be slow (constraint matrices are probably not designed for constraining all DoFs?).

Is there a better approach I do not see, or should we just stick with the current interface, and work around problems when (if) they appear?


source/mesh_deformation/function.cc
```diff
@@ -0,0 +1,124 @@
...
+      function(dim)
+    {}
+
+
+
+    template <int dim>
+    void
+    Function<dim>::update ()
+    {
+      function.set_time(this->get_time());
```
jdannberg: Just so we do not forget this later on: I think in the final version this should also take into account the parameter `Use years in output instead of seconds`.

MFraters:
Hey Rene,

Very cool, especially the core example! I am not an expert on the free surface code, but I have two comments on the code. 

The first comment is about your question. I think that a plugin should be as much self-contained as possible. If I understand the your explanation correctly, all plugins now contribute to a constrained matrix, which is then solved. I would be much more in favour of a structure where these kind of computations are handled internally by the plugins, and the interface would only give back a velocity/displacement/new position/new mesh. Like you say, this is also a bit more more intuitive. I do not fully understand what you mean with the interface `void mesh_deformation(Vector &mesh_velocity)`, so maybe we can talk about that.

The second comment is about the names. I see that you made a very nice namespace `MeshDeformation`, but then you call the handler `FreeSurfaceHandler` and in the parameters the section is also called `Free surface`. The problem is that you also named the free surface plugin freesurface, so it comes quite confusing to read the code. I would suggest that you change free surface in all places, which are not part of the free surface plugin to mesh deformation, which this is much more general. It would break the inputfile, but I think it would make the code (and input) file much more readable.

include/aspect/mesh_deformation/free_surface.h
```diff
@@ -0,0 +1,149 @@
...
+         * et. al. 2010 for more details.
+         */
+        double free_surface_theta;
+    };
+  }
+
+  namespace MeshDeformation
+  {
+    /**
+     * A plugin that computes the deformation of surface "
```
anne-glerum: rm "

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,398 @@
...
+         * in the parameter file.
+         */
+        std::vector<std::string> model_names;
+
+        /**
+         * Set the boundary conditions for the solution of the elliptic
+         * problem, which computes the displacements of the internal
+         * vertices so that the mesh does not become too distorted due to
+         * motion of the free surface.  Velocities of vertices on the free
+         * surface are set to be the normal of the Stokes velocity solution
```
anne-glerum: Maybe we should update the comments to include the "vertical" projection and the `Additional tangential mesh velocity boundary indicators`.

source/mesh_deformation/function.cc
```diff
@@ -0,0 +1,124 @@
...
+{
+  namespace MeshDeformation
+  {
+    ASPECT_REGISTER_MESH_DEFORMATION_MODEL(Function,
+                                           "function",
+                                           "A plugin, which prescribes the surface mesh to "
+                                           "deform according to an analytically prescribed "
+                                           "function. Note that the function prescribes a "
+                                           "deformation velocity, i.e. the return value of "
+                                           "the function is multiplied by the time step length "
```
anne-glerum: The way I understand it, the return value of the function is not multiplied by the time step length here, but the velocity solution computed with the constraints provided here is multiplied by the time step length to compute the mesh displacement. The way it is formulated here, it seems this plugin returns the displacement instead of the displacement velocity.

anne-glerum:
Hi Rene, 
very nice indeed! To me, the interaction of a new plugin of this type with some surface processing code would be very interesting. Such a code would provide a dim-1 grid of surface elevation/surface node location (which could be converted to a grid of surface velocities). I.e. there is no direct information for the motion of every vertex (approach 2.). Some interpolation will be necessary between the SPC grid and the vertices, which could perhaps be combined with approach 1., but your current implementation seems neatest to me.

I'm also a little confused by the MeshDeformation/FreeSurface naming :)

gassmoeller:
We discussed this some more, and plan to make an interface with a hierarchy of functions that can be overwritten by the user (inspired by deal.II manifold classes). One would be the current interface function, another one could be of the style:
`double surface_velocity (const Point<dim> &position)`

anne-glerum:
@gassmoeller The input file of the test needs to be updated (it still includes Model settings)

include/aspect/mesh_deformation/free_surface.h
```diff
@@ -0,0 +1,149 @@
...
+        virtual ~ApplyStabilization () {};
+
+        virtual
+        void
+        execute (internal::Assembly::Scratch::ScratchBase<dim>   &scratch,
+                 internal::Assembly::CopyData::CopyDataBase<dim> &data) const;
+
+      private:
+        /**
+         * Stabilization parameter for the free surface.  Should be between
```
anne-glerum: Remove space before Should

include/aspect/mesh_deformation/free_surface.h
```diff
@@ -0,0 +1,149 @@
...
+         * zero and one. A value of zero means no stabilization.  See Kaus
```
anne-glerum: Remove space before See

include/aspect/mesh_deformation/free_surface.h
```diff
@@ -0,0 +1,149 @@
...
+         * Project the Stokes velocity solution onto the
+         * free surface. Called by make_constraints()
+         */
+        void project_velocity_onto_boundary (const DoFHandler<dim> &free_surface_dof_handler,
+                                             const IndexSet &mesh_locally_owned,
+                                             const IndexSet &mesh_locally_relevant,
+                                             LinearAlgebra::Vector &output) const;
+
+        /**
+         * Stabilization parameter for the free surface.  Should be between
```
anne-glerum: Remove extra spaces

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,398 @@
...
+         */
+        void parse_parameters (ParameterHandler &prm);
+
+        /**
+         * A function that is used to register mesh deformation objects in such
+         * a way that the Manager can deal with all of them without having to
+         * know them by name. This allows the files in which individual
+         * plugins are implemented to register these plugins, rather than also
+         * having to modify the Manager class by adding the new initial
+         * temperature plugin class.
```
anne-glerum: temperature plugin -> mesh deformation plugin

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,398 @@
...
+         * A list of names of mesh deformation objects that have been requested
+         * in the parameter file.
+         */
+        std::vector<std::string> model_names;
+
+        /**
+         * Set the boundary conditions for the solution of the elliptic
+         * problem, which computes the displacements of the internal
+         * vertices so that the mesh does not become too distorted due to
+         * motion of the free surface.  Velocities of vertices on the
```
anne-glerum: Remove additional white space

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,398 @@
...
+         * deforming surface are fixed according to the selected deformation
+         * plugins.  Velocities of vertices on free-slip
```
anne-glerum: same here

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,398 @@
...
+         * boundaries are constrained to be tangential to those boundaries.
+         * Velocities of vertices on no-slip boundaries are set to be zero.
+         */
```
anne-glerum: Maybe add that when the no-slip boundaries are marked as additional tangential boundaries, their vertex velocities are constraint to be tangential.

tests/inner_core_growth.prm
```diff
@@ -0,0 +1,149 @@
...
+    set Reference temperature         = 0
+    set Thermal conductivity          = 1
+    set Thermal expansion coefficient = 1
+    set Viscosity                     = 1
+  end
+
+# The 'inner core material' model also contains a function that
+# represents the resistance to melting/freezing at the inner core
+# boundary. 
+# This phase boundary model combines a prescribed a normal
```
anne-glerum: 'a prescribed a normal velocity'

gassmoeller:
Superseded by #2964.

PR #2072, opened 2018-01-25
Add option to update adiabatic surface conditions

gassmoeller:
As described in the changelog. This is helpful for long-term models that nevertheless want to define properties in terms of variations from a changing reference profile (we want to set up such a model for the inner core).


source/adiabatic_conditions/initial_profile.cc
```diff
@@ -284,6 +306,18 @@ namespace aspect
                              "profile. The higher the number of points, the more accurate "
                              "the downward integration from the adiabatic surface "
                              "temperature will be.");
+          prm.declare_entry ("Use surface condition function", "false",
+                             Patterns::Bool(),
+                             "Whether to use the surface_conditions_function to determine surface "
```
tjhei: does it make sense to call it ``"Surface condition function"`` here instead?

source/adiabatic_conditions/initial_profile.cc
```diff
@@ -284,6 +306,18 @@ namespace aspect
...
+                             "conditions, or the adiabatic_surface_temperature and surface_pressure "
+                             "parameters. If this is set to true the reference profile is updated "
+                             "every timestep.");
```
tjhei: can you document here or somehow in the function below what the arguments and outputs of that function are?

gassmoeller:
I addressed the comments and renamed the plugin.

bangerth:
Looks like you need to update a test output

gassmoeller:
Done :smile:

gassmoeller:
@tjhei Did I address your comments?

tjhei:
How happy are you with the name? I would have picked something like ``ComputedProfile``, otherwise it sounds like an action.

gassmoeller:
I see your point, but in this case `ComputedProfile` could lead to misunderstandings, because it suggests the user should compute this profile before the computation, `ComputeProfile` makes clear ASPECT does it internally. Is this reasonable?

source/adiabatic_conditions/compute_profile.cc
```diff
@@ -79,8 +93,16 @@ namespace aspect
...
-              pressures[i] = this->get_surface_pressure();
-              temperatures[i] = this->get_adiabatic_surface_temperature();
+              if (!use_surface_condition_function)
+                {
+                  pressures[0] = this->get_surface_pressure();
+                  temperatures[0] = this->get_adiabatic_surface_temperature();
+                }
+              else
+                {
+                  pressures[0] = surface_condition_function.value(Point<1>(),0);
```
tjhei: why do you use the default constructor instead of ``Point<1>(0.0)``? This is supposed to be depth=0 or not?

source/adiabatic_conditions/compute_profile.cc
```diff
@@ -284,6 +306,21 @@ namespace aspect
                              "profile. The higher the number of points, the more accurate "
                              "the downward integration from the adiabatic surface "
                              "temperature will be.");
+          prm.declare_entry ("Use surface condition function", "false",
+                             Patterns::Bool(),
+                             "Whether to use the 'Surface condition function' to determine surface "
+                             "conditions, or the 'Adiabatic surface temperature' and 'Surface pressure' "
+                             "parameters. If this is set to true the reference profile is updated "
+                             "every timestep. The function expression of the function should be "
+                             "independent of space, but can depend on time 't'. The function must "
```
tjhei: in the header you comment that it is ``depth->p,t``, which one is correct?

source/adiabatic_conditions/compute_profile.cc
```diff
@@ -342,15 +399,18 @@ namespace aspect
...
+                                               "pressure and temperature in depth. "
                                                "The gravity is assumed to be in depth direction "
                                                "and the composition is either given by the initial "
                                                "composition at reference points or computed "
                                                "as a reference depth-function. "
                                                "All material parameters are computed by the "
-                                               "material model plugin.")
+                                               "material model plugin. The surface conditions are "
+                                               "either constant or changing over time as prescribed "
+                                               "by an user-provided function.")
```
tjhei: "by a user-provided"?

source/adiabatic_conditions/compute_profile.cc
gassmoeller: Aren't the two equivalent? Anyway, I changed it to make it more explicit.

source/adiabatic_conditions/compute_profile.cc
gassmoeller: This one is correct, I changed the header.

source/adiabatic_conditions/compute_profile.cc
gassmoeller: I never consciously thought about these rules, but it makes sense now. Thanks :smile:.

gassmoeller:
I addressed your comments.

source/adiabatic_conditions/compute_profile.cc
bangerth: Yes, it's not about the letter but how it's pronounced :-)

tjhei:
:+1:

PR #2073, opened 2018-01-26
enable free surface for all solver schemes

jdannberg:
A while ago, @gassmoeller restructured the solvers schemes (see #1850) so that the free surface computation is done outside of these schemes. This means that that there is no reason any more why any of the solver schemes should not work with free surface, the two things are now independent of each other. This pull request removes the Assert that was still there to prevent those things being used together. 

I also added a test for the iterated IMPES solver and made sure that the result is similar to the test that used iterated Stokes (the first 2-3 digits of RMS velocity and topography are the same).


gassmoeller:
This makes sense for our current implementation of the free surface (which is outside of the nonlinear iteration). One thing that we need to keep in mind is that the nonlinear solver will not iterate over the free surface, but will only find the correct velocities for the given topography, and topography will only be updated for the next timestep. This is something we can improve at some (undetermined) point in the future. Ready to merge if the tester passes.

PR #2074, opened 2018-01-26
Fix melt in parallel computations

gassmoeller:
Apparently, we had no test for parallel melt models, because parallel melt transport was likely broken since #1850. Should be fixed now though.

### DIFF
--- a/source/simulator/solver_schemes.cc
+++ b/source/simulator/solver_schemes.cc
@@ -266,7 +266,7 @@ namespace aspect
         // and will therefore be updated as well.
         const unsigned int fluid_velocity_block = introspection.variable("fluid velocity").block_index;
         const unsigned int fluid_pressure_block = introspection.variable("fluid pressure").block_index;
-        current_linearization_point.block(fluid_velocity_block) = solution(fluid_velocity_block);
+        current_linearization_point.block(fluid_velocity_block) = solution.block(fluid_velocity_block);
         current_linearization_point.block(fluid_pressure_block) = solution.block(fluid_pressure_block);
       }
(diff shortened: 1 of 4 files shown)


source/simulator/solver_schemes.cc
```diff
@@ -266,7 +266,7 @@ namespace aspect
         // and will therefore be updated as well.
         const unsigned int fluid_velocity_block = introspection.variable("fluid velocity").block_index;
         const unsigned int fluid_pressure_block = introspection.variable("fluid pressure").block_index;
-        current_linearization_point.block(fluid_velocity_block) = solution(fluid_velocity_block);
+        current_linearization_point.block(fluid_velocity_block) = solution.block(fluid_velocity_block);
```
bangerth: Did we just accidentally copy out a single element here and assign it to a whole vector? :-)

bangerth:
Please apply the changes from http://www.math.clemson.edu/~heister/cib/aspect-8.5-gcc-hd/5d4ee1ceaf1950972a5deb6d1dcb6d81eae96c68/changes-gcc.diff and then merge yourself!

source/simulator/solver_schemes.cc
gassmoeller: Yes, and I just understood that the result is luckily currently not used (the fluid velocity is not one of the primary solution variables, it is more a postprocessed field that is computed from the two pressures and the solid velocity). Thus, all models that ran since #1850 actually computed the correct results (everything beforehand was fine). Only parallel models after #1850 would crash, because they accessed this one element.

PR #2076, opened 2018-01-30
Number of steps between outputs

MarineLasbleis:
Define a new input parameter "Steps between graphical output" (corresponding to the maximum allowed timesteps between graphical output). 
The output files are written when the first criteria is reached: either the time since last output is smaller than the one input, or the number of steps since the last output is smaller than the "Steps between graphical output". 
(Default value is infinity)


include/aspect/postprocess/visualization.h
```diff
@@ -383,6 +383,22 @@ namespace aspect
...
+         */
+        double maximum_timesteps_between_outputs; 
+
+        /**
+         * Timestep at which the last graphical output was produced
+         * Used to check for the next necessary output time.
+         */
+        double last_output_timestep;
+
+
```
gassmoeller: Please only add a single free line between parameters.

include/aspect/postprocess/visualization.h
```diff
@@ -383,6 +383,22 @@ namespace aspect
          * to be produced. Used to check for the next necessary output time.
          */
         double last_output_time;
+	
+        /**
+         * Maximum number of steps between the generation of graphical output. 
+	 * This parameter
```
gassmoeller: please align the comment lines. You can automate this process by using the `astyle` software (see https://github.com/geodynamics/aspect/blob/master/CONTRIBUTING.md#making-aspect-better). If you have the software installed, you can reformat all your changes by running `make indent` in the folder you compiled ASPECT in.

source/postprocess/visualization.cc
```diff
@@ -623,6 +629,7 @@ namespace aspect
 
       // up the next time we need output
       set_last_output_time (this->get_time());
+      last_output_timestep= this->get_timestep_number();
```
gassmoeller: We usually have spaces before and after assignments `a = b`

source/postprocess/visualization.cc
```diff
@@ -729,6 +736,11 @@ namespace aspect
                              "'Use years in output instead of seconds' parameter is set; "
                              "seconds otherwise.");
 
+	  prm.declare_entry ("Steps between graphical output", boost::lexical_cast<std::string>(std::numeric_limits<int>::max()),
+                             Patterns::Integer(0),
+                             "The number of steps between each generation of "
```
gassmoeller: "The **maximum** number of **time** steps between ..."

tests/max_timesteps_btwn_output.prm
```diff
@@ -0,0 +1,177 @@
+# This is a test to check if output files are correctlywritten if the maximum number of steps
```
gassmoeller: space between correctly and written

source/postprocess/visualization.cc
```diff
@@ -326,11 +330,13 @@ namespace aspect
...
         {
           last_output_time = this->get_time() - output_interval;
+          last_output_timestep = this->get_timestep_number();
         }
 
       // return if graphical output is not requested at this time
       if ((this->get_time() < last_output_time + output_interval)
-          && (this->get_timestep_number() != 0))
+          && (this->get_timestep_number() != 0)
+	  && (this->get_timestep_number() < last_output_timestep + maximum_timesteps_between_outputs))
```
gassmoeller: Forgive me for being nitpicky, but could you change the order of the conditions? This would be a bit simpler to read if it would be:
```
if ((time < output_time)
&& (step < output_steps)
&& (timestep != 0))
```
Also make sure the `&&` align. With so complex code we learned how important it is to keep things as easy to read as possible.

gassmoeller:
Thanks, looks good from my side. Can somebody else have a look, and potentially merge?

source/postprocess/visualization.cc
```diff
@@ -326,10 +330,12 @@ namespace aspect
       if (std::isnan(last_output_time))
         {
           last_output_time = this->get_time() - output_interval;
+          last_output_timestep = this->get_timestep_number();
         }
 
       // return if graphical output is not requested at this time
       if ((this->get_time() < last_output_time + output_interval)
+          && (this->get_timestep_number() < last_output_timestep + maximum_timesteps_between_outputs)
```
bangerth: this doesn't work -- `maximum_timesteps_between_outputs` may be the largest integer around if the user did not specifically specify it, so the addition on the right is always going to wrap around. I suspect this condition to never be true. 

You'll have to write the condition in a way that doesn't overflow/wraps around, or limit the values the parameter can have to something so that "in all cases we can think of", there is no wrap around. For example, if we make sure the parameter is never larger than one billion, and we can reasonably assume that nobody ever runs a simulation with a billion time steps, then that would work because for integers, wrap around only happens at two billions.

include/aspect/postprocess/visualization.h
```diff
@@ -384,6 +384,20 @@ namespace aspect
          */
         double last_output_time;
 
+        /**
+         * Maximum number of steps between the generation of graphical output.
+         * This parameter
+         * is read from the input file and consequently is not part of the
+         * state that needs to be saved and restored.
+         */
+        double maximum_timesteps_between_outputs;
```
gassmoeller: This should be an unsigned int.

include/aspect/postprocess/visualization.h
```diff
@@ -384,6 +384,20 @@ namespace aspect
...
+
+        /**
+         * Timestep at which the last graphical output was produced
+         * Used to check for the next necessary output time.
+         */
+        double last_output_timestep;
```
gassmoeller: This should also be an unsigned int.

source/postprocess/visualization.cc
gassmoeller: Funnily enough, it will work as the tests show (although by accident, I did not think of your point when reviewing). The reason it works is that the default value is set to `std::numeric_limits<int>::max()`, but the value is stored in an `unsigned int`, so the one sign-bit (allowing 4 billion as max value, instead of the 2 billion of int) is saving us from an integer overflow, as long as nobody is running a model for more than 2 billion timesteps (a safe assumption). Are you fine with the rest of the PR?

bangerth: No, actually, the value is stored as a `double` as you remarked above. If it was stored as an `unsigned int`, that would work along the same lines as I mentioned above. In any case, though, I think this line deserves a comment that explains the issue with overflow, and in line 740, the pattern should be written as `Patterns::Integer(0,std::numeric_limits<int>::max())` to make clear that the value is not allowed to be larger than half the unsigned range.

gassmoeller:
I think we addressed the remaining issues.

source/postprocess/visualization.cc
```diff
@@ -729,6 +744,11 @@ namespace aspect
                              "'Use years in output instead of seconds' parameter is set; "
                              "seconds otherwise.");
 
+          prm.declare_entry ("Steps between graphical output", boost::lexical_cast<std::string>(std::numeric_limits<int>::max()),
```
bangerth: I didn't read this carefully -- please call it "Time steps between graphical output" to make clear what kind of steps it is.

source/postprocess/visualization.cc
```diff
@@ -875,6 +895,8 @@ namespace aspect
           if (this->convert_output_to_years())
             output_interval *= year_in_seconds;
 
+          maximum_timesteps_between_outputs = prm.get_integer("Steps between graphical output");
```
bangerth: need to adjust here

source/postprocess/visualization.cc
```diff
@@ -209,6 +211,8 @@ namespace aspect
       // initialize this to a nonsensical value; set it to the actual time
       // the first time around we get to check it
       last_output_time (std::numeric_limits<double>::quiet_NaN()),
+      maximum_timesteps_between_outputs (numbers::invalid_unsigned_int),
```
bangerth: I think you want this to be `std::numeric_limits<int>::max()` to match the parameter's default.

gassmoeller:
Ok, done.

PR #2079, opened 2018-02-01
Maintenance feb18

tjhei:


tjhei:
relevant changes are in the first commit

source/boundary_velocity/ascii_data.cc
```diff
@@ -130,9 +130,9 @@ namespace aspect
                                             "contain the number of grid points in each dimension as "
                                             "for example '# POINTS: 3 3'. "
                                             "The order of the data columns "
-                                            "has to be `x', `velocity_x', `velocity_y' in a 2d model "
-                                            "or `x', `y', `velocity_x', `velocity_y', "
-                                            "`velocity_z' in a 3d model. "
+                                            "has to be `x', `velocity$_x$', `velocity$_y$' in a 2d model "
+                                            "or `x', `y', `velocity$_x$', `velocity$_y$', "
+                                            "`velocity$_z$' in a 3d model. "
```
bangerth: err, how does this print? as an underscore? as a lower index? I'd just mark this up with a double backslash

tjhei: I stole that from source/prescribed_stokes_solution/ascii_data.cc and it looks correctly like a subscript x.

bangerth: OK if that works.

PR #2080, opened 2018-02-01
Fix up questionable LaTeX usage.

bangerth:
Follow-up to #2079.


gassmoeller:
Would not the standard usage be: `$_{x}$` and `$_{y}$`? I think the curly braces can be omitted here, because it is just a single letter, but if the subscript were longer it would need the curly braces around it.

bangerth:
What I'm going for here is that if you write `$_x$` then that's subsetting `x` to something, but you're omitting the something. If you write `${}_x$`, then it's clear that you're subsetting it to what's in the curly braces. That's empty here, but at least the *subject* of the sentence is provided :-)

gassmoeller:
Oh, ok makes sense now.

PR #2081, opened 2018-02-05
Update documentation of Simulator::solve_stokes().

bangerth:
While there, also use an empty line to better delineate two parts of a
comment.

@mfraters: this is what we talked about last week. Can you take a look?
I didn't update comments in the solve_stokes() function itself so as not to
step onto the changes you make in #2070.

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -618,7 +618,7 @@ namespace aspect
        * system involving the velocity and pressure variables).
        *
        * @note If this function is called from a nonlinear solver -- e.g., the
-       * iterated Stokes, iterated IMPES, or Newton solver --, then the
+       * iterated Stokes, or iterated IMPES --, then the
        * @p current_linearization_point is the solution of the previous
        * iteration (or the solution extrapolated from the previous time
        * steps, if this is the first nonlinear iteration). Let us call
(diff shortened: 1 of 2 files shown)


include/aspect/simulator.h
```diff
@@ -611,12 +611,43 @@ namespace aspect
       /**
        * Solve the Stokes linear system. Return the initial nonlinear
        * residual, i.e., if the linear system to be solved is $Ax=b$, then
-       * return $\|Ax_0-b\|$ where $x_0$ is the initial guess for the solution
-       * variable and is taken from the current_linearization_point member
+       * return $\|Ax_k-b\|$ where $x_k$ is the initial guess for the solution
+       * variable and is taken from the @p current_linearization_point member
        * variable. For the purpose of this function, this residual is computed
-       * only the velocity and pressure equations (i.e., for the 2x2 block
+       * only rom the velocity and pressure equations (i.e., for the 2x2 block
```
MFraters: from?

bangerth: Yes, good point. Fixed!

bangerth:
Hm. If you don't too violently disagree, can we merge this and I revisit what I write here once we've got #2070 sorted out?

MFraters:
Sure, it is correct with respect to what it is currently doing :)

PR #2084, opened 2018-02-08
add the exit timer in the linear solver catch

MFraters:
Based on a discussion in #2006. When the expensive linear sovler crashes, the section `   Solve Stokes system` is never exited. This can cause problems for the failsafe in #2006. I am just not sure what happens with this solution when the cheap solver fails with anything else then a non-convergence.

### DIFF
--- a/source/simulator/solver.cc
+++ b/source/simulator/solver.cc
@@ -841,6 +841,8 @@ namespace aspect
 
                     f.close();
 
+                    computing_timer.exit_section();
+
                     // avoid a deadlock that was fixed after deal.II 8.5.0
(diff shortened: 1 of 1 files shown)


source/simulator/solver.cc
```diff
@@ -815,6 +815,7 @@ namespace aspect
             // processors
             catch (const std::exception &exc)
               {
+                computing_timer.exit_section();
```
bangerth: please move the statement to line 843, right before the `// avoid a deadlock` comment. This way, it covers all of the code here as well.

PR #2086, opened 2018-02-09
Make SPD safety factor available as a parameter.

MFraters:
This pull request makes the SPD safety factor available as an input parameter.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -989,7 +989,7 @@ namespace aspect
     double compute_spd_factor(const double eta,
                               const SymmetricTensor<2,dim> &strain_rate,
                               const SymmetricTensor<2,dim> &dviscosities_dstrain_rate,
-                              const double safety_factor);
+                              const double SPD_safety_factor);
 
     /**
      * Converts an array of size dim to a Point of size dim.
--- a/source/simulator/newton.cc
+++ b/source/simulator/newton.cc
@@ -245,6 +245,14 @@ namespace aspect
   }
 
   template <int dim>
+  double
+  NewtonHandler<dim>::
+  get_SPD_safety_factor() const
+  {
+    return SPD_safety_factor;
+  }
+
+  template <int dim>
   std::string
   NewtonHandler<dim>::
   get_newton_stabilization_string(const NewtonStabilization preconditioner_stabilization) const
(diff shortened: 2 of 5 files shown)


source/simulator/newton.cc
```diff
@@ -331,6 +339,13 @@ namespace aspect
                            "When this parameter is true and the linear solver fails, we try again, but now with SPD stabilization "
                            "for both the preconditioner and the velocity block. The SPD stabilization will remain active untill "
                            "the next timestep, when the default values are restored.");
+
+
+        prm.declare_entry ("SPD safety factor", "0.9",
+                           Patterns::Double (0,1),
+                           "The SPD factor ensures positive semi-definiteness. So when the the SPD factor is smaller then one, it "
+                           "scales the eigenvalue back to zero. In practice we want to stay away from a zero eigenvalue, so we scale "
+                           "the SPD value by this value.");
```
bangerth: Write this as follows:
```
  When stabilizing the Newton matrix, we can encounter situations where the coefficient inside the elliptic (top-left)
  block becomes negative or zero. This coefficient has the form $1+x$ where $x$ can sometimes be smaller than
  $-1$. In this case, the top-left block of the matrix is no longer positive definite, and both preconditioners and
  iterative solvers may fail. To prevent this, the stabilization computes an $\alpha$ so that $1+\alpha x$ is
  never negative. This $\alpha$ is chosen as $1$ if $x\ge -1$, and $\alpha=-\frac 1x$ otherwise. (Note that
  this always leads to $0\le \alpha \le 1$.)  On the
  other hand, we also want to stay away from $1+\alpha x=0$, and so modify the choice of $\alpha$ to
  be $1$ if $x\ge -c$, and $\alpha=-\frac cx$ with a $c$ between zero and one. This way, if $c<1$, we are
  assured that $1-\alpha x>c$, i.e., bounded away from zero.
```

MFraters:
I guess it is a good idea to also adjust the spd computation directly in this pull request. So let me be clear what you propose for the function in utilities. Is this equivalent to it?
```
    template <int dim>
    double compute_spd_factor(const double eta,
                              const SymmetricTensor<2,dim> &dviscosities_dstrain_rate,
                              const double safety_factor)
    {
      // if the strain rate is zero, or the derivative is zero, then
      // the exact choice of alpha factor does not matter because the
      // factor that it multiplies is zero -- so return the best value
      // (i.e., one)
      if ((strain_rate.norm() == 0) || (dviscosities_dstrain_rate.norm() == 0))
        return 1;

      const double norm_a_b = std::sqrt((strain_rate*strain_rate)*(dviscosities_dstrain_rate*dviscosities_dstrain_rate));
      const double contract_a_b = (strain_rate*dviscosities_dstrain_rate);
      const double one_minus_part = 1 - (contract_a_b / norm_a_b);
      const double denom = one_minus_part * one_minus_part * norm_a_b;

      if (denom == 0)
        return 1.0;
      else
        {
          const double alpha = (2.0*eta)/denom;
          if (safety_factor * alpha >= 1.0)
            return 1.0;
          else
            return std::max(0.0, safety_factor * alpha);
        }
    }
```

bangerth:
This does not look correct to me. Let's talk this over later.

MFraters:
Ok, I have updated as discussed. Should I also change the input parameter name form `SPD safety factor` to `C safety`? It will be more consistent with the paper, but less descriptive.

bangerth:
No, keep the name of the parameter in the input file and in the program. It's just that in mathematical notation, multi-character variable names don't work well (though multi-character indices work ok.)

include/aspect/utilities.h
```diff
@@ -989,7 +989,7 @@ namespace aspect
     double compute_spd_factor(const double eta,
                               const SymmetricTensor<2,dim> &strain_rate,
                               const SymmetricTensor<2,dim> &dviscosities_dstrain_rate,
-                              const double safety_factor);
+                              const double C_safety);
```
bangerth: I think you can keep calling things here `SPD_safety_factor`. That also matches the name of the variable you pass in, which makes it easier to read.

source/utilities.cc
```diff
@@ -2406,7 +2406,7 @@ namespace aspect
     double compute_spd_factor(const double eta,
                               const SymmetricTensor<2,dim> &strain_rate,
                               const SymmetricTensor<2,dim> &dviscosities_dstrain_rate,
-                              const double safety_factor)
+                              const double C_safety)
```
bangerth: same here

PR #2087, opened 2018-02-09
Use scoped timers.

### LINKED ISSUE #2085: Make better use of TimerOutput::Scope
#2084 shows that we should really make use of the `TimerOutput::Scope` class to ensure that we're leaving timer sections when leaving scopes.

bangerth:
Fixes #2085.

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -396,7 +396,7 @@ namespace aspect
     if (parameters.use_direct_stokes_solver)
       return;
 
-    computing_timer.enter_section ("   Build Stokes preconditioner");
+    TimerOutput::Scope timer (computing_timer, "   Build Stokes preconditioner");
     pcout << "   Rebuilding Stokes preconditioner..." << std::flush;
 
     // first assemble the raw matrices necessary for the preconditioner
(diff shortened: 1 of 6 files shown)


source/simulator/assembly.cc
```diff
@@ -608,17 +608,15 @@ namespace aspect
...
-        else
-          computing_timer.enter_section ("   Assemble Stokes system newton");
-      }
-    else
-      computing_timer.enter_section ("   Assemble Stokes system rhs");
+    TimerOutput::Scope timer (computing_timer,
+                              (!assemble_newton_stokes_system ?
+                               "   Assemble Stokes system" :
+                               (assemble_newton_stokes_matrix ?
+                                (newton_handler->get_newton_derivative_scaling_factor() == 0 ?
```
tjhei: this is quite an ugly monstrosity. Would it be simpler if you leave the if block and assign a const char* instead?

bangerth: Yes, that place is ugly. I would need to create another variable with the same scope as the timer section for this. I can do that if you want me to.

tjhei: Shame that we can not use a lambda yet. I don't feel strongly about it. Either way, go ahead and merge yourself.

bangerth: Yes, a lambda would be nice here...

bangerth:
Since you don't feel strongly, let me go ahead and merge as is. We can always clean up these places later on when we are allowed to use lambdas. (Clean-ups is apparently all I do these days anyway...)

PR #2088, opened 2018-02-09
[Bugfix] Make models that use melt heating models work with operator splitting

jdannberg:
Due to some missing lines in the operator splitting routine, heating models did not get additional material model outputs (like melt outputs) and so any heating model that requires these outputs did not work with operator splitting. 
In addition, if melt transport was switched off, the reaction rate outputs were not filled correctly in the melt material models, and so they could not be used together with operator splitting. 

This pull request fixes both issues.

### DIFF
--- a/source/material_model/melt_global.cc
+++ b/source/material_model/melt_global.cc
@@ -214,8 +214,15 @@ namespace aspect
           else
             {
               out.viscosities[i] = eta_0;
+
+              // no melting/freezing is used in the model --> set all reactions to zero
               for (unsigned int c=0; c<in.composition[i].size(); ++c)
-                out.reaction_terms[i][c] = 0.0;
+                {
+                  out.reaction_terms[i][c] = 0.0;
+
+                  if (this->get_parameters().use_operator_splitting && reaction_rate_out != NULL)
+                    reaction_rate_out->reaction_rates[i][c] = 0.0;
+                }
             }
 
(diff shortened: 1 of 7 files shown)


source/material_model/melt_global.cc
```diff
@@ -213,9 +213,15 @@ namespace aspect
             }
           else
             {
+              // no melting/freezing is used in the model --> set all reactions to zero
```
tjhei: shouldn't this comment go below the next line?

source/material_model/melt_simple.cc
```diff
@@ -333,9 +333,15 @@ namespace aspect
             }
           else
             {
+              // no melting/freezing is used in the model --> set all reactions to zero
```
tjhei: same

source/material_model/melt_global.cc
jdannberg: Yes, I guess that makes more sense. Fixed in both places.

tjhei:
ready to be merged when the tester is happy.

jdannberg:
There seems to be a problem with the astyle tester, in a file I did not touch in this pull request (tests/particle_property_function) ...and the diff just tells me that there's a difference in a binary file. Does anyone know what could cause this?

tjhei:
I have no idea what happened. Reran and it succeeded now.

PR #2092, opened 2018-02-13
fix | bug in newton solver stablization.

MFraters:
Although in the output the correct stabilization appeared, it was not correctly applied. Somehow this bug slipped into #2006. This should fix it (I have tested it now, and am in the process of making tests for the stabilization).


bangerth:
Oh, yes, doh. How did I miss this?

PR #2094, opened 2018-02-14
Let solve_stokes() return something more uniform.

### LINKED ISSUE #2082: Let solve_stokes() return a pair of values
Discussing #2070 with @MFraters, we realized that the `solve_stokes()` function returns something (the starting nonlinear residual) in most situations and something else (the final linear residual) when using the Newton solver. That's awkward to document and more awkward to use.

The function should just always return a pair. Implement this when we're done with #2070.

bangerth:
Previously, solve_stokes() returned something in some contexts, and something else in others.
Make this more uniform by just returning a pair of values.

While there, also adjust some naming of variables to indicate that they refer to
*nonlinear* residuals.

Fixes #2082.

### DIFF
--- a/source/simulator/solver.cc
+++ b/source/simulator/solver.cc
@@ -522,7 +522,8 @@ namespace aspect
 
 
   template <int dim>
-  double Simulator<dim>::solve_stokes ()
+  std::pair<double,double>
+  Simulator<dim>::solve_stokes ()
   {
     TimerOutput::Scope timer (computing_timer, "   Solve Stokes system");
     pcout << "   Solving Stokes system... " << std::flush;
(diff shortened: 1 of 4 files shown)


bangerth:
@MFraters -- I'd like you take a close look at this and tell me if this makes sense. Start with the documentation of `solve_stokes()` and its implementation, and then see whether I'm using things consistently in the places where we call this function.

Because `solve_stokes()` returned different things before, depending on context, and now returns two values at once, I'm not sure whether I'm picking the correct one of the two values in all of the contexts. I think you know that part of the code better than I do.

source/simulator/solver.cc
```diff
@@ -522,19 +522,20 @@ namespace aspect
...
     TimerOutput::Scope timer (computing_timer, "   Solve Stokes system");
     pcout << "   Solving Stokes system... " << std::flush;
 
     // extract Stokes parts of solution vector, without any ghost elements
     LinearAlgebra::BlockVector distributed_stokes_solution (introspection.index_sets.stokes_partitioning, mpi_communicator);
 
-    double initial_residual = numbers::signaling_nan<double>();
+    double initial_nonlinear_residual = numbers::signaling_nan<double>();
     // the final linear residual is for the direct solver zero,
     // so we set it here to zero. For the iterative solver, we
```
MFraters: please update this.

source/simulator/solver.cc
```diff
@@ -919,7 +926,8 @@ namespace aspect
     if (parameters.include_melt_transport)
       melt_handler->compute_melt_variables(solution);
 
-    return initial_residual;
+    return std::pair<double,double>(final_linear_residual,
+                                    initial_nonlinear_residual);
```
MFraters: looking at the documentation in simulator.h and the code in solver_schemes, I think you mend to do this the other way around.

bangerth:
OK, let's see whether this round makes more sense now.

MFraters:
Looks good to me!

Issue #2083, opened 2018-02-07
Move Newton parameters into a separate struct


bangerth:
Reviewing #2006 and #2070 reminds me that the way the `NewtonHandler` deals with run-time parameters is different from the way we do things elsewhere. Elsewhere, we pack all parameters into a separate class whose member variables are all `public:`. But the `NewtonHandler` keeps these parameters in the main class, makes them `private:` and then provides accessor functions.

I think this is unnecessary because all of the setters and getters don't do anything other than accessing these variables -- so we may as well make the variables `public:`, and then to make sure they don't clutter up the main class, we may as well move them into a `NewtonHandler::Parameters` class or similar.

@MFraters -- FYI

PR #2096, opened 2018-02-14
Break the runtime parameters out of the NewtonHandler class.

bangerth:
This is the first part of what I want to do for #2083. This
patch does not remove the many getter/setter functions for
all of the parameters, but that will come next. One step
at a time.

### DIFF
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -57,15 +57,11 @@ namespace aspect
     };
   }
 
-  /**
-   * A Class which can declare and parse parameters and creates
-   * material model outputs for the Newton solver.
-   */
 
-  template <int dim>
-  class NewtonHandler: public SimulatorAccess<dim>
+  namespace Newton
   {
-    public:
+    struct Parameters
+    {
 
       /**
        * This enum describes the type of stabilization is used
@@ -75,7 +71,7 @@ namespace aspect
        * only symmetrized, and PD represents that we do the same as
        * what we do for SPD, but without the symmetrization.
        */
-      enum NewtonStabilization
+      enum Stabilization
       {
         none = 0,
         symmetric = 1,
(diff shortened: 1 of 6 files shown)


bangerth:
I should add that there should be no functional change -- it's really just a bit of refactoring.

gassmoeller:
Nice first step. 

I realize that we do the same for the main parameters, but out of curiosity: Can you give a reasoning for why in this case it is allowed to directly access public member variables of a class, instead of providing get and set functions? Is it because this class is only a support class for the Simulator, with the main task of storing these parameters? And would that be different for e.g. the MeltHandler, which also stores parameters, but needs them internally for the stuff the MeltHandler does?

The above is just a philosophical question, I do not oppose merging this PR. Feel free to merge.

bangerth:
You know that I don't like merging my own patches, so I'll leave this to someone else :-)

I think of these `Parameters` structures as just a way to better group things that conceptually belong together, rather than as an independent entity that supports code modularization. As such, it makes sense to allow direct access. A symptom of the fact that the Newton parameters are also sort of in this category is that the setters/getters we currently have all don't actually do anything. Setting a variable never has a side effect, there are no functions that set one parameter and as a side effect also need to do something to other variables, etc. In other words, if you have these sort of trivial getters and setters, these variables are *for all practical purposes* `public` anyway -- they're just unnecessarily complicated to access. 

As so many things, it's ultimately a judgement call whether these classes should be considered "just a way to group variables" or have independent functionality. My gut tells me that here, we're in the first category, not in the second, but I do agree that it's difficult to articulate this concisely. As a consequence, I would understand if others feel differently about this and are not in favor of the patch.

gassmoeller:
No I think this makes perfect sense, I just wanted to make sure, I am not missing some clear distinction. Good to go then.

PR #2097, opened 2018-02-16
Remove getters/setters for the Newton parameters

### LINKED ISSUE #2083: Move Newton parameters into a separate struct

bangerth:
This is the second step for #2083 and removes the getters and setters. I also renamed and moved the function that converts values of an enum to strings.

Fixes #2083.


include/aspect/newton.h
```diff
@@ -122,26 +123,53 @@ namespace aspect
...
       double              SPD_safety_factor;
     };
+
+
+    /**
+     * Get a std::string describing the stabilization type used for the
+     * preconditioner.
+     */
+    std::string
+    to_string(const Newton::Parameters::Stabilization preconditioner_stabilization);
```
MFraters: The name look a bit too generic to me, because in the code you now get `Newton::to_string(some_parameter)`. I would prefer to call it something like `stabilization_to_string` or `stabilization_type_to_string`

include/aspect/newton.h
```diff
@@ -122,26 +123,53 @@ namespace aspect
...
+       * derivative part of the Newton Stokes solver in the assembly.
        *
-       * See the get_newton_derivative_scaling_factor() function for an
-       * explanation of the purpose of this factor.
+       * The exact Newton matrix consists of the Stokes matrix plus a term
+       * that results from the linearization of the material coefficients.
+       * The scaling factor multiplies these additional terms. In a full
+       * Newton method, it would be equal to one, but it can be chosen
+       * smaller in cases where the resulting linear system has undesirable
+       * properties.
```
MFraters: So the main reason I add this parameter it to use it together with the `use_newton_residual_scaling_method`, because it helps the global convergence of the iteration. Because far way from the solution, you use mostly Picard, and the closer you get the the solution, the more Newton you use. And when it diverges for some reason it gets closer to Picard. It doesn't make it completely globally convergence, but it may help quite a bit. Although maybe part of this explanation should be at the `use_newton_residual_scaling_method` parameter. 

I can also update the documentation in a later pull request if you want.

include/aspect/newton.h
bangerth: I chose that name because C++11 has a function of that name for all sorts of data types: 
http://www.cplusplus.com/reference/string/to_string/
http://en.cppreference.com/w/cpp/string/basic_string/to_string
So this is sort of a standard name to use in all contexts where you want to convert a something to a string, and the distinction is really only in the type of the argument.

include/aspect/newton.h
bangerth: OK, I'll gladly let you make this update at a later time -- this is orthogonal to the current patch anyway.

include/aspect/newton.h
MFraters: ah, oke. I didn't know that. Then I am fine with this change.

include/aspect/newton.h
MFraters: sure

PR #2101, opened 2018-02-19
avoid clang warning

tjhei:
remove:
```
simulator_signals.cc:110:22: warning: explicit instantiation of
'call_connector_functions<2>' that occurs after an explicit
specialization has no effect [-Winstantiation-after-specialization]
```


bangerth:
Yes, this is the correct patch. Thanks!

PR #2102, opened 2018-02-20
Clean up a few lines of code for readability.

bangerth:

PR #2103, opened 2018-02-21
fix visco-plastic derivative

MFraters:
Fixes problem with #2102.


bangerth:

PR #2104, opened 2018-02-21
fix time step size for zero velocity

jdannberg:
This is something that annoyed me several times now: If a model has a zero velocity (either because that is the prescribed Stokes solution, or just because nothing is moving), the time step size is arbitrarily set to 1s. This doesn't make any sense, in particular as we have a parameter that allows the user to set a maximum time step. But because that setting of the time step size to 1 occurs before we apply the maximum time step, the maximum time step parameter is essentially not used in case of zero velocity, because 1s is pretty much always smaller than what you want to use as your maximum time step. 

I suggest that instead of setting the time step size to 1, it should just be set to the maximum time step (which is there for exactly that purpose).

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -605,10 +605,9 @@ namespace aspect
       {
         // If the velocity is zero and we either do not compute the conduction
         // timestep or do not have any conduction, then it is somewhat
-        // arbitrary what time step we should choose. In that case, do as if
-        // the velocity was one
-        new_time_step = (parameters.CFL_number /
-                         (parameters.temperature_degree * 1));
+        // arbitrary what time step we should choose. In that case, set the time
+        // step to the 'Maximum time step'.
+        new_time_step = parameters.maximum_time_step;
       }
 
     // make sure that the timestep doesn't increase too fast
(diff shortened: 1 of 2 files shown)


bangerth:
I'm for it -- the existing functionality clearly makes no sense. Do others agree?

MFraters:
Yes, this also gives more control over the behaviour (by the maximum timestep)

source/simulator/helper_functions.cc
```diff
@@ -605,10 +605,9 @@ namespace aspect
       {
         // If the velocity is zero and we either do not compute the conduction
         // timestep or do not have any conduction, then it is somewhat
-        // arbitrary what time step we should choose. In that case, do as if
-        // the velocity was one
-        new_time_step = (parameters.CFL_number /
-                         (parameters.temperature_degree * 1));
+        // arbitrary what time step we should choose. In that case, set the time
+        // step to the 'Maximum time step'.
```
MFraters: Maybe it is good to add in the description to add some examples when this may actually happen, like you state in the pull request description.

jdannberg: Okay, done.

gassmoeller:
Looks good.

PR #2105, opened 2018-02-21
Remove model settings subsection

gassmoeller:
This sits on top of #2099 and #2051, and moves the parameters in the 'Model settings' subsection into more appropriate subsections. When those other PRs are merged, I will rebase this one. It will be simplest to review the PR commit by commit (because the functional changes in the first two commits are small, and the last commit only contains the automatic renaming of all parameter files and is huge). After struggling with the arcane `sed` syntax to do this, I suggest to switch to something else if we do this again for another release (e.g. python, perl, awk).


gassmoeller:
Superseded by #2157

PR #2107, opened 2018-02-22
only ask for the derivatives when actually needed.

MFraters:
This pull request optimizes the defect correction Picard scheme in the Newton solver code. It does it by only adding the derivatives as additional output when they are actually needed.


gassmoeller:
Make sense. Tests pass.

PR #2108, opened 2018-02-22
add the Eisenstat Walker method to the defect correction Picard iteration.

MFraters:
Like the title says. It may slow down the convergence of the Picard iteration a bit, but I did a few test, and it is usually quite a bit faster.


MFraters:
somehow I can't upload the comparison image here to github, so I put it here: [Eisenstat Walker Picard comparison](https://mennofraters.nl/images/Eisenstat_Walker_Picard.png)

gassmoeller:
I can do a review later, but let me ask a question first: Why do you want to use that method? You say it is faster, but it converges slower? Why is it faster then?

MFraters:
It automatically computes an 'optimal' linear tolerance every nonlinear iteration, which is usually much lower then the default linear tolerance. That is why it is faster.

include/aspect/newton.h
```diff
@@ -156,6 +156,7 @@ namespace aspect
        */
       double              nonlinear_switch_tolerance;
 
+      bool                use_Eisenstat_Walker_method_for_Picard;
```
gassmoeller: please rename to `use_Eisenstat_Walker_method_for_Picard_iterations` to be consistent with the other parameters

source/simulator/newton.cc
```diff
@@ -246,6 +246,12 @@ namespace aspect
                              "$0\\le \\alpha \\le 1$.)  On the other hand, we also want to stay away from $1+\\alpha x=0$, and so modify the choice of "
                              "$\\alpha$ to be $1$ if $x\\ge -c$, and $\\alpha=-\\frac cx$ with a $c$ between zero and one. This way, if $c<1$, we are "
                              "assured that $1-\\alpha x>c$, i.e., bounded away from zero.");
+
+          prm.declare_entry ("Use the Eisenstat Walker method for Picard", "false",
```
gassmoeller: same here, and you can skip the 'the'. So: "Use Eisenstat Walker method for Picard iterations"

MFraters:
I addressed your comments and added two tests :)

bangerth:
@gassmoeller: To be precise, this patch makes the Picard iteration require more nonlinear iterations, but each iteration is far cheaper because we don't have to solve the linear system with the same accuracy. The sum is a *faster* time-to-solution, despite the larger number of iterations.

source/simulator/newton.cc
```diff
@@ -246,6 +246,12 @@ namespace aspect
                              "$0\\le \\alpha \\le 1$.)  On the other hand, we also want to stay away from $1+\\alpha x=0$, and so modify the choice of "
                              "$\\alpha$ to be $1$ if $x\\ge -c$, and $\\alpha=-\\frac cx$ with a $c$ between zero and one. This way, if $c<1$, we are "
                              "assured that $1-\\alpha x>c$, i.e., bounded away from zero.");
+
+          prm.declare_entry ("Use Eisenstat Walker method for Picard iterations", "false",
+                             Patterns::Bool(),
+                             "Turns on the Eisenstat Walker method for the defect correction Picard iteration, e.g. the Newton "
```
bangerth: The use of "e.g." is incorrect here I think. Maybe rewrite the whole description as follows:

"If set to true, the Picard iteration uses the Eisenstat Walker method to determine how accurately linear systems need to be solved. The Picard iteration is used, for example, in the first few iterations of the Newton method before the matrix is built including derivatives of the model, since the Picard iteration generally converges even from points where Newton's method does not.
\n\n
Once derivatives are used in a Newton method, \\aspect{} always uses the Eisenstat Walker method."

bangerth:
This looks good to me, with the exception of the text for the parameter. I think that if you address the one comment, then you can merge this yourself!

PR #2112, opened 2018-02-25
allow simple nonlinear material model without the deviator

MFraters:
allow simple nonlinear material model to be computed without the deviator. I noticed this may give slightly different behaviour for the Picard iteration, but a large difference in behaviour for the Newton solver.


benchmarks/nonlinear_channel_flow/simple_nonlinear.cc
```diff
@@ -197,7 +199,8 @@ namespace aspect
                   // to prevent a division-by-zero, and a floating point exception.
                   // Otherwise, calculate the square-root of the norm of the second invariant of the deviatoric-
                   // strain rate (often simplified as epsilondot_ii)
-                  const double edot_ii_strict = std::sqrt(0.5*deviator(in.strain_rate[i])*deviator(in.strain_rate[i]));
+                  const SymmetricTensor<2,dim> deviator_strain_rate = use_deviator_of_strain_rate ? deviator(in.strain_rate[i]) : in.strain_rate[i];
```
bangerth: I find this variable name confusing. Maybe just call it `edot`?

benchmarks/nonlinear_channel_flow/simple_nonlinear.cc
```diff
@@ -212,7 +215,10 @@ namespace aspect
                            * 0.5 * ((1/stress_exponent)-1) * std::pow(2,2) * out.viscosities[i] * (1/(edot_ii*edot_ii)) * deviator(in.strain_rate[i]),
                            * but we can (and do) simply it as:
                            */
-                          composition_viscosities_derivatives[c] = 2.0 * (stress_exponent_inv-1) * composition_viscosities[c] * (1.0/(edot_ii*edot_ii)) * deviator(in.strain_rate[i]);
+                          composition_viscosities_derivatives[c] = 2.0 * (stress_exponent_inv-1) * composition_viscosities[c] * (1.0/(edot_ii*edot_ii)) * deviator_strain_rate;
+
+                          if (use_deviator_of_strain_rate == true)
+                            composition_viscosities_derivatives[c]*deviator_tensor<dim>();
```
bangerth: This statement has no effect: you are computing the product of two things, but not assigning the result to anything. Can you explain what you wanted to do here?

benchmarks/nonlinear_channel_flow/simple_nonlinear.cc
```diff
@@ -338,6 +344,11 @@ namespace aspect
                              "This is the p value in the generalized weighed average equation: "
                              " $\\text{mean} = \\frac{1}{k}(\\sum_{i=1}^k \\big(c_i \\eta_{\\text{eff}_i}^p)\\big)^{\\frac{1}{p}}$. "
                              " Units: $Pa s$");
+
+          // strain-rate deviator parameter
+          prm.declare_entry ("Use deviator of strain-rate", "true",
+                             Patterns::Bool(),
+                             "This value determines wheter to use the deviator of the strain-rate in computing the viscosity.");
```
bangerth: add ", or simply the strain rate $\\varepsilon(\\mathbf u)$."

bangerth:
One of the tests is failing. Can you investigate?

MFraters:
Now I think of it, it makes sense that there is a difference, because the deviator derivative is now added. But somehow on my computer there is no difference in the screen output, while here there is. Will investigate further.

bangerth:
I think you should keep the default of the parameter so that it is backward compatible. Then the test should not change.

MFraters:
The problem is that it used to be with the deviator, but the derivative of the deviator was not added. Sot that makes a difference for the tests. Then it also adds the possibility for the material model without the deviator. I didn't add a test for that yet.

bangerth:
I see. I think the right would have been to do things in two PRs where one fixes the bug (and updates the existing test) and the other adds the new option. Splitting logically separate things into separate patches usually makes it much easier to merge things.

As for this patch, can you update the test output? Do you plan to still add a test for the case without the deviator?

MFraters:
I have update the tests and added two new test in #2114

PR #2113, opened 2018-02-26
Fix a LaTeX markup problem.

bangerth:
This is a follow-up to #2108. @mfraters: In order to get a backslash into the output,
you need to write two backslashes in the C++ code.


MFraters:
@bangerth Thanks, I also fixed it in #2112

PR #2115, opened 2018-02-27
Add spiegelman material model to benchmark section.

MFraters:
This is needed to implement some of the tests, and will serve as a start including the Spiegelman 2016 benchmark into ASPECT.

### DIFF
--- a/benchmarks/spiegelman_et_al_2016_gcubed/CMakeLists.txt
+++ b/benchmarks/spiegelman_et_al_2016_gcubed/CMakeLists.txt
@@ -1,0 +1,39 @@
+# Copyright (C) 2015 - 2018 by the authors of the ASPECT code.
+#
+# This file is part of ASPECT.
+#
+# ASPECT is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2, or (at your option)
+# any later version.
+#
+# ASPECT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with ASPECT; see the file doc/COPYING.  If not see
+# <http://www.gnu.org/licenses/>.
+
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.8)
+
+FIND_PACKAGE(Aspect 1.5.0 QUIET HINTS ${Aspect_DIR} ../ ../../ $ENV{ASPECT_DIR})
+
+IF (NOT Aspect_FOUND)
+  MESSAGE(FATAL_ERROR "\n"
+	"Could not find a valid ASPECT build/installation directory. "
+	"Please specify the directory where you are building ASPECT by passing\n"
+	"   -D Aspect_DIR=<path to ASPECT>\n"
+	"to cmake or by setting the environment variable ASPECT_DIR in your shell "
+	"before calling cmake. See the section 'How to write a plugin' in the "
+        "manual for more information.")
+ENDIF ()
+
+DEAL_II_INITIALIZE_CACHED_VARIABLES()
+
+SET(TARGET "spiegelman_material")
+PROJECT(${TARGET})
+
+ADD_LIBRARY(${TARGET} SHARED spiegelman_material.cc)
+ASPECT_SETUP_PLUGIN(${TARGET})
--- a/benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
+++ b/benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
@@ -1,0 +1,720 @@
+/*
+  Copyright (C) 2011 - 2018 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+#include <aspect/material_model/interface.h>
+#include <aspect/simulator_access.h>
+#include <aspect/newton.h>
+#include <aspect/utilities.h>
+#include <aspect/parameters.h>
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    /**
+     * The same material model as Drucker Prager, but this one supports multiple
+     * compositions in a way to reproduce the Spiegelman 2016 paper.
+     *
+     * The model is considered incompressible, following the definition
+     * described in Interface::is_compressible.
+     *
+     * The viscosity is computed according to the Drucker Prager frictional
+     * plasticity criterion based on a user-defined internal angle of friction $\phi$
+     * and cohesion $C$. In 3D:
+     * $\sigma_y = \frac{6 C \cos(\phi)}{\sqrt(3) (3+\sin(\phi))} +
+     * \frac{6 P \sin(\phi)}{\sqrt(3) (3+\sin(\phi))}$,
+     * where $P$ is the pressure.
+     * See for example Zienkiewicz, O. C., Humpheson, C. and Lewis, R. W. (1975),
+     * G&eacute;otechnique 25, No. 4, 671-689.
+     * With this formulation we circumscribe instead of inscribe the Mohr Coulomb
+     * yield surface.
+     * In 2D the Drucker Prager yield surface is the same
+     * as the Mohr Coulomb surface:
+     * $\sigma_y = P \sin(\phi) + C \cos(\phi)$.
+     * Note that in 2D for $\phi=0$, these criteria
+     * revert to the von Mises criterion (no pressure dependence).
+     * See for example Thieulot, C. (2011), PEPI 188, 47-68.
+     *
+     * Note that we enforce the pressure to be positive in the computation of
+     * the yield strength by replacing it with
+     * a zero value whenever it is negative to prevent negative
+     * yield strengths and viscosities.
+     * We then use the computed yield strength to scale back the viscosity on
+     * to the yield surface using the Viscosity Rescaling Method described in
+     * Kachanov, L. M. (2004), Fundamentals of the Theory of Plasticity,
+     * Dover Publications, Inc.
+     *
+     * To avoid numerically unfavourably large (or even negative) viscosity ranges,
+     * we regularize the viscosity by $\eta_{\text{eff}} = \eta_{\text{ref}} * \eta
+     * / (\eta_{\text{ref}} + \eta)$. Futhermore,
+     * we cut off the viscosity with a user-defined minimum and maximum viscosity:
+     * $\eta_eff = \frac{1}{\frac{1}{\eta_min + \eta}+
+     * \frac{1}{\eta_max}}$.
+     *
+     * Note that this model uses the formulation that assumes an incompressible
+     * medium despite the fact that the density follows the law
+     * $\rho(T)=\rho_0(1-\beta(T-T_{\text{ref}}))$.
+     *
+     *
+     * @ingroup MaterialModels
+     */
+
+    template <int dim>
+    class DruckerPragerCompositions : public MaterialModel::Interface<dim>, public ::aspect::SimulatorAccess<dim>
+    {
+      public:
+        std::vector<double> compute_volume_fractions( const std::vector<double> &compositional_fields) const;
+
+        double compute_second_invariant(const SymmetricTensor<2,dim> strain_rate, const double min_strain_rate) const;
+
+        double compute_viscosity(const double edot_ii,const double pressure,const int comp, const double constant_viscosity,const bool regularize, const double min_visc, const double max_visc) const;
+
+        virtual void evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
+                              MaterialModel::MaterialModelOutputs<dim> &out) const;
+
+        /**
+         * Return whether the model is compressible or not.  Incompressibility
+         * does not necessarily imply that the density is constant; rather, it
+         * may still depend on temperature or pressure. In the current
+         * context, compressibility means whether we should solve the continuity
+         * equation as $\nabla \cdot (\rho \mathbf u)=0$ (compressible Stokes)
+         * or as $\nabla \cdot \mathbf{u}=0$ (incompressible Stokes).
+         *
+         * This material model is incompressible.
+         */
+        virtual bool is_compressible () const;
+
+        virtual double reference_viscosity () const;
+
+        virtual double reference_density () const;
+
+        static
+        void
+        declare_parameters (ParameterHandler &prm);
+
+        virtual
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+
+        double reference_T;
+        std::vector<double> reference_T_list;
+
+
+        /**
+         * Whether or not to use the deviator of the strain-rate in the
+         * viscosity calculation.
+         */
+        bool use_deviator_of_strain_rate;
+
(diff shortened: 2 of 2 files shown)


MFraters:
This will also replace the drucker_prager_compositions material model in #2100

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
+/*
+  Copyright (C) 2011 - 2015 by the authors of the ASPECT code.
```
bangerth: please update

benchmarks/spiegelman_et_al_2016_gcubed/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+# Copyright (C) 2015 - 2016 by the authors of the ASPECT code.
```
bangerth: please update

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+     * We then use the computed yield strength to scale back the viscosity on
+     * to the yield surface using the Viscosity Rescaling Method described in
+     * Kachanov, L. M. (2004), Fundamentals of the Theory of Plasticity,
+     * Dover Publications, Inc.
+     *
+     * To avoid numerically unfavourably large (or even negative) viscosity ranges,
+     * we regularize the viscosity by $\eta_{\text{eff}} = \eta_{\text{ref}} * \eta
+     * / (\eta_{\text{ref}} + \eta)$. Futhermore,
+     * we cut off the viscosity with a user-defined minimum and maximum viscosity:
+     * $\eta_eff = \frac{1}{\frac{1}{\eta_min + \eta}+\\
```
bangerth: no need for the `\\` here

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+
+        /**
+         * Return whether the model is compressible or not.  Incompressibility
+         * does not necessarily imply that the density is constant; rather, it
+         * may still depend on temperature or pressure. In the current
+         * context, compressibility means whether we should solve the continuity
+         * equation as $\nabla \cdot (\rho \mathbf u)=0$ (compressible Stokes)
+         * or as $\nabla \cdot \mathbf{u}=0$ (incompressible Stokes).
+        *
+        * This material model is incompressible.
```
bangerth: odd indentation of the asterisks

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+      private:
+
+        double reference_T;
+        std::vector<double> reference_T_list;
+
+        /**
+         * Defining a minimum strain rate stabilizes the viscosity calculation,
+         * which involves a division by the strain rate. Units: $1/s$.
+         */
+        bool use_deviator_of_strain_rate;
```
bangerth: This comment does not seem to be attached to the right variable.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+
+        //optimize variables
+        const double sqrt_3 = std::sqrt(3);
+        const double sqrt_half = std::sqrt(0.5);
+        std::vector<double> sin_phi;
+        std::vector<double> cos_phi;
+
+
+
+    };
```
bangerth: just remove the empty variables

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+
+        unsigned int n_fields;
+        // averaging parameters
+        double viscosity_averaging_p;
+
+        bool use_analytical_derivative;
+
+        //optimize variables
+        const double sqrt_3 = std::sqrt(3);
+        const double sqrt_half = std::sqrt(0.5);
```
bangerth: Just use the exact formula in all of the places where you use the variables now. The compiler can optimize these things. (This is also not valid C++98, though it's valid C++11.)

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+        std::vector<double> sin_phi;
+        std::vector<double> cos_phi;
```
bangerth: what do these variables correspond to? We only have a single angle `phi`, so it's not clear to me how one would make a vector out of this.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+
+    };
+
+  }
+
+
+  namespace
+  {
+    std::vector<double>
+    get_vector_double (const std::string &parameter, const unsigned int n_fields, ParameterHandler &prm)
```
bangerth: I think this function now exists as `Utilities::possibly_extend_from_1_to_N`, so no need to duplicate the code.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+      return volume_fractions;
+    }
+
+    template <int dim>
+    double
+    DruckerPragerCompositions<dim>::
+    compute_second_invariant(const SymmetricTensor<2,dim> strain_rate, const double min_strain_rate) const
+    {
+      const double edot_ii_strict = std::sqrt(strain_rate*strain_rate);
+      const double edot_ii = std::max(edot_ii_strict, min_strain_rate*min_strain_rate);
```
bangerth: Same as discussed elsewhere: it should be only one power of `min_strain_rate`.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+      return edot_ii;
+    }
+
+    template <int dim>
+    double
+    DruckerPragerCompositions<dim>::
+    compute_viscosity(const double edot_ii,const double pressure,const int comp,const double prefactor,const bool regularize, const double min_visc, const double max_visc) const
```
bangerth: break these arguments onto separate lines

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+    {
+      double viscosity;
+      if (comp == 0)
+        {
+          const double strength = ( (dim==3)
+                                    ?
+                                    ( 6.0 * cohesion[comp] * cos_phi[comp] + 6.0 * std::max(pressure,0.0) * sin_phi[comp] )
+                                    / ( sqrt_3 * ( 3.0 + sin_phi[comp] ) )
+                                    :
+                                    cohesion[comp] * cos_phi[comp] + std::max(pressure,0.0) * sin_phi[comp] );
```
bangerth: You are here in a block where you know that `comp==0`, so we only ever use the first element of `cohesion`, `cos[phi]`, etc. But below, when `comp != 0`, we don't actually references these quantities at all and only return the prefactor. Is this a mistake?

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+
+                  if (derivatives != NULL)
+                    {
+                      if (use_analytical_derivative)
+                        {
+                          //analytic
+                          if (c == 0  && composition_viscosities[c] <= max_visc[c] && composition_viscosities[c] >= min_visc[c])
+                            {
+                              const double drucker_prager_viscosity = compute_viscosity(edot_ii,pressure,c,prefactor[c],false,min_visc[c],max_visc[c]);
+                              const double regulaization_adjustment = (ref_visc * ref_visc)
```
bangerth: typo. rename the variable to `regularization_adjustment`.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+                            }
+                          else
+                            {
+                              composition_viscosities_derivatives[c] = 0;
+                              composition_dviscosities_dpressure[c] = 0;
+                            }
+                        }
+                      else
+                        {
+                          // finite difference
```
bangerth: you checked that the two ways of computing things yield the same results, correct?

If so, let's just always use the analytic way of computing the derivatives because it's so much cheaper. In that case, just remove (i) this piece of code and (ii) the variable that decides which method to use.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+
+              if (derivatives != NULL)
+                {
+                  derivatives->viscosity_derivative_wrt_strain_rate[i] = Utilities::derivative_of_weighted_p_norm_average(out.viscosities[i],volume_fractions, composition_viscosities, composition_viscosities_derivatives, viscosity_averaging_p);
+                  derivatives->viscosity_derivative_wrt_pressure[i] = Utilities::derivative_of_weighted_p_norm_average(out.viscosities[i],volume_fractions, composition_viscosities, composition_dviscosities_dpressure, viscosity_averaging_p);
+
+
+#ifdef DEBUG
+                  for (int x = 0; x < dim; x++)
+                    for (int y = 0; y < dim; y++)
```
bangerth: * use `unsigned int` here
* instead of `x,y`, we like to use `d,e` as variables to loop over dimensions
* say `++x` instead of `x++`.

same in line 443/444 below

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+                      if (!dealii::numbers::is_finite(derivatives->viscosity_derivative_wrt_strain_rate[i][x][y]))
+                        std::cout << "Error: Averaged viscosity to strain-rate devrivative is not finite." << std::endl;
```
bangerth: I would just make this an `Assert`. Nothing good can happen downstream once you have a viscosity that is not finite

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+
+                  if (!dealii::numbers::is_finite(derivatives->viscosity_derivative_wrt_pressure[i]))
+                    {
+                      std::cout << "Error: Averaged viscosity to pressure devrivative is not finite. " << std::endl;
+                      for (unsigned int c=0; c < volume_fractions.size(); ++c)
```
bangerth: same here -- in fact, here you already have the assertion in line 442, so there is no need to output the text -- you'll get to this information in the debugger

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
bangerth: oh, in fact you already have the assertion in line 445, I believe

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+                             Patterns::Double(),
+                             "This is the p value in the generalized weighed average eqation: "
+                             " mean = \\frac{1}{k}(\\sum_{i=1}^k \\big(c_i \\eta_{\\text{eff}_i}^p)\\big)^{\\frac{1}{p}}. "
+                             " Units: $Pa s$");
+
+          // finite difference versus analytical
+          prm.declare_entry ("Use analytical derivative", "false",
+                             Patterns::Bool(),
+                             "A bool indicating wether to use finite differences to compute the derivative or to use "
+                             "the analytical derivative.");
```
bangerth: if you go with the suggestion above, then this is also no longer necessary.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+
+          prm.declare_entry ("Use deviator of strain-rate", "true",
+                             Patterns::Bool(),
+                             "Todo.");
```
bangerth: todo :-)

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+}
+
+// explicit instantiations
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    ASPECT_REGISTER_MATERIAL_MODEL(DruckerPragerCompositions,
+                                   "drucker prager compositions",
+                                   " An implementation of the spiegelman 2016 benchmark paper. It "
```
bangerth: no need for the leading whitespace

bangerth: Also, capitalize Spiegelman and maybe provide a complete reference to the paper

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,680 @@
...
+                                   "implements a regularized Drucker Prager material and a linear "
+                                   "material. The derivatives for the Newton solver have been "
```
bangerth: what does "linear material" reference here? Is it a DP material or a linear material?

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
MFraters: there is a phi per composition.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
MFraters: Good point. It has grown a bit like this since it was first intended as a Drucker Prager material model which supports different compositions, but then for the Spiegelman benchmark I needed a constant viscosity. I now changed it so that when the prefactor (which I renamed to constant_viscosity) is zero, it uses the Drucker Prager rhelology, otherwise it uses that value as the viscosity.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
MFraters: As discussed there are small difference between the two methods, but that should not affect the convergence. Since this is only a benchmark material, I think it is good to leave in both methods as a reference how they both can be implemented. I extended the documentation.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
MFraters: Yes, these where there go generate more output while testing, but they are not needed anymore, so I removed them.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
MFraters: The Spiegelman benchmark requires a combination of both. I clarified it a bit.

MFraters:
I think that I have addressed all your comments.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -0,0 +1,716 @@
...
+         * finite difference derivative for the Newton solver. Although
+         * the analytical derivative should be much faster to compute, the
+         * finite difference derivative can be useful when changing the
+         * rheology.
+         */
+        bool use_analytical_derivative;
+
+        //optimize variables
+        std::vector<double> sin_phi;
+        std::vector<double> cos_phi;
```
bangerth: Given your comment in the other place where I had a question: Can you move these two variables right below the `phi` vector above, and add a comment above `phi` of the sort "The friction angles phi for each of the compositional fields. Since we use their sines and cosines a lot, we only compute these values (for each of the phis) once and cache them."

This way, the purpose of these variables are clear.

MFraters:
Addresses comments :)

PR #2116, opened 2018-02-27
Clean up and document Newton Stokes preconditioner.

bangerth:


source/simulator/assemblers/newton_stokes.cc
```diff
@@ -109,16 +109,30 @@ namespace aspect
                 for (unsigned int j = 0; j < stokes_dofs_per_cell; ++j)
                   if (scratch.dof_component_indices[i] ==
                       scratch.dof_component_indices[j])
-                    data.local_matrix(i, j) += ((2.0 * eta * (scratch.grads_phi_u[i] * scratch.grads_phi_u[j]))
-                                                + one_over_eta * this->get_pressure_scaling() * this->get_pressure_scaling()
-                                                * (scratch.phi_p[i] * scratch.phi_p[j]))
+                    data.local_matrix(i, j) += (
+                                                 // top left block: approximate J^{uu}
```
MFraters: It might be good the make the distinction between this 'approximation' and the approximation of `J^{uu}` for the full Newton method. Something like: 'top left block: approximate J^{uu} by A^{uu} when using the defect correction Picard.'

bangerth: Well, it's *not* A^uu. The A^uu block uses the symmetric gradient, but this block here uses the regular gradient.

MFraters: You are right, so we are basically approximating A^uu

bangerth: Correct.

bangerth:
@MFraters -- updated version now online

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -109,16 +109,37 @@ namespace aspect
...
                   if (scratch.dof_component_indices[i] ==
                       scratch.dof_component_indices[j])
-                    data.local_matrix(i, j) += ((2.0 * eta * (scratch.grads_phi_u[i] * scratch.grads_phi_u[j]))
-                                                + one_over_eta * this->get_pressure_scaling() * this->get_pressure_scaling()
-                                                * (scratch.phi_p[i] * scratch.phi_p[j]))
+                    data.local_matrix(i, j) += (
+                                                 // top left block: for the current case with
+                                                 // derivative_scaling_factor==0 the top left block
+                                                 // of the system matrix only contains the usual
+                                                 // Stokes term. so approximate this block in the
```
MFraters: so->So

bangerth: Sure. Though we have many comments that start sentences with lower-case letters. For example, the first line of this comment starts with a lower-case letter. 

But happy to change -- new version now online.

MFraters: ah, I didn't notice that, I now see it everywhere... Anyway, not a big issue for me. Will merge it when the testers are done :)

PR #2119, opened 2018-02-27
Bring SPD safety factor in line with the one actually used in the computation.

MFraters:
It now uses the SPD safety factor use in the computation.


bangerth:

MFraters:
Ah, I had the postprocessor active them active for test with solvers which are not the Newton solver. I will change those tests.

PR #2122, opened 2018-03-01
World builder structure

MFraters:
@bangerth is this in the direction we discussed?

I also looked into translating the input into json, and I have currently two different options of which I think both would do, so that is the direction I will try out.


MFraters:
Btw, one of the downsides of the boost json parser is that is stores all values as strings, and therefore can't parse the types. There are way more advanced json libraries out there, but I think I can also just parse it myself (which is the option I am now going to try).

include/aspect/world_builder/interface.h
```diff
@@ -0,0 +1,214 @@
...
+    using namespace dealii;
+
+    /**
+     * Base class for classes that describe particular initial topographies
+     * for the domain.
+     *
+     * @ingroup WorldBuilder
+     */
+    template <int dim>
+    class Interface //: public SimulatorAccess<dim>
```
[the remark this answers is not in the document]
bangerth: It's not clear to me what this is actually an interface for. A description of a part of the earth? Maybe move the interface class into a (sub-)namespace with a name that makes this clear.

source/simulator/core.cc
```diff
@@ -152,6 +153,7 @@ namespace aspect
     parameters (prm, mpi_communicator_),
     melt_handler (parameters.include_melt_transport ? new MeltHandler<dim> (prm) : NULL),
     newton_handler (parameters.nonlinear_solver == NonlinearSolver::Newton_Stokes ? new NewtonHandler<dim> () : NULL),
+    world_builder (parameters.world_builder_directory != "" ? WorldBuilder::create_world_builder<dim> (prm) : NULL),
```
bangerth: I would always create such an object. It may simply not be used if no plugin references it.

The presence of the `world_builder_directory` variable in the main class confuses me. What is it used for? Why is that not a parameter of the world builder itself, and located in the corresponding structure?

source/simulator/parameters.cc
```diff
@@ -245,6 +246,10 @@ namespace aspect
                        "The name of the directory into which all output files should be "
                        "placed. This may be an absolute or a relative path.");
 
+    prm.declare_entry ("World builder file", "",
+                       Patterns::FileName (),
+                       "The location of the world builder json file.");
```
bangerth: JSON

source/simulator/parameters.cc
```diff
@@ -1055,6 +1060,9 @@ namespace aspect
                                  mpi_communicator,
                                  false);
 
+    world_builder_directory = prm.get("World builder file");
```
bangerth: is it a file or a directory? ;-)

source/world_builder/interface.cc
```diff
@@ -0,0 +1,227 @@
...
+            AssertThrow (module_name != "" && model_name != "unspecified",
+                         ExcMessage ("Please specify the module you want to use (It is not allowed to be empty)."));
+
+            model_names.push_back(module_name);
+
+          }*/
+      }
+      //prm.leave_subsection ();
+
+      sort(model_names.begin(), model_names.end());
```
bangerth: std::sort

source/world_builder/interface.cc
```diff
@@ -0,0 +1,227 @@
...
+      std::vector<std::string>::iterator it;
+      it = unique(model_names.begin(), model_names.end());
```
bangerth: std::unique

include/aspect/world_builder/interface.h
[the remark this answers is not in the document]
MFraters: It is an interface for the world builder plugins. I expended the description.

source/simulator/core.cc
MFraters: >I would always create such an object. It may simply not be used if no plugin references it.

sure.

>The presence of the `world_builder_directory` variable in the main class confuses me. What is it used for? Why is that not a parameter of the world builder itself, and located in the corresponding structure?

It should have been `world_builder_file`. But now I won't need it now we always create the world builder, so I removed it

MFraters:
I have now build it into a version which can actually do what I want it to do and addressed the comments. Because the parameter handler of deal.ii is currently not capable to what I need to properly read in JSON (made a start on it in https://github.com/dealii/dealii/pull/5993), but even if I would get everything in into deal.ii 9.0 before it is released, it won't be in deal.ii 8.5. So I decided for now to focus on implementing it here in aspect locally in a way that is designed to be as compatible as possible with the deal.ii parameter handler and the internal structure of the boost property tree in it. 

I added an example world builder file (wb3.json) and a empty plugin which declares values to show that the values are correctly declared and parsed.

build_9.0/wb3.json
```diff
@@ -0,0 +1,18 @@
+{
```
gassmoeller: We will probably need a new directory in `data/world-builder/` or `data/initial-temperature/world-builder` for these files.
Also: Is it possible to add comments to these files?

include/aspect/parameters.h
```diff
@@ -319,6 +319,7 @@ namespace aspect
     bool                           use_conduction_timestep;
     bool                           convert_to_years;
     std::string                    output_directory;
+    std::string                    world_builder_directory;
```
gassmoeller: Hmm, why do you need this variable? Could it not be a member of the world builder?

include/aspect/simulator_access.h
```diff
@@ -663,6 +668,13 @@ namespace aspect
       const NewtonHandler<dim> &
       get_newton_handler () const;
 
+      /**
+       * Return a reference to the world builder handler that controls the
```
gassmoeller: I would just call it world builder, the 'handler' is unnecessary.

include/aspect/world_builder/continental_plate.h
```diff
@@ -0,0 +1,83 @@
+/*
+  Copyright (C) 2011 - 2016 by the authors of the ASPECT code.
```
gassmoeller: 2018 (or whenever you started it until 2018)

include/aspect/world_builder/continental_plate.h
```diff
@@ -0,0 +1,83 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_initial_temperature_world_generator_modules_constant_h
```
gassmoeller: world_builder_continental_plate_h
the same in the line below

include/aspect/world_builder/continental_plate.h
```diff
@@ -0,0 +1,83 @@
...
+#include <aspect/world_builder/interface.h>
+
+namespace aspect
+{
+  namespace WorldBuilder
+  {
+    using namespace dealii;
+
+    /**
+     * A class that implements zero initial topography.
```
gassmoeller: Needs a new comment

include/aspect/world_builder/continental_plate.h
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @ingroup InitialTopographyModels
+     */
+    template <int dim>
+    class Constant : public Interface<dim>
```
gassmoeller: Needs a new class name?

include/aspect/world_builder/continental_plate.h
```diff
@@ -0,0 +1,83 @@
...
+    {
+      public:
+
+        /**
+         * Return the temperature of an object.
+         */
+        virtual
+        void temperature (const Point<dim> &position,
+                          const unsigned int object_number,
+                          const WorldBuilderParameters<dim> &world_generator_parameters,
```
gassmoeller: Hmm, I do not like this very much. Which `WorldBuilderParameters` might be needed by this plugin? Should not all things either be World builder parameter (in which case World Builder handles whatever should happen with them), or local parameters of this plugin?

gassmoeller: I personally prefer to hand over the previous temperature as `const double temperature` and return the new temperature as `double`. But that might be a personal preference.

source/world_builder/interface.cc
```diff
@@ -0,0 +1,681 @@
+/*
+  Copyright (C) 2011 - 2016 by the authors of the ASPECT code.
```
gassmoeller: 2018

source/world_builder/interface.cc
```diff
@@ -0,0 +1,681 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+ */
+
+
+#include <aspect/global.h>
+//#include <aspect/world_builder.h>
```
gassmoeller: Remove

include/aspect/world_builder/interface.h
```diff
@@ -0,0 +1,269 @@
...
+     * used to declare the parameters that this initial topography model wants
+     * to read from input files.
+     * @param factory_function A pointer to a function that can create an
+     * object of this initial topography model.
+     *
+     * @ingroup WorldBuilder
+     */
+    template <int dim>
+    void
+    register_world_builder (const std::string &name,
```
gassmoeller: This function and the `create_world_builder` function are not the best way to implement this. I suppose you took this from one of the other plugin systems without manager, but it is simpler to use the same structure as for the `InitialTemperature::Manager` class. All the stuff in `create_world_builder` can either move into the constructor of `Manager` (if it is related to the core of `Manager`), or into `parse_parameters` (if it is related to the plugins). `register_world_builder` can become a member function of `Manager`, and `declare_parameters` can also become member functions of `Manager`.

source/simulator/core.cc
```diff
@@ -152,6 +153,7 @@ namespace aspect
     parameters (prm, mpi_communicator_),
     melt_handler (parameters.include_melt_transport ? new MeltHandler<dim> (prm) : NULL),
     newton_handler (parameters.nonlinear_solver == NonlinearSolver::Newton_Stokes ? new NewtonHandler<dim> () : NULL),
+    world_builder (WorldBuilder::create_world_builder<dim> (prm)),
```
gassmoeller: If you move the functions as I mentioned above, you can change this to the same structure as for the NewtonHandler or MeltHandler.

build_9.0/wb3.json
MFraters: >We will probably need a new directory in `data/world-builder/` or `data/initial-temperature/world-builder` for these files.

I think it will be better to put it in the end into a cookbook, since it is not really data, just an example input.

>Also: Is it possible to add comments to these files?

Unfortunately for us, they removed support for comments from JSON because people where apparently abusing it as macro's, and some other reasons (see https://groups.yahoo.com/neo/groups/json/conversations/topics/156 ).  But what can be done is just adding a field `"comment":"and put the comment here",`. I could them exempt the comment field from needing declaration, or ignore the field when reading stuff in. 

An other thing I could manually parse the file in advance and remove comment lines, but then we would diverge from the JSON standard, making implementation into other codes slightly more difficult.

include/aspect/world_builder/continental_plate.h
MFraters: Yes, this had to do with the old structure, I still have to rethink the approach. I only added this file for now to show that the plugin structure with the declaration works.

build_9.0/wb3.json
bangerth: Yes, manually parsing is not an option and extending the file format with comments is also not an option. I do like the idea of a "comment" field.

MFraters:
Thanks for the comments! I have addressed them and implemented the manager temperature and composition function. I still have to do the parsing, and then I think I can get this out of the WIP phase.

build_9.0/wb3.json
gassmoeller: Ok, that is good enough for me. I just want to have the option to put in stuff like "this is model X with data taken from publication Y" or something of that sort.

MFraters:
The code is now ready for review. I will add a test tomorrow.

include/aspect/world_builder/interface.h
```diff
@@ -0,0 +1,316 @@
...
+         * Exception
+         */
+        DeclException2 (ExcValueDoesNotMatchPattern,
+                        std::string, std::string,
+                        << "The string <" << arg1
+                        << "> does not match the given pattern <" << arg2 << ">.");
+    };
+
+
+    using namespace dealii;
```
tjhei: redundant, see above.

include/aspect/world_builder/interface.h
```diff
@@ -0,0 +1,316 @@
...
+   */
+  namespace WorldBuilder
+  {
+    using namespace dealii;
+
+    class WorldBuilderParameterHandler
+    {
+      public:
+        WorldBuilderParameterHandler(const char path_separator = '.');
+        void set_tree(boost::property_tree::ptree &tree);
```
tjhei: please document these functions. I have no idea what they do.

include/aspect/world_builder/interface.h
```diff
@@ -0,0 +1,316 @@
...
+                        << "The string <" << arg1
+                        << "> does not match the given pattern <" << arg2 << ">.");
+    };
+
+
+    using namespace dealii;
+
+    /**
+     * Base class for classes that describe particular initial conditions.
+     * All plugins, which describe a plate tectonic features in terms of
```
tjhei: grammar, remove "a"?

include/aspect/world_builder/interface.h
```diff
@@ -0,0 +1,316 @@
...
+         * parameters. Consequently, derived classes do not have to overload
+         * this function if they do not take any runtime parameters.
+         */
+        virtual
+        void
+        parse_parameters (ParameterHandler &prm);
+
+
+
+      private:
```
tjhei: remove

include/aspect/world_builder/interface.h
```diff
@@ -0,0 +1,316 @@
...
+        /**
+         * Declare the runtime parameters of the registered world builder
+         * models.
+         *
+         * @ingroup WorldBuilder
+         */
+        static void
+        declare_parameters (ParameterHandler &prm);
+
+        ParameterHandler wb_prm;
```
tjhei: are these supposed to be public?

include/aspect/world_builder/interface.h
```diff
@@ -0,0 +1,316 @@
...
+        double top_crust_tickness;
+        unsigned int top_crust_composition;
+        std::vector<double> rotation_point;
+        double surface_temperature;
+        unsigned int number_of_objects;
+
+        double potential_mantle_temperature, thermal_expansion_coefficient_alfa, specific_heat_Cp;
+
+        //const std_cxx11::shared_ptr<GravityModel::Interface<dim> > gravity_model;
+        //const std_cxx11::shared_ptr<GeometryModel::Interface<dim> > geometry_model;
```
tjhei: remove

source/utilities.cc
```diff
@@ -2468,6 +2470,49 @@ namespace aspect
       return array;
     }
 
+    /**
+     * todo
```
tjhei: remove the comment here ;-)

source/utilities.cc
```diff
@@ -2468,6 +2470,49 @@ namespace aspect
...
+     */
+    template <int dim>
+    std::vector<Point<dim> >
+    vector_double_to_point(const std::vector<std::vector<double> > &s)
```
tjhei: these functions would make a good separate PR, don't you think?

source/utilities.cc
```diff
@@ -2573,7 +2618,7 @@ namespace aspect
     }
 
     template <int dim>
-    std_cxx11::array<double,dim> &NaturalCoordinate<dim>::get_coordinates()
+    const std_cxx11::array<double,dim> NaturalCoordinate<dim>::get_coordinates() const
```
tjhei: also good as a separate PR!

source/world_builder/interface.cc
```diff
@@ -0,0 +1,1082 @@
...
+      mangle (const std::string &s)
+      {
+        std::string u;
+
+        // reserve the minimum number of characters we will need. it may
+        // be more but this is the least we can do
+        u.reserve (s.size());
+
+        // see if the name is special and if so mangle the whole thing
+        const bool mangle_whole_string = (s == "value");
```
tjhei: I don't understand this, if s==value, couldn't you just return the mangled name of "value"?

source/world_builder/interface.cc
```diff
@@ -0,0 +1,1082 @@
...
+          else
+            {
+              Assert(i + 2 < s.size(),
+                     ExcMessage("Trying to demangle an invalid string."));
+
+              unsigned char c = 0;
+              switch (s[i + 1])
+                {
+                  case '0':
+                    c = 0 * 16;
```
tjhei: you can use something like ``(s[i+1]-'0')`` to get the value if it is a digit.

source/world_builder/interface.cc
```diff
@@ -0,0 +1,1082 @@
...
+                  case 'e':
+                    c = 14 * 16;
+                    break;
+                  case 'f':
+                    c = 15 * 16;
+                    break;
+                  default:
+                    Assert(false, ExcInternalError());
+                }
+              switch (s[i + 2])
```
tjhei: you can use a for loop to handle both chars with a single logic (and multiply by 16 in between)

source/world_builder/interface.cc
```diff
@@ -0,0 +1,1082 @@
...
+    Manager<dim>::initial_temperature(const dealii::Point<dim> &original_position) const
+    {
+      const GeometryModel::Interface<dim> *geometry_model= &this->get_geometry_model();
+
+      Point<dim> external_position;
+      Point<3> internal_position;
+      double temperature;
+      if (dim==2)
+        {
+          //Todo, convert from a 2d point of a cross section to a 3d point for the world generator
```
tjhei: put an assert in here.

Issue #2123, opened 2018-03-09
Diffusion Formulation Bug in Visco plastic material model


sibiaoliu:
Hi there,

I am running the intraplate compression model with visco-plastic material model.  Since I add both the diffusion and dislocation rheology, I found two questions.
1. The formulation of the diffusion creep viscosity is wrong: 
In visco_plastic.cc, Line 198, `// viscosity = 0.5 * A^(-1/n) * edot_ii^((1-n)/n) * d^(m/n) * exp((E +...`,   it should be **d^(-m/n)** if the grain size `d` is 1 mm (i.e., the default value in the code).

So, in Line 205, the equation should be `double viscosity_diffusion =...std::pow(grain_size, -grain_size_exponents_diffusion[j]);`. Another way is to `set Grain size = 1000` in the Input file. This is what I am doing now. It's a tiny error about missing the minus sign, but will make the viscosity value wrong :smile: 

2. As I am focusing on the lithosphere-upper mantle scale simulation, it is better to use the composite creep viscosity in each individual composition phase. It is not necessary, however, to add the diffusion creep to the crust. I am wondering that is it possible to set the diffusion parameters for only the lithospheric mantle and upper mantle?  For example, can I only set 3 values of diffusion parameters rather than 6 values if I have 5 compositional fields?  How about  `set Prefactors for diffusion creep =  1.e3, 1.e3, 1.e3`

By the way, just a remind: the current version of aspect seems to not work on the XCode version >9.0 on Mac OS, due to the missing SDK file "MacOSX10.12.sdk".

Thank you very much!

naliboff:
@sibiaoliu - Thanks for the questions! 

First, a quick note. The equations you refer to in `visco_plastic` model have been updated slightly since the last release. Are you using version 1.5? For reference, the current version can be seen here: [visco_plastic](https://github.com/geodynamics/aspect/blob/master/source/material_model/visco_plastic.cc) 

Question 1 - I am a bit confused as to why the equation would need to be altered to have `A^(-1/m)`. The current formulation looks correct. If one were to define the strain rate for diffusion creep, rather than viscosity, the equation would contain `A^(-1/m)`.

Question 2 - Yes, the current composition option is not particularly helpful if one does not want to use a composite viscosity in all the material models. A quick solution is to change the composite viscosity formulation to:
`std::pow( 1.0/viscosity_diffusion + 1.0/viscosity_dislocation, -1.0 )`
[...]

PR #2124, opened 2018-03-10
Update documentation of viscous prefactors in visco_plastic

naliboff:
Addresses #2123. 
@sibiaoliu - Does this clarify the prefactor term? If more documentation is needed, let me know!

### DIFF
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -745,7 +745,7 @@ namespace aspect
                              "List of viscosity prefactors, $A$, for background material and compositional fields, "
                              "for a total of N+1 values, where N is the number of compositional fields. "
                              "If only one value is given, then all use the same value. "
-                             "Units: $Pa^{-1} m^{1/m_{diffusion}} s^{-1}$");
+                             "Units: $Pa^{-1} m^{m_\\text{diffusion}} s^{-1}$");
           prm.declare_entry ("Stress exponents for diffusion creep", "1",
                              Patterns::List(Patterns::Double(0)),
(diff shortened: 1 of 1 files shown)


bangerth:
If there is disagreement in the literature on how exactly to define an exponent (with positive or negative signs), would it make sense to point this out in the documentation?

naliboff:
Yes, definitely worth pointing this out in the documentation. I've added a brief comment indicating that the grain size exponent should always be defined as positive and to carefully check how viscous flow law terms are defined in previous studies.

naliboff:
As other material models also use similar flow laws, perhaps it would be worth noting in the manual?

bangerth:
Thanks for the extra text!

You're right that in general, we ought to document these sorts of deviations between formulations used in the literature. If you come across a place that could use a similar sentence, feel free to add it!

jdannberg:
Thanks for adding this, @naliboff! 
Just one small thing: I noticed the spaces at the end of the documentation text are missing (so in the manual pdf, this will look like for example "useslightly" instead of "use slightly"). If you have time, maybe you could fix this? Otherwise I'll make another pull request later today. 

And I guess at some point we should also make similar changes to the diffusion dislocation model (as it uses the same formulation).

naliboff:
@jdannberg - Thanks for catching thtat! I'll add the spaces and add a few words to the diffusion dislocation model as well. Should I check the grain-size dependent material model as well?

jdannberg:
It seems to be correct in the `grain_size` material model (which also makes sense, because it does not use code from the diffusion dislocation model).

naliboff:
See #2128 for spacing fix in `visco_plastic` and additional documentation and formatting in `diffusion_dislocation`.

PR #2125, opened 2018-03-10
Only rebuild Stokes preconditioner if necessary

gassmoeller:
This came out of a discussion with @bangerth here at SIAM PP. Most of the time we do not need to rebuild the AMG preconditioner completely if the mesh has not changed. We just need to update it for the changed matrix values, which is much cheaper. I thought I give this a try (it is a small change), and it seems to work at least for a very small test model I ran on my laptop. I open this mostly to see if all tests work, and also to ask for comments from @tjhei and @bangerth. 
There are two motivations why I think this improves performance significantly:
1. This is a typical timing output of one of my large scale mantle convection models:
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |  4.23e+04s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble Stokes system          |      2878 |  7.17e+03s |        17% |
| Assemble composition system     |      5756 |  5.25e+03s |        12% |
| Assemble temperature system     |      2878 |  3.26e+03s |       7.7% |
| Build Stokes preconditioner     |      2878 |  8.19e+03s |        19% |
| Build composition preconditioner|      5756 |       610s |       1.4% |
| Build temperature preconditioner|      2878 |       314s |      0.74% |
| Solve Stokes system             |      2878 |   6.4e+03s |        15% |
| Solve composition system        |      5756 |       997s |       2.4% |
| Solve temperature system        |      2878 |       506s |       1.2% |
| Create snapshot                 |        23 |       530s |       1.3% |
| Initialization                  |         1 |      1.09s |         0% |
| Postprocessing                  |      1705 |  1.54e+03s |       3.6% |
| Refine mesh structure, part 1   |       569 |  5.04e+03s |        12% |
| Refine mesh structure, part 2   |       569 |       192s |      0.45% |
| Setup dof systems               |       570 |  1.71e+03s |         4% |
+---------------------------------+-----------+------------+------------+
```
Building the Stokes preconditioner is actually more expensive than solving the system. In this model I do mesh refinement every 3rd timestep, so reducing the time spent building the preconditioner by a factor of 2-3 (depending on how expensive the reinitialization is compared to a rebuild), would save me 10-15% of the runtime (10-15% of 40,000 CPU hours for this model alone).

2. Building the Stokes preconditioner can become the strong scaling limit for our models. Not for all of my scaling tests, but the worst case of one of my scaling models on Lonestar 5 (12,288 cores for 130 million dofs) looks like this:
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |       286s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble Stokes system          |         1 |     0.918s |      0.32% |
| Assemble composition system     |         1 |     0.365s |      0.13% |
| Assemble temperature system     |         1 |     0.498s |      0.17% |
| Build Stokes preconditioner     |         1 |       214s |        75% |
| Build composition preconditioner|         1 |    0.0516s |     0.018% |
| Build temperature preconditioner|         1 |    0.0456s |     0.016% |
| Solve Stokes system             |         1 |      49.7s |        17% |
| Solve composition system        |         1 |     0.167s |     0.058% |
| Solve temperature system        |         1 |     0.272s |     0.095% |
| Initialization                  |         2 |      15.3s |       5.3% |
| Postprocessing                  |         1 |    0.0232s |    0.0081% |
| Setup dof systems               |         1 |      4.08s |       1.4% |
+---------------------------------+-----------+------------+------------+
``` 
Reducing the numer of times we build the preconditioner does of course not avoid this scaling limit, but at least it makes it less significant if we call the function less often (for models that are not quite as dramatically bad). 

Opinions? Do I miss something, or do we get 10-15% speedup essentially for free?

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1679,6 +1679,7 @@ namespace aspect
       bool                                                      assemble_newton_stokes_matrix;
       bool                                                      assemble_newton_stokes_system;
       bool                                                      rebuild_stokes_preconditioner;
+      bool                                                      reinitialize_stokes_preconditioner;
 
       /**
        * @}
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -391,8 +391,18 @@ namespace aspect
   void
   Simulator<dim>::build_stokes_preconditioner ()
   {
-    if (rebuild_stokes_preconditioner == false)
+    if (rebuild_stokes_preconditioner == false
+        && reinitialize_stokes_preconditioner == false)
       return;
+    else if (rebuild_stokes_preconditioner == false
+             && reinitialize_stokes_preconditioner == true)
+      {
+        TimerOutput::Scope timer (computing_timer, "   Reinitialize Stokes preconditioner");
+        Amg_preconditioner->reinit();
+        Mp_preconditioner->initialize (system_preconditioner_matrix.block(1,1));
+        reinitialize_stokes_preconditioner = false;
+        return;
+      }
 
     if (parameters.use_direct_stokes_solver)
       return;
@@ -456,6 +466,7 @@ namespace aspect
                                       Amg_data);
 
     rebuild_stokes_preconditioner = false;
+    reinitialize_stokes_preconditioner = false;
 
     pcout << std::endl;
   }
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -394,15 +394,6 @@ namespace aspect
     if (rebuild_stokes_preconditioner == false
         && reinitialize_stokes_preconditioner == false)
       return;
-    else if (rebuild_stokes_preconditioner == false
-             && reinitialize_stokes_preconditioner == true)
-      {
-        TimerOutput::Scope timer (computing_timer, "   Reinitialize Stokes preconditioner");
-        Amg_preconditioner->reinit();
-        Mp_preconditioner->initialize (system_preconditioner_matrix.block(1,1));
-        reinitialize_stokes_preconditioner = false;
-        return;
-      }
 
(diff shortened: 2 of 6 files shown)


gassmoeller:
Hm, I might have been too optimistic there. It seems the reinitialized (instead of rebuild) preconditioner is significantly worse in some models. Could this be related to the fact that I can not update the constant modes for the updated preconditioner matrix? Or do I completely misunderstand the point here?

gassmoeller:
I have a theory for what is happening, but I would still be interested in opinions. Does the recomputed preconditioner take into account the updated matrix entries for smoothing and coarsening, but does not regenerate the coarsening pattern/hierarchy? If that is the case if viscosities have changed the preconditioner uses the correct preconditioner matrix values, but coarsens the wrong entries?
Somehow it seems unfair that we should not be able to exploit the knowledge that the matrix has only changed slightly since the last timestep / nonlinear iteration :-(

gassmoeller:
I am closing this for now since it does not seem to work (we have far too many stale PRs already, and this seems like a good candidate to become the next). If anyone has insights I am still interested, but it was a small side project anyway.

bangerth:
Hm, it took me a couple days to look at this, but I don't think I understand how this works. Can you post an example of where things become more expensive? I would have liked to see something along the lines of "let's rebuild the preconditioner every third time only, or if the mesh has changed". I don't think I understand how you make the decision here when to rebuild things. (Not your fault: the whole logic about when to rebuild the preconditioner is convoluted and probably ought to be rewritten.)

PR #2127, opened 2018-03-11
Optimize seismic ic

### LINKED ISSUE #962: Very slow S40RTS initial conditions
This initial condition plugin currently recomputes the spherical harmonics in every quadrature point taking 30 minutes or longer. Fix this.

gassmoeller:
Finally fix #962. Something I wanted to do for a long time (waiting time at an airport is so useful :smile:). Both plugins show a ~15X speedup with the changes in the first commit. They are still a big part of an instantaneous model (~50% for the tests, compared to 90% before), but at least they are now comparable to a Stokes solve, which should be good enough. The second commit adjusts the code style to our usual convention and uses more correct types for many indices.

@jaustermann or @Shangxin-Liu maybe one of you could take look and do a review of this PR, since you wrote the initial code? Did I mess anything up?

### DIFF
--- a/include/aspect/initial_temperature/S40RTS_perturbation.h
+++ b/include/aspect/initial_temperature/S40RTS_perturbation.h
@@ -130,7 +130,7 @@ namespace aspect
          * The maximum order the users specify, which is only valid when
          * "lower_max_order" is set to true.
          */
-        int max_order;
+        unsigned int max_order;
 
         /**
          * This parameter gives the reference temperature, which will be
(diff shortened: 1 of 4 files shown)


jaustermann:
I'm in the field until Tuesday but am happy to review this as soon as I'm back.

bangerth:
Ah, nice job -- all it took was hoisting some constants out of a loop :-)

Shangxin-Liu:
Nice speed-up insight @gassmoeller, it looks fine to me. I may need to also update ASPECT on our cluster:)

gassmoeller:
Ok, great. @jaustermann it would be great if you could take a look when you are back and have some time. I would like to merge this end of the week / first half of next week to make sure it makes it into the release.

jaustermann:
I went through all the changes and they look good to me. The S40 and Savani tests are also good checks that the content hasn't changed. Awesome speedup, thanks Rene! I can merge this if it's okay - maybe my first merge?!

gassmoeller:
Sure, go ahead :smile:

Issue #59, opened 2014-05-22
Exploit more parallelism


bangerth:
Currently, we run Aspect with one thread per MPI process. What we should really do is to run it with n_cpus/n_mpi_processes_on_this_machine threads, to make optimal use of it.

gassmoeller:
Is this still a relevant question? Since it is simple to activate I made a few runs on my laptop with a very small test model, and below are some results that show that replacing MPI processes with threads can indeed speed up things (on my laptop that is). I guess the two important question I would have about the topic are:
1. Which parts of ASPECT will benefit from threads, and more importantly which will not? Might those matter if we for example run only 1 MPI process per node on a cluster? It seems postprocessing does not benefit from threading?
2. Can we push the strong scaling limit a bit higher by using less MPI processes per node, because we create less MPI traffic and the individual work sizes per process are larger?

Clearly this needs more investigation on larger clusters, but here are the results from my laptop in Optimized mode for tests/simple_compressible.prm at a fixed frequency (no turbo):

1 process / 8 threads:
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      0.19s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble Stokes system          |         2 |   0.00918s |       4.8% |
| Assemble temperature system     |         2 |    0.0277s |        15% |
| Build Stokes preconditioner     |         1 |    0.0147s |       7.7% |
| Build temperature preconditioner|         2 |   0.00322s |       1.7% |
| Solve Stokes system             |         2 |    0.0195s |        10% |
| Solve temperature system        |         2 |   0.00166s |      0.87% |
| Initialization                  |         1 |      0.04s |        21% |
| Postprocessing                  |         2 |    0.0276s |        15% |
| Setup dof systems               |         1 |   0.00688s |       3.6% |
| Setup initial conditions        |         1 |    0.0138s |       7.3% |
+---------------------------------+-----------+------------+------------+

```

2 processes / 4 threads each:
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |     0.179s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble Stokes system          |         2 |   0.00925s |       5.2% |
| Assemble temperature system     |         2 |    0.0207s |        12% |
[...]

PR #2130, opened 2018-03-14
Threads

gassmoeller:
Implements #59. I decided to make it a cmake variable instead of a input parameter, because we need the value right at the beginning of main.cc:main() before even initializing anything, and we do not know the name of the parameter file at that time. Unfortunately now we need to recompile if we want to change the setting, but at least it is user-controllable. Is this an acceptable solution?

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -3887,16 +3887,16 @@ with for example two threads per process will offset some of the performance
 loss you will see in these situations.
 
 Multithreading is controlled by setting the command line parameter \texttt{-j}
-or \texttt{--threads}. If the parameter is not set \aspect{} will create
+or \texttt{-{}-threads}. If the parameter is not set, \aspect{} will create
 exactly one thread per MPI process, i.e. multithreading is disabled.  Appending
-it will automatically spawn several threads per MPI process (a typical example
-is ``2''). Note that the internally used TBB library will determine the number
-of threads as a reasonable number, i.e. if you start with 2~MPI processes on a
+the parameter allows \aspect{} to spawn several threads per MPI process. Note
+that the internally used TBB library will determine the number of threads based
+on the number of available cores, i.e., if you start 2~MPI processes on a
 quadcore machine with hyperthreading (8 logical cores), \aspect{} will spawn 4
 threads on each MPI process. Also note that there is no guarantee that the
 final number of threads will exactly match the number of available logical
 cores if you start with a number of processes that is not a divisor of your
-logical threads (e.g. 3 MPI processes for 8 logical cores).
+logical cores (e.g. 3 MPI processes for 8 logical cores).
 
 \subsection{Input parameter files}
 \label{sec:parameters-overview}
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -3886,18 +3886,17 @@ cores per node on a cluster to increase the available memory per core.  Running
 with for example two threads per process will offset some of the performance
 loss you will see in these situations.
 
-Multithreading is controlled by setting the environment variable
-\texttt{DEAL\_II\_NUM\_THREADS}. If the variable is not set \aspect{} will
-create exactly one thread per MPI process, i.e. multithreading is disabled.
-Setting it to any other number will spawn as many threads per MPI process as
-you select (a typical example is ``2''). Note that \aspect{} will limit the
-number of threads to a reasonable number, i.e. if you request 8 threads per
-process on an quadcore machine with hyperthreading (8 logical cores), but start
-with 2~MPI processes \aspect{} will only spawn 4 threads on each process. Also
-note that there is no guarantee that the final number of threads will exactly
-match the number of available logical cores if you start with a number of
-processes that is not a divisor of your logical threads (e.g. 3 MPI processes
-for 8 logical cores).
+Multithreading is controlled by setting the command line parameter \texttt{-j}
+or \texttt{--threads}. If the parameter is not set \aspect{} will create
+exactly one thread per MPI process, i.e. multithreading is disabled.  Appending
+it will automatically spawn several threads per MPI process (a typical example
+is ``2''). Note that the internally used TBB library will determine the number
+of threads as a reasonable number, i.e. if you start with 2~MPI processes on a
+quadcore machine with hyperthreading (8 logical cores), \aspect{} will spawn 4
+threads on each MPI process. Also note that there is no guarantee that the
+final number of threads will exactly match the number of available logical
+cores if you start with a number of processes that is not a divisor of your
+logical threads (e.g. 3 MPI processes for 8 logical cores).
 
 \subsection{Input parameter files}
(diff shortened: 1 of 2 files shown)


bangerth:
What if you made it a run time parameter, say -j N? You'd then specify it when calling the executable if you want anything other than the default.

tjhei:
1. anything except ``1`` will break PETSc (not that this has been tested recently, so I assume it is probably broken in other places) so keeping 1 as the default is a good idea.
2. I don't like that it is a cmake parameter (does that mean you have to recompile the whole code-base if you change it?)

gassmoeller:
Currently we need to recompile everything to change this, yes. Unfortunately everything else is more complicated, because when parsing our runtime parameters we already need information about the MPI communicator (e.g. -h only outputs on process 0) so MPI_InitFinalize needs to be called before parsing runtime parameters. I could delay executing these commands (and only set a flag when parsing), but that makes the whole parsing code more complicated. Do you have an idea how to work around this?

If this breaks PETSc I should probably add a check to CMakeLists that not both are active at the same time.

tjhei:
> Currently we need to recompile everything to change this, yes.

So this means it is more annoying to switch than editing main.cc. Hrm.

> Unfortunately everything else is more complicated

Hang on: we have an environment variable ``DEAL_II_NUM_THREADS`` that is used, see http://www.dealii.org/developer/doxygen/deal.II/classMultithreadInfo.html#ae253b8ff679c4119f9c5bff4439fe606

Wouldn't that be enough? This means we would need to set the value to invalid_unsigned_int for the command line value to work.

gassmoeller:
Hmm, I tried exporting DEAL_II_NUM_THREADS=1 and that indeed limits the number of threads to 1. But can we set this environment variable by default to 1? I would like to keep the current behavior (threading disabled until somebody explicitly enables it). Mostly to follow the principle of least surprise to the user. Or do you think we should just enable this by default? I could also ask around on the mailing list what people think about the feature.

tjhei:
> But can we set this environment variable by default to 1?

Not easily. Either we enable by default (I don't think this will be a problem) or safer:
Can you test early in main.cc:
```
const char *penv = getenv ("DEAL_II_NUM_THREADS");
unsigned int max_threads = (penv==nullptr) ? 1 : numbers::invalid_unsigned_int;

MPI::InitFinalize(...., max_threads);
```

gassmoeller:
Yes that seems to work. I like this solution, it is easy for people to set this if they want to, and everything stays as usual if you do nothing.

bangerth:
I'd still like to advocate for a command line option `-j` in the same way as it exists for `make` and `ninja`. (Though I'd also be happy if `-j` has no numeric argument and simply means "use as many threads as reasonable -- possibly limited by `DEAL_II_NUM_THREADS`. In other words, passing `-j` would set `max_threads` to `numbers::invalid_unsigned_int`, whereas the default is one.)

You can parse command line options before you initialize the MPI system.

gassmoeller:
I reorganized main() to parse the command line parameters before MPI initialization and made the parameter a command line option. Looks nicer and more intuitive now. I also removed the print_version() function as all of that information is in the header now and removed some duplication by introducing a new template function. Let's see what the tester says. Anything else? (I will follow up with a PR that further cleans main(), but that is unrelated).

bangerth:
One of the test fails by producing different output. I haven't looked at the code yet, but suspect you have something like
```
  if (dim>=2 && dim <=3)
    ...do something
  else
     Assert (false, ExcMessage(...));
```
To restore the previous error (and help the user a bit), just replace the assertion by
```
  Assert (dim >= 2 && dim <= 3, ...)
```
Since we're in the `else` block, the condition is of course always false.

doc/manual/manual.tex
```diff
@@ -3873,6 +3873,30 @@ \subsubsection{Regularizing models with large coefficient variation}
...
+speed up the assembly of the system matrices, e.g. by around 10-15\% if you
+utilize unused logical cores, or nearly linearly if you use otherwise
+unused physical cores. This can also reduce the performance cost if you are
+memory limited and need to run your model on less than the available number of
+cores per node on a cluster to increase the available memory per core.  Running
+with for example two threads per process will offset some of the performance
+loss you will see in these situations.
+
+Multithreading is controlled by setting the command line parameter \texttt{-j}
+or \texttt{--threads}. If the parameter is not set \aspect{} will create
```
bangerth: comma after "set"

doc/manual/manual.tex
```diff
@@ -3873,6 +3873,30 @@ \subsubsection{Regularizing models with large coefficient variation}
...
+exactly one thread per MPI process, i.e. multithreading is disabled.  Appending
+it will automatically spawn several threads per MPI process (a typical example
```
bangerth: what is "it" here?

doc/manual/manual.tex
```diff
@@ -3873,6 +3873,30 @@ \subsubsection{Regularizing models with large coefficient variation}
...
+is ``2''). Note that the internally used TBB library will determine the number
+of threads as a reasonable number, i.e. if you start with 2~MPI processes on a
```
bangerth: comma after "i.e."

source/main.cc
```diff
@@ -368,9 +368,11 @@ void print_help()
             << std::endl;
   std::cout << "    optional arguments [args]:"
             << std::endl
-            << "       --version              (for information about library versions)"
+            << "       -v, --version          (for information about library versions)"
             << std::endl
-            << "       --help                 (for this usage help)"
+            << "       -h, --help             (for this usage help)"
+            << std::endl
+            << "       -j, --threads          (to use multi-threading)"
```
bangerth: keep `-h` at either the top of the bottom of the list.

source/main.cc
```diff
@@ -641,65 +652,21 @@ int main (int argc, char *argv[])
...
-
+            run_simulator<3>(input_as_string,output_xml,output_plugin_graph);
             break;
           }
-
           default:
-            AssertThrow((dim >= 2) && (dim <= 3),
+            AssertThrow(false,
                         ExcMessage ("ASPECT can only be run in 2d and 3d but a "
                                     "different space dimension is given in the parameter file."));
```
bangerth: This is where the difference in output comes from. I'd say to just revert to the previous condition -- I know it's duplicative, but it's also instructive.

source/main.cc
```diff
@@ -641,65 +652,21 @@ int main (int argc, char *argv[])
...
-                if (i_am_proc_0)
-                  flow_problem.write_plugin_graph (std::cout);
-              }
-            else
-              {
-                aspect::Simulator<3> flow_problem(MPI_COMM_WORLD, prm);
-                flow_problem.run();
-              }
-
+            run_simulator<3>(input_as_string,output_xml,output_plugin_graph);
```
bangerth: nice simplification of code repetition

gassmoeller:
Done :smile:.

bangerth:

PR #2131, opened 2018-03-17
fix freezing in melt simple material model and improve documentation

jdannberg:
This was really confusing before (and I noticed this when @joeschools showed me one of his model setups). What used to happen was that there was a freezing rate parameter, but even if that was set to zero, melt could freeze, but only under certain circumstances (if it had not moved away from its source region). This update makes this behaviour more consistent (melt does not freeze if the freezing rate is set to zero; and when the freezing rate is larger than zero, it determines the rate of all freezing) and clearly documents how the amount of freezing is computed. I also improved some variable names.


bangerth:
I think @tjhei should probably look at this.

jdannberg:
I have updated the tests, and I've also tested the code with a simple setup for melting and freezing at a mid-ocean ridge that I'd like to add as a cookbook later (it only works with our new melt solver, so we have to wait until that is merged). 

I believe anyone could have a look at this, not necessarily just @tjhei, as it doesn't have to do anything with the melt solver, it is just a fix for the parametrization of melting/freezing (so it goes more in the direction of thermodynamics).

source/material_model/melt_simple.cc
```diff
@@ -258,7 +258,7 @@ namespace aspect
           out.densities[i] = (reference_rho_s + delta_rho)
                              * temperature_dependence * std::exp(compressibility * (in.pressure[i] - this->get_surface_pressure()));
 
-          if (this->include_melt_transport())
+          if (this->include_melt_transport() && in.strain_rate.size())
```
gassmoeller: > 0

source/material_model/melt_simple.cc
```diff
@@ -271,44 +271,71 @@ namespace aspect
...
+                                                                          * (1.0 - old_porosity[i]) / (1.0 - maximum_melt_fractions[i]),0.0);
+                    double freezing_amount = std::max(porosity_change_wrt_melt_fraction, porosity_change_wrt_depletion);
+
+                    if (eq_melt_fraction == 0.0)
+                      freezing_amount = - old_porosity[i];
+
+                    porosity_change += freezing_amount;
+
+                    // Adapt time scale of freezing with respect to melting.
+                    if (porosity_change < 0 && this->get_parameters().use_operator_splitting)
```
gassmoeller: Remove the non-operator split versions.

jdannberg:
Okay, I changed the material model so that it can only be used with operator splitting now and I could remove some lines of code. I adapted the tests accordingly.

doc/modules/changes/20180316_jdannberg
```diff
@@ -0,0 +1,9 @@
+Overhaul of the melt simple model.
```
gassmoeller: melt simple material model

source/material_model/melt_simple.cc
```diff
@@ -258,97 +236,110 @@ namespace aspect
...
+                    double freezing_amount = std::max(porosity_change_wrt_melt_fraction, porosity_change_wrt_depletion);
+
+                    if (eq_melt_fraction == 0.0)
+                      freezing_amount = - old_porosity;
+
+                    porosity_change += freezing_amount;
+
+                    // Adapt time scale of freezing with respect to melting.
+                    if (porosity_change < 0 )
+                      porosity_change *= freezing_rate * melting_time_scale;
```
gassmoeller: Can you add a comment here why you multiply by both? "E.g. we divide by the melting_time_scale later on in any case, but this is not necessary for this case, thus ...". That would make it easier to understand.

source/material_model/melt_simple.cc
```diff
@@ -258,97 +236,110 @@ namespace aspect
...
                 {
                   out.reaction_terms[i][c] = 0.0;
 
-                  if (this->get_parameters().use_operator_splitting && reaction_rate_out != NULL)
+                  if (reaction_rate_out != NULL)
                     reaction_rate_out->reaction_rates[i][c] = 0.0;
                 }
             }
 
-          out.entropy_derivative_pressure[i]    = entropy_change (in.temperature[i], this->get_adiabatic_conditions().pressure(in.position[i]), maximum_melt_fractions[i], NonlinearDependence::pressure);
```
gassmoeller: What happened to these lines? Are they not important for latent heat?

source/material_model/melt_simple.cc
```diff
@@ -531,14 +522,37 @@ namespace aspect
...
+                             "\n\n"
+                             "Note that melt does not freeze unless the 'Freezing rate' parameter is set "
+                             "to a value larger than 0.");
           prm.declare_entry ("Freezing rate", "0.0",
                              Patterns::Double (0),
-                             "Freezing rate of melt when in subsolidus regions."
-                             "Units: $1/yr$.");
+                             "Freezing rate of melt when in subsolidus regions. "
+                             "If this parameter is set to a number larger than 0.0, it specifies the "
+                             "fraction of melt that will freeze per year, as soon as the porosity "
```
gassmoeller: per year or per second

source/material_model/melt_simple.cc
```diff
@@ -531,14 +522,37 @@ namespace aspect
...
+                             "``Melting time scale for operator splitting''. The product of the "
+                             "``Freezing rate'' and the ``Melting time scale for operator splitting'' "
+                             "defines how fast freezing occurs with respect to melting (if the "
+                             "product is 0.5, melting will occur twice as fast as freezing). "
+                             "Units: 1/yr or 1/s, depending on the ``Use years "
+                             "in output instead of seconds'' parameter.");
           prm.declare_entry ("Melting time scale for operator splitting", "1e3",
                              Patterns::Double (0),
-                             "In case the operator splitting scheme is used, the porosity field can not "
+                             "Because the operator splitting scheme is used, the porosity field can not "
```
gassmoeller: Nice rewrite :smile:

source/material_model/melt_simple.cc
```diff
@@ -742,8 +759,7 @@ namespace aspect
     void
     MeltSimple<dim>::create_additional_named_outputs (MaterialModel::MaterialModelOutputs<dim> &out) const
     {
-      if (this->get_parameters().use_operator_splitting
-          && out.template get_additional_output<ReactionRateOutputs<dim> >() == NULL)
+      if (this->get_parameters().use_operator_splitting && out.template get_additional_output<ReactionRateOutputs<dim> >() == NULL)
```
gassmoeller: do you still need the first condition?

tests/shear_heating_with_melt.prm
```diff
@@ -187,12 +200,12 @@ subsection Melt settings
 end
 
 subsection Heating model
-  set List of model names                     = adiabatic heating, latent heat, shear heating, adiabatic heating of melt, shear heating with melt
+  set List of model names                     = adiabatic heating, latent heat melt, shear heating, adiabatic heating of melt, shear heating with melt
 end
 
 subsection Postprocess
 
-  set List of postprocessors = composition statistics,velocity statistics, mass flux statistics, heating statistics, temperature statistics
```
gassmoeller: the visualization was likely added by accident?

source/material_model/melt_simple.cc
jdannberg: I moved them further up into the individual branches (with and without melt transport). This was convenient because they need the depletion field, which we do not necessarily have in the case without melt, and so we compute it only in the part of the if statement that is executed with melt transport (otherwise, 0 is used, just as before these changes).

source/material_model/melt_simple.cc
jdannberg: Yes. I first removed it, but it caused some kind of problem. I don't remember what exactly it was, but I think it might have been related to the point that we still allow this material to be used without operator splitting if melt transport is switched off (in this case, there are no reactions, so it doesn't matter), and I got some error message in one of the tests that my material model asks for reaction outputs, but they are not filled, or something like that.

tests/shear_heating_with_melt.prm
jdannberg: I can remove it again, but if anyone needs to look at this test again in the future because something has changed, they will need the visualization output again, because there's no other way to really see what the test does and if its behaviour is still correct. So unless you feel strongly about removing it, I think it would be kind of useful to just leave it switched on.

jdannberg:
Alright, I think I addressed all of your comments.

gassmoeller:
Thanks, easier to understand now.

PR #1823, opened 2017-06-23
[WIP] To fix possible bug for problems with periodic boundary using DG and AMR 

yinghe616:
Issue #1822 gives an example showing that DG method can return a wrong solution for the periodic boundary problem on an AMR grid. After further investigating the code,   it does show a potential bug in the  flux construction. For the current implementation, the flux calculation will skip the case when the periodic neighbor cell is finer. One reason is that the periodic face in a coarsen cell is always on the boundary, therefore face->has_children() always return false. So face->has_children() can not be a criterion to indicate the current cell is coarser than the periodic neighbor. 
Todo: need to carefully go over the surfaces in the periodic case.

PR #2132, opened 2018-03-19
Fix dg periodic amr

gassmoeller:
Supersedes #1823 including all comments. I also included a test and an assert that prevents the parallel case (was not able to quickly find how to fix it).


bangerth:
Thanks for taking care of this old PR and making sure @yinghe616 's contribution isn't lost!

PR #2136, opened 2018-03-22
Speed up depth average postprocessor

gassmoeller:
This began as a small optimization project and ended up as nearly complete rewrite of the lateral averaging functionality. All external interfaces are kept, and all functionality should work as before, but by reducing duplication the depth average postprocessor is now nearly ten times faster than before. Instead of n loops over all cells for n properties, lateral averaging now performs 1 loop for n properties (by providing n functors that are all executed for each cell). The `compute_average` function no longer takes a `FUNCTOR` template argument, but instead each functor is derived from a common base class, which makes creating a vector of functors much simpler. I am quite happy with the current state so I would like to ask for a review. In particular I would like to ask for comments about the position of the base class `FunctorBase`. I would like to hide it from the header file, but I could not forward declare its type.

I have another improvement lined up, but that one changes the functionality by reducing the number of quadrature points if possible, and I will open that as a separate PR.

### DIFF
--- a/source/simulator/lateral_averaging.cc
+++ b/source/simulator/lateral_averaging.cc
@@ -36,14 +36,23 @@ namespace aspect
 
   template <int dim>
   template <class FUNCTOR>
-  void LateralAveraging<dim>::compute_lateral_average(std::vector<double> &values,
-                                                      FUNCTOR &fctr) const
+  void LateralAveraging<dim>::compute_lateral_averages(std::vector<std::vector<double> > &values,
+                                                       std::vector<FUNCTOR> &functors) const
   {
     Assert (values.size() > 0,
             ExcMessage ("To call this function, you need to request a positive "
+                        "number of properties to compute."));
+    Assert (values[0].size() > 0,
+            ExcMessage ("To call this function, you need to request a positive "
                         "number of depth slices."));
-    const unsigned int num_slices = values.size();
-    std::vector<double> volume(num_slices);
+
+    const unsigned int n_properties = values.size();
+    const unsigned int n_slices = values[0].size();
+    for (unsigned int i=0; i < n_properties; ++i)
+      Assert (values[i].size() == n_slices,
+              ExcMessage ("All properties need to have the same number of depth slices."));
+
+    std::vector<double> volume(n_slices);
 
     // this yields 10^dim quadrature points evenly distributed in the interior of the cell.
     // We avoid points on the faces, as they would be counted more than once.
@@ -59,7 +68,7 @@ namespace aspect
 
     std::vector<std::vector<double> > composition_values (this->n_compositional_fields(),std::vector<double> (n_q_points));
 
-    std::vector<double> output_values(quadrature_formula.size());
+    std::vector<std::vector<double> > output_values(n_properties,std::vector<double>(quadrature_formula.size()));
 
     typename DoFHandler<dim>::active_cell_iterator
     cell = this->get_dof_handler().begin_active(),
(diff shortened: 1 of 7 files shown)


source/simulator/lateral_averaging.cc
```diff
@@ -448,17 +250,279 @@ namespace aspect
...
+                                                       std::vector<std::vector<double> > &values) const
+  {
+    Assert (values.size() > 0,
+            ExcMessage ("To call this function, you need to request a positive "
+                        "number of properties to compute."));
+    Assert (values[0].size() > 0,
+            ExcMessage ("To call this function, you need to request a positive "
+                        "number of depth slices."));
+    Assert (functors.size() == values.size(),
+            ExcMessage ("To call this function, you need supply as many data "
```
bangerth: -> need TO  supply

source/simulator/lateral_averaging.cc
```diff
@@ -448,17 +250,279 @@ namespace aspect
...
+    for (unsigned int i=0; i < n_properties; ++i)
+      Assert (values[i].size() == n_slices,
+              ExcMessage ("All properties need to have the same number of depth slices."));
+
+    std::vector<double> volume(n_slices);
+
+    // this yields 10^dim quadrature points evenly distributed in the interior of the cell.
+    // We avoid points on the faces, as they would be counted more than once.
+    const QIterated<dim> quadrature_formula (QMidpoint<1>(),
+                                             10);
```
bangerth: Is this what we had before? Ten seems kind of a lot. If it's still slow, there would be an easy way to reduce the computational complexity.

source/simulator/lateral_averaging.cc
```diff
@@ -448,17 +250,279 @@ namespace aspect
...
+                                               this->n_compositional_fields());
+    MaterialModel::MaterialModelOutputs<dim> out(n_q_points,
+                                                 this->n_compositional_fields());
+
+    bool functors_need_material_output = false;
+    for (unsigned int i=0; i<n_properties; ++i)
+      {
+        functors[i]->setup(quadrature_formula.size());
+        if (functors[i]->need_material_properties())
+          functors_need_material_output = true;
```
bangerth: I look forward to the day where we can use `std::reduce` or `std::transform_reduce` for this kind of thing -- but that's only in C++17 :-(
http://en.cppreference.com/w/cpp/algorithm/reduce
http://en.cppreference.com/w/cpp/algorithm/transform_reduce

source/simulator/lateral_averaging.cc
```diff
@@ -448,17 +250,279 @@ namespace aspect
...
+                          << "**** Warning: When computing depth averages, there is at least one depth band"
+                          << std::endl
+                          << "     that does not have any quadrature points in it."
+                          << std::endl
+                          << "     Consider reducing the number of depth layers for averaging."
+                          << std::endl << std::endl;
+      }
+  }
+
+  template <int dim>
```
bangerth: here and below: 3 empty lines between functions

source/simulator/lateral_averaging.cc
```diff
@@ -448,17 +250,279 @@ namespace aspect
...
+  void LateralAveraging<dim>::get_temperature_averages(std::vector<double> &values) const
+  {
+    std::vector<std::vector<double> > vector_values(1,values);
+    std::vector<std::string> property_name(1,"temperature");
+    get_averages(property_name,vector_values);
+    values.swap(vector_values[0]);
+  }
+
+  template <int dim>
+  void LateralAveraging<dim>::get_composition_averages( const unsigned int c, std::vector<double> &values) const
```
bangerth: break arguments onto separate lines

source/simulator/lateral_averaging.cc
```diff
@@ -448,17 +250,279 @@ namespace aspect
...
+      }
+  }
+
+  template <int dim>
+  void LateralAveraging<dim>::get_temperature_averages(std::vector<double> &values) const
+  {
+    std::vector<std::vector<double> > vector_values(1,values);
+    std::vector<std::string> property_name(1,"temperature");
+    get_averages(property_name,vector_values);
+    values.swap(vector_values[0]);
```
bangerth: if you return the data from `get_averages` as mentioned earlier, you can just do
```
  values = get_averages (property_name);
```

source/simulator/lateral_averaging.cc
```diff
@@ -448,17 +250,279 @@ namespace aspect
...
-                                               this->introspection().extractors.temperature,
-                                               &this->get_gravity_model() );
+    std::vector<std::vector<double> > vector_values(1,values);
+    std::vector<std::string> property_name(1,"vertical_heat_flux");
+    get_averages(property_name,vector_values);
+    values.swap(vector_values[0]);
+  }
+
+  template <int dim>
+  void LateralAveraging<dim>::get_averages(std::vector<std::string> property_names,
```
bangerth: the first argument should be a `const` reference

source/simulator/lateral_averaging.cc
gassmoeller: Yes, this is what I meant in my initial comment. I thought about adaptively computing the number of quadrature points based on the largest cell diameter and the size of the depth slices to make sure we use the lowest possible resolution while still having at least one quadrature point per depth slice. Unfortunately this decreases the accuracy of the postprocessor (because the volume fraction that we add to the wrong depth slice increases). @jdannberg preference was to have a slower but more accurate postprocessor. I am not sure about this, we could also make that an input argument.

bangerth: Yes, that's reasonable. After your 10x speedup, does the depth averaging still show up in your profiles?

gassmoeller: It is still 3 times as expensive as visualization, in my example model that is 11% of total runtime, if you run it every timestep. However, I suspect this is not longer crucial as most people will run it at most every 20 timesteps or so. In other words not terribly concerning right now. I thought about better ways to do this, but could not come up with one. In theory a `QAnisotropic` would save a lot of time, since we only need a high resolution in depth direction, but I think we can not rely on the orientation of the cells. Anyway, maybe I just stop here and say it is good enough.

gassmoeller:
I addressed all comments. I kept the Functor operator as a reference argument (this is for internal use, and in one of the inner loops over all cells), but changed the other functions to return by value, which made the code much nicer.

source/simulator/lateral_averaging.cc
bangerth: Yes, good enough. If someone complains, then we'll drop the number of integration points to 8 in each direction and gain a factor of 2 :-)

gassmoeller:
Thanks!

PR #1600, opened 2017-05-12
New melt solver

tjhei:
supersedes #1516 and created from my repo so we can edit it

PR #2138, opened 2018-03-23
New melt solver rebase

jdannberg:
Supersedes #1600. For now, I mainly want to see which tests fail.


jdannberg:
@tjhei, do you know what's going on with the tester? It seems like it doesn't run tests any more...

jdannberg:
Okay, I have checked that all the tests are still working and the output makes sense (it seems the tester is currently not working, but @gassmoeller has set up a tester locally on his machine so that I could update the test output), so it would be great if someone could have a look now! Although the pull request changes many files, most of them are just tests (and because we now use a new type of solver for the melt transport problem, all melt tests have changed). 

The main advantages are that this pull request should make the solver for melt models much faster (in some cases it reduces the iteration count by an order of magnitude), in particular in cases with low or zero porosity in a part of the domain. So that would be a big step forward. 

There are a number of things that are not clear to me and that we should discuss:

1. Boundary terms: @tjhei wondered if we have to add special boundary terms between the regions with melt and without melt, but I've tried to do this and the results just didn't make sense any more (and without these terms, all tests I've done looked good to me)
2. We have implemented the cutoff between melt cells/cells without melt as a fixed number with respect to the reference permeability, and it basically replaces the parameter "Melt transport threshold" we used before. Do we want to keep it like this (and just remove the Melt transport threshold from all input files, as it is not used anymore), or do we instead want to reuse the melt transport threshold somehow and try to make the cutoff an input parameter?
3. How do we want to compute the nonlinear residual? We introduced a new scaled pressure (p_c_bar), which is now stored in the block of the solution vector where the compaction pressure was before. So we will use p_c_bar to compute the nonlinear residual. Is that what we want? Or is there a good way to use the real compaction pressure instead?

source/particle/property/melt_particle.cc
```diff
@@ -77,6 +77,47 @@ namespace aspect
...
+      template <int dim>
+      void
+      MeltParticle<dim>::parse_parameters (ParameterHandler &prm)
+      {
+        prm.enter_subsection("Postprocess");
+        {
+          prm.enter_subsection("Particles");
+          {
+            prm.enter_subsection("Melt particle");
+            threshold_for_melt_presence = prm.get_double ("Threshold for melt presence");
```
gassmoeller: missing {}

source/simulator/melt.cc
```diff
@@ -55,6 +61,45 @@ namespace aspect
...
     }
+
+    template <int dim>
+    double
+    MeltInterface<dim>::p_c_scale (const MaterialModel::MaterialModelInputs<dim> &inputs,
+                                   const MaterialModel::MaterialModelOutputs<dim> &outputs,
+                                   const MeltHandler<dim> &handler,
+                                   bool consider_is_melt_cell) const
+    {
+      Assert(inputs.current_cell.state() == IteratorState::valid, ExcMessage("sorry, I need a cell."));
```
gassmoeller: Assert message

source/simulator/melt.cc
```diff
@@ -55,6 +61,45 @@ namespace aspect
...
+
+      if (consider_is_melt_cell && !handler.is_melt_cell(inputs.current_cell))
+        return 0.0;
+
+      const MaterialModel::MeltOutputs<dim> *melt_outputs = outputs.template get_additional_output<MaterialModel::MeltOutputs<dim> >();
+      const double ref_K_D = this->reference_darcy_coefficient();
+
+      double K_D = 1.0;
+      double max_K_D = 0.0;
+      const unsigned int N = melt_outputs->permeabilities.size();
```
gassmoeller: we usually call this n_q_points

source/simulator/melt.cc
```diff
@@ -166,19 +212,12 @@ namespace aspect
           const double eta = scratch.material_model_outputs.viscosities[q];
           const double one_over_eta = 1. / eta;
           const double eta_two_thirds = scratch.material_model_outputs.viscosities[q] * 2.0 / 3.0;
+          const double K_D = this->get_melt_handler().limited_darcy_coefficient(melt_outputs->permeabilities[q] / melt_outputs->fluid_viscosities[q], p_c_scale > 0);
```
gassmoeller: please break into two lines

source/simulator/melt.cc
```diff
@@ -189,34 +228,51 @@ namespace aspect
...
+                                             scratch.grad_phi_p[j]
+                                             +
+                                             (1./eta + 1./viscosity_c) * p_c_scale * p_c_scale *
+                                             pressure_scaling *
+                                             pressure_scaling *
+                                             (scratch.phi_p_c[i] * scratch.phi_p_c[j])
+                                            )
+                                            * JxW;
+                // add S between p_c and p_f
+                if (true)
```
gassmoeller: not necessary

source/simulator/melt.cc
```diff
@@ -300,6 +352,11 @@ namespace aspect
       const unsigned int stokes_dofs_per_cell = data.local_dof_indices.size();
       const unsigned int n_q_points    = scratch.finite_element_values.n_quadrature_points;
 
+      Assert(dynamic_cast<const MaterialModel::MeltInterface<dim>*>(&this->get_material_model()) !=
+             NULL, ExcMessage("Error: The current material model needs to be derived from MeltInterface to use melt transport."));
+
+      const double p_c_scale = dynamic_cast<const MaterialModel::MeltInterface<dim>*>(&this->get_material_model())->p_c_scale(scratch.material_model_inputs, scratch.material_model_outputs, this->get_melt_handler(), true);
```
gassmoeller: break lines

source/simulator/melt.cc
```diff
@@ -378,17 +432,15 @@ namespace aspect
 
           const unsigned int porosity_index = introspection.compositional_index_for_name("porosity");
           const double porosity = std::max(scratch.material_model_inputs.composition[q][porosity_index],0.000);
-          const double K_D = (porosity > this->get_melt_handler().melt_transport_threshold
-                              ?
-                              melt_outputs->permeabilities[q] / melt_outputs->fluid_viscosities[q]
-                              :
-                              0.0);
+
+          const double K_D = this->get_melt_handler().limited_darcy_coefficient(melt_outputs->permeabilities[q] / melt_outputs->fluid_viscosities[q], p_c_scale > 0);
```
gassmoeller: Break line

source/simulator/melt.cc
```diff
@@ -483,8 +535,9 @@ namespace aspect
       const FEValuesExtractors::Scalar ex_p_f = introspection.variable("fluid pressure").extractor_scalar();
       const unsigned int p_f_component_index = introspection.variable("fluid pressure").first_component_index;
       const unsigned int p_c_component_index = introspection.variable("compaction pressure").first_component_index;
+      const double p_c_scale = dynamic_cast<const MaterialModel::MeltInterface<dim>*>(&this->get_material_model())->p_c_scale(scratch.material_model_inputs, scratch.material_model_outputs, this->get_melt_handler(), true);
```
gassmoeller: same here

source/simulator/melt.cc
```diff
@@ -505,15 +558,7 @@ namespace aspect
...
-
-          const unsigned int porosity_index = introspection.compositional_index_for_name("porosity");
-          const double porosity = std::max(scratch.face_material_model_inputs.composition[q][porosity_index],0.0);
-
-          const double K_D = (porosity > this->get_melt_handler().melt_transport_threshold
-                              ?
-                              melt_outputs->permeabilities[q] / melt_outputs->fluid_viscosities[q]
-                              :
-                              0.0);
+          const double K_D = this->get_melt_handler().limited_darcy_coefficient(melt_outputs->permeabilities[q] / melt_outputs->fluid_viscosities[q], p_c_scale > 0);
```
gassmoeller: same

source/simulator/melt.cc
```diff
@@ -1090,15 +1148,29 @@ namespace aspect
                       this->introspection(),
                       this->get_solution());
 
-
             this->get_material_model().evaluate(in, out);
 
+            const double p_c_scale = dynamic_cast<const MaterialModel::MeltInterface<dim>*>(&this->get_material_model())->p_c_scale(in,out,this->get_melt_handler(), true);
```
gassmoeller: break line

source/simulator/melt.cc
```diff
@@ -1180,7 +1265,12 @@ namespace aspect
       FEValues<dim> fe_values (this->get_mapping(),
                                this->get_fe(),
                                quadrature,
-                               update_quadrature_points | update_values);
+                               update_quadrature_points | update_gradients | update_values);
```
gassmoeller: check if you can remove the gradients and change `compute_material_model_input_values` argument to `false`

gassmoeller:
We discussed the mentioned questions with @jdannberg and @naliboff and came to the following conclusions:
1. Since  the benchmarks work without the flux terms, and do not with them (maybe wrong terms?) we go ahead without the terms for now. If we need them we can always add them later.

2. We will remove the outdated parameter and introduce a more appropriate one, something like 'Relative permeability threshold'.

3. Since the scaling factor remains the same during nonlinear iterations (although it is different per cell) it might make no difference which one to use. Lets stick with the simple assumption that it is sufficient to use the nonlinear residual of the variable we actually have in the equation (pcbar) unless we have evidence we would need to do something else.

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+
+      std::vector<types::global_dof_index>   nonzero_dof_indices;
+    };
+
+    template <int dim>
+    class PcNonZeroAssembler : public SimulatorAccess<dim>
+    {
+      public:
+        PcNonZeroAssembler(IndexSet &nonzero_entries,
+                           std::vector<bool> &is_melt_cell)
```
gassmoeller: put the input argument in front of the output, and make `is_melt_cell` const

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+            is_melt_cell (is_melt_cell)
+        {}
+
+
+        void
+        local_assemble (const typename DoFHandler<dim>::active_cell_iterator &cell,
+                        PcAssembleData<dim> &scratch,
+                        PcCopyData<dim> &data)
+        {
+
```
gassmoeller: remove empty line

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+        void
+        copy_local_to_global (const PcCopyData<dim> &data)
+        {
+          nonzero_entries.add_indices(data.nonzero_dof_indices.begin(),
+                                      data.nonzero_dof_indices.end());
+        }
+
+      private:
+
```
gassmoeller: remove empty line

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+    };
+
+  }
+
+  template <int dim>
+  void
+  MeltHandler<dim>::
+  add_current_constraints(ConstraintMatrix &constraints)
+  {
+    // First, fill the constrains matrix with the current constraints from the beginning
```
gassmoeller: constraints

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+
+          this->get_material_model().evaluate(scratch.material_model_inputs,
+                                              scratch.material_model_outputs);
+
+          const double p_c_scale = dynamic_cast<const MaterialModel::MeltInterface<dim>*>(&this->get_material_model())->p_c_scale(scratch.material_model_inputs,scratch.material_model_outputs,this->get_melt_handler(), true);
+          const bool is_melt_cell = (p_c_scale > 0.0);
+
+          if (is_melt_cell)
+            {
+              for (unsigned int j=0; j<this->get_fe().base_element(pc_base_index).dofs_per_cell; ++j)
```
gassmoeller: Add a line `data.nonzero_dof_indices.resize(this->get_fe().base_element(pc_base_index).dofs_per_cell);` outside of the loop and access `data.nonzero_dof_indices[j]` inside.

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+    IndexSet nonzero_pc_dofs(this->introspection().index_sets.system_relevant_set.size());
+
+    const QTrapez<dim> quadrature_formula;
+    const UpdateFlags cell_update_flags = update_quadrature_points | update_values | update_gradients;
+    const FiniteElement<dim> &fe = this->get_fe();
+
+    is_melt_cell_vector.resize(this->get_dof_handler().get_triangulation().n_active_cells());
+
+    {
+      // find our "melt cells" by looking at K_D
```
gassmoeller: Update comment -> `pc_scale`

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+            this->compute_material_model_input_values (this->get_current_linearization_point(),
+                                                       finite_element_values,
+                                                       cell,
+                                                       true,
+                                                       material_model_inputs);
+
+            this->get_material_model().evaluate(material_model_inputs,
+                                                material_model_outputs);
+
+            const double p_c_scale = dynamic_cast<const MaterialModel::MeltInterface<dim>*>(&this->get_material_model())->p_c_scale(material_model_inputs,material_model_outputs,this->get_melt_handler(), false /*=consider_is_melt_cell*/);
```
gassmoeller: break line

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+
+
+    PcNonZeroAssembler<dim> assembler(nonzero_pc_dofs, is_melt_cell_vector);
+    assembler.initialize_simulator(this->get_simulator());
+
+    typedef
+    FilteredIterator<typename DoFHandler<dim>::active_cell_iterator>
+    CellFilter;
+
+    unsigned int stokes_dofs_per_cell = dim * fe.base_element(this->introspection().base_elements.velocities).dofs_per_cell
```
gassmoeller: const

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+                                        + fe.base_element(this->introspection().base_elements.pressure).dofs_per_cell
+                                        + fe.base_element(this->introspection().base_elements.pressure).dofs_per_cell;
+
+    WorkStream::
```
gassmoeller: Add comment that we here call the assembler

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+                          &assembler,
+                          std_cxx11::_1),
+         PcAssembleData<dim> (fe, quadrature_formula,
+                              this->get_mapping(),
+                              cell_update_flags,
+                              this->introspection().n_compositional_fields,
+                              stokes_dofs_per_cell),
+         PcCopyData<dim> (stokes_dofs_per_cell));
+
+    // first pick all relevant p_c's:
```
gassmoeller: relevant p_c dof's

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+    IndexSet for_constraints = this->introspection().index_sets.system_relevant_set
+                               & Utilities::extract_locally_active_dofs_with_component<dim>(this->get_dof_handler(),
+                                   this->introspection().variable("compaction pressure").component_mask);
+
+    // now subtract the ones that are nonzero as computed above:
+    for_constraints.subtract_set(nonzero_pc_dofs);
+
+    // and constrain those. Note that we do not constrain p_c dofs that
+    // are on the interface between a cell "with melt" and a cell "without melt"
+    // (using the definition "at least one quadrature point with p_c_scale>0").
```
gassmoeller: Remove second comment sentence

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+  }
+
+
+  template <int dim>
+  double
+  MeltHandler<dim>::
+  limited_darcy_coefficient(const double K_D,
+                            const bool is_melt_cell) const
+  {
+    const double ref_K_D = dynamic_cast<const MaterialModel::MeltInterface<dim>*>(&this->get_material_model())->reference_darcy_coefficient();
```
gassmoeller: break this line

source/simulator/melt.cc
```diff
@@ -1258,11 +1623,12 @@ namespace aspect
     variables.insert(variables.begin()+2,
                      VariableDeclaration<dim>(
                        "compaction pressure",
-                       (parameters.use_locally_conservative_discretization)
+                       (parameters.use_locally_conservative_discretization || use_discontinuous_p_c)
```
gassmoeller: Do you allow discontinuous normal pressure with melt? Please check. Even if so, remove the first part of this condition as you silently overwrite a setting from the input file.

source/simulator/melt.cc
```diff
@@ -1258,11 +1623,12 @@ namespace aspect
...
                        ?
                        std_cxx11::shared_ptr<FiniteElement<dim> >(new FE_DGP<dim>(parameters.stokes_velocity_degree-1))
                        :
-                       std_cxx11::shared_ptr<FiniteElement<dim> >(new FE_Q<dim>(parameters.stokes_velocity_degree-1)),
+                       std_cxx11::shared_ptr<FiniteElement<dim> >(new FE_Q<dim>(parameters.stokes_velocity_degree-1))
+                       ,
```
gassmoeller: Move comma to line above

source/simulator/melt.cc
```diff
@@ -1363,6 +1752,11 @@ namespace aspect
     if (SimulatorAccess<dim> *sim = dynamic_cast<SimulatorAccess<dim>*>(boundary_fluid_pressure.get()))
       sim->initialize_simulator (simulator_object);
     boundary_fluid_pressure->initialize ();
+
+    if (use_discontinuous_p_c)
```
gassmoeller: Try moving this to `parse_parameters` if it does not work we need to look at core.cc (simulator constructor)

source/simulator/parameters.cc
```diff
@@ -1287,8 +1287,7 @@ namespace aspect
                       && !use_discontinuous_composition_discretization,
                       ExcMessage ("Using discontinuous elements for temperature "
                                   "or composition in models with melt transport is currently not implemented."));
-          // We can not have a DG p_f. While it would be possible to use a
-          // discontinuous p_c, this is not tested, so we disable it for now.
+          // We can not have a DG p_f.
           AssertThrow(!use_locally_conservative_discretization,
                       ExcMessage ("Discontinuous elements for the pressure "
                                   "in models with melt transport are not supported"));
```
gassmoeller: Add dot at end of sentence

source/simulator/solver.cc
```diff
@@ -280,10 +280,10 @@ namespace aspect
       // first solve with the bottom left block, which we have built
       // as a mass matrix with the inverse of the viscosity
       {
-        SolverControl solver_control(1000, src.block(1).l2_norm() * S_block_tolerance);
+        aspect::SolverControl solver_control(1000, src.block(1).l2_norm() * S_block_tolerance);
```
gassmoeller: do you need this? should be fixed on master already

source/simulator/solver.cc
```diff
@@ -280,10 +280,10 @@ namespace aspect
...
 
 #ifdef ASPECT_USE_PETSC
-        SolverCG<LinearAlgebra::Vector> solver(solver_control);
+        SolverGMRES<LinearAlgebra::Vector> solver(solver_control);
```
gassmoeller: revert this to CG as before

source/simulator/solver.cc
```diff
@@ -835,6 +836,7 @@ namespace aspect
                   {
                     // output solver history
                     std::ofstream f((parameters.output_directory+"solver_history.txt").c_str());
+                    f << std::setprecision(16);
```
gassmoeller: Leave that out

tests/rising_melt_blob.prm
```diff
@@ -0,0 +1,147 @@
+#########################################################
+# This is a test for the melting rate functionality
+# It uses a new material model similar to the melt
+# simple model, but with a different melting rate.
+
+# The melting rate has the form of a Gaussian function 
+# and every timestep a fixed amount of material is melting.
```
gassmoeller: Update description

source/simulator/melt.cc
jdannberg: I can set use_strain_rate to false, but we still need the gradients (as the material model inputs always compute the pressure gradients). 

Oh, and I changed the `compute_material_model_input_values` to `material_model_inputs.reinit`, as this is our new way to write this, correct?

source/simulator/melt.cc
```diff
@@ -1234,6 +1333,272 @@ namespace aspect
...
+      public:
+        PcNonZeroAssembler(IndexSet &nonzero_entries,
+                           std::vector<bool> &is_melt_cell)
+          : nonzero_entries (nonzero_entries),
+            is_melt_cell (is_melt_cell)
+        {}
+
+
+        void
+        local_assemble (const typename DoFHandler<dim>::active_cell_iterator &cell,
```
jdannberg: Now that I am looking over this, should I also give this a different name (something like `local_save_nonzero_pc_dofs`)?

source/simulator/melt.cc
gassmoeller: Ok then do that.

Yes I think it is.

source/simulator/melt.cc
gassmoeller: sure, sounds good

source/simulator/melt.cc
```diff
@@ -255,23 +263,19 @@ namespace aspect
...
-                       pressure_scaling *
-                       pressure_scaling)
-                      * scratch.phi_p[i] * scratch.phi_p_c[j]
-                      +
-                      (p_c_scale * one_over_eta *
-                       pressure_scaling *
-                       pressure_scaling)
-                      * scratch.phi_p_c[i] * scratch.phi_p[j]
-                    )
-                    * JxW;
```
gassmoeller: The above lines might be the cause for the test changes. Previously you always added these line (`if(true)`, although that condition is unnecessary), now you only add them if the condition above is true (component_index == component_index). Revert to the old state, but remove the `if(true)`. Its only the preconditioner though, not sure this is the only reason.

include/aspect/melt.h
```diff
@@ -271,27 +274,20 @@ namespace aspect
...
-   */
-  template <int dim>
-  class MeltHandler: public SimulatorAccess<dim>
+  namespace Melt
   {
-    public:
-      MeltHandler(ParameterHandler &prm);
-
+    template <int dim>
+    struct Parameters
```
gassmoeller: This struct should only contain the input or other parameters for MeltHandler, not internal variables. In particular `boundary_fluid_pressure`, `is_melt_cell_vector`, and `current_constraints` should be private member variables of the class and not in this struct. If you need them outside, make `get_` functions for them.

source/simulator/melt.cc
```diff
@@ -1689,92 +1692,114 @@ namespace aspect
...
-                         "and $\\phi$ go to zero for vanishing melt fraction. As the melt velocity is "
-                         "used for computing the time step size, and in models that use heat "
-                         "transport by melt or shear heating of melt, setting this parameter to true "
-                         "can speed up the model and make it mode stable. In computations where "
-                         "accuracy and convergence behavior of the melt velocity is important "
-                         "(like in benchmark cases with an analytical solution), this parameter "
-                         "should probably be set to 'false'.");
+      prm.enter_subsection ("Melt settings");
+      {
+        prm.declare_entry ("Darcy coefficient variation threshold", "1e-3",
```
gassmoeller: I do not like the `variation` a lot. What about `Relative Darcy coefficient threshold`? Variation sounds like you are measuring the spatial derivatives of darcy coefficients.

source/simulator/melt.cc
```diff
@@ -1689,92 +1692,114 @@ namespace aspect
...
 
-    BoundaryFluidPressure::declare_parameters<dim> (prm);
+    template <int dim>
+    void
+    Parameters<dim>::
+    parse_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection ("Melt settings");
+      {
+        K_D_variation_threshold = prm.get_double("Darcy coefficient variation threshold");
```
gassmoeller: `relative_K_D_threshold`?

source/simulator/parameters.cc
```diff
@@ -1278,20 +1278,6 @@ namespace aspect
...
-        {
-          // The additional terms in the temperature systems have not been ported
-          // to the DG formulation:
-          AssertThrow(!use_discontinuous_temperature_discretization
-                      && !use_discontinuous_composition_discretization,
-                      ExcMessage ("Using discontinuous elements for temperature "
-                                  "or composition in models with melt transport is currently not implemented."));
-          // We can not have a DG p_f.
-          AssertThrow(!use_locally_conservative_discretization,
-                      ExcMessage ("Discontinuous elements for the pressure "
```
gassmoeller: I like this part :smile:, less melt code in the general code.

tests/compaction_length_refinement.prm
```diff
@@ -110,10 +110,6 @@ subsection Model settings
   set Include melt transport                  = true
 end
 
-subsection Melt settings
-  set Melt transport threshold                = 0
```
gassmoeller: Could this be what is causing the changes? Or was the melt transport threshold not used before as well?

source/simulator/melt.cc
jdannberg: Ah, I see, I'll try it. Thanks!

source/simulator/melt.cc
jdannberg: I feel like it's not really clear what a relative threshold should be, so that name feels a bit weird to me. I thought one other way to express this would be to say, how much is K_D allowed to vary (or deviate) from the reference value? I could call it "Global Darcy coefficient variation threshold" to make clear it doesn't have anything to do with derivatives.

gassmoeller: Hmm, I am not really happy with that either. Lets discuss it later today.

jdannberg: I looked for parameters with a similar meaning in the code, and the closest thing I could find was something like the "Maximum lateral viscosity variation" in the Steinberger material model. But sure, let's discuss this later.

gassmoeller:
Thanks for working through the multiple reviews. I think if we missed anything, we need to address it as a fix later. Melt 2.0 done, :tada:

bangerth:
yay -- good work y'all!

PR #2139, opened 2018-03-23
Remove anisotropic viscosity from main code (move to tests)

gassmoeller:
It does not seem as if the anisotropic viscosity is used in practical applications a lot. On the other hand we now have all the infrastructure to remove it from the main assemblers, and restrict its use to the models that actually need it. The only model that uses it so far is our test/anisotropic viscosity (and we have no normal material model that supports it, only the plugin for anisotropic viscosity).
I did this mostly to clean up the code of the assemblers, but it turns out that the performance influence was bigger than expected. The change speeds up the assembly of the Stokes system and the preconditioner by about 25% in terms of instructions measured by callgrind (in 3d, optimized mode, using the geoid.prm test as reference). I am not sure why, but in the example we save about 10 instructions per innermost loop cycle, which is now down from 25 to 15 instructions per loop. It looks like the tensor contract function is inlined in the new version, but was not before, though that is a guess, it is not exactly clear to me what is happening.

Anyway, nice speedup, cleaner code, functionality still around if someone needs it.

### DIFF
--- a/tests/anisotropic_viscosity.cc
+++ b/tests/anisotropic_viscosity.cc
@@ -1,8 +1,594 @@
-#include <aspect/material_model/simple.h>
 #include <aspect/simulator_access.h>
+
+#include <aspect/material_model/simple.h>
+#include <aspect/material_model/grain_size.h>
+#include <aspect/heating_model/interface.h>
+#include <aspect/gravity_model/interface.h>
+#include <aspect/simulator/assemblers/stokes.h>
+#include <aspect/simulator_signals.h>
+
+#include <deal.II/base/symmetric_tensor.h>
+#include <deal.II/base/signaling_nan.h>
 
 namespace aspect
 {
+  namespace MaterialModel
+  {
+    /**
+      * Additional output fields for anisotropic viscosities to be added to
+      * the MaterialModel::MaterialModelOutputs structure and filled in the
+      * MaterialModel::Interface::evaluate() function.
+      */
+    template <int dim>
+    class AnisotropicViscosity : public NamedAdditionalMaterialOutputs<dim>
+    {
+      public:
+        AnisotropicViscosity(const unsigned int n_points);
+
+        virtual std::vector<double> get_nth_output(const unsigned int idx) const;
+
+        /**
+         * Stress-strain "director" tensors at the given positions. This
+         * variable is used to implement anisotropic viscosity.
+         *
+         * @note The strain rate term in equation (1) of the manual will be
+         * multiplied by this tensor *and* the viscosity scalar ($\eta$), as
+         * described in the manual section titled "Constitutive laws". This
+         * variable is assigned the rank-four identity tensor by default.
+         * This leaves the isotropic constitutive law unchanged if the material
+         * model does not explicitly assign a value.
+         */
+        std::vector<SymmetricTensor<4,dim> > stress_strain_directors;
+    };
+
+    template <int dim>
+    AnisotropicViscosity<dim>::AnisotropicViscosity (const unsigned int n_points)
+      :
+      NamedAdditionalMaterialOutputs<dim>(std::vector<std::string>(1,"anisotropic_viscosity")),
+      stress_strain_directors(n_points, dealii::identity_tensor<dim> ())
+    {}
+
+
+
+    template <int dim>
+    std::vector<double>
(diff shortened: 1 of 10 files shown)


gassmoeller:
Oh, and @jperryhouts you should be aware of this. I am not sure if you are still running models with the anisotropic viscosity, but if so, you can copy the assemblers from the test and use them as before. Let me know if you need help with that or in case this PR creates other problems for you.

jperryhouts:
Ah, thanks! I am still using this, so I'll update my plugins accordingly.

That speedup is surprising. The only reason this made it into the main branch in the first place is that it appeared to cause no noticeable slowdown at the time. I might poke around the tensor contraction commit history at some point to see if anything jumps out, it would be nice to have that speed back in all cases.

jperryhouts:
Actually, it shouldn't be a problem with tensor contraction because there's a check for whether or not the 4th order tensor is initialized, and avoids contracting it with grad(u) if not. So either that comparison wasn't working, or was being checked too frequently.

Oh well, in either case I agree that removing it is the right option, just confused about the speedup.

gassmoeller:
I checked this again in detail, but the only thing I can find is that without the anisotropy switch the symmetric tensor contraction (the one for the second rank tensors if `use_tensor=false`) is inlined, while it is not inlined if the switch is in there. The code inside of the tensor contraction actually uses the same cycles, so the only thing I can see is the cost of the function call, and maybe some optimization the compiler can do when inlining things. I remember the discussion we had when you implemented the feature and I know you did careful tests then. Maybe it depends on compiler versions or settings (I use gcc-4.8.4, release mode, no modification to compiler flags), but I would rather not depend on which compiler users are using. 
I would suggest moving it to tests for now. If more users want it, we can always create new regular assemblers that always use the anisotropic viscosity and activate them depending on an input parameter.

jperryhouts:
Huh, interesting. In-lining does make a lot of compiler optimization possible. I can't imagine a function call could cause a 60% increase in instructions.

In any case, I definitely don't have a problem with removing it. Good to know it won't be in 2.0 though, I'm about to submit a manuscript that almost preemptively claimed it would be :)

jdannberg:
As it speeds up the code so much to remove this from the main assemblers, and we currently do not have a material model that uses this functionality, I agree that we should move this into separate assembler plugins. However, I fear that if this remains in the tests folder, nobody will be able to find and use it. The best place for this would be in a cookbook or benchmark with a separate plugin. However, we just don't have the expertise to write a cookbook for this functionality (for me, I just don't know how I would convert for example the finite strain into an anisotropic viscosity). 

So @jperryhouts, it would be great if you have a simple input file & material model that you would be willing to contribute and that we could use to write a cookbook. If you don't want to/don't have the time to write a cookbook yourself, I could do that and we would of course give you the credit for providing the setup (and reference the manuscript you're about to submit). But I just feel it would be sad to ban this feature to a test file forever, as it would both be a good feature to advertise in ASPECT, and your paper might also gain some visibility if we link back to it from the cookbook.

jperryhouts:
I'd be happy to. Don't have time at the moment, but I'll write something up as soon as the paper's submitted. Hopefully before the end of the month.

jdannberg:
@jperryhouts That sounds great! So would it be okay for you if we move this to a test for now, and you'll move it to a plugin and write a cookbook when you have the time?

tests/anisotropic_viscosity.cc
```diff
@@ -1,8 +1,594 @@
-#include <aspect/material_model/simple.h>
 #include <aspect/simulator_access.h>
 
+#include <aspect/material_model/simple.h>
+#include <aspect/material_model/grain_size.h>
```
naliboff: Why do you need grain_size.h?

tests/anisotropic_viscosity.prm
```diff
@@ -64,7 +64,7 @@ subsection Material model
 end
 
 subsection Heating model
-  set List of model names = shear heating
+  set List of model names = anisotropic shear heating
```
naliboff: Update documentation for this test.

tests/anisotropic_viscosity.cc
gassmoeller: Because the anisotropic shear heating plugin needs the additional `DislocationViscosityOutputs` defined in grain_size.h.

gassmoeller:
I added documentation for the test.

PR #2140, opened 2018-03-28
futher cleanup spiegelman benchmark.

MFraters:
I noticed that I forgot to rename the class of the benchmark material. So that is done here.

PR #2141, opened 2018-03-28
Add spiegelman test

MFraters:
This is based on top of #2140, so you can ignore the changes to `benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc`. I am not sure wheter this should be merged in the end, since it is only a test of a material only used for a benchmark, but since @bangerth and I wanted to make sure that the derivatives of the benchmark material are working correctly I added this test. 

The tests show that in general the derivatives are in agreement, but only with very low strain-rates they become different. As I commented in the code, the differences disappear when larger and fully divergence free strain-rate tensor is used.

### DIFF
--- a/benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
+++ b/benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
@@ -79,7 +79,7 @@ namespace aspect
      */
 
     template <int dim>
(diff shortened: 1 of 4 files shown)


bangerth:
Can you please rebase this patch now? The same applies as for #2140: please create these branches in your own repository, rather than the main repository in the future. (But no need to migrate it now.)

The new test also seems to fail on the tester.

MFraters:
ah, sorry :( I somehow messed up my remotes in a new build directory, I will fix it now.

PR #2142, opened 2018-03-28
Add spiegelman benchmark material test.

MFraters:
I am not sure whether this should be merged in the end, since it is only a test of a material only used for a benchmark, but since @bangerth and I wanted to make sure that the derivatives of the benchmark material are working correctly I added this test.

The tests show that in general the derivatives are in agreement, but only with very low strain-rates they become different. As I commented in the code, the differences disappear when larger and fully divergence free strain-rate tensor is used.

### DIFF
--- a/benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
+++ b/benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
@@ -86,7 +86,7 @@ namespace aspect
 
         double compute_second_invariant(const SymmetricTensor<2,dim> strain_rate, const double min_strain_rate) const;
 
-        double compute_viscosity(const double edot_ii,const double pressure,const int comp, const double constant_viscosity,const bool regularize, const double min_visc, const double max_visc) const;
+        double compute_viscosity(const double edot_ii,const double pressure,const int comp, const double constant_viscosity,const bool compute_full_viscosity, const double min_visc, const double max_visc) const;
 
         virtual void evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
                               MaterialModel::MaterialModelOutputs<dim> &out) const;
@@ -250,11 +250,12 @@ namespace aspect
                       const double pressure,
                       const int comp,
                       const double constant_viscosity,
-                      const bool regularize,
+                      const bool compute_full_viscosity,
                       const double min_visc,
                       const double max_visc) const
     {
       double viscosity;
+
       if (constant_viscosity == 0)
         {
           const double strength = ( (dim==3)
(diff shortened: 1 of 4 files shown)


bangerth:
I'm always in favor of more tests as you know. But you'll have to adjust the output: http://www.math.clemson.edu/~heister/cib/aspect-8.5-gcc-hd/62833c0417b4b9645539cf2cc56e95cafda3405f/changes-gcc.diff

tests/spiegelman_material.cc
```diff
@@ -0,0 +1,289 @@
+#include <aspect/simulator.h>
```
gassmoeller: Is this a copy of the `benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc`? If so, replace the whole file by one with
`#include <../benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc` to avoid the two files getting out of sync.

tests/spiegelman_material.prm
```diff
@@ -0,0 +1,4 @@
+# This test checks wether the derivatives of the simple nonlinear
+# material model are the same as the finite difference derivatives 
+# created by the simple nonlinear material model.
+set Additional shared libraries = tests/libsimple_nonlinear.so
```
gassmoeller: Please update test description and replace the parameter by:
```
 set Additional shared libraries = libspiegelman_material.so 
```
Technically the test would work without the parameter anyway (because it is automatically added by the test system), but it makes more clear what is happening.

tests/spiegelman_material.cc
MFraters: no, it just has the same name. This one actually tests the derivatives of `../benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc` :)

tests/spiegelman_material.prm
MFraters: sure

MFraters:
This pull request is ready for an other review. It now also contains a bug fix for the Spiegelman benchmark material for when the minimum and maximum viscosities are activated.

benchmarks/spiegelman_et_al_2016_gcubed/spiegelman_material.cc
```diff
@@ -362,7 +366,7 @@ namespace aspect
                           //analytic
                           if (constant_viscosity[c] == 0  && composition_viscosities[c] <= max_visc[c] && composition_viscosities[c] >= min_visc[c])
                             {
-                              const double drucker_prager_viscosity = compute_viscosity(edot_ii,pressure,c,constant_viscosity[c],false,min_visc[c],max_visc[c]);
+                              const double drucker_prager_viscosity = compute_viscosity(edot_ii,pressure,c,constant_viscosity[c],false,1,1e100);
```
gassmoeller: Wait, shouldnt the last two parameters only be important if the bool flag before them is set to true? In other words: Why do you need to change their values if they are not used? Also: What do the magical numbers mean? How are they different from min_visc and max_visc? If they are unimportant  and just need to be 'something small' and 'something large' then please use `std::numeric_limits<double>::min()` and `max()` instead.

MFraters: Yes, you are right. What do you think of this solution? I also added some comments to explain it.

gassmoeller:
Looks good, thanks for the test.

PR #2144, opened 2018-03-30
Add forgotten subsection

gassmoeller:
I broke this in #2054. The update script now works again.


jdannberg:
Hm. I guess we don't even have to wait for the tester for this one. Maybe we should think about creating a test for the update script in the future?

PR #2147, opened 2018-03-30
Update script test

gassmoeller:
Adds the test requested in #2144.


tjhei:
Neat idea!

gassmoeller:
Thanks, it took a couple of broken versions of the update script until I was finally forced by @jdannberg to implement the test. Should have done it from the beginning.

PR #2148, opened 2018-04-02
Introduce conversion functions for plugins

gassmoeller:
One of the features of ASPECT I understood the least in the beginning was the use of `dynamic_cast` to determine the actual type of an interface plugin pointer. This is a proposal to make this more explicit and easier to understand. There are two main advantages to the new functions:
- No `dynamic_cast` that confuses users, instead function names that describe what is done
- No use of null pointers anymore. Therefore, nobody can forget to check for a null pointer before using the object. Always return references.

If this is something that we would like to do, there are dozens of places over the whole codebase which can be simplified with the new functions.
Comments?

### DIFF
--- a/include/aspect/postprocess/interface.h
+++ b/include/aspect/postprocess/interface.h
@@ -24,7 +24,6 @@
 
 #include <aspect/global.h>
 #include <aspect/plugins.h>
-#include <aspect/utilities.h>
 #include <aspect/simulator_access.h>
 
 #include <deal.II/base/std_cxx11/shared_ptr.h>
@@ -230,7 +229,8 @@ namespace aspect
          * if one of them has the desired type specified by the template
          * argument. If so, return a pointer to it. If no postprocessor is
          * active that matches the given type, return a NULL pointer.
-         * @deprecated: Use has_matching_postprocessor() and
+         *
+         * @deprecated Use has_matching_postprocessor() and
          * get_matching_postprocessor() instead.
          */
         template <typename PostprocessorType>
@@ -240,7 +240,7 @@ namespace aspect
         /**
          * Go through the list of all postprocessors that have been selected
          * in the input file (and are consequently currently active) and return
-         * if one of them has the desired type specified by the template
+         * true if one of them has the desired type specified by the template
          * argument.
          */
         template <typename PostprocessorType>
@@ -250,9 +250,10 @@ namespace aspect
         /**
          * Go through the list of all postprocessors that have been selected
          * in the input file (and are consequently currently active) and see
-         * if one of them has the desired type specified by the template
-         * argument. If so, return a reference to it. If no postprocessor is
-         * active that matches the given type, throw an exception.
+         * if one of them has the type specified by the template
+         * argument or can be casted to that type. If so, return a reference
+         * to it. If no postprocessor is active that matches the given type,
+         * throw an exception.
          */
         template <typename PostprocessorType>
         PostprocessorType &
@@ -411,7 +412,7 @@ namespace aspect
       for (typename std::vector<std_cxx11::shared_ptr<Interface<dim> > >::const_iterator
            p = postprocessors.begin();
            p != postprocessors.end(); ++p)
-        if (Utilities::plugin_type_matches<PostprocessorType>(*(*p)))
+        if (Plugins::plugin_type_matches<PostprocessorType>(*(*p)))
           return true;
 
       return false;
@@ -435,11 +436,11 @@ namespace aspect
       for (typename std::vector<std_cxx11::shared_ptr<Interface<dim> > >::const_iterator
            p = postprocessors.begin();
            p != postprocessors.end(); ++p)
-        if (Utilities::plugin_type_matches<PostprocessorType>(*(*p)))
-          return Utilities::get_plugin_as_type<PostprocessorType>(*(*p));
+        if (Plugins::plugin_type_matches<PostprocessorType>(*(*p)))
+          return Plugins::get_plugin_as_type<PostprocessorType>(*(*p));
 
       // We will never get here, because we had the Assert above. Just to avoid warnings.
-      return Utilities::get_plugin_as_type<PostprocessorType>(*(*postprocessor));
+      return Plugins::get_plugin_as_type<PostprocessorType>(*(*postprocessor));
     }
 
 
--- a/include/aspect/postprocess/interface.h
+++ b/include/aspect/postprocess/interface.h
@@ -24,6 +24,7 @@
 
 #include <aspect/global.h>
 #include <aspect/plugins.h>
+#include <aspect/utilities.h>
 #include <aspect/simulator_access.h>
 
 #include <deal.II/base/std_cxx11/shared_ptr.h>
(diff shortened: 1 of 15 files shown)


include/aspect/postprocess/interface.h
```diff
@@ -231,6 +232,28 @@ namespace aspect
...
+        /**
+         * Go through the list of all postprocessors that have been selected
+         * in the input file (and are consequently currently active) and see
+         * if one of them has the desired type specified by the template
+         * argument. If so, return a pointer to it. If no postprocessor is
+         * active that matches the given type, return a NULL pointer.
+         */
+        template <typename PostprocessorType>
+        PostprocessorType &
+        get_postprocessor_of_type () const;
```
bangerth: Would you want to deprecate `find_postprocess()` in that case?

Also, the docs still talk about pointers.

include/aspect/postprocess/interface.h
```diff
@@ -378,6 +401,46 @@ namespace aspect
...
+     */
+    template <int dim>
+    template <typename PostprocessorType>
+    inline
+    PostprocessorType &
+    Manager<dim>::get_postprocessor_of_type () const
+    {
+      AssertThrow(has_postprocessor_of_type<PostprocessorType> (),
+                  ExcMessage("You asked Postprocess:Manager::get_postprocessor_of_type() for a "
+                             "postprocessor type that could not be found in the current model. Activate this "
```
bangerth: I think you can print the actual type of the requested postprocessor using `typeid(PostprocessorType).name()`. You'll have to demangle this, though. I think there is a function for this in `dealii::Utilities`.

include/aspect/postprocess/interface.h
```diff
@@ -378,6 +401,46 @@ namespace aspect
...
+                             "postprocessor in the input file."));
+
+      typename std::vector<std_cxx11::shared_ptr<Interface<dim> > >::const_iterator postprocessor;
+      for (typename std::vector<std_cxx11::shared_ptr<Interface<dim> > >::const_iterator
+           p = postprocessors.begin();
+           p != postprocessors.end(); ++p)
+        if (Utilities::object_is_of_type<PostprocessorType>(*(*p)))
+          postprocessor = p;
```
bangerth: couldn't you just return right here? No need to keep the loop going.

include/aspect/utilities.h
```diff
@@ -1069,6 +1069,28 @@ namespace aspect
     template <int dim>
     SymmetricTensor<2,dim> nth_basis_for_symmetric_tensors (const unsigned int k);
 
+    template <typename TestType, typename ObjectType>
+    inline
+    bool
+    object_is_of_type (const ObjectType &object)
```
bangerth: I'm not fond of that name because the real object may actually be of type *derived* from `TestType`. But I also can't come up with a better name. Want to give it another try? Maybe `object_type_matches`?

gassmoeller:
I think I addressed your comments (I like the new name and the type information, and yes I intended to deprecate the `find_postprocessor` function), and added tests for the asserts. If you agree this is useful, should we merge this PR first, and I will create a follow-up that uses the new functions in all the other places?

include/aspect/postprocess/interface.h
```diff
@@ -226,10 +230,33 @@ namespace aspect
          * if one of them has the desired type specified by the template
          * argument. If so, return a pointer to it. If no postprocessor is
          * active that matches the given type, return a NULL pointer.
+         * @deprecated: Use has_matching_postprocessor() and
```
bangerth: add an empty line here to make sure we don't actually wrap the whole paragraph. I don't think you need the colon.

include/aspect/postprocess/interface.h
```diff
@@ -226,10 +230,33 @@ namespace aspect
...
          */
         template <typename PostprocessorType>
         PostprocessorType *
-        find_postprocessor () const;
+        find_postprocessor () const DEAL_II_DEPRECATED;
+
+        /**
+         * Go through the list of all postprocessors that have been selected
+         * in the input file (and are consequently currently active) and return
+         * if one of them has the desired type specified by the template
```
bangerth: return *true*

include/aspect/utilities.h
```diff
@@ -1069,6 +1071,51 @@ namespace aspect
...
+    TestType &
+    get_plugin_as_type (PluginType &object)
+    {
+      if (!plugin_type_matches<TestType>(object))
+        AssertThrow(false,
+                    ExcMessage("You have requested to convert a plugin of type <"
+                               + boost::core::demangle(typeid(PluginType).name())
+                               + "> into type <"
+                               + boost::core::demangle(typeid(TestType).name()) +
+                               "> in which it can not be converted."));
```
bangerth: this reads awkwardly. say "...into type <...>. But this cast cannot be performed."

include/aspect/utilities.h
```diff
@@ -1069,6 +1071,51 @@ namespace aspect
...
+     * plugin_matches_type() before calling this function. If the plugin is
+     * not convertible this function throws an exception.
+     */
+    template <typename TestType, typename PluginType>
+    inline
+    TestType &
+    get_plugin_as_type (PluginType &object)
+    {
+      if (!plugin_type_matches<TestType>(object))
+        AssertThrow(false,
```
bangerth: Couldn't you move the condition into the assertion where you currently have `false`?

include/aspect/postprocess/interface.h
```diff
@@ -226,10 +230,34 @@ namespace aspect
...
+         * argument.
+         */
+        template <typename PostprocessorType>
+        bool
+        has_matching_postprocessor () const;
+
+        /**
+         * Go through the list of all postprocessors that have been selected
+         * in the input file (and are consequently currently active) and see
+         * if one of them has the desired type specified by the template
```
tjhei: " or can be casted to that type"? This might be important if we have additional base classes.

include/aspect/utilities.h
```diff
@@ -1069,6 +1071,50 @@ namespace aspect
...
+     * a radial gravity model might only be implemented for spherical geometry
+     * models, and would want to check if the current geometry is in fact a
+     * spherical shell.
+     */
+    template <typename TestType, typename PluginType>
+    inline
+    bool
+    plugin_type_matches (const PluginType &object)
+    {
+      return (dynamic_cast<const TestType *> (&object) != 0);
```
tjhei: NULL?

gassmoeller:
I addressed your comments, but have one more question. The `Utilities` namespace does not feel quite right for this type of functions, and it would require me to include utilities.h in a lot of new plugins. These functions provide functionality at the very core of the plugin architecture, so should they go into plugins.h? That header is included everywhere (every interface includes it). And should I put them into a new namespace `aspect::Plugins`? Because the only namespace in there is currently `aspect::internal::Plugins`, but these functions are not intended for internal use, they should be used as widely as possible.

gassmoeller:
I understood that I had the ok to merge with the modifications, so I am going ahead with that. Let me know if there is need for more discussions, since this is not yet used a lot we can still change things.

PR #2152, opened 2018-04-04
Make radial linear more useful

gassmoeller:
The current 'radial linear' gravity profile is only useful for models of a full sphere (e.g. inner core), because it decreases linearly to 0 at the bottom of the domain (the gravity profile of a full sphere with constant density). Adding a simple input parameter makes it also useful for spherical shell models like mantle convection. Let me know if you prefer to keep the old value as default, I just thought it would be nice to have a reasonable profile for mantle convection as default. Below is a comparison of our currently available gravity profiles, with `ascii data` using the PREM profile (blue line) as the most realistic (the only one of the lines based on observations). 

As you see `radial linear` is not really good (a constant value of 10 would be closer to PREM than a radial linear profile with correct end points), but it can be a good model for the upper mantle. `radial earth-like` on the other hand is doing a terrible job, and I will open another pull request after this one that removes it (more discussion about that in the other PR)

comparison


jdannberg:
> radial earth-like on the other hand is doing a terrible job, and I will open another pull request after this one that removes it (more discussion about that in the other PR)

Yes, I totally agree! That has a huge potential to confuse people...

source/gravity_model/radial.cc
```diff
@@ -120,8 +120,8 @@ namespace aspect
         return Tensor<1,dim>();
 
       const double depth = this->get_geometry_model().depth(p);
-      return  (-magnitude_at_surface * p/p.norm() *
-               (1.0 - depth/this->get_geometry_model().maximal_depth()));
+      return  (-magnitude_at_surface  * p/p.norm() * (1.0 - depth/this->get_geometry_model().maximal_depth()))
+              + (-magnitude_at_bottom * p/p.norm() * (depth/this->get_geometry_model().maximal_depth()));
```
bangerth: You can pull the factor `p/p.norm()` out of the sum. It is really just the unit outward vector whereas the rest of the formula determines the magnitude.

source/gravity_model/radial.cc
```diff
@@ -137,6 +137,14 @@ namespace aspect
                              Patterns::Double (),
                              "Magnitude of the radial gravity vector "
                              "at the surface of the domain. Units: $m/s^2$");
+          prm.declare_entry ("Magnitude at bottom", "10.7",
+                             Patterns::Double (),
+                             "Magnitude of the radial gravity vector "
+                             "at the bottom of the domain. 'Bottom' means the"
```
bangerth: Since this goes into latex, use a backtick for the opening one. Same below.

source/gravity_model/radial.cc
```diff
@@ -137,6 +137,14 @@ namespace aspect
...
+                             "maximum depth in the chosen geometry, and for "
+                             "example represents the CMB in the 'spherical shell' "
```
bangerth: I know it's a common abbreviation, but can you spell out CMB anyway?

source/gravity_model/radial.cc
```diff
@@ -137,6 +137,14 @@ namespace aspect
...
+                             "geometry, and the center of the 'sphere' geometry. "
```
bangerth: -> in the case of the sphere geometry

gassmoeller:
Ok, addressed the comments. One of the tests had a slightly changed iteration count (probably floating point rounding differences), but that should be ok.

PR #2153, opened 2018-04-04
Remove radial earth like gravity plugin

gassmoeller:
As discussed in #2152 `radial earth-like` is a quadratic approximation to the real gravity profile under certain assumptions (constant density, and correct end points), but they are clearly violated in the earth, and therefore the resulting profile is not close to being `earth-like`. Thus the name is confusing, and we have better alternatives around (e.g. the `ascii data` plugin, which implements PREM by default). This PR removes the plugin and replaces every occurence with `ascii data`. 

There would be two alternatives, but I think this is the best option.
Alternative 1: Make `radial earth-like` an alias for `ascii data with default options (PREM)`. This is confusing.
Alternative 2: Make `radial earth-like a quadratic approximation (e.g. a least-squares fit) to PREM. This allows us keeping the clear name, but we might as well use the real PREM instead of a quadratic approximation.

The only downside to the current approach is that `ascii data` does not make clear we are using PREM, or anything earth like. But this is stated in the description of `ascii data` so hopefully that is clear enough.


jdannberg:
I agree it is sad that we will lose that name, but we can not just change the meaning of selecting `radial earth-like` in the input file because people who reuse their old input file might suddenly use a different gravity model without noticing. One alternative I could imagine where a quadratic approximation would have a place is that it uses the reference density, and under the assumption that the model is incompressible, and hence, the density is constant, computes the quadratic gravity profile that corresponds to this density. But if we do this, we might as well just add a different gravity profile that we call something like "adiabatic density integration", and that uses the density on the model's adiabatic profile to self-consistently compute a gravity profile, starting from 9.81 (or a value given in the input file) at the surface. That would be a bit more difficult to implement, but would be a lot more useful, in particular for other planets where we don't have a PREM profile. 

So I agree, I would just remove `radial earth-like`, as it's just the wrong name for what the plugin is doing, and even what it is doing is something that would have to be improved.

jaustermann:
I like Wolfgang's suggestion. I can imagine that this gravity profile is quite widely used so adding a description for why it has been changed could be quite useful (both for understanding what's going on and why results might change with the new profile). I also like your idea Juliane of an "adiabatic density integration" type profile. Maybe a task for someone at the hackathon ;)

jdannberg:
If you address @bangerth's suggestion, this should be good to go. 
And looking at the changes in the test outputs, the results are quite a bit different; as there's a difference of up to 5% between the old and the new gravity profile, I guess it makes sense to also expect that order of magnitude difference in the results. It may not change the dynamics very much, but the Rayleigh number scales with the gravity, so I think it's good that we're fixing this.

gassmoeller:
I agree that silently removing the model is not a good idea, and I implemented the assertion in the constructor. 
Another question: Should the update script automatically replace `radial earth-like` by `ascii data`? I implemented that to avoid people running into errors, but on the other hand it is a functional change, and with the new error message, maybe it is better to not replace it in input files and thus notify users about the change (by letting them see the error message)?

bangerth:
I think the latter -- let the users change this themselves.

gassmoeller:
Ok, done. I had to move the Assert into the `initialize` function though, because otherwise writing of the plugin graph fails (it creates an object, and the constructor throws an exception).

bangerth:
Looks good to me, but I would undo the changes to the plugin graph pdf since you left the plugin where it is for the moment.

gassmoeller:
Good point. I reverted the changes to the plugin graph and updated the parameters.tex to the current state. Should be good to go.

doc/manual/parameters.tex
```diff
@@ -4684,12 +4684,14 @@ \subsection{Parameters in section \tt Gravity model}
 
 `radial constant': A gravity model in which the gravity has a constant magnitude and the direction is radial (pointing inward if the value is positive). The magnitude is read from the parameter file in subsection 'Radial constant'.
 
+`radial earth-like': This plugins has been removed due to is misleading name. The included profile was hard-coded and was less earth-like than the `ascii data' plugin, which uses the profile of the Preliminary Reference Earth Model (PREM). Use `ascii data' instead of `radial earth-like'.
```
jdannberg: typo plugins --> plugin, is --> its

source/gravity_model/radial.cc
```diff
@@ -159,6 +185,14 @@ namespace aspect
                                   "is positive). The magnitude is read from the parameter "
                                   "file in subsection 'Radial constant'.")
 
+    ASPECT_REGISTER_GRAVITY_MODEL(RadialEarthLike,
+                                  "radial earth-like",
+                                  "This plugins has been removed due to is misleading name. "
```
jdannberg: here

Issue #2143, opened 2018-03-29
New test setup


gassmoeller:
As discussed with @tjhei a while ago in an email chain, CIG would like to help with setting up testers for software projects. Our current setup works well for normal development, but it is often overwhelmed during hackathons, it is not scalable, and it depends on one of the maintainer being able to run a machine with root access (for the docker containers). An earler try was Jenkins (http://blofeld.geodynamics.org/), but that got little attention, because it is not directly integrated into Github.
We are currently discussing a new approach. No decision has been reached yet, but I would like to put it out here just to let everyone know what is going on. The current plan is to keep the existing docker containers with small modifications to run the different test setups (the dockerfiles could move into our main repository), and set up a free drone server here at CIG (https://drone.io/) to start and manage the tests. Drone comes with native support for containers and integrates immediately with Github. A first glimpse of how that could look like can be seen here: https://github.com/gassmoeller/aspect/pull/9.
When we have finished the drone setup a click on 'Details' will bring you to a page similar to what we have at the moment that details what went wrong during the build (it outputs the screen output of the tester during the build). The nice thing about this approach is that we can parallelize testing different builds or prs by adding many machines as so called drone 'agents'. Also every user can easily run the test suite on their own (by running the docker containers with certain environment variables), and if at some point the drone server should go offline, it is free software and somebody else (e.g. one of the maintainers) can set up a new server. Also all of the configuration has to be part of the main repository, which makes reproducing all of this simpler.

I will update this issue as soon as we have more to show, but feel free to give feedback already.

tjhei:
> and it depends on one of the maintainer being able to run a machine with root acces

not true. I am using docker-machine and run the stuff inside virtualbox.

> earler try was Jenkins [...] it is not directly integrated into Github.

The important part is testing pull requests and posting commit status messages. This can be done using the github status plugin in jenkins, so we can use a jenkins setup. This can be the server run at Davis or mine at jenkins.tjhei.info (play ground for deal.II related testing for now).

> The nice thing about this approach is that we can parallelize testing different builds or prs by adding many machines as so called drone

The same can be done with jenkins. What is your reason to go with drone instead of jenkins? Can drone support non-linux slaves? It looks like the slave is always running inside docker so testing OSX or windows is difficult. This means for deal.II I would prefer to go the jenkins route...

Maybe we should talk about pros & cons at some point.

tjhei:
> Is the jenkins server what you plan on doing for deal.II?

[...]

PR #2154, opened 2018-04-09
Allow to run official ASPECT tester at home

gassmoeller:
This is related to #2143 and allows everybody to run the official ASPECT tester setup (currently only `aspect-8.5-gcc-hd`) on their local machine. This is super helpful if you know the tester is busy, or to check and update test results yourself if you are offline. It also allows with a little hack to `generate_reference_output` to only run certain tests you are interested in. This is a step towards having all the tester information as part of the official repository.

### DIFF
--- a/cmake/compile_and_update_tests.sh
+++ b/cmake/compile_and_update_tests.sh
@@ -1,17 +1,19 @@
 #!/bin/bash
 
-# This script can modify the test results in the source
+# This script can modify the test results in the tests/
 # directory to the official tester results, if docker is installed
-# on the system. Start this script by running the following command
-# after replacing ASPECT_SOURCE_DIR with the directory of your ASPECT
-# folder:
+# on the system. This script has to be started from the main ASPECT
+# folder (the one containing include/, cmake/, source/, and tests/).
+# Navigate to that folder and run the following command in a terminal:
 #
-# docker run -v ASPECT_SOURCE_DIR:/home/dealii/aspect --name=aspect-tester --rm -it dealii/dealii:v8.5.0-gcc-mpi-fulldepscandi-debugrelease bash /home/dealii/aspect/cmake/compile_and_update_tests.sh
+# docker run -v $PWD:/home/dealii/aspect --name=aspect-tester --rm -it dealii/dealii:v8.5.0-gcc-mpi-fulldepscandi-debugrelease bash /home/dealii/aspect/cmake/compile_and_update_tests.sh
 
 # This command executes the following shell script *inside* the docker container
 # that contains the official ASPECT test system. Note that by mounting your
 # ASPECT folder into the container you are actually changing reference test
-# results on the host system (i.e. your computer).
+# results on the host system (i.e. your computer) instead of inside the
+# container, and you also keep the build folder outside of the container in
+# the new directory 'tester-build'.
 
 echo "Starting official tester and compiling source ..."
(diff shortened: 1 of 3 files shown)


gassmoeller:
Wow, it seems the tester took this PR as a personal insult :smile:, @tjhei could you check what happened?

cmake/compile_and_update_tests.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# This script can modify the test results in the source
+# directory to the official tester results, if docker is installed
+# on the system. Start this script by running the following command
+# after replacing ASPECT_SOURCE_DIR with the directory of your ASPECT
+# folder:
+#
+# docker run -v ASPECT_SOURCE_DIR:/home/dealii/aspect --name=aspect-tester --rm -it dealii/dealii:v8.5.0-gcc-mpi-fulldepscandi-debugrelease bash /home/dealii/aspect/cmake/compile_and_update_tests.sh
```
tjhei: would it make sense to ask the user to execute this command in the source folder and then just specify ``$PWD`` here?

cmake/compile_and_update_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+# ASPECT folder into the container you are actually changing reference test
+# results on the host system (i.e. your computer).
+
+echo "Starting official tester and compiling source ..."
+
+SRC_PATH=`dirname $0`
+SRC_PATH=`cd $SRC_PATH/..;pwd`
+
+mkdir -p ${SRC_PATH}/tester-build
+cd ${SRC_PATH}/tester-build
```
tjhei: you are putting the build output and temporary files in the mounted folder, meaning they will be left around after running the container, right? This mean rerunning the compilation in the container can be fast, but you are not getting a clean build...

cmake/compile_and_update_tests.sh
gassmoeller: I do not have a preference. Both is possible. In the current form it requires some user action, but it allows running it from everywhere. Do you have a preference?

cmake/compile_and_update_tests.sh
gassmoeller: True, not sure what is better. Personally, I often rerun the tester multiple times fixing a bug or a test at a time and a clean compilation every time would take forever. So the current state is how I would use it, but I am not sure if this is something everybody would do. Alternatively, we could introduce an environment variable like `ASPECT_TESTER_CLEAN_BUILD` that can be set when running the docker container to force a clean build.

tjhei:
> "Wow, it seems the tester took this PR as a personal insult"

I triggered a retest and she reluctantly tested it now. ;-)

cmake/compile_and_update_tests.sh
tjhei: I prefer something copy&paste-able, but I don't feel strongly about it.

cmake/compile_and_update_tests.sh
tjhei: If this works reliably for you, then we should leave it as is.

gassmoeller:
I made the change to require running it from the source directory and extended the documentation a bit. Are you ok with the changes?

tjhei:
Looks good!

gassmoeller:
Thanks!

PR #2156, opened 2018-04-10
Clean Simulator::Simulator()

gassmoeller:
I am trying to work on something else, but this just keeps getting in the way and producing subtle bugs (and I think it was broken before already, because the velocity boundary ids are used for a check, before the velocity boundary manager has parsed its parameters). It also makes reading the constructor of Simulator a pain. It should really be a helper function. Btw, the Asserts I removed in core.cc are superfluous, that is why I did not bother to reimplement them: If we ask the boundary temperature manager for boundary ids on which boundary conditions are active it is the manager's responsibility to make sure the returned values are really boundary ids from the current geometry, not the responsibility of the calling function.

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1394,17 +1394,16 @@ namespace aspect
 
       /**
        * This function checks that the user-selected boundary conditions do not
-       * contain contradictions. If an incorrect
-       * selection is detected it throws an exception. This for example assures
-       * that not both velocity and traction boundary conditions are prescribed
-       * at the same boundary, and that no boundary temperatures are prescribed
-       * at a periodic boundary.
+       * contain contradictions. If an incorrect selection is detected it
+       * throws an exception. This for example assures that not both velocity
+       * and traction boundary conditions are prescribed at the same boundary,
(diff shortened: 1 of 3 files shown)


include/aspect/simulator.h
```diff
@@ -1385,10 +1385,27 @@ namespace aspect
...
+       * This function is implemented in
+       * <code>source/simulator/helper_functions.cc</code>.
        */
       void
       check_consistency_of_formulation ();
 
+      /**
+       * This function checks that the user-selected boundary conditions do not
+       * contain contradictions. If an incorrect
+       * selection is detected it throws an exception. This for example assures
```
naliboff: Remove extra spaces.

source/simulator/helper_functions.cc
```diff
@@ -1812,6 +1812,232 @@ namespace aspect
...
+      return false;
+    }
+  }
+
+
+
+  template <int dim>
+  void
+  Simulator<dim>::check_consistency_of_boundary_conditions()
+  {
```
naliboff: This function should be constant

PR #2159, opened 2018-04-12
[bugfix] fix average density implementation for boundary fluid pressure

jdannberg:
I don't know how we missed that in the melt solver pull request, but apparently in my rebase some lines got messed up and the new "average density" option for the boundary fluid pressure did not work. This pull request fixes this and also adds a test.


gassmoeller:
So the functionality was merged, but there was no way to activate it in the input file? Interesting. Anyway, looks good. Ready to merge if the tester is happy.

PR #2160, opened 2018-04-12
fix melt global cookbook

jdannberg:
We made several changes to how melt works in the last few weeks (both the operator splitting, and the melt solver), and I forgot to update the melt global cookbook accordingly, so it didn't run any more. Also, we introduced a bug in the melt global model in #2106. This pull request should fix both.


cookbooks/global_melt.prm
```diff
@@ -29,14 +29,14 @@ set Number of cheap Stokes solver steps    = 100
 # time scale than the flow of melt, so we want to decouple the advection
 # of melt (and temperature) and the melting process itself. To do that, 
 # we use the operator splitting scheme, and define that for every 
-# advection time step, we want to do at least 50 reaction time step, 
-# and in addition, we limit the maximum reaction time step size to 
-# 500,000 years. 
+# advection time step, we want to do at least at least 10 reaction time steps,
```
naliboff: "at least at least" -> "at least"

naliboff: Also, it might be cleaner to breakup this sentence. The second sentence could read "If the advection time step is larger than 10,000 years, one should do more reaction time steps".

In the sentence above, I am assuming the user would need to adjust the number of reaction time steps if the advection time step is larger than 10,000 years.

doc/manual/cookbooks/global_melt/global_melt.prm
```diff
@@ -21,14 +21,14 @@ set Nonlinear solver tolerance             = 1e-5
 # time scale than the flow of melt, so we want to decouple the advection
 # of melt (and temperature) and the melting process itself. To do that, 
 # we use the operator splitting scheme, and define that for every 
-# advection time step, we want to do at least 50 reaction time step, 
-# and if these time steps would be larger than 100,000 years, we will do 
+# advection time step, we want to do at least 10 reaction time steps, 
+# and if these time steps would be larger than 10,000 years, we will do
```
naliboff: If you like the suggest rewording in the parameter file, consider rewording here as well. Completely up to you.

naliboff:
@jdannberg - New text looks good, thanks!

PR #2161, opened 2018-04-12
Implemented clone() method in chunk and ellipsoidal_chunk geometries

lkaratun:
Implemented clone() method in chunk and ellipsoidal_chunk geometries (they are now pure virtual in the parent class)

### DIFF
--- a/source/geometry_model/chunk.cc
+++ b/source/geometry_model/chunk.cc
@@ -42,6 +42,13 @@ namespace aspect
     {}
 
     template <int dim>
+    Chunk<dim>::ChunkGeometry::ChunkGeometry(const ChunkGeometry &other)
+      :
+      ChartManifold<dim,dim>(other),
+      point1_lon(other.point1_lon)
+    {}
+
+    template <int dim>
     DerivativeForm<1,dim,dim>
     Chunk<dim>::ChunkGeometry::
     push_forward_gradient(const Point<dim> &chart_point) const
@@ -166,6 +173,16 @@ namespace aspect
     {
       point1_lon = p1_lon;
     }
+
+#if DEAL_II_VERSION_GTE(9,0,0)
+    template <int dim>
+    std::unique_ptr<Manifold<dim,dim> >
+    Chunk<dim>::ChunkGeometry::
+    clone() const
+    {
+      return std_cxx14::make_unique<ChunkGeometry>(*this);
+    }
+#endif
 
 #if !DEAL_II_VERSION_GTE(9,0,0)
     template <int dim>
@@ -540,7 +557,6 @@ namespace aspect
 
       return true;
     }
-
 
     template <int dim>
     void
--- a/source/geometry_model/ellipsoidal_chunk.cc
+++ b/source/geometry_model/ellipsoidal_chunk.cc
@@ -43,7 +43,7 @@ namespace aspect
      * the EllipsoidalChunkGeometry class
      */
 
-    // constructor
+    // Constructor
     template <int dim>
     EllipsoidalChunk<dim>::EllipsoidalChunkGeometry::EllipsoidalChunkGeometry()
       :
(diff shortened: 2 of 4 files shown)


source/geometry_model/chunk.cc
```diff
@@ -167,6 +167,14 @@ namespace aspect
       point1_lon = p1_lon;
     }
 
+    template <int dim>
+    std::unique_ptr<Manifold<dim,dim>>
+    Chunk<dim>::ChunkGeometry::
+    clone() const
+    {
+        return std_cxx14::make_unique<ChunkGeometry>();
```
bangerth: That satisfies the compiler, but it's not a copy of the current object. You'll have to somehow set the contents of the chunk geometry object to be a copy of the current object.

source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -184,6 +184,13 @@ namespace aspect
       return push_forward_ellipsoid (push_forward_topography(chart_point), semi_major_axis_a, eccentricity);
     }
 
+    template <int dim>
+    std::unique_ptr<Manifold<dim,3>>
+    EllipsoidalChunk<dim>::EllipsoidalChunkGeometry::clone() const
+    {
+        return std_cxx14::make_unique<EllipsoidalChunkGeometry>();
```
bangerth: same here

lkaratun:
I did another attempt, please have a look @bangerth

source/geometry_model/chunk.cc
```diff
@@ -41,6 +41,13 @@ namespace aspect
       point1_lon(0.0)
     {}
 
+    template <int dim>
+    Chunk<dim>::ChunkGeometry::ChunkGeometry(const ChunkGeometry &other, const double para_point1_lon)
+      :
+      ChartManifold<dim,dim>(other),
+      point1_lon(para_point1_lon)
```
bangerth: Couldn't you omit the second argument to the constructor and simply write
```
template <int dim>
￼ +    Chunk<dim>::ChunkGeometry::ChunkGeometry(const ChunkGeometry &other, const double para_point1_lon)
￼ +      :
￼ +      ChartManifold<dim,dim>(other),
￼ +      point1_lon(other.para_point1_lon)
```
? This way, you don't have to pass all of those arguments by hand. Same for the other class that has even more of these arguments.

lkaratun: Yes, I realized it right after I pushed the changes =( Fixed now

MFraters:
Looks good to me! Please run `make indent` and squash it into logical commit blocks.

lkaratun:
I actually ran `make indent` before pushing the changes. Got a warning regarding astyle version, but I thought it still worked. Can I not use astyle 3.1 with Aspect?

MFraters:
Unfortunately the version is important with astyle, you will need version 2.04

lkaratun:
astyle 2.04 doesn't seem to work: I'm getting a lot of `_: dos_to_unix: command not found` errors

bangerth:
If you follow the link to the failed astyle test (here: http://www.math.clemson.edu/~heister/cib/aspect-astyle-hd/05d5986d4dacd13019a0e260740c1191dac07971/), then you see a diff file that shows the differences between your indentation and the one we want.You can apply that by hand as well.

(I'll note that astyle 2.04 didn't know C++11, and it gets quite confused by `std::unique_ptr<Something<dim>>`. If you want a nicer indentation, write this in the pre-C++11 style: `std::unique_ptr<Something<dim> >` -- note the space between `> >`.)

lkaratun:
Does the checker want the lines marked with minuses or pluses? `a` and `b` aren't very descriptive names to figure out which is which =/
If it's the minuses, then I'm not sure what it's unhappy about -- both inspecting the file in the `Files changed` tab and in my text editor shows that all lines are equally indented.

P.s. astyle didn't work even after I inserted the spaces in the `std::unique_ptr`

lkaratun:
@gassmoeller it quite confusing because after the previous failed check I thought it wanted the lines marked with minuses, and made the corresponding  changes, and the checker is happy about all but one of those changes now. And my editor was set to spaces, 2 per indentation level, which is I think similar to the rest of the Aspect code. 
Anyway, I ran the `patch `command as you suggested, it's much easier than editing by hand, thank you.

gassmoeller:
Hmm, not sure what happened then. Do you by chance make the modifications on a windows system? That could introduce problems with the line endings, but apart from that I am not sure what could cause the issues. Anyway, the patch should fix it.

gassmoeller:
Thanks for the fix!

Issue #1969, opened 2017-10-27
Rename IMPES scheme to something more accurate


gassmoeller:
As @bangerth pointed out in https://github.com/geodynamics/aspect/pull/1850#discussion_r147316231 IMPES is actually no longer an accurate name for our default solver scheme. Since we are reorganizing things for ASPECT 2.0 anyway, find a better name and replace the use of IMPES (we can allow legacy use of the term in parameter files, but remove it from the source code).

bangerth:
We should really address this before the 2.0 release, though I don't know what better name we may want to choose.

We could just grandfather the name and carefully document what the name *actually* corresponds to today. What do you all think?

gassmoeller:
Closed by #2162

PR #2157, opened 2018-04-10
Remove model settings subsection

gassmoeller:
This is a rework of #2105, but without its dependencies on #2051 and #2099. It only depends on #2156, which should be easy to merge. It leaves all of the class structure intact, and only moves parameters in different subsections. The most complicated part is really the parameter update script.

PR #2162, opened 2018-04-12
rename IMPES and iterated IMPES

jdannberg:
This addresses #1969. 
This pull request renames the nonlinear solver scheme `IMPES` to `none` (my thought behind this is, if we set the nonlinear solver scheme to 'none', we just don't do any nonlinear iterations, which is exactly what happens); and it renames `iterated IMPES` to `iterated Advection and Stokes` (which is in line with the naming scheme for the other solver schemes, like iterated Stokes). 
In addition, I updated all the old documentation that referred to IMPES.

The old parameter can still be used in the input file, but I added a line to the documentation that it is deprecated (do we have a better way to do that?).


bangerth:
I'm not sure I like these names better. What if we rename "IMPES" to "advection then Stokes" or "advection and Stokes"? And "iterated IMPES" to "iterate advection and Stokes"?

Good finds with all of the typo fixes as well!

include/aspect/simulator.h
```diff
@@ -1423,8 +1422,8 @@ namespace aspect
 
       /**
        * This routine computes the initial (nonlinear) Stokes residual that is
-       * needed as a convergence criterion in models with the iterated IMPES
-       * solver. We calculate it in the same way as the tolerance for the linear
+       * needed as a convergence criterion in models with iterative solver
```
gassmoeller: "solver schemes that do nonlinear iterations"?

tests/diffusion.prm
```diff
@@ -92,11 +92,11 @@ set Use years in output instead of seconds = false  # default: true
 
 # Select one of the following solver schemes:
 #
-# 'IMPES': This is the IMplicit Pressure Explicit Saturation scheme. In a first
+# 'none': This is the IMplicit Pressure Explicit Saturation scheme. In a first
```
gassmoeller: Please update or remove

tests/diffusion_velocity.prm
```diff
@@ -92,11 +92,11 @@ set Use years in output instead of seconds = false  # default: true
 
 # Select one of the following solver schemes:
 #
-# 'IMPES': This is the IMplicit Pressure Explicit Saturation scheme. In a first
+# 'none': This is the IMplicit Pressure Explicit Saturation scheme. In a first
```
gassmoeller: same here

tests/postprocess_iterations_iterated_advection_and_stokes/solution.visit
```diff
@@ -0,0 +1,13 @@
+!NBLOCKS 1
```
gassmoeller: Please remove this file

MFraters:
Thanks for looking into this! It is good that it is decided before aspect 2.0 :) 

I agree with you that it is not a nonlinear solver scheme, and that therefore `none` would make sense, but I have two two problems with that name. 

Firstly, I think renaming `IMPES` to `none` might be not intuitive, because we also have the `stokes only` which is also a linear solver scheme. So we might want to think of renaming that too then. 

Secondly, this generates a large difference between how it is called in the input file and manual and how it is called in the code itself, which might not help the readability of the code. This last point could of course be solved by renaming everything internally to `none`, but that would not solve my first point.

So I think that I would prefer the option of renaming `IMPES` to `advection and Stokes`.

jdannberg:
To explain the choice of names a bit; the solver schemes we have now (other than the IMPES schemes) are called:

- `iterated Stokes`
- `Newton Stokes`
- `Stokes only`
- `Advection only`

So @bangerth if we want to call it `iterate advection and Stokes` instead of `iterated Advection and Stokes` that would not be consistent with the other schemes (we always call it iterate**d**, we always capitalize advection). That is why I chose these names, they are consistent with the other names we already have. 

To address @MFraters comments: the `Stokes only` solver scheme is a nonlinear solver scheme, it does not solve the advection equations, but it does nonlinear loops over the Stokes equation. So in this sense it might even be confusing to call it `Advection and Stokes` because that does not make it clear if it does iterations or not (or alternatively, if you all think that's a good idea, I could rename Stokes only to iterated Stokes only. I don't know if that makes it more clear what's actually going on). 
Second, if we really want to call it `Advection and Stokes`, we would also have to think about renaming the `Newton Stokes` to `iterated Advection and Newton Stokes`, and `iterated Stokes` to `Advection and iterated Stokes` because it also solves the advection equations. The naming scheme we have so far basically implicitly assumes that if something is not mentioned, and the name does not contain _only_, then it is solved once. 
Finally, I thought `none` is a really short name, so it is easy to use, and I feel it also makes it clear what happens (more so than `Advection and Stokes`). 

For the second point, I am happy to change the function in the code, I thought about renaming it, but couldn't find a good name. The code where it is called from, in core.cc, looks like this:

```
        case NonlinearSolver::none:
        {
          solve_advection_and_stokes();
          break;
        }
```
So I felt renaming it to none in this context wouldn't have made a lot of sense.

bangerth:
On 04/13/2018 10:35 AM, Juliane Dannberg wrote:
> So @bangerth <https://github.com/bangerth> if we want to call it 
> |iterate advection and Stokes| instead of |iterated Advection and 
> Stokes| that would not be consistent with the other schemes (we always 
> call it iterate*d*, we always capitalize advection). That is why I chose 
> these names, they are consistent with the other names we already have.

Oh, that's fine with me!

jdannberg:
Alright, so I guess we all agree on changing `iterated IMPES` to `iterated Advection and Stokes`. 

What about the `none` vs. `Advection and Stokes`? Does it make more sense now with my explanation why I chose that name?

gassmoeller:
I like `none`. It would be confusing to have a line like this in the input file: `set Nonlinear solver scheme = Advection and Stokes`, but no nonlinear iterations happen. Also the discrepancy in the code would not be too bad. It reads like this:
```
case NonlinearSolver::none:
  {
    solve_advection_and_stokes();
    break;
  }
```
That is still pretty self-explanatory.
I would vote for `none`.

tjhei:
None is also confusing. How about "linear" or "linearized" or "single evaluation" or something?

naliboff:
How about `single Advection and Stokes`? This way, the change in name is mini compared to `iterated Advection and Stokes`?

bangerth:
Or "one Advection and Stokes step". "None" indeed is confusing.

jdannberg:
So if you all don't like `none`, I think the next best solution for me would be something along the lines what @tjhei  suggested. So something like `single linear solve`.

gassmoeller:
@jdannberg, @naliboff and I had a discussion about the suggested alternatives. I think we all agree that the current names are inconsistent, and that makes it hard to find good names for renaming individual schemes. Our preferred solution we came up with is a renaming of all schemes in a consistent way. They would look like this:
```
IMPES -> single Advection single Stokes
Stokes only -> no Advection iterated Stokes
iterated Stokes -> single Advection iterated Stokes
Newton Stokes -> iterated Advection Newton Stokes
iterated IMPES -> iterated Advection iterated Stokes
Advection only -> single Advection no Stokes
```

This would catch all current solver schemes, and most of the new schemes people could come up with. Additionally, we can further support the old names for a few years to make transition simpler, and we can still allow special names if somebody comes up with a solver scheme that is not covered by this system. 
Would this be something everyone can agree on?

bangerth:
I like the general idea, but I think it's also not quite clear because, at least as far as I recall, `iterated IMPES` is actually "iterate out (advection, then Stokes)", not "(iterate out advection), then (iterate out Stokes)", right? `iterated Advection iterated Stokes` seems to suggest the latter.

You could make that clearer if you used commas. Maybe something like this:
```
IMPES -> single Advection, single Stokes
Stokes only -> no Advection, iterated Stokes
iterated Stokes -> single Advection, iterated Stokes
Newton Stokes -> iterated Advection, Newton Stokes
iterated IMPES -> iterated Advection and Stokes           ***
Advection only -> single Advection, no Stokes
```

MFraters:
> To address @MFraters comments: the `Stokes only` solver scheme is a nonlinear solver scheme

ah, yes ofcource. Also for the second point, I should have read the code changes more closely. 

I like the complete renaming, although the Newton one isn't clear yet. If I understand the new naming correctly it should be `iterated Advection itterated Newton Stokes` or `iterated Advection and Newton Stokes`, since the structure is similar to `iterated IMPES`

jdannberg:
Yes, I agree with @MFraters. So to summarize, the new names would be:
```
IMPES -> single Advection, single Stokes
Stokes only -> no Advection, iterated Stokes
iterated Stokes -> single Advection, iterated Stokes
Newton Stokes -> iterated Advection and Newton Stokes
iterated IMPES -> iterated Advection and Stokes
Advection only -> single Advection, no Stokes
``` 
Let me know if everyone agrees with that so that I can make the changes accordingly.

bangerth:
On 04/16/2018 10:25 AM, Juliane Dannberg wrote:
> |Newton Stokes -> iterated Advection and Newton Stokes|

I should know this but I don't: with |Newton Stokes|, do we really go 
back and solve the advection equation again once we're done with the 
Newton method on velocity/pressure? Or do we only do things only once?

jdannberg:
As @MFraters said, we iterate over both advection and Stokes, so the loop looks something like this:

```
for (nonlinear_iteration = 0; nonlinear_iteration < max_nonlinear_iterations; ++nonlinear_iteration)
      {
        assemble_and_solve_temperature();
        assemble_and_solve_composition();
        assemble_and_solve_newton_stokes();
      }
```

naliboff:
Is there any advantage in terms of parsing to making the naming scheme even more "formatted"? @jdannberg I'm thinking of our conversations with @gassmoeller where we discussed options for users easily mixing and matching different schemes.

A slightly simplified version of the naming scheme above could be:
```
IMPES -> single Advection, single Stokes
Stokes only -> no Advection, iterated Stokes
iterated Stokes -> single Advection, iterated Stokes
Newton Stokes -> iterated Advection, Newton Stokes
iterated IMPES -> iterated Advection, iterated Stokes
Advection only -> single Advection, no Stokes
```

Here, the advection and Stokes components are always separated by a comma and can have the values `no`, `single`,` iterated` or `Newton`. 

Probably only worth considering if this would be useful in terms of parsing and functionality added later on.

jdannberg:
@naliboff 
Yes, I think that was what we originally discussed, but @bangerth noted correctly that this would be confusing, because `iterated Advection, iterated Stokes` sounds like we're first iterating out the nonlinearity in the advection system and then in the Stokes system, not alternating between them. 

To make it clear for everyone what discussion @naliboff mentios here: @gassmoeller had the idea that we could read in the solver schemes not as one given name for a solver scheme, but as different keywords, so that users could mix and match (similarly to how you can mix and match formulations if you want to do that). This would not be something to be done for Aspect 2.0, but is just an idea for later (and would be backwards compatible). So for example for IMPES, one could split the string at the comma, and then read in the option for the advection solver scheme (single) and the Stokes solver scheme (single). In the code, this could look like

```
if (advection_scheme == single)
      {
        assemble_and_solve_temperature();
        assemble_and_solve_composition();
      }

if (stokes_scheme == single)
      assemble_and_solve_stokes();
```

In this way, adding new custom solver schemes would be easier. 
But I think we still have the option to do this even if we call it `iterated Advection and Newton Stokes` and `iterated Advection and Stokes`, we just would have to write the parsing algorithm accordingly.

bangerth:
@jdannberg -- thanks for confirming!

bangerth:
You could make the parsing algorithm more concise if you allowed parentheses: `iterated (Advection and Stokes)` is unambiguous. So would be `iterated(Advection) and Stokes`, which one could abbreviate as `iterated advection and Stokes`. In other words, `iterated` applies to the following item, which may be a single word or a parenthesized expression.

This is all a bit like regular expressions.

jdannberg:
@bangerth Yes, we thought about this too. But I also think it should be easy for beginners to just select the solver scheme they want, and for this purpose it would be much easier to have simple, obvious names without parenthesis for the existing solver schemes. So if we decide to change the parsing algorithm to read in keywords instead of the complete name later on, we should consider adding (optional) parenthesis to make the schemes unambiguous. 
At the moment, I think we should concentrate on finding good names for the existing solver schemes, so that we can get the release done before the end of the month.

bangerth:
I agree with the last two comments.

jdannberg:
Okay, great! Thanks for the discussion! I'll make the changes tomorrow.

gassmoeller:
I agree with the conclusion, lets take @bangerth's and @jdannberg's version from here: https://github.com/geodynamics/aspect/pull/2162#issuecomment-381664513

jdannberg:
I think I will need some help with the update script: 
If we automatically want to update, e.g. `iterated Stokes` to `single Advection, iterated Stokes`, it still has iterated Stokes in its name, so if one would run the update scrip a second time, it would then rename `single Advection, iterated Stokes` to `single Advection, single Advection, iterated Stokes`. @gassmoeller, do you have a way you usually address this? 
The only workaround I can think of is first renaming all occurrences of `single Advection, iterated Stokes` back to `iterated Stokes` (and the same for `no Advection, iterated Stokes`, which also has iterated Stokes as part of its new name), and then replacing all occurrences of `iterated Stokes` to `single Advection, iterated Stokes`. So something like:

```
s/single Advection, iterated Stokes/iterated Stokes/g
s/iterated Stokes/single Advection, iterated Stokes/g
```
Any better ideas?
(Its also the same for `Newton Stokes`)

doc/modules/changes/20180417_jdannberg
```diff
@@ -0,0 +1,10 @@
+Changed: All of the nonlinear solver schemes have been renamed:
+IMPES -> single Advection, single Stokes
+Stokes only -> no Advection, iterated Stokes
+iterated Stokes -> single Advection, iterated Stokes
+Newton Stokes -> iterated Advection and Newton Stokes
+iterated IMPES -> iterated Advection and Stokes
+Advection only -> single Advection, no Stokes
+The old names still work, but should be considered as deprecated. 
+<br>
+(Juliane Dannberg, 2018/04/06)
```
naliboff: Please fix the date.

source/simulator/parameters.cc
```diff
@@ -160,29 +162,46 @@ namespace aspect
                        "This parameter indicates whether the simulator should also use "
                        "heat conduction in determining the length of each time step.");
 
-    prm.declare_entry ("Nonlinear solver scheme", "IMPES",
-                       Patterns::Selection ("IMPES|iterated IMPES|iterated Stokes|Newton Stokes|Stokes only|Advection only"),
+    prm.declare_entry ("Nonlinear solver scheme", "single Advection, single Stokes",
+                       Patterns::Selection ("single Advection, single Stokes|iterated Advection and Stokes|single Advection, iterated Stokes|no Advection, iterated Stokes|iterated Advection and Newton Stokes|single Advection, no Stokes|IMPES|iterated IMPES|iterated Stokes|Newton Stokes|Stokes only|Advection only"),
```
gassmoeller: You can make this easier to read, if you specify the selection string above as follows:
```
const std::string allowed_solver_schemes = "single Advection, single Stokes|"
                                                       "iterated Advection and Stokes|"
 .... 
                                                       ;

and then use `Patterns::Selection(allowed_solver_schemes)`

File: source/simulator/parameters.cc [hunk 7]
```diff
@@ -160,29 +162,46 @@ namespace aspect
...
+                       "The `iterated IMPES' scheme is deprecated and only allowed for reasons of "
+                       "backwards compatibility, it is the same as `iterated Advection and Stokes'. "
+                       "The `iterated Stokes' scheme is deprecated and only allowed for reasons of "
+                       "backwards compatibility, it is the same as `single Advection, iterated Stokes'. "
+                       "The `Stokes only' scheme is deprecated and only allowed for reasons of "
+                       "backwards compatibility, it is the same as `no Advection, iterated Stokes'. "
+                       "The `Advection only' scheme is deprecated and only allowed for reasons of "
+                       "backwards compatibility, it is the same as `single Advection, no Stokes'. "
+                       "The `Newton Stokes' scheme is deprecated and only allowed for reasons of "
+                       "backwards compatibility, it is the same as `iterated Advection and Newton Stokes'.");
```
naliboff: Replace comma with period after "backwards compatibility".

source/simulator/parameters.cc
```diff
@@ -998,18 +1017,30 @@ namespace aspect
...
+    else if (prm.get ("Nonlinear solver scheme") == "iterated Advection and Newton Stokes")
+      nonlinear_solver = NonlinearSolver::iterated_Advection_and_Newton_Stokes;
     else if (prm.get ("Nonlinear solver scheme") == "Newton Stokes")
-      nonlinear_solver = NonlinearSolver::Newton_Stokes;
+      nonlinear_solver = NonlinearSolver::iterated_Advection_and_Newton_Stokes;
+    else if (prm.get ("Nonlinear solver scheme") == "single Advection, no Stokes")
+      nonlinear_solver = NonlinearSolver::single_Advection_no_Stokes;
     else if (prm.get ("Nonlinear solver scheme") == "Advection only")
-      nonlinear_solver = NonlinearSolver::Advection_only;
+      nonlinear_solver = NonlinearSolver::single_Advection_no_Stokes;
```
naliboff: Can you use "or" statements instead of so many "if" statements?

benchmarks/solitary_wave/solitary_wave.cc
```diff
@@ -368,13 +368,13 @@ namespace aspect
         {
           // Note that this number is based on the background porosity in the
           // solitary wave initial condition.
-          const SolitaryWaveInitialCondition<dim> *initial_conditions =
+          const SolitaryWaveInitialCondition<dim> *initial_temperature =
```
jdannberg: Note to self: remove this change, this is resolved differently in  #2157.

doc/manual/parameters.tex
```diff
@@ -12129,6 +12147,22 @@ \subsection{Parameters in section \tt Material model/Visco Plastic}
 {\it Description:} List of cohesions, $C$, for background material and compositional fields, for a total of N+1 values, where N is the number of compositional fields. The extremely large default cohesion value (1e20 Pa) prevents the viscous stress from exceeding the yield stress. Units: $Pa$.
 
 
+{\it Possible values:} A list of 0 to 4294967295 elements where each element is [A floating point number $v$ such that $0 \leq v \leq \text{MAX\_DOUBLE}$]
```
jdannberg: Note to self: remove this subsection.

source/simulator/parameters.cc
```diff
@@ -162,8 +162,9 @@ namespace aspect
                        "This parameter indicates whether the simulator should also use "
                        "heat conduction in determining the length of each time step.");
 
+    const std::string allowed_solver_schemes = "single Advection, single Stokes|iterated Advection and Stokes|single Advection, iterated Stokes|no Advection, iterated Stokes|iterated Advection and Newton Stokes|single Advection, no Stokes|IMPES|iterated IMPES|iterated Stokes|Newton Stokes|Stokes only|Advection only";
```
naliboff: Break into multiple lines.

bangerth:
Thank you all for making this happen and for fixing my mistakes from a decade ago!

Issue #2023, opened 2017-12-01
solver parameters


MFraters:
During the last hackaton we created a parameter section for solvers parameters, so I could put my Newton parameters in there. I think the idea was that also the other solver parameters would move to there. I think it would a good idea to do this before the release of aspect 2.0, because it would again break parameters files.

tjhei:
> so I could put my Newton parameters in there.

agreed.

> I think the idea was that also the other solver parameters would move to there.

I remember talking with @jdannberg about it last week. We have to balance breaking a lot of prm files with a neater layout, though. I think at least the less common things can go into that subsection.

naliboff:
I'm working on this right now.  Here are the following "solver related" parameters that are **not** in the `Solver parameters` subsection:

- `Max nonlinear iterations`
- `Nonlinear solver scheme`
- `Nonlinear solver tolerance`
- `Use direct solver for Stokes system`
- `Linear solver tolerance`
- `Linear solver A block tolerance`
- `Linear solver S block tolerance`
- `Number of cheap Stokes solver steps`
- `Maximum number of expensive Stokes solver steps`
- `Temperature solver tolerance`
- `Composition solver tolerance`

The obvious candidates to go into the Solver parameters subsection are `Linear solver A block tolerance` and  `Linear solver S block tolerance`.  Moving these also would not require running the update script.

How about the other parameters? @jdannberg and @gassmoeller - I recall that we thought `Nonlinear solver scheme` and  `Nonlinear solver tolerance` should remain in the current (global section).

Thoughts?

jdannberg:
Fixed by #2163 and #2174.

PR #2163, opened 2018-04-12
Create a GMRES subsection in Solver parameters

naliboff:
Partially addresses #2023. Moving these two parameters does not require running the update script, but moving any of the other solver parameters would require doing so.

### DIFF
--- a/cookbooks/global_no_melt.prm
+++ b/cookbooks/global_no_melt.prm
@@ -7,9 +7,7 @@ set Dimension                              = 2
 set Adiabatic surface temperature          = 1600
 set Maximum time step                      = 1e6
 set Output directory                       = no_melt
-set Linear solver tolerance                = 1e-8
 set Use years in output instead of seconds = true
-set Number of cheap Stokes solver steps    = 100
 
 # The end time of the simulation. Because we want to see how upwellings
 # and downwellings evolve over time, and if differences develop between 
@@ -193,4 +191,11 @@ end
 # point. 
 subsection Checkpointing
   set Time between checkpoint = 1700
+end
+
+subsection Solver parameters
+  subsection Stokes solver parameters
+    set Linear solver tolerance = 1e-8
+    set Number of cheap Stokes solver steps = 100
+  end
 end
--- a/cookbooks/global_no_melt.prm
+++ b/cookbooks/global_no_melt.prm
@@ -201,4 +201,3 @@ end
 subsection Checkpointing
   set Time between checkpoint = 1700
 end
-
--- a/cookbooks/global_no_melt.prm
+++ b/cookbooks/global_no_melt.prm
@@ -14,6 +14,15 @@ set Use years in output instead of seconds = true
 # the model with and without melt migration, we set the end time to 140 Ma.
 set End time                               = 1.4e8
 
+# We choose a stricter than default linear Stokes solver tolerance,
+# to improve the nonlinear convergence behavior.
+subsection Solver parameters
+  subsection Stokes solver parameters
+    set Linear solver tolerance = 1e-8
+    set Number of cheap Stokes solver steps = 100
+  end
+end
+
 # We fix the temperature at the top and bottom boundaries of
 # the box, and prescribe free-slip boundary conditions on all
 # sides. 
@@ -193,9 +202,3 @@ subsection Checkpointing
   set Time between checkpoint = 1700
 end
 
-subsection Solver parameters
-  subsection Stokes solver parameters
-    set Linear solver tolerance = 1e-8
-    set Number of cheap Stokes solver steps = 100
-  end
-end
--- a/cookbooks/global_no_melt.prm
+++ b/cookbooks/global_no_melt.prm
@@ -15,7 +15,7 @@ set Use years in output instead of seconds = true
 set End time                               = 1.4e8
 
 # We choose a stricter than default linear Stokes solver tolerance,
-# to improve the nonlinear convergence behavior.
+# to be consistent with the global_melt cookbook.
 subsection Solver parameters
   subsection Stokes solver parameters
     set Linear solver tolerance = 1e-8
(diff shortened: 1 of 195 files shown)


jdannberg:
Keep in mind that even though we may not use any of these parameters in our tests, other people might still use them in their input file, so we should add an entry in the update script anyway. Also, I would vote for moving `Number of cheap Stokes solver steps` and `Maximum number of expensive Stokes solver steps` too. Thoughts?

naliboff:
I agree we should move `Number of cheap Stokes solver steps` and `Maximum number of expensive Stokes solver steps`. Beyond these two, I think we would need to move all of the rest as a group. For example, I don't think it would make sense to have the `Nonlinear solver scheme` and `Nonlinear solver tolerance` in different sections.

For these (and any other) parameters I don't think there is any need to use an additional subsection under `Solver parameters`. 

Once we decide on the remaining parameters to move, I'll modify the update script.

source/simulator/parameters.cc
```diff
@@ -338,6 +321,26 @@ namespace aspect
 
     prm.enter_subsection ("Solver parameters");
     {
+      prm.enter_subsection ("GMRES parameters");
```
bangerth: Could I advocate for this to be renamed "Linear solver parameters"? The fact that we use GMRES is not really material in this regard -- it's just what we happen to use -- whereas the important fact is that it's about solving the *linear* problem.

bangerth: That makes sense.

naliboff:
If we keep 'Linear solver tolerance' outside of the Solver parameters subsection, I think it makes sense to keep `Use direct solver for Stokes system` in the same place. However, If the group consensus is to move it, no real objection.

naliboff:
I've updated the subsection name to `Linear solver parameters` and moved `Number of cheap Stokes solver steps` and` Maximum number of expensive Stokes solver steps` here. 

Unless there are more parameters to move into this subsection or `Solver parameters` in general, I'll revise the update script. Many of the tests should fail before this update.

bangerth:
On 04/13/2018 02:37 PM, naliboff wrote:
> If we keep 'Linear solver tolerance' outside of the Solver parameters 
> subsection, I think it makes sense to keep |Use direct solver for Stokes 
> system| in the same place. However, If the group consensus is to move it, no 
> real objection.

I'd say move both. They are both clearly related to the linear solver.

naliboff:
@gassmoeller - The last commit addressed your two comments. Should I squash everything down to one commit before changing the update script?

gassmoeller:
I think I fixed the update script for all special cases we had. @naliboff, sorry for taking over your PR, I rebased everything after #2168 was merged, so you will need to throw away your local branch and check out this one if you want to do any more changes. Hopefully that should not be necessary though. I removed some parameter lines that just set parameters to their default value, and made sure that the benchmarks and cookbooks are still nice to read.
Time for a final review and merge?

cookbooks/global_no_melt.prm
```diff
@@ -7,15 +7,22 @@ set Dimension                              = 2
...
 set Use years in output instead of seconds = true
-set Number of cheap Stokes solver steps    = 100
 
 # The end time of the simulation. Because we want to see how upwellings
 # and downwellings evolve over time, and if differences develop between 
 # the model with and without melt migration, we set the end time to 140 Ma.
 set End time                               = 1.4e8
 
+# We choose a stricter than default linear Stokes solver tolerance,
+# to improve the nonlinear convergence behavior.
```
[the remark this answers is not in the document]
jdannberg: This would also have the advantage that the files in the doc/manual/cookbooks and the cookbooks folder have the same documentation.

doc/manual/cookbooks/benchmarks/burstedde/burstedde.prm
```diff
@@ -27,3 +26,9 @@ end
 subsection Postprocess
   set List of postprocessors = visualization, velocity statistics, BursteddePostprocessor
 end
+
+subsection Solver parameters
+  subsection Stokes solver parameters
+    set Number of cheap Stokes solver steps = 0
+  end
+end
```
jdannberg: If this was not in the original file, you don't have to add this here (in particular as there's no documentation on why the parameter is chosen like this).

doc/manual/cookbooks/benchmarks/burstedde/burstedde.prm
```diff
@@ -1,4 +1,3 @@
-set Linear solver tolerance                = 1e-12
```
jdannberg: This is not the default value, so you can not just remove it. Please add it below to the Stokes solver parameters.

gassmoeller:
Done. @tjhei: It seems the tester is taking a break, can you have a look? Thanks!

source/simulator/parameters.cc
```diff
@@ -269,77 +269,6 @@ namespace aspect
...
-                       "preconditioner.");
-
-    prm.declare_entry ("Maximum number of expensive Stokes solver steps", "1000",
-                       Patterns::Integer(0),
-                       "This sets the maximum number of iterations used in the expensive Stokes solver. "
-                       "If this value is set too low for the size of the problem, the Stokes solver will "
-                       "not converge and return an error message pointing out that the user didn't allow "
-                       "a sufficiently large number of iterations for the iterative solver to converge.");
-
     prm.declare_entry ("Temperature solver tolerance", "1e-12",
```
jdannberg: I thought we had agreed to also move the temperature and composition solver tolerance if we move the linear solver tolerance? 
I am not insisting on this (and the last time we talked about this, I had assumed none of these tolerances would be moved, that is why I argued they should stay where they are), but I just think this can be quite confusing. Why is the Linear solver tolerance in its own subsection, while the temperature and composition solver tolerances aren't? 
We can also do that in a follow-up pull request, but we should at least discuss again what the final state for the release should be.

PR #2164, opened 2018-04-13
Fix update script and add test

gassmoeller:
Since #2152 there was an ugly bug in the update_prm script that repeatedly added lines if the script was applied more than once and the `radial linear` gravity profile subsection existed in the parameter file. The fix is some ugly sed-magic, but I also added a new test that checks this behavior. The test essentially copies the first update script test, but applies the script twice. I checked that the test fails, if one of the following happens:
- diff reports a difference in the parameter file between the first and second application of the update script
- aspect can not run the twice updated version of the parameter file
- updated2.prm (the twice updated version) differs from the stored reference version in the test output

### DIFF
--- a/doc/update_prm_files_to_2.0.0.sed
+++ b/doc/update_prm_files_to_2.0.0.sed
@@ -22,7 +22,25 @@ s/subsection Initial profile/subsection Compute profile/g
 s/set Model name = initial profile/set Model name = compute profile/g
 
 # Recover previous behavior of the 'Radial linear' gravity model
-s/subsection Radial linear/subsection Radial linear\n    set Magnitude at bottom = 0.0/g
+# In the following we read the whole subsection 'Radial linear'
+# when we find it (N adds the next line), check if 
+# 'Magnitude at bottom' is inside, and insert (i) it, if not.
+:jump_before_gravity
+/subsection Radial linear/,/\bend\b/ {
+n
+/^ *end/b jump_before_gravity
+
+:jump_in_gravity
+N
+/end/!b jump_in_gravity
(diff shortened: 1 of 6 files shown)


doc/update_prm_files_to_2.0.0.sed
```diff
@@ -22,7 +22,25 @@ s/subsection Initial profile/subsection Compute profile/g
...
+N
+/end/!b jump_in_gravity
+# If we get here, we found an end, either we found the new parameter,
+# or it needs to be added. Then jump up to finish this subsection
+/set Magnitude at bottom/!{
+i\    set Magnitude at bottom = 0.0
+}
+
+b jump_before_gravity
+}
```
naliboff: Can you expand the comments above? @jdannberg and I are still having a hard time interpreting the sed commands. Mission Impossible.

tests/update_script_2.prm
```diff
@@ -0,0 +1,138 @@
+# This is a test for the .prm update script in
+# doc/update_prm_files_to_2.0.0.sed. The content of this prm
+# file is only executable by ASPECT <=1.5, but the corresponding
+# shared library in update_script.cc will first execute the update
+# script on this .prm file before running ASPECT with the updated
+# version of the file. The success of this test means the script did
+# its job and ASPECT 2.0 is able to read the updated file. This
+# test differs from update_script.prm by applying the script twice,
+# and checking that it does not add additional lines after the
+# first run.
```
naliboff: Please move this sentence to the first line of the parameter file description.

gassmoeller:
Done, I hope it is more clear now.

Issue #1950, opened 2017-10-05
ellipsoidal chunk push and pull functions


MFraters:
While working on #1941 I noticed that the push and pull function for the ellipsoidal chunk do not exactly undo each others tranformation. Although the cartesian x and y are as far as I have seen exactly recovered, the z dimension is not. 

For the test use the ellipsoidal chunk defaults and I start with a cartesian point `(25000 25000 50000)`, which after a pull_back becomes `(0.785398163 1.2159859 -6295224.83)` and after a push_forward of this it becomes `(25000 25000 55272.2164)`. A pull_back of this results in `(0.785398163 1.22986595 -6291444.98)` and a push_forward of this results in `(25000 25000 59292.7979)`, etc..

Now it this a point, very deep in the earth, and I haven't tested it for more realistic Cartesian points. The equations used in this geometry model are from deal.ii step 53, but are not explained there where they exactly come from. But the ellipsoidal to Cartesian equation (in `push_forward`) is easy to find and seems to be correct. For the conversion from Cartesian to ellipsoidal coordinates (in `pull_back`), when I look on the internet, it seems to point at that usually an iteration is needed to get an precise solution and it could mean that this implementation is just an approximation. Although the equations which I have found for Cartesian to ellipsoidal coordinates are similar, I haven't been able to find this exact equation. 

@bangerth Do you know where this equation is based on? Or maybe Sarah Stamps or Luca Heltai?

bangerth:
The `ellipsoidal chunk` model uses the WGS84 geometry for which the equations (in `EllipsoidalChunkGeometry::push_forward_ellipsoid()` and `EllipsoidalChunkGeometry::pull_back_ellipsoid`) are implemented in `source/geometry_models/ellipsoidal_chunk.cc`. The formulas for this model originally came from @dsarahstamps and I don't know where she had them from.

@dsarahstamps -- do you recall a reference for these formulas?

jdannberg:
Let's check how big the error is for points that are in the Earth's mantle, further away from the origin. If we find the error is sufficiently small, let's postpone this until after ASPECT 2.0.

PR #2165, opened 2018-04-18
Fix unused warning

gassmoeller:
Supposedly introduced in #1941.


MFraters:
hmm, yes that looks weird. The only thing I can think of is that it is left over of experimenting with the conversion when I noticed that the ellispoidal chunk was off (see issue #1950). Good catch! 👍

PR #2169, opened 2018-04-20
Fix parameters.tex and manual

gassmoeller:
Fix some issues with parameters.tex and the manual. The only noticeable change to users is the renaming in the melt material postprocessor from `p_c` to `compaction pressure` (there is already a parameter `compaction viscosity` in the same postprocessor). Since the code is only 2 weeks in master (introduced in #2138), and there does not seem to be any parameter file that uses it directly (it is in the default list, so there are some tests that use it implicitly), I think it should be fine to rename without an entry in the update script, or the changelog. @jdannberg: do you have a lot of input files that use it?


jdannberg:
The idea behind naming it `p_c` instead of compaction pressure is that this is, in principal, a solution variable of our equations (in constrast to the compaction viscosity, which is a materiel property), and before we did the overhaul of the melt solver, people would have an output quantity called p_c (just like p, T and the other solution variables) that came from the solution vector. Now, we rescale the compaction pressure, and the variable that is stored in the solution vector is p_c_bar instead of p_c. But it's only a methodical change, people can still run the same models as before, and with the name as we have it right now, they will still get an output field that is called p_c, and is the same quantity that it was before the change. So although the code is only in master for two weeks, that output quantity itself is in Aspect for two years and follows the same naming style as all the other solution variables.

gassmoeller:
So this is to let users have the same p_c output field in their paraview output as before?
Hmm, seems reasonable, then I will just fix the parameter description in the source file, instead of renaming it. Thanks for the clarification :+1:

jdannberg:
> So this is to let users have the same p_c output field in their paraview output as before?

Yes. 

> then I will just fix the parameter description in the source file

Thank you! And sorry for breaking the file :(
Let me know once you've made the changes so I can review the pull request.

gassmoeller:
Hmm, actually this fix might be hard to do. I am not sure we allow tex-incompatible values for parameters. The problem is that I can easily change the description for parameters (adding \ to escape the latex usage), but not for their values, because they directly enter the parameter handler (it would change their value). I could however change the parameter in the input file to 'compaction pressure', but add a special case for the field name for Paraview, so that the output field would still be named p_c. See my last commit for an explanation. Would that be ok?

jdannberg:
Hm, do you think that might be confusing for the user? Alternatively, we could just always add p_c to the list of property names. I think that would make a lot of sense, as it's a solution variable.

tjhei:
> Would that be ok?

I think this is fixed in deal.II master, where we are now escape everything correctly. That doesn't help us now, though.

Could you just add a ``sed`` command to ``doc/update_parameters.sh``?

gassmoeller:
Hmm, the `sed` command would not work. Users would update their parameter file, but could still compile and run with deal.II 8.5. I think @jdannberg suggestion makes sense, just always add `p_c` as output field, if the `melt` visualization postprocessor is active.

tjhei:
> Hmm, the sed command would not work. Users would update their parameter file, but could still compile and run with deal.II 8.5.

I don't understand. You are trying to fix the manual from compiling, which fails because of ``p_c`` inside parameters.tex, right? You can do any replacement in ``update_parameters.sh``, for example ``p_c``->``p\_c`` or not?

Anyways, always outputting p_c would also be okay with me.

gassmoeller:
Oh, I misunderstood. I thought you meant replacing parameters in the parameter files. Yes, replacing things with sed in parameters.tex would work as well. Anyways, now I implemented the version that always outputs p_c and that seems reasonable.
By the way, Timo do you currently have access to the tester? It does not seem to trigger tests.

tjhei:
> Timo do you currently have access to the tester?

Clemson had a power outage today and I need to wait until tomorrow for somebody to turn the machines back on...

PR #2170, opened 2018-04-21
[WIP] Fix density used for thermal conductivity in visco_plastic and diffusion_dislocation models

naliboff:
In the` visco_plastic` and `diffusion_dislocation` models the thermal diffusivity is specified as an input file parameter instead of the thermal conductivity. The thermal conductivity is then calculated from the thermal diffusivity, density and heat capacity. 

This leads to an inconsistency when set `Temperature equation = reference density profile`. 

The same inconsistency is present in any heating model that uses the material model density output. @gassmoeller @jdannberg - I was planning to use the structure to enforce that the heating models use the correct density. Any objections before I do this in another commit?

Thanks and credit to Phillip Heron and Jeroen van Hunen for identifying an issue by running the `Blankenbach` benchmarks  with these two material models!

### DIFF
--- a/source/material_model/diffusion_dislocation.cc
+++ b/source/material_model/diffusion_dislocation.cc
@@ -20,7 +20,7 @@
 
 #include <aspect/material_model/diffusion_dislocation.h>
 #include <aspect/utilities.h>
-
+#include <aspect/adiabatic_conditions/interface.h>
 
 namespace aspect
 {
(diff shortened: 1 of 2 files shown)


naliboff:
@jdannberg - I've fixed the diffusivity in all the places above. Let me know if you would like further documentation or reformatting in these sections.

PR #2174, opened 2018-04-24
Move advection solver tolerances

### LINKED ISSUE #2171: Move temperature and composition solver tolerance into the Solver settings subsection
This would be a follow-up for #2163. I think it would be confusing for beginners, who may use the ASPECT gui, to find the composition and temperature solver tolerances in the global parameters, but not the linear (Stokes) solver tolerance. So I think they should be moved into the Solver settings subsection too (but they do not necessarily need their own subsection, they can just parameters in that section).

gassmoeller:
Fixes #2171 and #2023.

### DIFF
--- a/benchmarks/king2dcompressible/ala.prm
+++ b/benchmarks/king2dcompressible/ala.prm
@@ -72,7 +72,6 @@ subsection Discretization
   end
 end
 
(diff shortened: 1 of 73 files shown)


jdannberg:
Ready to merge when the tester is finished.

PR #2175, opened 2018-04-24
Add test for ellipsoidal chunk manifold conversion

### LINKED ISSUE #1950: ellipsoidal chunk push and pull functions

gassmoeller:
This closes  #1950. While the manifold conversion can produce significant errors very close to the center (e.g. 10% at a distance of 50 km from the center as mentioned in #1950), it is already negligible at the inner-core boundary (1e-7 at 1000km radius), and completely insignificant at the CMB (5e-11 at 2890 km depth).

### DIFF
--- a/tests/ellipsoidal_chunk_geometry.cc
+++ b/tests/ellipsoidal_chunk_geometry.cc
@@ -1,0 +1,95 @@
+#include <aspect/geometry_model/ellipsoidal_chunk.h>
+#include <aspect/geometry_model/initial_topography_model/zero_topography.h>
+
+#include <deal.II/base/exceptions.h>
+
+#include <iostream>
+
+using namespace aspect;
+using namespace dealii;
+
+bool test_point(const GeometryModel::EllipsoidalChunk<3>::EllipsoidalChunkGeometry ellipsoidal_manifold,
+                const Point<3> &test_point)
+{
+  const Point<3> converted_point = ellipsoidal_manifold.pull_back(test_point);
+  const Point<3> twice_converted_point = ellipsoidal_manifold.push_forward(converted_point);
+
+  const Tensor<1,3> difference = twice_converted_point - test_point;
+  const double error = difference.norm() / test_point.norm();
+
+  std::cout << "Point: " << test_point
+            << " becomes: " << converted_point
+            << ". Back transformed: " << twice_converted_point
+            << ". Error: " << error << std::endl;
+
+  return true;
+}
+
+/*
+ * Launch the following function when this plugin is created.
+ */
+int f()
+{
+
+  const unsigned int dim=3;
+
+  InitialTopographyModel::ZeroTopography<dim> topography;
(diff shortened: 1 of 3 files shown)


tests/ellipsoidal_chunk_geometry.cc
```diff
@@ -0,0 +1,98 @@
...
+  test_points.push_back(Point<3> (6371000.0,0,0));
+  test_points.push_back(Point<3> (6171000.0,0,0));
+  test_points.push_back(Point<3> (3000000.0,3000000.0,0));
+  test_points.push_back(Point<3> (3000000.0,3000000.0,3000000.0));
+  test_points.push_back(Point<3> (3000000.0,-3000000.0,-3000000.0));
+  test_points.push_back(Point<3> (25000.0,25000.0,3481000.0));
+  test_points.push_back(Point<3> (25000.0,25000.0,1000000.0));
+  test_points.push_back(Point<3> (25000.0,25000.0,500000.0));
+  test_points.push_back(Point<3> (25000.0,25000.0,50000.0));
+
```
MFraters: Can you for the sake of clarity in the output put a line in there which outputs something like "simple sphere test:". And also do that for the others?

tests/ellipsoidal_chunk_geometry.cc
```diff
@@ -0,0 +1,98 @@
...
+                                               eccentricity,
+                                               semi_minor,
+                                               500000.0,
+                                               corners);
+
+  for (unsigned int i=0; i<test_points.size(); ++i)
+    test_point(ellipsoidal_manifold,test_points[i]);
+
+
+  std::cout << "Ok" << std::endl;
```
MFraters: This line seems a bit useless, because no real check is done. Even if the output would not be oke, it would still output ok, which is a bit strange.

gassmoeller:
Thanks, I addressed the comments.

PR #2177, opened 2018-04-25
fix temperature dependence of compaction viscosity in melt simple

jdannberg:
This fixes the temperature dependence of the compaction viscosity in the melt simple material model in models without adiabatic heating (before, the computation just used the same parameter that is used for the shear viscosity, even though there is a separate parameter that allows it to prescribe the temperature dependence of the compaction viscosity).


gassmoeller:
Thanks for the fix. Please add a test and a changelog entry (this was a bug since the very first melt version).

PR #2180, opened 2018-04-27
Fix remaining cookbook that uses p_c in the input file

### LINKED ISSUE #2179: p_c is still used in some of the cookbooks
This causes problems because the files don't run any more after the changes introduced in #2169.

gassmoeller:
Closes #2179. It was only this cookbook (I checked cookbooks, benchmarks, and files in doc/), and the parameter existed in the main code for less than two weeks, so I do not think there is a need for an inclusion into the update script.

### DIFF
--- a/cookbooks/global_melt.prm
+++ b/cookbooks/global_melt.prm
@@ -347,7 +347,7 @@ subsection Postprocess
     end
 
     subsection Melt material properties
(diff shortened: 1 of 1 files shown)


jdannberg:
Okay, that was easy! Thanks!
Ready to merge when the tests pass.

PR #2187, opened 2018-04-27
Make some expressions simpler to read.

bangerth:
The current formulas are just very long and prone to run past the end
of the screen. Break them in a more sensible and readable way.


gassmoeller:
Good to go. Thanks.

PR #2189, opened 2018-04-27
Adjust default vs vp values

### LINKED ISSUE #1562: Default values for seismic vp and vs
At the moment, the vs and vp material properties (which are used for postprocessing) are initialized with a value of -1, and not with signalingNaNs, which would make more sense (if we want to output seismic velocities, we should be forced to fill them). 
However, changing the default value breaks approximately 50 tests. We should have a look what's going on there and think about if we can change the default.

gassmoeller:
Closes #1562. Lets see how many tests break.


doc/modules/changes/20180427_gassmoeller
```diff
@@ -0,0 +1,5 @@
+Changed: The default value for AdditionalSeismicOutputs was changed from -1 to
+a signaling Nan. This ensures every plugin that creates seismic outputs needs
+to use a material model that actually fills those outputs.
+<br>
+(Rene Gassmoeller, 2018/04/27)
```
naliboff: Would it be worth adding this documentation to the `seismic_anomalies` post-processor description? Alternatively, is there already an assert that produces a similar error message when the values are filled with Nan?

source/postprocess/visualization/seismic_anomalies.cc
```diff
@@ -405,6 +405,19 @@ namespace aspect
...
       {
+        // First make sure the material model actually provides seismic velocities
+        MaterialModel::MaterialModelOutputs<dim> out(1, this->n_compositional_fields());
+        this->get_material_model().create_additional_named_outputs(out);
+
+        const bool material_model_provides_seismic_output =
+          (out.template get_additional_output<MaterialModel::SeismicAdditionalOutputs<dim> >() != 0);
+
+        AssertThrow(material_model_provides_seismic_output,
+                    ExcMessage("You requested the 'Seismic' postprocessor, "
```
jdannberg: please correct the name of the postprocessor

source/postprocess/visualization/seismic_anomalies.cc
```diff
@@ -468,13 +481,25 @@ namespace aspect
...
```
jdannberg: Same here

source/postprocess/depth_average.cc
```diff
@@ -366,11 +366,40 @@ namespace aspect
...
+                {
+                  variables.push_back("Vs");
+                  variables.push_back("Vp");
+                }
 
-            if ( output_all_variables || std::find( output_variables.begin(), output_variables.end(), "Vp") != output_variables.end() )
-              variables.push_back("Vp");
+              if (std::find( output_variables.begin(), output_variables.end(), "Vs") != output_variables.end())
+                {
+                  AssertThrow(material_model_provides_seismic_output,
```
jdannberg: Can you move this assert above line 377 and combine it with the one below?

gassmoeller:
I addressed the comments and updated the test results. The one issue here is that the column numbers in the depth_average output files will change if no seismic velocities are supplied by the material model. This can break postprocessing scripts (only for columns viscosity and vertical heat flux). I do not see a way to prevent this though.

doc/modules/changes/20180427_gassmoeller_2
```diff
@@ -0,0 +1,5 @@
+Changed: The default value for AdditionalSeismicOutputs was changed from -1 to
+a signaling Nan. This ensures every plugin that creates seismic outputs needs
+to use a material model that actually fills those outputs.
```
jdannberg: If there's no way around changing the change in column number in the depth average file, that should be documented here too.

gassmoeller:
I ran the check.sh script in benchmarks and cookbooks, and checked that the seismic velocity columns are not referenced in the manual or somewhere else. Good to go?

PR #2191, opened 2018-04-28
Newton stabilization resets to the user settings at the end of the timestep

MFraters:
I have been able to find an other bug in the Newton solver. The user given parameters stabilization should reset every time a timestep is taken. I would like this fixed to be merged before 2.0 is released.

To make a test for this, but I would like to use the Spiegelman material for that. Should I just duplicate the code in a new test, or put it in a common place so the two tests can load it from there?


jdannberg:
You can include .cc files from other tests, so you could just add the line `#include "spiegelman_material.cc"`  in the .cc file of your new test. Wouldn't that work? 
That would also make it easy to move the file back to the benchmarks folder once you add the benchmark .prm files, as only that one line with the include would have to be changed.

MFraters:
> You can include .cc files from other tests, so you could just add the line #include "spiegelman_material.cc" in the .cc file of your new test. Wouldn't that work?

Unfortunately that wouldn't work because test cc file actually takes over the process and exits before aspect starts up. But I am now thinking that I could reverse it, and put the code in the new test, and then include it in the spiegelman_material.cc.

MFraters:
Thanks for the suggestion @jdannberg, it works now without duplication :)

MFraters:
hmm, the fail safe this test uses, requires at least deal.ii 9.0. So should I remove this test for now?

gassmoeller:
Keep the test in, but rename the parameter file to `spiegelman_fail_test.prm.deal.II.9.0`. This will disable the test for now (we have a few others like this). We just need to remember to rename them, when we require deal.II 9.0.

tests/spiegelman_fail_test.prm
```diff
@@ -0,0 +1,132 @@
+# This is the most difficult case of the Spiegelman benchmark. The
+# goal of this test is to show that the linear Newton solver fails
+# when the the matrix becomes non Symmetric Postive Definite, and 
+# is safed by foring the matrix to be SPD. The next timestep the
```
gassmoeller: "... saved by forcing ... . In the next timestep ..."

bangerth:
Good catch -- the thought of this had crossed my mind at one point, but like so many other things left my brain again before I could make a note of it...

PR #2192, opened 2018-04-29
Fix visco_plastic and diffusion_dislocation thermal diffusivity

naliboff:
Follow up to #2170. It turns out none of the `diffusion_dislocation` or `visco_plastic` tests use the `reference density profile` formulation for the temperature equation (I have no idea why I assumed they did .. careless error). 

As previously written, the two material models did not check whether adiabatic conditions are initialized before asking for the density from `adiabatic_conditions`. This is now fixed and I've added tests that use `reference density profile` in both material models. I also checked that the other changes (`basic statistics` & `thermal diffusivity` post processors) in #2170 work with `reference density profile`, but did not add any associated tests.

### DIFF
--- a/doc/modules/changes/20180430_naliboff
+++ b/doc/modules/changes/20180430_naliboff
@@ -1,0 +1,6 @@
+Fixed: The material models <code>visco plastic</code> and <code>diffusion 
+dislocation</code> now use the reference density to calculate thermal
+conductivity if the <code>Temperature equation</code> formulation is set
+to <code>reference density profile</code>.
+<br>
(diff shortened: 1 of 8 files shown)


gassmoeller:
Very nice. Please update the test results and add a changelog entry in the style of 'Fixed: The material models ... now correctly use the reference density for ... if the formulation ...  is used'.

naliboff:
@gassmoeller  - I added a changelog entry and fixed the test. Thanks for going through this and let me know if there is anything else you would like addressed.

gassmoeller:
Thanks! :+1:

PR #2193, opened 2018-04-30
Fix manual referencing removed file

gassmoeller:
This section of the manual referenced a file that was removed in #2157. I did not notice the missing file, because the corresponding .prm.out file was still in my folder, and so the manual compiled for me.


naliboff:
Looks good 👍. Ready to merge once the tester is finished.

PR #2197, opened 2018-05-01
Add release notice to tasklist

gassmoeller:
This is our preliminary compilation of important changes compared to 1.5.0. Can be modified up to the release if we find other points.


tjhei:
thank you! Should we leave this open for now?

gassmoeller:
Yes, lets merge this when we actually do the release.

Issue #1767, opened 2017-05-23
Refactor some material model functions


gassmoeller:
A number of material models (at least `DiffusionDislocation`, `ViscoPlastic`, and `DynamicFriction` implement the functions `compute_volume_fractions`, and `average_value`. These can be moved into the `Utilities` namespace, or into the `Interface` class to avoid code duplication.

bangerth:
#1707 adds at least one other occurrence of `compute_volume_fractions()`.

wanyingw:
@gassmoeller -- did this get completely fixed?

gassmoeller:
Sorry @wanyingw only saw your comment today. This is not quite fixed (compute_volume_fractions was merged, but average_value not). However, the remaining part is more tricky, so it would not make a good starter project in my opinion. I will remove the label for now.

naliboff:
@wanyingw @gassmoeller - I started to move `average_value` into the material model interface at the hack, but got distracted with other things. I was planning to finish this up after I got back from a trip in 10 days or so, but honestly what I have done so far was not that much work. 

@wanyingw - Feel free to go ahead with this if you would like to do so, but otherwise I will finish it up when I get back. I agree with @gassmoeller that moving other more complicated functions over is definitely not a started project and will likely require some discussion first.

PR #2199, opened 2018-05-02
Unify compute volume fractions

### LINKED ISSUE #2121: Unify compute_volume_fractions() implementation
We have a number of places where we define a function `compute_volume_fractions()`:
```
source/heating_model/compositional_heating.cc:    compute_volume_fractions( const std::vector<double> &compositional_fields) const
source/material_model/diffusion_dislocation.cc:    compute_volume_fractions( const std::vector<double> &compositional_fields) const
source/material_model/multicomponent.cc:    compute_volume_fractions( const std::vector<double> &compositional_fields) const
source/material_model/dynamic_friction.cc:    compute_volume_fractions( const std::vector<double> &compositional_fields) const
source/material_model/visco_plastic.cc:    compute_volume_fractions( const std::vector<double> &compositional_fields) const
```
I have not double-checked, but suspect that they are all copies of each other. We should unify all of these places by moving the function into namespace `Utilities`.

gassmoeller:
Closes #2121 and is a first step towards #1767. All of the unify_volume_fractions functions were the same (one minor deviation in visco_plastic and compositional_heating, that is why I introduced the component mask). Hopefully this stops the spreading of this function into ever more material models.


source/material_model/visco_plastic.cc
```diff
@@ -70,60 +70,31 @@ namespace aspect
...
-      // clip the compositional fields so they are between zero and one
-      std::vector<double> x_comp = compositional_fields;
-      for ( unsigned int i=0; i < x_comp.size(); ++i)
-        x_comp[i] = std::min(std::max(x_comp[i], 0.0), 1.0);
+      double averaged_parameter = 0.0;
 
-      // assign compositional fields associated with strain a value of 0
+      // Store which components to exclude during volume fraction computation.
+      ComponentMask composition_mask(this->n_compositional_fields(),true);
       if (use_strain_weakening == true)
```
jdannberg: What if the strain tensor is tracked, but not used for weakening? Is that not supported? (I know that technically this doesn't relate to your pull request...)

include/aspect/material_model/visco_plastic.h
```diff
@@ -160,8 +160,6 @@ namespace aspect
 
         double grain_size;
 
-        std::vector<double> compute_volume_fractions(
-          const std::vector<double> &compositional_fields) const;
```
jdannberg: Why are these all protected and not private?

source/material_model/visco_plastic.cc
```diff
@@ -70,60 +70,31 @@ namespace aspect
...
-
-    template <int dim>
-    double
-    ViscoPlastic<dim>::
-    average_value ( const std::vector<double> &composition,
-                    const std::vector<double> &parameter_values,
-                    const enum averaging_scheme &average_type) const
-    {
-      double averaged_parameter = 0.0;
       const std::vector<double> volume_fractions = compute_volume_fractions(composition);
```
jdannberg: Shouldn't that get the `composition_mask`? I am surprised that it's working like this...

benchmarks/nonlinear_channel_flow/simple_nonlinear.cc
```diff
@@ -55,9 +55,6 @@ namespace aspect
         parse_parameters (ParameterHandler &prm);
 
       private:
-        std::vector<double> compute_volume_fractions( const std::vector<double> &compositional_fields) const;
-
-
```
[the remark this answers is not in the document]
gassmoeller: there are a lot of nonlinear channel flow tests that use this material model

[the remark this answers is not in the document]
jdannberg: Ah, okay :+1:

source/material_model/visco_plastic.cc
gassmoeller: The `use_finite_strain_tensor` variable in this material model explicitly means use it for strain weakening. It is not meant for anything else.

gassmoeller:

include/aspect/heating_model/compositional_heating.h
```diff
@@ -99,7 +80,7 @@ namespace aspect
          * contribute to material properties like viscosity, heat production,
          * etc.
          */
-        std::vector<int> field_used_in_heat_production_averaging;
+        std::vector<bool> fields_used_in_heat_production_averaging;
```
bangerth: Is it correct that the length of this array is now equal to the number of compositional fields? What was it before -- the number of fields that are used, with each vector entry an index between zero and the number of compositional fields?

include/aspect/material_model/dynamic_friction.h
```diff
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2014 - 2018 by the authors of the ASPECT code.
+  Copyright (C) 2014 - 2017 by the authors of the ASPECT code.
```
bangerth: You probably want to do undo this.

include/aspect/material_model/dynamic_friction.h
```diff
@@ -177,8 +163,8 @@ namespace aspect
         std::vector<double> densities;
 
         /**
-         * The dynamic coefficient of friction
-         */
+        * The dynamic coefficient of friction
+        */
```
bangerth: and this

include/aspect/material_model/interface.h
```diff
@@ -779,6 +780,25 @@ namespace aspect
...
+     * For multicomponent material models: Given a vector of of compositional
+     * fields of length N, this function returns a vector of volume fractions
+     * of length N+1, corresponding to the volume fraction of a ``background
+     * material'' as the first entry, and volume fractions for each of the input
+     * fields as the following entries. The returned vector will sum to one.
+     * If the sum of the compositional_fields is greater than
+     * one, we assume that there is no background mantle (i.e., that field value
+     * is zero). Otherwise, the difference between the sum of the compositional
+     * fields and 1.0 is assumed to be the amount of background mantle.
+     * Optionally, one can input a component mask that determines, which of the
```
bangerth: no comma after 'determines'

include/aspect/material_model/interface.h
```diff
@@ -779,6 +780,25 @@ namespace aspect
...
+     * compositional fields to use during the computation (e.g. because
+     * some fields do contain non-volumetric quantities like strain,
```
bangerth: I would drop 'do' here because it puts emphasis on something that I find misleading.

source/heating_model/compositional_heating.cc
```diff
@@ -136,16 +89,23 @@ namespace aspect
...
-          field_used_in_heat_production_averaging = Utilities::possibly_extend_from_1_to_N (Utilities::string_to_int(Utilities::split_string_list(prm.get("Use compositional field for heat production averaging"))),
-                                                    n_fields,
-                                                    "Use compositional field for heat production averaging");
-
-          heating_values = Utilities::possibly_extend_from_1_to_N (Utilities::string_to_double(Utilities::split_string_list(prm.get("Compositional heating values"))),
-                                                                   n_fields,
-                                                                   "Compositional heating values");
-
-
+          std::vector<int> used_fields = Utilities::possibly_extend_from_1_to_N (
```
bangerth: I think you can make this `const`

source/heating_model/compositional_heating.cc
```diff
@@ -136,16 +89,23 @@ namespace aspect
...
+                                           Utilities::string_to_int(
+                                             Utilities::split_string_list(
+                                               prm.get("Use compositional field for heat production averaging"))),
+                                           n_fields,
+                                           "Use compositional field for heat production averaging");
+
+          fields_used_in_heat_production_averaging.resize(used_fields.size());
+          for (unsigned int i=0; i<used_fields.size(); ++i)
+            fields_used_in_heat_production_averaging[i] = used_fields[i];
```
bangerth: Is this correct? The right hand side are integers, the left are booleans. If this is correct, then I misunderstood what `used_fields` contained before. Can you make the cast explicit by putting a boolean expression on the rhs?

gassmoeller:
I addressed the comments and while doing that found an existing bug in compositional heating that interfered with this PR. It would be great if somebody could take a look at the new changelog entry, the compositional heating plugin, and the new test compositional_heating_exclude_background to make sure I got this right. In essence there was an off-by-one bug in the use of the `fields_used_in_heat_production_averaging` vector. The first entry was supposed to be used for the background field, but was actually used for the first compositional field when computing volume fractions.

naliboff:
Ugh, thanks for catching that Rene. The test I originally ran just worked by proxy of having the right combination of values.

PR #2195, opened 2018-05-01
add new cite.html

tjhei:

PR #2200, opened 2018-05-02
Add citation info to header

gassmoeller:
Adds to #2195 and #2196 a note to the header of the model output. I am aware we do not want to put too much into the header, but this seems like an important enough notice that many might forget otherwise.


tjhei:
Good idea and a step in the right dynamic-citation-thing direction. ;-)

gassmoeller:
Yes, we can extend this later with more info about the current model run.

PR #2202, opened 2018-05-03
[post-2.0] Automate test procedure

gassmoeller:
This is @tjhei, @tjesser-ucdavis-edu, and my current version of closing issue #2143. The Jenkinsfile that this PR adds to the repository documents every step in our testing procedure, and allows to easily set up a Jenkins server to replace our test setup. It will test pull requests from maintainers automatically, for each other pull request a user with access to the jenkins server will have to click a button on the jenkins server (replacing the current /run-tests command). I tested this script with a local jenkins installation, and it seems to work fine, but it might need some tweaking when we try to let it work with Timo's jenkins server. In particular it will need some changes when we switch to the Kubernetes cluster at CIG. Either way (kubernetes or stand-alone docker) will allow us to dedicate additional resources as testing agents that can run in parallel for different PRs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,74 @@
+#!groovy
+
+pipeline {
+  agent {
+    docker {
+        image 'dealii/dealii:v8.5.1-gcc-mpi-fulldepscandi-debugrelease'
+	label 'has-docker'
+    }
+  }
+
+  parameters {
+    booleanParam(defaultValue: false, description: 'Is the pull request approved for testing?', name: 'TRUST_BUILD')
+  }
+
+  stages {
+    stage('Check indentation') {
+      steps {
+        sh './doc/indent'
+        sh 'git diff | tee changes-astyle.diff'
+        archiveArtifacts artifacts: 'changes-astyle.diff', fingerprint: true
+        sh 'git diff --exit-code --name-only'
+        }
+    }
+
+    stage ("Check execution") {
+      when {
+	allOf {
+            environment name: 'TRUST_BUILD', value: 'false' 
+            not {branch 'master'}
+            not {changeRequest authorEmail: "rene.gassmoeller@mailbox.org"}
+            not {changeRequest authorEmail: "timo.heister@gmail.com"}
+            not {changeRequest authorEmail: "bangerth@colostate.edu"}
+            not {changeRequest authorEmail: "judannberg@gmail.com"}
+            not {changeRequest authorEmail: "ja3170@columbia.edu"}
+            not {changeRequest authorEmail: "john.naliboff@gmail.com"}
+            not {changeRequest authorEmail: "menno.fraters@outlook.com"}
+            not {changeRequest authorEmail: "acglerum@gfz-potsdam.de"}
+	    }
+      }
+      steps {
+	  echo "Please ask an admin to rerun jenkins with TRUST_BUILD=true"
+	    sh "exit 1"
+      }
+    }
+
+    stage('Build') {
+      steps {
+        sh '''
+          mkdir -p build-gcc-fast
+          cd build-gcc-fast
+          cmake -G "Ninja" gcc -D ASPECT_TEST_GENERATOR=Ninja -D ASPECT_USE_PETSC=OFF -D ASPECT_RUN_ALL_TESTS=ON -D ASPECT_PRECOMPILE_HEADERS=ON ..
+          ninja
+        '''
+      } 
+    }
+
+    stage('Run tests') {
+      steps {
+        sh '''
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,74 @@
...
+  agent {
+    docker {
+        image 'dealii/dealii:v8.5.1-gcc-mpi-fulldepscandi-debugrelease'
+	label 'has-docker'
+    }
+  }
+
+  parameters {
+    booleanParam(defaultValue: false, description: 'Is the pull request approved for testing?', name: 'TRUST_BUILD')
+  }
```
tjesser-ucdavis-edu: I haven't looked at the Jenkins settings for who can trigger tests, but I would hope that it would be entirely on the Jenkins side. Since the Jenkinsfile is in the repository, this is just as susceptible to malicious pull requests as everything else.

tjesser-ucdavis-edu: That's interesting. I wonder why Jenkins decided to do it that way. I'm not sure how much security you can get from using a trusted Jenkinsfile to build untrusted code.

It looks like you try to avoid that problem by erroring with this `TRUST_BUILD` whitelist, but you're still running the indentation checker on untrusted code.

Also, instead of killing the build with an error and telling the submitter to contact an admin, would it be better to use the Jenkinsfile [input directive](https://jenkins.io/doc/book/pipeline/syntax/#input)? ([Interface example](https://jenkins.io/doc/tutorials/create-a-pipeline-in-blue-ocean/#add-a-final-deliver-stage-to-your-pipeline), scroll down a bit)

tjhei: > but you're still running the indentation checker on untrusted code.

That is fine and I have been doing that since forever. Except that using ``doc/indent`` from untrusted code is not smart. Maybe reimplement this script? ;-)

> to use the Jenkinsfile input directive?

Yes, that is also an option except that it was somewhat finicky when I tried it. I also haven't figured out how to do this input check before a node was allocated. This way pull requests will block the CI servers.

tjhei:
Thank you. What is not covered here yet is time limits for the tester.

Jenkinsfile
```diff
@@ -0,0 +1,74 @@
...
+          ninja
+        '''
+      } 
+    }
+
+    stage('Run tests') {
+      steps {
+        sh '''
+          cd build-gcc-fast
+          ASPECT_TESTS_VERBOSE=1 ../cmake/generate_reference_output.sh
```
tjhei: does this work in that it reports failing tests and generate the output in one run?

Jenkinsfile
```diff
@@ -0,0 +1,74 @@
...
+        '''
+        archiveArtifacts artifacts: 'build-gcc-fast/changes.diff', fingerprint: true
+        sh 'git diff --exit-code --name-only'
+      }
+    }
+  }
+
+  post {
+    always {
+      deleteDir()
```
tjhei: is this necessary and what does it delete, exactly? I think this means the ASPECT repo will be cloned every time, right?

Jenkinsfile
```diff
@@ -0,0 +1,74 @@
...
+      steps {
+	  echo "Please ask an admin to rerun jenkins with TRUST_BUILD=true"
+	    sh "exit 1"
+      }
+    }
+
+    stage('Build') {
+      steps {
+        sh '''
+          mkdir -p build-gcc-fast
```
tjhei: this means the build directory is inside the workspace and so will write outside the container. Maybe not a good idea? In my current setup the docker instance I am building a separate build directory.

gassmoeller:
I think I covered all of the mentioned points. The behavior is now similar to the current tester. I also introduced time limits, swapped the order of permission check and indentation check, and moved the build folder into the container. One idea for future improvement is to not fail an untrusted build, but to await admin input, but we need to make sure that this happens on the jenkins master, so that it does not block an agent, and for that we need a working system first. Thus, I would like to get this merged (it does not hurt to have this file lying around and doing nothing), and then start playing with the server setup. @tjhei: You created a user account for me on your server, but I can not create or change projects, could you add that?

gassmoeller:
Very nice. It seems like pr-head passed the tests (i.e. the tests for this branch, without merging into master first). I am not sure why pr-merge failed, the error message looks like git had some problem checking out the master branch (i.e. before or while merging this pr into the master branch). Could this be related to the fact that master does not yet have a Jenkinsfile?

tjhei:
> I am not sure why pr-merge failed,

yeah, I don't know. That was the first run and I think the error message has something to do with out-of-memory on the machine this ran. I disabled the "pr-merge" for now.

PR #2208, opened 2018-05-04
Fix nullspace removal

gassmoeller:
This is the fix for the bug described in #2198. I am working on a test and postprocessor for the rotation properties, but it is more complicated than I expected, since I need to find a case where it matters (I will open a separate pull request for that). Since @Shangxin-Liu @ian-r-rose and I agree this is a bug, we should get this merged before the release (even if the test has to wait).

### DIFF
--- a/doc/modules/changes/20180503_gassmoeller
+++ b/doc/modules/changes/20180503_gassmoeller
@@ -1,0 +1,5 @@
+Fixed: The nullspace removal had a minor bug that caused a small
+remaining rotation in three dimensional models even if the net rotation
+or angular momentum should have been removed. The bug was fixed.
+<br>
+(Rene Gassmoeller, Shangxin Liu, 2018/05/03)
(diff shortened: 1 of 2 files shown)


Shangxin-Liu:
Yes. I agree with Ian. This subroutine is commonly used with free slip so maybe we can find a test that matters from this BC. But this of course should be firstly merged.  @gassmoeller

bangerth:
On 05/04/2018 08:41 AM, Ian Rose wrote:
> But I think it should be fairly easy to get a net rotation by having a 3D 
> model with free slip upper and bottom surfaces, and running it for 10-20 
> timesteps. Happy to help with that, but I don't have a working install at the 
> moment.

An easier way is to prescribe a gravity field that has a tangential component. 
Such fields are not physical (they do not result as the gradient of a 
potential), but ASPECT doesn't care about this. In any case, if you have a 
tangential force, you will get a tangential velocity.

PR #2209, opened 2018-05-04
clang 6 requires deprecation attribute in front

tjhei:
Fixes errors like:
```
../include/aspect/simulator_access.h:762:35: error: 'deprecated'
attribute cannot be applied to types
      find_postprocessor () const DEAL_II_DEPRECATED;
```


gassmoeller:
Ah weird. I know too little about the internals of compilers to understand this. Anyway, gcc does not seem to care (tests pass).

tjhei:
> Anyway, gcc does not seem to care (tests pass).

Yeah, I don't know what is going on either. The standard allows any location...

PR #2212, opened 2018-05-04
Add rotation statistics postprocessor

gassmoeller:
This is the full solution including test for the bug described in #2198. #2208 already contains the fix, but this PR adds tests and a new postprocessor 'rotation statistics' that will prevent this from happening again. The important part is the test rotation_statistics_3d_asymmetric.prm in which I show that the new term is analytically correct (while the old one had a sign error). Thanks @jdannberg for the idea for the analytical solution. Since this is a big piece of code we should go ahead with #2198 and I will rebase this afterwards.


include/aspect/postprocess/rotation_statistics.h
```diff
@@ -0,0 +1,87 @@
...
+         * Read the parameters this class declares from the parameter file.
+         */
+        virtual
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+        /**
+         * Whether to use a constant density of one for the computation of the
+         * angular momentum, and moment of inertia. This is an approximation
```
tjhei: remove comma

include/aspect/postprocess/rotation_statistics.h
```diff
@@ -0,0 +1,87 @@
...
+         * rotation. If this parameter is true this postprocessor computes
+         * 'net rotation' instead of 'angular momentum'.
+         */
+        bool use_constant_density;
+
+        /**
+         * Whether to write the full moment of inertia tensor into the
+         * statistics output instead of its norm for the current rotation
+         * axis. This is a second-order symmetric tensor with
+         * 6 components in 3D. (in 2D this option has no effect, because
```
tjhei: "(in" -> "In"

source/postprocess/rotation_statistics.cc
```diff
@@ -0,0 +1,260 @@
...
+    RotationStatistics<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Postprocess");
+      {
+        prm.enter_subsection("Rotation statistics");
+        {
+          prm.declare_entry("Use constant density of one","false",
+                            Patterns::Bool(),
+                            "Whether to use a constant density of one for the computation of the "
+                            "angular momentum, and moment of inertia. This is an approximation "
```
tjhei: remove comma

source/postprocess/rotation_statistics.cc
```diff
@@ -0,0 +1,260 @@
...
+                            "that assumes that the 'volumetric' rotation is equal to the 'mass' "
+                            "rotation. If this parameter is true this postprocessor computes "
+                            "'net rotation' instead of 'angular momentum'.");
+          prm.declare_entry("Output full moment of inertia tensor","false",
+                            Patterns::Bool(),
+                            "Whether to write the full moment of inertia tensor into the "
+                            "statistics output instead of its norm for the current rotation "
+                            "axis. This is a second-order symmetric tensor with "
+                            "6 components in 3D. (in 2D this option has no effect, because "
```
tjhei: again, "(in" -> "In"

gassmoeller:
I addressed all comments.

Shangxin-Liu:
I'm reading the new post-processor on the rotation statistics. Thanks for such a thorough check and fix-up! @gassmoeller 

One thing pops up in my mind and I'd like to further confirm: In this postprocessor, the velocity field (in.velocity[q]) you grab is the velocity solution after the net rotation (or angular momentum) has been removed or the original velocity solution before the net rotation (or angular momentum) has been removed? It seems that you are calculating the same thing as null_space.cc does to make a statistic on the rotation so the input velocity field is the original one before the removal of the net rotation (or angular momentum) has been applied, yes?  If so, how about the velocity field used in other postprocessor like in dynamic_topography and geoid to calculate strain rate? I know that the velocity solution used at each time-step to update energy system and composition should be the one with the removal of the pure rotation already done. I'm a little confused on the order of the removal of the pure rotation being applied.   @gassmoeller @bangerth

Shangxin-Liu:
Thanks, Wolfgang. @bangerth So it seems that I should not add "top sublayer rotation removal" in the null_space.cc file to prevent the discontinuous non-physical velocity field entering into the post processor. I'll remove it by myself after grabbing the velocity solution. 

I'm not sure whether my understanding is correct: So if the rotation is already set to be removed in the Null space removal subsection, the rotation statistics postprocessor here, in theory, should end up with zero rotation and we expect to see it tends to be close to zero as we increase the resolution. Is this what you are trying to do here? @gassmoeller

gassmoeller:
Yes, this postprocessor will output the rotation (or angular momentum, depending on input parameter) of the model. If this rotation has been removed by the nullspace removal, it shows values close to zero. In practice in the tests the nullspace removal reduces the net rotation / angular momentum by around 4 orders of magnitude compared to the non-removed version after a few timesteps.

Is this ready to merge?

source/postprocess/rotation_statistics.cc
```diff
@@ -0,0 +1,260 @@
...
+
+                    local_angular_momentum += r_cross_v * rho * fe_values.JxW(q);
+
+                    // calculate moment of inertia
+                    local_moment_of_inertia[0][0] += (position[1] * position[1] + position[2] * position[2]) * rho * fe_values.JxW(q);
+                    local_moment_of_inertia[1][1] += (position[0] * position[0] + position[2] * position[2]) * rho * fe_values.JxW(q);
+                    local_moment_of_inertia[2][2] += (position[0] * position[0] + position[1] * position[1]) * rho * fe_values.JxW(q);
+                    local_moment_of_inertia[0][1] -= (position[0] * position[1]) * rho * fe_values.JxW(q);
+                    local_moment_of_inertia[0][2] -= (position[0] * position[2]) * rho * fe_values.JxW(q);
+                    local_moment_of_inertia[1][2] -= (position[1] * position[2]) * rho * fe_values.JxW(q);
```
bangerth: I bet one could have done that in a dimension-independent way with only one line. Aren't these all just some outer  products and tensor products?

bangerth: Not that it really matters, though. So feel free to ignore my comment.

source/postprocess/rotation_statistics.cc
```diff
@@ -0,0 +1,260 @@
...
+                                   :
+                                   out.densities[q];
+
+                if (dim == 2)
+                  {
+                    // Get the velocity perpendicular to the position vector
+                    const Tensor<1,dim> r_perp = cross_product_2d(in.position[q]);
+                    // calculate a signed scalar angular momentum
+                    local_scalar_angular_momentum += in.velocity[q] * r_perp * rho * fe_values.JxW(q);
+                    local_scalar_moment_of_inertia += in.position[q].norm_square() * rho * fe_values.JxW(q);
```
bangerth: Can you leave a comment here that states that you are making the assumption (see the manual) that the 2d geometry is a cross-section through an infinite domain in z direction, with a zero z-velocity. As a consequence, the angular momentum you compute here is simply the zz-component of the full 3d tensor, and is consequently a scalar. (or attach that to the declaration of the variable.)

source/postprocess/rotation_statistics.cc
```diff
@@ -0,0 +1,260 @@
...
+      {
+        statistics.add_value(name,value);
+        statistics.set_precision (name, 8);
+        statistics.set_scientific (name, true);
+      }
+    }
+
+    template <int dim>
+    std::pair<std::string,std::string>
+    RotationStatistics<dim>::execute (TableHandler &statistics)
```
bangerth: I wonder whether  the function would be easier to read if you had specializations for the 2d and 3d cases separately. Opinions?

source/postprocess/rotation_statistics.cc
```diff
@@ -0,0 +1,260 @@
...
+                            "that assumes that the 'volumetric' rotation is equal to the 'mass' "
+                            "rotation. If this parameter is true this postprocessor computes "
+                            "'net rotation' instead of 'angular momentum'.");
+          prm.declare_entry("Output full moment of inertia tensor","false",
+                            Patterns::Bool(),
+                            "Whether to write the full moment of inertia tensor into the "
+                            "statistics output instead of its norm for the current rotation "
+                            "axis. This is a second-order symmetric tensor with "
+                            "6 components in 3D. In 2D this option has no effect, because "
+                            "the rotation axis is fixed and thus it is always a scalar.");
```
bangerth: it -> the angular momentum

gassmoeller:
I addressed the comments and made specializations for 2D and 3D, which is easier to read. I am sure there is a dimension-independent way to write this, but I could not figure it out in a few minutes, and the nullspace removal internally does it the same way, so I tried to keep things as similar as possible between the internal handling and the postprocessor.
Should be ready.

PR #2213, opened 2018-05-05
Prevent division by zero in bound preserving limiter

gassmoeller:
For some reason this does not trigger in deal.II 8.5, but it does in 9.0. Maybe we got more accurate and that is why one of the denominators now is zero. Anyway this fixes it and also makes the style a bit more uniform.


gassmoeller:
Looks like the tester broke while testing this Pr, @tjhei could you restart?

gassmoeller:
A fix not breaking any tests (on 8.5 that is). Good to go from my side.

PR #2214, opened 2018-05-05
Fix wrongly defined analytic function

gassmoeller:
This triggers an assert with deal.II 9.0, I guess because of a newly intoduced assert. Anyway, it is clearly a bug.


gassmoeller:
I added some more cases where analytic reference functions did not declare the correct number of components during construction.

PR #2218, opened 2018-05-06
add additional info to detailed.log

tjhei:
this doesn't need to go into 2.0, but it also won't hurt, I believe.


gassmoeller:
Looks good.

PR #2219, opened 2018-05-06
Remove info message

### LINKED ISSUE #2217: default ascii data gravity generates warning
```
subsection Gravity model
  set Model name = ascii data
end
```
for example in cookbooks/shell_simple_2d.prm generates the warning
```
   Ascii data file coordinates are not equidistant. 
```
I think this is confusing and we should remove the warning or suppress it in this situation.

gassmoeller:
Closes #2217. Is this what you thought about @tjhei? I agree that this message is confusing (since everything works even if the coordinates are non-equidistant). So it is more an info message than a warning. I can see that somebody might want to check though (i.e. to see if a data file was prepared correctly) that is why I introduced the get function. This might break test output.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -506,6 +506,14 @@ namespace aspect
         get_column_names() const;
 
         /**
+         * Returns whether the stored coordinates are equidistant. If
+         * coordinates are equidistant the lookup is more efficient. Returns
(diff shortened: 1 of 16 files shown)


bangerth:
(I would have merged, but I don't know whether this is what @tjhei had in mind, so I'll leave it to him.)

PR #2220, opened 2018-05-06
fix drucker prager 3d derivatives and extend test to catch this in the future.

MFraters:
Found this bug when I looking at the equations to make a analytical solution for the visco-plastic derivatives. I have extended the test to now also test the 3d implementation and fixed the bug (as the test shows). I will now also look at and update the other derivative tests.


gassmoeller:
Yes, looks correct.

MFraters:
ah, I now see that I forgot to enlarge the finite difference of the strain-rate tensor here and in #2221 and #2222. Should not matter for the result, but might be nice to update. Will do that tomorrow.

PR #2224, opened 2018-05-07
Better document the 'temperature' and 'velocity' refinement criteria.

### LINKED ISSUE #400: Temperature strategy needs more explanation
In the "Parameters in section mesh refinement" (believed to be 5.59) of the manual, under parameter name: strategy, there was a request from Shangxin to elaborate the 'temperature' strategy.

Him and I think that it is too brief comapred to the others and could use a little more explanation as to what it does.

bangerth:
Fixes #400.


MFraters:
I like the extension of the documentation. Since this is also used in the composition mesh refinement criteria, could you also add it there? Otherwise I think it is good to merge.

Issue #2211, opened 2018-05-04
floating point exception (clang6)


tjhei:
I am getting
```
Thread 1 "aspect" received signal SIGFPE, Arithmetic exception.
0x0000000001a1f19d in aspect::Assemblers::StokesIncompressibleTerms<2>::execute (this=0x603000556c38, scratch_base=..., data_base=...)
    at ../source/simulator/assemblers/stokes.cc:205
205	          for (unsigned int i=0; i<stokes_dofs_per_cell; ++i)
```
when running solcx in the second nonlinear solve. I am not sure what is going on here.

gassmoeller:
The fix for my problem is in #2214, does your error still occur after that fix?

gassmoeller:
I can reproduce exactly your error message on ubuntu 14.04 (clang 6 manually installed) and 18.04 (clang 6 installed from repository). I can not see what is happening though. Is there a tester for deal.II with clang 6 and this particular setup? Then we could at least narrow down if the problem is in aspect or deal.II.

gassmoeller:
This is the full callstack:
```
[cb45769cb27a:06347] *** Process received signal ***
[cb45769cb27a:06347] Signal: Floating point exception (8)
[cb45769cb27a:06347] Signal code: Invalid floating point operation (7)
[cb45769cb27a:06347] Failing at address: 0x135f71c
[cb45769cb27a:06347] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x12890)[0x7f0098b0a890]
[cb45769cb27a:06347] [ 1] ../aspect(_ZNK6aspect10Assemblers25StokesIncompressibleTermsILi2EE7executeERNS_8internal8Assembly7Scratch11ScratchBaseILi2EEERNS4_8CopyData12CopyDataBaseILi2EEE+0x3ec)[0x135f71c]
[cb45769cb27a:06347] [ 2] ../aspect(_ZN6aspect9SimulatorILi2EE28local_assemble_stokes_systemERKN6dealii18TriaActiveIteratorINS2_15DoFCellAccessorINS2_10DoFHandlerILi2ELi2EEELb0EEEEERNS_8internal8Assembly7Scratch12StokesSystemILi2EEERNSC_8CopyData12StokesSystemILi2EEE+0x396)[0x128c3c6]
[...]

PR #2225, opened 2018-05-07
Disable floating point exceptions for clang6

gassmoeller:
Title says it all. As described in #2211 clang6 passes our test for floating point exceptions during running cmake, but triggers floating point exceptions during runtime.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,6 +303,12 @@ ENDIF()
 # exceptions. This is done here:
 SET(ASPECT_USE_FP_EXCEPTIONS ON CACHE BOOL "If ON, floating point exception are raised in debug mode.")
 
+# Clang 6.0 throws random floating point exceptions, which we could not
+# track down. Disable the exceptions for now.
+IF("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 6.0.0)
+  SET(ASPECT_USE_FP_EXCEPTIONS OFF CACHE BOOL "" FORCE)
+ENDIF()
+
 IF (ASPECT_USE_FP_EXCEPTIONS)
   INCLUDE(${CMAKE_SOURCE_DIR}/cmake/fpe_check.cmake)
 


CMakeLists.txt
```diff
@@ -303,6 +303,12 @@ ENDIF()
 # exceptions. This is done here:
 SET(ASPECT_USE_FP_EXCEPTIONS ON CACHE BOOL "If ON, floating point exception are raised in debug mode.")
 
+# Clang 6.0 throws random floating point exceptions, which we could not
+# track down. Disable the exceptions for now.
+IF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 6.0)
```
tjhei: I think it makes sense to blacklist >=6.0, otherwise people will run into this issue with 6.1, 7.0, etc.. I doubt that this is a bug that will be "fixed".

bangerth: I think there is little harm in doing as you suggest, but they better do fix this -- it's clearly a bug to speculatively execute a statement that might trigger an exception!

(Out of curiosity, which continent are you on right now?)

tjhei: > but they better do fix this

I doubt this will be addressed, so I prefer to be on the safe side on this issue (better to not have FP exceptions than code that doesn't work).

> (Out of curiosity, which continent are you on right now?)

Europe for the next two weeks. :-)

bangerth:
On 05/08/2018 09:08 PM, Timo Heister wrote:
>     but they better do fix this
> 
> I doubt this will be addressed, so I prefer to be on the safe side on this 
> issue (better to not have FP exceptions than code that doesn't work).

I agree. All I wanted to say is that compilers are allowed to perform 
optimizations as long as these optimizations do not lead to a visibly 
different outcome. This one does.

It is possible that Intel's ICC has a switch that allows floating point 
optimizations that assume that all FP values are regular numbers -- can you 
post what compiler flags we are using in release mode?

tjhei:
> can you
> post what compiler flags we are using in release mode?

you mean clang and in debug mode? (we don't do FPE stuff in release mode) The flags are 
```
-pedantic -fPIC -Wall -Wextra -Wpointer-arith -Wwrite-strings -Wsynth -Wsign-compare -Wswitch -Woverloaded-virtual  -Qunused-arguments -Wno-unsupported-friend -Wno-undefined-var-template -std=c++17 -Wno-parentheses -Wno-unused-local-typedefs -Og -ggdb -Wa,--compress-debug-section
```
I assume that removing ``-Og`` helps, but this is the deal.II default.

gassmoeller:
I updated the PR to check for clang version greater than 5.0.

tjhei:
Not that it matters much, but that means 5.0.1 will also now trigger this?

gassmoeller:
Patched it. I really wish CMake had a VERSION_EQUAL_OR_GREATER option.

tjhei:
``NOT ... VERSION_LESS ...``?

gassmoeller:
Changed it. Unintuitive, but works.

tjhei:
> Unintuitive, but works.

That is a good way to describe cmake in general. @tamiko :-)

bangerth:
On 05/08/2018 11:39 PM, Timo Heister wrote:
>     can you
>     post what compiler flags we are using in release mode?
> 
> you mean clang and in debug mode? (we don't do FPE stuff in release mode)

Ah, I had missed that. Too bad that that's even a problem in debug mode :-(


> The flags are
> 
> |-pedantic -fPIC -Wall -Wextra -Wpointer-arith -Wwrite-strings -Wsynth 
> -Wsign-compare -Wswitch -Woverloaded-virtual -Qunused-arguments 
> -Wno-unsupported-friend -Wno-undefined-var-template -std=c++17 
> -Wno-parentheses -Wno-unused-local-typedefs -Og -ggdb 
> -Wa,--compress-debug-section |
> 
> I assume that removing |-Og| helps, but this is the deal.II default.

Good question. I can't seem to find documentation for clang -Ox flags that 
would help figure out whether there is another flag one can pass to suppress 
this kind of thing.

PR #2226, opened 2018-05-07
Restructure derivative tests

MFraters:
This is a large restructuring of the derivatives tests. This makes them cleaner, shorter, more uniform and better documented :)

The 2d values should be the same as before, but the order in which they appear is different. The 3d cases now use full 3d strain-rate tensors.


gassmoeller:
Are you ok postponing this to 2.1? As far as I can see it contains only test changes.

MFraters:
I will leave that up to you, since I do not know the exact status of the release. The main goal of this pull request is to extend the scope of the testing of the derivatives to full 3d. Along the way, I figured out that it would be a good idea to also do all the other things.

PR #2227, opened 2018-05-08
fix the reaction term in the Melt Stokes system for operator splitting

jdannberg:
In melt transport models, there is a term on the right-hand side of the mass conservation equation that describes the change in volume associated with reactions from melt to solid (and the other way around). However, it uses only the reaction terms, even if operator splitting is used and these terms are zero. This means that in operator splitting models, this term was not computed correctly. This pull request fixes the computation of this term. 
I'll add a test case for this too, at the moment I just want to see if any tests fail.


jdannberg:
I've added a test with an analytical solution that verifies that the new solution is correct (and the old solution was wrong), and updated all the tests. So it would be great if we could get this fix into aspect before the release.

source/simulator/helper_functions.cc
```diff
@@ -1510,9 +1510,6 @@ namespace aspect
     LinearAlgebra::BlockVector distributed_vector (introspection.index_sets.system_partitioning,
                                                    mpi_communicator);
 
-    LinearAlgebra::BlockVector distributed_reaction_vector (introspection.index_sets.system_partitioning,
```
tjhei: you are switching from a non-ghosted to a ghosted vector, is this on purpose?

tjhei: I don't think you should be allowed to write entries into it in parallel.

source/simulator/melt.cc
```diff
@@ -299,9 +300,19 @@ namespace aspect
         const unsigned int porosity_index = introspection.compositional_index_for_name("porosity");
         const unsigned int is_compressible = simulator_access->get_material_model().is_compressible();
 
-        const double melting_rate     = scratch.material_model_outputs.reaction_terms[q_point][porosity_index];
         const double solid_density    = scratch.material_model_outputs.densities[q_point];
         const double fluid_density    = melt_out->fluid_densities[q_point];
+        const double reaction_rate    = simulator_access->get_timestep() > 0
```
tjhei: can you put parens around the ? expression?

source/simulator/melt.cc
```diff
@@ -369,6 +380,12 @@ namespace aspect
 
       const double pressure_scaling = this->get_pressure_scaling();
 
+      std::vector<double> reactions(n_q_points);
+      const unsigned int porosity_index = introspection.compositional_index_for_name("porosity");
+      if (this->get_parameters().use_operator_splitting)
+        scratch.finite_element_values[introspection.extractors.compositional_fields[porosity_index]].get_function_values(this->get_reaction_vector(),
```
tjhei: what if ``use_operator_splitting`` is false, are we then using uninitialized ``reactions[q]`` below? Maybe initialize to NaN?

jdannberg:
Thanks for the review. I think I addressed all your comments. You were of course right about the code not working in parallel.

PR #2235, opened 2018-05-09
Fix visco plastic floating point exception

gassmoeller:
The visco plastic material model throws floating point exceptions when used with strain weakening, because it assumes to always get a strain rate, which is not the case. This was not covered by the existing test, because the test only solves the first timestep (for which strain weakening is explicitly disabled). The fix is simple, and might be important enough to be cherry picked to the release branch. But I will leave that decision up to @naliboff and @tjhei.

### DIFF
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -537,7 +537,7 @@ namespace aspect
           // which represents the second invariant of the strain tensor
           double edot_ii = 0.;
           double e_ii = 0.;
-          if  (use_strain_weakening == true && use_finite_strain_tensor == false && this->get_timestep_number() > 0)
+          if  (in.strain_rate.size() > 0 && use_strain_weakening == true && use_finite_strain_tensor == false && this->get_timestep_number() > 0)
             {
               edot_ii = std::max(sqrt(std::fabs(second_invariant(deviator(strain_rate)))),min_strain_rate);
               e_ii = edot_ii*this->get_timestep();
--- a/tests/visco_plastic_yield_strain_weakening.prm
+++ b/tests/visco_plastic_yield_strain_weakening.prm
@@ -1,7 +1,7 @@
 # Global parameters
 set Dimension                              = 2
 set Start time                             = 0
-set End time                               = 0
+set End time                               = 1
 set Use years in output instead of seconds = true
 set Nonlinear solver scheme                = single Advection, iterated Stokes
 set Max nonlinear iterations               = 1
--- a/tests/visco_plastic_yield_strain_weakening/screen-output
+++ b/tests/visco_plastic_yield_strain_weakening/screen-output
@@ -21,6 +21,24 @@ Number of degrees of freedom: 1,885 (882+121+441+441)
 +---------------------------------+-----------+------------+------------+
 +---------------------------------+-----------+------------+------------+
 
+*** Timestep 1:  t=1 years
+   Solving temperature system... 0 iterations.
+   Solving eii system ... 5 iterations.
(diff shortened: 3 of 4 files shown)


naliboff:
@gassmoeller - Thanks for catching this and I'm surprised that none of us had run into that yet. In any case, the test should have certainly gone to at least one time step in the first place. The fix looks good and this should be ready to merge as soon as the tester is done. 

I'll leave it up to @tjhei for whether or not this is incorporated into the release branch. It is an important fix, but the error does not produce any incorrect values.

naliboff:
@jdannberg - I originally checked manually to make sure that the friction and cohesion values were reduced correctly as a function of the accumulated strain. I have no idea why I did not do at least one time step to confirm this in the actual test. 

If we want to have a test that confirms the values are correct, there is an option to output the friction and cohesion values through AdditionalMaterialModelOutputs. however, I'm not sure what the best method would be to get them in human readable form.

jdannberg:
@naliboff I was more worried about the accumulated strain being computed correctly (specifically the version that only uses the second invariant, which is the option that caused the problem here). It would be good to have a test case where we know how the finite strain should evolve over time (and therefore how the friction and cohesion values would change) and compare to that. 

We have test cases for the finite strain in the `benchmarks/finite_strain` folder, so that could be a good starting point. 

> I'm not sure what the best method would be to get them in human readable form

There is the gnuplot output format (instead of vtk) that produces numbers that can be compared in the tests (you just need to add the gnuplot output file as one of the test outputs, and make sure it's not too big).

naliboff:
@jdannberg - Gotcha and I agree it would be good to have tests specifically for looking at the finite strain. We can use the tests in benchmarks/finite_strain for the cases with the full finite strain tensor and use a model with a fixed strain rate for the finite strain invariant. I'll put these tests together tonight or tomorrow morning.

jdannberg:
Thank you!

bangerth:
I think the question is where these reaction terms feed into? Are they used for a compositional model that is then used again in the material model?

jdannberg:
In this model, the reaction terms represent the finite strain occurring in a given time step and are used to compute the accumulated finite strain everywhere in the model, which then feeds back into the material model (the higher the accumulated strain, the lower the viscosity). 
So the only relevant place where we have to make sure we compute the term correctly is the assembly of the composition advection equation, and that gets the full material model inputs including the strain rate. So in principle, this part should now work as intended. But I still think we should have a test case to see if the second invariant of the finite strain is tracked correctly, and if it influences the material properties correctly (in a similar way as we have these tests for the full finite strain tensor, which are the tests I mentioned above).

bangerth:
On 05/10/2018 08:40 AM, Juliane Dannberg wrote:
> In this model, the reaction terms represent the finite strain occurring in a 
> given time step and are used to compute the accumulated finite strain 
> everywhere in the model, which then feeds back into the material model (the 
> higher the accumulated strain, the lower the viscosity).

I see.


> So the only relevant place where we have to make sure we compute the term 
> correctly is the assembly of the composition advection equation, and that gets 
> the full material model inputs including the strain rate.

Yes, that is correct. So this should work.


> So in principle, 
> this part should now work as intended. But I still think we should have a test 
> case to see if the second invariant of the finite strain is tracked correctly, 
> and if it influences the material properties correctly (in a similar way as we 
> have these tests for the full finite strain tensor, which are the tests I 
> mentioned above).

Yes, more tests are always better.

tjhei:
This looks harmless enough to include in 2.0...

tjhei:
I cherry-picked it for 2.0 (without a PR)

Issue #676, opened 2015-12-02
Clarify pressure scaling more carefully in manual


gassmoeller:
In particular make clear that a change of the reference viscosity changes the pressure scaling as well.

bangerth:
Maybe also mention this in the material model member function documentation.

gassmoeller:
@jdannberg @naliboff, @ludovicJnnt, and I just had a discussion about pressure scaling and reference viscosities in mantle convection vs lithosphere dynamics models and we would like to have a discussion about this at the hackathon to find out if there is a more consistent way to compute these values.

bangerth:
We could just use a cheap method to compute some reference viscosity, e.g., by evaluating the viscosity at some representative sample of cells and choosing the harmonic average.

jdannberg:
In principle, yes, but the other point that was unclear to us was if the scaling relation and the length scale we use for the pressure scaling that you determined for mantle convection in the Kronbichler 2012 paper is also relevant for lithosphere deformation (because of the very different rheology), or if the length scale should be different for those models (and how best to test this).

bangerth:
Good question too. Do you have an example of what these length scales would be in the two cases?

The length scale we use does not have to be exact. It is principally used to make sure the two equations have the same physical units and approximately the same magnitude of numbers. If you're off by a factor of 10 or 100, nothing terrible will likely happen. If you're off by a factor of 1e6, that's a bigger problem.

jdannberg:
In mantle convection, I would say typical length scales of features can be 100 km, whereas in lithosphere models, features can be on the scale of 1 km. 

But we were also wondering if the dynamic pressures in lithosphere models could become much larger than for mantle convection, and if that's important for the computation of the initial residual we use to determine convergence of the linear solver. We noticed that in some 3d models with large viscosity jumps, changing the reference viscosity by just one or two orders of magnitude would sometimes drastically influence the iteration count and convergence of the linear solver.

bangerth:
On 05/23/2018 12:27 PM, Juliane Dannberg wrote:
> But we were also wondering if the dynamic pressures in lithosphere models 
> could become much larger than for mantle convection, and if that's important 
> for the computation of the initial residual we use to determine convergence of 
> the linear solver. We noticed that in some 3d models with large viscosity 
> jumps, changing the reference viscosity by just one or two orders of magnitude 
> would sometimes drastically influence the iteration count and convergence of 
> the linear solver.

[...]

PR #2236, opened 2018-05-10
Better document the purpose of the reference viscosity in material models.

bangerth:
This addresses #676. While there, I also improved the latex markup of 'Pa s', where
latex would ordinarily eat the space between the two units.

### DIFF
--- a/source/material_model/diffusion_dislocation.cc
+++ b/source/material_model/diffusion_dislocation.cc
@@ -368,7 +368,26 @@ namespace aspect
           prm.declare_entry ("Effective viscosity coefficient", "1.0", Patterns::Double(0),
                              "Scaling coefficient for effective viscosity.");
           prm.declare_entry ("Reference viscosity", "1e22", Patterns::Double(0),
-                             "Reference viscosity for nondimensionalization. Units $Pa s$");
+                             "The reference viscosity that is used for pressure scaling. "
+                             "To understand how pressure scaling works, take a look at "
+                             "\\cite{KHB12}. In particular, the value of this parameter "
+                             "would not affect the solution computed by \\aspect{} if "
+                             "we could do arithmetic exactly; however, computers do "
+                             "arithmetic in finite precision, and consequently we need to "
+                             "scale quantities in ways so that their magnitudes are "
+                             "roughly the same. As explained in \\cite{KHB12}, we scale "
+                             "the pressure during some computations (never visible by "
+                             "users) by a factor that involves a reference viscosity. This "
+                             "parameter describes this reference viscosity."
+                             "\n\n"
+                             "For problems with a constant viscosity, you will generally want "
+                             "to choose the reference viscosity equal to the actual viscosity. "
+                             "For problems with a variable viscosity, the reference viscosity "
+                             "should be a value that adequately represents the order of "
+                             "magnitude of the viscosities that appear, such as an average "
+                             "value or the value one would use to compute a Rayleigh number."
+                             "\n\n"
+                             "Units: $Pa s$");
 
           // Viscosity iteration parameters
           prm.declare_entry ("Strain rate residual tolerance", "1e-22", Patterns::Double(0),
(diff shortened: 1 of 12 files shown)


source/material_model/diffusion_dislocation.cc
```diff
@@ -362,13 +362,25 @@ namespace aspect
...
           prm.declare_entry ("Effective viscosity coefficient", "1.0", Patterns::Double(0),
                              "Scaling coefficient for effective viscosity.");
           prm.declare_entry ("Reference viscosity", "1e22", Patterns::Double(0),
-                             "Reference viscosity for nondimensionalization. Units $Pa s$");
+                             "The reference viscosity that is used for pressure scaling. "
+                             "To understand how pressure scaling works, take a look at "
+                             "\\cite{KHB12}. In particular, the value of this parameter "
+                             "would not affect the solution computed by \\aspect{} if "
+                             "we could do arithmetic exactly; however, computers do "
+                             "arithmetic in finite precision, and consequently we need to "
```
tjhei: I would prefer we put this large blob into the manual only and then put a reference here instead of duplicating it in every material model.

bangerth: I tried to find a place where this would fit, but we don't discuss the pressure scaling anywhere in the manual. In fact, we don't discuss any of our numerical methods in the manual at all, and on purpose -- that's what the papers are there for.

So I agree with you, but I couldn't figure out how to do that without writing a section I didn't want to write and that I wouldn't know where to put.

source/material_model/diffusion_dislocation.cc
```diff
@@ -362,13 +362,25 @@ namespace aspect
...
+                             "scale quantities in ways so that their magnitudes are "
+                             "roughly the same. As explained in \\cite{KHB12}, we scale "
+                             "the pressure during some computations (never visible by "
+                             "users) by a factor that involves a reference density. This "
```
jdannberg: This (and in the line below) should be reference viscosity (not density). In addition, I think it would be useful to add a sentence that tells the user how to choose this value, something like: "For problems with constant viscosity, the reference viscosity should be chosen equal to the viscosity used in the model. When solving a problem with variable viscosity, the reference viscosity should be a value that adequately represents the order of magnitude of the viscosities that appear, such as an average value or the value one would use to compute a Rayleigh number." (I took this more or less straight from the Kronbichler, Heister, Bangerth 2012 paper).

source/material_model/diffusion_dislocation.cc
jdannberg: I am fine either way, but if you decide to put this in the manual, I think the end of section "2.4 Static or dynamic pressure?" could be a good place.

gassmoeller:
I think it is not a good idea to replicate this in every material model. Even all models that do not have an explicit reference viscosity use an implicit assumption about the reference viscosity, which can still be influenced by input parameters, and this is an important information for users. We can still refer to the paper (like you do), but I would prefer if we had a section like this in the manual (e.g. where @jdannberg suggested).

bangerth:
Oh, I made a mistake and @MFraters fell in the trap :-( When I created #2270, I forgot to switch back to `master` and instead branched off of the branch in question here. @MFraters merged it, which also brought in the two commits in question here, and then github decided that because they are merged, this PR can also be closed. So we accidentally merged this PR :-(

Do you want me to back all of this out again? I've put on my list of things to do for the hackathon to play with methods to automatically compute a reference viscosity. If we go with that, then I will remove these comments again because there will then no longer be a need to provide one by the user.

MFraters:
ah, I just thought it was a general update of comments. These looked all good to me since they described what is happening a lot better, so I merged it. I didn't know that you where thinking of going for an automatic way of computing it.

jdannberg:
@bangerth I think in general, we were happy with your changes, except for the one point that the documentation should be in the manual rather than duplicated in all the material models (as it's important even for the material models that do not have the reference viscosity explicitly as an input parameter). So I'd be fine if you just did a follow-up pull request that moves he text from the material models into the manual.

PR #2237, opened 2018-05-10
More small fixes to do with physical units.

bangerth:


source/boundary_temperature/dynamic_core.cc
```diff
@@ -128,24 +128,24 @@ namespace aspect
...
-                             "Heat capacity of the core. Units: $J/kg/K$");
+                             "Heat capacity of the core. Units: $J/kg/K$.");
           prm.declare_entry ("K0", "4.111e11",
                              Patterns::Double (0),
                              "Core compressibility at zero pressure. "
                              "Referring to Nimmo et al. (2004) for more details.");
           prm.declare_entry ("Rho0", "7.019e3",
                              Patterns::Double (0),
-                             "Core density at zero pressure. Units: $kg/m^3$. "
+                             "Core density at zero pressure. Units: $kg/m^3$."
```
tjhei: Can you add that space again?

bangerth: Oh, I totally missed the following sentence -- we always have the units at the end.

Fixed! (I also fixed language slightly.)

PR #2240, opened 2018-05-10
www: update links to 2.0

tjhei:


cite.html
```diff
@@ -41,7 +41,7 @@ <h1>
 
             <p><b>1. Body text</b>
 	      <pre><code>
-Computations were done using the ASPECT code version 1.5.0, see \cite{heister_aspect_methods2,...}.</code></pre>
+Computations were done using the ASPECT code version 2.0.0, see \cite{heister_aspect_methods2,...}.</code></pre>
```
bangerth: We should really remove the `,...` here. That's just lazy.

header.include
```diff
@@ -40,8 +40,9 @@
               <a href="#" class="dropdown-toggle" data-toggle="dropdown">Development<b class="caret"></b></a>
               <ul class="dropdown-menu">
                 <li><a href="../doc/doxygen/index.html">Documentation</a></li>
-                <li><a href="../doc/doxygen/changes_current.html">Changes after v1.5.0</a></li>
+                <li><a href="../doc/doxygen/changes_current.html">Changes after v2.0.0</a></li>
 
+		<li><a href="../doc/doxygen/changes_between_1_85_80_and_2_80_80.html">Changes v1.5&#10142;2.0</a></li>
```
bangerth: missing space for alignment

tjhei: fixed

cite.html
tjhei: This is there on purpose. Can we discuss maybe during the hackathon?

bangerth: Yes. I think I know what you mean, though if you hope that a user puts something specific into these dots, then you should say what exactly right here. (Unless that comes next -- github just shows me this one line right now.)

PR #2241, opened 2018-05-14
Better document what 'spherical' means in 2d.

bangerth:


bangerth:
Can the chunk geometry even be used in 2d? The discussion is specifically about how a geometry is to be interpreted in 2d. I thought the chunk geometry is specific to 3d.

MFraters:
The ellipsoidal chunk is only 3d, but the chunk has a 2d case.

bangerth:
I've added two more commits: the second one is just whitespace changes to use our common coding style of 3 empty lines between functions; the third one documents the issue in 2d for the chunk geometry.

source/geometry_model/chunk.cc
```diff
@@ -707,17 +747,30 @@ namespace aspect
...
                                    "edges of geographical quadrangle (in degrees)"
                                    "Chunk (inner || outer) radius: Radii at bottom and top of chunk"
                                    "(Longitude || Latitude || Radius) repetitions: "
-                                   "number of cells in each coordinate direction.")
+                                   "number of cells in each coordinate direction."
+                                   "\n\n"
+                                   "When used in 2d, this geometry does not imply the use of "
+                                   "a spherical coordinate system. Indeed, "
+                                   "in 2d the geometry is simply a sector of an annulus in a Cartesian "
+                                   "coordinate system and consequently would correspond to "
```
MFraters: I am rereading this, and I think the statement that we use a Cartesian coordinate system also holds for 3d geometry models. Although we have some wrappers which pretend everything is spherical, all the geometry models work in the end through Cartesian coordinates. Should we add that, or am I misinterpreting the text?

MFraters: Ok. Yes, I noticed this came from the mailing list :)

PR #2242, opened 2018-05-14
add a link to the ASPECT 2.0 webinar

jdannberg:
I thought we should link to the webinar so that people can go and watch it if they missed it and want to hear about the new features. If you think there's a better places than in the "News" column, let me know!


gassmoeller:
I think the indentation is off compared to the other items. Please fix, and then merge yourself.

PR #2245, opened 2018-05-15
update release tasklist

tjhei:


doc/release-tasklist
```diff
@@ -55,9 +55,10 @@
   . Tag the release:
       export TAG=$VER
       cd doc && ./bump_version.sh $TAG && cd .. && \
+      git commit -m "version $TAG" && \
       git tag -s v$TAG -m "version $TAG"
   . create a tar file:
-      cd doc && make manual.pdf && cp manual.pdf manual-$TAG.pdf && cd ..
+      cd doc && make manual.pdf && cp manual.pdf aspect-manual-$TAG.pdf && cd ..
```
bangerth: This is pre-existing, but doesn't this just create two copies of the same file in the .tar.gz file?

bangerth: Oh, then I was confused by the comments just one line above that says "create a tar file". And you create the tar file two lines down. So I'm probably confused what this whole block is actually doing...

bangerth: Looking at the link you show above, I think I understand a bit better what is happening here -- an extended comment might help. But I still don't understand what the `cp` operation is used for: where do we ever reference `aspect-manual-$TAG.pdf` again?

tjhei: We don't. This is the file I manually upload to the release later.

PR #2247, opened 2018-05-17
set required ASPECT version in plugin files

tjhei:


tjhei:
@gassmoeller turns out I am not trusted. ;-)

gassmoeller:
Oh, did you check if I put in the correct email address?

tjhei:
> Oh, did you check if I put in the correct email address?

I don't know. Maybe we should print the email address before bailing out?

I just triggered a trusted build manually...

tjhei:
> Maybe we should print the email address before bailing out?

see #2248

PR #2248, opened 2018-05-17
Jenkins: print some info before failing

tjhei:


tjhei:
1. why did this build without me having to allow the build?
2. I guess it uses my github default email address which is @clemson.edu. I guess I need to fix that. ;-)

gassmoeller:
Looks good.

PR #2253, opened 2018-05-18
Fix base model initialization for depth dependent material model

### LINKED ISSUE #2250: AsciiDataLookup error when using the 'Ascii reference profile' and 'Depth dependent' material models
This is a follow-up to an email conversation on aspect-devel (subject: 'Error on depth dependent material model').

When using the `Ascii reference profile` material model in combination with the `Depth dependent` material model a segmentation fault as a result of a null pointer in the `AsciiDataLookup` function.

This can be produced by modifying the `Material model` section of  `tests\burnman_test.prm to include
```
  subsection Depth dependent model
    set Base model = ascii reference profile
    set Depth dependence method = None
  end
``` 

The error traces back  to line 93 of ascii_reference_profile.cc, where the third value from the ascii table (thermal expansivity) is read in.
`out.thermal_expansion_coefficients[i] = profile.get_data_component(profile_position,thermal_expansivity_index);`

I haven't had a chance to go through the debugger again, but perhaps the variable for the ascii file name one of the two material model inputs (likely ascii_reference_profile input) is being overwritten?

gassmoeller:
This should fix #2250. The depth dependent material model did not initialize or update its base model, which is fine for most material models, but not for the ones that require initialization (like ascii reference profile). This is now done correctly, and at the same time allows to prescribe time-dependent viscosity functions.


tests/depth_dependent_box_function_time_dependent.prm
```diff
@@ -0,0 +1,175 @@
+# Test depth-dependent material model using simple material model as
+# base model with depth-dependence specified by user-provided function
+
+# A description of convection in a 2d box. See the manual for more information.
```
jdannberg: It looks like this is copied from somewhere. Can you either remove it, or just say which test/benchmark you copied?

tests/depth_dependent_box_function_time_dependent.prm
```diff
@@ -0,0 +1,175 @@
+# Test depth-dependent material model using simple material model as
+# base model with depth-dependence specified by user-provided function
```
jdannberg: Can you say that this tests the time dependence of the function in the depth dependent material model?

tests/depth_dependent_initialize_base_model.prm
```diff
@@ -0,0 +1,146 @@
+# A test using adiabatic conditions and a gravity 
+# profile generated by a mineral physics toolkit and read in
+# through an ascii data file. In this example, material properties
+# are based on a Birch-Murnaghan equation of state.
```
jdannberg: Can you update this comment?

bangerth:
+1 -- obscure bug not easy to figure out!

PR #2254, opened 2018-05-19
Update the citation page.

bangerth:
Specifically: `<pre>` blocks are not formatted, so long lines break awkwardly.
Also update the year of the manual.

### DIFF
--- a/cite.html
+++ b/cite.html
@@ -55,8 +55,10 @@ We thank the Computational Infrastructure for Geodynamics (geodynamics.org) whic
 	      The canonical reference for this software is:
               <pre><code>
 @Article{heister_aspect_methods2,
-  Title                    = {High Accuracy Mantle Convection Simulation through Modern Numerical Methods. {II}: {R}ealistic Models and Problems},
-  Author                   = {Heister, Timo and Dannberg, Juliane and Gassm{\"o}ller, Rene and Bangerth, Wolfgang},
+  Title                    = {High Accuracy Mantle Convection Simulation through Modern Numerical Methods.
+                              {II}: {R}ealistic Models and Problems},
+  Author                   = {Heister, Timo and Dannberg, Juliane and
+                              Gassm{\"o}ller, Rene and Bangerth, Wolfgang},
   Journal                  = {Geophysical Journal International},
   Year                     = {2017},
   Number                   = {2},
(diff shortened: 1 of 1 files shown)


bangerth:
And be more consistent in how we spell the bibtex fields.

cite.html
```diff
@@ -77,21 +79,20 @@ <h1>
...
   address      = {Davis, CA},
 }
 </code></pre>
               The manual's proper reference is:
               <pre><code>
 @Article{aspectmanual,
-  Title         = {{\textsc{ASPECT}: Advanced Solver for Problems in Earth's ConvecTion, User Manual}},
-  Author        = {W. Bangerth and J. Dannberg and R. Gassm{\"o}ller and T. Heister and others},
-  Year          = {2017},
-  Month         = {4},
```
tjhei: did you delete the month on purpose?

tjhei: this is what figshare has:
```
@misc{bangerth_dannberg_gassmoeller_heister__2018, title={ASPECT: Advanced Solver for Problems in Earth's ConvecTion, User Manual}, url={https://figshare.com/articles/ASPECT_Advanced_Solver_for_Problems_in_Earth_s_ConvecTion_User_Manual/4865333/4}, DOI={10.6084/m9.figshare.4865333.v4}, abstractNote={This is the manual of the ASPECT mantle convection code}, publisher={figshare}, author={Bangerth, Wolfgang and Dannberg, Juliane and Gassmoeller, Rene and Heister, Timo and others}, year={2018}, month={May}}
```

bangerth: I suspected that you would end up asking about this, because it was also not clear to me. I dropped the month because we update the manual all the time, and consequently with the month the date will essentially *always* be outdated. By providing only the current year, this is at least the case only once a year.

What is relevant to ask is whether the entry here must correspond to a particular figshare entry? The figshare entry is *also* permanently outdated. I don't know how to rectify that, and I don't know whether when we reference the *manual*, we should be tied to the *figshare* DoI date, or whether we are better off if we just take the DoI from the letter, but logically still imply that one is referencing the current state of the manual.

Guidance?

tjhei: > must correspond to a particular figshare entry?

Well, you use the DOI of the figshare release, so yes I think you should use their information.

> What is relevant to ask is whether the entry here must correspond to a particular figshare entry? The figshare entry is also permanently outdated.

We only upload the releases to figshare, so it is technically not "outdated" and the release month refers to the month of that release.

bangerth: OK, I can revert to the previous version of the entry.

But before I do this, shouldn't we then upload the 2.0 manual there, with the current date?

tjhei: > OK, I can revert to the previous version of the entry.

No, it should be May 2018 like I posted above.

> But before I do this, shouldn't we then upload the 2.0 manual there, with the current date?

I am confused. I did upload 2.0. That is the version 4 if you click on the figshare link.

bangerth: I was confused. I believed that because the entry here referred to last year's release, that the manual had not been updated on figshare. But you say that you did, and I just confirmed that you did. So I left the year at 2018 and put the month back in. 

Is this what you intended?

tjhei: yes, that looks good! :+1:

PR #2257, opened 2018-05-20
jenkins: fix info stage

tjhei:
not sure why it worked before, but the current code generates:

No such property: CHANGE_AUTHOR_EMAIL for class: groovy.lang.Binding

because it is interpreted as groovy instead of a shell variable.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,9 +19,9 @@ pipeline {
   stages {
     stage ("info") {
       steps {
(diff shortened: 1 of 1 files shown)


tjhei:
see https://stackoverflow.com/questions/42681184/change-author-email-and-change-id-environment-variables-return-no-such-property

gassmoeller:
Looks good. And yes I struggled with the same issue at some point.

Issue #2252, opened 2018-05-18
A bug in advection assemblers


YiminJin:
I believe that I have found a bug in assemblers of advection fields across all versions of ASPECT.

In struct internal::Assembly::Scratch::AdvectionSystem, there are pointers of FEFaceValues and FESubfaceValues, which will be newed by the constructor if discontinuous Galerkin method is chosen to solve the advection system. However, the copy constructor copies the pointer rather than FEFaceValues itself to the new object. So if the user use multi-threads, there will be several threads visiting the same auxiliary arrays in FEFaceValues at the same time, which will lead to mistake.

Since FEFaceValues cannot be copied, I think the correct way is to new a same one in the copy constructor.

Please let me know if it is really a bug, or I've made something wrong.

Thank you very much!

gassmoeller:
Hi YiminJin, welcome, and thanks for reporting this! This looks indeed like it could be a bug. We only introduced the multithreading support recently, and apparently this case was not tested. Could you post a minimal case that seems to cause the error? Is it enough to run aspect with -j on one of the discontinuous tests?
Thanks.

YiminJin:
You can add a line "set Use discontinuous composition discretization = true" in /tests/multicomponent_harmonic.prm, and try to run this test with -j. The output on my computer is:


-----------------------------------------------------------------------------
-- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.
--     . version 2.1.0-pre (master, 86d90e200)
--     . using deal.II 8.5.1
--     . using Trilinos 12.12.1
--     . using p4est 1.1.0
--     . running in DEBUG mode
--     . running with 1 MPI process
--     . using 8 threads 
-- How to cite ASPECT: https://aspect.geodynamics.org/cite.html
-----------------------------------------------------------------------------

Number of active cells: 4,096 (on 6 levels)
Number of degrees of freedom: 164,964 (33,410+4,257+16,705+36,864+36,864+36,864)

*** Timestep 0:  t=0 years
   Solving temperature system... 0 iterations.
   Solving C_1 system ... 0 iterations.
   Solving C_2 system ... 0 iterations.
   Solving C_3 system ... 0 iterations.
   Rebuilding Stokes preconditioner...
   Solving Stokes system... 30+12 iterations.


--------------------------------------------------------
An error occurred in line <3457> of file </home/.../build/dealii/dealii-8.5.1/source/fe/fe_values.cc> in function
[...]

PR #2258, opened 2018-05-21
fix bug described in [issue #2252]

YiminJin:
This branch fixes the bug described in [issue #2252]. The modifications mostly follow the comments of @bangerth . This branch runs well with /tests/discontinuous_composition_1.prm using multi-threads now, while the master branch would fail.

However, there is a small problem when running ./aspect -j $PATH_TO_TESTS/discontinuous_composition_1.prm.  The output on my computer is like this:

`-----------------------------------------------------------------------------
--     . version 2.1.0-pre (iss#2252, 5f43cb715)
--     . using deal.II 8.5.1
--     . using Trilinos 12.12.1
--     . using p4est 1.1.0
--     . using 8 threads 
-- How to cite ASPECT: https://aspect.geodynamics.org/cite.html

... (timesteps)

Termination requested by criterion: end time

+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      1.26s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble Stokes system          |        18 |    0.0698s |       5.5% |
| Assemble composition system     |        36 |     0.399s |        32% |
| Assemble temperature system     |        18 |     0.144s |        11% |
| Build Stokes preconditioner     |        18 |    0.0992s |       7.9% |
| Build composition preconditioner|        36 |    0.0498s |         4% |
| Build temperature preconditioner|        18 |   0.00621s |      0.49% |
| Solve Stokes system             |        18 |    0.0524s |       4.2% |
| Solve composition system        |        36 |    0.0209s |       1.7% |
| Solve temperature system        |        18 |   0.00952s |      0.76% |
| Initialization                  |         1 |     0.035s |       2.8% |
| Postprocessing                  |        17 |    0.0254s |         2% |
| Refine mesh structure, part 1   |         9 |    0.0585s |       4.6% |
| Refine mesh structure, part 2   |         9 |    0.0064s |      0.51% |
| Setup dof systems               |        10 |    0.0382s |         3% |
| Setup initial conditions        |         2 |   0.00735s |      0.58% |
+---------------------------------+-----------+------------+------------+

WARNING! There are options you set that were not used!
WARNING! could be spelling mistake, etc!
Option left: name:-j value: ../tests/discontinuous_composition_1.prm
`

I don't know what the warnings mean. I'm sure that multi-thread works because the program runs faster than running without -j.

I'm a beginner of github and I don't know if it is appropriate to say so much here. The same description of the new problem is also pasted in [issue #2252 ] and I hope that someone would figure it out.

Please check my modifications. Thank you!

### DIFF
--- a/include/aspect/simulator/assemblers/interface.h
+++ b/include/aspect/simulator/assemblers/interface.h
@@ -216,9 +216,9 @@ namespace aspect
 
           FEValues<dim> finite_element_values;
 
-          std_cxx11::shared_ptr<FEFaceValues<dim> >    face_finite_element_values;
-          std_cxx11::shared_ptr<FEFaceValues<dim> >    neighbor_face_finite_element_values;
-          std_cxx11::shared_ptr<FESubfaceValues<dim> > subface_finite_element_values;
+          std_cxx11::unique_ptr<FEFaceValues<dim> >    face_finite_element_values;
+          std_cxx11::unique_ptr<FEFaceValues<dim> >    neighbor_face_finite_element_values;
+          std_cxx11::unique_ptr<FESubfaceValues<dim> > subface_finite_element_values;
 
           std::vector<types::global_dof_index>   local_dof_indices;
 
--- a/source/simulator/assemblers/interface.cc
+++ b/source/simulator/assemblers/interface.cc
@@ -176,27 +176,27 @@ namespace aspect
           finite_element_values (mapping,
                                  finite_element, quadrature,
                                  update_flags),
-          face_finite_element_values ((face_quadrature.size() > 0
-                                       ?
-                                       new FEFaceValues<dim> (mapping,
-                                                              finite_element, face_quadrature,
-                                                              face_update_flags)
-                                       :
-                                       NULL)),
-          neighbor_face_finite_element_values ((face_quadrature.size() > 0
-                                                ?
-                                                new FEFaceValues<dim> (mapping,
-                                                                       finite_element, face_quadrature,
-                                                                       face_update_flags)
-                                                :
-                                                NULL)),
-          subface_finite_element_values ((face_quadrature.size() > 0
-                                          ?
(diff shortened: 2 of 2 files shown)


bangerth:
@YiminJin -- fantastic, many thanks! I'm so glad you told me about this issue, and that I prodded you into sending us a report about it!

The patch is ok to merge if the tester succeeds.
/run-tests

About the warnings: I think you are using PETSc, which has habit of warning about run-time flags it has no business warning about. You probably have these warnings for every run you do, right? If so, I would ignore them.

@geodynamics/aspect-developers: This patch has no testcase, but I don't know how one would create one either: the existing testcase only fails if run with more than one thread, but I don't think we have a way to ensure a test is run with more than one thread, right?

gassmoeller:
Yes you are right there is no way to test multithreading right now. We should probably set up a way to test it, e.g. have a tester that runs everything with multithreading, but we would first need to check if the test results actually stay the same (i.e. if order of assembly can lead to slightly different results). I will open a separate issue about that. @YiminJin: Thanks very much for the fix!. The failure in the new integration tester was resolved by #2257 and the old tester looks good, so I am going to merge this.

YiminJin:
@bangerth -- I'm glad to join the contributors of ASPECT. About the warnings, they only appears when using multi-threads. Besides, I have installed deal.II with both Trilinos and PETSc, and did not turn on the PETSc switch in ASPECT. It's just some nonsense produced by PETSc automatically. Gassmoeller has sent me a patch in comments of issue #2252  that fixes the problem, which filters out all of the ASPECT command line options before initializing PETSc in main(). I think this patch is worthy to be merged into the master branch to help people using PETSc get rid of those annoying warnings.

bangerth:
@YiminJin -- ok. I'll let @gassmoeller propose that patch if he think it's useful!

PR #2265, opened 2018-05-23
disable multithreading with PETSc

tjhei:


bangerth:
Oh, why is this? Don't we carefully synchronize the assembly of the linear systems, for example? Or are you thinking of the postprocessors accessing data vectors in parallel, for example?

tjhei:
read access can not happen on two threads because we (lock, read, unlock). This is asserted here: https://github.com/dealii/dealii/blob/master/source/lac/petsc_vector_base.cc#L119

bangerth:
?? So you say that we (lock, read, unlock), but that we also just error out if threading is enabled? Then why do we do the lock/unlock at all?

But if I understand you correctly, then you want to disable threading with PETSc to avoid the assertion in `VectorBase`. Is this correct?

tjhei:
> Then why do we do the lock/unlock at all?

Oh, not a mutex, but to read from a PETSc vector we use VecGetArray()/VecRestoreArray() and this operation is not thread-safe. To avoid this problem, we assert that multithreading is not enabled when you create a PETSc vector.

> But if I understand you correctly, then you want to disable threading with PETSc to avoid the assertion in VectorBase. Is this correct?

Yes.

PR #2270, opened 2018-05-30
Update a comment.

bangerth:
I noticed this while talking to @menno-fraters about the Newton solver.
The code in question here is used for both the preconditioner of the regular
Stokes solver as well as for the Newton-Stokes solver where that one matrix
block has a different meaning.


MFraters:
Yes, this makes the purpose of the reference viscosity way more clear :)

PR #2281, opened 2018-06-13
add some guidelines on how to write a cookbook

jdannberg:
This is just a suggestion, I don't think we ever discussed what the main points are when writing a cookbook, but I just thought we should have some general guidelines we can point people to in case they want to contribute a cookbook (e.g. at the hackathon).


tjhei:
Very well-written, thanks! I am wondering if it is better to be placed in the manual itself instead? As a bonus, you can also link to the correct sections that way.

Otherwise, maybe mention the ``benchmarks/`` folder too and that there can be other files like ``Makefile``s or scripts?

jdannberg:
Okay, I moved this to the .tex file and added the section you mentioned.

gassmoeller:
Looks good.

PR #2278, opened 2018-06-13
cleanup latex formatting of manual

tjhei:
to be able to compile to html, we need a couple of fixes:
- change x_\bla to x_{\bla}
- specify image extensions
- some minor changes

related to #2231

PR #2284, opened 2018-06-14
Avoid using the asterisk as multiplication operator in formulas.

bangerth:
They are not commonly used in mathematical notation and look quite ugly in latex output.
While there, also (i) avoid duplication of information, and (ii) fix a mistake in a
formula. These last two changes are in separate commits.


bangerth:
Found while looking at #2278.

source/particle/integrator/rk_4.cc
```diff
@@ -197,7 +197,8 @@ namespace aspect
       ASPECT_REGISTER_PARTICLE_INTEGRATOR(RK4,
                                           "rk4",
                                           "Runge Kutta fourth order integrator, where "
-                                          "$y_{n+1} = y_n + (1/6)*k_1 + (1/3)*k_2 + (1/3)*k_3 + (1/6)*k_4$ "
+                                          "$y_{n+1} = y_n + \\frac{1}{6} k_1 + \\frac{1}{3} k_2 "
+                                          "           + \\frac{1}{3} k_3 + \\frac{1}{6} k_4$ "
```
tjhei: why do you start with this much whitespace? This will look weird without the newline, doesn't it?

include/aspect/material_model/visco_plastic.h
```diff
@@ -63,8 +63,8 @@ namespace aspect
      *
      * Viscous deformation is defined by a viscous flow law describing
      * dislocation and diffusion creep:
-     *   $ v = 0.5 * A^{-\frac{1}{n}} * d^{\frac{m}{n}} *
-     *               \dot{\varepsilon}_{ii}^{\frac{1-n}{n}} *
+     *   $ v = 0.5   A^{-\frac{1}{n}} d^{\frac{m}{n}}
```
tjhei: \frac{1}{2} while you are here?

source/particle/integrator/rk_4.cc
bangerth: Latex don't care about spaces in formulas. I just wanted it to be readable in the source files :-)

bangerth:
I've pushed another commit concerning the 0.5 -> frac{1}{2} conversion.

source/particle/integrator/rk_2.cc
```diff
@@ -154,8 +154,8 @@ namespace aspect
       ASPECT_REGISTER_PARTICLE_INTEGRATOR(RK2,
                                           "rk2",
                                           "Second Order Runge Kutta integrator "
-                                          "$y_{n+1} = y_n + dt*v(t_{n+1/2}, y_{n} + 0.5*k_1)$ "
-                                          "where $k_1 = y_n + 0.5*dt*v(t_{n}, y_{n})$")
+                                          "$y_{n+1} = y_n + \\Delta t\\, v(t_{n+1/2}, y_{n} + \\frac{1}{2} k_1)$ "
+                                          "where $k_1 = \\Delta t\, v(t_{n}, y_{n})$")
```
tjhei: you need to write "t\\," instead of "t\,"

source/particle/integrator/rk_4.cc
tjhei: My point is that it looks like this in the prm and gui now:
```
    # `rk4': Runge Kutta fourth order integrator, where $y_{n+1} = y_n +
    # \frac{1}{6} k_1 + \frac{1}{3} k_2            + \frac{1}{3} k_3 +
    # \frac{1}{6} k_4$ and $k_1$, $k_2$, $k_3$, $k_4$ are defined as usual.
```

source/particle/integrator/rk_2.cc
bangerth: Ah, good grief. But good eyes! Fixed now!

source/particle/integrator/rk_4.cc
bangerth: Ah, hm. That's unfortunate. We really ought to come up with a way to print real formulas in the GUI -- Qt can do everything, can't it do that as well?

Either way, I've fixed this as well.

bangerth:
ping?

gassmoeller:
It seems like all comments have been addressed.

Issue #2262, opened 2018-05-22
2.0.1 Release


gassmoeller:
During the last two weeks we found a few bugs in the 2.0.0 release that might make it necessary to do a 2.0.1, because they cause crashes of valid input files, or silently produce wrong results. Specifically, I am thinking of the following:

- #2258 
- #2253 
- the bug in heating_model/compositional_heating that is fixed in #2199 (we might need a different smaller fix for the old version, or just pull the whole thing into 2.0.1, no functional change, just refactoring)
- the fix to the bug reported in #2243 (yet to be implemented in ASPECT, part in dealii/dealii#6630)
- another bug that prevents combining particles and free surface in ASPECT 2.0.0 (I need to open an issue and PR for that)

We might also decide to just drop support for combining particles and free surface (the last two issues) by explicitly disabling that for the moment (I have not tested it, it is just not explicitly disabled).

Should we wait for a few weeks and do a 2.0.1 before the hackathon?

jdannberg:
Yes, I think that makes sense.

bangerth:
Is anyone likely to run into any of these bugs? Things that just fail with an error are annoying, but at least they lead to an error one can report and be told to use the dev sources instead. (Though they may scare people away from using ASPECT.) The ones that silently produce wrong results concern me more. #2199 is of that sort.

The alternative is to stabilize for a month after the hackathon and then put out a 2.1. The people who go to the hackathon will all use the dev sources anyway, so they're not a concern. I do wonder, however, whether we're going to be able to stabilize soon after the hackathon. We've usually had a very long tail of things from the hackathon...

tjhei:
> that is fixed in #2199 (we might need a different smaller fix for the old version, or just pull the whole thing into 2.0.1

I think we can pull the whole thing...

gassmoeller:
We should now have everything that is necessary for 2.0.1. Lets move forward with it.

tjhei:
clang 6 with deal.II 9.1.0-pre fails with:
```
[...]

PR #2289, opened 2018-06-14
[2.0.1] pick #2199

tjhei:
see #2262


tjhei:
args, this seems to be a github bug, I can not reopen the PR, so see #2290

PR #2293, opened 2018-06-15
Add GRMES solver restart length as an input parameter

diogolourenco:
Add variable "GRMES solver restart length" as an input parameter. Solves convergence problems due to high viscosity jumps in the domain, and very likely will be helpful in other sorts of problems. 

Right now I left (though edited) the original version of increasing the number of restart iterations to 100 if melt transport is included, but maybe that it is not necessary anymore since it is now an input parameter?

### DIFF
--- a/source/simulator/parameters.cc
+++ b/source/simulator/parameters.cc
@@ -347,6 +347,13 @@ namespace aspect
                            "not converge and return an error message pointing out that the user didn't allow "
                            "a sufficiently large number of iterations for the iterative solver to converge.");
 
+	prm.declare_entry("GMRES solver restart length", "50",
+			  Patterns::Integer(0),
+			  "This is the number of iterations that define the GMRES solver restart length. It "
+			  "helps with convergence issues arising from high viscosity jumps in the domain. "
+			  "Increasing this number to 200 iterations allows convergence with several orders "
+			  "magnitude viscosity jumps.");
+
         prm.declare_entry ("Linear solver A block tolerance", "1e-2",
                            Patterns::Double(0,1),
                            "A relative tolerance up to which the approximate inverse of the $A$ block "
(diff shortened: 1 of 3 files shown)


diogolourenco:
Changes done! Let me know if something else is missing or should be fixed.

source/simulator/parameters.cc
```diff
@@ -1028,6 +1036,7 @@ namespace aspect
         n_expensive_stokes_solver_steps = prm.get_integer ("Maximum number of expensive Stokes solver steps");
         linear_solver_A_block_tolerance = prm.get_double ("Linear solver A block tolerance");
         linear_solver_S_block_tolerance = prm.get_double ("Linear solver S block tolerance");
+	stokes_gmres_restart_length     = prm.get_integer("GMRES solver restart length");
```
bangerth: You'll probably have to indent this line like the others.

source/simulator/solver.cc
```diff
@@ -806,7 +806,13 @@ namespace aspect
         // it in n_expensive_stokes_solver_steps steps or less.
         catch (SolverControl::NoConvergence)
           {
-            const unsigned int number_of_temporary_vectors = (parameters.include_melt_transport ? 100 : 50);
+            unsigned int number_of_temporary_vectors = parameters.stokes_gmres_restart_length;
+
+	    // use at least a restart length of 100 for melt models
+	    if (parameters.include_melt_transport &&
+		number_of_temporary_vectors < 100 )
+	      number_of_temporary_vectors = 100;
```
bangerth: You can combine all of this into one statement like
```
  const unsigned int number_of_... = (parameters.include_melt_transport == false
                                                           ?
                                                           parameters.stokes_gmres_restart_length
                                                           :
                                                           std::max (parameters.stokes_gmres_restart_length, 100U));
```
This has the advantage that you can make the variable `const` right away.

diogolourenco:
Changes done according to comments from Wolfgang Bangerth.

gassmoeller:
You will need to update some indentation according to our reference style. The easiest way to fix this is to click on `Details` next to the `aspect-astyle-hd` tester right below this comment. Then download the `astyle-changes.diff` file, save it into your aspect folder, and do a `patch -p 1 < changes-astyle.diff`. This should make the code conform to our formatting guidelines.

gassmoeller:
Nice, welcome to the list of ASPECT contributors :tada:

diogolourenco:
Thank you!

PR #2295, opened 2018-06-15
fix glob for changes.h in Makefile

tjhei:
fix my mess created in #2233

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -7,7 +7,7 @@ aspect-files      = $(shell echo ../include/aspect/*.h \
                             ../include/aspect/*/*/*.h   \
                             modules/*.h)
 aspect-manual-pix = $(shell find manual/cookbooks/ -name \*.png 2>/dev/null)
-change-files = $(shell ls modules/changes)
+change-files = $(shell find modules/changes/ -type f -not -name '.*')
 
 aspect-prms = $(shell find manual/cookbooks/ -name \*.prm 2>/dev/null)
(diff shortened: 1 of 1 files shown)


doc/Makefile
```diff
@@ -7,7 +7,7 @@ aspect-files      = $(shell echo ../include/aspect/*.h \
                             ../include/aspect/*/*/*.h   \
                             modules/*.h)
 aspect-manual-pix = $(shell find manual/cookbooks/ -name \*.png 2>/dev/null)
-change-files = $(shell ls modules/changes)
+change-files = $(shell find modules/changes/ -type f -not -name '.*')
```
bangerth: I don't think I understand what this does. It selects all files that do not have the form `.*`? What files are left in that case?

tjhei: I added a comment to explain

doc/Makefile
```diff
@@ -7,7 +7,11 @@ aspect-files      = $(shell echo ../include/aspect/*.h \
                             ../include/aspect/*/*/*.h   \
                             modules/*.h)
 aspect-manual-pix = $(shell find manual/cookbooks/ -name \*.png 2>/dev/null)
-change-files = $(shell ls modules/changes)
+
+# List all changelog files including the relative path and make sure to ignore
+# the .gitkeep file. Use find instead of ls because the directory might have
+# no changes in it:
+change-files = $(shell find modules/changes/ -type f -not -name '.*')
```
bangerth: Ah, doh -- `find` takes the point literally, not as a regex character...

PR #2296, opened 2018-06-15
Maintenance jun18

tjhei:
based on #2295


bangerth:
Take the first commit out. Merging the second and third is pre-approved.

PR #2298, opened 2018-06-16
manual: rename section

tjhei:
after longer discussion

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -330,7 +330,7 @@ which is supported by the National Science Foundation award EAR-1550901.
   \end{center}
 }
 
-\section{Geodynamic Models and Numerical Methods in \aspect{}}
+\section{Geodynamic modeling assumptions and numerical methods in \aspect{}}
 \label{sec:models}
 
 \subsection{Basic equations}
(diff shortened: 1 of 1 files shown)


doc/manual/manual.tex
```diff
@@ -330,7 +330,7 @@ \subsection{Acknowledgments}
   \end{center}
 }
 
-\section{Equations, models, coefficients}
+\section{Geodynamic Models and Numerical Methods in \aspect{}}
```
bangerth: For consistency with all other headings, don't capitalize any but the first word.

tjhei:
> Out of curiosity, how did you come up with "Geodynamic models"? That seems a rather broad term.

It is the best name we came up with that describes how we model the problem as a PDE and what approximation we use (density parameterization etc).

bangerth:
It seems too generic to me. Why not "Models for mantle convection"?

doc/manual/manual.tex
bangerth: @tjhei -- ping?

gassmoeller:
Because we are not longer only a code for mantle convection. What about "Geodynamic modelling assumptions and numerical methods"? We discussed a larger number of different options, and this was the best we came up with.

bangerth:
Yes, I like that title better. Can you make the change and merge yourself?

gassmoeller:
I would like @tjhei or @jdannberg to look over this and make the decision on merging it.

gassmoeller:
Could someone just push the merge button? I am happy with the current state, but I modified it, and dont want to merge my own change.

Issue #1476, opened 2017-05-01
Unify spelling of first names in the manual list of references


bangerth:
`doc/manual/manual.bib` uses many different ways to abbreviate or not abbreviate first names of authors. Make this uniform -- typically by only using initials, terminated by a period, and separated by a space.

bangerth:
Fixed by #2308.

PR #2305, opened 2018-06-17
Abbreviate first names for manuals.

bangerth:
In the list of references of the manual, we list the 1.5 and 2.0 manuals right
next to the figshare link to the generic manual. The latter uses abbreviated
first names, but the former written out first names. This looks awkward. Make
it look the same.

Somewhat related to #1476.

### DIFF
--- a/doc/manual/manual.bib
+++ b/doc/manual/manual.bib
@@ -11285,10 +11285,10 @@ year = {1992}
 
 @misc{aspect-doi-v1.5.0,
   title        = {{ASPECT} v1.5.0 [software]},
-  author       = {Wolfgang Bangerth and
-                  Juliane Dannberg and
-                  Rene Gassmoeller and
-                  Timo Heister and
+  author       = {W. Bangerth and
+                  J. Dannberg and
+                  R. Gassmoeller and
+                  T. Heister and
                   others},
   month        = {mar},
   year         = {2017},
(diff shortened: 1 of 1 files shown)


tjhei:
1. We need to keep cite.html in sync with any change  here.
2. I thought bibtex will (depending on style) shorten first names. Wouldn't it be better to provide the full names instead?

bangerth:
On 06/17/2018 09:17 AM, Timo Heister wrote:
>  1. We need to keep cite.html in sync with any change here.

I can do that if you want me to, but as before it has to be in a separate 
patch. Out of curiosity, why do the two have to be in sync?


>  2. I thought bibtex will (depending on style) shorten first names. Wouldn't
>     it be better to provide the full names instead?

The issue with full names is that you generally don't know them for all 
papers. In some papers, authors are only identified with initials, and in 
those cases you can't put anything other than initials into manual.bib. So 
this then also shows up in manual.pdf. My preference is therefore to simply 
use initials for *everyone* because that's the only way to stay consistent.

I had indeed thought that bibtex abbreviates first names, but that doesn't 
generally be the case. It's a question of the style file. We use the 
`alpha.bst` style, which apparently doesn't do this. But it could be easy to 
adjust, see
https://tex.stackexchange.com/questions/36660/only-authors-initials-in-bibtex-natbib-using-named-style
We'd just have to clone the alpha style.

tjhei:
> Out of curiosity, why do the two have to be in sync?

Well, I think that this is "a very good idea". For example, we only test the one in manual.bib, so the cite.html can be broken otherwise (and it was).



> My preference is therefore to simply
> use initials for *everyone* because that's the only way to stay consistent.

I see that it makes sense for our manual, but I prefer cite.html to list the full names.

PR #2306, opened 2018-06-17
Fix month field in a bibtex entry.

bangerth:
bibtex does not understand 'month=5'. It needs to be 'month={may}'.


tjhei:
Wait, didn't you just make the change may->5 in cite.html? Does ``month={5}`` work? Either way, please update cite.html too.

bangerth:
The old way was to use `month={5}` (that's the *before* state of this patch), but looking at `manual.pdf`, this doesn't work.

I'll fix this in `cite.html` as well, but because that's a different root in the repo, I have to do this in a different PR.

bangerth:
See #2307.

Issue #2231, opened 2018-05-09
html manual?


tjhei:
We discussed generating an html version of the manual in the past. I have played with ``make4ht`` and have a preliminary version here: http://www.math.clemson.edu/~heister/aspect-manual/manual.html

While many things are broken, it is quite useful to link people to a specific section in the manual, which is difficult in a pdf. Does it make sense to continue working on this? Note that some of the images/links/references/etc. are currently broken.

bangerth:
Neat. I think this is useful indeed, if you can get the currently broken parts going. 

If you want to link to individual sections, you also need stable links to sections. Does make4ht give you that?

tjhei:
> If you want to link to individual sections,

Section names are ``manualseX.html`` where X is the section number, which should be somewhat stable. Subsections of course change, so that is an issue. I will play with this a bit.

MFraters:
Looks cool Timo! Is this automatically loaded/linked from github every time you load a page, or do you have to compile it every now and then like the pdf?

tjhei:
> automatically loaded/linked

right now it is generated manually.

> to compile it every now and then like the pdf?

the pdf is generated every night.

tjhei:
how about this? http://www.math.clemson.edu/~heister/aspect-manual/nav.xhtml

One .html per chapter (otherwise the file is too big) and links are by name.

jdannberg:
That's really nice!
How do you want to continue with this (fixing the figures, etc.)? Also let me know if you reach a point where you need someone to look through the document to see if the formatting looks good everywhere!

tjhei:
> fixing the figures

It seems like embedded pdfs look terrible: http://www.math.clemson.edu/~heister/aspect-manual/ch006.xhtml#sec:finite-strain while other images looks mostly okay (except sizing issues).

> Also let me know if you reach a point where you need someone to look through the document to see if the formatting looks good everywhere!

Maybe it is a bit too early. I think we should decide if this is promissing enough, then merge it as is and put it online before we fix other stuff. Hackathon starter projects? ;-)

bangerth:
Nice indeed!

Something also goes wrong with tables, see http://www.math.clemson.edu/~heister/aspect-manual/ch006.xhtml#sec:sinker-with-averaging and the latex formatting instructions that leak into the HTML. For the figures, one of the problems is that the PDF or PNG versions are in whatever size they come, and latex rescales them. I don't think this is happening here.

jperryhouts:
[...]

PR #2309, opened 2018-06-19
manual: switch to named anchor labels

tjhei:
right now the internal links are "section_2.4", etc.. This means they
are not stable when the number of sections change. Currently we don't
care, but this change allows us to deeplink into section of the pdf.
Demo:
http://www.math.clemson.edu/~heister/aspect-pdf-manual/web/viewer.html?file=b.pdf#sec%3Apressure

related to: https://github.com/geodynamics/aspect/issues/2231

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -55,7 +55,7 @@
 % files in the aspect tree without having to specify the
 % location relative to the directory where the pdf actually
 % resides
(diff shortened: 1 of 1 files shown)


bangerth:
Documentation of the parameter in question: https://www.tug.org/applications/hyperref/manual.html

Issue #2090, opened 2018-02-12
Verify checkpoint/restart agree on the fields a simulation has


bangerth:
Based on a recent mailing list post:

When we do a restart, we can only do this successfully if a number of "critical" input parameters are the same as they were in the saved run. It's going to be hard to enumerate what parameters these are, but we could at least output the number and names of the fields into the restart file, and compare them with the values we have when we do the restart. This information should be relatively easy to get out of the `Introspection` object.

bangerth:
#2310 implemented the basic structure and checks that the number and names of compositional fields is the same between the checkpointed and the restarted computation.

We should go through the list of parameters and determine whether there are others we can check in the functions that #2310 now provides. I think this is a much easier task, so I'll mark this as a starter project.

bangerth:
The list of parameters worth checking should all be declared in `parameters.h` (with description in `source/simulator/parameters.cc`.

PR #2310, opened 2018-06-19
Save and load some critical parameters during checkpoint/restart.

bangerth:
This allows us to do a basic integrity check upon restart time.

This has the potential to earn me the Puckett Bucket on day 1 of the hackathon,
but I'd still like to see what happens during tests.

In reference to #2090.


bangerth:
I don't think this is ready yet, but I wanted to see what the tester does with it.

tjhei:
This looks good. I assume the tests will pass and we can merge it...

bangerth:
Oh, miraculously, it actually worked successfully through the testsuite! I had fully expected to break a whole bunch of tests :-)

MFraters:
That seems like a great opportunity to add a test which would have failed and now works :)

bangerth:
Yes. I need to figure out how to do this, but it's on my list.

bangerth:
I pushed another commit with a new test that should current fail, and I want to see it fail on the tester. If that's what happens, then I'll add the `EXPECT FAIL` tag to it.

bangerth:
OK, it failed as hoped for. I've updated the test to expect failure -- let's see how this does now.

tjhei:
the test still fails: https://jenkins.tjhei.info/blue/organizations/jenkins/aspect/detail/PR-2310/5/pipeline

bangerth:
Yes, I still needed to update the test output. Let's try this again.

tjhei:
can you rebase to master please? Otherwise "bangerth-down" won't like your PR.

bangerth:
I need help. I can't figure out why this one test fails -- it succeeds on my system and I have no idea why it doesn't on the tester. It complains about different output files, but the numbers look the same to me up to several digits down. It says
```

##28      #:3   <== 1.224703447855e+06
##28      #:3   ==> 1.224703422203e+06
@ Absolute error = 2.5652000000e-2, Relative error = 2.0945479154e-8
```
But that's basically identical. Don't we have a relative error of 1e-6?

bangerth:
I think I see -- I just have to take the output file that the tester wants, not the one I get on my own machine :-(

gassmoeller:
Looks good, thanks for fixing!

Issue #2256, opened 2018-05-20
Fix grammar mistakes in simulator_access.h


bangerth:
As found in #2255, `simulator_access.h` contains text of the kind "this can then i.e. be used to get the names of the..." in a number of places.

This is grammatically incorrect. It should be "this can then, for example, be used to get the names of the..."

SophieCoulson:
I will finish fixing this!

bangerth:
@SophieCoulson -- great, thank you! You'll want to exclude the one place I've fixed in #2313, but you don't have to wait for that to be merged to get started!

jdannberg:
As #2313 and #2317 are merged now, this should be fixed.

PR #2317, opened 2018-06-19
Fix grammar mistake in simulator_access.h

SophieCoulson:
In response to issue #2256 and in addition to Wolfgang's correction #2313

### DIFF
--- a/include/aspect/simulator_access.h
+++ b/include/aspect/simulator_access.h
@@ -579,7 +579,7 @@ namespace aspect
 
       /**
        * Return an reference to the manager of the boundary composition models.
-       * This can then i.e. be used to get the names of the boundary composition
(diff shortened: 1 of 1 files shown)


MFraters:
Good start! I did find that you missed one (line 554). Could you fix that one as well and squash your commits?

jdannberg:
@MFraters I think the one in line 554 is in @bangerth's pull request #2313.

MFraters:
ah, you are correct :)

bangerth:
Yes, nice! For completeness, I'm going to let the tester machine run on this with the following special command that only maintainers can use:
/run-tests
I don't expect this to turn up anything, but we like to make sure anyway!

bangerth:
All clear -- yay!

PR #2318, opened 2018-06-19
Track both plastic and viscous strain

anne-glerum:
Hi @naliboff, I've rebased on master, could you have another look?


naliboff:
@anne-glerum - Taking a look now!

include/aspect/material_model/visco_plastic.h
```diff
@@ -56,6 +56,12 @@ namespace aspect
          * the current object.
          */
         std::vector<double> friction_angles;
+
+        /**
+         * The area where the viscous stress exceeds the plastic yield strenght,
```
naliboff: strenght -> strength

include/aspect/material_model/visco_plastic.h
```diff
@@ -56,6 +56,12 @@ namespace aspect
...
+         * and viscosity is rescaled back to the yield envelope.
+         */
+        std::vector<double> yielding;
```
naliboff: Could this variable be switched to an `unsigned int`?

include/aspect/material_model/visco_plastic.h
```diff
@@ -221,16 +227,86 @@ namespace aspect
...
+         * compositional field.
+         */
+        double
+        calculate_viscous_weakening ( const double strain_ii,
+                                      const unsigned int j ) const;
 
+        /**
+         * Whether to use the accumulated strain to weaken
+         * plastic parameters cohesion and friction angle
+         * and viscous parameters diffusion and dislocation prefactor.
```
naliboff: It might be worth making this wording a bit more specific. How about:

> Whether to use the accumulated total strain or plastic strain to weaken the cohesion and friction angle, and the viscous strain to weaken the refactors for diffusion and dislocation creep.

include/aspect/material_model/visco_plastic.h
```diff
@@ -221,16 +227,86 @@ namespace aspect
...
         bool use_strain_weakening;
+        /**
+         * Whether to use the accumulated plastic strain to weaken
+         * plastic parameters cohesion and friction angle.
+         * When both use_plastic_strain_weakening and
+         * use_viscous_strain_weakening are true, it is assumed
+         * the first 2 compositional fields are tracking the
+         * plastic and viscous strain respectively.
+         * TODO: add check for fields named plastic_strain and
+         * viscous_strain?
```
naliboff: Yes, I think it would be worth enforcing a naming scheme and having these values listed as the first and second compositional fields. The asserts could take the following form:

```
if (use_strain_weakening == true)
  {
    AssertThrow(this->introspection().compositional_index_for_name("plastic_strain") == 0,
                            ExcMessage("Material model ViscoPlastic only works if the first "
                                                   "compositional field is called plastic_strain."));
    AssertThrow(this->introspection().compositional_index_for_name("viscous_strain") == 1,
                           ExcMessage("Material model ViscoPlastic only works if the first "
                                                 "compositional field is called plastic_strain."));
  }
```

However, should similar functionality be include for when `use_strain_weakening == false` (first field could be `total_strain`) and for when we use the full strain tensor?

source/material_model/visco_plastic.cc
```diff
@@ -861,23 +955,67 @@ namespace aspect
...
+          use_viscous_strain_weakening     = prm.get_bool ("Use viscous strain weakening");
+          if (use_viscous_strain_weakening)
+          {
+            AssertThrow(use_strain_weakening,
+                        ExcMessage("If viscous strain weakening is to be used, strain weakening should also be set to true. "));
+              AssertThrow(this->introspection().compositional_name_exists("viscous_strain") || this->introspection().compositional_name_exists("strain") ||
+                          this->introspection().compositional_name_exists("eii"),
+                          ExcMessage("Material model visco_plastic with viscous strain weakening only works if there is a "
+                                     "compositional field called viscous_strain or strain."));
+          }
```
naliboff: Nice! This commit came through just as I made a comment on the TODO in `visco_plastic.h`. As the `composition_mask` function make assumptions about the first 1 (total strain) or 2 compositional fields (plastic and viscous strain), should we also enforce that they are list in the correct order here? Alternatively, we could use the names to correctly mark them within `composition_mask`.  

If we do not modify  the current structure`composition_mask`, then I think we should enforce the listed order of the strain fields in the `AssertThrow`.

anne-glerum:
@naliboff I now enforce specific names for the strain field(s), so we don't need a specific order of the fields anymore, except for the order of the strain tensor components. Masks are also based on the names. Could you please check the logic?

naliboff:
@anne-glerum: The logic looks good after the first reading, thanks for doing all those changes! I think the last step is to add tests for all the different types of weakening. 

Since you've done all of the leg work here, I can make those series of tests tomorrow morning and then push them to your branch. At least I think this should work. After that, perhaps finally dove with this material model for a good while!

anne-glerum:
@naliboff I've changed the tests so that all combinations should be covered. Just need the updated results :)

naliboff:
@anne-glerum - thanks for updating all of the tests as well. I just looked through the failing tests and here are the issues:





visco_plastic_yield_plastic_strain_weakening.prm:
1. `set Number of cheap Stokes solver steps    = 0` needs to be moved to:
     subsection Solver parameters
      





















bsection Stokes solver parameters
         set Number of cheap Stokes solver steps = 0
       end
    end

naliboff:
@anne-glerum - FYI, I update the comment above a number of times.

naliboff:
@anne-glerum - I took a look at the failing tests and the issue (at least for one case) is a "divide-by-zero" in `average_value` for the harmonic averaging case. 

This stems from line 430:
`out.viscosities[i] = average_value(composition, composition_viscosities, viscosity_averaging);`
Here, `composition` should be `volume_fractions`.

I'm going to recompile with this change and see if the tests pass on my side.

naliboff:
@anne-glerum - The patch above did fix the issue for `tests/compositional_heating.prm` (and likely a few others). That switch from `volume_fractions` to `composition` must have introduced in the rebase?

I'm going through the other failing tests now as well.

naliboff:
In `visco_plastic_yield_plastic_viscous_strain_weakening.prm` `set Number of fields = 1` needs to be changed to 2.

anne-glerum:
@naliboff oh no I didn't see your comments, but I did address them. Hopefully now all is fine!

anne-glerum:
@naliboff ok, two tests still fail. It happens in the postprocessing in the second timestep, which is when strain has been added. Unfortunately, I can't reproduce it when I just run the test prm in debug mode. I can't run aspect gdb on the cluster on the login node, could you please have a look?

naliboff:
@anne-glerum - sure thing, taking a look now.

naliboff:
@anne-glerum - The issue was actually in calculating the strain rate invariant when the strain rate was not available. I guess the strain rate is not available for some post-processing steps when the material model is called?

Regardless, the issue is resolved by modifying line 563:
`if  (use_strain_weakening == true && use_finite_strain_tensor == false && this->get_timestep_number() > 0 `

to include `in.strain_rate.size()`:
`if  (use_strain_weakening == true && use_finite_strain_tensor == false && this->get_timestep_number() > 0 && in.strain_rate.size()`

Even if the tests pass with this change, perhaps we should have another careful look over tomorrow before merging?

anne-glerum:
Thanks @naliboff! Yes, the strain rate is set to zero when viscosity is not needed in a postprocessor. Again something that I missed in the rebase to master; let's indeed check one more time tomorrow (so you're back then?)

naliboff:
@anne-glerum - yep, I'll be back tomorrow morning. On to new things after this 👍.

naliboff:
Remaining test fixes:

1. Change the compositional field name `eii` to `total_strain` in `tests/visco_plastic_yield_strain_weakening/screen-output `.

2. The solution file `tests/visco_plastic_yield_strain_weakening/solution/solution-00000.0000.gnuplot` needs to be updated.

3. Repeat of issues 1 and 2 for `tests/visco_plastic_yield_strain_weakening_compositions.prm`

naliboff:
@anne-glerum - I had another look through the source and header files and did not see anything amiss. Once the tester is happy, feel free to merge as soon as you are ready.

naliboff:
@anne-glerum - It looks like two of the tests are still failing. Some of the test changes may have been reverted during the rebase? Sorry, my eagerness to try the fancy resolve conflicts features had ended up adding more work.

tjhei:
@naliboff jenkins fails because we need a newer Jenkinsfile which you could get by rebasing to the newest master.

anne-glerum:
@tjhei Something else is going wrong now after rebase to master...

bangerth:
Yes, the Jenkins server died apparently. We can ignore this. Let's just see whether the other tester works.

naliboff:
@bangerth - All of the tests pass for 8.5. Is it ok do go ahead and merge in this case? The last few iterations have just been dealing with the tests, so all of this has had a fairly careful look.

bangerth:
@gassmoeller already made the right call.

PR #2319, opened 2018-06-19
Add bib entry for dynamic core reference Nimmo 2004

### LINKED ISSUE #2238: Add reference to Nimmo et al. (2014) to the 'dynamic core' boundary temperature model
The `source/boundary_temperature/dynamic_core.cc` file repeatedly refers to a paper by "Nimmo et al. (2004)", but this paper is not referenced in the manual and not in the bibtex file either.

jonpkay:
Fixes #2238

### DIFF
--- a/source/boundary_temperature/dynamic_core.cc
+++ b/source/boundary_temperature/dynamic_core.cc
@@ -132,11 +132,11 @@ namespace aspect
           prm.declare_entry ("K0", "4.111e11",
                              Patterns::Double (0),
                              "Core compressibility at zero pressure. "
-                             "See Nimmo et al. (2004) for more details.");
+                             "See \\cite{NPB+04} for more details.");
           prm.declare_entry ("Rho0", "7.019e3",
                              Patterns::Double (0),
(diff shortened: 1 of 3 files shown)


source/boundary_temperature/dynamic_core.cc
```diff
@@ -160,16 +160,16 @@ namespace aspect
           {
             prm.declare_entry ("Tm0","1695",
                                Patterns::Double (0),
-                               "Melting curve (Nimmo et al. [2004] eq. (40)) parameter Tm0. Units: $K$.");
+                               "Melting curve (\cite{NPB+04} eq. (40)) parameter Tm0. Units: $K$.");
```
gassmoeller: You used the double backslash `\\cite` for the description above, but not in this case and the three occurences below. Please add it in those places as well. Just to explain: Everything in the double quotation marks is parsed through the deal.II parameter handler, and thus needs an extra escape character. That is why it is fine to write `\cite` in the documentation of the header file, but you need to write `\\cite` for every parameter description in the .cc file.

PR #2321, opened 2018-06-19
Added compaction length visualization

joeschools:


tjhei:
very nice. Can you add a changelog entry?  /run-tests

source/postprocess/visualization/melt.cc
```diff
@@ -166,6 +166,12 @@ namespace aspect
                   {
                     computed_quantities[q][output_index] = this->get_melt_handler().is_melt_cell(in.current_cell)? 1.0 : 0.0;
                   }
+                else if (property_names[i] == "compaction length")
+                  {
+                    const double compaction_length = std::sqrt((out.viscosities[q] + 4./3. * melt_outputs->compaction_viscosities[q])
+                                                               * melt_outputs->permeabilities[q] / melt_outputs->fluid_viscosities[q]);
```
bangerth: I can't say that this is the correct formula -- I guess we need someone with melt experience to double check :-)

bangerth:
Oh, the jenkins test killed itself. Can someone re-run it?

source/postprocess/visualization/melt.cc
jdannberg: Yes, I believe the formula is correct. But for the test, you should use the gnuplot graphical output instead of the vtk output so that the file is human readable.

tests/compaction_length_visualization.prm
```diff
@@ -0,0 +1,136 @@
+# Test the compaction length visualization plug in.
```
jdannberg: And this should be plugin (without a space).

gassmoeller:
All comments seem to be addressed. Ready to merge.

PR #2322, opened 2018-06-19
Jenkins: fix useless "gcc" in command

tjhei:


bangerth:
It looks alright to me. How did this even work before? I would have thought that `cmake` interprets the stray `gcc` as a source path. Does it just take the last path name for the source path?

tjhei:
yes, probably.

bangerth:
OK to merge. I don't know whether this will actually affect the tester, but I'll take a chance on it.

PR #2323, opened 2018-06-19
Jenkins werror

tjhei:


tjhei:
currently based on #2322, so I will need to rebase...

PR #2324, opened 2018-06-19
Fix reference to solver tolerance parameters in manual

bartniday:
changed a bit of the manual so that it correctly locates parameters in the "solver parameters" subsection


naliboff:
Nice! Ready to merge once the tester is finished.

bangerth:
The tester doesn't check the manual compilation, so I'll just go ahead with this in the interest of keeping the tester emptier.

PR #2325, opened 2018-06-19
allow compositional fields to be advected with the melt velocity

jdannberg:


source/simulator/parameters.cc
```diff
@@ -884,7 +884,7 @@ namespace aspect
                          Patterns::List(Patterns::Anything()),
                          "A user-defined name for each of the compositional fields requested.");
       prm.declare_entry ("Compositional field methods", "",
-                         Patterns::List (Patterns::Selection("field|particles|static")),
+                         Patterns::List (Patterns::Selection("field|particles|static|melt field")),
```
tjhei: is there a reason the enum is called differently?

source/simulator/parameters.cc
```diff
@@ -913,6 +913,16 @@ namespace aspect
                          "this way, then it does not evolve at all. Its values are "
                          "simply set to the initial conditions, and will then "
                          "never change."
+                         "\n"
+                         "\\item ``melt field'': If a compositional field is marked with this "
```
tjhei: it also includes diffusion, right?

tjhei:
Looks good. Can you fix the test output?

source/simulator/parameters.cc
jdannberg: What do you mean? The reason why the enum has a different name than the input parameter? I have no idea, but I think AdvectionFieldMethod makes a lot more sense as a name, and we can not change the name of the input parameter without breaking backwards compatibility.

source/simulator/parameters.cc
jdannberg: No, I don't think so. For compositional fields the diffusivity is always zero.

source/simulator/parameters.cc
tjhei: no, I mean why is the parameter "melt field" while the enum is "fem melt field"

source/simulator/parameters.cc
tjhei: oh. :-)

source/simulator/parameters.cc
jdannberg: Oh, I just copied the existing structure. For the normal compositional field, the parameter is "field", and the enum is "fem_field".

jdannberg:
I think I addressed all comments, and I updated the test output, but I have no idea what is going on with the jenkins tester...

gassmoeller:
The jenkins tester is sick at the moment, lets ignore it for now. The rest looks good.

jdannberg:
It's sick? I hope it gets well soon :)

PR #2326, opened 2018-06-19
jenkins: more helpful astyle error

tjhei:


Jenkinsfile
```diff
@@ -51,7 +51,10 @@ pipeline {
         sh './doc/indent'
         sh 'git diff | tee changes-astyle.diff'
         archiveArtifacts artifacts: 'changes-astyle.diff', fingerprint: true
-        sh 'git diff --exit-code --name-only'
+        sh '''
+	  git diff --exit-code --name-only || \
+	  { echo "Please check indentation, see artifacts in the top right corner!"; exit 1; }
```
bangerth: I had just seen this as well when looking at a PR. Why not just run `git diff` again here to show the actual diffs? I know you already do this above and output the result, but it's awkward to find because it shows up in a green tab. The failure output should show up in a red tab like it does here, so if you ran `git diff` instead of/in addition to the `echo` here, you can remove lines 52/53

tjhei: 52/53 has the advantage that you can download the .diff file directly and apply the diff.

tjhei: @gassmoeller what do you think. Should we just show the whole diff here?

bangerth: You could do what you do in lines 52/53 only in here where you know that things have failed.

tjhei: I can not conditionally archive the diff, but I think this does what you want now.

Jenkinsfile
```diff
@@ -49,9 +49,12 @@ pipeline {
     stage('Check indentation') {
       steps {
         sh './doc/indent'
-        sh 'git diff | tee changes-astyle.diff'
+        sh 'git diff > changes-astyle.diff'
         archiveArtifacts artifacts: 'changes-astyle.diff', fingerprint: true
-        sh 'git diff --exit-code --name-only'
+        sh '''
+	  git diff --exit-code || \
+	  { echo "Please check indentation, see artifacts in the top right corner!"; exit 1; }
```
bangerth: Yes, this now looks good in the jenkins test: https://jenkins.tjhei.info/blue/organizations/jenkins/aspect/detail/PR-2326/4/pipeline

Please change the "see artifacts..." text and remove the test error in the .cc file again.

tjhei: Do you want this text removed or do you want me to rephrase it?

bangerth: Your choice, I just don't want it to be factually wrong :-)

gassmoeller: But the text is correct, we archive the artifacts (the changes diff), and they are available in the top right corner under `Artifacts` for download.

bangerth: Ah, I see. Thanks.

PR #2329, opened 2018-06-19
Avoid a deprecation warning for ConstraintMatrix.

### LINKED ISSUE #2328: ConstraintMatrix is deprecated in deal.II 9.1
Consequently, we get a warning for every source file that uses it.

Working on a work-around.

bangerth:
Fixes #2328.

### DIFF
--- a/include/aspect/global.h
+++ b/include/aspect/global.h
@@ -36,6 +36,9 @@ DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
 #  include <deal.II/lac/trilinos_precondition.h>
 #endif
 
+#if DEAL_II_VERSION_GTE(9,1,0)
+#  include <deal.II/lac/affine_constraints.h>
+#endif
 #include <deal.II/lac/generic_linear_algebra.h>
 
 #include <boost/archive/binary_oarchive.hpp>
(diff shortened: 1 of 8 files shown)


tjhei:
wouldn't it make sense to replace ConstraintMatrix by AffineConstraints and create a ``using`` for deal.II <9.1.0?

tjhei:
on second thought: it will be a while until we switch to 9.1.0, so maybe it is okay to leave as is for now?

bangerth:
I thought about this, but what you suggest would mean to have the reverse typedef:
```
  template <typename Number>
  using AffineConstraints = ConstraintMatrix;
```
and then using `AffineConstraints<double>` everywhere. That could be done, but it's C++11 syntax and we can't use that yet in ASPECT. The alternative would have been to do
```
  template <typename Number>
  class AffineConstraints : public ConstraintMatrix
  {
     ...
  };
```
but that would require me to pass through all of the constructors of `ConstraintMatrix` class and seemed too cumbersome.

So I think this is the simplest solution.

tjhei:
do you want to do anything about:
```
include/aspect/simulator_signals.h:#include <deal.II/lac/constraint_matrix.h>
```

bangerth:
I've made the change to now also only `#include` the deprecated headers if we use an old deal.II version.

If this PR finds your favor, then I'll write a github issue that reminds us of what we need to do after we *require* deal.II 9.1 when we don't need all of this any more.

PR #2332, opened 2018-06-20
reinitialise solver solution vector before expensive solve

cedrict:


cedrict:
When running annulus problems, I observed that keeping the 200 cheap iterations in made the convergence worse. This is to see whether using the same guess vector solution for the expensive as for the cheap preconditioner helps.

cedrict:
so this breaks 32 tests, and looking in detail, the number of iterations is always lower when using the solution after the 200 cheap iterations than with the proposed fix. Closing this pull request.

bangerth:
This doesn't surprise me. In essence, you're throwing away the progress the cheap solver has made already when you switch to the expensive solver.

gassmoeller:
The reason @cedrict tested this was that we found at least one model where the solution after the cheap solver phase was worse than before, i.e. we needed *more* expensive iterations after the initial cheap iterations than if we only did expensive iterations. It does seem to be an uncommon case though.

bangerth:
On 06/21/2018 12:18 PM, Rene Gassmoeller wrote:
> The reason @cedrict <https://github.com/cedrict> tested this was that we found 
> at least one model where the solution after the cheap solver phase was worse 
> than before, i.e. we needed /more/ expensive iterations after the initial 
> cheap iterations than if we only did expensive iterations. It does seem to be 
> an uncommon case though.

Yes, I understood the motivation. It doesn't surprise me that there are 
exceptions, but I'm also not surprised that in the general, this is worse.

PR #2335, opened 2018-06-20
Fix grammar error in .prm files.

bangerth:


tjhei:
@gassmoeller do you understand why the jenkins tester failed?

gassmoeller:
I have no idea. It looks like the file `FAILED` was around in the build folder, although all tests passed. Maybe an earlier build was not properly cleaned up? But this PR had only a single test run.

tjhei:
seems like a connection timeout... Weird.

PR #2336, opened 2018-06-20
Add calculation of rms velocity over the area of each boundary. These…

pmbremner:
Add calculation of rms velocity over the area of each boundary. These values are written to both the screen output and new columns in the statistics output file.


source/postprocess/velocity_boundary_statistics.cc
```diff
@@ -42,12 +42,15 @@ namespace aspect
...
                                         quadrature_formula,
                                         update_values |
+                                        update_JxW_values |
                                         update_q_points);
 
       std::vector<Tensor<1,dim> > velocities (fe_face_values.n_quadrature_points);
 
       std::map<types::boundary_id, double> local_max_vel;
       std::map<types::boundary_id, double> local_min_vel;
+      std::map<types::boundary_id, double> local_boundary_sqvel;
```
gassmoeller: this name is not quite clear to me, what about `local_velocity_square_integral`? That name is used in the `velocity_statistics` postprocessor. I would say that `boundary` is implied by the name of the class

source/postprocess/velocity_boundary_statistics.cc
```diff
@@ -42,12 +42,15 @@ namespace aspect
...
+      std::map<types::boundary_id, double> local_boundary_area;
```
gassmoeller: this name is fine

source/postprocess/velocity_boundary_statistics.cc
```diff
@@ -76,48 +80,78 @@ namespace aspect
                 // extract velocities
                 fe_face_values[this->introspection().extractors.velocities].get_function_values (this->get_solution(),
                     velocities);
-                // determine the max and min velocity on the face
+                // determine the max, min, and squared velocity on the face
+                // also determine the face area
                 double local_max = -std::numeric_limits<double>::max();
                 double local_min = std::numeric_limits<double>::max();
+                double local_sqvel = 0.0;
+                double local_fe_face_area= 0.0;
```
gassmoeller: there is a space missing in this line

source/postprocess/velocity_boundary_statistics.cc
```diff
@@ -76,48 +80,78 @@ namespace aspect
...
 
+        std::vector<double> boundary_vel_rms_values (boundary_sqvel_values.size());
+        std::vector<double> boundary_area (boundary_area_values.size());
+        for (std::set<types::boundary_id>::const_iterator
+             p = boundary_indicators.begin();
+             p != boundary_indicators.end(); ++p)
+          {
+            boundary_vel_rms_values[*p]
+              = Utilities::MPI::sum (boundary_sqvel_values[*p], this->get_mpi_communicator());
+            boundary_area[*p]
```
gassmoeller: I think you can move the calls to `MPI::sum` out of the loop and remove the index `[*p]`. This function should be able to sum the entries of this vector entry-by-entry.

source/postprocess/velocity_boundary_statistics.cc
```diff
@@ -76,48 +80,78 @@ namespace aspect
...
                 local_min_vel[boundary_indicator] = std::min(local_min,
                                                              local_min_vel[boundary_indicator]);
+                local_boundary_sqvel[boundary_indicator] += local_sqvel;
+                local_boundary_area[boundary_indicator] += local_fe_face_area;
               }
 
       // now communicate to get the global values
       std::map<types::boundary_id, double> global_max_vel;
       std::map<types::boundary_id, double> global_min_vel;
+      std::map<types::boundary_id, double> boundary_vel_rms;
```
gassmoeller: this can be `global_rms_vel`, because that is what we actually put inside in the end.

source/postprocess/velocity_boundary_statistics.cc
```diff
@@ -76,48 +80,78 @@ namespace aspect
...
       {
         // first collect local values in the same order in which they are listed
         // in the set of boundary indicators
         std::vector<double> local_max_values;
         std::vector<double> local_min_values;
+        std::vector<double> boundary_sqvel_values;
+        std::vector<double> boundary_area_values;
```
gassmoeller: I would call these `local_squared_velocity_values` and `local_boundary_area_values`

source/postprocess/velocity_boundary_statistics.cc
gassmoeller: and I would also add a  `global_boundary area` for reasons outlined below.

source/postprocess/velocity_boundary_statistics.cc
```diff
@@ -76,48 +80,78 @@ namespace aspect
...
+            boundary_area_values.push_back (local_boundary_area[*p]);
           }
         // then collect contributions from all processors
         std::vector<double> global_max_values (local_max_values.size());
         Utilities::MPI::max (local_max_values, this->get_mpi_communicator(), global_max_values);
         std::vector<double> global_min_values (local_min_values.size());
         Utilities::MPI::min (local_min_values, this->get_mpi_communicator(), global_min_values);
 
+        std::vector<double> boundary_vel_rms_values (boundary_sqvel_values.size());
+        std::vector<double> boundary_area (boundary_area_values.size());
```
gassmoeller: then you can call this `global_rms_velocity_values` and `global_boundary_area_values`

source/postprocess/velocity_boundary_statistics.cc
gassmoeller: Also with the name changes this line would then read:
```
Utilities::MPI::sum(local_squared_velocity_values,this->get_mpi_communicator(),global_rms_velocity_values)
```
and equivalent for the boundary area. I would find this more intuitive to read.

source/postprocess/velocity_boundary_statistics.cc
pmbremner: Changed boundary_sqvel_values to local_velocity_square_integral_values to be more consistent with the map variable name.

source/postprocess/velocity_boundary_statistics.cc
pmbremner: Done

source/postprocess/velocity_boundary_statistics.cc
pmbremner: Done

source/postprocess/velocity_boundary_statistics.cc
pmbremner: Done

source/postprocess/velocity_boundary_statistics.cc
pmbremner: Done

source/postprocess/velocity_boundary_statistics.cc
pmbremner: Moved these out of the loop with the name changes, but had to leave the loop in place for calculating the rms at each boundary.

source/postprocess/velocity_boundary_statistics.cc
pmbremner: Changed boundary_vel_rms_values to global_velocity_square_integral_values. I removed rms from the variable name because the rms values are now calculated in the step that also put values back into maps (see above comment). The other variable is changed as suggested.

source/postprocess/velocity_boundary_statistics.cc
pmbremner: Changed again. Now the rms calculation is moved to the loop where the values are put into a map. That allowed me to remove the loop referenced in the code snippet above.

source/postprocess/velocity_boundary_statistics.cc
pmbremner: No longer needed global_boundary_area. Removed

gassmoeller:
You need to update the test output that changed, but except for that this seems ready.

PR #2337, opened 2018-06-20
New plugin system for heat flux boundary conditions

gassmoeller:
This is a proposal for a new plugin system for heat flux boundary conditions for the temperature equation. I need to look over this one more time to look for copy-paste errors, and so far the boundary condition is not actually used. This is just the structure to allow @jdannberg and @jonpkay to continue working on it.


jdannberg:
This would be the syntax for the input file:

```
subsection Boundary heat flux model
  set Fixed heat flux boundary indicators = bottom, top
  set Model name = function

  subsection Function
    set Function expression = if(y>0.5,1,-1)
  end
end
```

bangerth:
On 06/20/2018 05:16 PM, Juliane Dannberg wrote:
> |set Fixed heat flux boundary indicators = bottom, top|


Shouldn't this be in the same section where we declare the kinds of boundary 
conditions, i.e., in section "Boundary temperature model"?

jdannberg:
We also have `Boundary velocity model` and `Boundary traction model`, so I think it makes sense to have the same distinction for the energy equation.

bangerth:
We should definitely have the new section. My question is simply where exactly you declare which boundary has what type of boundary condition. Where do you currently say which boundaries are Dirichlet and which are Neumann?

gassmoeller:
When we removed the Model settings subsection we moved the specification of the boundary indicators into the separate subsection. There is no central place for the boundary conditions any more, if you prescribe dirichlet boundary conditions for the velocity you do it in `Boundary velocity model`, if you prescribe Neumann boundary conditions you do it in `Boundary traction model`. The same structure should hold for the temperature.

jdannberg:
Note to self: we have to add an Assert that makes sure we don't use the same boundary indicator for both temperature and heat flux boundaries.

jdannberg:
This should be ready for review now.

source/boundary_heat_flux/function.cc
```diff
@@ -0,0 +1,178 @@
...
+              const double depth = this->get_geometry_model().depth(position);
+              Point<dim> point;
+              point(0) = depth;
+
+              heat_flux[i] *= boundary_heat_flux_function.value(point);
+            }
+          else
+            {
+              AssertThrow(false, ExcNotImplemented());
+              return std::vector<Tensor<1,dim> > (n_evaluation_points,numbers::signaling_nan<Tensor<1,dim> >());
```
tjhei: is this line needed to compile? I would guess returning heat_flux below is enough.

source/boundary_heat_flux/function.cc
```diff
@@ -0,0 +1,178 @@
...
+
+      return heat_flux;
+    }
+
+
+    template <int dim>
+    void
+    Function<dim>::update()
+    {
+      // we get time passed as seconds (always) but may want
```
tjhei: not "passed"?!

source/boundary_heat_flux/function.cc
```diff
@@ -0,0 +1,178 @@
...
+// explicit instantiations
+namespace aspect
+{
+  namespace BoundaryHeatFlux
+  {
+    ASPECT_REGISTER_BOUNDARY_HEAT_FLUX_MODEL(Function,
+                                             "function",
+                                             "Implementation of a model in which the boundary "
+                                             "heat flux is given in terms of an explicit formula "
+                                             "that is elaborated in the parameters in section "
```
tjhei: elaborated -> given

source/simulator/parameters.cc
```diff
@@ -1204,7 +1234,7 @@ namespace aspect
         Utilities::split_string_list(prm.get("Remove nullspace"));
       AssertThrow(Utilities::has_unique_entries(nullspace_names),
                   ExcMessage("The list of strings for the parameter "
-                             "'Model settings/Remove nullspace' contains entries more than once. "
+                             "'Nullspace removal/Remove nullspace' contains entries more than once. "
```
tjhei: nice find!

source/simulator/parameters.cc
```diff
@@ -1668,6 +1698,27 @@ namespace aspect
...
+            = geometry_model.translate_symbolic_boundary_names_to_ids(Utilities::split_string_list
+                                                                      (prm.get ("Fixed heat flux boundary indicators")));
+          fixed_heat_flux_boundary_indicators
+            = std::set<types::boundary_id> (x_fixed_heat_flux_boundary_indicators.begin(),
+                                            x_fixed_heat_flux_boundary_indicators.end());
+        }
+      catch (const std::string &error)
+        {
+          AssertThrow (false, ExcMessage ("While parsing the entry <Boundary heat flux model/Fixed heat flux "
+                                          "boundary indicators>, there was an error. Specifically, "
```
tjhei: I don't like "there was an error": can you try to rephrase?

jdannberg: That's how it's done for all the other boundary indicators (Free surface, traction, etc.). Do you want me to change all of them? Maybe we just leave it as is?

source/boundary_heat_flux/function.cc
jdannberg: What do you mean, not "passed"? Again, this is just copied from all of the other plugins. Should I just make a separate pull request to change it for all of them? And to what?

source/boundary_heat_flux/function.cc
jdannberg: Same here. Change it everywhere?

source/simulator/parameters.cc
tjhei: then leave as is.

source/boundary_heat_flux/function.cc
tjhei: there is no argument that is "passed". Anyways, we can fix that later.

bangerth:
On 06/21/2018 12:13 PM, Rene Gassmoeller wrote:
> When we removed the Model settings subsection we moved the specification of 
> the boundary indicators into the separate subsection. There is no central 
> place for the boundary conditions any more, if you prescribe dirichlet 
> boundary conditions for the velocity you do it in |Boundary velocity model|, 
> if you prescribe Neumann boundary conditions you do it in |Boundary traction 
> model|. The same structure should hold for the temperature.

Hm, I did not pay attention to this at the time. I think it would have been 
nicer if we had had a central place where you say what boundaries have what 
kinds of boundary conditions, and then separate sections that describe the 
details. But that's too late now, so let's not worry about it here.

Issue #2331, opened 2018-06-19
Chunk Geometry Model - Error message for very small boundaries


SophieCoulson:
When running a chunk model for a region extending to 90 (or -90) degrees latitude the north (or south) boundary is a point, therefore the model will not run. The current error message for this is:

An error occurred in line <6072> of file </n/home08/slcoulson/newlib/dealii-8.5.0/include/deal.II
/numerics/vector_tools.templates.h> in function
    void dealii::VectorTools::compute_nonzero_normal_flux_constraints(const DoFHandlerType<dim, s
pacedim> &, unsigned int, const std::set<unsigned char, std::less<unsigned char>, std::allocator<
unsigned char>> &, dealii::FunctionMap<spacedim, double>::type &, dealii::ConstraintMatrix &, con
st dealii::Mapping<dim, spacedim> &) [with int dim = 3, int spacedim = 3, DoFHandlerType = dealii
::DoFHandler]
The violated condition was: 
    std::fabs(normal_vector.norm() - 1) < 1e-14

A more direct error message would be more helpful.

SophieCoulson:
I'm working on this now

bangerth:
Yes, nice idea. This should be easy enough to catch and give a good error message!

gassmoeller:
Was this fixed by #2339? If so: @SophieCoulson can you close this issue?

SophieCoulson:
Yes this is fixed!

PR #2339, opened 2018-06-20
Added error messages for boundaries that do not work in chunk geometry model

SophieCoulson:
Addressing issue #2331 - added error messages when lat =90/-90

### DIFF
--- a/source/geometry_model/chunk.cc
+++ b/source/geometry_model/chunk.cc
@@ -728,6 +728,10 @@ namespace aspect
 
               AssertThrow (point1[2] < point2[2],
                            ExcMessage ("Minimum latitude must be less than maximum latitude."));
+              AssertThrow (point1[2] > -0.5*numbers::PI,
+                           ExcMessage ("Minimum latitude needs to be larger than -90 degrees."));
+              AssertThrow (point2[2] < 0.5*numbers::PI,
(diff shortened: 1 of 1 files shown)


source/geometry_model/chunk.cc
```diff
@@ -728,6 +728,10 @@ namespace aspect
 
               AssertThrow (point1[2] < point2[2],
                            ExcMessage ("Minimum latitude must be less than maximum latitude."));
+              AssertThrow (point1[2] > -0.5*numbers::PI,
+                           ExcMessage ("Minimum latitude cannot be exactly -90 degrees"));
```
tjhei: maybe "... needs to be larger than -90 degrees" would be a better error message? Can you add a period at the end?

MFraters:
Thanks for adding this, this makes it a lot more friendly :)
/run-tests

jaustermann:
The wording has been updated. This looks good now if tester is happy.

tjhei:
@SophieCoulson it looks like your git settings on your machine do not have the same email address as your github account (see the commit looks gray and doesn't have your profile picture?). Would you mind changing this option and editing your commit? I can show you how to do that. As a result this will be correctly credited to you.

tjhei:
excellent. I think this is ready to be merged after the testers are happy.

PR #2341, opened 2018-06-20
Replace the word 'abort' by 'terminate' if we exit with a zero error code

bangerth:
`abort` has the connotation of being an error, but we just do `exit(0)`, which is totally acceptable way to say that we just end the program.


bangerth:
@tjhei approved this. Anyone willing to merge? (The Jenkins tester is currently broken, but the other three testers are ok with this patch.)

PR #2342, opened 2018-06-20
Add a test for the visco_plastic composite viscosity.

naliboff:
Something to add after #2318 is finished.


naliboff:
@anne-glerum - yes, I think that's a good idea. In fact, there is actually no test for the diffusion creep functionality. 

I'll add three more tests (diffusion, dislocation) with no yielding and constant P-T. I think the constant P could be achieved with gravity off and a specified surface pressure.

Does this sound reasonable?

anne-glerum:
Yes, I think it would be good to have a test for each case.

tests/visco_plastic_diffusion.prm
```diff
@@ -0,0 +1,105 @@
+# Global parameters
+set Dimension                              = 2
+set Start time                             = 0
+set End time                               = 0
+set Use years in output instead of seconds = true
+set Nonlinear solver scheme                = single Advection, iterated Stokes
+set Max nonlinear iterations               = 1
+set Output directory                       = visco_plastic_dislocation
```
anne-glerum: Setting the output directory is not necessary for the tests and is ignored. Maybe cleaner to remove it for all the tests?

tests/visco_plastic_composite.prm
```diff
@@ -0,0 +1,111 @@
...
+
+# Global parameters
+set Dimension                              = 2
+set Start time                             = 0
+set End time                               = 0
+set Use years in output instead of seconds = true
+set Nonlinear solver scheme                = single Advection, iterated Stokes
+set Max nonlinear iterations               = 1
+set Output directory                       = visco_plastic_composite
+set Timing output frequency                = 1
```
anne-glerum: This line is also unnecessary, 'cause we're doing only one timestep anyway.

tests/visco_plastic_composite.prm
```diff
@@ -0,0 +1,111 @@
...
+subsection Boundary velocity model
+  set Prescribed velocity boundary indicators = bottom y: function, top y: function, left x: function, right x: function
+end
+
+# Velocity on boundaries characterized by functions
+subsection Boundary velocity model
+  subsection Function
+    set Variable names      = x,y
+    set Function constants  = m=0.0005, year=1
+    set Function expression = if (x<50e3 , -1*m/year, 1*m/year); if (y<50e3 , 1*m/year, -1*m/year);
```
anne-glerum: This is probably too nit-picky, but I interpret the `m` and `year` function constants as scale factors for going to meters and years units. Right now that's not how `m` is being used. 
`0.5*mm/yr` with `mm=0.001` would be more clear to me.

tests/visco_plastic_diffusion.prm
```diff
@@ -0,0 +1,105 @@
...
+    set Activation energies for diffusion creep = 375.e3
+    set Activation volumes for diffusion creep = 10.e-6
+  end
+end
+
+# Gravity model
+subsection Gravity model
+  set Model name = vertical
+  subsection Vertical
+    set Magnitude = 0.0
```
anne-glerum: With the gravity magnitude set to zero and a constant temperature, we know what the viscosity should be here. Can you the viscosity value as a comment at the top just as you did for the composite test with zero parameter values?

tests/visco_plastic_diffusion.prm
naliboff: @anne-glerum - Agree. Should I just go ahead and remove it for all of the visco_plastic tests or just these three?

tests/visco_plastic_diffusion.prm
naliboff: 👍

naliboff:
@anne-glerum - If the changes look ok, this should be good to go.

Issue #2300, opened 2018-06-16
clarify "Use years in output"


tjhei:
Add a section to the manual (longer than the note in 2.3) about this setting. In particular that this influences the units for velocity boundary conditions and output of velocities but not much else.

Are we fully consistent?

Also, the part "in output" is not ideal...

(this came up at the tutorial)

hfmarkp:
I can work on this

bangerth:
Fixed by @hfmarkp in #2345.

PR #2345, opened 2018-06-20
Re-word manual sections on "Use years in output instead of seconds" flag

hfmarkp:
Addressing issue #2300 - I edited the note in Section 2.3 and added a short subsection 2.3.1 to clarify that the flag pertains to time and velocity units for input parameters as well as outputs.


tjhei:
fixes #2300

PR #2346, opened 2018-06-20
Break the computation of the reference viscosity into its own function.

bangerth:
Also make sure that it is initialized to an invalid value.


bangerth:
In its current form, this should be a no-op. The next steps are:
* Let the function be called at the beginning of the time step, before we assemble anything.
* Actually compute things differently than what we had before. Right now, I'm just moving the expression to a different place.

The comment for the new function is currently not accurate, but it will be once the two steps above are done.

source/simulator/core.cc
```diff
@@ -621,9 +616,20 @@ namespace aspect
...
          p != boundary_traction.end(); ++p)
       p->second->update ();
+
+    // Re-compute the pressure scaling factor. In some sense, it would be nice
+    // if we did this not just once per time step, but once for each solve --
+    // i.e., multiple times per time step if we iterate out the nonlinearity
+    // during during a Newton or Picard iteration. But that's more work,
+    // and the function would need to be called in more different places.
+    // Unless we have evidence that that's necessary, let's assume that
+    // the reference viscosity does not change tooo much between nonlinear
```
tjhei: "too"

include/aspect/simulator.h
```diff
@@ -404,6 +404,19 @@ namespace aspect
        */
       void compute_current_constraints ();
 
+      /**
+       * Compute the reference viscosity for the current time by
```
tjhei: this comments confused me for a second. Maybe start with "This function updates the pressure_scaling variable ..."

source/simulator/core.cc
```diff
@@ -621,9 +616,20 @@ namespace aspect
          p = boundary_traction.begin();
          p != boundary_traction.end(); ++p)
       p->second->update ();
+
+    // Re-compute the pressure scaling factor. In some sense, it would be nice
+    // if we did this not just once per time step, but once for each solve --
+    // i.e., multiple times per time step if we iterate out the nonlinearity
+    // during during a Newton or Picard iteration. But that's more work,
```
gassmoeller: one during too much

source/simulator/core.cc
gassmoeller: tooo -> too

source/simulator/helper_functions.cc
```diff
@@ -315,6 +315,20 @@ namespace aspect
...
 
+  template <int dim>
+  void
+  Simulator<dim>::
+  compute_pressure_scaling_factor()
+  {
+    // determine how to treat the pressure. we have to scale it for the solver
+    // to make velocities and pressures of roughly the same (numerical) size,
+    // and we may have to fix up the right hand side vector before solving for
+    // compressible models if there are no in-/outflow boundaries
```
gassmoeller: How does the second part of the second relate to this function?

source/simulator/helper_functions.cc
```diff
@@ -315,6 +315,20 @@ namespace aspect
 
 
 
+  template <int dim>
+  void
+  Simulator<dim>::
+  compute_pressure_scaling_factor()
+  {
+    // determine how to treat the pressure. we have to scale it for the solver
```
gassmoeller: Capitalize first words of sentences.

source/simulator/helper_functions.cc
bangerth: Not sure, actually. I copied it from the original statement, but I've taken that half sentence out now.

gassmoeller:
Wolfgang if you have a minute can you rebase this? It is essentially ready, except that the documentation is for functionality that is not yet implemented in the function, but should be. Should we merge and add the functionality, or modify the documentation?

bangerth:
I've rebased this. Let's merge this and leave the documentation "aspirational" for the moment.

gassmoeller:
That is ok with me. You will need to update the test results for the one test that you had to change for this PR.

bangerth:
Test output updated.

Issue #2266, opened 2018-05-24
Future features and improvements for viscoelastic material model


naliboff:
- [x] Use general function for computing volume fractions
- [x] Replace deprecated function 'double_contract'
- [ ] Change to averaging over compositions only after viscoelastic stresses have been calculated (e.g., stresses will be averaged rather than material properties used to calculate stresses.
- [x] Add option to track viscoelastic stresses through particles
- [x] Add a basic viscoelastic cookbook
- [ ] Determine if we can use operator splitting to enable this material model for all solver schemes
- [x] Add a comment that the elasticity can be switched off by setting the shear modulus to a very large value
- [x] Add a benchmark for brick-experiment with viscoelastic-plastic rheology (+/- Newton functionality)? 
- [ ] Implement elastic-viscous stress splitting (EVSS) method
- [x] Add a benchmark for Farrington et al. 2014 (G3) 2D shear test
- [x] Add a benchmark for Farrington et al. 2014 (G3) 3D torsion test
- [ ] Implement compressible elasticity

naliboff:
Closing this issue as all future VEP improvements are being tracked by the active project [Viscoelastoplasticity ](https://github.com/orgs/geodynamics/projects/4/views/1)

PR #2356, opened 2018-06-20
Use interface volume fraction function in the visco elastic material model

anne-glerum:
Item 1 of  #2266


anne-glerum:
@naliboff hmm all tests pass here

naliboff:
@anne-glerum - Must have been some other small change in the vep model. Thanks!

Issue #2010, opened 2017-11-21
assembler cleanup: Scratch::reinit()


tjhei:
Introduce ``Scratch::reinit(cell)`` and ``Scratch::reinit(cell,face)`` that calls ``fe_values.reinit(...)`` and (re)sets internal properties like ``cell``, ``face_no``, etc.
We can than also move other common operations into this like computing the stokes DoFs.

wanyingw:
@alarshi -- did this get completely fixed, or is there anything left to do?

alarshi:
This issue is resolved after the pull request #2357.

bangerth:
@alarshi -- thank you for checking!

PR #2357, opened 2018-06-20
Added a function, reinit, in assembler/interface.h 

alarshi:
To address the issue #2010 .

### DIFF
--- a/include/aspect/simulator/assemblers/interface.h
+++ b/include/aspect/simulator/assemblers/interface.h
@@ -107,12 +107,7 @@ namespace aspect
 
           FEValues<dim> finite_element_values;
 
-          void reinit ( const typename DoFHandler<dim>::active_cell_iterator &cell_ref)
-          {
-            this->cell = cell_ref;
-            this->face_number = numbers::invalid_unsigned_int;
-            finite_element_values.reinit (cell_ref);
-          }
+          void reinit (const typename DoFHandler<dim>::active_cell_iterator &cell_ref);
 
           std::vector<types::global_dof_index> local_dof_indices;
           std::vector<unsigned int>            dof_component_indices;
@@ -167,10 +162,10 @@ namespace aspect
 
           FEFaceValues<dim> face_finite_element_values;
 
-          void reinit (const typename DoFHandler<dim>::active_cell_iterator &cell_ref, const unsigned face_number_ref)
-          {
-            face_finite_element_values.reinit(cell_ref, face_number_ref);
-          }
+          using StokesPreconditioner<dim>::reinit;
+
+          void reinit (const typename DoFHandler<dim>::active_cell_iterator &cell_ref,
+                       const unsigned face_number_ref);
 
           std::vector<Tensor<1,dim> >          phi_u;
           std::vector<Tensor<1,dim> >          velocity_values;
@@ -228,12 +223,7 @@ namespace aspect
 
           FEValues<dim> finite_element_values;
 
-          void reinit ( const typename DoFHandler<dim>::active_cell_iterator &cell_ref)
-          {
-            this->cell = cell_ref;
-            this->face_number = numbers::invalid_unsigned_int;
-            finite_element_values.reinit (cell_ref);
-          }
+          void reinit (const typename DoFHandler<dim>::active_cell_iterator &cell_ref);
 
           std_cxx11::unique_ptr<FEFaceValues<dim> >    face_finite_element_values;
           std_cxx11::unique_ptr<FEFaceValues<dim> >    neighbor_face_finite_element_values;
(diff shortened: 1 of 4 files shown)


tjhei:
can I ask you to also do part II of this? Introduce ``reinit(cell, face)`` that replaces ``scratch.face_finite_element_values.reinit(...)`` Maybe in a separate PR?

include/aspect/simulator/assemblers/interface.h
```diff
@@ -84,8 +84,10 @@ namespace aspect
            * numbers::invalid_unsigned_int.
            */
           unsigned face_number;
+
```
tjhei: can you remove these newlines you introduced?

alarshi:
I removed the blank lines.

include/aspect/simulator/assemblers/interface.h
```diff
@@ -160,6 +167,13 @@ namespace aspect
 
           FEFaceValues<dim> face_finite_element_values;
 
+          using StokesPreconditioner<dim>::reinit;
+
+          void reinit (const typename DoFHandler<dim>::active_cell_iterator &cell_ref, const unsigned face_number_ref)
+          {
+            face_finite_element_values.reinit(cell_ref, face_number_ref);
```
tjhei: you need to set ``face_number``

include/aspect/simulator/assemblers/interface.h
```diff
@@ -160,6 +167,14 @@ namespace aspect
 
           FEFaceValues<dim> face_finite_element_values;
 
+          using StokesPreconditioner<dim>::reinit;
+
+          void reinit (const typename DoFHandler<dim>::active_cell_iterator &cell_ref, const unsigned face_number_ref)
+          {
+            this->face_number = face_number_ref;
```
tjhei: set this->cell

tjhei:
This looks good to me, but I coached her with the PR so I would appreciate if somebody else can take a look.

include/aspect/simulator/assemblers/interface.h
```diff
@@ -107,6 +107,13 @@ namespace aspect
 
           FEValues<dim> finite_element_values;
 
+          void reinit ( const typename DoFHandler<dim>::active_cell_iterator &cell_ref)
```
bangerth: Our usual style would not have a space after the opening parenthesis.

Also, we generally declare the function in the class declaration, but the body of the function (the *definition*) goes into the .cc file. See where the body of the constructor is, for example.

include/aspect/simulator/assemblers/interface.h
```diff
@@ -160,6 +167,15 @@ namespace aspect
 
           FEFaceValues<dim> face_finite_element_values;
 
+          using StokesPreconditioner<dim>::reinit;
+
+          void reinit (const typename DoFHandler<dim>::active_cell_iterator &cell_ref, const unsigned face_number_ref)
```
bangerth: same here for the body of the function.

Break the two arguments of the function onto separate lines.

include/aspect/simulator/assemblers/interface.h
```diff
@@ -216,6 +232,13 @@ namespace aspect
 
           FEValues<dim> finite_element_values;
 
+          void reinit ( const typename DoFHandler<dim>::active_cell_iterator &cell_ref)
+          {
```
bangerth: and again the same here with the body of the function

source/simulator/assemblers/stokes.cc
```diff
@@ -608,8 +608,6 @@ namespace aspect
           !=
           this->get_boundary_traction().end())
         {
-          scratch.face_finite_element_values.reinit (scratch.cell, scratch.face_number);
```
bangerth: Hm, why do we not need this any more? Because `reinit` is already called from the function that calls the current function?

source/simulator/assembly.cc
```diff
@@ -587,11 +585,10 @@ namespace aspect
       {
         // then also work on possible face terms. if necessary, initialize
         // the material model data on faces
-        for (scratch.face_number=0; scratch.face_number<GeometryInfo<dim>::faces_per_cell; ++scratch.face_number)
-          if (cell->at_boundary(scratch.face_number))
+        for (unsigned int face_number=0; face_number<GeometryInfo<dim>::faces_per_cell; ++face_number)
```
bangerth: yes, this is nice -- nice simplification!

source/simulator/assemblers/stokes.cc
tjhei: This was a bug, we used to call reinit twice (inside and outside of this function)!

alarshi:
I tried to address the styling issues.

Issue #2243, opened 2018-05-14
track the location of the rising stokes sphere


liaokp:
In my 2D simple model (a box), I have a buoyant rising stokes sphere, I want to track its location (Y-coordinate) at every time step and output it to a file. The way I tried is to put in a single passive particle at the center of the sphere (see model file line 176 - 187, complete model file and plugin is attached), but ASPECT failed to run my model. 

168 subsection Postprocess
169   set List of postprocessors = visualization,topography,velocity statistics,depth average,particles
170   subsection Visualization
171     set Time between graphical output = 1e5
172     set List of output variables = density, viscosity, shear stress, nonadiabatic pressure
173     # set Output format = gnuplot
174   end
175
176   subsection Particles
177     set Number of particles = 1
178     set Time between data output = 0.0
179     set Data output format = ascii
180     set List of particle properties = function, initial composition, initial position
181     set Particle generator name = ascii file
182
183     subsection Function
184       set Variable names      = x,y
185       set Function expression = if( sqrt( (x-1000.e3)^2 + (y-60.e3)^2 ) > 1.e3, 0, 1)
186     end
187   end
188 end

## 
error message is really long, repeat of following lines:

An error occurred in line <3520> of file </home1/04308/liaokp/progs/dealii-8.5.0/include/deal.II/dofs/dof_accessor.templates.h> in function
    void dealii::DoFCellAccessor<DoFHandlerType, lda>::get_dof_values(const InputVector&, ForwardIterator, ForwardIterator) const [with InputVector = dealii::TrilinosWrappers::MPI::Vector; ForwardIterator = double*; DoFHandlerType = dealii::DoFHandler<2>; bool level_dof_access = false]
The violated condition was: 
    this->is_artificial() == false
Additional information: 
    Can't ask for DoF indices on artificial cells.

Stacktrace:

#0  /home1/04308/liaokp/progs/deal.II-8.5/lib/libdeal_II.g.so.8.5.0: dealii::MappingQ1Eulerian<2, dealii::TrilinosWrappers::MPI::Vector, 2>::get_vertices(dealii::TriaIterator<dealii::CellAccessor<2, 2> > const&) const
[...]

PR #2358, opened 2018-06-21
Disable combining particle with free surface

gassmoeller:
The particles were never tested with the free surface and I am reasonably sure that at least something is broken in this case (see also #2243). Disable this feature for now until we further test it.


tjhei:
please update the test output (see jenkins)

PR #2359, opened 2018-06-21
Disable combining particle with free surface

gassmoeller:
Picks #2358 into the 2.0.1 release branch.

### DIFF
--- a/source/particle/world.cc
+++ b/source/particle/world.cc
@@ -826,6 +826,13 @@ namespace aspect
                              "diameter in one time step and therefore skip the layer "
                              "of ghost cells around the local subdomain."));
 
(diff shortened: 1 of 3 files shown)


anne-glerum:
Actually, another assert is triggered in the test (ExcMessage("The free surface scheme can only be used with no pressure normalization"), so set 
`set Pressure normalization                 = no`
Otherwise it looks good.

bangerth:
OK to merge after #2358.

gassmoeller:
Nice catch Anne, I totally did not check the test output :-(. Should be fixed now, same for #2358.

tjhei:
to double-check: is this up to date with the changes in #2358 ?

gassmoeller:
Yes this is exactly the same state.

Issue #2344, opened 2018-06-20
Creating an 'ascii data profile' plugin for composite initial temperature models


mnkajan:
I have been using my own initial temperature plugin that can read in different geotherms (depth and temperature values) and then add perturbations from a tomographic model (coordinates and temperature perturbation values). Since the "ascii data" initial temperature plugin already exists, and we can now create composite models, it seems unnecessary to have a totally separate model.

In fact, most of the initial temperature plugins (e.g., S40RTS perturbation, harmonic perturbation, etc.) add perturbations to a background temperature that is defined in the plugin. This is mainly limited to the adiabatic background temperature or a single constant reference temperature. I believe it would be simpler and more flexible to write a separate "ascii data profile" plugin for depth-dependent temperature input that could be combined with other plugins.

Let me know if you have any comments or suggestions to make this more useful for everyone!

gassmoeller:
Yes, and @mnkajan and I talked about this. I think this is a good idea. One alternative approach that is already possible is to prescribe a `ascii data` profile in the adiabatic conditions, and then use the `adiabatic` initial temperature plugin, but that has the disadvantage that it forces everyone to also prescribe a pressure and density profile. Having a initial temperature plugin `ascii data profile` seems reasonable and useful to me. The structure of this plugin could look very similar to `AsciiData` in adiabatic conditions.

mnkajan:
Addressed by #2360.

PR #2360, opened 2018-06-21
Add 'ascii profile' initial temperature plugin

mnkajan:
Reads in temperature values as a function of depth. Designed to provide a background temperature profile that can be combined with other plugins using the 'List of model operators' parameter. See #2344


jaustermann:
Adding a test for this would probably be good. Happy to help with that :)

gassmoeller:
The reference tester produces slightly different results than your machine. That is ok and happens quite often (different libraries, operating systems, ...). Ask one of use how to fix this (you can download a patch file when you click on 'details' next to the aspect-8.5-gcc-hd test results).

PR #2362, opened 2018-06-21
Only set time step number and timestep sizes after initialization is finished.

bangerth:
This way, the initialization code and other places have a way of determining
whether we are still in initialization, or whether we have already started time
stepping.

Right now, there is no way to find out, for example in the hook that
is called from the function that builds constraints, whether we
are already in the process of time stepping, or only in the
initialization phase. So move the initialization of the time
step number and time step variables to the very end of the
initialization phase.

### DIFF
--- a/source/material_model/viscoelastic.cc
+++ b/source/material_model/viscoelastic.cc
@@ -172,7 +172,12 @@ namespace aspect
       // On the first (0) time step the elastic time step is always equal to the value
       // specified in 'fixed_elastic_time_step', which is also used in all subsequent time
       // steps if 'use_fixed_elastic_time_step' is set to true.
-      const double dte = ( ( this->get_timestep_number() > 0 && use_fixed_elastic_time_step == false )
+      //
+      // We also use this parameter when we are still *before* the first time step,
+      // i.e., if the time step number is numbers::invalid_unsigned_int.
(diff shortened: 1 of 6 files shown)


bangerth:
I need this so that I can check, in the test that failed earlier today, whether the plugin is still in the initialization (where I can't compute a reference viscosity yet, because we don't have a temperature field), or whether I'm already in the time stepping part of the program.

tjhei:
> tests: 551 / 580 FAILED 

You broke a few tests with this change. :-) 

@ljhwang Bangerth bucket!?

bangerth:
Oh, oops :-)

bangerth:
Let's try again -- I moved three lines up by five lines :-)

source/material_model/viscoelastic.cc
```diff
@@ -144,7 +144,19 @@ namespace aspect
                                               const double average_elastic_shear_modulus,
                                               const double dte) const
     {
```
bangerth: @naliboff -- can I ask you to take a look at whether you think this change here is in this file is reasonable? I want to make a change where the time step is a signaling_NaN before we start the simulation (it used to be zero) because it really has not been computed or set at that time, and this now trips up the code here.

gassmoeller:
Looks good.

PR #2349, opened 2018-06-20
Doxygen: include bibliography

tjhei:

PR #2363, opened 2018-06-21
Use correct sentences. Also don't repeat the information from the reference.

bangerth:
Follow-up to #2349.


gassmoeller:
Looks good.

Issue #1097, opened 2016-07-02
Fix bibtex entries for ASPECT


bangerth:
There are several bibtex entries in the tree for ASPECT. Where we say things such as "Advanced Solver for Problems in Earth ConvecTion", all capital letters have to be enclosed in braces to make sure they are also capitalized after bibtex runs.

bangerth:
@egpuckett suggests that we use a style file, such as the Chicago Manual of Style.

njinju:
I will work on fixing the bibtex entries for ASPECT

bangerth:
Fixed by #2364.

PR #2364, opened 2018-06-21
Fix bibtex entries for ASPECT.

njinju:
There are several bibtex entries issues in the tree for ASPECT where some necessary capital letters such as in  "Advanced Solver for Problems in Earth ConvecTion" or in names like "Stokes" turns out to show small letters when bibtex runs. We fix this issue by enclosing all necessary capital letters into curly braces.

### DIFF
--- a/doc/manual/manual.bib
+++ b/doc/manual/manual.bib
@@ -129,8 +129,8 @@
 @ARTICLE{G09,
   Author = "Geenen, T. and ur Rehman, M. and MacLachlan, S. P. and Segal,
                   G. and Vuik, C. and van den Berg, A. P. and Spakman, W.",
-  Title = "{Scalable robust solvers for unstructured FE geodynamic modeling
-                  applications: Solving the Stokes equation for models with
+  Title = "{Scalable robust solvers for unstructured {FE} geodynamic modeling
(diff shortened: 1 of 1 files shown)


doc/manual/manual.bib
```diff
@@ -11315,7 +11315,7 @@ @misc{aspect-doi-v2.0.0
 }
 
 @article{aspectmanual,
-  title         = {{\textsc{ASPECT}: Advanced Solver for Problems in Earth's ConvecTion, User Manual}},
+  title         = {{\textsc{ASPECT}: {A}dvanced {S}olver for {P}roblems in {E}arth's {C}onvec{T}ion, User Manual}},
```
tjhei: this change is not necessary as there are already braces around the whole title.

tjhei:
related to #1097

tjhei:
@bangerth it this what you had in mind? Another option is to just put the whole title in braces.

doc/manual/manual.bib
bangerth: True, but there's also no harm here.

bangerth:
Yes, that's what I thought should be done.

Thanks, @njinju !

PR #2366, opened 2018-06-21
Improve documentation of output formats

### LINKED ISSUE #2055: Post a list of possible output formats
Post the list of supported output formats in the description of the Visualization/Output format input parameter.

gassmoeller:
Fixes #2055. I did not include the format options themselves, because that would duplicate them in the manual. Instead I just refer to the manual.

### DIFF
--- a/source/postprocess/depth_average.cc
+++ b/source/postprocess/depth_average.cc
@@ -294,9 +294,10 @@ namespace aspect
                              Patterns::Selection(DataOutBase::get_output_format_names().append("|txt")),
                              "The format in which the output shall be produced. The "
                              "format in which the output is generated also determines "
-                             "the extension of the file into which data is written. See "
-                             "the corresponding appendix of the manual for possible "
-                             "output formats.");
+                             "the extension of the file into which data is written. "
+                             "The list of possible output formats that can be given "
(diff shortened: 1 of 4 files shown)


tjhei:
this will be difficult to find. Can we add something like ``see Section \ref{parameters:Postprocess/Visualization/Output format}`` ? Or should we wait until we can link into the pdf and do this?

source/postprocess/depth_average.cc
```diff
@@ -294,7 +294,9 @@ namespace aspect
                              Patterns::Selection(DataOutBase::get_output_format_names().append("|txt")),
                              "The format in which the output shall be produced. The "
                              "format in which the output is generated also determines "
-                             "the extension of the file into which data is written.");
+                             "the extension of the file into which data is written. See "
+                             "the corresponding appendix of the manual for possible "
```
bangerth: I'm not sure it's clear here (and in the other places) what the corresponding appendix actually is. Maybe say "The list of possible output formats that can be given here is documented in the appendix of the manual where the current parameter is described."

tjhei:
astyle and jenkins tester report a bogus failure

PR #2369, opened 2018-06-21
add unit_testing framework

tjhei:
- include catch cpp testing framework (into contrib/)
- add a flag "./aspect --test" to run them
- add some example tests in unittests/
- create necessary machinery in cmake to generate one additional ctest test called "unittests"


gassmoeller:
Very nice, I have to take a closer look later, but would it make sense to include this right away as a new stage in Jenkins?

tjhei:
> right away as a new stage in Jenkins?

it should already run as there is now an additional test called "unittests" as part of "ctest". We could also do the unittests separately...

tjhei:
> We could also do the unittests separately...

pro: more visible in testing. disadvantage: it is not run automatically for users anymore when they type "make test" or "ctest".

bangerth:
(i) How do you test functions not defined in header files as long as we haven't broken ASPECT into a library and a `main` file?

(ii) Does the unittest framework allow running tests in parallel? I find it difficult in these frameworks to figure out how to run individual tests.

tjhei:
> (i) How do you test functions not defined in header files as long as we haven't broken ASPECT into a library and a main file?

The tests are part of the main executable so I can test any function.

> (ii) Does the unittest framework allow running tests in parallel?

We could make that happen but I expect unittests to be _very_ fast, especially because we don't do any IO.

> I find it difficult in these frameworks to figure out how to run individual tests.

You can just run all tests (if they are fast), but it is quite easy:
```
$ ./aspect --test
===============================================================================
All tests passed (12 assertions in 2 test cases)

$ ./aspect --test -h

Catch v2.2.3
usage:
  aspect [<test name|pattern|tags> ... ] options

where options are:
  -?, -h, --help                            display usage information
  -l, --list-tests                          list all/matching test cases
...

$ ./aspect --test -l
All available test cases:
  floating point checks
  Utilities weighted_p_norm_average
2 test cases

$ ./aspect --test "float*"
===============================================================================
All tests passed (1 assertion in 1 test case)
```

bangerth:
On 06/21/2018 05:34 PM, Timo Heister wrote:
>     (i) How do you test functions not defined in header files as long as we
>     haven't broken ASPECT into a library and a main file?
> 
> The tests are part of the main executable so I can test any function.

Oh, you're linking the tests into the executable? If we end up with a couple 
of hundred tests over time, isn't that going to lead to massive bloat?


>     I find it difficult in these frameworks to figure out how to run
>     individual tests.
> 
> You can just run all tests (if they are fast), but it is quite easy:

I see.

bangerth:
Probably more about executable size. If we get into the realm of deal.II, then there is about as much code in the test suite as there is anywhere else. I suspect that's not going to happen in ASPECT, but you never know. Though then we still have the option of separating ASPECT into a library.

I will note that we've made efforts to get the material models and boundary conditions out of ASPECT proper that are only used for testcases. It'd be a bit of a shame to move stuff back into ASPECT for testing, even though it is, of course, not visible to user.

tjhei:
Would you be happier if we only include the tests in debug mode?

bangerth:
Marginally.

But as I mentioned, we can always address this at a later time by breaking ASPECT into a library and a driver, and then we can put all of the tests into a separate driver.

CMakeLists.txt
```diff
@@ -159,8 +159,24 @@ ADD_CUSTOM_TARGET(setup_tests
 SET(ASPECT_TEST_GENERATOR "Unix Makefiles" CACHE STRING 
   "Generator to use for the test cmake project. Using ninja instead of make is not recommended.")
 
+FILE(WRITE ${CMAKE_BINARY_DIR}/CTestTestfile.cmake "# auto-generated ctest file\n")
```
bangerth: can you add a comment here that explains what the following code actually does?

source/main.cc
```diff
@@ -42,6 +42,8 @@
 #  endif
 #endif
 
+#define CATCH_CONFIG_RUNNER
+#include "catch.hpp"
```
bangerth: Could use a comment that explains what exactly is happening with the `#define`

source/main.cc
```diff
@@ -579,6 +582,11 @@ int main (int argc, char *argv[])
           use_threads = true;
 #endif
         }
+      else if (arg == "--test")
+        {
+          run_unittests = true;
+          break;
+        }
```
bangerth: Might as well run the tests right here and return, rather than defer to below.

source/main.cc
```diff
@@ -602,6 +610,19 @@ int main (int argc, char *argv[])
       // thrown to avoid MPI deadlocks.
       Utilities::MPI::MPI_InitFinalize mpi_initialization(n_remaining_arguments, remaining_arguments, use_threads ? numbers::invalid_unsigned_int : 1);
 
+      if (run_unittests)
+        {
+          int new_argc = n_remaining_arguments + 1;
+          std::vector<char *> args;
+          args.push_back(argv[0]);
```
bangerth: does this just append a string?

unittests/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+################### top matter ################
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.8)
```
bangerth: I don't like the name `unittests`. Use `unit_tests`.

bangerth: and think about replacing the merged word in all other places as well.

unittests/common.h
```diff
@@ -0,0 +1,27 @@
...
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_unittests_common_h
+#define _aspect_unittests_common_h
+
+#include <catch.hpp>
```
bangerth: Why not just `#include` this in all the places where we do `#include <common.h>` right now?

unittests/first.cc
```diff
@@ -0,0 +1,25 @@
+#include "common.h"
```
bangerth: standard header missing

source/main.cc
tjhei: I would like to do this after initializing MPI so we can later create vectors, etc..

bangerth: Ah, I see. OK then.

unittests/common.h
tjhei: You are right in that it is kind of useless right now, but I want to put extra utilities to be used in tests in this file in the very near future.

bangerth: OK

tjhei:

CMakeLists.txt
```diff
@@ -24,8 +24,8 @@ MESSAGE(STATUS "====================================================")
 # Set the name of the project and target:
 SET(TARGET "aspect")
 
-FILE(GLOB_RECURSE TARGET_SRC  "source/*.cc" "include/*.h")
-INCLUDE_DIRECTORIES(include ${CMAKE_BINARY_DIR}/include)
+FILE(GLOB_RECURSE TARGET_SRC  "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
```
bangerth: Do we need the header files here? I wonder whether `include/*.h` actually expands to anything, but it's something I don't know one way or the other anyway.

CMakeLists.txt
```diff
@@ -159,8 +159,51 @@ ADD_CUSTOM_TARGET(setup_tests
...
+  ENDIF()
+
+  # Finally hook up that "make test" will execute the tests:
+  ADD_CUSTOM_TARGET(run_unit_tests
+    COMMAND ${CMAKE_BINARY_DIR}/aspect --test
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/unit_tests
+    DEPENDS aspect
+    COMMENT "Running unit_tests ...")
+
+  ADD_DEPENDENCIES(test run_unit_tests)
```
bangerth: Ah, hum -- so calling `make test` checks the unit tests, but not the whole rest? I would have hoped that we can treat the unit tests the same way as all other tests -- i.e., by creating a particular test the same way we do for all of the .prm-based tests, just executing aspect differently.

doc/indent
```diff
@@ -46,10 +46,10 @@ fi
 # with up to 10 in parallel
 echo "--- Indenting all ASPECT header and source files"
 
-find tests include source benchmarks cookbooks \( -name '*.cc' -o -name '*.h' \) -print | xargs -n 50 -P 10 astyle --options=doc/astyle.rc
+find tests include source benchmarks cookbooks unittests \( -name '*.cc' -o -name '*.h' \) -print | xargs -n 50 -P 10 astyle --options=doc/astyle.rc
```
bangerth: wrong dir name. same in line 52

doc/indent
```diff
@@ -61,4 +61,4 @@ dos_to_unix()
     rm -f $f.tmp
 }
 export -f dos_to_unix
-find tests include source  \( -name '*.cc' -o -name '*.h' -o -name '*.prm' \) -print | xargs -n 1 -P 10 -I {} bash -c 'dos_to_unix "$@"' _ {} 
+find tests include source unittests \( -name '*.cc' -o -name '*.h' -o -name '*.prm' \) -print | xargs -n 1 -P 10 -I {} bash -c 'dos_to_unix "$@"' _ {}
```
bangerth: and here

CMakeLists.txt
tjhei: it is more complicated but I don't have a better way:
"make test" depends on "make run_unit_tests" and "make run_tests" (see the two calls to add_dependencies). They call "cmake build --target test" in the subdirectories.
Alternatively, you can call "ctest" in one of the subdirectories to run either the unit-tests or the other ones.
Finally, if you run ctest in the base directory (your build directory), the CTestTest.cmake file will let ctest to pick up tests in both subdirectories...

tjhei:
updated!

Issue #2365, opened 2018-06-21
ASCII data plugins have hard-coded subsection name


mnkajan:
This means that you cannot combine two plugins that call Utilities::AsciiDataBase::declare_parameters(). For example, you cannot create a composite initial temperature model that combines the 'ascii profile' and 'ascii data' plugins, because you can only define a single data file under the hard-coded subsection "Ascii data model" in the parameter file.

I've discussed this with @gassmoeller who suggested removing the hard-coded subsection name in utilities.cc. However, this would require entering section names for each plugin that uses the AsciiDataBase functions to declare and parse parameters.

Any alternative solutions or ideas are very welcome!

gassmoeller:
This was fixed by #2370 so @mnkajan you should be able to use the new functionality. Let us know when this can be closed or if you have questions.

mnkajan:
Yup, we can close this now!

PR #2370, opened 2018-06-21
Make the name of the subsection in AsciiData an input

jaustermann:
At the moment the fixed subsection name for any use of AsciiData is "Ascii data model". This might sometimes not be very descriptive. For example, I want to add an option to read in a vs to density scaling from a file. If I do this using AsciiData there is nothing in the parameter file that tells the user that the ascii data model is used for the vs to density scaling. This pull request makes the subsection name an input, which can be changed to e.g. "Ascii data model for vs to density scaling". The default is subsection name is set to "Ascii data model" to be backward compatible. This addresses issue #2365.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -615,13 +615,15 @@ namespace aspect
         void
         declare_parameters (ParameterHandler  &prm,
                             const std::string &default_directory,
-                            const std::string &default_filename);
+                            const std::string &default_filename,
+                            const std::string &subsection_name = "Ascii data model");
 
         /**
          * Read the parameters from the parameter file.
          */
         void
-        parse_parameters (ParameterHandler &prm);
+        parse_parameters (ParameterHandler &prm,
+                          const std::string &subsection_name = "Ascii data model");
 
         /**
          * Directory in which the data files are present.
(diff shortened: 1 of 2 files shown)


gassmoeller:
Not sure why @jaustermann is not on the list of trusted users, but this PR looks good, and should be ready to merge when the testers are finished.

tjhei:
@jaustermann please set your correct email address/username in git and update the commit. Let me know if you need help with that.

include/aspect/utilities.h
```diff
@@ -615,13 +615,15 @@ namespace aspect
         void
         declare_parameters (ParameterHandler  &prm,
                             const std::string &default_directory,
-                            const std::string &default_filename);
+                            const std::string &default_filename,
+                            const std::string &string = "Ascii data model");
```
tjhei: can we use a parameter name that is not ``string``? I find this really confusing.

jaustermann:
@tjhei I changed &string to a more descriptive &default_subsection_name in the header file and fixed it up with the previous commit.

source/utilities.cc
```diff
@@ -1501,9 +1501,10 @@ namespace aspect
     void
     AsciiDataBase<dim>::declare_parameters (ParameterHandler  &prm,
                                             const std::string &default_directory,
-                                            const std::string &default_filename)
+                                            const std::string &default_filename,
+                                            const std::string &default_subsection_name)
     {
-      prm.enter_subsection ("Ascii data model");
+      prm.enter_subsection (default_subsection_name);
```
bangerth: I think you chose the variable name incorrectly. It is not the "default name", but this really is the *name* of the section. So maybe the name of the function argument should be "subsection_name", and it can of course have a default value, but the name of the variable should no longer contain the part "default".

jaustermann:
@bangerth got it. I've addressed this and changed default_subsection_name to subsection_name throughout.

PR #2371, opened 2018-06-21
Add distance_to_line function and use it for distance_to_polygon

anne-glerum:
Functionality to setup initial conditions, for example temperature based on distance to a ridge.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -197,6 +197,18 @@ namespace aspect
     signed_distance_to_polygon(const std::vector<Point<2> > &point_list,
                                const dealii::Point<2> &point);
 
+
+    /**
+    * Given a 2d point and a list of two points that define a line, compute the smallest
+    * distance of the point to the line segment. Note that when the point's perpendicular
+    * base does not lie on the line segment, the smallest distance to the segment's end
+    * points is calculated.
+    */
+    template <int dim>
+    double
+    distance_to_line(const std::vector<dealii::Point<2> > &point_list,
+                     const dealii::Point<2> &point);
+
     /**
      * Given a vector @p v in @p dim dimensional space, return a set
      * of (dim-1) vectors that are orthogonal to @p v and to each
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -199,14 +199,13 @@ namespace aspect
 
 
     /**
-    * Given a 2d point and a list of two points that define a line, compute the smallest
-    * distance of the point to the line segment. Note that when the point's perpendicular
-    * base does not lie on the line segment, the smallest distance to the segment's end
-    * points is calculated.
-    */
-    template <int dim>
+     * Given a 2d point and a list of two points that define a line, compute the smallest
+     * distance of the point to the line segment. When the point's perpendicular
+     * base does not lie on the line segment, the smallest distance to the segment's end
+     * points is calculated.
+     */
     double
-    distance_to_line(const std::vector<dealii::Point<2> > &point_list,
+    distance_to_line(const std::array<dealii::Point<2>,2 > &point_list,
                      const dealii::Point<2> &point);
 
     /**
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -479,31 +479,60 @@ namespace aspect
 
       for (unsigned int i = 0; i < n_poly_points; ++i)
         {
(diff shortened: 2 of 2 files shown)


include/aspect/utilities.h
```diff
@@ -197,6 +197,18 @@ namespace aspect
     signed_distance_to_polygon(const std::vector<Point<2> > &point_list,
                                const dealii::Point<2> &point);
 
+
+    /**
+    * Given a 2d point and a list of two points that define a line, compute the smallest
```
bangerth: Please fix the indentation here.

include/aspect/utilities.h
```diff
@@ -197,6 +197,18 @@ namespace aspect
...
+    * distance of the point to the line segment. Note that when the point's perpendicular
```
bangerth: Just drop the "Note that" here -- it's an overused phrase :-)

include/aspect/utilities.h
```diff
@@ -197,6 +197,18 @@ namespace aspect
...
+    * base does not lie on the line segment, the smallest distance to the segment's end
+    * points is calculated.
+    */
+    template <int dim>
+    double
+    distance_to_line(const std::vector<dealii::Point<2> > &point_list,
```
bangerth: Since you know that it's only two lines, why not use a `std::array<Point<2>,2>`?

bangerth: Also, nothing in this function depends on the template argument `dim` -- I think you can just make it a non-template.

Unless, of course, you want to generalize this to general line segments in 3d as well :-)

source/utilities.cc
```diff
@@ -479,33 +479,62 @@ namespace aspect
...
-            }
+          std::vector<Point<2> > list = {point_list[i], shifted_point_list[i]};
+          distances[i] = distance_to_line<dim>(list, point);
         }
 
       // Return the minimum of the distances of the point to all polygon segments
       return *std::min_element(distances.begin(),distances.end()) * sign;
     }
 
+    template <int dim>
```
bangerth: two more empty lines here

anne-glerum:
@gassmoeller My build is not trusted on Jenkins yet, could you start the tester?

bangerth:
It passed three checks, that's probably good enough :-)

source/utilities.cc
```diff
@@ -479,33 +479,61 @@ namespace aspect
...
+       *
+       */
+
+      const unsigned int n_poly_points = point_list.size();
+      AssertThrow(n_poly_points == 2, ExcMessage("A list of points for a line segment should consist of 2 points."));
+
+      // Create vector along the polygon line segment P0 to P1
+      Tensor<1,2> vector_segment = point_list[1] - point_list[0];
+      // Create vector from point P to the second segment point
+      Tensor<1,2> vector_point_segment = point - point_list[0];
```
bangerth: make both of these `const`

source/utilities.cc
```diff
@@ -479,33 +479,61 @@ namespace aspect
...
-            }
+          std::array<Point<2>,2 > list = {point_list[i], shifted_point_list[i]};
+          distances[i] = distance_to_line(list, point);
         }
 
       // Return the minimum of the distances of the point to all polygon segments
       return *std::min_element(distances.begin(),distances.end()) * sign;
     }
 
+    double
```
bangerth: three empty lines between functions. same after this function

source/utilities.cc
```diff
@@ -479,33 +479,61 @@ namespace aspect
...
-          else if (c2 <= c1)
-            distances[i] = (Tensor<1,2> (shifted_point_list[i] - point)).norm();
-          // perpendicular base line lies on segment
-          else
-            {
-              const Point<2> point_on_segment = point_list[i] + (c1/c2) * vector_segment;
-              distances[i] = (Tensor<1,2> (point - point_on_segment)).norm();
-            }
+          std::array<Point<2>,2 > list = {point_list[i], shifted_point_list[i]};
+          distances[i] = distance_to_line(list, point);
```
bangerth: make the `list` variable `const`

source/utilities.cc
anne-glerum: The number of empty lines between the functions in this file varies, but is usually only 1, therefore I kept it to 1.

bangerth: I see. Then let's leave it at that.

PR #2379, opened 2018-06-22
Fix Jenkins email address

anne-glerum:


tjhei:
I guess we don't need to run tests...

PR #2380, opened 2018-06-22
improve comments in the S40RTS initial temperature model

jaustermann:
I added a comment to the beginning of S40 to point out that it uses a different normalization than is normally used. We already had this comment in the code but further down. The purpose of having it at the beginning is to alert people who might copy and paste this for some other spherical harmonic based functionality that they need to change this. In most cases the normalization in S40 is not what they will want.

### DIFF
--- a/source/initial_temperature/S40RTS_perturbation.cc
+++ b/source/initial_temperature/S40RTS_perturbation.cc
@@ -41,6 +41,11 @@ namespace aspect
         // Ritsema et al. choose real sine and cosine coefficients that follow the normalization
         // by Dahlen & Tromp, Theoretical Global Seismology (equations B.58 and B.99).
 
+        // NOTE: There is a factor of sqrt(2) difference between the standard orthonormalized
+        // spherical harmonics used by Dahlen & Tromp and that used for S40RTS (see PR # 966).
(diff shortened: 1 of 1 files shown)


source/initial_temperature/S40RTS_perturbation.cc
```diff
@@ -41,6 +41,11 @@ namespace aspect
         // Ritsema et al. choose real sine and cosine coefficients that follow the normalization
         // by Dahlen & Tromp, Theoretical Global Seismology (equations B.58 and B.99).
 
+        // NOTE: There is a factor of sqrt(2) difference between the standard orthonormilzed
```
anne-glerum: orthonormalized?

jaustermann: haha ... will edit and fix up

bangerth:
The jenkins build failed, but for spurious reasons. All of the other testers work fine, so I think the build at least is fine.

But I'll let @anne-glerum make the decision if this can be merged :-)

jaustermann:
Hm ... not sure what the issue is with jenkins. I only changed comments, no actual code. @anne-glerum what do you think? I'm about to send a new pull request for this file so it would be nice to rebase before if you're happy with this commit.

PR #2384, opened 2018-06-22
Remove ConstraintMatrix in global.h

gassmoeller:
This is a fixup to #2329. I do not like to include more things into global.h, we should only include what is necessary. In practice this will not make a big difference, because most plugins include simulator_access.h, but when we require deal.II 9.1 we can replace the include in the simulator_access.h by a forward declaration and probably save some compile time. For the moment I think this is a cleaner solution than just putting `ConstraintMatrix` into global.h.

### DIFF
--- a/include/aspect/global.h
+++ b/include/aspect/global.h
@@ -36,9 +36,6 @@ DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
 #  include <deal.II/lac/trilinos_precondition.h>
 #endif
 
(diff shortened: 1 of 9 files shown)


bangerth:
Yep, that's nicer!

tjhei:
@gassmoeller put this into 2.0.1? See #2262

gassmoeller:
No, I would prefer to not add this, as it only makes code nicer, and might break things I have not noticed.

PR #2386, opened 2018-06-22
As is usual, make a plugin name lower-case.

bangerth:
All plugins use lower-case letters when they are identified in the place
where one selects a plugin, and an initial upper-case letter as a section
name. However, the DynamicCore class uses 'Dynamic core' also as the
plugin name.

Fix this.

This is clearly an incompatible change. I don't know how we usually handle
this: it's a bit of an obscure plugin, so I'd venture the guess that not
many people will actually notice, and maybe we shouldn't be terribly
concerned with backward compatibility. But I still wanted to see what
people think about it first.

### DIFF
--- a/doc/modules/changes/20180620_bangerth-2
+++ b/doc/modules/changes/20180620_bangerth-2
@@ -1,0 +1,8 @@
+Incompatibility: When selecting the 'dynamic core' plugin for the
+boundary temperature, the plugin previously used the "Dynamic core"
+spelling with a leading upper-case letter. However, all other plugins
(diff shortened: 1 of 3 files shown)


jdannberg:
I am also okay with this if you add the script that @gassmoeller suggests (and mention in the changelog entry that this can be used to update the input file).

gassmoeller:
This is essentially ready, and the update script is not that important (it is a very rarely used plugin). Let's merge to not have this lying around another year. A recent test run merged with the current master succeeds.

bangerth:
Oh look, only took nearly a year ;-)

PR #2393, opened 2018-06-22
update release-tasklist

tjhei:


doc/release-tasklist
```diff
@@ -73,6 +84,7 @@
       gpg --detach-sign --armor manual-$TAG.pdf
       sha1sum aspect-$TAG.tar.gz manual-$TAG.pdf >sha1sum-$TAG.txt
   . create a release on github, upload tar.gz and manual-$VER.pdf
+  . readme / release notes: add zenodo DOI button
```
gassmoeller: Can you elaborate on this a bit more? I would not know where or how to add a DOI button.

doc/release-tasklist
```diff
@@ -57,13 +57,24 @@
       cd doc && ./bump_version.sh $TAG && cd .. && \
       git commit -m "version $TAG" && \
       git tag -s v$TAG -m "version $TAG"
-  . create a tar file:
+  . create a tar file: TODO: remove manual.pdf
```
gassmoeller: What about this TODO?

tjhei: I thought I don't spend the time right now to change and test the commands to not include the manual.pdf in the tarball but do this next time. Do you want me to update the bash commands now?

doc/release-tasklist
tjhei: yeah. I think this is something we should discuss before we put this into effect. Basically, @ljhwang said we could do something like this: https://github.com/geodynamics/pylith/releases/tag/v2.2.1

gassmoeller: Seems nice. Are there arguments against this?

doc/release-tasklist
gassmoeller: That would be great, yes.

tjhei:
updated. Okay like this?

PR #2398, opened 2018-06-23
Fix typos

bartniday:


gassmoeller:
The Jenkins tester actually succeeded but does not update its status :-(

tjhei:
living on the edge: I will just merge because testers are stuck...

PR #2403, opened 2018-06-23
Add new function to return the full list of compositional fields

pmbremner:
The change allows the full list of compositional fields to be returned in a vector. This change was needed for adding a different new function in utilities.cc to parse maps. The reference for this related function will be added soon.

### DIFF
--- a/include/aspect/introspection.h
+++ b/include/aspect/introspection.h
@@ -349,7 +349,6 @@ namespace aspect
 
       /**
        * A function that returns the full list of compositional field names.
-       * No arguments needed.
        */
       const std::vector<std::string>&
       get_composition_names () const;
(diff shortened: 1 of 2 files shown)


include/aspect/introspection.h
```diff
@@ -347,6 +347,13 @@ namespace aspect
       std::string
       name_for_compositional_index (const unsigned int index) const;
 
+      /**
+       * A function that returns the full list of compositional field names.
+       * No arguments needed.
```
bangerth: no need for that last sentence

source/simulator/introspection.cc
```diff
@@ -263,6 +263,14 @@ namespace aspect
     return composition_names[index];
   }
 
+  template <int dim>
```
bangerth: three empty lines between functions, before and after this function

PR #2406, opened 2018-06-23
Be extra careful when parsing the dimension in an input file.

### LINKED ISSUE #2367: Improve error message with DOS line endings
If a .prm file has DOS line endings, we get unintelligible error messages. They come out of the function that determines the space dimension, which is the first place we read anything from a .prm file. That's the place where we need to catch the C++ exception that comes out of `Utilities::string_to_int` and see whether we can print a better error message.

bangerth:
That's because the dimension is the first thing we read from an input file. Specifically,
pay attention with Windows-style line endings. The rest is explained in the comments.

Fixes #2367. I did check all possible mistakes by hand, but I don't know how to add a test 
that uses Windows line endings and that I can add to the test suite -- nor how to reliably
catch the resulting error.

### DIFF
--- a/source/main.cc
+++ b/source/main.cc
@@ -109,20 +109,64 @@ get_last_value_of_parameter(const std::string &parameters,
 }
 
 
-// extract the dimension in which to run ASPECT from the
-// the contents of the parameter file. this is something that
+// Extract the dimension in which to run ASPECT from the
+// the contents of the parameter file. This is something that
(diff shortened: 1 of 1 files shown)


tjhei:
> I don't know how to add a test
> that uses Windows line endings

I think this won't be necessary. This would also conflict with ./doc/indent which tries to detect incorrect endings in tests.

Issue #1690, opened 2017-05-15
Refine right away at first timestep


jaustermann:
It would be nice if we could check whether the refinement strategy depends on the solution of the first timestep and if it doesn't (e.g. boundary, topography) refine right away without calculating the solution first. This only matters for the initial refinement.

gassmoeller:
Hi Jacky,
if you are still interested in this, I think it should not be too hard to implement. I could imagine a parameter in the mesh refinement section called `Skip initial adaptive refinement solves`, which contains documentation that it is only save to set this to true if the refinement criteria are independent of the solution. Then one would only need to modify core.cc:2309-2314 to include a loop that calls `maybe_do_initial_refinement` until no more initial refinement is done, and then one final `goto start_time_iteration` to actually solve for the final mesh. Do you want to give this a try?

jaustermann:
Hi Rene, thanks for the pointers! I will give this a try.

bobmyhill:
@jaustermann do you want to assign this to yourself?

jaustermann:
@bobmyhill I haven't worked on this yet so feel free to tackle it if you want

bobmyhill:
Ok, this is mine!

jaustermann:
This is addressed by #2409.

PR #2409, opened 2018-06-23
Possibility to turn off the solvers and initial conditions setup during the initial refinement

LudovicJnnt:
For those who want to skip solvers and the initial conditions setup, this pull request adds 2 parameters for the subsection mesh refinement.

AssertThrow for solvers ON while setup initial conditions OFF does not work for some reason.
AssertThrow for skipping setup initial conditions while running postprocessors on initial refinement does not work either.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -1639,7 +1639,7 @@ namespace aspect
               {
                 const bool initial_refinement_done = maybe_do_initial_refinement(max_refinement_level);
                 if (initial_refinement_done)
-                goto start_time_iteration;
+                  goto start_time_iteration;
               }
           }
 
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -1591,6 +1591,13 @@ namespace aspect
 
     if (parameters.resume_computation == false)
       {
+        if (parameters.solvers_off_on_initial_refinement)
+          {
+            const bool initial_refinement_done = maybe_do_initial_refinement(max_refinement_level);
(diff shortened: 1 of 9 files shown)


jaustermann:
This addresses issue #1690. Looks like astyle test failed, @LudovicJnnt did you run astyle on the submitted code?

source/simulator/core.cc
```diff
@@ -1591,6 +1591,16 @@ namespace aspect
 
     if (parameters.resume_computation == false)
       {
+        // See if we want to skip the initial conditions setup during the pre-
+        // refinement phase. If set true, the solvers won't be executed. An
```
gassmoeller: If set true, the solvers won't be executed -> If set **to** true, the solver **will not** be executed/

LudovicJnnt:
I added 2 tests for skipping solvers and/or initial conditions setup during the pre-refinement
I applied the changes from Rene.

gassmoeller:
Let me see if I can simplfy the logic somewhat. After that this is ready to merge.

gassmoeller:
@LudovicJnnt: I updated the PR with a different logic that should produce the same result. Can you update your local branch and check that your models still work (do a git pull from this PR). The tests still seem to work.

LudovicJnnt:
Right, yes models still work. 
Thank you!

gassmoeller:
Then this is ready to merge when the tester is happy. Also please add an entry into doc/modules/changes and add a few sentences into the hackathon report.

PR #2414, opened 2018-06-24
jenkins: make more robust

tjhei:
- generate output while prebuilding test to hopefully avoid the timeout
disconnects happening
- ignore the result of the prebuild but use the output of ctest to check
if tests failed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -78,18 +78,24 @@ pipeline {
       } 
     }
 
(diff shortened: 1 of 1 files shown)


tjhei:
I am not sure if this will work but I guess these changes won't hurt. If you want me to add bogus commits here to see if it passes/fails correctly, let me know.

PR #2417, opened 2018-06-24
update zenodo badge in readme

tjhei:


bangerth:
What's different between these badges? What frustrates me a bit about these badges is that they link to totally obscure URLs.

PR #2418, opened 2018-06-24
Fix compile warning

tjhei:
fixes #2406

```
/home/dealii/jenkins/workspace/aspect_master-E2LGBV7TXCDSE4TFSDUSL7EULM3J3UXVDF7HPXGMHNZ7M2DF7ICQ/source/main.cc:171:1: warning: control reaches end of non-void function [-Wreturn-type]
```


tjhei:
we compile with -Werror on jenkins, so we better fix this quickly!

gassmoeller:
I take the blame if this one does not work.

PR #2420, opened 2018-06-24
Jenkins: disable openmpi output

tjhei:
on some machines mpirun will produce an OpenFabrics warning that will make all
tests fail. Just disable this transport protocol.


bangerth:
Thanks for updating the indentation ;-)

PR #2427, opened 2018-06-25
unit_tests: convert tests/additional_output.prm

tjhei:


tjhei:
oops, broke 2 tests. Will get back to this...

bangerth:
I guess we have a whole bunch of these kinds of tests. Is that worth anyone's time to convert them?

tjhei:
> I guess we have a whole bunch of these kinds of tests. Is that worth anyone's time to convert them?

We have 11 of those. This is not too much work and will serve as an example for future unit tests, I think. It is also a way to learn how to write tests like this.

tjhei:
this is ready!

gassmoeller:
This is correct!

PR #2428, opened 2018-06-25
unit_tests: add print info of --test

tjhei:


source/main.cc
```diff
@@ -493,23 +493,16 @@ parse_parameters (const std::string &input_as_string,
...
-            << "       --output-plugin-graph  (write a representation of all plugins to standard output and exit)"
-            << std::endl
+  std::cout << "    optional arguments [args]:\n"
+            << "       -h, --help             (for this usage help)\n"
+            << "       -v, --version          (for information about library versions)\n"
+            << "       -j, --threads          (to use multi-threading)\n"
+            << "       --output-xml           (print parameters in xml format to standard output and exit)\n"
+            << "       --output-plugin-graph  (write a representation of all plugins to standard output and exit)\n"
+            << "       --test                 (run the unit tests from unit_tests/, run --test -h for more info)\n"
             << std::endl;
```
bangerth: You're adding one empty line here (compared to the state before), but I don't think it matters.

PR #2429, opened 2018-06-25
Add AssertThrow to check 'uniform box' particle generator limits

bartniday:
I was getting floating point exceptions when I tried to generate particles using a uniform box with maximum y less than minimum y.

This adds an AssertThrow to prevent that.


gassmoeller:
Very nice! Thanks for fixing!

gassmoeller:
You might need to include `<deal.II/base/exceptions.h>` at the top of the source file to make the tester happy.

bartniday:
is this failed test from the old commit, or did that include not fix it?

bartniday:
oh. This fixes it.

PR #2432, opened 2018-06-25
jenkins: limit ninja processes

tjhei:
I think it is not useful to run N+2 processes with N cores (ninja default),
because some of our machines are short on RAM (at least tjhei-timo).


tjhei:
this shouldn't be necessary for the test pre-building as the processes use only a small amount of RAM.

gassmoeller:
Yes, this is really useful (should allow running jenkins on my laptop). Ready to merge (the tester will not test this).

PR #2436, opened 2018-06-25
www: update 2.0.1 zenodo in cite.html

tjhei:


bangerth:
Hm, it turns out that a dot release is actually a lot of work...

PR #2438, opened 2018-06-25
Clean up get_parameters().convert_to_years in material models and postprocessors

hfmarkp:
The flag for converting years to seconds can be checked with the wrapper function convert_output_to_years(), and all but three of the material models and postprocessors use the wrapper. @tjhei suggested cleaning up those three files to make things more consistent/searchable, so I've done that.


bangerth:
/run-tests

Issue #2388, opened 2018-06-22
Single timestep Stokes solver


bartniday:
In my personal version of Aspect I'm solving particle advection in a steady-state flow field, via a hack. I just skip the call to `solve_timestep()` on timesteps after the first.

After talking to @gassmoeller yesterday, it sounds like a solver (something like "one_timestep_Stokes_no_advection"?) that does this properly would be useful for other things.

bartniday:
fixed by #2441.

PR #2441, opened 2018-06-25
Solver scheme that only runs on the first timestep

bartniday:
Addresses #2388.

This runs Stokes on the initial timestep (with no nonlinear iterations) and never solves the temperature or composition systems. I think that is what was agreed on.


gassmoeller:
Looks good. Thanks!

Issue #2402, opened 2018-06-23
EllipsoidalChunk: move set/clear_manifold_ids into .cc


tjhei:
by making them global functions in an internal namespace...

PR #2442, opened 2018-06-25
Move the definition of the function into the header in ellipsoidal chunk.

alarshi:
Moved set/clear_manifold_ids into .cc to address issue #2402.


gassmoeller:
Looks good, thank you! Ready to merge if the tester is happy.

PR #2445, opened 2018-06-25
Make a local namespace anonymous.

bangerth:
The functions in it are only used in a single .cc file, so they don't need
to be exported from the object file.

This is another follow-up to #2442 and #2444, in pursuance of #2402.


tjhei:
how do anonymous namespaces interact with unity builds?

bangerth:
Good question. I suspect we will find out :-)

(If it doesn't work, then we need to enclose the anonymous namespace in one more level of namespaces.)

bangerth:
It seems to work :-)

PR #2401, opened 2018-06-23
read in vs to density scaling from file

jaustermann:
I've introduced an option to read in a vs to density scaling from a file. This is relevant for S40RTS and SAVANI. For now I've only included this in S40RTS but can update SAVANI accordingly once I've gotten feedback on this pull request.

PR #2446, opened 2018-06-25
Vs to density scaling savani

jaustermann:
This allows reading in a vs to density profile from a file instead of just setting it to constant. This pull request implements this for the SAVANI initial condition. This is analogous to the merged pull request #2401 for S40RTS.


jaustermann:
I've set the default directory for the vs to density file to data/initial-temperature/S40RTS and not data/initial-temperature/SAVANI. I did that because this way we don't have to have the vs_to_density_Steinberger.txt file twice. Will this lead to confusion?

gassmoeller:
Looks good, ready to merge.

PR #2448, opened 2018-06-25
ASPECT does not run in 4D (yet)

joeschools:
As far as I know, ASPECT won't run in four dimensions. Updated the manual accordingly.


bangerth:
Very interesting. I wonder how that statement got in there...

PR #2459, opened 2018-06-26
Check that spherical shell inner radius is less than outer radius

bartniday:


gassmoeller:
Yes, that makes sense :+1:

source/geometry_model/spherical_shell.cc
```diff
@@ -396,7 +396,6 @@ namespace aspect
                              "this geometry are 90, 180, and 360 in 2d; "
                              "and 90 and 360 in 3d. "
                              "Units: degrees.");
-
```
MFraters: why are you removing this line?

MFraters: hmm, interesting. I never realized we (also me if I look at other models) did that :s

PR #2463, opened 2018-06-26
Change default limits for 'uniform radial' particle generator

bartniday:
The default maximum longitude and latitude are pi, which does not make sense for angles measured in degrees.

### DIFF
--- a/doc/modules/changes/20180625_niday
+++ b/doc/modules/changes/20180625_niday
@@ -1,0 +1,4 @@
+Fixed: The default latitude and longitude limits for the 'uniform radial' particle generator were set to (0, pi) even though they are measured in degrees.
+The new default limits correspond to the entire surface of the sphere, and the manual clarifies that latitude is in (0, 180) and not (-90, 90).
+<br>
+(Bart Niday, 2018/06/25)
--- a/source/particle/generator/uniform_radial.cc
+++ b/source/particle/generator/uniform_radial.cc
@@ -171,18 +171,20 @@ namespace aspect
                                    Patterns::Double (0,360),
                                    "Minimum longitude coordinate for the region of particles "
                                    "in degrees. Measured from the center position.");
-                prm.declare_entry ("Maximum longitude", "3.1415",
+                prm.declare_entry ("Maximum longitude", "360",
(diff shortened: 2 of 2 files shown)


bangerth:
Not only did the previous default have the wrong units, it also only covered half of the world (pi = 180 degrees).

source/particle/generator/uniform_radial.cc
```diff
@@ -171,15 +171,15 @@ namespace aspect
...
                                    "Maximum longitude coordinate for the region of particles "
                                    "in degrees. Measured from the center position.");
                 prm.declare_entry ("Minimum latitude", "0",
                                    Patterns::Double (0,180),
                                    "Minimum latitude coordinate for the region of particles "
                                    "in degrees. Measured from the center position.");
-                prm.declare_entry ("Maximum latitude", "3.1415",
+                prm.declare_entry ("Maximum latitude", "180",
                                    Patterns::Double (0,180),
                                    "Maximum latitude coordinate for the region of particles "
```
bangerth: Isn't the maximal latitude 90 degrees? I don't know how this is used, but a latitude of 180 degrees doesn't make much sense.

bangerth:
I think this would deserve a changelog.

source/particle/generator/uniform_radial.cc
bartniday: It looks like the "maximum latitude" and "minimum latitude" are actually colatitude.
Does it make more sense to clarify that in the documentation, or to change things so that latitude is between -90 and 90?

bangerth: My preference is to keep things compatible, but to change the documentation to explain what it really is.

bartniday:
Added a phrase to clarify that "latitude" is not latitude.

doc/modules/changes/20180625_niday
```diff
@@ -0,0 +1,3 @@
+Fixed: The default latitude and longitude limits for the 'uniform radial' particle generator were set to (0, pi) even though they were measured in degrees.
```
gassmoeller: Please add how they are measured now, otherwise the reader will not know what is the fixed version.

PR #2381, opened 2018-06-22
Fix two performance problems in the Newton assembly.

MFraters:
This fixes two problems in the Newton solver assembly discovered by @gassmoeller. The first problem is that the functions call in the inner most loop the function `this->get_pressure_scaling()`. We now store this value in the beginning and reuse it. The second problem was that the tensor contraction for the symmetrisation was very expensive. We could use a similar trick here, making it a lot cheaper.

The stats before:
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      88.6s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble Stokes system          |         1 |      1.66s |       1.9% |
| Assemble Stokes system Newton   |         4 |      16.8s |        19% |
| Assemble Stokes system Picard   |         3 |      7.36s |       8.3% |
| Assemble Stokes system rhs      |         6 |      5.07s |       5.7% |
| Assemble temperature system     |         7 |      12.5s |        14% |
| Build Stokes preconditioner     |         7 |      19.5s |        22% |
| Solve Stokes system             |         7 |      12.6s |        14% |
| Initialization                  |         1 |       5.7s |       6.4% |
| Postprocessing                  |         1 |      2.55s |       2.9% |
| Setup dof systems               |         1 |     0.887s |         1% |
| Setup initial conditions        |         1 |      1.24s |       1.4% |
+---------------------------------+-----------+------------+------------+
```

stats after storing the pressure normalization:
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      75.2s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| Assemble Stokes system          |         1 |      1.41s |       1.9% |
| Assemble Stokes system Newton   |         4 |      10.9s |        15% |
| Assemble Stokes system Picard   |         3 |      4.29s |       5.7% |
| Assemble Stokes system rhs      |         6 |      5.09s |       6.8% |
| Assemble temperature system     |         7 |      11.9s |        16% |
| Build Stokes preconditioner     |         7 |      16.6s |        22% |
| Solve Stokes system             |         7 |      13.1s |        17% |
| Initialization                  |         1 |      5.23s |       6.9% |
| Postprocessing                  |         1 |       2.5s |       3.3% |
| Setup dof systems               |         1 |     0.816s |       1.1% |
[...]

PR #2465, opened 2018-06-26
fix newton assembly performance 2

MFraters:
This is a follow up to #2381.

### DIFF
--- a/source/simulator/assemblers/newton_stokes.cc
+++ b/source/simulator/assemblers/newton_stokes.cc
@@ -361,7 +361,7 @@ namespace aspect
                           {
                             data.local_matrix(i,j) += ( derivative_scaling_factor * alpha * (scratch.grads_phi_u[i] * (viscosity_derivative_wrt_strain_rate * scratch.grads_phi_u[j]) * strain_rate
                                                                                              + scratch.grads_phi_u[j] * (viscosity_derivative_wrt_strain_rate * scratch.grads_phi_u[i]) * strain_rate)
-                                                        + derivative_scaling_factor * this->get_pressure_scaling() * scratch.grads_phi_u[i] * 2.0 * viscosity_derivative_wrt_pressure * scratch.phi_p[j] * strain_rate )
+                                                        + derivative_scaling_factor * pressure_scaling * 2.0 * viscosity_derivative_wrt_pressure * scratch.phi_p[j] * scratch.grads_phi_u[i] * strain_rate )
                                                       * JxW;
 
                             Assert(dealii::numbers::is_finite(data.local_matrix(i,j)),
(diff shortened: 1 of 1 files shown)


bangerth:
There are a bunch of changes in the output. The first few pages look like the changes are small, but can you look at a more representative sample?

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -148,6 +148,11 @@ namespace aspect
                                    :
                                    1;
 
+              // pre-compute the tensor contractions
+              std::vector<SymmetricTensor<2,dim> > tensor_contraction(stokes_dofs_per_cell);
```
bangerth: that's maybe not a good name for the variable. How about `deta_deps_times_eps_times_phi_j`? That's a long name, but at least it's descriptive. Other suggestions also welcome :-)

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -353,15 +358,21 @@ namespace aspect
                                         :
                                         1;
 
+                  // pre-compute the tensor contractions
+                  std::vector<SymmetricTensor<2,dim> > tensor_contraction(stokes_dofs_per_cell);
```
bangerth: same here

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -148,6 +148,11 @@ namespace aspect
...
                                    1;
 
+              // pre-compute the tensor contractions
+              std::vector<SymmetricTensor<2,dim> > tensor_contraction(stokes_dofs_per_cell);
+              for (unsigned int i = 0; i < stokes_dofs_per_cell; ++i)
+                tensor_contraction[i] = (viscosity_derivative_wrt_strain_rate * scratch.grads_phi_u[i]) * strain_rate;
+
               // symmetrize when the stabilization is symmetric or SPD
               if ((preconditioner_stabilization & Newton::Parameters::Stabilization::symmetric) != Newton::Parameters::Stabilization::none)
                 {
```
bangerth: don't you actually only need these tensor contraction terms if you go into this `if` block here? If so, maybe move the declaration and initialization above after the `if` here.

source/simulator/assemblers/newton_stokes.cc
MFraters: I agree we need a better name, but we do not just use it for j, but also for i. So something like `viscosity_derivative_wrt_strain_rate_x_grads_phi_u_x_strain_rate` or `eta_derivative_wrt_eps_x_grads_phi_u_x_eps` or `eta_derivative_wrt_eps_times_grads_phi_u_times_eps` or just `deta_deps_times_eps_times_phi`

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -353,15 +358,21 @@ namespace aspect
...
                           {
-                            data.local_matrix(i,j) += ( derivative_scaling_factor * alpha * (scratch.grads_phi_u[i] * (viscosity_derivative_wrt_strain_rate * scratch.grads_phi_u[j]) * strain_rate
-                                                                                             + scratch.grads_phi_u[j] * (viscosity_derivative_wrt_strain_rate * scratch.grads_phi_u[i]) * strain_rate)
-                                                        + derivative_scaling_factor * this->get_pressure_scaling() * scratch.grads_phi_u[i] * 2.0 * viscosity_derivative_wrt_pressure * scratch.phi_p[j] * strain_rate )
+                            data.local_matrix(i,j) += ( derivative_scaling_factor * alpha * (scratch.grads_phi_u[i] * deta_deps_times_eps_times_phi[j]
+                                                                                             + scratch.grads_phi_u[j] * deta_deps_times_eps_times_phi[i])
+                                                        + derivative_scaling_factor * pressure_scaling * 2.0 * viscosity_derivative_wrt_pressure * scratch.grads_phi_u[i] * scratch.phi_p[j] * strain_rate )
```
MFraters: since it still fails, I will first try a commit where I just change the last few `this->get_pressure_scaling()` into `pressure_scaling` and reoder this line. If that works, I we can be sure the problem is related to the tensor contraction (`deta_deps_times_eps_times_phi`).

bangerth: good plan

MFraters:
This seems to pass as well, so the problem really is with the contraction somehow.

PR #2467, opened 2018-06-26
[bugfix] fix additional material model outputs on face assemblers

jdannberg:
In the assembly of the advection system for boundary faces, where we attach the additional outputs to the material model outputs, we used to attach the additional outputs to the "normal" material model outputs instead of the face material model outputs. I don't think anyone noticed this, probably because so far we haven't used additional outputs with the face assemblers. 

This pull request fixes this.


bangerth:
A test would have been nice, but I'm pretty sure it's correct as is.

PR #2473, opened 2018-06-26
switching output of solcx, solkz to MKS again

cedrict:
As they are now the two parameter files for SOLCX and SOLKZ generate velocity outputs in m/year which is a) nonsensical b) different than what is shown in the manual.


bangerth:
Jenkins seems to have lost the connection. Let's merge since the other tester succeeded.

PR #2474, opened 2018-06-26
unit-tests: add example that catches exception

tjhei:


unit_tests/first.cc
```diff
@@ -36,3 +36,16 @@ TEST_CASE("floating point std::vector checks")
 
   compare_vectors_approx(computed, expected);
 }
+
+TEST_CASE("Utilities::read_and_distribute_file_content")
+{
+  std::string content =
```
bangerth: make it `const`

PR #2400, opened 2018-06-23
Perplex live

bobmyhill:
This is a first attempt at allowing ASPECT to evaluate thermodynamic properties (potentially including phase proportions) on-the-fly using PerpleX. 

It is, of course, very slow; a second step will be to cache P-T-X-properties, probably using a KD tree (http://www.dealii.org/developer/doxygen/deal.II/classKDTree.html). 

This PR opened for code advice, and where to place parameter files / material models etc. 

(PerpleX should not be distributed with ASPECT, and most users of ASPECT will not have the PerpleX library installed.)

PR #2478, opened 2018-06-26
make perplex cmake search quiet

tjhei:
otherwise I am getting
```
CMake Warning at CMakeLists.txt:336 (FIND_PACKAGE):
Could not find a package configuration file provided by "PerpleX" with
any
of the following names:
```

related to #2400

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -333,7 +333,7 @@ ENDIF()
 
 DEAL_II_INVOKE_AUTOPILOT()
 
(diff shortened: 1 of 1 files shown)


bobmyhill:
@MFraters If the installer is run, aspect's CMakeLists.txt is configured to find the shared library. If aspect has already been made before installation, PerpleX won't be used.

tjhei:
@MFraters also check https://github.com/geodynamics/aspect/blob/master/contrib/perplex/README.md

Issue #1494, opened 2017-05-07
The Manual's Appendix numbers are overwriting the section titles


egpuckett:
For appendices numbered >= 100 the table of contents is not working correctly, at least with my LaTeX compiler on Ubuntu 16.04:

> A.100Parameters in section Mesh refinement/Boundary . . . . . . . . . . . . . . . . . . . . . . . . 348

See the following PDF file:

[Manual_Appendix.pdf](https://github.com/geodynamics/aspect/files/981910/Manual_Appendix.pdf)

bangerth:
Correct. We need to figure out how to teach `\tableofcontents` how to properly size the first column.

egpuckett:
@bangerth I just noticed that some of the titles are simply too long; e.g., 

> A.14 Parameters in section Boundary temperature model/Ascii data model . . . . . . . . . . . 236
A.15 Parameters in section Boundary temperature model/Box . . . . . . . . . . . . . . . . . . . . 238
A.16 Parameters in section Boundary temperature model/Box with lithosphere boundary indicators238

Is it reasonable to make 

> Box with lithosphere boundary indicators

a subsection of

> A.16 Parameters in section Boundary temperature model

Otherwise,  perhaps there is a way to make A 16 occupy two lines?

bangerth:
That may be a separate problem. But we also can't do as you suggest -- these titles are automatically generated from the parameter handler object that manages input files. We will just need to find a way to teach latex to deal with this.

egpuckett:
I'm willing to work on in when I'm too tired to do 'real' work, but I'll have to have you give me a quick overview of how the titles are automatically generated from the parameter handler object that manages input files.

bangerth:
There's nothing to it. Just look at the file `doc/manual/parameters.tex` to see how they look. How they are generated is not actually important -- some class in deal.II does it. The only relevant thing is that given how they are, how can we trick latex into making them look nice.

tjhei:
here is how to do it: https://tex.stackexchange.com/a/33842

egpuckett:
I just made a PR to fix this problem, but some lines are still too long; e.g., see A.10.  Should I close this issue and open another?

bangerth:
Yes, if there is a secondary issue, then open a second github issue. We try not to hijack issues for follow-up discussions.

You can always reference the current issue in the new one you open.

PR #2484, opened 2018-06-26
Aspect Manual Table of Contents

egpuckett:
This resolves (i.e., fixes) Issue #1494 
 
> The Manual's Appendix numbers are overwriting the section titles #1494 

The one line of code

    `\renewcommand{\numberline}[1]{#1~}`

adds a space between the Appendix number (e.g., A.101) and the Appendix title.

There is still one problem with the table of contents; namely, some long titles are not being broken into multiple lines.  For example, see 
    
    > A.10 Parameters in section Boundary composition model/Box with lithosphere ... 271


bangerth:
I verified that it works.

PR #2485, opened 2018-06-26
Fixup style and messages for #2405

gassmoeller:
This was necessary due an overeager merge of #2405 by myself. Thanks to @jdannberg for helping with the changes.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -107,8 +107,8 @@ namespace aspect
      *   to appear in the map. The order of these keys determines the order
      *   of values that are returned by this function.
      * @param[in] allow_background_field If true, expect N+1 values and allow
-     *   setting of the background using "background" or "bg".
-     * @param[in] field_name A name that identifies the type of information
+     *   setting of the background using the key "background".
+     * @param[in] property_name A name that identifies the type of property
      *   that is being parsed by this function and that is used in generating
      *   error messages if the map does not conform to the expected format.
      *
@@ -118,8 +118,8 @@ namespace aspect
     std::vector<double>
     parse_map_to_double_array (const std::string &key_value_map,
                                const std::vector<std::string> &list_of_keys,
-                               const bool allow_background_field,
-                               const std::string &field_name);
+                               const bool has_background_field,
+                               const std::string &property_name);
 
 
 
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -98,8 +98,7 @@ namespace aspect
      * - Three special keys are recognized:
      *      all --> Assign the associated value to all fields.
      *              Only one value is allowed in this case.
-     *      background (or bg) --> Assign associated value to
-     *                             the background.
+     *      background --> Assign associated value to the background.
      *
      * @param[in] key_value_map The string representation of the map
      *   to be parsed.
--- a/source/material_model/multicomponent.cc
+++ b/source/material_model/multicomponent.cc
@@ -218,7 +218,7 @@ namespace aspect
             AssertThrow(false, ExcMessage("Not a valid viscosity averaging scheme"));
 
           // Establish that a background field is required here
-          bool allow_background_field = true;
+          const bool has_background_field = true;
 
           // Retrieve the list of composition names
           const std::vector<std::string> list_of_composition_names = this->introspection().get_composition_names();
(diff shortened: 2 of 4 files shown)


tjhei:
doesn't compile, though.

bangerth:
Oh, it fails a test...

bangerth:
```
Utilities::parse_map_to_double_array FAIL ON PURPOSE
-------------------------------------------------------------------------------
/home/dealii/source/unit_tests/parse_map_to_double_array.cc:91
...............................................................................

/home/dealii/source/unit_tests/parse_map_to_double_array.cc:91: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  check fail 1: 
  check fail 2: 
  check fail 3: 
  check fail 4: 
  SIGSEGV - Segmentation violation signal
```

pmbremner:
That's strange. Not sure why this one failed. Is there something in this that's not compatible with dealii 9.0?

bangerth:
I don't know. It's a segmentation fault, which is generally pretty simple to find. Do you know how to run this in a debugger?

bangerth:
There's also a merge conflict, I think with #2493. Sorry about that.

tjhei:
this has nothing to do with deal 9.0 and with @pmbremner 's work. I assume @gassmoeller made a replacement error here.

Sadly, the exception triggered is not printed right now. This seems to be a bug in catch that I can reproduce but don't know how to fix yet.

gassmoeller:
Alright, this should be good to go. I rewrote the function from #2405, but kept the large-scale structure intact. @pmbremner: This is not meant to diminish your work, the function was absolutely correct, I just tried to improve the readability. In general, if you read the new code, you will find that I use less comments, because the code is more self-explanatory. I removed types from variable names and replaced them by the meaning of the variables (e.g. `split_parts` -> `key_and_value`). Additionally, I do not longer handle the background field as a special case. Instead I add it in the beginning, just like any other field, and then I could remove a lot of special checks for the background field. Thanks to your tests refactoring the function was relatively simple.

This should be ready for a review or merge.

pmbremner:
The rewrite looks good. Two questions though, can "bg" still be used for the background? And, it wasn't entirely clear to me how the new portion ensures that all fields are set to a value that are supposed to be set. Before it was a final loop to check that all values were really changed from NaN, how does it work now? I see the return values are initialized as NaN's, but it looked like the loop doing the final check was removed.

gassmoeller:
> Two questions though, can "bg" still be used for the background?

Yes, look at line 106. If it finds "bg" it replaces it by "background", then continue as usual. Question back: Is this necessary for backwards compatibility? Or is it just a cool feature? In my opinion we can also simply require that everyone names their background field "background", the only reason to keep this is, if there are already input files that rely on "bg".

> And,  it wasn't entirely clear to me how the new portion ensures that all  fields are set to a value that are supposed to be set. Before it was a  final loop to check that all values were really changed from NaN, how  does it work now? I see the return values are initialized as NaN's, but  it looked like the loop doing the final check was removed.

That is correct,  I removed the entire loop. I think this is possible, and makes it much easier to read. In fact I believe we guarantee to fill all values through the other checks that exist. In particular we check that:
- the list of key-value pairs is n_fields longs
- each key-value pair has a valid field name
- each key value is unique (by checking if its value has already been set)

This should guarantee that n_fields unique values are set. Does that make sense?

pmbremner:
> Yes, look at line 106. If it finds "bg" it replaces it by "background", then continue as usual. Question back: Is this necessary for backwards compatibility? Or is it just a cool feature? In my opinion we can also simply require that everyone names their background field "background", the only reason to keep this is, if there are already input files that rely on "bg".

Yes, I see it now. I don't believe "bg" has a history beyond what I did starting with work on the multicomponent material model last year. It is just a cool feature and can probably be removed if there is no desire to have it. I am fine with just simply requiring "background".

> This should guarantee that n_fields unique values are set. Does that make sense?

Yes, it does make sense, and I think you're correct.

One thing that could be improved now, though, is the error message starting at line 145. If the background field label is wrong, repeated, or missing, an exception will be thrown, which is the correct behavior. The message directs the user to check the list of field names, for example, given in the Compositional fields subsection. However, "background" won't be in that list. It would be helpful to the user to have something like the phrase "plus a background field when required" included in the message.

Otherwise, the changes retain the functionality and improve the readability of what was there. Looks good.

gassmoeller:
Ok, I addressed your comments by extending the error message, and removed the "bg" option.

pmbremner:
I don't have an official approval icon, but I approve these changes :)

gassmoeller:
Very nice! So I have one approval, but I need at least one from a person with write access. Would anyone be willing to press the merge button?

gassmoeller:
Thanks!

PR #2497, opened 2018-06-27
Fix viscoelasticity

naliboff:
This fixes an issue that was introduced in #2356. We thought it was fixed by #2495, but unfortunately it did resolve the issue. 
For safety, I've reverted back to the state in `viscoelastic.cc` and `viscoelastic.h` before #2356. I've checked that the benchmarks now run correctly with the changes in this branch.

At next chance, I'll add a series of tests that should have caught this issue in the first place and use the new `compute_volume_fractions` function.

@gassmoeller - This follows our discussion this evening. Lesson learned about not having a test for the  all the available options.

### DIFF
--- a/include/aspect/material_model/viscoelastic.h
+++ b/include/aspect/material_model/viscoelastic.h
@@ -226,6 +226,14 @@ namespace aspect
 
       private:
         /**
+         * The first 3 (2D) or 6 (3D) compositional fields are assumed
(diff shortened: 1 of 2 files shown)


naliboff:
Please don't merge this yet, I'm double checking to make sure  #2495 did not fix the issue.

naliboff:
#2495 did in fact fix the issue. Covered in glory, I am closing this PR.

PR #2498, opened 2018-06-27
[WIP] Sinker benchmark

gassmoeller:
This is a preliminary implementation of the NSinker benchmark as used in Rudi et al. 2017 (based on May et al 2015). I do not know if I find the time to finish the test during the hackathon, this is mostly to allow @tjhei to take a look.


tjhei:
I just looked at this PR again and I think we can merge this as is. @gassmoeller what do you think?

I would like to have this in mainline soon, so we can further test the solver...

gassmoeller:
The manual text was outdated and partly wrong (I never found the time to finish it), so I removed it for now from the PR (I will open another one for the text, I still have it in a local branch). The remaining part of this PR should be ready once the tester has finished.

benchmarks/nsinker/nsinker.cc
```diff
@@ -0,0 +1,291 @@
...
+      // Parameters for Sinker example
+      double delta = 200.0;
+      double omega = 0.1;
+      double chi = 1.0;
+
+      // Centers for the sinkers provided by Cedric Thielot (pers. comm. from Dave May)
+      std::vector<Point<dim> > centers(76);
+      if (dim == 3)
+        {
+          centers[1] = Point<3>(2.4257829890e-01, 1.3469574514e-02, 3.8313885004e-01);
```
tjhei: why do you not fill [0] and not use [0]?

benchmarks/nsinker/nsinker.cc
```diff
@@ -0,0 +1,291 @@
...
+      const unsigned dim = 3;
+
+      // Parameters for Sinker example
+      double delta = 200.0;
+      double omega = 0.1;
+      double chi = 1.0;
+
+      // Centers for the sinkers provided by Cedric Thielot (pers. comm. from Dave May)
+      std::vector<Point<dim> > centers(76);
+      if (dim == 3)
```
tjhei: this is a noop, right?

tjhei:
Looks good otherwise.

tjhei:
and thank you for looking at it again!

gassmoeller:
> why do you not fill [0] and not use [0]?

Because I got it numbered like this and was lazy. I fixed it now :smile:.

Addressed your comments.

PR #2499, opened 2018-06-27
Ice cookbook for 2D viscoplasticity 

jonpkay:
This is a new cook book for ice plasticity with a fixed conductivity and a high viscosity contrast. This cookbook also includes the new heat flux boundary condition.

### DIFF
--- a/cookbooks/simple_ice_box.prm
+++ b/cookbooks/simple_ice_box.prm
@@ -1,0 +1,177 @@
+# A simple setup for convection in a quarter of a 2d shell. See the
+# manual for more information.
+
+
+set Dimension                              = 2
+set Use years in output instead of seconds = true
+set End time                               = 1e6
+set Output directory                       = output-ice-box_2d
+set Nonlinear solver scheme                = single Advection, iterated Stokes 
+set Nonlinear solver tolerance             = 1e-5
+set Max nonlinear iterations               = 10
+set CFL number                             = 0.5
+set Pressure normalization                 = no
+set Use conduction timestep                = true
+# Material model
+# Rheology: Non-linear viscous flow and Drucker Prager Plasticity
+# Values for most rheological parameters are specified for a background material and
+# each compositional field.  Values for viscous deformation are based on dislocation
+# creep flow-laws, with distinct values for the upper crust (wet quartzite), lower
+# crust (wet anorthite) and mantle (dry olivine).  Table 1 of Naliboff and Buiter (2015),
+# Earth Planet. Sci. Lett., v.421, p. 58-67 contains values for each of these flow laws.
+subsection Material model
+  set Model name = visco plastic
+
+  subsection Visco Plastic
+
+    # Reference temperature and viscosity
+    set Reference temperature = 273 
+    set Reference viscosity = 1e11
+
+    # The minimum strain-rate helps limit large viscosities values that arise
+    # as the strain-rate approaches zero.
+    # The reference strain-rate is used on the first non-linear iteration
+    # of the first time step when the velocity has not been determined yet.
+    set Minimum strain rate = 1.e-20
+    set Reference strain rate = 1.e-12
+
+    # Limit the viscosity with minimum and maximum values
+    set Minimum viscosity = 1e10
+    set Maximum viscosity = 1e22
+
+    # Thermal diffusivity is adjusted to match thermal conductivities
+    # assumed in assigning the initial geotherm
+    set Thermal diffusivities =        2.6e-6
+    set Heat capacities       =        1000
+    set Densities             =        930
+    set Thermal expansivities =        0     
+
+    # Harmonic viscosity averaging
+    set Viscosity averaging scheme = harmonic
+
+    # Choose to have the viscosity (pre-yield) follow a dislocation
+    # diffusion or composite flow law.  Here, dislocation is selected
+    # so no need to specify diffusion creep parameters below, which are
+    # only used if "diffusion" or "composite" option is selected.
+    set Viscous flow law = dislocation
+
+    # Dislocation creep parameters for
+    # Note that the viscous pre-factors below are scaled to plane strain from unixial strain experiments.
+    set Prefactors for dislocation creep          =   1.667e-7  
+    set Stress exponents for dislocation creep    =      4.8
+    set Activation energies for dislocation creep =    60e+3
+    set Activation volumes for dislocation creep  =        0
+
+    # Plasticity parameters
+    set Angles of internal friction =   30.
+    set Cohesions                   =  1e7     
+
+  end
+end
+
+
+subsection Geometry model
+  set Model name = box
+
+  subsection Box
+    set X repetitions = 100
+    set Y repetitions = 25
+    set X extent = 100e3
+    set Y extent = 25e3
+  end
+end
+
+subsection Free surface
+  set Free surface boundary indicators        = top
+  set Surface velocity projection = vertical
+end 
+
+# Velocity on boundaries characterized by functions
+# The outward velocity (x-direction) on the left and right walls is 1.25 cm/year
+# The vertical velocity at the base is 0.125 cm/year (balances outflow on sides)
+# Velocity components parallel to the base (x-velocity) and side walls (y-velocity)
+# are unconstrained (i.e. 'free').
+subsection Boundary velocity model
+  set Prescribed velocity boundary indicators = left x: function, right x:function, bottom:function
+
+  subsection Function
+    set Variable names      = x,y
+    set Function constants  = cm=0.01, year=1
+    set Function expression = if (x<50e3 ,     0*cm/year,    0*cm/year) ; 0          
+  end
+end
+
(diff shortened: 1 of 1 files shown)


jdannberg:
Thanks for contributing this!
It would be great if you could also write a section in the manual about this, like we do for our other cookbooks. In the manual, we have a section with some guidelines on how to write a cookbook (Section 6.3 How to write a cookbook). If you've already made some figures or written some text for the report you can use that as a starting point.

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
+# A simple setup for using the boundary heat flux in a ice viscoplatic model.
```
naliboff: I also find it helpful to add a brief statement regarding what previous studies (if any) the cookbook is based on and a broad overview of the relevant application (e.g., icy planet, etc).

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+
+
+set Dimension                              = 2
+set Use years in output instead of seconds = true
+set End time                               = 1e6
+set Output directory                       = output-ice-box_2d
+set Nonlinear solver scheme                = single Advection, iterated Stokes 
+set Nonlinear solver tolerance             = 1e-5
+set Max nonlinear iterations               = 10
```
naliboff: Typically we set this value to a much higher number (50 or 100). In general, we never want to reach this limit (e.g., the model should always converge). Often, the limit is only used to ensure the code is not stuck doing an endless number of non-linear iterations. 

For now, I suggest changing this value to 50 and see if that is a sufficient number of non-linear iterations to achieve convergence. If not, I think this would be a great application for the Newton solver, which is enabled with `set Nonlinear solver scheme   = iterated Advection, Newton Stokes`

naliboff: We can talk more about the different nonlinear solver schemes later on.

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+set CFL number                             = 0.5
+set Pressure normalization                 = no
+set Use conduction timestep                = true
+# Material model
```
naliboff: Add a space above this line.

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+# Rheology: Non-linear viscous flow and Drucker Prager Plasticity
+# Values for most rheological parameters are specified for a background material and
+# each compositional field.  Values for viscous deformation are based on diffusion creep
```
naliboff: Should `diffusion creep` be changed to `dislocation creep` here?

naliboff: It may also be helpful to say how you scaled the prefactor value from unixial to trixial stress (or at least say this was done).

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+# each compositional field.  Values for viscous deformation are based on diffusion creep 
+# based on Dombard and McKinnon 2006. 
+subsection Material model
+  set Model name = visco plastic
+
+  subsection Visco Plastic
+
+    # Reference temperature and viscosity
+    set Reference temperature = 273 
+    set Reference viscosity = 1e11
```
naliboff: Just out of curiosity, what are the 'average' viscosities throughout the model simulation run. If they are significantly different from 1e11, changing this value may significantly reduce the Linear solver time.

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+    # The minimum strain-rate helps limit large viscosities values that arise
+    # as the strain-rate approaches zero.
+    # The reference strain-rate is used on the first non-linear iteration
+    # of the first time step when the velocity has not been determined yet.
+    set Minimum strain rate = 1.e-20
+    set Reference strain rate = 1.e-12
+
+    # Limit the viscosity with minimum and maximum values
+    set Minimum viscosity = 1e10
+    set Maximum viscosity = 1e22
```
naliboff: This is a fairly large range. Typically, we restrict the range to 6 or 8 orders of magnitude. Are there cases where the minimum and maximum values are actually present at a given point in time? If so, that speaks quite well to the robustness of the solvers 😄

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+    # Dislocation creep parameters for
+    # Note that the viscous pre-factors below are scaled to plane strain from unixial strain experiments.
+    set Prefactors for dislocation creep          =   1.667e-7  
+    set Stress exponents for dislocation creep    =      4.8
+    set Activation energies for dislocation creep =    60e+3
+    set Activation volumes for dislocation creep  =        0
+
+    # Plasticity parameters
+    set Angles of internal friction =   30.
+    set Cohesions                   =  1e7
```
naliboff: Typically, the nonlinear problem becomes much harder to solver as the Angle of internal friction increases. If it is not unreasonable to use a lower value for ice, it may be worth reducing this value.

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+  set Free surface boundary indicators        = top
+  set Surface velocity projection = vertical
+end 
+
+# Velocity on boundaries characterized by functions
+# The outward velocity (x-direction) on the left and right walls is 1.25 cm/year
+# The vertical velocity at the base is 0.125 cm/year (balances outflow on sides)
+# Velocity components parallel to the base (x-velocity) and side walls (y-velocity)
+# are unconstrained (i.e. 'free').
+subsection Boundary velocity model
```
naliboff: Update these comments to reflect that the imposed velocity is now zero. When the values are changed to include shortening, then the comments can be updated again.

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+    set Function expression = if (x<50e3 ,     0*cm/year,    0*cm/year) ; 0          
+  end
+end
+
+subsection Boundary temperature model
+  set Fixed temperature boundary indicators   = top
+  set List of model names = initial temperature
+end
+
+#This module introduces the heat flux boundary condition
```
naliboff: Can you add a brief description of what the negative heat flux implies? As I understand it, this means heat is flowing out of the system through the base?

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+subsection Boundary heat flux model
+  set Fixed heat flux boundary indicators = bottom
+  set Model name = function
+
+  subsection Function
+    set Function expression = -0.04
+  end
+end
+
+#A variable surface tempreature is prescribed
```
naliboff: Perhaps add a note indicating that this variable surface temperature produces a lateral viscosity contrast and helps initiate deformation through convection.

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+subsection Boundary temperature model
+  set Fixed temperature boundary indicators = top
+  set List of model names = function
+  subsection Function
+  set Variable names = x,y
+#    set Top temperature =  90
+   set Function expression = 60+30*sin(asin((x/100000)))^.25
+  end
```
naliboff: I think you can remove the lines above (initial temperature is set below) and use  'set List of model names = initial temperature'.

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+end
+
+subsection Solver parameters
+  subsection Stokes solver parameters
+    set Linear solver tolerance = 1e-8
+    end
+end
+
+
+# Gravity model
```
naliboff: Perhaps add a comment here indicating what this gravity value is based on (e.g., which planet)?

cookbooks/simple_ice_box.prm
```diff
@@ -0,0 +1,174 @@
...
+subsection Gravity model
+  set Model name = vertical
+
+  subsection Vertical
+    set Magnitude = 1.42
+  end
+end
+
+
+subsection Mesh refinement
```
naliboff: I suggest moving this subsection to just below the `Geometry model` subsection.

jonpkay:
Ok great, I will get to these shortly. Thanks.

cookbooks/simple_ice_box.prm
jonpkay: Average was 1e13 so changing to that.

cookbooks/simple_ice_box.prm
jonpkay: I need to figure out how to get the viscosity lower if I want there to be some overturn, at least that is what the analytical papers are suggesting.

naliboff:
@jonpkay - We are currently cleaning up stuff for the next Hackathon. If you would like to continue working on this PR, please go ahead and re-open!

PR #2501, opened 2018-06-27
A couple of minimalist changes.

bangerth:


bangerth:
I was briefly tempted to split this up into two commits, but then thought that that may just be too much ;-)

PR #2502, opened 2018-06-27
Move the check for reasonable parameters earlier in the function.

bangerth:
We should check first whether input parameters upon restart are reasonable before
we start to read the rest of the simulation data. That's because the rest may not make
sense to us if the parameters are non-sensical. 

In other words: Try to give error messages as early as possible.

This is a follow-up to #2310, in pursuance of #2090.


gassmoeller:
Makes sense. Ready to merge if tester is happy.

Issue #2491, opened 2018-06-27
TABLE OF CONTENTS LINE NUMBERS ARE TOO LONG


egpuckett:
There is a one problem with the table of contents; namely, some long titles are not being broken into multiple lines and hence the page number is pushed into the margin. This either primarily or entirely in the page numbers for entries in the Appendix.

This is related to closed issue #1494.

PR #2508, opened 2018-06-27
Aspect manual TOC

egpuckett:
This PR addresses Issue #2491

The solution was to adjust two internal LaTeX parameters that set the width of the right-hand TOC margin and the size of the box that contains the page numbers, so that now the page numbers in the TOC are aligned in a column.  There is addition documentation in a comment at the top of manual.tex.

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -11,6 +11,15 @@
 
 \renewcommand{\numberline}[1]{#1~}
 
+% If your page numbers stick out into the righthand margin but using lengths 
+% appropriate to your document.
+
+\makeatletter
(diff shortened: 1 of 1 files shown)


doc/manual/manual.tex
```diff
@@ -11,6 +11,16 @@
 
 \renewcommand{\numberline}[1]{#1~}
 
+% If your page numbers stick out into the righthand margin but using lengths 
+% appropriate to your document. See the Introduction to the "The tocloft 
+% package" for additional information.
+
+\makeatletter
```
bangerth: One of my favorite commands ever! :-)

PR #2509, opened 2018-06-27
Initial temperature condition - patch on S40RTS

SophieCoulson:
Add initial temperature condition patch_on_S40RTS which combines S40RTS and an ascii grid containing an upper mantle tomography. Data is input in Vs and converted to temperature within the function (using the same method as S40RTS_perturbation). Includes an option to smooth by averaging between the two tomographies. Also added test for this function.

Changes to S40RTS_perturbation.h allow calling all functions from S40RTS_perturbation.
Default data added into the data directory - upper_shell_3d.txt


SophieCoulson:
Have fixed the outputs for my test

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,204 @@
...
+    PatchOnS40RTS<dim>::parse_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Initial temperature model");
+      {
+        Utilities::AsciiDataBase<dim>::parse_parameters(prm);
+
+        max_grid_depth   = prm.get_double ("Maximum grid depth");
+        smoothing_length_scale   = prm.get_double ("Smoothing length scale");
+
+        AssertThrow (max_grid_depth != 0,
```
anne-glerum: What is the reason for max_grid_depth not being allowed to be zero? Is it allowed to be negative?
Instead of testing for not being zero, maybe test that the max_grid_depth is bigger than the smoothing length?

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,204 @@
...
+    double
+    PatchOnS40RTS<dim>::
+    ascii_grid_vs (const Point<dim> &position) const
+    {
+      const double vs_perturbation = Utilities::AsciiDataInitial<dim>::get_data_component(position,0);
+      return vs_perturbation;
+    }
+
+
+    //read in ascii grid vs data above max_grid_depth and S40RTS vs data below max_grid_depth
```
anne-glerum: Please move this comment to the header file as it's useful info for the user.

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,204 @@
...
+    {
+      Utilities::AsciiDataInitial<dim>::initialize(1);
+      if (s40rts.vs_to_density_method == s40rts.file)
+        {
+          s40rts.profile.initialize(this->get_mpi_communicator());
+          s40rts.vs_to_density_index = s40rts.profile.get_column_index_from_name("vs_to_density");
+        }
+    }
+
+    //read in and return vs perturbation from ascii grid
```
anne-glerum: Usually we don't add comments above the function in the source file. This info is already given in the header file for the user, and the function itself is very clear for the programmer, so here it's not really needed. In case you want to keep it, move it to line 54.

source/initial_temperature/patch_on_S40RTS.cc
SophieCoulson: Here I was trying to make it so the user has to declare a maximum grid depth - I thought it didn't make that much sense to have a default when I have no idea what size grids people will use. Perhaps there is a better way to do this?

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,204 @@
...
+      if (this->get_geometry_model().depth(position) < max_grid_depth - smoothing_length_scale)
+        {
+          vs_perturbation = ascii_grid_vs(position);
+        }
+      //add smoothing between the two models
+      else if (this->get_geometry_model().depth(position) > max_grid_depth - smoothing_length_scale && this->get_geometry_model().depth(position) < max_grid_depth)
+        {
+          vs_perturbation = (s40rts.get_Vs(position) + ascii_grid_vs(position))/2;
+        }
+
```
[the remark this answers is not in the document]
SophieCoulson: I chatted with Jacky and Paul a bit about this... As we have no measure of the accuracy of the tomographies in this region (i.e. which is most accurate at these specific depths), I'm not sure whether it is actually more physically realistic to have a linear combination. Although, it would probably make the transition smoother. What do you think? Maybe we can chat about this?

SophieCoulson:
Think I've addressed all your comments now!

SophieCoulson:
Hopefully this passes the astyle test now!

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,197 @@
...
+      double vs_perturbation;
+      if (depth <= max_grid_depth - smoothing_length_scale)
+        {
+          vs_perturbation = ascii_grid_vs(position);
+        }
+      //add smoothing between the two models
+      else if (depth > max_grid_depth - smoothing_length_scale && depth < max_grid_depth)
+        {
+          const double scale_factor = (depth-(max_grid_depth-smoothing_length_scale))/smoothing_length_scale;
+          vs_perturbation = s40rts.get_Vs(position)*(scale_factor) + ascii_grid_vs(position)*(1-scale_factor);
```
anne-glerum: Thanks for putting in the smoothing!

Although the integer 1 is promoted to a double here automatically, I usually write these numbers as doubles (1.0) for clarity.

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,197 @@
...
+        }
+      else
+        {
+          vs_perturbation = s40rts.get_Vs(position);
+        }
+
```
anne-glerum: remove one empty line

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,197 @@
...
+      // get the Vs to density conversion
+      double vs_to_density = 0.0;
+      if (s40rts.vs_to_density_method == s40rts.file)
+        vs_to_density = s40rts.profile.get_data_component(Point<1>(depth), s40rts.vs_to_density_index);
+      else if (s40rts.vs_to_density_method == s40rts.constant)
+        vs_to_density = s40rts.vs_to_density_constant;
+      else
+        // we shouldn't get here but instead should already have been
+        // kicked out by the assertion in the parse_parameters()
+        // function
```
anne-glerum: This assert is located in the s40rts parse_parameters function, not below. Could you add that?

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,197 @@
...
+        Assert (false, ExcNotImplemented());
+
+      // scale the perturbation in seismic velocity into a density perturbation
+      // vs_to_density is an input parameter
```
anne-glerum: remove this comment (line 102), `vs_to_density is` not strictly one input parameter, as it can be derived from a profile.

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,197 @@
...
+      const double density_perturbation = vs_to_density * vs_perturbation;
+
+      double temperature_perturbation;
+      if (depth > s40rts.no_perturbation_depth)
```
anne-glerum: I think the `s40rts.no_perturbation_depth` should only work on the s40rts data. Now it could also remove the ascii grid data, which defeats the purpose of adding a upper mantle model. Maybe move this check to line 79 and set the vs_perturbation from s40rts to zero there when above the `s40rts.no_perturbation_depth`.

Then you don't need the conditional statements and just go for `const double temperature_perturbation = ...`

Do you actually need the `s40rts.no_perturbation_depth`? It is not used above `max_grid_depth-smoothing_length_scale` and quite strange if it lies below that depth.

Or do you need it because you would still like to describe the temperature in some upper part of the mantle in a different way than through any seismic velocity conversion (i.e. nor through the upper mantle model nor through s40rts)? In that case, I think it's more clear to add a separate `no_perturbation_depth` for this combined model, and to return the background_temperature for depths smaller than `no_perturbation_depth` early on.

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,197 @@
...
+                                                          "$ASPECT_SOURCE_DIR/data/initial-temperature/patch-on-S40RTS/test/",
+                                                          "upper_shell_3d.txt");
+        prm.declare_entry ("Maximum grid depth", "700000.0",
+                           Patterns::Double (0),
+                           "The maximum depth of the Vs ascii grid. The model will read in  "
+                           "Vs from S40RTS below this depth.");
+        prm.declare_entry ("Smoothing length scale", "200000.0",
+                           Patterns::Double (0),
+                           "The depth range (above maximum grid depth) over which to smooth. "
+                           "Smoothing is done by averaging the values of the ascii grid and S40RTS at each point.");
```
anne-glerum: Please update the statement on averaging.

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,197 @@
...
+// explicit instantiations
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    ASPECT_REGISTER_INITIAL_TEMPERATURE_MODEL(PatchOnS40RTS,
+                                              "patch on S40RTS",
+                                              "Implementation of a model in which the initial "
+                                              "temperature is derived from a file containing data "
+                                              "in ascii format (e.g. a high resolution upper mantle tomography) "
```
anne-glerum: Please add a line about the type of data in the file, e.g. seismic velocity perturbations.

include/aspect/initial_temperature/patch_on_S40RTS.h
```diff
@@ -0,0 +1,115 @@
...
+
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    using namespace dealii;
+
+    /**
+     * A class that implements a prescribed temperature field determined from
+     * a Vs model given by a PatchOnS40RTS input file.
```
anne-glerum: Please add that below a certain depth, the S40RTS model is used. And is `PatchOnS40RTS input file` a search and replace typo?

source/initial_temperature/patch_on_S40RTS.cc
SophieCoulson: Not specifying file/constant would actually already be caught when declaring the parameters as these are the only patterns allowed. I have therefore edited the comment and made the error message more direct for this back up throw.

source/initial_temperature/patch_on_S40RTS.cc
SophieCoulson: So we do actually want the s40rts.no_perturbation_depth to work on the ascii grid data too - the idea is that because the continental lithosphere has really high/low temperature values due to the Vs to temperature scalings we tend to use, we want to have the option to zero out the top 100 or 200 km in order to remove this bias. The upper mantle tomography is still useful as it goes down to 600/700 km.
We are thinking about better ways to handle the continents but this is an okay option for now.

Let me know what you think!

include/aspect/initial_temperature/S40RTS_perturbation.h
```diff
@@ -186,6 +189,7 @@ namespace aspect
          */
         unsigned int vs_to_density_index;
 
+        template <int dim2> friend class PatchOnS40RTS;
```
gassmoeller: Is there a reason this has to be `dim2`? Does `dim` not work?

include/aspect/initial_temperature/patch_on_S40RTS.h
```diff
@@ -0,0 +1,116 @@
...
+        /**
+         * Read the parameters this class declares from the parameter file.
+         */
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+        /**
+        * Allows us to call functions from S40RTS_perturbation.cc.
+        * Declares class.
```
gassmoeller: I would reword this whole comment as:
```
Declare a member variable of type S40RTSPerturbation that allows us to call functions from S40RTS_perturbation.cc.
```

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,206 @@
...
+    template <int dim>
+    void
+    PatchOnS40RTS<dim>::initialize ()
+    {
+      Utilities::AsciiDataInitial<dim>::initialize(1);
+      if (s40rts.vs_to_density_method == s40rts.file)
+        {
+          s40rts.profile.initialize(this->get_mpi_communicator());
+          s40rts.vs_to_density_index = s40rts.profile.get_column_index_from_name("vs_to_density");
+        }
```
gassmoeller: From what I see of S40RTS in the master branch, the code in the `if` statement above should have no effect. The same code is executed in `S40RTSPerturbation::initialize`, which is called by `S40RTSPerturbation::parse_parameters` (weird, but that is what it is at the moment), which is called by `PatchOnS40RTS::parse_parameters`, i.e. before we get to this particular line. In other words: Everything should be ready to go, we just repeat these lines here.

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,206 @@
...
+                                              "Implementation of a model in which the initial "
+                                              "temperature is derived from a file containing shear wave velocity perturbations "
+                                              "in ascii format (e.g. a high resolution upper mantle tomography) "
+                                              "combined with S40RTS. Note the required format of the "
+                                              "input ascii input data: The first lines may contain any number of comments "
+                                              "if they begin with '#', but one of these lines needs to "
+                                              "contain the number of grid points in each dimension as "
+                                              "for example '# POINTS: 3 3 3'. "
+                                              "The order of the data columns has to be "
+                                              " `x', `y', `z', 'Temperature [K]' in a 3d model, which means that "
```
gassmoeller: The 'Temperature' has to be replaced by 'Vs anomaly' (or another suitable name) and the corresponding unit.

gassmoeller:
I also feel like this plugin handles a more general case, namely the depth averaging of two arbitrary plugins. Before implementing more plugins like this one for other combinations of tomography models, we should thinks about a more general plugin (or manager function) that can average arbitrary plugins. But that is something to consider for the future, not for this PR.

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -0,0 +1,206 @@
...
+        // we shouldn't get here but instead should already have been
+        // kicked out when declaring the parameter, as the pattern won't match ("file|constant")
+        Assert(false, ExcMessage("Unknown method for vs to density scaling."));
+
+      // scale the perturbation in seismic velocity into a density perturbation
+      // vs_to_density is read in from input file
+      const double density_perturbation = vs_to_density * vs_perturbation;
+
+      double temperature_perturbation;
+      if (depth > s40rts.no_perturbation_depth)
```
anne-glerum: Unfortunately I can't find my comment about this `if` statement any more. It seems unclear to me that the `s40rts.no_perturbation_depth` should have effect on the perturbation of the patch. I think it should only work on the s40rts perturbation (e.g. in line 79), or be removed completely (as the upper part of s40rts is replaced by the patch anyway), or there should be a separate no_perturbation_depth for the upper mantle model.

include/aspect/initial_temperature/S40RTS_perturbation.h
SophieCoulson: Yes, dim was throwing an error, so Wolfgang had me change it to dim2

source/initial_temperature/patch_on_S40RTS.cc
SophieCoulson: Sorry this is causing confusion. 
I've tested it and the no_perturbation_depth does definitely work on the ascii grid too, we are just saying that any temperature perturbation above this depth (S40RTS or the replaced ascii grid) is zeroed out. 
This is exactly what we want if we are trying to remove the contribution from the density contrasts in the lithosphere (i.e. to zero out the top 100 or 200 km). Although this does zero out some of the new data we have put in with the ascii grid, we still have a lot of it left as this upper mantle tomography goes to a depth of 700 km.
@jaustermann am I explaining this correctly? Do you agree?

I guess what may make this slightly clearer is if we have a no_perturbation_depth parameter for this material model as a whole and declare it in patch_on_S40RTS rather than parsing it from S40RTS_perturbation? Would this help?

jaustermann: I agree @SophieCoulson, I think it is currently doing the right thing but I also agree that this could be confusing. Maybe making no_perturbation_depth a parameter within patch_on_S40RTS rather than S40RTS_perturbation (as you suggest Sophie) is the way to go.

gassmoeller:
Yes please resolve the merge conflict the following way:

- Switch to master: `git checkout master`
- update master: `git pull upstream master` (replace upstream with the name of the git remote that represents the official ASPECT repo)
- switch back to your branch: `git checkout combined-grid-smoothing`
- rebase to master: `git rebase master`

Now you should get a merge conflict. Check ` include/aspect/initial_temperature/S40RTS_perturbation.h ` for a line with `<<<<<` and `>>>>` and fix the part in between to a working state.
- Then do: `git add  include/aspect/initial_temperature/S40RTS_perturbation.h `
- and finish the rebase: `git rebase --continue`
- and upload to github: `git push -f origin combined-grid-smoothing` (replace origin by the name of the git remote that repesents your github repo)

Let me know if you run into problems.

Did you address the other comments?

SophieCoulson:
Okay, I've resolved the merge conflict and think I've addressed all comments. no_perturbation_depth_patch is now a parameter within patch_on_S40RTS rather than S40RTS_perturbation.
Hope everything is good to go now.

gassmoeller:
Looks ready to go. i will let @anne-glerum push the button, it was her initial review :+1:

Issue #2504, opened 2018-06-27
Add section to manual related to thermodynamic self-consistency


bobmyhill:
In ASPECT, material models can be implemented where the material properties are not compatible with each other. For example, one could implement a material model where thermal expansivity is constant, but the density formulation does not satisfy the expression alpha = -1/rho (d rho/d T)|P. There are both implicit and explicit constraints. 

We should add a section to the manual that covers the relations and potential pitfalls in designing material models.

bobmyhill:
I'm working on this now...

bangerth:
See also #2468.

bobmyhill:
Addressed in #2513.

PR #2513, opened 2018-06-27
added thermodynamic self-consistency section to the manual

bobmyhill:
Addresses #2504.

I don't know whether this is sufficient to be added as a contributor to the manual - I'm happy to take my name off if not!

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -165,6 +165,7 @@ COMPUTATIONAL INFRASTRUCTURE FOR GEODYNAMICS (CIG)
     Marine Lasbleis,
     Shangxin Liu,
     Elvira Mulyukova,
+    Bob Myhill,
     John Naliboff,
     Jonathan Perry-Houts,
     Elbridge Gerry Puckett,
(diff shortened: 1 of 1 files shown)


doc/manual/manual.tex
```diff
@@ -10666,6 +10667,71 @@ \subsubsection{Material models}
...
+\paragraph{Isobaric heat capacity}
+
+We start by taking the partial derivative of the isobaric heat capacity~\eqref{eq:mm_isobaric_heat_capacity} with respect to pressure at constant temperature:
+\begin{eqnarray}
+  \pddc{C_p^{*}}{p}{T} &=& -T \pddcbc{\mathcal{G}^{*}}{T}{p} \\
+  &=& -T\pddc{\left(\alpha / \rho \right)}{T}{p} \label{eq:heat_capacity_p_dependence}
+\end{eqnarray}
+From this expression it becomes clear that if $\alpha / \rho$ has any temperature dependence, the heat capacity \emph{cannot} be globally constant. One way to solve this issue is to define heat capacity at constant pressure, and then integrate~\eqref{eq:heat_capacity_p_dependence} with respect to pressure:
+\begin{equation}
+  C_p^{*}(p, T) = C_p^{*}(p_{\textrm{ref}}, T) + \int_{p_{\textrm{ref}}}^p \pddc{C_p^{*}}{p}{T} dp
```
gassmoeller: The equation you reference here (101) is not the one that is written down (100).

gassmoeller: Also, does that expression get any simpler if you actually put it in?

doc/manual/manual.tex
```diff
@@ -10666,6 +10667,71 @@ \subsubsection{Material models}
...
+\end{equation}
+
+\paragraph{Isentropic gradient}
+
+The isentropic gradient is defined as the change in temperature with pressure at constant entropy. Using the cyclic relation, we can define this in terms of partial differentials of the entropy with respect to pressure and temperature:
+\begin{eqnarray}
+  \pddc{T}{p}{S} &=& - \pddc{T}{S}{p} \pddc{S}{p}{T} \\
+  &=& - \left( \frac{T}{C_p^{*}} \right) \left( - \frac{\alpha}{\rho} \right) \\
+  &=& \frac{\alpha T}{\rho C_p^{*}} \label{eq:mm_isentropic_gradient}
+\end{eqnarray}
```
gassmoeller: Can you explain a bit more? I don't see how that poses a constraint on the material properties (as long as you compute the adiabat following this relation)

doc/manual/manual.tex
```diff
@@ -10666,6 +10667,71 @@ \subsubsection{Material models}
...
+
+
+For complex material models, obtaining analytical functions which obey all these relations may be a non-trivial exercise. Indeed, there are many equations of state which exist in the thermodynamic literature which do no obey these relations! 
+
+\emph{A final note of warning: some compressible formulations in \aspect{} (Section~\ref{sec:mass-conservation-approximation}) use the isothermal compressibility, while others use the isentropic compressibility. Fully self-consistent material models must have a switch so that they use the correct compressibility. Conversion between the compressibilities can be achieved using ~\eqref{eq:mm_isentropic_compressibility}.}
```
gassmoeller: --> must either specify what approximation of the compressible equations they are consistent with (see Section~\ref{sec:approximate-equations}), or have a switch so that they use the correct compressibility for each of the different approximations.

bobmyhill:
Thanks @gassmoeller (or @jdannberg!) ... I've made the changes you suggested. 

Sadly, there is no easy way to construct a self-consistent model - I think this is why there are so many inconsistent formulations in the literature. I'm working on a multicomponent compressible model now which is probably one of the simpler examples (and simplifies to an existing model with certain parameter choices).

bobmyhill:
Can we run tests?

PR #2517, opened 2018-06-27
cmake: disallow in-source test runs

tjhei:
Running the whole testsuite in an in-source build has several problems.
Most importantly, the test build and source directory are not separate,
so the project can not be easily reset.
Rather than trying to fix this, disallow running the "all tests". Of
course running our single "quick test" is still okay.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,6 +163,13 @@ ADD_CUSTOM_TARGET(setup_tests
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
   COMMENT "Enabling all tests ...")
 
+# Disable the ability to run all tests in an in-source build:
+IF (("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}") AND ASPECT_RUN_ALL_TESTS)
+  MESSAGE(WARNING "\nEnabling all tests is not supported in in-source builds. Please create a separate build directory!\n")
+  SET(ASPECT_RUN_ALL_TESTS OFF CACHE BOOL "" FORCE)
+ENDIF()
+
+
 SET(ASPECT_TEST_GENERATOR "Unix Makefiles" CACHE STRING 
   "Generator to use for the test cmake project. Using ninja instead of make is not recommended.")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -165,8 +165,8 @@ ADD_CUSTOM_TARGET(setup_tests
 
 # Disable the ability to run all tests in an in-source build:
 IF (("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}") AND ASPECT_RUN_ALL_TESTS)
-  MESSAGE(WARNING "\nEnabling all tests is not supported in in-source builds. Please create a separate build directory!\n")
   SET(ASPECT_RUN_ALL_TESTS OFF CACHE BOOL "" FORCE)
+  MESSAGE(FATAL_ERROR "\nEnabling all tests is not supported in in-source builds. Please create a separate build directory!\n")
 ENDIF()
 
 


CMakeLists.txt
```diff
@@ -163,6 +163,13 @@ ADD_CUSTOM_TARGET(setup_tests
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
   COMMENT "Enabling all tests ...")
 
+# disable the ability to run all tests in an in-source build:
```
bangerth: I'm surprised you start a comment with a lower-case letter :-)

CMakeLists.txt
```diff
@@ -163,6 +163,13 @@ ADD_CUSTOM_TARGET(setup_tests
...
+IF (("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}") AND ASPECT_RUN_ALL_TESTS)
+  MESSAGE(WARNING "\nEnabling all tests is not supported in in-source builds. Please create a separate build directory!\n")
```
bangerth: Just make it FATAL and not worry about disabling the variable in the next line.

bangerth: Where is `ASPECT_RUN_ALL_TESTS` set, by the way? When you do `make setup_tests`, or also otherwise?

doc/manual/manual.tex
```diff
@@ -2756,19 +2756,15 @@ \subsubsection{Obtaining \aspect{} and initial configuration}
...
- cmake -DDEAL_II_DIR=/u/username/deal-installed/ .
+ cmake -DDEAL_II_DIR=/u/username/deal-installed/ ..
 \end{verbatim}
 
-An alternative would be to configure \aspect{} as an out-of-source build. 
-You would need to create a separate build directory and
-specify \aspect{}'s source directory using \texttt{cmake PATH\_TO\_ASPECT\_SOURCE}
-from within the build directory. The
-instructions in the following sections assume an in-source build.
+This will create an ``out-of-source`` build, where the build directory is different than the source directory. While in-source builds (where you run \texttt{cmake .} in your source directory), are supported, we strongly recommend and out-of-source build as described above. Specifically, running the whole test suite (see Section~\ref{sec:running_tests}) is only supported this way.
```
bangerth: -> different FROM the source

bangerth: strongly recommend AND -> we strong recommend AN

doc/manual/manual.tex
```diff
@@ -2780,8 +2776,8 @@ \subsubsection{Compiling \aspect{} and generating documentation}
 \end{verbatim}
 on the command line (or \texttt{make -jN} if you have multiple processors in
 your machine, where \texttt{N} is the number of processors). This builds the
-\aspect{} executable which will reside in
-the main directory and will be named \texttt{./aspect}. If you intend to
+\aspect{} executable which will reside in the \textt{build} directory
+and will be named \texttt{./aspect}. If you intend to
```
bangerth: the `./` is now no longer true. It'll be `./build/aspect`.

Are there other references to `./aspect` in the manual?

tjhei:
as a beginner github user, I accidentally combined two commits. I am pulling the manual update out, but addressed the comments. see #2521

CMakeLists.txt
tjhei: touché!

CMakeLists.txt
tjhei: It s a cmake flag you can set (``cmake -D ASPECT_RUN_ALL_TESTS=ON``) or it is enabled by running ``make setup_tests``. I am not sure failing here would be useful as this will make ``make setup_tests`` fail, but I can change it if you prefer.

bangerth: Yes, please. Silently doing something other than the person intended is not nice :-)

bangerth:
I'm still in favor of a fatal error :-)

tjhei:
okay, it actually works quite nicely if I set it to OFF before failing.

CMakeLists.txt
```diff
@@ -163,6 +163,13 @@ ADD_CUSTOM_TARGET(setup_tests
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
   COMMENT "Enabling all tests ...")
 
+# Disable the ability to run all tests in an in-source build:
+IF (("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}") AND ASPECT_RUN_ALL_TESTS)
+  SET(ASPECT_RUN_ALL_TESTS OFF CACHE BOOL "" FORCE)
+  MESSAGE(FATAL_ERROR "\nEnabling all tests is not supported in in-source builds. Please create a separate build directory!\n")
```
bangerth: That's fine too, but isn't the SET a dead store in that case because we're about to die anyway?

tjhei: No. Otherwise you end up in a state where the variable is set and cmake fails when you run it until you figure out how to fix it (``cmake -D ASPECT_RUN_ALL_TESTS=OFF``).

bangerth: Ah, the crazy cache...

PR #2519, opened 2018-06-27
Add test based on bending beam benchmark

naliboff:
A test to ensure the viscoelastic material model works with models that use a compositional field to modify material properties (e.g., other than fields for viscoelastic stresses).


gassmoeller:
Nice! Ready to merge if the tester is happy.

PR #2520, opened 2018-06-27
Let initial conditions use thermal expansivity from material model

bartniday:
This adds an option so that S40RTS or SAVANI initial conditions can read in thermal expansivity from the material model, instead of inputing a constant value in the parameter file. By default the option is off.

### DIFF
--- a/source/initial_temperature/S40RTS_perturbation.cc
+++ b/source/initial_temperature/S40RTS_perturbation.cc
@@ -22,6 +22,9 @@
 #include <aspect/initial_temperature/S40RTS_perturbation.h>
 #include <aspect/adiabatic_conditions/interface.h>
 #include <aspect/utilities.h>
+#include <aspect/simulator_access.h>
+#include <aspect/initial_composition/interface.h>
+#include <aspect/material_model/interface.h>
 #include <fstream>
 #include <iostream>
 #include <deal.II/base/std_cxx11/array.h>
@@ -337,8 +340,28 @@ namespace aspect
 
       double temperature_perturbation;
       if (depth > no_perturbation_depth)
-        // scale the density perturbation into a temperature perturbation
-        temperature_perturbation =  -1./thermal_alpha * density_perturbation;
+        {
+          // scale the density perturbation into a temperature perturbation
+          // see if we need to ask material model for the thermal expansion coefficient
+          if (use_material_model_thermal_alpha)
+            {
+              MaterialModel::MaterialModelInputs<3> in(1, this->n_compositional_fields());
+              MaterialModel::MaterialModelOutputs<3> out(1, this->n_compositional_fields());
+              in.position[0] = position;
+              in.temperature[0] = background_temperature;
+              in.pressure[0] = this->get_adiabatic_conditions().pressure(position);
+              in.velocity[0] = Tensor<1,3> ();
+              for (unsigned int c=0; c<this->n_compositional_fields(); ++c)
+                in.composition[0][c] = this->get_initial_composition_manager().initial_composition(position, c);
+              in.strain_rate.resize(0);
+
+              this->get_material_model().evaluate(in, out);
+
+              temperature_perturbation = -1./(out.thermal_expansion_coefficients[0]) * density_perturbation;
+            }
+          else
+            temperature_perturbation = -1./thermal_alpha * density_perturbation;
+        }
       else
         // set heterogeneity to zero down to a specified depth
         temperature_perturbation = 0.0;
@@ -356,9 +379,9 @@ namespace aspect
       {
         prm.enter_subsection("S40RTS perturbation");
         {
-          prm.declare_entry("Data directory", "$ASPECT_SOURCE_DIR/data/initial-temperature/S40RTS/",
-                            Patterns::DirectoryName (),
-                            "The path to the model data. ");
+          prm.declare_entry ("Data directory", "$ASPECT_SOURCE_DIR/data/initial-temperature/S40RTS/",
+                             Patterns::DirectoryName (),
+                             "The path to the model data. ");
           prm.declare_entry ("Initial condition file name", "S40RTS.sph",
                              Patterns::Anything(),
                              "The file name of the spherical harmonics coefficients "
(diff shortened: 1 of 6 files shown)


bangerth:
@jaustermann -- feel like reviewing this?

jaustermann:
Yes! I don't think I have /run-tests permission so if you could start the tests I'll take a look at the code.

source/initial_temperature/S40RTS_perturbation.cc
```diff
@@ -337,8 +340,28 @@ namespace aspect
...
+                in.composition[0][c] = this->get_initial_composition_manager().initial_composition(position, c);
+              in.strain_rate.resize(0);
+
+              this->get_material_model().evaluate(in, out);
+
+              temperature_perturbation = -1./(out.thermal_expansion_coefficients[0]) * density_perturbation;
+            }
+          else
+            // scale the density perturbation into a temperature perturbation
+            temperature_perturbation =  -1./thermal_alpha * density_perturbation;
```
jaustermann: could you make the spaces around the = sign consistent? This is re: 
in.pressure[0]=this-...
in.velocity[0]= Te...
temperature_perturbation =  -1./th...

source/initial_temperature/SAVANI_perturbation.cc
```diff
@@ -297,8 +300,28 @@ namespace aspect
...
+                in.composition[0][c] = this->get_initial_composition_manager().initial_composition(position, c);
+              in.strain_rate.resize(0);
+
+              this->get_material_model().evaluate(in, out);
+
+              temperature_perturbation = -1./(out.thermal_expansion_coefficients[0]) * density_perturbation;
+            }
+          else
+            // scale the density perturbation into a temperature perturbation
+            temperature_perturbation =  -1./thermal_alpha * density_perturbation;
```
jaustermann: same comment on spacing around = sign

source/initial_temperature/S40RTS_perturbation.cc
```diff
@@ -380,6 +403,11 @@ namespace aspect
                              Patterns::Double (0),
                              "The value of the thermal expansion coefficient $\\beta$. "
                              "Units: $1/K$.");
+          prm.declare_entry ("Use thermal expansion coefficient from material model", "false",
+                             Patterns::Bool (),
+                             "Option to use the material model to determine the thermal expansion "
+                             "coefficient, instead of the thermal expansion coefficient specified in "
+                             "this model.");
```
jaustermann: Maybe this is a bit clearer: "Option to take the thermal expansion coefficient from the material model instead of from what is specified in this section."

source/initial_temperature/SAVANI_perturbation.cc
```diff
@@ -340,6 +363,11 @@ namespace aspect
                              Patterns::Double (0),
                              "The value of the thermal expansion coefficient $\\beta$. "
                              "Units: $1/K$.");
+          prm.declare_entry ("Use thermal expansion coefficient from material model", "false",
+                             Patterns::Bool (),
+                             "Option to use the material model to determine the thermal expansion "
+                             "coefficient, instead of the thermal expansion coefficient specified in "
+                             "this model.");
```
jaustermann: same comment as on S40 file.

source/initial_temperature/S40RTS_perturbation.cc
```diff
@@ -337,8 +340,28 @@ namespace aspect
...
+              for (unsigned int c=0; c<this->n_compositional_fields(); ++c)
+                in.composition[0][c] = this->get_initial_composition_manager().initial_composition(position, c);
+              in.strain_rate.resize(0);
+
+              this->get_material_model().evaluate(in, out);
+
+              temperature_perturbation = -1./(out.thermal_expansion_coefficients[0]) * density_perturbation;
+            }
+          else
+            // scale the density perturbation into a temperature perturbation
```
jaustermann: this comment should probably stay at the beginning of these changes. Can you move it up, e.g. in the line before 
// see if we need ...

source/initial_temperature/SAVANI_perturbation.cc
```diff
@@ -297,8 +300,28 @@ namespace aspect
...
+              for (unsigned int c=0; c<this->n_compositional_fields(); ++c)
+                in.composition[0][c] = this->get_initial_composition_manager().initial_composition(position, c);
+              in.strain_rate.resize(0);
+
+              this->get_material_model().evaluate(in, out);
+
+              temperature_perturbation = -1./(out.thermal_expansion_coefficients[0]) * density_perturbation;
+            }
+          else
+            // scale the density perturbation into a temperature perturbation
```
jaustermann: same comment on this as in S40 file

bartniday:
Changed to address comments and (I think) fix all the spacing

source/initial_temperature/S40RTS_perturbation.cc
```diff
@@ -337,8 +340,28 @@ namespace aspect
...
+        {
+          // scale the density perturbation into a temperature perturbation
+          // see if we need to ask material model for the thermal expansion coefficient
+          if (use_material_model_thermal_alpha)
+            {
+              MaterialModel::MaterialModelInputs<3> in(1, this->n_compositional_fields());
+              MaterialModel::MaterialModelOutputs<3> out(1, this->n_compositional_fields());
+              in.position[0] = position;
+              in.temperature[0] = background_temperature;
+              in.pressure[0] = this->get_adiabatic_conditions().pressure(position);
```
gassmoeller: Is it consistent to use the pressure from the adiabatic profile, but the background temperature for the temperature to compute the thermal expansivity? I do not think it matters much as thermal_expansivity is not often set to be temperature depedent (although in theory it is), but have you thought about which one to choose, and do you think this is the better choice?

bartniday: I copied the MaterialModelInputs initialization from another file, changed background_temperature, and did not think about that - I'll get back to you.

bartniday: It looks like `background_temperature` is set to the adiabatic temperature in compressible models and the reference temperature in incompressible models. That sounds reasonable, unless maybe someone uses an incompressible material model with temperature dependent thermal expansion (which sounds unlikely).

gassmoeller: Ok, sounds reasonable. I will let @jaustermann push the merge button.

jaustermann:
All of our comments are addressed, thank you! The only thing left is adding a test.

bangerth:
@bartniday -- you saw @jaustermann 's last comment, right?

bartniday:
I did. Test is incoming tomorrow.

bartniday:
Added a test for the S40RTS initial temperature model

gassmoeller:
Very nice, you need to update the test results with the official tester results, but afterwards this is ready.

bartniday:
Updated the test results now.

gassmoeller:
Thanks! :+1:

PR #2521, opened 2018-06-27
Manual: promote cmake as out-of-source

tjhei:


doc/manual/manual.tex
```diff
@@ -2780,8 +2781,10 @@ \subsubsection{Compiling \aspect{} and generating documentation}
 \end{verbatim}
 on the command line (or \texttt{make -jN} if you have multiple processors in
 your machine, where \texttt{N} is the number of processors). This builds the
-\aspect{} executable which will reside in
-the main directory and will be named \texttt{./aspect}. If you intend to
+\aspect{} executable which will reside in the \textt{build} directory
```
bangerth: texttt

doc/manual/manual.tex
```diff
@@ -2780,8 +2781,10 @@ \subsubsection{Compiling \aspect{} and generating documentation}
...
+and will be named \texttt{./aspect}. To run \aspect{} from the main source directory
```
bangerth: that's still misunderstandable. what does `.` refer to here?

tjhei: I don't know. That was the sentence before. I am removing it.

bangerth: No, it was clear before because at the time source==build. In that case, it's clear what `.` corresponds to.

PR #2523, opened 2018-06-28
Verify parameter/variable settings when restart, for issue #2090

### LINKED ISSUE #2090: Verify checkpoint/restart agree on the fields a simulation has

wanyingw:
Fixes #2090.

### DIFF
--- a/source/simulator/checkpoint_restart.cc
+++ b/source/simulator/checkpoint_restart.cc
@@ -82,7 +82,6 @@ namespace aspect
                                    aspect::oarchive &oa)
     {
       oa << parameters.convert_to_years;
-      oa << parameters.output_directory;
       oa << parameters.surface_pressure;
       oa << parameters.use_operator_splitting;
       oa << parameters.include_melt_transport;
@@ -115,15 +114,6 @@ namespace aspect
       ia >> convert_to_years;
       AssertThrow (convert_to_years == parameters.convert_to_years,
                    ExcMessage ("The value provided for `Use years in output instead of seconds' that was stored "
-                               "in the checkpoint file is not the same as the one "
-                               "you currently set in your input file. "
-                               "These need to be the same during restarting "
-                               "from a checkpoint."));
-
-      std::string output_directory;
-      ia >> output_directory;
-      AssertThrow (output_directory == parameters.output_directory,
-                   ExcMessage ("The name of the output directory that was stored "
                                "in the checkpoint file is not the same as the one "
                                "you currently set in your input file. "
                                "These need to be the same during restarting "
--- a/source/simulator/checkpoint_restart.cc
+++ b/source/simulator/checkpoint_restart.cc
@@ -138,42 +138,9 @@ namespace aspect
                                "These need to be the same during restarting "
                                "from a checkpoint."));
 
-      bool use_operator_splitting;
-      ia >> use_operator_splitting;
-      AssertThrow (use_operator_splitting == parameters.use_operator_splitting,
-                   ExcMessage ("The value provided for `Use operator splitting' that was stored "
-                               "in the checkpoint file is not the same as the one "
-                               "you currently set in your input file. "
-                               "These need to be the same during restarting "
-                               "from a checkpoint."));
-
-      bool include_melt_transport;
-      ia >> include_melt_transport;
-      AssertThrow (include_melt_transport == parameters.include_melt_transport,
-                   ExcMessage ("The value provided for `Include melt transport' that was stored "
-                               "in the checkpoint file is not the same as the one "
-                               "you currently set in your input file. "
-                               "These need to be the same during restarting "
-                               "from a checkpoint."));
-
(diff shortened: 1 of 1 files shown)


source/simulator/checkpoint_restart.cc
```diff
@@ -96,10 +111,127 @@ namespace aspect
...
+                               "in the checkpoint file is not the same as the one "
+                               "you currently set in your input file. "
+                               "These need to be the same during restarting "
+                               "from a checkpoint."));
+
       unsigned int n_compositional_fields;
       ia >> n_compositional_fields;
       AssertThrow (n_compositional_fields == parameters.n_compositional_fields,
-                   ExcMessage ("The number of compositional fields that were stored "
+                   ExcMessage ("The number of compositional fields that was stored "
```
bangerth: nice catch!

source/simulator/checkpoint_restart.cc
```diff
@@ -96,10 +111,127 @@ namespace aspect
...
+      ia >> composition_degree;
+      AssertThrow (composition_degree == parameters.composition_degree,
+                   ExcMessage ("The composition polynomial degree that was stored "
+                               "in the checkpoint file is not the same as the one "
+                               "you currently set in your input file. "
+                               "These need to be the same during restarting "
+                               "from a checkpoint."));
+
+      std::string pressure_normalization;
+      ia >> pressure_normalization;
```
tjhei: I think this is okay to be changed, at least the computation should still work.

source/simulator/checkpoint_restart.cc
```diff
@@ -96,10 +111,127 @@ namespace aspect
...
+      ia >> surface_pressure;
+      AssertThrow (surface_pressure == parameters.surface_pressure,
+                   ExcMessage ("The value of surface pressure that was stored "
+                               "in the checkpoint file is not the same as the one "
+                               "you currently set in your input file. "
+                               "These need to be the same during restarting "
+                               "from a checkpoint."));
+
+      bool use_operator_splitting;
+      ia >> use_operator_splitting;
```
tjhei: is this something we can/should be able to change?

source/simulator/checkpoint_restart.cc
bangerth: @tjhei -- I thought about that, but it will result in a simulation in which the pressure will likely have a drastic jump from one time step to the next. I don't think that's a useful mode of usage.

source/simulator/checkpoint_restart.cc
bangerth: @tjhei -- here, too, it seemed like a thing we're at best unsure whether it works. If someone really wants to do this, we can always reconsider this, but it seems like a dubious proposition.

Or do you feel strongly about it?

source/simulator/checkpoint_restart.cc
```diff
@@ -81,8 +81,23 @@ namespace aspect
     void save_critical_parameters (const Parameters<dim> &parameters,
                                    aspect::oarchive &oa)
     {
+      oa << parameters.convert_to_years;
+      oa << parameters.output_directory;
```
gassmoeller: I disagree with including this parameter. I can imagine plenty of applications for changing this parameter during restarts (e.g. parameter variations from a common checkpoint) and it is really not critical, as long as all files are in the new directory I am always able to restart from there.

bangerth:
I added the missing comments and removed the output directory in commits of my own. Let's see what happens now.

gassmoeller:
It is stil failing some tests. Do you want me to take a look?

tjhei:
checkpoint_01 fails with
```
An error occurred in file <checkpoint_restart.cc> in function
The violated condition was: 
    composition_degree == parameters.composition_degree
```
the reason is that you are not serializing and deserializing the same thing (``use_operator_splitting`` is the third variable to serialize but is not deserialized). @bangerth can you maybe do the serialization and deserialization in the same function to not duplicate the logic?

bangerth:
I have this on my list of things to look at. Give me a couple of days.

bangerth:
Indeed, all of these are written but not read:
```
     oa << parameters.use_operator_splitting;
     oa << parameters.include_melt_transport;
     oa << parameters.enable_additional_stokes_rhs;
     oa << parameters.stokes_velocity_degree;
```

bangerth:
@wanyingw -- apologies for letting this sit for a whole year, but we have now taken your work (with minor modifications) and fixed the issue. The commit that does this is still credited to you.

wanyingw:
Thank you for letting me know! Wish everything is well with you!

PR #2530, opened 2018-06-28
New visualization postprocessor for seismic velocities in models with melt

joeschools:
My main hackathon work. Created a new visualization post processor that outputs the seismic velocities and velocity anomalies, as a function of porosity, in models with melt. The test output is mostly non dimensional nonsense, but when applied to the mid-ocean ridge cookbook the output roughly matches observations.


source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,310 @@
...
+                                 "Units: $GPa$.");
+
+              prm.declare_entry ("Liquid bulk modulus", "40",
+                                 Patterns::Double (),
+                                 "The bulk modulus of the liquid. "
+                                 "Units: $GPa$.");
+
+              prm.declare_entry ("Shear modulus", "64",
+                                 Patterns::Double (),
+                                 "The shear modulus of the solid material"
```
tjhei: space at the end

joeschools:
I addressed Timo's comments. I took out internal porosity thresholds in favor of putting limits on the contiguity so that it must be between zero and one. Contiguity is essentially a measure of how grains are touching, so either all grain sides are touching, with no melt or void space (one), or none of the grains are touching and are floating in the melt (zero).

In the spirit of more openness and user input I added in the option to use different coefficients for the contiguity equation in the parameter file. The defaults are from Wimert and Hier-Majumder (2012) https://doi.org/10.1029/2011JB009012 which has associated error with the fit to data. Now the user can adjust within error or some new estimates, assuming it is a polynomial fit. I plan on extending the user input further in the future, possibly including options for different relations. Since the contiguity is dependent on user input here, I am closing pull request #2531 and instead making contiguity a visual output in this postprocessor.

Also fixed a bug based on my misunderstanding of how array and matrix indexing is handled. Previously I was referencing elements stating from one instead of zero.

bangerth:
Can someone with melt experience take another look at this?

source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,372 @@
...
+        AssertThrow(this->include_melt_transport()==true,
+                    ExcMessage("'Include melt transport' has to be on when using melt transport postprocessors."));
+
+        const unsigned int n_quadrature_points = input_data.solution_values.size();
+        Assert (computed_quantities.size() == n_quadrature_points,    ExcInternalError());
+        Assert (input_data.solution_values[0].size() == this->introspection().n_components,   ExcInternalError());
+
+        // Set use_strain_rates to true since the compaction viscosity might also depend on the strain rate.
+        MaterialModel::MaterialModelInputs<dim> in(input_data,
+                                                   this->introspection(), true);
```
jdannberg: I think you can set this to `false` instead of `true`, as you only use the densities (not the viscosities) from the material model outputs, and change the comment above to something like:
'We don't need the strain rate, because we only evaluate the solid and melt densities'

source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,372 @@
...
+        MaterialModel::MeltOutputs<dim> *melt_outputs = out.template get_additional_output<MaterialModel::MeltOutputs<dim> >();
+        AssertThrow(melt_outputs != NULL,
+                    ExcMessage("Need MeltOutputs from the material model for computing the melt properties."));
+
+        const unsigned int porosity_idx = this->introspection().compositional_index_for_name("porosity");
+
+        for (unsigned int q=0; q<n_quadrature_points; ++q)
+          {
+            unsigned output_index = 0;
+            for (unsigned int i=0; i<property_names.size(); ++i, ++output_index)
```
jdannberg: Can you move this loop down to line 193?
The way you do it right now, you do the computation again and again for every in the list, while in realisty, you only need to do it once and then just fill the different properties.

source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,372 @@
...
+        this->get_material_model().evaluate(in, out);
+        MaterialModel::MeltOutputs<dim> *melt_outputs = out.template get_additional_output<MaterialModel::MeltOutputs<dim> >();
+        AssertThrow(melt_outputs != NULL,
+                    ExcMessage("Need MeltOutputs from the material model for computing the melt properties."));
+
+        const unsigned int porosity_idx = this->introspection().compositional_index_for_name("porosity");
+
+        for (unsigned int q=0; q<n_quadrature_points; ++q)
+          {
+            unsigned output_index = 0;
```
jdannberg: You shouldn't need the `output_index` here. In the melt properties postprocessor, it was needed because one of the output properties was a vector, so each of the vector components needed a separate index, but all the components belonged to one property. You have only scalars, so in your case the `output_index` is always equal to `i`, and you can just use `i` instead.

source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,372 @@
...
+        // Set use_strain_rates to true since the compaction viscosity might also depend on the strain rate.
+        MaterialModel::MaterialModelInputs<dim> in(input_data,
+                                                   this->introspection(), true);
+        MaterialModel::MaterialModelOutputs<dim> out(n_quadrature_points, this->n_compositional_fields());
+        MeltHandler<dim>::create_material_model_outputs(out);
+
+        this->get_material_model().evaluate(in, out);
+        MaterialModel::MeltOutputs<dim> *melt_outputs = out.template get_additional_output<MaterialModel::MeltOutputs<dim> >();
+        AssertThrow(melt_outputs != NULL,
+                    ExcMessage("Need MeltOutputs from the material model for computing the melt properties."));
```
jdannberg: Please update the comment to refer to 'melt seismic properties' instead of 'melt properties'.

source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,372 @@
...
+            for (unsigned int i=0; i<property_names.size(); ++i, ++output_index)
+              {
+                double porosity = std::max(in.composition[q][porosity_idx],0.0);
+
+                if (porosity < 0)
+                  porosity = 0.0;
+
+                double Vp = 0.0;
+                double Vs = 0.0;
+                double contiguity = 0.0;
```
jdannberg: You can just initialize the contiguity as 1.0 here, then you don't need the statement in line 188.

source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,372 @@
...
+                                 "of Wimert and Hier-Majumder (2012).");
+
+              prm.declare_entry ("p5", "-19.77",
+                                 Patterns::Double (),
+                                 "The fifth coefficient, associated with "
+                                 "porosity to the first power, in the fifth "
+                                 "order polynomial contiguity equation "
+                                 "of Wimert and Hier-Majumder (2012).");
+
+              prm.declare_entry ("p6", "1.00",
```
jdannberg: This should always be one, correct?
If the porosity approaches 0, you want the contiguity to approach 1. I guess anything else wouldn't make any sense physically, would it? If that's the case, just don't make it a variable at all, but just replace it by 1.0 in line 144.

source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,372 @@
...
+                double a_3 = 0.0;
+                double b_1 = 0.0;
+                double b_2 = 0.0;
+
+                if (porosity > 0)
+                  {
+                    //contiguity relation from Wimert and Hier-Majumder 2011
+                    contiguity = p1 * std::pow(porosity,5) + p2 * std::pow(porosity,4) + p3 * std::pow(porosity,3) + p4 * std::pow(porosity,2) + p5 * porosity + p6 ;
+                    if (contiguity < 0)
+                      contiguity = 0.0;
```
jdannberg: You can make this a bit shorter by using std::max(p1*...., 0.0) in line 144.

source/postprocess/visualization/seismic_melt.cc
```diff
@@ -0,0 +1,372 @@
...
+
+                    const double b_hat [2][2] =
+                    {
+                      { -0.3238, 0.2341},
+                      { -0.1819, 0.5103},
+                    };
+
+                    a_1 = a_hat[0][0] * std::exp(((a_hat[0][1] * (poisson - 0.25)) + a_hat[0][2] * std::pow((poisson - 0.25),3))) + a_hat[0][3] ;
+                    a_2 = a_hat[1][0] * std::exp(((a_hat[1][1] * (poisson - 0.25)) + a_hat[1][2] * std::pow((poisson - 0.25),3))) + a_hat[1][3] ;
+                    a_3 = a_hat[2][0] * std::exp(((a_hat[2][1] * (poisson - 0.25)) + a_hat[2][2] * std::pow((poisson - 0.25),3))) + a_hat[2][3] ;
```
jdannberg: It is a bit more elegant and shorter if you write this as a loop from 1...3 instead of copying the expression three times.

PR #2531, opened 2018-06-28
Contiguity

joeschools:
Added contiguity to the melt material properties visualization plugin.

Contiguity is a property of melting materials, related to the area of the faces of touching grains making up the solid matrix. I had to put these equations into ASPECT for #2530 but I think there are melt minded people out there that may want to see the contiguity output directly.


joeschools:
Sorry, yes I rolled it into #2530. I mentioned it there but not here. It makes more sense as a part of the new post processor in that request than on it's own.

bangerth:
Ah, thanks for the explanation!

PR #2533, opened 2018-06-28
Postprocessor to calculate intermediate quantities for grain orientation lag parameter

hfmark:
My holy grail here was the grain orientation lag parameter defined by Kaminski and Ribe (2002, Gcubed). I can't yet calculate it within ASPECT, so instead I've written a postprocessor to output two intermediate quantities that can be used to get the grain orientation lag with some external processing.


source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+        // Set up material models
+        MaterialModel::MaterialModelInputs<dim> in(n_q_points,
+                                                   this->n_compositional_fields());
+        MaterialModel::MaterialModelOutputs<dim> out(n_q_points,
+                                                     this->n_compositional_fields());
+
+        // Set up cell iterator for looping
+        typename DoFHandler<dim>::active_cell_iterator cell =
+          this->get_dof_handler().begin_active(), endc =
+            this->get_dof_handler().end();
```
bangerth: That's slightly funny indentation for the two variables. Just say
```
  typename DoF...iterator
     cell = ...,
     endc = ...;
```

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                  this->get_solution(), velocities);
+                fe_values[this->introspection().extractors.velocities].get_function_gradients(
+                  this->get_solution(), velocity_gradient);
+
+                // Calculate eigenvalues of strain rate and take maximum (absolute value)
+                // to get tauISA, the timescale for grain rotation toward the infinite strain axis
+                const SymmetricTensor<2, dim> strain_rate = in.strain_rate[0];
+                std::array<double, dim> strain_rate_eigenvalues = eigenvalues(
+                                                                    strain_rate);
+                double lamda1 = std::max(std::abs(strain_rate_eigenvalues[0]),
```
bangerth: lambda (with a b)

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                                                               Tensor<2, dim>());
+                fe_values[this->introspection().extractors.velocities].get_function_values(
+                  this->get_solution(), velocities);
+                fe_values[this->introspection().extractors.velocities].get_function_gradients(
+                  this->get_solution(), velocity_gradient);
+
+                // Calculate eigenvalues of strain rate and take maximum (absolute value)
+                // to get tauISA, the timescale for grain rotation toward the infinite strain axis
+                const SymmetricTensor<2, dim> strain_rate = in.strain_rate[0];
+                std::array<double, dim> strain_rate_eigenvalues = eigenvalues(
```
bangerth: you can make this variable `const`. same for most of the following variables

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                                                                    strain_rate);
+                double lamda1 = std::max(std::abs(strain_rate_eigenvalues[0]),
+                                         std::abs(strain_rate_eigenvalues[dim]));
```
bangerth: This looks like an invalid array access -- you are accessing element `dim`, but the last value index is `dim-1` (there are `dim` elements, starting at zero).

bangerth: In fact, it's not clear to me what you are doing here -- what do you do in 3d? assume that the largest eigenvalue is either the first or the last one in the array?

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                double tauISA = 1.0 / lamda1;
+
+                // Get the deformation gradient tensor in the limit of "infinite" time
+                // from the velocity gradient tensor. As per Kaminski and Ribe (2002),
+                // tmax = 75*tauISA is infinite enough for most applications.
+                unsigned int maxorder = 4;
+                Tensor<2, dim> F = Tensor<2, dim>(
+                                     unit_symmetric_tensor<dim, double>());
```
bangerth: I think you should be able to write this as
```
  Tensor<2,dim> F = unit_symmetric_tensor<dim>();
```

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                double tmax = 75 * tauISA;
+
+                for (unsigned int i = 0; i < maxorder; ++i)
+                  {
+                    Tensor<2, dim> multiplier = Tensor<2, dim>(
+                                                  unit_symmetric_tensor<dim, double>());
```
bangerth: same here

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                    double factor = 1.0;
+                    for (unsigned int j = 0; j < i; ++j)
+                      {
+                        multiplier = multiplier
+                                     * Tensor<2, dim>(velocity_gradient[0]);
+                        factor *= tmax / (j + 1);
```
bangerth: I'm not sure I understand this algorithm. Can you add a comment that explain what this does?

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                      }
+                    multiplier = factor * multiplier;
+                    F += multiplier;
+                  }
+
+                // The "left stretch" tensor is computed from the deformation gradient tensor
+                // Note that Kaminski and Ribe incorrectly wrote U=Ft*F (typo, I guess).
+                Tensor<2, dim> Ft = transpose(F);
+                SymmetricTensor<2, dim> U = SymmetricTensor<2, dim>(F * Ft);
```
bangerth: again, almost all of these variables can be made `const`.

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                // Calculate theta by dotting the ISA (ehat) and the velocity (scaled to a unit vector)
+                // theta is limited to be less than pi/2 (stay in the first quadrant)
+                double umag = std::sqrt(velocities[0] * velocities[0]);
+                double umagmin = 1.0e-40;  // to avoid problems with /0 in acos
+                double theta_val = 0;
+                if (umag > umagmin)
+                  {
+                    theta_val = std::acos(std::abs((velocities[0] * ehat) / umag));
+                    if (theta_val > 1.57079632679)
+                      theta_val = fmod(theta_val, 1.57079632679);
```
bangerth: the numeric value here could better be written as `numbers::PI/2`

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+        // Set up material models
+        MaterialModel::MaterialModelInputs<dim> in(n_q_points,
+                                                   this->n_compositional_fields());
+        MaterialModel::MaterialModelOutputs<dim> out(n_q_points,
+                                                     this->n_compositional_fields());
+
+        // Set up cell iterator for looping
+        typename DoFHandler<dim>::active_cell_iterator cell =
+          this->get_dof_handler().begin_active(), endc =
+            this->get_dof_handler().end();
```
bangerth: here and below, may of the same comments apply as for your first postprocessor. I won't repeat them, but please go through the code and address what you see fits in the same style as my questions/comments above

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+
+// explicit instantiations
+namespace aspect
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      ASPECT_REGISTER_VISUALIZATION_POSTPROCESSOR(Theta, "Theta",
+                                                  "A visualization output object that generates output"
```
bangerth: missing space at the end of the line

source/postprocess/visualization/theta_tauisa.cc
```diff
@@ -0,0 +1,215 @@
...
+                                                  "showing the angle between the infinite strain axis "
+                                                  "(approximated by a truncated series solution) "
+                                                  "and the flow velocity.")
```
bangerth: If you had one available, a formulaic description of these quantities would be much appreciated here.

The text you have here will eventually become input to latex, so you can use all latex comments for formulas, etc. The only thing to pay attention to is that in latex you use one `\`, whereas here you need to double up on all backslashes.

The same applied below for the other postprocessor.

bangerth: Also, much of the commentary you have in the big comment at the top of the .h file would fit very well here. Most people read the manual (which is generated using information such as the one here), not the doxygen generated documentation, so having the documentation here (or ideally in both places) is useful!

bangerth:
I forgot to also mention this: the names you have for your postprocessors -- `tauISA` and `theta` -- are quite generic. They probably don't mean very much to people who are looking around what postprocessors are there and who are wondering whether certain functionality is already available. Do you think you can come up with longer, more descriptive names?

hfmarkp:
Thanks for the comments, Wolfgang! I think I addressed most of them here, though I may have missed a few spots where things should be 'const's. To be honest I'm not totally sure what that does, though I get the impression it has something to do with memory management?

With the indexing of the eigenvalues, I did mean to use [dim-1] - I want the eigenvalue with the largest magnitude, and since I think the eigenvalues() function returns the values sorted by size, that'll either be the first or the last depending on whether there are any negative values.

source/postprocess/visualization/grain_lag_angle.cc
```diff
@@ -0,0 +1,178 @@
...
+                const SymmetricTensor<2, dim> U = SymmetricTensor<2, dim>(F * Ft);
+                // The ~infinite strain axis is the first eigenvector of the "left stretch" tensor.
+                const Tensor<1, dim> ehat = eigenvectors(U)[0].second;
+
+                // Calculate theta by dotting the ISA (ehat) and the velocity (scaled to a unit vector)
+                // theta is limited to be less than pi/2 (stay in the first quadrant)
+                const double umag = std::sqrt(velocities[0] * velocities[0]);
+                const double umagmin = 1.0e-40;  // to avoid problems with /0 in acos
+                double theta_val = 0;
+                if (umag > umagmin)
```
gassmoeller: there are libraries that make it unnecessary to use small arbitrary numbers. I would suggest rewriting this as:
```
if (umag > 100.0 * std::numeric_limits<double>::min())
```
This way it will be tied to the precision of the system you are using.

source/postprocess/visualization/grain_lag_angle.cc
```diff
@@ -0,0 +1,178 @@
...
+                  {
+                    theta_val = std::acos(std::abs((velocities[0] * ehat) / umag));
+                    if (theta_val > numbers::PI/2)
+                      theta_val = fmod(theta_val, numbers::PI/2);
```
gassmoeller: please use `std::fmod` to make it more clear

source/postprocess/visualization/grain_lag_angle.cc
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// explicit instantiations
+namespace aspect
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      ASPECT_REGISTER_VISUALIZATION_POSTPROCESSOR(GrainLagAngle, "grain lag angle",
```
gassmoeller: please move the second argument to the next line

hfmarkp:
@gassmoeller - Thanks for looking this over, and sorry it took me so long to get back to! I think I addressed all your comments.

gassmoeller:
There are some errors compiling your changes, could you take a look (click on 'details' next to the 'aspect-8.5-gcc-hd' tester)? Let me know if you get stuck with one of the errors.

hfmarkp:
@gassmoeller There were some typos in there - my bad. I fixed those (and also fixed the tests) so now it all works on my local machine but I don't think it'll compile on the tester because I'm using the eigen*() functions from dealii 9.0.0. Is there a way to specify dealii versions here, or should this just be shelved until aspect switches from 8.5.0?

bangerth:
You could enclose everything (the whole class declaration in the .h file, and the class definition in the .cc file) with
```
#if DEAL_II_VERSION_GTE(9,1,0)
...
#endif
```
But that has the disadvantage that users of older versions don't get to see the postprocessor at all. (Of course they don't get to *use* it anyway.) My preference would be if you just enclosed the calls to the eigenvalue functions with this and did
```
#if DEAL_II_VERSION_GTE(9,1,0)
...  // call to eigenvalue functions
#else
  Assert (false, ExcMessage ("This postprocessor cannot be used with deal.II versions before 9.1."));
#endif
```
This also makes it clear to us what we need to change once we require deal.II 9.1.

gassmoeller:
Wolfgang's suggestion seems the right way to solve this. It will mean that the tester will compile, but the test will fail at the moment. You can fix this by renaming the test from `*.prm` to `*.prm.deal.II.9.0`. This way the tester will not find the test. We will rename the test back when we change the tester to use 9.0. This is what we have done for some other tests that also require 9.0.

hfmarkp:
Thanks for the suggestions, @bangerth and @gassmoeller - I added the if statements and renamed the tests.

gassmoeller:
Hi Hannah,
sorry for coming back so late to this PR. There are two little things remaining:
- I should have seen this earlier, but please make the `Assert` that you added because of Wolfgang's comment an `AssertThrow`. This way we do the check for both debug and release mode of ASPECT.
- You will need to update the indentation of your file. You can do so by downloading the `changes.diff` file you find when you click on 'Details' next to the `aspect-astyle-hd` test results at the bottom of this PR. Save this file into your ASPECT folder, and use the following terminal command to update your files: `patch -p 1 < changes-astyle.diff`.

hfmark:
Thanks for checking it again, Rene. I changed the assertion and (hopefully) fixed the indentation.

gassmoeller:
Yes, looks great! Ready to merge. :+1:

PR #2534, opened 2018-06-28
[WIP] first version of bfbt

tjhei:
compiles, runs, converges (slowly). Opening so @bangerth and @gassmoeller can take a look.


tjhei:
Note to self:
- try making the AMG solve preserve mean 0
- the approximation of (B C^-1 BT) should probably scale with the lumped matrix C instead of assembling the coefficient directly

PR #2537, opened 2018-06-29
doc: added small section on coefficient averaging

bobmyhill:
...for multicomponent material models.

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -969,6 +969,36 @@ of state which exist in the thermodynamic literature which do not obey these rel
   compressibility for each of the different approximations. The conversion between isothermal
   and isentropic compressibilities is given in ~\eqref{eq:mm_isentropic_compressibility}.}
 
+
+\subsubsection{Coefficient averaging}
+In multiphase rocks, or multirock areas in convection simulations, properties must be averaged
+because the length scales at which the rock types vary is far smaller than the resolution of the mesh.
+As a consequence, we need to use ``effective coefficients'', i.e., coefficients that do not
+correspond to any particular rock, but that lead to a macroscopic response that is a good match
+to the response of the correct, but unresolvable mixture of rocks.
+For viscosity and conductivity, there is no single expression that describes how averaging should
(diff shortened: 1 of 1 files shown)


doc/manual/manual.tex
```diff
@@ -969,6 +969,31 @@ \subsubsection{Coefficient self-consistency}
   compressibility for each of the different approximations. The conversion between isothermal
   and isentropic compressibilities is given in ~\eqref{eq:mm_isentropic_compressibility}.}
 
+
+\subsubsection{Coefficient averaging}
+In multiphase rocks, or multirock areas in convection simulations, properties must be averaged.
```
bangerth: would you be willing to add a half sentence of the form "...because the length scales at which the rock types vary is far smaller than the resolution of the mesh. As a consequence, we need to use ``effective coefficients'', i.e., coefficients that do not correspond to any particular rock, but that lead to a macroscopic response that is a good match to the response of the correct, but unresolvable mixture of rocks."

doc/manual/manual.tex
```diff
@@ -969,6 +969,31 @@ \subsubsection{Coefficient self-consistency}
...
+For viscosity and conductivity, there is no single expression that describes how averaging should
+be performed; indeed, these properties are dependent on rock texture and mineral alignment,
+both of which may change through time as strain accumulates, and chemical diffusion and
+reactions take place.
+
+In the case of density, thermal expansivity, heat capacity and bulk compressibility, there is
+one correct way of averaging. Here we must consider conservation of mass and composition in a
+multicomponent rock $r$. If component $i$ has masses $M_i$ and densities $\rho_i$, we
```
bangerth: do you mean "fractional masses"? From the formula I believe you mean that they should add up to one, right?

bobmyhill: No, I mean masses. x_i are the fractional masses (in the second of the following two equations). The division by the total mass is trivial, so I can remove it if you want. I just find this way marginally easier to follow.

doc/manual/manual.tex
bobmyhill: Sure. Done!

PR #2540, opened 2018-07-02
test: function parser and if()

tjhei:
This test confirms the suspicision about the fact that if() will
evaluate the then and the else branch, which can lead to floating point
exceptions. Check that the ternary operator works.

### DIFF
--- a/unit_tests/first.cc
+++ b/unit_tests/first.cc
@@ -20,6 +20,7 @@
 
 #include "common.h"
 #include <aspect/utilities.h>
+#include <deal.II/base/function_parser.h>
 
 TEST_CASE("floating point check example")
 {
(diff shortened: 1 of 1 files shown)


tjhei:
This was prompted from a discussion on the mailing list. Not a super useful test, but I thought I would upload instead of deleting it.

More importantly, should we use the ``?`` operator instead of ``if`` in the examples and manual?

gassmoeller:
Looks good. Yes, I think we should gradually change all uses of the function parser if() to the new syntax. I remember vaguely that we had this problem years ago, and it is just too obscure to figure out for users.
How would a nested if look like in the new syntax? `if() ? if () ? if () ? a : b : c : d` ? @jdannberg you use the function parser extensively, would this be acceptable for the benefit of not crashing because of division-by-zero?

tjhei:
> How would a nested if look like in the new syntax?

``(x>0) ? ( (y>0) ? 1 : 0 ) : -1`` vs ``if(x>0, if(y>0, 1, 0), -1)``

gassmoeller:
That looks reasonable. Lets merge this, and gradually replace existing expressions by the new syntax. Could you open an issue for that?

bangerth:
Oh, bummer. We should really report this to the authors of the fparser library. I agree that the syntax of `if(a,b,c)` is difficult, but `(a ? b : c)` is only marginally better and just as difficult if you happen to not be a C programmer.

bangerth:
See https://github.com/beltoforion/muparser/issues/51

mibillen:
Hi there, I was hoping that I might help by providing a simple cookbook example using the nested lazy-expression syntax (and a python script that does the same thing) and to modify the section of the manual "A note on syntax of formulas in input files" to explain a bit better about the syntax and point to the new example (and how to find other example in cookbook files etc.).  I have this all ready, but I have never contributed and I need someone to walk me through the process the first time (maybe we can start with just editing the manual section, and then I can add the cookbook?) Can someone walk me through this?

tjhei:
> Oh, bummer. We should really report this to the authors of the fparser library.

This has nothing to do with that library. ``if`` is a function I added for backwards compatibility with our earlier library we used: https://github.com/dealii/dealii/blob/master/source/base/function_parser.cc#L130

bangerth:
Oh, and because it's just a regular function it needs to evaluate its second and third argument :-(

gassmoeller:
@mibillen: A short explanation for our process is here: https://github.com/geodynamics/aspect/blob/master/CONTRIBUTING.md#making-aspect-better, but I will also be at CIDER and can walk you through the steps of creating a pull request if you want.

PR #2542, opened 2018-07-02
Add option to use full A block as preconditioner

gassmoeller:
In order to finish #2498 I would like to be able to use the same solver setup as the paper Rudi et al (2017), "Weighted BFBT Preconditioner for Stokes Flow Problems with Highly Heterogeneous Viscosity". However, they use the full A block for the preconditioner, instead of our simplified version. This PR adds a parameter that allows changing this behavior, which might also be of interest for other models that struggle with convergence issues in the linear solver. I also tried to avoid some unnecessary work in that case (there is not much use in assembling a preconditioner matrix block that will never be used). 
In theory there is more potential for savings, we could even think about not allocating the preconditioner A block at all, but that would require more changes to the constraints, so i skipped it for now. 
Also we could remove the arbitrary `if (parameters.free_surface_enabled || parameters.include_melt_transport)` conditions and simply set `use_full_A_block_preconditioner` if we have a free surface or melt. Would that be useful?

### DIFF
--- a/source/simulator/assemblers/stokes.cc
+++ b/source/simulator/assemblers/stokes.cc
@@ -99,17 +99,18 @@ namespace aspect
             }
           else
             {
+              const unsigned int pressure_component_index = this->introspection().component_indices.pressure;
               for (unsigned int i = 0; i < stokes_dofs_per_cell; ++i)
-                for (unsigned int j = 0; j < stokes_dofs_per_cell; ++j)
-                  if (scratch.dof_component_indices[i] ==
-                      scratch.dof_component_indices[j])
-                    {
-                      data.local_matrix(i, j) += (one_over_eta * pressure_scaling
-                                                  * pressure_scaling
-                                                  * (scratch.phi_p[i]
-                                                     * scratch.phi_p[j]))
-                                                 * JxW;
-                    }
+                if (scratch.dof_component_indices[i] == pressure_component_index)
+                  for (unsigned int j = 0; j < stokes_dofs_per_cell; ++j)
+                    if (scratch.dof_component_indices[j] == pressure_component_index)
+                      {
+                        data.local_matrix(i, j) += (one_over_eta * pressure_scaling
+                                                    * pressure_scaling
+                                                    * (scratch.phi_p[i]
+                                                       * scratch.phi_p[j]))
+                                                   * JxW;
+                      }
             }
         }
     }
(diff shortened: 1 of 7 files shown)


source/simulator/assemblers/stokes.cc
```diff
@@ -78,17 +80,37 @@ namespace aspect
...
+
+            }
+          else
+            {
+              for (unsigned int i = 0; i < stokes_dofs_per_cell; ++i)
+                for (unsigned int j = 0; j < stokes_dofs_per_cell; ++j)
+                  if (scratch.dof_component_indices[i] ==
+                      scratch.dof_component_indices[j])
+                    {
+                      data.local_matrix(i, j) += (one_over_eta * pressure_scaling
```
tjhei: this will still do a lot of unnecessary work because it loops over all velocity dofs and multiplies zeros. If you don't want to fix it (for example by looking at ``fe.system_to_component_index(i).first``), can you leave a TODO?

source/simulator/parameters.cc
```diff
@@ -367,6 +367,24 @@ namespace aspect
                            "block preconditioner for the Stokes equation can be found in "
                            "\\cite{KHB12}.");
 
+        prm.declare_entry ("Use full A block as preconditioner", "false",
+                           Patterns::Bool(),
+                           "This parameter determines whether we use an simplified approximation of the"
```
tjhei: missing spaces.

source/simulator/parameters.cc
```diff
@@ -367,6 +367,24 @@ namespace aspect
...
+                           "$A$ block as preconditioner for the Stokes solver, or the full $A$ block. The "
+                           "simplified approximation only contains the terms that describe the coupling of"
+                           "identical components (plus boundary conditions) as described in "
+                           "\\cite{KHB12}. The full block is closer to the description in "
+                           "\\cite{rudi2017weighted}. There is no clear way to determine, which preconditioner "
+                           "performs better in runtime, the default value (simplified approximation) requires "
+                           "more GMRES iterations, but is faster to apply in each iteration. On the downside it "
+                           "requires storing an additional matrix block, and such requires more memory, although "
+                           "the block is by a factor of 2 or 3 smaller than the original A block (depending on spatial "
```
tjhei: one of the reasons to use the simplified A is that AMG needs a lot less memory, so this is less clear to me. Also: right now we are not saving memory (but I could do that soon as a follow-up).

gassmoeller:
Good points, I now only assemble the pressure components, and removed the description about the memory consumption (I forgot about the additional memory consumption of the AMG). Does this address your comments?

PR #2543, opened 2018-07-03
Clean docker images on jenkins agents

gassmoeller:
Currently the jenkins agents accumulate stopped docker containers until their disk space runs out. I suspect that that was a big part of the problems we had during the hackathon. I started my old laptop again this morning and it build 5 PRs or so, until it filled 12 GB of its 20 GB with docker containers, then it crashed on any further tests, because of missing disk space. The docker prune at the post stage in the Jenkinsfile should not exactly be necessary, but the start option in the beginning will not remove already accumulated containers, so I left it in for the moment.


gassmoeller:
It seems that was not the only problem, the agent still randomly reports a read-only file system. I will try again outside of a VM tomorrow.

tjhei:
I am pretty sure pipeline issues a ``docker rm`` (see bottom of https://jenkins.tjhei.info/job/aspect/job/PR-2485/7/console for example). But I agree that adding this won't hurt. Want to merge?

tjhei:
so the build fails with "java.io.IOException: Failed to rm container "

gassmoeller:
You are right, it seems there is a `docker rm` command there already. But then why did I find all of these orphaned containers on my tester? Maybe the containers are not removed properly if the jenkins pipeline aborts at some point (e.g. because of a lost connection)? Anyway, we should not merge this until we know more.

tjhei:
> Maybe the containers are not removed properly if the jenkins pipeline aborts at some point (e.g. because of a lost connection)?

Not sure, but this is likely what happened.

bangerth:
Does it issue the `docker rm` even in cases where the build failed in one stage? In other words, is the `docker rm` just another (final) stage in the pipeline, or is it (or should be) always executed?

PR #2545, opened 2018-07-03
Add Jenkinsfile for kubernetes Jenkins

gassmoeller:
@tjesser-ucdavis-edu and I are working on a CIG jenkins server that uses kubernetes. Unfortunately, we have not yet found a way to write a Jenkinsfile that works for kubernetes and docker agents, so for now I copied the existing one and modified it slightly.

### DIFF
--- a/Jenkinsfile.k8s
+++ b/Jenkinsfile.k8s
@@ -1,0 +1,133 @@
+#!groovy
+
+pipeline {
+  agent {
+    kubernetes {
+      //cloud 'kubernetes'
+      label 'mypod'
+      containerTemplate {
+        name 'dealii'
+        image 'gassmoeller/aspect-tester:8.5.0'
+        ttyEnabled true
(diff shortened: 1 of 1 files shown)


gassmoeller:
Ok, it looks like we are having a running CIG jenkins tester :tada:. I woud like to merge this as is, although it duplicates our Jenkinsfile, because I would like to observe the new tester behavior a bit.
With that data we can think further about how to handle the tester setup and try answering questions like: Should we spent time trying to set everything up in a single Jenkinsfile? Do we want to completely switch to just one of the Jenkins servers? How stable is the new tester?

jdannberg:
As this only adds a new jenkins file, I will merge this for now so that @gassmoeller can observe and test the new behaviour. But I don't really know how these jenkins files work, so @tjhei, it would be great if you could have a look at this at some point!

PR #2547, opened 2018-07-04
Minor update to the documentation of a parameter.

bangerth:
Follow-up to #2542, while reading through what has happened over the past couple
of days in the repository :-)


gassmoeller:
Makes sense.

PR #2481, opened 2018-06-26
Donea and Huerta book 2D cartesian benchmark

cedrict:
These are the files for a new benchmark: 2D, cartesian, manufactured solution which resembles a convection cell in a unit square, taken from the Donea and Huerta FEM book.

PR #2550, opened 2018-07-04
Minor updates to the Donea-Huerta benchmark.

bangerth:
This benchmark was added by @cedrict in #2481. I thought it was faster to make
these fixes myself, rather than ask him for another round of changes.


gassmoeller:
Looks good.

PR #2554, opened 2018-07-04
Avoid creating too many jenkins kubernetes agents

gassmoeller:
So far this is a test to avoid a deprecation warning and a few issues with the kubernetes jenkins tester. Do not merge, if this works I will add more detailed explanations.

### DIFF
--- a/Jenkinsfile.k8s
+++ b/Jenkinsfile.k8s
@@ -10,6 +10,7 @@ pipeline {
         image 'gassmoeller/aspect-tester:8.5.0'
         ttyEnabled true
         command 'cat'
+        idleMinutes 20
       }
     }
   }


gassmoeller:
Ok, this should hopefully fix the issue that the new jenkins tester allocates too many new test runs simultaneously. It keeps a running jenkins agent alive for 20 minutes after it finishes. Runs in the queue will be assigned to this running agent instead of starting new agents. This circumvents the problem that the jenkins master spins up a new agent for every job in the queue, because it takes some time for the agents to come alive, and the master only checks for already started agents when deciding whether the limit of simultaneously running agents is reached.

tjhei:
please merge yourself if you are happy with it.

gassmoeller:
This is not what the parameter should do. The parameter should keep a finished runner online for 20 minutes after it finishes testing. If new jobs appear they will be assigned to this runner and jenkins does not start new runners, because the current runner limit is 1 (this is why your second PR waited for the first to finish). This mostly circumvents the situation when one runner finishes, jenkins sees we are not longer at the runner cap, starts 10 runners for the 10 jobs in the queue (because they are not yet online) and overloads the resources. I just started a few runs and will check if they work.

gassmoeller:
Oh well, this has not worked, it still generates several agents at the same time, even though the limit is 1 if more than one job is in the queue when another job finishes. I will need to look more into what is actually happening there (@tjesser-ucdavis-edu if you are interested I would be happy about suggestions how to hard limit the number of agents).

Easiest way to reproduce the problem:
Start one test run
Wait until the runner is online
Start two more test runs (they are queued)
Wait until the first one finishes
Expected result: One job from the queue gets started. Real result: Both jobs are started simultaneously on different runners.

PR #2555, opened 2018-07-05
Point values only at specific timesteps

anne-glerum:
Add option to specify the time interval between the executions of the point values postprocessor. Default behavior is still to execute every timestep.


anne-glerum:
correct

source/postprocess/point_values.cc
```diff
@@ -162,6 +188,14 @@ namespace aspect
       {
         prm.enter_subsection("Point values");
         {
+          prm.declare_entry ("Time between point values output", "0",
+                             Patterns::Double (0),
+                             "The time interval between each generation of "
+                             "graphical output files. A value of zero indicates "
```
gassmoeller: `point values output files`? graphical does not seem right.

tests/point_value_specific_output_times.prm
```diff
@@ -0,0 +1,132 @@
+# A variation of cell_in_circumference to test that the point values
+# postprocessor also works with non-rectangular meshes
```
tjhei: can you update this comment please? Otherwise good to merge.

PR #2557, opened 2018-07-05
Update a note in the manual.

bangerth:
The note showed the divergence of the stress tensor. But it talked about the
strain rate tensor (which is only a part of it) and it is in fact not necessary
to use the divergence of it at all.


gassmoeller:
Makes sense.

PR #2560, opened 2018-07-06
Add cookbook example for using lazy-expression syntax AND add to manual

mibillen:
I've created a simple cookbook example that uses the lazy-expression syntax ( ? : ) for nested if-else statements. The cookbook includes a python script that has the same if-else statement so that new users can see how to go from a more familiar way of writing these expressions to what is needed in the function expressions.

I've also modified the manual to explain the lazy-expression and point to this new cookbook.


doc/manual/manual.tex
```diff
@@ -4181,12 +4181,34 @@ \subsubsection{A note on the syntax of formulas in input files}
...
+The muparser uses lazy-expression syntax \texttt{(if-condition ?\ true-expression :\ false-expression )}
+for if-else statements. This lazy-expression only evaluates the expression that meets the
+if-condition, rather than evaluating both expressions, which can be useful if one of the
+expression is not defined (e.g., has a divide by zero) when the if-condition is not met.
+Note it is also possible to use the syntax \texttt{if(condition, true-expression, false-expression)},
+but in this case both expressions are always evaluated.
+Therefore, the lazy-expression syntax is recommended.
+
+As a simple example using the lazy-expression syntax, the statement ``if $1<x<4$ then output 1, else output 0''
+can be expressed as \texttt{(1<x\ \&\&\ x<4\ ?\ 1\ :\ 0)}.  Multiple, nested if-else expressions
```
tjhei: Can you write this as ``(1<x \&\& x<4) ? (1) : (0)`` instead? Otherwise you are relying on the operator precedence rules between && and ? (and I have no idea what they are for the mu parser).

mibillen:
Okay, I made the change to the manual file, then added, commitded and pushed... and I see that my comment from the commit is showing up... so I'm guessing that I did this step of making a change to a pull request correctly. Let me know if I need to do something else. 

The reason I added the python file example is that I found the idea of writing these nested if-else statements in muparser syntax very intimidating until I did exactly what this example does... I wrote it out in python and then translated it to the one-line expression. From there I could see better how to think through the logic.  I do think it will be helpful to students, and those with less c++ programming experience like me.

jaustermann:
Thanks @mibillen  for contributing this! I wonder if adding this cookbook to section 5 (cookbooks) in the manual would be useful. I worry that as the manual is growing people tend to just go to the cookbooks rather than reading through things before that and therefore might miss this. It seems like it could fit quite well in section 5.3 (Geophysical setups). That way you could also discuss things like this: "In addition, the parameter file shows how to run aspect to just check initial conditions are set correctly (by using and End time of zero and setting the solver tolerances to large values)." This is a useful tip but where it is now most people would never see it. 
[Related to that, Pull Request #2409 might be relevant to you, which got merged during the hackathon. You can now do initial adaptive refinement without solving the equations and / or without assigning the initial temperature IF the refinement criterium doesn't depend on the solution and / or the initial temperature.]

mibillen:
@jaustermann  - thanks for suggesting to actually add this within the cookbook section of the manual. I thought about whether i should do this also and wasn't sure if the cookbook was too simple. But I agree that helping new users (like me) past the apparently simple things (like how to I just check in the initial condition) is useful.

Is there an 'ordering' logic to the cookbooks? Do I just add at the end of the section you suggested (chronological order). Or, since this is a very simple example, should I add it at the beginning?

jaustermann:
Great, I think it would be a worthwhile addition! Looking through the cookbooks again - maybe tagging it on at the end of 5.2. (Simple setups) would make sense, so this would then be 5.2.13? I'm not sure if there is a specific order (chronological or otherwise). 
You might have already seen this, but there is also a section in the manual on how to write a cookbook (section 6.3) in case that is useful. Thanks for contributing this!

PR #2563, opened 2018-07-08
Improve prm not found error

tjhei:
Instead of showing a 16 line long and hard to understand blurb about MPI
deadlock, Exceptions, and failing assertitions in line xyz, output the
error directly.

### DIFF
--- a/source/main.cc
+++ b/source/main.cc
@@ -366,9 +366,18 @@ read_parameter_file(const std::string &parameter_file_name)
       std::ifstream parameter_file(parameter_file_name.c_str());
       if (!parameter_file)
         {
+          if (parameter_file_name=="parameter-file.prm"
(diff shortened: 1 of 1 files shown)


source/main.cc
```diff
@@ -366,9 +366,18 @@ read_parameter_file(const std::string &parameter_file_name)
       std::ifstream parameter_file(parameter_file_name.c_str());
       if (!parameter_file)
         {
+          if (parameter_file_name=="parameter-file.prm"
+              || parameter_file_name=="parameter_file.prm")
+            {
+              std::cout << "***          You should not take everything literally!          ***\n"
+                        << "*** Please pass the name of an existing parameter file instead. ***" << std::endl;
```
bangerth: here and below, please use `std::cerr` instead.

gassmoeller:
Thanks!

PR #2564, opened 2018-07-08
[WIP] "Richardson extrapolation" Error estimation postprocessing plugin

class4kayaker:
This PR begins adds a plugin to export FEM state variable data and quadrature weights at the Gauss-Legendre quadrature points for the highest refinement level, or one more refinement level for a convergence calculation. This is useful for testing convergence behavior for problems where there is not a known solution that can be used for comparison. The data is currently exported using an ASCII data format, but adding an option for the HDF5 format is likely reasonable.

Note that some of the plugin structure is based on better behavior with VOF fields [when the plugin is extended to handle them](/class4kayaker/aspect/tree/vof_richardson_update/source/postprocess/richardson_extrapolation.cc).

The actual error estimation requires a separate post-processing step, which must correctly associate the quadrature point lists. This is the most expensive part of the computation, but in some cases it can be done more quickly by exploiting some aspect of the data set.

The current implementation is functional enough to test on special cases such as the solkz benchmark. The new copy of that benchmark includes a prototype comparison script that makes use of the box geometry to efficiently get a consistent order. The script also does a basic check to ensure that the points being compared are correctly associated.

Improvements likely needed for final inclusion are:
- Add headers to output files to make some necessary information available. This should probably include some basic geometry data for use by the post-processing script.
- Identify where best to include the python post-processing script, if it is to be included
- Improve structure of the post-processing script
- Improve the post-processing script to be suitable for at least general box geometries


egpuckett:
I would like to be added to the list of people who can comment on and review this PR. The ideas that have been implemented by Ying, Harsha, and now Jon, who finally seems to have gotten it working on a uniform grid came from me.  How do I go about becoming one of the people who can comment on and review this PR.

jdannberg:
> I would like to be added to the list of people who can comment on and review this PR.

You can comment on any pull request and leave a review, the only thing you can not do is _merge_ the pull request. So just go ahead!

bangerth:
In other words: once you've left one comment, you will get all following comments on a pull request. So you're already on, @egpuckett. 

For future cases, somewhere along the top menu or the right menu, there is a button called "Watch" or "Notifications" where you can sign up to follow a pull request. (I can't see this myself because I'm automatically following all PRs in aspect.)

egpuckett:
Thanks!

> For future cases, somewhere along the top menu or the right menu, there is a button called "Watch" or "Notifications" where you can sign up to follow a pull request. (I can't see this myself because I'm automatically following all PRs in aspect.)

I found that part.  For some reason yesterday I didn't see the comment box.

class4kayaker:
I have currently refactored my "Richardson extrapolation" post-processing python script into a python package.

Would this be the preferable way to include the post-processing script, or would there be another preferred manner in which to include it?

class4kayaker:
Added cookbook files, and began writeup on example cookbook.

I would appreciate advice on where the cookbook would work best in the manual, and also the appropriate level of detail as far as getting the postprocessing python script funcitoning and applied. The cookbook for this uses a python script (rather than a bash script) to automate the full convergence computation.

PR #2565, opened 2018-07-09
Minor cleanups of the seismic anomalies postprocessor.

bangerth:

PR #2566, opened 2018-07-09
Move variable declarations right before the loop they control.

bangerth:
This is just like #2565, but I had not realized that the class is duplicated twice
in one file. So I should have made the changes of #2565 in two more places. This
is done here.


gassmoeller:
Looks good. These classes need a bigger overhaul, but this is already a step in the right direction.

bangerth:
I agree. But I'm an incrementalist -- fix what you see while you see it, rather than wait till you can do it all and thereby delaying it to Sanktnimmerleinstag (a word that can truly not be translated ;-) ).

PR #2567, opened 2018-07-11
Replace the use of a separate index by cell->active_cell_index().

bangerth:
I am pretty sure that this should work, but didn't want to put it into the
previous patches because it's probably useful to test this one
change separately.


bangerth:
Well, I thought I might as well do the same change in all of the other postprocessor files as well...

bangerth:
It turns out that this seems to work :-)

PR #2469, opened 2018-06-26
cookbook for adiabatic boundary as initial temperature condition

njinju:
This is a cookbook for the implementation of the adiabatic boundary as initial temperature condition in ASPECT.

PR #2569, opened 2018-07-12
Adiabatic boundary cookbook njinju

njinju:
This is a cookbook for the implementation of the adiabatic boundary as initial temperature condition in ASPECT. It is a revision of the earlier submitted pull request.


naliboff:
@njinju - Thanks for submitting the updated PR. Before going through this, can you take a look at the comments I made in #2469 and see if any of them still apply? 

You will also need to rebase as there is a merge conflict with the current manual. If you would like specific instructions on how to do the rebase, just let me know!

gassmoeller:
Same comment as #2570. As mentioned this is not a rejection, please make your changes to #2469 instead. We would be happy to get this cookbook in.

PR #2462, opened 2018-06-26
WIP: World builder 2.0

MFraters:
This replaces #2122. It is now a semi-separate library, which doesn't use any external functions except boost. The license is set a lgplv3, so that is possible for other codes to use it without changing their licence. They do have to share any modifications to the library which they made. Is anyone sees any copyright concerns in the files, please let me know.

I have removed the support for ellipsoidal for now, because I would need to contact Sarah Stamps first for permission. The only copied code is I think the cartesian to spherical and back, but as I understood from @tjhei, it is too small and trivial to be copyrighted. Is this correct, or should I also rewrite this?

The astyle doesn't work yet, so I will take a look at that, and I will also add tests and try to add unit tests.

If there are issues of any other kind, please let me know.

And ofcourse some figures of what this can do (temperature plots):
screenshot from 2018-06-25 17-52-28
screenshot from 2018-06-25 17-52-44

PR #2571, opened 2018-07-16
World builder 3: adding it to the core of asepct

MFraters:
Since it took too long for the review of #2462 to be done, I decided to put the world builder code in a own repository so I could develop the code faster. Now that the basics are done, I thought it would be a good idea (and useful to me) to merge it with aspect in way smaller pieces than I proposed in #2462. I hope this will help the reviewing of the code.

I have added it as a git submodule which is automatically loaded by cmake. This way the version used by aspect can be controlled, and the installation is done automatically with the installation of aspect.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,8 @@ MESSAGE(STATUS "====================================================")
 # Set the name of the project and target:
 SET(TARGET "aspect")
 
-FILE(GLOB_RECURSE TARGET_SRC  "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
-INCLUDE_DIRECTORIES(include ${CMAKE_BINARY_DIR}/include contrib/catch)
+FILE(GLOB_RECURSE TARGET_SRC  "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp" "contrib/WorldBuilder/source/*.cc")
+INCLUDE_DIRECTORIES(include ${CMAKE_BINARY_DIR}/include contrib/catch contrib/WorldBuilder/include/)
 
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.8)
 
@@ -354,6 +354,19 @@ ELSE()
   SET(ASPECT_WITH_PERPLEX OFF)
 ENDIF()
 
+EXECUTE_PROCESS(COMMAND ls ${CMAKE_SOURCE_DIR}/contrib/WorldBuilder/VERSION RESULT_VARIABLE result OUTPUT_QUIET ERROR_QUIET)
+IF (result)
+	message("-- World Builder not found.")
+	SET(ASPECT_WITH_WORLD_BUILDER OFF CACHE BOOL "" FORCE)
+ELSE()
+	file (STRINGS "${CMAKE_SOURCE_DIR}/contrib/WorldBuilder/VERSION" WorldBuilderVersion)
+	message("-- World Builder version ${WorldBuilderVersion} found.")
+	SET(ASPECT_WITH_WORLD_BUILDER ON CACHE BOOL "If ON the World Builder is compiled togheter with ASPECT.")
+	    FOREACH(_source_file ${TARGET_SRC})
+      SET_PROPERTY(SOURCE ${_source_file}
+	      APPEND PROPERTY COMPILE_DEFINITIONS ASPECT_USE_WORLD_BUILDER=1)
+ENDFOREACH()
+ENDIF()
 
 IF (ASPECT_USE_SHARED_LIBS)
   # some systems need to explicitly link to some libraries to use dlopen
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1571,6 +1571,9 @@ namespace aspect
       InitialComposition::Manager<dim>                                        initial_composition_manager;
       InitialTemperature::Manager<dim>                                        initial_temperature_manager;
       const std::unique_ptr<AdiabaticConditions::Interface<dim> >             adiabatic_conditions;
+#ifdef ASPECT_USE_WORLD_BUILDER
+      const std::unique_ptr<WorldBuilder::World>                              world_builder;
+#endif
       BoundaryVelocity::Manager<dim>                                          boundary_velocity_manager;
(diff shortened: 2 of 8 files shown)


MFraters:
It looks like the testers are not able to use git (which is needed to load in the submodule).

gassmoeller:
The testers have git (it is reported as found in the output here: https://singleb.cse.ucdavis.edu/blue/organizations/jenkins/aspect/detail/PR-2571/1/pipeline), but the aspect repository is not a git repository (because it just downloads one version instead of the whole repository). And this is a situation that is pretty common, e.g. everybody downloading the release tarball from the website or geodynamics.org will not be able to compile if we handle it this way. If you want to keep it as a submodule, you need to make sure that ASPECT still compiles and works even if there is no WorldBuilder available (i.e. by checking in CMakeLists.txt if the submodule init is successful, and then create a preprocessor variable that is available in the source files, see e.g. CMakeLists.txt:379-392 for how we do that for the `USE_FP_EXCEPTIONS` variable. Sorry for the complications with this.

MFraters:
That will mean that most of this code and all of the next pull requests for integrating the world builder will be outside of the tester coverage. Is this oke?

An other option would be to do a git clone in the correct directory when the init fails. This way the integration can be tested.

I an not personally bound to it being a submodule, but I liked the suggestion of @tjhei, because it doesn't require a lot of code copying.

include/aspect/simulator.h
```diff
@@ -1571,6 +1571,9 @@ namespace aspect
       InitialComposition::Manager<dim>                                        initial_composition_manager;
       InitialTemperature::Manager<dim>                                        initial_temperature_manager;
       const std::unique_ptr<AdiabaticConditions::Interface<dim> >             adiabatic_conditions;
+#if (ASPECT_USE_WORLD_BUILDER)
```
gassmoeller: We usually write this without parentheses around the variable.
And: In the CMakeLists you define the variable as `ASPECT_WITH_WORLD_BUILDER`, but here you say `ASPECT_USE_WORLD_BUILDER`. Did you check that both the version with and without world builder work?

include/aspect/simulator_access.h
```diff
@@ -734,6 +739,15 @@ namespace aspect
       const NewtonHandler<dim> &
       get_newton_handler () const;
 
+#if (ASPECT_USE_WORLD_BUILDER)
```
gassmoeller: same as above and in all other similar places

MFraters:
Yes, there was a bug in there, which took me a while to figure out. I changed it a bit, and it now couples correctly again with the world builder through the plugins of #2574.

tjhei:
@bangerth @gassmoeller are we okay with a submodule? This requires cloning with ``--recurse-submodule`` to enable it. 

This means that we currently have no test coverage of the world builder, which is problematic unless we find a way to pull in the submodule in the tester (I can do this on the aspect-8.5 tester but I don't know how to do this in jenkins, is there a recursive clone option?).

gassmoeller:
It might be enough to add `sh 'git submodule update --init --recursive'` to the pipeline (taken from here: https://stackoverflow.com/questions/42290133/jenkins-pipeline-git-command-submodule-update), but we can test this. I am fine with the current state in the interest of allowing Menno to iterate quickly on the development of the world builder.

tjhei:
> It might be enough to add sh 'git submodule update --init --recursive' to the pipeline

That makes sense, yes.

We also need to make sure this works as expected when we create release tarballs.

CMakeLists.txt
```diff
@@ -354,6 +354,19 @@ ELSE()
   SET(ASPECT_WITH_PERPLEX OFF)
 ENDIF()
 
+execute_process(COMMAND ls ${CMAKE_SOURCE_DIR}/contrib/WorldBuilder/VERSION RESULT_VARIABLE result OUTPUT_QUIET ERROR_QUIET)
```
gassmoeller: Please capitalize all of the cmake functions as in the other parts of this file (EXECUTE_PROCESS, MESSAGE, FILE, ENDIF).

MFraters:
I changed it, and update the submodule to the newest version while I rebasing anyway :)

gassmoeller:
I get the following warning when compiling aspect with the world builder with gcc 7.3:
```
/home/rene/software/aspect/contrib/WorldBuilder/source/parameters.cc:649:33: warning: explicit
      instantiation of 'get_point<2>' that occurs after an explicit specialization has no effect
      [-Winstantiation-after-specialization]
  template Point<2> Parameters::get_point<2>(const std::string &name) const;
                                ^
/home/rene/software/aspect/contrib/WorldBuilder/source/parameters.cc:530:15: note: previous template
      specialization is here
  Parameters::get_point(const std::string &name) const
              ^
/home/rene/software/aspect/contrib/WorldBuilder/source/parameters.cc:650:33: warning: explicit
      instantiation of 'get_point<3>' that occurs after an explicit specialization has no effect
      [-Winstantiation-after-specialization]
  template Point<3> Parameters::get_point<3>(const std::string &name) const;
                                ^
/home/rene/software/aspect/contrib/WorldBuilder/source/parameters.cc:539:15: note: previous template
      specialization is here
  Parameters::get_point(const std::string &name) const
              ^
```

Could you fix this and update the version of the world builder submodule?

gassmoeller:
Otherwise good to go.

MFraters:
>    I get the following warning when compiling aspect with the world builder with clang-6.0.0:

Fixed it. That is weird, I didn't get those warning with gcc 7 (through aspect and directly) and gcc 6.4.0 (directly). Did you use some special flags?

gassmoeller:
No, but different compilers can report different warnings. Ready to merge.

include/aspect/simulator_access.h
```diff
@@ -734,6 +739,15 @@ namespace aspect
...
       get_newton_handler () const;
 
+#ifdef ASPECT_USE_WORLD_BUILDER
+      /**
+       * Return a reference to the world builder that controls the setup of
+       * initial conditions.
+       */
+      const WorldBuilder::World &
+      get_world_builder () const;
+#endif
```
bangerth: I think this is the wrong solution. We try to be compatible by providing the same interface (if we can) whether or not we have an external library. My preference would be to *always* declare the function, and if there is no WorldBuilder available, just throw an exception.

bangerth: See #2614.

PR #2574, opened 2018-07-19
World builder3 add plugins

MFraters:
Follow up to #2571. This add the plugins which do the actual linking to the world builder.

### DIFF
--- a/source/initial_composition/world_builder.cc
+++ b/source/initial_composition/world_builder.cc
@@ -52,8 +52,8 @@ namespace aspect
                                               "Specify the composition in through the World Buider. "
                                               "More information on the World Builder can be found at "
                                               "https://geodynamicworldbuilder.github.io ."
-											  "Make sure to specify the location of the World Builder file "
-											  "in the parameter 'World builder file'.")
+                                              "Make sure to specify the location of the World Builder file "
+                                              "in the parameter 'World builder file'.")
   }
 }
 #endif
--- a/source/initial_composition/world_builder.cc
+++ b/source/initial_composition/world_builder.cc
@@ -1,0 +1,69 @@
+/*
+  Copyright (C) 2011 - 2017 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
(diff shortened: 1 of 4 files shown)


source/initial_composition/world_builder.cc
```diff
@@ -0,0 +1,69 @@
...
+  {
+
+    template <int dim>
+    double
+    WorldBuilder<dim>::
+    initial_composition (const Point<dim> &position, const unsigned int n_comp) const
+    {
+      return this->get_world_builder().composition(Utilities::convert_point_to_array(position),
+                                                   this->get_geometry_model().depth(position),
+                                                   n_comp) ? 1.0 : 0.0;
```
gassmoeller: I do not understand the final `? 1.0 : 0.0`. Why does the world build not simply return the correct compositional value? E.g. what if you want to have a material that is 0.5  of two compositions at the same time?

source/initial_composition/world_builder.cc
```diff
@@ -0,0 +1,69 @@
...
+
+// explicit instantiations
+namespace aspect
+{
+  namespace InitialComposition
+  {
+    ASPECT_REGISTER_INITIAL_COMPOSITION_MODEL(WorldBuilder,
+                                              "world builder",
+                                              "Specify the composition in through the World Buider located"
+                                              "in the ASPECT contrib directory. Make sure to specify the "
```
gassmoeller: Same here, remove the reference to the directory. Do we already have a place with more documentation about the world builder? If so we should link to it from here. Also add to the second sentence the name of the world builder file parameter: `Make sure to specify the location of the World Builder file in the parameter ``World builder file'').

source/initial_temperature/world_builder.cc
```diff
@@ -0,0 +1,72 @@
...
+
+// explicit instantiations
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    ASPECT_REGISTER_INITIAL_TEMPERATURE_MODEL(WorldBuilder,
+                                              "world builder",
+                                              "Specify the initial temperature in through the World Buider located"
+                                              "in the ASPECT contrib directory. Make sure to specify the "
```
gassmoeller: Same as for the composition.

source/initial_composition/world_builder.cc
MFraters: hmm, I have changed my mind on this several times now. It is through this mechanism possible to have several compositions at the same time, but they all return true or false. I could change it back, so that it returns a double, but I have a hard time coming up with cases where you want to start a model with multiple compositions at the same location. Do you have something in mind?

gassmoeller: Sorry for the late reply, was travelling. The prime example would be having grain size (or finite strain) in some compositional fields, and actual (chemical) compositions in the other fields. Also porosity in melt models is a separate field that is independent of the other fields.

bangerth:
The first three commits have already been merged. Can you rebase?

MFraters:
Sorry it took so long. I have been working focusing on first implementing on several crucial features like subduction zones and layers. Especially the implementation of suduction zones took a lot of time to figure out the right way of implementing it in spherical coordinates. I think that I have addressed all the comments, and it is ready for an other round of reviews.

MFraters:
Can anyone review this? It would be nice to have this merged soonish.

gassmoeller:
Looks good. Sorry for the delay.

MFraters:
Thanks!

PR #2575, opened 2018-07-20
Max yield strength

mibillen:
This pull requests add an input variable to the visco_plastic material model to allow the user to limit the maximum value of the yield stress. This is necessary for models that include subducting lithosphere: the drucker-prager equations predicts that the yield stress just keeps increasing with depth, but this is not physically realistic.


include/aspect/material_model/visco_plastic.h
```diff
@@ -313,6 +313,8 @@ namespace aspect
         std::vector<double> angles_internal_friction;
         std::vector<double> cohesions;
         std::vector<double> exponents_stress_limiter;
+        
+         double max_yield_strength; // limit maximum yield stress from drucker-prager
```
gassmoeller: Please put the comment in the doxygen format above the variable declaration. That way it shows up in our online documentation at: https://aspect.geodynamics.org/doc/doxygen/classaspect_1_1MaterialModel_1_1ViscoPlastic.html

The format is as above:
```
/**
 * Comments
 */
double max_yield_strength;

File: source/material_model/visco_plastic.cc [hunk 2]
```diff
@@ -252,12 +252,13 @@ namespace aspect
...
+          // Uses max_yield_strength to limit the yield strength for slab beneath the lithosphere
           double yield_strength = ( (dim==3)
                                     ?
-                                    ( 6.0 * coh * std::cos(phi) + 6.0 * std::max(pressure,0.0) * std::sin(phi) )
-                                    / ( std::sqrt(3.0) * (3.0 + std::sin(phi) ) )
+                                    std::min(( 6.0 * coh * std::cos(phi) + 6.0 * std::max(pressure,0.0) * std::sin(phi) )
+                                    / ( std::sqrt(3.0) * (3.0 + std::sin(phi) ) ), max_yield_strength)
                                     :
-                                    coh * std::cos(phi) + std::max(pressure,0.0) * std::sin(phi) );
+                                    std::min(coh * std::cos(phi) + std::max(pressure,0.0) * std::sin(phi), max_yield_strength) );
```
gassmoeller: I would prefer if we make this not more compilcated than it already is. Maybe just put it in a new line:
```
double yield_strength = ....

yield_strength = std::min(yield_strength, max_yield_strength);
```

source/material_model/visco_plastic.cc
```diff
@@ -1106,7 +1114,9 @@ namespace aspect
           // Stress limiter parameter
           exponents_stress_limiter  = Utilities::possibly_extend_from_1_to_N (Utilities::string_to_double(Utilities::split_string_list(prm.get("Stress limiter exponents"))),
                                                                               n_fields,
-                                                                              "Stress limiter exponents");
+                                                                             "Stress limiter exponents");
```
gassmoeller: I think you accidentally modified the indentation here.

gassmoeller:
Please address the comments in #2578, I will close this PR for now.

source/material_model/visco_plastic.cc
naliboff: Agree with @gassmoeller, the formulas are already quite long.

naliboff:
@billen - Agree again with Rene, the code looks good. 

The last step would just be to add a new test for this feature. I think you could likely do this with just one two lines by deriving your test with one of the `visco_plastic` tests that including yielding. 

For an example of how to do this, see the test `viscoelastic_fixed_elastic_time_step`. In that case the test parameter file modifies a cookbook parameter file.

PR #2578, opened 2018-07-20
Add a maximum yield strength value for drucker-prager rheology needed…

mibillen:
this pull request re-opens #2575


mibillen:
I made the test by copying the test for visco_plastic_yield. After figuring out the stress level that occurs in this test case (1.0e6 Pa), I set the the parameter Maximum yield stress = 0.5e6. With this
value of the stress, and the applied strain-rate (3.17e-16) the viscosity after yielding is 5.0e5/2/3.17e-16 = 7.88e20.  I modified the output to write the gnuplot file. And here you can see
that the viscosity is correct.

tests/visco_plastic_yield_max.prm
```diff
@@ -0,0 +1,111 @@
+# Global parameters
```
naliboff: Can you  add a few brief sentences at the top of the parameter file describing what this parameter file and testing and what the expected output is? Effectively what you outlined in your earlier comment?

gassmoeller:
Hi Magali,
I think you accidentally included some changes that were not meant for this pull request. I can show you how to fix it, but I do not know your office number or where to find you.

tests/visco_plastic_yield_max.prm
mibillen: This is done.

include/aspect/material_model/visco_plastic.h
```diff
@@ -314,6 +314,8 @@ namespace aspect
         std::vector<double> cohesions;
         std::vector<double> exponents_stress_limiter;
 
+        double max_yield_strength; // limit maximum yield stress from drucker-prager
```
gassmoeller: Please put this comment before the parameter, like this:
```
/**
 * Limit maximum yield stress from drucker-prager.
 */
```

gassmoeller:
Ready to merge. Thank you! :+1:

Issue #2579, opened 2018-07-22
viscoplastic model uses velocity element to evaluate reaction terms


jdannberg:
The viscoplastic material model uses an FeValues object to evaluate the gradients of the velocity, and subsequently writes it into reaction terms for the compositional fields. As it does not know the quadrature formula, it assumes `QGauss<dim> quadrature_formula (this->get_fe().base_element(this->introspection().base_elements.velocities).degree+1)`. This will only be correct if the order of the finite element of velocity and composition match, and there is no check for that. In the viscoelastic model, this is done differently, using the current quadrature points, which I think is probably a good solution. So we should (a) fix this and (b) move as much of this code as we can into utilities.

naliboff:
@jdannberg - Good catch! Should we do a fix first following how the viscoelastic model works and then move the code into utilities in a separate PR later on? I'm happy to do the fix (a) if you have not already started.

jdannberg:
Yes, that would be great, thank you! (And I think for (b) we already have a separate issue, it just reminded me that we should really do this...)

jdannberg:
This was fixed by #2580.

PR #2580, opened 2018-07-22
Update viscoplastic FEValues quadrature

naliboff:
Fix for the `visco_plastic` quadrature formula assumption in `FEValues`, which addresses #2579. 

@jdannberg - Let me know if this is not what you had in mind and thanks again for catching this!


gassmoeller:
This looks reasonable to me, but I will let @jdannberg take a look.

PR #2584, opened 2018-07-26
Do not assemble stokes boundary assemblers on periodic faces

jdannberg:
In the assembly, where we assemble the boundary terms for the stokes assemblers, we do not check for periodic boundaries, so boundary terms are assembled on periodic faces too. This is not a problem for Stokes or traction boundary conditions, because for these terms we have Asserts that make sure no velocity/traction boundary conditions are prescribed on periodic boundaries. However, for other boundary terms from additional assemblers, we probably don't want them to be assembled on periodic faces. Or is there any situation where one might want to do that, and the assemblers should check for that individually?

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -597,7 +597,7 @@ namespace aspect
         // then also work on possible face terms. if necessary, initialize
         // the material model data on faces
         for (unsigned int face_number=0; face_number<GeometryInfo<dim>::faces_per_cell; ++face_number)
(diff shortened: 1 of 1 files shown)


gassmoeller:
This makes sense and is consistent with what we do for the advection system. A face at a periodic boundary does not have boundary face terms (if at all it has interior face terms, but we have no such thing for the Stokes equation at the moment).

bangerth:
For the record, this is indeed the correct solution.

PR #578, opened 2015-08-18
fix log_average FPE

tjhei:
Log averaging generates FPE when encountering 0 values. We can not get
rid of all of them (compressibilities, entropy_derivative_pressure,
...), so handle 0 appropriately here.

PR #2585, opened 2018-07-26
add an adiabatic temperature gradient to the flow-law in visco_plastic

mibillen:
This option is being added because the choice of activation volume depends on whether the temperature field includes a temperature gradient in depth. Most incompressible models have constant temperature in depth, and so the activation volume needs to be lower compared to models that are compressible (or extended boussinesq). To make comparison of models easier, this option allows the user to choose the activation volume that is appropriate for a earth-like model, and then add the temperature gradient in the flow-law.  The default value is zero, so there is no change to the model-defined temperature.


include/aspect/material_model/visco_plastic.h
```diff
@@ -319,6 +319,11 @@ namespace aspect
         */
         double max_yield_strength;
 
+        /**
+         * temperature gradient added to temperature used in the flow law.
+         */
+        double temperature_adiabat_for_viscosity;
```
naliboff: On it's own (e.g., excluding comments), the name `temperature_adiabat_for_viscosity` implies (to me) that the value is the absolute adiabatic temperature at a given point, rather than the temperature gradient.

I think renaming the variable to something like `adiabatic_temperature_gradient_for_viscosity` (or something similar), would be a bit clearer and could be deciphered even without a comment.

source/material_model/visco_plastic.cc
```diff
@@ -1124,6 +1137,10 @@ namespace aspect
 
           // Limit maximum value of the drucker-prager yield stress
           max_yield_strength = prm.get_double("Maximum yield stress");
+
+          // Include an adiabat temperature gradient in flow laws
+          adiabatic_temperature_gradient_for_viscosity = prm.get_double("Adiabat temperature gradient for viscosity");
+
```
gassmoeller: Please add the following lines below this line:
```
if (this->get_heating_model_manager().adiabatic_heating_enabled())
AssertThrow (adiabatic_temperature_gradient_for_viscosity == 0.0,
  ExcMessage("If adiabatic heating is enabled you should not add another adiabatic gradient 
                         to the temperature for computing the viscosity, because the ambient 
                         temperature profile already includes the adiabatic gradient.")
```

It will make sure that if you use the extended Boussinesq approximation you are not allowed to set this parameter.

source/material_model/visco_plastic.cc
```diff
@@ -149,6 +149,11 @@ namespace aspect
                                std::max(std::sqrt(std::fabs(second_invariant(deviator(strain_rate)))),
                                         min_strain_rate) );
 
+      // Choice of activation volume depends on whether there is an adiabatic temperature
+      // gradient used when calculating the viscosity. This allows the same activation volume
+      // to be used in incompressible and compressible models. Default value is 0. Units [K/Pa]
```
gassmoeller: Please remove the default value and unit in the comment, this has been documented in the function below (and if we ever change this, we will likely forget to update this comment).

source/material_model/visco_plastic.cc
```diff
@@ -916,6 +921,14 @@ namespace aspect
                              "is not automatically used. Values of 100e6--1000e6 $Pa$ have been used "
                              "in previous models. Units: $Pa$");
 
+          // Temperature in viscosity laws to include an adiabat (note units of K/Pa)
+          prm.declare_entry ("Adiabat temperature gradient for viscosity", "0.0", Patterns::Double(0),
+                             "Add an adiabatic temperature gradient to the temperature used in the flow law "
+                             "so that the activation volume is consistent with what one would use in a "
+                             "earth-like (compressible) model. Default is set so this is off. "
+                             "Using a pressure gradient of 32436 Pa/m, then a value of "
```
gassmoeller: Could you add a note here that this is a linear approximation of the real adiabatic gradient? In practice that will probably be good enough for the upper mantle, but for whole mantle convection models this would not be very helpful (there one would use the real adiabatic gradient instead).

source/material_model/visco_plastic.cc
mibillen: Thanks for writing this out for me. I've added it.

source/material_model/visco_plastic.cc
mibillen: okay. done.

source/material_model/visco_plastic.cc
mibillen: Yes, this is a good point and I have added a note to the description.

mibillen:
@gassmoeller  - I'm unclear about what I need to do at this point. I was expecting to need to patch the test results, but I'm unsure about what the  failed test means. 
@naliboff it also looks like you need to approve the requested changes since you asked for specific changes (am I understanding what GIT is telling me correctly).

source/material_model/visco_plastic.cc
```diff
@@ -1140,7 +1142,11 @@ namespace aspect
 
           // Include an adiabat temperature gradient in flow laws
           adiabatic_temperature_gradient_for_viscosity = prm.get_double("Adiabat temperature gradient for viscosity");
-
+          if (this->get_heating_model_manager().adiabatic_heating_enabled())
+              AssertThrow (adiabatic_temperature_gradient_for_viscosity == 0.0,
+              ExcMessage("If adiabatic heating is enabled you should not add another adiabatic gradient 
+                         to the temperature for computing the viscosity, because the ambient 
+                         temperature profile already includes the adiabatic gradient.")
```
naliboff: It looks like there is a missing close parenthesis and semicolon ( `);` ) at the end of this statement.

naliboff: The compiler error is also related to missing quotation marks. The code should read as follows:
```
          if (this->get_heating_model_manager().adiabatic_heating_enabled())
               AssertThrow (adiabatic_temperature_gradient_for_viscosity == 0.0,
               ExcMessage("If adiabatic heating is enabled you should not add another adiabatic gradient" 
                          "to the temperature for computing the viscosity, because the ambient" 
                          "temperature profile already includes the adiabatic gradient.")
```

mibillen:
@naliboff thanks for finding those typos - I should have recompiled after pasting in the lines that Rene gave me. Sorry I forgot to do that. This time, I recompiled and reran my tests and checked that
the solution files are the same.

mibillen:
@gassmoeller I'm unclear on what you mean by updating the test results. There is not a changes-test-results.diff listed when I click on Artifacts on the jenkins page.  The test results that are in the pull request are correct (the results haven't changed, and I checked that those files are correct). Am I missing something?

mibillen:
Kept looking around and found the changes-gcc.diff file and saw that this was changing things in the file that are related to running in the tester (like different output directory name). So I used patch to update the two test files.

mibillen:
There is an error in running the test for the visco_plastic_adiabat_temperature, but it has something to do with a perl script, so I don't know how/what to do to fix this:

571/612 Test #578: visco_plastic_adiabat_temperature ..................................***Failed    0.24 sec
*** "Test visco_plastic_adiabat_temperature failed": ***
[1/2] Generating output-visco_plastic_adiabat_temperature/solution-00000.0000.gnuplot.notime
FAILED: output-visco_plastic_adiabat_temperature/solution-00000.0000.gnuplot.notime 
cd /home/dealii/build-gcc/tests && /usr/bin/perl /home/dealii/source/tests/normalize.pl /home/dealii/build-gcc/tests/output-visco_plastic_adiabat_temperature/solution-00000.0000.gnuplot /home/dealii/source >/home/dealii/build-gcc/tests/output-visco_plastic_adiabat_temperature/solution-00000.0000.gnuplot.notime
Can't read file: No such file or directory at /home/dealii/source/tests/normalize.pl line 21.
ninja: build stopped: subcommand failed.

CMake Error at /home/dealii/source/tests/run_test.cmake:14 (MESSAGE):
  *** Test aborted.

gassmoeller:
Hi Magali,
the test fails, because it does not find the file output-..../solution-00000.0000.gnuplot in the test output. The reason for that is that the file name is actually output-..../solution/solution-00000.0000.gnuplot. You just need to move the file in your tests directory into the correct subfolder ('solution') then the test should succeed.

mibillen:
I'm so sorry about missing that... I have fixed this.

gassmoeller:
No need to be sorry, it was somewhat tricky to find.

Looks good now, I will just wait for the last tester and then merge.

Issue #2551, opened 2018-07-04
Fix invalid references in manual.tex


bangerth:
There are a bunch of invalid references when compiling the manual, including but not limited to things like this:
```
LaTeX Warning: Hyper reference `parameters:Material_20model/Simpler_20with_20cr
ust_20model' on page 97 undefined on input line 3.

[97 <./cookbooks/free_surface/free_surface_blob.png> <./cookbooks/free_surface/
free_surface_topography.png>]

LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Reference density' on page 98 undefined on input line 4.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Reference specific heat' on page 98 undefined on input line 5.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Reference temperature' on page 98 undefined on input line 6.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Thermal conductivity' on page 98 undefined on input line 7.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Thermal expansion coefficient' on page 98 undefined on input line 8.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Lower viscosity' on page 98 undefined on input line 13.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Upper viscosity' on page 98 undefined on input line 14.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Jump height' on page 98 undefined on input line 15.
LaTeX Warning: Hyper reference `parameters:Material_20model/Simpler_20with_20cr
ust_20model' on page 97 undefined on input line 3.

[97 <./cookbooks/free_surface/free_surface_blob.png> <./cookbooks/free_surface/
free_surface_topography.png>]

LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Reference density' on page 98 undefined on input line 4.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Reference specific heat' on page 98 undefined on input line 5.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
del/Reference temperature' on page 98 undefined on input line 6.


LaTeX Warning: Hyper reference `parameters:Material model/Simpler with crust mo
[...]

PR #2588, opened 2018-07-29
Fix manual references

alarshi:
This PR addresses the issue #2551.
All but one of the warnings for invalid references were for parameters used in benchmarks and cookbooks source files, which I do not know how to fix. I commented a sentence in the manual.tex which refers to an older parameter in Gplates cookbook.

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -7610,10 +7610,6 @@ movements \textit{below} a lithosphere model. All plugin specific options may be
 section~\ref{parameters:Boundary_20velocity_20model}. Possible options include the data directory
 and file name of the velocity file/files, the time step (in model units, mostly seconds or years depending on the 
 ``\texttt{Use years in output instead of seconds}'' flag) and the points that define the 2D plane.
-%The parameter ``\texttt{Interpolation width}'' is used to smooth the provided velocity files by
-%a moving average filter. All velocity data points within this distance are averaged to determine the
-%actual boundary velocity at a certain mesh point. This parameter is usually set to 0 (no interpolation, use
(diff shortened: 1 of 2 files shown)


doc/manual/manual.tex
```diff
@@ -7610,10 +7610,10 @@ \subsubsection{Using reconstructed surface velocities by GPlates}
 section~\ref{parameters:Boundary_20velocity_20model}. Possible options include the data directory
 and file name of the velocity file/files, the time step (in model units, mostly seconds or years depending on the 
 ``\texttt{Use years in output instead of seconds}'' flag) and the points that define the 2D plane.
-The parameter ``\texttt{Interpolation width}'' is used to smooth the provided velocity files by
-a moving average filter. All velocity data points within this distance are averaged to determine the
-actual boundary velocity at a certain mesh point. This parameter is usually set to 0 (no interpolation, use
-nearest velocity point data) and is only needed in case the original setting is unstable or slowly converging.
+%The parameter ``\texttt{Interpolation width}'' is used to smooth the provided velocity files by
```
tjhei: It is not a good idea to comment out text like this. If it is no longer relevant, we should delete it. I don't understand why this parameter got removed. Maybe @gassmoeller remembers?

gassmoeller: Yes we should delete this whole sentence. The parameter was removed a while ago when GPlates implemented this ability (so users can now select in GPlates whether they want their velocities to be smoothened. I apparently forgot to update the manual when the parameter was removed.

alarshi: I understand, I have removed the sentence.

tjhei:
testers are unhappy but this is just documentation...

PR #2589, opened 2018-07-29
Fix bug steinberger material

gassmoeller:
As described in the changelog entry. A pretty annoying bug (wrong velocity and pressure solution without warning), but only for a very specific combination of parameters. The test provokes exactly this bug, and was wrong before this fix.


naliboff:
@gassmoeller - What is the difference between the continuous-integration/jenkins/pr-head and continuous-integration/jenkins/pr-merge testers? Are they just different machines? If so and "pr-head" is not available for some reason, this is ready to merge.

gassmoeller:
pr-head is @tjhei jenkins tester and pr-merge is the jenkins tester we have at CIG. Timo's machine does currently not report its errors, and since the old tester and the CIG tester are happy I am ignoring the failure (both jenkins testers are currently in a experimental state).

source/material_model/steinberger.cc
```diff
@@ -393,7 +393,7 @@ namespace aspect
         {
           dRhodp = material_lookup[0]->dRhodp(temperature,pressure);
         }
-      if (material_lookup.size() == compositional_fields.size() + 1)
+      else if (material_lookup.size() == compositional_fields.size() + 1)
```
bangerth: damn!

PR #2590, opened 2018-07-31
update section on how to write a test

jdannberg:
This is based on feedback I got from @cedrict at the hackathon. Before, it was not really clear in which directory the commands outlined in the manual should be run, so I made the instructions more clear.

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -12484,7 +12484,7 @@ The test suite is run using the \texttt{ctest} program that comes with
 compiled \aspect{}.
 
 After running \texttt{cmake} and then compiling \aspect{}, you can run the
-testsuite by saying \texttt{ctest}. By default, this will only run a small
+testsuite by using the command \texttt{ctest} in your build directory. By default, this will only run a small
 subset of all tests given that both setting up all tests (several hundred) and
 running them takes a non-trivial amount of time. To set up the full test suite,
 you can run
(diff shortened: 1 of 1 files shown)


doc/manual/manual.tex
```diff
@@ -12484,7 +12484,7 @@ \subsubsection{Running tests}
 compiled \aspect{}.
 
 After running \texttt{cmake} and then compiling \aspect{}, you can run the
-testsuite by saying \texttt{ctest}. By default, this will only run a small
+testsuite by using the command \texttt{ctest} in your build directory. By default, this will only run a small
 subset of all tests given that both setting up all tests (several hundred) and
 running them takes a non-trivial amount of time. To set up the full test suite,
 you can run
```
naliboff: Missing text after "you can run" ?

jdannberg: There is text after "you can run", it's just a command:

```
you can run
\begin{lstlisting}[frame=single,language=ksh]
    make setup_tests
\end{lstlisting}
in the build directory. 
```
I don't know why these lines are not shown on github, maybe the manual.tex file is too long.

naliboff: Weird. If that text is present on your local copy of manual.tex, it must indeed be a github issue?

naliboff: Makes sense. I just gave this a second read through and it looks good to go. I'm not sure why the jenkins tester is not running, but for this case it does not really matter. Unless you would like another set of eyes on the changes, feel to merge when ready.

jdannberg: The jenkins tester is not running because someone has to start it and I don't have the permission to do that... but as you pointed out, all changes are in the manual.tex, so it can't break the tester anyway.

jdannberg: And I believe if you think it's ready to merge, you should just be brave and hit the merge bottun :grinning:

naliboff: merged 😄

PR #2591, opened 2018-08-01
manual: add missing bib entry

tjhei:
see #2590


tjhei:
oops. I meant #2542

jdannberg:
Ah, that makes sense.

PR #2592, opened 2018-08-01
update parameters, fix manual

tjhei:


gassmoeller:
Alright, looks good. It seems like two of the testers are having problems, but this is only a documentation change anyway, and the cig one is doing ok, so this is good to go.

source/initial_temperature/patch_on_S40RTS.cc
```diff
@@ -186,7 +186,7 @@ namespace aspect
                                               "contain the number of grid points in each dimension as "
                                               "for example '# POINTS: 3 3 3'. "
                                               "The order of the data columns has to be "
-                                              " `x', `y', `z', 'Vs Perturbation [%]' in a 3d model, which means that "
+                                              " `x', `y', `z', 'Vs Perturbation' in a 3d model, which means that "
```
bangerth: do we say anywhere that this column is interpreted as a percentage perturbation?

PR #2596, opened 2018-08-08
www: fixes for citing.html

tjhei:


jdannberg:
Thanks! What about the `Cite!` button in index.html (line 116)? That should also get the href="citing.html?src=www&ver=release", right?

tjhei:
yes, fixed.

PR #2598, opened 2018-08-08
Add option to output gravity anomalies to geoid postprocessor

jaustermann:
Calculating the free-air gravity anomalies from the geoid postprocessor is just a simple scaling in the spectral space. I've added this scaling and the option to output gravity anomalies on a grid or as spherical harmonic coefficients. The result should be comparable / complementary to the gravity_point_values postprocessor by @LudovicJnnt but limited to the spherical shell geometry and points at the surface. 
As a comparison that this is working I compared results to a solution from a spectral code. The run setup is S20RTS with constant scaling from vs to density (0.15) and a simple isoviscous material (the parameter file is in benchmarks/S20RTS_gravity_anomaly/S20RTS_gravity_anomaly.prm). I first compared the surface and CMB topography, which are very similar and then the geoid, which is a little smaller in magnitude, particularly at long wavelengths. This difference might be because the spectral code includes a self-consistent gravity calculation that accounts for internal gravity perturbations and surface deflections. Aspect just uses constant g.
The gravity anomalies compare well to the results from the spectral code but are also slightly smaller in magnitude, which is expected since they directly depend on the geoid. Attached a figure showing the comparison:

screen shot 2018-08-08 at 1 41 39 pm

@LudovicJnnt Maybe it's worth doing another comparison with your setup? I have added the parameter file (and geoid and gravity anomaly output) in the benchmarks folder. It's run at quite high resolution, maybe such a high resolution is not necessary. Could you have a look at whether you get comparable results? Feel free to also run a different setup. I don't think this parameter file and output need to be pushed into aspect but added the commit it here to transfer the files.


LudovicJnnt:
@jaustermann I will run the same setup and/or add a finer resolution for the surface and use the gravity_point_values postprocessor and share the results. Looking good!

bangerth:
I think that if you believe this sort of experiment is something other people may want to do as well, or that that is something they could learn something from regarding setting up input files, then this would still make for a nice cookbook!

source/postprocess/geoid.cc
```diff
@@ -739,6 +739,125 @@ namespace aspect
...
+                         ?
+                         surface_cell_spherical_coordinates.at(i).second*(180./numbers::PI)
+                         :
+                         surface_cell_spherical_coordinates.at(i).second*(180./numbers::PI) - 360.);
+                  stored_values_lon_lat.emplace_back(std::make_pair(lon,lat),gravity_anomaly.at(i));
+                }
+              // Write the solution to the stream output
+              for (unsigned int i=0; i<stored_values_lon_lat.size(); ++i)
+                {
+                  output_gravity_anomaly << stored_values_lon_lat.at(i).first.first
```
gassmoeller: Could you not just compute the geographical coordinate for every point inside this loop and immediately output it point-by-point? Then you would not need to create a new container holding all the converted coordinates and gravity values (you could remove the whole for-loop above.

source/postprocess/geoid.cc
```diff
@@ -739,6 +739,125 @@ namespace aspect
...
+            {
+              std::vector<std::pair<Point<dim>,double> > stored_values_xyz;
+              for (unsigned int i=0; i<surface_cell_locations.size(); ++i)
+                {
+                  stored_values_xyz.push_back(std::make_pair(surface_cell_locations.at(i),gravity_anomaly.at(i)));
+                }
+              // Write the solution to the stream output
+              for (unsigned int i=0; i<stored_values_xyz.size(); ++i)
+                {
+                  output_gravity_anomaly << stored_values_xyz.at(i).first
```
gassmoeller: Same comment here, the loop above is not really necessay, you could do all that inside this loop and save the additional container.

LudovicJnnt:
@jaustermann and @gassmoeller: I succeeded to get a decent gravity acceleration map from the gravity post-processor (and it requires a lot of resolution in the surface (GMR 5 + AMR 2 for the 300 first km at surface). However, I am not able to get the gravity anomalies because I need data of "observed gravity" to compare my prediction with. (right?)
How does it work using the geoid post-processor to get gravity anomalies, and only for the mantle? I remember it was just a derivative from the geoid (?) Somehow now I am confused how gravity anomalies can be obtained from a geoid because it is an anomaly compared to - what? 
Or maybe I am missing a point here and need to read more literature about it.

gassmoeller:
Ludo: It is an anomaly compared to a spherical shell of a constant density. For your comparison you should be able to subtract the average gravity from the points of a shell at the surface and hopefully the results will agree with the map from the geoid postprocessor.

Jacky: Sounds good, I am a bit surprised that the difference between the geoid postprocessor and the spectral code is so big (it looks like on the order of 50%)? This would suggest something else than self-gravity (as density only changes by a few percent and consequently gravity variations should be small). Maybe the spectral code used a density profile from PREM (i.e. a depth-dependent density), while ASPECT's run was incompressible? Just speculating though.

LudovicJnnt:
Rene: thanks for your answer, below it is what the gravity postprocessor predicts for gravity anomalies. The anomalies are in the same order of magnitude as the two other methods although they are look quite different. I reused the input file of Jacky and added AMR of 2 for the surface above 6300km (otherwise I had just noise). You can still see the noise from the mesh with this setup. 
I am now afraid that the difference of gravity anomalies between the two methods may be due to not enough resolution either in depth or at the surface. It is a pity that the gravity postprocessor requires very high resolution to obtain mildly accurate results. 
Also, the predictions below are obtained at height 10 km above the surface (initially it avoids the mesh to have to much effect on the results - which has to be proven - test on the go) 

gravity_anomaly

jaustermann:
@LudovicJnnt: Your results are quite different and look very similar to dynamic topography (not the magnitude but the pattern - see also plots below). Is there any chance you're only calculating gravity anomalies due to the surface topography (I'm wouldn't think that's the case, but just thinking of possible issues)? I don't think it's a resolution issue, mine are global refinement 5, which is pretty good. Have you done any analytical benchmarks or comparison to other papers (other than just the full sphere calculation to check that you get ~9.81)?

@gassmoeller: I've asked Mark Hoggard - who is running the spectral code in this comparison - to remove self-gravity, i.e. remove delta g x rho in the momentum equation. Below are the results: The aspect output is now very close to the spectral code output. It seems to show that even though the density changes are only on the percent level, they do affect the magnitude of the geoid at low wavelength. They are less important for gravity anomalies and dynamic topography. I also double checked that the background density is the same (see details on the run setup below). 
I think I've addressed all your comments including making the output SI. I'm happy to put something in the manual about this (maybe adding it to cookbook 5.3.4?) if you think that's useful.

screen shot 2018-10-03 at 10 52 13 am
screen shot 2018-10-03 at 10 52 38 am
screen shot 2018-10-03 at 10 53 01 am


Just fyi - since you asked about the density profile, here is the spectral code setup:
Generate density field using S20RTS Vs % variations, scaled to density % variations using constant factor of 0.15, and converted into absolute density variations using constant 3300 kg/m3 
Run instantaneous flow with compressibility turned off, the value of g is set to constant 10.0 m/s2, background density = 3300 kg/m3, core density = 9900 kg/m3, surface fluid density = 0 kg/m3 
One difference between the two codes is resolution, especially vertical resolution, however we tried to make those as similar as possible.

gassmoeller:
Oh, maybe one comment: Since the benchmark does not have a lot of explanation on how to compare it  to published results or analytical results, this feels more like a testcase than a benchmark. Could you move the parameter file and results into the tests/ directory?

Shangxin-Liu:
Hi Jacky, Rene, and Ludo,

Allow my late coming into this discussion. This comparison is a nice test between the geoid and gravity anomaly post-processor. I agree with Jacky that Ludo's results are dominated by the effect of the topography. Ludo, you mentioned about that you added AMR 2 to the shallower region. I feel that in your model the very fine resolution at the shallower region makes the surface topography dominating the results somehow. Maybe worth a double check into the code that whether the calculation is perturbed by AMR. Jacky's figures are pretty sound to prove the accuracy of the dynamic topography and geoid code. I'd like to add one more point that, as Rene knew last year, I already benchmarked my geoid code with the semi-analytic kernels and it agree very well with the spectral code generally within 1% error in 64 layers. I took a snapshot from my document below:

screen shot 2018-10-03 at 8 12 26 pm

So the accuracy of my geoid code and hence the derived gravity anomalies should be reliable.

I also agree with Rene that this pull request should be better regarded as a test case instead of a benchmark. I'm happy to add my response function (kernel) benchmarks into ASPECT manual later, which should be a strong support for the accuracy of ASPECT's dynamic topography and geoid calculation.

tjhei:
> Since the benchmark does not have a lot of explanation on how to compare it to published results or analytical results, this feels more like a testcase than a benchmark. Could you move the parameter file and results into the tests/ directory?

I think it would be a shame if this is hidden as a test only. I think a minimal readme.txt in the same folder or a short section in the manual with a picture or two would be great.

LudovicJnnt:
Dear all,

Cedric is going to run his gravity post-processor on S20RTS with his software to check if him and I got similar results on gravity acceleration and gravity potential. If these are the same, then I will be able to check how we can predict gravity anomalies from gravity acceleration.

The similarity between my gravity anomalies and dynamic topography may be just a coincidence: in fact there is rather a similarity between my gravity anomalies with the shear-wave variations of S20RTS at depth 100km, which makes me think that the gravity anomalies I obtain via my _own calculation_* are more reflecting the shallower region of S20RTS than the whole domain. This can be justified because a mass near the surface (therefore closer to the satellite) has more attraction contribution than a mass at the bottom of the mantle. 
I do not really understand why or how surface topography would be involved here (or at least I am not sure to understand what you mean by surface topography) - I just reused the same parameter file than Jacky but added +2 AMR. I cannot calculate gravity directly at the surface but at certain height above the surface but I do not think 20km altitude difference can generate such a difference in our gravity anomalies. 

*It is highly possible that I am indeed not calculating the same thing than you. Maybe I should explain again how the gravity post-processor works: for a satellite at altitude x-km above the surface I compute gravity acceleration using Newton's integrals. Basically, I compute at one point-satellite the local gravity acceleration contribution from a cell and then sum gravity acceleration of all the cells. This is repeated for each position of a satellite. Once I have gravity acceleration, I subtract it with the mean gravity of all my satellites (maybe here there is an issue). At the moment, I am running a model where I compute gravity directly on relative density - which should give me gravity anomalies directly from the post-processor. 

I will wait for Cedric to redo our experiments using his code and see what the outcome is. So unfortunately, there is no analytical benchmark done on the gravity post-processor other than the full sphere. I am still trying to understand what is gravity anomalies from the geoid post-processor (surface gravity anomalies?). I hope my explanation of how I proceeded may help you understand what I am doing differently?

Shangxin-Liu:
Ludo: So I read your gravity code a bit and also from your description, it seems that your gravity post-processor calculates the gravity from the contribution of only density anomaly (no topography). The perturbation to the gravitational potential is determined by Poisson's equation and the density perturbation term in the equation include both the contribution from the density anomaly (buoyancy part) and surface & CMB topography. This is what is done in my geoid code. There is one parameter in my geoid code called "Also output the spherical harmonic coefficients of density anomaly contribution", I think another useful comparison is to turn this parameter on and calculate the geoid and the derived gravity anomaly from only density contribution, and then compare this to the output from your code. This comparison may yield the same results between the two post-processors.

I also feel that the definition of the gravity anomaly sometime causes misleading. The common definition of the free-air gravity anomaly is the gravity anomaly after the free-air correction. I saw the equation (97) in Forte 2007, it's called the "free-air gravity anomaly from geoid anomaly. But in some other papers, I also noticed that it's directly called "gravity anomaly" (no free-air). I still need to better understand this to make the definition consistent with each other in physics. Maybe Jacky can give a comment on this?

LudovicJnnt:
@Shangxin-Liu : thank you for your explanations. So indeed we were not calculating the same thing. At the moment I can only compute gravity anomalies from density contribution (graph below). Cedric obtained the same results than me with his code. If there is a way to calculate gravity anomalies from only density contribution with the geoid post-processor, it would be great to see the differences. 

**The results in the figure that I presented above (which looked like dynamic topography) had a mistake in the norm calculation; it was not reflecting gravity anomalies but a sort of gravity acceleration "difference".  

gravity_anomaly_s20rts

Shangxin-Liu:
Ludo: If you turn on the parameter "Also output the spherical harmonic coefficients of density anomaly contribution" in my geoid code, it can output the geoid only from the density perturbation. But it's necessary to further sum over them back into the Cartesian domain. I think this will output the same thing as calculated by your gravity code. @LudovicJnnt 

One more point I'm a little bit concerned is how to describe this. Because the gravity anomaly scaled from geoid code clearly includes the topography contribution whereas the gravity post-processor does not. Maybe it's better to clarify this in the manual to avoid confusion.

bangerth:
Something went wrong with the tester. Can you just rebase to current master and force push again? This should re-start the tester.

jaustermann:
Thanks Wolfgang for checking. I rebased and pushed, testers are running.

gassmoeller:
Looks all good. No worries about the delay, we all have a lot of other things to take care of. With the added description and figure the `spectral-comparison` might actually be better in the benchmark section (since we now have a test with the geoid test already). But I can take care of that in a follow-up, it is just a minor change. Nice new feature :+1:

PR #2460, opened 2018-06-26
Remove cxx11 includes

gassmoeller:
One step towards #2375. This is work in progress, I need to go over the code and look for broken indentation, I just want to see if tests fail as well.

PR #2599, opened 2018-08-08
Fix update source file script

gassmoeller:
This was broken since #2460. By starting from the backup file $file.bak the first changes that were made by the sed script were always ignored. Instead the perl scripts need to be applied to the already partially upgraded file $file.


tjhei:
:+1:

Issue #1462, opened 2017-04-27
cmake: remove ctest dependency to aspect target


tjhei:
We currently have a dependency from each test target to ``aspect``. The motivation is that a test needs to re-run whenever the ``aspect`` binary changes (because it was rebuilt). Sadly, cmake does not add a dependency to the binary ``aspect`` but to the target to build ``aspect`` instead. This means that for every test cmake will check all dependencies recursively over and over again, which is very slow.

Another consequence of this is that when you run ``ctest -j 8`` and aspect is out of date (maybe because one source file of aspect is newer), we try to compile ``aspect`` 8 times at the same time. This has brought down my machine a couple of times over the years.

This first problem (slowness) is not visible if you use ninja instead of make, because ninja is very efficient in checking dependencies. The problem is that using ninja for the testsuite is somewhat broken (we disable ninja for the deal.II testsuite, see https://github.com/dealii/dealii/pull/2767 ).

I am not sure how to tell cmake that a target should depend on the aspect binary instead of the aspect target. Maybe breaking the tests into a separate project like we do in deal.II is enough to make this work. This would have the advantage that we can reliably use ninja to compile again.

gassmoeller:
Sounds great, 7min is much better than 21. I do know too little about make to make suggestions about how to improve the 4s, but this seems like a nice speedup already. Wouldnt the tests need to be rerun for every pull request anyway, so that the 4s case usually only applies if you work on aspect on your machine? Or does the tester keep files between different PRs?

tjhei:
> Wouldnt the tests need to be rerun for every pull request anyway, so that the 4s case usually only applies if you work on aspect on your machine? Or does the tester keep files between different PRs?

Well, most important is for normal development, but the tester currently does the following:

1. make -j $NPROC
2. ctest -j $NPROC
3. ASPECT_GENERATE_REFERENCE_OUTPUT=1 ctest -j $NPROC

Step 3 will rerun all the tests a second time. The current tester uses ninja for step 2&3, which causes it to rerun the examples in 3) (taking the additional 7 minutes or so). I can fix this by using make, but then I am still paying 4s*400 tests / 16 cores = 1.5 minutes.

Maybe I can fold 2&3 into a single computation...

bangerth:
Would there be a way to create one sub-project per test, rather than a subproject for all tests? In that case, we could use `ninja` for them again. Or does `ctest` not like this?

tjhei:
[...]

PR #2604, opened 2018-08-09
Use post step to create test results instead of '|| true'

tjesser-ucdavis-edu:
Remove Prebuild Test stage, as CTest seems to use ninja anyway. Also use
Jenkins xunit to display test results in Jenkins.

I talked with @gassmoeller and he was unsure if there was another reason to use the Prebuild Test Stage. @tjhei Do you remember what, if anything, it was?

I reduced the thread count to half of the available cpus because even if we are only running one ASPECT Jenkins build, there may be other repos running tests on CIG infrastructure.


tjhei:
> Remove Prebuild Test stage, as CTest seems to use ninja anyway.

ctest uses ninja because we specify ``ASPECT_TEST_GENERATOR=Ninja`` for performance reasons. The prebuild is there for several reasons:
1. If you do ``ctest -j X`` ninja will be executed concurrently and each ninja process might claim up to N cores.
2. Concurrent ninja execution is not supported by ninja and will cause a race condition in the database generation. This causes sporadic rebuilds and errors.
3. Running one ninja instance is much better in fully utilizing the machine to build and run the tests than ctest.

> I reduced the thread count to half of the available cpus
> there may be other repos running tests on CIG infrastructure.

That doesn't make much sense to me. Why is a single machine used for more than one run?

tjhei:
so in summary: ``ASPECT_TEST_GENERATOR=Ninja`` is faster but very fragile and is not meant to be used without prebuilding.

tjesser-ucdavis-edu:
> If you do ctest -j X ninja will be executed concurrently and each ninja process might claim up to N cores.
Concurrent ninja execution is not supported by ninja and will cause a race condition in the database generation. This causes sporadic rebuilds and errors.

That seems to match what happened on the tester. In my opinion, that's a bug in CMake. I wonder if later versions of CMake have a fix. Thanks for the explanation.

> That doesn't make much sense to me. Why is a single machine used for more than one run?

Currently, the Kubernetes cluster that the CIG Jenkins is on only has one server node. You're probably right that decreasing build latency is more important than increasing build throughput, though, so I removed the limit.

tjhei:
> In my opinion, that's a bug in CMake

No, that is a known limitation of ninja. See https://github.com/geodynamics/aspect/issues/1462 when we introduced this.

tjesser-ucdavis-edu:
I can't access Timo's Jenkins, but it looks like the tests are getting added to the CIG Jenkins interface: https://singleb.cse.ucdavis.edu/blue/organizations/jenkins/aspect/detail/PR-2604/16/tests

None of them are failing, so I can't tell if the useful output associated with each test is getting lost because of the ninja prerun before ctest.

---

> No, that is a known limitation of ninja. See #1462 when we introduced this.

Ninja doesn't need multiple instances running to do parallel tests. Since CTest officially supports Ninja, it should know how to correctly run Ninja in parallel.

tjesser-ucdavis-edu:
@gassmoeller helped me test the interface for failing tests. You can see them here: https://singleb.cse.ucdavis.edu/blue/organizations/jenkins/aspect/detail/PR-2604/32/tests

I think this is ready for review/merge as long as the testers are happy.

tjesser-ucdavis-edu:
Oh, I forgot to mention, I squashed the commits into one since I was only working with one file.

Jenkinsfile.k8s
```diff
@@ -112,18 +98,35 @@ pipeline {
...
             export OMPI_MCA_btl=self,tcp
-            rm -f /home/dealii/build-gcc-fast/FAILED
-            cd /home/dealii/build-gcc-fast
-            ctest --output-on-failure -j4 || { touch FAILED; }
-            echo "Generating reference output..."
-            ninja generate_reference_output
-            cd $WORKSPACE
-            git checkout tests/CMakeLists.txt
+            cd /home/dealii/build-gcc-fast/tests
+            ninja -k 0 tests || true
```
bangerth: I have no idea what the rest of all of this does, but is this really what you want to do? It succeeds even if there are failures. Or are failures expected with the tests?

gassmoeller:
@bangerth: To answer your question with a very short answer: Yes, this is what we want to happen (it was there already in the old version in the 'Prebuild tests' stage). 
The long answer is: This file is now a masterpiece in arcane Kubernetes/Jenkins/Docker/Ninja/Ctest/MPI interaction syntax, and I have tried to explain the most obscure parts in the last commit that adds comments. Maybe you and @tjesser-ucdavis-edu could take a look and let me know if the comments clarify things enough to make it readable.

Jenkinsfile.k8s
```diff
@@ -78,7 +78,14 @@ pipeline {
             export NP=`grep -c ^processor /proc/cpuinfo`
             mkdir -p /home/dealii/build-gcc-fast
             cd /home/dealii/build-gcc-fast
+
+            # We need to add this option to mpirun, otherwise
+            # openmpi complains about running as root. We need
+            # to run as root, because we are inside a docker container,
+            # and the surrounding containers assume we have root access.
```
tjesser-ucdavis-edu: I'm pretty sure it's the Jenkins/Kubernetes interaction that requires running as root, and not the fact that it's a Docker container.

Jenkinsfile.k8s
```diff
@@ -89,41 +96,71 @@ pipeline {
...
-            git checkout tests/CMakeLists.txt
+
+           cd /home/dealii/build-gcc-fast
+
+            # Output the test results using ctest. Since
+            # the tests were prebuild in the previous shell
+            # command, this will be fast although it is not
+            # running in parallel.
+	    ctest \
+              --no-compress-output \
```
tjhei: You are no longer running ctest in parallel. This is not a big issue, but will slow down this step a little bit. Is this needed?

Jenkinsfile.k8s
```diff
@@ -89,41 +96,71 @@ pipeline {
...
-      options {timeout(time: 90, unit: 'MINUTES')}
-      steps {
-        container('dealii'){
+            # Let ninja prebuild the test libraries and run
+            # the tests to create the output files in parallel. We
+            # want this to always succeed, because it does not generate
+            # useful output (we do this further down using 'ctest', however
+            # ctest can not run ninja in parallel, so this is the
+            # most efficient way to build the tests).
+            ninja -k 0 tests || true
```
tjhei: We used to have this in a separate step. This has the advantage that the output won't show up as only the last failed step is typically shown. How does that work now?

tjhei:
Finally, your tester it not running on this PR. Why?

Jenkinsfile.k8s
tjesser-ucdavis-edu: On the latest build of this PR, the ninja test sub-step takes 15 minutes and this ctest sub-step takes 1 minute. It's not parallel because of the issues with ctest parallelizing ninja we previously discussed.

Basically, all ctest is doing here is checking that the tests are already completed and formatting them into a Jenkins readable .xml file.

Jenkinsfile.k8s
tjesser-ucdavis-edu: It's part of the "Run tests" step, but since it can't fail, the ctest sub-step will be displayed (unfurled?) in the Jenkins BlueOcean interface. (If it failed.)

tjesser-ucdavis-edu: It's part of the "Run tests" step, but since it can't fail, the ctest sub-step will be displayed (unfurled?) in the Jenkins BlueOcean interface. (If it failed.)

tjesser-ucdavis-edu:
@tjhei 
> Finally, your tester it not running on this PR. Why?

Unfortunately, our Kubernetes cluster has been having DNS issues that cause it to sometimes not be able to communicate with github.com. What that means is that builds that finish, builds that start, or often during cloning for the build container, Jenkins will not be able to connect to Github. We've been trying different things, but haven't been able to diagnose the issue yet.

The build was completed and passed: https://singleb.cse.ucdavis.edu/blue/organizations/jenkins/aspect/detail/PR-2604/37/pipeline/66

Jenkinsfile.k8s
```diff
@@ -78,7 +78,14 @@ pipeline {
             export NP=`grep -c ^processor /proc/cpuinfo`
             mkdir -p /home/dealii/build-gcc-fast
             cd /home/dealii/build-gcc-fast
+
+            # We need to add this option to mpirun, otherwise
+            # openmpi complains about running as root. We need
+            # to run as root, because the surrounding Jenkins container
+            # assumes we have root access.
```
bangerth: @#$%^&

Who designed that system? Even in containers, running as root is a poor way of dealing with things...

Jenkinsfile.k8s
bangerth: I see now. I don't think I had realized quite what `ninja tests` does exactly.

Jenkinsfile.k8s
gassmoeller: From what I have read in forums discussing containerization, there are quite some people who would disagree with you on that :smirk:. 

Also I want to point out that of course in theory you can use normal users in containers, only our combination (having a jenkins image with a jenkins user id of 10000) that mounts a directory into the dealii image with the dealii user, bundled in containers that run inside kubernetes somewhere produces the problem. Tyler and I spend the better part of two days figuring out a workaround (synchronizing user-ids between the containers, setting access rights on the host filesystem, switching users inside the container), but the only solution we could find was to run as root.

I have no preference about the whole running as root in containers issue, but what I really find a bit frustrating here is that the openmpi project was informed about the complications that their decision to create this command-line option causes, but so far has not decided on a future course of action, because they want to make it intentionally hard to run MPI as root (see also issue 4451 of open-mpi/ompi, I am on purpose not linking to the issue to not get dragged into the discussion :smile:). I can understand their arguments, but it just leads to more conflicts for anyone who wants to run openmpi inside a container image that is designed to run as root. And using an openmpi specific command-line flag that is not necessary in and not even supported by other mpi implementations does not help the problem either.

Jenkinsfile.k8s
gassmoeller: @tjhei: You can see the current output here: https://singleb.cse.ucdavis.edu/blue/organizations/jenkins/aspect/detail/PR-2604/41/pipeline
The prebuild output does only show up if you click on the first line of the tests stage. But the best place to see the test results is now here:
https://singleb.cse.ucdavis.edu/blue/organizations/jenkins/aspect/detail/PR-2604/41/tests
It will show all failed tests at the top, including their output, which is even better than before (where we had to search through the ctest output of all tests for the output of the failed tests).

gassmoeller:
I am reasonably happy with the current state. Did we address your comments sufficiently?

Jenkinsfile.k8s
bangerth: Oh, I hope it was clear that I didn't want to criticize you for this setup. I understood from your previous description that that choice was made for you.

It's exactly the kind of issues you describe why I think running anything as root is a bad idea. Apart from simply making it too easy for anyone to shoot themselves into the foot. But that's definitely not your fault!

bangerth: I left a comment on the open-mpi issue as well.

gassmoeller: Don't worry, I was aware that you did not criticize our setup but the docker practice of running as root. I just needed to vent a bit of frustration about the fact that the whole jenkins setup got so much more complicated than anticipated. I am curious to see what happens in open-mpi :smile:.

bangerth: This, now, is a little bit funny. I commented on the openmpi issue 4451 with something I didn't think was *completely* serious (though also not entirely stupid), but that's what they implemented and just merged :-)

gassmoeller: I had to laugh when I saw the discussion :smile:. And I had to read your proposal twice to understand what you meant, so it is definitely complicated enough that nobody does it by accident.

bangerth: Apparently "complicated" is something they value ;-)

PR #2607, opened 2018-08-14
Allow negative scale factor for boundary velocity

### LINKED ISSUE #2605: Changing the sign of ascii gravity and boundary velocity
The ascii data plugins for gravity and boundary velocity don't allow you to change the sign (for backward advection) without making a separate input file, since the built-in scale factor is required to be >= 0.

Checking that the scale factor is non-negative makes sense for other ascii data plugins, but it would be nice to have a way of inverting gravity and boundary velocity within the parameter file.

bartniday:
Fixes #2605.

### DIFF
--- a/source/boundary_velocity/gplates.cc
+++ b/source/boundary_velocity/gplates.cc
@@ -823,7 +823,7 @@ namespace aspect
                              "in the input file, this number is either interpreted as seconds or as years. "
                              "The default is one million, i.e., either one million seconds or one million years.");
           prm.declare_entry ("Scale factor", "1",
-                             Patterns::Double (0),
+                             Patterns::Double (),
                              "Scalar factor, which is applied to the boundary velocity. "
                              "You might want to use this to scale the velocities to a "
                              "reference model (e.g. with free-slip boundary) or another "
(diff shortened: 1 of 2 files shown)


gassmoeller:
Thanks, this is exactly what I meant. Ready to merge if the testers are happy.

bangerth:
The tester seems to refuse to build this. Let's see if this still works:
/run-tests

tjesser-ucdavis-edu:
Neither of the webhooks are set up to alert about label changes:

webhooks

gassmoeller:
@tjesser-ucdavis-edu : I do not think that is the issue. The option you showed triggers when the labels itself are changed or a new label category is created, but the "Pull requests" option that is checked at the moment explicitly includes the "Pull request labeled/unlabeled" event. That suggests to me that our webhook currently triggers, but Jenkins does not recognize that it should rebuild a pull requests only because it has been labeled. Do you know if there is an option to do this? Otherwise we will just have to do it manually whenever we label a pull request (I might have forgotten to do it here).

tjesser-ucdavis-edu:
Yeah, you're right, my mistake.

We discussed this in person, but there does appear to be a plugin that can do that. Unfortunately, it seems to have a scary warning on the plugin page that it's still using unsafe plugin practices: https://plugins.jenkins.io/ghprb

It might not be affected anymore, as the last update was only three months ago: https://github.com/jenkinsci/ghprb-plugin. It's unclear whether the warning is out-of-date or not.

gassmoeller:
Ok, the tests succeeded (the jenkins tester are having trouble again, but at least the CIG one also succeeded, it just did not update the status correctly). Thanks for the fix Bart.

PR #2609, opened 2018-08-15
Add a termination criteria based on wall time.

SiqiZhang:
Many HPCs have wall time limitations for jobs (e.g. 24 hours). I created a termination criteria base on the wall time used. I thought it would be useful for those running long jobs on those HPCs have limited wall time.

### DIFF
--- a/include/aspect/termination_criteria/end_walltime.h
+++ b/include/aspect/termination_criteria/end_walltime.h
@@ -1,0 +1,86 @@
+/*
+  Copyright (C) 2011- 2018 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef __aspect__termination_criteria_end_walltime_h
+#define __aspect__termination_criteria_end_walltime_h
+
+#include <aspect/termination_criteria/interface.h>
+#include <aspect/simulator.h>
+
+namespace aspect
+{
+  namespace TerminationCriteria
+  {
+
+    /**
+     * A class that terminates the simulation when a specified end time is
+     * reached.
+     *
+     * @ingroup TerminationCriteria
+     */
+    template <int dim>
+    class EndWalltime : public Interface<dim>, public SimulatorAccess<dim>
+    {
+      public:
+
+        /**
+         * Evaluate this termination criterion.
+         *
+         * @return Whether to terminate the simulation (true) or continue
+         * (false).
+         */
+        virtual
+        bool
+        execute (void);
+
+        /**
+         * Declare the parameters this class takes through input files.
+         */
+        static
+        void
+        declare_parameters (ParameterHandler &prm);
+
+        /**
+         * Read the parameters this class declares from the parameter file.
+         */
+        virtual
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+        /**
+         * The maximum walltime duration in seconds. The program will be terminated
+         * once this value is reached.
+         */
+        unsigned int walltime_duration;
+
+        /**
+         * The start_walltime is not stored into checkpoint, so it will get reset at restart.
+         * Make start_walltime as a static variable that it gets initialized pretty much
+         * right away as the program starts, rather than several seconds in once we get
+         * to creating the plugin. It gives better estimate of the actual start time.
+         */
+        static std::time_t start_walltime;
+    };
+  }
+}
+
+#endif
--- a/source/termination_criteria/end_walltime.cc
+++ b/source/termination_criteria/end_walltime.cc
@@ -1,0 +1,80 @@
+/*
+  Copyright (C) 2011 - 2018 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#include <aspect/termination_criteria/end_walltime.h>
+
+namespace aspect
+{
+  namespace TerminationCriteria
+  {
+    // The start_walltime is made as static and initialized here to
+    // make sure it is initialized right way as the programe starts.
+    template <int dim>
+    std::time_t
+    EndWalltime<dim>::start_walltime = std::time(NULL);
+
+    template <int dim>
+    bool
+    EndWalltime<dim>::execute()
+    {
+      return ( (std::time(NULL)-start_walltime) >= (int)walltime_duration);
+    }
+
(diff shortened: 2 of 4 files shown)


source/termination_criteria/end_walltime.cc
```diff
@@ -0,0 +1,83 @@
...
+    EndWalltime<dim>::start_walltime = std::time(NULL);
+
+    template <int dim>
+    bool
+    EndWalltime<dim>::execute()
+    {
+      // Measure the time at process 0 to detarmin termination.
+      // This makes all the processors terminate together.
+      int global_do_terminate = ((std::time(NULL)-start_walltime) >= end_wall_time);
+      MPI_Bcast(&global_do_terminate, 1, MPI_INT, 0, this->get_mpi_communicator());
```
tjhei: this is not necessary as this happens in the termination criteria manager. Just do the if check here.

include/aspect/termination_criteria/end_walltime.h
```diff
@@ -0,0 +1,75 @@
...
+        /**
+         * Read the parameters this class declares from the parameter file.
+         */
+        virtual
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+        double end_wall_time;
+        static time_t start_walltime;
```
tjhei: why is this variable static? I don't think this is necessary.

source/termination_criteria/end_walltime.cc
```diff
@@ -0,0 +1,83 @@
...
+
+#include <aspect/termination_criteria/end_walltime.h>
+
+namespace aspect
+{
+  namespace TerminationCriteria
+  {
+    template <int dim>
+    time_t
+    EndWalltime<dim>::start_walltime = std::time(NULL);
```
tjhei: remove.

include/aspect/termination_criteria/end_walltime.h
```diff
@@ -0,0 +1,75 @@
+/*
+  Copyright (C) 2011, 2012 by the authors of the ASPECT code.
```
bangerth: You'll want to use 2018 here :-)

source/termination_criteria/end_walltime.cc
```diff
@@ -0,0 +1,83 @@
+/*
+  Copyright (C) 2011 - 2015 by the authors of the ASPECT code.
```
bangerth: same here

source/termination_criteria/end_walltime.cc
```diff
@@ -0,0 +1,83 @@
...
+  {
+    template <int dim>
+    time_t
+    EndWalltime<dim>::start_walltime = std::time(NULL);
+
+    template <int dim>
+    bool
+    EndWalltime<dim>::execute()
+    {
+      // Measure the time at process 0 to detarmin termination.
```
bangerth: -> determine

include/aspect/termination_criteria/end_walltime.h
bangerth: This should be `std::time_t`. Same in the place where you initialize the variable.

source/termination_criteria/end_walltime.cc
```diff
@@ -0,0 +1,83 @@
...
+
+
+    template <int dim>
+    void
+    EndWalltime<dim>::parse_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Termination criteria");
+      {
+        end_wall_time = prm.get_double ("Wall time") * 3600; // Change from hours to seconds.
+        start_walltime = std::time(NULL);
```
bangerth: If you make the `start_walltime` a `static` variable, then you shouldn't overwrite it here but use the initialization above.

include/aspect/termination_criteria/end_walltime.h
```diff
@@ -0,0 +1,75 @@
...
+
+        /**
+         * Read the parameters this class declares from the parameter file.
+         */
+        virtual
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+        double end_wall_time;
```
bangerth: The way you use this variable is as a *duration*, not as an *end time*. This should be reflected in the variable name.

By the way, the difference between `std::time_t` objects is almost certainly an integral type, so you can probably make this `unsigned int`. That gives you approximately 1 million hours as a max, enough for most purposes :-)

SiqiZhang:
Thanks for the comments @tjhei @bangerth . Those are addressed with the new commits.

include/aspect/termination_criteria/end_walltime.h
```diff
@@ -0,0 +1,87 @@
...
+         * once this value is reached.
+         */
+        unsigned int walltime_duration;
+
+        /**
+         * Make start_walltime as a static variable that it gets initialized pretty much 
+         * right away as the program starts, rather than several seconds in once we get 
+         * to creating the plugin. It gives better estimate of the actual start time.
+		 * The start_walltime is not stored into checkpoint, so it will get reset at 
+		 * restart.
```
bangerth: Please move these two lines to the top of the comment. They describe what the variable does, the three lines that are currently above it only describe implementation details.

You may also want to make sure everything is indented to the same level, using spaces instead of tabs.

source/termination_criteria/end_walltime.cc
```diff
@@ -0,0 +1,80 @@
...
+
+    template <int dim>
+    void
+    EndWalltime<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Termination criteria");
+      {
+        prm.declare_entry ("Wall time",
+                           "0",
+                           Patterns::Double (),
```
bangerth: I missed this yesterday: make this `Patterns::Double(0)` to ensure that the user-provided value is not negative.

SiqiZhang:
@bangerth New comments has been addressed and the PR is squashed into one commit.

gassmoeller:
The tests succeeded, but you will need to update the code style (see the details of the continuous-integration/jenkins/pr-merge tester). The problems are mostly spaces at the end of lines, and some whitespace issues. Should be easy to fix, but let us know if you get stuck.

tjhei:
Would it  be possible to add a test that terminates after 1 second?

tjhei:
> manages to run in 1 second

the time starts when aspect is launched. I think it would be easy to have a test where the setup takes more than one second to do, so we would always quit immediately.

bangerth:
Let's give @SiqiZhang the benefit of the doubt and go without a test: I think we can all come up with ways to test this in sort of roundabout ways, but it's not like we're likely going to break core functionality. By the way, it doesn't matter how long *setup* takes, the criterion is only evaluated at the end of the time step anyway.

Unless, of course, @SiqiZhang wants to write such a test (or learn how to write tests)?

SiqiZhang:
@bangerth @gassmoeller @tjhei I could set up a test for this.

bangerth:
That would then of course be very much appreciated. Do you need help with this?

SiqiZhang:
@bangerth Should be OK. I will let you know when it's done.

SiqiZhang:
@bangerth @tjhei @gassmoeller A test has been added with some bug fixes. I set the walltime to zero, so it always stops at the first step. If everything is OK and you all happy with this, I can squash them into a single commit and we are ready to merge.

bangerth:
Nice. Out of curiosity, why did you change the default for the wall time time-out to 24 hours?

source/termination_criteria/end_walltime.cc
```diff
@@ -0,0 +1,79 @@
...
+    // make sure it is initialized right way as the programe starts.
+    template <int dim>
+    std::time_t
+    EndWalltime<dim>::start_walltime = std::time(NULL);
+
+    template <int dim>
+    bool
+    EndWalltime<dim>::execute()
+    {
+      return ( (std::time(NULL)-start_walltime) >= (int)walltime_duration);
```
tjhei: can you use ``nullptr`` instead of ``NULL``?

source/termination_criteria/end_walltime.cc
```diff
@@ -0,0 +1,79 @@
...
+
+namespace aspect
+{
+  namespace TerminationCriteria
+  {
+    // The start_walltime is made as static and initialized here to
+    // make sure it is initialized right way as the programe starts.
+    template <int dim>
+    std::time_t
+    EndWalltime<dim>::start_walltime = std::time(NULL);
```
tjhei: same here.

SiqiZhang:
@bangerth While the default value set to 0, if somebody select the 'wall time' plugin without setting the wall time, the code will terminate after the initial step. So I think 24 hours might be a better choice.  It's the wall time limit for small jobs I usually has from a HPC system.

source/termination_criteria/end_walltime.cc
SiqiZhang: It's done.

bangerth:
Yay!

PR #2614, opened 2018-08-17
Make a function unconditionally available and conditionally fail.

bangerth:
This allows writing code that calls SimulatorAccess::get_world_builder()
without having to worry whether the WorldBuilder is actually available.
The check then simply happens at run-time.

This is the follow-up promised in #2571.


gassmoeller:
Looks correct, and this is indeed the better approach.

PR #2616, opened 2018-08-19
Fix prm formatting in the manual

jperryhouts:
Unimportant, but I just noticed that some of the prm excerpts in the manual weren't properly formatted.


tjhei:
thanks!

bangerth:
Nice find!

PR #2617, opened 2018-08-20
modify visco_plastic to allow for different grain size and activation…

mibillen:
… volume in the lower mantle and add test

This is the last change to the visco_plastic material model to allow for realistic composite rheology in whole mantle convection and full-mantle subduction models. This change will allow the user to use a different grain size and different activation volumes in the lower mantle in the lower mantle. This is needed to allow for both a step in viscosity (grain-size changes the pre-exponent) and a different radial viscosity gradient in the lower mantle (activation volume). Without this change, typical activation volumes end up predicting mantle viscosity exceeding 1e24 Pa-s by 1000 km depth. The user can also use this option to set the activation volume for dislocation creep to a large value in the lower mantle, which will cause the viscosity to follow the diffusion flow law instead (allows for composite viscosity in the upper mantle, but diffusion creep in the lower mantle).  For the grain-size, only one value is allowed (same as in the upper mantle), but for the activation volumes there is one value for each composition and both deformation mechanisms. The transition into the lower mantle is defined by a pressure given by an input parameter (default is set to large number so this option is not used).


source/material_model/visco_plastic.cc
```diff
@@ -173,6 +173,22 @@ namespace aspect
                                                   (constants::gas_constant*temperature*stress_exponents_dislocation[j])) *
                                          std::pow(edot_ii,((1. - stress_exponents_dislocation[j])/stress_exponents_dislocation[j]));
 
+          // Allow for different flow-law parameters (grain-size and activation volume) in the lower mantle
+          if (pressure > pressure_lower_mantle)
```
naliboff: Quick question/comment before a full review. How about moving this conditional statement to the beginning of the loop through each compositional field (e.g., right below explanation of viscosity equation)? 

Rather than repeating the equations, `if (pressure > pressure_lower_mantle)` would then define whether the activation volume and grain size would be the 'normal' values or those for the lower mantle. The new variable names would be 'activation_volume' and something like `current_grain_size`?

mibillen: I'm fine with this suggestion - I had tried to make the modification with the smallest possible change to the current code. 

Is there a guiding principal to use in making this decision... when is it better to introduce 2 new variables that are standing in for something else?  Maybe clarity and/or efficiency?

Here it seems like it might be okay, because you define them and immediately use them, so you don't have to search through the code to see where something got defined. This would make the code clearer (I think).  I can't use "activation_volume" because there will still be two - for diffusion and dislocation creep, but maybe current_activation_volume_diffusion and current_activation_volume_dislocation?

mibillen:
@naliboff and @gassmoeller  - I think I should close this pull request because I will use the new pressure-transition functionality to make these changes. Please confirm that this is what I should do and I will close this.

mibillen:
Closing this pull request because we will address this using the new phase transition option being added in #2656.

PR #2619, opened 2018-08-20
Jenkinsfile.k8s : Build in Jenkins workspace instead of /home/dealii and other build step cleanup

tjesser-ucdavis-edu:
Some more refactoring to remove complexity.

Regarding https://github.com/geodynamics/aspect/commit/bbe3c56225d3e1b3178c3fc78a6dee0626376be5, I'm not sure how Aspect detects the Deal.ii installation in the container, so that change may need to be reverted.

### DIFF
--- a/Jenkinsfile.k8s
+++ b/Jenkinsfile.k8s
@@ -97,9 +97,8 @@ pipeline {
               ..
           '''
           sh '''
-            export NP=`grep -c ^processor /proc/cpuinfo`
             cd build-gcc-fast
-            ninja -j "${NP}"
+            ninja
           '''
         }
       }
--- a/Jenkinsfile.k8s
+++ b/Jenkinsfile.k8s
@@ -83,24 +83,25 @@ pipeline {
             # openmpi complains about running as root. We need
             # to run as root, because the surrounding Jenkins container
             # assumes we have root access.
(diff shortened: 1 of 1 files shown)


gassmoeller:
Did you check if this makes Jenkins preserve all the build folders, which would quickly clutter the hard disk of the jenkins master? I remember that we looked into this last week, but dont remember our conclusion.

tjesser-ucdavis-edu:
The files on the Jenkins master server just come from the `git clone` it does before creating the jnlp container. The files created by building never leave the jnlp container (except as artifacts).

gassmoeller:
Ok, looks good. Since we are sure the files remain inside the container (and are thrown away after the test run) this is good to go. For reference: ASPECT finds deal.II via an environment variable DEAL_II_DIR that is set inside the container, so we do not need to worry about where we build ASPECT.

tjhei:
Are you sure? I would assume that the workspace directory is mounted in the ``docker run`` command from the outside. Building inside it could a) be slower or b) leave the files behind.
What is the advantage of this approach?

gassmoeller:
It is mounted from the outside, but the `outside` here means the jenkins-agent (jnlp) container, which was started by the jenkins-master. If we understood everything correctly, the jenkins-master should only store the artifacts. Maybe that is different to your jenkins setup, because we routinely destroy our agents after a test run has finished (because the kubernetes pod ends), and each pull request gets its new jenkins-agent container.

Issue #2375, opened 2018-06-22
c++11 changes and clang-tidy


tjhei:
now that we require c++11 we should update the code base:
[ ] use range-based for
[ ] remove pre c++11 stuff from compat.h
[ ] use nullptr
[ ] use lambda instead of bind
[ ] look at move constructors

Many of these things can be fixed by running clang-tidy, which I am running right now. We should also create an official clang-tidy config.

tjhei:
related: #2378

PR #2620, opened 2018-08-21
Use nullptr instead of NULL.

bangerth:
Since we now use C++11, we can safely do this.

I have not attempted to find places where we compare a pointer again the
integer '0', which should eventually also be fixed to 'nullptr'. But one
step at a time.


tjhei:
```
/home/dealii/source/source/particle/output/hdf5.cc:162:66: error: â€˜MPI_INFO_nullptrâ€™ was not declared in this scope
         H5Pset_fapl_mpio(plist_id, this->get_mpi_communicator(), MPI_INFO_nullptr);
```

tjhei:
related to  #2375

gassmoeller:
Looks good.

PR #2623, opened 2018-08-21
Refactor visco plastic

gassmoeller:
This is a response to the features that @mibillen and @naliboff want to add to the ViscoPlastic material model. Before we can make that model any more complicated we need to clean up some things, otherwise it will be impossible to keep track of its features.

Somewhere I broke a test during the refactoring, but I would like to see what the official tester says, and also give @naliboff and @mibillen a chance to comment, before I proceed.

### DIFF
--- a/include/aspect/material_model/visco_plastic.h
+++ b/include/aspect/material_model/visco_plastic.h
@@ -24,6 +24,8 @@
 #include <aspect/material_model/interface.h>
 #include <aspect/simulator_access.h>
 
+#include<deal.II/fe/component_mask.h>
+
 namespace aspect
 {
   namespace MaterialModel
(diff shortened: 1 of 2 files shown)


mibillen:
This is a very useful re-organization of code into functions in order to make the whole calculation (and separate behaviors) much clearer. I am not very familiar with the parts of the code that you put into new functions (fill_plastic_outputs, compute_viscosity_derivatives, compute_finite_strain_reaction_terms), so I can't really comment on whether there is anything missing here, I'll leave that to @naliboff to check more carefully.

gassmoeller:
Ok, found it, it was a weird refactoring mistake, not a bug in the original code. I called `fill_plastic_outputs` from within the loop over all points, but had another loop over all points inside (apparently because I thought it was called from outside that loop over i in `evaluate`). Should be fixed now and ready for review if the testers are happy.

naliboff:
@gassmoeller - I looked through this again and am a bit embarrassed to say found nothing specific to comment on. Unless you would like to go over this together on next week, this look ready to merge.

PR #2625, opened 2018-08-21
Use label for jenkins tester

gassmoeller:
This is just a suggestion, but I think the approach of using the 'ready to test' label to indicate pull requests that should be tested has worked quite well for the cig jenkins tester (it is essentially equivalent to our old /run-tests command, except it does not immediately trigger a build the first time it is added). I would like to add it to @tjhei's tester too, because this avoids having to manually restart that test run whenever a pull request is updated. Timo would you be fine with this?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,10 +12,6 @@ pipeline {
     timeout(time: 2, unit: 'HOURS') 
   }
 
-  parameters {
-    booleanParam(defaultValue: false, description: 'Is the pull request approved for testing?', name: 'TRUST_BUILD')
-  }
-
   stages {
     stage ("info") {
       steps {
@@ -30,7 +26,6 @@ pipeline {
     stage ("Check permissions") {
       when {
 	allOf {
-            environment name: 'TRUST_BUILD', value: 'false' 
             not {branch 'master'}
             not {changeRequest authorEmail: "rene.gassmoeller@mailbox.org"}
             not {changeRequest authorEmail: "heister@clemson.edu"}
@@ -43,8 +38,10 @@ pipeline {
 	    }
       }
       steps {
-	  echo "Please ask an admin to rerun jenkins with TRUST_BUILD=true"
-	    sh "exit 1"
+        sh '''
+          wget -q -O - https://api.github.com/repos/geodynamics/aspect/issues/${CHANGE_ID}/labels | grep 'ready to test' || \
+          { echo "This commit will only be tested when it has the label 'ready to test'"; exit 1; }
+        '''
       }
     }
 


tjhei:
Hah. I was thinking about that earlier today! Thanks! :+1:

bangerth:
Yes, we need something like this. The system with the tag does not reliably trigger a test run on all testers right now.

Jenkinsfile
```diff
@@ -43,8 +38,10 @@ pipeline {
 	    }
       }
       steps {
-	  echo "Please ask an admin to rerun jenkins with TRUST_BUILD=true"
-	    sh "exit 1"
+        sh '''
+          wget -q -O - https://api.github.com/repos/geodynamics/aspect/issues/${CHANGE_ID}/labels | grep 'ready to test' || \
+          { echo "This commit will only be tested when it has the label 'ready to test'"; exit 1; }
```
bangerth: So this is in the "Check permissions" stage, but how often does this run for a given PR? What triggers a re-run of this part of the script when the tag is applied?

tjhei: Right now only adding a new commit (or manually triggering the build on jenkins), sadly. We haven't found a better solution for this, yet. Note that this is even worse with the current version: here you have to manually rerun and check "trusted_build" after every commit.

gassmoeller: This stage is run for every re-run of the testsuite for a given PR (i.e. you can also remove the label, which means from now on the tester will not run anymore for this PR). The application of the label itself does unfortunately not trigger a re-run, but a push from someone does. As an example take the following situation:

- a PR is opened -> the testers run, but fail due to missing label
- someone with write permissions does a review, adds the label, and requests changes -> the testers do not run automatically (although the reviewer can manually start them, if he has an account on the jenkins server)
- the comments are addressed and the contributor updates the PR -> the testers run automatically, because the label is set
- from here on testers run automatically every time someone pushes to the PR (as long as the label is not removed).

The only unfortunate part is that the run is not triggered when the label is first added, everything afterwards works just as the old system did. I suppose the git plugin of Jenkins controls when to rebuild a pull request, and it is not obvious how to change that (at least there does not seem to be an option in the interface). There is a jenkins plugin that triggers runs based on comments (pull request builder), but it does not seem to be supported at the moment and has security warnings, so we would prefer to not use it.

bangerth: That's sort of what I understood.

What confuses me is that ours can't be a particularly unorthodox workflow. How do other projects deal with this? Do they just *always* build things, regardless of author, and say "we're inside a VM, nothing serious can happen"?

tjhei: > Do they just always build things

Good, question. I think so?

gassmoeller: Additionally, many now use cloud services like Travis-CI, Circle-CI, etc. that build everything either in fresh VMs for each test run, or inside containers. I guess that is why security becomes less of an issue.

PR #2630, opened 2018-08-23
Add material statistics postprocessor

gassmoeller:
This adds a postprocessor that computes the average density, average viscosity and total mass in a model. One could compute this from the solution output in Paraview, but I need accurate numbers for some benchmark purposes, and thought others might find it useful as well. This can potentially be extended in the future to also output the other material properties into the statistics file, but I think density and viscosity are the most useful ones.


tests/material_statistics.prm
```diff
@@ -0,0 +1,84 @@
+# This is a benchmark to check that the 'Material statistics'
+# postprocessor works correctly. It creates a model with lateral
+# flow. The density is the model is constant and the volume is known,
```
naliboff: is -> in

source/postprocess/material_statistics.cc
```diff
@@ -0,0 +1,126 @@
...
+namespace aspect
+{
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(MaterialStatistics,
+                                  "material statistics",
+                                  "A postprocessor that computes some statistics about "
+                                  "the material properties. In particular, it computes the "
+                                  "volume-averaged density, volume-averaged viscosity, and "
+                                  "total mass in the model.")
```
naliboff: Can you specify above that an arithmetic averaging is being used?

gassmoeller:
i adressed the comments.

naliboff:
Thanks, looks good to go once the tester is finished 👍

PR #2634, opened 2018-08-23
Add parse input table function

gassmoeller:
This is a function that allows to parse two-dimensional input arrays. One application would be to assign different densities to different compositions, with each composition having different phases (e.g. dependent on the depth). As discussed with @naliboff, I think this can be used to make #2586 more generic. It is not yet used anywhere, I just added tests, and we can discuss about how to integrate it best tomorrow.


gassmoeller:
Is this ready?

tjhei:
I think so? :)

PR #2635, opened 2018-08-27
Separate dashes in manual.tex

gassmoeller:
Just a minor fix for the manual to prevent merging the two dashes.


gassmoeller:
Stackoverflow is your friend :smirk:.

Issue #2282, opened 2018-06-13
manual section that explains the flow of the program


jdannberg:
This is something that came up in a discussion at the CGU/CIG meeting. 
Several participants mentioned that they would find it useful to get some more documentation on the core of ASPECT, namely, what happens when the program is started and what are the main functions that are called from there. This should probably include a flowchart that links the flow of the program to the files and functions where things are happening. 

Traditionally, this is something that we do not expose very much, but people who have used other codes before find it useful to see where the time stepping/assembly/etc. is done.

bangerth:
The problem is how to keep this kind of document up to date. We do change things in the code base, and it's easy to forget that a certain section in the manual needs to be updated as well. Maybe we can limit this to the big picture, rather than technical details?

gassmoeller:
We should not put in a lot of details that require frequent updates, but many participants in our tutorial found it very valuable to hear about where the program starts, how it flows, and what is called from where. I think steps that should be included in the chart would be:
- position of `int main()` and that it creates a `Simulator`, and then calls `Simulator.run()`
- position of `Simulator` and that `run()` contains the main time loop, and calls functions like `start_timestep`,`solve_timestep`, `postprocessing`, etc.
- position of the solver schemes, solver functions, and assembly functions
- some information about how our plugin system relates to the directory structure

These are not things that are of interested to participants of a very basic tutorial (like students who just started geodynamic modelling), but we had a large crowd of people developing other codes, and they are right, that we are not doing a good job of explaining how the internals of ASPECT work.

tjhei:
duplicate of #1029 ? The plugin graph there might be a good place to start...

gassmoeller:
@tjhei: Maybe, but we were explicitly asked: Where is this code? Not: How does this conceptually work?

tjhei:
I disagree. If you look at my figure, it is already a flow graph. If you annotate this with class/function names and .cc files, you would be quite close to what I would do. I think having the places where plugins are "active" in the graph is very useful.

this is a sketch for what I would do:
```
main() -- see main in main.cc

  construct Simulator -- ::Simulator()

    create geometry  -> plugin: GeometryModel

    set_assemblers   -> plugin: ...


  Simulator::run() -- run() in simulator/core.cc

    initial adaptive refinement loop:
      set_initial_temperature_and_compositional_fields -> plugin: initial conditions
      maybe_refine_mesh

    loop:
      start_timestep()
        ...

      solve_timestep()
        ...

[...]

PR #2636, opened 2018-08-27
Add section about program flow to the manual

gassmoeller:
We were asked at several tutorials about the large-scale flow of ASPECT (e.g. where is the main time loop, where does the program start), and I realized that we indeed nowhere explain the concept of `main`, and the `Simulator` class. I tried to come up with a simple explanation and flow chart, but I am not yet satisfied with the design and text (but currently out of creative ideas). I would be happy about suggestions for improvements.


tjhei:
I suggested something similar in #2282. Did you look at this?

doc/manual/manual.tex
```diff
@@ -10530,6 +10530,20 @@ \section{Extending and contributing to \aspect}
...
...Simulator.html}{aspect::Simulator}, and calls the function \texttt{Simulator::run()} that contains the main loop that advances the model in time (see Panel 3 in Figure~\ref{fig:program-structure}), until its termination criteria are fulfilled. You can imagine the \texttt{Simulator} class as \textbf{the} particular model you are running, it is meant to represent everything that encompasses your model, its matrices, solution vectors, geometries, material properties, and so on. All of this is created in the constructor of the \texttt{Simulator} class (Panel 2 of Figure \ref{fig:program-structure}). However, the \texttt{Simulator} class does not contain all of this functionality in itself (that would be overwhelming and bad software design). Instead it owns a number of other objects that do many sub-tasks, from setting boundary conditions, to assembling specific terms in the matrices, to prescribing the material properties. The main simulator object only orchestrates all of these member objects to create a succesful model run. Many of these sub-objects need to look very differently for different models, and therefore they are designed as plugin systems as described in the next section.
```
naliboff: A few stylistic suggestions:

`This function has a number of initialization tasks (see Figure~\ref{fig:program-structure}) and finally creates an object of the class \href{https://aspect.geodynamics.org/doc/doxygen/classaspect_1_1Simulator.html}{aspect::Simulator}. {aspect::Simulator} calls the function \texttt{Simulator::run()}, which contains the main loop that advances the model in time (see Panel 3 in Figure~\ref{fig:program-structure}) until its termination criteria are fulfilled`

`it is meant to` -> `as it is meant to`

`its matrices, solution vectors, geometries, material properties, and so on` -> `(matrices, solution vectors, geometries, material properties, and so on)`

gassmoeller:
@jdannberg: Could you take another look at the figure? I kept the design, but rearranged the boxes. If we keep it vertical it would cover a full page of the manual. Also the background boxes had too little contrast to the text, which made it hard to read. I made the background lighter, which should help with the readability of the text. Do you agree?

@tjhei: Sorry, no I had completely forgotten about that version. What do you think about the new version?

tjhei:
> I think it is okay to have the boxes next to each other, but I would keep the arrows, and keep the other boxes inside of the blue main.cc-box. The colours look good.

Can we also commit the source of this diagram or even better use a tool like graphviz so that we can generate the output automatically? I am sure we need to update diagrams like this over time.
I find that a lot more important than a pretty layout or the colors used. :-)

ljhwang:
I don’t know which version I am looking at now, but if it returns to main.cc <http://main.cc/> to terminate should the blue boxes within the blue boxes be the same size as the simulator/core.cc <http://core.cc/> boxes? Not necessarily clear in the arrangement I am looking at but I do like the colors ;)

Best,
-Lorraine

*****************************
Lorraine Hwang, Ph.D.
Associate Director, CIG
530.752.3656

tjhei:
> I find that a lot more important than a pretty layout or the colors used. :-)

Oh, I would also think just writing pseudo-code like in #2282 with clickable links would be okay. The flow-chart is currently more or less representable as:
```
setup
while (true)
{
time_step
post_process
if (terminate)
  break;
}
```

jdannberg:
Of course you could also represent the flowchart as pseudocode, but I think that especially for new users, a flowchart might be more intuitive. And the colours/and boxes are not only to make it pretty, I thought they would also make it easier to read, as everything in the same file has the same colours. 

And of course it would be nice to do this in graphviz, but I haven't used that before, and I am not sure it would be the best use of my time to redo the whole thing. But if you think that's the way to go, I can try and make a plot in graphviz once we're done discussing what should actually be in the flowchart.

tjhei:
> I think that especially for new users, a flowchart might be more intuitive

I am no opposed to a flowchart but I think it is a lot of work, difficult to update, and doesn't add much value.

> And the colours/and boxes are not only to make it pretty, I thought they would also make it easier to read, as everything in the same file has the same colours.

Oh, that's what the colors are for! I find dark green on light green difficult to read though.

gassmoeller:
I have to say my main concern is not really to get this perfect the first time around. Anything is better than nothing, and even only the text I wrote is better than our current non-existent documentation of the flow. Should I split this PR into one for the manual text, and another one for the discussion about the figure? I feel like investing more time in the figure is currently not the best use of my time :smirk:.

bangerth:
Nice! Something I noticed in the flow chart (on which I can't comment in the code diffs): parsing parameters already happens in `main()` (in fact, in `run_simulator()`, also in `main.cc`) and not in the constructor of class `Simulator` as shown in the graph.

bangerth:
In the `run()` box (green), it's not quite correct to say "assemble matrices -> solve system" because we do this repeatedly for each of the variables/blocks.

doc/manual/manual.tex
bangerth: If you break this paragraph into individual lines of ~80 characters, it becomes easier to comment :-)

Either way: 
* "you will most likely not need to change something" -> "you will most likely not need to change anything"
* "an object of the class" -> "an object of class"
* this sentence has "and finally..., and" -- remove one "and", but the "finally" also appears wrong :-)
* no comma in front of "until"; I would move the parenthetical remark after the "until" part to make reading easier
* \textbf{the} -> \textem{the}

doc/manual/manual.tex
```diff
@@ -10530,6 +10530,20 @@ \section{Extending and contributing to \aspect}
...
+
+\begin{figure}[tbp]
+  \centering
+  \includegraphics[width=0.95\textwidth]{program_structure.pdf}
+  \caption{\it \aspect{}'s main control flow starts in \url{source/main.cc}, which creates an object of type Simulator.
```
bangerth: no comma

bangerth: and markup "Simulator"

doc/manual/manual.tex
```diff
@@ -10530,6 +10530,20 @@ \section{Extending and contributing to \aspect}
...
+  This object orchestrates the model, and its run function contains the main time loop.}
```
bangerth: -> \texttt{run()}

bangerth:
I think having the source file from which you create the svg/pdf would be quite useful. I don't really care what program you used for it, but it would be good to have something we can edit.

gassmoeller:
I addressed the text comments. The .svg file of the graph was already part of the pull request and can be used to edit it, but before I do any changes on it, we should agree on whether to proceed with this version or the one in #2282.

bangerth:
On 09/06/2018 11:51 AM, Rene Gassmoeller wrote:
> The .svg file of the graph was already part of the pull request and can 
> be used to edit it,

Out of curiosity, how do you edit svg files?

bangerth:
On 09/06/2018 12:23 PM, Timo Heister wrote:
> in constructor:
> 
>     Call Simulator::parse_parameters()
> 
> really? that doesn't happen in the constructor

I had already mentioned that previously. Neither the first nor the first 
half of the fifth box in the middle graph actually happens in 
Simulator::Simulator().

jdannberg:
> Out of curiosity, how do you edit svg files?

I used [Inkscape](https://inkscape.org/en/) to make the first version of the chart, which is a vector graphics editor for Windows, Mac OS X and Linux that's free and open source.

bangerth:
On 09/11/2018 04:43 AM, Juliane Dannberg wrote:
> I used Inkscape <https://inkscape.org/en/> to make the first version of the 
> chart, which is a vector graphics editor for Windows, Mac OS X and Linux 
> that's free and open source.

Can Inkscape then read the .svg file, or does it have its own file format and 
.svg is then just what you export?

PR #412, opened 2015-05-23
[WIP] Adding SUPG to ASPECT

rrgrove6:
Rough draft of adding SUPG to ASPECT.

Attached is a picture (I think) that shows the comparison between my code (left) and ASPECT (right) and SUPG is falling a little short of expectations (I assume it has something to do with the residual, but the problem could be anywhere).  It is still a long way from being pretty, but I figured the sooner I get eyes on it to help, the better.

comparison

PR #2638, opened 2018-08-30
Reimplement supg #412

gassmoeller:
This is a very preliminary reimplementation of the SUPG stabilization method as proposed by @rrgrove6 in PR #412. It is in its very early stages, but compiles, runs, and at a very superficial glance also gives results in the test that are somewhat promising (stable in the sense that oscillations are present, but do not grow, less diffusive than our current default EV). There is still a lot of cleanup and benchmarking to do, but this version could allow @maxrudolph and others to compare SUPG to EV.


source/simulator/entropy_viscosity.cc
```diff
@@ -428,13 +429,105 @@ namespace aspect
...
+             :
+             parameters.composition_degree
+            );
+
+        // Compute tau for SUPG
+        if (parameters.use_supg == true)
+          {
+            double h = cell->diameter();
+            //nu = .1*h*h*norm_of_advection_field; //RG: temp
+            if (norm_of_advection_field>1e-8)
```
tjhei: this messy logic needs to be rewritten. I am also convinced these hard-coded 1e-8 won't work for anything dimensional.

gassmoeller:
I got a paper from @cedrict (PEPI, 2011 about his FANTOM code), and reimplemented SUPG as described in appendix A there. @tjhei: Do you have a reference for the implementation Ryan did in #412? It was very different from the implementation Cedric suggested, and it also does not seem to work as good.

gassmoeller:
Ah, I should mention that for now I have fixed tau, so it needs to be changed for every model. This is not ready for use as is, but it allows me to run some benchmarks and comparisons.

tjhei:
> Do you have a reference for the implementation Ryan did

The parameter design for tau? The paper is John, Knobloch 2008 ("On spurious oscillations at layers ...") equation 3 but here you obviously don't have these 1e-8 cut-offs Ryan implemented (this was work in progress to make things work).

One could also just pick tau=C*h as proposed by John and Novo 2011 ("Error analysis of the SUPG ...") and see what happens.

gassmoeller:
Actually, I was referring to the implementation of the terms in the assembler in (e.g. https://github.com/geodynamics/aspect/pull/2638/commits/cc619eae6bb25db98f95f9a1d3e872b48f288b0b#diff-2aa07bdd5ceb4101a47002236c944655L172 and https://github.com/geodynamics/aspect/pull/2638/commits/cc619eae6bb25db98f95f9a1d3e872b48f288b0b#diff-2aa07bdd5ceb4101a47002236c944655R212) . They had some funny units (like (rho*cp)^2) and had right-hand side terms, that according to the description in Appendix A of https://www.sciencedirect.com/science/article/pii/S0031920111001336 should not exist. However, I did not look up the original SUPG description and maybe Cedric's implementation in FANTOM is different, because of the different time-stepping scheme (although as far as I understand SUPG only enters in the advection term?).

tjhei:
> although as far as I understand SUPG only enters in the advection term?

No. SUPG adds the residual of your equation tested with a special test function (in streamline direction). I am not saying that Ryan did this correctly (we decided to implement this outside ASPECT and abandon the PR) or that Cedric is incorrect. I would have to double-check the equations.

tjhei:
See [scanned-aspect-supg.pdf](https://github.com/geodynamics/aspect/files/2341059/scanned-aspect-supg.pdf)

I can type it up if you want, but this was quicker.

tjhei:
> for some reason the dT/dt term is not modified

I think you can think of SUPG in two different ways:
1. Apply it to the time-dependent PDE so the residual contains dT/dt, then discretize in time. (my suggestion)
2. Apply time discretization first, so the PDE is stationary but contains a reaction term and an additional term on the RHS. Then apply SUPG.

Not sure what difference this makes.

bangerth:
Part of #2713.

include/aspect/parameters.h
```diff
@@ -249,6 +249,40 @@ namespace aspect
       };
     };
 
+    /**
+     *
```
gassmoeller: Add some documentation here.

source/simulator/assemblers/advection.cc
```diff
@@ -198,6 +224,29 @@ namespace aspect
...
+                {
+                  for (unsigned int j=0; j<advection_dofs_per_cell; ++j)
+                    {
+                      data.local_matrix(i,j)
+                      += tau *
+                         (
+                           (current_u * (density_c_P + latent_heat_LHS)) *
+                           scratch.grad_phi_field[i] *
+                           (
+                             -time_step * conductivity * trace(scratch.hessian_phi_field[j])
```
gassmoeller: Is `trace(hessian)` equivalent to the laplacians of the test function? I could imagine that `get_function_laplacians` is cheaper than `get_function_hessians` if we only use the trace anyway. Or am I mistaken?

source/simulator/assemblers/advection.cc
```diff
@@ -91,6 +91,7 @@ namespace aspect
             {
               if (fe.system_to_component_index(i).first == solution_component)
                 {
+                  scratch.hessian_phi_field[i_advection]        = scratch.finite_element_values[solution_field].hessian (i,q);
```
gassmoeller: Since we only need this with SUPG, can you move `use_supg` up and only do this if necessary? It will make assembly more expensive.

source/simulator/assembly.cc
```diff
@@ -1140,6 +1140,7 @@ namespace aspect
 
     const UpdateFlags update_flags = update_values |
                                      update_gradients |
+                                     update_hessians |
```
gassmoeller: Can we make this dependent on using SUPG? This will make assembly more expensive.

source/simulator/entropy_viscosity.cc
```diff
@@ -547,13 +548,67 @@ namespace aspect
...
+                                                                              global_max_velocity,
+                                                                              global_field_range.second - global_field_range.first,
+                                                                              0.5 * (global_field_range.second + global_field_range.first),
+                                                                              global_entropy_variation,
+                                                                              cell->diameter(),
+                                                                              advection_field);
+          }
+        else if (parameters.advection_stabilization_method == Parameters<dim>::AdvectionStabilizationMethod::supg)
+          {
+            // things needed for the calculating of tau for SUPG loop
```
gassmoeller:
```suggestion
            // things needed for calculating tau for SUPG loop
```

source/simulator/parameters.cc
```diff
@@ -884,13 +884,19 @@ namespace aspect
 
       prm.enter_subsection ("Stabilization parameters");
       {
+        prm.declare_entry ("Stabilization method", "entropy viscosity",
+                           Patterns::Selection("entropy viscosity|SUPG"),
+                           "Select the method for stabilizing the advection equation. The original "
+                           "method implemented is 'entropy viscosity' as described in \\cite {KGB12}.");
```
gassmoeller:
```suggestion
                           "method implemented is 'entropy viscosity' as described in \\cite {KHB12}.");
```

Also add a sentence that says that SUPG is currently experimental.

source/simulator/assemblers/advection.cc
tjhei: Sadly, there is no ``fe_values[field].laplacian``, but see how you like my change I will push in a minute.

tjhei:
This is ready for re-review...

gassmoeller:
Thanks, looks good.

PR #2642, opened 2018-08-31
Fix parallel viscosity smoothing

gassmoeller:
Found this while working on the stabilization. The logic in the old form does not smooth artificial viscosity across processor domains, because for ghost cells `!cell->is_locally_owned()` is true and the cell is skipped no matter what the second condition says. I think the old form missed a parentheses around the two conditions (`!(is_locally_owned || smoothing && is_ghost)`). However, instead of keeping this complicated logic I propose we just always compute the entropy viscosity on ghost cells (whether we need it for smoothing or not). It should not be very expensive, and is much simpler to understand.

### DIFF
--- a/source/simulator/entropy_viscosity.cc
+++ b/source/simulator/entropy_viscosity.cc
@@ -286,7 +286,9 @@ namespace aspect
     typename DoFHandler<dim>::active_cell_iterator cell = dof_handler.begin_active();
     for (; cell<dof_handler.end(); ++cell)
       {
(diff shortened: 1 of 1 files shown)


tjhei:
I agree, the logic is certainly incorrect. I wouldn't mind fixing it with the correct logic, though.

gassmoeller:
I do not like complicated negated if-statements, they always give me headaches :smile:.
What about this new version?

PR #2643, opened 2018-08-31
Some code cleanup in entropy viscosity

gassmoeller:
Only cleanup of little things I saw when going over entropy_viscosity.cc. The only functional change is that we now use the correct polynomial degree for computing the global entropy variation for compositions, if the composition and temperature have different polynomial degrees. Since this only computes a global reference this should have no effect even in the unlikely case somebody ran into this problem.


gassmoeller:
Look at that, we indeed had some tests that were (very mildly) affected by this exotic bug. I updated the test results and added a changelog entry.

PR #2644, opened 2018-09-03
jenkins: clean workspace

tjhei:
Each build generates about 250MB on the slave (source checkout) that is
never removed. Clean this in a post-build step.


tjhei:
please wait with merging so I can check if this works.

PR #2646, opened 2018-09-03
Fix plugin graph generation

tjhei:
aspect now happily prints the citation info, which generates a broken
plugin-graph. Fix this by suppressing the output.


tjhei:
I was afraid one of you would ask for that. :-)

I didn't see an obvious way to do it as the printing happens in the constructor, but let me see.

tjhei:
updated.

gassmoeller:
Thanks! :+1:

PR #2649, opened 2018-09-04
Only apply entropy viscosity in streamline direction

gassmoeller:
This is an experimental change to the entropy viscosity stabilization method that only applies the stabilization along streamlines (by multiplying the velocity with the advected gradient). Compared to SUPG in #2638 it has only a minimal change in the code, and we can keep all the other properties of EV. In first tests by myself and @maxrudolph this seems to fix several of the problems reported in the recent discussion on the mailing list:
- artificial diffusion across boundary layers is much smaller now, resulting in more correct average temperatures for a given Rayleigh number
- the reduced artificial diffusion across boundary layers reduces the inconsistency between top and bottom boundary heat flux (because the artificial diffusion is the part that does not appear in the heat flux postprocessor, and it is cell-size dependent that was the reason for the apparent imbalance)
The attached figure is from Max, and shows how in EV the heat flux does not converge towards the same value (black lines), while without stabilization (red lines) and SEV (green lines) it does.

image

- artificial diffusion of internal structures also is smaller than with EV (because structures are only diffused in one dimension?)

I attach a figure of the same model Ryan used in #412, but with EV, SEV, and SUPG. The bottom left panel shows the initial state, the other three panels the state after one rotation. Note that black and pink colors are over-/undershoots (most prominent in SUPG), and how in EV the structures tend to diffuse in azimuthal and radial direction, while in SEV they primarily diffuse in azimuthal direction. Also note that there are some oscillations in radial direction in SEV that are no longer damped as well as in EV. 

ev-sev-supg-comparison

There is certainly still room for improvements (currently we do not apply any stabilization perpendicular to streamlines which causes some of the oscillations, and also a high residual does lead to a high stabilization, even if the oscillations are in radial direction, thus the stabilization can not help against it, which leads to overdiffusion in azimuthal direction). Still the current state looks like a big step forward to finally be able to reproduce all the benchmarks multiple groups have been struggling with for so long, and explain the wrong average temperature.

I would appreciate comments, both about the mathematical theory, and how well this works in practice. If you had a benchmark that failed because of weird heat flux / average temperature differences, please give it a try with this change (all tests I have done so far look much better now).

One final remark: Of course if you run a model with sufficiently high resolution all the differences between the methods disappear and you converge towards the correct solution, but in 3D spherical models it is very worthwhile to reduce the *sufficiently* high resolution as much as possible, and with the current EV this means using at least 5 (better 6 or 7) refinements in the boundary layer, which is just absurdly expensive if you want to reproduce a benchmark of say Rayleigh number 7000. SEV should allow much lower resolutions, but how much we would need to test.

### DIFF
--- a/source/simulator/assemblers/advection.cc
+++ b/source/simulator/assemblers/advection.cc
@@ -106,12 +106,22 @@ namespace aspect
                   ExcMessage ("The product of density and c_P needs to be a "
                               "non-negative quantity."));
 
-          const double conductivity =
+          double conductivity =
             ((advection_field_is_temperature)
              ?
              scratch.material_model_outputs.thermal_conductivities[q]
              :
              0.0);
+
+          if (!this->get_parameters().use_streamline_entropy_viscosity)
+            conductivity += scratch.artificial_viscosity;
+
+          const double anisotropic_artificial_viscosity = (this->get_parameters().use_streamline_entropy_viscosity)
+                                                          ?
+                                                          scratch.artificial_viscosity
+                                                          :
+                                                          0.0;
+
           const double latent_heat_LHS =
             ((advection_field_is_temperature)
              ?
@@ -154,6 +164,11 @@ namespace aspect
           if (this->get_parameters().free_surface_enabled)
             current_u -= scratch.mesh_velocity_values[q];
 
+          Tensor<1,dim> current_unit_u = current_u;
+
+          if (current_u.norm() > std::numeric_limits<double>::min())
+            current_unit_u /= current_u.norm();
+
           const double JxW = scratch.finite_element_values.JxW(q);
 
           // do the actual assembly. note that we only need to loop over the advection
@@ -174,8 +189,10 @@ namespace aspect
                 {
                   data.local_matrix(i,j)
                   += (
-                       (time_step * (conductivity + scratch.artificial_viscosity)
+                       (time_step * conductivity
                         * (scratch.grad_phi_field[i] * scratch.grad_phi_field[j]))
+                       + time_step * anisotropic_artificial_viscosity
+                       * ((current_unit_u * scratch.grad_phi_field[i]) * (current_unit_u * scratch.grad_phi_field[j]))
                        + ((time_step * (scratch.phi_field[i] * (current_u * scratch.grad_phi_field[j])))
                           + (bdf2_factor * scratch.phi_field[i] * scratch.phi_field[j])) *
                        (density_c_P + latent_heat_LHS)
--- a/source/simulator/entropy_viscosity.cc
+++ b/source/simulator/entropy_viscosity.cc
@@ -225,6 +225,12 @@ namespace aspect
                                max_velocity * max_residual /
                                (global_u_infty * global_field_variation));
 
+        // If there is physical conductivity it already stabilizes the equation,
+        // thus reduce the artificial stabilization by this amount.
+        if (entropy_viscosity >= max_conductivity)
+          entropy_viscosity -= max_conductivity;
+        else
+          entropy_viscosity = 0.0;
 
         return std::min (max_viscosity, entropy_viscosity);
       }
--- a/include/aspect/parameters.h
+++ b/include/aspect/parameters.h
@@ -446,6 +446,8 @@ namespace aspect
     std::vector<double>            stabilization_c_R;
     std::vector<double>            stabilization_beta;
     double                         stabilization_gamma;
+    bool                           use_streamline_entropy_viscosity;
+
     double                         discontinuous_penalty;
     bool                           use_limiter_for_discontinuous_temperature_solution;
     bool                           use_limiter_for_discontinuous_composition_solution;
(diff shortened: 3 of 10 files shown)


source/simulator/assemblers/advection.cc
```diff
@@ -174,8 +189,10 @@ namespace aspect
                 {
                   data.local_matrix(i,j)
                   += (
-                       (time_step * (conductivity + scratch.artificial_viscosity)
+                       (time_step * conductivity
                         * (scratch.grad_phi_field[i] * scratch.grad_phi_field[j]))
+                       + time_step * anisotropic_artificial_viscosity
+                       * ((current_unit_u * scratch.grad_phi_field[i]) * (current_unit_u * scratch.grad_phi_field[j]))
```
bangerth: That's a lot of extra terms to multiply out. How does this affect run time?

source/simulator/assemblers/advection.cc
```diff
@@ -154,6 +164,11 @@ namespace aspect
           if (this->get_parameters().free_surface_enabled)
             current_u -= scratch.mesh_velocity_values[q];
 
+          Tensor<1,dim> current_unit_u = current_u;
+
+          if (current_u.norm() > std::numeric_limits<double>::min())
+            current_unit_u /= current_u.norm();
+
```
bangerth: I think you might as well add
```
  else
    current_unit_u = Tensor<1,dim>();
```

source/simulator/entropy_viscosity.cc
```diff
@@ -225,6 +225,12 @@ namespace aspect
                                max_velocity * max_residual /
                                (global_u_infty * global_field_variation));
 
+        // If there is physical conductivity it already stabilizes the equation,
+        // thus reduce the artificial stabilization by this amount.
+        if (entropy_viscosity >= max_conductivity)
+          entropy_viscosity -= max_conductivity;
+        else
+          entropy_viscosity = 0.0;
```
bangerth: Nah, let's do this as in #2650, instead of modifying the formula here.

gassmoeller:
> That's a lot of extra terms to multiply out. How does this affect run time?
I guess significantly for the assembly. I was planning to pre-compute the scalar products outside of the loop, but did not want to make the code more complicated right now.

> I don't know whether that's going to be the case, but experience tells me that someone is going to come up with cases where our default values are going to be too small now.

Yes, I was not planning on asking for this to be merged before we have *a lot* of testing results (I am writing the mail for the mailing list right now that asks everyone to check their failing benchmarks with this branch). However, I am running the blankenbach benchmarks at the moment, and it looks like even in 2D boxes the vrms values are **much** closer to the reference results (as in we need 1-3 refinements less for the same deviation from the reference vrms value). If this holds true across other benchmarks it will mean a significant decrease in the resolution needed to get accurate results. Nusselt numbers are worse than before and actually too high (previously they were too low), i wonder if this is the accuracy problem in the postprocessor related to the gradient computation on the surface.

bangerth:
You could set up the testcase from step-31, which showed all sorts of pathologies when the two plumes going down on the sides meet again. Hot plumes meeting head on were the cases that always led to blow-up. That's also why the movie stops at the point where the two plumes meet...

maxrudolph:
I am seeing numerical instabilities in the temperature field using this PR for our more complex models with imposed plate motions. I am not sure that I can produce a minimal complexity case that fails. The attached figure shows the temperature field (3-slice). The first shows the full range of the color axis. The second shows a truncated axis [0,2500].

visit0003
visit0002

gassmoeller:
No, I see the same issues in high Rayleigh number flow. The benchmark results for low Rayleigh numbers seem to be better, but it is not stable for high Rayleigh numbers (=realistic models). Bummer, I will need to do more testing.

maxrudolph:
For high Ra calculations, I am still seeing much faster cooling with this implementation than in CitcomS.

gassmoeller:
One thing you could also try is to run the aspect model with branch #2650 and the original (lower) stabilization parameters:

```
subsection Discretization
  subsection Stabilization parameters
    set cR = 0.11
    set beta = 0.078
  end
end
```
Let me know how that goes.

maxrudolph:
I don't think that #2650 will solve any problems, even with lower values of the stabilization parameters. I would be more interested in trying this SUEV approach with the lower stabilization parameter values, or trying SUPG, though we will have to determine a reasonable choice of tau. I would also like to try to use DG for temperature without stabilization, but not sure how to work past the failures to converge. I don't think that I had the bound preserving limiter enabled, and this may make a difference.

bangerth:
#2650 isn't meant as a magic bullet, but more like a first step towards a shared goal. I think it is independent of what's proposed here.

maxrudolph:
I am happy to run the simple 3D convection example with #2650 and we can see whether it makes any difference.

bangerth:
That would be great -- thanks, @maxrudolph !

bangerth:
Part of #2713.

gassmoeller:
I think we agreed my idea would not be stable, thanks Max for the testing! I will keep the branch and we can always reopen the PR, but there is no point in keeping this open at the moment.

PR #2650, opened 2018-09-04
Use the max of physical and artificial viscosity.

bangerth:
There is really no need to choose the sum: one is always sufficient, and
then that's the one we should choose.

This is one of the ideas that came out of the recent discussion on how
to choose the artificial viscosity. It is likely not going to make a
significant difference, but small steps are also useful. I'd expect it
to break a *lot* of tests -- I'm ready to receive the Puckett Bucket
if so.


bangerth:
CC: @gassmoeller @maxrudolph

gassmoeller:
Yes, I agree. This is already a part of #2649, but it is a feature on its own, so we can split it into this PR. By the way: I am opposed to blaming someone for breaking tests if that was expected.

bangerth:
On 09/04/2018 04:36 PM, Rene Gassmoeller wrote:
> By the way: I am opposed to blaming someone for breaking tests if that was 
> expected.

You're just too nice ;-)

But I do agree that there are cases where someone foresees that tests will be 
broken and that that should be ok.

gassmoeller:
> You're just too nice ;-)

Pure self preservation, I broke 30% of all tests in #2649, and that was much less than I expected ;-).

I am not quite sure how to test this. It is not as critical as #2649, but we should probably run at least a few benchmarks to see that nothing explodes. Maybe the series of blankenbach cases? They come with a plugin and a makefile that automatically compares the results to reference results.

bangerth:
Nice improvement.

As mentioned in the other PR, I think a testcase that has plumes meeting head-on was the one that broke most things. Also ones where a compositional field is truly discontinuous.

gassmoeller:
All results I have seen suggest that this is already a nice improvement. It is stable (unlike #2649 at the moment), and the blankenbach benchmark vrms velocity converge much faster to the reference values than before. So, I would be in favor of merging this separately (after updating the tests).

I have one more question about the stabilization: The artificial viscosity scales with the residual, but the residual that we compute does not include the artificial viscosity term (see simulator/assemblers/advection.cc:241). So we solve a different equation than we use to compute the residual. Will that not lead to unnecessarily large artificial viscosities, especially in areas where we apply stabilization? 
I would test what happens if we compute the real residual, but currently the artificial viscosity of the old timestep is not stored. Maybe we should?

tjhei:
> The artificial viscosity scales with the residual, but the residual that we compute does not include the artificial viscosity term

The residual is not the residual of the PDE but the entropy residual. I went back to the paper Guermond 2011 and they do not include the artificial viscosity in the entropy residual but they also don't consider physical diffusion at all. My understanding is that the entropy residual won't be smaller by adding the entropy viscosity to it but this might lead to feedback cycles (you want the same size of the residual in every timestep if the solution stays constant). @bangerth ?

bangerth:
Yes, the method uses the physical entropy, which is based on the original equation but not the same. I think it is correct to not include the artificial viscosity in this.

maxrudolph:
image

bangerth:
Do I interpret this correctly as saying that the approach in #2650 leads to faster convergence at the bottom than any of the other methods, but is no good (or at least not *much* better) at the surface?

maxrudolph:
@bangerth, yes I think that this is a fair interpretation. This figure shows the artificial viscosity at the last timestep. It is not entirely obvious why the cmb heat flow is resolved so much better than the surface since the artificial viscosity exceeds the thermal conductivity throughout much of both boundary layers.

image

This figure shows only regions where the artificial viscosity is greater than the thermal conductivity.
image

same as above but spherical slice at cmb:
image

spherical slice at surface:
image

bangerth:
Fun pictures :-)

So I think we agree that things are at least "somewhat" better than before. How do we proceed from here?

maxrudolph:
I advocate for merging this PR as it clearly represents an improvement over the current implementation but it would probably be a good idea to re-run at least some of the thermal convection benchmarks (blakenbach cases). What do you think?

gassmoeller:
I took the liberty of updating the test results (this PR affects nearly every test). Since we all agree this is a good idea, I guess we can proceed with it? Should I add a changelog entry, since this will affect all of our benchmark results as well?

For reference these are the old Blankenbach benchmark results:

# Nu                     Vrms                    name (refinement level):
4.78661864e+00 4.34590432e+01 case1a_ref4.stat
4.87927972e+00 4.29377468e+01 case1a_ref5.stat
4.88993106e+00 4.28733838e+01 case1a_ref6.stat
4.88680525e+00 4.28659548e+01 case1a_ref7.stat
4.88440900e+00 4.28649470e+01 case1a_reference.stat

Both Nu and Vrms converge, but rather slowly for the very low Rayleigh number (10^4). Below are the values with this PR (taking the max of artificial diffusion and physical diffusion instead of the sum):

# Nu                     Vrms                    name (refinement level):
5.30885322e+00 4.28499932e+01 case1a_ref3.stat
5.06735289e+00 4.28656773e+01 case1a_ref4.stat
4.93712396e+00 4.28650353e+01 case1a_ref5.stat
4.88440900e+00 4.28649470e+01 case1a_reference.stat

As you can see the Vrms is now much closer to the reference value already at low resolutions (even at refinement level 3, which is only 8x8 cells). But the Nusselt number is now worse, and converging from above the reference value instead of from below. As discussed on the mailing list, this is caused by the postprocessor that was hidden before by the artificial diffusion term, and is adressed in #2660. In short: The actual solution (Vrms) is much better now, but the postprocessed quantity (Nusselt number) is worse because of a separate issue.

tjhei:
> I guess we can proceed with it?

yes.

> Should I add a changelog entry, since this will affect all of our benchmark results as well?

yes.

gassmoeller:
Done.

tjhei:
:+1:

PR #2651, opened 2018-09-04
Better explain what kinds of checks run with --validate.

bangerth:
I think the text is ok as is, but I've been looking at something else in this file
and thought I might as well explain this better. In the process, I've also made sure
that the message fits into an 80-character wide terminal.


tjhei:
Message looks good. Is there a reason you are manually breaking the lines instead of letting the terminal window do this automatically?

PR #2652, opened 2018-09-05
Provide formulas for the material-statistics postprocessor.

bangerth:
Better be specific with formulas than rely on a reader's understanding of
a text description.

I'm not sure there is a reason why one would ever want to use anything
other than the usual arithmetic averaging for the density and mass,
given that these have physical interpretations, so I dropped that remark.

I'm tacking on a change in wording in @gassmoeller's original changelog
to make it clear that the postprocessor computes 'the average density',
(passive) rather than 'averages the density' (an action).

Follow-up to #2630.

### DIFF
--- a/source/postprocess/material_statistics.cc
+++ b/source/postprocess/material_statistics.cc
@@ -120,8 +120,12 @@ namespace aspect
                                   "material statistics",
                                   "A postprocessor that computes some statistics about "
                                   "the material properties. In particular, it computes the "
-                                  "volume-averaged density, volume-averaged viscosity, and "
-                                  "total mass in the model. Currently, only arithmetic "
-                                  "averaging is supported.")
+                                  "volume-averages of the density and viscosity, and the "
+                                  "total mass in the model. Specifically, it implements "
+                                  "the following formulas in each time step: "
+                                  "$\\left<\\rho\\right> = \\frac{1}{|\\Omega|} \\int_\\Omega \\rho(\\mathbf x) \\, \\text{d}x$, "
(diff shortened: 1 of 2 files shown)


source/postprocess/material_statistics.cc
```diff
@@ -120,8 +120,12 @@ namespace aspect
                                   "material statistics",
                                   "A postprocessor that computes some statistics about "
                                   "the material properties. In particular, it computes the "
-                                  "volume-averaged density, volume-averaged viscosity, and "
-                                  "total mass in the model. Currently, only arithmetic "
-                                  "averaging is supported.")
+                                  "volume-averages of the density and viscosity, and the "
+                                  "total mass in the model. Specifically, it implements "
+                                  "the following formulas in each time step: "
+                                  "$\\left<\\rho\\right> = \\frac{1}{|\\Omega|} \\int_\\Omega \\rho(\\mathbf x) \\, \\text{d}x, "
```
tjhei: I think you are missing a lot of $ signs in those lines.

bangerth:
I fixed the formulas. I guess I am just so used that in emacs latex mode, when you type `$` it automatically creates the matching closing `$` as well :-)

gassmoeller:
Hm, I am ok dropping that remark. @naliboff asked for it in the original pull request, because it might make more sense to compute a different average for the viscosity later on (e.g. averaging the logarithm of the viscosity). Anyway, we can always add it later on if we actually use different averaging schemes for different properties.

bangerth:
ping?

PR #2660, opened 2018-09-11
Average heat flux in boundary cells for more accurate heat flux computation

gassmoeller:
This PR builds on features from #2659 and #2650. It changes the computation of the heat flux: Instead of taking the surface integral of the conductive heat flux, it computes the average heat flux normal to the boundary face across the entire volume of the boundary cell (as suggested by Scott King on the mailing list, and first discussed in Ho-Liu, Phyllis, Bradford H. Hager, and Arthur Raefsky. "An improved method of Nusselt number calculation." Geophysical Journal International 88.1 (1987): 205-215.). This reduces the influence of temperature oscillations at the boundary and avoids computing the gradient of the solution on a face.

First tests have shown it to be vastly more accurate for incompressible models like the blankenbach benchmark:

These are results of the blankenbach benchmark with #2650:

# Nu                     Vrms                    name (refinement level):
5.30885322e+00 4.28499932e+01 case1a_ref3.stat
5.06735289e+00 4.28656773e+01 case1a_ref4.stat
4.93712396e+00 4.28650353e+01 case1a_ref5.stat
4.88440900e+00 4.28649470e+01 case1a_reference.stat

These are the results with the improved postprocessor of this PR (note that the solution is not affected, only the postprocessing of the Nusselt number):

# Nu                     Vrms                    name (refinement level):
4.89728221e+00 4.28499932e+01 case1a_ref3.stat
4.88535143e+00 4.28656773e+01 case1a_ref4.stat
4.88443365e+00 4.28650353e+01 case1a_ref5.stat
4.88440900e+00 4.28649470e+01 case1a_reference.stat

The other cases of the blankenbach benchmark show similar (though somewhat less spectacular) improvements.

Unfortunately, the method does not work well for compressible models like the benchmarks in the king2dcompressible folder. I suspect that this is because it is missing the additional heating terms in the energy equation, but no matter what I tried I could not find a combination of terms that give a more correct result than our reference results (computed before #2650). I tried:
- not including any heating terms
- only including the adiabatic heating (as it is the only term that removes 'heat' by doing volume expansion) in streamline direction (assuming that the additional heating only matters if the material particle cools while approaching the boundary), or isotropically (assuming all of this heat would make it to the boundary)
- include the sum of all heating terms in streamline direction or isotropically
- modifying the temperature in the advection term (rho * cp * velocity * temperature) by the temperature difference in the adiabatic profile from the current location to the boundary face location, this should account for the integrated adiabatic heating 'on the way' to the boundary

So at the moment I am stuck. Does somebody have a theoretical understanding of how to compute the 'heat flux' (as in heat that will eventually leave through the boundary face) as a volume average in a boundary cell? Should we use this as a separate (accurate) way to compute heat flux in incompressible models, and leave everything as is for compressible models?


SiqiZhang:
I wonder if we extrapolate the heat flux from quadrature point to the boundary instead of using cell average will stay effective and could help us to get rid of variation due to heating terms. 

> Should we use this as a separate (accurate) way to compute heat flux in incompressible models, and leave everything as is for compressible models?
In current approach, I guess incompressible models with internal heating will also have trouble.

tjhei:
Can we evaluate internal heating and density related terms on the surface while averaging the flux over the cell?

SiqiZhang:
@gassmoeller I did a few tests on this pull request. If the temperature gradient has systematical variation within the cell (can be due to heating term or curvature), the averaging is not doing so well. 
I run this postprocessor on the initial condition (conductive profile) of a small core 3D non-dimensional model (core ratio f=R0/R1=0.1, R0=0.1111,R1=1.1111). The initial condition is T(r)=f/(f-1)*(1-R1/r), analytical solution for the heat flow is ~1.5513883 through the whole mantle. The velocity is small due to no perturbation, so only the temperature gradient is the main contributor of the heat flow.
For this new postprocessor:
GMR=4:
Heat fluxes through boundary parts:              -0.9296 W, 1.641 W
GMR=5:
Heat fluxes through boundary parts:              -1.186 W, 1.596 W

For the old postprocessor from master(d01deca300b9a8a0052e54bcec33c56506fc2508):
GMR=4:
Heat fluxes through boundary parts:              -1.428 W, 1.549 W

Sadly, the old treatment works much better compare to the new implementation in this test case.
I did some calculation by hand for GMR=4 case, if using temperature gradient near cell center I get ~0.9 heat flow for the bottom, if using temperature gradient extrapolated to the boundary I get ~1.4 which confirms heat flow values I get in the test.
I would vote for extrapolating the revised heat flux from cell quadrature points to boundary instead of a cell average. I will try to implement this and test, but it is a bit of struggle to do the extrapolation.

gassmoeller:
Sorry for being absent from this PR for a while, and thanks for the suggestions!

@tjhei: To your first point: I do think I implemented the reference density correctly, and this should not be an issue, or were you referring to something specifically? I am not sure about the second suggestion. If we want to evaluate heat flux at the surface we do not need to consider internal heating processes (because there is only a conductive heat flux through the surface), and the density only appears in the advective term, which also goes to zero at the boundary. The main purpose of this change is to avoid evaluating the temperature gradients at the boundary, but if we evaluate the gradients in the cell volume instead, we need to compensate for the fact that there are other heat flux processes at work there.

@SiqiZhang: Thanks for the tests! It is unfortunate that they seem to be worse than the old state, but I am not sure I understood completely what you mean by 
>  If the temperature gradient has systematical variation within the cell (can be due to heating term or curvature), the averaging is not doing so well.

I think I understood what you mean by extrapolating the heat flux, though I would probably call it average the face area differently. Since we are computing a volume average heat flux, we should also compute a *volume-averaged* face area (sounds a bit funny). In essence this should be something like the average of the boundary face area and the area of the face opposite to the boundary of this cell. Could you check if my last commit improved your results? I had no opportunity to test it, but if you have a good testcase feel free to attach that as well.

Still, the problems with compressible models seem unrelated to the curved geometry. If I find some time, I will come back to how to formulate the shear heating terms that are currently missing for compressible models.

source/postprocess/heat_flux_map.cc
```diff
@@ -121,6 +132,22 @@ namespace aspect
...
 
+                    // The opposite face according to the deal.II face indexing
+                    // is always the next or previous face index.
+                    unsigned int opposite_face_index = f;
+                    if (f%2 == 0)
+                      opposite_face_index += 1;
+                    else
+                      opposite_face_index -= 1;
+
+                    fe_face_values.reinit (cell, f);
```
SiqiZhang: > `fe_face_values.reinit (cell, f);`
Should be:
`fe_face_values.reinit (cell, opposite_face_index);`

gassmoeller: Good catch. Fixed. Does it work for you?

SiqiZhang:
@gassmoeller: Thanks for the update. After fixing the error in the new commit, it fixed the problem with the curvature quite well.
The heat flow for my test:
 GMR=4:
Heat fluxes through boundary parts:              -1.6 W, 1.552 W
It is pretty close to analytical solution value ~1.5513883
Here is a prm file for my test. It is 3D test with 1.7M DoFs, but it only runs the initial step:
[f01v0_no.prm.txt](https://github.com/geodynamics/aspect/files/2413768/f01v0_no.prm.txt)

Yes, averaging the face area should solve the problem with the curvature, however with heating terms, the heat flow through a cell is not longer expect to be a constant. That's why I was suggesting to do extrapolation instead of averaging.
However, I tried to extrapolate the heat flux to the face, it yields the similar results as the old postprocessor. No problem with the curvature, but no significant improvement with blankenbach benchmark either:
# Nu           Vrms           name:
5.05516070e+00 4.28656773e+01 case1a_ref4.stat
4.93622447e+00 4.28650353e+01 case1a_ref5.stat
4.89801954e+00 4.28649625e+01 case1a_ref6.stat

My implementation is here if you would like to take a look:
https://github.com/SiqiZhang/aspect/tree/extrapolate_heat_flux
It seems to be a failed approach anyway.

Hope we could figure out a way to deal with the heating terms with the averaging approach.

gassmoeller:
Superseded by #2675. @SiqiZhang: If you have the time and compute time, would you have a chance to test your example with the code in #2675? I did not test it in spherical geometries yet, although there should not be anything different.

PR #2661, opened 2018-09-11
allow /rebuild in Jenkins

tjhei:
Every time a comment that contains "/rebuild" is added to a PR, a fresh build will be triggered.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,9 +38,14 @@ pipeline {
 	    }
       }
       steps {
+        // For /rebuild to work you need to:
+        // 1) select "issue comment" to be delivered in the github webhook setting
+        // 2) install "GitHub PR Comment Build Plugin" on Jenkins
+        // 3) in project settings select "add property" "Trigger build on pr comment" with
+        //    the phrase ".*/rebuild.*" (without quotes)
         sh '''
           wget -q -O - https://api.github.com/repos/geodynamics/aspect/issues/${CHANGE_ID}/labels | grep 'ready to test' || \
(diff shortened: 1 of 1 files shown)


tjhei:
@gassmoeller maybe also worth doing on the CIG Jenkins? Instructions are in the comment of this PR.

Jenkinsfile
```diff
@@ -38,9 +38,14 @@ pipeline {
 	    }
       }
       steps {
+        // For /rebuild to work you need to:
+        // 1) select "issue comment" to be delivered in the github webook setting
```
bangerth: webhook

tjhei:
> Does the plugin allow filtering based on who made the comment?

I don't think so, at least there is no such setting.

We could try to use https://github.com/jenkinsci/pipeline-github-plugin#triggers instead of https://plugins.jenkins.io/github-pr-comment-build (as I have done here). That plugin is a lot fancier and "will only allow collaborators of the repository in question to trigger builds".

Thoughts?

tjhei:
> It doesn't seem to have the property either.

not sure. If you scroll up it says "Currently this trigger will only allow collaborators of the repository in question to trigger builds."

I guess we could try this out. Anyways: this not a huge problem because the tests will only run if the label is set (which requires one of us to set it). Worst case is somebody can queue a bunch of harmless runs.

gassmoeller:
I agree this looks reasonable. I will merge it for now (it only affects Timo's tester at the moment). @tjesser-ucdavis-edu: Would you be ok including this into the CIG tester? Let's discuss this in person some time.

PR #2662, opened 2018-09-12
fix umlauts in citation files

jdannberg:
I noticed this while trying to copy the citation information from the Aspect website into a manuscript. 
@gassmoeller, let me know if this makes sense to you!


tjhei:
actually, that is how his name is spelled in manual/manual.bib and for zenodo that is how it is defined over there: https://zenodo.org/record/1297145#.W5kUZnUzrTY which is based on his orcid id.

I think the reasoning behind it was that zenodo would mess up the formatting but we are welcome to try to change it. I just think that this location is the wrong place to edit it.

jdannberg:
Oh, I see. So I'll just close this? 
I just noticed that it's not consistent, and that in places where ö is used (like in the Heister et al., 2017 paper) the formatting is messed up because html does not get the proper backslash, so it's actually not usable with bibtex. Let's ask @gassmoeller what spelling he would prefer.

PR #2663, opened 2018-09-13
www: citing.html: fix escaping

### LINKED ISSUE #2611: Fix bibtex entry in citing.html
#2594 added `citing.html`, which has a bibtex entry that is rendered as follows:
```
@article{aspectmanual, 
title = {{	extsc{ASPECT}: Advanced Solver for Problems in Earth's ConvecTion, User Manual}}, 
author = {W. Bangerth and J. Dannberg and R. Gassm{"o}ller and T. Heister and others}, 
```
That's missing backslashes and text in `\testsc` and `"o`.

tjhei:
fixes #2611
also see #2662

### DIFF
--- a/database.js
+++ b/database.js
@@ -1,51 +1,54 @@
+// Auto-generated file by https://github.com/geodynamics/aspect/doc/make_cite_html.py
+// Do not edit!
+
 var papers = {entry1:{
-text: "Bangerth, W., Dannberg, J., Gassmoeller, R., Heister, T., &amp; Others. (2017, March 1). Aspect V1.5.0. Zenodo. https://doi.org/10.5281/zenodo.344623",
-bibtexkey: "aspect-doi-v1.5.0",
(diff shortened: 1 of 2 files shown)


database.js
```diff
@@ -1,51 +1,54 @@
+// Auto-generated file by https://github.com/geodynamics/aspect/doc/make_cite_html.py
```
jdannberg: That link doesn't work for me. Although it's clear what you mean, maybe either put the full link https://github.com/geodynamics/aspect/blob/master/doc/make_cite_html.py, or just say aspect/doc/make_cite_html.py.

tjhei: fixed

jdannberg:
Actually, the comment I had is still not fixed. The plain text list of publications is missing now [on the citing website](https://aspect.geodynamics.org/citing.html), and instead there's just an empty box.

tjhei:
> Can you check what's going on there?

argh. Let me check.

tjhei:
> I opted for a voluntary root canal operation instead

oh, the file is automatically generated, you can just convince yourself that the script is changed in a sensible way in #2664

PR #2664, opened 2018-09-13
update script for database.js for citing.html

tjhei:
- correctly escape umlauts and backslashes
- retain formatting for bibtex as used in manual.bib

used in #2663


tjhei:
as #2663 has been merged, I think we can do the same for this PR.

PR #750, opened 2016-02-11
Postprocessor for executing shell commands.

jperryhouts:
I think this is handy. I have a python script for generating simple figures based on the output, and this plugin allows me to automatically run that script at the end of each timestep. It could also be used for all kinds of other stuff (progress notifications, etc)

One question: I remember there being a mechanism for forcing the order of plugins. Is that the case, and how could I use it to force this plugin to run last?

PR #2666, opened 2018-09-14
Add cpp material model

jperryhouts:
A while ago someone mentioned that it'd be cool to have a 'Function' material model, that uses muparser to calculate material properties. -- This plugin tries to add the same functionality, but keep the speed of a native plugin.

It basically just puts the user's C++ code from the parameter file into a template, compiles that template, and dlopen's the resulting object. The whole thing adds about 15 seconds to the initialization stage of the model, but thereafter should be pretty fast. I haven't done any benchmarks to quantify that, though.

This code feels like a hack, but it works so far, and I think it's kinda elegant to be able to define material properties inside the parameter file (even though the resulting parameter file looks awkward). Here's a working example:

```
subsection Material model
  set Model name = cpp
  set Material averaging = none
  subsection C plus plus
    set Includes = math.h
    set Is compressible = false
    set Reference viscosity = 1.0
    set Global variable definitions = double r, x, y;
    set Update function = x=position[0];y=position[1];r=sqrt(x*x+y*y);
    set Viscosity function = return 1.0;
    set Density function = return (r < 1.5) ? 1.0 : 0.0;
    set Thermal conductivity function  = return 4.7;
    set Thermal expansivity function = return 2e-5;
    set Specific heat function = return 1250;
    set Compressibility function = return 0.0;
    set Entropy derivative pressure function = return 0.0;
    set Reaction function = return 0.0;
  end
end
```

Let me know if this seems worthwhile to include with Aspect, and I can write some tests.

-JPH


jperryhouts:
> I see potential for having input files which can execute any code, potentially harmful to the system which it runs on

There's already a 'command' postprocessor that runs an arbitrary command via `system()`, but it's a good point. We had the same discussion about that plugin as well. I think the consensus was that we should just avoid running parameter files that came bundled with some random movie torrent. And to be careful if we ever decide to expose aspect to some sort of web interface.

I'll get to your other suggestions soon. Thanks again for the review!

source/material_model/cpp.cc
```diff
@@ -0,0 +1,674 @@
...
+              << "    using namespace dealii;\n"
+              << "    class Local_\n"
+              << "    {\n"
+              << "      public:\n"
+              << "        const unsigned int dim = "<<dim<<";\n"
+              <<          user_variable_defs << ";\n"
+              <<          function_defs
+              << "    } local;\n"
+              << "\n"
+              << "    extern \"C\" void\n"
```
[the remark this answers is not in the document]
jperryhouts: Necessary for loading a single function from a shared library (rather than linking a cpp class for example). `extern "C"` prevents name mangling.

source/material_model/cpp.cc
```diff
@@ -0,0 +1,674 @@
...
+      // Make sure the user did not specify a list that included 'none' and
+      // other dependencies.
+      AssertThrow(list.size() == 1
+                  || (std::find(list.begin(),list.end(),"none") == list.end()),
+                  ExcMessage("The property 'none' in the parameter '"+param_name+"' "
+                             "cannot be combined with any other options."));
+    }
+
+    int
+    execute (std::string cmd, bool assert_success)
```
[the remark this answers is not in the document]
jperryhouts: Lots of other places to inject code, too. But aspect runs with user permissions, so we're not exposing anything the user couldn't do anyway.

source/material_model/cpp.cc
```diff
@@ -0,0 +1,674 @@
...
+      return reference_viscosity_param;
+    }
+
+    template <int dim>
+    void
+    CPP<dim>::
+    evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
+             MaterialModel::MaterialModelOutputs<dim> &out) const
+    {
+      if (needs_simulator)
```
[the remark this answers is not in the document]
jperryhouts: Yeah, I went a bit overboard (prematurely) optimizing things. The thought was to make every function pass the bare minimum number of parameters. I don't know if calling get_simulator has any cost associated with it, beyond what's already used just checking this conditional.

There are a bunch of places inside the generated cpp code that do something similar (e.g. generate different function declarations depending on what parameters are actually used). In that code it's probably even less necessary, since the compiler can see whether or not each parameter is actually used.

I wrote it both ways (see [commit 15df124](https://github.com/geodynamics/aspect/tree/15df124ca5e5424568d0eae5d47df4851ec1069d) ), and ended up going with the convoluted one because I spent time on it, and it works. But it definitely adds more complexity.

I'm open to either way, depending on what you and the other developers prefer.

include/aspect/material_model/cpp.h
```diff
@@ -0,0 +1,116 @@
...
+    using namespace dealii;
+
+    /**
+     * A material model that compiles user-defined C++
+     * code at runtime.
+     *
+     * @ingroup MaterialModels
+     */
+    template <int dim>
+    class CPP : public MaterialModel::Interface<dim>, public ::aspect::SimulatorAccess<dim>
```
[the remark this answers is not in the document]
jperryhouts: My thought was that the class name doesn't matter so much, but it might be worthwhile making the class name, the model name, and the parameter subsection consistent with one another. Right now they're `CPP`, `cpp`, and `C plus plus`, respectively.

source/material_model/cpp.cc
```diff
@@ -0,0 +1,674 @@
...
+          needs_simulator = depends(all_user_code, "simulator");
+
+          // Generate and compile the evaluation function
+          std::string cdir = this->get_output_directory() + "cpp_material/";
+          std::string sfname = cdir + "material.cc";
+          std::string ofname = cdir + "libmaterial.so";
+
+          if (dealii::Utilities::MPI::this_mpi_process (MPI_COMM_WORLD) == 0)
+            {
+              execute ("mkdir -p " + cdir);
```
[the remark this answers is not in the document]
jperryhouts: * see last comment. Hopefully if the user has permissions to remove `/`, then they can do that anyway. I could see it being a problem if it were possible to break something here by accident, but I can't think of a case where that would be a problem.

On the other hand, you did remind me that there's already functionality to make a directory, in Utilities. I'll use that instead.

source/material_model/cpp.cc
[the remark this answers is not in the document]
MFraters: I also do that sometimes ;)

The function get_simulator, just returns the pointer to the simulator (see https://github.com/geodynamics/aspect/blob/master/source/simulator/simulator_access.cc#L58) without any checks, so my guess would be that that would not be very slow. I would except it to be something which could be very easily optimized by a compiler. 

On the other hand, an if/else statement is something which may slow down you code as well. That is because the cpu will probably have to wait till the comparison is done, or somehow speculate which branch is more likely to be executed, which might go wrong. 

But as you may see from the language I am using, it is in general very hard to know what will be faster without testing it. My intuition would actually say that using the if/else statement is slower, but I could be completely wrong ;)

I think that for the clarity of the code it would be good simplify it, and only if you know for sure that the more complex code is significantly faster, and it located in a place where it matters (which is the case here), use more complex code.

source/material_model/cpp.cc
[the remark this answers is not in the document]
MFraters: hmm, I am a bit confused here. I thought that calling cpp functions from a cpp library in a cpp program was fine to do without using extern c. That's also what aspect does with the deal.ii library as far as I know. What am I missing here?

source/material_model/cpp.cc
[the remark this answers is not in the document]
MFraters: That seems to be a good solution for this line :)

MFraters:
> There's already a 'command' postprocessor that runs an arbitrary command via `system()`

Ah, I had missed that one. Just for reference for others, it is pull request #750. I see that that one is even more general, allowing users to literally execute anything from the input file. 

I think that it might be a good idea then, to have some cmake variable, which can turn this kind of functionality on and off. By the way, if we would decide we would want something that, it doesn't have to happen in this pull request :) 

I would like to hear some other opinions on this issue.

source/material_model/cpp.cc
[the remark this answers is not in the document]
jperryhouts: The `extern "C"` keyword doesn't mean that the code has to be plain C, it means that the compiler doesn't mangle symbol names, which makes it possible to find the right one with `dlsym()`. It's basically just promising the compiler that you won't overload symbol names or do anything tricky with templates or namespaces.

Aspect links against deal.ii at compile time, so mangled symbol names somehow get magically resolved by the linker. Dynamically loaded plugins are a closer analogy to what's going on here.

Others who understand this better than I do, please correct me here, but as I understand it plugins have special requirements like declaring and registering multiple C++ objects with overloaded symbols (`dim==2` vs `dim==3` for example) so the compiler needs to mangle symbol names.

I'm sure there are other ways to load a shared function at run time, but this is the only one I'm familiar with, and `extern "C"` doesn't hurt anything.

jperryhouts:
I just made a bunch of changes that fix *almost* all of those comments (haven't done anything with the class name yet). Let me know what you think. Thanks again, @MFraters

jperryhouts:
I discussed closing this PR with @gassmoeller at AGU, and we agreed that it shouldn't be included in Aspect because it's system-dependent and will be difficult to maintain over time. We decided that maybe it could be included as a cookbook instead, so that it could be used if someone was particularly interested in it.

After more thought I still agree that it shouldn't be officially supported, but I'm a little disappointed that using it will always require moving files around and maintaining a branch that's not synchronized with upstream. Especially if multiple people are separately maintaining the same changes in their own repositories. Because of that I want to suggest an alternative: What if we add a CMake flag for "experimental" features that are not officially supported? Sort of the opposite of what I've included here (right now this patch adds an `ASPECT_SAFE` flag that is turned off by default). The "prescribed internal velocities" cookbook might fit this category as well, which would keep the cookbooks more consistently focused on use cases rather than code hacks.

I don't feel particularly strongly about this idea, but it does seem more elegant than the alternative. @gassmoeller et al.: What do you think?

gassmoeller:
As discussed, I am in favor of keeping this somehow, e.g. as a cookbook, but i do not quite understand your reasoning. Why would you need to move files around if you would want to use a cookbook material model? If you want to use it for a different model, you can still compile them in the cookbook directory (e.g. `cookbooks/finite_strain`) and then provide the path to the plugin in the input file of the model that uses it. Is it because you need the path to the build folder that is set in the CMakeLists? Could that be done in the cookbook specific CMakeLists file (e.g. `cookbooks/finite_strain/CMakeLists.txt`)? You are right that if you want to have the shared library in the folder of the model you would need to copy it, but that seems similar to copying a parameter file to a new folder to me (and would not require changes to the global CMakeLists or the introduction of a new compile flag).

jperryhouts:
I agree that having people copy it from a cookbook isn't difficult, or even complicated, which is why I don't feel very strongly about it. The main practical benefit would be that it might help users who didn't compile Aspect themselves, or don't know how to compile plugins (which is one of the target audiences of this plugin).

Beyond that it's basically an aesthetic choice. I like the elegance of telling a user "compile with this flag in order to try this feature" over "compile this extra thing, copy it to your project folder, and add the library's name to the Additional shared libraries parameter". Also the thing about narrowing the focus of cookbooks to use-cases rather than code samples would make the manual's organizational structure a bit more obvious.

include/aspect/material_model/cpp.h
```diff
@@ -0,0 +1,215 @@
...
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    /**
+     * A structure for organizing user code snippets.
+     */
+    struct UserCode
```
bangerth: Can you put this into a namespace specific to the CPP material model? Like `aspect::MaterialModel::internal::CPP::UserCode`? This is lengthy, but hides the rather generic name from all other uses in the namespace.

include/aspect/material_model/cpp.h
```diff
@@ -0,0 +1,215 @@
...
+      std::string reaction_function;
+    };
+
+    /**
+     * Converts a string describing a nonlinear dependency
+     * to its equivalent `Dependence'.
+     */
+    inline
+    NonlinearDependence::Dependence
+    str2dep (const std::string depstr)
```
bangerth: This function would probably be better located right next to the `Dependence` enum. I'd name it `string_to_dependence()` to keep with out usual naming conventions.

include/aspect/material_model/cpp.h
bangerth: Alternatively, and maybe better, why not put it into the material model itself? Or are these global parameters in the `.prm` file?

include/aspect/material_model/cpp.h
```diff
@@ -0,0 +1,215 @@
...
+        }
+      return NonlinearDependence::uninitialized;
+    }
+
+    /**
+     * Execute a command using the system() command processor. Return the
+     * exit code.
+     */
+    int
+    execute (const std::string &cmd)
```
bangerth: This function would probably be better suited in the .cc file.

include/aspect/material_model/cpp.h
```diff
@@ -0,0 +1,215 @@
...
+         */
+
+      private:
+        double reference_viscosity_param;
+        bool is_compressible_param;
+
+        /**
+         * Define function pointer for the new evaluate function.
+         */
+        typedef void (*eval_t)(const MaterialModel::MaterialModelInputs<dim> &in,
```
bangerth: We've switched to C++11 now, where this can be more easily expressed as
```
  using eval_t = void (const ....);
```

source/material_model/cpp.cc
```diff
@@ -0,0 +1,623 @@
...
+
+      return "";
+    }
+
+    /*
+     * Make sure user didn't specify 'none' along with other nonlinear
+     * dependencies.
+     */
+    void
+    assert_valid_deps (std::vector<std::string> list, std::string param_name)
```
bangerth: make the input arguments const references

source/material_model/cpp.cc
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    /*
+     * Write a simple CMakeLists.txt for the generated code, to simplify
+     * linking with Aspect and Deal.II. Basically, treat the code as a plugin,
+     * since it occupies the same namespace, and should expose most of the same
+     * interfaces as a native plugin.
+     */
+    void
+    generate_cmakelists (const std::string cdir)
```
bangerth: same here and for the other functions below: non-scalar input arguments should be const reference

source/material_model/cpp.cc
```diff
@@ -0,0 +1,623 @@
...
+    {
+      std::ofstream cmfile;
+      cmfile.open((cdir + "CMakeLists.txt").c_str(), std::ios_base::out);
```
bangerth: might want to check that `cdir` ends in a slash

bangerth: also, just write
```
  std::ofstream cmfile (cdit + "CMakeLists.txt");
```
;-)

source/material_model/cpp.cc
```diff
@@ -0,0 +1,623 @@
...
+    } while (false)
+
+      MKFUNCTIONS(viscosities,user_code.viscosity_function);
+      MKFUNCTIONS(densities,user_code.density_function);
+      MKFUNCTIONS(thermal_conductivities,user_code.thermal_conductivity_function);
+      MKFUNCTIONS(thermal_expansion_coefficients,user_code.thermal_expansivity_function);
+      MKFUNCTIONS(specific_heat,user_code.specific_heat_function);
+      MKFUNCTIONS(compressibilities,user_code.compressibility_function);
+      MKFUNCTIONS(entropy_derivative_pressure,user_code.entropy_derivative_p_function);
+      MKFUNCTIONS(entropy_derivative_temperature,user_code.entropy_derivative_t_function);
```
bangerth: You can do `#undef MKFUNCTIONS` here to avoid inadvertent use later on.

source/material_model/cpp.cc
```diff
@@ -0,0 +1,623 @@
...
+
+      // Write the code to an output file. Make liberal use of semicolons
+      // after user code, since it's easy to forget them in parameter files,
+      // and extra ones don't bother non-pedantic compilers.
+      std::ofstream srcfile;
+      srcfile.open((fname).c_str(), std::ios_base::out);
```
bangerth: same here

source/material_model/cpp.cc
```diff
@@ -0,0 +1,623 @@
...
+              MPI_Bcast(&error, 1, MPI_INT, 0, MPI_COMM_WORLD);
+              AssertThrow (error == 0,
+                           ExcMessage (std::string("Compiling material model failed. See build log at <")
+                                       + cdir + "build.log>."));
+            }
+          else
+            {
+              MPI_Bcast(&error, 1, MPI_INT, 0, MPI_COMM_WORLD);
+              if (error != 0)
+                throw aspect::QuietException();
```
bangerth: Nice job with the quiet exception!

jperryhouts:
> Very interesting approach!
> 
> One of the things you'll figure out is that you compile the shared lib on the compute nodes, not the front-end node. But on many systems, the compute nodes don't have the compilers installed, and probably not cmake either. I don't know whether that's something that can be detected. Do you know where error output goes when you call `execute()`?

@bangerth Wow I put this on the back burner after it looked like it was heading towards cookbookland. I apparently missed the notification that you reviewed it too- sorry about that! It resurfaced for me when I was cleaning up some unused branches on github.

Do you still think this is worth finishing, or should I close the PR? When I discussed it with @gassmoeller (at AGU 2018, I think!) he wasn't keen on the idea except maybe as a cookbook, and I agreed that it could be challenging to provide ongoing user support for a feature like this. I'd still need to address your review comments in any case.

jperryhouts:
More digging about this turned up the [Tiny C Compiler (TCC)](https://savannah.nongnu.org/projects/tinycc), which could be an interesting way to do something similar in a more self-contained way. It wouldn't be as powerful as a properly linked shared library, but it would still enable much more flexibility than muparser, like loops, table look-ups, more complex conditionals, etc. The upside would be that it would be fast, and would not rely on access to a compatible compiler suite on compute nodes.

I'll also mention that full HPC deployments are not necessarily the primary target of this material model. I think it would still be useful as a low-barrier access point for users to begin experimenting with the plugin architecture without the boilerplate code. If this were distributed within the docker image, then new users could start writing low-level code immediately that could later be transplanted directly into a proper plugin or a pull request. -- But of course if it works for HPC as well that would be great.

jperryhouts:
I hacked together an inelegant proof of concept using the TCC compiler ([jperryhouts:add_tcc_material_model](https://github.com/jperryhouts/aspect/tree/add_tcc_material_model))

It's way less flexible than the dynamically linked C++ version in this pull request, but it doesn't have any runtime dependencies, and it's *really* fast-- no noticeable compiling delay during initialization. It compiles to static object code in memory immediately after parsing the parameter file.

It's like a much more flexible muparser rather than a full-blown plugin templater, which turns out to be a lot more useful than I expected for doing actual work, not just as a teaching or demo tool. It allows you to do things like:

    set Viscosity function = const double e2inv = \
                                 strain_rate[0][0] * strain_rate[1][1] \
                               - strain_rate[0][1] * strain_rate[0][1]; \
                                \
                             const double min_edot = 1e-4, min_eta = 1e0, max_eta = 1e8, max_e2inv2 = 1e8; \
                                \
                             if (e2inv == 0) { return max_eta; } \
                                \
                             const double e2inv2 = e2inv*e2inv; \
                             if (e2inv2 > max_e2inv2) { e2inv2 = max_e2inv2; } \
                             double eta = 1e-2 / (5e-12*e2inv2 + min_edot*min_edot); \
                                \
                             if (compositions[0] > 0.5) { eta *= 0.1; } \
                                \
                             if (eta > max_eta) {  return max_eta; } \
                             else if (eta < min_eta) { return min_eta; } \
                             else { return eta; }

which looks a bit messy, but it implements a functional (albeit non-physical) composition-dependent nonlinear rheology that can be modified without recompiling anything. Also, because it's part of the parameter file you never lose track of exactly which version of your code was used for any particular model run.

The figure below is from a couple time steps into a toy model using that function.
Screenshot from 2020-11-19 18-46-57

MFraters:
Wow, that is really cool! Not only in functionality, but also in relative ease of use and speed. It would make aspect a lot easier to just tinker with. Here are some initial thoughts I have on this approach: 
1. I am a little bit afraid of overuse of this in the long run, making the input files partial cpp files. With that I mean that small scripts tend to grow and may lead to duplicated efforts and in the form the code would be hard to read, debug and test the code. On the other hand, the fact that it probably becomes unwieldy quite quickly might give it a natural limit. I don't think that should be a reason to not get this into aspect, but it might be good to mention in the documentation the intended scope. 
2. What are the limitations of this? For example, can you use Assert macro's or to get back to an earlier discussion, can you for example call the execute function from this?
3. You could make it look prettier and more readable if you would put the code in a separate file and in the input file you just define what file you want to load. The downside of that would be that you get quite a few files and that the code is no longer directly in the input file, so that would do away with one of your upsides.
4. It looks like TCC is available for all operating systems, so that is great :)

jperryhouts:
> Wow, that is really cool! Not only in functionality, but also in relative ease of use and speed. It would make aspect a lot easier to just tinker with. Here are some initial thoughts I have on this approach:
> 
>     1. I am a little bit afraid of overuse of this in the long run, making the input files partial cpp files. With that I mean that small scripts tend to grow and may lead to duplicated efforts and in the form the code would be hard to read, debug and test the code. On the other hand, the fact that it probably becomes unwieldy quite quickly might give it a natural limit. I don't think that should be a reason to not get this into aspect, but it might be good to mention in the documentation the intended scope.

It's certainly tempting to use this a lot, but I agree that it's pretty self limiting in both unwieldyness and in its actual capabilities. The original version (in this PR) would have been more problematic in that sense because it has fewer limitations.

>     2. What are the limitations of this? For example, can you use Assert macro's or to get back to an earlier discussion, can you for example call the execute function from this?

I don't really know the limitations at this point. You can't use macros, since those are typically handled by a separate preprocessor, and you can't use any C++. It seems to let me include libraries, but I don't know how that works to be honest, since it doesn't have access to a linker. I've just been avoiding touching that issue while experimenting with this, which is why the example in my comment above doesn't use any math.h functions. I wasn't positive they would work correctly.

>     3. You could make it look prettier and more readable if you would put the code in a separate file and in the input file you just define what file you want to load. The downside of that would be that you get quite a few files and that the code is no longer directly in the input file, so that would do away with one of your upsides.

I agree, but loading separate files would definitely detract from my favorite part of this, which is keeping everything self-contained. It would be nice if deal.ii could parse multi-line parameter inputs (e.g. triple quotes like Python or triple backticks like markdown), rather than appending backslashes to every line. That would make things a lot more readable and less bug prone.

>     4. It looks like TCC is available for all operating systems, so that is great :)

I noticed on the main developer's webpage they say they're no longer working on it, which was a little concerning, but then I looked at the git repository and the latest commit was earlier that same day. It looks like others have stepped in and development is still very active. It's also encouraging that the Debian/Ubuntu package manager has the latest release version, so presumably it isn't just a backwater abandoned project.

One other downside I haven't figured out a good solution for is handling compiler errors, and more gracefully failing if the input does not run correctly. The way it's implemented right now will spit out some terse compiler warnings to stderr during the initialization step, and then fail with segmentation faults if your code doesn't work. Really annoying especially now that I'm used to Aspect's assert statements giving me readable and helpful error messages. At the very least it would be nice to catch compile errors early. The seg fault thing might just come with the territory of handling raw C code, and at least it would usually become obvious on the first time step.

MFraters:
I currently see no problem with this finding it's way into the main version of aspect in some form, but it would be good to get some more opinions on this early on in the process. Since the code will be very different from what is in this pull request, I would suggest you close this pull request and open a new one as WIP. I think that would be a better place to discuss this different approach than here.

jperryhouts:
> I currently see no problem with this finding it's way into the main version of aspect in some form, but it would be good to get some more opinions on this early on in the process. Since the code will be very different from what is in this pull request, I would suggest you close this pull request and open a new one as WIP. I think that would be a better place to discuss this different approach than here.

Agreed. It's going to take some work to even decide how the interface to this should work. It's too bad there's no official developers meeting during AGU, but maybe whoever does show up would have some feedback on that. I'll hold off on doing anything else with this until I get a chance to talk to others.

jperryhouts:
After reading through the docs (aka source code comments-- the docs are pretty sparse), it looks like TCC does bundle its own macro preprocessor and its own linker (seems obvious in retrospect, but compilers are all sort of magic black boxes to me. Who's to say you can't have a compiler without a linker?) I'm hoping that static linking will still work to make this independent of particular libraries installed on the compute node. Or perhaps bundling like what dealii does with muparser.

Anyway, just stuff to think about for now. I'll pick this back up once I get a chance to talk it over with others.

tjhei:
Considering how old the PR is and that it did not get any updates, we can probably close this PR.

PR #2667, opened 2018-09-15
Fix reference temperature for discontinuous element face terms

gassmoeller:
This should fix a bug reported by @maxrudolph if you combine ALA or Boussinesq models with DG temperature elements. It can lead to oscillations and non-convergence of the advection solver. It still needs a test, but I wanted to make it available.


gassmoeller:
Added a test that crashes on master, but succeeds and gives expected results with this pull request. Should be ready for a review.

source/simulator/assemblers/advection.cc
```diff
@@ -613,6 +613,16 @@ namespace aspect
               this->get_material_model().evaluate(scratch.neighbor_face_material_model_inputs,
                                                   scratch.neighbor_face_material_model_outputs);
 
+              if (parameters.formulation_temperature_equation ==
+                  Parameters<dim>::Formulation::TemperatureEquation::reference_density_profile)
+                {
+                  const unsigned int n_q_points = scratch.neighbor_face_finite_element_values->n_quadrature_points;
```
tjhei: I think the ``n_q_points`` variable in the outer scope would be correct too, so delete this line.

maxrudolph:
I ran our more complex test case with prescribed plate motions and internal heating and there are no more convergence problems with this branch, but we still see very large temperature over/undershoots even with the bound preserving limiter enabled. In the attached figure, the blue curve shows the final temperature profile from the DG run. The orange curve shows the depth average for the DG case after 150 Myr, and somehow at the surface and CMB the temperatures have departed from the prescribed boundary temperature.
image

gassmoeller:
Ok, great. It sounds like the changes have resolved the crashes. I think I found a different issue with the bound preserving limiter that could explain the remaining oscillations, but I will open that as a separate PR. 

I have removed the `n_q_point` variable altogether, since it does not really matter that these points are quadrature points, we simply want to replace all densities by reference densities. @tjhei: Does that address your comments?

tjhei:
Is the compiler smart enough to see that some_vector.size() is constant in the for loop condition? My suggestion was to just use the ``_n_q_point`` variable that already existed in an outer scope.

gassmoeller:
The problem with the outer scope is that currently the different FEFaceValues (for own cell, neighbor cell, and subface) are created with the same quadrature, but it does not need to be that way. That is why I would like to use the size of the current density object (or the number of quadrature points of the currently used FEFaceValues object) instead of relying on the fact that they use the same quadrature. But you are right, I should not put that into the loop condition. I can move it before the loop, e.g. `const unsigned int evaluation_points = .... .size(); for (...)`. Do you agree?

tjhei:
> Do you agree?

This is equivalent to what you had in the beginning right?

> re created with the same quadrature, but it does not need to be that way.

I don't understand. You are filling ``scratch.neighbor_face_material_model_outputs`` with a loop (for example in line 622). You then access this array (for example in line 711) using an index ``q` which is 0<=q<=n_q_points. This means those two loop bounds should always be the same or not?

gassmoeller:
Ah, I understand what you were aiming at. The place where we use the material model outputs already assumes that all three `FEFaceValues` have the same number of quadrature points, so it is unnecessary to prepare the place where we fill the MaterialModelOutputs for the case where this is not true. I simplified the loops using the existing `n_q_points` variable, but added Assertions that will trigger if anyone ever creates the FEFaceValues with inconsistent quadrature rules.

tjhei:
> but added Assertions

:+1:

tjhei:
This is ready to merge once testers are happy.

gassmoeller:
It looks like the old astyle tester had a hickup, but the astyle tester of both jenkins instances passed, so I am going to merge.

PR #2668, opened 2018-09-17
Negative longitude in uniform radial particle generator

bartniday:
This allows it to make particle distributions that cross the prime meridian.

### DIFF
--- a/source/particle/generator/uniform_radial.cc
+++ b/source/particle/generator/uniform_radial.cc
@@ -168,11 +168,11 @@ namespace aspect
                                    "Maximum radial coordinate for the region of particles. "
                                    "Measured from the center position.");
                 prm.declare_entry ("Minimum longitude", "0",
-                                   Patterns::Double (0,360),
+                                   Patterns::Double (-180,360),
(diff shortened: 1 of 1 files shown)


bartniday:
This just changes the longitude limits for that particular particle generator, so I don't think any other plugins will see it. The only other place I've run into this was the chunk geometry model, which already allows negative longitude.

It might be worth going through everything that uses spherical coordinates to check that the limits and defaults make sense and the coordinates are defined in a consistent way though. I remember fixing a few similar things at the hackathon.

naliboff:
@bartniday - Yes, as long as the relevant geometry model allows negative longitudes, then there should be no direct compatibility issues elsewhere.

tjhei:
> Looking for a way to re-started the jenkins testers.

I did that manually.

gassmoeller:
@naliboff: Just to clarify: The longitude is measured relative to the center position of this particle generator, not relative to any global coordinate. That is why it should work with all geometries (not just spherical models). This looks good to go. Thanks!

PR #2669, opened 2018-09-20
Fix assert two merged boxes in point_in_domain on free surface.

MFraters:
When running a model with this geometry model and a free surface the assert was triggered before the first timestep, which is not consistent with the assert message. I think this should fix it. @anne-glerum Could you look at this?

### DIFF
--- a/source/geometry_model/two_merged_boxes.cc
+++ b/source/geometry_model/two_merged_boxes.cc
@@ -277,8 +277,8 @@ namespace aspect
     bool
     TwoMergedBoxes<dim>::point_is_in_domain(const Point<dim> &point) const
     {
-      AssertThrow(this->get_free_surface_boundary_indicators().size() == 0 ||
-                  this->get_timestep_number() == 0,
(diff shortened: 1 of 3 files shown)


gassmoeller:
I do not quite believe this would be the correct solution. Could this error be related to #2362 - the fact that the timestep number before the first timestep is now set to `invalid_unsigned_int` instead of 0? There was a change to `Box` in that PR, but not to `TwoMergedBoxes`, presumably, because there is no test for this case.

anne-glerum:
@MFraters I'd agree with Rene that this is not the correct solution. Changing the timestep number is probably the way to go.

MFraters:
Sorry for my late reply. I think that i should work. I will add a simple test.

bangerth:
Once #2972 , you could probably use the function introduced there to check whether we have started time stepping.

MFraters:
ok, I have now added the test. I can update the code once #2972 is merged.

gassmoeller:
Looks good. Thanks!

PR #2670, opened 2018-09-21
Fix input check for bound preserving limiter option

gassmoeller:
It makes no sense to check for the correct length of the `global_composition_max_preset` variable if only the temperature limiter is selected. The check is only relevant if the composition limiter is selected.


tjhei:
ready to merge when testers are happy!

PR #2671, opened 2018-09-22
Fix bound preserving limiter in 3d

gassmoeller:
I think there were indexing errors in the bound preserving limiter 3D case that could explain why @maxrudolph sees values outside of the bounds of the limiter in his models. The first commit fixes the indexing errors. The second and third commit of this PR simply refactor the code to avoid the index computation altogether and prevent future mistakes.

I tried to create a testcase, but could not find a model where the bound preserving limiter is actually needed. @egpuckett or @yinghe616 do you maybe have testcases that test the bound preserving limiter?

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -1452,19 +1452,19 @@ namespace aspect
                 if (std::abs(max_solution_local-local_solution_average) > std::numeric_limits<double>::min())
                   {
                     theta = std::min(theta, (max_solution_exact_global-local_solution_average)
-                                                     / (max_solution_local-local_solution_average));
+                                     / (max_solution_local-local_solution_average));
                   }
                 if (std::abs(min_solution_local-local_solution_average) > std::numeric_limits<double>::min())
                   {
                     theta = std::min(theta, (min_solution_exact_global-local_solution_average)
-                                                     / (min_solution_local-local_solution_average));
+                                     / (min_solution_local-local_solution_average));
                   }
 
                 AssertThrow(theta>=0.0,ExcMessage("The bound preserving factor is smaller than zero. "
-                    "This should not happen, please contact the developers. "
-                    "Minimum solution in cell is: " + Utilities::to_string(min_solution_local) +
-                    ". Maximum solution in cell is: " + Utilities::to_string(max_solution_local) +
-                    ". Average solution in cell is: " + Utilities::to_string(local_solution_average)));
+                                                  "This should not happen, please contact the developers. "
+                                                  "Minimum solution in cell is: " + Utilities::to_string(min_solution_local) +
+                                                  ". Maximum solution in cell is: " + Utilities::to_string(max_solution_local) +
+                                                  ". Average solution in cell is: " + Utilities::to_string(local_solution_average)));
 
 
                 /* Modify the advection degrees of freedom of the numerical solution.
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -1357,6 +1357,8 @@ namespace aspect
         default:
           Assert (false, ExcNotImplemented());
       }
+
+    Assert (quadrature_points.size() == n_q_points, ExcInternalError());
     Quadrature<dim> quadrature_formula(quadrature_points);
 
     // Quadrature rules only used for the numerical integration for better accuracy
@@ -1440,6 +1442,7 @@ namespace aspect
                     local_solution_average += values_0[q]*fe_values_0.JxW(q);
                   }
                 local_solution_average /= local_area;
+
                 /*
                  * Define theta: a scaling constant used to correct the old solution by the formula
                  *   new_value = theta * (old_value-old_solution_cell_average)+old_solution_cell_average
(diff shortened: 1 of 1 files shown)


maxrudolph:
Thank you @gassmoeller. I will run a test with this to see whether the BPL now works for the energy equation in a 3D spherical case.

source/simulator/helper_functions.cc
```diff
@@ -1457,19 +1457,26 @@ namespace aspect
...
                   {
-                    theta = std::min(theta, std::abs((min_solution_exact_global-local_solution_average)
-                                                     / (min_solution_local-local_solution_average)));
+                    theta = std::min(theta, (min_solution_exact_global-local_solution_average)
+                                                     / (min_solution_local-local_solution_average));
                   }
 
+                AssertThrow(theta>0.0,ExcMessage("The bound preserving factor is smaller than zero. "
+                    "This should not happen, please contact the developers. "
+                    "Minimum solution in cell is: " + Utilities::int_to_string(min_solution_local) +
```
tjhei: why not ``to_string``? int does not seem appropriate here?

tjhei: oops, that was only in the first commit...

tjhei:
The code for combining the quadratures is still messy, but I agree that it is much better now.

tjhei:
Can you please add a check that the ``size()`` after the loops is indeed equal to ``n_q_points``?

egpuckett:
I don't have a test case in mind off hand.  Ying tested the algorithm in 3D with a 3D version of the falling box problem.  I am communicating with her via email this weekend. I'll ask if she has still has the initial conditions for that problem.  @maxrudolph do you want to collaborate on creating a problem in a spherical domain that needs the limiter?

gassmoeller:
I addressed the comments, but our test cases show another problem: Apparently `theta` can become negative if the average value of a cell becomes negative. I got the following output:
```
The bound preserving factor is smaller than zero. This should not happen, please contact the developers. Minimum solution in cell is: -0.0002205584259760636. Maximum solution in cell is: 5.3791808445420153e-05. Average solution in cell is: -1.3282081848247357e-06
```
for the test discontinuous_composition_bound_preserving_limiter_3d. We need to think about what the limiter should do in such cases. In this case the average should not be below 0 (which is the minimum bound for the limiter), but apparently oscillations have caused it nevertheless. Should we bound `theta` to the range of 0,1 even if the average of the cell is outside the bounds? Should we adjust the average?

egpuckett:
@gassmoeller Can we meet tomorrow so you can show me the  'offending' cell or cells. I may have some ideas, but I need to see the values in the neighboring cells at at least one previous time step and possibly more.

source/simulator/helper_functions.cc
```diff
@@ -1457,19 +1454,26 @@ namespace aspect
...
+                    theta = std::min(theta, (max_solution_exact_global-local_solution_average)
+                                     / (max_solution_local-local_solution_average));
                   }
                 if (std::abs(min_solution_local-local_solution_average) > std::numeric_limits<double>::min())
                   {
-                    theta = std::min(theta, std::abs((min_solution_exact_global-local_solution_average)
-                                                     / (min_solution_local-local_solution_average)));
+                    theta = std::min(theta, (min_solution_exact_global-local_solution_average)
+                                     / (min_solution_local-local_solution_average));
                   }
```
yinghe616: Hi, Rene, the theta defined originally using absolute value to make sure it to be a positive number was correct.  For you better understanding the definition, you can check the equation (2.8) at [On maximum-principle-satisfying high order schemes for scalar conservation laws](https://www.math.purdue.edu/~zhan1966/research/paper/maximum_principle.pdf)

yinghe616:
There could be several reasons to see the values on the outside surface is beyond the ideally prescribed bound for the temperature equation. 1) Theoretically, the limiter will preserve the pre-scribed bound if the time advance scheme is strongly stability preserved scheme, which unfortunately has to be explicit scheme like forward Euler but ASPECT using semi or fully implicit scheme which one I couldn't recall, so this will downgrade the performance of DGBP limiter. But  one can still expect that using DGBP limiter can reduce the overshoot/undershoot produced by using the DG method only. So if using DG method only to run the example also give very big error at the outside surface, it might be the best DGBP limiter can do already. 2) Inconsistent boundary conditions or added reaction term; the limiter is designed based on the conservation laws for the current implementation. That's being said, if the temperature equation can be rewritten in the equation format of satisfying conservation law, the DG/DGBP limiter should work well. It also means the DGBP limiter doesn't always work well for the temperature equation but it should work well for the compositional equation without reaction term under an incompressible flow as it can be always rewritten in the conservative form. 3) Perhaps there exists index error as you pointed out, but I didn't see the essential difference of initializing the quadrature points between before and now. However, it's much clean now as Timo mentioned. Did you print out the size info to see any difference?  4) A good example can be to modify any current temperature model problem such that the diffusivity to be relatively small compared to the advection term. Also making discontinuity of either material property or  the underline velocity flow, this will give a good example with strong discontinuity and possible high overshoot/undershoot at beginning used to test the DG/DGBP method. I might have an example configure file in 2D. If I can find it, I'll update it to you later.

egpuckett:
> I addressed the comments, but our test cases show another problem: Apparently `theta` can become negative if the average value of a cell becomes negative. I got the following output:

@gassmoeller For what problem does the average value of a cell become negative? If the lower bound on the exact solution is 0,  then it strikes me that something is not correct.  Perhaps the computation is under-resolved.  If not, I'm concerned that there is a problem with either the algorithm or the implementation.

gassmoeller:
@yinghe616: Thanks for the reference, you are right the old formulation was the one that was in the paper. I still wonder if this is correct if the average cell value is outside of the prescribed bounds though. Obviously the limiter can not correct the average, because it is designed to not change the average, but does it compute a reasonable `theta` in these cases, or can it introduce oscillations in these cells (by computing `theta` as some oscillating value)?
Btw: The change between the old indexing and the new one lies here: https://github.com/geodynamics/aspect/pull/2671/commits/32a6556797cace65d05f31b2d51b0ef83b3a306c#diff-bbf48e1dd20b08ffc36659d03d91bc8aL1321 and here: https://github.com/geodynamics/aspect/pull/2671/commits/32a6556797cace65d05f31b2d51b0ef83b3a306c#diff-bbf48e1dd20b08ffc36659d03d91bc8aL1354. 
Basically, the old implementation computed all the right points, but when it stored them in the `quadrature_points` vector it overwrote some of the entries multiple times, and left others at the default constructed value (Point 0,0,0). The new implementation simply appends all the points to the vector so that nothing can be overwritten. I also added an Assert that checks that exactly as many points are appended as expected.

@egpuckett: The test that produces the problem is `tests/discontinuous_composition_bound_preserving_limiter_3d.prm`. We could certainly look into what causes the strange average in this (possibly under-resolved) model, but since this sounds like a bigger project lets keep it out of this PR. I will update the PR to only contain the bug fixes, and remove the functional change. I will open an issue for the functional question and link it to this PR.

gassmoeller:
This now only contains the fixes, and should be ready if the testers are happy.

tjhei:
the CIG tester result is missing. Otherwise okay to merge.

gassmoeller:
Had a DNS lookup issue, I restarted the job and it is running now.

PR #2672, opened 2018-09-25
Fix typo in documentation

bartniday:


gassmoeller:
Looks like one of the testers had a connection issue, but the other succeeded, and it is only a documentation fix anyway. Ready to merge, and thanks for the fix! :+1:

Issue #2613, opened 2018-08-16
Rethink inclusion of the WorldBuilder via a git submodule


bangerth:
It is not lost on me that the reason WorldBuilder was added to ASPECT via a git submodule (in #2571) is because I took (or had to take) a break from ASPECT development and didn't get to reviewing @MFraters's patches. That's my fault.

Be that as it may, we may want to rethink this structure. Submodules have the distinct disadvantage that (i) we need to update build instructions to include flags people don't typically use (`--recursive`), see #2587 , and (ii) that a `git pull upstream master` does not get the newest version of the WorldBuilder into a working tree. Both are annoying and confusing. The latter is also going to present a source of awkward cases where the WorldBuilder changes its interface, ASPECT is correspondingly adjusted, but someone just pulls from upstream and can't compile anymore because they forgot to *also* update the submodule.

For this reason, we've generally treated external projects as external to the tree. My preference would therefore be if, in the long run (before the next release), we could undo the current structure and either treat WorldBuilder as entirely external (like PETSc, Trilinos, ... in deal.II) or if we could import a fixed version that is updated periodically (like we do with BOOST or the TBB, for example).

What do others think?

tjhei:
> What do others think?

Let's see:

submodules:
- better separation (including history) (+)
- making it a separate project at some point is more natural (+)
- cloning/updating etc. is less obvious for users (-)
- potential dependency problems when not updating the submodule (-)
- other changes to release tasks/testing/etc. needed (-)

external project:
- no dependency problems (+)
- users might not install or able to use it (-)
- extra effort when installing (-)
- no testing unless we put in extra effort to set it up (-)

bundled:
- easy setup/usage/testing (+)
- no separation of history, copying over files often is not great (-)

I have to admit that I don't see a clear winner here. I think it depends a bit on the development speed of the WorldBuilder. If we don't have to update very often, the "bundled" approach looks best. On the other hand, if a lot of development happens, the other two options look better.

So, I am fine with either approach but I prefer a solution that allows for a solid testing infrastructure.

gassmoeller:
I agree with Timo that I see not one solution that is uniformly better than the others. Long-term I am fine with all options, but would add that we were not very good at reviewing the large chunks of code that the world builder pull requests included in the past. @MFraters needs to be able to make fast progress on this since he is on a tight schedule, so the bundled version is at the moment out of the question (at least until he has finished his thesis). If we want to allow for easy testing we need to stick with the submodule for the moment.

[...]

PR #2673, opened 2018-09-26
Fix some typos

gassmoeller:
Some fixes for #2574, nothing critical, just things I noticed but did not want to delay the PR further for. 

@MFraters: When I updated the world builder submodule git commited the change to the version of the submodule as well. Does that mean we need to manually update the submodule every time we want to have a newer version of the world build? Or the other way around: Does the version of the world build in ASPECT stay the same if we do not update it? I guess this ties into #2613, and we should continue the discussion there at some point in the future, but here I was just curious. I can remove the update to the world builder when that is wanted, but the current version of the world build that is part of ASPECT has some serious bugs (like it returns the composition as a bool instead of a double), so maybe we want to keep the update in this PR?

### DIFF
--- a/source/initial_composition/world_builder.cc
+++ b/source/initial_composition/world_builder.cc
@@ -49,9 +49,9 @@ namespace aspect
   {
     ASPECT_REGISTER_INITIAL_COMPOSITION_MODEL(WorldBuilder,
                                               "world builder",
-                                              "Specify the composition in through the World Buider. "
+                                              "Specify the initial composition through the World Builder. "
                                               "More information on the World Builder can be found at "
-                                              "https://geodynamicworldbuilder.github.io ."
+                                              "https://geodynamicworldbuilder.github.io. "
                                               "Make sure to specify the location of the World Builder file "
                                               "in the parameter 'World builder file'.")
   }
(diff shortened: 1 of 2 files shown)


gassmoeller:
Ok, I removed the world builder update. Only typos now, and should be ready to merge when the testers are happy.

source/initial_composition/world_builder.cc
```diff
@@ -49,9 +49,9 @@ namespace aspect
   {
     ASPECT_REGISTER_INITIAL_COMPOSITION_MODEL(WorldBuilder,
                                               "world builder",
-                                              "Specify the composition in through the World Buider. "
+                                              "Specify the initial composition through the World Builder. "
                                               "More information on the World Builder can be found at "
-                                              "https://geodynamicworldbuilder.github.io ."
+                                              "https://geodynamicworldbuilder.github.io. "
```
tjhei: the space after "io" is probably there so that is it not seen as part of the url. How are we handling URLs in parameters, anyway? With or without ``{url}``?

gassmoeller: I do not think there is a policy yet, but including links is reasonable. I have checked that `\url` works in the pdf, and included it.

PR #2677, opened 2018-10-01
design a structure for interpolating material model outputs onto comp…

jdannberg:
…ositional fields

This is something I started to work on during the hackathon together with @alarshi because she wanted to copy something from the material model outputs into compositional fields and then solve a diffusion problem for this field, but it turned out that interpolating properties onto compositional fields is also useful for a number of other things. In this case, we wanted to compute something expensive in the material model that will be used in several different places (the equilibrium grain size), so it is convenient to store it in a compositional field so that it can be accessed easily. 

In the future, I see this as a method that could be combined with other compositional field advection/diffusion methods, and it is a first step to getting part of the changes suggested in #2532 back into the main version of ASPECT.

### DIFF
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -798,6 +798,35 @@ namespace aspect
 
 
     /**
+     * Additional output fields for prescribed field outputs to be added to
+     * the MaterialModel::MaterialModelOutputs structure and filled in the
+     * MaterialModel::Interface::evaluate() function.
+     *
+     * If the advection scheme "prescribed field" is selected, then these
+     * material model outputs will be interpolated onto the corresponding
+     * compositional field.
+     */
+    template <int dim>
+    class PrescribedFieldOutputs : public NamedAdditionalMaterialOutputs<dim>
+    {
+      public:
+        PrescribedFieldOutputs (const unsigned int n_points,
+                                const unsigned int n_comp);
+
+        virtual std::vector<double> get_nth_output(const unsigned int idx) const;
+
+        /**
+         * Prescribed field outputs for all compositional fields at the evaluation points
+         * that are passed to the instance of MaterialModel::Interface::evaluate()
+         * that fills the current object.
+         * copy_properties[q][c] is the prescribed field output at the evaluation point q
+         * for the compositional field with the index c.
+         */
+        std::vector<std::vector<double> > copy_properties;
+    };
+
+
+    /**
      * A class for additional output fields to be added to the RHS of the
      * Stokes system, which can be attached to the
      * MaterialModel::MaterialModelOutputs structure and filled in the
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -822,7 +822,7 @@ namespace aspect
          * copy_properties[q][c] is the prescribed field output at the evaluation point q
          * for the compositional field with the index c.
          */
-        std::vector<std::vector<double> > copy_properties;
+        std::vector<std::vector<double> > prescribed_field_outputs;
     };
 
 
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -805,6 +805,11 @@ namespace aspect
      * If the advection scheme "prescribed field" is selected, then these
      * material model outputs will be interpolated onto the corresponding
      * compositional field.
+     *
+     * However, note that this structure always has as many prescribed field
+     * outputs as there are compositional fields, even if not all of them
+     * are using the "prescribed field" method. It is the responsibility
+     * of the individual material models to fill the correct entries.
      */
     template <int dim>
     class PrescribedFieldOutputs : public NamedAdditionalMaterialOutputs<dim>
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -819,7 +819,7 @@ namespace aspect
          * Prescribed field outputs for all compositional fields at the evaluation points
          * that are passed to the instance of MaterialModel::Interface::evaluate()
          * that fills the current object.
-         * copy_properties[q][c] is the prescribed field output at the evaluation point q
+         * prescribed_field_outputs[q][c] is the prescribed field output at the evaluation point q
          * for the compositional field with the index c.
          */
         std::vector<std::vector<double> > prescribed_field_outputs;
--- a/source/material_model/interface.cc
+++ b/source/material_model/interface.cc
@@ -854,6 +854,15 @@ namespace aspect
 
         return names;
       }
+
+
+      std::vector<std::string> make_prescribed_field_output_names(const unsigned int n_comp)
+      {
+        std::vector<std::string> names;
+        for (unsigned int c=0; c<n_comp; ++c)
+          names.push_back("prescribed_field_output_C" + Utilities::int_to_string(c));
+        return names;
+      }
     }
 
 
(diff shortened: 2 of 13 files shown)


source/material_model/interface.cc
```diff
@@ -854,6 +854,15 @@ namespace aspect
 
         return names;
       }
+
```
gassmoeller: Add another empty line for consistency

source/simulator/helper_functions.cc
```diff
@@ -1725,6 +1725,93 @@ namespace aspect
...
 
+  template <int dim>
+  void Simulator<dim>::interpolate_material_output_into_field ()
+  {
+    // we need some temporary vectors to store our updates to composition in
+    // before we copy them over to the solution vector in the end
+    LinearAlgebra::BlockVector distributed_vector (introspection.index_sets.system_partitioning,
+                                                   mpi_communicator);
+
+    pcout << "   Copying material properties into prescribed compositional fields."
```
gassmoeller: I find the `material properties` a bit confusing, this sounds to the user as if you would copy things like density or viscosity. What about `Copying properties into prescribed compositional fields`? Then the user would need to know themselves what is copied.

jdannberg:
Okay, I think I have addressed your comments.Can you have a look, and also start the tester again?

source/simulator/solver_schemes.cc
```diff
@@ -214,6 +215,14 @@ namespace aspect
           }
       }
 
+    // Update the copy fields *after* we have solved all the other fields
+    for (unsigned int c=0; c < introspection.n_compositional_fields; ++c)
+      {
+        const AdvectionField adv_field (AdvectionField::composition(c));
+        if (adv_field.advection_method(introspection) == Parameters<dim>::AdvectionFieldMethod::prescribed_field)
+          interpolate_material_output_into_field();
```
tjhei: this looks wrong. I think you should call this method only once even if you have several prescribed fields.

source/simulator/parameters.cc
```diff
@@ -1004,6 +1004,12 @@ namespace aspect
                          "advected with the melt velocity instead of the solid velocity. "
                          "This method can only be chosen if melt transport is active in the "
                          "model."
+                         "\n"
+                         "\\item ``prescribed field'': If a compositional field is marked "
```
tjhei: I am not a big fan of "prescribed field" as this remind me of "prescribed Stokes solution". Not sure if you can think of a better name? If not, we can leave it as is.

source/simulator/parameters.cc
```diff
@@ -1004,6 +1004,12 @@ namespace aspect
...
+                         "with this method, then the value of a specific additional material "
```
tjhei: can you add at the beginning that the value of these fields is determined in each timestep from the material model?

include/aspect/material_model/interface.h
```diff
@@ -797,6 +797,35 @@ namespace aspect
...
+        PrescribedFieldOutputs (const unsigned int n_points,
+                                const unsigned int n_comp);
+
+        virtual std::vector<double> get_nth_output(const unsigned int idx) const;
+
+        /**
+         * Prescribed field outputs for all compositional fields at the evaluation points
+         * that are passed to the instance of MaterialModel::Interface::evaluate()
+         * that fills the current object.
+         * copy_properties[q][c] is the prescribed field output at the evaluation point q
```
tjhei: update name

source/simulator/helper_functions.cc
```diff
@@ -1725,6 +1725,93 @@ namespace aspect
...
+              if (parameters.compositional_field_methods[c] == Parameters<dim>::AdvectionFieldMethod::prescribed_field)
+                {
+                  const unsigned int composition_idx
+                    = dof_handler.get_fe().component_to_system_index(introspection.component_indices.compositional_fields[c],
+                                                                     /*dof index within component=*/ j);
+
+                  // skip entries that are not locally owned:
+                  if (dof_handler.locally_owned_dofs().is_element(local_dof_indices[composition_idx]))
+                    {
+                      distributed_vector(local_dof_indices[composition_idx]) = prescribed_field_out->prescribed_field_outputs[j][c];;
```
tjhei: remove ";'

include/aspect/material_model/interface.h
```diff
@@ -797,6 +797,35 @@ namespace aspect
...
+         * for the compositional field with the index c.
```
tjhei: Do we really want n entries here even if we have less than n prescribed fields? This might be confusing because you have to make sure to fill the correct entry. Otherwise your values will be silently ignored.

source/simulator/helper_functions.cc
```diff
@@ -1725,6 +1725,93 @@ namespace aspect
...
+                           "(it does not create PrescribedFieldOutputs, which are required for this "
+                           "advection field type)."));
+
+    // Make a loop first over all cells, and then over all degrees of freedom in each element
+    // to interpolate material properties onto a solution vector.
+
+    // Note that the values for some degrees of freedom are set more than once in the loop
+    // below where we assign the new values to distributed_vector (if they are located on the
+    // interface between cells), as we loop over all cells, and then over all degrees of freedom
+    // on each cell. But even though we touch some DoF twice, we always compute the same value,
```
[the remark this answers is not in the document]
jdannberg: This was just copy and paste, I am changing it in the original place too.

source/simulator/solver_schemes.cc
jdannberg: Yes, you're right.

source/simulator/parameters.cc
jdannberg: But it is something very similar to "prescribed Stokes solution". Instead of solving an equation, we use an output of the material model to fix the compositional field to a prescribed value that can also change over time (exactly as in the prescribed Stokes solution). I think that's how we came up with the name. 
The only difference is that for the prescribed Stokes solution, you use a new plugin to prescribe the solution, while here, you use a material model output. But the output doesn't even have to depend on any material properties, you could also just use any function that changes over time, so the functionality is exactly the same.

include/aspect/material_model/interface.h
jdannberg: I thought using as many entries as we have compositional fields makes sense because then it is easy to assign values for the correct field in the material model by using `this->introspection().compositional_index_for_name("field_name")`. If we choose a smaller number, we would have to do some conversion between the indices somewhere. 

And if you happen to put your values into the wrong field (and not fill the field you wanted), it will try and copy quietNaNs into the prescribed field (which is what the PrescribedFieldOutputs are initialized to), so the user should notice something is wrong. 

The only problem you get is if you fill the outputs in the material model, but then do not actually use prescribed fields. But that is the same for other plugins and additional outputs, you can fill reaction rates, but if you do not use operator splitting, they will just be ignored. 
And I do not see a good way to check if all fields that are not prescribed fields still have their original quietNaN values in the PrescribedFieldOutputs.

jdannberg:
Okay, I think I addressed all of your comments, @tjhei. Can you have another look?

include/aspect/material_model/interface.h
tjhei: > And if you happen to put your values into the wrong field (and not fill the field you wanted), it will try and copy quietNaNs into the prescribed field

They are quiet, so no error will happen, right? How do you notice? Can we add a ``is_finite(vector.max())`` assert or something?

Question: can you visualize this additional output with the postprocessor or does this break because of the NaNs?

tjhei: oh, and can you leave a note in this class (or here) that this is _all_ compositional fields not only the prescribed ones?

tjhei:
> Can you have another look?

This looks good, but take a look at the questions above.

include/aspect/material_model/interface.h
jdannberg: > Question: can you visualize this additional output with the postprocessor or does this break because of the NaNs?

At the moment, if you don't fill the outputs, it will copy NaNs into your prescribed compositional field, so even without the postprocessor, if you ever look at the output, you will notice it's all NaN. But you can also use the postprocessor to look at the prescribed field outputs, and they will also all be NaNs. So it should be very obvious that something is wrong. 

I am adding an Assert (so it's only triggered in debug mode) that will check the entries that are used are finite, and in release mode, it will just copy NaNs into the fields as described above.

jdannberg:
> This looks good, but take a look at the questions above.

I think I've addressed your questions.

jdannberg:
Yay :smiley: 
Thank you for reviewing!

PR #2678, opened 2018-10-02
Implement maximum first timestep parameter.

MFraters:
I noticed that in models with a free surface and open boundaries all around, it can help a lot if the first timestep can be capped manually. I thought it might also be useful to others.

### DIFF
--- a/source/simulator/parameters.cc
+++ b/source/simulator/parameters.cc
@@ -155,7 +155,7 @@ namespace aspect
                        "``Maximum relative increase in time step''."
                        "The default value is a value so that when converted from years into seconds "
                        "it equals the largest number representable by a floating "
-                       "point number, implying an unlimited time step."
+                       "point number, implying an unlimited time step. "
                        "Units: Years or seconds, depending on the ``Use years "
                        "in output instead of seconds'' parameter.");
 
--- a/source/simulator/parameters.cc
+++ b/source/simulator/parameters.cc
@@ -143,6 +143,22 @@ namespace aspect
                        "Units: Years or seconds, depending on the ``Use years "
                        "in output instead of seconds'' parameter.");
 
+    prm.declare_entry ("Maximum first time step",
+                       /* boost::lexical_cast<std::string>(std::numeric_limits<double>::max() /
+                                                           year_in_seconds) = */ "5.69e+300",
+                       Patterns::Double (0),
+                       "Set a maximum time step size for only the first timestep. Generally the time step "
+                       "based on the CFL number should be sufficient, but for complicated models "
+                       "or benchmarking it may be useful to limit the first time step to some value, "
+                       "especially when using the free surface, which needs to settle to prevent "
+                       "instabilities. This should in that case be combined with a value set for "
+                       "``Maximum relative increase in time step''."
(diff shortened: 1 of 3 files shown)


tjhei:
1. Couldn't you implement this as a ``TerminationCriteria`` plugin? The documentation says that 
```
A reduced time step size may be returned for the last time
         * step. For all other time steps, the current time step size
         * (provided as argument) will be returned.
```
but I don't see a reason why that should be the case.

2. I agree with @naliboff that you could implement a FunctionParser that returns the desired time step based on ``timestep_no, time_step_size, t``. So your functionality could be done as
```
set expression = if(timestep_n==0, 1e-5, time_step_size)
```
or something.

source/simulator/parameters.cc
```diff
@@ -143,6 +143,22 @@ namespace aspect
...
+    prm.declare_entry ("Maximum first time step",
+                       /* boost::lexical_cast<std::string>(std::numeric_limits<double>::max() /
+                                                           year_in_seconds) = */ "5.69e+300",
+                       Patterns::Double (0),
+                       "Set a maximum time step size for only the first timestep. Generally the time step "
+                       "based on the CFL number should be sufficient, but for complicated models "
+                       "or benchmarking it may be useful to limit the first time step to some value, "
+                       "especially when using the free surface, which needs to settle to prevent "
+                       "instabilities. This should in that case be combined with a value set for "
+                       "``Maximum relative increase in time step''."
```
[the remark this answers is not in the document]
gassmoeller: John this looks correct to me (compare to the lines below ` ``Use years in output instead of seconds'' `).

naliboff:
@gassmoeller - Yes, sounds good to me.

source/simulator/parameters.cc
```diff
@@ -143,6 +143,22 @@ namespace aspect
...
+                       "The default value is a value so that when converted from years into seconds "
+                       "it equals the largest number representable by a floating "
+                       "point number, implying an unlimited time step."
```
tjhei:
```suggestion
                       "point number, implying an unlimited time step. "
```

source/simulator/parameters.cc
tjhei:
```suggestion
                       "``Maximum relative increase in time step''. "
```

tjhei:
> I would not want to ask him for doing a complete rework of the timestep computation architecture

I can see that but I would think it would take a few minutes to check if this can be done inside TerminationCriteria. I am not opposed to merging as is.

gassmoeller:
I just checked. You can implement a termination plugin that never terminates, but uses a function parser to return reduced time steps if necessary. It would be a bit weird though, because the term termination criteria is then misleading :-).

MFraters:
Sorry for the late response. I like the function parser idea of @tjhei a lot, but I don't think I will have the time in the next few months to implement it. Although it might be possible to do something similar through the TerminationCriteria, it feels illogical and counter intuitive to me to put it there. 

So I am fine with either merging this approach, or the function parser approach.

gassmoeller:
@tjhei: Would you be ok if I merge this for now?

PR #2679, opened 2018-10-04
Fix multiple additional named outputs postprocessor

gassmoeller:
As described in the changelog. This was only a problem if a material model creates multiple named additional outputs and only if this postprocessor is active.

### DIFF
--- a/doc/modules/changes/20181004_gassmoeller
+++ b/doc/modules/changes/20181004_gassmoeller
@@ -1,0 +1,5 @@
+Fixed: The 'named additional outputs' postprocessor used to only support
+a single named output from the material model due to an indexing bug. This
+is fixed now.
+<br>
+(Rene Gassmoeller, 2018/10/04)
--- a/source/postprocess/visualization/named_additional_outputs.cc
+++ b/source/postprocess/visualization/named_additional_outputs.cc
@@ -128,6 +128,7 @@ namespace aspect
         this->get_material_model().create_additional_named_outputs(out);
         this->get_material_model().evaluate(in, out);
 
+        unsigned int field_index = 0;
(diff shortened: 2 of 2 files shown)


gassmoeller:
Thanks! The testers are happy (though the cig tester did not report its result, it has finished successfully if you click on 'Details').

bangerth:
@jdannberg -- do we have a test that checks this?

jdannberg:
I don't think so. Do you volunteer to make one?

bangerth:
I don't actually know how to do that. Would one need a material model that creates multiple additional outputs? How did you find this bug?

jdannberg:
I don't remember how it came up, I thought @gassmoeller found it (after all, it's his pull request). So @gassmoeller, do you still have the setup that caused the problem?

bangerth:
Oh, sorry. I tagged you by accident then -- I keep getting confused by the text at the top of merged pull requests that says "jdannberg merged 2 commits into geodynamics:master" which does not imply that you actually wrote the PR.

So, @gassmoeller, do you recall?

gassmoeller:
Yes, the problem appears if a material model creates more than one additional named output object (this is the index `k` in the changed code). Suppose there is a material model that creates 2 named additional output objects with 1 entry each, then you would expect to output them as:
output field 1 = named additional output object 1 - field 1
output field 2 = named additional output object 2 - field 1

but with the old indexing you would get
output field 1 = named additional output object 1 - field 1
output field 1 = named additional output object 2 - field 1
and output field 2 would not be filled.

There is no material model that does this as the moment (that is why we did not notice it when this was introduced), but I found this with an plugin material model. I created a test for it in #2730.

PR #2683, opened 2018-10-11
Remove link to legal notice of deal.ii on website

gassmoeller:
I think this was appropriate when our website was still hosted at dealii.org, but now it is outdated (and confusing).

### DIFF
--- a/index.html
+++ b/index.html
@@ -236,19 +236,6 @@
 	    </p>
 
 	  </div>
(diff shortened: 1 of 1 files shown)


bangerth:
I think it was always confusing and legally probably no longer necessary for many years already. I removed a similar snippet from the deal.II website a while back.

PR #2684, opened 2018-10-11
Fix typo in website title.

gassmoeller:
This must have been around since the birth of ASPECT.


index.html
```diff
@@ -41,7 +41,7 @@ <h2>
 	    <td width="10%"><b>What it is:</b></td>
 	    <td>
 	      An extensible code written in C++ to support research
-	      in simulating convection in the Earth mantle and elsewhere. 
+	      in simulating convection in Earth's mantle and elsewhere.
```
jdannberg: Shouldn't it be `in the Earth's mantle`?

bangerth:
I think both is correct. It's a question how specific you want to be and what the focus is on. If you say "the Earth mantle", then the emphasis is on the mantle. If you say "the Earth's mantle", then the emphasis is on Earth.

PR #2688, opened 2018-10-18
Replace aspect-devel references

naliboff:
@gassmoeller - If we are able to import the aspect-devel archives into discus, then we can remove the aspect-devel references or just update that aspect-devel discussions are archived on the forum.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,9 +14,7 @@ guide on the process of contributing to an open-source project
 [here](https://opensource.guide/how-to-contribute/).
 
 ## Asking and answering questions about ASPECT
-The ASPECT community maintains an active mailing list hosted by CIG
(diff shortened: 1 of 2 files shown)


naliboff:
The new github feature that @gassmoeller used is perfect for minor suggestions to something like documentation. The only downside is that each suggested changes is saved as a different commit. Perhaps github will allow merging multiple suggestions into a single commit at some point.

bangerth:
@naliboff -- you always have the opportunity to merge @gassmoeller 's suggested changes, locally do a `git pull`, then squash/fix-up things, and push the merged whole thing. That avoids having lots and lots of very small commits that just implement suggested wording changes.

naliboff:
@bangerth - yep, that is straightforward enough :+1:

PR #2689, opened 2018-10-23
Balto plugin

kneumiller:


bangerth:
@kneumiller -- you closed this pull request again but I looked through it anyway. I would prefer if one did not need to specify the `read_from_url` flag in the input file, but if the `load_file()` function just figures this out automatically by looking at the given file name: if the file name starts with `http://` or something else that makes sense, then the file should be opened via whatever process you have for that, otherwise it's assumed to be the name of a local file.

I prefer this because ASPECT may be reading a substantial number of files during startup, depending on what kinds of plugins one uses. The way you have this right now, using a *global* flag `read_from_url`, implies that either all or none of them are read via the internet. But that's unlikely to be a useful approach: in many cases, one may want to use some input files from the local disk and maybe others via the internet. To make this work, one would have to have a flag `read_from_url` for *each plugin that can read data from somewhere*. That seems excessive. The easier approach is that if someone provides a URL with a transport protocal prefix (such as `http://` or `https://` or `webdav://`, etc), then the `load_file()` function should simply assume that the file is not a local file name. This also makes it more flexible if one eventually wanted to implement *different* ways of accessing web-served files without making `read_from_url` ternary or even higher-order switches everywhere.

kneumiller:
@bangerth -- I took a look at the solution you suggested and I noticed that there are some functions that call `read_and_distribute_file_content()` directly rather than through `load_file()` (such as initial_temperature/adiabatic_boundary.cc). I could just be misinterpreting the way the code interacts, and if I am please let me know, but I am wondering if the check for `http://` should be inside `read_and_distribute_file_content()` instead since it is called directly by some functions and `load_file()` calls it at the beginning of its function. 

I could also look at moving the global parameter `read_from_url` to be a parameter of the parent class (AsciiDataBase/AsciiDataBoundary) so that the default value is false unless someone wishes to use a url for one of their subsections in which they can set the `read_from_url` parameter to be true for just that subsection.

PR #2690, opened 2018-10-24
Instantiate additional dislocation viscosities output object

gassmoeller:
The `DislocationViscosityOutputs` are only implicitly used in the grain size material model, but not explicitly instantiated. This can lead to linker errors if they are used in an external plugin (very rare, but happened to me with clang 6.0).


bangerth:
Ha, interesting failure mode!

PR #2691, opened 2018-10-24
[bugfix] fix restarting melt models

jdannberg:
I just noticed that this was broken and that models with melt transport couldn't be restarted any more. I think we broke it in #2419, where we moved the line 
`melt_handler->add_current_constraints (new_current_constraints);`
into `compute_current_constraints()`, where we then close the constraints, so we can't and shouldn't let the melt handler add its constraints again outside of compute_current_constraints() when we restart the computation. 

I added a test to make sure we don't break this in the future.


gassmoeller:
Looks good. :+1:

PR #2693, opened 2018-10-27
add diffusion of prescribed fields

jdannberg:
This pull request adds an advection method that allows it to first copy a material model output into a compositional field, and then to diffuse it using a fixed length scale. In order to make this work, I had to change the way the prescribed_field method works: It will not be called after all of the other fields are solved, but instead will be called in the order the fields are listed in (so if the input file lists the prescribed_field last, it would still be solved last). As I just added this functionality two weeks ago, I don't think anyone has started using it yet. 
If we agree that this makes sense, I'll also add a changelog entry explaining this.

### DIFF
--- a/doc/modules/changes/20181029_jdannberg
+++ b/doc/modules/changes/20181029_jdannberg
@@ -5,10 +5,5 @@ given as an input parameter. This option can be used by selecting the
 'prescribed field with diffusion' compositional field method in the 
 input file, and allows it to smooth quantities that are computed in 
 the material model.
-Changed: If the 'prescribed field' compositional field method is used 
-for a field, this field will now be solved in the order it is listed
-in in the input file (as is done for all compositional fields) instead 
-of being updated after all of the other fields are solved (the 
-previous behaviour). 
 <br>
 (Juliane Dannberg, 2018/10/29)
--- a/doc/modules/changes/20181029_jdannberg
+++ b/doc/modules/changes/20181029_jdannberg
@@ -1,0 +1,14 @@
+New: There is now a new method for compositional fields that allows
+to prescribe them to a value that is computed in the material model 
+and then, in a second step, to diffuse them based on a length scale 
+given as an input parameter. This option can be used by selecting the 
+'prescribed field with diffusion' compositional field method in the 
+input file, and allows it to smooth quantities that are computed in 
+the material model.
+Changed: If the 'prescribed field' compositional field method is used 
+for a field, this field will now be solved in the order it is listed
+in in the input file (as is done for all compositional fields) instead 
+of being updated after all of the other fields are solved (the 
+previous behaviour). 
+<br>
+(Juliane Dannberg, 2018/10/29)
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1149,10 +1149,12 @@ namespace aspect
 
 
       /**
-       * Interpolate material model outputs onto compositional fields in case the
-       * 'prescribed field' compositional field method is used. For every field that
-       * uses this method, this function interpolates additional material model outputs
-       * called 'PrescribedFieldOutputs' and copies these values into the solution vector.
+       * Interpolate material model outputs onto a compositional field. For the field
+       * whose index is given in the input argument compositional_index, this function
+       * interpolates additional material model outputs called 'PrescribedFieldOutputs'
+       * and copies these values into the solution vector.
+       * This is useful for fields that use the 'prescribed field' compositional field
+       * method.
        *
        * This function also updates the old solution vectors with these interpolated
        * values so that this compositional field method can be combined with time-dependent
@@ -1161,7 +1163,7 @@ namespace aspect
        * This function is implemented in
        * <code>source/simulator/helper_functions.cc</code>.
        */
-      void interpolate_material_output_into_fields ();
+      void interpolate_material_output_into_fields (const unsigned int compositional_index);
 
 
       /**
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1150,7 +1150,7 @@ namespace aspect
 
       /**
        * Interpolate material model outputs onto a compositional field. For the field
-       * whose index is given in the input argument compositional_index, this function
+       * whose index is given in the @p compositional_index, this function
        * interpolates additional material model outputs called 'PrescribedFieldOutputs'
        * and copies these values into the solution vector.
        * This is useful for fields that use the 'prescribed field' compositional field
@@ -1163,7 +1163,7 @@ namespace aspect
        * This function is implemented in
        * <code>source/simulator/helper_functions.cc</code>.
        */
-      void interpolate_material_output_into_fields (const unsigned int compositional_index);
+      void interpolate_material_output_into_field (const unsigned int compositional_index);
 
 
       /**
--- a/source/simulator/assemblers/advection.cc
+++ b/source/simulator/assemblers/advection.cc
@@ -77,6 +77,10 @@ namespace aspect
 
       const FEValuesExtractors::Scalar solution_field = advection_field.scalar_extractor(introspection);
 
+      if (!advection_field_is_temperature && advection_field.advection_method (introspection)
+          == Parameters<dim>::AdvectionFieldMethod::prescribed_field_with_diffusion)
+        return;
+
       for (unsigned int q=0; q<n_q_points; ++q)
         {
           // precompute the values of shape functions and their gradients.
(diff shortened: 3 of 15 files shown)


include/aspect/simulator.h
```diff
@@ -1149,10 +1149,12 @@ namespace aspect
 
 
       /**
-       * Interpolate material model outputs onto compositional fields in case the
-       * 'prescribed field' compositional field method is used. For every field that
-       * uses this method, this function interpolates additional material model outputs
-       * called 'PrescribedFieldOutputs' and copies these values into the solution vector.
+       * Interpolate material model outputs onto a compositional field. For the field
+       * whose index is given in the input argument compositional_index, this function
```
tjhei: ``given in @p compositional_index``

include/aspect/simulator.h
```diff
@@ -1161,7 +1163,7 @@ namespace aspect
        * This function is implemented in
        * <code>source/simulator/helper_functions.cc</code>.
        */
-      void interpolate_material_output_into_fields ();
+      void interpolate_material_output_into_fields (const unsigned int compositional_index);
```
tjhei: maybe rename to ``..._field(``?

source/simulator/assemblers/advection.cc
```diff
@@ -255,6 +259,84 @@ namespace aspect
...
+            {
+              data.local_rhs(i)
+              += scratch.old_field_values[q] * scratch.phi_field[i]
+                 *
+                 JxW;
+
+              for (unsigned int j=0; j<advection_dofs_per_cell; ++j)
+                {
+                  data.local_matrix(i,j)
+                  += (std::pow(parameters.diffusion_length_scale, 2) * (scratch.grad_phi_field[i] * scratch.grad_phi_field[j])
```
tjhei: 2.0

tests/prescribed_field_with_diffusion/screen-output
```diff
@@ -0,0 +1,31 @@
...
+-----------------------------------------------------------------------------
+-----------------------------------------------------------------------------
+Number of active cells: 400 (on 2 levels)
+Number of degrees of freedom: 10,628 (4,010+603+2,005+2,005+2,005)
+
+*** Timestep 0:  t=0 seconds
+   Solving temperature system... 0 iterations.
+   Skipping normal_field composition solve because RHS is zero.
+   Copying properties into prescribed compositional field diffused_field.
+   Solving diffused_field system ... 159 iterations.
```
tjhei: why is this taking so many iterations?

jdannberg:
To reply to your questions:

> why do you need so many linear iterations? That looks weird.

I have no idea... I don't know enough about the solver to really say anything about that, what should I look out for? Do you have any ideas?

> what happens if you use a nonlinear loop, will it add diffusion in each nonlinear iteration or does it behave correctly?

It copies a material property into the field before solving the diffusion equation, so the diffusion should not add up, and instead should be recomputed in each iteration (unless you write a material model that copies the prescribed field with diffusion itself into the field). Do we need another test case with a nonlinear loop?

> You never describe the equation you are solving, can we put this somewhere?

I added it in the description of the compositional field method. 

> Why does this not depend on the timestep size? Wouldn't this mean that you apply more or less smoothing if you change your timestep size?

No, this means that you apply exactly the same amount of diffusion independent of the time step. The field is not advected (in other words, it has no history), it is just a material model output that is diffused. So the desired behaviour is smoothing it for a fixed amount that does not depend on the time step size. The idea is that it can be used to smooth out for example the strain rate before it is used to compute the viscosity in the Stokes equation, which might help with convergence of the linear and nonlinear solvers.

tjhei:
> Do you have any ideas?

Is it building/updating the preconditioner for this system? What are your boundary conditions? What is the effective coefficient for the diffusion in this example?

doc/modules/changes/20181029_jdannberg
```diff
@@ -0,0 +1,14 @@
+New: There is now a new method for compositional fields that allows
+to prescribe them to a value that is computed in the material model 
+and then, in a second step, to diffuse them based on a length scale 
+given as an input parameter. This option can be used by selecting the 
+'prescribed field with diffusion' compositional field method in the 
+input file, and allows it to smooth quantities that are computed in 
+the material model.
+Changed: If the 'prescribed field' compositional field method is used
```
tjhei: do we need two separate changelog files for this? This will look weird.

jdannberg:
Okay, I tested a number of things together with @gassmoeller, and we came to the following conclusions:

> Is it building/updating the preconditioner for this system? 

It is building the preconditioner, and we use exactly the same preconditioner as for the advection equation (so it is using the matrix block). 

> What are your boundary conditions? 

I tried both prescribed and reflective boundaries, and it does not change the iteration count. 

> What is the effective coefficient for the diffusion in this example?

I think this might be the problem. In my example, the diffusion length scale is around 0.2, while the cell size is only 0.05. So we are diffusing over much bigger distances than the cell size. I have run a number of tests; increasing the diffusion length scale increases the number of iterations, decreasing the cell size increases the number of iterations, but if I keep the ratio of diffusion length scale and cell size constant, the iteration count stays the same. So I assume we need this many iterations because we apply more diffusion than one would normally do in one time step. 

However, I think this shouldn't be too much of a problem for the applications this is intended for. The idea is that one may want to smooth properties like the strain or the strain rate over a distance of a few cells, so that the length scale of shear bands is determined by the diffusion length scale instead of the mesh size. So usually, the diffusion length scale should not be bigger (relative to the cell size) than what I have chosen here.

doc/modules/changes/20181029_jdannberg
jdannberg: fixed

tjhei:
> I think this might be the problem. In my example, the diffusion length scale is around 0.2, while the cell size is only 0.05. So we are diffusing over much bigger distances than the cell size.

Ohhh. We are using ILU. If you have a large coefficient we are basically solving a Laplace equation. Here it would be much better to use AMG instead of ILU. ILU is only good for mass-matrix like things.

tjhei:
Anyways, I think we can merge this as is and worry about solvers some other time. 

> I tried both prescribed and reflective boundaries, and it does not change the iteration count.

What is it set to right now?

jdannberg:
> What is it set to right now?

The test has one boundary where the composition is fixed, and three boundaries that are reflective (the reason is that it's basically a 1-D setup of a Gaussian that is diffused, so I wanted to have a reflective boundary at the center of the Gaussian, but I prescribed the boundary composition on the opposite end of the model).

tjhei:
oh, so the normal boundary conditions work correctly, great!

bangerth:
I know I'm 45 days late with this, but I'd like to come back to a thing @tjhei had previously asked:

> Why does this not depend on the timestep size? Wouldn't this mean that you apply more or less smoothing if you change your timestep size?

This is still the case, see https://github.com/geodynamics/aspect/blob/master/source/simulator/assemblers/advection.cc#L318
What this implies is that you will get different solutions depending on the time step you choose, and of course also a different solution if you refine the mesh once (because then the time step will be different). In other words, this scheme does not converge to the solution of any equation. In other words, there *is* no equation.

@jdannberg -- is this really what you want? The description in the documentation of the length scale parameter is consistent with the implementation, but I'm questioning whether this is a useful operation to do. Why would one want to compute a field smoothed in such a way?

jdannberg:
Yes, this is exactly what I want. 

The prescribed fields are not advected fields, so for them, it doesn't make any difference what the time step size is. They are just copied from a material model output, which could, for example, be the stress or the strain rate. After they are copied, a certain amount of diffusion is applied, and it should not depend on the time step size, because we want to apply the same amount of diffusion no matter what the time step size is. (Or a different way you could think about it is that for solving the diffusion equation, we just do not use ASPECT's normal time step, we choose a different time step that does not have anything to do with real time, because what we want to do is smooth the field by a certain amount, not diffuse it in real time.) 

I will not get a different solution if the time step size is different or the mesh is refined, because I don't advect the field, so the diffusion does not add up. The field is copied over from the material model outputs again in each time step, and so I want the exact same amount of diffusion independent of the time step size, exactly because I want the field smoothed in the same way every time. 

So if we come back to the example of the strain rate, the idea is to make the solution independent of the mesh resolution by smoothing the strain rate that is used to compute the viscosity, because then the amount of diffusion determines the smallest length scale that faults or other strain-rate dependent features can have rather than the cell size (this is a common problem for models with plastic rheology: fault zones will always be exactly one cell size wide, because there is nothing to keep them from localizing).

PR #2694, opened 2018-10-30
Fix benchmark plugin crash and modify test to actually use it

gassmoeller:
The benchmark plugin for the solcx particle benchmark wasnt compiling (no functional bug, just a mismatch of components). We did not notice, because the test designed to test it did not actually use this plugin. I fixed the bug, and modified the test to use the plugin (and in general behave more like the real solcx benchmark).


tjhei:
looks good, but the CIG tester is dead. Merge without?

gassmoeller:
Yeah, merge without, we are still having the DNS hickups, and if they appear right when a PR is opened, the tester does not see the PR.

PR #2695, opened 2018-10-31
jenkins: set reference repo

tjhei:
this speed ups cloning of the repo by using a local copy of the
repository as a reference

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,10 @@ pipeline {
     docker {
         image 'dealii/dealii:v8.5.1-gcc-mpi-fulldepscandi-debugrelease'
 	label 'has-docker'
+	// We mount /repos into the docker image. This allows us to cache
+	// the git repo by setting "advanced clone behaviors". If the
+	// directory does not exist, this will be ignored.
+	args '-v /repos:/repos:ro'
     }
   }
 


gassmoeller:
Looks good to me. How much does it speed up the tester? Or is this mostly useful on slow internet connections?

tjhei:
> How much does it speed up the tester?

Checkout takes 1-2 minutes without and a few seconds with this change. It also saves quite a bit of bandwidth if you think about it. Not a big deal, but it makes a larger difference for deal.II compared to ASPECT.

> Or is this mostly useful on slow internet connections?

I hope that this will also cause fewer failing builds due to failing clones.

bangerth:
I'm going through old PRs and this one caught my eyes. If I see this right, then you are mounting a directory external to the container from inside the container, right? This is awkward because now you are requiring a certain setup from everyone who wants to run the container. This makes it more difficult for others to deploy Jenkins testers if that's what they would like to do.

Or do I see this wrongly?

gassmoeller:
In a way you are right. This argument is used by Jenkins when it starts the docker container, i.e. it is an argument to the `docker run` command on the host. It is true that this requires a certain directory structure for the jenkins tester though.

bangerth:
Can we document this somewhere? Or adapt the scripts in such a way that if the directory is not found/mounted, that they check out the repo from scratch? 

We could also restrict the clone to a subset of all commits. We don't actually need the entire ASPECT history in the clone.

PR #2700, opened 2018-11-02
move material utility functions to a new file

jdannberg:
As discussed in #2696, I am moving the utility functions out of the material model interface and into their own file and namespace. I'd like to keep the name as short as possible, as the namespace is used in many places, so I went with `MaterialUtilities`. 

This pull request also moves the computation of the Drucker-Prager viscosity into a utility function. To do that, I defined separate Inputs and Outputs to be used with this function. If this model seems reasonable, similar things could be done with the diffusion/disclocation viscosity computation, phase transitions, etc. 

Note that in the utilities files, the `compute_volume_fractions`, `CompositionalAveragingOperation`, and `average_value` functions are just moved from the interface, while the `parse_compositional_averaging_operation`, `DruckerPragerInputs/Outputs` and `compute_drucker_prager_yielding` function are new.


naliboff:
@jdannberg - Let me know when this is ready for review.

jdannberg:
@naliboff This is ready for review. If there are still tests failing, it's only because they're not finding the compute_volume_fractions function because I moved it to a different namespace, so that should be easy to fix.

include/aspect/material_model/utilities.h
```diff
@@ -0,0 +1,179 @@
...
+#include <deal.II/base/parameter_handler.h>
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    /**
+     * An namespace in which we define utility functions that
```
naliboff:
```suggestion
     * A namespace in which we define utility functions that
```

source/material_model/visco_plastic.cc
```diff
@@ -198,63 +193,44 @@ namespace aspect
               if (use_viscous_strain_weakening == true)
                 strain_ii = composition[this->introspection().compositional_index_for_name("viscous_strain")];
 
-              viscous_weakening = calculate_viscous_weakening(strain_ii, j);
+              // Apply strain weakening of the viscous viscosity
+              viscosity_pre_yield *= calculate_viscous_weakening(strain_ii, j);
```
naliboff: Is this term is going to be moved into the "if" statement above in a later a pull request? That's what I recall, but wanted to double check.

jdannberg: Yes. For this pull request I didn't want to change any functionality, because I change lots of lines, so I want to make sure that none of the tests are changed and everything still works. Once we merge this, I have a follow-up pull request ready that actually fixes this.

include/aspect/material_model/utilities.h
jdannberg: Copy-and-paste error :grin:
I fixed it in the original place too.

gassmoeller:
Looks good to me, mostly moved code. User plugins will now need to use the `MaterialUtilities` namespace if they want to use these functions, but the functions were only added recently anyway.

PR #2701, opened 2018-11-05
Make installation FAQ more visible.

gassmoeller:
The current link to the installation FAQ is not really visible inside the instructions if someone gets stuck at a particular point (because it is in the introductory text of the section). Put another reference to the point where it is most likely needed (when running candi). This was suggested by Björn Heyn.
Also update the reference to the mailing list by the forum address.


jdannberg:
What's going on with the testers?
I'll merge this, as the changes are only in manual.tex, but it would be good to have at least one tester work reliably.

PR #2704, opened 2018-11-13
add two merged boxes to initial lithostatic pressure plugin.

MFraters:
The two merged boxes should also not convert spherical to cartesian coordinates.


gassmoeller:
Our testers seem to have missed this PR, but I tested it locally and it passed all tests. I will take the blame if something goes wrong :smile:.

PR #2705, opened 2018-11-14
fix strain weakening logic

jdannberg:
When looking at the strain weakening code with @naliboff and @gassmoeller we found that there was a specific combination of strain weakening options one could choose that would lead to the wrong strain weakening computation under these special circumstances. This pull request (hopefully) fixes this, and also makes the logic more intuitive by just using one parameter that defines the strain weakening mechanism. For backwards compatibility, the old options are still allowed (which meant I had to add some complicated logic, but this code can be removed after the next release or so). 

@naliboff, can you have a look if I translated from the old to the new options correctly, if all of the new options make sense (there might be combinations that one shouldn't use, in which case I would just not allow them), and if the new options are all the ones one would use or if there are options missing. 

If we agree this makes sense and all of the tests still work, I will update all parameter files to use the new options (and add a changelog entry).


source/material_model/visco_plastic.cc
```diff
@@ -158,50 +158,73 @@ namespace aspect
...
-              // Apply strain weakening of the viscous viscosity
-              viscosity_pre_yield *= calculate_viscous_weakening(strain_ii, j);
             }
 
 
           // Third step: plastic yielding
           // Calculate Drucker-Prager yield strength (i.e. yield stress)
-          // Use max_yield_strength to limit the yield strength for depths beneath the lithosphere
-          const MaterialUtilities::DruckerPragerInputs plastic_in(coh, phi, std::max(pressure,0.0), edot_ii, max_yield_strength);
+          const MaterialUtilities::DruckerPragerInputs plastic_in(yield_parameters.first, yield_parameters.second, std::max(pressure,0.0), edot_ii);
```
naliboff: Should `max_yield_strength` be the last entry for `DruckerPragerInputs`?

naliboff:
```suggestion
          const MaterialUtilities::DruckerPragerInputs plastic_in(yield_parameters.first, yield_parameters.second, std::max(pressure,0.0), edot_ii, max_yield_strength);
```

source/material_model/visco_plastic.cc
```diff
@@ -715,29 +738,91 @@ namespace aspect
...
+                             "\\item ``viscous weakening with viscous strain only'': The second invariant of "
+                             "the strain accumulated by viscous deformation is tracked, and is used to "
+                             "weaken the viscous viscosity. The plastic viscosity is not weakened."
+                             "\n\n"
+                             "\\item ``default'': The default option has the same behavior as ``none'', "
+                             "but is there to make sure that the original parameters for specifying the "
+                             "strain weakening mechanism (``Use plastic/viscous strain weakening'') are still allowed, "
+                             "but to guarantee that one uses either the old parameter names or the new ones, "
+                             "never both.");
+          prm.declare_entry ("Use strain weakening", "default",
```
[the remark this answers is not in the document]
naliboff: Given the amount of code required to keep backward compatibility, I'm inclined towards outright removal. However, that can always be done in a follow-up PR as it may be quite a bit of work.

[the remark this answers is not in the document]
jdannberg: @naliboff, what do you think how many people use this at the moment? The feature was introduced before the last release (the changelog says 2017/04/14) and I am not sure we want to break people's prm files again without even marking the option as deprecated...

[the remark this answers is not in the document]
naliboff: On second thought, I agree. A fair number of people are likely using the feature by now, so probably best to not break functionality until we get close to the next major release.

naliboff:
It looks like some of the strain-weakening tests are changing, so perhaps some of the related logic has changed? I'll take another look and see if I can spot anything.

jdannberg:
Yes, that would be good. I definitely changed some of the logic (as we agreed that it was not correct for some specific cases) but mostly I was not quite sure what the different combinations of Use viscous/plastic strain weakening were supposed to do, so it is very well possible that the logic is not yet doing what it should. If you think this would be easier, we can also sit together and discuss this at some point.

bangerth:
If you leave removal of backward compatible code for later, can you open a github issue to remind us of it, and tag it with the 2.2 release?

source/material_model/visco_plastic.cc
```diff
@@ -158,50 +158,73 @@ namespace aspect
               }
             }
 
-          double phi = angles_internal_friction[j];
-          double coh = cohesions[j];
+          std::pair<double, double> yield_parameters (cohesions[j], angles_internal_friction[j]);
```
gassmoeller: This logic needs to be repeated in `fill_plastic_outputs` as @naliboff found. Instead of repeating the whole block we should put it into a new function of the following type:
```
template <int dim>
std::array<double, 3>
ViscoPlastic::compute_weakened_yield_parameters(const double viscosity_pre_yield)
```
which returns cohesion, internal friction angle, and modified viscosity. This way we remove the duplication and avoid future bugs like the one we had to search for this time.

jdannberg: @naliboff made some changes on his branch, but I think not quite what you suggest here, and I think he also rebased to a more current ASPECT version. Do you just want me to force-push all of his changes to this pull request, or should we close this one and @naliboff opens a new one?

jdannberg: Actually, I don't think you can, because my aspect is not forked from geodynamics/aspect. But I am happy to do it once your done. Whatever seems easier for you.

naliboff: @jdannberg After I finish the change @gassmoeller and I discussed today and update the tests, I'll rebase again to the current master. If you also rebase to the current master on your branch I think you should be able pull in the commits from my branch directly? Maybe there is a cleaner way?

naliboff: @jdannberg @gassmoeller - ok, my branch (below) has a new commit and was rebased a minute ok to the current ASPECT master. @jdannberg - can you try pulling in my two commits or rebasing relative to my branch?
https://github.com/naliboff/aspect/tree/strain_weakening_logic

Hopefully all that is left is updating the tests and doing a final check on documentation/logic.

jdannberg: Thanks! I just took your branch and force-pushed all of the changes. If you still have to make changes and don't rebase any more, I will simply be able to pull your changes to my branch.

naliboff: @jdannberg - Doh, I forgot to run astyle. I just added that commit to my branch.

naliboff: @jdannberg @gassmoeller - Hmm, three tests failing, but two appear to be due to small changes in screen-output. The third test actually has a small change to the solution. I guess we need a final look at that test (visco_plastic_yield_strain_weakening_compositions) to see if something is off with the new logic.

jdannberg:
I think it makes sense that the tests fail. At least the one with the composition runs into the bug that this pull request fixes. In the old version, even though use_viscous_strain_weakening == false, it would always apply the viscous strain weakening (see old line 205). So do we have a way to figure out if the new logic is correct?

naliboff:
@gassmoeller - Would it be possible to restart the CIG tester for this PR. Timo's tester is passing and I'm not sure what went wrong with the CIG tester ... it looks like it did not run?

@jdannberg - Halfway to having passing tests! I'll take another look at the code and then maybe someone with a fresh set of eyes (@anne-glerum) can take a look?

jdannberg:
Yes! I mean it seems that the tests work, so it would be good if we could get a review and finally merge this.

gassmoeller:
All tests pass, and we have looked over this more times than I can remember. @jdannberg if you fix the minor comments by John feel free to merge.

jdannberg:
Alright, I addressed the comments. Just waiting for the tester.

jdannberg:
Yay! Finally!
But @naliboff, didn't you still want to add something? Do you want to make a separate pull request?

naliboff:
@jdannberg - @gassmoeller was too quick for me! I submitted the updated documentation in a separate PR #2974, which @bangerth merged. Glad this is finally resolved!

PR #2706, opened 2018-11-14
cleanup function comments in utilities.cc

tjhei:


naliboff:
@tjhei - Just to be sure before merging, was the intent to remove all of the comments here?  Is the long-term to plan to remove most of the comments in `utilities.cc` and preserve the comments in `utilities.h`?

tjhei:
> Just to be sure before merging, was the intent to remove all of the comments here

Yes, this is a weirdly formatted entry in front of the function in the cc. It will not show up in doxygen. I think the function is adequately documented in the header. I would put implementation details that might be interesting inside the function itself.

>  Is the long-term to plan to remove most of the comments in 

Code comments or internal details need to stay but I did not think this was a useful comment to keep.

naliboff:
Gotcha 👍

PR #2707, opened 2018-11-19
avoid FGMRES unused argument

tjhei:
This will be deprecated in deal.II
(https://github.com/dealii/dealii/pull/7451) and the parameter has no
meaning to begin with.


bangerth:
The argument was already defaulted in 8.5, so removing the explicit value is backward compatible.

Issue #2715, opened 2018-11-28
Inconsistent code for compositional fields set by material models?


bangerth:
#2677 extended some code in `solver_schemes.cc` where we now also allow compositional fields to be set from material model outputs. The code here looks like this:
```
        switch (method)
          {
            case Parameters<dim>::AdvectionFieldMethod::fem_field:
            case Parameters<dim>::AdvectionFieldMethod::fem_melt_field:
            case Parameters<dim>::AdvectionFieldMethod::prescribed_field_with_diffusion:
            {
              // if this is a prescribed field with diffusion, we first have to copy the material model
              // outputs into the prescribed field before we assemle and solve the equation
              if (method == Parameters<dim>::AdvectionFieldMethod::prescribed_field_with_diffusion)
                interpolate_material_output_into_compositional_field(c);

              assemble_advection_system (adv_field);

              if (compute_initial_residual)
                (*initial_residual)[c] = system_rhs.block(introspection.block_indices.compositional_fields[c]).l2_norm();

              current_residual[c] = solve_advection(adv_field);

              // free matrix:
              const unsigned int block_idx = adv_field.block_index(introspection);
              if (adv_field.compositional_variable!=0)
                system_matrix.block(block_idx, block_idx).clear();
              break;
            }

            case Parameters<dim>::AdvectionFieldMethod::particles:
            {
              interpolate_particle_properties(adv_field);
              break;
            }

            case Parameters<dim>::AdvectionFieldMethod::prescribed_field:
            {
              interpolate_material_output_into_compositional_field(c);

              // Call the signal in case the user wants to do something with the variable:
              SolverControl dummy;
              signals.post_advection_solver(*this,
                                            adv_field.is_temperature(),
                                            adv_field.compositional_variable,
                                            dummy);
              break;
            }

            case Parameters<dim>::AdvectionFieldMethod::static_field:
            {
              // Do nothing here, but at least call the signal in case the
              // user wants to do something with the variable:
              SolverControl dummy;
              signals.post_advection_solver(*this,
                                            adv_field.is_temperature(),
                                            adv_field.compositional_variable,
                                            dummy);
            }
```
What's a little bit strange is that we call the `post_advection` signal for `prescribed_field`, but not for the other types above. Is this on purpose?

jdannberg:
For the fields that actually call the advection solver, the signal is called inside of the solve_advection() function. So only if we don't call the function, we have to call the post_advection signal separately. So yes, this is on purpose. 
[...]

PR #2716, opened 2018-11-28
Fix a syntactic oddity.

bangerth:
Found while looking at #2677 and #2715. The code is not wrong,
but unnecessarily hard to read.


tjhei:
ready to merge when tests pass.

PR #2722, opened 2018-11-28
Replace a plain pointer by a std::unique_ptr.

bangerth:
This also removes the current memory leak we have because the pointer is never
released.

Also in relation to #2375.


tjhei:
okay to merge when testers turn green

Issue #2719, opened 2018-11-28
Question about interpolate_material_output_into_compositional_field()


bangerth:
*Question 1:*
After a slight rewrite (patch to be submitted tomorrow), the documentation of this function now states this:
```
       * This function also sets the previous solution vectors (corresponding to the
       * solution from previous time steps) to the same interpolated
       * values. This implies that the compositional field method can then be
       * combined with time-dependent problems like advection or diffusion of
       * the field.
```
I don't think I understand the sentence. The only case I can think of where it could be used in a time dependent scheme is if we checkpoint the solution, and the restart it with an input file that makes the compositional field that was previously `prescribed_field` a regular `fem_field`. Or do you mean to say that because we set the old value, that it can now also participate in the time dependent solution of *other fields* that evaluate both the current and the previous solution? If the latter, would it not be the correct thing to leave the old state untouched so that we keep the previous time step's values in the `old_solution` vector?

*Question 2:*
The function sets `old_solution.block(...) = solution.block(...)`, but it does not touch `old_old_solution`. Intentional or oversight?

Related to #2677.

jdannberg:
Somehow I missed your question, @bangerth, but to come back to it:

_Question 1:_
This refers to what is done when we use the `prescribed_field_with_diffusion` compositional field method. We first interpolate a material model output onto a compositional field, and then we solve a diffusion equation for it to smooth the quantity we interpolated. Maybe "time-dependent" is not really the right word in this context, as the intent is just to smooth a quantity, but the idea is that you can still do other things with the interpolated quantities after putting them into the compositional fields. 

_Question 2:_
The reason we only do this for the old solution is that at the moment, this is only used in one place, the `prescribed_field_with_diffusion` method mentioned above, and there we only need the old solution. In this case, we need the old solution for the right-hand side of the diffusion equation (this is implemented in `DiffusionSystem<dim>::execute` in advection.cc), and because the idea is only to smooth the solution, it does not make sense to use the higher-order time-stepping we usually use in the advection equation. 
So this was intentional, but i am open to changing it if we think can come up with cases where it would be useful to also have these updated values in the `old_old_solution`. 

Does that answer your questions?

Issue #2718, opened 2018-11-28
Make Simulator<dim>::assemble_and_solve_composition() faster


bangerth:
The `Simulator<dim>::assemble_and_solve_composition` function works compositional field by compositional field to update each compositional field at a time. That's fine if we want to advect it, but it's inefficient for at least two of the cases handled there:

* When interpolating from particles to compositional fields
* When using data from material models (see #2677) because it calls `interpolate_material_output_into_compositional_field(c)` separately for each compositional field `c`.

For sure in the latter case it is unfortunate to call this function separately for each field because the function is exceedingly expensive: It needs to evaluate the solution at quadrature points, call the material model that computes *all* output variables, and then it throws away all but one of the output elements and copies that one into the compositional field.

The function, as written, is quite elegant. But I think it could be made significantly faster by checking whether any of the compositional fields need to be set from another source, and then do this all at once for all affected compositional fields.

bangerth:
I'll mention that the original implementation in #2677 did exactly that. I haven't gotten far enough through my patch backlog to see which PR changed it to the current behavior ;-)

PR #2724, opened 2018-11-28
Updates to the code for copying material model data into compositional fields.

bangerth:
These are all updates for code introduced mostly in #2677. The only functional
change is to rename interpolate_material_output_into_field() to
interpolate_material_output_into_compositional_field() to make clear
where the data goes.

I have some more questions about this code, but will ask that separately
once these cleanups are merged.

Related also to #2715, #2716, #2718, #2719.

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -1727,8 +1727,9 @@ namespace aspect
   }
 
 
+
   template <int dim>
-  void Simulator<dim>::interpolate_material_output_into_field (const unsigned int c)
+  void Simulator<dim>::interpolate_material_output_into_compositional_field (const unsigned int c)
   {
     // we need some temporary vectors to store our updates to composition in
     // before we copy them over to the solution vector in the end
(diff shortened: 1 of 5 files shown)


source/simulator/helper_functions.cc
```diff
@@ -1784,34 +1789,36 @@ namespace aspect
...
-
+    // Put the final values into the solution vector, also
+    // updating the ghost elements of the 'solution' vector.
     const unsigned int block_c = introspection.block_indices.compositional_fields[c];
     distributed_vector.block(block_c).compress(VectorOperation::insert);
     solution.block(block_c) = distributed_vector.block(block_c);
 
     // We also want to copy the values into the old solution, because it might
-    // be used in other parts of the code
+    // be used in other parts of the code.
```
tjhei: Wait, did I just manage to win our 5 year long battle on punctuation in comments? :-D

jdannberg: @tjhei, I am not so sure. Lines 1746 and 1792 do not have the proper punctuation :grin: 
But we should at least try to be consistent...

bangerth: Yes, you did :-) But that already happened several weeks ago :-))

@jdannberg  We are all inconsistent anyway. Do you want me to change other places as well?

jdannberg: It's not really important to me. 
But if we decide that from now on, we always want to have comments starting capitalized and using punctuation, then I would write all of my comments like this too, and if we just keep on doing it long enough, at some point most comments should be written that way. So I guess what I am saying is, we should at least be consistent in all new comments, but I don't care enough about it to delay merging a pull request (there are more useful things that can be done with that time).

bangerth: I agree and appreciate you not holding this PR up over it.

I have for 20 years written comments starting with lower-case letters and no final period. But @tjhei has berated me for a sufficiently long time that I'm willing to change my habits going forward. I have no real preference for one way or the other, but he seems to have.

jdannberg: Makes sense to me. For me, text is just easier to read with punctuation. 

So from my side this is ready to merge.

Issue #703, opened 2016-01-11
Time after loading checkpoint is not time when writing final checkpoint


gassmoeller:
I need to look more into this issue, but we had the case were we wrote a checkpoint upon termination of a model at t=100 Ma, and when restarting the code reported to restart at 540 Ma. Maybe there is something wrong with the logic of updating the internal time at final model timesteps.

bangerth:
That definitely sounds wrong :-) A testcase would help for sure.

gassmoeller:
I more or less convinced myself that the above reported behavior is expected (and I do not longer have the model to reproduce it). In our main loop we calculate the length of the next timestep before saving the checkpoint, and then use the saved timestep length after restarting for the next timestep. In the example I mentioned above we changed the velocity boundary conditions during the restart and wanted to first solve the new stokes system and then do the new advection steps, but that is not possible with our current loop layout. I may need to look at this again when it comes up again, but it is definitely not a bug, just an (inconvenient) feature.

PR #2725, opened 2018-11-28
Gravity postprocessor improvement 

LudovicJnnt:
The gravity postprocessor now computes gravity anomalies (relative to a constant density specified by the user) and gravity gradients (tensor in xx, yy, zz, xy, xz and yz). The theoretical gravity gradients of a hollow sphere filled with a constant density is also calculated. 
In addition, it is now possible to use the gravity postprocessor either for a "list" of points prescribed in the .prm or - as before - generate a "map" of points.
Finally, I added the possibility to run the gravity postprocessor at certain time or time steps using the same formulation than in the visualization postprocessor.

This figure shows the theoretical gravity gradients_xx for a map of points and the numerical values obtained at 10 specific points (above famous cities). The values obtained have been checked and the results are similar to Cedric's code. 
[g_gradients.pdf](https://github.com/geodynamics/aspect/files/2625936/g_gradients.pdf)

This figure shows gravity anomalies at 10km altitude above the surface model using S20RTS for initial temperature.
gravity_anomalies_s20rts


source/postprocess/gravity_point_values.cc
```diff
@@ -300,35 +486,88 @@ namespace aspect
     void
     GravityPointValues<dim>::parse_parameters (ParameterHandler &prm)
     {
+      std::string gravity_subdirectory = this->get_output_directory() + "output_gravity/";
+      Utilities::create_directory (gravity_subdirectory,
+                                   this->get_mpi_communicator(),
+                                   true);
+
+      end_time = prm.get_double ("End time");
```
gassmoeller: Ah, the termination criteria manager can not return its plugins? Then you need to do this. But you should still be able to remove the declaration of this input parameter from the `declare_parameters` function above. It is sufficient to declare a parameter once.

source/postprocess/gravity_point_values.cc
```diff
@@ -300,35 +486,88 @@ namespace aspect
...
+          if (this->convert_output_to_years())
+            output_interval *= year_in_seconds;
+          maximum_timesteps_between_outputs = prm.get_integer("Time steps between gravity output");
         }
         prm.leave_subsection();
       }
       prm.leave_subsection();
     }
 
+    // IS THIS NEEDED ???
```
gassmoeller: This is needed if you want to have correct behavior during checkpoint/restart cycles, so yes. Please remove comment. Did you test if the checkpoint/restart works? I.e. does the postprocessor write at the correct time after a restart?

source/postprocess/gravity_point_values.cc
```diff
@@ -76,46 +150,24 @@ namespace aspect
         }
       else
         {
-          Assert (false, ExcMessage ("This initial condition can only be used if the geometry "
```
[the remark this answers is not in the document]
LudovicJnnt: This assert was actually duplicated in the parse_parameter handler so I think it can be safely removed, right?

LudovicJnnt:
I have made all the changes and tests but I have two comments: 

1). I cannot use the density from the adiabatic function because it has a different value than the density from the material model. This actually confused me a lot in the past because the user specifies a reference density in the material model but on the screen-output log, the post-processing part writes a reference density which is from the adiabatic function. And the density from the graphical output shows the density from material model. 
Gravity is for me computed from the density of the material model because I am using an input for density directly to ASPECT and temperature is not involved. Moreover, the post-processor also computes the analytical solution of gravity of a geometry filled with a fixed density which is easier defined if the user defines it itself. 
Therefore i would suggest to keep a density input but instead of calling it _reference_density_, I called it _absolute_density_. That is because the absolute density is not only used for the analytical solution of gravity but also to calculate the density anomalies which give gravity anomalies. I have been thinking a lot of a name for that density. And I am still not sure that _absolute_density_ is the right term to use.

2). The checkpoint test gave me two results to discuss: 
   + resuming computation using a checkpoint which was the last at end model time (2000yr) results in continuing computation one step ahead of the end time (~2488yr). Is this a normal behavior?
   + writing the gravity output does not take into account the file number of the last output. I was not able to see where I went wrong on that.

gassmoeller:
To 1.: I agree that the current behavior is confusing, but adding more reference density values to more plugins is not going to make it simpler in the future. Lets instead try to find out what the problem with the current setup is, I did not quite understand your situation. Usually adiabatic conditions computes a reference density by asking the material model for a density with the input parameters of `Adiabatic surface temperature` and `Surface pressure`. This can be different from the reference density input to the material model, if the `Adabatic surface temperature` is different from the `Reference temperature` input parameter of the material model. You can check for all of the parameters of the adiabatic conditions by adding the `adiabat` option to the `List of output variables` parameter of the visualization output. Also check which value the reference temperature for the material model is given in the parameter file (or its default value). Can you also post an input file that produces different densities in the adiabatic conditions than in the material model? Because this discrepancy most likely means you have some inconsistency in the model setup.

To 2.: 
- This is known behavior. When you resume from a checkpoint ASPECT does not repeat the last timestep, but continues with the next one, therefore the time is already set to the next timestep. Also see #703.
- What do you mean by `does not take into account`? Does the first output of the resumed run overwrite the last output of the first run (i.e. is `output_file_number` not properly incremented)? Or does it start from 0 again (i.e. it is not properly stored and loaded)?

Also please go over the code one more time to make sure you removed all commented code lines. These can be absolutely confusing when they get accidentally merged.

LudovicJnnt:
To 1.: 
It took me a while to understand what is going on. According to how you described the `adiabatic surface temperature`, it seems indeed that there is an inconsistency in my model setup. I used to have the default value `adiabatic surface temperature` of 0 all the time while the `reference temperature` in the material model was 1600. For a reference density of 3300 set in material model, this resulted in a reference density different with the adiabat. 
Is there any cases where a different `adiabatic surface temperature` from the material model is fine? or would it be necessary to have a warning somewhere that this is happening? I have check the manual but I did not find that clear somewhere. Or maybe I missed it.
 
I agree on one point is that we should not add an additional reference density. However, the density user-defined in the gravity post-processor would not be a reference density because it has a different purpose:
+ purpose 1: benchmark the gravity postprocessor with computing analytically gravity acceleration and gradients in the domain filled with a constant density. 
+ purpose 2: calculate the density difference from an absolute density given from an ascii file, and a use-defined average density. 
I am now using a dataset of density loaded as an `initial composition`, and here I must put the reference density in `material model` to 1 which I suppose add up with the density from the ascii file (reference density in material model cannot be 0, and setting it to 3300 results in density in model of about 6600) . Therefore I cannot use the density of adiabat (or any other reference density) because it returns the value 1 and not a density of 3300. Unless I am setting up the model wrong? 
I find quite difficult to find another name for the density in this post-processor. At some point I came up with _Medium density for gravity anomalies and uniform earth gravity_ but it seems quite long

To 2.: 
regarding the second point about gravity output file number, if I resume computation, it starts from 0 again. I wrote this part follow the example of the visualization post-processor and I just had a look again but I cannot see where I missed something.

gassmoeller:
To 1: You know what? Before I wrote this comment I had authored a two page reply to your comment about how to rectify the architecture, modify the material model and adiabatic condition to finally be able to use their densities and had discussed any corner case to work around to make that work. When I looked at my reply I figured it would be easier to simply have a another parameter called `Reference density` for this postprocessor as you originally suggested. It is not ideal, but simpler than fixing everything else in the architecture that currently forces you to do that. Could you please change the name back to reference density (to at least be consistent with the other parameters)? Sorry for the hassle, I thought it would be simpler to come up with a better solution. Still remove the code lines that are commented at the moment. Thanks for the patience!

To 2: I just saw: You will also need to copy the `save` and `load` functions from the `Visualization` class to make serialization work. Their content and interface can stay exactly the same (remove the comment about the visualization plugins). Let me know if that works.

LudovicJnnt:
To 1.: Oh ok, that seems like a lot of work and thinking! No worries. I added further explanations in the instantiations and the description of the head file to be as clear as possible about the use of this reference density. 

To 2.: now it works, thanks!

Let me know if there is anything else to improve. Otherwise I am happy with it.

gassmoeller:
I modified the explanation a bit, but otherwise I think this is ready to merge when the testers are done.

gassmoeller:
Thanks for tolerating all the reviews, I think the postprocessor has improved significantly
 :+1:

LudovicJnnt:
Thanks for modifying the explanations for the better.
👍 :)

PR #2727, opened 2018-11-29
Use std::make_shared/unique instead of plain 'new'.

bangerth:
Part of using C++11 better, see #2375.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -109,15 +109,16 @@ namespace aspect
      * computation.
      */
     template <int dim>
-    Mapping<dim> *construct_mapping(const GeometryModel::Interface<dim> &geometry_model,
-                                    const InitialTopographyModel::Interface<dim> &initial_topography_model)
+    std::unique_ptr<Mapping<dim>>
+                               construct_mapping(const GeometryModel::Interface<dim> &geometry_model,
+                                                 const InitialTopographyModel::Interface<dim> &initial_topography_model)
     {
       if (geometry_model.has_curved_elements())
-        return new MappingQ<dim>(4, true);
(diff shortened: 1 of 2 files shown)


source/simulator/core.cc
```diff
@@ -143,10 +144,14 @@ namespace aspect
...
                              ParameterHandler &prm)
     :
-    assemblers (new Assemblers::Manager<dim>()),
+    assemblers (std_cxx14::make_unique<Assemblers::Manager<dim>>()),
     parameters (prm, mpi_communicator_),
-    melt_handler (parameters.include_melt_transport ? new MeltHandler<dim> (prm) : nullptr),
-    newton_handler (parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_Newton_Stokes ? new NewtonHandler<dim> () : nullptr),
+    melt_handler (parameters.include_melt_transport ?
+                  std_cxx14::make_unique<MeltHandler<dim>>(prm) :
+                  std::unique_ptr<MeltHandler<dim>>()),
```
tjhei: can you use ``  ... : nullptr`` here? If yes, wouldn't that be more clear and more concise?

bangerth:
Like this? I was concerned that using different types in the `?:` expression would make things ambiguous, but it seems to compile for me.

tjhei:
> expression would make things ambiguous

``nullptr`` is convertible to a ``unique_ptr<...>`` so the resulting type should be the common type of the two args.

tjhei:
> But `std::unique_ptr` is also convertible to a regular pointer, isn't it?

But it can not be converted to ``nullptr_t``, while the second arg can be converted to the first type.

bangerth:
Excellent point!

PR #2731, opened 2018-12-01
Move benchmark, add test

gassmoeller:
Follow up to #2598. The comparison is really more of a benchmark with the description and figure, but since I changed my mind about that before (sorry Jacky), I did not want to delay the PR further. This only moves a few files to benchmarks and adds a test that uses the benchmark prm file as base. The test did not work properly on my system (it always passed no matter what output I provided) and I could not figure out what was going wrong, lets see what the tester says.


jaustermann:
This looks good to me, thanks Rene for merging the previous PR and moving the test to the benchmark. It's not clear to me either why the test would always pass independent of the provided output.

bangerth:
@jaustermann : If you're happy with it, you can merge it -- you've got that privilege!

PR #2734, opened 2018-12-07
Fix the documentation of the compositional smoothing method.

bangerth:
This is a follow-up to #2693. The formula is just not right: It applies the
Laplace operator to the right hand side when it should apply its inverse.

There was also a missing underscore. Finally, because the length scale is
a constant, I simplified the formula a bit.


source/simulator/parameters.cc
```diff
@@ -1046,7 +1046,7 @@ namespace aspect
                          "the ``prescribed field'' method. Afterwards, the field is diffused based on "
                          "a solver parameter, the diffusion length scale, smoothing the field. "
                          "Specifically, the field is updated by solving the equation "
-                         "$(I-\\nabla \\cdot l^2 \\nabla) C_\\text{smoothed} = C_\\text{prescribed}$, "
+                         "$(I-l^2 \\Delta) C_\\text{smoothed} = C_\\text{prescribed}$, "
```
jdannberg: I like your other changes, but for this one, I just find `\\nabla \\cdot l^2 \\nabla` easier to read (as the temperature diffusion equation is written in the same way). Here, it doesn't make a difference because `l^2` is constant, but for our normal diffusion equation it does. 
But if you feel strongly about this, we can change it like you suggested.

bangerth:
I squashed the three commits, even though they were logically independent and addressed separate problems :-)

I think I like the use of the Laplace operator better than the `div l^2 grad` thing. Most people will have an intuition what the former means, whereas the latter is more of a specialized thing you learn when you start to deal with spatially variable coefficients. So I left it as is.

PR #2738, opened 2018-12-11
switch to deal.II 9.0

tjhei:
- require deal.II 9.0.0
- switch to new docker image for the tester
- update test output


tjhei:
@gassmoeller this is interesting: switching the docker image to the new one based on the same deal.II version only breaks a single test where we see a difference of 1 iteration. This image should have very similar (same?) software versions. The only difference that matters here should be mpich instead of openmpi.
I don't think we need to make the switch now, but rather discuss switching to deal.II 9.0 immediately. Should we do this and if yes, disable 8.5 support at the same time?

gassmoeller:
Hmm, different iterations on different MPI versions? Interesting, but at least it is only 1 test. I agree, instead of switching the image twice, lets just move on to deal.II 9.0, and also require 9.0 at the same time.

bangerth:
Didn't we have a discussion in which we argued that we wanted to stick with 8.5 for one more ASPECT release?

tjhei:
okay, this passes for deal.II 9.0 now! @gassmoeller do you want to check if the diffs look good? :-)

tjhei:
I updated the branch. One problem: melt_transport_compressible_iterative crashes with ``Iterative method reported convergence failure in step 1000``.

gassmoeller:
Yes that test always tend to create problems (it's underresolved). Let me take a look at the setup.

gassmoeller:
I checked the test. It converges for me using deal.II 9.0.1, trilinos 12.10.1, p4est 2.0 (though barely, using 200+627 iterations at one time). But if you increase the resolution by one (to 5) it converges nicely, and is actually faster, because of the better convergence. I would suggest to simply increase the resolution by one to make the test more useful. Since the test produces no visualization output it will not increase the output size.

gassmoeller:
I took the liberty of updating the cig Jenkinsfile as well to see how our tester reacts to the changes.

gassmoeller:
I updated melt_transport_compressible_iterative. Timo could you rebase this PR so that it includes the changes and new tests on master (do not want to force push to your banch). I can take care of updating the rest then.

tjhei:
thanks, increasing the resolution for the test makes sense.

Rebased...

gassmoeller:
Looks good to go to me. Also seems like the CIG tester is reacting nicely to your new docker image. I will create a follow-up next week that merges our Jenkinsfiles to avoid the duplication. 

Merge if you think this is ready.

tjhei:
Wow, looking back, it is quite easy to move to a new deal.II version. :+1: 
I am not concerned moving to 9.1 at some point at all.

tjhei:
> week that merges our Jenkinsfiles to avoid the duplication.

We can discuss during the hackathon: It might make sense to keep different configurations. No reason to run the identical things...

tjhei:
well, then. Let's move to the future. :-)

Issue #2740, opened 2018-12-13
Bug in maximum horizontal compressive stress postprocessor


gassmoeller:
@MFraters discovered a bug in the maximum horizontal compressive stress postprocessor that is likely affecting results of this postprocessor, for a discussion see: https://community.geodynamics.org/t/problem-with-maximum-horizontal-compressive-stress-computation/245/6

This issue is just to keep track of the bug.

gassmoeller:
Fixed by #2833.

PR #2743, opened 2018-12-17
[WIP] Fix maximum horizontal compressive stress and add eigenvalue postprocessors

MFraters:
work in progress related to #2740.

### DIFF
--- a/include/aspect/postprocess/visualization/shear_stress_eigenvectors.h
+++ b/include/aspect/postprocess/visualization/shear_stress_eigenvectors.h
@@ -1,0 +1,89 @@
+/*
+  Copyright (C) 2016 - 2017 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_postprocess_visualization_shear_stress_eigenvectors_h
+#define _aspect_postprocess_visualization_shear_stress_eigenvectors_h
+
+#include <aspect/postprocess/visualization.h>
+#include <aspect/simulator_access.h>
+
+#include <deal.II/numerics/data_postprocessor.h>
+
+
+namespace aspect
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      /**
+       * A class that computes a field of horizontal vectors that
+       * represent the direction of maximal horizontal compressive
+       * stress. For an exact definition, see the documentation of
+       * this plugin in the manual.
+       *
+       * The member functions are all implementations of those declared in the
+       * base class. See there for their meaning.
+       */
+      template <int dim>
+      class ShearStressEigenvectors
+        : public DataPostprocessor<dim>,
+          public SimulatorAccess<dim>,
+          public Interface<dim>
+      {
+        public:
+          virtual
+          void
+          evaluate_vector_field(const DataPostprocessorInputs::Vector<dim> &input_data,
+                                std::vector<Vector<double> > &computed_quantities) const;
+
+          /**
+           * Return the vector of strings describing the names of the computed
+           * quantities. Given the purpose of this class, this is a vector
+           * with entries all equal to the name of the plugin.
+           */
+          virtual std::vector<std::string> get_names () const;
+
+          /**
+           * This functions returns information about how the individual
+           * components of output files that consist of more than one data set
+           * are to be interpreted. The returned value is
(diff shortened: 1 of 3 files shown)


gassmoeller:
Is the `Fix maximum horizontal compressive stress` part of this PR already addressed in #2833? If so please change the title of the PR. Otherwise I always get nervous when reading old open bug fixes that have not been merged.

punch_box_3d.prm
```diff
@@ -0,0 +1,119 @@
+
```
bangerth: can you add a couple of lines of explanation of what the test does?

shear_box_3d.prm
```diff
@@ -0,0 +1,116 @@
+
```
bangerth: same here

source/postprocess/visualization/maximum_horizontal_compressive_stress.cc
```diff
@@ -111,7 +111,7 @@ namespace aspect
                   const double b = orthogonal_directions[1] *
                                    (compressive_stress *
                                     orthogonal_directions[1]);
-                  const double c = orthogonal_directions[0] *
+                  const double c = 2.0*orthogonal_directions[0] *
```
bangerth: The comment above this block references the formulas in the documentation. Do they also have the factor of 2?

Does this change affect the output of any tests?

include/aspect/postprocess/visualization/shear_stress_eigenvectors.h
```diff
@@ -0,0 +1,89 @@
...
+
+namespace aspect
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      /**
+       * A class that computes a field of horizontal vectors that
+       * represent the direction of maximal horizontal compressive
```
bangerth: shear?

source/postprocess/visualization/shear_stress_eigenvectors.cc
```diff
@@ -0,0 +1,340 @@
...
+            double z = p[2];
+            double e = 1.1e-1;
+            Tensor<1,dim> ev0 = eigenvector_array[0].second;
+            Tensor<1,dim> ev1 = eigenvector_array[1].second;
+            Tensor<1,dim> ev2 = eigenvector_array[2].second;
+            //if(fabs(x-0.5) < e && fabs(y-0.25) < e && fabs(z-0.375) < e || fabs(x-0.625) < e && fabs(y-0.375) < e && fabs(z-0.5) < e)
+            {
+              //std::cout << x << "," << y << "," << z << ", strain rate = " << compressible_strain_rate << std::endl;
+              //std::cout << x << "," << y << "," << z << ": tau_magnitude = " << tau_magnitude <<  ", eigenvector 0 = " << eigenvector_array[0].second << ", 1 = " << eigenvector_array[1].second<< ", 2 = " << eigenvector_array[2].second << " = " << tau_direction << ", norm = " << tau_direction/tau_direction.norm() << std::endl;
+              std::cout << x << " " << y << " " << z << " "
```
bangerth: remove all of this debug output

source/postprocess/visualization/shear_stress_eigenvectors.cc
```diff
@@ -0,0 +1,340 @@
...
+      std::vector<std::string>
+      ShearStressEigenvectors<dim>::get_names () const
+      {
+        std::vector<std::string> names;
+        switch (dim)
+          {
+            case 2:
+              names.emplace_back("shear_stress_eigenvectors_sigma_1");
+              names.emplace_back("shear_stress_eigenvectors_sigma_1");
+              names.emplace_back("shear_stress_eigenvectors_sigma_1");
```
bangerth: is this a mistake? shouldn't these only be repeated twice instead of three times? same for the next three lines, of course

source/postprocess/visualization/shear_stress_eigenvectors.cc
```diff
@@ -0,0 +1,340 @@
...
+namespace aspect
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      ASPECT_REGISTER_VISUALIZATION_POSTPROCESSOR(ShearStressEigenvectors,
+                                                  "shear stress eigenvectors",
+                                                  "A plugin that computes the direction and magnitude of the "
+                                                  "maximum horizontal component of the compressive stress as a vector "
```
bangerth: You'll have to go through the documentation here and update it. I think this is copied from the other class?

It would probably also be good to add a sentence to both this and the other plugin, referencing each other.

source/postprocess/visualization/shear_stress_eigenvectors.cc
```diff
@@ -0,0 +1,340 @@
...
+                                                  "    in front-back direction, the compressive strength "
+                                                  "    in front-back direction is zero, making the along-ridge "
+                                                  "    direction the dominant one. At the center of the "
+                                                  "    convection rolls, both horizontal directions yield "
+                                                  "    the same stress; the plugin therefore chooses an "
+                                                  "    essentially arbitrary horizontal vector, but then "
+                                                  "    uses a zero magnitude given that the difference "
+                                                  "    between the maximal and minimal horizontal stress "
+                                                  "    is zero at these points.}"
+                                                  "  \\label{fig:max-horizontal-compressive-stress}"
```
bangerth: all of this doesn't seem like it was updated from the other plugin.

source/postprocess/visualization/maximum_horizontal_compressive_stress.cc
bangerth: This is #2833, right?

PR #2744, opened 2018-12-19
Nonlinear solver: deal with failure

tjhei:
This introduces a new setting to decide what should happen if the
nonlinear solver fails to converge.

The old behavior was to "just continue", which is an unsafe choice.

### DIFF
--- a/include/aspect/parameters.h
+++ b/include/aspect/parameters.h
@@ -98,6 +98,38 @@ namespace aspect
     }
 
     /**
+     * A struct that contains the enums to decide what to do when a nonlinear solver fails.
+     */
+    struct NonlinearSolverFailureStrategy
+    {
+      enum Kind
+      {
+        continue_with_next_timestep,
+        cut_timestep_size,
+        abort_program
+      };
+
+      /**
+       * Parse the enum value from a string.
+       */
+      static
+      Kind
+      parse(const std::string &input)
+      {
+        if (input == "continue with next timestep")
+          return continue_with_next_timestep;
+        else if (input == "cut timestep size")
+          return cut_timestep_size;
+        else if (input == "abort program")
+          return abort_program;
+        else
+          AssertThrow(false, ExcNotImplemented());
+
+        return Kind();
+      }
+    };
+
+    /**
      * @brief The NullspaceRemoval struct
      */
     struct NullspaceRemoval
@@ -484,6 +516,7 @@ namespace aspect
      * @{
      */
     typename NonlinearSolver::Kind nonlinear_solver;
+    typename NonlinearSolverFailureStrategy::Kind nonlinear_solver_failure_strategy;
 
     typename AdvectionStabilizationMethod::Kind advection_stabilization_method;
     double                         nonlinear_tolerance;
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -169,6 +169,14 @@ namespace aspect
   };
 
   /**
+   * Exception to be thrown when the nonlinear solver needs too many iterations to converge.
+   */
+  DeclExceptionMsg(ExcNonlinearSolverNoConvergence,
+                   "Nonlinear solver failed to converge in the prescribed number of steps. "
+                   "Consider changing `Max nonlinear iterations` or `Nonlinear solver failure "
+                   "strategy`.");
+
+  /**
    * This is the main class of ASPECT. It implements the overall simulation
    * algorithm using the numerical methods discussed in the papers and manuals
    * that accompany ASPECT.
@@ -1995,6 +2003,7 @@ namespace aspect
       unsigned int                                              timestep_number;
       unsigned int                                              pre_refinement_step;
       unsigned int                                              nonlinear_iteration;
+      unsigned int                                              nonlinear_solver_failures;
       /**
        * @}
        */
--- a/include/aspect/time_stepping/interface.h
+++ b/include/aspect/time_stepping/interface.h
@@ -203,7 +203,6 @@ namespace aspect
         void
         write_plugin_graph (std::ostream &output_stream);
 
-
         /**
          * A function that is used to register time stepping model objects in such
          * a way that the Manager can deal with all of them without having to
@@ -258,6 +257,8 @@ namespace aspect
          */
         TerminationCriteria::Manager<dim> termination_manager;
     };
+
+
 
     /**
      * Given a class name, a name, and a description for the parameter file, register it with the
--- a/include/aspect/time_stepping/repeat_on_nonlinear_fail.h
+++ b/include/aspect/time_stepping/repeat_on_nonlinear_fail.h
@@ -1,0 +1,106 @@
+/*
+  Copyright (C) 2018 - 2023 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_time_stepping_repeat_on_nonlinear_fail_h
+#define _aspect_time_stepping_repeat_on_nonlinear_fail_h
+
+#include <aspect/time_stepping/interface.h>
+
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    using namespace dealii;
+
+    /**
+     * A class that implements a time stepping plugin to repeat a time step if the
+     * nonlinear solver failed to converge in the specified number of iterations.
+     * The timestep size will be reduced by the given amount specified by the user
+     * and hopefully results in the nonlinear solver converging. If necessary, the
+     * timestep size will be repeatedly reduced.
+     *
+     * This class is automatically enabled if the user specifies that he/she wants
+     * to do this.
+     *
+     * @ingroup TimeStepping
+     */
+    template <int dim>
+    class RepeatOnNonlinearFail : public Interface<dim>, public SimulatorAccess<dim>
+    {
+      public:
+        /**
+         * Constructor.
+         */
+        RepeatOnNonlinearFail ();
+
+        /**
+         * @copydoc aspect::TimeStepping::Interface<dim>::execute()
+         */
+        double
+        execute() override;
+
+        /**
+         * This function notifies the plugin that the nonlinear solver
+         * failed.
+         */
+        void nonlinear_solver_has_failed() const;
+
+        /**
+         * The main execute() function.
+         */
+        std::pair<Reaction, double>
+        determine_reaction(const TimeStepInfo &info) override;
+
+        static
+        void
+        declare_parameters (ParameterHandler &prm);
+
+        void
+        parse_parameters (ParameterHandler &prm) override;
+
+      private:
+        /**
+         * Parameter to determine how much smaller the time step should be
+         * repeated as.
+         */
+        double cut_back_factor;
+
+        /**
+         * Enabled by nonlinear_solver_has_failed() to signal that this
+         * plugin needs to act in the current timestep;
+        */
+        mutable bool nonlinear_solver_just_failed;
+
+        /**
+         * How many times should we try cutting the timestep size before giving up?
+        */
+        unsigned int maximum_number_of_repeats;
+
+        /**
+         * How many times have we been repeating already in this timestep?
+        */
+        unsigned int current_number_of_repeats;
+    };
+  }
+}
+
+
+#endif
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -37,6 +37,7 @@
 #include <aspect/simulator/assemblers/interface.h>
 #include <aspect/geometry_model/initial_topography_model/zero_topography.h>
 #include <aspect/material_model/rheology/elasticity.h>
+#include <aspect/time_stepping/repeat_on_nonlinear_fail.h>
 
 #include <deal.II/base/index_set.h>
 #include <deal.II/base/conditional_ostream.h>
@@ -195,6 +196,7 @@ namespace aspect
     old_time_step (numbers::signaling_nan<double>()),
     timestep_number (numbers::invalid_unsigned_int),
     nonlinear_iteration (numbers::invalid_unsigned_int),
+    nonlinear_solver_failures (0),
 
     // We need to disable eliminate_refined_boundary_islands as this leads to
     // a deadlock for deal.II <= 9.2.0 as described in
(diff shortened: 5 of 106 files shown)


tjhei:
@jdannberg something like this?

source/simulator/core.cc
```diff
@@ -1767,15 +1767,53 @@ namespace aspect
...
-        // Only solve if we are not in pre-refinement, or we do not want to skip
-        // solving in pre-refinement.
+        // During pre-refinement, do not solve if we are asked to skip it:
         if (! (parameters.skip_solvers_on_initial_refinement
                && pre_refinement_step < parameters.initial_adaptive_refinement))
           {
             start_timestep ();
 
             // then do the core work: assemble systems and solve
-            solve_timestep ();
```
gassmoeller: I somewhat dislike adding so many lines in the main run loop that are unrelated to it. Conceptionally it is the responsibility of the solve_timestep function to handle non-convergence. Could we move the whole try-catch block into solve_timestep, and modify the code in `run` to the following:
```
const bool retry_timestep = solve_timestep();
if (retry_timestep)
  continue;
```

All of the exception handling can stay the same, but it keep this loop readable.

source/simulator/parameters.cc
```diff
@@ -221,6 +221,14 @@ namespace aspect
                        "The `Newton Stokes' scheme is deprecated and only allowed for reasons of "
                        "backwards compatibility. It is the same as `iterated Advection and Newton Stokes'.");
 
+    prm.declare_entry ("Nonlinear solver failure strategy", "abort program",
+                       Patterns::Selection("continue with next timestep|cut timestep size|abort program"),
+                       "Select the strategy on what to do if the nonlinear solver scheme fails to "
+                       "converge. The options are:\n"
+                       "`continue with next timestep`: ignore error and continue to the next timestep\n"
+                       "`cut timestep size`: reduce the current timestep size and redo the timestep\n"
```
gassmoeller: maybe `by a factor of two`?

source/simulator/core.cc
```diff
@@ -1767,15 +1767,53 @@ namespace aspect
...
+            try
+              {
+                solve_timestep ();
+              }
+            catch (...)
+              {
+                pcout << "ERROR: the solver in the current timestep failed to converge." << std::endl;
```
gassmoeller: What about
```
pcout << "WARNING: The nonlinear solver in the current timestep failed to converge." << std::endl
          << "Acting according to the parameter 'Nonlinear solver failure strategy'." << std::endl;
```

source/simulator/core.cc
```diff
@@ -1767,15 +1767,53 @@ namespace aspect
...
+                    == Parameters<dim>::NonlinearSolverFailureStrategy::continue_with_next_timestep)
+                  {
+                    // do nothing and continue
+                  }
+                else if (parameters.nonlinear_solver_failure_strategy
+                         == Parameters<dim>::NonlinearSolverFailureStrategy::cut_timestep_size)
+                  {
+                    if (timestep_number == 0)
+                      {
+                        pcout << "Note: we can not cut the timestep in step 0, so we are aborting."
```
gassmoeller: I would call this an `ERROR:...`

include/aspect/simulator.h
```diff
@@ -123,6 +123,14 @@ namespace aspect
     template <int dim>      class Manager;
   }
 
+  /**
+   * Exception to be thrown when the nonlinear solver needs to many iterations to converge.
```
gassmoeller: to -> too

source/simulator/core.cc
```diff
@@ -1767,15 +1767,53 @@ namespace aspect
...
+                              << std::endl;
+                        throw;
+                      }
+
+                    // reduce timestep size and update time to "go back":
+                    const double new_time_step = 0.5*time_step;
+                    pcout << "\tReducing timestep to " << new_time_step << '\n' << std::endl;
```
jdannberg, line 1802: This output should be in years or seconds, depending on the parameter.

source/simulator/core.cc
```diff
@@ -1767,15 +1767,53 @@ namespace aspect
...
+
+                    time -= time_step - new_time_step;
+                    time_step = new_time_step;
```
jdannberg, line 1805: There's one thing that I think doesn't work well with the current implementation: The time step size is reduced for the current time step, until the solver converges, but the next time step bases the time step size on the CFL number (rather than on what time step size was necessary to actually get the nonlinear solver to converge). So if the problem stays as hard to solve, we need to go though a number of failing nonlinear solver loops for every time step, rather than continuing with the new (smaller) time step that is adapted to the nonlinear problem.

tests/nonlinear_failure_strategy_cut.cc
```diff
@@ -0,0 +1 @@
+#include "../benchmarks/nonlinear_channel_flow/simple_nonlinear.cc"
```
jdannberg, line 1: It looks like this test problem is some sort of benchmark. So would it make sense to do some postprocessing, so that we see that our new method actually still gives us the correct solution?

source/simulator/core.cc
tjhei: That makes sense, but how do you want to decide when to make bigger timestep again? I don't think there is a good answer here. One thing that might help in an extreme case is not increasing it by more than a factor of say 2 in each step. But in general, this seems to be something that would require more user input.

tjhei:
I am resurrecting this old PR because @danieldouglas92 is interested in it. I rebased and I am curious to see if things still work, but I know I have to make some more changes.

tjhei:
This has been updated to the current main branch and now uses the TimeStepping infrastructure. The test added here is working correctly.

I would be grateful if somebody can take a look at the current state.

tjhei:
@jdannberg ? @gassmoeller ?

gassmoeller:
I will take a look asap.

tjhei:
it helps to include all files. :-)

include/aspect/simulator.h
gassmoeller: too

include/aspect/time_stepping/interface.h
```diff
@@ -234,6 +234,36 @@ namespace aspect
         void
         parse_parameters (ParameterHandler &prm);
 
+        /**
+         * Go through the list of all mesh refinement strategies that have been selected
```
gassmoeller, line 238:
```suggestion
         * Go through the list of all time stepping plugins that have been selected
```

include/aspect/time_stepping/interface.h
```diff
@@ -234,6 +234,36 @@ namespace aspect
...
+         * This function can only be called if the given template type (the first template
+         * argument) is a class derived from the Interface class in this namespace.
+         */
+        template <typename PluginType,
+                  typename = typename std::enable_if_t<std::is_base_of<Interface<dim>,PluginType>::value>>
+        bool
+        has_matching_plugin () const;
+
+        /**
+         * Go through the list of all mesh refinement strategies that have been selected
```
gassmoeller, line 252:
```suggestion
         * Go through the list of all time stepping plugins that have been selected
```

include/aspect/time_stepping/interface.h
```diff
@@ -234,6 +234,36 @@ namespace aspect
...
+         * in the input file (and are consequently currently active) and see
+         * if one of them has the type specified by the template
+         * argument or can be casted to that type. If so, return a reference
```
gassmoeller, line 255:
```suggestion
         * argument or can be cast to that type. If so, return a reference
```

include/aspect/time_stepping/interface.h
```diff
@@ -308,6 +337,50 @@ namespace aspect
...
+
+    template <int dim>
+    template <typename PluginType, typename>
+    inline
+    bool
+    Manager<dim>::has_matching_plugin () const
+    {
+      for (typename std::list<std::unique_ptr<Interface<dim>>>::const_iterator
+           p = active_plugins.begin();
+           p != active_plugins.end(); ++p)
```
gassmoeller, line 350: want to make this a range-based loop?

include/aspect/time_stepping/interface.h
```diff
@@ -308,6 +337,50 @@ namespace aspect
...
+    {
+      AssertThrow(has_matching_plugin<PluginType> (),
+                  ExcMessage("You asked TimeStepping::Manager::get_matching_plugin() for a "
+                             "plugin of type <" + boost::core::demangle(typeid(PluginType).name()) + "> "
+                             "that could not be found in the current model. Activate this "
+                             "plugin in the input file."));
+
+      for (typename std::list<std::unique_ptr<Interface<dim>>>::const_iterator
+           p = active_plugins.begin();
+           p != active_plugins.end(); ++p)
```
gassmoeller, line 373: range-based loop?

include/aspect/time_stepping/repeat_on_nonlinear_fail.h
```diff
@@ -0,0 +1,102 @@
...
+{
+  namespace TimeStepping
+  {
+    using namespace dealii;
+
+    /**
+     * A class that implements a time stepping plugin to repeat a time step if the
+     * nonlinear solver failed to converge in the specified number of iterations.
+     * This class is automatically enabled if the user specifies that he/she wants
+     * to do this.
```
gassmoeller, line 37: Can you add a sentence why the result for the next try would be different from the previous one? Currently it reads as if you may get a different result if you only reran an exactly identical problem.

source/simulator/core.cc
gassmoeller: please check my comment from 2019: https://github.com/geodynamics/aspect/pull/2744/files#r244753861
does this still apply?

source/simulator/parameters.cc
```diff
@@ -265,6 +265,15 @@ namespace aspect
                        "The `Newton Stokes' scheme is deprecated and only allowed for reasons of "
                        "backwards compatibility. It is the same as `iterated Advection and Newton Stokes'.");
 
+    prm.declare_entry ("Nonlinear solver failure strategy", "abort program",
+                       Patterns::Selection("continue with next timestep|cut timestep size|abort program"),
+                       "Select the strategy on what to do if the nonlinear solver scheme fails to "
+                       "converge. The options are:\n"
+                       "`continue with next timestep`: ignore error and continue to the next timestep\n"
+                       "`cut timestep size`: reduce the current timestep size by a factor of 2 and redo "
```
gassmoeller, line 273: I think it would be better to not hard-code the factor of 2, because it is an input parameter of that plugin. Please update the documentation.

include/aspect/parameters.h
```diff
@@ -97,6 +97,38 @@ namespace aspect
              false;
     }
 
+    /**
+     * A struct that contains the enums to decide what to do when a nonlinear solver fails.
+     */
+    struct NonlinearSolverFailureStrategy
```
bangerth, line 103: This might be more easily done with a namespace. Your `struct` does not have data members.

bangerth: I recognize that there are other structs of the same kind. I'm going to fix those too right next.

source/simulator/solver_schemes.cc
```diff
@@ -814,6 +814,7 @@ namespace aspect
       }
     while (nonlinear_solver_control.check(nonlinear_iteration, relative_residual) == SolverControl::iterate);
 
+    AssertThrow(nonlinear_solver_control.last_check() != SolverControl::failure, ExcNonlinearSolverNoConvergence());
     signals.post_nonlinear_solver(nonlinear_solver_control);
```
gassmoeller, line 818: So whether we throw the assert before or after the signal may actually have some impact on the behavior. `post_nonlinear_solver` e.g. informs the postprocessor about the number of nonlinear iterations, and the postprocessor probably keeps accumulating the number of Stokes iterations until it receives the signal that the nonlinear solver has finished. That being said I dont know if the other order would be better, it could mean that one timestep appears twice in the statistics file. Have you check what the output statistics file looks like for a repeated timestep?

include/aspect/parameters.h
bangerth: Ah, drats. These are class members. You can't have a namespace be a class member :-( (That's one of my long-running complaints about C++...)

include/aspect/parameters.h
```diff
@@ -97,6 +97,38 @@ namespace aspect
...
+    {
+      enum Kind
+      {
+        continue_with_next_timestep,
+        cut_timestep_size,
+        abort_program
+      };
+
+      /**
+       * parse the enum value from a string
```
bangerth, line 113:
```suggestion
       * Parse the enum value from a string.
```

source/time_stepping/repeat_on_nonlinear_fail.cc
```diff
@@ -0,0 +1,125 @@
...
+    std::pair<Reaction, double>
+    RepeatOnNonlinearFail<dim>::determine_reaction (const TimeStepInfo &/*info*/)
+    {
+      if (nonlinear_solver_just_failed)
+        {
+          nonlinear_solver_just_failed = false;
+          ++current_number_of_repeats;
+
+          if (current_number_of_repeats >= maximum_number_of_repeats)
+            throw "waaa!";
```
gassmoeller, line 57: waaa?

source/time_stepping/repeat_on_nonlinear_fail.cc
```diff
@@ -0,0 +1,125 @@
...
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    ASPECT_REGISTER_TIME_STEPPING_MODEL(RepeatOnNonlinearFail,
+                                        "repeat on nonlinear solver failure",
+                                        "This time stepping plugin will react when the nonlinear solver "
+                                        "does not converge in the specified maximum number of iterations "
+                                        "and repeats the current timestep with a smaller step size. "
+                                        "This plugin is enabled automatically if \"\" "
```
gassmoeller, line 122: parameter name missing in description

tests/nonlinear_failure_strategy_cut/statistics
```diff
@@ -0,0 +1,17 @@
...
+# 6: Number of temperature degrees of freedom
+# 7: Number of degrees of freedom for all compositions
+# 8: Number of nonlinear iterations
+# 9: Iterations for temperature solver
+# 10: Iterations for composition solver 1
+# 11: Iterations for Stokes solver
+# 12: Velocity iterations in Stokes preconditioner
+# 13: Schur complement iterations in Stokes preconditioner
+0 0.000000000000e+00 0.000000000000e+00 512 4851 2145 2145 2  0 0  58  60  59 
+1 6.625491997652e+03 6.625491997652e+03 512 4851 2145 2145 8 83 0 338 346 346
```
gassmoeller, line 15: So this looks like the statistics file records all nonlinear iterations (2x4) and all Stokes iterations, etc. from all tries. I guess that is fair since we actually made all these iterations?

gassmoeller (commented):
Very nice, this will be useful for a bunch of models. Just some smallish comments and a question about the best way to react in the statistics postprocessor.

Also please add a changelog entry.

bangerth (commented):
Not getting to the rest of it fast enough, but here's a couple comments.

bangerth (dismissed):
I can't seem to get the uninterrupted time to finish this review, so here's what I have so far. I'll get back to the rest later.

tjhei:
Comments addressed. There are still many tests that need updating, though.

tjhei:
Something is still wrong with nonlinear problems:
```
WARNING: The nonlinear solver in the current timestep failed to converge.
Acting according to the parameter 'Nonlinear solver failure strategy'...
   Postprocessing:
     RMS, max velocity:                               7.49 m/s, 56.3 m/s
     Temperature min/avg/max:                         0 K, 0.5048 K, 1 K
     Heat fluxes through boundary parts:              0 W, 0 W, -3.127 W, 1 W
     Max, min, and rms velocity along boundary parts: 11.29 m/s, 0.001015 m/s, 4.795 m/s, 58.32 m/s, 0.001293 m/s, 23.75 m/s, 30.68 m/s, 0.02087 m/s, 11.42 m/s, 0.04261 m/s, 0.0008854 m/s, 0.03199 m/s

*** Timestep 2:  t=0.0021 seconds, dt=9.27806e-05 seconds
   Solving temperature system... 5 iterations.
   Initial Newton Stokes residual = 3.59864e+06, v = 3.59864e+06, p = 0

   Solving Stokes system... 14+0 iterations.
      Relative nonlinear residual (total Newton system) after nonlinear iteration 1: 1, norm of the rhs: 3.59864e+06

   Solving temperature system... 95 iterations.
[0b3d367e5bc0:34578] *** Process received signal ***
[0b3d367e5bc0:34578] Signal: Floating point exception (8)
[0b3d367e5bc0:34578] Signal code: Floating point divide-by-zero (3)
[0b3d367e5bc0:34578] Failing at address: 0x7fc1f8c5cb14
[0b3d367e5bc0:34578] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x14420)[0x7fc2048bc420]
[0b3d367e5bc0:34578] [ 1] ./libtosi_benchmark_newton_solver.debug.so(_ZNK6aspect13TosiBenchmark12TosiMaterialILi2EE9viscosityEddRKSt6vectorIdSaIdEERKN6dealii15SymmetricTensorILi2ELi2EdEERKNS8_5PointILi2EdEE+0x114)[0x7fc1f8c5cb14]
[0b3d367e5bc0:34578] [ 2] ./libtosi_benchmark_newton_solver.debug.so(_ZNK6aspect13TosiBenchmark12TosiMaterialILi2EE8evaluateERKNS_13MaterialModel19MaterialModelInputsILi2EEERNS3_20MaterialModelOutputsILi2EEE+0x2ff)[0x7fc1f8c5e24f]
[0b3d367e5bc0:34578] [ 3] /__w/aspect/aspect/build/aspect(_ZNK6aspect9SimulatorILi2EE31compute_pressure_scaling_factorEv+0x222)[0x563654f78982]
[0b3d367e5bc0:34578] [ 4] /__w/aspect/aspect/build/aspect(_ZN6aspect9SimulatorILi2EE22assemble_stokes_systemEv+0x101)[0x563655027a71]
[0b3d367e5bc0:34578] [ 5] /__w/aspect/aspect/build/aspect(_ZN6aspect9SimulatorILi2EE36do_one_defect_correction_Stokes_stepERNS_25DefectCorrectionResidualsEb+0x1fa)[0x563654e8b02a]
[0b3d367e5bc0:34578] [ 6] /__w/aspect/aspect/build/aspect(_ZN6aspect9SimulatorILi2EE42solve_iterated_advection_and_newton_stokesEv+0x195)[0x563654e8d835]
[0b3d367e5bc0:34578] [ 7] /__w/aspect/aspect/build/aspect(_ZN6aspect9SimulatorILi2EE14solve_timestepEv+0x1a5)[0x563655a14b25]
[0b3d367e5bc0:34578] [ 8] /__w/aspect/aspect/build/aspect(_ZN6aspect9SimulatorILi2EE3runEv+0x4ef)[0x563655a21ccf]
[0b3d367e5bc0:34578] [ 9] /__w/aspect/aspect/build/aspect(_Z13run_simulatorILi2EEvRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES7_bbbb+0x2cf)[0x5636559ef89f]
[0b3d367e5bc0:34578] [10] /__w/aspect/aspect/build/aspect(main+0x5ad)[0x5636549fb66d]
[0b3d367e5bc0:34578] [11] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7fc2046da083]
[0b3d367e5bc0:34578] [12] /__w/aspect/aspect/build/aspect(_start+0x2e)[0x5636549fd11e]
```

tjhei:
@gassmoeller @jdannberg can you remind me what the conclusion of our discussion regarding default of the new setting is?

gassmoeller:
I think we agreed on `continue_with_next_timestep`, because it is closest to the old behavior, and at least provides the user with a warning. I think we also talked about somehow tracking if there are any non-converging timesteps and then issue a warning at the end, because warnings for individual timesteps might get lost in the long output text.

naliboff:
@tjhei - I tested out these changes on top of #4370, and so far using `set Nonlinear solver failure strategy      = cut timestep size` is allowing a complex model to get through non convergent time steps after repeating the time step anywhere from 1-3 times. 

I will report back after additional testing, but so far good news!

jdannberg:
@tjhei From what I remember, this was almost finished, except for the default. Or was there still a problem? 
@naliboff found that this helps with convergence in lithosphere-scale models (and that the results are actually different), so it would be great to get this merged soon!

source/simulator/core.cc
```diff
@@ -1850,91 +1851,125 @@ namespace aspect
     if (parameters.use_operator_splitting)
       compute_reactions ();
 
-    switch (parameters.nonlinear_solver)
+    try
```
gassmoeller, line 1854: @naliboff tested this PR for a model, and we just realized that it looks like the mesh deformation above is not reset if the nonlinear solver fails. it would be nice to make the repeat work with mesh deformation as well, but at least we should probably document this. (also I havent tested this extensively yet, maybe it is reset in a place I dont see).

tjhei:
@jdannberg Do you remember what we decided regarding the default? I vaguely remember some kind of statistic to be printed at the end of the run as well? Sorry, I don't remember any of the details.

gassmoeller:
I think we decided to continue on non-convergence, but print a warning (and possibly another warning at the end of the model run if at least one timestep didnt converge). I dont remember about the statistics, maybe it would be useful to report how many timesteps didnt reach the nonlinear solver tolerance?

source/simulator/core.cc
tjhei: I thought this is done here:
https://github.com/geodynamics/aspect/blob/994f686e0d05ea995689c0f19600a11255850e37/source/simulator/core.cc#L2133-L2134

tests/global_melt/screen-output
```diff
@@ -95,3 +101,5 @@ Termination requested by criterion: end time
 
 
 
+
+WARNING: During this computation 2 nonlinear solver failures occurred!
```
tjhei, line 105: Here is an example of the new output.

tjhei:
Ok, I changed the default to continue and I added printing of information. Any other comments?

source/simulator/core.cc
naliboff: @gassmoeller @tjhei - Yes, the example in question was a version of the continental extension cookbook using #4370 and the commits from this PR. Indeed, in some time steps where the time step was repeated 4-5 times and the time size got quite small (10,000 years -> 500 years) the solver eventually converged, but the solution looked quite different even the nonlinear solve eventually converged. I ran the same model without mesh deformation and the same issue persisted, and I think it is due to the dependence of the rheology on the time step size rather than an issue with this PR. In general, the functionality here helped quite a bit when the final time step size only ends up being a factor of 2 or 4 lower than previous time steps.

naliboff:
> Ok, I changed the default to continue and I added printing of information. Any other comments?

Not for this PR, but based on my testing so far I think it would be good to implement options in the future where:
1. There is a maximum  number of repeated time steps (or minimum time step size) to prevent spurious behavior in models with viscoelasticity
2. An option to repeat the time step and increase the maximum number of nonlinear iterations. This is based on a few cases where the model would have converged with a few more iterations if the maximum number of nonlinear iterations had been set to a higher value. I'll think about ways that this could be automated.

Thanks again for this functionality @tjhei, after a bit of initial testing it also helping in models with two-phase flow and reactions that can lead to rapid changes in porosity.

source/simulator/core.cc
```diff
@@ -1889,91 +1891,127 @@ namespace aspect
...
+                  pcout << "Error: We can not cut the timestep in step 0, so we are aborting."
+                        << std::endl;
+                  throw;
+                }
+              time_stepping_manager.template get_matching_active_plugin<TimeStepping::RepeatOnNonlinearFail<dim>>().nonlinear_solver_has_failed();
+              break;
+            }
+            case Parameters<dim>::NonlinearSolverFailureStrategy::abort_program:
+            {
+              // rethrow the current exception
```
gassmoeller, line 2008: the other two options have an additional line that says what is happening next (continue, cut), but this one doesnt, it just throws the error. Can you add another line here, saying that we are now aborting the model run? This would explain to the user we purposefully abort (they would see that from the error message as well, but this makes it more obvious).

source/simulator/core.cc
```diff
@@ -2173,6 +2210,9 @@ namespace aspect
     // throwing an exception. Therefore, we have to do this manually here:
     computing_timer.print_summary ();
 
+    if (nonlinear_solver_failures>0)
```
gassmoeller, line 2213:
```suggestion
    if (nonlinear_solver_failures > 0)
```

source/time_stepping/repeat_on_nonlinear_fail.cc
```diff
@@ -0,0 +1,135 @@
...
+
+    template <int dim>
+    void
+    RepeatOnNonlinearFail<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Time stepping");
+      {
+        prm.enter_subsection("Repeat on nonlinear solver failure");
+
+        prm.declare_entry("Cut back amount", "0.5",
```
gassmoeller, line 92: Maybe and the same for the internal variable name:
```suggestion
        prm.declare_entry("Cut back factor", "0.5",
```

gassmoeller (commented):
Thank you! This is almost ready, could you look into my last remaining comments? I think all of @bangerth's comments have been addressed, unless we hear something more I am willing to merge this.

tjhei:
updated.

gassmoeller (approved):
:+1: Thanks.

PR #2745, opened 2018-12-19
remove some old-style C casts

tjhei:


source/termination_criteria/end_walltime.cc
```diff
@@ -35,7 +35,7 @@ namespace aspect
     bool
     EndWalltime<dim>::execute()
     {
-      return ( (std::time(nullptr)-start_walltime) >= (int)walltime_duration);
+      return std::difftime(std::time(nullptr),start_walltime) >= walltime_duration;
```
bangerth: nice find -- I didn't know about `std::difftime`!

PR #2746, opened 2018-12-20
www: add google analytics to citing.html

tjhei:


tjhei:
as discussed with @jdannberg and @gassmoeller so we can track the number of people visiting this URL.

PR #2747, opened 2018-12-23
Introduce parameters.xml

gassmoeller:
This change makes `doc/update_parameters.sh` also create a parameters.xml file that can be used for an extremely convenient view of all parameters (better interface than the long list in the manual, and easier to use than the parameter gui). Since it is included in the usual update cycle, it will always be in sync with the parameters.tex in the repo and the latest manual. The xml file is combined with an .xsl stylefile that allows viewing the file in a browser offline or online. For an online preview see: https://gassmoeller.github.io/parameters/parameters.xml, but the file also works offline (just open the .xml file in your browser, I think Chrome has a security feature that disallows the offline view, but firefox works). In a follow up PR I will link to this file from our website. This might be the optimal solution for tutorials, and for users who want to see what options there are.

There are numerous options for improvements, but for now I would prefer to just get this in and make adjustments later. I already always use this option when I am checking parameter names or default values.

### DIFF
--- a/doc/parameter_view/parameters.js
+++ b/doc/parameter_view/parameters.js
@@ -1,13 +1,34 @@
+/*
+  Copyright (C) 2018 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
 demangleStrings();
 
 function demangleStrings() {
-mangledStrings = document.getElementsByClassName("mangled");
-var j;
-for (j = 0; j < mangledStrings.length; j++) {
-mangledStrings[j].innerHTML = mangledStrings[j].innerHTML.replace(/_20/g," ");
-mangledStrings[j].innerHTML = mangledStrings[j].innerHTML.replace(/_2d/g,"-");
-}
-return;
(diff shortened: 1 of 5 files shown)


doc/update_parameters.sh
```diff
@@ -64,6 +64,27 @@ grep '[^\\]%' parameters.tex && echo "Error, please remove '%'!" && exit 1
 
 cd ../..
 
+
+# next, run ASPECT so that it produces the parameters.xml file that
+# documents all parameters and that we use for the web view of parameters
+echo Creating parameters.xml
+rm -f output/parameters.xml
+$ASPECT --output-xml doc/manual/empty.prm >doc/parameter_view/parameters.xml 2>/dev/null \
```
tjhei: why not pipe into .bak, then you can skip the ``cp`` below.

doc/update_parameters.sh
```diff
@@ -64,6 +64,27 @@ grep '[^\\]%' parameters.tex && echo "Error, please remove '%'!" && exit 1
...
+
+echo Patching parameters.xml
+cd doc/parameter_view
+cp parameters.xml parameters.bak
+head -n 1 parameters.bak > parameters.xml
+echo '<?xml-stylesheet type="application/xml" href="parameters.xsl"?>' >> parameters.xml
+echo '' >> parameters.xml
+tail -n +2 parameters.bak >> parameters.xml
+rm parameters.bak
+sed -i 's/\(.\)</\1\n</g' parameters.xml
```
tjhei: this will fail on OSX or does it? I forgot how to do the sed command on OSX.

doc/parameter_view/parameters.js
```diff
@@ -0,0 +1,27 @@
+demangleStrings();
+
+function demangleStrings() {
+mangledStrings = document.getElementsByClassName("mangled");
```
tjhei: indent?

doc/parameter_view/parameters.js
```diff
@@ -0,0 +1,27 @@
+demangleStrings();
```
tjhei: can you add a comment at the top with the ASPECT copyright?

gassmoeller:
I think I addressed all issues. It looks like Timo's tester has some issues (something about the mounting process of the disks), but I think it is unrelated to this PR. The CIG tester succeeds, and this PR should not change anything in the code anyway (only documentation).

PR #2751, opened 2019-01-06
Extend parameter view

gassmoeller:
This is a reopened #2749. I accidentally changed the base branch of that one, and github could not revert that. Description and commits just like in #2749.


tjhei:
Looks good!

PR #2755, opened 2019-01-08
Improve error message of ascii data plugin while loading file

### LINKED ISSUE #2753: misleading error message in ascii data plugin
When there are invalid entries in an ascii data file (in this case, this came up with the initial topography ascii data plugin), the error message still says

> The violated condition was: field_index == (components + dim) * data_table.n_elements()
>  Additional information: Number of read in points does not match number of expected points. File corrupted?

This can be reproduced by setting one of the numbers in `data/geometry-model/initial-topography-model/ascii-data/test/box_3d_top.0.txt` to NaN and then running the `box_initial_topography_ascii_data` test.

gassmoeller:
Fix #2753.


gassmoeller:
I simplified the logic and improved the error message. Is this more clear?

source/utilities.cc
```diff
@@ -1643,9 +1643,17 @@ namespace aspect
...
 
+      AssertThrow(in.eof(),
+                  ExcMessage (std::string("While reading the data file '" + filename + "' the ascii data "
+                                          "plugin has encountered an error before the end of the file. "
+                                          "Please check for malformed data values (e.g. NaN) or superfluous "
+                                          "lines at the end of the data file.")));
 
       AssertThrow(field_index == (components + dim) * data_table.n_elements(),
-                  ExcMessage (std::string("Number of read in points does not match number of expected points. File corrupted?")));
+                  ExcMessage (std::string("While reading the data file '" + filename + "' the ascii data "
```
tjhei: I think @jdannberg  wants this assert() condition (which appears in the output) to be easier to understand (right?). Maybe we can do 
```suggestion
      const n_total_values = (components + dim) * data_table.n_elements();
      AssertThrow(field_index == n_total_values,
```
and ``... expected number of floating point values based on the dimension, the number of components, and the number of lines given by POINTS in the header of the file``.

source/utilities.cc
```diff
@@ -1643,9 +1643,17 @@ namespace aspect
         }
       while (in >> temp_data);
 
+      AssertThrow(in.eof(),
+                  ExcMessage (std::string("While reading the data file '" + filename + "' the ascii data "
```
bangerth: no need for the explicit conversion to `std::string`. The addition of `"..." + std::string` already does that.

gassmoeller:
Is this better? I renamed the variables to make the condition easier to understand.

tjhei:
Looks good to me!

PR #2760, opened 2019-01-10
Link to the 2019 hackathon.

bangerth:


gassmoeller:
We should also announce it on the forum.

PR #2761, opened 2019-01-10
add assert and notes about inhomogeneous constraints

tjhei:
We use constraints.distribute_local_to_global() without a matrix in the
Stokes assembly. This gives incorrect results when we have inhomogeneous
constraints. Make sure that we never use this. Also add documentation
for this.

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -686,6 +686,15 @@ namespace aspect
     if (rebuild_stokes_matrix == true)
       system_matrix = 0;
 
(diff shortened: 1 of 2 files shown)


tjhei:
This came up after a long discussion with @tcclevenger . I hope that my assumption is correct. :)

tjhei:
> Did you run into any wrong results?

No, but I wasn't convinced just by reading the code. The assert does not trigger for me or for the tester so I think this is safe to merge.

gassmoeller:
Ok, I guess there is no harm in being too careful and the tester passes.

Issue #2762, opened 2019-01-11
Better document how to add publications using ASPECT to the list on our website


gassmoeller:
This is currently done using a Mendeley group https://www.mendeley.com/community/aspect-1/ which is exported and shown on our website. Document this process on our website https://aspect.geodynamics.org/publications.html, or find a simpler way.

tjhei:
We would need a volunteer to try to add to the group, who is not an admin. If this does work, we can document it that way. If not: we could just create a github issue where we ask people to post a comment for a new entry...

ljhwang:
I volunteer!  But  I think I will need further instructions and discussions with you.

tjhei:
> I will need further instructions

Go to https://www.mendeley.com/community/aspect-1/ and try to add some publication to the group. I assume you will need to sign up for a mendeley account as a first step. See if you can find a button to add an entry after clicking on "group documents".

tjhei:
Oh, it looks like this is an "invite only" group. Not sure how to change this...

tjhei:
closing this in place of #2831

PR #2763, opened 2019-01-11
update publications

tjhei:
(Sorry for the change in order of the entries. They should be sorted by "date added" from now on)

### DIFF
--- a/doc/manual/citing_aspect.bib
+++ b/doc/manual/citing_aspect.bib
@@ -222,6 +222,21 @@ url = {https://www.sciencedirect.com/science/article/pii/S0031920116301388},
 volume = {273},
 year = {2017}
 }
+@article{Bredow2018,
+author = {Bredow, Eva and Steinberger, Bernhard},
+doi = {10.1002/2017GL075822},
+journal = {Geophysical Research Letters},
+title = {{Variable melt production rate of the Kerguelen hotspot due to long-term plume-ridge interaction}},
+url = {http://onlinelibrary.wiley.com/doi/10.1002/2017GL075822/full},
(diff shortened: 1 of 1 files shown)


tjhei:
hang on, there are no new entries in there...
related to  #2762

tjhei:
updated. Check the second commit for the new entries. The first one only reorders them.

doc/manual/citing_aspect.bib
```diff
@@ -306,3 +321,67 @@ @techreport{Bangerth2017a
 url = {https://doi.org/10.6084/m9.figshare.4865333},
 year = {2017}
 }
+@article{Bredow2018a,
```
gassmoeller: I think I accidentally duplicated this entry in the Mendeley group and now it appears twice in the bib file. I removed the duplicate copy from Mendeley, could you do the same here? Sorry for that.

doc/manual/citing_aspect.bib
```diff
@@ -306,3 +321,67 @@ @techreport{Bangerth2017a
...
+year = {2018}
+}
+@article{Heron2018,
+author = {Heron, P J and Pysklywec, R N and Stephenson, R and van Hunen, J},
+journal = {Geology},
+publisher = {Geological Society of America},
+title = {{Deformation driven by deep and distant structures: Influence of a mantle lithosphere suture in the Ouachita orogeny, southeastern United States}},
+year = {2018}
+}
+@article{Bredow,
```
gassmoeller: This is actually a thesis not an article.

tjhei:
:+1:

PR #2764, opened 2019-01-11
reorder solver parameters by subsection

tjhei:
- reorder by prm subsection
- remove unused nonlinear_switch_tolerance (there is a variable with the
same name in NewtonHandler)


tjhei:
This has little to do with doxygen (but yes, we do). I was trying to get a sense of the solver parameters and got confused so I reordered by the prm sections...

PR #2770, opened 2019-01-15
udpate world builder to version 0.2.0-pre.

MFraters:
I have update the submodule, and make it run with the new version of the world builder. 

I have been thinking about issue #2613, and I think once the world builder is stable, we should just move all the files directly in there. But for now it updating it when issues arise is easier for me through submodules.


CMakeLists.txt
```diff
@@ -368,6 +368,8 @@ ELSE()
   MESSAGE("-- World Builder version ${WorldBuilderVersion} found.")
   SET(ASPECT_WITH_WORLD_BUILDER ON CACHE BOOL "If ON the World Builder is compiled togheter with ASPECT.")
 
+  configure_file("${CMAKE_SOURCE_DIR}/contrib/WorldBuilder/source/config.cc.in" "${CMAKE_SOURCE_DIR}/contrib/WorldBuilder/source/config.cc" @ONLY)
```
tjhei: how does this work? the .in file has a bunch of version strings that are not defined in here or does it?

MFraters: Yes, you where correct that it was not complete. Sorry for that. This was a quick fix to make it compile for someone who was planning to use aspect and the world builder. I now fully tested it and I managed to get it working on a complex 3d example.

tjhei:
This looks more reasonable. I wonder if you could move the long blob of cmake code into worldbuilder/ and do a ``INCLUDE(worldbuilder/version.cmake)`` here, instead. Would that work?

MFraters:
You mean that it would become part of the world builder repository?

tjhei:
Yes. You have that code sitting in your CMakeLists.txt already, right? You could just put it into a separate .cmake file that you include here and there.

MFraters:
Yes, that is a good solution. Fixed it now :)

PR #2774, opened 2019-01-17
disable FPEs while averaging

tjhei:
This disables FPEs while averaging quantities, which allows us to
average NaNs (values the user didn't fill) without crashing. This is
useful if the values are unused. Errors will still trigger on usage of
the averaged quantity. A good example is
output.entropy_derivative_pressure.

supersedes #2025

### DIFF
--- a/source/material_model/interface.cc
+++ b/source/material_model/interface.cc
@@ -32,6 +32,11 @@
 
 #include <list>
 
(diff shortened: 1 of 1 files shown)


gassmoeller:
Seems reasonable to me.

PR #2775, opened 2019-01-18
Fix latent_heat phase transitions loop

naliboff:
It looks like the iteration over the number of points in `latent_heat.cc` using iterator `i`:
https://github.com/geodynamics/aspect/blob/a19b94bc9ca75891c3bcead9a808ab5327f01eed/source/material_model/latent_heat.cc#L148

is unfortunately later overwritten by the first loop over the number of phase transitions:
https://github.com/geodynamics/aspect/blob/a19b94bc9ca75891c3bcead9a808ab5327f01eed/source/material_model/latent_heat.cc#L229

If `i` is indeed overwritten, the fix in this PR should break a number of tests.

### DIFF
--- a/source/material_model/latent_heat.cc
+++ b/source/material_model/latent_heat.cc
@@ -226,30 +226,30 @@ namespace aspect
             // entry for the first layer), so we have to use i+1 as index
             if (composition.size()==0)      // only one field
               {
-                for (unsigned int i=0; i<number_of_phase_transitions; ++i)
+                for (unsigned int phase=0; phase<number_of_phase_transitions; ++phase)
                   {
                     const double phaseFunction = phase_function (position,
                                                                  temperature,
                                                                  pressure,
-                                                                 i);
+                                                                 phase);
 
-                    phase_dependence += phaseFunction * density_jumps[i];
-                    viscosity_phase_dependence *= 1. + phaseFunction * (phase_prefactors[i+1]-1.);
+                    phase_dependence += phaseFunction * density_jumps[phase];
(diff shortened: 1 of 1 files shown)


source/material_model/latent_heat.cc
```diff
@@ -226,30 +226,30 @@ namespace aspect
...
                     const double phaseFunction = phase_function (position,
                                                                  temperature,
                                                                  pressure,
-                                                                 i);
-                    if (transition_phases[i] == 0)     // 1st compositional field
-                      phase_dependence += phaseFunction * density_jumps[i] * (1.0 - composition[0]);
+                                                                 phase);
+                    if (transition_phases[phase] == 0)     // 1st compositional field
+                      phase_dependence += phaseFunction * density_jumps[phase] * (1.0 - composition[0]);
                     else if (transition_phases[i] == 1) // 2nd compositional field
```
jdannberg: You forgot an _i_ here. This is why the test changes.

jdannberg:
Alright, I think I figured it out (and learned something new). 
This was actually not a bug before (just bad coding style). You can use a variable with the same name in nested for loops, and it will work because they're in different scopes. Somehow. But I agree that we should rename the `i` to `phase` inside of the loop. If you fix the last `i` that you missed, this shoulnd't actually change any tests.

tjhei:
> This was actually not a bug before (just bad coding style). You can use a variable with the same name in nested for loops, and it will work because they're in different scopes. Somehow.

Correct. The correct term for this is to "shadow": https://en.wikipedia.org/wiki/Variable_shadowing#C++

tjhei:
Regardless, this is cleaner than before. :)

PR #2777, opened 2019-01-21
world builder cmake cleanup

tjhei:
- use EXIST instead of ls
- upper case cmake functions


MFraters:
@tjhei Can this pull request be close so that I only have to look at #2778, or do you want to merge them separately?

PR #2778, opened 2019-01-21
world builder update 2

tjhei:
- only include files if enabled
- allow disabling after the fact
- put config file in binary dir instead

based on #2777.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,8 @@ MESSAGE(STATUS "====================================================")
 # Set the name of the project and target:
 SET(TARGET "aspect")
 
-FILE(GLOB_RECURSE TARGET_SRC  "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp" "contrib/WorldBuilder/source/*.cc")
-INCLUDE_DIRECTORIES(include ${CMAKE_BINARY_DIR}/include contrib/catch contrib/WorldBuilder/include/)
+FILE(GLOB_RECURSE TARGET_SRC  "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
+INCLUDE_DIRECTORIES(include ${CMAKE_BINARY_DIR}/include contrib/catch)
 
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.8)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -337,26 +337,27 @@ IF (ASPECT_PRECOMPILE_HEADERS)
...
 
-  configure_file("${WORLD_BUILDER_SOURCE_DIR}/source/config.cc.in" "${WORLD_BUILDER_SOURCE_DIR}/source/config.cc" @ONLY)
+  # add source and include dirs:
+  FILE(GLOB_RECURSE _files "contrib/WorldBuilder/source/*.cc")
+  LIST(APPEND TARGET_SRC ${_files})
+  INCLUDE_DIRECTORIES("${WORLD_BUILDER_SOURCE_DIR}/include/")
 
-  LIST(INSERT TARGET_SRC 0 "${WORLD_BUILDER_SOURCE_DIR}/source/config.cc")
+  # generate config.cc and include it:
+  CONFIGURE_FILE("${WORLD_BUILDER_SOURCE_DIR}/source/config.cc.in" "${CMAKE_BINARY_DIR}/world_builder_config.cc" @ONLY)
```
MFraters: This creates a problem for me (multiple definitions of version etc.), because I also use the world builder directory directly. Is it a problem for you to have the config file in "${WORLD_BUILDER_SOURCE_DIR}/source/config.cc"?

tjhei: You should not put temporary files into the source directory. If you have done so in the past, you will get linker errors (because you have two different .cc files) with this change. So I guess this means you also need to change the location in your CMakeLists.txt.
The reasoning is that an out-of-source build should not touch the source directory. Imagine you have two out-of-source builds. Now both will start overwriting or using the same file. Even if they will have the same content (for now at least unless other configuration options will be in that file!), configuring a new build directory will cause a rebuild of another build directory as the file is modified.
In summary: I think it is a bad design to modify/clutter the source directory.

MFraters: Hmm, that is a good point. I always viewed it as a source file, and that it therefore should be in the source directory. But you are right, that it makes more sense to put it in the build directory.

I have change the configuration on the and tested it and it seems to work fine now. Could you update the world builder, so that it links to a version which has separated out the config file? Otherwise it is good to merge.

tjhei:
updated.

Issue #2779, opened 2019-01-23
check handling or periodic and hanging node constraints


tjhei:
See discussion in https://github.com/dealii/dealii/pull/7487

Matthias says:
```
It is absolutely mandatory to set periodic boundary conditions first, and hanging node constraints second.
```
and our code in core.cc currently does:
```
    DoFTools::make_hanging_node_constraints (dof_handler,
                                             constraints);
    // Now set up the constraints for periodic boundary conditions
    // ...
```

gassmoeller:
Since #2780 is merged this is hopefully solved.

PR #2780, opened 2019-01-23
constraints: periodic before hanging

tjhei:
This PR is to check if #2779 makes a difference.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -1283,14 +1283,14 @@ namespace aspect
       free_surface->setup_dofs();
 
 
-    // reinit the constraints matrix and make hanging node constraints
+    // Reconstruct the constraint-matrix:
     constraints.clear();
     constraints.reinit(introspection.index_sets.system_relevant_set);
 
-    DoFTools::make_hanging_node_constraints (dof_handler,
-                                             constraints);
+    // Set up the constraints for periodic boundary conditions:
 
-    // Now set up the constraints for periodic boundary conditions
+    // Note: this has to happen _before_ we do hanging node constraints,
+    // because inconsistent contraints could be generated in parallel otherwise.
     {
       typedef std::set< std::pair< std::pair< types::boundary_id, types::boundary_id>, unsigned int> >
       periodic_boundary_set;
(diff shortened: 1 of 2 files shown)


gassmoeller:
Does not seem to make a difference. On the other hand: Do we have a test with hanging nodes along  periodic boundaries in parallel? Maybe it is just untested. I remember we had an issue with DG elements on parallel AMR periodic boundaries, which we disallowed in #2132. Maybe that is related?

tjhei:
> Maybe it is just untested

Likely. It also only happens rarely (specific parallel distribution needs to occur). I just wanted to see if it does make a difference by running the tester.

> I remember we had an issue with DG elements on parallel AMR periodic boundaries

Unlikely as you don't have hanging node constraints with DG.

gassmoeller:
So what do we do? Merge this? Add a test even if we do not know if it was broken before?

gassmoeller:
What is the status on this PR? Anything missing to merge it? Or has the topic been solved in deal.II, and this is a leftover?

gassmoeller:
Should this go into 2.1? We have not seen an effect, but it is a possible bug fix.

gassmoeller:
@tjhei: I will defer this to you. The change does not seem to affect anything, and if we consider this new version to be the correct one, then lets merge it for the release.

tjhei:
Gosh, sorry for ignoring your comment for the last 4 months. One problem is that I am not 100% if this will break anything. I don't think we have great test coverage for parallel, 3d problems. Maybe we should leave this out of 2.1? Not sure. @bangerth ?

gassmoeller:
No problem. Ok, lets leave it out of 2.1. However, then lets at least merge it now. Tests are passing, and if we find any problems with it until the release of 2.2 we have time to fix them.

tjhei:
Agreed!

bangerth:
Apologies for ignoring this for that long. The order only makes sense if:
* You have adaptive mesh refinement
* Some of the hanging nodes are located on the boundary
* Which can only happen in 3d.
In essence, you need to determine whether the value of a hanging node at the boundary is supposed to be 1/2 of the adjacent nodes, or equal to the correct boundary values. In the former case you end up with a continuous solution that violates the boundary conditions; in the latter the solution is discontinuous at refined faces but has the correct boundary values.

If you wanted to check this, you'll have to have a 3d geometry and boundary values that are not part of the solution space.

PR #835, opened 2016-04-29
Add VoF Interface tracking

class4kayaker:
This is an implementation of the VoF based interface tracking implemented in #823, modified to link more closely with the main model. Currently the majority of the logic is implemented in the Simulator class due to relative ease of initial implementation. Moving most of these into a separate encapsulating class is planned, though some modification of approach may be necessary. Some of the modifications may be dependent on whether and how the results of the implementation are desired by the main model.

Current features:
- Signed distance Level Set or composition function initialization
- Dimensionally split, explicit update based on computed velocity field (Standard for VoF based methods due to problems from double fluxing induced by geometric flux calculation)
- Creation of internal per cell LS field for the interface reconstruction, primarily useful for visualization of reconstructed interface (left over from considering storing the reconstructed interface as a LS, not currently used for computation)

Current limitations:
- Only functional for 2D
- Can only handle parallelogram meshes
- Cannot handle variations in refinement level

Planned improvements:
- Make interface LS data available to MaterialModel (as composition?)

Possible improvements:
- Couple to level set method
- Reduce computation to narrow band around interface (Probably not worth additional code complexity)

PR #2781, opened 2019-01-24
Remove indentation in some timing sections

gassmoeller:
As was pointed out in #835 some of ASPECT's timing sections have strange indentation in their names, likely because the names were copied from the screen output string close by that is indented. The majority of sections is not indented, and it is not visible anyway, because the `Timer` class does not print leading spaces in the timing output. Therefore this PR removes the indentation and unifies the naming of our timing sections.

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -426,7 +426,7 @@ namespace aspect
     if (parameters.use_direct_stokes_solver)
       return;
 
-    TimerOutput::Scope timer (computing_timer, "   Build Stokes preconditioner");
+    TimerOutput::Scope timer (computing_timer, "Build Stokes preconditioner");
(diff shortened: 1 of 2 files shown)


tjhei:
Why do you have changes in the worldbuilder submodule, did you forget to update it?

tjhei:
> class does not print leading spaces in the timing output

huh, interesting. I did not know that.

bangerth:
The lines in question were last changed in #2087. Before that, these lines appear to predate the git repository and I remember that leading spaces were printed and allowed us to show individual sections under (less indented) broader categories.

tjhei:
> under (less indented) broader categories

But Rene claims that whitespace is trimmed when showing?

gassmoeller:
Maybe the TimerOutput class changed later on? Either way, I think the indentation was kept internally in the name of the Timing section, but upon output leading spaces are trimmed.

PR #2783, opened 2019-01-25
Allow cell-wise constant composition element

gassmoeller:
I wanted to do some testing with a DGQ0 composition element and found that it was not allowed. Fixed this and added a test.


tjhei:
:+1:

PR #2784, opened 2019-01-25
Slightly improve accuracy of particle interpolation for active particles in time-varying flow.

gassmoeller:
This PR fixes a small problem with the particle interpolation. 

The same function was used to interpolate the particle properties for the initial condition (where solution, old_solution, and old_old_solution need to be initialized) and the current time step (where only solution should be interpolated), and it always interpolated the current and the two previous time steps. 

This is not a problem for the Stokes equation, because it only uses the current solution to evaluate material properties. But it is a small problem for the temperature equation, which uses the bdf2 extrapolation to determine material properties at the middle of the timestep. Therefore, the part of the material properties that depends on the composition would have looked like taken from a method that is only first-order accurate in time. 

In most models the part of the material property that is used in the temperature equation and depends on composition is small, I had trouble creating a test-case that is actually sensitive to this, but maybe in a model with composition-dependent viscosity and strong shear heating this might have reduced accuracy in the temperature slightly. Either way, it is fixed now.


source/simulator/initial_conditions.cc
```diff
@@ -303,8 +303,12 @@ namespace aspect
     // overwrite the relevant composition block only
     const unsigned int blockidx = advection_field.block_index(introspection);
     solution.block(blockidx) = particle_solution.block(blockidx);
-    old_solution.block(blockidx) = particle_solution.block(blockidx);
-    old_old_solution.block(blockidx) = particle_solution.block(blockidx);
+
+    if (timestep_number == 0)
```
naliboff: Perhaps write a comment saying that on time step 0 `solution`, `old_solution`, `old_old_solution` are by default the same, but we still need to fill the solution blocks with values.

tests/particle_interpolator_time_dependence.prm
```diff
@@ -0,0 +1,114 @@
...
+  end
+end
+
+
+# This is the new part: We declare that there will
+# be two compositional fields that will be
+# advected along. Their initial conditions are given by
+# a function that is one for the lowermost 0.2 height
+# units of the domain and zero otherwise in the first case,
+# and one in the top most 0.2 height units in the latter.
```
naliboff: Consider removing this comment, as the features below are not the "new part" of this test.

gassmoeller:
Thanks, I addressed your comments.

gassmoeller:
Ah yes, thanks. Fixed now.

For future reference: I now use `git config --global submodule.recurse true`, which automatically updates to the current submodule version whenever I switch branches.

PR #2785, opened 2019-01-27
Add assert when the variable temperature_for_viscosity becomes zero.

MFraters:
Prevents division by zero when viscosity_diffusion is computed.

### DIFF
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -105,6 +105,14 @@ namespace aspect
       // gradient used when calculating the viscosity. This allows the same activation volume
       // to be used in incompressible and compressible models.
       const double temperature_for_viscosity = temperature + adiabatic_temperature_gradient_for_viscosity*pressure;
+      Assert(temperature_for_viscosity != 0, ExcMessage(
+               "The temperature used in the calculation of the visco platic rheology is zero. "
+               "This is not allowed, because this value is used to divide through. It is probably "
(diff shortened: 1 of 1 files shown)


source/material_model/visco_plastic.cc
```diff
@@ -105,6 +105,14 @@ namespace aspect
       // gradient used when calculating the viscosity. This allows the same activation volume
       // to be used in incompressible and compressible models.
       const double temperature_for_viscosity = temperature + adiabatic_temperature_gradient_for_viscosity*pressure;
+      Assert(temperature_for_viscosity != 0, ExcMessage(
+               "The temperature used in the caluclation of the visco platic rheology is zero. "
```
tjhei: typo "calculation"

tjhei:
Is this for a situation where one incorrectly sets up her problem and you really get precisely 0.0 here? Otherwise a check with ``(abs(x)<eps)`` might be more appropriate?

MFraters:
In the case I encountered this, the temperature and the adiabatic_temperature_gradient_for_viscosity where exactly zero, which led to an SIGFPE which was quite hard to track. So I made this mainly with the intention to prevent other people from wasting time on tracking the floating point exception. I think that is it is not exactly 0.0, there won't be an SIGFPE, although other problems may arise of course. So I am not sure which one is more appropriate.

tjhei:
Well, this is better than nothing and probably not worth worrying about.

PR #2787, opened 2019-01-28
Add a higher resolution heat flux visualization postprocessor

gassmoeller:
This adds a new visualization postprocessor called `heat flux density` that is similar to the existing `heat flux map` except it does not output heat flux as cell-wise constant, but computes it at the patch vertices. I wrote this for an application of @ebredow, and the effect in output resolution is pretty staggering. Below is a comparison of the S20RTS cookbook with the two postprocessors for the same resolution:

s20rts_heat_flux_comparison

I would like to replace `heat flux map` with this new postprocessor, but unfortunately it does not support Neumann boundary conditions, or non-tangential velocity boundary conditions at the moment. The problem is that `compute_heat_flux_through_boundary_faces` is currently not prepared for both the statistics postprocessors, which require accurate heat flux values per face, and the visualization postprocessors, which only allow a single value per vertex (i.e. at edges between faces you have to decide on one value). It is also not possible to have `heat flux map` decide on different techniques based on the type of boundary, because depending on the decision it needs to be derived from two different classes.

Since I do not have the time to rewrite the interface, and I do not want to duplicate the code again (which I just unified), I propose to keep this as a separate postprocessor with a clear description that the new one is superior unless one has heat flux boundary conditions, or non-tangential velocities.

If someone is interested in making this generally applicable and replace `heat flux map` I am more than happy to provide guidance, I just do not have the time myself at the moment. Also see my related issue #2788 about dynamic topography and geoid for which the same apply.


gassmoeller:
I found a way to combine the new and old approach into a single postprocessor. For that I converted the existing 'heat flux map' postprocessor into a `DataPostprocessorScalar` and if it should write cell data, it simply outputs the same cell-wise constant value into all evaluation points. This way I can make the new higher resolution output an input parameter. Once we have figured out how in a future PR how to use the new process for all models, we can change the default of the parameter or even drop the option altogether. This should now be ready for review. It essentially consists of two parts:
1. The refactoring in source/postprocess/heat_flux_map.cc extracts the CBF computation from the rest of the heat flux computation into its own function. Functionality remains unchanged, because the existing function now calls the new function. Since all tests that use any heat flux postprocessor are unchanged I assume this is correct.
2. The added option to source/postprocess/visualization/heat_flux_map.cc to use `compute_dirichlet_boundary_heat_flux_solution_vector` instead of `compute_heat_flux_through_boundary_faces` to compute the heat flux and therefore output in higher resolution. I added a test, and the result can also be seen in the screenshot above.

gassmoeller:
Ping. 
The review should not be too complicated (the code change is around 300 lines, mostly moving things and boiler plate. The changes in `source/postprocess/heat_flux_map` only separate the part of the code that is relevant for fixed temperature (Dirichlet) boundaries from the rest, and the changes in `source/postprocess/visualization/heat_flux_map` use this now separated part to change the postprocessor from a CellDataVectorCreator postprocessor to a DataPostprocessorScalar. 

@ebredow is already actively using this in a hacked version and it would be great to get it into master.

include/aspect/postprocess/visualization/heat_flux_map.h
```diff
@@ -33,22 +33,49 @@ namespace aspect
...
       {
         public:
+          HeatFluxMap();
+
+          /**
+           * @copydoc Interface<dim>::update()
+           */
+          void update();
+
+          virtual
```
naliboff: Please add a comment saying what this function does.

include/aspect/postprocess/visualization/heat_flux_map.h
```diff
@@ -33,22 +33,49 @@ namespace aspect
...
-          execute () const;
+          void
+          parse_parameters (ParameterHandler &prm);
+
+        private:
+          bool output_point_wise_heat_flux;
+
+          LinearAlgebra::BlockVector heat_flux_density_solution;
+
+          std::vector<std::vector<std::pair<double, double> > > heat_flux_and_area;
```
naliboff: Please provide a brief description of the two variables above.

tests/heat_flux_map_vis_point_wise.prm
```diff
@@ -0,0 +1,106 @@
+# This is a copy of the heat_flux_map_vis postprocessor that
```
naliboff: "postprocessor" --> "test"

naliboff: Rather than copy most of the parameter file over, this seems like a good case to derive from the existing `heat_flux_map_vis` test.

gassmoeller:
Thanks for the review! I addressed your comments. Ready from my side.

naliboff:
Looks good to me :+1:  @MFraters - did you want to take another look?

PR #2789, opened 2019-01-28
"no Advection, no Stokes" solver scheme

LudovicJnnt:
I added a solver_scheme that I called "no Advection, no Stokes" to be consistent with other solver schemes (although jperryhouts in the ASPECT forum suggested to call it simply "none" ?).

The idea is that the user can directly go to post-processing after setting up the initial condition, which save a lot of time if not required. For instance it can be used to post-process a model based on its most recent checkpoint.

In solver_scheme.cc, the section regarding this new solver_scheme is sort of *too* empty, but it works for the purpose I want to use it. Please let me know if assert-throws or more conditions are required in this section. 

For the test case, I reuse the initial_temperature_profile test. This was to be sure the initial conditions were properly set.


source/simulator/solver_schemes.cc
```diff
@@ -906,6 +906,19 @@ namespace aspect
...
     return;
   }
+
+
+
+  template <int dim>
+  void Simulator<dim>::solve_no_advection_no_stokes ()
+  {
+    // This section is now empty but I guess I need some assertThrows?
+    // NOT SURE IF THIS IS NEEDED:
```
gassmoeller: Currently, I can not think of AssertThrows that are necessary. You would not get useful results with this scheme if you set end time to something else but 0, but who knows maybe someone has an application even for that, and I do not think it breaks anything.

And you need the line about the postprocessing, otherwise you will get no output if you set the `Run postprocessors on nonlinear iterations` parameter (if that is true, the solver scheme is responsible for calling the postprocessors).
Just remove the comment in capitals and this should be fine.

LudovicJnnt: Ok thanks. That is done. 

Procedural question: bellow saying the tester failed means the tester actually ran but failed or do you still have to manually ask the tester to run when all is ready?

gassmoeller:
If you click on details you will see that the tester ran, but your new test failed. See manual section 6.8.2 on how to fix that. Let us know if the section is not clear enough.

gassmoeller:
Tester looks good now. Are you done with the adjustments?

LudovicJnnt:
All good, I am done with it! Thanks

gassmoeller:
Ready to merge then. Thanks!

PR #2790, opened 2019-01-30
allows 2d sinker

tcclevenger:
I edited the nsinker.cc file to only use the first two coordinates from the center points when running in 2d. Previously not implemented.


gassmoeller:
Looks good to me. Could you add a testcase for this to make sure we do not break it in the future? Just copy tests/nsinker* to tests/nsinker_2d* and modify to use the 2D case. Then update the test results as described in manual section 6.8.2.

benchmarks/nsinker/nsinker.cc
```diff
@@ -158,33 +163,27 @@ namespace aspect
...
-      double delta = 200.0;
-      double omega = 0.1;
-      double chi = 1.0;
-
-      // Centers for the sinkers provided by Cedric Thielot (pers. comm. from Dave May)
-      std::vector<Point<dim> > centers(76);
+      delta = 200.0;
+      omega = 0.1;
 
+      centers.resize(76);
```
tjhei: 75?

tcclevenger: right, typo.

benchmarks/nsinker/nsinker_2d.prm
```diff
@@ -0,0 +1,86 @@
+# This is the "NSinker" benchmark defined in May et al. (2015,G3) in the
```
tjhei: change this comment to just say that this is a 2d version of the originally 3d problem

tcclevenger: done.

Issue #2788, opened 2019-01-28
Output dynamic topography on vertices instead of cell-wise averaged


gassmoeller:
Since we use the CBF method for computing dynamic topography there is really no reason to keep outputting the dynamic topography (or potentially also the geoid) as a cell-wise constant value. Instead we can use the solution vector to output it point-wise, just as I did with the heat flux in #2787. This should drastically increase the resolution of the output with no additional cost.

This might be of interest to anyone especially interested in dynamic topography or the geoid, maybe @jaustermann, @SophieCoulson, @mnkajan? I will likely not get to this for a while (maybe during the hackathon), but let me know if you want to give it a try and need pointers.

gassmoeller:
Closed by #2791.

PR #2791, opened 2019-01-30
Output dynamic topography as point-wise vector

gassmoeller:
This addresses the dynamic topography part of #2788 and was really simpler than I expected. Even without using interpolated visualization output the effect is nice (side-by-side comparison of the S20RTS cookbook with and without the new output):

s20rts_dynamic_topo_comparison


include/aspect/postprocess/visualization/dynamic_topography.h
```diff
@@ -44,10 +44,13 @@ namespace aspect
        */
```
MFraters: Please update the class description

include/aspect/postprocess/visualization/dynamic_topography.h
```diff
@@ -66,8 +69,9 @@ namespace aspect
            * to.
```
MFraters: Same here.

source/postprocess/visualization/dynamic_topography.cc
```diff
@@ -28,19 +28,60 @@ namespace aspect
...
+
         const Postprocess::DynamicTopography<dim> &dynamic_topography =
           this->get_postprocess_manager().template get_matching_postprocessor<Postprocess::DynamicTopography<dim> >();
 
-        std::pair<std::string, Vector<float> *>
-        return_value ("dynamic_topography",
-                      new Vector<float>(dynamic_topography.cellwise_topography()));
+        auto cell = input_data.template get_cell<DoFHandler<dim> >();
+
+        bool cell_at_top_or_bottom_boundary = false;
```
MFraters: Maybe the code could use some more comment to make reading it easier. You could for example put a comment here like: "We are only interested in the faces at the top and bottom of the domain, so we check that here."

source/postprocess/visualization/dynamic_topography.cc
```diff
@@ -28,19 +28,60 @@ namespace aspect
...
+        for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
+          if (cell->at_boundary(f) &&
+              (this->get_geometry_model().translate_id_to_symbol_name (cell->face(f)->boundary_id()) == "top" ||
+               this->get_geometry_model().translate_id_to_symbol_name (cell->face(f)->boundary_id()) == "bottom"))
+            cell_at_top_or_bottom_boundary = true;
+
+        if (cell_at_top_or_bottom_boundary)
+          {
+            std::vector<Point<dim>> quadrature_points(input_data.evaluation_points.size());
```
MFraters: And here something like: "This cell is at the top or bottom, now compute the values of the dynamic topography in this cell." 
It might be just my personal style, but I like very verbose comments. If you feel it clutters the code too much, feel free to leave it out.

source/postprocess/visualization/dynamic_topography.cc
```diff
@@ -28,19 +28,60 @@ namespace aspect
...
-        return return_value;
+            FEValues<dim> fe_volume_values (this->get_mapping(),
+                                            this->get_fe(),
+                                            quadrature_formula,
+                                            update_values);
+
+            fe_volume_values.reinit(cell);
+
+            std::vector<double> dynamic_topography_values(quadrature_formula.size());
+            fe_volume_values[this->introspection().extractors.temperature].get_function_values(dynamic_topography.topography_vector(),
```
naliboff: Might be worth putting a short comment here explaining why one needs to use the temperature quadrature.

gassmoeller:
I addressed your comments.

PR #2793, opened 2019-01-31
memory statistics: add VmPeak

tjhei:


tjhei:
I will need to update test output...

tjhei:
well, except that this value is not going to be stable between runs. :-(

tjhei:
@bangerth I don't know what to do about this. This is useful information, but it is not reliable between runs or different machines. The options are: disable vmpeak by default (annoying), not merge, or add a postprocess script for the unit test that deletes the vmpeak value. I think the last option is the easiest, but I haven't found time to do it...

bangerth:
Let the same script do it that also deletes the timing information from the output files. It should be a one-line addition to some perl something.

tjhei:
I hacked something together, but it is a bit more complicated than a "one-line addition".

tjhei:
how about this?

source/postprocess/memory_statistics.cc
```diff
@@ -44,17 +44,60 @@ namespace aspect
...
+
+    template <int dim>
+    void
+    MemoryStatistics<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Postprocess");
+      {
+        prm.enter_subsection("Memory statistics");
+        {
+          prm.declare_entry ("output virtual memory peak", "true",
```
gassmoeller: Capitalize the input parameter

source/postprocess/memory_statistics.cc
```diff
@@ -44,17 +44,60 @@ namespace aspect
...
+
+    template <int dim>
+    void
+    MemoryStatistics<dim>::parse_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Postprocess");
+      {
+        prm.enter_subsection("Memory statistics");
+        {
+          output_vmpeak = prm.get_bool ("output virtual memory peak");
```
gassmoeller: same here

source/postprocess/memory_statistics.cc
bangerth: Close -- we usually start parameter names with upper case letters.

bangerth: Matter of fact, I think the proper name should be "Output peak virtual memory". Or maybe "Output peak virtual memory (VmPeak)".

gassmoeller:
This must be the first simultaneous double review we had in a while, and we even had the same comment :smile:

tjhei:
I decided that it is much more useful to do a ``max`` over all MPI ranks.

Issue #2794, opened 2019-02-01
Using deal.II dev crashes when loading any shared libraries


gassmoeller:
I am in the process of fixing #2602 by rebasing and fixing #1957. I found a number of small things, but I have a more severe problem that I could not debug in several hours today. I will take another look tomorrow or next week, but maybe @tjhei or @bangerth immediately see the problem, so I thought I could ask.

With the recent deal.II master and my rebased ASPECT branch that fixes #2602 (and works normal models) all models loading shared libraries into ASPECT fail immediately (when the library is loaded) with 
```
Loading shared library <./libsimple_nonlinear.so>
malloc(): memory corruption
[rene-laptop:27704] *** Process received signal ***
[rene-laptop:27704] Signal: Aborted (6)
[rene-laptop:27704] Signal code:  (-6)
```

gdb does not give helpful output, and when I run valgrind/memcheck for this model I get the following output (shortened, the part I left out is repeating the error message many times for different finite element classes):

```
==14966== Memcheck, a memory error detector
==14966== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==14966== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==14966== Command: ../aspect simple_nonlinear.x.prm
==14966== 
==14966== Warning: set address range perms: large range [0x4e3c000, 0x1ccb4000) (defined)
--14966-- warning: DiCfSI 0xe64e800 .. 0xed4011f is huge; length = 7280928 (libdeal_II.g.so.9.1.0-pre)
==14966== Conditional jump or move depends on uninitialised value(s)
==14966==    at 0x25568375: opal_value_unload (in /usr/lib/x86_64-linux-gnu/openmpi/lib/libopen-pal.so.20.10.1)
==14966==    by 0x1DED897A: ompi_proc_complete_init (in /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so.20.10.1)
==14966==    by 0x1DEDC8A4: ompi_mpi_init (in /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so.20.10.1)
==14966==    by 0x1DEFD404: PMPI_Init_thread (in /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so.20.10.1)
[...]

PR #2796, opened 2019-02-01
Make sure tests find ASPECT from build directory

gassmoeller:
This fixes an issue I had with the test suite as mentioned in #2794. If I have an environment variable $Aspect_DIR that points to aspect_build_folder_1 (linking to deal.II version 1), but create an aspect_build_folder_2 (linking to deal.II version 2) and try to run the test suite in that second build folder the test suite would still link to the deal.II version that was given from aspect_build_folder_1. I suspect that
```
FIND_PACKAGE(Aspect 2.1.0 QUIET HINTS ${ASPECT_DIR})
```
simply looks first for $Aspect_DIR (pointing to the wrong aspect), before it uses the hint to $ASPECT_DIR (pointing to the right aspect). I use cmake version 3.10.2.

I am unsure about our use of $ASPECT_DIR vs $Aspect_DIR, I hope I did the right thing by replacing $ASPECT_DIR by $Aspect_DIR, and adjusting the hints to the $FIND_PACKAGE call to the same hints we use in doc/plugin-CMakeLists.txt (this fixes my problem at least). If there is a better solution let me know.


tjhei:
This is a super weird situation. I am surprised that FIND_PACKAGE would check an ENV variable even though we don't tell it to do so.

tests/CMakeLists.txt
```diff
@@ -6,10 +6,10 @@ IF (CMP0058)
 ENDIF()
 
 SET(ASPECT_BINARY "${ASPECT_BINARY}" CACHE STRING "" FORCE)
-SET(ASPECT_DIR "${ASPECT_DIR}" CACHE STRING "" FORCE)
+SET(Aspect_DIR "${Aspect_DIR}" CACHE STRING "" FORCE)
 SET(ASPECT_COMPARE_TEST_RESULTS "${ASPECT_COMPARE_TEST_RESULTS}" CACHE BOOL "" FORCE)
 
-FIND_PACKAGE(Aspect 2.1.0 QUIET HINTS ${ASPECT_DIR})
+FIND_PACKAGE(Aspect 2.1.0 QUIET HINTS ${Aspect_DIR} ../ ../../ $ENV{ASPECT_DIR})
```
tjhei: I don't think the additional hints are necessary or particularly helpful here?

gassmoeller:
I have removed the additional hints, they are not necessary to fix my problem. What do you want me to do about the rest? Is there a reason the variable handed over to cmake when setting up tests has to be called ASPECT_DIR instead of Aspect_DIR? Otherwise I think there is no harm in renaming, in particular if it fixes the issue I observe (without this change, if I run `ccmake .` in the tests directory, I see a `ASPECT_DIR` variable and a `Aspect_DIR` variable and they can be different as in my case).

tjhei:
> up tests has to be called ASPECT_DIR instead of Aspect_DIR?

No, this is not the problem. My understanding is that FIND_PACKAGE will automatically check the environment variable with the same name. In your case, this would happen before the hint (ASPECT_DIR) would be checked.

PR #2800, opened 2019-02-04
Change make debug/release behavior

tjhei:
inspired from the mailing list (make -j 4 release) doesn't work
correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,17 +263,72 @@ ADD_CUSTOM_TARGET(indent
   COMMENT "Indenting all ASPECT header and source files..."
   )
 
+IF(CMAKE_GENERATOR MATCHES "Ninja")
+  SET(_make_command "$ ninja")
+ELSE()
+  SET(_make_command " $ make")
+ENDIF()
+
 # Provide "release" and "debug" targets to switch compile mode
+IF(${DEAL_II_BUILD_TYPE} MATCHES "DebugRelease")
 ADD_CUSTOM_TARGET(release
   COMMAND ${CMAKE_COMMAND} -D CMAKE_BUILD_TYPE=Release .
+  COMMAND ${CMAKE_COMMAND} -E echo "***"
+  COMMAND ${CMAKE_COMMAND} -E echo "*** Switched to Release mode. Now recompile with: ${_make_command}"
+  COMMAND ${CMAKE_COMMAND} -E echo "***"
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  VERBATIM
   COMMENT "switching to RELEASE mode..."
   )
 ADD_CUSTOM_TARGET(debug
   COMMAND ${CMAKE_COMMAND} -D CMAKE_BUILD_TYPE=Debug .
+  COMMAND ${CMAKE_COMMAND} -E echo "***"
+  COMMAND ${CMAKE_COMMAND} -E echo "*** Switched to Debug mode. Now recompile with: ${_make_command}"
+  COMMAND ${CMAKE_COMMAND} -E echo "***"
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  VERBATIM
   COMMENT "switching to DEBUG mode..."
   )
+ENDIF()
+
+# Provide a "distclean" target (like it is done in deal.II):
+ADD_CUSTOM_TARGET(distclean
+  COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --target clean
+  COMMAND ${CMAKE_COMMAND} -E remove_directory CMakeFiles
+  COMMAND ${CMAKE_COMMAND} -E remove
+    CMakeCache.txt cmake_install.cmake Makefile
+    build.ninja rules.ninja .ninja_deps .ninja_log
+  COMMENT "distclean invoked"
+  )
+
+
+
+FILE(WRITE ${CMAKE_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/print_usage.cmake
+"MESSAGE(
+\"###
+#
+#  Project ${TARGET} set up with  ${DEAL_II_PACKAGE_NAME}-${DEAL_II_PACKAGE_VERSION}  found at
+#      ${DEAL_II_PATH}
+#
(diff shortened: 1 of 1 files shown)


tjhei:
I just realized we can rewrite ``make release`` and ``make debug`` as well, so please wait with merging...

tjhei:
So, I guess we can decide if we want to add two new targets that only switch or if we want to change the semantics of ``release`` and ``debug`` (commit 2). Thoughts?

CMakeLists.txt
```diff
@@ -263,6 +263,17 @@ ADD_CUSTOM_TARGET(indent
...
+# Provide "release" and "debug" targets to switch compile mode
+ADD_CUSTOM_TARGET(release
+  COMMAND ${CMAKE_COMMAND} -D CMAKE_BUILD_TYPE=Release .
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  COMMENT "switching to RELEASE mode..."
+  )
+ADD_CUSTOM_TARGET(debug
+  COMMAND ${CMAKE_COMMAND} -D CMAKE_BUILD_TYPE=Debug .
+  WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+  COMMENT "switching to RELEASE mode..."
```
tjesser-ucdavis-edu: The comment should probably be "switching to DEBUG mode..."

gassmoeller:
I like the second option more, if you modify the comment to let the user know that they still need to recompile. E.g. "Switching to ... mode ... Remember to recompile before running ASPECT."

CMakeLists.txt
```diff
@@ -364,7 +375,8 @@ IF (ASPECT_WITH_WORLD_BUILDER)
   ENDFOREACH()
 ENDIF()
 
-DEAL_II_INVOKE_AUTOPILOT()
+ADD_EXECUTABLE(${TARGET} ${TARGET_SRC})
+DEAL_II_SETUP_TARGET(${TARGET})
```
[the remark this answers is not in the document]
tjhei: autopilot creates the following:
```
#  You can now run                                                                                                                          
#      $ ninja                - to compile and link the program                                                                             
#      $ ninja run            - to (compile, link and) run the program                                                                      
#                                                                                                                                           
#      $ ninja debug          - to switch the build type to 'Debug'                                                                         
#      $ ninja release        - to switch the build type to 'Release'                                                                       
#                                                                                                                                           
#      $ ninja strip_comments - to strip the source files in this                                                                           
#                               directory off the documentation comments                                                                    
#      $ ninja clean          - to remove the generated executable as well as                                                               
#                               all intermediate compilation files                                                                          
#      $ ninja runclean       - to remove all output generated by the program                                                               
#      $ ninja distclean      - to clean the directory from _all_ generated                                                                 
#                               files (includes clean, runclean and the removal                                                             
#                               of the generated build system)                                                                              
#      $ ninja info           - to view this message again 
```
I think the only useful thing is ``make info``, which we could write ourselves.

> And why is this not how autopilot sets up these targets?

The autopilot targets switch the mode _and_ compile. As deal.II tutorials do not need parallel compilation, it never was an issue there.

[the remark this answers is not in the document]
tjhei: also see https://github.com/dealii/dealii/blob/master/cmake/macros/macro_deal_ii_invoke_autopilot.cmake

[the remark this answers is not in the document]
bangerth: I think `make clean` and `make distclean` are also useful. I use them frequently.

Does cmake have a way of "overwriting" an existing target? Then we could let the autopilot do its work and then just re-define the `debug` and `release` targets.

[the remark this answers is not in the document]
tjhei: Not that I know, no.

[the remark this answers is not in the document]
tjhei: ``clean`` is not made by the autopilot and I can easily add the distclean if we want.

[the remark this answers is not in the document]
bangerth: I'm ok with it then if you add `distclean`.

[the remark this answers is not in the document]
tjhei: done.

CMakeLists.txt
```diff
@@ -263,6 +263,72 @@ ADD_CUSTOM_TARGET(indent
   COMMENT "Indenting all ASPECT header and source files..."
   )
 
+IF(CMAKE_GENERATOR MATCHES "Ninja")
+  SET(_make_command "$ ninja")
+ELSE()
+  SET(_make_command " $ make")
```
gassmoeller:
```suggestion
  SET(_make_command "$ make")
```

CMakeLists.txt
```diff
@@ -263,6 +263,72 @@ ADD_CUSTOM_TARGET(indent
...
+  )
+ENDIF()
+
+# Provide a "distclean" target (like it is done in deal.II):
+ADD_CUSTOM_TARGET(distclean
+  COMMAND ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --target clean
+  COMMAND ${CMAKE_COMMAND} -E remove_directory CMakeFiles
+  COMMAND ${CMAKE_COMMAND} -E remove
+    CMakeCache.txt cmake_install.cmake Makefile
+    build.ninja rules.ninja .ninja_deps .ninja_log
```
gassmoeller: What happens if these files are not there?

tjhei: it works and does not show any errors.

CMakeLists.txt
tjhei: it is made such that both commands have the same width (so that the info message is aligned nicely without a double space between ``make`` and ``release`` for example).

bangerth: isn't there a cmake variable that spells out the name of the generator? I thought @tamiko used it in the deal.II patch...

bangerth: I'm saying this because there may be other generators we could use, e.g. the MS VC internal one at one point.

tjhei: This is copied verbatim from deal.II:
https://github.com/dealii/dealii/blob/d9bf195abe45827205522a31094d814ad416c4fd/cmake/macros/macro_deal_ii_invoke_autopilot.cmake#L41

tjhei:
is this good to be merged or anything I can improve, @bangerth ?

gassmoeller:
Looks good to me, we can improve the process later if necessary.

PR #2801, opened 2019-02-05
Provide a version of std_cxx14::make_unique() that takes arguments.

bangerth:
We don't need this if the compiler supports C++14, as apparently most used by the developers
do. But when using a compiler that only supports C++11, we need to provide a version of
std_cxx14::make_unique() that accepts a variable number of arguments.

I have no way of testing this on any of my machines, and neither is this going to be
tested by the CI machines. I am pretty sure that this should work, though.

### DIFF
--- a/include/aspect/compat.h
+++ b/include/aspect/compat.h
@@ -152,16 +152,17 @@ namespace dealii
 #ifdef DEAL_II_WITH_CXX14
     using std::make_unique;
 #else
-    /* This is a simplified form of std::make_unique that only allows the default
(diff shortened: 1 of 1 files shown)


tjhei:
I will try this out here. Please wait with merging.

tjhei:
works :+1:

bangerth:
Nice to know. I had suspected the day would arrive one day sooner or later when we would need that...

bangerth:
Part of #2375.

PR #2802, opened 2019-02-05
fix several clang warnings

tjhei:
Interesting how often we ``get_double`` even though we want to do ``get_integer``...


gassmoeller:
Looks reasonable.

PR #2807, opened 2019-02-12
Use std::unique_ptr instead of std::shared_ptr.

bangerth:
I suspect that this is a left-over from the time when we used boost's shared_ptr
but there was no equivalent for shared_ptr.


bangerth:
Part of #2375.

PR #2809, opened 2019-02-13
Fix formatting problems in the documentation.

bangerth:
While there, also don't copy things literally, but use doxygen facilities.


tjhei:
Great. I am a big fan of not duplicating documentation!

PR #2810, opened 2019-02-13
Avoid bar 'new' operations in more places.

bangerth:


bangerth:
Yes, I suppose I could do that. Let's hold off on this then for the moment.

bangerth:
Actually, let me change my mind on this: I'd like to merge this as is because it should be a no-op at the moment. I will come back to changing `unique_ptr`s to `shared_ptr`s at some later time once I've gotten rid of all of the bare `new` operations. The change you ask for is not a no-op, and so I'd like to do that in a separate step -- will open a separate issue for that.

bangerth:
Part of #2375.

PR #2813, opened 2019-02-14
Fix a memory leak.

bangerth:
While there, also clean up around it.

### DIFF
--- a/source/boundary_temperature/dynamic_core.cc
+++ b/source/boundary_temperature/dynamic_core.cc
@@ -303,21 +303,24 @@ namespace aspect
     {
       data_OES.clear();
       if (name_OES.size()==0) return;
-      std::istringstream in(Utilities::read_and_distribute_file_content(name_OES.c_str(), this->get_mpi_communicator()));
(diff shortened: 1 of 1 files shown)


bangerth:
Part of #2375.

source/boundary_temperature/dynamic_core.cc
```diff
@@ -303,21 +303,24 @@ namespace aspect
...
       data_OES.clear();
       if (name_OES.size()==0) return;
-      std::istringstream in(Utilities::read_and_distribute_file_content(name_OES.c_str(), this->get_mpi_communicator()));
+      std::istringstream in(Utilities::read_and_distribute_file_content(name_OES.c_str(),
+                                                                        this->get_mpi_communicator()));
       if (in.good())
         {
           str_data_OES data_read;
-          const int buff_size = 1024;
-          char *line = new char [buff_size];
```
tjhei: important to do a dynamic allocation here. ;-)

tjhei:
:+1:

source/boundary_temperature/dynamic_core.cc
bangerth: Yes, one could have done without that here :-)

Issue #2811, opened 2019-02-14
Change shared_ptr to unique_ptr


bangerth:
We have `std::shared_ptr` objects primarily because there was a boost version we used for that a long time ago. But in reality, we probably want `std::unique_ptr` in most places -- but there was no corresponding boost version for that class. Go through all of the places where we're currently using shared pointers and evaluate whether that is necessary.

bangerth:
Part of #2375.

PR #2817, opened 2019-02-15
Use std::unique_ptr instead of std::shared_ptr for the boundary tractions.

bangerth:
Part of #2375 and #2811.

### DIFF
--- a/source/boundary_traction/ascii_data.cc
+++ b/source/boundary_traction/ascii_data.cc
@@ -35,15 +35,10 @@ namespace aspect
     void
     AsciiData<dim>::initialize ()
     {
-      const std::map<types::boundary_id,std::shared_ptr<BoundaryTraction::Interface<dim> > >
-      bvs = this->get_boundary_traction();
-      for (typename std::map<types::boundary_id,std::shared_ptr<BoundaryTraction::Interface<dim> > >::const_iterator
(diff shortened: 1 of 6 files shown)


bangerth:
The `traction_ascii_data` test fails. I'll have to investigate what is happening, so ignore this PR for the moment.

source/boundary_traction/ascii_data.cc
```diff
@@ -35,14 +35,11 @@ namespace aspect
     void
     AsciiData<dim>::initialize ()
     {
-      const std::map<types::boundary_id,std::shared_ptr<BoundaryTraction::Interface<dim> > >
-      bvs = this->get_boundary_traction();
-      for (typename std::map<types::boundary_id,std::shared_ptr<BoundaryTraction::Interface<dim> > >::const_iterator
-           p = bvs.begin();
-           p != bvs.end(); ++p)
+      std::set<types::boundary_id> traction_bi;
```
tjhei: you introduced a local set here that you are filling but not using.

bangerth: Yes, good point. Stupid copy-paste mistake.

bangerth:
That did the trick, thanks @tjhei for the pointer!

PR #2820, opened 2019-02-18
Change order of spherical unit vectors from theta, phi to phi,theta

anne-glerum:
When setting up a chunk model with boundary velocities using the `function` plugin, I noticed that although the variables in the spherical coordinate system are interpreted as radius, longitude and latitude, the spherical unit vectors point in the direction of radius, latitude and then longitude. It seems more intuitive and in line with the general order of spherical coordinates in aspect to switch to radius, longitude, latitude. Did I understand this correctly and would you agree @bartniday ?


source/utilities.cc
```diff
@@ -547,13 +547,13 @@ namespace aspect
               const double u_theta = spherical_vector[2];
 
               cartesian_vector[0] = std::cos(phi)*std::sin(theta)*u_r
-                                    - std::sin(phi)*u_theta
-                                    - std::cos(phi)*std::cos(theta)*u_phi; // X
+                                    + std::cos(theta)*std::cos(phi)*u_theta
+                                    - std::sin(phi)*u_phi; // X
```
gassmoeller: Could you make the reordering consistent by also here and below always using `a*u_r + b*u_phi + c*u_theta`, and order the individual terms as `cos(phi)cos(theta)` instead of `cos(theta)cos(phi)`?

anne-glerum:
@gassmoeller I've addressed your comment :)

gassmoeller:
Thanks :+1:

anne-glerum:
@gassmoeller Just a question, existing prm files from users will be interpreted differently now, do we note that down somewhere?

gassmoeller:
Good point. Usually we note that in the changelog entries, and if possible also provide an update script in `doc/update_scripts`. However, this feature was added after the latest release, so we do not need to do either. But could you update @bartniday 's changelog entry with the new coordinate directions in a follow up PR? It is in doc/modules/changes/20180624_niday

PR #2822, opened 2019-02-20
Update citation templates.

bangerth:
I've apparently had this in my repository for a while already. It takes over jabref style files from deal.II.


tjhei:
Let's try it!

PR #2823, opened 2019-02-21
Complete a bibtex entry.

bangerth:


tjhei:
this won't work. This file is automatically generated from mendeley. I will make the change there.

tjhei:
see #2824

bangerth:
Ah, that explains the funny formatting. I forgot how this all works...

PR #2826, opened 2019-02-22
www: publications.html: remove pdf

tjhei:
the PDF links to files are not working and do nothing (because these are local files in
mendeley). Remove them.


tjhei:
See the first entry: https://aspect.geodynamics.org/publications.html

PR #2827, opened 2019-02-22
Fix the formatting of '@misc' entries.

bangerth:
In particular, this affects the entry at
  https://aspect.geodynamics.org/publications.html
for the ASPECT manual (the second one in 2015). The format totally doesn't
match any of the other entries, and also misses the newline at the end to separate
itself from Tosi et al.


tjhei:
This looks good. Are you keeping track about the changes inside ASPECT vs deal.II for this?

bangerth:
I was going to sync at one point.

PR #2829, opened 2019-02-25
avoid warning in gravity point postprocessor

tjhei:
I am getting warnings about un-initialized variables for the dim=2
instantiation that doesn't work anyways. Fix this.

### DIFF
--- a/source/postprocess/gravity_point_values.cc
+++ b/source/postprocess/gravity_point_values.cc
@@ -53,6 +53,7 @@ namespace aspect
     {}
 
 
+
     template <int dim>
     std::pair<std::string,std::string>
     GravityPointValues<dim>::execute (TableHandler &)
@@ -60,6 +61,7 @@ namespace aspect
       AssertThrow(false, ExcNotImplemented());
       return std::pair<std::string,std::string>();
     }
+
 
 
     template <>
@@ -408,6 +410,8 @@ namespace aspect
       return std::pair<std::string, std::string> ("gravity computation file:",filename);
     }
 
+
+
     template <int dim>
     void
     GravityPointValues<dim>::declare_parameters (ParameterHandler &prm)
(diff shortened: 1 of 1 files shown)


source/postprocess/gravity_point_values.cc
```diff
@@ -52,10 +52,22 @@ namespace aspect
...
+      AssertThrow(false, ExcNotImplemented());
+      return std::pair<std::string,std::string>();
+    }
+
+
+    template <>
+    std::pair<std::string,std::string>
+    GravityPointValues<3>::execute (TableHandler &)
+    {
+      const int dim = 3;
```
gassmoeller: `const unsigned int`

source/postprocess/gravity_point_values.cc
```diff
@@ -52,10 +52,22 @@ namespace aspect
       output_file_number (numbers::invalid_unsigned_int)
     {}
 
+
```
gassmoeller: I think we usually have 3 empty lines between functions (here and below).

source/postprocess/gravity_point_values.cc
tjhei: Interesting point. We normally use unsigned ints, but our template parameters are always
```
template <int dim>
```
It looks like we mostly use signed ints in ASPECT in this situation, but ~75% unsigned inside deal.II:
```
$ grep "const int dim" -r deal-git/source/ | wc -l
25
$ grep "const unsigned int dim" -r deal-git/source/ | wc -l
87
```
I will summon @bangerth for his opinion. :-)

gassmoeller:
Ok, lets not delay this PR further because of a not crucial discussion about `int` vs `unsigned int`. The content of the PR is definitely correct. We can think separately about unifying the templates if we ever want to. Thanks for the fix.

PR #2833, opened 2019-03-01
fix the maximum horizontal compressive stress plugin.

MFraters:
This is a follow up to #2743 to fix issue #2740. 

This pull request fixes the problem and the documentation. If no test fails, I will add a test so that we will be able to see if something changes in the future.


MFraters:
I had two sort of test cases when I investigated the problem. I will add them as tests.  The difference is not large in these cases, but visible.

First case velocity:
case one velocity

First case new (green) and old (red) maximum horizontal shear stress:
case two velocity

Second case velocity:
case one max horizontal compressive stress
Second case new (green) and old (red) maximum horizontal shear stress:
case two max horizontal compressive stress

MFraters:
I am not sure why the tester gets significantly different values for one of the two new tests the I get on my computer. Tester it shows this diff for the log:
```
##17      <==      RMS, max velocity:        1.07 m/year, 2 m/year

##17      ==>      RMS, max velocity:        1.09 m/year, 2.08 m/year


##17      #:4   <== 1.07

##17      #:4   ==> 1.09

@ Absolute error = 2.0000000000e-2, Relative error = 1.8691588785e-2

##17      #:6   <== 2

##17      #:6   ==> 2.08

@ Absolute error = 8.0000000000e-2, Relative error = 4.0000000000e-2
```
And the gnpulot file, of which one entry is:
```
----------------

##8       <== 0 0 0 0 0.745201 -1 -3.63459e+18 0 3319.34 5e+24 1.55603e+16 -6.68293e+18 0

##8       ==> 0 0 0 0 0.839453 -1 -3.63459e+18 0 3319.34 5e+24 1.55603e+16 -6.68293e+18 0


##8       #:5   <== 0.745201

##8       #:5   ==> 0.839453

@ Absolute error = 9.4252000000e-2, Relative error = 1.2647862791e-1

```

gassmoeller:
It looks like the difference is in the velocity of one of the corner vertices (xyz = 0,0,0). I would guess that you ran the tests on a different deal.II version than the tester (8.5.0)? So what I could imagine is that the boundary conditions are applied in a different order, leading to slightly different test results. I would be fine if you update the test results and we merge as is.
If you want to be super careful, you can run the test on both deal.II versions with increasing resolution and make sure they converge to the same result, or find a set of boundary conditions where the order does not matter.

gassmoeller:
@MFraters any chance we can get this into the release? Otherwise we need a list of known bugs.

MFraters:
ah, I forgot about this one. I will just update the results. I will try to do it today.

tjhei:
The new test also generates quite a lot of output (150k lines of gnuplot). Is it possible to change to a lower resolution?

MFraters:
ok, will do that.

MFraters:
It won't be done tonight (currently compiling), so I will fix it tomorrow.

MFraters:
How do I access the patch file in Jenskins?

gassmoeller:
There is a tab called `Artifacts` in the top right corner of the screen that holds the patch files.

MFraters:
Thanks :)

tjhei:
tests pass, things look good! :+1:

PR #2835, opened 2019-03-04
add cmake compile commands output.

MFraters:
This will automatically create a compile_commands.json file when cmake is run. This file can then be used by IDEs like vscode to automatically setup the project correctly.

I think the default of setting it on should be fine, since it will help people to setup their project more easily and it only adds one file to the build directory (which is added to the .gitingore file). I also added the .vscode directory to the .gitignore.


tjhei:
This is also useful for clang-tidy, so I like it! I agree that there is no problem in generating this file unconditionally.

tjhei:
Let me check if cmake 2.8 supports this...

MFraters:
> Let me check if cmake 2.8 supports this...

[This](https://stackoverflow.com/questions/23960835/cmake-not-generating-compile-commands-json) post says it should be supported form 2.8.5 (we require at least 2.8.8), but I have trouble reading the documentation for cmake before 3.0. So if you could find it, it would be great :)

MFraters:
I can find the command in  [the documentation of cmake 3.5.2](https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html?highlight=cmake_export_compile_commands) but not in version before it. Will it harm if the cmake variable is set, but not used?

tjhei:
> Will it harm if the cmake variable is set, but not used?

I tried it out and it works without errors!

PR #2836, opened 2019-03-04
set minimum required cmake version

tjhei:
It turns out that we do not support 2.8.8 as advertised, but that we
need 2.8.12 (confirmed working).
While here, fix a few weird looking CMakeLists.txt files (missing
headers)


MFraters:
just out of curiosity, what goes wrong with version 2.8.8?

tjhei:
```

CMake Error at /scratch/libs-candi-v8.5.1-r1/install/deal.II-v8.5.1/lib/cmake/deal.II/deal.IITargets.cmake:62 (message):
  This file relies on consumers using CMake 2.8.12 or greater.
Call Stack (most recent call first):
  /scratch/libs-candi-v8.5.1-r1/install/deal.II-v8.5.1/share/deal.II/macros/macro_deal_ii_setup_target.cmake:64 (INCLUDE)
  /scratch/aspect-git/buildtemp4/AspectConfig.cmake:38 (DEAL_II_SETUP_TARGET)
  CMakeLists.txt:175 (ASPECT_SETUP_PLUGIN)


CMake Error at CMakeLists.txt:241 (MESSAGE):
  ERROR: tests/ project could not be configured.
```
this means any plugin (including tests) requires 2.8.12 because deal.II 8.5 requires it.

PR #2839, opened 2019-03-05
update copyright years

tjhei:
We normally do this before a release, but I saw we are not up to date when looking at the CMake files the other day.


bangerth:
This is the sort of patch to hide an Easter egg in. But it looks like you've been a good boy here :-)

PR #2845, opened 2019-03-08
fix clang warning

tjhei:
clang 6 warns about an array access out of bounds when instantiating for
dim=2. This is of course not a problem as this code is only executed for
dim=3. Silence the warning by using dim-1 instead of 2.


gassmoeller:
Looks reasonable to me.

PR #2846, opened 2019-03-08
remove uncaught exceptions handling

tjhei:
I think this logic is unnecessary as .reset() is safe to call in any
situation (we do not have any "on destruction" printing enabled).

I can into a problem with clang 6 with gcc's libstdc++ with version 5.4:
uncaught_exceptions() is not declared, but the version is reported as
c++17.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -486,42 +486,13 @@ namespace aspect
     // object (set from the output_statistics() function)
     output_statistics_thread.join();
 
-    // Detect if we are being destroyed because an uncaught exception has
-    // been triggered. If so, reset() the TimerOutput class. This will clear
-    // the currently active timing sections. Otherwise, its destructor will try to
-    // do MPI communication when leaving the open sections, which can cause one of the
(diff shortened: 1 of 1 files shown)


gassmoeller:
Ok, seems good to me.

source/simulator/core.cc
```diff
@@ -486,42 +486,13 @@ namespace aspect
...
-    if (std::uncaught_exception())
-#endif
-      computing_timer.reset();
+    // If an exception is being thrown (for example due to AssertThrow()), we
+    // might end up here with currently active timing sections. The destructor
+    // of TimerOutput does MPI communication, which can lead to deadlocks,
+    // hangs, or confusing MPI error messages. To avoid this, we can call
+    // reset() to remove all open sections. In a normal run, we won't have any
+    // active sessions, so this won't hurt to do:
+    computing_timer.reset();
```
bangerth: Oh, doh. I must have spent an hour on the patch that figured all of the previous code out...

PR #2848, opened 2019-03-11
Fix Volume of Fluid visualization plugin

class4kayaker:
While working on putting together a cookbook for the VoF composition advection method, I found that I had not corrected the names of the parameters for the vis postprocessor when reading from the parameter file. This PR makes that correction.

### DIFF
--- a/source/postprocess/visualization/volume_of_fluid_values.cc
+++ b/source/postprocess/visualization/volume_of_fluid_values.cc
@@ -153,8 +153,8 @@ namespace aspect
           {
             prm.enter_subsection("Volume of Fluid");
             {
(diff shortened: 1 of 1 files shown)


tjhei:
Seems like none of your tests generate the output. I changed that here: #2849

class4kayaker:
To clarify, are you indicating that I should modify one of the tests to generate the output in this PR, or is the existing PR intended to solve that issue separately?

tjhei:
> or is the existing PR intended to solve that issue separately?

Yes, no worries. Take a look at the diffs.

bangerth:
In reference to #835.

PR #2849, opened 2019-03-12
VoF: add test for graphical output

tjhei:
none of the tests trigger graphical output of "volume of fluid values".
Fix this.


tjhei:
test failed crashed before being based on #2848 and passes now. Ready to be merged, I think.

gassmoeller:
Yes looks good. I guess it was on purpose that you did not add the binary output files? We could also choose to include low resolution gnuplot files to the test data. What do you think?

gassmoeller:
Anyway this is better than before, so I am going to merge. Thanks!

bangerth:
In reference to #835.

PR #2850, opened 2019-03-13
print more MPI and deal.II info

tjhei:
- in detailed.log
- and some at runtime

### DIFF
--- a/cmake/write_config.cmake
+++ b/cmake/write_config.cmake
@@ -69,5 +69,12 @@ _detailed("#
 #        COMPILE_FLAGS:             ${_COMPILE_FLAGS}
 ")
 
+_detailed("#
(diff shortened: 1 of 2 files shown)


gassmoeller:
Did you check if this works with 8.5? I do not know if the tests would catch it, because they filter the header. Otherwise good to go.

tjhei:
> Did you check if this works with 8.5?

```
$ ./aspect
--     . version 2.1.0-pre (print_dealii_info, 7f5647e)
--     . using deal.II 8.5.1
--     . using p4est 2.0.0
```

> Oh, and does the vectorization level make a difference?

Only for a limited number of deal.II internal functions unless you do matrix-free computations.

gassmoeller:
Alright, good to go.

PR #2851, opened 2019-03-14
Add option for higher order output

### LINKED ISSUE #2601: Eventually implement higher order output for paraview
Just saw this by chance: Deal.II and paraview now support higher order output as discussed for example here: dealii/dealii#6931. We should implement the option to use this feature and at some point in the far future (when everyone uses at least paraview 5.5) make it the new default. It would save disk space and allow visualizations of larger models (less memory requirements for visualization).

gassmoeller:
As discussed in #2841 this creates the option to write visualization data as higher order polynomials. It also adds a test that uses the option and a figure that shows the difference. First tests using this for some application models look good, and I think we should at some point enable this by default (likely when we require deal.II 9.1 at some point in the future). 

Closes #2601.


source/postprocess/visualization.cc
```diff
@@ -573,6 +573,10 @@ namespace aspect
           vtk_flags.cycle = this->get_timestep_number();
           vtk_flags.time = time_in_years_or_seconds;
 
+#if DEAL_II_VERSION_GTE(9,1,0)
+          vtk_flags.write_higher_order_cells = write_higher_order_output;
```
MFraters: Shouldn't this be dependent on whether you set the option `Write higher order output` to true?

doc/modules/changes/20190313_gassmoeller
```diff
@@ -0,0 +1,8 @@
+New: There is now an option to output visualization data as higher order
+polynomials. This is an improvement in accuracy and requires less disk space
+than the 'Interpolate output' option that was available before (the new option
+requires the old option to be set, which it is by default). However the new
```
MFraters: What is the reason you choose to make this option dependent on an other option? Can't we just deal with this internally?

include/aspect/postprocess/visualization.h
```diff
@@ -462,6 +462,19 @@ namespace aspect
          */
         bool filter_output;
 
+        /**
+         * deal.II offers the possibility to write vtu files with higher order
```
MFraters: Capital D in Deal.II?

source/postprocess/visualization.cc
gassmoeller: It is. `write_higher_order_output` is only true if it is set in the input file, otherwise it is false and nothing changes.

doc/modules/changes/20190313_gassmoeller
gassmoeller: We could, but then setting `Interpolate output` would not change anything if `Write higher order output` is set. That might be confusing as well. Would you prefer the internal solution?

include/aspect/postprocess/visualization.h
gassmoeller: Nope, it is usually deal.II, see https://dealii.org/. For reasons you would need to ask Wolfgang :smile:. It looks a bit funny at the beginning of a sentence, but I think it is consistent with how it is usually used.

source/postprocess/visualization.cc
MFraters: ah, I now see. I overlooked it :(

include/aspect/postprocess/visualization.h
MFraters: Since you where so consistent with it, I already suspected something like that. I just wanted to make sure :)

doc/modules/changes/20190313_gassmoeller
MFraters: I Ithink the best option would be to have the user choose of the three options in one parameter, but that would not be backwards compatible. 

I would personally prefer to have this done internally, because it is one thing less for the user to worry about/check, but I do not feel strongly about it and am also fine with leaving it this way. I will leave it up to you.

gassmoeller: In that case I prefer the current way :smile:. In a few years the question will become  irrelevant anyway, as we can then simply always use the higher order output if the interpolate option is set to true.
Maybe leave some time for others to comment (e.g. until tomorrow or so) and then feel free to merge.

MFraters:
Should this wait for #2841, since you are stating that `Interpolate output` is on by default?

gassmoeller:
Probably yes, but #2841 is also ready from my side, so feel free to merge both if you feel comfortable with them.

gassmoeller:
I rebased and resolved the conflict. Should be ready to go.

PR #2855, opened 2019-03-19
Multi VoF field correction

class4kayaker:
I ran into another volume of fluid edge case that I had not explicitly checked due to assuming it was covered by existing tests.
If multiple VoF fields were used, they were not associated with the correct compositional fields due to using incorrect indexes in the map (source was forgetting about the indexes including temperature).

This both corrects the implementation, and adds a postprocessor to an existing test to ensure that it remains correct.


gassmoeller:
Ok, looks good to me. You will need to `make indent` to fix the style.

class4kayaker:
Force pushed  indentation correction commit to force new check.

gassmoeller:
Looks good. Thanks!

bangerth:
In reference to #835.

PR #2857, opened 2019-03-22
Add entropy viscosity statistics postprocessor

gassmoeller:
This PR adds a new statistics postprocessor for the entropy viscosity stabilization. I used it to debug the EV method, and it allowed me to find some fixes that I will propose in a follow-up PR. It is also a nice way to find out how much a model is affected by the EV stabilization.


naliboff:
After a first pass this looks good to me, thanks!. 

My only suggestion is to derive `entropy_viscosity_statistics.prm` directly from `artificial_viscosity.prm`, assuming the only change is indeed adding the new statistics post processor.

Issue #2713, opened 2018-11-27
[Meta] Improve accuracy of ASPECT solutions


bangerth:
We've had a number of discussions over the years on how to improve the accuracy of the solutions ASPECT computes. Some of these have been implemented, for example @ian-r-rose's implementation of the consistent boundary flux method for the stress in #1572, or @gassmoeller's implementation for the heat flux in #2675.

Other ideas are proposed in #2711 and #2712. Let us collect what else we can come up here or in separate issues linked here.

maxrudolph:
I had some discussions with @tjhei, @Shangxin-Liu, and Scott King at AGU about how to address these issues. To summarize, Scott and Shangxin have found through their efforts to reproduce the benchmarks in Zhong et al. 2008 that they can only reproduce the CitcomS benchmark values for temperature and heat flux if the stabilization parameters for the entropy viscosity scheme are set to zero. However, the benchmarks in Zhong et al. 2008 are run at low to moderate Rayleigh number. At higher Ra, turning off stabilization will lead to numerical instabilities. I have basically five ideas about how to improve the accuracy of the solution to the advection diffusion equation.

1. Attempt to find a more suitable choice of the stabilization parameters that would make the existing entropy viscosity implementation viable. 
2. Explore the use of the DG discretization for temperature. My understanding is that the DG discretization does not require stabilization. Correct? We tried this already and encountered failures to converge as well as inconsistency of boundary values. Someone with deeper knowledge of DG would have to help with this.
3. Re-implement in ASPECT SUPG the way that it is implemented in CitcomS. I have read the implementation in CitcomS and would be happy to write out a detailed note on how it is implemented but I will need help implementing it in aspect/deal.ii. This may be more difficult than it appears at first glance because the elements used in ASPECT and CitcomS are different.
4. Advect temperature on particles following Gerya and Yuen (2007, 2009), also now done in StagYY apparently. Gerya and Yuen add an additional subgrid scale temperature relaxation scheme which is physically motivated but potentially unsatisfactory to some as it is somewhat ad-hoc.
5. Implement a different, yet-to-be-identified scheme for the advection-diffusion equation. In choosing a scheme, we should consider not just the convergence properties but also the accuracy at the levels of resolution that are accessible with current computing resources.

[...]

Issue #2797, opened 2019-02-01
Error in solution.xdmf file when simulation resumes from restart files


ricitron:
Hello,

I noticed a strange error that appears in the solution.xdmf file when I restart an ASPECT simulation. The error only occurs when a simulation stops and then is resumed from the restart files. If a simulation completes in one go then the solution.xdmf file is fine.

When I start the simulation, the solution.xdmf file begins:

```
<?xml version="1.0" ?>
<!DOCTYPE Xdmf SYSTEM "Xdmf.dtd" []>
<Xdmf Version="2.0">
  <Domain>
    <Grid Name="CellTime" GridType="Collection" CollectionType="Temporal">
      <Grid Name="mesh" GridType="Uniform">
        <Time Value="0"/>
        <Geometry GeometryType="XY">
          <DataItem Dimensions="1769472 2" NumberType="Float" Precision="8" Format="HDF">
            solution/mesh-00000.h5:/nodes
          </DataItem>
```

However, when the simulation stops and then I resume it from the restart files, the solution.xdmf file is changed so it appears:

```
<?xml version="1.0" ?>
<!DOCTYPE Xdmf SYSTEM "Xdmf.dtd" []>
<Xdmf Version="2.0">
  <Domain>
    <Grid Name="CellTime" GridType="Collection" CollectionType="Temporal">
      <Grid Name="mesh" GridType="Uniform">
        <Time Value="0"/>
        <Geometry GeometryType="XYZ">
          <DataItem Dimensions="1769472 4294967295" NumberType="Float" Precision="8" Format="HDF">
            solution/mesh-00000.h5:/nodes
          </DataItem>
        </Geometry>
```

Any line that previously had ' <DataItem Dimensions="1769472 2" ...' now has '<DataItem Dimensions="1769472 4294967295" ...'. All the number '2' values are replaced with '4294967295' after the simulation is restarted from the restart files. 

This only occurs when I restart the simulation from the restart files. I noticed this on two clusters I had the software installed on, and a completely different user tested this with a different install on a different cluster and noticed the same error. 

Because of the error in the solution.xdmf file, I cannot load the data into Paraview. However, if I find and replace '4294967295' with '2' then I can load the .xdmf file into paraview just fine. It appears that everything else in the .xdmf file is fine except for this strange error. 

Thanks.

gassmoeller:
Hi Robert,
[...]

PR #2858, opened 2019-03-23
Improve entropy viscosity at conduction dominated boundaries

gassmoeller:
This is a step towards #2713 and might be able to fix a lot of the weird advection benchmark results we got. It sits on top of #2857 (so you also see the changes of that one).

I went through the implementation of the EV method and noticed two things that should be addressed in this PR (I packed them in one PR, because they are both going to change all test results): 

1. We had a bug in the maximum viscosity computation for timestep 1. Instead of using the real velocity we only used half of its value (because `old_old_solution` was not initialized in timestep 1). That seems small, but it leads to too little stabilization in timestep 1 and was the original reason why the van Keken overturn model showed so much oscillations that I changed the default for the stabilization parameters in 2012 in commit a9b886a3df599a3bc9c3ff80c6d5eeadbda02cac. Now that the bug is fixed we can safely return to the former (lower) defaults that are also described in the original ASPECT paper. I also added a test that analytically checks the correct value for the maximum stabilization. For comparison: with the old defaults we had maximum overshoots of 15% before fixing the bug, and 5% with the fix (and they are stable and decrease over time). This alone reduces the stabilization significantly.

2. The residuals that scale the entropy viscosity are always large at Dirichlet boundaries. Previously, I thought that was because of the strong gradients in the boundary layer, but even in well resolved models they are way bigger than I would expect. This is also consistent with the deviations @GEuen, @Shangxin-Liu, @maxrudolph and @sdkatvt saw in convection benchmarks where lots of stabilization is applied at boundaries even when steady-state is reached. I think the reason is that our residual computation for EV does not take into account the heat flux that results from the boundary. When solving the equation we apply the constraints to the solution vector to prescribe the boundary condition, but when computing the residual for the entropy viscosity we do not take that into account. The result is a serious overstabilization of conduction dominated boundary layers. Unfortunately, I do not see an easy way to compute a more correct residual (because we no longer have the boundary conditions for past time steps), but in general the boundary cells are conduction dominated anyway if there is a thermal boundary layer. Therefore we should be able to safely switch off stabilization in boundary cells with prescribed temperatures. I ran tests with the blankenbach benchmarks and the gplates cookbooks and the benchmark results show significant improvements for underresolved models (e.g. 8x8 cells for Ra=1e6) with a bit more overshoots than before, generally less than 5% of the temperature jump. In the application models I see very little overshoots (~2% at the sharpest temperature jumps above plumes), with a significant decrease in stabilization and a much more accurate boundary heat flux.

This PR is work in progress, because I have not added enough tests and documentation, and have not updated the benchmark results in the benchmarks folder, but the implementation is ready for testing. In particular, I would like to ask @GEuen and @maxrudolph to rerun their benchmarks with this version (If you do not have the time, please send me some input files, and I can run them as well). Also if anyone finds a model that is now not longer sufficiently stabilized we need to know about it.


maxrudolph:
Thank you for working on this Rene. I am cloning this branch and building now. I will try to get the tests running shortly.

GEuen:
Thanks from me as well Rene. I have the branch cloned and tests are running. I have a few preliminary ones done and they look promising. I'll have plenty of numbers next week as the last tests finish.

gassmoeller:
Ok sounds great. Really curious to see what data you get. Let me know once you have it.

maxrudolph:
I also ran a bunch of tests but haven't had time to collate the results
yet. Will try to get it to you later this week.

Max

gassmoeller:
Any news on the benchmark results? I do not need a full analysis, but would like to get a feeling if it is worth pursuing this further: 1. Do the new results look better than before? 2. Have you found any models that are unstable in the sense that there are oscillations that grow over time? Some minor over/undershoots are expected, but they should not grow.

GEuen:
Hey, Rene. Sorry for the late reply. Results are definitely better on my
end. I'm still analyzing, and I'll have some tables in the next day or two
with specific numbers. Using default stabilization, agreement with the
Zhong benchmark is very good. These cases are more heavily influenced by
refinement of mesh, but even coarse cases are in good agreement. I haven't
seen any unstable cases.

gassmoeller:
Awesome, thanks for the work! Max is this consistent with your results? Then I would go ahead and clean this up.
@tjhei you mentioned you had a different idea for how to improve the residual computation at boundaries? Would you still be willing to consider the current solution, or should I break this up into two parts (1. fixing the first timestep bug and changing the defaults, and 2. fixing the prescribed boundary residuals)?

maxrudolph:
@gassmoeller I finished pulling together the results of the first benchmark test. The problem setup is a spherical shell with constant inner and outer temperatures of 1973 and 973, respectively. The initial condition is uniformly hot at 1973. I ran tests using 'head' (latest aspect development branch as of a couple of weeks ago) and also your version ('e6af77b'). I ran models at global refinement levels 3,4, and 5. I run each model long enough to reach steady state and plot T(r). There is an analytic solution at steady state of the form T(r) = a/r + b. Your new version matches the analytic solution for T(r) at steady state much better than the current development version for global refinement 5.
T_vs_r

Next, I plot the surface and cmb heat flow reported in the statistics file. The analytic solution value for the steady state heat flow is indicated with the blue horizontal line. There is no internal heating, so qsurf=qcmb at steady state (note that I actually plot -qcmb reported by aspect). For global refinement level 5, the solutions approach the correct value quite nicely with your new code. For global refinement levels 3-4, the results are very unsatisfactory.

heat_flow_vs_time

Note that because this is a purely conductive solution, no stabilization should be necessary anywhere within the domain. The temperature gradients in this problem are very small relative to what we expect in a thermal convection problem with thin boundary layers, and I surprised that aspect performs so poorly even at global refinement levels 3-4, which should be more than adequate to resolve a conduction problem. This suggests to me that there is still too much artificial diffusion being introduced throughout the domain.

All of the plotting scripts and input files can be found here: maxrudolph/aspect-energy-equation

I have a second set of tests for thermal convection that have been run but not analyzed. Will try to get it done today.

gassmoeller:
Thanks for sharing the results Max. It seems the changes in this PR are a good step forward. Still there might be something funny going on at low resolutions in spherical geometries. Since the temperature is reaching steady-state it suggests the temperature equation is solved correctly in the sense that the solution is fulfilling the terms in the equation including the stabilization. But the stabilization is higher than necessary. SUPG would not have that issue in this particular benchmark, because its stabiiization is tied to the velocity, which is 0 in this benchmark (just noting this for my future self). This suggests we should either:
1) take another closer look at the computation of the advection equation residual, which is used for the entropy viscosity computation, and figure out why it is large even when the solution is smooth and has reached steady state (since all terms of the residual should be small in steady state, we can check which one is not)
2) try implementing SUPG as discussed.

Do I read your first graph correctly that at very coarse resolutions the new temperature solution is actually worse than before, and that most of the deviation happens close to the boundary? This would be unfortunate, because it means we might need that stabilization in the boundary cells after all.

Just for reference, can you upload your parameter file and ASPECT and deal.II versions to this PR? I know it is a super simple setup, but I want to make sure to reproduce it exactly when trying to figure out what happens to the advection residual (i.e. it could be something about the accuracy of the spherical mapping in deal.II and I would like to know the version for that before going guessing around).

sdkatvt:
A puzzling aspect of Max’s result compared with Grant’s is that Grant's radial average plots are all nearly on top of each other for 3, 4, and 5 refinement steady cases.   I was surprised how good the 3 refinement case was.   I would have predicted a diffusion problem would be spot on.  

I see two differences: 1) Grant uses a (nearly) uniform mesh to more closely compare with CitcomS and 2) he used non-dimensional values for radius and temperature.  I suppose it is possible that there is a small scaling problem that is exacerbated by the the dimensional scaling and doesn’t impact the non-dimensional case (as much).   In the past Grant has tried some problems scaled dimensionally.  We could try one or two of those again.

It would probably be helpful if Grant uploaded his config files as well so that everyone can work from a consistent set.  The 3 and 4 refinement cases run in the debug queue on our system so they can be done pretty easily.   He does need to make two minor changes to the code, one to make the rheology Frank-Komenetskii and one add a second perturbation for the C1 cases.

gassmoeller:
> It would probably be helpful if Grant uploaded his config files as well so that everyone can work from a consistent set. 

Yes that would be great. We would love to have that as a benchmark case or cookbook in the official repository, but uploading as an attachment to this pull request would also be good.

>  I was surprised how good the 3 refinement case was.

After years of bugfixing the use of higher order methods finally pays off :smile:

>  I would have predicted a diffusion problem would be spot on.

Not necessarily. The entropy viscosity method was not designed with pure diffusion problems in mind. Since its stabilization is applied to the conduction term (instead of the advection term like in SUPG) it does not automatically go to 0 for a problem without advection. The pure conduction case might be somewhat degenerate for the method. However, I still find it suspicious that in the case of a steady-state solution without flow, where essentially every term in the equation should go to zero we keep having a residual that triggers some stabilization. I think our best bet is to track down where that residual appears (by looking at the artificial viscosity postprocessor and where it shows stabilization), and which term in the residual is still big (e.g. by individually commenting out the residual terms in `source/simulator/assemblers/advection.cc::AdvectionSystem::compute_residual()` and looking when the stabilization disappears). That should give an idea where the problem is.

tjhei:
> @tjhei you mentioned you had a different idea for how to improve the residual computation at boundaries?

Maybe? I think we should sit down at the hackathon and discuss all of this. 

Maybe we need to rethink the residual computation again.

maxrudolph:
I have some preliminary results for the thermal convection testcase now. This is running at global refinement 3, so poorly resolved, but I do not entirely understand the differences that I'm seeing between this branch and 'head'. For some reason the timesteps using this version are larger, so despite the fact that these jobs were started within seconds of each other running on the same number of cores, the model time is nearly twice as large using this PR as in 'head'.

heat_flow_vs_time

The above figure shows the surface and CMB heat flow, along with average internal temperature. It looks like this branch cools less quickly and has much lower surface and cmb heat flows than the current development version. Should I attribute this primarily to having lower artificial diffusion in the boundary layers? It looks like you have changed the default values of the stabilization parameters relative to the main development branch, so there may be less artificial diffusion in the interior as well.

```
diff output-shell_simple_3d_aspect-e6af77b_3/parameters.prm output-shell_simple_3d_aspect-head_3/parameters.prm
48c48
< set Dimension                                  = 3                                       # default: 2
---
> set Dimension                                  = 3                                    # default: 2
55c55
< set End time                                   = 2e10                                    # default: 5.69e+300
---
> set End time                                   = 2e10                                 # default: 5.69e+300
95c95
< set Maximum time step                          = 1e7                                     # default: 5.69e+300
---
> set Maximum time step                          = 1e7                                  # default: 5.69e+300
141c141
< set Output directory                           = output-shell_simple_3d_aspect-e6af77b_3 # default: output
---
> set Output directory                           = output-shell_simple_3d_aspect-head_3 # default: output
197c197
< set Use conduction timestep                    = true                                    # default: false
---
> set Use conduction timestep                    = true                                 # default: false
2090c2090
<     # itself or its residual. An appropriate value for 2d is 0.052 and 0.78
---
>     # itself or its residual. An appropriate value for 2d is 0.078 and 0.117
2101,2102c2101,2105
<     # of the paper.) Units: None.
<     set beta                                               = 0.052
---
>     # of the paper. After further experiments, we have also chosen to use a
>     # different value than described there: It can be chosen as stated there
>     # for uniformly refined meshes, but it needs to be chosen larger if the
>     # mesh has cells that are not squares or cubes.) Units: None.
>     set beta                                               = 0.078
2117,2118c2120,2123
<     # formulas following equation (15) of the paper.) Units: None.
<     set cR                                                 = 0.11
---
>     # formulas following equation (15) of the paper. After further
>     # experiments, we have also chosen to use a different value than described
>     # there.) Units: None.
>     set cR                                                 = 0.33
7344,7347d7348
<   # `entropy viscosity statistics': A postprocessor that computes the maximum
<   # and volume averagedentropy viscosity stabilization for the temperature
<   # field.
<   #
8388,8389c8389,8390
<     # (i.e., the option is switch off, as is the default), and the right
<     # picture shows the same simulation with the option switched on. The
---
>     # (i.e., the option is switched off), and the right picture shows the same
>     # simulation with the option switched on (which is the default). The
8398c8399
<     set Interpolate output                  = true       # default: false
---
>     set Interpolate output                  = true
```

I also plotted the depth average of the temperature after 2.5e9 years. It looks like this PR is not cooling as quickly, which is probably more reasonable (but I don't have an analytic solution).

T_vs_r

I'm happy to run this again with Global refinement 4 from the beginning, and to run a case with higher refinement in the boundary layers, if it will help.

maxrudolph:
Per @sdkatvt suggestion that the issues there could be issues associated with nondimensionalization, I re-ran the conductive cooling tests for a dimensionless model. 

T_vs_r
This shows dimensionless temperature vs. radius for all cases. 

heat_flow_vs_time
This shows surface and CMB heat flow, along with the analytic value.

From this, I conclude that the behavior is indistinguishable from the dimensional case. Rene's fix does appear to improve the convergence, though global refinement level 5 is still required to obtain a reasonably accurate solution.

maxrudolph:
Here an update on the convection tests as well. I am running models at global refinement level 5, which is exceedingly slow on 256 cores. However, the results at global refinement level 5 look identical so far to level 4, so I do not anticipate significant differences if these models are allowed to run further. At global refinement level 5, the 'head' branch looks identical to this PR. At lower refinements, this PR produces results that approach more closely the higher refinement level results.   

heat_flow_vs_time

Additionally, to assess how well the models are conserving energy, I calculated the secular cooling rate and the average surface and cmb heat flows over the last 1 Gyr of each model time. In the table below, I show the energy non-conservation as a fraction of the CMB heat flow in the last column. For all of the models, this is on the order of a few percent, which is encouraging. This is not an entirely fair comparison since not all of the models have reached the desired end time (note that this is especially true of the `head_5` case). I also realized in these tests that I forgot to switch the formulation from ASPECT default to Boussinesq.

| Model | averaging period (yr) | Qsurf (W) | Qcmb (W) | Qsecular  (W) | imbalance/Qcmb |
|--------|---------------------|-----------|------------|---------------|------------------|
|results/statistics-aspect-e6af77b_3|9.99e+08|7.20e+12|4.88e+12|-2.01e+12|-6.24e-02|
|results/statistics-aspect-e6af77b_4|1.00e+09|6.54e+12|5.66e+12|-9.45e+11|1.16e-02|
|results/statistics-aspect-e6af77b_5|1.00e+09|1.22e+13|4.40e+12|-8.18e+12|8.46e-02|
|results/statistics-aspect-head_3|1.00e+09|2.87e+13|2.71e+13|-1.60e+12|2.15e-03|
|results/statistics-aspect-head_4|1.00e+09|7.21e+12|6.69e+12|-5.45e+11|3.44e-03|
|results/statistics-aspect-head_5|6.64e+08|1.23e+13|3.89e+12|-8.64e+12|5.04e-02|

From my point of view, this pull request is a step in the right direction. I think that we may want to do some tests at higher Ra as well, but those are likely to be more costly so perhaps we should discuss a bit to make sure that we're all in agreement about which model setups will be most enlightening first.

GEuen:
Here are my latest results for this new branch. Tests are cases A1, A3, A7, and C1 from the Zhong benchmark. I tested values for beta and cR at 0, default, and their old default values. In the A cases this seems to make little difference. C1 has a bit of difference in values in my coarsest mesh test. All of these results are in good agreement with the benchmark.
[Table_of_Results__April_2019 (1).pdf](https://github.com/geodynamics/aspect/files/3091271/Table_of_Results__April_2019.1.pdf)

sdkatvt:
So in addition to the results Grant posted, he has have A LOT of additional tests.  It’s hard to organize them because there are so many variations and I think would have been confusing.

These results are significantly better than anything we saw previously.  In fact it is remarkable how good the 3 refinement cases are.  I’d say the cases are about as good as one can imagine.  It’s not clear to me how much we should burn cycles trying to exactly match some of Shijie’s results because the published CitcomS results don’t have Nu_top = Nu_bottom.  In fact Aspect has better agreement between top and bottom Nu in many/most cases.  

I will propose that we consider for a higher Ra 3D test case, a couple of stagnant lid calculations I did for Venus.  These were amazingly stable (from my 2018 JGR Planets paper).   It would be possible to computer Geoid/Topography/Heatflow/Vrms/Nu/…  I have all the CitcomS results.  Since Grant is going to be working on Venus it would be good to test those.

I was hoping Grant could work with others on this at the Hackathon.   I could skype in and chat with others.  

Cheers,

Scott

maxrudolph:
I would also like to see results at high Ra with mobile lid. Because velocity affects the value of the artificial diffusion, accuracy may depend on both boundary layer thickness and velocity.

sdkatvt:
Good point.  Yet, it is the diffusion problem the method is having the most trouble with.  Consistent with that, if you look at our A7 results, those results (especially for the smaller grids) have the most problematic results (although they are not bad).  But the point is well taken, many of you earthlings want mobile lid models.  It would be good to collect as many types of problems as we can that exercise the code under different conditions.

maxrudolph:
Here are preliminary results for a more complex testcase at higher Ra with imposed time-dependent plate motions and a chemical layer. I computed a version of this case in CitcomS at a resolution of 65^3*12, and then dimensionalized everything and ran an identical model in aspect. Essential features:
- Isothermal surface and CMB, with temperature increase of 2500K
- Boussinesq approximation
- Plate motions from Zhang et al. 2010 for 458-200 Ma, Seton et al. 2012 for 200 Ma-present.
- 150 Myr spinup with initial (458 Ma) plate stage
- Rayleigh number (using CitcomS convention) of 2.0x10^8 
- Internal heating number of Q0=100
- Buoyancy number of 0.5
- CitcomS case was run with radial refinement in lithosphere and lowermost mantle
- Aspect case was run at global refinement 4 (also tried 5 but running too slow...)
- Basal layer is treated using particles in CitcomS, and  as a compositional field with DG and bound-preserving limiter in aspect.

I plot radial profiles for the initial condition and after 150 Myr below. What I see here is that the 'head' version of aspect shows much more rapid cooling, and the entrainment of most of the chemical material within 150 Myr of the run. With this PR (e6af77b), secular cooling is slower but still quite a bit faster than what we predict with CitcomS. I looked at the visualization output for the entropy viscosity, and not surprisingly, the slabs in the aspect model have entropy viscosity between 10-100 times larger than the physical thermal conductivity, so they are most likely artificially cooling the mantle. It is on my to-do list to run a case with AMR (I am thinking global refinement level 3 + 3 levels of AMR as a starting point). Any input on this setup would be welcome. My starting point would likely be to refine based on composition approximate gradient, entropy viscosity, and near the boundaries. Not sure if it would help to add temperature gradient to this as well.

image
image
image

maxrudolph:
An update on the test case. After 400 Myr the radial temperature profiles from the two models are looking quite different. It's interesting that the radial profiles through the bottom boundary layer (including chemical piles) are nearly identical, but the upper mantle temperatures differ by ~250K. 

image

sdkatvt:
This puzzles me because the gradient of the temperature near the top and bottom are nearly identical, so the heat flux *should* be the same.  To the extent that you are correct about the artificial diffusivity being too large this would explain why the Aspect results are cooler than the sitcoms results.  I wonder now about adding the diffusivity only in the direction of flow in this branch of the code.   Grant’s work showed that (only adding diffusion in the direction of the flow) was helpful but not as good as the PR (e6af77b <https://github.com/geodynamics/aspect/commit/e6af77baf60b1a51dad64f1c9be57a90c8e1b90e>), branch for the very low Ra problems.

Your first plots worried me because the biggest difference i saw was the the temperature near the bottom where the pile is.  If the shape of the pile changed, that could cause a lot of problems.  This is a pretty complex case (chemical piles, imposed velocities from a plate model, etc. and there are many places things could go wrong).  One of the lessons we learned in the thermochemical benchmark was that you have to be really simple.   U li Christensen didn’t want to do the RT problems and started with the themo chemical problems.  He said, the RT problem was “too easy” and "everyone’s code would get it right.”  LOL.  Seven years later…    

Can I just say I hate auto correction in typing???

maxrudolph:
@sdkatvt When I first encountered the problems with too-rapid cooling in aspect solutions, it was with a case similar to this. Rene did try implementing 'streamline upwind entropy viscosity', i.e. artificial diffusion only applied along streamlines. The results were at first promising but ultimately lead to numerical instabilities. It might be possible to add a smaller amount of cross-stream artificial diffusion.

gassmoeller:
Hi Max, you are right, this is quite a significant difference. We might have discussed this before, but the Citcom results seems to have a sub-adiabatic radial temperature profile with higher temperatures close to the upper boundary. I would expect this if there were insulating continents (at low Rayleigh number), or if there were internal heating terms. Could there be a difference in the inclusion of radiogenic heating between the two models? That would also explain ASPECT's faster cooling.
If there is no difference: Do you have visualization output for the end state of the ASPECT model, in particular about the magnitude of the entropy viscosity? What are the typical values (a screenshot would be very interesting)?

maxrudolph:
I tried to generate some images of the solution this morning, but there is
an issue with the visualization output. I restarted the calculation and the
xdmf file is unreadable by visit. I see that this is a known issue that can
be avoided by using the development version of deal.II. Maybe it is
possible to repair the damaged xdmf? Otherwise I have to start the model
over, and it's been running for a week. Is Timo's GMG solver going to get
merged sometime soon? I'm hoping that it makes these types of calculations
quite a bit faster.

Both models should have the same amount of internal heating. I have tried
to make the model setups absolutely identical.
Have you compared plume excess temperatures in CitcomS vs. ASPECT? I am
willing to bet that plumes in aspect, at least at this resolution, cool far
more rapidly.

gassmoeller:
> I restarted the calculation and the xdmf file is unreadable by visit.

You can probably fix it by reverting the change mentioned in https://github.com/geodynamics/aspect/issues/2797. Essentially look for `<DataItem Dimensions` and replace the second number by 3 for every line that it appears. Alternatively, maybe you can open the .h5 files directly (I never tried that).

> Have you compared plume excess temperatures in CitcomS vs. ASPECT?

No I have not. But I ran 2D models ASPECT models with this branch with realistic Rayleigh numbers and an equivalent resolution of 4 (which means 5 in 2d because the coarse grid is different) and saw nearly no cells that had artificial diffusion that was higher than physical diffusion, so I expected the influence in 3D to be equally small. Maybe 3D is different though.

maxrudolph:
You cannot open the .h5 files directly (at last not with visit). That was
also the first thing that I tried :). It needs the metadata to know what
fields to look for in the .h5 file. I may  try  building a new copy of
deal.ii-master and aspect and restarting the calculation. This might fix
the xdmf file?

gassmoeller:
Yes rebuilding deal.II and restarting might fix it (not 100% sure, but likely).
Also I noticed my advice above is likely wrong. If you want to fix the existing xdmf, you do not need to replace all `<DataItem Dimensions` but only the ones with `4294967295` as the second number. Can you post the current xdmf file so I can take a look?

maxrudolph:
Here is the .xdmf file
[solution.xdmf.txt](https://github.com/geodynamics/aspect/files/3136099/solution.xdmf.txt)

gassmoeller:
Hi Max, try this file:
[solution_fixed.xdmf.txt](https://github.com/geodynamics/aspect/files/3136356/solution_fixed.xdmf.txt)

Does it work?

maxrudolph:
Thank you Rene. That did work, and it's a good thing too because I killed the run and now can't seem to restart it from checkpoint files after re-compiling aspect with the development version of deal.ii.

Here are some images of the artificial viscosity at the last timestep for which I have output.

visit0002
visit0001
visit0000

gassmoeller:
> Thank you Rene. That did work, and it's a good thing too because I killed the run and now can't seem to restart it from checkpoint files after re-compiling aspect with the development version of deal.ii.

Ah too bad. Usually checkpoints are still usable, unless something changed in the checkpoint format. I was not aware that something had changed in the development version (though now that I think about it, the fix we were looking for might have changed the format, I should have thought of that :disappointed: ).


Either way it is good to have the pictures. So it seems like you were right, locally the stabilization is still significantly bigger than the physical diffusion. Visit's color scale is a bit unfortunate, but it looks like the stabilization in up/downwellings is mostly somewhere between 15 and 80 W/mK? In earlier models I have seen the artificial stabilization drop between h^2 and h^3 in 3D models. That would suggest increasing the adaptive refinement level by one to a a level of 5 / 64 cells in radial direction would be required to reduce the stabilization to approximately the level of the physical diffusion, but only level 6 would be sufficient to get rid of any stabilization effect  for earth-like Rayleigh numbers. Well at least we know what we need now.  

> It is on my to-do list to run a case with AMR (I am thinking global refinement level 3 + 3 levels of AMR as a starting point). Any input on this setup would be welcome. My starting point would likely be to refine based on composition approximate gradient, entropy viscosity, and near the boundaries.

Sorry, I somehow missed your earlier post asking for AMR settings for the model. Below is what we used for most of our global models (e.g. the ones in Heister et al, 2017). It should reliably resolve up/downwellings, and generally resolves at least with level 4 in the upper mantle, level 3 in the lower mantle, and level 5 in the boundary layers.

```
subsection Mesh refinement
  set Initial adaptive refinement        = 1

  set Initial global refinement          = 4
  set Minimum refinement level           = 3
  set Refinement fraction                = 0.4
  set Coarsening fraction                = 0.05

  set Strategy                           = nonadiabatic temperature,minimum refinement function, composition, viscosity

  set Time steps between mesh refinement = 3

    set Function constants  =
    set Function expression = if (depth > 2500000,5,if(depth > 660000,3,if(depth > 80000,4,5)))
    set Variable names      = depth,y,z
  end
end
```

maxrudolph:
Thank you @gassmoeller. For the first timestep and initial adaptive
refinement, do you use a different refinement/coarsening fraction from the
rest of the run?

gassmoeller:
No, we just use the one that is in the parameters I posted. The model becomes quite big though, between 50 and 100 million dofs. On 1500 cores it requires about a day for 250 million years (for a compressible model though; it might be faster incompressible in particular if there is no nonlinear iteration). I attach the parameter file we used (still in the format of pre 2.0 so some section names have changed).

[global_convection_3d.prm.txt](https://github.com/geodynamics/aspect/files/3140159/global_convection_3d.prm.txt)

gassmoeller:
Max let's continue the discussion about your model in a forum post instead. Since we all agreed that this PR is at least a step forward I would like to move ahead with the merge and discuss other improvements separately.

I would like to ask for a review about these changes from someone with experience in the EV method, e.g. @tjhei, @bangerth, or @jdannberg.

source/simulator/entropy_viscosity.cc
```diff
@@ -319,6 +325,81 @@ namespace aspect
...
+                      continue;
+                    }
+
+                  scratch.face_finite_element_values->reinit (cell, face_no);
+                  (*scratch.face_finite_element_values)[introspection.extractors.velocities].get_function_values(old_solution,
+                      face_old_velocity_values);
+                  (*scratch.face_finite_element_values)[introspection.extractors.velocities].get_function_values(old_old_solution,
+                      face_old_old_velocity_values);
+
+                  // ... check if the face is an boundary with normal flow by integrating the normal velocities
```
naliboff: `a boundary`

source/simulator/entropy_viscosity.cc
```diff
@@ -319,6 +325,81 @@ namespace aspect
...
+                    {
+                      cell_at_conduction_dominated_dirichlet_boundary = false;
+                      break;
+                    }
+                }
+
+            if (cell_at_conduction_dominated_dirichlet_boundary)
+              {
+                viscosity_per_cell[cell->active_cell_index()] = 0.0;
+                continue;
```
tjhei: add comment:
// if viscosity is zero, we don't need any further computation on this cell

source/simulator/entropy_viscosity.cc
```diff
@@ -319,6 +325,81 @@ namespace aspect
...
+                                     scratch.face_finite_element_values->JxW(q);
+                      flow += ((face_old_velocity_values[q]+face_old_old_velocity_values[q])/2.0).norm() *
+                              scratch.face_finite_element_values->JxW(q);
+                      area += scratch.face_finite_element_values->JxW(q);
+                    }
+
+                  // Disable stabilization for boundaries with slow flow, or tangential flow.
+                  // Break the loop in case a face is at multiple boundaries, some with flow, some without.
+                  // In those cases we can not disable stabilization.
+                  if ((std::abs(flow/area) < std::sqrt(std::numeric_limits<double>::epsilon()) * area) ||
```
tjhei: bound should not be an area, but a velocity.

gassmoeller:
I addressed your comments.

tjhei:
I think this is a step forward, but we should discuss what the correct EV on the surface should be.

gassmoeller:
I agree, and we should also look into SUPG. But at least we now reproduce the Zhong 2008 benchmarks with a resonable resolution. Thanks!

bangerth:
All, I'll add that I appreciate all of the work that went into this PR. My bottom fell out a bit this spring and I got to spend little time in ASPECT. I enjoy working in a community in which things just move along even if I can't contribute. Thanks y'all!

Issue #2859, opened 2019-03-26
The  morency_doin.prm couldn't be execute


cgw0814:
I found the  morency_doin.prm in cookbooks couldn't be execute because of ./libmorency_doin.so doesn't exist in that file

tjhei:
You need to compile the shared library, see section "5.4.1 Running benchmarks that require code" in the manual.

tjhei:
I am closing the issue. Let us know if you need further help.

cgw0814:
I did as you said "cmake . and then make" but still run fail as following massage

![screenshot 2019-03-27 10 59 32 a m](https://user-images.githubusercontent.com/38973778/55047375-842b2700-507f-11e9-8082-49a64b38cb6a.png)

cgw0814:
There is another problem for executing benchmarks/buiter_et_al_2008_jgr/figure_6_1e20.prm

The massage appeared
`Line <153> of file <input string>: No entry with name <Start strain weakening intervals> was declared in the current subsection. `

I read the manual and found there is no variable named `Start strain weakening intervals` in the section Material model/Visco Plastic but have  `Start plasticity strain weakening intervals` and `Start prefactor strain weakening intervals`

I modified `Start strain weakening intervals`  and then the massage appeared

`An error occurred in line <998> of file </home/dealii/aspect/source/material_model/visco_plastic.cc> in function
    void aspect::MaterialModel::ViscoPlastic<dim>::parse_parameters(dealii::ParameterHandler&) [with int dim = 2]
The violated condition was: 
    this->introspection().compositional_name_exists("total_strain")
Additional information: 
    Material model visco_plastic with total strain weakening only works if there is a compositional field called total_strain.`

and then I modified the eii in the composition field into `total strain` and it could be executed.

about the information buiter_et_al_2008_jgr/figure_6_1e20.prm is less in manual @@"

I thought it should be modified in docker or github @@

naliboff:
@cgw0814 - good catch on the buiter_et_al_2008 benchmark. I'll put in a fix for that tomorrow.

tjhei:
> I did as you said "cmake . and then make" but still run fail as following massage

Can you post the commands you typed in and the output that got produced? Is there a libmorency_doin.so in the directory?

naliboff:
@cgw0814 - see pull request #2863 for updates to the Buiter et al. benchmarks. Thanks for catching those issues and let us know if you have any other questions. 

[...]

PR #2862, opened 2019-03-27
benchmarks: various fixes

tjhei:
I discovered various errors while running benchmarks. This is the first batch of fixes. It would be good if somebody could review things, as things are a bit suspicious (unused ``div_v`` for example).

PR #2863, opened 2019-03-27
Update Buiter et al. 2008 benchmarks

naliboff:
This pull request addresses part of #2859, while also improving the documentation and formatting of the `buiter_et_al_2008_jgr` benchmarks.

The key change to these benchmarks is to update the strain weakening options, which currently use options that are no longer supported (e.g., they do not run). The strain weakening has also been updated to use the plastic finite strain invariant, which is consistent with the Buiter et al. 2008 study.


tjhei:
Thanks that looks good! I am currently going through the benchmark folder fixing other problems. I don't think we need to run the tester on this, so I will merge and rebase #2862

PR #2865, opened 2019-04-03
timing sparsity pattern setup

tcclevenger:
Here was a significant section of code @tjhei and I found that wasn't being included in the timings.

I'll give an example of why this is a significant section using other setup sections that are being timed:
Ex. 3d, 4 cores, 1.2M DoFs:
Total runtime: 73.1s
Setup dof systems: 0.842s
Setup initial conditions: 2.42s
**_Setup sparsity patterns: 2.56s_**

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -629,6 +629,8 @@ namespace aspect
     // constraints. Of course we need to force assembly too.
     if (rebuild_sparsity_and_matrices)
       {
+        TimerOutput::Scope timer (computing_timer, "Setup matrices");
+
         rebuild_sparsity_and_matrices = false;
         setup_system_matrix (introspection.index_sets.system_partitioning);
         setup_system_preconditioner (introspection.index_sets.system_partitioning);


gassmoeller:
I agree that this should be included in the timing information. Have you thought about other names though? I feel `Setup sparsity patterns` will not tell the average user much (you need to know quite a bit about parallel computing before knowing what a sparsity pattern is). Do you think `Setup matrices` would be as precise (this is also more similar to the name of the functions that are actually called in that section). I am not adamant about this, just wanted to ask your opinion.

tcclevenger:
> I agree that this should be included in the timing information. Have you thought about other names though? I feel `Setup sparsity patterns` will not tell the average user much (you need to know quite a bit about parallel computing before knowing what a sparsity pattern is). Do you think `Setup matrices` would be as precise (this is also more similar to the name of the functions that are actually called in that section). I am not adamant about this, just wanted to ask your opinion.

I agree, I think `Setup matrices` is a better name. Changes are pushed.

tjhei:
agreed, this makes sense. :+1:

gassmoeller:
Ok thanks for the improvement :+1:. Changes make sense and testers are happy.

PR #2866, opened 2019-04-03
Adjust location of imposing bounds on vol fractions in composition init

class4kayaker:
The clipping on the composition quadrature based init was in the wrong place (if cells were large, then the bounds were imposed incorrectly, resulting in incorrect volume fractions). This corrects that.


gassmoeller:
This fix makes sense, but it did not change any tests. Could you add a test that uses this feature? That way we avoid any future bugs.

class4kayaker:
I have added a test case.

gassmoeller:
Thanks for adding the test.
As an outsider I do not understand how the test works though. The composition statistics reports the composition is 0 everywhere. Does that mean you chose the initial composition function to be negative and the test makes sure it is bounded to 0? But how does your fix affect this, if it was negative before it would also have been bounded. We would need a test that specifically checks that the composition is limited as 0.0 to 1.0, instead of 0.0 to cell_volume as this bound was broken before. Or do I miss how the test works?

class4kayaker:
Checking the test, the issue was I did not adjust the initial condition so that there would be some initial fluid using the specified condition. I have made an adjustment and am currently getting the corrected output.

gassmoeller:
Yes that looks more reasonable. Thanks!

bangerth:
In reference to #835.

PR #2872, opened 2019-04-08
Fix shear bands benchmark input and FPE

gassmoeller:
The shear bands benchmark had a leftover pre 2.0 input file, because the file is generated from a script and not stored in the repository. Also fixed a floating-point exception in the associated postprocessor that occurs if the analytical growth rate of the shear bands is 0.


gassmoeller:
Done. Ready from my side.

PR #2873, opened 2019-04-10
Create Jenkinsfile.cig

tjesser-ucdavis-edu:
Scaling up the CIG Jenkins service using Kubernetes has not been successful, so we're switching to standard Docker containers. This Jenkinsfile is basically the same as the k8s one, but for Docker.

### DIFF
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -1,0 +1,150 @@
+#!groovy
+
+pipeline {
+  agent {
+    docker {
+      image 'gassmoeller/aspect-tester:8.5.0'
+    }
+  }
+
+  options {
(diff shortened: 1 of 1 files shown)


tjhei:
Looks good to me. We will need to switch to a different docker image soon, but I guess that can wait.

tjhei:
@gassmoeller you are welcome to merge.

gassmoeller:
Looks good. Yes the plan was for @tjesser-ucdavis-edu to activate the new jenkins (server is already running, just need to add the webhook), then tear down the current kubernetes option, and when we switch to requiring 9.0 we can try to unify the Jenkinsfile.cig with Jenkinsfile to go back to a single file that controls both testers (yours and ours).

PR #2876, opened 2019-04-10
Use arithmetic instead of geometric averaging for determining melt region

gassmoeller:
I am rerunning the melt cookbooks for an unrelated reason, but while doing that I noticed that flagging cells as 'melt cells' is not always consistent. In particular at boundaries of melt regions with sharp gradients in porosity there are cells that are incorrectly flagged as non-melt cells, although the porosity on some of their vertices exceeds 1% (I have an example case with >20% on one vertex and essentially 0% on another vertex).

The cause of this is that the decision whether a cell is a melt cell is based on the maximum and the average darcy coefficient of a cell. Unfortunately, the average is computed as a geometric average, which means as soon as the porosity at one point in the cell is 0, the average is considered 0 and the cell is flagged as non-melt cell. We should use an arithmetic average instead (an alternative would be to set K_D to a minimum threshold instead, but we can not exclude points below a threshold, because we compare the average to a threshold to determine if the cell is a melt cell).

I have not yet updated tests and I am currently rerunning cookbooks, but I would like to ask @jdannberg and @tjhei for opinions.


tjhei:
> Unfortunately, the average is computed as a geometric average, which means as soon as the porosity at one point in the cell is 0, the average is considered 0 and the cell is flagged as non-melt cell.

Yes, that sounds like a mistake. So the decision needs to be done using the arithmetic average. As for what average to use in the computation, I don't think it matters much.

tjhei:
should this go into 2.1? It should be safe to do so...

gassmoeller:
Yes, the current behavior looks really incorrect, e.g. during the initial rise of material in the mid ocean ridge cookbook. I will run some more testing overnight, and update the tests tomorrow (mostly iteration numbers, but one crashes now, need to find out why).

gassmoeller:
We should also wait for @jdannberg's opinion. She is busy this week, so lets give it a few days.

jdannberg:
Yes, I guess that makes sense, the arithmetic average is the only thing we can use. 
In any case, we should use the same average for deciding if a cell is a melt cell and the actual computation of K_D (and this is what your current changes would do, I believe), just to be consistent. If we use different averaging schemes, the average K_D in the cell may actually turn out to be below the threshold. 

And I think we should merge this soon, because we actually spell out in the paper which averaging scheme we use (at the moment, we say it's geometric averaging), and we want to be sure about what our plan is when we resubmit.

tjhei:
> Yes, I guess that makes sense, the arithmetic average is the only thing we can use.

well then, let's go ahead with this change.

This changes quite a few test results. Many only seem to have minor differences, while some are quite different:
```
--- a/tests/average_density_boundary_fluid_pressure/screen-output
+++ b/tests/average_density_boundary_fluid_pressure/screen-output
-     Mass fluxes through boundary parts: 0 kg/yr, -12.37 kg/yr, 10.82 kg/yr, 0 kg/yr
+     Mass fluxes through boundary parts: 0 kg/yr, -89.22 kg/yr, 85.73 kg/yr, 0 kg/yr
```

gassmoeller:
> In any case, we should use the same average for deciding if a cell is a melt cell and the actual computation of K_D (and this is what your current changes would do, I believe), just to be consistent.

I think this happens automatically, because `p_c_scale` is the function that is called both to determine if a cell is a melt cell, and to compute the scaling factor.

> This changes quite a few test results. Many only seem to have minor differences, while some are quite different.

Yes I am not surprised. I checked the `average_density_boundary_fluid_pressure` test (the first one that shows the significant differences) and it has exactly this problem. In the first timestep there are lots of cells that have a porosity of 2% but are flagged as non-melt cells. The new version fixes that. Also note that the difference becomes smaller in later timesteps. I think these changes are consistent and the new version is working.

tjhei:
> The new version fixes that.

great!

Issue #2314, opened 2018-06-19
cotire errors, warnings and slow compile with gcc 7.3.0


bobmyhill:
With precompiled headers enabled (`ASPECT_PRECOMPILE_HEADERS:BOOL=ON` in CMakeCache.txt), cotire no longer appears to work correctly (tested on machines of @gassmoeller, @jdannberg and myself). 

When CMakeFiles/ doesn't exist (i.e. on first compile), the following error appears
`CMake Error in CMakeLists.txt:`
`  Evaluation file to be written multiple times with different content.  This`
`  is generally caused by the content evaluating the configuration type,`
`  language, or location of object files:`

`   /home/bob/software/aspect_mainbranch/aspect_CXX_Debug_cotire.cmake`

After the first compile, compilation doesn't result in an error, but building every CXX object results in the following warning:
`cc1plus: warning: /home/bob/software/aspect/cotire/aspect_CXX_prefix.hxx.gch: created with -gnone, but used with -gdwarf-2`

Compilation is very slow.

tjhei:
what is your ``cmake --version`` ?

bobmyhill:
cmake version 3.10.2

tjhei:
cmake 3.11 with gcc 5.4 works for me. So is this related to your gcc version?

tjhei:
@gassmoeller are you going to try a new cotire version?

bobmyhill:
@tjhei I already tried the new cotire version; it didn't solve the problems. I assume this is an issue with gcc; it's difficult for me to check whether cmake 3.11 fixes the issue, as that would require I uninstall a lot of other software (including fenics) that I'd rather not reinstall.

bangerth:
I don't think I understand how the `CMakeFiles` can or can not exist. It's a result of running cmake, so it should always be there.

tjhei:
I get a different and funny error with gcc 7.3 and cmake 3.11:
```
error: ‘::fenv_t’ has not been declared
```
only if cotire is on.

bangerth:
For the latter, you're probably missing an `#include` file somewhere. Which file is this for?

tjhei:
helper_functions.cc but I assume this is cotire's fault.

bangerth:
Can you show the whole error message?

bangerth:
?? The error comes out of a system header? And I don't see `helper_functions.cc` making use of fp exceptions to begin with. I think you'll have to look into the file being compiled at the time (the combined one).

[...]

PR #2877, opened 2019-04-10
Fix and update cotire

gassmoeller:
This is a workaround for #2314. It updates cotire to the latest release, and disables it for cmake version >= 3.9 due to the errors with the generator expressions that are used in deal.II as discussed in that issue. If we fix the issues otherwise we can again enable cotire for later cmake versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -396,9 +396,8 @@ ENDIF()
 # machines, therefore it is disabled by default. Also cotire currently
 # has issues with cmake generator expressions used in deal.II for cmake
 # version 3.9 and above. Disable cotire for cmake >= 3.9.
+SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "Precompile external header files using cotire to speedup compile time. Currently only supported for cmake versions older than 3.9.")
 IF(CMAKE_VERSION VERSION_LESS 3.9)
-  SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "Precompile external header files using cotire to speedup compile time.")
-
   IF (ASPECT_PRECOMPILE_HEADERS)
     # For the unity build we need to compile helper_functions.cc before core.cc
     # ensure this by reordering the files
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -393,15 +393,19 @@ ENDIF()
 
 # Check if we want to use cotire to precompile header files and prepare a
 # unity build target. This speeds up compile time, but can fail on some
-# machines, therefore it is disabled by default.
-SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "Precompile external header files using cotire to speedup compile time.")
```
tjhei: I think you should always define this variable and force it to off, like this:
```
SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "Precompile external header files using cotire to speedup compile time.")
IF(CMAKE_VERSION VERSION_LESS 3.9)
... existing code
ELSE()
  SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "" FORCE)
ENDIF()
```

CMakeLists.txt
```diff
@@ -486,7 +490,7 @@ IF (ASPECT_USE_FP_EXCEPTIONS)
   ENDIF()
 ENDIF()
 
-IF (ASPECT_PRECOMPILE_HEADERS)
+IF (CMAKE_VERSION VERSION_LESS 3.9 AND ASPECT_PRECOMPILE_HEADERS)
```
tjhei: then you don't need this change

tjhei:
This should go into 2.1...

gassmoeller:
Like this? I added it to the description of the parameter to avoid confusion why this variable keeps switching itself off.

tjhei:
> I added it to the description of the parameter to avoid confusion why this variable keeps switching itself off.

Yes, this is what I thought. You make a good point: we could also report an error if somebody enables cotire and has a cmake version that is too new. 
```
SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "Precompile external header files using cotire to speedup compile time. Currently only supported for cmake versions older than 3.9.")
￼	IF(CMAKE_VERSION VERSION_LESS 3.9)
￼	  IF (ASPECT_PRECOMPILE_HEADERS)
￼	    # For the unity build we need to compile helper_functions.cc before core.cc
￼	    # ensure this by reordering the files
￼	    FIND_FILE(HELPER_PATH helper_functions.cc HINTS "${CMAKE_SOURCE_DIR}/source/simulator")
￼	    LIST(REMOVE_ITEM TARGET_SRC ${HELPER_PATH})
￼	    LIST(INSERT TARGET_SRC 0 ${HELPER_PATH})
￼	  ENDIF()
￼	ELSE()
￼	  MESSAGE(FATAL "Sorry, you need CMake version 3.9 or less for this feature to work.")
￼	ENDIF()
```
I leave this to you.

tjhei:
Looks good. Ready to merge when the testers are happy.

PR #2879, opened 2019-04-11
 clang-tidy: more checks

tjhei:
fix a couple of clang-tidy performance complaints (missing references) reported with checks:
performance-unnecessary-value-param
performance-unnecessary-copy-initialization

PR #2880, opened 2019-04-11
update clang-tidy settings

tjhei:
- enable other checks
- update instructions

see #2878 and #2879


tjhei:
> OK to merge but I'm not sure whether this should happen before the other two have been merged.

doesn't matter. We are not running clang-tidy automatically right now.

PR #2881, opened 2019-04-11
Replace std::shared_ptr by std::unique_ptr.

bangerth:
Obvious for the most part. What isn't has gotten a couple of lines more comments.


bangerth:
In reference to #2811.

bangerth:
Also #2375.

PR #2882, opened 2019-04-11
Replace std::shared_ptr by std::unique_ptr in the heating models.

bangerth:
Strictly speaking, this is an incompatible change because the heating model manager
exports its list of heating models. I suspect that few plugins we don't have
in the tree ourselves (which I have all converted) make use of this.

Related to #2375 and #2811.

### DIFF
--- a/benchmarks/tangurnis/code/tangurnis.cc
+++ b/benchmarks/tangurnis/code/tangurnis.cc
@@ -400,7 +400,7 @@ namespace aspect
 
     std::vector<std::vector<double> > composition_values (this->n_compositional_fields(),std::vector<double> (n_q_points));
 
-    const std::list<std::shared_ptr<HeatingModel::Interface<dim> > > &heating_model_objects = this->get_heating_model_manager().get_active_heating_models();
+    const std::list<std::unique_ptr<HeatingModel::Interface<dim> > > &heating_model_objects = this->get_heating_model_manager().get_active_heating_models();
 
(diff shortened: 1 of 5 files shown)


tjhei:
```
The following tests FAILED:
	312 - material_model_dependencies_tangurnis (Failed)
	584 - tangurnis_ba (Failed)
	585 - tangurnis_ba_custom (Failed)
	586 - tangurnis_tala (Failed)
	587 - tangurnis_tala_c (Failed)
	588 - tangurnis_tala_implicit (Failed)
``
should we delay this change until after 2.1? I am more happy to break things then.

bangerth:
Let's try this again. 
/rebuild

It's ok with me if we want to postpone this until after the release. I realize that it's an incompatible change. Anyone got an opinion on a changelog entry?

gassmoeller:
Unless you have checked all benchmarks and cookbooks I would vote for postponing this after the release. Who knows what hidden plugins rely on this (ideally we should have a test for all cookbooks and benchmarks of course).

gassmoeller:
Since we branched off the release we can merge this. About the changelog entry: Do we require changelog entries that change the API, or only entries if we change the parameter files? I do not think we had many entries with interface changes, so maybe we just go ahead and merge this? Users who found and used this function can likely find it again and change the type.

bangerth:
Let me know what you prefer and I'll do it. I don't feel strongly.

gassmoeller:
Lets just merge it. I think the function is not used widely.

PR #2887, opened 2019-04-14
clang-tidy: suppress warnings

tjhei:
Suppress the three locations where we want to ignore performance-
unnecessary-value-param.

followup to #2879 and required for clang-tidy to pass again.


tjhei:
This is a bit ugly, but I don't mind it too much as this will remind us why we pass by value here.

gassmoeller:
I agree.

PR #2890, opened 2019-04-15
Use smoother language.

bangerth:
The current text says 'common...frequently...frequently' in two successive
sentences.

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -1640,7 +1640,7 @@ give a detailed description and rationale for the various building blocks.
 \label{sec:approximate-equations}
 
 There are a number of common variations to equations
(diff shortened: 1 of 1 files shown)


doc/manual/manual.tex
```diff
@@ -1638,7 +1638,7 @@ \subsection{Approximate equations}
 \label{sec:approximate-equations}
 
 There are a number of common variations to equations
-\eqref{eq:stokes-1}--\eqref{eq:temperature} that are frequently used in the
+\eqref{eq:stokes-1}--\eqref{eq:temperature} that used in the
```
tjhei: missing "are"? Not very smooth as is. :-)

bangerth:
Ah, good grief. How about this?

PR #2893, opened 2019-04-17
Update world builder with fix against warnings in gcc 5.

MFraters:
fix to issue https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/118 as discussed in #2613.


tjhei:
:+1: I will merge and then test here, thanks!

PR #2895, opened 2019-04-21
cmake: switch to CXX only project

tjhei:
Declare the project as a CXX only project and change one of the tests to
be a c++ check (instead of a C check). This change will:
- not check the existence of a c compiler (speed up cmake!)
- stop printing confusing messages (finding gcc for the C language,
while we compile with intel for CXX for example)
- actually perform the include check with CXX


gassmoeller:
Looks good to me. Thanks!

PR #2875, opened 2019-04-10
Post release 2.1.0 version bump

tjhei:
this can be merged after we create the branch for 2.1, but it can be reviewed already.

Issue #2867, opened 2019-04-05
Release 2.1.0 planning


tjhei:
I discussed with @gassmoeller yesterday, that we should try to make a release in the near future and then switch to requiring deal.II 9.0 (#2738). This would allow us to iron out issues and help transitioning to the new version for the participants of the hackathon in May. Dragging it out to after the hackathon sounds like a bad idea and we need to make sure the testers are running...

Things we need to do or look at before the release:
- [x] ~~#1455~~
- [x] test on newest gcc, clang
- [x] test on intel
- [x] compile with deal.II master, 9.0, 8.5
- [x] cotire: #2314
- anything else?

gassmoeller:
We need to take a look at the 2.1 milestone issues. I think many of them are not critical for 2.1. though.
https://github.com/geodynamics/aspect/milestone/5

tjhei:
I looked through all of them and none are urgent and some require deal.II 9.0, so they have to wait.

tjhei:
master builds cleanly with:
- deal.II 8.5.1, 9.0.1, master
- gcc 5.4, clang 6, gcc 8.3, clang 8
- PETSc works
- clang-tidy

TODO: intel compiler

otherwise, I think we are good to go...

tjhei:
I noticed we get compile warnings with the world builder: https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/118

tjhei:
rc1 in case you want to try it out: https://github.com/geodynamics/aspect/releases/tag/v2.1.0-rc1

bangerth:
Not related to the release, but I think we ought to just require the current deal.II dev version. Or maybe wait for deal.II 9.1 (hopefully not all that far in the future either) and then require that. People will have to update anyway, might as well have to update to 9.1.

tjhei:
> I think we ought to just require the current deal.II dev version.

The hackathon will be before the 9.1 release and I would prefer to have the CI testers run with a release instead of a dev version. What is in the current dev version that is not in 9.0?
My suggestion:
- release 2.1.0 (branch is already created)
- switch to require 9.0 and switch CI machines (before the hackathon)
- during the hackathon: help people transition, clean up the code base (remove 8.5 hacks)
Then, before we release the next version of ASPECT we can decide to require 9.1 instead. This way, we move away from 8.5 soon and we never require a dev version for an ASPECT release. Thoughts?

gassmoeller:
> Thoughts?

[...]

PR #2896, opened 2019-04-24
Add release announcement

gassmoeller:
This is a tentative release announcement. I filtered the changelog for changes important to users, and shortened some entries. Let me know if you think anything important is missing.

### DIFF
--- a/doc/modules/to-2.1.0.h
+++ b/doc/modules/to-2.1.0.h
@@ -205,6 +205,11 @@
  * <br>
  * (Menno Fraters and Rene Gassmoeller, 2018/06/26)
  *
+ * <li> New: Created a new plugin system that allows it to prescribe a fixed
+ * heat flux (instead of prescribing the temperature) at the model boundaries.
+ * <br>
+ * (Juliane Dannberg and Rene Gassmoeller, 2018/06/25)
+ *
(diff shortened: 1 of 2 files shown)


gassmoeller:
Related to #2867.

MFraters:
Isn't this also the first release which includes the geodynamic world builder?

gassmoeller:
Hah, very good point. But you never added a changelog entry, that is why it did not appear in the list of changes. Can you write a changelog entry? We need to figure out how to best add this, as I just merged #2875, which deleted all of the old changelog entries. Maybe add it directly to `doc/modules/to-2.1.0.h` and `doc/release-tasklist` and do not create a file in `doc/modules/changes`.

jdannberg:
Seems like we also forget the changelog entry for something else:
a new plugin system that allows it to prescribe a fixed heat flux (as an alternative to prescribing the temperature) at the model boundaries

gassmoeller:
Ok, heat flux boundary conditions are included. Let me know in case you find anything else. Otherwise this is good to go.

tjhei:
Looks good, thanks! I will merge so I can cherry-pick the change to the 2.1 branch. We can edit the changelog in a followup if needed.

PR #2898, opened 2019-04-25
check constraints are consistent

tjhei:
extend on #2780 (and might detect situations where things break)


tjhei:
we should switch to 9.0 first to make sure all tests pass

tjhei:
update: rebased.

gassmoeller:
Looks good.

PR #2900, opened 2019-04-29
update DOI info

tjhei:


gassmoeller:
This just changes documentation, one tester succeeded and the other aborted because of a timeout. I will change the timeout in a follow up, but in the meantime this is good to go. Thanks for the release work! :+1:

PR #2903, opened 2019-04-29
no temperature matrix for Stokes only

tcclevenger:
We are still reserving memory for the temperature matrix even when it is not needed. Adding an if-statement that ignores the Stokes-only case.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -905,7 +905,12 @@ namespace aspect
               coupling[x.pressure][x.velocities[d]] = DoFTools::always;
             }
         }
-      coupling[x.temperature][x.temperature] = DoFTools::always;
+      // Do not allocate a temperature matrix if no temperature
+      // solves are going to be performed.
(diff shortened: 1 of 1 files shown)


source/simulator/core.cc
```diff
@@ -905,7 +905,8 @@ namespace aspect
               coupling[x.pressure][x.velocities[d]] = DoFTools::always;
             }
         }
-      coupling[x.temperature][x.temperature] = DoFTools::always;
+      if (!(parameters.nonlinear_solver == NonlinearSolver::Kind::no_Advection_iterated_Stokes))
```
tjhei: can you leave a comment? Maybe something like "Do not allocate a temperature matrix if no temperature solves are going to be performed"?

gassmoeller: Looks good, but could you also include `NonlinearSolver::Kind::no_Advection_no_Stokes`? I know that one would need additional adjustments elsewhere, but lets start here. Otherwise ready.

tcclevenger: Done.

gassmoeller:
Thanks!

PR #2906, opened 2019-04-30
Jenkins update

gassmoeller:
Some maintenance on the CIG Jenkinsfile's. Some of our newly added nodes are somewhat slow (but numerous :-)), so we need to give them a bit more time to finish (usually they finish in 90 minutes, but one was a bit slower). These nodes are used with less priority, so the fast nodes get used as much as possible for quicker turnaround. Also removed the old Kubernetes Jenkinsfile.

Finally, added a line for workspace cleanup. @tjhei: After adding this line (`cleanWs()`) did you ever have problems about Jenkins-VMs filling up their disk space? Or did this resolve the issue?


Jenkinsfile.cig
```diff
@@ -132,6 +132,9 @@ pipeline {
           sh 'git diff tests > changes-test-results.diff'
           archiveArtifacts artifacts: 'changes-test-results.diff', fingerprint: true
         }
+        cleanup {
+          cleanWs()
+        }
```
tjesser-ucdavis-edu: I wonder if which scope this is in makes a difference. My understanding of what Jenkins is doing is that:

1. The Jenkins server receives the webhook.
2. The server tells an agent to handle the build.
3. The agent pulls the PR repo.
4. And launches the build container, copying the repo into the container (creating the container workspace).
5. The build container follows the build instructions.
6. The agent grabs the whole container workspace (taking up a lot of space on the agent).
7. And then removes the build container.
8. The agent tells the server the result.
9. The server posts the result.

I think this is a reasonable approximation. When Rene and I examined the agent nodes, there were no leftover build containers, just large volumes attached to the agent.

My question then is, is that post step being run in the container or in the agent?

gassmoeller: Did we agree in our discussion that this is the right way forward? If so, do you think this is ok to merge?

tjesser-ucdavis-edu: Yes, when we checked, it had the desired effect.

tjhei:
> did you ever have problems about Jenkins-VMs filling up their disk space? Or did this resolve the issue?

unless the machine crashes, I have not encountered problems. Note that you can also go to the job configuration and select "Orphaned Item Strategy" and enter a small number there.

Jenkinsfile.cig
tjhei: I think this should be a "post" step in the top level "pipeline" section, otherwise it only runs if you make it to that step. See my Jenkinsfile...

gassmoeller: Good catch. That was my intention, but I confused myself with the existing post section of the test stage. Fixed.

gassmoeller:
> Note that you can also go to the job configuration and select "Orphaned Item Strategy" and enter a small number there.
Good point. @tjesser-ucdavis-edu could you check that for our server?
Otherwise this is ready from my side once the tester agrees.

tjesser-ucdavis-edu:
The Aspect Jenkins configuration is currently set to discard all orphaned items.

gassmoeller:
Ok, that should to it. Ready from my side.

PR #2908, opened 2019-05-02
Convert more shared_ptr -> unique_ptr.

bangerth:
Related to #2375 and #2811

### DIFF
--- a/include/aspect/boundary_temperature/interface.h
+++ b/include/aspect/boundary_temperature/interface.h
@@ -247,7 +247,7 @@ namespace aspect
          * Return a list of pointers to all boundary temperature models
          * currently used in the computation, as specified in the input file.
          */
-        const std::vector<std::shared_ptr<Interface<dim> > > &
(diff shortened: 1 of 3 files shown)


include/aspect/boundary_temperature/interface.h
```diff
@@ -247,7 +247,7 @@ namespace aspect
          * Return a list of pointers to all boundary temperature models
          * currently used in the computation, as specified in the input file.
          */
-        const std::vector<std::shared_ptr<Interface<dim> > > &
+        const std::vector<std::unique_ptr<Interface<dim> > > &
```
gassmoeller: Did you also change the type of the `boundary_temperature_objects` member?

bangerth: Yes -- that's in line 325 in the next part of the diff :-)

gassmoeller: No idea what I was thinking when doing this review :smile:, I was sure I checked your whole diff. Anyway, good to go then.

Issue #1671, opened 2017-05-15
Residual velocities in Chunk Geometry model


gassmoeller:
There are a number of superseded pull requests and issues that describe this problem, but since a part of it is addressed I closed the other issues and opened a new clean discussion about the remaining problem.

To summarize the problem and our findings:
Tangential velocity boundaries of the Chunk geometry model show residual velocities even without driving forces that are higher than what we expect (~1cm/year for 8 cell resolution in depth). This only occurs on curved boundaries (i.e. not the longitude boundaries), and converges to 0 with increased resolution. They also:
- scale with radius (for a nondimensional radius of 1 they are 1e-16)
- rapidly decay with increased finite element degree
- do not vanish if we internally scale the radius of the chart coordinate system of the manifold (i.e. they are created by something in the real space, not by the manifold)
- scale with gravity, therefore they are likely related to the absolute pressure in the model

Related to:
#822 
#1016 
#1018 
#1361
#1533

bangerth:
@gassmoeller and @anne-glerum : Can you attach an input file here for easy reproducibility?

gassmoeller:
Added the file.
- Also the residual velocities do not scale with linear solver tolerence

PR #2910, opened 2019-05-03
Create anisotropic quadrature for lateral averaging

gassmoeller:
This is a continuation of the speedup of the lateral averaging procedures (see #2136), in response to an email from @maxrudolph that showed that running the depth_average postprocessor in 3d for every 17 timesteps consumes about 70% of the compute time.

The lateral averaging so far used a very high resolution quadrature formula with 1000 quadrature points per cell (10x10x10, 3D) to ensure that even coarse cells contribute to their depth layers. We only really need this resolution in depth though, laterally a normal quadrature degree is sufficient. Therefore an anisotropic quadrature with higher resolution in depth direction can save a lot of compute time. Previously I thought this is not possible, because we do not know if all cells have the same dimension in the unit coordinate system pointing in depth direction. Turns out most of our geometry models create the mesh so that the same direction points in depth direction for every cell (in particular box and spherical shell, our two most common geometries). I created a postprocessor to check this (output unit coordinates for every vertex of every cell), but I think it is not worth having that in the main version. For anyone interested, it is in this branch: https://github.com/gassmoeller/aspect/tree/add_unit_coordinates_postprocessor.

In the current form the postprocessor is faster by a factor of 3 in 2d and 10 in 3d, but if I am correct in this discussion: https://community.geodynamics.org/t/appropriate-quadrature-degree-for-postprocessors/609, we could gain another factor of 2 in 3D (or 50% in 2D) by reducing the lateral quadrature degree further. 

I expect a lot of test results to change, but I have verified that for very high lateral quadrature degrees the tests remain essentially unchanged (= the new version is also correct, just less expensive, and slightly less accurate for the lateral integrals).

I still need to take a look at the test results, but otherwise this is ready for review.


source/simulator/lateral_averaging.cc
```diff
@@ -312,22 +360,54 @@ namespace aspect
...
+                                                           Plugins::plugin_type_matches<GeometryModel::Chunk<dim> >(this->get_geometry_model()) ||
+                                                           Plugins::plugin_type_matches<GeometryModel::EllipsoidalChunk<dim> >(this->get_geometry_model()) ||
+                                                           Plugins::plugin_type_matches<GeometryModel::SphericalShell<dim> >(this->get_geometry_model()) ||
+                                                           Plugins::plugin_type_matches<GeometryModel::TwoMergedBoxes<dim> >(this->get_geometry_model());
+
+    unsigned int depth_direction = numbers::invalid_unsigned_int;
+
+    // The Chunk geometry model has depth as first dimension (radius, lon, lat),
+    // all others with unique direction have it last.
+    if (geometry_model_has_unique_depth_direction)
```
tjhei: you could remove this if and assign to depth_direction above unconditionally.

source/simulator/lateral_averaging.cc
```diff
@@ -312,22 +360,54 @@ namespace aspect
...
+                          dim;
+      }
+
+    const unsigned int max_fe_degree = std::max(this->introspection().polynomial_degree.velocities,
+                                                std::max(this->introspection().polynomial_degree.temperature,
+                                                         this->introspection().polynomial_degree.compositional_fields));
+
+    std::unique_ptr<Quadrature<dim> > quadrature_formula;
+    if (geometry_model_has_unique_depth_direction)
+      quadrature_formula = std::make_unique<Quadrature<dim> >(internal::get_quadrature_formula<dim>(max_fe_degree+1,
```
tjhei: can you please leave a comment for the ``max_fe_degree+1`` logic? I assume that the +1 is also quite conservative, but maybe we should leave it here (curved geometry).

gassmoeller:
Thanks, please wait until I come back to you, something funny is happening with the Chunk and possibly EllipsoidalChunk geometries. I need to make a test first, and if that confirms the problems, I will leave them at the old state for now.

tjhei:
> something funny is happening with the Chunk and possibly EllipsoidalChunk geometries.

Thinking about this: we could pick the lateral degree based on the type of geometry as well. I would assume that a box geometry (without curvature) requires fewer points.

Finally: what about free surface computations? Do we need to be more careful there as well?

gassmoeller:
> Thinking about this: we could pick the lateral degree based on the type of geometry as well. 
> Finally: what about free surface computations? Do we need to be more careful there as well?

Both true, that might be interesting to investigate at some point. Sounds like a bigger project to me though, because I have no intuition for which degree would be appropriate. In the current form and if I understand the discussion in https://community.geodynamics.org/t/appropriate-quadrature-degree-for-postprocessors/609/2 correctly, the error of the current quadrature converges with the same order as the FE degree for arbitrary geometries (the error constant might be influenced by the geometry though), so it should be sufficient. Nevertheless, for `Chunk` I saw significant changes in the output, and for `EllipsoidalChunk` we do not have a test that is why I left them out of the changes for now (I do not have the time for getting to the root of that problem, but it might be related to #1671).

Do you think the current state is good enough for a merge?

tjhei:
> Do you think the current state is good enough for a merge?

Yes, except depth_average_rounding_bug is failing right now

gassmoeller:
Fixed, I messed up the update of the test results.

source/simulator/lateral_averaging.cc
```diff
@@ -312,22 +360,55 @@ namespace aspect
                                              std::vector<double>(n_slices,0.0));
     std::vector<double> volume(n_slices,0.0);
 
-    // this yields 10^dim quadrature points evenly distributed in the interior of the cell.
-    // We avoid points on the faces, as they would be counted more than once.
-    const QIterated<dim> quadrature_formula (QMidpoint<1>(),
-                                             10);
-    const unsigned int n_q_points = quadrature_formula.size();
+    // We would like to use a quadrature formula that is appropriately accurate laterally,
+    // but has a high resolution in depth (to populate all depth slices for the averages).
```
naliboff: Consider adding in a brief explanation of why there could be more than one depth slice in a cell (AMR).

doc/modules/changes/20190505_gassmoeller
```diff
@@ -0,0 +1,7 @@
+Changed: The lateral averaging functions, e.g. used in the 'depth average'
+postprocessor were optimized for speed. Their lateral averaging is now slightly
+less accurate (the difference is typically on the order of 1e-3 at the coarsest
```
naliboff: `less accurate` ... `for specific geometry models (Spherical Shell, Box, Two Merged Boxes)`

gassmoeller:
I resolved the conflict and added the requested comments.

naliboff:
@gassmoeller - Thanks for adding in the new comments. One test is still failing, but after that is fixed this is good to go.

gassmoeller:
Yes I messed up the rebase. Fixed now.

PR #2914, opened 2019-05-06
[DoNotMerge] Keep build container /home in memory only

tjesser-ucdavis-edu:
This Jenkinsfile change will put the build container's `/home` directory in a tmpfs partition, effectively making the build occur only in memory. This is to try and narrow down the reason why Timo's server is building so much faster than CIG's servers.

### DIFF
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -4,6 +4,8 @@ pipeline {
   agent {
     docker {
       image 'tjhei/dealii:v9.0.1-full-v9.0.1-r5-gcc5'
+      label 'tmpfs_capable'
+      args '--mount type=tmpfs,destination=/home'
     }
   }
 


tjhei:
what kind of hardware are you running on and what is your software setup on the builder?

tjesser-ucdavis-edu:
__ | projector | singleb
-- | --------- | -------
CPU | 4-core Intel i5 | 4-core 8-thread Intel Xeon
Memory | 32GB | 64GB
Disk | 256GB SSD | 4TB 2 HDD RAID1

I used these scripts to set up the server and agents: https://github.com/tjesser-ucdavis-edu/jenkins-static-container-scripts

tjhei:
4 cores is not a lot.
Additionally, I assume you run docker natively on the machine?

Can you check ``docker info | grep Storage``? I found ``aufs`` to be a lot slower compared to ``overlay2``. See https://docs.docker.com/storage/storagedriver/overlayfs-driver/

tjesser-ucdavis-edu:
Yep, both are base hardware, not VMs.

Both also return `Storage Driver: overlay2`.

gassmoeller:
We made some progress on the difference in tester runtime. In general the test runtimes between running the tester container on my laptop (6 cores/12 threads) and the projector machine (4cores/4threads) are within the expected differences (~30-50%), but the `periodic_box_mpi10` is incredibly slow if the machine has less than 10 threads available. We tested that this is independent of Jenkins, Ninja or ctest. Even if I run the test manually on the projector with `mpirun -np 4 ...prm` It finished in 0.25 secs, while with `mpirun -np 5 ...` it takes several minutes. Coincidentally, on my laptop the change occurs from 12 to 13 MPI processes. We suspect this is an issue with the switch from openmpi to mpich for the test container, maybe MPICH behaves differently if the number of MPI processes is bigger than the available threads. If that is the case (final tests running now), we should look into options that can improve the MPI behavior for oversubscribed jobs.

tjhei:
This could be related to https://wiki.mpich.org/mpich/index.php/Frequently_Asked_Questions#Q:_Why_does_my_MPI_program_run_much_slower_when_I_use_more_processes.3F yes.

gassmoeller:
Since #2919 was merged this can be closed. @thei: Yes that FAQ entry seems related. Oversubscribing MPI jobs seems to get harder these days, my OpenMPI requires a flag to allow it (otherwise it crashes), and MPICH is just unusably slow. Might be a problem if someone with an older dual-core wants to run our test suite.

PR #2917, opened 2019-05-07
Fix unity build

gassmoeller:
The unity build functionality was broken due to a namespace that was only unique in the non unity build. Fix that, and also add a test stage for the unity build. This is currently just to compare the two stages, we could also decide to switch the CIG tester over to the unity build altogether if it saves a bit of time.

### DIFF
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -68,17 +68,14 @@ pipeline {
 
           # Set up build system and compile ASPECT
           cmake \
-            -G 'Ninja' \
-            -D ASPECT_TEST_GENERATOR='Ninja' \
             -D ASPECT_USE_PETSC='OFF' \
-            -D ASPECT_RUN_ALL_TESTS='ON' \
             -D ASPECT_PRECOMPILE_HEADERS='ON' \
             ..
         '''
 
         sh '''
           cd build-gcc-unity
-          ninja aspect_unity
+          make aspect_unity
         '''
       }
     }
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -56,6 +56,33 @@ pipeline {
       }
     }
 
+    stage('Unity Build') {
+      options {
+        timeout(time: 30, unit: 'MINUTES')
+      }
+      steps {
+        sh 'mkdir build-gcc-unity'
+
+        sh '''
+          cd build-gcc-unity
+
+          # Set up build system and compile ASPECT
+          cmake \
+            -G 'Ninja' \
+            -D ASPECT_TEST_GENERATOR='Ninja' \
+            -D ASPECT_USE_PETSC='OFF' \
+            -D ASPECT_RUN_ALL_TESTS='ON' \
+            -D ASPECT_PRECOMPILE_HEADERS='ON' \
+            ..
+        '''
+
+        sh '''
+          cd build-gcc-unity
+          ninja aspect_unity
+        '''
+      }
+    }
+
     stage('Build') {
       options {
         timeout(time: 30, unit: 'MINUTES')
--- a/source/postprocess/interface.cc
+++ b/source/postprocess/interface.cc
@@ -236,7 +236,7 @@ namespace aspect
                      "all") != postprocessor_names.end())
         {
           postprocessor_names.clear();
-          for (typename std::list<typename internal::Plugins::PluginList<Interface<dim> >::PluginInfo>::const_iterator
+          for (typename std::list<typename aspect::internal::Plugins::PluginList<Interface<dim> >::PluginInfo>::const_iterator
                p = std::get<dim>(registered_plugins).plugins->begin();
                p != std::get<dim>(registered_plugins).plugins->end(); ++p)
             postprocessor_names.push_back (std::get<0>(*p));


tjhei:
I don't understand. I thought the other builds already use ``ASPECT_PRECOMPILE_HEADERS=ON`` ?!?

gassmoeller:
`ASPECT_PRECOMPILE_HEADERS` only precompiles all header files outside of ASPECT into a single `.gch` compiled header that gets included for every .cc file. `make aspect_unity` additionally concatenates all .cc files into a single (or just a few, e.g. number of cores) files that need to be compiled. This way the even the duplication of including the .gch 240 times gets removed, and I suppose the compiler can also work more efficiently if most of the code is in a single file. The build time on the tester using precompiled headers is 3min25s, while creating a unity build with only as many .cc files as the number of cores brings down the compile time to 1min25s. The disadvantage of the unity build is that you need to recompile the whole project no matter which file you change (but that is not relevant for the tester).

tjhei:
> `make aspect_unity` additionally concatenates all .cc

huh, this is the first time I learn about this target. :-)

tjhei:
> while creating a unity build with only as many .cc files as the number of cores

Does this mean that the compile can also break depending on how many cores your machine has?

gassmoeller:
> huh, this is the first time I learn about this target. :-)

It is created by cotire only if `ASPECT_PRECOMPILE_HEADERS` is on.

> Does this mean that the compile can also break depending on how many cores your machine has?

It is possible (though unlikely) to have a code that compiles correctly with `ninja -j6 aspect_unity`, but fails while compiling with `ninja -j1 aspect_unity`. That happens if the conflicting parts of the code are split between the concatenated .cc files. Maybe we should run a serial unity build to be sure (i.e. `ninja -j1 aspect_unity`).

The one problem the unity build has currently is that it triggers a bunch of warnings, unused argument warnings in AdolC, and deprecation warnings in deal.II, of the sort:

```
/home/rene/software/deal.II-v9.0.1/include/deal.II/base/utilities.h: At global scope:
/home/rene/software/deal.II-v9.0.1/include/deal.II/base/utilities.h:332:33: warning: ‘template<int a, int N> struct dealii::Utilities::fixed_int_power’ is deprecated [-Wdeprecated-declarations]
     static const int value = a *fixed_int_power<a,N-1>::value;
                                 ^~~~~~~~~~~~~~~
/home/rene/software/deal.II-v9.0.1/include/deal.II/base/utilities.h:330:29: note: declared here
   struct DEAL_II_DEPRECATED fixed_int_power
                             ^~~~~~~~~~~~~~~
```

and a lot of:
```
In file included from /home/rene/software/aspect/source/simulator/nullspace.cc:33,
                 from /home/rene/software/aspect/debug/cotire/aspect_CXX_unity.cxx:2598:
/home/rene/software/deal.II-v9.0.1/include/deal.II/lac/pointer_matrix.h: At global scope:
/home/rene/software/deal.II-v9.0.1/include/deal.II/lac/pointer_matrix.h:117:49: warning: ‘template<class VectorType> class dealii::PointerMatrixBase’ is deprecated [-Wdeprecated-declarations]
 class DEAL_II_DEPRECATED PointerMatrix : public PointerMatrixBase<VectorType>
                                                 ^~~~~~~~~~~~~~~~~
/home/rene/software/deal.II-v9.0.1/include/deal.II/lac/pointer_matrix.h:53:26: note: declared here
 class DEAL_II_DEPRECATED PointerMatrixBase : public Subscriptor
                          ^~~~~~~~~~~~~~~~~
/home/rene/software/deal.II-v9.0.1/include/deal.II/lac/pointer_matrix.h:198:33: warning: ‘template<class MatrixType, class VectorType> class dealii::PointerMatrix’ is deprecated [-Wdeprecated-declarations]
   SmartPointer<const MatrixType,PointerMatrix<MatrixType,VectorType> > m;
                                 ^~~~~~~~~~~~~
/home/rene/software/deal.II-v9.0.1/include/deal.II/lac/pointer_matrix.h:117:26: note: declared here
 class DEAL_II_DEPRECATED PointerMatrix : public PointerMatrixBase<VectorType>
                          ^~~~~~~~~~~~~
```

tjhei:
> Maybe we should run a serial unity build to be sure (i.e. `ninja -j1 aspect_unity`).

Isn't this quite slow?

> The one problem the unity build has currently is that it triggers a bunch of warnings

:-(

gassmoeller:
The serial unity build takes about as much time as the parallel build with precompiled headers on a 4core/8thread machine (3m40s vs 3m20s). In so far it is acceptable. However, the warnings prevent us from using it instead of the current build for the tester. I am currently not sure what to do best. I would like to have it included somewhere in the tests so that at least we do not break it accidentally.

tjhei:
> I am currently not sure what to do best. I would like to have it included somewhere in the tests so that at least we do not break it accidentally.

should we merge for now and discuss at some point how to deal with the warnings etc.?

gassmoeller:
Yes merging sound like a good idea. At the moment it slows down the CIG tests by 4 mins, but we have enough machines to make up for that, and merging it creates some pressure to fix the warnings. We can discuss options at the hackathon.

bangerth:
On 5/7/19 3:20 PM, Timo Heister wrote:
> Does this mean that the compile can also break depending on how many cores 
> your machine has?

I think the typical case where it would break is if you had
   namespace {
     void f() { do something }
   }
in one .cc file and
   namespace {
     void f() { do something else }
   }
in another .cc file. C++ says that this is ok: The function exists only in 
that one .cc file that it is defined in, and so in each file, the 
corresponding function is called.

You get into trouble if you concatenate different .cc files because you now 
have these conflicting function declarations. We have cases in deal.II where 
that would happen, principally in the source/fe/ directory where different 
elements implement equally named (and similar) functions, but differently.

You would see these issues in a build where you concatenate *all* files (i.e., 
-j1) but might get lucky with higher -jN, depending on the choice of N.

PR #2919, opened 2019-05-07
Simplify test that caused testers to break because of too many MPI processes

gassmoeller:
If our discussion in #2914 is correct, this should significantly speed up the CIG test machines. As we found MPICH significantly deteriorates in performance if more MPI processes than available hardware threads are used (significantly deteriorates == using 2100 seconds instead of 0.21 seconds). Since the only other solution is to only test on large machines (which we have not available), or to recompile MPICH for the container (which would be complicated), or to update the container to Ubuntu 18.04 (which would allow switching back to openmpi), limiting tests to <= 4 MPI ranks seems the easiest option.


gassmoeller:
Yes, looks like this change brings down the total test time on the projector node from 48 to 26 minutes. Now the test infrastructure should be ready for the hackathon.

tjhei:
I think this is the easiest solution for now. Switching to OpenMPI has various other issues.

bangerth:
This one, single test, caused a degradation of performance by a factor of 10,000?

gassmoeller:
The time for the single test increases by a factor of 10,000 once you use too many processes, yes.

bangerth:
On 5/23/19 11:45 AM, Rene Gassmoeller wrote:
> The time for the single test increases by a factor of 10,000 once you use too 
> many processes, yes.

Bizarre!

PR #2922, opened 2019-05-08
Increase GMRES restart length for advection solver

### LINKED ISSUE #1243: pure Neumann temperature solve fails
If you prescribe no fixed boundary for the temperature, the solver fails. We should raise an error or make the solver work (by enforcing a mean value?).

[lev_solver_prob.prm.txt](https://github.com/geodynamics/aspect/files/514646/lev_solver_prob.prm.txt)

gassmoeller:
This fixes #1243 for me. It seems there are hard problems for the temperature solver where a GMRES restart length of 30 is not enough. Most production problems I checked only use 15-20 iterations for the temperature solver, but the problem in #1243 needs 40, because it has an essentially constant temperature that does not change over time. I saw a few other models over the years where essentially uniform initial temperatures, combined with a very long timestep could lead to convergence issues of the temperature solver, hopefully this change should fix them all.


naliboff:
Initial thought - increasing the restart length to 100 may end up being more of a problem (memory increase) than the relatively rare (to my knowledge) issue in #1243. My vote would be to increase the restart length to 50 and/or provide an input parameter for the restart length.

gassmoeller:
I did both, and I agree. With 50 the advection solver should not use more memory than the Stokes solver (also has a default of 50), so there should be no effective change in the memory usage, while still fixing the convergence issues.

PR #2923, opened 2019-05-08
Fix issue 2315

### LINKED ISSUE #2315: Velocity discontinuity along 0 longitude meridan for GPlates boundary conditions
When running a 3D spherical model with GPlates boundary conditions, there is a discontinuity in the velocity magnitude when crossing the 0 longitude meridan. 

screen shot 2018-06-19 at 10 38 34
Zoom in with mesh:
screen shot 2018-06-19 at 10 38 01

gassmoeller:
This PR fixes #2315. The gplates files only contain velocity values for (0,2pi - dphi), but the gplates plugin assumed they contained values for (0,2pi). As a consequence all surface points with longitudes between 2pi-dphi and 2pi were assigned the surface velocities of 2pi-dphi, the highest value of the table. The fix is to insert the velocity values for phi=0 twice, once for phi=0 and once for phi=2pi. This way the existing interpolation works correctly.

Comparison of before/after (the problem is exaggerated in the figure compared to a production model, because the colorscale was chosen to emphasize the jump, and this model uses a coarse gplates file):

comparison


source/boundary_velocity/gplates.cc
```diff
@@ -208,10 +208,16 @@ namespace aspect
             i++;
           }
 
+        for (unsigned int i=0; i<n_theta; ++i)
```
naliboff: Can you add a brief comment explaining why this is necessary?

gassmoeller:
Done.

PR #2924, opened 2019-05-09
Remove particle code that is now in deal.II

gassmoeller:
Final step of #1897. Remove all of the dead particle code that is no longer needed in ASPECT, because it is now in deal.II >=9.0.


naliboff:
Thanks for going through and cleaning all of that up! Looks good.

PR #2925, opened 2019-05-10
Deduplicate material output postprocessors

gassmoeller:
Remove a number of visualization plugins that all only provide a subset of the functionality of the `material properties` postprocessor. I kept parameter file compatibility, and did not update any parameter files for now. If we want we can keep the old names around, but I dislike all the duplicated calls to the material model, and the duplicated code. Plus this makes compiling slightly faster.


source/postprocess/visualization.cc
```diff
@@ -1018,6 +1021,48 @@ namespace aspect
...
+
+            std::set<std::string> deprecated_postprocessors = {"density",
+                                                               "specific heat",
+                                                               "thermal conductivity",
+                                                               "thermal diffusivity",
+                                                               "thermal expansivity",
+                                                               "viscosity"
+                                                              };
+
+            auto plugin_name = viz_names.begin();
```
naliboff: Please add in a few explanatory comments in the conditional statements below.

gassmoeller:
I explained the logic. Ready from my side.

Issue #2803, opened 2019-02-05
 particle changes [post switch to deal.ii 9.0]


tjhei:
- [ ] remove "ascii" output format from parameter options, document all formats, update tests
- [ ] remove all compatibility code

PR #2928, opened 2019-05-10
Remove some more ifdefs

gassmoeller:
Follow up on #2924 removing some more not longer needed checks. Also related to #2803.


tjhei:
Great, but these would make great starter projects for next week!

gassmoeller:
I just removed them because they were related to the work in #2924, which I think was too complicated for a starter project. I will leave all the remaining ifdefs for the hackathon.

Issue #2926, opened 2019-05-10
Tests without result directories pass without notice


gassmoeller:
Something that @naliboff found: We have some tests that have no directories with reference results, and they always pass (he is working on a fix for the tests). But the permament solution would be to let the tester fail in such cases to let us know that the test is not working as intended.

tjhei:
Yes, that makes sense. Will do.

PR #2930, opened 2019-05-13
add missing output directory for tests/melt_track.prm

tjhei:
also see #2926


gassmoeller:
Surprise :smile:. Thanks for fixing.

PR #2931, opened 2019-05-13
fail if directory of a test is missing

### LINKED ISSUE #2926: Tests without result directories pass without notice

tjhei:
fixes #2926


tjhei:
currently based on #2930

gassmoeller:
Thanks, I suspect that will be useful during the hackathon. :+1:

PR #2936, opened 2019-05-21
Implement the stabilized Q1-Q1 element combination

bangerth:


bangerth:
First pull request of the hackathon?

tjesser-ucdavis-edu:
@bangerth You had to jinx it, huh?

@gassmoeller For some reason, Jenkins isn't detecting the Jenkinsfile.cig in this PR.

From the webhook events: https://jenkins.geodynamics.org/job/aspect/indexing/events

>   Getting remote pull request #2936...
>
>    Checking pull request #2936
>      ‘Jenkinsfile.cig’ not found
>    Does not meet criteria

And from the repository scan log: https://jenkins.geodynamics.org/job/aspect/indexing/console

> Getting remote pull requests...
>
>    Checking pull request #2936
>      ‘Jenkinsfile.cig’ not found
>    Does not meet criteria

GitHub reports the webhook arriving correctly. I'm not sure what would cause this.

bangerth:
I need to add that this is joint work with @cedrict . A paper about this is 2/3 done.

source/simulator/core.cc
```diff
@@ -924,6 +920,15 @@ namespace aspect
...
         coupling[x.temperature][x.temperature] = DoFTools::always;
 
+      // For equal-order interpolation, we need a stabilization term
+      // in the bottom right of Stokes matrix. Make sure we have the
+      // necessary entries.
+      if (parameters.use_equal_order_interpolation_for_stokes == true)
+        coupling[x.pressure][x.pressure] = DoFTools::always;
+
+
+      coupling[x.temperature][x.temperature] = DoFTools::always;
```
tjhei: this line looks like it might have been a mistake during rebase. Can you remove it?

source/simulator/parameters.cc
```diff
@@ -1513,17 +1554,37 @@ namespace aspect
...
                                "a linear element for the velocity, you'd need a continuous "
-                               "element of degree zero for the pressure, which does not exist."))
+                               "element of degree zero for the pressure, which does not exist. "
+                               "On the other hand, if using equal-order interpolation, "
+                               "choosing the polynomial degree as one yields a Q1-Q1 "
+                               "element; using a locally conservative discretization "
+                               "with polynomial degree of one yields a Q1-P0 "
+                               "element."));
+
+      AssertThrow (! (use_locally_conservative_discretization &&
```
tjhei: would it make sense to do the check that the order is 1 if use_equal_order_interpolation_for_stokes is true here?

tests/q1_q1.prm
```diff
@@ -0,0 +1,97 @@
+############### Global parameters
```
tjhei: can you add a comment at the top along the lines of "test the q1q1 stabilization". Where did you copy the test from?

source/simulator/parameters.cc
bangerth: The test here is this:
```
      AssertThrow (! (use_locally_conservative_discretization &&
                      use_equal_order_interpolation_for_stokes),
                   ExcMessage ("You have tried to use both the 'Use locally "
                               "conservative discretization' and 'Use equal order "
                               "interpolation for Stokes' parameters in the input "
                               "file. However, their use is incompatible: you "
                               "can only select one of the two."));
```
What I want to check is that you can't at the same time select to use the $Q_k x DGP_{k-1}$ element and the $Q_k x Q_k$. These two things just don't go together, independent of the degree $k$.

tjhei: I understand. My question was: can you please additionally add
```
AssertThrow(!use_equal_order_interpolation_for_stokes || stokes_velocity_degree==1,
ExcMessage("you need to use degree=1 if you use equal order interpolation.")
```

bangerth: Nah, you don't. You can use equal order elements also in the form of Q2 x Q2. It just requires a different implementation of the projection in the stabilization term. I haven't implemented that yet, but I have the corresponding assertion in the assembly term. That's where the missing functionality is, and so that's where the assertion should be.

The assertions we have in the current place (in `parameters.cc`) are logical errors, rather than things not implemented.

tjhei: I would prefer not to trigger an exception somewhere deep inside a workstream assembly, but I don't need to start a fight over this.

tjhei:
```
The following tests FAILED:
	495 - q1_q1 (Failed)
	496 - q1_q1_6 (Failed)
	550 - sol_cx_2_equal_order (Failed)
	555 - sol_cx_4_equal_order (Failed)
```

source/simulator/assemblers/stokes.cc
```diff
@@ -56,6 +56,52 @@ namespace aspect
...
+                                                           numbers::signaling_nan<double>());
+      if (this->get_parameters().use_equal_order_interpolation_for_stokes)
+        {
+          // Check that we are really only using a Q1-Q1 element and
+          // not a Q2-Q2 element. This is because in the latter case, the
+          // projection isn't just on the piecewise constants, but onto
+          // the piecewise (bi,tri)linears, and this is going to be a bit
+          // more involved than just computing a single number per shape
+          // function.
+          Assert (this->get_parameters().stokes_velocity_degree==1,
```
bangerth: @tjhei: FYI, this is the place where we trigger the assertion. There is a similar one further down in the file.

bangerth:
Let's try this again.

tjhei:
````
The following tests FAILED:
	495 - q1_q1 (Failed)
	496 - q1_q1_6 (Failed)
```

bangerth:
I wished I could fix this, but the testers offer me no artifacts or outputs. Ideas?

bangerth:
I've got it figured out. I think this should finally work.

bangerth:
Works now! Hallelujah!

bangerth:
Ah, nice to see it in there! @cedrict -- thanks for the collaboration!

PR #2937, opened 2019-05-22
Verify parameter/variable settings when restart, for issue #2090

### LINKED ISSUE #2090: Verify checkpoint/restart agree on the fields a simulation has

bangerth:
This PR supersedes #2523. It simply moves the commit there to the current master and makes the adjustments necessary to make things run (or, at least, we'll find out whether that works).

Fixes #2090. Closes #2523.

### DIFF
--- a/source/simulator/checkpoint_restart.cc
+++ b/source/simulator/checkpoint_restart.cc
@@ -85,7 +85,6 @@ namespace aspect
       oa << parameters.surface_pressure;
       oa << parameters.use_operator_splitting;
       oa << parameters.include_melt_transport;
-      oa << parameters.enable_additional_stokes_rhs;
       oa << parameters.stokes_velocity_degree;
       oa << parameters.use_locally_conservative_discretization;
       oa << parameters.use_discontinuous_temperature_discretization;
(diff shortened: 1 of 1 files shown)


bangerth:
/restart

Let's see what happens here! :-)

bangerth:
If you run without free surface for a while, you will accumulate stresses at the surface that are proportional to the time you have run (or something like this), not just proportional to the length of the last time step (because you will have relaxed the surface at the end of the previous time step).

So I would expect that you will get a very large deformation in the first time step after restart. I don't think that's going to be useful.

(But I don't feel particularly strongly about it. I'd be more comfortable, though, if someone tested this before we allow it to happen.)

gassmoeller:
Ok, then lets merge this. I do not quite agree with your statement that the stresses will be proportional to the time though. The Stokes equation is time-independent, as long as the density field is comparable (in terms of density variations in the model) the surface stresses will stay approximately constant. But you are right, better only allow it for users who know what they are doing.

bangerth:
It's in the commit authorship for the first commit, but just for the record: This is really work by @wanyingw.

Issue #2921, opened 2019-05-08
Remove all occurences of DEAL_II_VERSION_GTE(9,0,0)


gassmoeller:
This is a great starter project for the hackathon, also as a first pull request for everyone. Lets make this a joint exercise, everyone could pick a file that contains at least one occurrence and open a pull request that fixes that file.

tjhei:
There is still plenty of places left up for grabs!

PR #2944, opened 2019-05-22
Remove dealii compatibility fix

sibiaoliu:
Remove a not longer necessary compatibility fix for deal.ii 8.5.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -788,7 +788,6 @@ namespace aspect
       {
         bool constrained_dofs_set_changed = false;
 
-#if DEAL_II_VERSION_GTE(9,0,0)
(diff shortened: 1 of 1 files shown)


source/simulator/core.cc
```diff
@@ -797,23 +797,6 @@ namespace aspect
                 break;
               }
           }
-#else
```
bangerth: You are removing both the `#else` and the `#endif`. You will then also have to remove the `#if` above.

sibiaoliu: Removed.

tjhei:
related to #2921

sibiaoliu:
We merged the two commits into one.

PR #2951, opened 2019-05-22
Manual changes in inner core cookbook

MarineLasbleis:
In the inner core cookbook explanations: 
- clarifying the self-gravitation (that is not truly self-gravitation)
- added how to compile the shared library


bangerth:
Only manual, so merging without running the tester.

PR #2954, opened 2019-05-22
Use consistent ParaView capitalization.

Djneu:
Changes to keep all capitalization of ParaView consistent.


bangerth:
It's only in the manual. The tester will not uncover anything useful.

anne-glerum:
@bangerth just testing our permissions :)

bangerth:
@anne-glerum -- yes, you did the right thing. It's just that the tester doesn't check the creation of the manual, and so won't find anything about this PR. You couldn't know that.

PR #2959, opened 2019-05-22
corrections in cookbook global_melt

MarineLasbleis:
The value in the comment was not the same than the one used in the code.

### DIFF
--- a/cookbooks/global_melt.prm
+++ b/cookbooks/global_melt.prm
@@ -304,9 +304,7 @@ subsection Mesh refinement
 
   # minimum of 4 global refinements
   subsection Minimum refinement function
-    set Coordinate system   = depth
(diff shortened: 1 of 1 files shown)


cookbooks/global_melt.prm
```diff
@@ -304,12 +304,10 @@ subsection Mesh refinement
 
   # minimum of 4 global refinements
   subsection Minimum refinement function
-    set Coordinate system   = depth
     set Function expression = 4
-    set Variable names      = depth,phi
```
bangerth: Can you say why these changes are correct?

bangerth: Ah, thanks for the explanation across the table: The function expression is a constant, there is no point in naming coordinates explicitly.

PR #2960, opened 2019-05-22
Change more std::shared_ptr to std::unique_ptr.

bangerth:


tjhei:
Looks good, let's wait for the tester...

PR #2964, opened 2019-05-22
Mesh deformation

Djneu:
Addressing some comments from Pr #2071


source/simulator/parameters.cc
```diff
@@ -1791,13 +1788,13 @@ namespace aspect
   parse_geometry_dependent_parameters(ParameterHandler &prm,
                                       const GeometryModel::Interface<dim> &geometry_model)
   {
-    prm.enter_subsection ("Free surface");
+    prm.enter_subsection ("Mesh deformation");
     {
       try
         {
           const std::vector<types::boundary_id> x_free_surface_boundary_indicators
```
anne-glerum: For neatness, rename these variables as well?

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+
+
+    template <int dim>
+    void MeshDeformationHandler<dim>::make_constraints()
+    {
+      if (!sim.parameters.free_surface_enabled)
+        return;
+
+      // Now construct the mesh displacement constraints
+      mesh_displacement_constraints.clear();
```
anne-glerum: We solve for the mesh velocity, so shouldn't `mesh_displacement_constraints` be `mesh_velocity_constraints'?

bangerth:
Ugh, this patch moves around a thousand lines of code. @gassmoeller: Can you point me at where I should look?

Also, these changes are all credited to @Djneu. How did you create the five commits in this PR?

Djneu:
> Also, these changes are all credited to @Djneu. How did you create the five commits in this PR?

I wasn't sure how to deal with git conflicts when updating Rene's branch, so I ended up creating a new branch that was up to date with the current master and copying over the files he made. The first commit is all the new files copied over from his branch, is there a way to change that so its attributed to him?

source/mesh_deformation/free_surface.cc
```diff
@@ -0,0 +1,417 @@
...
+      mass_matrix.compress(VectorOperation::add);
+
+      // Jacobi seems to be fine here.  Other preconditioners (ILU, IC) run into troubles
+      // because the matrix is mostly empty, since we don't touch internal vertices.
+      LinearAlgebra::PreconditionJacobi preconditioner_mass;
+      preconditioner_mass.initialize(mass_matrix);
+
+      SolverControl solver_control(5*rhs.size(), this->get_parameters().linear_stokes_solver_tolerance*rhs.l2_norm());
+      SolverCG<LinearAlgebra::Vector> cg(solver_control);
+      cg.solve (mass_matrix, dist_solution, rhs, preconditioner_mass);
```
bangerth: This is pre-existing, but it seems a bit fishy that we are solving a linear system that really only exists on the boundary. All rows of the matrix corresponding to interior nodes are zero. It is a bit of a miracle that the solver even succeeds, and that the Jacobi preconditioner does not error out when it takes the inverse of a zero diagonal entry.

I suspect that this could be improved somehow.

source/mesh_deformation/free_surface.cc
```diff
@@ -0,0 +1,417 @@
...
+    {
+      // For the free surface indicators we constrain the displacement to be v.n
+      LinearAlgebra::Vector boundary_velocity;
+
+      const IndexSet mesh_locally_owned = free_surface_dof_handler.locally_owned_dofs();
+      IndexSet mesh_locally_relevant;
+      DoFTools::extract_locally_relevant_dofs (free_surface_dof_handler,
+                                               mesh_locally_relevant);
+      boundary_velocity.reinit(mesh_locally_owned, mesh_locally_relevant, this->get_mpi_communicator());
+      project_velocity_onto_boundary(free_surface_dof_handler,mesh_locally_owned,mesh_locally_relevant,boundary_velocity);
```
bangerth: break these arguments onto several lines so they fit into 80-character lines. Put a space between successive arguments

source/mesh_deformation/free_surface.cc
```diff
@@ -0,0 +1,417 @@
...
+
+
+    template <int dim>
+    void FreeSurface<dim>::declare_parameters(ParameterHandler &prm)
+    {
+      prm.enter_subsection ("Mesh deformation");
+      {
+        prm.declare_entry("Free surface stabilization theta", "0.5",
+                          Patterns::Double(0,1),
+                          "Theta parameter described in Kaus et. al. 2010. "
```
bangerth:
```suggestion
                          "Theta parameter described in Kaus et al. (2010). "
```
Extra credit if you add the complete reference to `doc/manual/manual.bib` and `\\cite{...}` it here

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+
+
+
+namespace aspect
+{
+  namespace MeshDeformation
+  {
+    template <int dim>
+    Interface<dim>::~Interface ()
+    {}
```
bangerth: Yes, you can omit this if you put `=default` after the declaration of the function in the .h file.

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+
+    template <int dim>
+    MeshDeformationHandler<dim>::~MeshDeformationHandler ()
+    {
+      // Free the Simulator's mapping object, otherwise
+      // when the MeshDeformationHandler gets destroyed,
+      // the mapping's reference to the mesh displacement
+      // vector will be invalid.
+      sim.mapping.reset();
```
bangerth: Who actually destroys the mesh deformation handler? Shouldn't that place be in charge of resetting the variable?

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+      sim.rebuild_stokes_matrix = sim.rebuild_stokes_preconditioner = true;
+    }
+
+
+
+    template <int dim>
+    void MeshDeformationHandler<dim>::make_constraints()
+    {
+      if (!sim.parameters.free_surface_enabled)
+        return;
```
bangerth: what happens if you have other mesh deformation plugins? Shouldn't you execute these as well?

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+            fe_values.reinit (cell);
+
+            cell_vector = 0;
+            cell_matrix = 0;
+            for (unsigned int point=0; point<n_q_points; ++point)
+              for (unsigned int i=0; i<dofs_per_cell; ++i)
+                {
+                  for (unsigned int j=0; j<dofs_per_cell; ++j)
+                    cell_matrix(i,j) += scalar_product( fe_values[extract_vel].gradient(j,point),
+                                                        fe_values[extract_vel].gradient(i,point) ) *
```
bangerth: reverse these two lines so that it is first `i` then `j`

bangerth: This only yields something nonzero if the components for `i` and `j` are the same. This could be checked above the computation here to skip the computation if possible.

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+    void MeshDeformationHandler<dim>::interpolate_mesh_velocity()
+    {
+      // Interpolate the mesh vertex velocity onto the Stokes velocity system for use in ALE corrections
+      LinearAlgebra::BlockVector distributed_mesh_velocity;
+      distributed_mesh_velocity.reinit(sim.introspection.index_sets.system_partitioning, sim.mpi_communicator);
+
+      const std::vector<Point<dim> > support_points
+        = sim.finite_element.base_element(sim.introspection.component_indices.velocities[0]).get_unit_support_points();
+
+      Quadrature<dim> quad(support_points);
```
bangerth:
```suggestion
      const Quadrature<dim> quad(support_points);
```

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+      UpdateFlags update_flags = UpdateFlags(update_values | update_JxW_values);
```
bangerth:
```suggestion
      const UpdateFlags update_flags = UpdateFlags(update_values | update_JxW_values);
```

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+          {
+            cell->get_dof_indices (cell_dof_indices);
+
+            fe_values.reinit (cell);
+            fs_fe_values.reinit (fscell);
+            fs_fe_values[extract_vel].get_function_values(fs_mesh_velocity, velocity_values);
+            for (unsigned int j=0; j<n_q_points; ++j)
+              for (unsigned int dir=0; dir<dim; ++dir)
+                {
+                  unsigned int support_point_index
```
bangerth:
```suggestion
                  const unsigned int support_point_index
```

bangerth: But it's not actually the support point index, but the local DoF index

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+      distributed_mesh_velocity.compress(VectorOperation::insert);
+      mesh_velocity = distributed_mesh_velocity;
+    }
+
+
+    template <int dim>
+    void MeshDeformationHandler<dim>::setup_dofs()
+    {
+      if (!sim.parameters.free_surface_enabled)
+        return;
```
bangerth: same as before

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+
+      // these live in the same FE as the velocity variable:
+      mesh_velocity.reinit(sim.introspection.index_sets.system_partitioning,
+                           sim.introspection.index_sets.system_relevant_partitioning,
+                           sim.mpi_communicator);
+
+
+      free_surface_dof_handler.distribute_dofs(free_surface_fe);
```
bangerth: needs to be renamed

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+
+      sim.pcout << "Number of free surface degrees of freedom: "
```
bangerth: get this through the simulator access base class

source/mesh_deformation/interface.cc
```diff
@@ -0,0 +1,639 @@
...
+      mesh_vertex_constraints.reinit(mesh_locally_relevant);
+
+      DoFTools::make_hanging_node_constraints(free_surface_dof_handler, mesh_vertex_constraints);
+
+      // We can safely close this now
+      mesh_vertex_constraints.close();
+
+      // Now reset the mapping of the simulator to be something that captures mesh deformation in time.
+      sim.mapping.reset (new MappingQ1Eulerian<dim, LinearAlgebra::Vector> (free_surface_dof_handler,
+                                                                            mesh_displacements));
```
bangerth: same question here as above

source/simulator/core.cc
```diff
@@ -344,8 +344,11 @@ namespace aspect
         AssertThrow ( parameters.pressure_normalization == "no",
                       ExcMessage("The free surface scheme can only be used with no pressure normalization") );
 
-        // Allocate the FreeSurfaceHandler object
-        free_surface = std_cxx14::make_unique<FreeSurfaceHandler<dim>>(*this, prm );
+        // Allocate the MeshDeformationHandler object
+        //free_surface = std_cxx14::make_unique<MeshDeformation::MeshDeformationHandler<dim>>(*this);
+        free_surface.reset(new MeshDeformation::MeshDeformationHandler<dim>(*this));
```
bangerth: I think the previous (commented out) line is the canonical syntax we are using these days

source/simulator/parameters.cc
```diff
@@ -591,13 +591,13 @@ namespace aspect
     }
     prm.leave_subsection();
 
-    prm.enter_subsection ("Free surface");
+    prm.enter_subsection ("Mesh deformation");
```
bangerth: That's going to be a non-backward compatible change

source/simulator/simulator_access.cc
```diff
@@ -658,7 +658,7 @@ namespace aspect
 
 
   template <int dim>
-  const FreeSurfaceHandler<dim> &
+  const MeshDeformation::MeshDeformationHandler<dim> &
   SimulatorAccess<dim>::get_free_surface_handler () const
```
bangerth: rename the function

bangerth:
> I wasn't sure how to deal with git conflicts when updating Rene's branch,
> so I ended up creating a new branch that was up to date with the current
> master and copying over the files he made. The first commit is all the new
> files copied over from his branch, is there a way to change that so its
> attributed to him?

We can fix this in the end. You chose a rather clumsy way of merging a file
with conflicts :-) The problem is that your approach lost the history of the 
changes. Let us help with this sort of stuff in the future :-)

source/mesh_deformation/function.cc
```diff
@@ -0,0 +1,129 @@
...
+#include <aspect/mesh_deformation/function.h>
+
+#include <deal.II/numerics/vector_tools.h>
+
+namespace aspect
+{
+  namespace MeshDeformation
+  {
+    template <int dim>
+    Function<dim>::Function()
```
bangerth: Since it's only the velocity at the boundary, maybe rename the class to `BoundaryFunction` or similar?

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,400 @@
...
+
+        /**
+         * A function that creates constraints for the velocity of certain mesh
+         * vertices (e.g. the surface vertices). The calling class will respect
+         * these constraints when computing the new vertex positions.
+         */
+        virtual
+        void
+        deformation_constraints(const DoFHandler<dim> &free_surface_dof_handler,
+                                ConstraintMatrix &mesh_velocity_constraints) const = 0;
```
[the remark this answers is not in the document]
bangerth: `ConstraintMatrix` is now `AffineConstraints<double>`

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,400 @@
...
+        MeshDeformationHandler(Simulator<dim> &simulator);
+
+        /**
+         * Destructor for the free surface handler.
+         */
+        ~MeshDeformationHandler();
+
+        void initialize();
+
+        void update();
```
bangerth: these two functions need comments

source/mesh_deformation/free_surface.cc
anne-glerum: It's already there :)

include/aspect/mesh_deformation/free_surface.h
```diff
@@ -0,0 +1,149 @@
...
+        execute (internal::Assembly::Scratch::ScratchBase<dim>   &scratch,
+                 internal::Assembly::CopyData::CopyDataBase<dim> &data) const;
+
+      private:
+        /**
+         * Stabilization parameter for the free surface. Should be between
+         * zero and one. A value of zero means no stabilization. See Kaus
+         * et. al. 2010 for more details.
+         */
+        double free_surface_theta;
```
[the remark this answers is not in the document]
anne-glerum: Nope, it's read from the prm file.

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,400 @@
...
+  /**
+   * A namespace that contains everything that is related to the deformation
+   * of the mesh vertices over time.
+   */
+  namespace MeshDeformation
+  {
+    /**
+     * A base class for mesh deformation plugins. Each derived class should
+     * implement a function that determines the deformation velocity for certain
+     * mesh vertices and store them in a ConstraintMatrix object. The velocities
```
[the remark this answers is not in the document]
anne-glerum: We currently require dealii 9.0.0 for ASPECT, but dealii 9 does not include AffineConstraints yet. Should we use compiler statements to test for dealii version and depending on the dealii version, use either ConstraintMatrix or AffineConstraints?

source/mesh_deformation/interface.cc
anne-glerum: So it's safer/better to use the getters when available, even though we have direct access to the simulator?

bangerth: I think so. I don't particularly like the fact that we're having `friend` classes and would want to eventually get rid of that. Not directly accessing variables is going to make it easier later on to achieve that goal.

include/aspect/mesh_deformation/interface.h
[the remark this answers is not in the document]
bangerth: Oh, I hadn't realized that it wasn't in 9.0. I was quite sure that it was!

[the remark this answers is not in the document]
bangerth: Ignore the issue in that case!

anne-glerum:
As soon as @Djneu has accepted my PR, could @bangerth or @gassmoeller have a look? I did just notice I can replace some more direct accesses to simulator, but sleep first :)

Djneu:
I've merged @anne-glerum latest pull request.

include/aspect/mesh_deformation/interface.h
```diff
@@ -0,0 +1,450 @@
...
+        const std::set<types::boundary_id> &
+        get_active_mesh_deformation_boundary_indicators () const;
+
+        /**
+         * Return the boundary id of the surface that has a free surface
+         * mesh deformation object. If no free surface is used,
+         * numbers::invalid_boundary_id is returned.
+         */
+        types::boundary_id
+        get_free_surface_boundary_indicator () const;
```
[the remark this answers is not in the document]
anne-glerum: The function in SimulatorAccess was updated to `get_mesh_deformation_boundary_indicators`, but I found that in some places, only the actual free_surface boundary ids were needed, so I added the `get_free_surface_boundary_indicator` function. I will keep both, but return a set of boundary ids for the latter function as well.

Djneu:
Rebased to master and updated with @anne-glerum's latest changes.

gassmoeller:
I will take a look and make a hopefully final review. Please take a look at the test results once it is done.

/rebuild

gassmoeller:
You will need to fix the indentation, let us know if you need help with that.

source/simulator/assembly.cc
```diff
@@ -525,7 +513,8 @@ namespace aspect
         Mp_preconditioner_AMG->initialize (system_preconditioner_matrix.block(1,1), Amg_data);
       }
 
-    if (parameters.free_surface_enabled || parameters.include_melt_transport || parameters.use_full_A_block_preconditioner)
+    // TODO only needed for free surface, or for all mesh deformations?
```
gassmoeller: Probably for all mesh deformations that not just move the whole boundary uniformly. Lets be safe and use it always when mesh deformation is enabled. Remove the TODO

Djneu:
Okay, I've ran make indent to fix the indent issues.

Djneu:
Somehow missed the other comments, but the styling issues should all be fixed now.

gassmoeller:
Ok, it looks like some of the tests fail. There must be a subtle bug in the new structure. Try to look at the common features of the tests that fail, and guess from there where the problem occurs. I will try to take a look as well, but can not promise that I will have enough time for it in the next week.

gassmoeller:
A quick first look at the test results: It seems the mesh velocity system is not properly solved anymore (or not correctly assembled). The mesh moves somewhat correctly, but the results for topography are off. I already confirmed that the stabilization assembler is called. It also seems that removing the additional calls in `set_assemblers` is not the cause for the test failures. I might be able to come back to this in a week or so, but it would be great if you could take a look and try to collect some info about what works and what does not.

anne-glerum:
@gassmoeller I'll look at it this week.

anne-glerum:
The get_free_surface_boundary_indicators().find() did not work correctly, so Stokes velocities were projected onto all boundaries, instead of just the free surface. Also, the change to the matrix in compute_mesh_displacements  had to be removed. Waiting for @Djneu to merge my PR :)

Djneu:
I’ve merged @anne-glerum latest changes.

include/aspect/mesh_deformation/interface.h
```diff
@@ -228,7 +228,7 @@ namespace aspect
          * mesh deformation object. If no free surface is used,
          * an empty set is returned.
          */
-        const std::set<types::boundary_id>
+        const std::set<types::boundary_id> &
```
gassmoeller: Ouch. This was the reason the `find` function failed? The call created two different copies of the same object so the returned iterator of the one object was never the `end` iterator of the other one? Nice find. Could you try removing the `tmp` objects you created and see if it still works?

anne-glerum: Yep, that was it. It still works without the `tmp` objects, but I thought it was neater to just ask for the `free_surface_boundary_indicators` once outside the loop. Do you agree?

anne-glerum:
> You will need to run `make indent` to let the tests run.

This would be one of those cases where I directly push to @Djneu branch right? Would that be something like `git push Djneu/aspect mesh_deformation`?

include/aspect/mesh_deformation/interface.h
gassmoeller: It looks a bit nicer, yes. But then please make the `tmp` objects into references as well, otherwise you copy the whole set anyway. Like this:
```
const std::set<types::boundary_id> &tmp_free_surface_boundary_indicators = this->...
```

gassmoeller:
`git push Djneu mesh_deformation` should be sufficient if `Djneu` is the name of the remote to Derek's repo (otherwise you need to add his repo as a new remote beforehand with `git remote add Djneu git@github.com:...`).

anne-glerum:
Alright, only 2 of the 24 failing tests are really failing (`chunk_lithostatic_pressure_2d` and `two_merged_boxes_free_surface`), the rest fails on the renaming of `free surface dofs` to `mesh deformation dofs`. I'll try to fix that today/tomorrow.

gassmoeller:
I had a look through the remaining test results, and I think they are purely related to the changes in the output text. The few cases where timings got removed or the nonlinear residuals changed a bit are related to how we test. Small deviations still pass, but if a big change happens (e.g. the text changes) then the whole file gets replaced so that the small deviations also appear. The only apparently bigger change for `free_surface_iterated_IMPES` is also already happening on master so not related to this PR. So if you go ahead and just replace the test output this should be ready.

gassmoeller:
Looks like Timo's tester is happy. I had to restart the CIG tester so it will take a while. Can you add a changelog entry?

gassmoeller:
Ok, looks ready to me. Anything else you want to do, or should I push the button?

anne-glerum:
No more todo's from my side. Are you happy with the changes to the geometry and mesh refinement models I made to make the tests run?

gassmoeller:
Yes the changes look reasonable to me. We might still find small things to fix later, but if necessary we can do that in a follow-up. :tada: for one of the bigger changes in the last year.

anne-glerum:
Great! 🎉

PR #2967, opened 2019-05-22
Viscosity in no melting/freezing set-up

MarineLasbleis:
In the case of no melting/freezing, it's still possible to have a non-zero porosity (if some at the initial conditions or boundary conditions)

Since porosity is used in both cases of the if/else, then we grab the porosity value outside of the loop.


source/material_model/melt_global.cc
```diff
@@ -157,16 +157,19 @@ namespace aspect
                                    0.0;
           out.densities[i] = (reference_rho_s + delta_rho) * temperature_dependence
                              * std::exp(compressibility * (in.pressure[i] - this->get_surface_pressure()));
+           
+	  AssertThrow(this->introspection().compositional_name_exists("porosity"),
```
jdannberg: This may break some things, because the Assert is now executed even if `this->include_melt_transport() == false` and in this case the model may not have a porosity field. 

So what I would suggest is to split this up and put the logic back into the loop, but change the condition to 
`if (this->include_melt_transport())`, and then have a separate if-statement before line 175 (`// Calculate the melting rate...`) saying `if (include_melting_and_freezing && in.strain_rate.size())`. 
Then you can compute the viscosity as `out.viscosities[i] = eta_0;` outside of the loop, `out.viscosities[i] *= exp(- alpha_phi * porosity);` for models with melt transport, and additionally multiply with `out.viscosities[i] *= depletion_strengthening;` for models that also have melting and freezing. 

Sorry for not noticing this earlier.

MarineLasbleis: Since it's a "global melt", it would makes sense to require a porosity field to exist anyway, no? But maybe then it would needs to go somewhere else?

MarineLasbleis: OK, we actually need it for one cookbook! So let's keep it. I'm trying to modify to put the different if statements.

jdannberg:
The astyle tester complains that the indentation is not correct? Do you know how to fix that?

jdannberg:
Alright, this changes the output of one of the tests. I looked at the test, and it makes sense that this output changes (it is a test that uses the melt global model with melt, but no melting and freezing, so the viscosity is slightly lower now, which makes the velocities slightly higher). So in order for the tests to pass, you have to update the test output. You can do that in the same way as for the indentation changes (going to the tester, downloading the .diff file with the changes, and applying it using patch).

gassmoeller:
@MarineLasbleis when you address some comments, always leave a comment about it, otherwise we will not see it. @jdannberg: This is ready for you to merge.

MarineLasbleis:
Yes, sorry!

jdannberg:
Looks good!

PR #2968, opened 2019-05-22
Replace if valid operator for initial temperature conditions

SophieCoulson:
Create new 'replace if valid' operator for initial temperature conditions. It checks if the value in the second model in a nan, if so it replaces with the value from the first model.

Test also included based on the tests for the other operators.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1122,7 +1122,7 @@ namespace aspect
           subtract,
           minimum,
           maximum,
-		  replace_if_valid
+          replace_if_valid
         };
 
         /**
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1121,7 +1121,8 @@ namespace aspect
           add,
           subtract,
           minimum,
-          maximum
+          maximum,
+		  replace_if_valid
         };
 
         /**
--- a/source/initial_temperature/interface.cc
+++ b/source/initial_temperature/interface.cc
@@ -204,7 +204,7 @@ namespace aspect
                           std::get<dim>(registered_plugins).get_description_string());
 
         prm.declare_entry("List of model operators", "add",
-                          Patterns::MultipleSelection("add|subtract|minimum|maximum"),
+                          Patterns::MultipleSelection("add|subtract|minimum|maximum|replace if valid"),
                           "A comma-separated list of operators that "
                           "will be used to append the listed temperature models onto "
                           "the previous models. If only one operator is given, "
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -2795,11 +2795,11 @@ namespace aspect
             return std::max(x,y);
           }
           case Utilities::Operator::replace_if_valid:
-		  {
-        	  if (std::isnan(y))
-			  return x;
-			  else return y;
-		  }
+          {
+            if (std::isnan(y))
+              return x;
+            else return y;
+          }
           default:
           {
             Assert (false, ExcInternalError());
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -2837,7 +2837,7 @@ namespace aspect
           else
             AssertThrow(false,
                         ExcMessage ("ASPECT only accepts the following operators: "
-                                    "add, subtract, minimum, maximum and replace if valid. But your parameter file "
+                                    "add, subtract, minimum, maximum, and replace if valid. But your parameter file "
                                     "contains: " + operator_names[i] + ". Please check your parameter file.") );
         }
 
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -2794,6 +2794,12 @@ namespace aspect
           {
             return std::max(x,y);
           }
+          case Utilities::Operator::replace_if_valid:
+		  {
+        	  if (std::isnan(y))
+			  return x;
+			  else return y;
+		  }
           default:
           {
             Assert (false, ExcInternalError());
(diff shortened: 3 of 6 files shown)


include/aspect/utilities.h
```diff
@@ -1121,7 +1121,8 @@ namespace aspect
           add,
           subtract,
           minimum,
-          maximum
+          maximum,
+		  replace_if_valid
```
bangerth: You'll have to run the indent script I guess :-)

source/utilities.cc
```diff
@@ -2794,6 +2794,12 @@ namespace aspect
           {
             return std::max(x,y);
           }
+          case Utilities::Operator::replace_if_valid:
+		  {
+        	  if (std::isnan(y))
+			  return x;
+			  else return y;
```
bangerth: same here

source/utilities.cc
```diff
@@ -2826,10 +2832,12 @@ namespace aspect
             operator_list[i] = Operator(Operator::minimum);
           else if (operator_names[i] == "maximum")
             operator_list[i] = Operator(Operator::maximum);
+          else if (operator_names[i] == "replace if valid")
+            operator_list[i] = Operator(Operator::replace_if_valid);
           else
             AssertThrow(false,
                         ExcMessage ("ASPECT only accepts the following operators: "
-                                    "add, subtract, minimum and maximum. But your parameter file "
+                                    "add, subtract, minimum, maximum and replace if valid. But your parameter file "
```
bangerth:
```suggestion
                                    "add, subtract, minimum, maximum, and replace if valid. But your parameter file "
```

tests/initial_temperature_list_using_replace_if_valid.prm
```diff
@@ -0,0 +1,86 @@
+##### simple test for using multiple initial temperature plugins
```
bangerth: can you add a sentence that describes what's specific about this test?

SophieCoulson:
Thanks Wolfgang - these things should be fixed now

tjhei:
I have one comment, that we can do as a separate PR: The operator classs is used in several places in the code and those places need adjusting (to add the additional option in the ``prm.declare_entry`` call):
```
$ grep create_model_operator_list -r source/ include/
source/boundary_temperature/interface.cc:        model_operators = Utilities::create_model_operator_list(model_operator_names);
source/initial_composition/interface.cc:        model_operators = Utilities::create_model_operator_list(model_operator_names);
source/boundary_composition/interface.cc:        model_operators = Utilities::create_model_operator_list(model_operator_names);
source/initial_temperature/interface.cc:        model_operators = Utilities::create_model_operator_list(model_operator_names);

```
Ideally, we move the string into a helper function in Utilities (something like ``const std::string get_model_operator_options()`` that returns the string ""add|subtract|minimum|maximum|replace if valid". That way, they can not become out of sync. Again, this can be a followup as the code is correct right now.

tests/initial_temperature_list_using_replace_if_valid.prm
```diff
@@ -0,0 +1,87 @@
...
+# This tests whether the 'replace if valid' operator is working
+'
```
tjhei: remove this please, it causes the test to fail.

source/utilities.cc
```diff
@@ -2794,6 +2794,12 @@ namespace aspect
           {
             return std::max(x,y);
           }
+          case Utilities::Operator::replace_if_valid:
+          {
+            if (std::isnan(y))
+              return x;
+            else return y;
```
bangerth:
```suggestion
            else
              return y;
```

SophieCoulson:
I've accepted Wolfgang's change and removed the header lines in the test file. Will do the other pull request in due course.

tjhei:
The test fails with
``Message:  <Unexpected token "Nan" found at position 10.>``. Are you sure it works on your PC?

SophieCoulson:
Changed Nan to lower case nan, the tester is running again

tjhei:
So, on my machine (and the tester), we can not parse any of "NaN, nan, inf" and I am also unable to define these as additional constants, as you would need to write something like "NaN=NaN", which would require the thing on the right to be parseable. :-(
@bangerth can we discuss what to do at some point? I see the following options:
1. also allow something like -1e300 to register as invalid.
2. Ignore the fact that this new settings doesn't work with function parser. Modify the test to use a different plugin.
3. figure out a way to extend the function parser somehow. Maybe just search and replace "NaN" before parsing? :-)

gassmoeller:
What was the result of the discussion about the NaNs?

source/initial_temperature/interface.cc
```diff
@@ -204,7 +204,7 @@ namespace aspect
                           std::get<dim>(registered_plugins).get_description_string());
 
         prm.declare_entry("List of model operators", "add",
-                          Patterns::MultipleSelection("add|subtract|minimum|maximum"),
+                          Patterns::MultipleSelection("add|subtract|minimum|maximum|replace if valid"),
```
bangerth: We should document "somewhere" what these operators actually are. Is there a good place for this?

bangerth:
Ugh. Not sure what to do. Does it work on *some* systems at least?

I'd be ok to treat everything less than -1e300 as a nan. @SophieCoulson -- do you think you could make that addition (i.e., extend the test in the operator)? I think while you're there, you might also want to check whether the value is `inf` or `-inf`, since these are presumably also invalid values?

SophieCoulson:
@bangerth @tjhei @gassmoeller 
Timo tested inf and -inf and neither of these work on microsoft either. I solved the problem by never specifying 'nan' in a parameter file (since I don't see why we'd ever need to do this in practice anyway). See PR #3007 
I'll close this pull request if you think that solution is okay.

PR #2970, opened 2019-05-23
New: Realistic compressible equation of state (Modified Tait)

bobmyhill:
A material model that implements the thermal modified Tait equation of state as written in the paper of Holland and Powell, 2011 ("An improved and extended internally consistent thermodynamic dataset for phases of petrological interest, involving a new equation of state for solids", J. metamorphic Geol., 2011, 29, 333-383). See here: http://dx.doi.org/10.1111/j.1525-1314.2010.00923.x

Constant values are used for the thermal conductivity and viscosity. The defaults for all coefficients are chosen to be similar to what is believed to be correct for Earth's mantle.


source/material_model/modified_tait.cc
```diff
@@ -0,0 +1,288 @@
...
+    void
+    ModifiedTait<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Material model");
+      {
+        prm.enter_subsection("Modified Tait model");
+        {
+          prm.declare_entry ("Reference pressure", "1e5",
+                             Patterns::Double (0),
+                             "Reference pressure $\\P_0$. Units: $Pa$.");
```
bangerth: The `\\` in front of the `P_0` looks wrong. In principle, the `Pa` should be typeset as `\\text{Pa}` (as I do in #3044, for example), but I won't hold you to actually doing that if you don't want.

Same below for the parameters that follow.

source/material_model/modified_tait.cc
```diff
@@ -0,0 +1,288 @@
...
+  {
+    ASPECT_REGISTER_MATERIAL_MODEL(ModifiedTait,
+                                   "modified tait",
+                                   "A material model that implements the thermal modified Tait "
+                                   "equation of state as written in the paper of "
+                                   "Holland and Powell, 2011 "
+                                   "(``An improved and extended internally consistent "
+                                   "thermodynamic dataset for phases of petrological interest, "
+                                   "involving a new equation of state for solids'', "
+                                   "J. metamorphic Geol., 2011, 29, 333-383, "
```
bangerth: Can you put this citation into `doc/manual/manual.bib` and just `\\cite{...}` it here?

tests/modified_tait_forsterite_box.prm
```diff
@@ -0,0 +1,92 @@
+# This test ensures that the Modified Tait
```
bangerth: you'll need an output file for this test to be useful :-)

source/material_model/modified_tait.cc
```diff
@@ -0,0 +1,288 @@
...
+
+    template <int dim>
+    void
+    ModifiedTait<dim>::
+    evaluate(const MaterialModelInputs<dim> &in,
+             MaterialModelOutputs<dim> &out) const
+    {
+      for (unsigned int i=0; i < in.temperature.size(); ++i)
+        {
+          const double temperature = std::max(1e-10,in.temperature[i]);
```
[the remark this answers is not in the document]
bobmyhill: The model for the heat capacity is a polynomial with negative powers in T, so it breaks down at very low temperatures. Should I instead make sure that ASPECT throws an error if T is small? Or is there an alternative?

source/material_model/modified_tait.cc
```diff
@@ -0,0 +1,288 @@
...
+          out.thermal_conductivities[i] = k_value;
+
+          // Einstein equations for the heat capacity at constant volume
+          const double x_einstein = einstein_temperature/temperature;
+          const double x_einstein0 = einstein_temperature/reference_temperature;
+
+          // The following Einstein energies and heat capacities are divided through by 3nR.
+          // This doesn't matter, as the equation of state relies only on ratios of these quantities
+          double E_th0, E_th, C_V0, C_V;
+          if (x_einstein0 < 35)
```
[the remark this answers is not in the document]
bobmyhill: No, this is just where exp(x_einstein0) becomes extremely big. At this temperature we can simplify the expressions for thermal energy and Cv. Again, this is only true at low temperatures.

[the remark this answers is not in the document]
bobmyhill: I've added a note to this effect, but happy to change further if you want.

tests/modified_tait_forsterite_box.prm
bobmyhill: Oops! Thanks, this also allowed me to catch a bug.

source/material_model/modified_tait.cc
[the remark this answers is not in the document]
bangerth: Why not cut off at the lower end of the data set used to fit the function (say, 100K) and add an `Assert` with an error message explaining the issue?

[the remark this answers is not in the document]
bobmyhill: @bangerth So it turns out that the first call (calculating the adiabat) is set to evaluate by default at 0 K. This works only if we have non-zero temperatures and have the modified approximation for E_th that we discussed earlier.

bobmyhill:
@bangerth I didn't have any other bright ideas about the low temperature issue - I think we need to make sure that we can run the models at 0 K (for the sake of the adiabatic condition). Is this ok?

bangerth:
Yes, sorry for dropping the ball. You say that the adiabat computation requires zero temperature, but why is this so? Do we start the adiabatic temperature at zero at the top? It seems wrong to me to ever evaluate anything at zero temperature, and so I'm surprised you get hit by that. Can we talk about this tomorrow to see how you get to a point where you're asked to evaluate something at zero temperature?

bobmyhill:
@bangerth Fixed! The default for the parameter Adiabatic surface temperature is 0 [K]. I don't think this is ever sensible, but it was an easy fix. I think this might be ready to be tested!

source/material_model/modified_tait.cc
```diff
@@ -0,0 +1,282 @@
...
+          const double temperature = in.temperature[i];
+          const double pressure = in.pressure[i];
+
+          AssertThrow(temperature > 100., ExcMessage("The current temperature is <100 K. "
+                                                     "The modified Tait equations of state is "
+                                                     "not designed for such temperatures. "
+                                                     "If this exception is triggered at the "
+                                                     "beginning of a run, you may need to "
+                                                     "set Adiabatic surface temperature "
+                                                     "to a value > 100 in the parameter file."));
```
bangerth: nice!

source/material_model/modified_tait.cc
```diff
@@ -0,0 +1,282 @@
...
+                                                    "equation of state is less than 1/35th "
+                                                    "of the Einstein temperature. This is not "
+                                                    "expected for geological materials."));
+
+          // The following Einstein energies and heat capacities are divided through by 3nR.
+          // This doesn't matter, as the equation of state relies only on ratios of these quantities.
+          // At large values of x_einstein0, we can simplify the calculation of the thermal energy and Cv.
+          double E_th0, E_th, C_V0, C_V;
+          E_th0 = einstein_temperature * (0.5 + 1. / (std::exp(x_einstein0) - 1.0));
+          C_V0 = x_einstein0 * x_einstein0 * std::exp(x_einstein0) / std::pow(std::exp(x_einstein0) - 1.0, 2.0);
```
bangerth: Ah, this definitely makes the code easier, doesn't it? ;-)

Since you don't modify the variables any more, declare and initialize them in line. Then you can make them `const` as well.

PR #2972, opened 2019-05-23
Provide a way to check whether the simulator has been completely initialized

bangerth:
This should provide @anne-glerum and @naliboff a way to do what they want to do, and more elegantly than they try to do in #2529.

### DIFF
--- a/doc/modules/changes/20190524_bangerth
+++ b/doc/modules/changes/20190524_bangerth
@@ -1,0 +1,6 @@
+Remove: The
+SimulatorAccessor::simulator_is_initialized() function has been
+removed. It was not used anywhere before. It has been superseded by
+the SimulatorAccess::simulator_is_past_initialization() function.
+<br>
+(Wolfgang Bangerth, 2019/05/24)
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -333,6 +333,25 @@ namespace aspect
 
     private:
 
+      /**
+       * A member variable that tracks whether we are completely done
+       * with initialization and have started the time loop. This
+       * variable needs to be the first one that is initialized in
+       * the constructor, so that its value correctly tracks the
+       * status of the overall object. As a consequence, it has to
+       * be the first one declared in this class.
+       *
+       * The variable is set to @p true just before we start the time
+       * stepping loop, but may be temporarily reset to @p false
+       * if, for example, we are during the initial mesh refinement
+       * steps where we start the time loop, but then go back to
+       * initialization steps (mesh refinement, interpolation of initial
+       * conditions, etc.) before re-starting the time loop.
+       *
+       * This variable is queried by
+       * SimulatorAccess::simulator_is_past_initialization().
+       */
+      bool simulator_is_initialized;
 
       /**
        * A class that is empty but that can be used as a member variable and
--- a/include/aspect/simulator_access.h
+++ b/include/aspect/simulator_access.h
@@ -775,12 +775,6 @@ namespace aspect
       get_current_constraints () const;
 
       /**
-       * Return whether or not the current object has been initialized by providing it with
-       * a pointer to a Simulator class object.
-       */
-      bool simulator_is_initialized () const;
-
-      /**
        * Return whether the Simulator object has been completely initialized
        * and has started to run its time stepping loop.
        *
--- a/include/aspect/simulator_access.h
+++ b/include/aspect/simulator_access.h
@@ -781,6 +781,40 @@ namespace aspect
       bool simulator_is_initialized () const;
 
       /**
+       * Return whether the Simulator object has been completely initialized
+       * and has started to run its time stepping loop.
+       *
+       * This function is useful to determine in a plugin whether some
+       * of the information one can query about the Simulator can be trusted
+       * because it has already been set up completely. For example,
+       * while the Simulator is being
+       * set up, plugins may already have access to it via the current
+       * SimulatorAccess object, but data such as the current time, the
+       * time step number, etc, may all still be in a state that is not
+       * reliable since it may not have been initialized at that time. (As
+       * an example, at the very beginning of the Simulator object's existence,
+       * the time step number is set to numbers::invalid_unsigned_int, and
+       * only when the time step loop is started is it set to a valid
+       * value). Similar examples are that at some point the Simulator
+       * sets the solution vector to the correct size, but only at a later
+       * time (though before the time stepping starts), the *contents* of
+       * the solution vector are set based on the initial conditions
+       * specified in the input file.
+       *
+       * Only when this function returns @p true is all of the information
+       * returned by the SimulatorAccess object reliable and correct.
+       *
+       * @note This function returns @p true starting with the moment where the
+       *   Simulator starts the time stepping loop. However, it may
+       *   temporarily revert to returning @p false if, for example,
+       *   the Simulator does the initial mesh refinement steps where
+       *   it starts the time loop, but then goes back to
+       *   initialization steps (mesh refinement, interpolation of initial
+       *   conditions, etc.) before re-starting the time loop.
+       */
+      bool simulator_is_past_initialization () const;
+
+      /**
        * Return the value used for rescaling the pressure in the linear
        * solver.
        */
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -145,6 +145,7 @@ namespace aspect
   Simulator<dim>::Simulator (const MPI_Comm mpi_communicator_,
                              ParameterHandler &prm)
     :
+    simulator_is_initialized (false),
     assemblers (std_cxx14::make_unique<Assemblers::Manager<dim>>()),
     parameters (prm, mpi_communicator_),
     melt_handler (parameters.include_melt_transport ?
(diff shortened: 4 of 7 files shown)


include/aspect/simulator_access.h
```diff
@@ -781,14 +781,32 @@ namespace aspect
...
+       * of the information one can query about the Simulator is reliable
+       * or has already been set. For example, while the Simulator is being
+       * set up, plugins may already have access to it via the current
+       * SimulatorAccess object, but data such as the current time, the
+       * time step number, etc, may all still be in a state that is not
+       * reliable since it may not have been initialized at that time.
+       * Only when this function returns @p true is all of the information
+       * returned by the SimulatorAccess object reliable and correct.
+       */
+      bool simulator_is_past_initialization () const;
```
tjhei: I don't particularly like the name. How strongly do you feel about it? Some suggestions: ``is_initialized``, `is_initialization_complete``, ``!currently_in_initialization``,  ...

include/aspect/simulator_access.h
```diff
@@ -781,14 +781,32 @@ namespace aspect
...
       bool simulator_is_initialized () const;
 
+      /**
+       * Return whether the Simulator object has been completely initialized
+       * and has started to run its time stepping loop. Obviously, for this
+       * function to work, the simulator_is_initialized() function needs
+       * to return @p true.
+       *
+       * This function is useful to determine in a plugin whether some
+       * of the information one can query about the Simulator is reliable
```
tjhei: what is the difference between "reliable" or "has already been set"? I find the word "reliable" confusing. Maybe state that initialization of all plugins and settings is completed? I guess the main difference between ``simulator_is_initialized`` and this function is that the first only says that initialization started and some information is not valid/set yet.

include/aspect/simulator_access.h
```diff
@@ -781,14 +781,32 @@ namespace aspect
...
+       * or has already been set. For example, while the Simulator is being
+       * set up, plugins may already have access to it via the current
+       * SimulatorAccess object, but data such as the current time, the
+       * time step number, etc, may all still be in a state that is not
+       * reliable since it may not have been initialized at that time.
```
tjhei: replace with "in a state that is not valid since ..."

include/aspect/simulator_access.h
```diff
@@ -781,14 +781,32 @@ namespace aspect
        * constraints for the time step we are currently solving.
        */
       const ConstraintMatrix &
-      get_current_constraints() const;
+      get_current_constraints () const;
 
       /**
        * Return whether or not the current object has been initialized by providing it with
        * a pointer to a Simulator class object.
```
tjhei: can you add a comment that this means that the initialization started? Maybe we should rename this function?

include/aspect/simulator.h
```diff
@@ -333,6 +333,15 @@ namespace aspect
...
 
+      /**
+       * A member variable that tracks whether we are completely done
+       * with initialization and have started the time loop. This
+       * variable needs to be the first one that is initialized in
+       * the constructor, so that its value correctly tracks the
+       * status of the overall object. As a consequence, it has to
+       * be the first one declared in this class.
+       */
+      bool simulator_is_initialized;
```
tjhei: this is confusing because I expect the function simulator_is_initialized to return this bool.

source/simulator/core.cc
```diff
@@ -1802,7 +1803,9 @@ namespace aspect
           }
       }
 
-    // start the principal loop over time steps:
+    // Start the principal loop over time steps. At this point, everything
+    // is completely initialized, so set that status as well
+    simulator_is_initialized = true;
```
tjhei: so this variable means "initialization completed"?!

source/simulator/simulator_access.cc
```diff
@@ -706,6 +706,18 @@ namespace aspect
...
   }
 
+
+
+  template <int dim>
+  bool
+  SimulatorAccess<dim>::simulator_is_past_initialization () const
+  {
+    return (simulator_is_initialized()
+            &&
```
tjhei: that logic is very confusing and points at inconsistent naming of functions/variables.

include/aspect/simulator_access.h
```diff
@@ -781,14 +781,32 @@ namespace aspect
...
+       *
+       * This function is useful to determine in a plugin whether some
+       * of the information one can query about the Simulator is reliable
+       * or has already been set. For example, while the Simulator is being
+       * set up, plugins may already have access to it via the current
+       * SimulatorAccess object, but data such as the current time, the
+       * time step number, etc, may all still be in a state that is not
+       * reliable since it may not have been initialized at that time.
+       * Only when this function returns @p true is all of the information
+       * returned by the SimulatorAccess object reliable and correct.
```
tjhei: you never mention that this will be false during mesh refinement later in the simulation, can you add a sentence?

tjhei:
@gassmoeller this PR might be a good example for PR review. I can delete my comments if you want. :-)

include/aspect/simulator_access.h
bangerth: The problem is that the `simulator_is_initialized()` function talks about the initialization of the `SimulatorAccess` object, whereas the new function talks about the initialization state of the `Simulator` object. I'm of course open to other names.

include/aspect/simulator.h
bangerth: Yes. The "simulator" in `SimulatorAccess::simulator_is_initialized()` doesn't talk about the `Simulator` object itself, but whether the `simulator` member variable (a pointer) of the `SimulatorAccess` object has been set to something non-`nullptr`.

gassmoeller:
I think this is a good PR for a public review. We can even leave your comments in here Timo that way we show what happens if reviewers disagree :wink:

bangerth:
I even added two typos :-)

include/aspect/simulator.h
```diff
@@ -333,6 +333,15 @@ namespace aspect
 
     private:
 
+      /**
+       * A mmeber variable that tracks whether we are completely done
```
tjhei: Please fix typo: mmeber -> member

include/aspect/simulator.h
```diff
@@ -333,6 +333,15 @@ namespace aspect
...
+       * with initialization and have started the time lopp. This
```
tjhei:
```suggestion
       * with initialization and have started the time loop. This
```

include/aspect/simulator.h
tjhei: Rene agrees. He thinks it would be best to align the name of the variable with the function that actually returns the value.

include/aspect/simulator_access.h
```diff
@@ -781,14 +781,32 @@ namespace aspect
        * constraints for the time step we are currently solving.
        */
       const ConstraintMatrix &
-      get_current_constraints() const;
+      get_current_constraints () const;
```
tjhei: This is actually unrelated to the PR. Please put it into a separate PR.

bangerth:
I've fixed the typos.

include/aspect/simulator_access.h
bangerth: See #2995.

bangerth:
I think I've addressed all comments.

doc/modules/changes/20190524_bangerth
```diff
@@ -0,0 +1,6 @@
+Remove: The
+SimulatorAccessor::simulator_is_initialized() function has been
```
tjhei: what about source/material_model/drucker_prager.cc ?

bangerth: I had replaced the call to the function in the file you mention.

include/aspect/simulator_access.h
```diff
@@ -772,13 +772,41 @@ namespace aspect
...
+       * value). Similar examples are that at some point the Simulator
+       * sets the solution vector to the correct size, but only at a later
+       * time (though before the time stepping starts), the *contents* of
+       * the solution vector are set based on the initial conditions
+       * specified in the input file.
+       *
+       * Only when this function returns @p true is all of the information
+       * returned by the SimulatorAccess object reliable and correct.
+       *
+       * @note The returns @p true starting with the moment where the
```
tjhei: "This"?

bangerth:
I fixed the typo. Thanks!

* Naming: It's a function that returns information it gets from a different object. I think it's not different from `SimulatorAccess::get_fe()`, for example, which internally calls something complicated to figure out what the finite element is that the simulator uses. Here, we have a function that tells us something about the state the `Simulator` object is in; how it determines that is not particularly relevant to the caller, and so there is really no reason to name the function anything that's reminiscent of any particular variable name used inside the simulator -- which is `private` anyway.

* Deprecation: Since it's used in only one place inside ASPECT, I assumed that it was not widely used outside. That may be wrong. I can leave the function in and deprecate it, if you want -- your call.

gassmoeller:
How strongly to do you feel about 1 Timo? 

2. Should be fine, I just checked the log and the function was only added 2 years ago, and never used much anyway. It was useful to figure out if we are in a unit test (which does not have a Simulator object), or not (in which case it should always return true).

tjhei:
> How strongly to do you feel about 1 Timo?

It is okay if you both like it this way. Not a big deal...

I would name  the variable  ``simulator_is_past_initialization`` like the function.

PR #2973, opened 2019-05-23
This is a supplementary explanation of using the minimum refinement function in manual

sibiaoliu:
Since I am using the minimum refinement function strategy in my model, I think it may be a bit unclear for explanation of minimum refinement function in manual, particularly in the section 5.2.13 Using lazy expression syntax for if-else-statements in function expressions.

For example: 
subsection Mesh refinement
    set Initial global refinement = 3
    set Initial adaptive refinement = 4
    set Strategy = minimum refinement function
          set Coordinate system = cartesian
          set Variable names = x,yset Function constants = ymax=1e6, lith=1.5e5
          set Function expression = ((ymax-y<=lith) ? 7 : 0)
...

In this function expression of if-then-else, the else statement shows minimum refinement level is 0, which does not exactly mean no mesh refinement. In fact, if the refinement level in the else statement (e.g., 0 here) is lower than initial global refinement level, the rest part of the mesh is 1 level down coarsening from initial global refinement level.


doc/manual/manual.tex
```diff
@@ -6754,7 +6754,7 @@ \subsubsection{Using lazy expression syntax for if-else-statements in function e
...
 \lstinputlisting[language=prmfile]{cookbooks/muparser-temperature-example/repetitions.part.prm}
-Further refinements will divide each sub-region multiple times keeping the aspect ratio of the sub-region. In this case, we refine the elements in each subregion 3 more times. We then use the \texttt{minimum refinement function} strategy and use the function expression to refine 4 more times to a refinement level of 7, but only where the depth is less than 150 km.
+Further refinements will divide each sub-region multiple times keeping the aspect ratio of the sub-region. In this case, we refine the elements in each subregion 3 more times. We then use the \texttt{minimum refinement function} strategy and use the function expression to refine 4 more times to a refinement level of 7, but only where the depth is less than 150 km. An appropriate value of minimum refinement level in function expression (7 here) could be the sum of initial global refinement level (3) and initial adaptive refinement level (4).
```
naliboff: Suggested rewording: An appropriate value of **the** minimum refinement level in **the** function expression

cookbooks/muparser_temperature_example/muparser-temperature-example.prm
```diff
@@ -63,6 +63,9 @@ end
 # Refine the upper 150 km of the mesh so lithosphere structure is resolved.
 # Function expression asks if the depth is less than the lithosphere depth and
 # then refinement level = 5, else  refinement level = 0
+# Note that this refinement level 0 means the refinement level below 150 km 
+# depth is the same as initial global refinement level (3), rather than no 
+# mesh refinement
```
naliboff: Just to be safe, can you check that this in fact the case in this specific cookbook? I know you have tested this in other examples.

sibiaoliu: Tested the case and corrected the words

sibiaoliu:
Thanks for reviewing. I modified the prm file of the muparser-temperature-example in the cookbook and updated related part .prm files in the manual. Please take another look :-)

cookbooks/muparser_temperature_example/muparser-temperature-example.prm
```diff
@@ -64,6 +64,11 @@ end
 # Function expression asks if the depth is less than the lithosphere depth and
 # then refinement level = 5, else  refinement level = 0
 
+# Note that if the refinemet level in the else statement (e.g., 0 here) in the 
+# function expression is lower than initial global refinement level, the lower 
+# 850 km of the mesh is 1 level down coarsening from initial global refinement level.
```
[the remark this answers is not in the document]
mibillen: You could, but I think it would still want to explain how this value is specifically chosen to match the initial global refinement.  The way in which different parameter choices combine is actually quite confusing.

PR #2975, opened 2019-05-23
Update tests

naliboff:
Follow-up to #2705 with an update of the strain-weakening parameters used in the `visco_plastic` test suite. @jdannberg @gassmoeller


naliboff:
The updated tests are all passing, so this should be ready to merge.

naliboff:
@jdannberg - I updated `visco_plastic_complex`, should be ready to go.

PR #2978, opened 2019-05-23
add coverage testing.

MFraters:
I though it would be useful to test how well the tests actually cover the code. This is a work in progress.

### DIFF
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -111,7 +111,7 @@ pipeline {
         # compiling...
         cd build-gcc-fast
         ninja
-        ''' 
+        '''
       }
     }
 
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -51,13 +51,7 @@ pipeline {
     }
 
     stage('Check Indentation') {
-      environment {
-        TEST_SECRET = credentials('dummy-credential')
-      }
       steps {
-        sh '''python -c "import os; print os.environ['TEST_SECRET'][3:10]"'''
-        sh '''printf $TEST_SECRET | md5sum -'''
-        sh 'false'
         sh './contrib/utilities/indent'
         sh 'git diff > changes-astyle.diff'
         archiveArtifacts artifacts: 'changes-astyle.diff', fingerprint: true
@@ -135,6 +129,9 @@ pipeline {
       options {
         timeout(time: 150, unit: 'MINUTES')
       }
+      environment {
+        COVERALLS_REPO_TOKEN = credentials('Coveralls-Repo-Token-ASPECT')
+      }
       steps {
         // Let ninja prebuild the test libraries and run
         // the tests to create the output files in parallel. We
@@ -183,6 +180,10 @@ pipeline {
           archiveArtifacts artifacts: 'changes-test-results.diff'
 
           // post the coverage results to coveralls
+          sh '''
+          cd build-gcc-fast
+          coveralls -r ../ -b . -i source --gcov-options '\\-lp'
+          '''
         }
       }
     }
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -51,7 +51,13 @@ pipeline {
     }
 
     stage('Check Indentation') {
+      environment {
+        TEST_SECRET = credentials('dummy-credential')
+      }
       steps {
+        sh '''python -c "import os; print os.environ['TEST_SECRET'][3:10]"'''
+        sh '''printf $TEST_SECRET | md5sum -'''
+        sh 'false'
         sh './contrib/utilities/indent'
         sh 'git diff > changes-astyle.diff'
         archiveArtifacts artifacts: 'changes-astyle.diff', fingerprint: true
@@ -61,6 +67,7 @@ pipeline {
         '''
       }
     }
+   
 
     stage('Unity Build') {
       options {
@@ -99,7 +106,7 @@ pipeline {
 
         cmake \
         -G 'Ninja' \
-        -D CMAKE_CXX_FLAGS='-Werror' \
+        -D CMAKE_CXX_FLAGS='-Werror --coverage' \
         -D ASPECT_TEST_GENERATOR='Ninja' \
         -D ASPECT_USE_PETSC='OFF' \
         -D ASPECT_RUN_ALL_TESTS='ON' \
@@ -111,7 +118,7 @@ pipeline {
         # compiling...
         cd build-gcc-fast
         ninja
-        '''
+        ''' 
       }
     }
 
@@ -174,6 +181,8 @@ pipeline {
           // used to update the test results
           sh 'git diff tests > changes-test-results.diff'
           archiveArtifacts artifacts: 'changes-test-results.diff'
+
+          // post the coverage results to coveralls
         }
       }
     }
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -61,7 +61,6 @@ pipeline {
         '''
       }
     }
-   
 
     stage('Unity Build') {
       options {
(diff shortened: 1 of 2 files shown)


MFraters:
Is it correct that there is no python on the tester?

bangerth:
I think this would be an interesting feature. I know nothing about the correctness of the setup, though.

What is the coverage we get so far? Do you happen to know?

tjhei:
> Is it correct that there is no python on the tester?

```
$ docker run --rm -it tjhei/dealii:v9.0.1-full-v9.0.1-r5-gcc5
dealii@9e8cf75a45bf:~$ python --version
Python 2.7.12
```

gassmoeller:
python is there, but pip is not installed. Is there a way to install cpp-coveralls without pip?

tjhei:
We will need to:
1. change the docker image to include python to be able to install coveralls or use codecov instead.
2. configure jenkins to supply the submission token.
3. configure the githhub repo and access to it for one of the services.

Together, this is a bit time consuming and tricky to do during the hack, but we should definitely do something like this later.

gassmoeller:
@mfraters if you want to move forward with this let me know what to do on the jenkins side.

MFraters:
thanks! I am going to sit down with tyler today to see how far we can get.

tjesser-ucdavis-edu:
Here is a link to pipeline credential usage: https://jenkins.io/doc/book/pipeline/syntax/#supported-credentials-type

Adding credentials can be done from `Credentials` > `Jenkins (Scope)` > `Global (Domain)`.

MFraters:
So thanks to Tyler we have been able to pass a secret to the docker environment. We now need to decide which website we want to use. 

For coveralls we need to have some pip module installed, but it will give us more options:
Screenshot from 2019-06-03 12-22-58

codecov works through a bash script, but it only seems reports through leaving a comment which is updated. 

My preference is coveralls because I found it to be more stable and it has more options. The downside is that we need to update the docker with pip and the coveralls pip module.

gassmoeller:
#3162 would allow us to add packages to the tester image (e.g. the coveralls modules).

tjhei:
I looked into both options and they request access to private email addresses, commit status access, and writing to webhooks on all organizations I am part of. Not sure I want to click that button. :-(

image

MFraters:
I had the same problem back then, but I really needed it for my project and I have seen these coverage testers with many project I trusted that it wouldn't do weird things (see for example Specfem3D Globe). With coveralls you can turn them on and off in their panel. I can turn it on in my panel and give you the token.

tjhei:
Yes, but I can not click this button because this will grant access for the other projects.

One other question: we would need to run the tester on the master branch as well to generate useful statistics, right? Right now, we are not running on master... @gassmoeller what do you think?

gassmoeller:
I agree that the permissions are an issue. If you had only your own repos it is not important but if you have to give access to many organizations (as Timo and I would have to do) that is just not really acceptable. I have read that CodeCov as Github App allows more fine-grained access (https://github.com/apps/codecov), but I do not have the time to test right now. And yes, currently we do not test master (i disabled it during the hackathon, because it took too long). We can think about running it on master nightly.

MFraters:
Now I am thinking of it, I should not have clicked that button. I assumed that because I had no administrator rights on the geodynamics organization, that it would pass the question on to an admin first or just don't give the access. I should have done more research on this and directly contacted the the geodynamics admins to discuss, and I am sorry for that. 

What I think I should have done (there was no github app back then I think) is directly create a world builder organization (it was a personal repository) and create a new special user which was only a member of that world builder organization (like the geodynamics-jenkins and geodynamics-dockerhub).

tjhei:
@gassmoeller should we try to install the codecov app? I can request access just for the cig/aspect repo.

gassmoeller:
Yes that sounds like a good plan.

tjhei:
> Yes that sounds like a good plan.

except that codecov.io still wants me to create an account, which will ask for the same organizational permissions. Grr.

MFraters:
Rene and me discussed what would be the best solution today, and Rene is going to arrange access to the geodynamic-jenkins account or to a new geodynamics-cov account. 

The only thing we need is to make a decision on whether to use codecov or coveralls. My personal peference goes to coveralls because it is is more stable and allows for the use of the status API so that it doesn't have to leave a comment in every pull request. The advantage of codecov over coveralls is that coveralls requires a certain python script to be installed with pip, while codecov just uses a bash script which can be downloaded. If I remember correctly, this is the reason why you preferred codecov Timo?

tjhei:
> whether to use codecov or coveralls.

I don't have a strong preference.

> allows for the use of the status API so that it doesn't have to leave a comment in every pull request

I feel like the status API implies that we would not merge a PR that reduces the coverage (a failure). Is that something we want to do/enforce? I would prefer a single comment that is updated (that is how the tools work, right?)

> while codecov just uses a bash script which can be downloaded.

We are using a custom docker image now, so we could do either without too much trouble.

MFraters:
> I feel like the status API implies that we would not merge a PR that reduces the coverage (a failure). Is that something we want to do/enforce? I would prefer a single comment that is updated (that is how the tools work, right?)

I would say that if the coverage has significantly decreased in a pull request it is something which should have a bit of weight. In the end it is only an advice. I personally think it is a bit cleaner with the api and without the comment, but on the other hand the comment will provide me info in the pull request itself. Both options a possible with coveralls and I am fine with either.

> We are using a custom docker image now, so we could do either without too much trouble.

Nice!

gassmoeller:
> I feel like the status API implies that we would not merge a PR that reduces the coverage (a failure). Is that something we want to do/enforce? I would prefer a single comment that is updated (that is how the tools work, right?)

I would not want pull requests that reduce coverage to fail, that seems too restrictive to me. And I do not want to get used to merging PRs with failing status checks either. So maybe we should go with the comment.
@MFraters: You mentioned you prefer coveralls, and it offers both options (comments or API)? Then lets just go with that and see how it works out. I will let you know once I have access to a suitable user account.

MFraters:
Worked yesterday together with @gassmoeller to fix this and the coverage computation works now (see https://coveralls.io/github/geodynamics/aspect/, coverage is nearly 90%!). It just doesn't post the results yet, but that might have to do with the fact that there is no reference wrt master. 

Anyway, I think that it is in a good enough state to get it into master, so it is ready for further review.

Thanks Rene and Tyler for the help with getting this to work with the CIG Jenkins servers!

Jenkinsfile.cig
```diff
@@ -111,7 +111,7 @@ pipeline {
         # compiling...
         cd build-gcc-fast
         ninja
-        '''
+        '''
```
tjhei: remove this

tjhei:
some notes:
- Something seems to be weird in the coverage output: files are listed 4 times each
- Does this mean that we will need to build master now?
- how is coveralls configured: will it post comments, status, or something?

gassmoeller:
I fixed the space.

> some notes:

>    Something seems to be weird in the coverage output: files are listed 4 times each

Yes in the build for this PR they are listed four time (as often as the PR was tested maybe?). But in the individual jobs they are only listed once: https://coveralls.io/jobs/53511893. I am not sure if this is a common thing, but as long as the results would link to the individual jobs that should be fine.

>    Does this mean that we will need to build master now?

I think if we want to have a general status badge then we may need to do that (maybe nightly)? That would also be a nice way to test deal.II-dev at least daily.

>    how is coveralls configured: will it post comments, status, or something?

At the moment both (although it does not seem to work in this PR). I have no strong feelings about that, should we activate both and then see which one we like more?

tjhei:
> At the moment both (although it does not seem to work in this PR). I have no strong feelings about that, should we activate both and then see which one we like more?

I just would like to avoid being spammed with annoying stuff that I don't know how to turn off. :-)

tjhei:
alright, please merge after tests pass.

PR #2979, opened 2019-05-23
Add plugin to track plastic, viscous, or total strain using particles.

Djneu:
A pull request that adds a new particle plugin that works with the visco-plastic material model to calculate the integrated strain, similar to the integrated strain invariant plugin, and assigns it to the plastic, viscous, and/or total strain.


include/aspect/material_model/visco_plastic.h
```diff
@@ -156,6 +156,16 @@ namespace aspect
 
         double get_min_strain_rate() const;
 
+        /**
+         * A function that computes the plastic yielding
+         * if needed outside material model (e.g., particles).
```
gassmoeller: I think the name of the function currently does not make clear what `plastic yielding` means, how do you return it as a bool? What about the following function name: `is_yielding`? You could then update the documentation to:
```
A function that returns whether the material is plastically yielding at the given pressure, temperature, composition, and strain rate.
```

include/aspect/particle/property/viscoplastic_strain_invariants.h
```diff
@@ -0,0 +1,135 @@
+/*
+ Copyright (C) 2015 - 2017 by the authors of the ASPECT code.
```
gassmoeller: 2019

include/aspect/particle/property/viscoplastic_strain_invariants.h
```diff
@@ -0,0 +1,135 @@
...
+           * Initialization function. This function is called once at the
+           * creation of every particle for every property to initialize its
+           * value.
+           *
+           * @param [in] position The current particle position.
+           * @param [in,out] particle_properties The properties of the particle
+           * that is initialized within the call of this function. The purpose
+           * of this function should be to extend this vector by a number of
+           * properties.
+           */
```
gassmoeller: you do not need to copy this documentation. Either write a documentation that is specific for this plugin, or use the following:
```
/**
 * @copydoc aspect::Particle::Property::Interface::initialize_one_particle_property()
**/
```

The same applies to all the other functions below.

source/material_model/visco_plastic.cc
```diff
@@ -42,6 +42,27 @@ namespace aspect
...
+    {
+      bool plastic_yielding = false;
+
+      const std::vector<double> volume_fractions = MaterialUtilities::compute_volume_fractions(composition, get_volumetric_composition_mask());
+
+      const std::pair<std::vector<double>, std::vector<bool> > calculate_viscosities =
+        calculate_isostrain_viscosities(volume_fractions, pressure, temperature, composition, strain_rate, viscous_flow_law, yield_mechanism);
+
+      std::vector<double>::const_iterator max_composition = std::max_element(volume_fractions.begin(),volume_fractions.end());
+      plastic_yielding = calculate_viscosities.second[std::distance(volume_fractions.begin(),max_composition)];
```
gassmoeller: So this function in reality returns whether the material with the highest volume fraction is plastically yielding? That definitely needs to go into the documentation. Is there a reason you do not check if *any* composition is yielding?

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -0,0 +1,195 @@
...
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with ASPECT; see the file LICENSE.  If not see
+ <http://www.gnu.org/licenses/>.
+ */
+
+#include <aspect/particle/property/viscoplastic_strain_invariants.h>
+#include <aspect/material_model/visco_plastic.h>
+#include <aspect/initial_composition/interface.h>
```
gassmoeller: Do you need this last include? I think you can remove it.

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -0,0 +1,195 @@
...
+        // Velocity gradients
+        Tensor<2,dim> grad_u;
+        for (unsigned int d=0; d<dim; ++d)
+          grad_u[d] = gradients[d];
+
+        // Calculate strain rate from velocity gradients
+        const SymmetricTensor<2,dim> strain_rate = symmetrize (grad_u);
+
+
+        //Put compositional fields into single variable
```
gassmoeller: Missing space before `Put`

source/material_model/visco_plastic.cc
anne-glerum: This bit is a repetition of some lines in `evaluate`, where I did provide some comments:
` // Decide based on the maximum composition if material is yielding.
   // This avoids for example division by zero for harmonic averaging (as plastic_yielding
   // holds values that are either 0 or 1), but might not be consistent with the viscosity
  // averaging chosen.`

naliboff: I agree that choosing the maximum composition is the right approach. If many of the functions in visco_plastic are moved to MaterialUtilities, then hopefully things can be consistent between the compositional field and particle approach.

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -0,0 +1,195 @@
+/*
+  Copyright (C) 2015 - 2017 by the authors of the ASPECT code.
```
naliboff: 2017 -> 2019

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -0,0 +1,195 @@
...
+        const double new_strain = old_strain + dt*edot_ii;
+
+
+        // Put strain into correct data position.
+        if (this->introspection().compositional_name_exists("plastic_strain") && plastic_yielding == true)
+          data[data_position] = new_strain;
+
+        if (this->introspection().compositional_name_exists("viscous_strain") && plastic_yielding == false)
+          data[data_position+(n_components-1)] = new_strain;
+        else if (this->introspection().compositional_name_exists("total_strain") && !this->introspection().compositional_name_exists("plastic_strain"))
```
naliboff: Why do you need to know if  a field for `plastic_strain` exists here?

anne-glerum:
> This is in pretty good shape. I have some documentation and minor comments that are easy to solve. But I also have one major concern that I want to talk to you and @anne-glerum about: As soon as something is yielding, you assign the full strain that is accumulated to any of the strain fields. What is the reasoning for assuming all of that strain is plastic? 
If we're in the plastic regime, we indeed assume that all strain is plastic. This is in line with the computation of the viscosity, where we assume all strain rate is used in the calculation of the viscosity when on yield. 

And I think it confusing that plastic strain is assigned to a field called `viscous_strain`. I know this is all just copy-paste from the benchmark, but let us talk about this.
This I don't see, can you point it out?

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -0,0 +1,195 @@
...
+namespace aspect
+{
+  namespace Particle
+  {
+    namespace Property
+    {
+      ASPECT_REGISTER_PARTICLE_PROPERTY(ViscoPlasticStrainInvariant,
+                                        "viscoplastic strain invariants",
+                                        "A plugin which calculates the the integrated strain invariant "
+                                        "(as described in the integrated_strain_invariant particle property) "
```
gassmoeller: please copy the comment from the header file here instead of referring to it.

source/particle/property/viscoplastic_strain_invariants.cc
Djneu: The particles are currently initialized with the initial composition. Is there a way to do that without the include or is it needed?

gassmoeller: Oh you are right, I did not see that. Keep it as is.

Djneu:
Thank you for the comments, I addressed them and added some tests. Please, take another look and let me know if I missed anything or you notice any other issues.

naliboff:
@Djneu - Thank you for adding the tests! Let's run the tester and see what is says 👍 

A few general suggestion - since these tests are duplicating existing tests and only changing a few lines, it might be worth deriving the parameter files from the existing tests. This enables only including the changes to the specified parameter file in the new test parameter file. For an example, see the test added in #2992.

Djneu:
@naliboff Updated the test prm files.

Djneu:
Okay I think I've added the changelog correctly, but let me know if there are any issues.

gassmoeller:
Ready to go. Thanks!

PR #2981, opened 2019-05-23
Clean up an unused parameter file subsection.

gassmoeller:
This benchmark has no temperature boundary conditions (there are no indicators prescribed). Additionally, it is in a spherical shell, not a box.


gassmoeller:
Thanks

PR #2986, opened 2019-05-23
Convert the majority of the remaining std::shared_ptr to std::unique_ptr

bangerth:
With regards to #2811 and #2375.

I'm sure this is going to fail a bunch of tests, but the tester will reveal which and how many.


gassmoeller:
Looks like the tester is in a generous mood today :smile:. Thanks for cleaning up.

bangerth:
Damn, that was a surprise!

PR #2987, opened 2019-05-23
Initial GMG Stokes solver

tcclevenger:
GMG preconditioner to solve Stokes. Matrix-free for entire Stokes system (system matrix, A-block preconditioning, mass matrix solve). 

Limitations include
- BCs: Dirichlet for any geometry or No normal flux only for box
- Q2 X Q1 only for velocity and pressure
- Must use some type of viscosity cell averaging
- Melt transport, free surface, and locally conservative disretization not yet implemented

I added 3 tests: solcx benchmark with 2 and 4 refinements (no normal flux boundary, 2D) and hollow_sphere benchmark with 1 refinement (inhomogeneous Dirichlet boundary, shell mesh, 3D)


tests/hollow_sphere_gmg.prm
```diff
@@ -0,0 +1,111 @@
+# The 3D HollowSphere benchmark, for which an
```
tjhei: for all your test/*.prm , add a line at the top, what it tests like "solve the hollow_sphere.prm with GMG"

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,652 @@
...
+      : public MatrixFreeOperators::Base<dim, dealii::LinearAlgebra::distributed::BlockVector<number> >
+    {
+      public:
+        StokesOperator ();
+        void clear ();
+        void fill_viscosities_and_pressure_scaling(const dealii::LinearAlgebra::distributed::Vector<number> &visc_vals,
+                                                   const double scaling,
+                                                   const Triangulation<dim> &tria,
+                                                   const DoFHandler<dim> &dof_handler_for_projection);
+        Table<2, VectorizedArray<number> > get_visc_table();
```
tjhei: do you really want to return a copy and not a const ref?

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,652 @@
...
+  namespace MatrixFreeStokesOperators
+  {
+
+    template <int dim, int degree_v, typename number>
+    class StokesOperator
+      : public MatrixFreeOperators::Base<dim, dealii::LinearAlgebra::distributed::BlockVector<number> >
+    {
+      public:
+        StokesOperator ();
+        void clear ();
```
tjhei: these functions need a short doxygen documentation

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,652 @@
...
+        void fill_viscosities_and_pressure_scaling(const dealii::LinearAlgebra::distributed::Vector<number> &visc_vals,
+                                                   const double scaling,
```
tjhei: can some of these made private?

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,652 @@
...
+
+        void local_compute_diagonal (const MatrixFree<dim,number>                     &data,
+                                     dealii::LinearAlgebra::distributed::Vector<number>  &dst,
+                                     const unsigned int                               &dummy,
+                                     const std::pair<unsigned int,unsigned int>       &cell_range) const;
+
+        Table<2, VectorizedArray<number> > one_over_viscosity;
+        double pressure_scaling;
+    };
+    template <int dim, int degree_p, typename number>
```
tjhei: empty lines

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,652 @@
...
+      for (unsigned int i=0; i<local_size; ++i)
+        {
+          Assert(inverse_diagonal.local_element(i) > 0.,
+                 ExcMessage("No diagonal entry in a positive definite operator "
+                            "should be zero"));
+          inverse_diagonal.local_element(i)
+            =1./inverse_diagonal.local_element(i);
+        }
+    }
+    template <int dim, int degree_p, typename number>
```
tjhei: add empty lines

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,652 @@
...
+
+
+
+  template<int dim>
+  class StokesMatrixFreeHandler
+  {
+    public:
+      /**
+                 * Initialize this class, allowing it to read in
+                 * relevant parameters as well as giving it a reference to the
```
tjhei: indentation is wrong

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,652 @@
...
+      typedef MatrixFreeStokesOperators::ABlockOperator<dim,2,double> ABlockMatrixType;
+
+      DoFHandler<dim> dof_handler_v;
+      DoFHandler<dim> dof_handler_p;
+      DoFHandler<dim> dof_handler_projection;
+
+      FESystem<dim> stokes_fe;
+      FESystem<dim> fe_v;
+      FESystem<dim> fe_p;
+      const FESystem<dim> fe_projection;
```
tjhei: why const?

include/aspect/stokes_matrix_free.h
tcclevenger: Both `compute_diagonal()` and `fill_viscosities_and_pressure_scaling()` must (I believe) be public since we call them in `stokes_matrix_free.cc`. We never actually use `clear()' so maybe that can be moved. Or it may not be needed at all?

include/aspect/stokes_matrix_free.h
tcclevenger: No reason, copied and pasted from somewhere. Should all these be const or not? We never change them.

tcclevenger:
The first round of questions have been answered.

source/simulator/stokes_matrix_free.cc
```diff
@@ -0,0 +1,1223 @@
...
+      {
+        const unsigned int n_blocks = in.n_blocks();
+        for (unsigned int b=0; b<n_blocks; ++b)
+          copy(out.block(b),in.block(b));
+      }
+    }
+
+
+    /**
+             * Implement the block Schur preconditioner for the Stokes system.
```
tjhei: fix indentation

source/simulator/stokes_matrix_free.cc
```diff
@@ -0,0 +1,1223 @@
...
+             */
+    template <class ABlockMatrixType, class StokesMatrixType, class MassMatrixType, class PreconditionerMp,class PreconditionerA>
+    class BlockSchurPreconditioner : public Subscriptor
```
tjhei: This has the same name as the one in solver.cc. What is the precise difference, just template arguments for the types?
If you think we need to keep both, can you rename it? Otherwise a unity build fails: https://jenkins.geodynamics.org/blue/organizations/jenkins/aspect/detail/PR-2987/8/pipeline/#step-45-log-730

tcclevenger: The main difference is that
{
        dealii::LinearAlgebra::distributed::BlockVector<double>  dst_tmp(dst);
        dst_tmp.block(0) = 0.0;
        stokes_matrix.vmult(utmp, dst_tmp); // B^T
        utmp.block(0) *= -1.0;
        utmp.block(0) += src.block(0);
}
Is how we have to compute `B^T`. The other just does `stokes_matrix.block(0,1).vmult()`

Should we just rename it `BlockSchurGMGPreconditioner`?

tjhei: you need to give it a different name, otherwise unity builds fail (that is the reason the tester is unhappy)

tcclevenger:
It looks like my tests failed. They pass when I run them on my machine..

gassmoeller:
Talk to me about how to update the tests to the official tester results (if that is what you mean by the tests failed). Small differences between your machine and the official tester are expected.

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,781 @@
...
+          for (unsigned int q=0; q<velocity.n_q_points; ++q)
+            {
+              velocity.submit_symmetric_gradient
+              (viscosity_x_2(cell,q)*velocity.get_symmetric_gradient(q),q);
+            }
+          velocity.integrate (false, true);
+          velocity.distribute_local_to_global (dst);
+        }
+    }
+    template <int dim, int degree_v, typename number>
```
gassmoeller: More empty lines between functions, go once through all the functions in this file and check that they are limited by empty lines.

tcclevenger:
> Except for my one stylistic comment I have only one remaining question: Do you need the implementation of `StokesOperator` in `stokes_matrix_free.h` instead of `stokes_matrix_free.cc` (e.g. for inlining?)? If not, please move them to `stokes_matrix_free.cc`

Yes, they are needed in `stokes_matrix_free.h` because on line 761 (I'll add a comment) we need to define the class variable `stokes_matrix`, `mass_matrix`, etc. which are dependent on these types.

include/aspect/stokes_matrix_free.h
```diff
@@ -0,0 +1,787 @@
...
+      DoFHandler<dim> dof_handler_p;
+      DoFHandler<dim> dof_handler_projection;
+
+      FESystem<dim> stokes_fe;
+      FESystem<dim> fe_v;
+      FESystem<dim> fe_p;
+      FESystem<dim> fe_projection;
+
+      // TODO: velocity degree not only 2, Choosing quadrature degree?
+      typedef MatrixFreeStokesOperators::StokesOperator<dim,2,double> StokesMatrixType;
```
tcclevenger: Here is where we must define the `MatrixFreeStokesOperators`

gassmoeller: Well but that only requires that there is a declaration in the header file, the definition (= implementation) can move to the .cc file like for all the other functions. Or am I missing something?

tcclevenger:
I moved all the implementation to the `.cc` file. All checks/tests have passed. I will rebase, let me know if you have any more comments.

tcclevenger:
> Except for my one stylistic comment I have only one remaining question: Do you need the implementation of `StokesOperator` in `stokes_matrix_free.h` instead of `stokes_matrix_free.cc` (e.g. for inlining?)? If not, please move them to `stokes_matrix_free.cc`

I moved to the `.cc` file. Let me know if there are anymore comments and if not I will rebase.

PR #2992, opened 2019-05-24
Continental geotherm initial temperature condition

anne-glerum:
This sets up a laterally homogeneous continental geotherm as initial temperature condition for for example rift models.

### DIFF
--- a/include/aspect/initial_temperature/continental_geotherm.h
+++ b/include/aspect/initial_temperature/continental_geotherm.h
@@ -65,6 +65,10 @@ namespace aspect
         virtual
         double temperature (const double depth) const;
 
+        virtual
+        double temperature (const double depth,
+                            const std::vector<double> thicknesses) const;
+
         /**
          * Declare the parameters this class takes through input files.
          */
--- a/include/aspect/initial_temperature/continental_geotherm.h
+++ b/include/aspect/initial_temperature/continental_geotherm.h
@@ -87,7 +87,7 @@ namespace aspect
 
         /**
          * Vector for the thicknesses of the compositional fields
-         * representing the layers `upper`, `lower` and `mantle_L`.
+         * representing the layers 'upper_crust', 'lower_crust' and 'lithospheric_mantle'.
          */
         std::vector<double> thicknesses;
 
--- a/include/aspect/initial_temperature/continental_geotherm.h
+++ b/include/aspect/initial_temperature/continental_geotherm.h
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2014 - 2016 by the authors of the ASPECT code.
+  Copyright (C) 2014 - 2019 by the authors of the ASPECT code.
 
   This file is part of ASPECT.
 
@@ -93,7 +93,8 @@ namespace aspect
         double LAB_isotherm;
 
         /**
-         * Vector for field depths
+         * Vector for the thicknesses of the compositional fields
+         * representing the layers `upper`, `lower` and `mantle_L`.
          */
         std::vector<double> thicknesses;
 
--- a/include/aspect/initial_temperature/continental_geotherm.h
+++ b/include/aspect/initial_temperature/continental_geotherm.h
@@ -1,0 +1,117 @@
+/*
+  Copyright (C) 2014 - 2016 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+
+#ifndef _aspect_initial_temperature_continental_geotherm_h
+#define _aspect_initial_temperature_continental_geotherm_h
+
+#include <aspect/initial_temperature/interface.h>
+#include <aspect/simulator_access.h>
+
+namespace aspect
+{
+
+
+  namespace InitialTemperature
+  {
+
+    /**
+     *
+     * @ingroup InitialTemperatures
+     */
+    template <int dim>
+    class ContinentalGeotherm : public Interface<dim>, public ::aspect::SimulatorAccess<dim>
+    {
+      public:
+        /**
+         * Constructor.
+         */
+        ContinentalGeotherm ();
+
+        /**
+         * Initialization function. This function is called once at the
+         * beginning of the program. Checks preconditions.
+         */
+        void
+        initialize ();
+
+        /**
+         * Return the initial temperature as a function of position.
+         */
+        virtual
+        double initial_temperature (const Point<dim> &position) const;
+
+        /**
+         * Return the initial temperature as a function of position.
+         */
+        virtual
+        double temperature (const double depth) const;
+
+        /**
+         * Declare the parameters this class takes through input files.
+         */
+        static
+        void
+        declare_parameters (ParameterHandler &prm);
+
+        /**
+         * Read the parameters this class declares from the parameter file.
+         */
+        virtual
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+        /**
+         * Surface temperature
+         */
+        double T0;
+
+        /**
+         * LAB isotherm temperature
+         */
+        double LAB_isotherm;
+
+        /**
+         * Vector for field heat production rates, read from parameter file .
+         */
+        std::vector<double> heat_productivities;
+
+        /**
+         * Vector for thermal conductivities, read from parameter file .
+         */
+        std::vector<double> conductivities;
+
+        /**
+         * Vector for field densities, read from parameter file .
+         */
+        std::vector<double> densities;
+
+        /**
+         * Vector for field depths, read from parameter file .
+         */
+        std::vector<double> thicknesses;
+    };
+  }
+}
+
+
+#endif
--- a/include/aspect/initial_temperature/continental_geotherm.h
+++ b/include/aspect/initial_temperature/continental_geotherm.h
@@ -28,8 +28,6 @@
 
 namespace aspect
 {
-
-
   namespace InitialTemperature
   {
 
@@ -54,18 +52,12 @@ namespace aspect
         initialize ();
 
         /**
-         * Return the initial temperature as a function of position.
-         */
-        virtual
-        double initial_temperature (const Point<dim> &position) const;
-
-        /**
          * Return the initial temperature as a function of depth,
          * based on the solution of the steady-state heat conduction
          * differential equation.
          */
         virtual
-        double temperature (const double depth) const;
+        double initial_temperature (const Point<dim> &position) const;
 
         /**
          * Declare the parameters this class takes through input files.
@@ -88,7 +80,8 @@ namespace aspect
         double T0;
 
         /**
-         * LAB isotherm temperature
+         * Value of the isotherm defining the
+         * Lithosphere-Asthenosphere boundary.
          */
         double LAB_isotherm;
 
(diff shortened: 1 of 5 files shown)


include/aspect/initial_temperature/continental_geotherm.h
```diff
@@ -0,0 +1,119 @@
+/*
+  Copyright (C) 2014 - 2016 by the authors of the ASPECT code.
```
naliboff: Update copyright

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,257 @@
+/*
+  Copyright (C) 2011 - 2016 by the authors of the ASPECT code.
```
naliboff: update copyright

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,257 @@
...
+    void
+    ContinentalGeotherm<dim>::
+    initialize ()
+    {
+      // Check that the required radioactive heating model ("compositional heating") is used
+      const std::vector<std::string> &heating_models = this->get_heating_model_manager().get_active_heating_model_names();
+      AssertThrow(std::find(heating_models.begin(), heating_models.end(), "compositional heating") != heating_models.end(),
+                  ExcMessage("The continental geotherm initial temperature plugin requires the compositional heating plugin."));
+
+      // Check that the required material model ("visco plastic") is used
```
naliboff: Would it also make sense to use this with `DiffusionDislocation` or any other material model?

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+
+    template <int dim>
+    double
+    ContinentalGeotherm<dim>::
+    temperature (const double depth) const
+    {
+      // Compute some constants to calculate the temperatures T1 and T2 at the interfaces
+      // between the layers upper crust/lower crust and lower crust/lithospheric mantle.
```
naliboff: Are these specific constants referred to in Turcotte and Schubert? If so, would it be possible to reference the equations?

naliboff: Same for the equations below. I find these a bit harder to read/interpret than the Chapman equations, so it would be nice to have some specific references (if possible).

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,257 @@
...
+      const double a = 0.5*densities[0]*heat_productivities[0]*thicknesses[0] + 0.5*densities[1]*heat_productivities[1]*thicknesses[1] + conductivities[0]/thicknesses[0]*T0;
+      const double b = 1./(conductivities[0]/thicknesses[0]+conductivities[1]/thicknesses[1]);
+      const double c = 0.5*densities[1]*heat_productivities[1]*thicknesses[1] + conductivities[2]/thicknesses[2]*LAB_isotherm;
+      const double d = 1./(conductivities[1]/thicknesses[1]+conductivities[2]/thicknesses[2]);
+
+      //Temperature at boundary between layer 1 and 2
+      const double T1 = (a*b + conductivities[1]/thicknesses[1]*c*d*b) / (1.-(conductivities[1]*conductivities[1])/(thicknesses[1]*thicknesses[1])*d*b);
+      //Temperature at boundary between layer 2 and 3
```
naliboff: Formatting: add space between `//` and`Temperature`

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,257 @@
...
+      // Temperature in layer 1
+      if (depth <= thicknesses[0])
+        return -0.5*densities[0]*heat_productivities[0]/conductivities[0]*std::pow(depth,2) + (0.5*densities[0]*heat_productivities[0]*thicknesses[0]/conductivities[0] + (T1-T0)/thicknesses[0])*depth + T0;
+      // Temperature in layer 2
+      else if (depth <= thicknesses[0]+thicknesses[1])
+        return -0.5*densities[1]*heat_productivities[1]/conductivities[1]*std::pow(depth-thicknesses[0],2.) + (0.5*densities[1]*heat_productivities[1]*thicknesses[1]/conductivities[1] + (T2-T1)/thicknesses[1])*(depth-thicknesses[0]) + T1;
+      // Temperature in layer 3
+      else if (depth <= thicknesses[0]+thicknesses[1]+thicknesses[2])
+        return (LAB_isotherm-T2)/thicknesses[2] *(depth-thicknesses[0]-thicknesses[1]) + T2;
+      // Return a constant sublithospheric temperature of 10*LAB_isotherm.
```
naliboff: Will a factor  10 always be sufficient to exceed the adiabatic temperature? I imagine so, but perhaps 100 to be safe?

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,257 @@
...
+
+    template <int dim>
+    void
+    ContinentalGeotherm<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection ("Initial temperature model");
+      {
+        prm.enter_subsection("Continental geotherm");
+        {
+          prm.declare_entry ("Layer thicknesses", "30000.",
```
naliboff: Will this only work if you have 3 layers within the lithosphere? If this is the case, can you state this in the documentation below and add an exception in the parse parameter section?

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,257 @@
...
+                                                                3,
+                                                                "Layer thicknesses");
+        }
+        prm.leave_subsection();
+      }
+      prm.leave_subsection();
+
+
+      // Retrieve the indices of the fields that represent the lithospheric layers.
+      AssertThrow(this->introspection().compositional_name_exists("upper"),ExcMessage("We need a compositional field called 'upper' representing the upper crust."));
```
naliboff: Hmm, enforcing the names to be `"upper"` and `"lower"` is really specific, but I know this was done to be consistent with the continental extension cookbook. Could we allow the name to be `upper` or `upper_crust`, etc?

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,257 @@
...
+      AssertThrow(this->introspection().compositional_name_exists("lower"),ExcMessage("We need a compositional field called 'lower' representing the lower crust."));
+      AssertThrow(this->introspection().compositional_name_exists("mantle_L"),ExcMessage("We need a compositional field called 'mantle_L' representing the lithospheric part of the mantle."));
+
+      // For now, we assume a 3-layer system with an upper crust, lower crust and lithospheric mantle
```
naliboff: Would it be possible to generalize the plugin to have an infinite number of layers within the lithosphere? I'll go back and look at the Chapman equations to see if this is possible.

naliboff: The equations can also be reformulated such that the surface heat flow and temperature are provided, but the basal temperature is not specified. However, further generalization may be beyond the scope here.

tests/continental_extension.prm
```diff
@@ -0,0 +1,246 @@
...
+set Use years in output instead of seconds = true
+set Nonlinear solver scheme                = single Advection, iterated Stokes
+set Nonlinear solver tolerance             = 1e-4
+set Max nonlinear iterations               = 10
+set CFL number                             = 0.5
+set Pressure normalization                 = no
+
+#### Parameters describing the model
+
+# Model geometry (400x100 km, 2 km spacing)
```
naliboff: 10 km spacing here

tests/continental_extension.prm
```diff
@@ -0,0 +1,246 @@
+####
```
naliboff: You could remove a good chunk of the parameter file by using:
`include $ASPECT_SOURCE_DIR/cookbooks/continental_extension.prm`

tests/continental_extension.prm
```diff
@@ -0,0 +1,246 @@
...
+  subsection Box
+    set X repetitions = 40
+    set Y repetitions = 10
+    set X extent      = 400e3
+    set Y extent      = 100e3
+  end
+end
+
+# Mesh refinement specifications (no mesh refinement, 
+# but the coarse mesh is already 200x50, see above)
```
naliboff: `mesh is already 40x10`

tests/continental_extension.prm
```diff
@@ -0,0 +1,246 @@
...
+  set Fixed temperature boundary indicators = bottom, top
+  set List of model names = box
+
+  subsection Box
+    set Bottom temperature = 1573
+    set Top temperature    =  273
+  end
+end
+
+# Initial temperature field
```
naliboff: Can you remove this documentation and the `function` section?

source/initial_temperature/continental_geotherm.cc
anne-glerum: This will only work for 3 layers. I've added that to the documentation. There is no need to make an additional exception, because the `possibly_extend_from_1_to_N ` function will throw for any other length than 1 or 3.

source/initial_temperature/continental_geotherm.cc
anne-glerum: I can allow more names, but the logic for getting the composition index, e.g. ` const unsigned int id_upper = this->introspection().compositional_index_for_name("upper");`, would get much more complicated as well.

source/initial_temperature/continental_geotherm.cc
anne-glerum: That would make sense, any material model that provides the density and diffusivity for 3 compositions could be used. However, by restricting it to `visco plastic`, I know which parameters to read in `parse_parameters`. How would I otherwise decide which ones to read there?

anne-glerum:
Hi @naliboff I addressed/answered your comments.

source/initial_temperature/continental_geotherm.cc
naliboff: What happens if only one layer thickness is specified? It looks like there will be an error when the code tries to access `thicknesses[1]` and `thicknesses[2]`.

source/initial_temperature/continental_geotherm.cc
naliboff: You are right, this will get more complicated really quickly. More thoughts on this in the review summary.

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,276 @@
...
+          const std::vector<double> temp_heat_capacities = Utilities::possibly_extend_from_1_to_N (Utilities::string_to_double(Utilities::split_string_list(prm.get("Heat capacities"))),
+                                                           n_fields+1,
+                                                           "Heat capacities");
+
+          densities.push_back(temp_densities[id_upper+1]);
+          densities.push_back(temp_densities[id_lower+1]);
+          densities.push_back(temp_densities[id_mantle_L+1]);
+
+          // Thermal diffusivity kappa = k/(rho*cp), so thermal conducitivity k = kappa*rho*cp
+          conductivities.push_back(temp_thermal_diffusivities[id_upper+1] * densities[0] * temp_heat_capacities[id_upper+1]);
```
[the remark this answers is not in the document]
anne-glerum: It's correct. The densities have already been put in the right order in lines 210-212 by selecting the needed densities from `temp_densities`. The diffusivities still need to be selected with the ids of the compositional fields however.

anne-glerum:
I'm not attached to the names of the fields, I just stuck to those in the cookbook. I've changed them to what you suggested and addressed the rest of the comments.

include/aspect/initial_temperature/continental_geotherm.h
```diff
@@ -0,0 +1,116 @@
...
+
+        /**
+         * Value of the isotherm defining the
+         * Lithosphere-Asthenosphere boundary.
+         */
+        double LAB_isotherm;
+
+        /**
+         * Vector for the thicknesses of the compositional fields
+         * representing the layers `upper`, `lower` and `mantle_L`.
```
gassmoeller: Please update names here.

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,269 @@
...
+        }
+        prm.leave_subsection();
+      }
+      prm.leave_subsection();
+
+
+      // Retrieve the indices of the fields that represent the lithospheric layers.
+      AssertThrow(this->introspection().compositional_name_exists("upper_crust"),ExcMessage("We need a compositional field called 'upper' representing the upper crust."));
+      AssertThrow(this->introspection().compositional_name_exists("lower_crust"),ExcMessage("We need a compositional field called 'lower' representing the lower crust."));
+      AssertThrow(this->introspection().compositional_name_exists("lithospheric_mantle"),ExcMessage("We need a compositional field called 'mantle_L' representing the lithospheric part of the mantle."));
```
gassmoeller: The three Asserts need updates to the names

source/initial_temperature/continental_geotherm.cc
```diff
@@ -0,0 +1,269 @@
...
+                                              "\n"
+                                              "For any depths below the depth of the LAB, a unrealistically high "
+                                              "temperature is returned, such that this plugin can be combined with "
+                                              "another temperature plugin through the 'minimum' operator. "
+                                              "\n"
+                                              "Note that the current implementation only works for a 3-layer lithosphere, "
+                                              "even though in principle the heat conduction equation can be solved "
+                                              "for any number of layers. The naming of the compositional fields "
+                                              "that represent the layers is also very specific, namely 'upper', "
+                                              "'lower', and 'mantle_L'. "
```
gassmoeller: Need to update names.

anne-glerum:
@gassmoeller done :)

gassmoeller:
Great thanks!

PR #2997, opened 2019-05-24
Clean up a piece of cmake scripts.

bangerth:
Follow-up to #2870.


bangerth:
With the `BEFORE` option, I'm now putting certain paths *before* all of the system paths where the existing version put them all *after* the system include paths. I think that before is the correct choice. We don't want to be preempted by a system include file that happens to have the same name, with the exception of the unit test framework case where I think that would make sense. (This is why I put that onto a separate line; that part is unchanged from current behavior.)

PR #2998, opened 2019-05-24
Improved: Ascii datafile end-read warning message

bobmyhill:
The original message for the end of Ascii datafile read-in was very confusing, and made it seem like ASPECT had encountered a problem. This 2-line PR improves that message.

### DIFF
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -2149,9 +2149,12 @@ namespace aspect
       time_dependent = false;
       // Give warning if first processor
       this->get_pcout() << std::endl
-                        << "   Loading new data file did not succeed." << std::endl
-                        << "   Assuming constant boundary conditions for rest of model run."
-                        << std::endl << std::endl;
+                        << "   From this timestep onwards, ASPECT will not attempt to load new Ascii data files." << std::endl
+                        << "   This is either because ASPECT has already read all the files necessary to impose" << std::endl
+                        << "   the requested boundary condition, or that the last available file has been read." << std::endl
+                        << "   If the Ascii data represented a time-dependent boundary condition," << std::endl
+                        << "   that time-dependence ends at this timestep  (i.e. the boundary condition" << std::endl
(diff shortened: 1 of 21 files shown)


source/utilities.cc
```diff
@@ -2149,8 +2149,8 @@ namespace aspect
       time_dependent = false;
       // Give warning if first processor
       this->get_pcout() << std::endl
-                        << "   Loading new data file did not succeed." << std::endl
-                        << "   Assuming constant boundary conditions for rest of model run."
+                        << "   From this timestep onwards, ASPECT will not attempt to load new Ascii data files." << std::endl
```
bangerth: The original message said that reading another file did not succeed. I suspect that this is because there are no more such files. Can you mention that in the error message on *why* you won't attempt to load new files?

source/utilities.cc
```diff
@@ -2149,8 +2149,8 @@ namespace aspect
...
+                        << "   If the Ascii data represented a time-dependent boundary condition, that dependence ends at this timestep."
```
bangerth: Maybe add something like "..., i.e., the boundary condition will simply continue unchanged from the last known state into the future"?

source/utilities.cc
bobmyhill: Good idea. I've added a couple of lines which cover all the present uses of the function.

source/utilities.cc
bobmyhill: Done :)

tjhei:
bob, can you please fix the indentation?

bobmyhill:
Ok, indentation fixed :)

gassmoeller:
Thanks, please update all the tests that display this message.

tjhei:
This is a smart way to generate a lot of changed lines. I am kidding, thanks for improving this!

bangerth:
Says the man who just changed 300 lines by converting tabs into spaces in #3035  ;-)

PR #3000, opened 2019-05-24
Use lower-case letters for cmake variables.

bangerth:
We generally use upper-case names for cmake commands and keywords, and lower-case letters for our own variables.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ MESSAGE(STATUS "====================================================")
 # Set the name of the project and target:
 SET(TARGET "aspect")
 
-FILE(GLOB_RECURSE TARGET_SRC
+FILE(GLOB_RECURSE target_src
      "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
 
 # Set up include directories. Put the ASPECT header files
@@ -97,7 +97,7 @@ DEAL_II_INITIALIZE_CACHED_VARIABLES()
 PROJECT(${TARGET} CXX)
 
 IF (ASPECT_USE_PETSC)
-FOREACH(_source_file ${TARGET_SRC})
+FOREACH(_source_file ${target_src})
   SET_PROPERTY(SOURCE ${_source_file}
     APPEND PROPERTY COMPILE_DEFINITIONS ASPECT_USE_PETSC="1")
 ENDFOREACH()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ MESSAGE(STATUS "====================================================")
 # Set the name of the project and target:
 SET(TARGET "aspect")
 
-FILE(GLOB_RECURSE TARGET_SRC  "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
+FILE(GLOB_RECURSE target_src  "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
 INCLUDE_DIRECTORIES(${CMAKE_BINARY_DIR}/include include  contrib/catch)
```
MFraters: I am used for cmake stuff to make it capital, so it is consistent with with cmake does in for example this line and also the deal.ii cmake variables like _DEALII_GOOD in line 81, but looking at for example line 95, we are not really consistent anyway.

I am not sure what the best approach is here, but currently capital variables for cmake has my preference.

bangerth: cmake is case insensitive, so it doesn't matter. It's really just a convention that all commands and keywords are upper case and all of our own variables are lower-case. I agree that we're probably still inconsistent -- we're just moving in the "more consistent" direction with this patch.

tjhei: > cmake is case insensitive

That is not correct: https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#variables

We use lower and upper case variables all over the place in deal.II and in ASPECT. I don't feel strongly about it, but I think this is too much churn if we want to change that everywhere.

The TARGET_SRC variable is capitalized the same way in the deal.II CMakeFiles. :-)

bangerth: I was imprecise: cmake command names are case insensitive. 

But `TARGET_SRC` is not a built-in cmake variable -- it's one exclusively used in user projects, and I find it useful to distinguish between cmake commands and built-in stuff and user stuff using the case in which things are written. We may not be entirely consistent about it, but we can at least strive for it and get closer to it :-)

tjhei: What an important change for pull request #3000 :-)

I still think that it is weird to not consistently change capitalization in deal.II as well. I don't think you want to do this in the deal.II cmake code-base?

bangerth: Happy to do it in deal.II as well -- give me a minute!

bangerth: See https://github.com/dealii/dealii/pull/8283

tjhei: My point is: Matthias uses uppercase variables all over the place inside deal.II as well. Good luck convincing him. I am not opposed to the change, but I also don't see great value in doing so.

bangerth: Fair enough. Let's put this on hold for the moment. I didn't realize it was something so much larger than just this one variable.

bangerth:
Let's not worry about it, given the deal.II discussion.

bangerth:
Shame that that's what happens to the PR with the round number...

MFraters:
don't worry, we will have an other change in a 1000 pull requests ;)

PR #3002, opened 2019-05-24
Simply prohibit copying material model inputs/outputs.

bangerth:
Follow-up to #2993.


tjhei:
Good to know that we never made any copies...

PR #3006, opened 2019-05-24
Add a new structure for rheology models

gassmoeller:
This is a first stab at the new structure for outsourcing and deduplicating viscosity computations between material models. I am grateful for comments.

### DIFF
--- a/include/aspect/material_model/rheology/constant.h
+++ b/include/aspect/material_model/rheology/constant.h
@@ -1,0 +1,64 @@
+/*
+  Copyright (C) 2019 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _aspect_material_model_rheology_constant_h
+#define _aspect_material_model_rheology_constant_h
+
+#include <aspect/global.h>
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    namespace Rheology
+    {
+      class Constant
+      {
+        public:
+          /**
+           * Declare the parameters this function takes through input files.
+           */
+          static
+          void
+          declare_parameters (ParameterHandler &prm);
+
+          /**
+           * Read the parameters from the parameter file.
+           */
+          virtual
+          void
+          parse_parameters (ParameterHandler &prm);
+
+          /**
+           * Compute the viscosity, in this case just a constant.
+           */
+          double
+          compute_viscosity () const;
+
+        private:
+
+          double eta;
+      };
+    }
+  }
+}
+#endif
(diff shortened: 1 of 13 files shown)


include/aspect/material_model/rheology/constant.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <aspect/global.h>
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    namespace Rheology
```
MFraters: RheologyModel?

naliboff:
@gassmoeller - Thank for you taking the first crack at this! 

A general comment about the naming schemes for different properties. Here, you used `Constant` to represent a constant viscosity. 

However, there are potential use cases (ex: viscoelasticity) where the value being returned is a force or stress. Another example is grain-size - would there be a separate function to compute grain size or should this be done in the material model. Yet another use case is strain-softening, but perhaps this should go in the `MaterialUtilities`.

If the classes under `Rheology` are really only designed to return viscosities, should the namespace be changed to `Viscosity`? For me, "rheology" implies anything related to the constitutive behavior (stress, strain, viscosity, etc).

include/aspect/material_model/rheology/constant.h
gassmoeller: Personally I want to get rid of as many `model` as possible (we renamed all the boundary conditions and initial conditions for 2.0). I can live with `MaterialModel`, but I do not want to introduce more of them :smile:.

gassmoeller:
I think I addressed all comments, and also mentioned in the changelog that the deduplication changed a default value.

gassmoeller:
Yes my idea was to keep the function as simple as possible and not worry about an interface.

include/aspect/material_model/rheology/constant_viscosity.h
```diff
@@ -0,0 +1,64 @@
...
+          static
+          void
+          declare_parameters (ParameterHandler &prm);
+
+          /**
+           * Read the parameters from the parameter file.
+           */
+          virtual
+          void
+          parse_parameters (ParameterHandler &prm);
```
bangerth: If you make the function `virtual`, then that is only useful if you want to use the class in a class hierarchy. But then you also want a virtual destructor.

In the end, I think you probably don't want that, and could just remove the `virtual`.

include/aspect/material_model/rheology/constant_viscosity.h
```diff
@@ -0,0 +1,64 @@
...
+
+          /**
+           * Compute the viscosity, in this case just a constant.
+           */
+          double
+          compute_viscosity () const;
+
+        private:
+
+          double eta;
```
bangerth: Maybe a one-line comment? For example how this variable is initialized?

bangerth: The class could use a constructor that sets the variable to an invalid value.

source/material_model/rheology/constant_viscosity.cc
```diff
@@ -0,0 +1,56 @@
...
+      {
+        return eta;
+      }
+
+
+
+      void
+      ConstantViscosity::declare_parameters (ParameterHandler &prm)
+      {
+        prm.declare_entry ("Viscosity", "1e21",
```
bangerth: What was the default value before and why can't we use it?

include/aspect/material_model/rheology/constant_viscosity.h
```diff
@@ -0,0 +1,64 @@
...
+    {
+      class ConstantViscosity
+      {
+        public:
+          /**
+           * Declare the parameters this function takes through input files.
+           */
+          static
+          void
+          declare_parameters (ParameterHandler &prm);
```
bangerth: You could avoid the problem with the change of default if you let this function take an argument for the constant viscosity. This way, each model that uses this rheology could just use its own preferred value.

gassmoeller:
I addressed the comments and added a optional argument that allows to choose the default viscosity. Ready from my side.

PR #3010, opened 2019-05-25
Get rid of the remaining uses of std::shared_ptr.

### LINKED ISSUE #2811: Change shared_ptr to unique_ptr

bangerth:

This in fact closes #2811.


tjhei:
```
The following tests FAILED:
	 23 - advection_reaction (Failed)
	210 - exponential_decay (Failed)
```

bangerth:
I fixed the remaining failing tests.

PR #3011, opened 2019-05-25
Remove reference to deal.II version 9.0.0.

mnkajan:
Addresses issue #2921.


naliboff:
@tjhei - The CIG tester is passing here, but yours is failing with the error:
`./contrib/utilities/indent: not found`
Is there something @mnkajan needs to do here or does the tester need to be restarted?

tjhei:
This could be fixed by reading. We recently changed the location of the indent script. My tester is testing the pr while Rene is testing the merged result, hence the difference.

PR #3013, opened 2019-05-25
Allow particles to be applied to fields when particles aren't distributed throughout model.

Djneu:
These changes allow the cell/harmonic average, and nearest neighbor particle interpolators to apply the particles as a field even when there are areas with no particles. 

A few questions, as of now this assumes that any area with no particles should be given a value of zero. Would it be better to have this be user defined? Also, should I add a flag for this so that, generally, an assert would still be thrown unless this feature is specifically wanted?

### DIFF
--- a/include/aspect/particle/interpolator/cell_average.h
+++ b/include/aspect/particle/interpolator/cell_average.h
@@ -52,6 +52,29 @@ namespace aspect
 
           // avoid -Woverloaded-virtual:
           using Interface<dim>::properties_at_points;
+
+          /**
+          * @copydoc Interface<dim>::declare_parameters()
+          **/
+          static
+          void
+          declare_parameters (ParameterHandler &prm);
+
+          /**
+           * @copydoc Interface<dim>::parse_parameters()
+           **/
+          virtual
+          void
+          parse_parameters (ParameterHandler &prm);
+
+        private:
+          /**
+           * Generally, particles need to be distributed throughout the
+           * model domain when being applied to a field. This parameter
+           * allows particles to be applied as a field even if some cells
+           * have no particles.
+           */
+          bool allow_cells_without_particles;
       };
     }
   }
--- a/include/aspect/particle/interpolator/cell_average.h
+++ b/include/aspect/particle/interpolator/cell_average.h
@@ -69,10 +69,9 @@ namespace aspect
 
         private:
           /**
-           * Generally, particles need to be distributed throughout the
-           * model domain when being applied to a field. This parameter
-           * allows particles to be applied as a field even if some cells
-           * have no particles.
+           * By default, every cell needs to contain particles to use this interpolator
+           * plugin. If this parameter is set to true, cells are allowed to have no particles,
+           * in which case the interpolator will return 0 for the cell's properties.
            */
           bool allow_cells_without_particles;
       };
--- a/include/aspect/particle/interpolator/harmonic_average.h
+++ b/include/aspect/particle/interpolator/harmonic_average.h
@@ -52,6 +52,29 @@ namespace aspect
 
           // avoid -Woverloaded-virtual:
           using Interface<dim>::properties_at_points;
+
+          /**
+           * @copydoc Interface<dim>::declare_parameters()
+           */
+          static
+          void
+          declare_parameters (ParameterHandler &prm);
+
+          /**
+           * @copydoc Interface<dim>::parse_parameters()
+           */
+          virtual
+          void
+          parse_parameters (ParameterHandler &prm);
+
+        private:
+          /**
+           * Generally, particles need to be distributed throughout the
+           * model domain when being applied to a field. This parameter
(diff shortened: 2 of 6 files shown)


gassmoeller:
I like this feature, but it needs a bit of work. To your questions:
1. Do *you* need a value different from 0? If not, do not worry about things somebody else might need in the future. Often these things are never used and just lie around unused.
2. I would want to have this as a feature that is switched off by default. So by default the models should crash if there a no particles in a cell and its neighbors. I have two reasons for that: 1. To preserve the current behavior for old parameter files. 2. For most models that use the particles so far it is essential to have them in every cell (e.g. if they determine the viscosity in a cell). So continueing a run with empty cells will likely lead to bad results, and we should crash rather than continue. You can add a new input parameter (something like `Allow cells without particles`) and only ignore empty cells if this is set to `true`.

Djneu:
Thanks for the comments! I've added a flag for allowing cells without particles that defaults to false.

include/aspect/particle/interpolator/cell_average.h
```diff
@@ -52,6 +52,29 @@ namespace aspect
...
+          /**
+           * @copydoc Interface<dim>::parse_parameters()
+           **/
+          virtual
+          void
+          parse_parameters (ParameterHandler &prm);
+
+        private:
+          /**
+           * Generally, particles need to be distributed throughout the
```
gassmoeller: This wording is not really clear to someone who is not familiar with how particles work. What about the following:
```
           * By default, every cell needs to contain particles to use this interpolator
           * plugin. If this parameter is set to true, cells are allowed to have no particles,
           * in which case the interpolator will return 0 for the cell's properties.
```

include/aspect/particle/interpolator/harmonic_average.h
```diff
@@ -52,6 +52,29 @@ namespace aspect
...
+
+          /**
+           * @copydoc Interface<dim>::parse_parameters()
+           */
+          virtual
+          void
+          parse_parameters (ParameterHandler &prm);
+
+        private:
+          /**
```
gassmoeller: Same comment.

include/aspect/particle/interpolator/nearest_neighbor.h
```diff
@@ -53,6 +53,29 @@ namespace aspect
...
+
+          /**
+           * @copydoc Interface<dim>::parse_parameters()
+           */
+          virtual
+          void
+          parse_parameters (ParameterHandler &prm);
+
+        private:
+          /**
```
gassmoeller: and here

source/particle/interpolator/cell_average.cc
```diff
@@ -115,17 +115,65 @@ namespace aspect
...
+                // Assume property is zero for any areas with no particles
+                else if (allow_cells_without_particles && non_empty_neighbors == 0)
+                  cell_properties[i] = 0;
+              }
+
           }
 
         return std::vector<std::vector<double> > (positions.size(),cell_properties);
       }
+
```
gassmoeller: We usually use 3 empty lines between functions, please change.

source/particle/interpolator/cell_average.cc
```diff
@@ -115,17 +115,65 @@ namespace aspect
...
+      void
+      CellAverage<dim>::declare_parameters (ParameterHandler &prm)
+      {
+        prm.enter_subsection("Postprocess");
+        {
+          prm.enter_subsection("Particles");
+          {
+            prm.declare_entry ("Allow cells without particles", "false",
+                               Patterns::Bool (),
+                               "Generally, particles need to be distributed throughout the "
```
gassmoeller: Same text change here.

source/particle/interpolator/cell_average.cc
```diff
@@ -115,17 +115,65 @@ namespace aspect
...
+      template <int dim>
+      void
+      CellAverage<dim>::parse_parameters (ParameterHandler &prm)
+      {
+        prm.enter_subsection("Postprocess");
+        {
+          prm.enter_subsection("Particles");
+          {
+            allow_cells_without_particles = prm.get_bool("Allow cells without particles");
+
```
gassmoeller: Remove empty line. Also in the other plugins.

source/particle/interpolator/harmonic_average.cc
```diff
@@ -118,17 +118,64 @@ namespace aspect
...
+                // Assume property is zero for any areas with no particles
+                else if (allow_cells_without_particles && non_empty_neighbors == 0)
+                  cell_properties[i] = 0;
+              }
+
           }
 
         return std::vector<std::vector<double> > (positions.size(),cell_properties);
       }
+
```
gassmoeller: three empty lines here and at every other gap between functions.

Djneu:
Okay, I believe I addressed all the styling issues.

gassmoeller:
Ready to merge once the tester is finished.

PR #3014, opened 2019-05-25
disable adaptive refinement by default

tjhei:
It is often a mistake that adaptive refinement is enabled by default and
more likely than not, that is not what is intended. Let's discuss if we should change the default...

### DIFF
--- a/source/simulator/parameters.cc
+++ b/source/simulator/parameters.cc
@@ -712,7 +712,7 @@ namespace aspect
                          "The number of adaptive refinement steps performed after "
                          "initial global refinement but while still within the first "
                          "time step.");
-      prm.declare_entry ("Time steps between mesh refinement", "10",
+      prm.declare_entry ("Time steps between mesh refinement", "0",
                          Patterns::Integer (0),
                          "The number of time steps after which the mesh is to be "
                          "adapted again based on computed error indicators. If 0 "


bangerth:
Hm, I don't particularly like this. Now, if you want to use AMR, you have to figure out what a useful frequency is. The 10 isn't a random value, but one we've come up with after a good amount of testing. I don't think this is the right direction.

gassmoeller:
I am not really convinced either. I now that it is sometimes surprising, especially for benchmark cases, but as I mentioned to Timo, ASPECT is known as "the AMR code" and I think most users expect AMR to happen. We could talk about making it more clear when AMR is happening on the screen instead of just posting the number of cells.

jdannberg:
I can see both points. I've often seen beginners running their models with weird meshes (before they've even found out how to look at the mesh in ParaView) and wondering what's going on. And it's a bit confusing that there is no separate parameter like "Use adaptive mesh refinement = true/false" that would allow to switch it on or off (users have to search for the time steps between mesh refinement parameter). So maybe it would be worth it to add such a variable?
We can still argue if the default should be true or false, but it would allow us to leave the default value of "Time steps between mesh refinement" as it is right now. 

By the way, I would be interested in knowing more about the details of those tests. I almost always set "Time steps between mesh refinement" to a number smaller than 10, because I want to resolve the fastest moving features (plumes or slabs) and I am worried they will just move out of the refined region otherwise.

gassmoeller:
Maybe this would be a good topic for a larger discussion at one of our group meetings to hear everyone's opinion?

PR #2999, opened 2019-05-24
Advection benchmark in 2d annulus

GEuen:
Unfinished benchmark modified from benchmarks/annulus. Used to test cooling due to various parameters for advection stabilization.

PR #3017, opened 2019-05-26
new SUPG parameter design

tjhei:
This is the more robust parameter design as discussed with @tcclevenger over the last few days.

Synthetic benchmarks show little to no difference, but I expect it to be more robust if one of the parameters (\|u\|, eps) goes to zero.

### DIFF
--- a/source/simulator/entropy_viscosity.cc
+++ b/source/simulator/entropy_viscosity.cc
@@ -560,7 +560,6 @@ namespace aspect
           }
         else if (parameters.advection_stabilization_method == Parameters<dim>::AdvectionStabilizationMethod::supg)
           {
-            // things needed for calculating tau for SUPG loop
             double norm_of_advection_term = 0.0;
             double max_conductivity_on_cell = 0.0;
 
@@ -587,8 +586,6 @@ namespace aspect
                       max_conductivity_on_cell = 0.0;
                     }
                 }
-              // TODO: this needs replacing with a more sophisticated check
-              norm_of_advection_term = std::max(norm_of_advection_term,1e-8);
             }
 
             const double fe_order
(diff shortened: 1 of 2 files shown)


tjhei:
@tcclevenger can you check if this makes sense to you?

tjhei:
Note, #2999 produces somewhat different results for the max temperature:
image

source/simulator/entropy_viscosity.cc
```diff
@@ -599,14 +596,32 @@ namespace aspect
...
+              {
+                // Diffusion dominant case, no stabilization needed:
+                viscosity_per_cell[cell->active_cell_index()] = 0.0;
+              }
+            else
+              {
+                // To avoid a division by zero, increase eps slightly. The actual value is not
+                // important, as long as the result is still a valid number. Note that this
+                // is only important if \|u\| and eps are zero.
+                const double peclet = peclet_times_eps / (eps + 1e-100);
```
tcclevenger: Do we need `eps+1e-100` here? If `eps=0`, the `peclet_times_eps==0.0` above will be true, correct? I guess it doesn't really matter though...

source/simulator/entropy_viscosity.cc
```diff
@@ -599,14 +596,32 @@ namespace aspect
...
+                const double coth_of_peclet = (1.0 + exp(-2.0*peclet)) / (1.0 - exp(-2.0*peclet));
+                const double delta = h/2.0/norm_of_advection_term * (coth_of_peclet - 1.0/peclet);
```
tcclevenger:
```suggestion
                const double delta = h/2.0/norm_of_advection_term/fe_order * (coth_of_peclet - 1.0/peclet);
```

tjhei: Wow, good catch. Thanks!

tjhei:
Okay, now this PR passes the tests without a change in the results! :+1:

tcclevenger:
Really great work Timo!

tjhei:
with this fixed version and #3020 I am now getting the same results (better than the old SUPG):
image

tjhei:
With this + #3020:
image

gassmoeller:
Do you want me to merge the current version? Looks reasonable to me.

tjhei:
> Looks reasonable to me.

yes.

PR #3019, opened 2019-05-26
Add option to track noninitial plastic strain

naliboff:
In models where the initial plastic strain is used to initialize deformation through strain-softening, it is often difficult to identify discrete faulting patterns accumulating on top of the initial strain. This PR adds on option to track an additional compositional field that tracks the plastic strain, but does not include the initial plastic strain.

@anne-glerum @Djneu maybe this of interest?


source/material_model/visco_plastic.cc
```diff
@@ -654,6 +657,8 @@ namespace aspect
                   else
                     out.reaction_terms[i][this->introspection().compositional_index_for_name("viscous_strain")] = e_ii;
                 }
+              if (track_noninitial_plastic_strain == true && plastic_yielding == true)
```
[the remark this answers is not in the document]
naliboff: @anne-glerum - That is what I originally did, but thought it might be useful to track the plastic strain even in cases where strain weakening is not used? Thoughts?

source/material_model/visco_plastic.cc
```diff
@@ -546,6 +546,9 @@ namespace aspect
             }
         }
 
+      if (track_noninitial_plastic_strain)
+        composition_mask.set(this->introspection().compositional_index_for_name("noninitial_plastic_strain"),false);
+
```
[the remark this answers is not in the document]
naliboff: This certainly would work, but my preference would be to leave the current input flag so it appears in parameter.prm.

source/material_model/visco_plastic.cc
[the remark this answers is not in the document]
naliboff: After a bit more thought my preference would be to keep `track_noninitial_plastic_strain` separate from strain-weakening. I just tried using the option with the continental extension cookbook (no strain softening) and the `noninitial_plastic_strain` field nicely illustrates the evolution of the rift basin, including where the ductile layers are located. 

Another scenario to use this variable is with the finite strain tensor, which works with total strain. One may want to do strain-softening with the finite strain tensor invariant, but still track the time-dependent location of faults.

Are you comfortable moving forward with the current structure?

include/aspect/material_model/visco_plastic.h
```diff
@@ -321,6 +321,12 @@ namespace aspect
 
         bool use_finite_strain_tensor;
 
+        /**
+         * Whether to track the accumulation of plastic strain in a
+         * field that does not include the initial plastic strain.
```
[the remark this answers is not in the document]
naliboff: Yes, `starts off zero initial condition` is the right interpretation. I will update the documentation.

source/material_model/visco_plastic.cc
```diff
@@ -887,6 +892,11 @@ namespace aspect
                              "for a total of N+1 values, where N is the number of compositional fields. "
                              "If only one value is given, then all use the same value.  Units: None");
 
+          prm.declare_entry ("Track noninitial plastic strain", "false",
+                             Patterns::Bool (),
+                             "Track the accumulation of plastic strain in a field that does not include "
+                             "the initial plastic strain. Units: None");
```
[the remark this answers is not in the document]
naliboff: Yes, I'll add some documentation stating when and why one would use this option.

naliboff:
@jdannberg - I certainly understand not wanting to make the material model more complicated, but unfortunately I also don't see an alternative to adding it directly into `visco_plastic`. 

I don't think a plugin would work as we need to set the `component_mask` prior to that point. We also need access to whether the material is yield, but I think this could be accesses in a plugin. I'm open to any alternatives for implementation in the code structure.

Going forward, a number of groups I'm working with will be using this option and I'm not sure it is ideal for them to be working off my branch rather than the main repository. I am also concerned about where we "stop", as I can envision a number of other minor but widely used features being added in later on.

naliboff:
@gassmoeller - coming back to to this PR, would it make sense to put this bit of code in `rheology/strain_dependent` for the time being? 

The other option is to proceed with making a new submodule for composition fields as discussed above.

gassmoeller:
Na, I do not think `rheology/strain_dependent` would be the correct place for this (it is not influencing viscosity so it is not a rheology). I would say with the new structure of visco plastic it would be ok for now to put it directly into visco_plastic. Just rebase to the newest version, and we will find a better place for it once we have an additional type of submodules for composition reactions.

naliboff:
@gassmoeller - I rebased, significantly reduced the amount of code, and also addressed some of @anne-glerum initial comments. 

While all of the code is now in strain_dependent.cc, the following line could be moved (slight modifications needed) to visco_plastic.cc after the call to update the compositional fields for strain:

```
 if (this->introspection().compositional_name_exists("noninitial_plastic_strain") && plastic_yielding == true && weakening_mechanism == plastic_weakening_with_plastic_strain_only)
              out.reaction_terms[i][this->introspection().compositional_index_for_name("noninitial_plastic_strain")] = e_ii;
```

The `strain_mask` line could also be moved to visco_plastic, but it fits quite logically in its current location.

Thoughts?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -425,6 +427,8 @@ namespace aspect
                 else
                   out.reaction_terms[i][this->introspection().compositional_index_for_name("viscous_strain")] = e_ii;
               }
+            if (this->introspection().compositional_name_exists("noninitial_plastic_strain") && plastic_yielding == true && weakening_mechanism == plastic_weakening_with_plastic_strain_only)
```
gassmoeller: Why the third condition? Should this not be happening whenever plastic yielding happens, even if it is not weakening the material?
If you change the condition remember to also update the documentation change above.

naliboff: @gassmoeller - the idea behind the third condition was to make sure there was no confusion about whether strain weakening was being used even if  `strain weakening mechanism = none`. With said, anyone who uses this option will certainly have carefully read through the documentation and knows exactly what the field represents. I've updated accordingly.

source/material_model/rheology/strain_dependent.cc
```diff
@@ -88,7 +88,12 @@ namespace aspect
                            "but is there to make sure that the original parameters for specifying the "
                            "strain weakening mechanism (``Use plastic/viscous strain weakening'') are still allowed, "
                            "but to guarantee that one uses either the old parameter names or the new ones, "
-                           "never both.");
+                           "never both."
+                           "\n\n"
+                           "If a compositional field named noninitial_plastic strain is "
```
gassmoeller: you need to escape this underscore that is why the documentation does not build at the moment

gassmoeller: Sorry, should have been more precise, double backslash like this: `\\_`

naliboff: Just saw the failed test and realized I needed a double backslash as well. Sorry about that.

naliboff:
@gassmoeller - Tests are updated and I added a change log entry. Let me know if you would like any further changes and thanks for the reviews!

source/material_model/rheology/strain_dependent.cc
```diff
@@ -88,7 +88,12 @@ namespace aspect
                            "but is there to make sure that the original parameters for specifying the "
                            "strain weakening mechanism (``Use plastic/viscous strain weakening'') are still allowed, "
                            "but to guarantee that one uses either the old parameter names or the new ones, "
-                           "never both.");
+                           "never both."
+                           "\n\n"
+                           "If a compositional field named noninitial\\_plastic strain is "
```
tjhei: the field name has an underscore in the code. Is this a mistake here? Can you also put the name in quotes please?

naliboff: There should have another underscore to give `noninitial_plastic_strain`. Thanks for catching that. The name is also in quotes now as well.

PR #3020, opened 2019-05-26
SUPG: do not use EV conduction boundary fix

tjhei:
Disable the EV logic when using SUPG


bangerth:
You'll need to update the tests:

	591 - supg-compositional-passive (Failed)

	592 - supg (Failed)

tests/supg/screen-output
```diff
@@ -8,7 +8,7 @@ Number of degrees of freedom: 948 (578+81+289)
    Postprocessing:
      RMS, max velocity:                  0.816 m/s, 1.37 m/s
      Temperature min/avg/max:            0 K, 0.1027 K, 1 K
-     Heat fluxes through boundary parts: -6.27e-08 W, 0 W, -4.888e-07 W, -5.556e-06 W
+     Heat fluxes through boundary parts: -1.633 W, 0 W, -12.73 W, -144.7 W
```
tjhei: I am a bit surprised by this change...

tjhei:
This gives better results in the advection benchmark:
image

gassmoeller:
Did you mix up the SUPG old and SUPG new figures? To me the left and middle figure look the same, while the right one looks weird.

tjhei:
> Did you mix up the SUPG old and SUPG new figures? To me the left and middle figure look the same, while the right one looks weird.

No, I am pretty sure I annotated this correctly. See https://www.youtube.com/watch?v=KSk7zA--_ZQ&feature=youtu.be
SUPG with this fix has clearly smaller overshoots.

gassmoeller:
Oh, I see what you mean. I thought the bulges were convection features, but they are overshoots that should not be there. Ready then.

PR #2983, opened 2019-05-23
[WIP] Adjust particle vel free surface2

anne-glerum:
Opened to discuss with @gassmoeller. This PR should enable using the free surface i.c.w. particles.

We still need to test whether the particle location in the reference cell is correct at all times when needed, or needs to be updated somewhere.

PR #3023, opened 2019-05-26
[WIP] Move set pointer to mapping

anne-glerum:
This would supersede #2983 . 

At the moment I want to see what the tester says, because on my machine all free surface tests fail, while on @gassmoeller 's machine only 1. Only one of my tests seems to really produce different results tho, but this is another test than Rene's.  

Edit: All free surface tests pass on the tester except where a computation is resumed from snapshot. Some of the tests include adaptive mesh refinement. This therefore seems to work with the new location of the setting of the pointer.

### DIFF
--- a/source/mesh_deformation/interface.cc
+++ b/source/mesh_deformation/interface.cc
@@ -69,7 +69,11 @@ namespace aspect
       : sim(simulator),  // reference to the simulator that owns the MeshDeformationHandler
         mesh_deformation_fe (FE_Q<dim>(1),dim), // Q1 elements which describe the mesh geometry
         mesh_deformation_dof_handler (sim.triangulation)
-    {}
+    {
+      // Now reset the mapping of the simulator to be something that captures mesh deformation in time.
+      sim.mapping.reset (new MappingQ1Eulerian<dim, LinearAlgebra::Vector> (mesh_deformation_dof_handler,
+                                                                            mesh_displacements));
+    }
 
     template <int dim>
     MeshDeformationHandler<dim>::~MeshDeformationHandler ()
@@ -678,10 +682,6 @@ namespace aspect
 
       // We can safely close this now
       mesh_vertex_constraints.close();
-
-      // Now reset the mapping of the simulator to be something that captures mesh deformation in time.
-      sim.mapping.reset (new MappingQ1Eulerian<dim, LinearAlgebra::Vector> (mesh_deformation_dof_handler,
-                                                                            mesh_displacements));
     }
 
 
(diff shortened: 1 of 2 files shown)


gassmoeller:
I ran the tests not on my local machine, but inside the official tester docker image that is why the tests succeeded. If you want to do the same you can install docker and then use the script I just moved in #3034. If you want to test the adaptive refinement you would not necessarily need that though, if it does not work the model should just crash in debug mode (because it tries to access invalid memory).

anne-glerum:
Time to revisit this PR. See also https://community.geodynamics.org/t/particles-and-free-surface/1246

I just tried this fix on the current dev version for the van-keken-smooth.prm benchmark, and free surface + particles runs. The free surface model evolution is pretty much identical to the free slip surface run, apart from 293 out of 1e5 particles getting lost at t1. After this, the number of particles remains constant. In the vtu output, at t1 two regions can be seen at the surface where there are no particles anymore. These regions are advection along and gather along the right vertical boundary. 
t0 - left free slip, right free surface
Screenshot 2020-06-03 at 16 50 41
t1
Screenshot 2020-06-03 at 16 51 00
t328
Screenshot 2020-06-03 at 16 50 31

I'll update the PR and try the restart, as this used to crash before.

Particle parameters:

    subsection Particles
      set Number of particles = 1e5 
      set Time between data output = 100 
      set Data output format = vtu 
      set List of particle properties = function, initial composition, initial position
      set Interpolation scheme = nearest neighbor
      set Particle generator name = random uniform

      subsection Function
        set Variable names      = x,z 
        set Function expression = 0.5*(1+tanh((0.2+0.02*cos(pi*x/0.9142)-z)/0.02))
      end 
    end

anne-glerum:
@gassmoeller Swapping the call to GridTools and setup_dofs() gets rid of crashing after restart.

anne-glerum:
@bangerth The problem was that mesh deformation resets the FE mapping after the ParticleHandler has a pointer to it. By changing when to reset it in the mesh_deformation/interface.cc we avoid that. A much longer alternative that I implemented in #2983 is to clear and set the pointer before and after resetting the mapping by mesh deformation.

bangerth:
Is there a test we could add?

anne-glerum:
At the moment, adding particles to a free surface model will hit an AssertThrow, so that combination cannot be tested yet. Unless there is a way to go around the assert?

bangerth:
No way. If there isn't already a github issue, can you open one? We've often attached a failing testcase to these github issues to make sure that a test already exists if someone wanted to tackle an issue.

anne-glerum:
There is a test that checks whether aspect crashes for free surface + particles due to the AssertThrow (`particle_free_surface.prm`). I thought you meant adding a test for what was changed here.

PR #3024, opened 2019-05-26
fix missing ElasticAdditionalOutputs instantiation

tjhei:
clang + lld complains about a missing instantiation. I guess gcc doesn't
require it, but it is certainly correct to do so.

### DIFF
--- a/source/material_model/viscoelastic.cc
+++ b/source/material_model/viscoelastic.cc
@@ -520,6 +520,11 @@ namespace aspect
 {
   namespace MaterialModel
   {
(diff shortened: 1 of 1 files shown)


bangerth:
Huh, how could this have worked with gcc?...

tjhei:
@gassmoeller do you know why the CIG tester is not touching this PR?

gassmoeller:
Must have missed the webhook. I rescanned the repo, which fixed it. Next time it happens I will try the rebuild command first to see if that is sufficient. Anyway, tests pass.

PR #3025, opened 2019-05-26
Replace using std::shared_ptr in tests and benchmarks.

bangerth:
Follow-up to #3010.

### DIFF
--- a/benchmarks/annulus/annulus.cc
+++ b/benchmarks/annulus/annulus.cc
@@ -428,7 +428,7 @@ namespace aspect
     std::pair<std::string,std::string>
     AnnulusPostprocessor<dim>::execute (TableHandler &)
     {
-      std::shared_ptr<Function<dim> > ref_func;
+      std::unique_ptr<Function<dim> > ref_func;
       {
         const AnnulusMaterial<dim> &material_model
           = Plugins::get_plugin_as_type<const AnnulusMaterial<dim> >(this->get_material_model());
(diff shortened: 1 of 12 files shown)


tjhei:
tests are failing, can you check? I am sure you compiled all benchmarks and cookbooks too, right?

bangerth:
I fixed the test.

I find it awkward to let the tests run on my laptop, because it takes quite a long time and also because so many tests fail on my local setup. So I use the tester to do part of the work I *should* probably be doing before uploading something as a pull request...

tjhei:
> I find it awkward to let the tests run on my laptop, because it takes quite a long time and also because so many tests fail on my local setup.

I am not talking about the tests. My point is that we should check if benchmarks/* still work with your changes as they are not running on the tester. Yes, it takes a while to run them but I think it is necessary. Let me know and I can do it...

bangerth:
Ah, fair. I have to admit that I don't know how that works. If you have a few cycles left-over while talking to someone, I'd appreciate it!

tjhei:
> If you have a few cycles left-over while talking to someone, I'd appreciate it!

FYI, this is done using ``benchmarks/check.sh``. Will run them over night...

tjhei:
There are several things that need to be fixed for the benchmarks, but your changes are not breaking anything. :+1:

bangerth:
OK, thanks for checking. I guess you're willing to fix these things?

tjhei:
> OK, thanks for checking. I guess you're willing to fix these things?

of course, see #3050

PR #3027, opened 2019-05-26
Add a note to the benchmark section of the manual.

bangerth:
While there, also reference the Newton paper.


tjhei:
documentation only...

PR #3028, opened 2019-05-26
cmake: do not configure test project every time

tjhei:
The wait is kind of annoying, so only configure the tests when it is
actually needed.


tjhei:
This is quite an important change. @gassmoeller @jdannberg are you okay with this?

gassmoeller:
I am ok with the test (though I need to change my workflow for it), but only if you update the instructions in the manual in the `Writing tests` subsection, they currently just state you have to run `cmake .` to add a new test.

tjhei:
updated.

gassmoeller:
I talked to @jdannberg and she is fine with the change.

PR #3031, opened 2019-05-26
Provide a base class with a virtual destructor.

bangerth:
The class has a virtual member function but a non-virtual destructor. This is generally
a bad idea, though it doesn't matter here because we never store pointers to this
particular type of class (we use this class as one of several base classes of
the viz postprocessor classes, and only store pointers to the Interface class). But
it is bad practice, so fix this anyway.

### DIFF
--- a/include/aspect/postprocess/visualization.h
+++ b/include/aspect/postprocess/visualization.h
@@ -224,6 +224,12 @@ namespace aspect
       {
         public:
           /**
+           * Destructor.
+           */
+          virtual
+          ~CellDataVectorCreator () = default;
+
+          /**
            * The function classes have to implement that want to output
            * cell-wise data.
            *


bangerth:
I confused myself: The class is actually derived from the `Interface` class. So its destructor is already `virtual`. I guess it doesn't hurt to avoid the confusion by just making that explicit again as done here.

tjhei:
> I confused myself: The class is actually derived from the `Interface` class. So its destructor is already `virtual`. I guess it doesn't hurt to avoid the confusion by just making that explicit again as done here.

how is that with other classes? We should have hundreds derived from an ``Interface`` class.

bangerth:
My confusion was because the class in question still serves as a base class to other classes (where the rule is that base classes ought to have `virtual` destructors. All of the other classes you mention are `final` (something which we could mark now) and for these, there is generally no confusion possible.

tjhei:
> All of the other classes you mention are `final` (something which we could mark now) and for these, there is generally no confusion possible.

:+1:

Issue #2929, opened 2019-05-11
reorganize folders/scripts in doc/


tjhei:
We place all kind of scripts and utilities in the doc/ folder. I think this is an unfortunate situation and I would like to move things around.
Suggestion:
./contrib/utilities/ for indent, update_prm_files.sh, update_source_files.sh
./contrib/release/ for update_copyright, release-tasklist, bump_version.sh
./contrib/docker/ for the docker files (from ./docker)

gassmoeller:
I like it. `/contrib/utilities` will also be useful for other scripts like mentioned in #1234

bangerth:
On 5/11/19 5:34 PM, Timo Heister wrote:
> 
> ./contrib/utilities/ for indent, update_prm_files.sh, update_source_files.sh
> ./contrib/release/ for update_copyright, release-tasklist, bump_version.sh
> ./contrib/docker/ for the docker files (from ./docker)

Good idea.

PR #3034, opened 2019-05-26
Update tester image in run tests script

gassmoeller:
We updated the image for the tester, but forgot to update the script that can run all tests automatically. Found this while looking at #3023.


tjhei:
related to #2929

tjhei:
no tester run required...

PR #3035, opened 2019-05-26
replace tabs with spaces and enforce using indent script

tjhei:
modify indent script:
- replace tab in source and cmake files with two spaces
- also run dos to unix endings on cookbooks/benchmarks
- update files


bangerth:
Was the change to the King et al. file accidental? It doesn't quite belong here...

tjhei:
> Was the change to the King et al. file accidental? It doesn't quite belong here...

yes, the table contained tabs and it would have been messed up, so I had to move it to the readme before applying this indentation. I think the readme is a better place anyways...

PR #3036, opened 2019-05-26
Equation of state model

jdannberg:
This is a first version of the Equation of state model we discussed a few days ago. For now, I've only added one simple equation of state that can be used in a number of material models. 
As we discussed, it is not a plugin system, but for now each equation of state would be a separate class. On the other hand, the functions are so general that it would be possible to make an interface for the equation of state model in the future, if that's what we want to do.

One big problem I have at the moment is that my equation of state needs an additional parameter (in my case, the number of compositional fields), because this is different between the different material models. Because we don't want to declare parameters for a material models that then can never be used, this number is already needed in the declare_parameters function (which may be before there is an instance of the class), so we can not simply set it as part of some initialize function. What I've done now is to use a static variable (and function to set it), but that's not really pretty or easy to read. Another alternative would be to pass this number as a function argument, but as the variable is needed in so many different places, I don't really like that solution either (and it would also make it more difficult to later make an interface for all equations of state). So I would be happy about suggestions for improvement.


include/aspect/material_model/equation_of_state/interface.h
```diff
@@ -0,0 +1,98 @@
+/*
+  Copyright (C) 2011 - 2018 by the authors of the ASPECT code.
```
gassmoeller: 2019

include/aspect/material_model/equation_of_state/interface.h
```diff
@@ -0,0 +1,98 @@
...
+      /**
+       * Thermal expansion coefficients for each composition. It is defined
+       * as $\alpha = - \frac{1}{\rho} \frac{\partial\rho}{\partial T}$
+       */
+      std::vector<double> thermal_expansion_coefficients;
+
+      /**
+       * Specific heat for each composition.
+       */
+      std::vector<double> specific_heat;
```
gassmoeller: Maybe we can fix the naming here: `specific_heat_capacities`? Would be inconsistent with the `MaterialModelOutputs`, but more consistent with the other names. What do you think?

include/aspect/material_model/equation_of_state/interface.h
```diff
@@ -0,0 +1,98 @@
...
+       * as $\kappa = \frac{1}{\rho} \frac{\partial\rho}{\partial p}$.
+       */
+      std::vector<double> compressibilities;
+
+      /**
+       * The product of the change of entropy $\Delta S$ at a phase transition
+       * and the derivative of the phase function $X=X(p,T,\mathfrak c,\mathbf
+       * x)$ with regard to pressure for each composition.
+       */
+      std::vector<double> entropy_derivative_pressure;
```
gassmoeller: This means you expect phase transitions to move into the EoS as well? I guess that makes sense.

include/aspect/material_model/equation_of_state/linearized_incompressible.h
```diff
@@ -0,0 +1,91 @@
...
+       * $\rho(p,T,\mathfrak c) = \left(1-\alpha (T-T_0)\right)\rho_0 + \sum_i \Delta\rho_i \; \mathfrak c_i.$ "
+       * There is no pressure-dependence of the density, and all other material properties
+       * relating to the equation of state are assumed to be constant and identical for each
+       * composition.
+       */
+      template <int dim>
+      class LinearizedIncompressible
+      {
+        public:
+          void evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
```
gassmoeller: Add a short documentation. `Function that computes the output of the equation of state for all compositions` should be sufficient.

include/aspect/material_model/equation_of_state/linearized_incompressible.h
```diff
@@ -0,0 +1,91 @@
...
+           * does not necessarily imply that the density is constant; rather, it
+           * may still depend on temperature or pressure. In the current
+           * context, compressibility means whether we should solve the continuity
+           * equation as $\nabla \cdot (\rho \mathbf u)=0$ (compressible Stokes)
+           * or as $\nabla \cdot \mathbf{u}=0$ (incompressible Stokes).
+           */
+          bool is_compressible () const;
+
+          /**
+           * Declare the parameters this class takes through input files.
```
gassmoeller: Mention the optional parameter and what it controls.

include/aspect/material_model/equation_of_state/linearized_incompressible.h
```diff
@@ -0,0 +1,91 @@
...
+          /**
+           * Read the parameters this class declares from the parameter file.
+           */
+          void
+          parse_parameters (ParameterHandler &prm,
+                            const unsigned int n_compositions = 0);
+
+
+        private:
+          double reference_rho;
```
gassmoeller: Documentation for the member variables.

source/material_model/equation_of_state/linearized_incompressible.cc
```diff
@@ -0,0 +1,159 @@
+/*
+  Copyright (C) 2011 - 2018 by the authors of the ASPECT code.
```
gassmoeller: 2019

source/material_model/equation_of_state/linearized_incompressible.cc
```diff
@@ -0,0 +1,159 @@
...
+               const unsigned int q,
+               MaterialModel::EquationOfStateOutputs<dim> &out) const
+      {
+
+        Assert(number_of_compositions+1 >= out.densities.size(),
+               ExcMessage("Error: You are trying to evaluate the equation of state with "
+                          + Utilities::to_string(out.densities.size()-1) +
+                          " compositional fields, which is larger than "
+                          + Utilities::to_string(number_of_compositions) +
+                          ", the number of fields the equation if state was set up with "
```
gassmoeller:
```suggestion
                          ", the number of fields the equation of state was set up with "
```

gassmoeller: And there is no longer a `set_number_of_compositions` function

source/material_model/equation_of_state/linearized_incompressible.cc
```diff
@@ -0,0 +1,159 @@
...
+        for (unsigned int c=0; c < out.densities.size(); ++c)
+          {
+            out.densities[c] = reference_rho * (1 - thermal_alpha * (in.temperature[q] - reference_T));
+            if (c>0)
+              out.densities[c] += compositional_delta_rhos[c-1];
+
+            out.thermal_expansion_coefficients[c] = thermal_alpha;
+            out.specific_heat[c] = reference_specific_heat;
+            out.compressibilities[c] = 0.0;
+            // Pressure derivative of entropy at the given positions.
```
gassmoeller: Remove these comments, they belong in the header file for this variable.

source/material_model/composition_reaction.cc
```diff
@@ -63,16 +64,18 @@ namespace aspect
...
-                            :
-                            0.0;
-          const double c2 = composition.size()>1?
-                            std::max(0.0, composition[1])
-                            :
-                            0.0;
-          out.densities[i] = reference_rho * (1 - thermal_alpha * (temperature - reference_T))
-                             + compositional_delta_rho_1 * c1 + compositional_delta_rho_2 * c2;
+          // The Composition reaction model has up to two compositional field that can influence the density
+          EquationOfStateOutputs<dim> eos_outputs (in.composition[i].size()>2 ? 3 : in.composition[i].size() + 1);
```
gassmoeller: Creating this object is somewhat slow. Can we move it outside of the loop over `i`? Same for the other material models.

include/aspect/material_model/equation_of_state/interface.h
jdannberg: Yes, that's the idea. As the latent heat release can be modelled by modifying either alpha and c_p, or by filling the entropy derivatives, and these two ways are more or less equivalent, I think they should both be in the equation of state.

include/aspect/material_model/equation_of_state/linearized_incompressible.h
jdannberg: What did you have in mind?

They didn't have any documentation in the original material model (and don't have a documentation in most other models). I always liked it that way because I thought it was pretty self-explanatory (the header of the class lists the equation of the density), and I don't really know what to document here except for something like 
```
// The reference density $\rho_0$ used in the computation of the density. 
double reference_rho;
```
which doesn't add much information.

source/material_model/equation_of_state/linearized_incompressible.cc
jdannberg: Oh, can I do that for other material models as well?
I'll get lots of pull requests in!

include/aspect/material_model/equation_of_state/linearized_incompressible.h
gassmoeller: I guess you are right. There is a limit where documentation for obvious things is just not useful anymore. Ignore my comment.

source/material_model/equation_of_state/linearized_incompressible.cc
gassmoeller: Yes, I think there are a lot of comments that actually should be in the header, and that will hopefully be reduced by deduplicating all the EoS code.

include/aspect/material_model/equation_of_state/interface.h
```diff
@@ -0,0 +1,98 @@
...
+    {
+      /**
+       * Constructor. Initialize the various arrays of this structure with the
+       * given number of compositions.
+       *
+       * @param n_comp The number of vector quantities (in the order in which
+       * the Introspection class reports them) for which input will be
+       * provided, and outputs should be filled.
+       */
+      EquationOfStateOutputs (const unsigned int n_comp);
```
tjhei: maybe rename as n_compositions? This has nothing to do with an ASPECT composition, so maybe we can find a better name? You are calling it ``n_points`` in the .cc...

source/material_model/composition_reaction.cc
```diff
@@ -36,6 +37,11 @@ namespace aspect
     {
       ReactionRateOutputs<dim> *reaction_rate_out = out.template get_additional_output<ReactionRateOutputs<dim> >();
 
+      // The Composition reaction model has up to two compositional fields (plus one background field)
+      // that can influence the density
+      const unsigned int n_compositions_for_eos = std::min(this->n_compositional_fields()+1, static_cast<unsigned int>(3));
```
tjhei:
```suggestion
      const unsigned int n_compositions_for_eos = std::min(this->n_compositional_fields()+1, 3u);
```

source/material_model/composition_reaction.cc
```diff
@@ -63,16 +69,16 @@ namespace aspect
...
-                            0.0;
-          out.densities[i] = reference_rho * (1 - thermal_alpha * (temperature - reference_T))
-                             + compositional_delta_rho_1 * c1 + compositional_delta_rho_2 * c2;
+          equation_of_state.evaluate(in, i, eos_outputs);
+
+          std::vector<double> volume_fractions (n_compositions_for_eos, 1.0);
+          for (unsigned int c=0; c<n_compositions_for_eos-1; ++c)
+            {
+              volume_fractions[c+1] = std::max(0.0, in.composition[i][c]);
+              volume_fractions[0] -= volume_fractions[c+1];
```
tjhei: do we need to ensure this value stays positive? You could add a max(..) after the loop.

source/material_model/equation_of_state/interface.cc
```diff
@@ -0,0 +1,51 @@
...
+
+#include <aspect/material_model/equation_of_state/interface.h>
+
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    template <int dim>
+    EquationOfStateOutputs<dim>::EquationOfStateOutputs(const unsigned int n_points)
```
tjhei: use the same variable name as in header.

source/material_model/equation_of_state/linearized_incompressible.cc
```diff
@@ -0,0 +1,156 @@
...
+    {
+      template <int dim>
+      void
+      LinearizedIncompressible<dim>::
+      evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
+               const unsigned int q,
+               MaterialModel::EquationOfStateOutputs<dim> &out) const
+      {
+
+        Assert(number_of_compositions+1 >= out.densities.size(),
```
tjhei: can you check if you can make this work with ``==``? If that does not work, would be more appropriate to rename it as ``maximum_number_of_compositions``?

tests/ascii_data_initial_composition_2d_box/screen-output
```diff
@@ -115,7 +115,7 @@ Number of degrees of freedom: 349 (162+25+81+81)
 
 *** Timestep 9:  t=732284 years
    Solving temperature system... 20 iterations.
-   Solving C_1 system ... 70 iterations.
+   Solving C_1 system ... 61 iterations.
```
tjhei: what the?!

source/material_model/composition_reaction.cc
jdannberg: We didn't do that in the old logic, and I didn't want to change functionality. If you think we should change it, I can make a separate pull request.

jdannberg:
I think I addressed all of your comments.

include/aspect/material_model/equation_of_state/linearized_incompressible.h
bangerth: I don't think I understand what `q` is supposed to be. You say "which entry of the vector of inputs is used", but the input is a structure and not an array. Can you clarify?

include/aspect/material_model/equation_of_state/linearized_incompressible.h
bangerth: Since you're already documenting these variables with `//`, why not just use `/** */` instead?

source/material_model/composition_reaction.cc
```diff
@@ -241,7 +213,7 @@ namespace aspect
 
       // Declare dependencies on solution variables
       this->model_dependence.viscosity = NonlinearDependence::none;
-      this->model_dependence.density = NonlinearDependence::none;
+      this->model_dependence.density = NonlinearDependence::temperature | NonlinearDependence::compositional_fields;
```
bangerth: Do you want to ask the EoS for the dependency? Maybe this whole kind of computation also needs to go into the EoS

source/material_model/simple.cc
```diff
@@ -192,12 +181,10 @@ namespace aspect
           if ( minimum_thermal_prefactor == 0.0 ) minimum_thermal_prefactor = std::numeric_limits<double>::min();
 
           k_value                    = prm.get_double ("Thermal conductivity");
-          reference_specific_heat    = prm.get_double ("Reference specific heat");
-          thermal_alpha              = prm.get_double ("Thermal expansion coefficient");
-          compositional_delta_rho    = prm.get_double ("Density differential for compositional field 1");
 
           if (thermal_viscosity_exponent!=0.0 && reference_T == 0.0)
             AssertThrow(false, ExcMessage("Error: Material model simple with Thermal viscosity exponent can not have reference_T=0."));
+
```
bangerth: no need for the empty line :-)

source/material_model/simple.cc
```diff
@@ -208,17 +195,12 @@ namespace aspect
       this->model_dependence.specific_heat = NonlinearDependence::none;
       this->model_dependence.thermal_conductivity = NonlinearDependence::none;
       this->model_dependence.viscosity = NonlinearDependence::none;
-      this->model_dependence.density = NonlinearDependence::none;
+      this->model_dependence.density = NonlinearDependence::temperature|NonlinearDependence::compositional_fields;
```
bangerth: same issue here, and maybe in other material models further down

tests/artificial_viscosity_list/solution/solution-00001.0000.gnuplot
```diff
@@ -5,21 +5,21 @@
 # For a description of the GNUPLOT format see the GNUPLOT manual.
 #
 # <x> <y> <velocity> <velocity> <p> <T> <C_1> <artificial_viscosity> <artificial_viscosity_composition> 
-6.336e+06 0 0 1.36111e-11 -2.67379 973 1 0 0.000152425 
-6.12011e+06 1.63988e+06 -1.34172e-09 5.00735e-09 2.75252 973 0.999998 0 0.000152425 
+6.336e+06 0 0 1.35985e-11 -2.67379 973 1 0 0.000152425
```
bangerth: do you know why this changes?

source/material_model/composition_reaction.cc
jdannberg: I don't think these dependencies are very useful. I can't see why anyone would run a model where the density does not depend on at least the temperature or the composition, and all other material models simply assume that this is the case, instead of checking the parameter values. I am not even sure where this is used...

Conceptually, this is something that belongs to the equation of state, but I simply don't think it's worth the effort, and it would add more code.

bangerth: Fair enough :-)

tests/artificial_viscosity_list/solution/solution-00001.0000.gnuplot
jdannberg: It doesn't change by much, so my suspicion is that it may just be due to the change in the order we compute things in. I think all of the tests that change are with the simple model and have at least one compositional field (but not necessarily one that changes the density). Before, we computed the density as rho = rho(T) + Delta_rho * C. Now we compute the density as rho0 = rho(T), rho1=rho(T) + Delta_rho, rho=average(rho0, rho1). 

Analytically, the old and the new computations should be the same, but numerically, there could be small differences that lead to small changes in iteration count etc. as we see in the tests.

jdannberg:
> You are duplicating fields between the MaterialModelOutputs and EquationOfStateOutputs. Why not derive the former from the latter?

I am not sure that would work in the way you suggest. 
The vectors in the EquationOfStateOutputs have one entry per compositional field, whereas the MaterialModelOutputs have one entry per location/quadrature point.

tests/artificial_viscosity_list/solution/solution-00001.0000.gnuplot
jdannberg: I looked into this a bit more, and there are cookbooks that use the simple model, compositional fields, and a density difference for that compositional field, that remain unchanged (like point_value_01, which is a variation of the sinker benchmark). So I think the density computation is correct.

bangerth:
On 5/29/19 1:00 PM, Juliane Dannberg wrote:
> 
> The vectors in the EquationOfStateOutputs have one entry per compositional 
> field, whereas the MaterialModelOutputs have one entry per location/quadrature 
> point.

Ah, I see. I missed that.

tests/artificial_viscosity_list/solution/solution-00001.0000.gnuplot
bangerth: Hm, ok.

bangerth:
Please address my remaining comments (mostly doc issues), squash/fixup, and then this is good to go. I've looked at the changes again and they do seem small. Since all testers use numdiff, it's also possible that most of these changes actually predate your patch and were just not picked up before.

jdannberg:
I've addressed your changes and squashed the commits into logical units.

PR #3037, opened 2019-05-26
Refactor writing data in the viz postprocessor.

bangerth:
This is the first step towards being able to writing face data (on the surface of the domain). There are no functional changes in this patch. The only thing I'm doing is breaking out a substantial chunk of work from `Visualization::execute()` into a function of its own (`Visualization::write_data_out_data()`).

I want to get this part out of the way before I make functional changes because the large-scale code motion without functionality change will be difficult to review if I also mix in functionality changes.

(The new function has a template argument that I don't currently use but will in the follow-up patches.)

### DIFF
--- a/include/aspect/postprocess/visualization.h
+++ b/include/aspect/postprocess/visualization.h
@@ -490,6 +490,14 @@ namespace aspect
         bool output_mesh_velocity;
 
         /**
+         * File operations can potentially take a long time, blocking the
+         * progress of the rest of the model run. Setting this variable to
+         * 'true' moves this process into a background thread, while the
+         * rest of the model continues.
+         */
+        bool write_in_background_thread;
+
+        /**
          * Set the time output was supposed to be written. In the simplest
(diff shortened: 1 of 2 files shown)


include/aspect/postprocess/visualization.h
```diff
@@ -562,6 +563,16 @@ namespace aspect
                                  const std::string &solution_file_prefix,
                                  const std::vector<std::string> &filenames);
 
+
+        /**
+         * A function, called from the execute() function, that takes a
+         * DataOut object, does some preliminary work with it, and then
+         * writes the result out to files via the writer() function (in the
+         * case of VTU output) or through the XDMF facilities.
+         */
```
tjhei: please document the return value

bangerth: I've already got follow-up patches in the queue so have put the documentation into the following commits. I hope that's ok with you.

bangerth:
Nevermind, I've made the change in this patch already.

PR #3038, opened 2019-05-26
Replace world builder submodule

MFraters:
This replaces the world builder submodule with the files itself. This way no git clone --recrusive is needed anymore and we can make tests for the coupling between aspect and the world builder.


MFraters:
This is a follow up to #2613

CMakeLists.txt
```diff
@@ -418,18 +418,16 @@ IF (ASPECT_PRECOMPILE_HEADERS)
 ENDIF()
 
 # Check whether we can find the WorldBuilder and enable it.
-SET(ASPECT_WITH_WORLD_BUILDER ON CACHE BOOL "If ON the World Builder is compiled togheter with ASPECT.")
-IF (NOT EXISTS ${CMAKE_SOURCE_DIR}/contrib/WorldBuilder/VERSION)
-  MESSAGE("-- World Builder not found.")
-  SET(ASPECT_WITH_WORLD_BUILDER OFF CACHE BOOL "" FORCE)
+SET(WORLD_BUILDER_SOURCE_DIR "" CACHE PATH "Provide an external World Builder directory to be compiled with ASPECT. If the path is not not provided or the World Builder is not found in the provided location, the version in the contrib folder is is used.")
```
tjhei: break line please

CMakeLists.txt
```diff
@@ -418,18 +418,16 @@ IF (ASPECT_PRECOMPILE_HEADERS)
...
+IF (NOT EXISTS ${WORLD_BUILDER_SOURCE_DIR}/VERSION)
+  MESSAGE(STATUS "World Builder not found using internal version.")
+  SET(WORLD_BUILDER_SOURCE_DIR "${CMAKE_SOURCE_DIR}/contrib/world_builder/" CACHE PATH "" FORCE)
```
tjhei: I thought the path is ``WorldBuilder`` right now? Is there a reason you are changing it?

CMakeLists.txt
```diff
@@ -442,7 +440,6 @@ IF (ASPECT_WITH_WORLD_BUILDER)
     SET_PROPERTY(SOURCE ${_source_file}
                  APPEND PROPERTY COMPILE_DEFINITIONS ASPECT_USE_WORLD_BUILDER=1)
   ENDFOREACH()
-ENDIF()
```
tjhei: we will need to fix the indentation in this file...

CMakeLists.txt
MFraters: I changed it to `world_builder` to be more consistent with the naming conventions of aspect

MFraters:
> Another thing worth discussion is if we should filter some of the files that are not needed. The total size if 4MB right now.

hmm, is 4MB a problem? The four largest folders are: include (1,2MB), source (632KB), tests (1,1MB) and doc (1,1MB). So the bare minimum is ~2MB. I personally would think it would be a good idea to also link the tests to the aspect testers, so that would be ~3MB. Furthermore, I also think it would be good having the documentation supplied with the code. But if there are strong opinions on it I am willing to remove parts.

tjhei:
please rebase to current master.

tjhei:
can you rebase again (we had a problem on master)

MFraters:
oke, I have rebased it

tjhei:
```
CMake Error at CMakeLists.txt:430 (IF):

  if given arguments:

    "NOT" "EXISTS" "CACHE" "PATH" "Provide an external World Builder directory to be compiled with ASPECT. " "If the path is not not provided or the World Builder is not found in the provided location, the version in the contrib folder is is used./VERSION"

  Unknown arguments specified
```

MFraters:
This is confusing. Somehow the unity build doesn't work, while the normal build does work. And at least some of the errors in the unity build are completely unrelated to the world builder. Any ideas?

tjhei:
> Any ideas?

Below is the error. Basically, WorldBuilder::Point and dealii::Point are conflicting:
```
/home/jenkins/agent/workspace/aspect_PR-3038/include/aspect/initial_temperature/adiabatic.h:52:43: error: reference to ‘Point’ is ambiguous
         double initial_temperature (const Point<dim> &position) const;
                                           ^
In file included from /home/dealii/dealii-v9.0.1/include/deal.II/base/array_view.h:23:0,
                 from /home/dealii/dealii-v9.0.1/include/deal.II/base/mpi.h:20,
                 from /home/dealii/dealii-v9.0.1/include/deal.II/base/timer.h:21,
                 from /home/jenkins/agent/workspace/aspect_PR-3038/include/aspect/simulator.h:25,
                 from /home/jenkins/agent/workspace/aspect_PR-3038/source/simulator/helper_functions.cc:22,
                 from /home/jenkins/agent/workspace/aspect_PR-3038/build-gcc-unity/cotire/aspect_CXX_unity.cxx:15:
/home/dealii/dealii-v9.0.1/include/deal.II/base/tensor.h:40:43: note: candidates are: template<int dim, class Number> class dealii::Point
 template <int dim, typename Number> class Point;
                                           ^
In file included from /home/jenkins/agent/workspace/aspect_PR-3038/contrib/world_builder/include/world_builder/parameters.h:32:0,
                 from /home/jenkins/agent/workspace/aspect_PR-3038/contrib/world_builder/include/world_builder/world.h:23,
                 from /home/jenkins/agent/workspace/aspect_PR-3038/source/initial_temperature/world_builder.cc:23,
                 from /home/jenkins/agent/workspace/aspect_PR-3038/build-gcc-unity/cotire/aspect_CXX_unity.cxx:363:
/home/jenkins/agent/workspace/aspect_PR-3038/contrib/world_builder/include/world_builder/point.h:37:9: note:                 template<int dim> class WorldBuilder::Point
```

gassmoeller:
The crucial part of the error messages is this:
```
In file included from /home/rene/software/aspect/debug/cotire/aspect_CXX_unity.cxx:855:
/home/rene/software/aspect/source/initial_composition/world_builder.cc:35:32: error: reference to ‘Point’ is ambiguous
     initial_composition (const Point<dim> &position, const unsigned int n_comp) const
```
You have a `Point` class in the WorldBuilder and there is a Point class in deal.II, and in many functions we use an unqualified `Point` when referring to the deal.II point class. There is a `using dealii` statement in many of our headers, so I do not quite know why the unity build fails, but maybe this helps you figure out the problem.

MFraters:
I guess because with the unity build both the using deal.ii (from the aspect files) and using worldbuilder (from the world builder fuiles) are put into one file, which causes the confusion

tjhei:
> are put into one file, which causes the confusion

correct. Let me know if you need help in fixing this.

MFraters:
It seems to be fixed now!

MFraters:
So I think there are a few things which still have to/can be done in this pull request.

1. make tests for the world builder plugins
2. connect the world builder tests to the aspect tester (?)
3. add a changelog entry

New cookbooks can be done in later pull requests. Anything else?

MFraters:
Talked with @tjhei and we decided to only do 3 for this pull request (which is now done), and do the other points in different pull requests. So this pull request should be ready to merge now.

gassmoeller:
Hmm, I would want to move forward with this, but I am really afraid of either reviewing 80k lines of code, or merging 80k lines of code without know what is happening inside. Did you discuss this with Timo at the hackathon already? Also: Is the version of the world build that is in here a certain release? Or is this version of the world build the official development version of it? Do we every know and then update the included version or is it regularly updated?

tjhei:
> . Did you discuss this with Timo at the hackathon already?

We talked about it. But I have similar concerns as it is not tested thoroughly and it is a huge changeset.

The reason for this PR is #2613. As this was started by @bangerth, I think he should comment on this PR before we move on.

> Also: Is the version of the world build that is in here a certain release?

This is a recent development version.

> Do we every know and then update the included version or is it regularly updated?

I think the thought is to update this occasionally (when major changes happen). If a user needs a newer version, she can always point to a custom checkout.

MFraters:
> Hmm, I would want to move forward with this, but I am really afraid of either reviewing 80k lines of code, or merging 80k lines of code without know what is happening inside. Did you discuss this with Timo at the hackathon already? Also: Is the version of the world build that is in here a certain release? Or is this version of the world build the official development version of it? Do we every know and then update the included version or is it regularly updated?

I understand you hesitation. In reality most of those lines are probably testing results since the coverage testers reports that there are 2542 relevant lines in the code. This may be an underestimation since it won't count lines with a single bracket, but still quite a bit less than 80k :)

What I would propose is that I explain the structure of the code to you when you are back in Davis, because that will probably also help a lot. The most relevant part of the code for aspect are probably a few 100 lines.

I am planning to release verison 0.2.0 when the paper is accepted. I am fine with waiting until then to merge the code so that aspect ships with a full release and not a development version.

>The reason for this PR is #2613. As this was started by @bangerth, I think he should comment on this PR before we move on.

My understanding was that we agreed that this was the best approach for when the code became stable, since the whole submodule thing is not really working well.

> Do we every know and then update the included version or is it regularly updated?
>>I think the thought is to update this occasionally (when major changes happen). If a user needs a newer version, she can always point to a custom checkout.
Yes, the plan is to no update for every small fix, but probably only for every release.

gassmoeller:
That sounds reasonable, lets talk about the details when I am back (next week Friday evening, so maybe the week after). I am not extremely concerned about the review, as long as we use a reasonable version, and make sure it works well with the current ASPECT version. Having this as a bundled package (like the muparser or boost in deal.II) sounds reasonable.

tjhei:
> That sounds reasonable, lets talk about the details when I am back

sound good. Menno has my okay with this PR but we should wait for @bangerth to comment.

bangerth:
I was thinking of this in the same way as we do in the `bundled/` directory of deal.II: These are packages that are externally maintained, with their own styles, formatting, bugs, etc. I suspect that *WorldBuilder* would benefit from a code review, but I don't see it as necessary for ASPECT that we review every line of what we import here -- it's just a copy of an upstream package.

It's possible that we let WorldBuilder off the hook too easily this way, but I'm ok with importing things into directories where we clearly state that this is not ASPECT code.

MFraters:
Since version 0.2.0 is officially released, I have now updated this pull request. I also changed the setup a bit so that even though the wb files are now always there, the user can still decide whether to compile aspect with the world builder or not.

I have structured my commits i a way which should help review. The first one just deletes the submodule, the second commit just add the files and the commits after change aspect files. I have tested building locally both the version where ASPECT_WITH_WORLD_BUILDER is ON and OFF and the world builder test works fine when it is ON and is removed when it is OFF.

MFraters:
Hey Rene,

Thanks for the quick review! I think I have addressed all your comments on the code itself, and found and fix another problem I found. 

For the files, I think that it can be useful to have the documentation included, but I do not feel to strongly about it since I suspect that in practice it is easier to look up the manual online that to compile the pdf's yourself (as long as you have internet access). I was thinking to add the possibility of adding the world builder test to aspect, but I found out that that requires some restructuring of the cmake files in the world builder, so that will not happen for this version of the world builder.

MFraters:
I had missed one comment (about the test description). Now they are all addressed :)

PR #3040, opened 2019-05-26
Further refactor the viz postprocessor.

bangerth:
The patch looks complicated, but really does just three things: 
* It groups a bunch of member variables of the viz postprocessor class into a structure
* It makes a couple of member functions work on the members of this structure (obtained via a reference argument) instead of the former member variables of the surrounding class
* Create a member variable of this structure.

I do this because it will, in the next step, allow me to create a second one of these member variables of struct type so that I can deal with both cell and face output.

### DIFF
--- a/source/postprocess/visualization.cc
+++ b/source/postprocess/visualization.cc
@@ -27,6 +27,7 @@
 
 #include <deal.II/dofs/dof_tools.h>
 #include <deal.II/numerics/data_out.h>
+#include <deal.II/numerics/data_out_faces.h>
 
 #include <boost/archive/text_oarchive.hpp>
 #include <boost/archive/text_iarchive.hpp>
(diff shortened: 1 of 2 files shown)


source/postprocess/visualization.cc
```diff
@@ -343,8 +373,13 @@ namespace aspect
     template <int dim>
     template <typename DataOutType>
     std::string
-    Visualization<dim>::write_data_out_data(DataOutType &data_out)
+    Visualization<dim>::write_data_out_data(DataOutType &data_out,
+                                            OutputHistory                  &output_history) const
```
tjhei: this space before the variable name looks weird, can you fix?

bangerth:
Spacing fixed.

PR #3047, opened 2019-05-27
Fix: avoid MPI deadlock in constraints

tjhei:
In rare cases, the code to check if the ConstraintMatrix changed can
cause a deadlock (only some processes would invoke the MPI::sum to check
if the mesh changed).
For this to happen, the total number of DoFs has to stay the same during
refinement. Fix this by unconditionally doing the communication step.

### DIFF
--- a/doc/modules/changes/20190527_tjhei
+++ b/doc/modules/changes/20190527_tjhei
@@ -1,0 +1,3 @@
+New: Fix a rare deadlock caused by adaptive refinement in parallel runs.
+<br>
+(Timo Heister, 2019/05/27)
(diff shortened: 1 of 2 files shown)


tjhei:
@MFraters please check if this fixes your hang. If you have a small repo, we should make this a unit test (but please not with 480 MPI ranks).

MFraters:
Thanks for fixing this! I am now building a small testcase which hopefully reproduces the problem and I will then try the fix.

MFraters:
After a discussion with Timo, we decided not to invest time in making a specific fail case, because that will be hard. 

Although the pull request looks alright to me, and the tests pass, I do not feel confident enough with the code to merge it. So I think it would be good if someone else takes a look at it. I will rerun the model with this code, but it will take a about 3.63e5 seconds (~4.2 days) before I know whether it works...

tjhei:
I think this is okay to merge?

PR #3048, opened 2019-05-27
Added missing grid interpolator

bobmyhill:
InterpolatedUniformGridData is needed for some planned ascii data plugins, but is not yet implemented in deal.ii. This PR is a workaround for that problem. 

It has been tested using the new functionality in PR #3039.

### DIFF
--- a/include/aspect/compat.h
+++ b/include/aspect/compat.h
@@ -42,7 +42,7 @@ namespace aspect
    * deal.II (dealii/include/deal.II/base/function_lib.h)
    */
   template <int dim>
-  class InterpolatedUniformGridData : public dealii::Functions::InterpolatedUniformGridData<dim>
+  class InterpolatedUniformGridData : public dealii::Function<dim>
   {
     public:
       /**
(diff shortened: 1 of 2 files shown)


gassmoeller:
Specifically, this is for the `gradient` function of the class, correct? Also is there a corresponding PR for deal.II that you can link to from here?

include/aspect/compat.h
```diff
@@ -28,6 +28,353 @@
 #include <functional>
 #include <memory>
 
+#if !DEAL_II_VERSION_GTE(9,2,0)
```
tjhei: as it doesn't exist in deal.II yet (right?), I think we need to use this unconditionally here. Otherwise, you need to have a ``using `` directive in the ``#else`` branch, right?

include/aspect/compat.h
```diff
@@ -28,6 +28,353 @@
...
+       * The data that is to be interpolated.
+       */
+      const Table<dim, double> data_values;
+
+  };
+}
+
+namespace aspect
+{
+  DEAL_II_DISABLE_EXTRA_DIAGNOSTICS
```
tjhei: what warnings are you suppressing?

source/utilities.cc
```diff
@@ -1727,9 +1727,9 @@ namespace aspect
         {
           if (coordinate_values_are_equidistant)
             data[i]
-              = std_cxx14::make_unique<Functions::InterpolatedUniformGridData<dim>> (grid_extent,
-                                                                                     table_intervals,
-                                                                                     data_tables[dim+i]);
+              = std_cxx14::make_unique<aspect::InterpolatedUniformGridData<dim>> (grid_extent,
```
bangerth: If you move the function into a namespace `aspect::Functions::`, then you can avoid this change and, once the class is in deal.II, we can just disable it in aspect without having to change this place.

include/aspect/compat.h
bobmyhill: this is to suppress "functions unused" warnings. We don't know exactly why they appear - the functions are definitely used.

bobmyhill:
> Specifically, this is for the `gradient` function of the class, correct? Also is there a corresponding PR for deal.II that you can link to from here?

Correct. There isn't a deal.II PR yet.

include/aspect/compat.h
tjhei: can you try marking your functions ``inline`` instead?

include/aspect/compat.h
bobmyhill: Ok, so I'll remove this for the current PR, and we'll add it again when the changes are adding to deal.II.

bangerth: Yes.

gassmoeller:
Is this a blocker for #3039? If so, we should merge it and add the deal.II ifdef later. Otherwise we can leave it open until the deal.II PR is in.

bobmyhill:
> Is this a blocker for #3039? If so, we should merge it and add the deal.II ifdef later. Otherwise we can leave it open until the deal.II PR is in.

@gassmoeller I think we should merge this now.

gassmoeller:
That is not quite an answer to my question :wink:. Do you need this PR for #3039? If not, I would like to wait until the PR for deal.II is opened and do this right (including the ifdef). Otherwise somebody will have to come back in a few months and fix it, or it gets lost and this code sticks around forever.

bobmyhill:
We need it for a full implementation of what is intended for #3039. Without it, the ASCII data topography input will not work for some geometry models.

bangerth:
There's a dealii pull request here now: https://github.com/dealii/dealii/pull/8294

gassmoeller:
Alright, @bobmyhill please rebase and uncomment the ifdefs. The deal.II PR should hopefully be merged soon.

bangerth:
Thinking about it, we need to keep the class in unconditionally (without any `#if`) for the moment or those of us who have a deal.II dev version will have failures because the `#if` succeeds but the dev version doesn't have the patch in yet, or at least the version we have on our hard drive doesn't have it in yet.

include/aspect/compat.h
```diff
@@ -31,5 +31,357 @@
 // for std_cxx14::make_unique:
 #include <deal.II/base/std_cxx14/memory.h>
 
+#if !DEAL_II_VERSION_GTE(9,2,0)
```
bangerth: In other words, put a `//` in front of this line (and its corresponding ones below) and we remove it in a couple of weeks when deal.II dev has the missing functionality.

bobmyhill:
@gassmoeller Ok, done! :)

PR #3053, opened 2019-05-27
Rename a variable.

bangerth:
In #2972, @tjhei asked for a variable to be renamed but then merged without waiting for this to happen. This PR does the renaming.


bangerth:
Like many other PRs, only the supg test fails.

PR #3056, opened 2019-05-27
Provide an error message for a case we ran into this morning.

bangerth:
Would have saved us ten minutes of debugging.

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -848,6 +848,13 @@ namespace aspect
       double temp[2];
       Utilities::MPI::sum (my_temp, mpi_communicator, temp);
 
+      Assert (temp[1] > 0,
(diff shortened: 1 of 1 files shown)


source/simulator/helper_functions.cc
```diff
@@ -848,6 +848,13 @@ namespace aspect
       double temp[2];
       Utilities::MPI::sum (my_temp, mpi_communicator, temp);
 
+      Assert (my_area > 0,
```
tjhei: what if one processor has no boundary part? 
```suggestion
      Assert (my_temp[1] > 0.0,
```

bangerth: Oh! Yes, fixed!

tjhei:
Risky click, but only SUPG fails...

PR #3062, opened 2019-05-27
Use the \degree symbol for degrees in the parameter descriptions.

bangerth:
Requires #3061.

Done with the command
```
perl -pi -e 's/\{\}\^\\\\circ/\\degree/g;' `find source | egrep '\.cc$'`
```
which is one of those things that seems completely normal to me...


bangerth:
See also #3044 and #3045.

bangerth:
Should be fixed now.

Issue #2831, opened 2019-02-26
[META] adding/fixing a publication


tjhei:
Do you want to fix or add an entry to https://aspect.geodynamics.org/publications.html ?
You are at the right place. 

We track publications in the Mendeley group https://www.mendeley.com/community/aspect-1/ which we can add the entries to, if you leave a comment in this issue (or by requesting to be part of that group so you can add the entry yourself).

After adding to Mendeley, one of us needs to follow the step here:
https://github.com/geodynamics/aspect/blob/www/README.md#publication-list

We should regularly check google scholar for updates:
> https://scholar.google.com/scholar?cites=15800282755444821039&as_sdt=4005&sciodt=0,6&hl=en
> https://scholar.google.com/scholar?cites=14219378401242979903&as_sdt=4005&sciodt=0,6&hl=en

Pedrogea08:
Hi Timo!

I have just finished my master's thesis in which I have studied the Gibraltar slab dynamics through thermo-mechanical numerical modelling using ASPECT.  The work has been supervised by Flor de Lis Mancilla (University of Granada) and Ana Negredo (Complutense University of Madrid). I would like to contribute to the list of publications of ASPECT by adding my work. The BibTeX source to create the entry is:

@MastersThesis{Gea2022,
  author = {Pedro J. Gea},
  school = {University of Granada},
  title  = {Numerical modelling of physical processes involved in the recycling of the lithosphere},
  year   = {2022},
}

Also, attached you will find the full work in pdf format.

Thank you in advance,

Pedro 

[Numerical modelling of physical processes involved in the recycling of the lithosphere.pdf](https://github.com/geodynamics/aspect/files/8096313/Numerical.modelling.of.physical.processes.involved.in.the.recycling.of.the.lithosphere.pdf)

bangerth:
Hi Pedro:
first, congratulations for finishing your Master's!

Would you like to write a patch to add the entry yourself? It needs to be added to the file `doc/manual/citing_aspect.bib`.

Pedrogea08:
Thank you! I will add the entry to the bibtex file once the work is uploaded to my university repository in the next few weeks weeks, so I can add the URL to the full work.

tjhei:
TODO:
- [ ] https://se.copernicus.org/articles/13/849/2022/

Minerallo:
Hi everyone, 

is it fine if I leave a comment here for a new publication entry ?  
thank you in advance for your help !

[...]

PR #3064, opened 2019-05-27
The bibtex generated by Mendeley incorrectly categorizes several mast…

ljhwang:
The bibtex generated by Mendeley incorrectly categorizes several masters theses as phd theses.  I was unable to edit the metadata as Admin on Mendeley to correct.  After discussions with Timo, I corrected the file manually.  An issue will be open to change the workflow going forward to updating this file directly and NOT using Mendeley.

Updates the citing_aspect.bib file with the following corrections:
	▪	removes duplicate thesis entry for Blom
	▪	Changes Fraters2014, Schuurmans2018, Zelst2015, and Ellowitz2018 to @mastersthesis
	▪	changes JonathanPerry-Hout2019 URL to a URI. URL was not resolving

### DIFF
--- a/doc/manual/citing_aspect.bib
+++ b/doc/manual/citing_aspect.bib
@@ -1,4 +1,4 @@
-@phdthesis{Fraters2014,
+@mastersthesis{Fraters2014,
 author = {Fraters, Menno},
 file = {:home/heister/.local/share/data/Mendeley Ltd./Mendeley Desktop/Downloaded/Fraters - 2014 - Thermo-mechanically coupled subductionmodelling with ASPECT(2).pdf:pdf},
 number = {August},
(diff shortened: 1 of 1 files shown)


bangerth:
Ugh. Does Mendeley only classify things as "thesis", rather than a qualified versions thereof?

ljhwang:
Yes, that is what I surmised after brief research. When you “+Add" a new document using" New manual entry", the pull down only gives you the choice “Thesis” with no finer granularity.

Best,
-Lorraine

*****************************
Lorraine Hwang, Ph.D.
CIG (co-PI)
530.752.3656
Out of office: May 20 - June 1

tjhei:
We currently generate this from mendeley: #2831 
before we merge this we need to decide that we want to get rid of the mendeley workflow and update all instructions. Do we agree on this? I talked about this with @ljhwang and I am okay with this change.

PR #3067, opened 2019-05-28
fix melt global model for zero temperature

jdannberg:
The global melt model triggered a floating point exception for a zero reference temperature. Usually, a zero reference temperature is not allowed, except for the case where the thermal_viscosity_exponent is also zero (there is an Assert that checks that if the reference temperature is zero, the thermal_viscosity_exponent also has to be zero). 
This pull request makes this case actually work (without triggering the floating point exception).

### DIFF
--- a/source/material_model/melt_global.cc
+++ b/source/material_model/melt_global.cc
@@ -240,7 +240,7 @@ namespace aspect
               const double delta_temp = in.temperature[i]-this->get_adiabatic_conditions().temperature(in.position[i]);
               visc_temperature_dependence = std::max(std::min(std::exp(-thermal_viscosity_exponent*delta_temp/this->get_adiabatic_conditions().temperature(in.position[i])),1e4),1e-4);
             }
-          else
+          else if (thermal_viscosity_exponent != 0.0)
             {
               const double delta_temp = in.temperature[i]-reference_T;
               visc_temperature_dependence = std::max(std::min(std::exp(-thermal_viscosity_exponent*delta_temp/reference_T),1e4),1e-4);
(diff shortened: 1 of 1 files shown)


source/material_model/melt_global.cc
```diff
@@ -240,7 +240,7 @@ namespace aspect
               const double delta_temp = in.temperature[i]-this->get_adiabatic_conditions().temperature(in.position[i]);
               visc_temperature_dependence = std::max(std::min(std::exp(-thermal_viscosity_exponent*delta_temp/this->get_adiabatic_conditions().temperature(in.position[i])),1e4),1e-4);
             }
-          else
+          else if (thermal_viscosity_exponent != 0.0)
```
gassmoeller: Why do you check for `thermal_viscosity_exponent` and not for `reference_T`?

jdannberg: We initialize the temperature-dependence with `visc_temperature_dependence = 1.0`, which is the value that corresponds to no temperature-dependence of the viscosity (because `out.viscosities[i] *= visc_temperature_dependence`). 

Accordingly, if the thermal viscosity exponent is zero (meaning we don't want any temperature-dependence), we can just leave visc_temperature_dependence at its initial value. This also has the advantage that we never get to the case where we divide by `reference_T`, so we don't have to deal with adding another if statement. Logically, I found this easier to read. 

Technically, it probably wouldn't make a difference to check for `reference_T` instead, because there is an Assert in parse_parameters that makes sure that thermal_viscosity_exponent is zero whenever reference_T is zero. So if you want me to change it, I can do that.

gassmoeller:
Ok your explanation makes sense. I was not aware of the other Assert in the parse_parameters.

PR #3069, opened 2019-05-28
use natural_coordinates in boundary get_data_component

bobmyhill:
Spherical geometry models (e.g. sphere, chunk) can accept information which runs from -pi to pi or 0 to 2pi. For the new initial topography/ascii data plugin, there is 
1) a call to `spherical_to_cartesian_coordinates` before 
2) a call to `Utilities::get_data_coordinates`, which then calls 
3) `spherical_to_cartesian_coordinates`.

This results in a loss of information; we no longer know whether our ascii data runs from -pi to pi or 0 to 2pi. In this particular case, we don't need the conversions at all.

This PR fixes this problem.

### DIFF
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -2167,18 +2167,11 @@ namespace aspect
     {
       if (this->get_time() - first_data_file_model_time >= 0.0)
         {
-          Point<dim> internal_position = position;
+          const std::array<double,dim> natural_position = this->get_geometry_model().cartesian_to_natural_coordinates(position);
 
-          if (dynamic_cast<const GeometryModel::SphericalShell<dim>*> (&this->get_geometry_model()) != nullptr
-              || dynamic_cast<const GeometryModel::Chunk<dim>*> (&this->get_geometry_model()) != nullptr
-              || dynamic_cast<const GeometryModel::Sphere<dim>*> (&this->get_geometry_model()) != nullptr)
-            {
-              const std::array<double,dim> spherical_position =
-                Utilities::Coordinates::cartesian_to_spherical_coordinates(position);
-
-              for (unsigned int i = 0; i < dim; i++)
-                internal_position[i] = spherical_position[i];
-            }
+          Point<dim> internal_position;
+          for (unsigned int i = 0; i < dim; i++)
+            internal_position[i] = natural_position[i];
 
           const std::array<unsigned int,dim-1> boundary_dimensions =
             get_boundary_dimensions(boundary_indicator);
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -2173,6 +2173,12 @@ namespace aspect
           for (unsigned int i = 0; i < dim; i++)
             internal_position[i] = natural_position[i];
 
+          // The chunk model has latitude as natural coordinate. We need to convert this to colatitude
+          if (dynamic_cast<const GeometryModel::Chunk<dim>*> (&this->get_geometry_model()) != nullptr && dim == 3)
+            {
+              internal_position[2] = numbers::PI/2. - internal_position[2];
+            }
+
           const std::array<unsigned int,dim-1> boundary_dimensions =
             get_boundary_dimensions(boundary_indicator);
 
--- a/include/aspect/geometry_model/sphere.h
+++ b/include/aspect/geometry_model/sphere.h
@@ -129,6 +129,23 @@ namespace aspect
         bool
         point_is_in_domain(const Point<dim> &point) const;
 
+        /**
+               * Takes the Cartesian points (x,z or x,y,z) and returns standardized
+               * coordinates which are most 'natural' to the geometry model. For a sphere
+               * this is (radius, longitude) in 2d and (radius, longitude, latitude) in 3d.
+               */
+        virtual
+        std::array<double,dim> cartesian_to_natural_coordinates(const Point<dim> &position) const;
+
+        /**
+         * Undoes the action of cartesian_to_natural_coordinates, and turns the
+         * coordinate system which is most 'natural' to the geometry model into
+         * Cartesian coordinates.
+         */
+        virtual
+        Point<dim> natural_to_cartesian_coordinates(const std::array<double,dim> &position) const;
+
+
         static
         void
         declare_parameters (ParameterHandler &prm);
--- a/source/geometry_model/sphere.cc
+++ b/source/geometry_model/sphere.cc
@@ -23,6 +23,7 @@
 #include <aspect/geometry_model/initial_topography_model/zero_topography.h>
 
 #include <deal.II/grid/grid_generator.h>
+#include <aspect/utilities.h>
 
 #if !DEAL_II_VERSION_GTE(9,0,0)
 #include <deal.II/grid/tria_boundary_lib.h>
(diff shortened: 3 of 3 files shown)


gassmoeller:
I am not sure this is the correct approach. It is an incompatible change to the utility function, which might be used in other plugins (not in ASPECT). Also it seems like a reasonable convenience functionality to automatically convert the coordinates. Is there no way to fix the behavior in the initial topography/ascii data plugin without affecting other plugins?

bobmyhill:
> I am not sure this is the correct approach. It is an incompatible change to the utility function, which might be used in other plugins (not in ASPECT). Also it seems like a reasonable convenience functionality to automatically convert the coordinates. Is there no way to fix the behavior in the initial topography/ascii data plugin without affecting other plugins?

@gassmoeller  Is there a good way to tell the utilities function that we're setting Chunk initial topography? A fudge would be to pass a very large int for the component (the last variable in the function) and add an if statement for that int in the utilities function.

bobmyhill:
@gassmoeller How about this one?

gassmoeller:
Could you instead replace Utilities:2171-... : 
```
          if (dynamic_cast<const GeometryModel::SphericalShell<dim>*> (&this->get_geometry_model()) != nullptr
              || dynamic_cast<const GeometryModel::Chunk<dim>*> (&this->get_geometry_model()) != nullptr
              || dynamic_cast<const GeometryModel::Sphere<dim>*> (&this->get_geometry_model()) != nullptr)
            {
              const std::array<double,dim> spherical_position =
                Utilities::Coordinates::cartesian_to_spherical_coordinates(position);

              for (unsigned int i = 0; i < dim; i++)
                internal_position[i] = spherical_position[i];
            }
```
by
```
const std::array<double,dim> natural_position = this->get_geometry_model().cartesian_to_natural_coordinates(point);

    for (unsigned int i = 0; i < dim; i++)
                internal_position[i] = natural_position[i];

File: source/geometry_model/initial_topography_model/ascii_data.cc [hunk 1]
```diff
@@ -78,19 +78,16 @@ namespace aspect
         {
           // No need to set the radial coordinate correctly,
           // because it will be thrown away in get_data_component anyway
-          std::array<double, dim> point;
-          point[0] = 6371000.0;
+          global_point[0] = 6371000.0;
```
bangerth: This is pre-existing, but we should probably query the geometry model for the proper radius. Though in the comment above, it says that we don't really care, so we may set it to zero.

source/geometry_model/initial_topography_model/ascii_data.cc
```diff
@@ -78,19 +78,16 @@ namespace aspect
...
-          global_point = Utilities::Coordinates::spherical_to_cartesian_coordinates<dim>(point);
+            global_point[d+1] = surface_point[d];
         }
       else
         AssertThrow(false, ExcNotImplemented());
 
       const double topo = Utilities::AsciiDataBoundary<dim>::get_data_component(surface_boundary_id,
                                                                                 global_point,
-                                                                                0);
+                                                                                1000000);
```
bangerth: This was from a test, right?

bangerth: In other words, I think you want to revert this.

source/utilities.cc
```diff
@@ -2168,9 +2168,19 @@ namespace aspect
...
-          if (dynamic_cast<const GeometryModel::SphericalShell<dim>*> (&this->get_geometry_model()) != nullptr
-              || dynamic_cast<const GeometryModel::Chunk<dim>*> (&this->get_geometry_model()) != nullptr
-              || dynamic_cast<const GeometryModel::Sphere<dim>*> (&this->get_geometry_model()) != nullptr)
+          // In the initial topography ascii data plugin, a component variable value of 1000000
+          // is used to indicate that the conversion from cartesian to spherical coordinates
+          // is unnecessary even for models based on spherical symmetry.
+          // However, we still need to use the actual component value for the lookup.
+          unsigned int use_component = component;
+          if (component == 1000000)
+            use_component = 0;
```
bangerth: Oh, I see. No, this is awkward. Let's not do this.

bangerth:
I don't think this is the right way to go.

I have to admit that I don't understand the problem. Can you explain again which function calls which, and what coordinate systems they all use? In my view, the AsciiData classes don't actually care -- they just use a set of coordinates in whatever coordinate system one wants to interpret the data that was read from file.

source/geometry_model/initial_topography_model/ascii_data.cc
bobmyhill: zero would be a bad idea for a radius, as then phi and theta would not be defined. I was considering a value of 1, but actually we don't need to change this text now.

source/utilities.cc
bobmyhill: Ok! natural_coordinate conversion provide a much neater solution anyway.

bobmyhill:
> I don't think this is the right way to go.
> 
> I have to admit that I don't understand the problem. Can you explain again which function calls which, and what coordinate systems they all use? In my view, the AsciiData classes don't actually care -- they just use a set of coordinates in whatever coordinate system one wants to interpret the data that was read from file.

@gassmoeller and @anne-glerum provided a neat solution to this problem (now pushed).

gassmoeller:
@bangerth: Yes this is a complicated issue. @anne-glerum, @bobmyhill and I had a discussion about it, but in summary the Chunk Material model allows longitude coordinates between -pi and 2pi (and automatically detects if we use -pi to pi, or 0 to 2pi longitudes), and an AsciiData table can contain values (e.g. -pi/4 to +pi/4), but after converting several times between spherical and cartesian coordinates we limit longitudes to 0 to 2pi, so the data table has a different coordinate system than the model internally. The solution is to not use the `cartesian_to_spherical` function in Utilities, but the `cartesian_to_natural_coordinates` function from the chosen geometry model, because they know about such peculiarities (like allowing -pi to 2pi).

@bobmyhill you need to run `make indent`

source/utilities.cc
```diff
@@ -2167,17 +2167,10 @@ namespace aspect
...
-              || dynamic_cast<const GeometryModel::Sphere<dim>*> (&this->get_geometry_model()) != nullptr)
-            {
-              const std::array<double,dim> spherical_position =
-                Utilities::Coordinates::cartesian_to_spherical_coordinates(position);
-
-              for (unsigned int i = 0; i < dim; i++)
-                internal_position[i] = spherical_position[i];
-            }
+          for (unsigned int i = 0; i < dim; i++)
+            internal_position[i] = natural_position[i];
```
bangerth: This looks easier :-)

You initialize `internal_position` above, but then overwrite it here. Can you just move the declaration of the variable right above its use here, and omit the initializer?

bobmyhill: Yes!

gassmoeller:
You updated the test output, but did you make sure the new output is correct (or more reasonable than the old one)? I am just afraid we broke something accidentally.

bobmyhill:
The changes in the output are small, but I've just checked the vtu files and the solution has definitely changed. Eugh. To be continued...

gassmoeller:
While you are at it #3104 has been merged, so if you rebase that will at least fix the spherical shell tests.

source/utilities.cc
```diff
@@ -2167,17 +2167,18 @@ namespace aspect
...
-            {
-              const std::array<double,dim> spherical_position =
-                Utilities::Coordinates::cartesian_to_spherical_coordinates(position);
+          Point<dim> internal_position;
+          for (unsigned int i = 0; i < dim; i++)
+            internal_position[i] = natural_position[i];
 
-              for (unsigned int i = 0; i < dim; i++)
-                internal_position[i] = spherical_position[i];
+          if ((dynamic_cast<const GeometryModel::SphericalShell<dim>*> (&this->get_geometry_model()) != nullptr
```
gassmoeller: I would like to keep this function free of geometry specific stuff. Didnt the spherical shell cases work, and only the Chunk model had a change in test results?

gassmoeller:
Thanks for the effort, that was more complicated than expected. Let's take this solution. At some point I would like to get rid of the Chunk tests, maybe by modifying the according functions in the chunk geometry model, but that is not urgent.

PR #3050, opened 2019-05-27
/benchmark/ fixes

tjhei:
- fix a few warnings and errors in the benchmark folder
- make geoid-spectral-comparison run in a reasonable time

PR #3070, opened 2019-05-28
Fix test result

gassmoeller:
To fix #3050.


gassmoeller:
This is already fixed in #3054, so I will let Timo fix it.

PR #3073, opened 2019-05-29
parameter view: implement filtering

tjhei:
add a text box that filters parameters when pressing return inside it.

### DIFF
--- a/doc/parameter_view/parameters.js
+++ b/doc/parameter_view/parameters.js
@@ -23,7 +23,7 @@ demangleStrings();
 reorderList();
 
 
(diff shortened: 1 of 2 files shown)


doc/parameter_view/parameters.js
```diff
@@ -22,6 +22,44 @@
 demangleStrings();
 reorderList();
 
+
+// grab text from the search input field and only show parameters that contain the text:
```
bangerth: I though you were a strong advocate of starting with a capital letter :-)

tjhei:
fixed.

gassmoeller:
Let's start simple. Good to go.

PR #3074, opened 2019-05-29
Parameter view: + / -

tjhei:
show a + in front of expandable things and a - if it is expanded.

### DIFF
--- a/doc/parameter_view/parameters.css
+++ b/doc/parameter_view/parameters.css
@@ -36,6 +36,15 @@ ul {
   background-color: #ccc;
 }
 
+
+/* show a + in front of expandable things and a - if it is expanded: */
+.collapsible:not(.active)::before {
+    content: "[+]";
+}
+.active::before {
+    content: "[-]";
+}
+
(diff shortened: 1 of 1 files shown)


doc/parameter_view/parameters.css
```diff
@@ -36,6 +36,15 @@ ul {
   background-color: #ccc;
 }
 
+
+/* show a + in front of expandable things and a - if it is expanded: */
+.collapsible:not(.active)::before {
+    content: "[+]";
+}
+.active::before {
+    content: "[-]";
```
jdannberg: Can you make this a longer dash so that the plus and minus sines take up the same space?
```suggestion
    content: "[−]";
```

bangerth: Or be explicit: `&ndash;` or `&mdash;`

tjhei: > Or be explicit: `&ndash;` or `&mdash;`

that doesn't work, but @jdannberg had the right idea.

tjhei:
fixed.

PR #3075, opened 2019-05-29
Multicomponent equation of state

jdannberg:
This is another pull request based on #3036 (so only the last commit is new). 
It adds an equation of state for multicomponent models and uses it in 4 different material models. There is another material model where this could potentially be used, which is the diffusion_dislocation model. The difference here is that diffusion_dislocation only has one value for the specific heat (but a vector with entries for each composition for the densities and thermal expansivities). This may be worth changing. 
What do you think?


jdannberg:
This is now rebased to the current master.

bangerth:
There are two failing tests:
```
	656 - viscoelastic_bending_beam (Failed)
	659 - viscoelastic_plastic_bending_beam (Failed)
```

jdannberg:
I updated the test results. The only difference was a change in iteration count.

include/aspect/material_model/equation_of_state/multicomponent_incompressible.h
```diff
@@ -0,0 +1,118 @@
...
+           * density.
+           */
+          static
+          void
+          declare_parameters (ParameterHandler &prm,
+                              const double default_thermal_expansion = 3.5e-5);
+
+          /**
+           * Read the parameters this class declares from the parameter file.
+           * The optional parameter @p n_compositions determines the maximum
```
gassmoeller: There is no optional parameter here.

include/aspect/material_model/equation_of_state/multicomponent_incompressible.h
```diff
@@ -0,0 +1,118 @@
...
+           * may still depend on temperature or pressure. In the current
+           * context, compressibility means whether we should solve the continuity
+           * equation as $\nabla \cdot (\rho \mathbf u)=0$ (compressible Stokes)
+           * or as $\nabla \cdot \mathbf{u}=0$ (incompressible Stokes).
+           */
+          bool is_compressible () const;
+
+          /**
+           * Declare the parameters this class takes through input files.
+           * The optional parameter @p n_compositions determines the maximum
```
gassmoeller: There is a different optional parameter here.

include/aspect/material_model/equation_of_state/multicomponent_incompressible.h
```diff
@@ -0,0 +1,118 @@
...
+           * number of compositions the equation of state is set up with,
+           * and should have the same value as the parameter with the same
+           * name in the declare_parameters() function.
+           */
+          void
+          parse_parameters (ParameterHandler &prm);
+
+
+        private:
+          // Vector of reference densities $\rho_0$ with one entry per composition,
```
gassmoeller: Make these comments into doxygen documentation: /** ... */. Same for all parameters below.

source/material_model/equation_of_state/multicomponent_incompressible.cc
```diff
@@ -0,0 +1,136 @@
...
+                                                                      list_of_composition_names,
+                                                                      has_background_field,
+                                                                      "Thermal expansivities");
+
+        specific_heats = Utilities::parse_map_to_double_array (prm.get("Heat capacities"),
+                                                               list_of_composition_names,
+                                                               has_background_field,
+                                                               "Specific heats");
+      }
+
```
gassmoeller: Remove empty line

jdannberg:
I addressed your comments.

PR #3078, opened 2019-05-29
benchmarks/check.sh: skip original.prm

tjhei:
also cookbooks/check.sh while I am here...


bangerth:
I'd merge this if I understood what you're trying to do. Does the script execute for all .prm files it can find and you want to exclude some? Where are the ones from that you exclude?

bangerth:
@tjhei explained it to me in person. I've got a follow-up in #3083.

Issue #3065, opened 2019-05-27
Change Workflow to update /publications.html from Mendeley to direct edit of citing_aspect.bib


ljhwang:
The bibtex generated by Mendeley cannot distinguish between thesis types, in this case Masters vs. PhD.  I was unable to edit the metadata as Admin on Mendeley to correct.  Manually adding an entry in Mendeley only gives you the choice of "Thesis" from the drop down menu without further granularity (e.g. Ph.D., Master's, or Bachelors).

The workflow will need to be changed going forward FROM export of all entries from Mendeley TO updating the citing_aspect.bib file directly.

tjhei:
I agree. How about @geodynamics/aspect-developers ?

Also see #2831 #3064

bangerth:
So have you decided that you don't want to use Mendeley any more? I've never cared much for that workflow, but it did have advantages...

tjhei:
> So have you decided that you don't want to use Mendeley any more? I've never cared much for that workflow, but it did have advantages...

I think it doesn't provide a lot of advantages (any?) but I would like to hear other opinions about it.

gassmoeller:
I am fine either way. If Mendeley does not provide benefits over having a .bib file in the repo, then the latter is preferable, because simpler.

tjhei:
> simpler

yeah, it means I don't need to run mendeley on my computer, sync, sort, export, create a PR. We originally thought Mendeley would make it easier for others to add entries, but I don't think this is true. I like that creating a PR to the .bib allows us to review things properly.

bangerth:
I agree. As I mentioned, I've never particularly cared for the Mendeley part of the workflow. All it did was simply importing things into bibtex entries. But we can always use doi2bib.org for this now.

gassmoeller:
I think with #3080 and #3064 this is adressed. Anything else left to do? Do we document anywhere how to add papers to the list of ASPECT publications? I did not find any place, but that suggests we probably should. Maybe in the README.md, or on the website, and/or the manual?

tjhei:
I modified the instructions in #2831 so I consider this done. Thanks, @ljhwang !

PR #3080, opened 2019-05-29
[www] remove references to mendeley [WIP]

tjhei:
part of #3065, which need to be discussed first.


bangerth:
Let's come to a conclusion in #3065 before merging this.

bangerth:
Looks like we agreed over in #3065.

PR #3083, opened 2019-05-29
Better document a shell function.

bangerth:
Follow-up to #3078.

### DIFF
--- a/benchmarks/check.sh
+++ b/benchmarks/check.sh
@@ -47,7 +47,11 @@ run_coarse_prm ()
 }
 
 
-# run aspect on all .prm files in the current folder or any subdirectory
+# Run aspect on all .prm files in the current folder or any
+# subdirectory; however, exclude files named parameters.prm and
+# original.prm as these are created by previous ASPECT runs and placed
+# in the output directories (where they can't be run from since paths
+# don't match up any more).
 run_all_prms ()
 {
     for prm in `find . -name "*prm"`;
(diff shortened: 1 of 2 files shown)


benchmarks/check.sh
```diff
@@ -47,7 +47,11 @@ run_coarse_prm ()
 }
 
 
-# run aspect on all .prm files in the current folder or any subdirectory
+# run aspect on all .prm files in the current folder or any
```
tjhei:
```suggestion
# Run ASPECT on all .prm files in the current folder or any
```

cookbooks/check.sh
```diff
@@ -17,23 +17,29 @@ then
 fi
 
 
-# run aspect on all .prm files in the current folder
+# run aspect on all .prm files in the current folder; however,
```
tjhei:
```suggestion
# Run ASPECT on all .prm files in the current folder; however,
```

benchmarks/check.sh
bangerth: @tjhei: I can do that, but it would be the only comment in the whole file that satisfies this style. Which way do you want me to go?

tjhei: Well, you turned this into a sentence with a period at the end. I think it should be capitalized in that case. You can just push the "commit suggestion" button...

bangerth:
All fixed. @tjhei?

tjhei:
> All fixed. @tjhei?

:+1:

PR #3086, opened 2019-05-29
make manual compile

tjhei:


doc/manual/manual.tex
```diff
@@ -10630,10 +10630,10 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
 \end{equation}
 where $\rho_{0} = 3291~\text{kg m}^{-3}$ and $\alpha = 3.59 \times 10^{-5}~\text{K}^{-1}$ are the density and thermal expansivity corresponding to $T_{0} = 873~\text{K}$, $P$ = pressure and $K = 115.2~\text{GPa}$ is the bulk modulus. $J_1$ and $J_2$ are expressed as
 \begin{equation}
-\begin{align*}
+\begin{aligned}
 J_1(\tau_S^{\prime})= & J_U \left[ 1 + \frac{A_B[\tau_S^{\prime}]^{\alpha_B}}{\alpha_B} + \frac{\sqrt{2\pi}}{2} A_P~\sigma_P \left\{ 1-\text{erf}\left(\frac{\text{ln}[\tau_P^{\prime}/\tau_S^{\prime}]}{\sqrt{2}\sigma_P}\right)\right\}\right] \\
 J_2(\tau_S^{\prime}) = & J_U \frac{\pi}{2} \left[ A_B[\tau_S^{\prime}]^{\alpha_B} + A_P~\text{exp} \left(-\frac{\text{ln}^2[\tau_P^{\prime}/\tau_S^{\prime}]}{2\sigma_P^2}\right)\right] + J_U \tau_S^{\prime}
-\end{align*}
+\end{aligned}
```
bangerth: No, that's not the correct spelling.

tjhei: I think this is how you spell ``aligned``. At least latex seems to like it and I think this is what the author intended. What would you suggest then?

bangerth: We have hundreds of formulas in the manual that use the `{align}` or `{align*}` environment, and none with `{aligned}`. In other words, I don't know what would be wrong with the existing spelling?

tjhei: this doesn't compile as it is inside an ``begin{equation}``

bangerth: Ah, I missed that it was inside `\begin/end{equation}`. I had also never heard of `{aligned}`.

But then the `{equation}` seems unnecessary. Just stick with the `{align*}` environment and remove the `{equation}` environment surrounding it.

tjhei: sure, that makes sense. Fixed.

PR #3087, opened 2019-05-29
Add get_model_operator_options function in utilities so operator options don't get out of sync

SophieCoulson:
I added a get_model_operator_options function in utilities which returns "add|subtract|minimum|maximum|replace if valid", and then replaced the list with a call to this function in all the relevant interface.cc files. This way they cannot become out of sync if someone adds a new operator.


tjhei:
can you please fix the indentation? (``make indent``)

SophieCoulson:
Done!

tjhei:
Thanks. :+1:

PR #3088, opened 2019-05-29
Python contrib

MarineLasbleis:
I added a file with routines for python + a notebook to show how to use them. 
They help to read the statistics file + the parameter file + the outputs (if gnuplot) files. THere are routines to read the header, so that the user only needs to know the names of the variables they want to read or plot.

Since I added the notebook, I had to add an example of outputs. It is maybe not ideal, and I should maybe remove most of the outputs? I am not sure what is the best in that case. Maybe I could also provide an example of .prm file that would produce some output files to be read. Let me know what's the best.


MarineLasbleis:
I updated the PR. Now, the notebook run an instance of aspect with a very simple convection-box.prm and look at the data produced. No need to have them on a folder here then.

MarineLasbleis:
I updated the PR. Now, the notebook runs an instance of aspect with a very simple convection-box.prm and look at the data produced. No need to have them on a folder here then.

contrib/python/aspect_data.py
```diff
@@ -0,0 +1,136 @@
...
+    param_value = line[debut+1:]
+    
+    if '#' in param_value: 
+        fin = param_value.index("#")
+        param_value = param_value[:fin]
+
+    return param_name, param_value   
+    
+def read_parameter_file(file):
+    """ Read parameter file, and return a dictionnary with all values.
```
gassmoeller:
```suggestion
    """ Read parameter file, and return a dictionary with all values. 
```

contrib/python/aspect_data.py
```diff
@@ -0,0 +1,136 @@
...
+                section = section[:section.rindex(":")]
+            else:    
+                section = ""
+        elif line[0] == "subsection":
+            name_section = ' '.join(word for word in line[1:])
+            if section == "":
+                section = name_section
+            else:
+                section = section+": "+name_section
+            #print(section)
```
gassmoeller: I think you had this for debugging purposes, right? Please remove now.

contrib/python/convection-box.prm
```diff
@@ -0,0 +1,165 @@
+# A description of convection in a 2d box. See the manual for more information.
```
gassmoeller: Is this just a copy of the cookbook? Could you then instead reference the original .prm file in the python notebook instead of making a copy? This way we avoid the duplication. Duplicated files can always create problems if they get out of sync.

If it is just nearly a copy with some changes, you can do the following: Remove everything below the top comment, and then put in:

```
set Dimension = 2

include $ASPECT_SOURCE_DIR/cookbooks/convection-box.prm

set all parameters that have been changed from the original cookbook here, below the include statement
```

This uses the original cookbook as a base model, and then overwrites any parameters you want to change.

contrib/python/README.md
```diff
@@ -0,0 +1,11 @@
+Python contributions for ASPECT 
+==============================
+
+This folder contains some useful routines to read files with Python,
+using the numpy and pandas packages.
+
+To install Python with anaconda: 
+https://www.anaconda.com/distribution/
+
```
gassmoeller: Could you add a short description of how to start the ipython notebook?

contrib/python/convection-box.prm
MarineLasbleis: Oh, I did not know this trick! I'll do that.

MarineLasbleis: Do we need to keep the set Dimension = 2 for a specific reason? Or can I only include the prm file?

contrib/python/README.md
MarineLasbleis: done!

MarineLasbleis:
Thanks! If I end up using it more extensively and add more functions, I may create a separate github repository and a proper python package to be installed with pip/conda. But for now, I think this is useful as it is!

contrib/python/convection-box.prm
gassmoeller: Yes you need to keep the Dimension parameter, because it is needed before the `include` directive is resolved (the current version should not run).

gassmoeller:
I agree. The current version looks ready to merge once you add the Dimension parameter back into the .prm file.

contrib/python/convection-box.prm
MarineLasbleis: The current version is running correctly... So I'm surprised, it seems to be OK without including the Dimension in the prm file. Could you test?

gassmoeller: You are right, sorry for the confusion, I misremembered that. The problem is as I described, but if ASPECT does not find the `Dimension` parameter in such a file, it simply assumes 2 (i.e. if that were a 3D model it would have crashed). Then this is good to go.

PR #3089, opened 2019-05-29
Material model that sets a constant viscosity in the lithopshere

SophieCoulson:
New material model plug-in that sets a constant specified lithosphere viscosity above the lithsphere-asthenosphere boundary and takes the viscosity from a base model below this. The LAB can be specified by an ascii file or a maximum lithosphere depth value. All other material properties are taken from the base model.


bangerth:
I think you need to make the test run on a smaller mesh to avoid the addition of 16,722 lines of text :-)

bangerth:
Hm, actually the 16k lines are in a data file. Do we need this data only for the test, or is this generally useful?

SophieCoulson:
The data file is an example LAB map. But actually its already in aspect in the data/initial_temperature/lithosphere_mask from my previous PR, so i definitely don't need to repeat it. Should I just call it from there or move it to somewhere different?

gassmoeller:
Just use the file that is already in there. 16k lines is at the upper end of what we would like to have as data in the repo, but since I thought it is a useful dataset for potentially quite a number of users I merged the other PR that contained the data. I will take a closer look at the code next.

include/aspect/material_model/replace_lithosphere_viscosity.h
```diff
@@ -0,0 +1,148 @@
...
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    /**
+     * A material model that applies a given constant viscosity in the lithosphere.
+     * Viscosity below this is taken from a ''base mode''chosen from any of the
```
gassmoeller:
```suggestion
     * Viscosity below this is taken from a ''base model'' chosen from any of the
```

include/aspect/material_model/replace_lithosphere_viscosity.h
```diff
@@ -0,0 +1,148 @@
...
+     * other available material models. The replace lithosphere viscosity
```
gassmoeller:
```suggestion
     * other available material models. The 'replace lithosphere viscosity'
```

include/aspect/material_model/replace_lithosphere_viscosity.h
```diff
@@ -0,0 +1,148 @@
...
+
+        /**
+         * Function to compute the material properties in @p out given the
+         * inputs in @p in.
+         */
+        virtual
+        void
+        evaluate (const typename Interface<dim>::MaterialModelInputs &in,
+                  typename Interface<dim>::MaterialModelOutputs &out) const;
+        /**
```
gassmoeller: Add empty line

include/aspect/material_model/replace_lithosphere_viscosity.h
```diff
@@ -0,0 +1,148 @@
...
+         * Method to declare parameters related to depth-dependent model
```
gassmoeller: depth-dependent -> Replace ...

include/aspect/material_model/replace_lithosphere_viscosity.h
```diff
@@ -0,0 +1,148 @@
...
+         */
+        static void
+        declare_parameters (ParameterHandler &prm);
+
+        /**
+         * Method to parse parameters related to depth-dependent model
```
gassmoeller: same here

include/aspect/material_model/replace_lithosphere_viscosity.h
```diff
@@ -0,0 +1,148 @@
...
+         */
+        virtual void
+        parse_parameters (ParameterHandler &prm);
+
+        /**
+         * Method that indicates whether material is compressible. Depth dependent model is compressible
```
gassmoeller: Depth dependent -> ...

SophieCoulson:
I deleted the duplicated data file and also change the folder name to "-" rather than "_" so its in-keeping with the others.

include/aspect/material_model/replace_lithosphere_viscosity.h
```diff
@@ -0,0 +1,148 @@
...
+         * Method to calculate reference viscosity for the model. The reference
+         * viscosity is simply the reference  viscosity from the base model.
+         */
+        virtual double reference_viscosity () const;
+
+      private:
+        /**
+         * Reads in file containing input data in ascii format.
+         */
+        Utilities::AsciiDataLookup<2> lab_depths;
```
gassmoeller: This feels like a lot of duplication with the `LithosphereMask` initial temperature plugin. We can avoid that if you would create a new helper class in `include/aspect/initial_temperature.h` that looks like this:
```
namespace LABDepth
{
  class LABDepthLookup: public SimulatorAccess<dim>
    public:
        /**
         * Return LAB depth as a function of position (latitude and longitude). For the
         * current class, this function returns value from the text file. We read in
         * two dimensions so the third column (depth) is treated as data.
         */
        double
        ascii_lab (const Point<2> &position) const;

        /**
         * Declare the parameters this class takes through input files.
         */
        static
        void
        declare_parameters (ParameterHandler &prm);

        /**
         * Read the parameters this class declares from the parameter file.
         */
        void

private:
     all of the variables the class needs
}
```
You can then use this class from the initial temperature plugin *and* from the material model, and the two implementation always stay the same (you will need to include the initial temperature header here in the material model). Plus if somebody adds a new way to define the LAB both plugins immediately get access. 
Splt this up in two pull requests. The first only creates the class for initial temperature as I wrote above. The second is this one, which can then just use the new class.

source/initial_temperature/lithosphere_mask.cc
```diff
@@ -44,6 +44,7 @@ namespace aspect
         {
           const std::string filename = data_directory+LAB_file_name;
 
+          if (Utilities::MPI::this_mpi_process(this->get_mpi_communicator()) == 0)
```
gassmoeller: Instead of adding this line, change the next one to:
```
this->get_pcout() << ...
```
`pcout` is made to do exactly what you do with the if above (it stands for parallel cout).

PR #3090, opened 2019-05-30
manual: improve cmake plugin help

tjhei:
Clarify the instructions on how to configure a plugin (after helping a
user today).

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -8233,8 +8233,7 @@ The required heating model and changes to the material model are implemented in 
 (\url{cookbooks/inner_core_convection/inner_core_convection.cc}). To compile the file, do
 
 \begin{verbatim}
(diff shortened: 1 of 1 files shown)


doc/manual/manual.tex
```diff
@@ -11169,7 +11169,7 @@ \subsection{How to write a plugin}
   \end{itemize}
   You can then just run the commands
     \begin{verbatim}
- cmake -DAspect_DIR=/path/to/aspect .
+ cmake -DAspect_DIR=/path/to/aspect/build/ .
```
bangerth: Above you have omitted the trailing slash.

tjhei: good point, I prefer it with a slash. I hope this is acceptable.

bangerth: I've got no preference. This is ok with me.

PR #3091, opened 2019-05-30
use VectorTools::compute_global_error

tjhei:


tjhei:
I fixed a typo in the screen output "L2" instead of "L", so I had to update the test results.

PR #3092, opened 2019-05-30
AsciiDataLookup improvements

tjhei:


include/aspect/utilities.h
```diff
@@ -579,7 +579,8 @@ namespace aspect
          *
          * @param position The current position to compute the data (velocity,
          * temperature, etc.)
-         * @param component The index of the data column to be returned.
+         * @param component The index between 0 and @p components of the data
```
bangerth: What does `@p components` refer to here? It's not clear from the context.

tjhei:
I think this addresses your comment.

PR #3096, opened 2019-05-30
Initial temperature model based on anelastic Vs-T conversion

freddrichards:
A new initial temperature model to convert absolute shear wave velocity to temperature, accounting for non-linear behaviour related to anelastic deformation. This code is essentially the same as that written for the "yamauchi_takei_2016_anelasticity" benchmark.

### DIFF
--- a/include/aspect/initial_temperature/anelasticity_temperature.h
+++ b/include/aspect/initial_temperature/anelasticity_temperature.h
@@ -1,0 +1,145 @@
+/*
+  Copyright (C) 2016 - 2019 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _aspect_initial_temperature_anelasticity_temperature_h
+#define _aspect_initial_temperature_anelasticity_temperature_h
+
+#include <aspect/initial_temperature/interface.h>
+#include <aspect/utilities.h>
+#include <aspect/simulator_access.h>
+
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    using namespace dealii;
+
+    /**
+     * A class that describes an initial temperature field for a 2D or 3D shear wave velocity (Vs) model.
+     * Vs values are converted to temperature using the anelasticity parameterization of Yamauchi & Takei (2016).
+     *
+     * @ingroup InitialTemperatures
+     */
+    template <int dim>
+    class AnelasticVs2T : public Utilities::AsciiDataInitial<dim>, public Interface<dim>
+    {
+      public:
+        /**
+        * Constructor. Initialize variables.
+        */
+        AnelasticVs2T ();
+
+        /**
+        * Initialization function. This function is called once at the
+        * beginning of the program. Checks preconditions.
+        */
+        void
+        initialize ();
+
+        // Avoid -Woverloaded-virtual:
+        using Utilities::AsciiDataInitial<dim>::initialize;
+
+        /**
+        * Return the boundary temperature as a function of position. For the
+        * current class, this function returns value from the text files.
+        */
+        double
+        initial_temperature (const Point<dim> &position) const;
+
+        /**
+        * Declare the parameters for the input files.
+        */
+        static
+        void
+        declare_parameters (ParameterHandler  &prm);
+
+        /**
+        * Read the parameters from the parameter file.
+        */
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+
+        /**
+        * Function that assesses difference between input and calculated Vs for Brent minimization
+        */
+        double
+        fVs(double x, double depth, double absolute_Vs, double mu0, double dmudT,
+            double dmudP, double viscosity_prefactor, double activation_energy, double activation_volume,
+            double solidus_gradient, bool density_model_flag) const;
+
+        /**
+        * Function that assesses difference between input and calculated pressure for Brent minimization
+        */
+        double
+        fdV(double x, double bulk_modulus, double bulk_modulus_pressure_derivative, double pressure) const;
+
+        /**
+          * Function to calculate Vs using Yamauchi & Takei 2016 anelasticity parameterization
+        */
+        double
+        yamauchi_takei_Vs(double temperature, double depth, double mu0, double dmudT,
+                          double dmudP, double viscosity_prefactor, double activation_energy,
+                          double activation_volume, double solidus_gradient, bool density_model_flag) const;
+
+        /**
+        * Whether to remove temperature heterogeneity upper parts of model
+        */
+        double no_perturbation_depth;
+
+        /**
+        * Whether to use Yamauchi & Takei (2016) anelasticity parameterization
+        */
+        bool use_yamauchi_takei;
+
+        /**
+        * Whether to use original parameters published in Yamauchi & Takei (2016)
+        * or a more up-to-date version that accounts for non-linear pressure
+        * dependence of thermal expansivity and depth dependence of shear wave period
+        */
+        bool use_original_model;
+
+        /**
+        * Constant temperature to set where variations have been removed
+        */
+        double upper_temperature;
+
+
+        /**
+        * Shear modulus parameters
+        */
+        double mu0;
+        double dmudT;
+        double dmudP;
+
+        /**
+        * Viscosity parameters
+        */
+        double viscosity_prefactor;
+        double activation_energy;
+        double activation_volume;
+        double solidus_gradient;
+
+    };
+  }
+}
+
+#endif
--- a/source/initial_temperature/anelasticity_temperature.cc
+++ b/source/initial_temperature/anelasticity_temperature.cc
@@ -1,0 +1,421 @@
+/*
+  Copyright (C) 2016 - 2019 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+#include <aspect/global.h>
+#include <aspect/initial_temperature/anelasticity_temperature.h>
+#include <aspect/material_model/interface.h>
+#include <aspect/initial_composition/interface.h>
+#include <aspect/adiabatic_conditions/interface.h>
+#include <aspect/material_model/anelasticity_temperature.h>
+#include <aspect/simulator_access.h>
+
+#include <boost/lexical_cast.hpp>
+#include <boost/math/tools/minima.hpp>
+
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    template <int dim>
+    AnelasticVs2T<dim>::AnelasticVs2T ()
+    {}
+
+    template <int dim>
+    void
+    AnelasticVs2T<dim>::initialize ()
+    {
+      Utilities::AsciiDataInitial<dim>::initialize(1);
+    }
+
+     // set up Vs function that Brent minimization operates on
+    template <int dim>
+    double
+    AnelasticVs2T<dim>::
+    fVs (double x,
+         const double      depth,
+         const double      absolute_Vs,
+         const double      mu0,
+         const double      dmudT,
+         const double      dmudP,
+         const double      viscosity_prefactor,
+         const double      activation_energy,
+         const double      activation_volume,
+         const double      solidus_gradient,
+         const bool      use_original_model) const
+    {
+      return std::abs(yamauchi_takei_Vs(x,depth,mu0,dmudT,dmudP,viscosity_prefactor,activation_energy,activation_volume
+                                        ,solidus_gradient,use_original_model)-absolute_Vs);
+    }
+
+    // set up volume change function that Brent minimization operates on
+    template <int dim>
+    double
+    AnelasticVs2T<dim>::
+    fdV (double x, const double bulk_modulus, const double bulk_modulus_pressure_derivative, const double pressure ) const
+    {
+      return std::abs((bulk_modulus*(3./2.)*(std::pow(x,7./3.)-std::pow(x,5./3.))*(1+(((3./4.)*
+                                                                                      (bulk_modulus_pressure_derivative-4))*(std::pow(x,2./3.)-1))))-pressure);
+    }
+
+    // set up initial temperature
+    template <int dim>
+    double
+    AnelasticVs2T<dim>::
+    initial_temperature (const Point<dim> &position) const
+    {
+      // determine depth
+      double depth;
+      depth = this->get_geometry_model().depth(position);
+
+      // declare temperature
+      double temperature;
+
+      // read absolute Vs in from ascii file
+      const double absolute_Vs = Utilities::AsciiDataInitial<dim>::get_data_component(position,0);
+
+      if (depth >= no_perturbation_depth)
+        {
+          // convert absolute Vs into temperature
+          // check if using Yamauchi & Takei 2016 parameterization
+          if (use_yamauchi_takei == true)
+            {
+              // specify Brent algorithm parameters
+              const double a=273;
+              const double b=3273;
+              typedef std::pair<double, double> Result;
+              // create fVs function to use in Brent minimization and calculate temperature
+              auto bfunc = [ &,this] (double x)
+              {
+                return fVs(x, depth, absolute_Vs, mu0, dmudT,dmudP,
+                           viscosity_prefactor,activation_energy,activation_volume,solidus_gradient,use_original_model);
+              };
+              // determine maximum Vs
+              double maximum_Vs;
+              int fail;
+              maximum_Vs=yamauchi_takei_Vs(273.,depth,mu0,dmudT,dmudP,viscosity_prefactor,activation_energy,activation_volume
+                                           ,solidus_gradient,use_original_model);
+
+              // set number of fails to zero
+              fail=0;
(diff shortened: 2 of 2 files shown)


include/aspect/initial_temperature/anelasticity_temperature.h
```diff
@@ -0,0 +1,145 @@
...
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+
+        /**
+        * Function that assesses difference between input and calculated Vs for Brent minimization
+        */
+        double
+        fVs(double x, double depth, double absolute_Vs, double mu0, double dmudT,
```
gassmoeller: please make all parameters `const` to make sure they are only used for input

include/aspect/initial_temperature/anelasticity_temperature.h
```diff
@@ -0,0 +1,145 @@
...
+            double dmudP, double viscosity_prefactor, double activation_energy, double activation_volume,
+            double solidus_gradient, bool density_model_flag) const;
+
+        /**
+        * Function that assesses difference between input and calculated pressure for Brent minimization
+        */
+        double
+        fdV(double x, double bulk_modulus, double bulk_modulus_pressure_derivative, double pressure) const;
```
gassmoeller: same here and for the other functions below

include/aspect/initial_temperature/anelasticity_temperature.h
gassmoeller: Also unless you plan to use the functions outside of this class: since the functions are part of the class, you do not need to hand over member variables of this class to the function. You can access the member variables directly inside the function.

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+
+    // set up initial temperature
+    template <int dim>
+    double
+    AnelasticVs2T<dim>::
+    initial_temperature (const Point<dim> &position) const
+    {
+      // determine depth
+      double depth;
+      depth = this->get_geometry_model().depth(position);
```
gassmoeller: simplify this to
```suggestion
      const double depth = this->get_geometry_model().depth(position);
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+      if (depth >= no_perturbation_depth)
+        {
+          // convert absolute Vs into temperature
+          // check if using Yamauchi & Takei 2016 parameterization
+          if (use_yamauchi_takei == true)
+            {
+              // specify Brent algorithm parameters
+              const double a=273;
+              const double b=3273;
+              typedef std::pair<double, double> Result;
```
gassmoeller: I do not think you need this typedef. See my comment below.

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+    AnelasticVs2T<dim>::initialize ()
+    {
+      Utilities::AsciiDataInitial<dim>::initialize(1);
+    }
+
+     // set up Vs function that Brent minimization operates on
+    template <int dim>
+    double
+    AnelasticVs2T<dim>::
+    fVs (double x,
```
gassmoeller:
```suggestion
    fVs (const double x,
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+         const double      depth,
+         const double      absolute_Vs,
+         const double      mu0,
+         const double      dmudT,
+         const double      dmudP,
+         const double      viscosity_prefactor,
+         const double      activation_energy,
+         const double      activation_volume,
+         const double      solidus_gradient,
+         const bool      use_original_model) const
```
gassmoeller: please align with the other parameters

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+    {}
+
+    template <int dim>
+    void
+    AnelasticVs2T<dim>::initialize ()
+    {
+      Utilities::AsciiDataInitial<dim>::initialize(1);
+    }
+
+     // set up Vs function that Brent minimization operates on
```
gassmoeller: you have this comment already in the header file, no need to repeat here. Same for the other functions below.

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+              // create fVs function to use in Brent minimization and calculate temperature
+              auto bfunc = [ &,this] (double x)
+              {
+                return fVs(x, depth, absolute_Vs, mu0, dmudT,dmudP,
+                           viscosity_prefactor,activation_energy,activation_volume,solidus_gradient,use_original_model);
+              };
+              // determine maximum Vs
+              double maximum_Vs;
+              int fail;
+              maximum_Vs=yamauchi_takei_Vs(273.,depth,mu0,dmudT,dmudP,viscosity_prefactor,activation_energy,activation_volume
```
gassmoeller:
```suggestion
              const double maximum_Vs=yamauchi_takei_Vs(273.,depth,mu0,dmudT,dmudP,viscosity_prefactor,activation_energy,activation_volume
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+                                           ,solidus_gradient,use_original_model);
+
+              // set number of fails to zero
+              fail=0;
+              // where absolute Vs exceeds maximum Vs, set temperature to 273 K
+              if (absolute_Vs>maximum_Vs)
```
gassmoeller: Nice error check! We have a cleaner way for doing this though:
```
AssertThrow(absolute_Vs < maximum_Vs, 
                      ExcMessage("Computed Vs of " + std::to_string(absolute_Vs) + " too fast for sensible temperature"));
```
This will not give you the information about the number of failing points, but how much does that help you anyway? It will only report back the number of failing points in each cell separately. Do you need this information?

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+                {
+                  temperature=273.;
+                  fail=fail+1;
+                  std::cout << "Vs too fast for sensible temperature for " << fail << " points!" << std::endl;
+                }
+              else
+                {
+                  Result r1 = boost::math::tools::brent_find_minima(bfunc,a,b,16);
```
gassmoeller: You can save the typedef by doing:
```suggestion
                  temperature = boost::math::tools::brent_find_minima(bfunc,a,b,16).first;
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+        }
+      // return the absolute temperature in Kelvin
+      return temperature;
+    }
+
+    // set up function to calculate shear wave velocity from Yamauchi & Takei 2016 parameterisation
+    template <int dim>
+    double
+    AnelasticVs2T<dim>::
+    yamauchi_takei_Vs (double temperature,
```
gassmoeller:
```suggestion
    yamauchi_takei_Vs (const double temperature,
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+                       double depth,
```
gassmoeller:
```suggestion
                       const double depth,
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+      if (use_original_model == true)
+        {
+          // calculate density using original parameters
+          density=original_density*(1-(original_thermal_expansivity*((temperature-273)-600))+((pressure*1e-9)/original_bulk_modulus));
+        }
+      else
+        {
+          // create fdV function to use in Brent minimization and calculate isothermal_volume_change and density using
+          // expressions in Grose & Afonso 2013
+          typedef std::pair<double, double> Result2;
```
gassmoeller: use the same approach as above to avoid the typedef

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+      const double gruneisen_parameter=6;
+      const double reference_density=3330;
+      // specify original density parameters
+      const double original_density=3291;
+      const double original_thermal_expansivity=3.59e-5;
+      const double original_bulk_modulus=115.2;
+      // initialize solidus
+      const double T_solidus = 1326.0 + 273 + (((depth-50000)/1e3)*solidus_gradient);
+      // initialize homologous_temperature
+      double homologous_temperature = temperature/T_solidus;
```
gassmoeller: make `const`

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+      // initialize pressures
+      double pressure = depth/pressure_gradient;
```
gassmoeller: same here

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+      // declare other parameters
+      double viscosity,viscosity_reduction_factor,peak_amplitude,peak_width,isothermal_volume_change;
```
gassmoeller: I guess it makes sense to define all of the parameters with constant values above, but please move all of these uninitialized parameters, and the ones in the following lines, in the smallest scope they can be in. I.e. declare them as close to their usage places as possible, if possible as `const`. This makes reading the code easier, and avoids scrolling to the top of the function to look up available names.

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+      double compressibility,pressure_dependent_density,integrated_thermal_expansivity,density,anharmonic_Vs;
+      double unrelaxed_compliance,loss_compliance,storage_compliance,attenuation,period,anelastic_Vs;
+      // begin calculation of Vs
+      if (homologous_temperature<critical_homologous_temperature)
+        {
+          viscosity_reduction_factor=1;
```
gassmoeller:
```suggestion
          viscosity_reduction_factor=1.0;
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+        }
+      else if ((homologous_temperature>=critical_homologous_temperature) && (homologous_temperature<1))
+        {
+          viscosity_reduction_factor=std::exp((-1*((homologous_temperature-critical_homologous_temperature)/(homologous_temperature-
```
gassmoeller: Here and in all other place, clearly separate integers from doubles. Not super important (as C++ automatically promotes integers), but important if you work in more than one language (makes your life easier), and clarifies the types of the variables in the equation (no need to worry if this might be an integer division).
```suggestion
          viscosity_reduction_factor=std::exp((-1.0*((homologous_temperature-critical_homologous_temperature)/(homologous_temperature-
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+        {
+          // set shear wave period as constant
+          period=100;
+        }
+      else
+        {
+          // calculate shear wave period incorporating depth dependence of Forsyth 1992
+          period=(3*depth)/4200;
+        }
+      double normalised_period=period/(2*M_PI*maxwell_relaxation_time);
```
gassmoeller:
```suggestion
      const double normalised_period=period/(2*M_PI*maxwell_relaxation_time);
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+    template <int dim>
+    void
+    AnelasticVs2T<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection ("Initial temperature model");
+      {
+        prm.declare_entry ("Remove temperature heterogeneity down to specified depth",
+                           boost::lexical_cast<std::string>(-std::numeric_limits<double>::max()),
+                           Patterns::Double (),
+                           "This will set the heterogeneity prescribed by S20RTS or S40RTS to zero "
```
gassmoeller: S20RTS / S40RTS is not longer accurate, is it?

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+                           "down to the specified depth (in meters). Note that your resolution has "
+                           "to be adequate to capture this cutoff. For example if you specify a depth "
+                           "of 660km, but your closest spherical depth layers are only at 500km and "
+                           "750km (due to a coarse resolution) it will only zero out heterogeneities "
+                           "down to 500km. Similar caution has to be taken when using adaptive meshing.");
+        prm.declare_entry ("Use Yamauchi and Takei parameterization", "true",
+                           Patterns::Bool(),
+                           "This parameter determines whether to use the anelasticity model of "
```
gassmoeller: Is there an alternative implemented? If no, you might just remove the parameter. If there is, it makes more sense to have this as a string argument that gets parsed to the correct choice. With a `bool` you limit your options to 2 cases, which might become a problem in the future.

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+// explicit instantiations
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    ASPECT_REGISTER_INITIAL_TEMPERATURE_MODEL(AnelasticVs2T,
+                                              "anelastic Vs to temperature",
+                                              "Implementation of a model in which the initial temperature is calculated "
+                                              "from files containing absolute shear wave velocity (Vs) data in ascii format. "
+                                              "This plug-in allows you to select from a number of different models that"
```
gassmoeller: If there are several options, mention them here shortly (e.g. their publications), if there is currently just one, them just say it currently only implements one conversion.

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+                                              "convert Vs into temperature, accounting for the anelastic behaviour of mantle material."
+                                              "Note the required format of the "
+                                              "input data: The first lines may contain any number of comments "
+                                              "if they begin with `#', but one of these lines needs to "
+                                              "contain the number of grid points in each dimension as "
+                                              "for example `# POINTS: 3 3'. "
+                                              "The order of the data columns "
+                                              "has to be `x', `y', `Temperature [K]' in a 2d model and "
```
gassmoeller:
```suggestion
                                              "has to be `x', `y', `Velocity [km/s]' in a 2d model and "
```

source/initial_temperature/anelasticity_temperature.cc
```diff
@@ -0,0 +1,421 @@
...
+                                              " `x', `y', `z', `Temperature [K]' in a 3d model, which means that "
```
gassmoeller: same here

source/initial_temperature/anelasticity_temperature.cc
freddrichards: Have removed for now for but in the long term I plan to add more anelasticity models e.g. Faul & Jackson 2011 and Goes et al. 2012. Will change it to "int" and add different cases when I eventually get round to this.

source/initial_temperature/anelasticity_temperature.cc
freddrichards: Nope, changed!

include/aspect/initial_temperature/anelasticity_temperature.h
```diff
@@ -0,0 +1,145 @@
+/*
```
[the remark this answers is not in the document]
freddrichards: Agree and now changed

include/aspect/initial_temperature/anelasticity_temperature.h
```diff
@@ -0,0 +1,145 @@
...
+
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    using namespace dealii;
+
+    /**
+     * A class that describes an initial temperature field for a 2D or 3D shear wave velocity (Vs) model.
+     * Vs values are converted to temperature using the anelasticity parameterization of Yamauchi & Takei (2016).
```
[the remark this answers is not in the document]
freddrichards: done on both counts

include/aspect/initial_temperature/anelasticity_temperature.h
freddrichards: I have now changed the function declaration so the input variables are no longer member variables. Was this what you had in mind?

source/initial_temperature/anelasticity_temperature.cc
freddrichards: You're right, have changed accordingly.

source/initial_temperature/anelasticity_temperature.cc
freddrichards: The number of failed points isn't that relevant, most useful is the warning that some of the Vs values you're reading in are too high to give sensible temperatures. I have now implemented your suggested change as I agree it's much cleaner and does what's needed.

source/initial_temperature/anelasticity_temperature.cc
freddrichards: I've now clarified that this plugin only implements the Yamauchi & Takei 2016 version.

freddrichards:
Have made requested changes, will write a test in the next few days.

source/initial_temperature/anelasticity_vs_to_temperature.cc
```diff
@@ -94,43 +90,29 @@ namespace aspect
...
-                }
-              else
-                {
-                  Result r1 = boost::math::tools::brent_find_minima(bfunc,a,b,16);
-                  temperature=r1.first;
-                }
+          // where absolute Vs exceeds maximum Vs, set temperature to 273 K
+          if (absolute_Vs>maximum_Vs)
+            {
+              AssertThrow(absolute_Vs > maximum_Vs,
```
gassmoeller: Hmm I am still not quite sure what you want to do here. If the absolute Vs is bigger than the maximum Vs, do you (i) want the model to crash (because the computation makes no sense anymore), or (ii) use 273 because that is a reasonable temperature for this case? If (i) then change then remove the `if-else` altogether and change the  AssertThrow to the following `AssertThrow(absolute_Vs < maximum_Vs, ExcMessage...)`. If (ii) then remove the AssertThrow. We currently have no nice way of writing warnings so in that case the plugin should simply continue.

source/initial_temperature/anelasticity_vs_to_temperature.cc
```diff
@@ -271,18 +257,18 @@ namespace aspect
         {
           // create fdV function to use in Brent minimization and calculate isothermal_volume_change and density using
           // expressions in Grose & Afonso 2013
-          typedef std::pair<double, double> Result2;
-          auto vfunc = [ &,this] (double x)
+//          typedef std::pair<double, double> Result2;
```
gassmoeller: Remove this line

source/initial_temperature/anelasticity_vs_to_temperature.cc
freddrichards: The second case is preferable. Occasionally you get a stray Vs value that is anomalously fast and its better that this doesn't crash the model.

gassmoeller:
Hey, this looks good now, but the tester complains about unused variables in your code (you can see the messages if you click on 'Details' next to the red cross below the PR). Could you take a look and fix the warnings?

PR #3101, opened 2019-05-30
[WIP] Plugin to use Frank-Kamenetskii rheology

GEuen:
Plugin that modifies the viscosity law in the Simple material model. Also adds a few new parameters, specifically for allowing a different reference temperature for the viscosity law.


gassmoeller:
This looks generally in good shape, but usually we do not merge single plugins without their according parameter files, because they are not very useful for others. What do you think about keeping this open for a while until you want to add the corresponding benchmarks?

GEuen:
That would be fine with me. I have a simplified benchmark I could add in the meantime.

gassmoeller:
#3698 merged a Frank-Kamenetskii rheology, so I think we can close this PR. Feel free to reopen if necessary.

PR #3102, opened 2019-05-30
WIP: change nonlinear Picard solver to defect correction Picard.

MFraters:
only changed the one solver at the moment.

I currently just expended the `solve_iterated_advection_and_newton_stokes` function so that it can be forced to only do Picard iterations. I am thinking of just keep expanding this function, because otherwise there will be a lot of code duplication. @tjhei @bangerth @jdannberg @gassmoeller what are your opinions on this approach?


MFraters:
This pull request will be split up, so I will close this one.

PR #3105, opened 2019-05-30
Rearrange savani_perturbation to be the same as s40rts_perturbation

SophieCoulson:
Here I've simply rearranged the savani_perturbation plug-in so there is a function to return Vs, and this can be called from another plug in. It is now in exactly the same format as s40rts_perturbation.

### DIFF
--- a/source/initial_temperature/SAVANI_perturbation.cc
+++ b/source/initial_temperature/SAVANI_perturbation.cc
@@ -191,7 +191,7 @@ namespace aspect
 
     template <>
     double
-	SAVANIPerturbation<2>::
+    SAVANIPerturbation<2>::
     get_Vs (const Point<2> &/*position*/) const
     {
       Assert (false, ExcNotImplemented());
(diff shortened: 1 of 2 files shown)


source/initial_temperature/SAVANI_perturbation.cc
```diff
@@ -188,101 +189,120 @@ namespace aspect
       return 0;
     }
 
+    template <>
+    double
+	SAVANIPerturbation<2>::
+    get_Vs (const Point<2> &/*position*/) const
```
bangerth: You'll have to run the indent script.

SophieCoulson:
Done

source/initial_temperature/SAVANI_perturbation.cc
```diff
@@ -176,6 +176,7 @@ namespace aspect
 
     }
 
+
```
gassmoeller: while you are at it, add another empty line. We usually have 1 empty line between functions in header files, and 3 empty lines between functions in .cc files.

gassmoeller: Thinking about this again, it is really not necessary to hold up this PR for an empty line that was missing before already. Ready to merge.

PR #3106, opened 2019-05-30
fix visco plastic get timestep number problem.

MFraters:
found the problem in and needed this for #3102.

### DIFF
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -661,7 +661,7 @@ namespace aspect
           // If plastic strain is tracked (so not the total strain), only overwrite
           // when plastically yielding.
           // If viscous strain is also tracked, overwrite the second reaction term as well.
-          if  (this->get_timestep_number() > 0 && in.strain_rate.size())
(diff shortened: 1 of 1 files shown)


source/material_model/visco_plastic.cc
```diff
@@ -661,7 +661,7 @@ namespace aspect
           // If plastic strain is tracked (so not the total strain), only overwrite
           // when plastically yielding.
           // If viscous strain is also tracked, overwrite the second reaction term as well.
-          if  (this->get_timestep_number() > 0 && in.strain_rate.size())
+          if  (this->get_timestep_number() != numbers::invalid_unsigned_int && this->get_timestep_number() > 0 && in.strain_rate.size())
```
gassmoeller: is this equivalent to `this->simulator_is_past_initialization() && this->get_timestep_number() > 0 && in.strain_rate.size())`? That seems a bit easier to understand.

MFraters: Yes, I think so. I will change it.

bangerth: Thanks, @gassmoeller, for seeing this. I had noticed it too but forgot to mark it up.

MFraters:
addressed the comments

PR #1593, opened 2017-05-12
Add viscoelastic material model and elastic outputs functionality

naliboff:
The next step is to incorporate the right-hand side terms.

PR #3107, opened 2019-05-30
move elasticity to rheology model

jdannberg:
This is a suggestion for moving the elasticity to a new rheology class. Although this looks like a lot of lines, most of these lines are just moved from viscoelastic to the new file. It would be great if I could get some feedback on the functions, if they make sense like this, or if they should be different/have different arguments. 

In particular, the `fill_elastic_force_outputs` and `fill_reaction_outputs` right now simply get the material model outputs and use some of these values as inputs (the viscosity), but fill other properties (force outputs, reaction outputs).

### DIFF
--- a/include/aspect/material_model/rheology/elasticity.h
+++ b/include/aspect/material_model/rheology/elasticity.h
@@ -111,7 +111,7 @@ namespace aspect
           get_elastic_shear_moduli () const;
 
           /**
-           * Given the (viscous) viscosity and the shear modulus, compute the viscoelastic
+           * Given the (viscous or visco-plastic) viscosity and the shear modulus, compute the viscoelastic
            * viscosity.
            */
           double
--- a/include/aspect/material_model/rheology/elasticity.h
+++ b/include/aspect/material_model/rheology/elasticity.h
@@ -1,0 +1,159 @@
+/*
+  Copyright (C) 2019 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _aspect_material_model_rheology_elasticity_h
+#define _aspect_material_model_rheology_elasticity_h
+
+#include <aspect/global.h>
+#include <aspect/material_model/interface.h>
+#include <aspect/simulator_access.h>
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    namespace Rheology
+    {
+      /**
+       * Additional output fields for the elastic shear modulus to be added to
+       * the MaterialModel::MaterialModelOutputs structure and filled in the
+       * MaterialModel::Interface::evaluate() function.
+       */
+      template <int dim>
+      class ElasticAdditionalOutputs : public NamedAdditionalMaterialOutputs<dim>
+      {
+        public:
+          ElasticAdditionalOutputs(const unsigned int n_points);
+
+          virtual std::vector<double> get_nth_output(const unsigned int idx) const;
+
+          /**
+           * Elastic shear moduli at the evaluation points passed to
+           * the instance of MaterialModel::Interface::evaluate() that fills
+           * the current object.
+           */
+          std::vector<double> elastic_shear_moduli;
+      };
+
+
+
+      template <int dim>
+      class Elasticity : public ::aspect::SimulatorAccess<dim>
+      {
+        public:
+          /**
+           * Declare the parameters this function takes through input files.
+           */
+          static
+          void
+          declare_parameters (ParameterHandler &prm);
+
+          /**
+           * Read the parameters from the parameter file.
+           */
+          void
+          parse_parameters (ParameterHandler &prm);
+
+          /**
+           * Create the additional material model outputs object that contains the
+           * elastic shear moduli.
+           */
+          void
+          create_elastic_outputs (MaterialModel::MaterialModelOutputs<dim> &out) const;
+
+          /**
+           * Given the stress of the previous time step in the material model inputs @p in,
+           * the elastic shear moduli @p average_elastic_shear_moduli a each point,
+           * and the (viscous) viscosities given in the material model outputs object @p out,
+           * fill an additional material model outputs objects with the elastic force terms.
+           */
+          void
+          fill_elastic_force_outputs (const MaterialModel::MaterialModelInputs<dim> &in,
+                                      const std::vector<double> average_elastic_shear_moduli,
+                                      MaterialModel::MaterialModelOutputs<dim> &out) const;
+
+          /**
+           * Given the stress of the previous time step in the material model inputs @p in,
+           * the elastic shear moduli @p average_elastic_shear_moduli a each point,
+           * and the (viscous) viscosities given in the material model outputs object @p out,
+           * compute an update to the elastic stresses and use it to fill the reaction terms
+           * material model output property.
+           */
+          void
+          fill_reaction_outputs (const MaterialModel::MaterialModelInputs<dim> &in,
+                                 const std::vector<double> average_elastic_shear_moduli,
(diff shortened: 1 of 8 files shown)


include/aspect/material_model/rheology/elasticity.h
```diff
@@ -0,0 +1,159 @@
...
+
+          /**
+           * Return the values of the elastic shear moduli for each composition used in the
+           * rheology model.
+           */
+          std::vector<double>
+          get_elastic_shear_moduli () const;
+
+          /**
+           * Given the (viscous) viscosity and the shear modulus, compute the viscoelastic
```
gassmoeller:
```suggestion
           * Given the (viscous or visco-plastic) viscosity and the shear modulus, compute the viscoelastic
```

jdannberg:
I added the viscoelasto-plastic model, so it would be great to get a more thorough review, @gassmoeller. 

Also, @naliboff, these material model used different averaging schemes! Please tell me which one is the correct one (averaging elastic moduli, or averaging viscoelastic viscosities) and then I will implement this once and for all and remove the other option.

naliboff:
@jdannberg - The overall structure looks really good! More comments later today.

Regarding the averaging, after the discussion with @bobmyhill I think the best approach is to average the elastic shear moduli before calculating the viscoelastic viscosities (e.g, method in viscoelastic material model).  Would you like me to do a PR for this fix or implement it here?

jdannberg:
@naliboff:

> I think the best approach is to average the elastic shear moduli before calculating the viscoelastic viscosities

That's what I do in this pull request. So there's no need to change anything :smiley:

naliboff:
Then lets stick with the averaging method you have implemented 👍 Unlikely anyone has started using the VEP model for a production run since it was merged, so no reason not to change it now.

source/material_model/rheology/elasticity.cc
```diff
@@ -0,0 +1,381 @@
...
+        // The elastic time step (dte) is equal to the numerical time step if the time step number
+        // is greater than 0 and the parameter 'use_fixed_elastic_time_step' is set to false.
+        // On the first (0) time step the elastic time step is always equal to the value
+        // specified in 'fixed_elastic_time_step', which is also used in all subsequent time
+        // steps if 'use_fixed_elastic_time_step' is set to true.
+        //
+        // We also use this parameter when we are still *before* the first time step,
+        // i.e., if the time step number is numbers::invalid_unsigned_int.
+        const double dte = ( ( this->get_timestep_number() > 0 &&
+                               this->get_timestep_number() != numbers::invalid_unsigned_int &&
```
gassmoeller: This should be equivalent to 

```suggestion
                               this->simulator_is_past_initialization() &&
```

include/aspect/material_model/rheology/elasticity.h
```diff
@@ -0,0 +1,159 @@
...
+
+          /**
+           * Given the stress of the previous time step in the material model inputs @p in,
+           * the elastic shear moduli @p average_elastic_shear_moduli a each point,
+           * and the (viscous) viscosities given in the material model outputs object @p out,
+           * fill an additional material model outputs objects with the elastic force terms.
+           */
+          void
+          fill_elastic_force_outputs (const MaterialModel::MaterialModelInputs<dim> &in,
+                                      const std::vector<double> average_elastic_shear_moduli,
```
tjhei: add reference

include/aspect/material_model/rheology/elasticity.h
```diff
@@ -0,0 +1,159 @@
...
+
+    namespace Rheology
+    {
+      /**
+       * Additional output fields for the elastic shear modulus to be added to
+       * the MaterialModel::MaterialModelOutputs structure and filled in the
+       * MaterialModel::Interface::evaluate() function.
+       */
+      template <int dim>
+      class ElasticAdditionalOutputs : public NamedAdditionalMaterialOutputs<dim>
```
tjhei: Is putting this into the Rheology namespace intentional?

include/aspect/material_model/rheology/elasticity.h
```diff
@@ -0,0 +1,159 @@
...
+      {
+        public:
+          ElasticAdditionalOutputs(const unsigned int n_points);
```
tjhei:
```suggestion
          explicit ElasticAdditionalOutputs(const unsigned int n_points);
```

include/aspect/material_model/rheology/elasticity.h
```diff
@@ -0,0 +1,159 @@
...
+          /**
+           * Given the stress of the previous time step in the material model inputs @p in,
+           * the elastic shear moduli @p average_elastic_shear_moduli a each point,
+           * and the (viscous) viscosities given in the material model outputs object @p out,
+           * compute an update to the elastic stresses and use it to fill the reaction terms
+           * material model output property.
+           */
+          void
+          fill_reaction_outputs (const MaterialModel::MaterialModelInputs<dim> &in,
+                                 const std::vector<double> average_elastic_shear_moduli,
```
tjhei: same here

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,16 @@ namespace aspect
 
 
 
+      /**
+       * Used for calculating average elastic shear modulus and viscosity
```
tjhei: can you explain what it does?

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,16 @@ namespace aspect
...
+       */
+      template <int dim>
+      double calculate_average_vector (const std::vector<double> &composition,
```
tjhei: it doesn't calculate an "average vector" but it averages a vector into a double, right? Can you think of a better name? calculate_average_of_vector? average_vector?
And sorry, I realize you are just moving things around.

source/material_model/rheology/elasticity.cc
```diff
@@ -0,0 +1,381 @@
...
+        NamedAdditionalMaterialOutputs<dim>(make_elastic_additional_outputs_names()),
+        elastic_shear_moduli(n_points, numbers::signaling_nan<double>())
+      {}
+
+      template <int dim>
+      std::vector<double>
+      ElasticAdditionalOutputs<dim>::get_nth_output(const unsigned int idx) const
+      {
+        AssertIndexRange (idx, 1);
+        switch (idx)
```
tjhei: you can delete the whole switch.

source/material_model/rheology/elasticity.cc
```diff
@@ -0,0 +1,381 @@
...
+                                   "compositional field is called stress_xy."));
+            AssertThrow(this->introspection().compositional_index_for_name("stress_xz") == 4,
+                        ExcMessage("Rheology model Elasticity only works if the fifth "
+                                   "compositional field is called stress_xz."));
+            AssertThrow(this->introspection().compositional_index_for_name("stress_yz") == 5,
+                        ExcMessage("Rheology model Elasticity only works if the sixth "
+                                   "compositional field is called stress_yz."));
+          }
+        else
+          ExcNotImplemented();
```
tjhei: that doesn't do what you want without the assert. ;-)

include/aspect/material_model/rheology/elasticity.h
jdannberg: Well I thought elastic outputs are something that's really part of the rheology model, so I just put it into that namespace. What would be a reason for moving it out of the namespace?

tjhei: The reason I mention this is that you are moving this class from the MaterialModel namespace. I am not sure if others are using this class, but you would break those plugins.

include/aspect/material_model/utilities.h
jdannberg: I was just moving this function, but after checking what it actually does, I found it was just duplicating code, so I removed it. That also solves your comment about the name below.

jdannberg:
I think I addressed all comments (and I also rebased, as there were conflicts with master). 
@tjhei and @gassmoeller: Thank you for your reviews. Next time I will read the code that I am moving more carefully! You caught lots of things that didn't make any sense.

source/material_model/rheology/elasticity.cc
jdannberg: Wow, I didn't even see that...

tjhei:
> Next time I will read the code that I am moving more carefully!

sorry for nitpicking code that you just wanted to move...

jdannberg:
> sorry for nitpicking code that you just wanted to move...

I know I usually complain about this, but this time I feel it was totally justified. The whole pull request was about cleaning up the material models, and this actually saved us 50 lines of code and made everything much more readable.

tjhei:
Looks good, but I still think we should keep the AdditionalOutputs in the same namespace.

include/aspect/material_model/rheology/elasticity.h
jdannberg: I checked, an the functionality was added more than a year ago (#1593). I'll move this out of the namespace.

jdannberg:
Yes, I agree. Done.

PR #3108, opened 2019-05-30
manual: document SUPG

tjhei:


tjhei:
@tcclevenger can you take a quick look please?

tjhei:
image

tcclevenger:
@tjhei It looks good to me!

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
 \end{itemize}
 
 
+Both add additional terms to the temperature (or compositional field equation). We will discuss
```
jdannberg:
```suggestion
Both add additional terms to the temperature (or compositional field) equation. We will discuss
```

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
...
+ \rho C_p \frac{\partial T}{\partial t} + \rho C_p \mathbf{u} \cdot \nabla T - \nabla \cdot k\nabla T = F,
+\]
+where $F$ is the combination of source and reaction terms,
+while the weak form (with test function $\varphi$) becomes:
+\[
+a(T,\varphi) =
+ \left(\rho C_p \frac{\partial T}{\partial t}, \varphi \right)
+ + \left(\rho C_p \mathbf{u} \cdot \nabla T, \varphi \right)
+ + \left( k \nabla T, \nabla \varphi \right) = (F,\varphi) = f(\varphi).
+\]
```
jdannberg: There is only one other place in the manual where we write about the weak form (in the inner core cookbook), and there we write out the integrals. So if you want to use this shorter form here, it would make sense to define what you mean by that.

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
...
+
+\subsubsection{SUPG Stabilization}
+
+For streamline upwind/Petrov-Galerkin (SUPG) (see for example \cite{JohnKnobloch2006,step-63}), we amend the weak form with a cell-wise
```
jdannberg: I your screenshot, this reference does not work?

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
...
+defined weak form
+\[
+a_{\text{SUPG}} (T, \varphi) =
+ \sum_{K \in \mathcal{T}_h}
+  \delta_K \left( \rho C_p \frac{\partial T}{\partial t} - k \triangle T + \mathbf{\beta} \cdot \nabla T - F, \mathbf{\beta} \cdot \nabla \varphi \right)_K,
+\]
+where $K \in \mathcal{T}_h$ are the cells in the computation, $\delta_K \geq 0$ is a stabilization coefficient
+defined on each cell, $\mathbf{\beta} = \rho C_p \mathbf{u}$ is the effective advection velocity.
```
jdannberg: Is there a reason you replace this by beta?

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
...
+\]
+
+\subsubsection{SUPG Stabilization}
+
+For streamline upwind/Petrov-Galerkin (SUPG) (see for example \cite{JohnKnobloch2006,step-63}), we amend the weak form with a cell-wise
+defined weak form
+\[
+a_{\text{SUPG}} (T, \varphi) =
+ \sum_{K \in \mathcal{T}_h}
+  \delta_K \left( \rho C_p \frac{\partial T}{\partial t} - k \triangle T + \mathbf{\beta} \cdot \nabla T - F, \mathbf{\beta} \cdot \nabla \varphi \right)_K,
```
jdannberg: Why is this k \triangle T instead of /nabla k /nabla T?

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
...
+\]
+where $K \in \mathcal{T}_h$ are the cells in the computation, $\delta_K \geq 0$ is a stabilization coefficient
+defined on each cell, $\mathbf{\beta} = \rho C_p \mathbf{u}$ is the effective advection velocity.
+The first argument in the inner product is the strong form of the residual of PDE, which
+is tested with the expression $\mathbf{\beta} \cdot \nabla \varphi$.
```
jdannberg: I don't really understand what the second half of the sentence is supposed to mean.

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
...
+
+We use the parameter design presented in \cite{JohnKnobloch2006} for $\delta_K$:
+\[
+ \delta_K = \frac{h}{2d\|\mathbf{\beta}\|} \left( \coth(Pe)-\frac{1}{Pe} \right)
+\]
+where the Peclet number is given by
+\[
+ Pe = \frac{ \| \mathbf{\beta} \| h}{2 k d},
+\]
+$d$ is the polynomial degree (typically 2) and
```
jdannberg: ...polynomial degree of the temperature (or composition) element

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
...
+$
+ \coth(x) = (1+\exp(-2x)) / (1-\exp(-2x)).
+$
+If $Pe<1$, the equation is diffusion-dominated and no stabilization is needed, so we
+set $\delta_K=0$. Care needs to be taken in the definition if $\| \beta \|$ or $k$ become zero.
```
jdannberg: So what do you do in this case?

doc/manual/manual.tex
jdannberg: What does that mean, you "amend" the weak form? This is just added on top of the original left- and right-hand side of the equation?

doc/manual/manual.tex
```diff
@@ -2286,6 +2286,58 @@ \subsection{Advection Stabilization}
...
+
+
+
+
+
+
+
```
jdannberg: Do you need all those empty lines?

doc/manual/manual.tex
tjhei: This is an interesting point! It is the strong form of the residual and there is no way to compute the derivatives of k, so it is assumed to be constant.

doc/manual/manual.tex
tjhei: Because I don't want to write this thing that many times and it helps to know that this is, effectively, the velocity we are advecting with (the normal SUPG literature doesn't have rho C_p, just a velocity. I made the mistake the first time not to include the constants and that makes the method terrible. I am adding a sentence.

PR #3118, opened 2019-05-31
add deviatoric stress postprocessing

jvanhunen:
This is a small addition to postprocess/visualisation: 
the stress.cc (and stress.h) postprocessor was outputting all components of the (full) stress tensor. 
I created a slightly modified set of files (devstress.cc and devstress.h) that output the deviatoric stress: for each component, the full stress = viscosity*strainrate + pressure, and the deviatoric stress is simply devstress = viscosity*strainrate (i.e. without the pressure term). 

This is useful to (e.g.) visualize the stresses that drive shortening or extension in the lithosphere.

### DIFF
--- a/include/aspect/postprocess/visualization/devstress.h
+++ b/include/aspect/postprocess/visualization/devstress.h
@@ -1,0 +1,93 @@
+/*
+  Copyright (C) 2015 - 2017 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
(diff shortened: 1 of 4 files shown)


source/postprocess/visualization/devstress.cc
```diff
@@ -0,0 +1,146 @@
...
+
+// explicit instantiations
+namespace aspect
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      ASPECT_REGISTER_VISUALIZATION_POSTPROCESSOR(DevStress,
+                                                  "devstress",
```
tjhei: we typically spell out names, so please use "deviatoric stress"

jvanhunen: I changed the following files names: 
source/postprocess/visualization/devstress.cc --> source/postprocess/visualization/deviatoric_stress.cc and 
include/aspect/postprocess/visualization/devstress.h --> include/aspect/postprocess/visualization/deviatoric_stress.h

Furthermore: 
- Class name: DevStress to DeviatoricStress
- Postprocessing name in parameter file is now "deviatoric_stress"
- visualisation option names changed from "devstress_xx" to deviatoric_stress_xx", and similar for all other components of stress tensor.

include/aspect/postprocess/visualization/devstress.h
```diff
@@ -0,0 +1,93 @@
+/*
```
gassmoeller: I think you included the files twice, once as `devstress.h` and once as `deviatoric_stress.h`, same for the .cc files.

jvanhunen: you are right: shear stress does exactly the same. So no need anymore for this postprocessor

PR #3119, opened 2019-05-31
prescribed compression [WIP]

tjhei:


tjhei:
@bangerth can you please take a look if you can find any mistakes in this new assembly? You can ignore the changes in benchmarks/ (this is using it in the burstedde material model as an example).

benchmarks/burstedde/burstedde.cc
```diff
@@ -214,6 +233,12 @@ namespace aspect
               for (unsigned int c=0; c<in.composition[i].size(); ++c)
                 out.reaction_terms[i][c] = 0.0;
 
+              if (prescribed_compression != nullptr)
+                {
+                  const double pi=numbers::PI;
+                  prescribed_compression->prescribed_compression[i]=3;
```
bangerth: I know it's just a test, but you don't actually *use* `pi` :-)

include/aspect/parameters.h
```diff
@@ -495,6 +495,7 @@ namespace aspect
      */
     bool                           include_melt_transport;
     bool                           enable_additional_stokes_rhs;
+    bool                           enable_prescribed_compression;
```
bangerth: why have a flag? can't you just test whether the material model attached the appropriate additional output?

source/simulator/assemblers/stokes.cc
```diff
@@ -269,6 +269,12 @@ namespace aspect
       const MaterialModel::ElasticOutputs<dim>
       *elastic_outputs = scratch.material_model_outputs.template get_additional_output<MaterialModel::ElasticOutputs<dim> >();
 
+      const MaterialModel::PrescribedCompressionOutputs<dim>
+      *prescribed_compression =
+        (this->get_parameters().enable_prescribed_compression)
+        ? scratch.material_model_outputs.template get_additional_output<MaterialModel::PrescribedCompressionOutputs<dim> >()
+        : nullptr;
```
bangerth: Just don't use a parameter and instead test whether `get_additional_output()` returns anything that's non-null to determine whether or not to assemble anything.

source/simulator/assemblers/stokes.cc
```diff
@@ -369,6 +379,19 @@ namespace aspect
                 data.local_rhs(i) += (scalar_product(elastic_outputs->elastic_force[q],Tensor<2,dim>(scratch.grads_phi_u[i])))
                                      * JxW;
 
+              if (prescribed_compression != nullptr)
+                data.local_rhs(i) += (
+                                       // RHS of - (div u,p) = - (R,p)
+                                       - pressure_scaling
+                                       * prescribed_compression->prescribed_compression[q]
```
bangerth: had to do a double-take here :-)

bangerth (commented):
Nothing stands out as wrong. The code could be simplified, but that's not the question you asked me.

tjhei:
@naliboff I added the logic to automatically switch between the explicit and implicit version (you can toggle this from ``is_compressible`` in the material model).

Can you let me know if this PR works for you as is or if I need to push the changes to the 2nd invariant of the strain tensor?

source/simulator/assemblers/stokes.cc
tjhei: This makes the logic below (``const bool update_strain_rate = ...``) complicated as I would need to create a fake material model.

tjhei:
explicit:
image

disabled:
image

convergence:
image

naliboff:
@tjhei - Thanks for doing all these updates! Are the results you posted above with the normal method we use for calculating the second invariant? I'm about to run the tests locally to make sure I get the same results.

tjhei:
> Are the results you posted above with the normal method we use for calculating the second invariant?

No, this was run with the changes to the computation of the second invariant. Do you want me to push those? I don't know if they are needed or not.

naliboff:
New results using the explicit formulation and the second invariant (standard deal.II definition) of the deviatoric strain rate (rather than full strain rate).  

Not as clean as the results above, but we are still getting roughly the right angle at low resolutions. Still, probably warrants a full comparison between the two definitions of the strain rate invariant to start.

explicity_second_invariant_deviatoric_strain_rate

naliboff:
@tjhei, @cedrict  - I did a few tests with different assumptions regarding the strain rate invariant and it appears we do need to use the full strain rate (e.g., the method we used at the Hackathon). 

Oddly, using the deal.II `second_invariant`  function causes all sorts of convergence problems when using the full strain rate in place of the deviatoric strain rate. Perhaps we need to take an absolute value somewhere?

The figures below show the extension results for global refinement 3 or 5 and four difference scenarios for calculating the strain rate invariant:

1. top panel: deal.II second_invariant and deviatoric strain rate 
2. second panel: second_moment_invariant and deviatoric strain rate
3. third panel: deal.II second invariant and full strain rate
4. bottom panel: second_moment_invariant and full strain rate

The bottom panel is the same method we used at the Hackathon and appears to give a slightly better fit to the correct angle (60 degrees, white line) for the high resolution case. Fortunately, we are still obtaining roughly the right angle at coarse resolutions.

Thoughts on how to proceed? 

My vote would be to update the PR with the different options for calculating the strain rate invariant (I'll do this) and then re-run these results with the Spiegelman 2016 brick experiment (e.g., setup that was used in the Newton paper). I can write some scripts to automatically measure the shear band angle and width from these experiments.

Global refinement 3 (16 vertical elements):
global_refinement_3

Global refinement 5 (64 vertical elements):
global_refinement_5

tjhei:
@naliboff can we close this or do you want to grab anything else from here?

bangerth:
Is this PR superseded by the recent development of dilation plugins?

PR #3120, opened 2019-05-31
Correction for initial guess/tol in GMRES solve

tcclevenger:
This is a correction for https://github.com/geodynamics/aspect/pull/2987. The previous way to get the initial solution and tolerance for the GMRES solve was 1) incorrect for time dependent and 2) ugly. I cleaned it up a bit.

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1202,7 +1202,7 @@ namespace aspect
     // linearized_stokes_variables has a different
     // layout than current_linearization_point, which also contains all the
     // other solution variables.
-    if (!sim.assemble_newton_stokes_system)
+    if (sim.assemble_newton_stokes_system == false)
       {
         linearized_stokes_initial_guess.block (block_vel) = sim.current_linearization_point.block (block_vel);
         linearized_stokes_initial_guess.block (block_p) = sim.current_linearization_point.block (block_p);
@@ -1229,7 +1229,7 @@ namespace aspect
     linearized_stokes_initial_guess.block (block_p) /= sim.pressure_scaling;
 
     double solver_tolerance = 0;
-    if (!sim.assemble_newton_stokes_system)
+    if (sim.assemble_newton_stokes_system == false)
       {
         // (ab)use the distributed solution vector to temporarily put a residual in
         // (we don't care about the residual vector -- all we care about is the
@@ -1464,7 +1464,7 @@ namespace aspect
 
     // do some cleanup now that we have the solution
     sim.remove_nullspace(sim.solution, distributed_stokes_solution);
-    if (!sim.assemble_newton_stokes_system)
+    if (sim.assemble_newton_stokes_system == false)
       sim.last_pressure_normalization_adjustment = sim.normalize_pressure(sim.solution);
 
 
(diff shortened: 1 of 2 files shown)


source/simulator/stokes_matrix_free.cc
```diff
@@ -1176,16 +1176,26 @@ namespace aspect
...
                                  : sim.introspection.block_indices.pressure;
+
+    LinearAlgebra::BlockVector distributed_stokes_solution (sim.introspection.index_sets.stokes_partitioning,
+                                                            sim.mpi_communicator);
+    // extract Stokes parts of rhs vector
+    LinearAlgebra::BlockVector distributed_stokes_rhs(sim.introspection.index_sets.stokes_partitioning,
+                                                      sim.mpi_communicator);
+
+    distributed_stokes_rhs.block(block_vel) = sim.system_rhs.block(block_vel);
+    distributed_stokes_rhs.block(block_p) = sim.system_rhs.block(block_p);
```
bangerth: why do you need to copy these vectors? The `sim.system_rhs` is already a completely distributed vector

source/simulator/stokes_matrix_free.cc
```diff
@@ -1218,9 +1228,8 @@ namespace aspect
     sim.current_constraints.set_zero (linearized_stokes_initial_guess);
     linearized_stokes_initial_guess.block (block_p) /= sim.pressure_scaling;
 
-
     double solver_tolerance = 0;
-    if (sim.assemble_newton_stokes_system == false)
+    if (!sim.assemble_newton_stokes_system)
```
bangerth: nah, let's stick with the system we had -- that's what we do almost everywhere. it's just easier to read

source/simulator/stokes_matrix_free.cc
tcclevenger: It is a Trilinos BlockVector, and we need a `dealii::distributed::BlockVector`. Also, `sim.system_rhs` contains more blocks than just the stokes blocks.

source/simulator/stokes_matrix_free.cc
tcclevenger: Ok, I will change the ones in `simulator/solver.cc`. This one here was changed to be consistent with the others.

tcclevenger:
The issue is in the order of the calls `initialize_dof_vector()` and `collect_sizes()`. An example here is that the order used to be:

`dealii::LinearAlgebra::distributed::BlockVector<double> solution_copy(2);`
 
`solution_copy.collect_sizes();`

`stokes_matrix.initialize_dof_vector(solution_copy);`

and now order is:

`dealii::LinearAlgebra::distributed::BlockVector<double> solution_copy(2);`
 
`stokes_matrix.initialize_dof_vector(solution_copy);`

`solution_copy.collect_sizes();`

This only shows up when you do multiple timesteps (otherwise the initial vectors are zero anyways)

tjhei:
I guess it would be good to have a time dependent test case...

gassmoeller:
Timo's tester is failing because of an old wrong test result on master. Since the CIG tester is happy, I am going to merge.

tcclevenger:
Sorry guys, I've been away since the hackathon. I see you already merged this. If y'all would like, per @tjhei suggestion, I can create a time-dependent test case in a new pull request?

gassmoeller:
No problem, we were all somewhat distracted. A time-dependent GMG test case would still be appreciated.

PR #3124, opened 2019-05-31
[WIP] Add an interface and material model for the CRUST 1.0 data set

naliboff:
This PR add an initial composition model, material model and cookbook (eventually) for the CRUST 1.0 data set, which provides a global description of the density and boundaries of different crustal layers.

@cedrict @LudovicJnnt and @mnkajan contributed significantly at various points.

The code is currently not ready for review and a PR was opened solely for the purpose of discussing the large scale structure and benchmarking against gravity calculations performed by @LudovicJnnt and @cedrict. 

At present, gravity calculations using the provided .prm file show first-order agreement with @LudovicJnnt calculations using a pre-processed CRUST 1.0 data set, but there are definitely second- and third-order variations to work out.

A list of ToDos before the code is ready for review:

- [ ] Make sure CRUST 1.0 is cited in all places it is mentioned by name.
- [ ] Major cleanup for hardwired values, lazy coding, documentation, etc
- [ ] Add in more Asserts to make sure all options are cross compatible
- [ ] Check the accuracy of the interpolation method in the new initial composition plugin
- [ ] Determine if there is a faster, more elegant and/or accurate method for loading data 
- [ ] Add an option to prescribed a compositional field for each layer in CRUST1.0
- [ ] Add additional benchmarks and methods to assess accuracy


LudovicJnnt:
Below some gravity plot for testing the new crust1.0 plugin. 

GRAVITY - from gravity postprocessor using new crust1.0 plugin 
crust1-john-gm

GRAVITY - from gravity postprocessor using crust1.0 density from an ascii file. This model has the same resolution as above (GMR 7 - 25 000 000 cells and ~800m 
crust1-ascii-gm

GRAVITY - from spherical harmonic software
crust1-SH-gm

In terms of gravity magnitude, gravity from the new crust1.0 plugin is the closest to the gravity obtained with the SH software. But in terms of pattern, gravity from loading density ascii file is the closest to the gravity obtained with the SH software.

source/initial_composition/crust1.cc
```diff
@@ -0,0 +1,225 @@
...
+          const double depth = (position.norm() - 6371000.)/1000.;
+
+          int ilat = std::min(int(colat),179);
+          int ilon = std::min(int(lon),359);
+
+          int cr1_index = (ilat+1)*360 + ilon;
+          double cr1_rho = 0.;
+          if (depth < bnds[cr1_index][8])
+            {
+              cr1_rho = 3.3;
```
bobmyhill: There is a discrepancy here with the Crust1.0 input files, which have non-constant mantle densities below the Moho. There should be an easily accessible comment for the user to know that mantle densities are being set as constant here.

Perhaps add an input parameter for the user to set the mantle density?

source/initial_composition/crust1.cc
```diff
@@ -0,0 +1,225 @@
...
+// explicit instantiations
+namespace aspect
+{
+  namespace InitialComposition
+  {
+    ASPECT_REGISTER_INITIAL_COMPOSITION_MODEL(Crust1,
+                                              "crust1",
+                                              "A class that implements an initial compositional field containing "
+                                              "densities from the CRUST 1.0 data set. Note that this plugin only "
+                                              "works if there is a compositional field called 'crust1_densities'.")
```
bobmyhill: Perhaps also worth noting that the crust1.0 model includes data above sea level, which will be cut off if the model doesn't include topography.

source/initial_composition/crust1.cc
```diff
@@ -0,0 +1,225 @@
...
+      AssertThrow(dynamic_cast<const MaterialModel::Crust1<dim>*> (&this->get_material_model()) != nullptr,
+                  ExcMessage("The initial composition plugin `crust1' only works with "
+                             "the 'Crust1' material model."));
+
+      const unsigned int crust1_index = this->introspection().compositional_index_for_name("crust1_density");
+      if (compositional_index == crust1_index)
+        {
+          
+          // Get coordinates (surface elevation, longitude (0-360), latitude (-90 - 90))
+          std::array<double,dim> wcoord = Utilities::Coordinates::WGS84_coordinates(position);
```
bobmyhill: If a spherical shell is being used, wouldn't cartesian_to_spherical be a better choice than WGS84?
Also, const?

source/initial_composition/crust1.cc
```diff
@@ -0,0 +1,225 @@
...
+                
+          // Define location of current point
+          const double lon    = wcoord[1];       
+          const double lat    = wcoord[2];
+          const double colat  = 90. - lat;
+
+          // The reference radius is hardwired in here and needs to be
+          // an input parameter.
+          const double depth = (position.norm() - 6371000.)/1000.;
```
bobmyhill: Assert that Spherical shell.outer radius = 6371.e3, then use that value?

source/initial_composition/crust1.cc
mnkajan: Good catch. The variable name should probably be changed to `scoord` for consistency as well, e.g.:
```suggestion
          const std::array<double,dim> scoord = Utilities::Coordinates::cartesian_to_spherical_coordinates(position);
```

gassmoeller:
What is the current status of this PR, in particular when compared to #3149? Do they both add the same functionality? I like that this one already has a cookbook and figures, but I like about #3149 that it does not seem to need additional very specific plugins. If both offer the same functionality, could we use the parameter file approach from #3149 in this one to avoid adding the plugins?

One comment about the data: Both PRs currently contain a huge amount of data. Can we cut this down to something reasonable (e.g. < 10,000 lines)? E.g. by only adding the original data, and then using a script to generate the extra files for ASPECT? How big is the original data set?

bobmyhill:
Hi @gassmoeller 
To allow #3149 to do everything this PR can do, one would still have to create a material model plugin that takes the gravity data from the initial composition plugin. But the rest is there.

The original dataset is ~200k lines. In conjunction with #3149 one could potentially add a "crust10" datafile, which would reduce the dataset to ~2k lines. We'd have to decide whether we wanted to antialias that smaller dataset (I'd argue yes).

naliboff:
@bobmyhill thanks for the reminder about these two PRs. I recall one issue for both PRs was the size of the ascii data files being added into the repository. Could we use some of the libdap functionality implemented by @kneumiller and others to circumvent this issue?

bangerth:
What is the size if you run bzip2 on it? We can read compressed files with no problem through `boost::iostreams`.

If that's not small enough, one could think about https://github.com/LLNL/fpzip

bobmyhill:
> What is the size if you run bzip2 on it? We can read compressed files with no problem through `boost::iostreams`.

That helps a lot. bzip2 takes each of my crust1-derived files down to 140 kb (there are ten files in total).
bzip2 takes each of the raw crust1 files down to 440 kb (there are four files in total).

So there isn't any significant difference between the two in terms of size. ~1.5 Mb is still a largish footprint, but probably reasonable.

I like the suggestion from @gassmoeller (to take components from both of these PRs), but I'll be happy with any decision that allows me to close my PR :D.

bangerth:
I have no particular stake in this and so will happily let others make the call, just wanted to point out the possibility of compressing data.

@gassmoeller ?

naliboff:
@bobmyhill @bangerth - great that bzip2 gets the files down to a reasonable level! I think 1.5 Mb total is probably reasonable to have in the repository, but the maintainers should make that decision.

@bobmyhill - I prefer your approach #3149 given that it does not require an entire new interface for reading in the CRUST1.0 data. And we don't necessarily need an entirely new material model to merge #3149. Perhaps the most useful thing would be to have a cookbook with a self-contained material model plugin that composites on top of an existing material model. Alternatively, perhaps there could be a new equation_of_state module for CRUST1.0? In the end, I think people will want use the CRUST1.0 data sets in a number of different ways, and simply showing one example of _how to use the data_ is more than sufficient.

To sum up, I think the approach in #3149 + a new simple cookbook plugin or EOS module is the best path forward.

bangerth:
On 8/11/20 8:59 AM, naliboff wrote:
> @bobmyhill <https://github.com/bobmyhill> @bangerth 
> <https://github.com/bangerth> - great that bzip2 gets the files down to a 
> reasonable level! I think 1.5 Mb total is probably reasonable to have in the 
> repository, but the maintainers should make that decision.

Uh, you're one of them, @naliboff :-)

bobmyhill:
@bangerth Is there a bzip2-file-reading functionality in ASPECT/dealII, or is this something I'll have to add to the AsciiData plugin?

bangerth:
It would work similarly to here: https://github.com/dealii/dealii/blob/master/include/deal.II/base/utilities.h#L1301-L1305

bobmyhill:
> It would work similarly to here: https://github.com/dealii/dealii/blob/master/include/deal.II/base/utilities.h#L1301-L1305

Thanks @bangerth ... I've got a new branch started just for the bzip read (https://github.com/bobmyhill/aspect/tree/allow_bzip), but at the linker stage of compilation I get lots of undefined reference errors like:

`.../deal.II-v9.2.0/include/deal.II/bundled/boost/iostreams/filter/bzip2.hpp:357: error: undefined reference to 'boost::iostreams::detail::bzip2_base::bzip2_base(boost::iostreams::bzip2_params const&)'`

Is this likely to be a problem with my deal.II install (a missing dependency?)? Or am I forgetting something trivial?

bangerth:
Hm. Does the same happen with the gzip compression?

bobmyhill:
> Hm. Does the same happen with the gzip compression?

No, gzip works fine.

bobmyhill:
@bangerth the files are ~60% bigger when gzip compressed, so bzip2 would be preferable if the problem is easily fixed.

bobmyhill:
@bangerth Any ideas? If not, I'll go with gzip for the time being, and we can add bzip2 later.

tjhei:
I am not sure if bzip is a useful path to take:
1. We are concerned about the size of the git repository not the storage of the current version on disk. Git compresses snapshots using zlib, if I remember correctly.
2. A more sensible question might be to look into netcdf files that store things in binary form (the support for that in AsciiData is in progress but I can move that along)
3. Not storing in the main repo is an option as well (libdap or some other repo)

bobmyhill:
Hi @tjhei ... ok, in that case I'll concentrate on updating the PR and we can worry about file size (or not) later.

naliboff:
Closing this PR as it will be superseded by #3149. @bobmyhill - Please extract and use any code in this PR that will be useful in #3149, and please ping as soon as a review would be helpful there!

PR #3126, opened 2019-05-31
Unstructured lookup

pmbremner:
I added a new class to read in unstructured data tables. The class is modeled off of the ASCII data reader. The purpose in mind was to read in multiple unstructured material properties files, but this new class is not restricted to this purpose. The data files should contain columns of coordinates, and columns of data. The coordinates are stored in a searchable kdtree object, and the data are stored in vectors. The user can query the n-nearest neighbor data points from a target point. When applied to material properties, this allows the ability to update compositional  properties on-the-fly based on current pressure and temperature conditions.


pmbremner:
Made redundant by pull request #3128

PR #3128, opened 2019-05-31
Addition of an unstructured lookup class

pmbremner:
Follow-up for #3126. I added a new class to read in unstructured data tables. The class is modeled off of the ASCII data reader. The purpose in mind was to read in multiple unstructured material properties files, but this new class is not restricted to this purpose. The data files should contain columns of coordinates, and columns of data. The coordinates are stored in a searchable kdtree object, and the data are stored in vectors. The user can query the n-nearest neighbor data points from a target point. When applied to material properties, this allows the ability to update compositional properties on-the-fly based on current pressure and temperature conditions.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1018,14 +1018,16 @@ namespace aspect
      * Note the required format of the input data: The first lines may contain
      * any number of comments if they begin with '#', but one of these lines
      * needs to contain the number of columns and lines in the file:
-    * '#POINTS: TOTALNUMBERDATALINES NUMBERCOORDCOLUMNS NUMBERDATACOLUMNS'.
-    * For example '# POINTS: 5 4 3' would be 5 lines of data, 4 columns of coordinates,
-    * and 3 columns of data. The comments can optionally be followed by a
+     * '#POINTS: TOTALNUMBERDATALINES NUMBERCOORDCOLUMNS NUMBERDATACOLUMNS'.
+     * For example '# POINTS: 5 4 3' would be 5 lines of data, 4 columns of coordinates,
+     * and 3 columns of data. The comments can optionally be followed by a
      * single line, which does not start with '#', containing the names of
      * the data columns.
      * The order of the following data columns has to be
      * 'coordinates data' with @p querydim coordinate columns and @p components
-     * data columns. The coordinates do not need to be equidistant.
+     * data columns. @p querydim and @p components need to
+     * match NUMBERCOORDCOLUMNS NUMBERDATACOLUMNS, respectivily.
+     * The coordinates do not need to be equidistant.
      */
     template <int querydim>
     class UnstructuredDataLookup
@@ -1033,8 +1035,7 @@ namespace aspect
       public:
 
         /**
-         * The following variables are properties of the material files
-         * we read in.
+         * The following are member variables and functions of the class
          */
         unsigned int n_data_files;
 
@@ -1049,7 +1050,7 @@ namespace aspect
 
         virtual void initialize (const MPI_Comm &comm,
                                  std::string datadirectory,
-                                 std::vector<std::string> material_file_names);
+                                 std::vector<std::string> file_name_list);
 
 
 
@@ -1083,13 +1084,13 @@ namespace aspect
 
 
         /**
-         * Returns the data (velocity, temperature, etc. - according
+         * Returns the data (e.g., velocity, temperature, etc. - according
          * to the used plugin) for the n closest points to the target point.
-        * The function returns a vector of pairs of < data_return , Euclidean distance >,
-        * where data_return is a vector containing all the data properties
-        * associated with a single line in the file -- all the data columns.
-        * The Euclidean distance is the distance that the data points
-        * are from the target point.
+         * The function returns a vector of pairs of < data_return , Euclidean distance >,
+         * where data_return is a vector containing all the data properties
+         * associated with a single line in the file -- all the data columns.
+         * The Euclidean distance is the distance that the data points
+         * are from the target point.
          */
         std::vector< std::pair<std::vector<double>, double> >
         get_data (const Point<querydim> &target_point, const unsigned int n_points) const;
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -31,6 +31,7 @@
 #include <deal.II/base/function_lib.h>
 #include <deal.II/dofs/dof_handler.h>
 #include <deal.II/fe/component_mask.h>
+#include <deal.II/numerics/kdtree.h>
 
 #include <aspect/coordinate_systems.h>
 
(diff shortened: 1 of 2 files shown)


include/aspect/utilities.h
```diff
@@ -1011,6 +1012,139 @@ namespace aspect
...
 
+
+    /**
+     * UnstructuredDataLookup reads in files containing input data in ascii format.
+     * Note the required format of the input data: The first lines may contain
+     * any number of comments if they begin with '#', but one of these lines
+     * needs to contain the number of columns and lines in the file:
+    * '#POINTS: TOTALNUMBERDATALINES NUMBERCOORDCOLUMNS NUMBERDATACOLUMNS'.
+    * For example '# POINTS: 5 4 3' would be 5 lines of data, 4 columns of coordinates,
+    * and 3 columns of data. The comments can optionally be followed by a
```
MFraters: could you align the stars here?

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,139 @@ namespace aspect
...
+     * single line, which does not start with '#', containing the names of
+     * the data columns.
+     * The order of the following data columns has to be
+     * 'coordinates data' with @p querydim coordinate columns and @p components
```
MFraters: so x,y,z,c1,c2,...? And should `guerydim` be `dim`?

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,139 @@ namespace aspect
...
+     * data columns. The coordinates do not need to be equidistant.
+     */
+    template <int querydim>
+    class UnstructuredDataLookup
+    {
+      public:
+
+        /**
+         * The following variables are properties of the material files
```
MFraters: What do you mean with material files?

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,139 @@ namespace aspect
...
+        std::vector< std::vector<double> > data;
+
+
+        // Create an instance of a kdtree object as a member variable of the class
+        KDTree<querydim> coordinate_kdtree;
+
+
+        virtual void initialize (const MPI_Comm &comm,
+                                 std::string datadirectory,
+                                 std::vector<std::string> material_file_names);
```
MFraters: same, what do you mean with material?

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,139 @@ namespace aspect
...
+         * exist, if the data file format is incorrect or if the file grid
+         * changes over model runtime.
+         */
+        void
+        load_file(const std::string &filename,
+                  const MPI_Comm &communicator);
+
+
+        /**
+         * Returns the data (velocity, temperature, etc. - according
```
MFraters: e.g. velocity, ....?

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,139 @@ namespace aspect
...
+         * to the used plugin) for the n closest points to the target point.
+        * The function returns a vector of pairs of < data_return , Euclidean distance >,
```
MFraters: please align stars :)

MFraters (commented):
I only looked at the header file yet, will do the rest later. Looks well documented to me! Just some minor fixes.

pmbremner:
Addressed comments

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+     * needs to contain the number of columns and lines in the file:
+     * '#POINTS: TOTALNUMBERDATALINES NUMBERCOORDCOLUMNS NUMBERDATACOLUMNS'.
+     * For example '# POINTS: 5 4 3' would be 5 lines of data, 4 columns of coordinates,
+     * and 3 columns of data. The comments can optionally be followed by a
+     * single line, which does not start with '#', containing the names of
+     * the data columns.
+     * The order of the following data columns has to be
+     * 'coordinates data' with @p querydim coordinate columns and @p components
+     * data columns. @p querydim and @p components need to
+     * match NUMBERCOORDCOLUMNS NUMBERDATACOLUMNS, respectivily.
```
tjhei:
```suggestion
     * match NUMBERCOORDCOLUMNS and NUMBERDATACOLUMNS, respectively.
```

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+         */
+        unsigned int n_data_files;
+
+        std::vector< Point<querydim> > coordinate_values;
+
+        std::vector< std::vector<double> > data;
+
+
+        // Create an instance of a kdtree object as a member variable of the class
+        KDTree<querydim> coordinate_kdtree;
```
tjhei: this should be private

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+
+
+        virtual void initialize (const MPI_Comm &comm,
+                                 std::string datadirectory,
```
tjhei:
```suggestion
                                 const std::string &datadirectory,
```

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+                                 std::vector<std::string> file_name_list);
```
tjhei:
```suggestion
                                 const std::vector<std::string> &file_name_list);
```

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+        /**
+         * Constructor that explicitly prescribes the number of data columns
+         * in the data file. If a list of data components is provided in the
+         * data file it is checked that the length of this list is consistent
+         * with this number of components. This constructor is mostly provided
+         * for backwards compatibility. Not prescribing the number of components
+         * and instead reading them from the input file allows for more
+         * flexible files.
+         */
+        UnstructuredDataLookup(const unsigned int components);
```
tjhei:
```suggestion
        explicit UnstructuredDataLookup(const unsigned int components);
```

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+
+
+
+        /**
+         * Constructor that explicitly prescribes the number of data columns
+         * in the data file. If a list of data components is provided in the
+         * data file it is checked that the length of this list is consistent
+         * with this number of components. This constructor is mostly provided
+         * for backwards compatibility. Not prescribing the number of components
+         * and instead reading them from the input file allows for more
```
tjhei: Do we need this constructor? I think the same one in AsciiDataLookup is only there for backwards compatibility.

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+         */
+        std::vector<std::string> data_component_names;
+
+        /**
+         * Number of points in the data as specified in the data file.
+         * This is declared in the header as "#POINTS:", and
+         * must have three compnents:
+         * TOTAL_NUMBER_DATA_LINES NUMBER_COORD_COLUMNS NUMBER_DATA_COLUMNS.
+         */
+        std::vector<int> point_dimensions;
```
tjhei: this might not be necessary to store as a member but a local variable might be enough, right?

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+         * Returns the data (e.g., velocity, temperature, etc. - according
+         * to the used plugin) for the n closest points to the target point.
+         * The function returns a vector of pairs of < data_return , Euclidean distance >,
+         * where data_return is a vector containing all the data properties
+         * associated with a single line in the file -- all the data columns.
+         * The Euclidean distance is the distance that the data points
+         * are from the target point.
+         */
+        std::vector< std::pair<std::vector<double>, double> >
+        get_data (const Point<querydim> &target_point, const unsigned int n_points) const;
```
tjhei: I think it would be great to have a second function that only returns the closest point to make the return value simpler.

include/aspect/utilities.h
```diff
@@ -1011,6 +1012,140 @@ namespace aspect
...
+
+        /**
+         * Returns a vector that contains the names of all data columns in the
```
tjhei:
```suggestion
         * Return a vector that contains the names of all data columns in the
```

tjhei (commented):
This looks pretty good already!

tjhei:
Can you add a unit test similar to the one you showed us a couple of days ago?

include/aspect/utilities.h
pmbremner: I don't know. Perhaps I can try with and without this constructor in the unit_test, and decide from there.

include/aspect/utilities.h
pmbremner: Same here

include/aspect/utilities.h
tjhei: ``dealii::RTree<std::size_t, boost::geometry::index::linear<16>,     dealii::IndexableGetterFromIndices<std::vector>> ``

source/utilities.cc
```diff
@@ -2416,6 +2416,294 @@ namespace aspect
...
+      n_data_files = file_name_list.size();
+
+      // Read in values from each file in the list
+      for (unsigned int i = 0; i < n_data_files; i++)
+        {
+          load_file (datadirectory+file_name_list[i], comm);
+        }
+
+      // Set the coordinate points into the kdtree object
+      coordinate_kdtree.set_points (coordinate_values);
```
tjhei, line 2434: pack_rtree_of_indices(const ContainerType &container);

bobmyhill:
@Arijitchkc Is this of value to you? I know we talked about unstructured lookups with reference to MageMin intgration.

Arijitchkc:
Yes I think this is really useful. I also have something similar myself with a kd-tree and a hash. This is useful.

bangerth:
@Arijitchkc We'd be happy to take a patch with a completed implementation! :-)

PR #3129, opened 2019-05-31
additions to the changelog

pmbremner:
Added changelog files for the pull-requests made during the hackathon.


doc/modules/changes/20190531_bremner
```diff
@@ -0,0 +1,2 @@
+New: Addition of an unstructured data lookup class within aspect/utilities.<br>
```
tjhei: this should be part of #3128, which is not merged yet.

PR #3142, opened 2019-06-02
Initial isotherm plugin

bobmyhill:
This PR adds an Initial temperature plugin which allows the user to specify the depths of isotherms as a function of "horizontal" coordinates (x, y or phi, theta) in a set of ASCII files. The depth of the isotherms are interpolated linearly as a function of horizontal natural coordinates (either x-y or lat-lon), and then the temperature structure is interpolated vertically as piecewise constant or linear between isotherms.

The plugin uses AsciiDataLayered, a new class in Utilities. This could be used for a variety of different plugins (for example, to parameterize Crust 1.0).


include/aspect/initial_temperature/ascii_data_layered.h
```diff
@@ -0,0 +1,85 @@
...
+         * beginning of the program. Checks preconditions.
+         */
+        void
+        initialize ();
+
+        // avoid -Woverloaded-virtual:
+        using Utilities::AsciiDataLayered<dim>::initialize;
+
+        /**
+         * Return the boundary temperature as a function of position. For the
```
tjhei: copy and pasted?

include/aspect/initial_temperature/ascii_data_layered.h
```diff
@@ -0,0 +1,85 @@
...
+
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    using namespace dealii;
+
+    /**
+     * A class that implements a prescribed temperature field determined from
+     * a AsciiData input file.
```
tjhei: please update

source/initial_temperature/ascii_data_layered.cc
```diff
@@ -0,0 +1,117 @@
...
+#include <aspect/global.h>
+#include <aspect/initial_temperature/ascii_data_layered.h>
+
+namespace aspect
+{
+  namespace InitialTemperature
+  {
+    template <int dim>
+    AsciiDataLayered<dim>::AsciiDataLayered ()
+    {}
```
tjhei: can the constructor be deleted?

source/utilities.cc
```diff
@@ -2277,6 +2277,182 @@ namespace aspect
 
 
 
+    template <int dim>
+    AsciiDataLayered<dim>::AsciiDataLayered ()
+    {}
```
tjhei: remove?

source/utilities.cc
```diff
@@ -2277,6 +2277,182 @@ namespace aspect
...
+    template <int dim>
+    void
+    AsciiDataLayered<dim>::initialize(const unsigned int components)
+    {
+      AssertThrow ((dynamic_cast<const GeometryModel::SphericalShell<dim>*> (&this->get_geometry_model()))
+                   || (dynamic_cast<const GeometryModel::Chunk<dim>*> (&this->get_geometry_model())) != nullptr
+                   || (dynamic_cast<const GeometryModel::Sphere<dim>*> (&this->get_geometry_model())) != nullptr
+                   || (dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model())) != nullptr,
+                   ExcMessage ("This ascii data plugin can only be used when using "
+                               "a spherical shell, chunk or box geometry."));
```
tjhei: or sphere

source/utilities.cc
```diff
@@ -2277,6 +2277,182 @@ namespace aspect
...
+
+      // Create the lookups for each file and a vector of doubles corresponding to the layer_values
+      std::string filename;
+      for (unsigned int i=0; i<layer_boundary_names.size(); ++i)
+        {
+          filename = data_directory + data_file_names[i];
```
tjhei: const std::string filename =

data/initial-temperature/ascii-data/test/initial_isotherm_1000K_box_2d.txt
```diff
@@ -0,0 +1,7 @@
+# Test data for ascii data initial temperature conditions.
```
tjhei: you need to update the comments on all of these files.

source/utilities.cc
```diff
@@ -2277,6 +2277,182 @@ namespace aspect
...
+                                             const std::string &subsection_name)
+    {
+      Utilities::AsciiDataBase<dim>::parse_parameters(prm, subsection_name);
+
+      prm.enter_subsection(subsection_name);
+      {
+
+        data_directory = Utilities::expand_ASPECT_SOURCE_DIR(prm.get ("Data directory"));
+        data_file_names    = Utilities::split_string_list(prm.get ("Data file names"), ',');
+        layer_boundary_names = Utilities::split_string_list(prm.get("Layer boundary values"), ',');
```
tjhei: are those names or values?!

bobmyhill: Ooops, I used this parameter when I had a different file-finding structure. Changed!

source/initial_temperature/ascii_data_layered.cc
bobmyhill: I don't think so? At least, when I try, I get a bunch of `error: type ‘aspect::Utilities::AsciiDataLayered<2>’ is not a base type for type ‘aspect::InitialTemperature::AsciiDataLayered<2>’` errors

bobmyhill:
> I am not quite sure what the difference between this and the existing asciidata plugin is. Maybe this because some of the docs are missing? Can you explain (or even better extend the comments)?

In the other AsciiData plugin, the temperature is read in using a <dim>D spatial grid that spans the domain. In this plugin, the data is read in as a series of <dim-1>D grids, which each represent an isotherm. For each point in the domain, the temperature is evaluated by first finding the isotherms bounding the data point, then finding the depth of those isotherms, and finally interpolating between them (either linearly or piece-wise constant). 

This is a much more compact data structure than the full <dim>D grid for many geological applications (because vertical/radial temperature profiles are often well-approximated by a few linear T-z segments).

bobmyhill:
I just noticed that if we want to use this for Crust 1 and other similar composition initial conditions, we need to allow AsciiDataLayered to interpolate properties on the layer boundaries. The new commit implements that change.

include/aspect/utilities.h
```diff
@@ -941,6 +941,85 @@ namespace aspect
...
+        void
+        initialize (const unsigned int components);
+
+
+        /**
+         * Returns the data component at the given position.
+         */
+        double
+        get_data_component (const Point<dim>                    &position,
+                            const unsigned int                   component) const;
```
gassmoeller: Please remove the many spaces between the parameters and their types.

source/initial_temperature/ascii_data_layered.cc
```diff
@@ -0,0 +1,118 @@
...
+
+
+    template <int dim>
+    void
+    AsciiDataLayered<dim>::initialize ()
+    {
+      Utilities::AsciiDataLayered<dim>::initialize(2);
```
gassmoeller: Hmm, why 2? Should it not be a single temperature component?

source/utilities.cc
```diff
@@ -2277,6 +2277,174 @@ namespace aspect
...
+
+
+    template <int dim>
+    void
+    AsciiDataLayered<dim>::initialize(const unsigned int components)
+    {
+      AssertThrow ((dynamic_cast<const GeometryModel::SphericalShell<dim>*> (&this->get_geometry_model()))
+                   || (dynamic_cast<const GeometryModel::Chunk<dim>*> (&this->get_geometry_model())) != nullptr
+                   || (dynamic_cast<const GeometryModel::Sphere<dim>*> (&this->get_geometry_model())) != nullptr
+                   || (dynamic_cast<const GeometryModel::Box<dim>*> (&this->get_geometry_model())) != nullptr,
```
gassmoeller: I know this is still common in other parts of the code, but could you please replace this with the new and easier structure: `Plugins::plugin_type_matches<GeometryModel::Box<dim> >(this->get_geometry_model())` ? Same for all the other checks.

source/utilities.cc
```diff
@@ -2277,6 +2277,174 @@ namespace aspect
...
+        }
+    }
+
+
+
+    template <int dim>
+    double
+    AsciiDataLayered<dim>::
+    get_data_component (const Point<dim>                    &position,
+                        const unsigned int                   component) const
```
gassmoeller: please remove spaces

source/utilities.cc
```diff
@@ -2277,6 +2277,174 @@ namespace aspect
...
+        }
+
+      // Find which layer we're in
+      unsigned int layer_boundary_index=0;
+      // if position < position of the first boundary layer, stop
+      double old_difference_in_vertical_position = vertical_position - lookups[layer_boundary_index]->get_data(horizontal_position,0);
+      double difference_in_vertical_position = old_difference_in_vertical_position;
+      while (difference_in_vertical_position > 0. && layer_boundary_index < number_of_layer_boundaries-1)
+        {
+          layer_boundary_index++;
```
gassmoeller:
```suggestion
          ++layer_boundary_index;
```

source/utilities.cc
```diff
@@ -2333,8 +2501,6 @@ namespace aspect
       return lookup->get_data(internal_position,component);
     }
 
-
-
```
gassmoeller: Please leave these line in here.

bobmyhill:
Thanks @gassmoeller , I've addressed your comments, so this should be good to go as and when the tester is done.

gassmoeller:
Thanks!

Issue #3144, opened 2019-06-03
Add missing #DEAL_II_VERSION_GTE


bangerth:
#3048 and https://github.com/dealii/dealii/pull/8294 have now both been merged. As a consequence, we should put the `#ifdef DEAL_II_VERSION_GTE(9,2,0)` back into the code. There is really no hurry in this, and maybe we should give people a bit of time to update their deal.II versions if they work from deal.II dev, but we should do it before the next release.

@bobmyhill -- FYI.

gassmoeller:
Closed by #3145.

PR #3145, opened 2019-06-03
added if !DEAL_II_VERSION_GTE(9,2,0) back into compat.h

bobmyhill:
gradient/value is now implemented for InterpolatedUniformGridData in deal.II and will be contained in the 9.2 release. In this PR, we add an if statement in compat.h to only use the compatibility functions if the deal.II version is <9.2.
Fixes Issue #3144

### DIFF
--- a/include/aspect/compat.h
+++ b/include/aspect/compat.h
@@ -31,7 +31,7 @@
 // for std_cxx14::make_unique:
 #include <deal.II/base/std_cxx14/memory.h>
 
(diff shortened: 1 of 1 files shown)


bangerth:
Ready to merge, but let's sit on this for a week before merging so that everyone who uses deal.II dev versions has time to update.

Thanks, @bobmyhill, for the quick fix!

bobmyhill:
@bangerth, you're welcome! Anything for a break from marking exam scripts!

PR #3154, opened 2019-06-12
Introduce a reset function for heating model outputs

jdannberg:
This fixes a bug that occurred when a model used several heating models and some of them filled the rates of temperature change heating model outputs (needed for operator splitting models with latent heat), but some of them didn't. Because many heating models just don't touch the rates of temperature change heating model outputs at all, we need to reset these values to zero before we evaluate those heating models, otherwise they will contain rates of temperature change outputs from a heating model that was evaluated previously, and these outputs will be wrong and will affect the temperature.
I have also added a test where this is important (the melting rate was only half of its intended value before due to the wrong entry in the heating outputs).


tjhei:
It looks like shear_heating_with_melt needs updated output. Ready to merge after the testers are happy.

jdannberg:
Wow, I am surprised that the test didn't catch this (it runs into the exact problem I describe above). 
When we switched over from the old melting model to operator splitting, this must have changed the test results. But I guess the changes were just too small to notice something was wrong.

tjhei:
excellent! :-)

Do you want to add a changelog entry? Is this important enough? If not, please hit merge yourself.

jdannberg:
You are right, there should be a changelog entry. I added one.

PR #3157, opened 2019-06-14
Rename isothermal to isentropic compression

gassmoeller:
This changes the name of the isothermal compression approximation to isentropic compression. As discussed in our recent manuscript the current name is misleading and leads to inconsistent models. I have thought about keeping backward compatibility by still allowing the old name, but (1) I think nearly nobody specified this approximation explicitly in the input file (it was the default), and (2) if they did, crashing with an error will force everyone to think about the change and make sure they fix their models.


tjhei:
Looks good and I am okay with forcing people to change the name.

PR #3159, opened 2019-06-17
Add pull request template

gassmoeller:
How does everyone feel about adding a default pull request template to the repository? I feel like it can save some time doing reviews, and I constantly forget to add (or ask for) changelog entries, or wait for half an hour for the tester only to discover I forgot to run `make indent` again. I admit it can be a bit annoying to always have the text field prefilled with a lot of text, but maybe if we keep the template short enough it is worth it? Opinions?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,7 @@
 Pull Request Checklist. Please read and check each box with an X. Feel free to delete any part not applicable to your PR.
 
+### Describe the content of this pull request, and why you did these changes:
+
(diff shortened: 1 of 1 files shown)


naliboff:
I'm in favor of adding the template, although I'm not exactly not sure how the template works as it relates to opening a new PR. Does one only have to check the boxes when the pull request is first opened?

gassmoeller:
Yes, it prepopulates the description text with the text of the template. So it does not help if you make changes and forget to run `make indent` again, or if test results change later. But at least it tells everyone (including myself) to do it right from the start, which is in particular important for the change log entries, because I keep forgetting to ask for them.

PR #3162, opened 2019-06-18
[CI]: test manual on CIG tester

gassmoeller:
We keep breaking our manual, because we do not have a test for it. This PR creates a new docker image for the tester (including latex) and should test the creation of the manual for every PR. We can also extend it later to update the input parameters.

Having the tester as a separate image from the deal.II image will also be helpful for #2978.

### DIFF
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -110,7 +110,9 @@ pipeline {
 
     stage('Build Documentation') {
       steps {
-        sh 'cd doc && make manual.pdf'
+        sh 'cd doc && make manual.pdf || touch ~/FAILED-DOC'
+        archiveArtifacts artifacts: 'doc/manual/manual.log', allowEmptyArchive: true
+        sh 'if [ -f ~/FAILED-DOC ]; then exit 1; fi'
       }
     }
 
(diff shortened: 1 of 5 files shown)


gassmoeller:
i have added commits that show how the tester works. My questions would be:
- Do we want to make this change? I am in favor of it.
- Do we want to change both testers? I think it would be nice to have them doing the same things.
- Should I include updating the parameters file for the manual immediately in this PR?

I can remove the test commits once others have seen how they work.

tjhei:
Great, that looks good!

tjhei:
you need to do ``cd doc && ./update_parameters.sh ~/build*/aspect``

Jenkinsfile.cig
```diff
@@ -108,6 +108,15 @@ pipeline {
       }
     }
 
+    stage('Build Documentation') {
+      steps {
+        sh './doc/update_parameters.sh ./build-gcc-fast/aspect || touch ~/FAILED-DOC'
```
tjhei: you can remove the ``|| touch ...``. Then the command will fail immediately (below I do this so that we get the .log file).

gassmoeller:
Already found and fixed some parameter problems :-). I assume you approval means I can merge?

Issue #3161, opened 2019-06-18
Compositing material model does not allow for the underlying material models to provide additonal output


anne-glerum:
The `visco plastic` material model fills several additional outputs, e.g. `plastic_yielding`. When using the `compositing` material model to refer to the `visco_plastic` model to compute the viscosity, these outputs are not filled. This came up when asking for the `named additional outputs` visualization postprocessor. 

Although there is no guarantee that the components requested from the underlying material model correspond to the additional output that this material model computes (think `thermal_conductivity` versus `plastic_yielding`), I think in principle it should be an option to ask the underlying material models for their additional output when using the `compositing` material model.

gassmoeller:
That seems indeed wrong. Is the problem that it is not possible to ask for the additional outputs, or is the problem that they will never be created in the first place (because `Compositing` does not forward the call to `create_additional_named_outputs` to the base models)? Because if they are created they should also be available from `MaterialModelOutputs`.

anne-glerum:
`Compositing` does not forward the `create_additional_named_outputs` call to the base models, nor does it do anything else with AdditionalOutput. 

The error message when activating the pp:
`Exception 'ExcMessage("You activated the postprocessor <named additional outputs>, but there are no additional outputs " "provided by the material model. Either remove the postprocessor, or check why no output is provided.")' on rank 0 on processing: 
`
`An error occurred in line <68> of file </Users/ACG/Documents/PhD/Aspect/source/25-05-2019/aspect/source/postprocess/visualization/named_additional_outputs.cc> in function
    virtual void aspect::Postprocess::VisualizationPostprocessors::NamedAdditionalOutputs<2>::initialize() [dim = 2]
The violated condition was: 
    out.additional_outputs.size() > 0
Additional information: 
    You activated the postprocessor <named additional outputs>, but there are no additional outputs provided by the material model. Either remove the postprocessor, or check why no output is provided.`

gassmoeller:
> Compositing does not forward the create_additional_named_outputs call to the base models, nor does it do anything else with AdditionalOutput.

Then it might be enough to create such a function in `Compositing` and let it call the corresponding functions of all the base models it owns?

anne-glerum:
> Then it might be enough to create such a function in `Compositing` and let it call the corresponding functions of all the base models it owns?

[...]

PR #3164, opened 2019-06-21
Add additional named outputs to the compositing material model

anne-glerum:
This pr addresses #3161.


gassmoeller:
Yes, this is nearly what I had in mind in #3161. Your version needs a bit less code changes, but also is a bit less efficient than my suggestion of making `base_out` a member and only generating the outputs once in `create_additional_named_outputs` instead of once per cell in `evaluate`. I am fine either way.
But I noticed one additional problem: This will only address all of the additional material model outputs that are generated by material models. There are other places that can generate additional outputs, e.g. the `MeltHandler` and the `NewtonHandler` classes. If your current PR works for you let's go with this for now (it solves a part of the problem) and think about the other cases separately. I do not see an easy way to fix them, except for changes to the general structure of the `AdditionalMaterialModelOutputs`.

anne-glerum:
> Your version needs a bit less code changes, but also is a bit less efficient than my suggestion of making `base_out` a member and only generating the outputs once in `create_additional_named_outputs` instead of once per cell in `evaluate`. 

Sorry, I should have explained why I did it this way. I did not make `base_out` a member of the `Compositing` class because the `create__additional_named_outputs` function is `const` and there is no copy for `MaterialModelOutputs` objects (only move). What would be your way around that?

> There are other places that can generate additional outputs, e.g. the `MeltHandler` and the `NewtonHandler` classes. If your current PR works for you let's go with this for now (it solves a part of the problem) and think about the other cases separately.
Yes, the current fix works for me. I can update the issue to the left cases?

gassmoeller:
> Sorry, I should have explained why I did it this way. I did not make base_out a member of the Compositing class because the create__additional_named_outputs function is const and there is no copy for MaterialModelOutputs objects (only move). What would be your way around that?

Ah, I had not thought of that. You are right then your way is better.

> Yes, the current fix works for me. I can update the issue to the cases that are left?

Ok, good. Yes please update the issue.

source/material_model/compositing.cc
```diff
@@ -86,16 +86,31 @@ namespace aspect
     {
       typename Interface<dim>::MaterialModelOutputs base_output(out.viscosities.size(),
                                                                 this->introspection().n_compositional_fields);
+      create_additional_named_outputs(base_output);
```
gassmoeller: Please leave a comment here that this is an incomplete fix, as it can only create the additional outputs generated by material models, but not by other places.

PR #3166, opened 2019-06-24
[CI]: unify and comment files

tjhei:
- make files more similar
- lead sh section with a comment (displayed while running)


gassmoeller:
Looks good. At some point we can just use the same file.

PR #3167, opened 2019-06-25
Use the same initialization for operator splitting as for normal time stepping

gassmoeller:
If using the operator splitting: After updating the solution vectors with the reaction terms, but before solving for a timestep we need to reinitialize the current linearization point to account for the changes that happened to the solution vector. This was done before already, however, only as a simple copy of the old solution, instead of the linear extrapolation that the BDF2 scheme assumes. This probably affected accuracy, but definitely affected the initial guess for the solution, needing on average more solver iterations. The fix unifies the normal initialization behavior with the one we use after the operator splitting to avoid this inconsistency in the future.


tjhei:
Extracting it makes a lot of sense. :+1:
Iteration numbers are not always lower, but within +-2, so I am okay with that.

Issue #1867, opened 2017-07-21
make operator splitting more effective


jdannberg:
In the compute_reactions() function that is used by the operator splitting solver option, the values for some degrees of freedom are set more than once in the loop where we assign the new values to distributed_vector (if they are located on the interface between cells), as we loop over all cells, and then over all degrees of freedom on each cell. The results are still correct, but we do some unnecessary work. 
At the moment, we also do all operations twice, once for the temperature and once for the compositional fields. As they might use different finite elements, we loop through their elements separately, and update the temperature and the compositions for both. This also duplicates some code. 
In addition, we use a simple forward Euler scheme with a fixed time step, instead of a more sophisticated ODE solver. 

TODO:
- [ ] touch each DoF only once
- [ ] see if there is a way to handle temperature and composition together
- [ ] improve ODE solver

PR #3168, opened 2019-06-25
Refactor operator splitting function

gassmoeller:
This refactors the operator splitting functionality and includes the changes from #3167. #3167 should be merged first, as it contains changes to the test results. 
This PR should not create any changes. It usually saves 50% of the operator splitting time, because if the temperature and composition element are identical we can skip a lot of the duplicated computations.

This addresses a part of #1867.

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -1538,12 +1538,18 @@ namespace aspect
 
     // we have to update the old solution with our reaction update too
     // so that the advection scheme will have the correct time stepping in the next step
-    LinearAlgebra::BlockVector tmp(distributed_vector);
+    LinearAlgebra::BlockVector tmp;
+    tmp.reinit(distributed_vector, false);
 
+    // What we really want to do is
+    //     old_solution.block(block_index) += distributed_reaction_vector.block(block_index);
+    // but because 'old_solution' is a ghosted vector, we can't write into it directly. Rather,
+    // we have to go around with a completely distributed vector.
     tmp.block(block_index) = old_solution.block(block_index);
     tmp.block(block_index) +=  distributed_reaction_vector.block(block_index);
     old_solution.block(block_index) = tmp.block(block_index);
 
+    // Same here with going through a distributed vector.
     tmp.block(block_index) = old_old_solution.block(block_index);
     tmp.block(block_index) +=  distributed_reaction_vector.block(block_index);
     old_old_solution.block(block_index) = tmp.block(block_index);
@@ -1580,7 +1586,7 @@ namespace aspect
     Assert (reaction_time_step_size > 0,
             ExcMessage("Reaction time step must be greater than 0."));
 
-    pcout << "   Solving composition reactions ";
+    pcout << "   Solving composition reactions... " << std::flush;
 
     // make one fevalues for the composition, and one for the temperature (they might use different finite elements)
     const Quadrature<dim> quadrature_C(dof_handler.get_fe().base_element(introspection.base_elements.compositional_fields).get_unit_support_points());
@@ -1652,12 +1658,10 @@ namespace aspect
     for (const auto &cell : dof_handler.active_cell_iterators())
       if (cell->is_locally_owned())
         {
-          cell->get_dof_indices (local_dof_indices);
-
           fe_values_C.reinit (cell);
           in_C.reinit(fe_values_C, cell, introspection, solution);
 
-          if (!temperature_and_composition_use_same_fe)
+          if (temperature_and_composition_use_same_fe == false)
             {
               fe_values_T.reinit (cell);
               in_T.reinit(fe_values_T, cell, introspection, solution);
@@ -1718,6 +1722,8 @@ namespace aspect
                 }
             }
 
+          cell->get_dof_indices (local_dof_indices);
+
           // copy reaction rates and new values for the compositional fields
           for (unsigned int j=0; j<dof_handler.get_fe().base_element(introspection.base_elements.compositional_fields).dofs_per_cell; ++j)
             for (unsigned int c=0; c<introspection.n_compositional_fields; ++c)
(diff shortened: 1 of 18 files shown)


tjhei:
can you rebase, please?

gassmoeller:
I rebased and all tests pass. Ready for a review.

gassmoeller:
Ping. Could I get a review for this PR? The logic is a bit complicated, but the changes are purely refactoring and it speeds up the operator splitting significantly.

bangerth:
I'll take a look at this today.

source/simulator/helper_functions.cc
```diff
@@ -1529,13 +1529,39 @@ namespace aspect
...
+  {
+    solution.block(block_index) = distributed_vector.block(block_index);
+
+    // we have to update the old solution with our reaction update too
+    // so that the advection scheme will have the correct time stepping in the next step
+    LinearAlgebra::BlockVector tmp(distributed_vector);
+
+    tmp.block(block_index) = old_solution.block(block_index);
+    tmp.block(block_index) +=  distributed_reaction_vector.block(block_index);
+    old_solution.block(block_index) = tmp.block(block_index);
```
bangerth: So if I see this right, you're doing
```
  A = B
  A += C
  B = A
```
This would be equivalent to `B += C`. Why does that not work? Because the distribution of vectors is different? If that's the case, could you add a comment of the form
```
  // What we really want to do is 
  //     old_solution.block(block_index) += distributed_reaction_vector.block(block_index);
  // but because 'old_solution' is a ghosted vector, we can't write into it directory. Rather,
  // we have to go around with a completely distributed vector.
```

source/simulator/helper_functions.cc
```diff
@@ -1529,13 +1529,39 @@ namespace aspect
...
+  template <int dim>
+  void Simulator<dim>::update_solution_vectors_with_reaction_results (const unsigned int block_index,
+                                                                      const LinearAlgebra::BlockVector &distributed_vector,
+                                                                      const LinearAlgebra::BlockVector &distributed_reaction_vector)
+  {
+    solution.block(block_index) = distributed_vector.block(block_index);
+
+    // we have to update the old solution with our reaction update too
+    // so that the advection scheme will have the correct time stepping in the next step
+    LinearAlgebra::BlockVector tmp(distributed_vector);
```
bangerth: Could you do
```suggestion
    LinearAlgebra::BlockVector tmp(distributed_vector, false);
```
to avoid the actual copying of vector elements (which you don't need)?

source/simulator/helper_functions.cc
```diff
@@ -1529,13 +1529,39 @@ namespace aspect
...
+
+    tmp.block(block_index) = old_solution.block(block_index);
+    tmp.block(block_index) +=  distributed_reaction_vector.block(block_index);
+    old_solution.block(block_index) = tmp.block(block_index);
+
+    tmp.block(block_index) = old_old_solution.block(block_index);
```
bangerth:
```suggestion
    // Same here with going through a distributed vector.
    tmp.block(block_index) = old_old_solution.block(block_index);
```

source/simulator/helper_functions.cc
```diff
@@ -1624,12 +1652,16 @@ namespace aspect
     for (const auto &cell : dof_handler.active_cell_iterators())
       if (cell->is_locally_owned())
         {
-          fe_values_C.reinit (cell);
           cell->get_dof_indices (local_dof_indices);
```
bangerth: can you just push this line down to immediately above the first block that uses `local_dof_indices`?

source/simulator/helper_functions.cc
```diff
@@ -1624,12 +1652,16 @@ namespace aspect
...
+
+          fe_values_C.reinit (cell);
           in_C.reinit(fe_values_C, cell, introspection, solution);
 
-          fe_values_T.reinit (cell);
-          in_T.reinit(fe_values_T, cell, introspection, solution);
+          if (!temperature_and_composition_use_same_fe)
```
bangerth:
```suggestion
          if (temperature_and_composition_use_same_fe == false)
```

source/simulator/helper_functions.cc
```diff
@@ -1698,64 +1736,43 @@ namespace aspect
...
-        distributed_vector.block(block_c) = old_solution.block(block_c);
-        distributed_vector.block(block_c) +=  distributed_reaction_vector.block(block_c);
-        old_solution.block(block_c) = distributed_vector.block(block_c);
-
-        distributed_vector.block(block_c) = old_old_solution.block(block_c);
-        distributed_vector.block(block_c) +=  distributed_reaction_vector.block(block_c);
-        old_old_solution.block(block_c) = distributed_vector.block(block_c);
-
-        operator_split_reaction_vector.block(block_c) = distributed_reaction_vector.block(block_c);
+        update_solution_vectors_with_reaction_results(block_c,distributed_vector,distributed_reaction_vector);
```
bangerth: just inline the initializer of `block_c` into the call here and get rid of the curly braces

source/simulator/helper_functions.cc
```diff
@@ -1554,10 +1580,7 @@ namespace aspect
     Assert (reaction_time_step_size > 0,
             ExcMessage("Reaction time step must be greater than 0."));
 
-    pcout << "   Solving composition reactions in "
-          << number_of_reaction_steps
-          << " substep(s)."
-          << std::endl;
+    pcout << "   Solving composition reactions ";
```
bangerth: This will likely not show on the screen until you complete the line. You might have to add a `std::flush` to make it show.

Do we have other places where we place preliminary output on a line and later complete it? I think that sometimes we end the preliminary text with `...` but I'm not sure whether we're consistent.

source/simulator/helper_functions.cc
gassmoeller: There is no copy constructor like that, but I used `reinit (vector, false)` instead, which does not copy entries.

gassmoeller:
Thanks for the review! I addressed your comments.

bangerth:
You'll have to update a few output files where it was "reactions" before and is now "reactions...".

Ok otherwise.

PR #3169, opened 2019-06-26
Better Compositing model and AdditionalOutputs

tjhei:
This avoid two evaluations of the contained material models by creating
a function that allows moving additional outputs between objects.

See #3164


anne-glerum:
Neat!

gassmoeller:
Anne did you make a test for your scenario that failed before? If so, feel free to merge. If not, can you check if your model works with this branch, and if so, merge?

tjhei:
related to #3164

anne-glerum:
I don't think I added a test. Let me check this today.

include/aspect/material_model/interface.h
```diff
@@ -537,6 +537,11 @@ namespace aspect
        */
       template <class AdditionalOutputType>
       const AdditionalOutputType *get_additional_output() const;
+
+      /**
+       * Steal the additional outputs from @p other.
+       */
+      void take_additional_outputs_from(MaterialModelOutputs<dim> &other);
```
bangerth: Since you're moving things, may I suggest naming the function `move_...` instead of `take_...`? Using standard nomenclature is useful, I think.

bangerth:
Thinking about it, it would also be useful to document that this overwrites any additional material model output that might already have been associated with the recipient.

anne-glerum:
- There is no test for the `compositing` material model i.c.w. `named additional outputs`
- This pr works for combination of `compositing` material model with `named additional outputs` 
- This pr gives a segmentation fault for switching to the `compositing` material model in test steinberger_seismic_velocities.prm

anne-glerum:
@tjhei This pr will fix the problem reported by Sungho on the Forum that properties are not mapped correctly. It was probably introduced in #3164.

tjhei:
- added a test with named outputs that looks correct
- renamed the function
- assert and document that destination needs to be empty

gassmoeller:
@anne-glerum: Did Timo address your comments? If so, can you merge?

anne-glerum:
@tjhei When I tested your PR, I noticed the following:

> This pr gives a segmentation fault for switching to the compositing material model in test steinberger_seismic_velocities.prm (a test with different additional outputs) when named additional outputs are requested.

If I remember correctly, this also happened before your changes. So is this something that should be addressed in this PR? Or merge this and then address any other problems?

gassmoeller:
If it is not directly related to the changes of this PR then lets move it to a follow-up PR. Can you make an issue out of that Anne?

tjhei:
> Can you make an issue out of that Anne?

I know what the bug is and it is unrelated to my PR. I have a fix incoming as well.

gassmoeller:
This is ready then. @anne-glerum feel free to merge.

anne-glerum:
@tjhei Unfortunately the tester now fails on `compositing_02`.

tjhei:
This is weird, why would the output of #3225 be changed by this PR?

tjhei:
> This is weird, why would the output of #3225 be changed by this PR?

I think this is because the old values were incorrect, as Steinberger was not initialize()ed when I generated the output before #3225 was merged. Results updated.

anne-glerum:
I'm slightly confused as `compositing_02.prm` was introduced in #3225 right? So you're saying you didn't create the test results for #3225 with the changes in #3225?

tjhei:
> I'm slightly confused

So am I, you are right, this can't be true. I have no idea. Let me look into it some more...

anne-glerum:
It's probably because of the bug that this PR fixes, right? As reported on the forum, when 2 different material models are used in `Compositing`, parameters are only taken from the first material model. This should be easy to check, whether `Reference specific heat` for example is set from the `Simple` model or the `Steinberger`model.

anne-glerum:
@tjhei have you had a chance to check it?

tjhei:
okay, so #3164 incorrectly overwrites all outputs in ``out``. The test was introduced in #3225 based on this incorrect data.
So, I think this can be merged as is.

PR #3172, opened 2019-07-04
Fix deprecation message with deal.II 9.1

gassmoeller:
Fixes some deprecation messages caused by the new mesh deformation plugin system (the branch overlapped with the fixes that were already done in all other places where we use constraint matrices).


tjhei:
Does this really fix it? I think if you use ``dealii::ConstraintMatrix`` you will get a warning. 

Instead, we need a ``using`` statement like the one in simulator_access.h.

gassmoeller:
Well, it did fix the message. I suppose because `simulator_access.h` was included in all the files that used `ConstraintMatrix`? Anyway, I added the using statement to interface.h as well, just to be sure.

PR #3173, opened 2019-07-04
Replace dynamic cast in cookbooks and benchmarks

gassmoeller:
Since #2148 there is not really a use any more for casting pointers to determine the type of a plugin, and it is confusing for new users. This PR replaces all occurences of `dynamic_cast` with the appropriate functions. In a few cases I had to change the order of checks, and I removed a bit outdated code.

### DIFF
--- a/cookbooks/anisotropic_viscosity/av_material.cc
+++ b/cookbooks/anisotropic_viscosity/av_material.cc
@@ -193,6 +193,7 @@ namespace aspect
     {
       internal::Assembly::Scratch::StokesPreconditioner<dim> &scratch = dynamic_cast<internal::Assembly::Scratch::StokesPreconditioner<dim>& > (scratch_base);
       internal::Assembly::CopyData::StokesPreconditioner<dim> &data = dynamic_cast<internal::Assembly::CopyData::StokesPreconditioner<dim>& > (data_base);
+
       const MaterialModel::AnisotropicViscosity<dim> *anisotropic_viscosity =
         scratch.material_model_outputs.template get_additional_output<MaterialModel::AnisotropicViscosity<dim> >();
 
@@ -525,13 +526,13 @@ namespace aspect
     {
       for (unsigned int i=0; i<assemblers.stokes_preconditioner.size(); ++i)
         {
-          if (dynamic_cast<Assemblers::StokesPreconditioner<dim> *>(assemblers.stokes_preconditioner[i].get()) != nullptr)
+          if (Plugins::plugin_type_matches<Assemblers::StokesPreconditioner<dim>>(*(assemblers.stokes_preconditioner[i])))
             assemblers.stokes_preconditioner[i] = std_cxx14::make_unique<Assemblers::StokesPreconditionerAnisotropicViscosity<dim> > ();
         }
 
       for (unsigned int i=0; i<assemblers.stokes_system.size(); ++i)
         {
-          if (dynamic_cast<Assemblers::StokesIncompressibleTerms<dim> *>(assemblers.stokes_system[i].get()) != nullptr)
+          if (Plugins::plugin_type_matches<Assemblers::StokesIncompressibleTerms<dim>>(*(assemblers.stokes_system[i])))
             assemblers.stokes_system[i] = std_cxx14::make_unique<Assemblers::StokesIncompressibleTermsAnisotropicViscosity<dim> > ();
         }
     }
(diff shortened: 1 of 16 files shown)


cookbooks/anisotropic_viscosity/av_material.cc
```diff
@@ -191,8 +191,11 @@ namespace aspect
     execute (internal::Assembly::Scratch::ScratchBase<dim>   &scratch_base,
              internal::Assembly::CopyData::CopyDataBase<dim> &data_base) const
     {
-      internal::Assembly::Scratch::StokesPreconditioner<dim> &scratch = dynamic_cast<internal::Assembly::Scratch::StokesPreconditioner<dim>& > (scratch_base);
-      internal::Assembly::CopyData::StokesPreconditioner<dim> &data = dynamic_cast<internal::Assembly::CopyData::StokesPreconditioner<dim>& > (data_base);
+      internal::Assembly::Scratch::StokesPreconditioner<dim> &scratch =
```
tjhei: are you sure this works? This is certainly not a plugin, so I would leave the old code.

cookbooks/anisotropic_viscosity/av_material.cc
```diff
@@ -276,8 +279,10 @@ namespace aspect
     execute (internal::Assembly::Scratch::ScratchBase<dim>   &scratch_base,
              internal::Assembly::CopyData::CopyDataBase<dim> &data_base) const
     {
-      internal::Assembly::Scratch::StokesSystem<dim> &scratch = dynamic_cast<internal::Assembly::Scratch::StokesSystem<dim>& > (scratch_base);
-      internal::Assembly::CopyData::StokesSystem<dim> &data = dynamic_cast<internal::Assembly::CopyData::StokesSystem<dim>& > (data_base);
+      internal::Assembly::Scratch::StokesSystem<dim> &scratch =
+        Plugins::get_plugin_as_type<internal::Assembly::Scratch::StokesSystem<dim>> (scratch_base);
```
tjhei: same here

cookbooks/anisotropic_viscosity/av_material.cc
```diff
@@ -525,13 +530,13 @@ namespace aspect
     {
       for (unsigned int i=0; i<assemblers.stokes_preconditioner.size(); ++i)
         {
-          if (dynamic_cast<Assemblers::StokesPreconditioner<dim> *>(assemblers.stokes_preconditioner[i].get()) != nullptr)
+          if (Plugins::plugin_type_matches<Assemblers::StokesPreconditioner<dim>>(*(assemblers.stokes_preconditioner[i])))
```
tjhei: this is also not a plugin

cookbooks/inner_core_convection/inner_core_assembly.cc
```diff
@@ -71,8 +71,10 @@ namespace aspect
       execute (internal::Assembly::Scratch::ScratchBase<dim>   &scratch_base,
                internal::Assembly::CopyData::CopyDataBase<dim> &data_base) const
       {
-        internal::Assembly::Scratch::StokesSystem<dim> &scratch = dynamic_cast<internal::Assembly::Scratch::StokesSystem<dim>& > (scratch_base);
-        internal::Assembly::CopyData::StokesSystem<dim> &data = dynamic_cast<internal::Assembly::CopyData::StokesSystem<dim>& > (data_base);
+        internal::Assembly::Scratch::StokesSystem<dim> &scratch =
+          Plugins::get_plugin_as_type<internal::Assembly::Scratch::StokesSystem<dim>> (scratch_base);
```
tjhei: same here

gassmoeller:
You are right about the scratch objects, I was in a copy-paste meditation at the time and just replaced every dynamic_cast :-). 
However, I would argue that the assemblers are plugins: They have an interface and can be selected through run-time options in the parameter file. They can not be selected explicitly by name at the moment, and the using file still needs to include their header file to use them, but I would say that still qualifies as plugin. Also maybe the name plugin in the function is a bit misleading, in fact the function does more something like: Does a dynamic_cast succeed? Anyway, would you be ok with the current version?

tjhei:
> Also maybe the name plugin in the function is a bit misleading, in fact the function does more something like: Does a dynamic_cast succeed?

If you rename the function, then you might as well just do a dynamic_cast and check the result. :-)

> However, I would argue that the assemblers are plugins:

In my mind plugins are the things we select from a PluginList and that the user can register and then select, which means the assemblers do not fit into this (right now).
Anyways, I am okay with your choice.

PR #3174, opened 2019-07-05
Correct behavior of VoF interface refinement algorithm near MPI process boundaries

class4kayaker:
### Describe the content of this pull request, and why you did these changes:

When running a test problem, I encountered a case where the VoF advection algorithm was not encountering the maximally refined region near the fluid interface. Upon farther testing, the issue was localized to being a case where a cell that should have been marked as neighboring the interface in the first pass was not causing cells sharing vertices to be refined. This is intended to fix this case by adding a communication step that will ensure that any cell which is marked by the owning process will be marked on all processes. Because the owning process must have all neighboring ghost cells, this should make certain that any cell which would be marked on any process will be marked on all processes that need to reference it.

### Before submitting your first pull request:

* [x] I have read the guidelines in our [CONTRIBUTING.md](../CONTRIBUTING.md) document.

### For all Pull Requests:

* [x] I have followed the instructions for indenting my code according to ASPECT's code style as described in the last paragraph of the ['Making ASPECT better'](../CONTRIBUTING.md#making-aspect-better) section of CONTRIBUTING.md.

### For Feature/Benchmark Submissions or Changes of Existing Behavior:


class4kayaker:
Is there an easy way to induce this state for testing? The line in question uses a function documented to refer to the global dof count, and is needed to bypass the case where only the triangulation exists, but not the state data.

bangerth:
@class4kayaker: The way to test this is to use more processors than cells. So use a mesh with 4 cells and 5 processors, or similar. This way, at least one processor can not own a cell.

bangerth:
In reference to #835.

class4kayaker:
Test added.

gassmoeller:
@tjhei: Did Jon address your concerns?

tjhei:
> @tjhei: Did Jon address your concerns?

Yes, this is correct. I assumed that the ``n_dofs`` check at the beginning of the function could happen on some processors, but this is not true (it returns the global number of DoFs). :+1:

PR #3175, opened 2019-07-05
Fix floating point exception caused by accessing timestep prematurely

gassmoeller:
This fixes a floating point exception reported by @KiralyAgi in the anisotropic viscosity cookbook. Agi could you take a look and see if this version fixes your problems? You can see the code changes by clicking on `Files changed`, and you can download this version by doing:

```
git remote add gassmoeller
git fetch gassmoeller fix_floating_point_exception_in_av_cookbook
git checkout gassmoeller/fix_floating_point_exception_in_av_cookbook
```

Then recompile and see if it works. If it works, and when this pull request has been merged you can switch back to the main branch and update to the newest version to get the changes as usual, i.e. `git checkout master; git pull upstream master`. 

Also if that succeeds, can you create a test case for your cookbook, so that we know if something breaks in future versions? The procedure is described here: http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests. You can for example take the files in `tests/time_dependent_annulus.prm/.cc/...` and copy them to `tests/anisotropic_viscosity_cookbook...`.

### DIFF
--- a/cookbooks/anisotropic_viscosity/av_material.cc
+++ b/cookbooks/anisotropic_viscosity/av_material.cc
@@ -617,91 +617,96 @@ namespace aspect
           for (unsigned int i=0; i<dim; ++i)
             n[i] = in.composition[q][c_idx_n[i]];
 
-          Tensor<1,dim> n_dot;
-          if (n.norm() >0.5 && in.composition[q][c_idx_gamma] > 0.8)
+          // The computation of the viscosity tensor is only
+          // necessary after the simulator has been initialized.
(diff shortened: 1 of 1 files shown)


KiralyAgi:
Thanks @gassmoeller! It resolved the problem, now it runs also on Abel (UiO cluster). I will proceed with creating a test.

gassmoeller:
Let's merge and do the test  as a follow-up. Better to get the cookbook fixed as soon as possible.

PR #3176, opened 2019-07-06
Move assignment to old_solution to the only place it is used

gassmoeller:
Currently the `interpolate_material_output_into_compositional_field` function interpolates the material output both into `solution`and `old_solution`, because the `prescribed_field_with_diffusion` advection method needs it that way (because it afterwards solves a diffusion equation as if the interpolation had not happened). This does not matter for all current applications, but I want to use the prescribed field to compute a time-derivative of the composition, and the current assignment to `old_solution` prevents that.

Also it feels weird to me that the function modifies the solutions of previous time steps, because it is intended to create the current compositional field. This PR solves this, by moving the assignment out of the function into the advection method. Technically that changes the behavior of the function, but it is private and only used twice inside `Simulator`.

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1248,13 +1248,8 @@ namespace aspect
        * interpolated to the finite element space and consequently into the
        * solution vector.
        * This is useful for compositional fields whose advection mode is set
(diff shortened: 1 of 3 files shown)


tjhei:
That is certainly not intuitive, but I agree that it is cleaner now.

PR #3181, opened 2019-07-15
Reorder writing checkpoints

gassmoeller:
This is a reliability improvement for ASPECT's snapshot/restart ability. We occasionally have models that get cancelled or crash during writing a snapshot (e.g. disk full, wallclock time reached). Because we rename the old snapshot before writing the new one that leaves ASPECT in a non-restartable state, which is very annoying if you want to chain a lot of model runs on a cluster. This PR reverses the order of operations. We now first write the new checkpoint into a temporary file, then rename the old checkpoint, then rename the new checkpoint. There is still a window where ASPECT is not restartable (during renaming), but it should be much more reliable now.

### For all Pull Requests:

* [X] I have followed the instructions for indenting my code according to ASPECT's code style as described in the last paragraph of the ['Making ASPECT better'](../CONTRIBUTING.md#making-aspect-better) section of CONTRIBUTING.md.

### For Feature/Benchmark Submissions or Changes of Existing Behavior:


source/simulator/checkpoint_restart.cc
```diff
@@ -388,6 +366,41 @@ namespace aspect
 
     }
 
+    // Now rename the snapshots to put the new one in place of the old one.
```
tjhei: Can you add an MPI_Barrier before this line, please? I am a little nervous that not all processors might be done writing the ``restart.mesh`` when processor 0 gets here. This wasn't a problem with the old code, as the ``mv`` happened before the next snapshot only.

gassmoeller:
Good point, I added a MPI barrier. I also noticed we did not have a parallel checkpoint test so I added one. Could you take another look?

tjhei:
great! :+1:

PR #3182, opened 2019-07-15
Streamline pull request template

gassmoeller:
After a few pull requests with the new template, I would like to shorten it a bit in some places, and add a message for how to get help with the process. I hope this makes it less annoying for experienced developers, and less intimidating for new developers. I am open for suggestions about improvements.


tjhei:
I am not sure yet if I like having a template in place. My brain is confused reading useless information at the top of every PR I am reviewing. Maybe it will grow on me over time. :-)
That said, this is certainly an improvement of the current status.

gassmoeller:
Yes, I agree. I am also not decided yet (if we ever decide we want to get rid of it I am fine with that). But on the plus side I noticed that I already improved some of my PRs because of it, e.g. I remembered that I forgot to indent. Let's keep it in place for a few weeks or months and see how it plays out.

PR #3184, opened 2019-07-23
Add a rheological parameter to visco_plastic

naliboff:

* [X] I have read the guidelines in our [CONTRIBUTING.md](../CONTRIBUTING.md) document.

### For all pull requests:

This PR add a new rheological parameter to the Visco Plastic material model, which allows one to multiply the pre-yield viscosity by a specified value. 

This can be useful if one wants to make specific compositional fields stronger or weaker relative to other fields that otherwise have the same pre-yield viscosity (e.g., diffusion, dislocation, etc) parameters. 

A natural example is that faults may be inherently weaker than the surrounding rocks in the ductile portions of the lithosphere, but simply changing the flow law (or flow law parameters) may not make sense.


jdannberg:
But couldn't you achieve the same thing by changing the diffusion and dislocation creep prefactors in the input file? Yes, if you use a composite rheology, you would have to change both prefactors by the same amount (instead of just this one number), but do you think this is really worth adding a new parameter?

Or are there any specific cases you're thinking of when you say that changing the flow law parameters may not make sense?

naliboff:
Yes, one absolutely could change the prefactors to achieve the same result. 

In my view, the main argument for this approach is that one is much less likely to make errors when dealing with a large number of compositional fields and potentially multiple fields with different pre -yield factors. 

For the case on dislocation and composite creep, it certainly would not be clear (to me) what the different pre-yield viscosity factors are based on the prefactor terms in the equations. I would handle this in parameter files by adding a series of comments explaining what each of the prefactor terms represent, but at that point perhaps having a dedicated parameter is a simpler and more error-proof approach?

naliboff:
@anne-glerum @Djneu @MFraters (and others) - Thoughts on this? If you think the most straightforward approach is to just change the prefactor term, we should go with that.

MFraters:
I am in favor of this change. 

You could do it through changing the prefactors, but than you manually have to account for the the stress factors, which makes quick testing of different options more time consuming can confusing than needed. It also shows more clearly the intend of the composition in the input file (a two times weaker/stronger material)

naliboff:
@jdannberg @MFraters - Coming back to this now that other `visco_plastic` PRs are almost ready to merge.

How about instead of having the parameter `pre_yield_viscosity_factors` embedded `visco_plastic`, a new rheology model called `multiply_viscosity` could simply multiply the viscosity by a given factor. I think other existing material models that have composition dependent viscosity could take advantage of this.

Thoughts?

gassmoeller:
> How about instead of having the parameter pre_yield_viscosity_factors embedded visco_plastic, a new rheology model called multiply_viscosity could simply multiply the viscosity by a given factor. I think other existing material models that have composition dependent viscosity could take advantage of this.

I like this approach. We could even call the rheology module `constant_viscosity_prefactors`, and copy everything from `constant_viscosity` with little modifications. I would suggest the following interface for the `compute_viscosity` function:

```
double
ConstantViscosityPrefactors<dim>::compute_viscosity(const double viscosity, const unsigned int composition_index) const;
```
Viscosity is then multiplied with the prefactor with index `composition_index`. Do you think that sounds reasonable?

naliboff:
Closing this PR as it is superseded by #3444, which is based on the discussion above with @gassmoeller.

PR #3191, opened 2019-08-04
Compressible computations with GMG

tcclevenger:
Adds functionality for compressible flow computations with the block GMG precoditioner. I verified with the tangurnis benchmarks that we are solving for the same solution as with the block AMG preconditioner and added a test based on tangurnis/tala

I've also changed some function names to be more appropriate to their functionality:
`evaluate_viscosity`->`evaluate_material_model` 
and 
`fill_viscosities_and_pressure_scaling`->`fill_cell_data`
so if you have any thoughts there, let me know.

Note: Does not yet include functionality for implicit reference density computations.

### DIFF
--- a/tests/tangurnis_tala_gmg.prm
+++ b/tests/tangurnis_tala_gmg.prm
@@ -1,0 +1,18 @@
+# TanGurnis benchmark TALA using geometric multigrid preconditioner
+
+include $ASPECT_SOURCE_DIR/benchmarks/tangurnis/tala/tan.prm 
+
+subsection Material model
+  set Material averaging = harmonic average
+end
+
+subsection Solver parameters
+  subsection Stokes solver parameters
+    set Stokes solver type = block GMG
+  end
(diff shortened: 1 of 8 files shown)


tests/tangurnis_tala_gmg.prm
```diff
@@ -0,0 +1,331 @@
+# TanGurnis benchmark TALA
+
```
gassmoeller: instead of copying all this, can you replace it by the following:
```
# TanGurnis benchmark TALA using geometric multigrid preconditioner

include $ASPECT_SOURCE_DIR/benchmarks/tangurnis/tala/tan.prm 

Make all the GMG modifications here
```

tcclevenger:
Sorry, I forgot about the tests being run on different machines. All is fixed now and tests pass.

tjhei:
can you please squash down to 1 or 2 commits?

PR #3192, opened 2019-08-05
Simplify some logic

gassmoeller:
Just the simplification of some logic I saw when reviewing something else. Not sure if I disentangled this correctly, but the tester will tell.

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -672,23 +672,27 @@ namespace aspect
   template <int dim>
   void Simulator<dim>::assemble_stokes_system ()
   {
+    std::string timer_section_name = "Assemble Stokes system";
+
     // Matrix-free, only assemble RHS
     if (stokes_matrix_free)
       {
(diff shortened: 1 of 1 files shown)


source/simulator/assembly.cc
```diff
@@ -672,23 +672,27 @@ namespace aspect
...
+    std::string timer_section_name = "Assemble Stokes system";
+
     // Matrix-free, only assemble RHS
     if (stokes_matrix_free)
       {
         rebuild_stokes_matrix = false;
+        timer_section_name += " rhs";
+      }
+
+    if (assemble_newton_stokes_system)
```
tjhei: should this be "else if"?

gassmoeller:
You are right. Currently it does not matter because the settings are mutually exclusive, but I fixed it anyway.

tjhei:
> Currently it does not matter because the settings are mutually exclusive

You are right, but we can hope, right? :-)

PR #3212, opened 2019-08-27
Add pictures of an actual sandbox.

bangerth:
I came by an actual sandbox in the lab of Dennis Harry. I thought it might be neat to take a couple of pictures and put them into the manual to show that what we compute is not actually so dissimilar to what the real world produces in similar situations, even though I don't have the exact same scenarios to show that are in Buiter et al.

@sibiaoliu , @chronictectonic , @cedrict , @naliboff -- FYI, given that you are the original authors of that section.


bangerth:
On 8/27/19 4:47 PM, Timo Heister wrote:
> same here, it is weird how things have changed since the SVN days. :-)

Ha, yes, for sure!

PR #3213, opened 2019-09-01
add cmake PROJECT directive

tjhei:
Newer versions of cmake produce a warning that the unittest cmake
project doesn't have a PROJECT() statement. Fix this.


gassmoeller:
Alright, lets do this: :smile:

`make cmake-happy`

PR #3216, opened 2019-09-05
add some information on the 2.1 release that was missing on the website

jdannberg:
I couldn't test this, because the header doesn't show up in my browser, so let me know if you think any of this won't work.


tjhei:
Thanks, how did we forget that?!

Looks correct, let's see what happens!

bangerth:
Thanks. Don't we have a release checklist somewhere in the `doc/` subdirectory where this step should have been an item?

PR #3220, opened 2019-09-13
do not create temperature matrix for first_timestep_only_single_Stokes

tjhei:


tjhei:
@tcclevenger FYI

tjhei:
> Looks good. Somehow the upload of the coverage failed, lets see if a rebuild fixes that.

maybe because the commit is older?

gassmoeller:
Hmm, maybe. All other stages succeed and your tester is fine as well so I will merge. We should maybe look into making the coverage upload into a separate stage to clearly distinguish test failures from failure to post results.

PR #3221, opened 2019-09-17
Factor out diffusion and dislocation creep

anne-glerum:

This PR creates two new Rheology models for viscous creep and just entails moving code related to diffusion and dislocation to the respective Rheology models. For now it just removes code from visco_plastic. 

Question: I've made diffusion and dislocation creep a separate Rheology model, because I figured it would be easiest to for example refactor `diffusion_dislocation.cc` as well. However, one Rheology model called e.g. `viscous_creep` that includes the compositing of diffusion and dislocation also makes sense to me. What are your preferences @naliboff , @jdannberg ?

* [X] I have read the guidelines in our [CONTRIBUTING.md](../CONTRIBUTING.md) document.

### For all pull requests:

### DIFF
--- a/include/aspect/material_model/rheology/diffusion_creep.h
+++ b/include/aspect/material_model/rheology/diffusion_creep.h
@@ -56,7 +56,7 @@ namespace aspect
           parse_parameters (ParameterHandler &prm);
 
           /**
-           * Compute the viscosity, in this case just a constant.
+           * Compute the viscosity based on the diffusion creep law.
            */
           double
           compute_viscosity (const double pressure,
@@ -64,10 +64,30 @@ namespace aspect
                              const unsigned int composition) const;
 
         private:
+
+          /**
+           * List of diffusion creep prefactors A.
+           */
           std::vector<double> prefactors_diffusion;
+
+          /**
+           * List of diffusion creep grain size exponenents m.
+           */
           std::vector<double> grain_size_exponents_diffusion;
+
+          /**
+           * List of diffusion creep activation energies E.
+           */
           std::vector<double> activation_energies_diffusion;
+
+          /**
+           * List of diffusion creep activation volumes V.
+           */
           std::vector<double> activation_volumes_diffusion;
+
+          /**
+           * Diffusion creep grain size d.
+           */
           double grain_size;
       };
     }
--- a/include/aspect/material_model/rheology/diffusion_creep.h
+++ b/include/aspect/material_model/rheology/diffusion_creep.h
@@ -1,0 +1,76 @@
+/*
+  Copyright (C) 2019 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _aspect_material_model_rheology_diffusion_creep_h
+#define _aspect_material_model_rheology_diffusion_creep_h
+
+#include <aspect/global.h>
+#include <aspect/material_model/interface.h>
+#include <aspect/simulator_access.h>
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    namespace Rheology
+    {
+      template <int dim>
+      class DiffusionCreep : public ::aspect::SimulatorAccess<dim>
+      {
+        public:
+          /**
+           * Constructor.
+           */
+          DiffusionCreep();
+
+          /**
+           * Declare the parameters this function takes through input files.
+           */
+          static
+          void
+          declare_parameters (ParameterHandler &prm);
+
+          /**
+           * Read the parameters from the parameter file.
+           */
+          void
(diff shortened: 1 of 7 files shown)


naliboff:
@anne-glerum - @gassmoeller and I had a related discussion yesterday regarding how to move plastic behavior into the` Rheology` namespace. The options would be to put all types of plasticity (Drucker Prager, Stress Limiter, etc) into a single Rheology model (e.g., `Plastic`) or make a series of separate models.

I think one advantage of having a single model (`viscous_creep`, `plastic`) is the public functions within those models can easily trade information back and forth, thus making adding a `composite` option extremely straightforward. This should also reduce the amount of code in each material model.

In the case of plasticity, I'm envisioning moving things like `plastic_additional_outputs`, is_yielding functions, etc to the `plastic` model.

A disadvantage to this approach, which @gassmoeller pointed out, is each Rheology model then has the potential to become fairly complicated and difficult to follow (e.g., what we are trying to fix in `visco_plastic`).

My current preference is to have a single rheology model for each type of behavior (`viscous_creep`, `plastic`), but this seems like a decision where we should reach a group consensus.

gassmoeller:
Thanks Anne, I really like this change. I will do a more detailed review later. As John already mentioned, after a quick chat yesterday we did not come to a conclusion as to how fine-grained we want to split the rheologies. However, I tend to like your approach. The reasoning would be to separate everything for which the input parameters can be separated, i.e. there is no depedence from diffusion creep on dislocation creep or vice-versa so they should be separate modules. The same can be said of the different types of plasticity. Composite rheologies are indeed more complicated, but we could make another class for that (composite_viscous, composite_plastic) that could rely on the simple elements.

What opinions does everyone else have on this question?

naliboff:
If there is a straightforward way to have the composite rheology classes rely on the simpler elements, I am all for this approach.

mibillen:
I'll comment here that I think making very basic, independent elements, and then allow users to combine these pieces however they wish is the best approach.  For example, one might think that viscous creep is just diffusion + dislocation, but what about dislocation-accommodated grain boundary  sliding (disGBS) which is another viscous creep mechanism that may be important in the lithosphere, or low-temperature plasticity... this is a viscous mechanism also.  If each deformation mechanism is defined independently, then this allows individuals to decide which ones they want to include and how they want to combine them.   

In terms of how mechanisms are combined, there are also only a few choices - I can think of 2 common approaches: 1) strain-rates for all mechanisms are summed (including yielding ... commonly done, although may not really make sense physically), and  2) strain-rate for viscous mechanisms sum, and  yielding is applied after.   It may make sense to create two modules for these two approaches and then  allow the user to choose which viscous mechanisms to turn on/off and which type of yielding to turn on/off.

naliboff:
Thanks for the feedback @mibillen! For clarification, in either of the scenarios outlined above things like diffusion creep, dislocation creep, etc will be still be treated as independent elements, but the file organization and class structure would differ. Aside, I agree we should give the users maximum flexibility to choose how things are combined.

anne-glerum:
From the above I take it that there's consensus for having these small modules diffusion and dislocation, so I'll finish-up the documentation etc now before a formal review.

include/aspect/material_model/rheology/diffusion_creep.h
```diff
@@ -0,0 +1,96 @@
...
+    using namespace dealii;
+
+    namespace Rheology
+    {
+      template <int dim>
+      class DiffusionCreep : public ::aspect::SimulatorAccess<dim>
+      {
+        public:
+          /**
+           * Constructor.
```
naliboff: For consistency with `ConstantViscosity`, maybe add in `Initializes viscosity to NaN.` after `Constructor.`?

include/aspect/material_model/rheology/dislocation_creep.h
```diff
@@ -0,0 +1,92 @@
...
+    using namespace dealii;
+
+    namespace Rheology
+    {
+      template <int dim>
+      class DislocationCreep : public ::aspect::SimulatorAccess<dim>
+      {
+        public:
+          /**
+           * Constructor.
```
naliboff: Same comment as above.

include/aspect/material_model/rheology/diffusion_creep.h
anne-glerum: I removed it, because nothing is done in the Constructor actually, there is no viscosity parameter to initialize.

source/material_model/visco_plastic.cc
```diff
@@ -145,24 +145,11 @@ namespace aspect
...
-                                                (constants::gas_constant*temperature_for_viscosity)) *
-                                       std::pow(grain_size, grain_size_exponents_diffusion[j]);
-
-          // For dislocation creep, viscosity is grain size independent (m=0) and strain-rate dependent (n>1)
-          double viscosity_dislocation = 0.5 * std::pow(prefactors_dislocation[j],-1/stress_exponents_dislocation[j]) *
-                                         std::exp((activation_energies_dislocation[j] + pressure*activation_volumes_dislocation[j])/
-                                                  (constants::gas_constant*temperature_for_viscosity*stress_exponents_dislocation[j])) *
-                                         std::pow(edot_ii,((1. - stress_exponents_dislocation[j])/stress_exponents_dislocation[j]));
+          // Compute viscosity from iffusion creep law
+          double viscosity_diffusion = diffusion_creep.compute_viscosity(pressure, temperature_for_viscosity, j);
```
gassmoeller:
```suggestion
          const double viscosity_diffusion = diffusion_creep.compute_viscosity(pressure, temperature_for_viscosity, j);
```

source/material_model/visco_plastic.cc
```diff
@@ -145,24 +145,11 @@ namespace aspect
...
+
+          // Compute visocisty from dislocation creep law
+          double viscosity_dislocation = dislocation_creep.compute_viscosity(edot_ii, pressure, temperature_for_viscosity, j);
```
gassmoeller:
```suggestion
          const double viscosity_dislocation = dislocation_creep.compute_viscosity(edot_ii, pressure, temperature_for_viscosity, j);
```

PR #3219, opened 2019-09-12
Add a strain dependent rheology model

naliboff:
Pull Request Checklist. Please read and check each box with an X. Delete any part not applicable. Ask on the [forum](https://community.geodynamics.org/c/aspect) if you need help with any step.

This PR moves code related to tracking of finite strain and strain softening behavior from the `ViscoPlastic` material model to a new model `StrainDependent` in the `Rheology` class. By and large, very little code had to be modified during the transfer. Looking through the existing code I see lots of room for improvement, but I think that should be done in follow-up pull requests.

All the relevant tests pass on my local machine with the current changes, but we'll see what the tester says.

### For all pull requests:

* [x] I have followed the [instructions for indenting my code](../CONTRIBUTING.md#making-aspect-better).

### For new features/models or changes of existing features:

* [X] I have tested my new feature locally to ensure it is correct.
* [ ] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../tests/) directory.
* [X] I have added a changelog entry in the [doc/modules/changes](../doc/modules/changes) directory that will inform other users of my change.

PR #3223, opened 2019-09-17
remove duplicate strain softening parameters

naliboff:
This PR is a follow-up to #3219 and removes input parameters and related checks for the old strain softening syntax. This removes backward compatibility for the prior strain softening, but removes a significant amount of code and tricky logic. 

### DIFF
--- a/doc/modules/changes/20190917_naliboff
+++ b/doc/modules/changes/20190917_naliboff
@@ -1,5 +1,6 @@
-Remove: Input parameters and functionality related to the old syntax
-for strain softening in the Rheology::StrainDependent namespace. 
+Removed: Input parameters and functionality related to the old syntax
+for strain softening in the ViscoPlastic material model and the 
+Rheology::StrainDependent namespace. 
 <br>
 (John Naliboff, 2019/09/17)
 
--- a/doc/modules/changes/20190917_naliboff
+++ b/doc/modules/changes/20190917_naliboff
@@ -1,0 +1,5 @@
+Remove: Input parameters and functionality related to the old syntax
+for strain softening in the Rheology::StrainDependent namespace. 
+<br>
+(John Naliboff, 2019/09/17)
+
--- a/source/material_model/rheology/strain_dependent.cc
+++ b/source/material_model/rheology/strain_dependent.cc
@@ -90,46 +90,6 @@ namespace aspect
                            "but to guarantee that one uses either the old parameter names or the new ones, "
                            "never both.");
 
-        prm.declare_entry ("Use strain weakening", "default",
-                           Patterns::Selection("true|false|default"),
-                           "Apply strain weakening to viscosity, cohesion and internal angle "
-                           "of friction based on accumulated finite strain.  Units: None. "
-                           "By default, this parameter is set to false. The default option is there "
-                           "to make sure that the parameter is not used at the same time as the "
-                           "``Strain weakening mechanism'' parameter, which has the same functionality. "
-                           "This parameter is deprecated; please use ``Strain weakening mechanism'' "
-                           "instead!");
-
-        prm.declare_entry ("Use plastic strain weakening", "default",
-                           Patterns::Selection("true|false|default"),
-                           "Apply strain weakening to cohesion and internal angle "
-                           "of friction based on accumulated finite plastic strain only. Units: None. "
-                           "By default, this parameter is set to false. The default option is there "
-                           "to make sure that the parameter is not used at the same time as the "
-                           "``Strain weakening mechanism'' parameter, which has the same functionality. "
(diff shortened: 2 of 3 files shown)


source/material_model/rheology/strain_dependent.cc
```diff
@@ -329,16 +226,20 @@ namespace aspect
...
           {
             AssertThrow((this->get_parameters().nonlinear_solver ==
                          Parameters<dim>::NonlinearSolver::single_Advection_single_Stokes
                          ||
                          this->get_parameters().nonlinear_solver ==
-                         Parameters<dim>::NonlinearSolver::single_Advection_iterated_Stokes),
+                         Parameters<dim>::NonlinearSolver::single_Advection_iterated_Stokes
+                         ||
+                         this->get_parameters().nonlinear_solver ==
+                         Parameters<dim>::NonlinearSolver::single_Advection_iterated_Newton_Stokes),
```
jdannberg: This is a change in functionality. I guess it makes sense to allow this. But have we thought about how this works? As far as I can see, the Newton solver does not compute derivatives with respect to composition, but in a case with strain weakening, the viscosity could depend strongly on the different strain fields, correct?
Using a Newton solver without these terms probably still works much better than using the standard iterated Stokes, but do we think this is something that would be useful to add at some point in the future? Should we note that somewhere?

naliboff: Yes, good point! At minimum, I think we should add a comment noting this. @MFraters - Would this be difficult to implement?

naliboff:
@jdannberg - Thanks for the review! Yes, I think we should definitely include an update script. I'll work on this tomorrow.

gassmoeller:
Hmm, This update script might be tricky to write (because the new parameter can be a combination of older ones. You would essentially need to take the logic below, and reimplement it in a perl/sed/python script in doc/update_scripts/prm_files. If this turns out to be too complicated we should probably skip the update script.

naliboff:
Hmm, if the logic is going to be incredibly complicated for the update script, perhaps we could just make an announcement on the forum?

source/material_model/rheology/strain_dependent.cc
MFraters: Sorry for the late reply, I still have to find a way where github messages where I am mentioned will go to my main mail box :)

The Newton solver currently only uses the viscosity derivatives wrt the strain-rate and the pressure. In the visco-plastic plugin it does this for every composition and then averages the viscosities derivatives for the composition in the same way the viscosities for the compositions are averages. I haven't thought about what would happen with a composition which contains the strain, but by guess would be that it should be alright, because like a normal composition it influences the viscosity.

This means that the Newton solver currently only gives possible quadratic convergence for the nonlinearities from strain-rate and pressure dependence in viscosity. The rest of the nonlinearies have linear/Picard like convergence. 

If the viscosity is really strongly dependent on strain, and you want near quadratic convergence, we would need to compute the derivatives (this can be very easy if you just do a finite difference) and I think we would also need to solve it as a coupled system. Putting everything in a coupled system (we would then probably also put temperture and the other compositions in there) would be probably quite a bit of work because it changes the properties of the matrix we are solving and the way we need to precondition it. @bangerth might have a better idea on how difficult this would be and how that should be done.

relevant lines in visco-plastic:
https://github.com/geodynamics/aspect/blob/f457e03d3d805ae5022adc1b15063053566522e2/source/material_model/visco_plastic.cc#L304-L382

doc/modules/changes/20190917_naliboff
```diff
@@ -0,0 +1,5 @@
+Remove: Input parameters and functionality related to the old syntax
```
gassmoeller:
```suggestion
Removed: Input parameters and functionality related to the old syntax
```

doc/modules/changes/20190917_naliboff
```diff
@@ -0,0 +1,5 @@
...
+for strain softening in the Rheology::StrainDependent namespace.
```
gassmoeller:
```suggestion
for strain softening in the ViscoPlastic material model and the Rheology::StrainDependent namespace. 
```

naliboff:
@gassmoeller - I updated the change log based on your suggestions. If everyone else is comfortable using the forum in replace of an update script, I'll do that when the PR is merged.

@MFraters - Thanks for the explanation and good to know the Newton solver should still work even if we don't explicitly include the strain fields in the coupled system. However, in some models the viscosity is certainly strongly dependent on the strain fields, which are likely a first-order control in certain locations (shallow brittle crust, etc).

naliboff:
A forum announcement about the new syntax and removing backwards compatibility was poster here:
https://community.geodynamics.org/t/parameter-file-syntax-change-for-strain-weakening-behavior/816

Issue #3226, opened 2019-09-18
improve error message on constant temperture model using thermal energy density mesh refinement?


MFraters:
I was playing around with the convection box 3d cookbook and turning it into a shear model when I encountered the following error message:

```
----------------------------------------------------
Exception on MPI process <0> while running mesh refinement plugin <N6aspect14MeshRefinement20ThermalEnergyDensityILi3EEE>: 

--------------------------------------------------------
An error occurred in line <900> of file <***/2019-09-11-dealii-manual/dealii/source/numerics/derivative_approximation.cc> in function
    void dealii::DerivativeApproximation::internal::approximate_cell(const dealii::Mapping<dim, spacedim>&, const DoFHandlerType<dim, spacedim>&, const InputVector&, unsigned int, const dealii::TriaActiveIterator<dealii::DoFCellAccessor<DoFHandlerType<dim, spacedim>, false> >&, typename DerivativeDescription::Derivative&) [with DerivativeDescription = dealii::DerivativeApproximation::internal::Gradient<3>; int dim = 3; DoFHandlerType = dealii::DoFHandler; InputVector = dealii::TrilinosWrappers::MPI::BlockVector; int spacedim = 3; typename DerivativeDescription::Derivative = dealii::Tensor<1, 3>]
The violated condition was: 
    determinant(Y) != 0
Additional information: 
    We have encountered a cell on which the number of linearly independent directions that span the matrix Y (discussed in the documentation of the DerivativeApproximation class) is not equal to dim. The matrix Y then is rank deficient and can not be inverted.
--------------------------------------------------------

Aborting!
----------------------------------------------------
```

Thinking out it, it made sense to me, because I set the temperature to be constant everywhere  in the model, which means that there is no gradient. But since this model is the default mesh refinement strategy (if nothing is given), I thought this message might be a bit cryptic for new users tying aspect and I am wondering whether we can improve this.

One option would be to somehow check in advance whether the temperature is constant everywhere, which I think will probably be difficult. An other option would be to catch the deal.ii error and provide an extra suggestion for the user to make sure that there model actually has a temperature gradient or set an other refinement strategy. I am also fine with leaving it this way, if that is the preference of other people.

https://github.com/geodynamics/aspect/blob/2fcc82139062ee80a185f2b42b985a3809e27b6c/source/mesh_refinement/thermal_energy_density.cc#L103-L109

PR #3228, opened 2019-09-20
Improve mesh ref thermal desn grad error message

MFraters:
This improves the error message in one of the ways as suggested in issue #3226. Discussed the options with @gassmoeller and we agreed that improving the error message with a catch was the best option.

* [ X] I have read the guidelines in our [CONTRIBUTING.md](../CONTRIBUTING.md) document.

### For all pull requests:

* [ X] I have followed the [instructions for indenting my code](../CONTRIBUTING.md#making-aspect-better).

* [ X] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../tests/) directory.


source/mesh_refinement/thermal_energy_density.cc
```diff
@@ -101,11 +101,36 @@ namespace aspect
...
+        {
+          std::cerr << std::endl << std::endl
+                    << "----------------------------------------------------"
+                    << std::endl;
+          std::cerr << "Exception on MPI process <"
+                    << Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)
+                    << "> while running mesh refinement plugin <"
+                    << "thermal energy density"
+                    << ">: " << std::endl
+                    << "Mesh refinement plugin thermal energy density could failed to find a gradient. "
```
gassmoeller: I would reword this a bit, e.g. like this:
```suggestion
                    << "The mesh refinement plugin failed to compute a thermal energy density gradient. "
```

tests/thermal_energy_density_error_message.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env perl
```
gassmoeller: please rename this file to `thermal_energy_density_error_message.pl` since it is a perl script. Also add a comment at the top explaining what this script does (it seems like you need to replace some output part, but why?)

tests/thermal_energy_density_error_message.sh
```diff
@@ -0,0 +1,18 @@
...
+
+$filename=$ARGV[0];
+my $to_find1="An error occurred in line <";
+my $to_find2="> of file ";
+my $to_find3="/source/numerics/derivative_approximation.cc> in function";
+my $to_replace1="An error occurred in line <";
+my $to_replace2="> of file <DEALII_DIR/source/numerics/derivative_approximation.cc> in function\n(line in output replaced by thermal_energy_density.sh script)";
```
gassmoeller:
```suggestion
my $to_replace2="> of file <DEAL_II_DIR/source/numerics/derivative_approximation.cc> in function\n(line in output replaced by thermal_energy_density.pl script)";
```

tests/thermal_energy_density_error_message.prm
```diff
@@ -0,0 +1,173 @@
+
+# Failing test for ascii data that checks the error message
```
gassmoeller: Update the test description.

tests/thermal_energy_density_error_message.sh
MFraters: Oke, I will add a comment explaining what the script does. 

The normal output contains the whole path starting at root, which will be different for every testing machine. The other lines are just taken from the default script which is used for every test output. The tester succeeds locally for me, but I have to find out why it doesn't work with the online testers.

I could change the extension to pl, but then we would need to change it for all scripts (since they all use the perl) and change the `CMakeLists.txt` in the text directory:

https://github.com/geodynamics/aspect/blob/9faf7ec832a1ab002b348f65e08b275e1ec2bb95/tests/CMakeLists.txt#L245-L252

MFraters:
I have changed the script a bit so that it is simpler and just removes everything which is not relevant but can break the test if run on a different computer. 

The extension `.sh` is set by the `tests/CMakeLists.txt`, so changing it to `.pl` is I think beyond the scope of this pull request.

@gassmoeller I think that I have addressed all your comments.

tjhei:
@MFraters are you sure this is because the temperature is constant? The error you describe is explained in https://www.dealii.org/developer/doxygen/deal.II/namespaceDerivativeApproximation.html with the comment that this is related to "very coarse grids".
How does your mesh look like? From the test it looks like that you have 5 cells after coarsening (a single row?). Maybe we should just return 0 if you have an active cell with a single neighbor instead?

MFraters:
@tjhei That is a good point. I had looked at the description of the function, but not at the overhead description. The error message indeed looks very similar.

I did some more testing to be sure and when I set the temperature to non-constant in the domain the error goes away, but when I increase the resolution the error stays (I upped it from global 1 to 2 and 3). So I don't think that in this case the problem is resolution. But we could include it as a possible problem in the error message. 

Are you proposing to set the indicators to zero and not fail in the catch?

tjhei:
Do you still see a sequence of coarsening steps before it fails? Can you give me a prm? I will think about how to avoid the problem then.

MFraters:
The prm I used is the one included in this pull request. I will post here the differences I made, and the results it gave:

basis case (no changes):
```

# Failing test for the thermal energy density plugin that checks the error message.
#
# EXPECT FAILURE

# A description of convection in a 3d box. See the manual for more information.
set CFL number = 0.1

# At the top, we define the number of space dimensions we would like to
# work in:
set Dimension                              = 3

# There are several global variables that have to do with what
# time system we want to work in and what the end time is. We
# also designate an output directory.
set Use years in output instead of seconds = false
set End time                               = 5.0

# Then there are variables that describe how the pressure should
# be normalized. Here, we choose a zero average pressure
# at the surface of the domain (for the current geometry, the
# surface is defined as the top boundary).
set Pressure normalization                 = surface
set Surface pressure                       = 0


# Then come a number of sections that deal with the setup
# of the problem to solve. The first one deals with the
# geometry of the domain within which we want to solve.
# The sections that follow all have the same basic setup
# where we select the name of a particular model (here,
# the box geometry) and then, in a further subsection,
# set the parameters that are specific to this particular
# model.
subsection Geometry model
  set Model name = box

  subsection Box
    set X extent = 1
    set Y extent = 5
    set Z extent = 1
    set Y repetitions = 5
  end
end


# The next section deals with the initial conditions for the
# temperature (there are no initial conditions for the
# velocity variable since the velocity is assumed to always
# be in a static equilibrium with the temperature field).
# There are a number of models with the 'function' model
# a generic one that allows us to enter the actual initial
# conditions in the form of a formula that can contain
# constants. We choose a linear temperature profile that
# matches the boundary conditions defined below plus
# a small perturbation:
subsection Initial temperature model
  set Model name = function

  subsection Function
    set Variable names      = x,y,z
    set Function constants  = p=0.01, L=1, pi=3.1415926536, k=1
    set Function expression = 1 #(1.0-z) - p*cos(k*pi*x/L)*sin(pi*z)*y^3
  end
end


# Then follows a section that describes the boundary conditions for the
# temperature. The model we choose is called 'box' and allows to set a constant
# temperature on each of the four sides of the box geometry. In our case, we
# choose something that is heated from below and cooled from above. 
# All other parts of the boundary are insulated (i.e., no heat flux through
# these boundaries; this is also often used to specify symmetry boundaries).
subsection Boundary temperature model
  set List of model names = box
  set Fixed temperature boundary indicators   = bottom, top

  subsection Box
    set Bottom temperature = 1
    set Left temperature   = 1
    set Right temperature  = 1
    set Top temperature    = 1
    set Front temperature  = 1
    set Back temperature   = 1
  end
end


# The next parameters then describe on which parts of the
# boundary we prescribe a zero or nonzero velocity and
# on which parts the flow is allowed to be tangential.
# Here, all six sides of the box allow tangential
# unrestricted flow but with a zero normal component:
subsection Boundary velocity model
#  set Tangential velocity boundary indicators = left, right, front, back, bottom, top
  set Prescribed velocity boundary indicators = left: function, right: function, front: function, back: function, bottom: function, top: function # velocity bc
end


subsection Boundary velocity model
  subsection Function
    set Function constants = n = 5
    set Variable names = x,y,z
    # For velocity boundary conditions both are used, for pressure boundary conditions only the first (x) component 
    set Function expression = 0;1-z;0;  
  end
end


# The following two sections describe first the
# direction (vertical) and magnitude of gravity and the
# material model (i.e., density, viscosity, etc). We have
# discussed the settings used here in the introduction to
# this cookbook in the manual already.
subsection Gravity model
  set Model name = vertical

  subsection Vertical
    set Magnitude = 1e16   # = Ra / Thermal expansion coefficient
  end
end


subsection Material model
  set Model name = simple # default:

  subsection Simple model
    set Reference density             = 1
    set Reference specific heat       = 1
    set Reference temperature         = 1
    set Thermal conductivity          = 1
    set Thermal expansion coefficient = 1e-10
    set Viscosity                     = 1
  end
end


# The settings above all pertain to the description of the
# continuous partial differential equations we want to solve.
# The following section deals with the discretization of
# this problem, namely the kind of mesh we want to compute
# on. We here use a globally refined mesh without
# adaptive mesh refinement.
subsection Mesh refinement
  set Initial global refinement                = 1
  set Initial adaptive refinement              = 1
  set Time steps between mesh refinement       = 1
end


# The final part is to specify what ASPECT should do with the
# solution once computed at the end of every time step. The
# process of evaluating the solution is called `postprocessing'
# and we choose to compute velocity and temperature statistics,
# statistics about the heat flux through the boundaries of the
# domain, and to generate graphical output files for later
# visualization. These output files are created every time
# a time step crosses time points separated by 0.01. Given
# our start time (zero) and final time (0.5) this means that
# we will obtain 50 output files.
subsection Postprocess
  set List of postprocessors = velocity statistics, temperature statistics

  subsection Visualization
    set Time between graphical output = 0.025
    set List of output variables = density, viscosity, strain rate
  end
end

subsection Solver parameters
  set Temperature solver tolerance = 1e-10
end
```

run `./aspect thermal_energy_density_error_message.prm`
```
--     . version 2.2.0-pre (improve_mesh_ref_thermal_desn_grad_error_message, 9656bd2bc)
--     . using deal.II 9.0.1
--     . using p4est 2.0.0

-- For information on how to cite ASPECT, see:
--   https://aspect.geodynamics.org/citing.html?ver=2.2.0-pre&sha=9656bd2bc&src=code
Number of active cells: 40 (on 2 levels)
Number of degrees of freedom: 2,199 (1,575+99+525)

*** Timestep 0:  t=0 seconds
   Solving temperature system... 0 iterations.
   Solving Stokes system... 31+0 iterations.

Number of active cells: 5 (on 1 levels)
Number of degrees of freedom: 420 (297+24+99)

*** Timestep 0:  t=0 seconds
   Solving temperature system... 0 iterations.
   Solving Stokes system... 7+0 iterations.

   Postprocessing:
     RMS, max velocity:       0.578 m/s, 0.901 m/s
     Temperature min/avg/max: 1 K, 1 K, 1 K



Exception on MPI process <0> while running mesh refinement plugin <thermal energy density>: 
Mesh refinement plugin thermal energy density could failed to find a gradient. This could be caused by having a constant temperature in the model. If this is the case, please set an other mesh refinement strategy. DEAL.II error:
An error occurred in line <912> of file </home/fraters/Documents/PhD/code/deal.ii/2019-04-19-dealii-candi/tmp/unpack/deal.II-v9.0.1/source/numerics/derivative_approximation.cc> in function
    void dealii::DerivativeApproximation::internal::approximate_cell(const dealii::Mapping<dim, spacedim>&, const DoFHandlerType<dim, spacedim>&, const InputVector&, unsigned int, const dealii::TriaActiveIterator<dealii::DoFCellAccessor<DoFHandlerType<dim, spacedim>, false> >&, typename DerivativeDescription::Derivative&) [with DerivativeDescription = dealii::DerivativeApproximation::internal::Gradient<3>; int dim = 3; DoFHandlerType = dealii::DoFHandler; InputVector = dealii::TrilinosWrappers::MPI::BlockVector; int spacedim = 3; typename DerivativeDescription::Derivative = dealii::Tensor<1, 3>]
The violated condition was: 
    determinant(Y) != 0
Additional information: 
    We have encountered a cell on which the number of linearly independent directions that span the matrix Y (discussed in the documentation of the DerivativeApproximation class) is not equal to dim. The matrix Y then is rank deficient and can not be inverted.

Aborting!
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
```

when I run `mpirun -np 6 ./aspect thermal_energy_density_error_message.prm` it actually runs fine till the end.

When I have the following diff:
```
<   set Initial global refinement                = 1
---
>   set Initial global refinement                = 3
```

I get with `./aspect thermal_energy_density_error_message.prm` the following result:

```
--     . version 2.2.0-pre (improve_mesh_ref_thermal_desn_grad_error_message, 9656bd2bc)
--     . using deal.II 9.0.1
--     . using p4est 2.0.0

-- For information on how to cite ASPECT, see:
--   https://aspect.geodynamics.org/citing.html?ver=2.2.0-pre&sha=9656bd2bc&src=code
Number of active cells: 2,560 (on 4 levels)
Number of degrees of freedom: 96,957 (70,227+3,321+23,409)

*** Timestep 0:  t=0 seconds
   Solving temperature system... 0 iterations.
   Solving Stokes system... 83+0 iterations.

Number of active cells: 320 (on 3 levels)
Number of degrees of freedom: 13,809 (9,963+525+3,321)

*** Timestep 0:  t=0 seconds
   Solving temperature system... 0 iterations.
   Solving Stokes system... 79+0 iterations.

   Postprocessing:
     RMS, max velocity:       0.574 m/s, 0.979 m/s
     Temperature min/avg/max: 1 K, 1 K, 1 K

Number of active cells: 40 (on 2 levels)
Number of degrees of freedom: 2,199 (1,575+99+525)

*** Timestep 1:  t=0.0125 seconds
   Solving temperature system... 0 iterations.
   Solving Stokes system... 38+0 iterations.

   Postprocessing:
     RMS, max velocity:       0.577 m/s, 0.957 m/s
     Temperature min/avg/max: 1 K, 1 K, 1 K

Number of active cells: 5 (on 1 levels)
Number of degrees of freedom: 420 (297+24+99)

*** Timestep 2:  t=0.0375 seconds
   Solving temperature system... 0 iterations.
   Solving Stokes system... 13+0 iterations.

   Postprocessing:
     RMS, max velocity:       0.579 m/s, 0.904 m/s
     Temperature min/avg/max: 1 K, 1 K, 1 K



Exception on MPI process <0> while running mesh refinement plugin <thermal energy density>: 
Mesh refinement plugin thermal energy density could failed to find a gradient. This could be caused by having a constant temperature in the model. If this is the case, please set an other mesh refinement strategy. DEAL.II error:
An error occurred in line <912> of file </home/fraters/Documents/PhD/code/deal.ii/2019-04-19-dealii-candi/tmp/unpack/deal.II-v9.0.1/source/numerics/derivative_approximation.cc> in function
    void dealii::DerivativeApproximation::internal::approximate_cell(const dealii::Mapping<dim, spacedim>&, const DoFHandlerType<dim, spacedim>&, const InputVector&, unsigned int, const dealii::TriaActiveIterator<dealii::DoFCellAccessor<DoFHandlerType<dim, spacedim>, false> >&, typename DerivativeDescription::Derivative&) [with DerivativeDescription = dealii::DerivativeApproximation::internal::Gradient<3>; int dim = 3; DoFHandlerType = dealii::DoFHandler; InputVector = dealii::TrilinosWrappers::MPI::BlockVector; int spacedim = 3; typename DerivativeDescription::Derivative = dealii::Tensor<1, 3>]
The violated condition was: 
    determinant(Y) != 0
Additional information: 
    We have encountered a cell on which the number of linearly independent directions that span the matrix Y (discussed in the documentation of the DerivativeApproximation class) is not equal to dim. The matrix Y then is rank deficient and can not be inverted.

Aborting!
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
```

When I have the following diff: 
```
<     set Function expression = 1 #(1.0-z) - p*cos(k*pi*x/L)*sin(pi*z)*y^3
---
>     set Function expression = (1.0-z) - p*cos(k*pi*x/L)*sin(pi*z)*y^3
```
it runs fine until the end.

So looking at it again it seems that if there is no temperature gradient **and** cells are on there coarest (I am assuming the second case reaches the coarsest grid after 2 timesteps) **and** it is run on a single mpi process it will fail (this of course doesn't exclude be other combinations possible where it fails). 

Thinking about it, what may happen is that with no temperature gradient, everything is coarsened, which leads to this error, while if there is a temperature gradient it will never be coarsened enough to reach that state.

tjhei:
> Thinking about it, what may happen is that with no temperature gradient, everything is coarsened, which leads to this error, while if there is a temperature gradient it will never be coarsened enough to reach that state.

Exactly. You end up with 1x5 cells and then the code crashes...

source/mesh_refinement/thermal_energy_density.cc
```diff
@@ -101,11 +101,36 @@ namespace aspect
...
+                    << "The mesh refinement plugin failed to compute a thermal energy density gradient. "
+                    << "This could be caused by having a constant temperature in the model. If this is "
+                    << "the case, please set an other mesh refinement strategy. DEAL.II error:"
+                    << exc.what() << std::endl
+                    << "Aborting!" << std::endl
+                    << "----------------------------------------------------"
+                    << std::endl;
+
+          // terminate the program!
+          MPI_Abort (MPI_COMM_WORLD, 1);
```
bangerth: I would prefer if we handled aborting the program in one central place. If you take a look at `source/initial_temperature/function.cc` (or any other of the similar classes), we just call `throw` there and let things further up deal with it.

source/mesh_refinement/thermal_energy_density.cc
```diff
@@ -101,11 +101,36 @@ namespace aspect
...
+                    << "----------------------------------------------------"
+                    << std::endl;
+          std::cerr << "Exception on MPI process <"
+                    << Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)
+                    << "> while running mesh refinement plugin <"
+                    << "thermal energy density"
+                    << ">: " << std::endl
+                    << "The mesh refinement plugin failed to compute a thermal energy density gradient. "
+                    << "This could be caused by having a constant temperature in the model. If this is "
+                    << "the case, please set an other mesh refinement strategy. DEAL.II error:"
```
bangerth: Is that really the exception you get? I would think that that might happen because you have too few cells (no neighbors in all coordinate directions) to compute a finite difference approximation. The solution *happens* to be constant if you have constant boundary values, but that's a coincidence.

bangerth:
```suggestion
                    << "the case, please set an other mesh refinement strategy. deal.II error:"
```

tests/thermal_energy_density_error_message/screen-output
```diff
@@ -0,0 +1,40 @@
...
+--------------------------------------------------------
+(line in output replaced by thermal_energy_density.sh script)
+    void dealii::DerivativeApproximation::internal::approximate_cell(const dealii::Mapping<dim, spacedim>&, const DoFHandlerType<dim, spacedim>&, const InputVector&, unsigned int, const dealii::TriaActiveIterator<dealii::DoFCellAccessor<DoFHandlerType<dim, spacedim>, false> >&, typename DerivativeDescription::Derivative&) [with DerivativeDescription = dealii::DerivativeApproximation::internal::Gradient<3>; int dim = 3; DoFHandlerType = dealii::DoFHandler; InputVector = dealii::TrilinosWrappers::MPI::BlockVector; int spacedim = 3; typename DerivativeDescription::Derivative = dealii::Tensor<1, 3>]
+The violated condition was: 
+    determinant(Y) != 0
+Additional information: 
+    We have encountered a cell on which the number of linearly independent directions that span the matrix Y (discussed in the documentation of the DerivativeApproximation class) is not equal to dim. The matrix Y then is rank deficient and can not be inverted.
```
bangerth: That's a different way of saying what I mentioned above.

PR #3230, opened 2019-09-20
suppress warning in release mode in elasticity.cc

tjhei:


gassmoeller:
All Asserts (except AssertThrows) are removed in release mode that is why the variable is not used then and the compiler complains. Looks good to go.

PR #3232, opened 2019-09-23
Make particle properties visualization optional

MFraters:
This pull request allows the user to set exactly what particle fields are included for visualization. The default is set to `all` so all the tests should still pass. I have tested it locally, but will make real tests once the structure has been determined/approved.

This dependent on to pull request #3231. 

* [x] I have read the guidelines in our [CONTRIBUTING.md](../CONTRIBUTING.md) document.

### For all pull requests:


MFraters:
interesting, coveralls has finally appeared with the checks :)

MFraters:
@gassmoeller This should address all the comments you had and solves the initial problem I had. Thanks for the discussions and your help on this! We only need to address the Todo in the description of the new parameter, any idea's for that? I can also add some tests later.

tjhei:
I am coming a bit late to this discussion, but:
1. What is the motivation to make a list of things to exclude instead of a list of properties to output?
2. I don't understand why it is useful to have "all" as one of the strings in the list. Why not only allow the whole string to be "all"?

MFraters:
Hey Timo,

1. The reason for this is that the lpo particle plugin I am writing has a lot of numbered fields per particle. It has sort of the following structure: particle water fraction, grain 0 vf, grain 0 a matrix element 1, ... grain 0 a matrix element 9, grain 1 vf, ...,  grain n a matrix element 9, average a vector 1, average a vector 2, ...., average c vector 3. Now, most of the time I only want to output the average vectors, while the rest is only being used for the computation. My reasoning for going for an exclude list, instead of a include list is that for this case I think it is the easiest way of removing what I do not need, and if I would need the rest, of for example all the vf's, but not the a matrix, I do not have to include them specifically. You are right that I could have inverted this and made a contain search and then include what is found, but than a user would have to know the names of all the particle properties of the plugins they are using, now they can just look up what to exclude for for example the lpo particle property plugin.

2. I could make a check that if there is a `all` in the vector list, the vector list may only be one long, but I do not see why we should enforce this. I can see a case where the user may want to quickly switch between just looking at the particle positions, and a specific set of particle properties and then they just have to add and remove `all` at the end. But I do not feel really strong about this.

source/particle/property/interface.cc
```diff
@@ -97,6 +97,7 @@ namespace aspect
       std::string
       ParticlePropertyInformation::get_field_name_by_index(const unsigned int field_index) const
       {
+        Assert(field_names.size() > field_index, ExcMessage("The amount of field names (" + std::to_string(field_names.size()) + ") is smaller than the requested field index (" + std::to_string(field_index) + ")."));
```
gassmoeller:
```suggestion
        Assert(field_names.size() > field_index, ExcMessage("The number of field names (" + std::to_string(field_names.size()) + ") is smaller than the requested field index (" + std::to_string(field_index) + ")."));
```

source/postprocess/particles.cc
```diff
@@ -38,17 +38,32 @@ namespace aspect
...
       ParticleOutput<dim>::build_patches(const dealii::Particles::ParticleHandler<dim> &particle_handler,
                                          const aspect::Particle::Property::ParticlePropertyInformation &property_information,
+                                         const std::vector<std::string> &exclude_output_properties,
                                          const bool only_group_3d_vectors)
       {
         // First store the names of the data fields
         dataset_names.reserve(property_information.n_components()+1);
         dataset_names.push_back("id");
 
+        // just output all properties
```
gassmoeller: Should this not be: `output all properties that are not excluded`?

source/postprocess/particles.cc
```diff
@@ -651,6 +715,13 @@ namespace aspect
                              "move this file to a network file system. If this variable is "
                              "set to a non-empty string it will be interpreted as a "
                              "temporary storage location.");
+
+          prm.declare_entry ("Exclude output properties", "",
+                             Patterns::Anything(),
+                             "A comma seperated list of strings which exclude all particle"
+                             "property fields which contain these strings. If one of the "
+                             "entries is 'all', only a id will we provided for every point."
```
gassmoeller:
```suggestion
                             "entries is 'all', only a id will be provided for every point."
```

source/postprocess/particles.cc
```diff
@@ -651,6 +715,13 @@ namespace aspect
...
+                             "Todo: do we need a way to generate all posibilites?");
```
gassmoeller: It would be nice to document which options are available, but I do not see an easy way to do it. So do not worry about it for now.

source/postprocess/particles.cc
```diff
@@ -66,18 +81,31 @@ namespace aspect
...
         for (unsigned int field_index = 0; field_index < property_information.n_fields(); ++field_index)
           {
             const unsigned n_components = property_information.get_components_by_field_index(field_index);
 
+            const std::string field_name = property_information.get_field_name_by_index(field_index);
+
+            bool found = false;
+            for (unsigned int i = 0; i < exclude_output_properties.size(); ++i)
+              {
+                if (exclude_output_properties[i] == "all" ||field_name.find(exclude_output_properties[i]) != std::string::npos)
```
gassmoeller:
```suggestion
                if (exclude_output_properties[i] == "all" || field_name.find(exclude_output_properties[i]) != std::string::npos)
```

source/postprocess/particles.cc
```diff
@@ -102,15 +131,49 @@ namespace aspect
...
+
+                unsigned int output_index = 1;
+                unsigned int field_index = 0;
+                unsigned int do_not_increase_for = 0;
+                std::string field_name = "";
+
                 for (unsigned int property_index = 0; property_index < properties.size(); ++property_index)
-                  patches[i].data(property_index+1,0) = properties[property_index];
+                  {
+                    if (do_not_increase_for > 0)
```
gassmoeller: This may work, but I find the logic very hard to follow, keeping track of all of these different indices is error prone, and doing the string searches of which properties to exclude twice also seems unnecessary. What about the following modification: You could create a `std::vector<unsigned int> property_index_to_output_index`, and fill it in the first loop above (lines 96-101). Each entry in the vector is either 0 (if the property should not be written into `patches`) or an index that replaces `output_index` below in line 174. That way you only need to fill this vector once and can use it without additional search when copying the properties. The `if` in line 172 would then be replaced by `if (property_index_to_output_index[property_index] != 0) then copy properties[property_index] into patches[property_index_to_output_index[property_index]]; else do nothing`. You might need to set several entries in the vector to 0 if one excluded field has several components, but that should be fine. Do you think that would work and be better?

MFraters:
Thanks for the review Rene. It took me a while because the method you proposed was a bit more complicated than expected. I don't think it is more (or much less) complicated than the other method and it saves two loops which do string comparisons, so I think this is probably the better way to do it.

I also added two to test excluding one and excluding all particle properties.

gassmoeller:
Hi Menno, sorry for messing with your PR, but when I read through your new version I noticed I could simplify some of the pre-existing loops. Before asking you to do more work, I sat down and did it myself (because I was not sure it would work). Could you take a look at the new version I created and see if it fits for your purpose? I think it is much easier to read and it passes all the tests, so hopefully everything works.

tjhei:
> Could you take a look at the new version I created and see if it fits for your purpose? I think it is much easier to read and it passes all the tests, so hopefully everything works.

Looks good and clean to me. 
Quick question: is this performance sensitive? You are making a bunch of string comparisons for each property on each patch, right? I guess one could compute the indices of all properties ahead of time...

gassmoeller:
> Quick question: is this performance sensitive? You are making a bunch of string comparisons for each property on each patch, right? I guess one could compute the indices of all properties ahead of time...

Not really, it is actually only done once per timestep. Initially, I thought it was done once per cell, and had the initialize function half written until I realized `build_patches` is actually only called once per timestep. It builds the patches for all local particles in one call, so the parsing of the output properties is also only done once per timestep, which should be negligible.

MFraters:
@gassmoeller Yes, this simplifies the code a bit, and since the new tests still pass it should be alright. Thanks for doing that work and extending the inline documentation!

gassmoeller:
So it sounds like I am allowed to merge? If you want further changes I will address them in a follow-up.

PR #3233, opened 2019-09-25
Prescribe temperature field

gassmoeller:
As the changelog entry says this adds the option to 'prescribe' the temperature from the material model instead of solving the temperature equation. I have a model with a different way of computing temperature, but would still like to solve for compositions that is why I need to individually switch off temperature. There might be some test failures, because the output of `PrescribedFieldOutputs` now contains an additional entry for temperature, but no solution should have changed. I added a test case modelled after the existing one for composition.


source/simulator/parameters.cc
```diff
@@ -1265,6 +1287,18 @@ namespace aspect
         AssertThrow (false, ExcNotImplemented());
     }
 
+    {
```
jdannberg: Can you put this into a separate subsection ("Temperature field"?), and then move it to above or below where the same is done for composition?

source/simulator/solver_schemes.cc
```diff
@@ -188,7 +211,7 @@ namespace aspect
               // outputs into the prescribed field before we assemle and solve the equation
```
jdannberg:
```suggestion
              // outputs into the prescribed field before we assemble and solve the equation
```

tests/prescribed_field_temperature.prm
```diff
@@ -0,0 +1,86 @@
+# A testcase that demonstrates that interpolating material model
+# outputs into compositional fields works.
```
jdannberg:
```suggestion
# outputs into the temperature field works.
```

include/aspect/simulator.h
```diff
@@ -1255,23 +1255,23 @@ namespace aspect
 
 
       /**
-       * Interpolate material model outputs onto a compositional field. For the field
-       * whose index is given in the @p compositional_index, this function
+       * Interpolate material model outputs onto a advection field (temperature
+       * or composition). For the field indicated by @p adv_field, this function
```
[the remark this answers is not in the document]
gassmoeller: But `adv_field` is not an index, it is a struct of type AdvectionField that contains much more than the index. I changed it to: `For the field identified by the AdvectionField @p adv_field,`

gassmoeller:
I addressed all comments. Lets see what the tester says.

PR #3234, opened 2019-09-26
Fix the logic for checking the parameter file order of 3D strain compositional fields 

naliboff:
This PR adds a test and fixes the logic for determining how compositional fields associated with a 3D strain tensor should be listed in the parameter file. Thanks to @MFraters for pointing out the issue!

### For all pull requests:


source/material_model/rheology/strain_dependent.cc
```diff
@@ -199,20 +199,23 @@ namespace aspect
...
                 const unsigned int n_s31 = this->introspection().compositional_index_for_name("s31");
                 const unsigned int n_s32 = this->introspection().compositional_index_for_name("s32");
                 const unsigned int n_s33 = this->introspection().compositional_index_for_name("s33");
-                AssertThrow(n_s23 > n_s13 && n_s31 > n_s23 && n_s32 > n_s31 && n_s33 > n_s32,
+                AssertThrow(n_s12>n_s11 && n_s13>n_s12 && n_s21>n_s13 && n_s22>n_s21 && n_s23>n_s22 && n_s31>n_s23 && n_s32>n_s31 && n_s33>n_s32,
                             ExcMessage("Material model visco_plastic with strain weakening using the full strain tensor only works if there "
                                        "are compositional fields called sij, with i=1,..,dim and j=1,...,dim in the order s11,s12,s21 etc."));
-                AssertThrow(n_s33 == n_s13+s-1, ExcMessage("The strain tensor components should be represented by consecutive fields."));
+                AssertThrow(n_s33 == n_s11+s-1, ExcMessage("The strain tensor components should be represented by consecutive fields."));
```
tjhei: why are you checking this only in 3d?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -199,20 +199,23 @@ namespace aspect
             const unsigned int n_s12 = this->introspection().compositional_index_for_name("s12");
             const unsigned int n_s21 = this->introspection().compositional_index_for_name("s21");
             const unsigned int n_s22 = this->introspection().compositional_index_for_name("s22");
-            AssertThrow(n_s12>n_s11 && n_s21>n_s12 && n_s22 > n_s21,
-                        ExcMessage("Material model visco_plastic with strain weakening using the full strain tensor only works if there "
-                                   "are compositional fields called sij, with i=1,..,dim and j=1,...,dim in the order s11,s12,s21 etc."));
+            if (dim==2)
+              {
+                AssertThrow(n_s12>n_s11 && n_s21>n_s12 && n_s22 > n_s21,
```
tjhei: I would find it easier to read if you write ``s11 < s12 && s12 < s21 ...``

source/material_model/rheology/strain_dependent.cc
```diff
@@ -199,20 +199,23 @@ namespace aspect
...
+                            ExcMessage("Material model visco_plastic with strain weakening using the full strain tensor only works if there "
+                                       "are compositional fields called sij, with i=1,..,dim and j=1,...,dim in the order s11,s12,s21 etc."));
```
MFraters: Could you update this error message with the actual order you require in 2d in a way users can just copy and paste? Same for 3d.

source/material_model/rheology/strain_dependent.cc
naliboff: agree :+1:

source/material_model/rheology/strain_dependent.cc
naliboff: Yes, this should also be present for the 2D case, but I think this check is actually not needed ... the assert above will be thrown if the total number is off, the naming is wrong, or the ordering is off.

source/material_model/rheology/strain_dependent.cc
naliboff: How does the new error message look?

source/material_model/rheology/strain_dependent.cc
MFraters: I thought about this as well, but I came to the conclusion that it catches the case `s11, mantle, s12, s13, s21, s22, s23, s31, s32, s33`, which I believe is not caught by the above assertion.

source/material_model/rheology/strain_dependent.cc
MFraters: Looks good to me, thanks!

source/material_model/rheology/strain_dependent.cc
naliboff: Yep, makes sense :+1: Now present for 2D and 3D cases.

naliboff:
@MFraters - last three commits are squashed and tests are passing. The error with the CIG tester appears to be related to the tester coverage.

MFraters:
hmm, yes, we can ignore that for now. Not sure why it fails. Maybe we need to put the coverage in a separate stage where it can't fail the tester.

I am fine with this being merged. @tjhei, are you oke with the changes?

PR #3240, opened 2019-10-03
matrix-free: use AssertThrow

tjhei:
Also check preconditions coming from prms for matrix-free in release
mode by using AssertThrow instead of Assert.


tjhei:
@tcclevenger FYI

PR #3246, opened 2019-10-05
GMG: allow no cheap iterations

tjhei:
The code in the Stokes solver was missing the throw to skip the cheap
solver if so desired by the user. As a consequence, we were always doing
1 cheap solver step. Fix this and add a test.


tjhei:
@tcclevenger FYI. Can you review please?

tcclevenger:
Yes, this looks good to me!

PR #3247, opened 2019-10-05
implement higher order Stokes GMG

tjhei:
First working version of higher order FEM support for the geometric multigrid solver. This requires introducing a base class for the StokesMatrixFreeHandler to be able to have the degree be a compile time template argument.

This PR works, but needs some cleanup (ordering of functions in .cc, possibly different class names, etc.) and tests. I can confirm that we get identical errors for the burstedde benchmark for Q3 between matrix-free and matrix-based.

@tcclevenger can you please take a look?

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1914,7 +1914,7 @@ namespace aspect
       friend class MeshDeformation::MeshDeformationHandler<dim>;   // MeshDeformationHandler needs access to the internals of the Simulator
       friend class VolumeOfFluidHandler<dim>; // VolumeOfFluidHandler needs access to the internals of the Simulator
       friend class StokesMatrixFreeHandler<dim>;
-      template <int dimi, int velocity_degree>
+      template <int dimension, int velocity_degree>
       friend class StokesMatrixFreeHandlerImplementation;
       friend struct Parameters<dim>;
   };
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -97,6 +97,9 @@ namespace aspect
   template <int dim>
   class StokesMatrixFreeHandler;
 
+  template <int dim, int velocity_degree>
+  class StokesMatrixFreeHandlerImplementation;
+
   namespace MeshDeformation
   {
     template <int dim>
@@ -1912,7 +1915,7 @@ namespace aspect
       friend class VolumeOfFluidHandler<dim>; // VolumeOfFluidHandler needs access to the internals of the Simulator
       friend class StokesMatrixFreeHandler<dim>;
       template <int dimi, int velocity_degree>
-      friend class StokesMatrixFreeHandlerImpl;
+      friend class StokesMatrixFreeHandlerImplementation;
       friend struct Parameters<dim>;
   };
 }
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1911,6 +1911,8 @@ namespace aspect
       friend class MeshDeformation::MeshDeformationHandler<dim>;   // MeshDeformationHandler needs access to the internals of the Simulator
       friend class VolumeOfFluidHandler<dim>; // VolumeOfFluidHandler needs access to the internals of the Simulator
       friend class StokesMatrixFreeHandler<dim>;
+      template <int dimi, int velocity_degree>
+      friend class StokesMatrixFreeHandlerImpl;
       friend struct Parameters<dim>;
   };
 }
--- a/include/aspect/stokes_matrix_free.h
+++ b/include/aspect/stokes_matrix_free.h
@@ -306,23 +306,10 @@ namespace aspect
       virtual void setup_dofs()=0;
 
       /**
-       * Evalute the MaterialModel to query for the viscosity on the active cells,
-       * project this viscosity to the multigrid hierarchy, and cache the information
-       * for later usage. Also sets pressure scaling and information regarding the
-       * compressiblity of the flow.
+       * Evaluate the material model and update internal data structures before the
+       * actual solve().
        */
-      virtual void evaluate_material_model()=0;
-
-      /**
-       * Get the workload imbalance of the distribution
-       * of the level hierarchy.
-       */
-      virtual double get_workload_imbalance()=0;
-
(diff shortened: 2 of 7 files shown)


include/aspect/stokes_matrix_free.h
```diff
@@ -274,11 +274,72 @@ namespace aspect
...
+   * setup, assembly and solving the Stokes system.
+   *
+   * We need to derive from StokesMatrixFreeHandler to be able to introduce a
+   * second template argument for the degree of the Stokes finite
+   * element. This way, the main simulator does not need to know about the
+   * degree and we can pick the desired class to use at runtime.
+   */
+  template<int dim, int velocity_degree>
+  class StokesMatrixFreeHandlerImpl: public StokesMatrixFreeHandler<dim>
   {
```
tcclevenger: To me `StokesMatrixFreeHandlerImpl` looks like `StokesMatrixFreeHandlerImp1`. Is ` StokesMatrixFreeHandlerImplemntation` too long? Maybe we could remove the word "Handler"?  Or maybe we call the base class `StokesMatrixFreeHandlerBase`?

tjhei: > Is StokesMatrixFreeHandlerImplemntation` too long?

no, I think this would be fine to use.

tjhei:
Okay, cleanup up the code, added a test, and changed the class name...

tcclevenger:
Looks good to me!

include/aspect/simulator.h
```diff
@@ -1911,6 +1914,8 @@ namespace aspect
       friend class MeshDeformation::MeshDeformationHandler<dim>;   // MeshDeformationHandler needs access to the internals of the Simulator
       friend class VolumeOfFluidHandler<dim>; // VolumeOfFluidHandler needs access to the internals of the Simulator
       friend class StokesMatrixFreeHandler<dim>;
+      template <int dimi, int velocity_degree>
```
gassmoeller:
```suggestion
      template <int dim, int velocity_degree>
```

tjhei:
> But these member variables do not even use the template in their class declarations, they are just used in the implementation of some functions. So would it not make more sense to remove the templates from these member variables?

This is a good point and I didn't question if the template argument is needed for StokesOperator. We do need to have the degree as a template argument at some place, as the matrix-free framework needs to know the degree at compile time.


> So would it not make more sense to remove the templates from these member variables? They could still get the degrees as arguments for the functions that need them, or as constructor arguments that are stored in a private const variable

A const variable is not enough as they will eventually need to be template arguments. I don't think we can make them template member functions as we need to take the address of local_apply for example. @tcclevenger ?

tcclevenger:
> > But these member variables do not even use the template in their class declarations, they are just used in the implementation of some functions. So would it not make more sense to remove the templates from these member variables?
> 
> This is a good point and I didn't question if the template argument is needed for StokesOperator. We do need to have the degree as a template argument at some place, as the matrix-free framework needs to know the degree at compile time.
> 
> > So would it not make more sense to remove the templates from these member variables? They could still get the degrees as arguments for the functions that need them, or as constructor arguments that are stored in a private const variable
> 
> A const variable is not enough as they will eventually need to be template arguments. I don't think we can make them template member functions as we need to take the address of local_apply for example. @tcclevenger ?

The `StokesOperator` will need velocity_degree as a template argument for several function implementation, but does not need it in the class declaration. But won't some of these functions (like `apply_add`)  be called outside of our code and so the matrix-free object will have been required to be templated by the class objects? I also think Timo is correct that we cannot make templated member functions for matrix-free class.

tjhei:
(I will rebase after #3251 is merged)

gassmoeller:
> The StokesOperator will need velocity_degree as a template argument for several function implementation, but does not need it in the class declaration. But won't some of these functions (like apply_add) be called outside of our code and so the matrix-free object will have been required to be templated by the class objects? I also think Timo is correct that we cannot make templated member functions for matrix-free class.

But I do not see why this requires them to be template parameters. The StokesOperator has other private variables (like the pressure_scaling) that are set at some point and then used whenever functions are called. Why not for example have a private variable velocity_degree in the `StokesOperator` class that is set in its constructor? Its current (default) constructor is implicitly called by the constructor of StokesMatrixFreeHandler when it creates the `stokes_matrix` member variable, so we could easily make a custom constructor that stores the velocity degree (it would be called from here: https://github.com/geodynamics/aspect/blob/master/source/simulator/stokes_matrix_free.cc#L850). The new variable can even be `const`. And analogous for the `MassMatrixOperator` and `ABlockOperator` classes? I am not saying we do not need the information, I am just saying I do not see why we need them as templates and having them as private members would be much easier to understand.

tjhei:
> But I do not see why this requires them to be template parameters. The StokesOperator has other private variables (like the pressure_scaling) that are set at some point and then used whenever functions are called. Why not for example have a private variable velocity_degree in the `StokesOperator` class that is set in its constructor?

Eventually, degree needs to be a template argument so it needs to be known at compile time, which a ``const int`` can not do.

See here for an example: https://godbolt.org/z/TBAtB6

At some place, I need something like:
```
switch (degree):
case 1:
    do_something<1>();
case 2:
    do_something<2>();
...
```
to convert from a runtime parameter to a template argument. I can move it inside one of the functions though (which I did not think about earlier). See the last example in the link above. Problem: This needs to be done for every function that uses the template argument right now(3x fill_cell_data,  local_apply, local_compute_diagonal, ...). Which is more messy and likely less performant (not sure by how much).

gassmoeller:
Ah ok, I see your point, I leave this decision up to you then. But at least fix the typo I commented on above :wink:

include/aspect/simulator.h
tjhei: that doesn't work:
``
../include/aspect/simulator.h:145:13: error:  shadows template parm ‘int dim’
``

tjhei: but I changed it to ``dimension``

gassmoeller: Ok then. Do you think the current state is ready?

tjhei:
This ready to go from my side.

PR #3249, opened 2019-10-06
use c++ override instead of virtual

tjhei:
- replace virtual with override
- enable clang-tidy rule
- rationale is here:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -26,7 +26,8 @@
 
 
 # enabled checks:
-Checks: '-*,modernize-use-emplace,modernize-make-shared,mpi-*,modernize-use-nullptr,modernize-replace-autoptr,modernize-make-unique,performance-*,-performance-inefficient-string-concatenation'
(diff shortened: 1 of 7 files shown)


tjhei:
> (when should I use `virtual` when should I use `override`?)

they basically never need virtual (only interfaces have virtual).

> Related question: You only modify to `override` in a couple of places, but should not every plugin now use `override` instead of `virtual` for almost every function they implement? Did you leave this for a future PR?

This is a good point. I actually didn't modify anything, I just ran the clang-tidy check and applied the fixes automatically. ;-) I have to admit that I did not check why only a small number of places were reported.

gassmoeller:
> they basically never need virtual (only interfaces have virtual).

Yes, I was more worried about the transition than about the end state.

> This is a good point. I actually didn't modify anything, I just ran the clang-tidy check and applied the fixes automatically. ;-) I have to admit that I did not check why only a small number of places were reported.

Hm, mysterious, do you have an idea why that is? I created issue #3258 to remind us, so we could go forward with this and postpone the other plugins to a later time.

tjhei:
> Hm, mysterious, do you have an idea why that is?

Figured it out:
For most cases I am only getting a warning because clang-tidy doesn't know if final or override is the right thing to do.

We can merge this and I will incrementally do the other steps.

PR #3257, opened 2019-10-10
add drucker prager rheology module

naliboff:
This PR adds a new model in the `Rheology` namespace for Drucker Prager plasticity, which will replace the similar functionality in `MaterialUtilities`.

This is currently work in progress and I wanted to let @anne-glerum take a look before proceeding further. 

Remaining tasks and topics for discussion:
- Use the new functions in the `drucker_prager` and `viscoelastic_plastic` material models.
- Remove the related functionality in MaterialUtilities
- Consistently use "yield stress" or "yield strength"
- Rename various functions to be more intuitive/descriptive
- Simplify existing documentation and make code more readable

### DIFF
--- a/include/aspect/material_model/rheology/drucker_prager.h
+++ b/include/aspect/material_model/rheology/drucker_prager.h
@@ -1,0 +1,125 @@
+/*
+  Copyright (C) 2019 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _aspect_material_model_rheology_drucker_prager_h
+#define _aspect_material_model_rheology_drucker_prager_h
+
+#include <aspect/global.h>
+#include <aspect/material_model/interface.h>
+#include <aspect/simulator_access.h>
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    namespace Rheology
+    {
+      template <int dim>
+      class DruckerPrager : public ::aspect::SimulatorAccess<dim>
+      {
+        public:
+          /**
+           * Constructor.
+           */
+          DruckerPrager();
+
+          /**
+           * Declare the parameters this function takes through input files.
+           */
+          static
+          void
+          declare_parameters (ParameterHandler &prm);
+
+          /**
+           * Read the parameters from the parameter file.
+           */
+          void
+          parse_parameters (ParameterHandler &prm);
+
+          /**
+           * Compute the plastic yield stress based on the Drucker Prager yield criterion.
+           */
+          double
+          compute_stress (const double cohesion,
+                          const double angle_internal_friction,
+                          const double pressure) const;
+
+          /**
+           * Compute the plastic viscosity with the yield stress and effective strain rate.
+           */
+          double
+          compute_viscosity (const double cohesion,
+                             const double angle_internal_friction,
+                             const double pressure,
+                             const double effective_strain_rate) const;
+
+          /**
+           * Compute the derivative of the plastic viscosity with respect to pressure.
+           */
+          double
+          compute_derivative (const double angle_internal_friction,
+                              const double effective_strain_rate) const;
+
+
+          /**
+           * Return the value of the maximum yield stress for each composition used in
+           * the rheology model.
+           */
+          double
+          get_max_yield_strength () const;
+
+          /**
+           * Return the values of the cohesions for each composition used in the
+           * rheology model.
+           */
+          std::vector<double>
+          get_cohesions () const;
+
+          /**
+           * Return the values of the angles of internal friction for each
+           * composition used in the rheology model.
+           */
+          std::vector<double>
+          get_angles_internal_friction () const;
+
+        private:
+
+          /**
+           * List of internal friction angles (phi).
+           */
+          std::vector<double> angles_internal_friction;
+
+          /**
+           * List of the cohesions.
+           */
+          std::vector<double> cohesions;
+
+          /**
+           * Limit maximum yield stress from drucker prager yield criterion.
+           */
+          double max_yield_strength;
+
+      };
+    }
+  }
+}
+#endif
--- a/include/aspect/material_model/rheology/drucker_prager.h
+++ b/include/aspect/material_model/rheology/drucker_prager.h
@@ -23,7 +23,6 @@
 
 #include <aspect/global.h>
 #include <aspect/material_model/interface.h>
-#include <aspect/simulator_access.h>
 
 namespace aspect
 {
@@ -55,7 +54,7 @@ namespace aspect
       };
 
       template <int dim>
-      class DruckerPrager : public ::aspect::SimulatorAccess<dim>
+      class DruckerPrager
       {
         public:
           /**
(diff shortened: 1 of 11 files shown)


anne-glerum:
Hi @naliboff , thanks for this! I like the new structure. Concerning your discussions points:

- I'd go with `yield_stress`. This goes with most of the other function and variable names, like `viscous_stress` and `compute_stress`. Also, I associate stress sooner with the symbol sigma that we use in the comments than `strength` I think. But I'm also fine with `strength` :)
- Instead of `compute_stress` I'd name the function `compute_yield_stress`. 

Looking through it, I thought it might also be nice to have the DruckerPrager rheology store the updated cohesions and friction angles, instead of recomputing them to fill the AdditionalOutputs using the StrainDependent rheology, but maybe that's too much.

naliboff:
@anne-glerum - Thanks for looking this over! I agree that `yield_stress` fits in better with the other function and variable names. 

I like the idea of outputting the modified cohesion and friction angles through the DruckerPrager rheology. Eventually, we could change the visco plastic AdditionalOutputs variable from `is_yielding` to a specific deformation mechanism name (Plastic, Diffusion, Composite, etc).

naliboff:
The test `drucker_prager_derivatives` is failing and it appears to be to related to the the parsing of parameters in the `DruckerPrager` rheology module (see errors below). 

Looking at `tests/drucker_prager_derivatives.cc`, it's not clear to me what the issue is as the `EquationOfState` parameters are also parsed in the `drucker_prager` material model in a similar fashion.

Suggestions for what might be going on? I'll continue digging.

from GDB:

> [ltt:12081] [ 1] /data1/Software/aspect/drucker_prager_rheology/aspect/debug/aspect(_ZNK6aspect15SimulatorAccessILi2EE22n_compositional_fieldsEv+0x4)[0x55798b907da2]
> [ltt:12081] [ 2] /data1/Software/aspect/drucker_prager_rheology/aspect/debug/aspect(_ZN6aspect13MaterialModel8Rheology13DruckerPragerILi2EE16parse_parametersERN6dealii16ParameterHandlerE+0x32)[0x55798b4b5a02]
> [ltt:12081] [ 3] /data1/Software/aspect/drucker_prager_rheology/aspect/debug/aspect(_ZN6aspect13MaterialModel13DruckerPragerILi2EE16parse_parametersERN6dealii16ParameterHandlerE+0xe3)[0x55798b41f0dd]
>

source/material_model/drucker_prager.cc
```diff
@@ -224,6 +224,7 @@ namespace aspect
                              "Units: $W/m/K$.");
           prm.enter_subsection ("Viscosity");
           {
+
```
jdannberg: @naliboff, I don't know if this causes the problem, but you want to remove the Angle of internal friction/Cohesion below (and same in parse parameters). These things are now read in the Rheology::DruckerPrager<dim>::declare_parameters(prm) function, right?

naliboff: @jdannberg - Thanks for taking a look! I should have added a comment on Friday after figuring out the solution. 

I actually intentionally left `Angle of internal friction` and `Cohesion` in `drucker_prager` as  using the equivalent parameters in `Rheology::DruckerPrager` would have required adding a decent amount of new code and adjusting tests. This is due to the  parameters in `Rheology::DruckerPrager` being lists, rather than single values.

@gassmoeller and I talked briefly on Friday about this and the issue is due to not having simulator access in the` drucker_prager_derivatives` test, which we need to get the number of compositional fields when parsing parameters. I checked with @MFraters and the correct path forward is to adjust the test (or make new tests) following the `visco_plastic_derivatives_` tests, which have simulator access. This does indeed work and I'll submit the new tests in a separate PR.

include/aspect/material_model/rheology/drucker_prager.h
```diff
@@ -0,0 +1,125 @@
...
+           * Read the parameters from the parameter file.
+           */
+          void
+          parse_parameters (ParameterHandler &prm);
+
+          /**
+           * Compute the plastic yield stress based on the Drucker Prager yield criterion.
+           */
+          double
+          compute_yield_stress (const double cohesion,
```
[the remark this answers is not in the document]
gassmoeller: Ah, I saw below why you are doing this. Still I think this is not a good solution. You pass back the parameters of this class, then modify them outside (using the strain weakening module) and then hand them over again to this function. This breaks the concept of encapsulation though, either these parameters are mostly used inside this class (then they shouldnt be modified outside of it), or they should not be members of this class. I would suggest the following interface instead:
```
double
compute_yield_stress (const double pressure,
                                     const unsigned int composition_index,
                                     const double weakening_cohesion = 0.0,
                                     const double weakening_angle_of_friction = 0.0) const;
```

This way you do not need to pass the parameters out and in again, you just hand over all the information that is needed inside this function. You can modify the default value for the weakening parameters from 0.0 to 1.0 if that fits better with the formula.
Did I understand this correctly?

include/aspect/material_model/rheology/drucker_prager.h
```diff
@@ -0,0 +1,125 @@
...
+                                const double angle_internal_friction,
+                                const double pressure) const;
+
+          /**
+           * Compute the plastic viscosity with the yield stress and effective strain rate.
+           */
+          double
+          compute_viscosity (const double cohesion,
```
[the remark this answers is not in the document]
gassmoeller: Same as above. The inputs should be all inputs for the function above, plus `effective_strain_rate`.

include/aspect/material_model/rheology/drucker_prager.h
```diff
@@ -0,0 +1,125 @@
...
+           */
+          double
+          get_max_yield_stress () const;
+
+          /**
+           * Return the values of the cohesions for each composition used in the
+           * rheology model.
+           */
+          std::vector<double>
+          get_cohesions () const;
```
gassmoeller: If the above comment works, we might not need this function anymore.

include/aspect/material_model/rheology/drucker_prager.h
```diff
@@ -0,0 +1,125 @@
...
+
+          /**
+           * Return the values of the angles of internal friction for each
+           * composition used in the rheology model.
+           */
+          std::vector<double>
+          get_angles_internal_friction () const;
```
gassmoeller: same for this one

naliboff:
@gassmoeller - thanks for the review! A quick recap from our discussion:

1. I agree with the suggestion for improving the interface.
2. We can handle the duplicate parameters in the DruckerPrager material model through the `declare_alias()` function ([link](https://www.dealii.org/current/doxygen/deal.II/classParameterHandler.html#aecb847561585089c4c7fcf51eddece16))
3. For now, we will hold off on moving the ViscoPlastic AdditionalOutputs to the DruckerPrager rheology module and revisit this when we add a second plasticity rheology module (e.g., dynamic friction, etc).

naliboff:
@gassmoeller - It turns out there is an issue with the new proposed interface. In the dynamic_friction material model we calculate the steady-state friction coefficient, which is used to compute the Drucker Prager yield stress. Other material models in the future may want to calculate various permutations of the friction and cohesion angle as well (e.g., beyond simple strain weakening).

With this in mind, I don't think the new proposed interface is a good long-term solution and the best format is to keep with previous practice and have the material model pass the cohesion and internal friction value to the DruckerPrager rheology module.

In order to deal with the encapsulation issue, perhaps the most straightforward path is to move the `Cohesions` and `Angles of internal Friction` variables back to ViscoPlastic for now? As an alternative, we could also make a new rheology module for declaring these variables? For now, the first option seems like the most straightforward one.

naliboff:
:+1: - thanks for the change @gassmoeller, this is a good solution. I made the associated changed to `drucker prager `and `dynamic friction`, and merged those changes into your commit. I think I've now addressed all of your previous comments, nearly all of which were fixed by the last commit. So, likely ready for a final review.

@MFraters - any thoughts on why the drucker_prager_derivatives_2d test is failing, but the 3D case is not?

gassmoeller:
The test was failing because the max_yield_stress for that test was changed. Previously the default was infity (no max yield stress), but the default for the input parameter is 1e12. I reverted that in another commit. I also removed the use of the DruckerPragerParameters in the drucker_prager material model and the dynamic_friction material model, because they both have a different structure for their input parameters. We can later think about how to unify them.

gassmoeller:
When the tester agrees this is ready from my side.

naliboff:
@gassmoeller - yikes, I'm surprised the change in maximum yield stress value affected only 1 out of 4 of the drucker prager tests. Thanks for pushing the other changes and agree that it is best to deal with unification of the interface and parameter declarations later on.

Issue #3258, opened 2019-10-10
Derived classes should use `override` instead of `virtual`


gassmoeller:
The `virtual` keyword should only be used when declaring new virtual functions. Implementations of virtual functions that overwrite the base class function should use `override` instead as explained in #3249. Currently only a small set of ASPECT's plugin use `override`, and nearly all uses of `virtual` should be replaced by `override`.

PR #3259, opened 2019-10-15
use override instead of virtual

tjhei:
part of #3258


tjhei:
... with a few minor, uncontroversial other changes. Look for "(void)" if you want to see a little surprise.

tjhei:
> Interesting that the compiler does not complain, member variables are not allowed to have the type `void`

This is allowed in C++ for backwards-compatibility with C. :-)

PR #3260, opened 2019-10-15
cleanup virtual destructors

tjhei:
- use =default if possible
- empty destructors in derived classes are not needed at all


tjhei:
rebased.

PR #3261, opened 2019-10-16
matrix-free free-slip and no-slip conflict

tjhei:
This fixes a crash caused by confliciting constraints between no-normal
flux and Dirichlet as observed with a quarter_shell.


tjhei:
FYI @tcclevenger

tcclevenger:
Yes, good catch. I had an issue with this a while back on an old branch and the change evidently didn’t make it to the new one.

tjhei:
> Since Conrad agrees and there is a test I am going to merge.

Yes, but the tests don't run as they require 9.2...

PR #3264, opened 2019-10-18
Add option to turn off dynamic topography contribution in geoid calculation

Shangxin-Liu:

As we discussed in Lovelace meeting @LudovicJnnt @cedrict , adding an option to remove dynamic topography contribution in geoid post-processor will enable a straightforward benchmark test between the spectral calculation post-processor and newton integral calculation post-processor for the gravity in 3D spherical shell.

I created another test case based on the previous geoid test prm file to remove dynamic topography contribution to verify this change. The test directly performs the density integral from the input buoyancy field. The output spectral results agree well with the analytic solution determined from input buoyancy field, i.e., reference density (0,0) and (2,0) perturbation components.


include/aspect/postprocess/geoid.h
```diff
@@ -119,6 +119,11 @@ namespace aspect
         double density_above;
         double density_below;
 
+        /**
+         * A parameter to control whether to turn off the dynamic topography contribution on geoid
+         */
+        bool remove_dynamic_topo_contribution;
```
gassmoeller: May I suggest to invert the logic of this parameter? I would prefer to call the parameter `include_dynamic_topo_contribution` with a default value of `true`. The reasoning for this is that below you have a lot of conditions that say `if remove_dynamic_topo_contribution == false`, this suggests you really want to test for `if include_dynamic_topo_contribution == true`. Also my brain hurts every time I try to understand double negations :smile:. You would need to update the descriptions and the if statements below, but I think it will be easier to understand afterwards.

source/postprocess/geoid.cc
```diff
@@ -326,15 +326,25 @@ namespace aspect
...
-      std::pair<double, std::pair<std::vector<double>,std::vector<double> > > SH_surface_dyna_topo_coes = SH_dyna_topo_coes.first;
-      std::pair<double, std::pair<std::vector<double>,std::vector<double> > > SH_CMB_dyna_topo_coes = SH_dyna_topo_coes.second;
-
-      // Get the density contrast at the surface and CMB
-      const double surface_delta_rho =  SH_surface_dyna_topo_coes.first - density_above;
-      const double CMB_delta_rho = density_below - SH_CMB_dyna_topo_coes.first;
+      std::pair<double, std::pair<std::vector<double>,std::vector<double> > > SH_surface_dyna_topo_coes;
+      std::pair<double, std::pair<std::vector<double>,std::vector<double> > > SH_CMB_dyna_topo_coes;
+      // Initialize the surface and CMB density contrasts with NaNs becasue they may be unused in case of no dynamic topography contribution.
+      double surface_delta_rho = std::numeric_limits<double>::quiet_NaN();
```
gassmoeller: Please replace this and the line below by:
```suggestion
      double surface_delta_rho = numbers::signaling_nan<double>();
```
Signaling NaNs ensure that this number is not used by crashing if someone tries to use them.

source/postprocess/geoid.cc
```diff
@@ -489,98 +511,104 @@ namespace aspect
...
+                  surface_dynamic_topo_contribution_SH_coes_file << "surface density contrast(kg/m^3): "
+                                                                 << output_surface_delta_rho.str()
+                                                                 << std::endl;
+                  // write out the data on processor 0
+                  surface_dynamic_topo_contribution_SH_coes_file << output_surface_dynamic_topo_contribution_SH_coes.str();
+                }
+            }
+        }
+
+      if (remove_dynamic_topo_contribution == false)
```
gassmoeller: I think you can remove this `if` and the braces in line 562 above and 565 below. The result should be the same.

tests/geoid_no_topo.prm
```diff
@@ -0,0 +1,104 @@
+# A simple pure math setup to compute geoid in single degree 2 order 0 perturbation density field.
+# The perturbation is added to a constant density, i.e., degree 0 order 0 component.
+# The analytic solution of the denbsity integral contribution of this case should be -0.53059320038
+# for (2,0) component and 14.1570019498 for degree 0 order 0.
```
gassmoeller: Please add a sentence about excluding dynamic topography.

Shangxin-Liu:
Thanks for the suggestions. @gassmoeller   I add a new version here with the minor changes. In addition to the changes above, I also fix one previous spelling typo and added this new parameter to the previous two geoid test prm files. I also find that I should parse this new parameter before the AssertThrow check to avoid the bug of default false for bool variable. This is my previous omission.😄 Let me know if there are more things to improve.

gassmoeller:
Looks good, ready to go!

Issue #3254, opened 2019-10-08
Restart does not take into account initial topography of the mesh.


anne-glerum:
When restarting a model that had initial topography added to the mesh, this topography is not taken into account and thus you'll have a different mesh upon restart. See the figure of a model with an initial hill that is or is not restarted at the second timestep. I have a PR coming up to remedy this that adds the initial topography to the stored mesh displacements.

![image](https://user-images.githubusercontent.com/7631624/66401517-8cc3c900-e9e3-11e9-945d-111ef4a52ab7.png)

gassmoeller:
I think this was fixed in #3265.

PR #3265, opened 2019-10-18
Mesh def initial topo

anne-glerum:
This PR in part addresses #3254. For Box geometries, it stores the initial topography that is added to the mesh nodes after all initial refinement on the same structure as the mesh deformation is stored. After restart, this initial topography is added to these mesh displacements, so that the mesh after restart is the same as before. 

There are some TODOs left in the code, where I think things could be more elegant, but I didn't see a better way to do them. I would appreciate your insight on this. Especially,
1. Where and how to call the set_initial_topography() function. It has to be called _before_ the topography is actually set. 
2. Where to check for the unwanted combination of writing checkpoints + initial topography - mesh deformation, as in this special case the mesh will still be wrong after a restart. Now it's done after deciding whether there is mesh deformation, but then there is no access to the initial_topography_model yet.
3. Once a restart has occurred, we don't care for the initial topography any more, as it was added to the mesh deformations and I set it to zero. It will still be redistributed etc after mesh_refinement. Is there a way to not do this? 

### For all pull requests:

* [x] I have created a testcase


anne-glerum:
PS the tester fails on one test because I changed it.

gassmoeller:
Hi Anne, sorry for not looking into this earlier. How weird. I had a look at it (I saw your PR, but wanted to check the behavior myself first). It is possible that your PR fixes part of the problem, but why is the problem there in the first place? In my understanding this is the order in which things happen:
1. We create the coarse grid
2. We make all initial adaptive refinements and solutions
3. We create initial topography after the final refinement and before calculating the finest initial solution. This is done by actually deforming the vertices of the triangulation (no trickery with the MappingEulerian or similar, so there should be no interaction with the free surface plugin).
4. We store everything into a checkpoint (including the deformed vertex positions).
5. We read everything from the checkpoint (including the deformed vertex positions)

So where does the information about the initial topography get lost? The deformation definitely happens (as we see in the output of timestep 0). Is it not stores in the checkpoint? Is it not properly read when resuming? Does it get overwritten by something else?

anne-glerum:
I see what you mean. I’ll dig into the checkpointing/restarting some more.

anne-glerum:
I wrote the triangulation to vtk right before saving it during checkpointing and right after loading  using GridOut. This shows that before saving it does indeed include the _initial_ topography, but after loading it doesn't. It does not take into account any topography induced by the internal dynamics after t0, as expected.

The dealii documentation states that although vertex locations, indices and connections are stored, signals and pointers to Manifold objects are not. 

Adding the topography is done through a custom `pre_set_initial_state` signal, so that transformation is probably not stored and applied to the mesh upon loading?

gassmoeller:
Ah, I think we found the problem then. Your new solution in your last comment is correct, and I think I finally understand why: It is a problem in our expectation of the `triangulation.save()` function, and a problem of documentation in deal.II. While the save() function of a serial triangulation actually stores the mesh (vertices, faces, and so on), the save() function of a parallel::distributed::Triangulation only stores the refinement information (i.e. which cells to refine how many times, based on the assumption of an already existing coarse mesh). Therefore when restarting ASPECT creates the coarse mesh and refines it appropriately, but does not generate initial topography, because we are not setting the initial condition anymore. Great, this makes the solution much simpler.

> Adding the topography is done through a custom pre_set_initial_state signal, so that transformation is probably not stored and applied to the mesh upon loading. Indeed when I connect the initial_topography function to the signals.post_resume_load_user_data signal, the loaded mesh then has the correct topography. Should I update the PR to this?

Yes, connecting the topography function for all geometry models that support initial topography (is it only box?) to both the `pre_set_initial_state`, and the `post_resume_load_user_data` should fix the problem. Can you also add a changelog that says that initial mesh topography with checkpointing was broken before and is fixed now? Just to let people know.

> For the other geometries, we will have to set Manifold objects again after resuming then.

No we do not. Since we recreate the geometry anyway when restarting we already set up all the manifold connections just like when starting the model. Resuming then only refines the mesh according to the state it had before, and loads all the solutions (and the free surface mesh deformation vectors) onto the mesh.

anne-glerum:
> Yes, connecting the topography function for all geometry models that support initial topography (is it only box?) to both the pre_set_initial_state, and the post_resume_load_user_data should fix the problem. Can you also add a changelog that says that initial mesh topography with checkpointing was broken before and is fixed now? Just to let people know.

Alright, I've reverted the first semi-fix and just connected to the `post_resume_load_user_data` signal. The ellipsoidal and normal chunk don't use the `pre_set_initial_state` signal for the initial topography, so no changes there. Also added a change log.

anne-glerum:
Just to check: The test succeeds on my machine. Do you think the differences in iterations, time step and topography are just do to the different machines?

gassmoeller:
Usually it is just the different library versions, but here the changes in velocity and topography look somewhat large. If you want to be sure you could test if the difference is smaller with deal.II 9.0.1 (that is the one running on the tester) and also test if the general behavior under 9.0.1 looks correct.

anne-glerum:
I just ran the test locally with deal.II 9.0.1 and only a gnuplot solution file fails, the timestep, topography and velocity are the same...

gassmoeller:
Then the only remaining difference is the trilinos version (12.10 on the tester, 12.14 on your machine). It makes sense that this can influence the iteration count and solution slightly. I am a bit surprised by how much in this case, but there is not much point in looking for that here since you tested the functionality of this PR works for both deal.II 9.0 and 9.1. If you update the test results with the ones from the tester I think this is ready.

anne-glerum:
Alright, I've updated the test results, thanks!

PR #3267, opened 2019-10-18
Extend parse function to allow multiple values per key

gassmoeller:
This PR extends Utilities::parse_map_to_double_array to allow multiple entries per key using the syntax: `key1: value1|value2|value3, key2: value1, key3: value1|value2`. We would like to use this syntax to make progress with defining phase transitions for multiple compositional fields (like #2656). The unit tests should cover all cases, so I can make this as a separate PR without actual implementations using it (it will be used by several rheologies, equations of state and material models). 

* [x] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../tests/) directory.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -79,7 +79,7 @@ namespace aspect
      * This function takes a string argument that is interpreted as a map
      * of the form "key1 : value1, key2 : value2, etc", and then parses
      * it to return a vector of these values where the values are ordered
-     * in the same order as a given set of keys. If @p allow_subentries
+     * in the same order as a given set of keys. If @p allow_multiple_values_per_key
      * is set to 'true' it also allows entries of the form
      * "key1: value1|value2|value3, etc", in which case the returned
      * vector will have more entries than the provided
@@ -93,7 +93,8 @@ namespace aspect
      *   where "key1", "key2", ... are exactly the keys provided by the
      *   @p list_of_keys in the same order as provided. In this situation,
      *   if a background field is required, the background value is
-     *   assigned to the first element of the output vector.
+     *   assigned to the first element of the output vector. This form only
+     *   allows a single value per key.
      * - Whether or not a background field is required depends on
      *   the parameter being parsed. Requiring a background field
      *   increases the size of the output vector by 1. For example,
@@ -114,25 +115,36 @@ namespace aspect
      * @param[in] property_name A name that identifies the type of property
      *   that is being parsed by this function and that is used in generating
      *   error messages if the map does not conform to the expected format.
-     * @param [in] allow_subentries If true allow having multiple subentries
-     *   for each key. If false only allow a single entry per key. In either
+     * @param [in] allow_multiple_values_per_key If true allow having multiple values
+     *   for each key. If false only allow a single value per key. In either
      *   case each key is only allowed to appear once.
-     * @param [out] subentry_structure If handed over this
-     *   vector will be filled with as many components as
-     *   keys (+1 if there is a background field). Each value represents
-     *   how many subentries were found for this key. The sum over all
-     *   entries is the length of the returned vector of values.
+     * @param [in,out] n_values_per_key A pointer to a vector of unsigned
+     *   integers. If no pointer is handed over nothing happens. If a pointer
+     *   to an empty vector is handed over, the vector
+     *   is resized with as many components as
+     *   keys (+1 if there is a background field). Each value then stores
+     *   how many values were found for this key. The sum over all
+     *   entries is the length of the return value of this function.
+     *   If a pointer to an existing vector with one or more entries is
+     *   handed over (e.g. a n_values_per_key vector created by a
+     *   previous call to this function) then this vector is used as
+     *   expected structure of the input parameter, and it is checked that
+     *   @p key_value_map fulfills this structure.
      *
      * @return A vector of values that are parsed from the map, provided
      *   in the order in which the keys appear in the @p list_of_keys argument.
+     *   If multiple values per key are allowed, the vector contains first all
+     *   values for key 1, then all values for key 2 and so forth. Using the
+     *   @p n_values_per_key vector allows the caller to sort entries in the
+     *   returned vector to specific keys.
      */
     std::vector<double>
     parse_map_to_double_array (const std::string &key_value_map,
                                const std::vector<std::string> &list_of_keys,
                                const bool has_background_field,
                                const std::string &property_name,
-                               const bool allow_subentries = false,
-                               std::shared_ptr<std::vector<unsigned int> > subentry_structure = nullptr);
+                               const bool allow_multiple_values_per_key = false,
+                               std::shared_ptr<std::vector<unsigned int> > n_values_per_key = nullptr);
 
     /**
      * This function takes a string argument that is assumed to represent
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -79,7 +79,11 @@ namespace aspect
      * This function takes a string argument that is interpreted as a map
      * of the form "key1 : value1, key2 : value2, etc", and then parses
      * it to return a vector of these values where the values are ordered
-     * in the same order as a given set of keys.
+     * in the same order as a given set of keys. If @p allow_subentries
+     * is set to 'true' it also allows entries of the form
+     * "key1: value1|value2|value3, etc", in which case the returned
+     * vector will have more entries than the provided
+     * @p list_of_keys.
      *
      * This function also considers a number of special cases:
      * - If the input string consists of only a comma separated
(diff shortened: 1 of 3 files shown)


gassmoeller:
@jdannberg, @naliboff, @mibillen can one of you take a look if this is what we talked about?

source/utilities.cc
```diff
@@ -230,7 +254,7 @@ namespace aspect
                        ExcMessage ("The required format for field <"
                                    + property_name
                                    + "> was not found. Specify a comma separated "
-                                   "list of `<double>' or `<id> : <double>'."));
+                                   "list of `<double>' or `<id> : <double>|<double|...'."));
```
naliboff:
```suggestion
                                   "list of `<double>' or `<id> : <double>|<double>|...'.")); 
```
Maybe a missing `>`?

gassmoeller:
Thanks for the quick review. I have fixed the comment and added one more functionality we will need. That is to return the actual substructure of the returned vector. As discussed we need this to ensure that all entries actually have the same number of phases for each individual composition (e.g. to distinguish `a:1|2,b:3` from `a:1,b:2|3`).

include/aspect/utilities.h
```diff
@@ -79,7 +79,11 @@ namespace aspect
...
      * it to return a vector of these values where the values are ordered
-     * in the same order as a given set of keys.
+     * in the same order as a given set of keys. If @p allow_subentries
+     * is set to 'true' it also allows entries of the form
+     * "key1: value1|value2|value3, etc", in which case the returned
+     * vector will have more entries than the provided
+     * @p list_of_keys.
      *
      * This function also considers a number of special cases:
      * - If the input string consists of only a comma separated
```
jdannberg: Do you want to add  comment here that mentions that you can not do this if `allow_subentries == true` because then the notation does not provide enough information?

include/aspect/utilities.h
```diff
@@ -110,6 +114,14 @@ namespace aspect
      * @param[in] property_name A name that identifies the type of property
      *   that is being parsed by this function and that is used in generating
      *   error messages if the map does not conform to the expected format.
+     * @param [in] allow_subentries If true allow having multiple subentries
+     *   for each key. If false only allow a single entry per key. In either
+     *   case each key is only allowed to appear once.
+     * @param [out] subentry_structure If handed over this
```
jdannberg: To me, this name does not seem as descriptive as it could be. What about something like `values_per_key` or `n_values_per_key`? That would also be consistent with the naming of the other input parameters.

unit_tests/parse_map_to_double_array.cc
```diff
@@ -82,6 +82,36 @@ TEST_CASE("Utilities::parse_map_to_double_array")
...
+    {0.0,100.0,200.0,100.0,300.0,400.0,500.0});
+
+    REQUIRE(*subentry_structure == std::vector<unsigned int>({1,1,2,1,1,1}));
+  }
+
+  {
+    INFO("check 10: ");
+    auto subentry_structure = std::make_shared<std::vector<unsigned int>>();
+
+    compare_vectors_approx(aspect::Utilities::parse_map_to_double_array ("C1:100, C2:200|, C3:300, C4:400, C5:500",
```
jdannberg: Oh, we actually allow to use the | but then not have an entry? Is that intended?

unit_tests/parse_map_to_double_array.cc
```diff
@@ -135,6 +165,32 @@ TEST_CASE("Utilities::parse_map_to_double_array FAIL ON PURPOSE")
   {"C1","C2","C3","C4","C5"},
   true,
   "TestField"), Contains("The keyword `all' is expected but is not found"));
+
+  // Subentries not allowed
+  INFO("check fail 7: ");
+  REQUIRE_THROWS_WITH(
+    aspect::Utilities::parse_map_to_double_array ("C1:100, C2:|200, C3:300, C4:400, C5:500",
```
jdannberg: Wouldn't this test be more helpful and easier to read if the output format was correct (and only the subentries are not allowed)? This way, how do you know which of the two it is?

source/utilities.cc
```diff
@@ -206,22 +231,40 @@ namespace aspect
...
+                        ++(*subentry_structure)[field_index];
+                    }
                 }
             }
+
+          for (const auto &entry: return_map)
+            for (const auto &sub_entry: entry)
+              return_values.push_back(sub_entry);
         }
       else if (Patterns::List(Patterns::Double(),1,n_fields).match(input_string))
```
jdannberg: We should make sure that `allow_subentries` is false in this case.

source/utilities.cc
```diff
@@ -89,21 +89,41 @@ namespace aspect
...
+
+      const auto key_pattern = (allow_subentries)
+                               ?
+                               Patterns::List(Patterns::Double(),
+                                              1,
+                                              std::numeric_limits<unsigned int>::max(),
+                                              "|")
+                               :
+                               Patterns::Double()
+                               ;
```
jdannberg: Move that to the end of the line before?

source/utilities.cc
```diff
@@ -146,7 +166,12 @@ namespace aspect
```
jdannberg: There is a space missing in line 162 above.

source/utilities.cc
```diff
@@ -146,7 +166,12 @@ namespace aspect
 
                   // Assign all the elements to the "all" value
                   for (unsigned int field_index=0; field_index<n_fields; ++field_index)
-                    return_values[field_index] = Utilities::string_to_double(key_and_value[1]);
+                    {
+                      return_map[field_index].push_back(Utilities::string_to_double(key_and_value[1]));
+
+                      if (subentry_structure)
+                        (*subentry_structure)[field_index] = 1;
+                    }
```
jdannberg: So I guess that mean we don't allow entries in the format `all:100|200`. Is that something we want, or, alternatively, should we state somewhere you can't do that?

include/aspect/utilities.h
```diff
@@ -110,6 +114,14 @@ namespace aspect
...
+     *   vector will be filled with as many components as
+     *   keys (+1 if there is a background field). Each value represents
+     *   how many subentries were found for this key. The sum over all
+     *   entries is the length of the returned vector of values.
      *
      * @return A vector of values that are parsed from the map, provided
      *   in the order in which the keys appear in the @p list_of_keys argument.
```
jdannberg: You should add some documentation here that explains the new structure with subentries. That could also be a good place to explain why the `subentry_structure` is useful (it means whoever calls the function can figure out which value belongs to which key).

unit_tests/parse_map_to_double_array.cc
gassmoeller: That is allowed by deal.II (it depends on the implementation of the `split_string_list` function, which explicitly allows this case). I think this is fine, likely nobody wants to use it.

source/utilities.cc
gassmoeller: I would not want to do that. Consider the case that somebody wants to allow multiple values per key, but does not know the input structure the user puts into the .prm file. You would still want to allow a simple list of only a single value per key. The parameter is `allow_multiple_values_per_key`, not `enforce_multiple_values_per_key`. But you are right, I should document that this format expects exactly one value per key (and we already enforce that, because `possibly_extend_from_1_to_N` expects either 1 entry, or `n_fields` entries).

gassmoeller:
I think I addressed all comments, and also added the functionality to hand over an existing vector of `n_values_per_key` that can be used as the expected template for the input parameter. This saves us from implementing Asserts for every parameter to check that its structure is the same as the last one that was parsed. Every change is in the last commit.

source/utilities.cc
```diff
@@ -94,27 +94,37 @@ namespace aspect
                                const std::vector<std::string> &input_field_names,
```
jdannberg: For some reason, the names of the first two function arguments are different in the declaration and in the implementation here.

source/utilities.cc
```diff
@@ -260,10 +287,18 @@ namespace aspect
...
+                (*n_values_per_key)[i] = 1;
+            }
+
+          if (check_structure)
             {
               for (unsigned int i=0; i<n_fields; ++i)
-                (*subentry_structure)[i] = 1;
+                AssertThrow((*n_values_per_key)[i] == 1,
+                            ExcMessage("The input parameter " + property_name + " does not have "
+                                       + "the expected number of values for field index " + std::to_string(i) + "."));
```
jdannberg: I find this error message a bit confusing. If this error is thrown, it means someone specified a list of doubles in the input file with one value per key, but some other parameter was specified with multiple values per key, provided in the key:value format (or there is some other reason why the function expects more than one value per key). 

So to help the user a bit more, I would tell the user how many entries were expected, and tell them something like: "To specify more than one value per field for this input parameter, you need to use the format <key>:value1|value2|..."

gassmoeller:
I addressed the comments. Thanks for the review!

PR #3269, opened 2019-10-21
more c++ override

tjhei:
part of #3258

### DIFF
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -765,7 +765,7 @@ namespace aspect
         /**
          * Destructor.
          */
-        virtual ~NamedAdditionalMaterialOutputs();
+        ~NamedAdditionalMaterialOutputs() override;
 
         /**
          * Return a reference to the vector of names of the additional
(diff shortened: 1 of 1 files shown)


include/aspect/material_model/interface.h
```diff
@@ -765,7 +765,7 @@ namespace aspect
         /**
          * Destructor.
          */
-        virtual ~NamedAdditionalMaterialOutputs();
+        ~NamedAdditionalMaterialOutputs() override;
```
bangerth: Isn't our usual style to leave the `virtual` in place even when we add `override` to the function?

tjhei: I don't know about "our usual style" with regard to that. I enabled ``modernize-use-override`` in #3249, which flags these places. Can you take a look at #3258?

gassmoeller: The C++ core guidelines specify that there should be exactly one keyword: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c128-virtual-functions-should-specify-exactly-one-of-virtual-override-or-final

PR #3271, opened 2019-10-21
Let phase function handle compositions

gassmoeller:
This PR builds on #3267 and allows the `MaterialModel::Utilities::PhaseFunction` class to parse more complex phase transition parameters. It is still up to the material model to correctly interpret these phase transitions (and we have not implemented that yet in any material model), but at least now we can get started using this object in more places.


tests/latent_heat_viscosity.prm
```diff
@@ -17,11 +17,11 @@ subsection Material model
...
     set Reference temperature         = 1600
-    set Phase transition depths       = 410000, 660000
-    set Phase transition widths       = 20000, 20000
-    set Phase transition temperatures = 1600, 1600
-    set Phase transition Clapeyron slopes = 3e6, -2e6
-    set Phase transition density jumps = 100, 200
+    set Phase transition depths           = all:410000|660000
+    set Phase transition widths           = all:20000|20000
+    set Phase transition temperatures     = all:1600|1600
+    set Phase transition Clapeyron slopes = all:3e6|-2e6
```
jdannberg: Ah, so this is not backwards compatible. I am fine with that, but you should add a changelog entry. 
This is something we can not address in the update script, right? Is there anything we can do about this?

tests/phase_function_syntax.prm
```diff
@@ -0,0 +1,154 @@
...
+  set Model name = vertical
+  subsection Vertical
+    set Magnitude = 10.0
+  end
+end
+
+
+# The parameters below this comment were created by the update script
+# as replacement for the old 'Model settings' subsection. They can be
+# safely merged with any existing subsections with the same name.
```
jdannberg: You can remove this.

tests/phase_function_syntax.prm
```diff
@@ -0,0 +1,154 @@
...
+    # Clapeyron slope, this is what determines the entropy change.
+    set Phase transition density jumps                 = 115.6, 200.0
+    set Corresponding phase for density jump           = 0, 0
+
+    # If the temperature is equal to the phase transition temperature, the 
+    # phase transition will occur at the phase transition depth. However, 
+    # if the temperature deviates from this value, the Clapeyron slope 
+    # determines how much the pressure (and depth) of the phase boundary
+    # changes. Here, the phase transition will be in the middle of the box
+    # for T=T1.
```
jdannberg: This comment is outdated.

tests/phase_function_syntax.prm
```diff
@@ -0,0 +1,154 @@
...
+    # determines how much the pressure (and depth) of the phase boundary
+    # changes. Here, the phase transition will be in the middle of the box
+    # for T=T1. 
+    set Phase transition depths                        = all:500000|700000
+    set Phase transition temperatures                  = background:1000|1000
+    set Phase transition Clapeyron slopes              = all:1e7|1e7
+
+    # We set the width of the phase transition to 5 km. You may want to 
+    # change this parameter to see how latent heating depends on the width
+    # of the phase transition.
```
jdannberg: Same here.

tests/phase_function_syntax.prm
```diff
@@ -0,0 +1,154 @@
+# This is a copy of latent_heat.prm, meant to test the input
+# syntax of the PhaseFunction object.
```
jdannberg: Can you be a bit more specific here and say what exactly is tested? 
Moreover, how is this test conceptually different from latent_heat_viscosity? Wouldn't it make more sense to have a test with more than one compositional field?

tests/latent_heat_viscosity.prm
jdannberg: And have you tested that all cookbooks still work?

gassmoeller:
I think I addressed all comments and took over the part of #3273 that was relevant for this PR. You were right the test was not more elaborate than the existing one, and since `latent_heat` is somewhat problematic as a test case (because it has its own interpretation of how to do phase transitions) I would rather add more test cases to #3273 than here. I removed the new test.

include/aspect/utilities.h
```diff
@@ -130,6 +130,8 @@ namespace aspect
      *   previous call to this function) then this vector is used as
      *   expected structure of the input parameter, and it is checked that
      *   @p key_value_map fulfills this structure.
+     * @param [in] allow_empty_keys Whether to allow empty keys that have
+     *   no associated entries. This also allows a completely empty map.
```
jdannberg: I find this name and description not quite clear, and related to that, it's not quite clear to me what we actually want to allow here. Allow empty key sounds like 'allow no values per key'. But that's not quite what this is, correct? Here, someone can just leave out one of the compositional fields from their list in the parameter file? Something like `background:1, field1:1, field3:1` (leaving out field2). Then our list of keys does have a different number of entries than `nfields`. But we still do all of the checks, like, if you have just one entry in the list and more than one field, the keyword has to be `all`. Or do you just want to allow something like `field:` (specifying the key but not the value)?

It works for a completely empty map, but just having some of the keys specified makes this logic even more complicated. (I am not saying you should take this out, I get that we want to allow this for users, I just find it very hard to make sure all the logic is correct)

gassmoeller:
Thanks for pointing out the logic problems. I have completely reworked the parse_map_to_double function and separated it into two function. The first one is an internal helper function that just parses the map (only checking for correct formatting, not checking for structure). In this new function I also simplified the logic for checking the 'all' keyword. After parsing, conditions like how many values per key are allowed or expected are checked separately. I think the new layout simplifies reading the functions. I also added new unit tests for the `allow_missing_keys` parameter.
I checked that the existing 'latent heat' cookbook still works as documented (I checked and that is the the only model using this material model outside of the tests).

Please take another look at the reworked functions. Thanks for reviewing this complicated stuff!

source/utilities.cc
```diff
@@ -95,225 +195,111 @@ namespace aspect
...
+        std::vector<unsigned int> values_per_key(n_fields, 0);
+
+        if (check_structure)
+          AssertThrow(n_values_per_key->size() == n_fields,
+                      ExcMessage("When providing an expected structure for input parameter " + property_name + " you need to provide "
+                                 + "as many entries in the structure vector as there are input field names (+1 if there is a background field). "
+                                 + "The current structure vector has " + std::to_string(n_values_per_key->size()) + " entries, but there are "
+                                 + std::to_string(n_fields) + " field names." ));
+
+        for (const auto &entry: parsed_map)
```
jdannberg: Would you be okay with renaming entry to map_entry? I find it hard to remember what entry.first is, so maybe that would help.

source/utilities.cc
```diff
@@ -95,225 +195,111 @@ namespace aspect
...
+      for (const auto &field_name: field_names)
         {
-          // No Patterns matches were found!
-          AssertThrow (false,
-                       ExcMessage ("The required format for field <"
-                                   + property_name
-                                   + "> was not found. Specify a comma separated "
-                                   + "list of `<double>' or `<key1> : <double>|<double>|..., "
-                                   + "<key2> : <double>|... , ... '."));
+          const auto entry_range = parsed_map.equal_range(field_name);
```
jdannberg: Can you type out the variable type of `entry_range`? I find this really hard to read... I hadn't seen the equal_range function before, but if you tell me that the return type is a pair of iterators, it would be more clear what's happening.

source/utilities.cc
```diff
@@ -87,6 +87,106 @@ namespace aspect
...
+                    for (const auto &key: list_of_keys)
+                      for (const auto &value : values)
+                        {
+                          parsed_map.emplace(key, Utilities::string_to_double(value));
+                        }
+                  }
+                // Handle lists of multiple unique entries
+                else
+                  {
+                    AssertThrow (parsed_map.find(key_and_value[0]) == parsed_map.end(),
```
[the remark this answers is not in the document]
gassmoeller: check fail 1 triggers this assert already.

source/utilities.cc
```diff
@@ -87,6 +87,106 @@ namespace aspect
...
+                                 ExcMessage ("The keyword <"
+                                             + key_and_value[0]
+                                             + "> in "
+                                             + property_name
+                                             + " is listed multiple times. "
+                                             "Check that you have only one value for "
+                                             "each field id in your list."));
```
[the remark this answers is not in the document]
gassmoeller: That is correct. Originally, I wanted to not even hand over `list_of_keys` to this function, but then I had to, to support the list without key names, and the keyword `all`. The purpose of this function is only to convert a string into a multimap, with as little error checking as possible.

source/utilities.cc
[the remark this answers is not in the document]
gassmoeller: Yes, the reason is I want to make the `parse_string_to_map` as generic and with as little error checking as possible. Additionally, to move this over into the other function I need to hand over more input parameters like `has_background_field`, which is really an irrelevant parameter for that function. I imagined the streamlined process as:

1. Set up things like has_background_field
2. Call `parse_string_to_map` to do the parsing.
3. Do all specific error checking concerning the many options in `parse_map_to_double_array`.

That is why I would prefer to keep this assert here.

gassmoeller:
Thanks for the review. I think I addressed all of your comments. As explained I would like to keep that assert in the outer function to make the inner function easier to understand and avoid handing over unnecessary parameters to the inner function.

source/utilities.cc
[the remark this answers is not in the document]
jdannberg: Oh, alright, I see. It's just a test that already existed.

source/utilities.cc
```diff
@@ -221,15 +226,15 @@ namespace aspect
                                  + "The current structure vector has " + std::to_string(n_values_per_key->size()) + " entries, but there are "
                                  + std::to_string(n_fields) + " field names." ));
 
-        for (const auto &entry: parsed_map)
+        for (const std::pair<std::string, double> &key_and_value: parsed_map)
```
jdannberg: I really like this change. I always struggle with pairs because from the name it's usually not clear what pair.first and pair.second are. This makes it really easy to read.

source/utilities.cc
```diff
@@ -221,15 +226,15 @@ namespace aspect
...
           {
-            const auto field_name = std::find(field_names.begin(),field_names.end(),entry.first);
+            const auto field_name = std::find(field_names.begin(),field_names.end(),key_and_value.first);
```
jdannberg: Just for consistency, as you changed it in all the other places too: Maybe change this one too: 
`auto` --> `const std::string`

gassmoeller: unfortunately this is not a `std::string`, it is a `std::vector<std::string>::iterator`. Still want me to change it?

jdannberg: Hm. I am not sure. 
But maybe just the fact that I didn't know the return type of the functions means changing it would make the code clearer?

gassmoeller:
Alright, I updated the type. Should be ready once the tester is done.

PR #3272, opened 2019-10-21
Add ASPECT publications

gassmoeller:
Since we are all adding recent publications, here are a few more :smile:


ljhwang:
Adding doi's to all references would be useful information.

gassmoeller:
I added the dois. It is really annoying that Google Scholar does not automatically include them in the generated bibtex entries :-(.

ljhwang:
I will typically generate this from the journal as the “official” version.  I do not know what Google Scholar does!

Best,
-Lorraine

*****************************
Lorraine Hwang, Ph.D.
530.752.3656

Issue #3001, opened 2019-05-24
Ellipsoidal chunk pull function is not the inverse of push function


bobmyhill:
The pullback function in the ellipsoidal chunk model provides the correct longitude, but only the first iterate of the latitude (for which there is no closed-form solution). See here:
https://gssc.esa.int/navipedia/index.php/Ellipsoidal_and_Cartesian_Coordinates_Conversion

This explains the discrepancy spotted by @MFraters in Issue #1950.

I will add some comments to the geometry model (in the pullback function), but do we want to iterate a couple more times to get a more accurate solution?

bobmyhill:
Hmmm, ok, so ASPECT's current equations are indeed the analytical solution for a uniaxial ellipsoid (they are only the first iterate for a triaxial ellipsoid, for which there is no analytical solution). The problem must be elsewhere. I'll close this PR and continue hunting.

PR #3274, opened 2019-10-25
Reverse order natural coordinates

anne-glerum:
While updating my models, I realized that the order of the coordinates that are read in for `AsciiData` with an `EllipsoidalChunk` has changed; the `cartesian_to_natural_coordinates` function outputs radius, lat, lon. Looking into the geometry model, the comments (radius, lon, lat) do not seem to match the order of the coordinates used in the `cartesian_to_natural_coordinates` and `natural_to_cartesian_coordinates` functions (radius, lat, lon). 

As I think the order should be as written in the comments (radius, lon, lat), I've reordered the coordinates. However, the failure of the `natural_coordinates_conversion` test is now even worse, so @MFraters, do you think these changes are correct? Note that for an eccentricity of zero, the functions recover the original coordinates. 

@bobmyhill pinging you as well as we introduced the natural coordinates during the hack.

* [x] I have updated the test results.
* [x] I have added a changelog entry in the [doc/modules/changes](../doc/modules/changes) directory that will inform other users of my change.

### DIFF
--- a/source/geometry_model/ellipsoidal_chunk.cc
+++ b/source/geometry_model/ellipsoidal_chunk.cc
@@ -775,8 +775,8 @@ namespace aspect
       const double radius = get_radius(position_point);
       std::array<double,dim> position_array;
       position_array[0] = radius + transformed_point(2);
-      position_array[1] = transformed_point(1);
-      position_array[2] = transformed_point(0);
+      position_array[1] = transformed_point(0);
+      position_array[2] = transformed_point(1);
 
       return position_array;
     }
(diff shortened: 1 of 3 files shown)


source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -797,10 +797,10 @@ namespace aspect
       // We receive radius, longitude, latitude and we need to turn it first back into
       // longitude, latitude, depth for internal use, and push_forward to cartesian coordiantes.
       Point<3> position_point;
-      position_point(0) = position_tensor[2];
-      position_point(1) = position_tensor[1];
+      position_point(0) = position_tensor[1];
+      position_point(1) = position_tensor[2];
 
-      const double radius = semi_major_axis_a / (std::sqrt(1 - eccentricity * eccentricity * std::sin(position_point(1)) * std::sin(position_point(1))));
+      const double radius = semi_major_axis_a / (std::sqrt(1 - eccentricity * eccentricity * std::sin(position_point(0)) * std::sin(position_point(0))));
```
MFraters: I do not think this change is correct, since you only need to switch the long and lat intake. I also tested it, and if you remove just this, the new point will be the same as before. 

Looking at it, it might be worth implementing a few more iterations for the pull back function, but that is a separate issue.

anne-glerum: Agreed! This matches the functions in the reference and returns the same point as before.

bobmyhill: @MFraters - the documentation was added by me. I have no idea where the original implementation came from, so I had to decide between keeping the description from the literature, modifying the description to fit the code, or rewriting the code. I did check that the two algorithms were equivalent, but I agree that this is not perfect. 

I guess finding the original literature or rewriting the code would be the optimal solutions.

bobmyhill:
Thanks @anne-glerum!

source/geometry_model/ellipsoidal_chunk.cc
MFraters: The original implementation came from @dsarahstamps (see issue #1950). 

@bobmyhill Thanks for adding that documentation! That made checking this pull request a lot easier :)

I made an issue for the rewrite: #3277 

@bobmyhill, I now also saw your issue #3001. Are you saying there that iterating would not help because it is already the analytical solution?

PR #3275, opened 2019-10-29
add tests

naliboff:
This PR adds two new tests for the Drucker Prager derivatives, which follow the structure of the visco plastic derivatives tests (e.g, `visco_plastic_derivatives_2d` and `visco_plastic_derivatives_3d`). These changes are needed, as #3257  requires simulator access during the tests. 


naliboff:
@MFraters - can you take a look at this when you have a chance? If the new tests look reasonable, I'll do a follow-up commit removing the original test.

naliboff:
@MFraters - thanks! Let me know if you would like these two commits merged. 

@gassmoeller - Is the practice now to add change log entries for these types of minor test updates?

gassmoeller:
No, we only need changelog entries for:
- new features
- bugfixes
- changes to existing features that affect users

So no need to document changed tests.

Feel free to merge once the testers are happy.

tests/drucker_prager_derivatives_2d.cc
```diff
@@ -5,13 +5,20 @@
...
 #include <iostream>
 
 template<int dim>
-int f()
+void f(const aspect::SimulatorAccess<dim> &simulator_access,
+       aspect::Assemblers::Manager<dim> &)
 {
-  std::cout << std::endl << "Test with dimension " << dim << std::endl;
+
+  std::cout << std::endl << "Testing DruckerPrager derivatives against analytical derivatives " << std::endl;
```
MFraters: could you check the above sentence? Shouldn't it be: `Testing DruckerPrager derivatives against finite difference derivatives`? Same for the other test.

naliboff: @MFraters - yes, the wording you used above is correct. Thanks for catching that!

naliboff:
@MFraters - comment addressed, I think should be ready to merge unless you would like to have another look.

PR #3281, opened 2019-11-01
doc: remove incorrect @copydoc link

tjhei:


gassmoeller:
This PR looks good. 

At some point over the last weeks the unity build memory footprint must have increased over 7.x GB. Some of our Jenkins agents only have 8 GB RAM and if a PR is tested on them they now fail to compile within the time limit :-(. Not sure what to do here, I could try splitting the unity build into two build units and compile them serially.

tjhei:
Can you do the unity build without parallelism or does it run out of memory even then?

gassmoeller:
It runs out of memory even on a single process. What I can do is split the unity build into a few chunks and then compile them serially. That should not make it much slower, but it might work, while a single chunk would fail.

Issue #3248, opened 2019-10-05
inconsistent constraints crash


tjhei:
I ran into this scary bug when testing Q3 discretization for the Stokes system for the hollow_sphere benchmark. To reproduce, put the attached .prm into benchmarks/hollow_sphere/ and run with 3
 MPI processes.

[inconsistent_hollow_sphere.prm.txt](https://github.com/geodynamics/aspect/files/3693717/inconsistent_hollow_sphere.prm.txt)

gassmoeller:
I can confirm this is an issue. I attach my screen output:
```
-----------------------------------------------------------------------------
-- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.
--     . version 2.2.0-pre (fix_latent_heat_viscosity, f2f88270b)
--     . using deal.II 9.2.0-pre (master, 24fc4f42ed)
--     .       with 32 bit indices and vectorization level 2 (256 bits)
--     . using Trilinos 12.10.1
--     . using p4est 2.0.0
--     . running in DEBUG mode
--     . running with 3 MPI processes
-----------------------------------------------------------------------------

Loading shared library <./libhollow_sphere.so>

-----------------------------------------------------------------------------
-- For information on how to cite ASPECT, see:
--   https://aspect.geodynamics.org/citing.html?ver=2.2.0-pre&sha=f2f88270b&src=code
-----------------------------------------------------------------------------
Number of active cells: 768 (on 3 levels)
Number of degrees of freedom: 81,330 (67,470+6,930+6,930)

*** Timestep 0:  t=0 seconds
   Rebuilding Stokes preconditioner...
   Solving Stokes system... 0+20 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 1: 1

   Rebuilding Stokes preconditioner...
   Solving Stokes system... 0+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 2: 8.87017e-07


   Postprocessing:
     RMS, max velocity:                     3.67 m/s, 16.6 m/s
     System matrix memory consumption:      101.72 MB
     Pressure at top/bottom of domain:      2.64e-13 Pa, 500 Pa
     Computing dynamic topography           
     Writing graphical output:              output/solution/solution-00000
     Errors u_L1, p_L1, u_L2, p_L2 topo_L2: 1.215723e+00, 6.541060e+00, 7.145619e-01, 4.057621e+00, 1.056476e-02

[...]

PR #3282, opened 2019-11-04
remove constraint consistency check

tjhei:
It turns out that constraints.is_consistent_in_parallel() is too eager
and gives false alarm if constraints differ by machine epsilon. For now,
just remove this call.


tjhei:
see #3248

bangerth:
This is an acceptable stop-gap, but I'd still like to find out more about why you get different constraints. I think we ought to be computing them in the exact same way on all processors.

tjhei:
updated.

PR #3287, opened 2019-11-08
Allow multiple output formats for depth average postprocessor

gassmoeller:
I repeatedly had to explain the depth average output, because the gnuplot format is so unintuitive to read. While it is nice for plotting, many users (and myself) want to do further analysis with the data and for that an easier format that can be read into python, bash or other scripts is more useful.
This PR adds the possibility to write the depth_average output into multiple formats and enables the .txt format by default. I think this can avoid a lot of confusion and makes the results more useful.


tests/initial_porosity/screen-output
```diff
@@ -18,7 +18,7 @@ Number of degrees of freedom: 5,641 (1,666+801+1,666+225+833+225+225)
    Rebuilding Stokes preconditioner...
    Solving Stokes system... 0+36 iterations.
    Solving for u_f in 1 iterations.
-      Relative nonlinear residuals (temperature, compositional fields, Stokes system): 2.02861e-16, 1.20111e-16, 0, 0.752886
+      Relative nonlinear residuals (temperature, compositional fields, Stokes system): 2.14876e-16, 4.47907e-17, 0, 0.752886
```
tjhei: I notice a few unrelated output changes (these are all within tolerance of course). Are these from the blessed docker image?

gassmoeller: I think these are changes that usually fall below numdiff's thresholds and happened before, but they are updated anyway when the file is updated for other changes.

PR #3289, opened 2019-11-08
Add a mass flux statistics for the depth average postprocessor.

gassmoeller:
This adds a mass flux statistics computation to the depth average postprocessor. One could already compute an approximation for this by multiplying the sinking velocity with the reference density and multiplying by two to account for upward mass movement, but that is less accurate for varying densities - downwellings tend to have different densities than upwellings and for various reasons the two might not be balanced (open boundaries, free surface, etc.). This new field can be used to measure layering in a convecting system.


tests/depth_average_06.prm
```diff
@@ -0,0 +1,90 @@
+# Test the 'depth average' postprocessor on a coarse mesh version of
+# convection_box.prm. This test makes no sense since we average over 10 depth
+# intervals but we only compute on a 4x4 mesh -- This is sufficient to produce
+# reasonable data, but not enough for actual resolution.
```
jdannberg: Can you update the documentation?
This is a copy of test..., with the only difference that...
(or even better, just include the other input file and then only list your changes, if there aren't that many)

gassmoeller:
I updated the test comment (it is a pretty heavily modified copy, so I kept it separate from other tests and wrote a new documentation) and updated the test results. Should be ready when the tester is done.

tests/ascii_data_adiabatic_conditions/depth_average.gnuplot
```diff
@@ -4,74 +4,74 @@
 #
 # For a description of the GNUPLOT format see the GNUPLOT manual.
 #
-# <depth> <time> <temperature> <adiabatic_temperature> <adiabatic_pressure> <adiabatic_density> <adiabatic_density_derivative> <velocity_magnitude> <sinking_velocity> <viscosity> <vertical_heat_flux> 
-0 0 3.05 3 1 5 3.46021e-07 3.17701e-13 3.0917e-14 1e+21 1.6263e-06 
-289000 0 3.05 3 1 5 3.46021e-07 3.17701e-13 3.0917e-14 1e+21 1.6263e-06 
+# <depth> <time> <temperature> <adiabatic_temperature> <adiabatic_pressure> <adiabatic_density> <adiabatic_density_derivative> <velocity_magnitude> <sinking_velocity> <viscosity> <vertical_heat_flux> <vertical_mass_flux> 
+0 0 3.05 3 1 5 3.46021e-07 3.17104e-13 3.08398e-14 1e+21 1.6263e-06 6.89975e-18
```
jdannberg: Any idea why these numbers have changed? That must have happened in an earlier change, and is probably so small that numdiff is fine with it?

gassmoeller: Yes that is the best explanation I have as well.

PR #3291, opened 2019-11-12
reduce storage for matrix-free gmg

tcclevenger:
Reduce the storage of the matrix-free GMG Stokes setup by only storing one viscosity value for each cell instead of one viscosity value for each quadrature point. This was left over from old code where I was querying viscosity values from some functional representation on each level, not using averaging. When we update the code in the future to not requiring averaging, this may look different. In the meantime, this greatly reduces storage for the case of cell-wise coefficient averaging.

I did not create any new tests since this is not a new feature, and all tests currently using GMG should pass.

### For all pull requests:

### DIFF
--- a/include/aspect/stokes_matrix_free.h
+++ b/include/aspect/stokes_matrix_free.h
@@ -84,7 +84,7 @@ namespace aspect
         /**
          * Returns the viscosity table.
          */
-        const Table<2, VectorizedArray<number> > &
+        const Table<1, VectorizedArray<number> > &
         get_viscosity_x_2_table();
 
         /**
@@ -114,7 +114,7 @@ namespace aspect
         /**
          * Table which stores the viscosity on each quadrature point.
          */
-        Table<2, VectorizedArray<number> > viscosity_x_2;
+        Table<1, VectorizedArray<number> > viscosity_x_2;
 
         /**
          * Pressure scaling constant.
@@ -191,7 +191,7 @@ namespace aspect
         /**
          * Table which stores the viscosity on each quadrature point.
          */
-        Table<2, VectorizedArray<number> > one_over_viscosity;
+        Table<1, VectorizedArray<number> > one_over_viscosity;
 
         /**
          * Pressure scaling constant.
@@ -270,7 +270,7 @@ namespace aspect
         /**
          * Table which stores the viscosity on each quadrature point.
          */
-        Table<2, VectorizedArray<number> > viscosity_x_2;
+        Table<1, VectorizedArray<number> > viscosity_x_2;
 
         /**
           * Information on the compressibility of the flow.
--- a/include/aspect/stokes_matrix_free.h
+++ b/include/aspect/stokes_matrix_free.h
@@ -112,7 +112,7 @@ namespace aspect
                           const std::pair<unsigned int, unsigned int> &cell_range) const;
 
         /**
-         * Table which stores the viscosity on each quadrature point.
+         * Table which stores a viscosity value for each cell.
          */
         Table<1, VectorizedArray<number> > viscosity_x_2;
 
@@ -189,7 +189,7 @@ namespace aspect
                                      const std::pair<unsigned int,unsigned int>       &cell_range) const;
 
         /**
-         * Table which stores the viscosity on each quadrature point.
+         * Table which stores a viscosity value for each cell.
          */
         Table<1, VectorizedArray<number> > one_over_viscosity;
 
@@ -268,7 +268,7 @@ namespace aspect
                                      const std::pair<unsigned int,unsigned int>       &cell_range) const;
 
         /**
-         * Table which stores the viscosity on each quadrature point.
+         * Table which stores a viscosity value for each cell.
          */
         Table<1, VectorizedArray<number> > viscosity_x_2;
 
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -709,7 +709,7 @@ namespace aspect
 
     for (unsigned int cell=cell_range.first; cell<cell_range.second; ++cell)
       {
-        VectorizedArray<number> cell_viscosity_x_2 = viscosity_x_2(cell);
+        const VectorizedArray<number> &cell_viscosity_x_2 = viscosity_x_2(cell);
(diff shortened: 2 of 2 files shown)


include/aspect/stokes_matrix_free.h
```diff
@@ -114,7 +114,7 @@ namespace aspect
         /**
          * Table which stores the viscosity on each quadrature point.
```
tjhei: you will need to update the comments

include/aspect/stokes_matrix_free.h
```diff
@@ -191,7 +191,7 @@ namespace aspect
...
          */
-        Table<2, VectorizedArray<number> > one_over_viscosity;
+        Table<1, VectorizedArray<number> > one_over_viscosity;
```
tjhei: are these tables in the three operators the same? if yes, you could only compute and store it once and only keep a reference here.

source/simulator/stokes_matrix_free.cc
```diff
@@ -984,8 +981,7 @@ namespace aspect
 
     for (unsigned int cell=cell_range.first; cell<cell_range.second; ++cell)
       {
-        AssertDimension(viscosity_x_2.size(0), data.n_macro_cells());
-        AssertDimension(viscosity_x_2.size(1), velocity.n_q_points);
+        VectorizedArray<number> cell_viscosity_x_2 = viscosity_x_2(cell);
```
tjhei: do you need to make a copy here? I would think a const ref would be enough

source/simulator/stokes_matrix_free.cc
```diff
@@ -1056,6 +1052,8 @@ namespace aspect
     FEEvaluation<dim,degree_v,degree_v+1,dim,number> velocity (data, 0);
     for (unsigned int cell=cell_range.first; cell<cell_range.second; ++cell)
       {
+        VectorizedArray<number> cell_viscosity_x_2 = viscosity_x_2(cell);
```
tjhei: same here?

source/simulator/stokes_matrix_free.cc
```diff
@@ -1367,14 +1365,16 @@ namespace aspect
     sim.current_constraints.distribute(u0);
     u0.update_ghost_values();
 
-    const Table<2, VectorizedArray<double>> viscosity_x_2_table = stokes_matrix.get_viscosity_x_2_table();
+    const Table<1, VectorizedArray<double> > viscosity_x_2 = stokes_matrix.get_viscosity_x_2_table();
```
tjhei: why do you make a copy of the table here?

source/simulator/stokes_matrix_free.cc
```diff
@@ -1367,14 +1365,16 @@ namespace aspect
...
     FEEvaluation<dim,velocity_degree,velocity_degree+1,dim,double>
     velocity (*stokes_matrix.get_matrix_free(), 0);
     FEEvaluation<dim,velocity_degree-1,velocity_degree+1,1,double>
     pressure (*stokes_matrix.get_matrix_free(), 1);
 
     for (unsigned int cell=0; cell<stokes_matrix.get_matrix_free()->n_macro_cells(); ++cell)
       {
+        VectorizedArray<double> cell_viscosity_x_2 = viscosity_x_2(cell);
```
tjhei: same here

include/aspect/stokes_matrix_free.h
tcclevenger: I think the issue is that the cells withing the matrix-free object can be iterated through in a different ordering depending on the matrix-free operator and so the indexing could change for each table.

source/simulator/stokes_matrix_free.cc
tcclevenger: Is having a const ref in the cell loop better than looking up the value each time in the quadrature loop below (where the value is used)?

tcclevenger:
I think I may have accidentally started the tester?

include/aspect/stokes_matrix_free.h
tjhei: I guess we are now talking only about a couple of MB, so I am okay with it. I would be surprised if the order is different though?

source/simulator/stokes_matrix_free.cc
tjhei: I don't know to be honest and it is hard to figure out without looking at assembly.

include/aspect/stokes_matrix_free.h
tcclevenger: You are right, it is different for DG finite elements and continuous finite elements, but this isn't an issue for us since all operators are based on continuous elements.

I could add a function to each operator that sets a viscosity table?

tcclevenger: I say we don't worry right now with the reduction of the viscosity tables into 1. This whole structure will look different when we add in the higher order projection.

tjhei: ok, let's do that later.

tjhei:
I am okay with the changes, but I let others have a chance at looking at it before merging.

gassmoeller:
I am fine with the changes as well. Could you give a quick update on the state of using variable viscosities within the cells? This is mostly so that I know how to plan future work, and what to say if I am asked about it.

tcclevenger:
Thanks @gassmoeller!

I just finished defending my thesis. Now I will have much more time to work on the coefficient averaging. The new averaging is the top priority now. Timo and I will discuss this soon and I'll be able to give you a better answer, but I would like to have significant progress on it by the middle of December.

The idea is still the same: Cellwise averaging where the viscosity is discontinuous, Q2 or DGQ2 interpolation where the viscosity is continuous. For all the other coefficients (density etc), we never need to touch them so I imagine leaving them as quadrature point representations is ok?

gassmoeller:
Awesome, congratulations on your defense! You did some great work with the GMG!

Yeah the plans sound perfect. Let me know when you want input or model problems to toy around with.

tjhei:
I would like to add that viscosity averaging is high on our list, but there is other important stuff that will keep us busy for the rest of the year.
A realistic example would be very useful to have soon, especially if it would allow us to experiment with different ways to average / not average viscosities.

PR #3292, opened 2019-11-13
Reorder viscoelastic-plastic viscosity calculation

naliboff:
This PR reorders the viscoelastic plastic material model so that the viscoelastic viscosity is calculated prior to yielding and viscoelastic stresses are included when evaluating whether plastic yielding is occurring. This approach is consistent with Moresi et al. (2003), while the previous approach deviates from the approach of Moresi and others (Kaus, etc).

The reformulation and testing was done in conjuction with @dansand and @anne-glerum.

@dansand - would you mind taking a look and seeing if all of the new terminology (and formulation) is consistent with our last discussions?

Other developers - please note that there will be a follow up PR in a few days that removes the VEP material and adds elasticity to the visco plastic material model. That PR uses the VEP formulation updated in this PR and for consistency I thought we should update this material model as well.


doc/modules/changes/20191013_naliboff
```diff
@@ -0,0 +1,6 @@
+Changed: The viscoelastic plastic material has been reordered so
+that the viscoelastic viscosity is calculated prior to yielding
+and viscoelastic stresses are taken into account when comparing
+the current stress against the plastic yield stress.
```
gassmoeller: Could you add the comment that this is now consistent with Louis's paper that you cite in the code?

source/material_model/viscoelastic_plastic.cc
```diff
@@ -77,62 +79,55 @@ namespace aspect
...
 
-                  // If the viscous stress is greater than the yield strength, rescale the viscosity back to yield surface.
-                  // If the viscous stress is less than the yield stress, the yield viscosity is equal to the pre-yield value.
-                  if ( stresses_viscous[j] >= stresses_yield[j]  )
+                  // If the viscoelastic stress is greater than the yield strength, rescale the viscosity back to yield surface.
+                  // If the viscoelastic stress is less than the yield stress, the yield viscosity is equal to the pre-yield (viscoelastic) value.
+                  if ( stresses_ve[j] >= stresses_yield[j]  )
                     {
-                      viscosities_viscoplastic[j] = stresses_yield[j] / (2.0 * edot_ii);
+                      viscosities_vep[j] = stresses_yield[j] / std::sqrt(std::fabs(second_invariant((2. * (deviator(strain_rate)) + stress_old / (elastic_shear_moduli[j] * dte) ) ) ) );
```
gassmoeller: You divide by the strain rate invariant here, but you removed the security measure above that replaces the strain rate with a reference strain rate if it is 0. Is this no longer necessary, because if the strain rate is 0 the stresses_ve can never be larger than the stresses_yield? Otherwise we might run into an divide by zero here.

naliboff:
@gassmoeller - Thanks for the review. 

The short answer is we have been doing additional benchmarking for the implementation, although the one true benchmark (e.g., analytical solution) works for either implementation. Specifically, @dansand has been comparing the results of different codes for VEP problems and is going to write some related cookbooks. I'll check with @dansand to see if we can include one his benchmarks here or in a follow-up PR.

source/material_model/viscoelastic_plastic.cc
naliboff: Yes, this security measure absolutely should still be there. I'm surprised we did not get a crash on the first nonlinear iteration when the elastic stresses are 0  (time step 0). The code has been adjusted accordingly.

naliboff:
@gassmoeller - I took a look at the test and don't have a good physical explanation for why with the new ordering no linear Stokes solves are need on the second non-linear iteration of the second time step. I looked at the residuals (via stokes_residuals.txt) and the first value of the residual on the second non-linear iteration is the same as the last value of the previous non-linear iteration (e.g., no iterations needed).

This test is a derivative of the `viscoelastic_stress_buildup-yield` benchmark and I reran that to see if the same behavior is present there. For the few first time steps (10 or 20) there are a few non-linear iterations with that require 0 Stokes solves, but after the elastic stresses start to build-up and the yield stress is reached at least a few linear Stokes solves are required.
 
I'm not sure if this behavior should be cause for further concern and diagnosis. Odd that it occurs in just one of the tests for the same problem (e..g, only difference is how we define the elastic time step).

source/material_model/viscoelastic_plastic.cc
```diff
@@ -77,62 +79,66 @@ namespace aspect
...
+                  const double edot_ii = ( (this->get_timestep_number() == 0 && strain_rate.norm() <= std::numeric_limits<double>::min() )
+                                           ?
+                                           reference_strain_rate
+                                           :
+                                           std::max(std::sqrt(std::fabs(second_invariant(edot))), minimum_strain_rate) );
+
+                  // Calculate viscoelastic viscosity (equation 28 from Moresi et al., 2003)
+                  viscosities_ve[j] = viscosities_pre_yield[j] * dte / (dte + (viscosities_pre_yield[j]/elastic_shear_moduli[j]));
+
+                  // Calculate updated viscoselastic stress magnitude, which will be compared with the plastic yield stress
```
anne-glerum:
```suggestion
                  // Calculate updated viscoelastic stress magnitude, which will be compared with the plastic yield stress
```

source/material_model/viscoelastic_plastic.cc
```diff
@@ -434,6 +440,14 @@ namespace aspect
                                    "This force term is added onto the right-hand side force vector in the "
                                    "system of equations. "
                                    "\n\n "
+                                   "When plastic yielding occurs, the effective viscosity in equation 29 and 30 is the "
+                                   "plastic viscosity (equation 35). If the current stress is below the plastic "
+                                   "yield stress, the effective viscosity is still as defined in equation 28. "
+                                   "During non-linear iterations, we define the current stress prior to yielding "
+                                   "(e.g., value compared to yield stress) as "
```
anne-glerum:
```suggestion
                                   "(i.e., value compared to yield stress) as "
```

tests/viscoelastic_plastic_numerical_elastic_time_step_yield/statistics
```diff
@@ -15,5 +15,5 @@
 # 15: Schur complement iterations in Stokes preconditioner
 # 16: RMS velocity (m/year)
 # 17: Max. velocity (m/year)
-0 0.000000000000e+00 0.000000000000e+00 400 3803 441 5043 1 0  0  0  0 26 28 28 2.57523019e-02 4.43529491e-02 
-1 1.000000000000e+03 1.000000000000e+03 400 3803 441 5043 1 0 11 11 12 14 16 16 2.57523019e-02 4.43549287e-02 
+0 0.000000000000e+00 0.000000000000e+00 400 3803 441 5043 1 0  0  0  0         26 28 28 2.57523019e-02 4.43529491e-02 
+1 1.000000000000e+03 1.000000000000e+03 400 3803 441 5043 1 0 11 11 12 4294967295  0  0 2.57523019e-02 4.43529491e-02
```
tjhei: unrelated, but do we really output -1 here for the number of Stokes iterations?

tjhei:
This looks good to me. Two comments:
1. Rene is right in that a test that shows the difference between the two ways to implement it, would be great.
2. Does this the wrong shear band angle we have been seeing in the past?

tests/viscoelastic_plastic_numerical_elastic_time_step_yield/statistics
gassmoeller: I think that is a bug with a specific deal.II version (9.0) if there are no Stokes iterations. We then output `invalid_unsigned_int`. I saw this a while ago, but never got around to fixing it. I think it is fixed for deal.II 9.1.

naliboff:
@gassmoeller - this test is a bit odd as the benchmark it is derived from has a fixed strain rate throughout the model (boundary driven) and is only designed to check whether the viscoelastic stresses build up as predicted by the analytical solution. The VEP rheology is utilized in the test as the yield stress is reduced to a value (1e6 Pa) that should be reached after the first time step. The updated test shows the stresses are still below 1e6 Pa, so that at least is working.

However, I agree we absolutely need another test case with yielding and I'm going to submit a separate PR (using original VEP formulation) that derives from the VEP brick benchmark experiment. 

@tjhei - Unfortunately, I don't think this will have any effect on the shear band angles.

dansand:
The figure below shows a simple shear benchmark using the VEP model before and after the reordering. Also shown are results from Underworld2.

In this model the strain rate is 0.025, and the background viscosity is 100. The cohesion is 3.5.  The shearing velocity is switched to zero at time = 1. 

With a visco-plastic rheology, the viscous stress (5) exceeds the cohesion and yielding will occur at the first timestep.  

In a visco-elastic-plastic rheology, it takes a finite time for stresses to build up, and yielding does not take place before the shearing velocity is switched to zero.  

This model shows the inconsistency in the original VEP implementation (prior to reordering), where the stress evaluation is determined from the viscous stress relationship. Because this predicts yielding, the background viscosity is replaced with a reduced plastic value.  

The changes that @naliboff  and I have implemented address this problem, and the results are now much closer to the analytical solution, as well as the results from Underworld2. 

The input files can be found here: https://github.com/dansand/vep_benchmarks/tree/master/simple_shear

I will also make a pull request to add these to the ASPECT benchmarks collection. 

simple_shear_comparison

naliboff:
@dansand - Excellent! Would you expect the Underworld and revised ASPECT formulation to converge toward the analytical solution further with increasing resolution and/or smaller time steps?

@gassmoeller - I looked at the other VEP benchmark (viscoelastic_plastic_shear_bands) and the same odd convergence behavior (0 stokes iterations) occurs in the first few time steps as well until localization begins. I'll add in the test case for that benchmark, but this behavior appears to be a consistent feature with the revised implementation.

dansand:
@naliboff I will check this. Based on Farrington, 2014 it looks like closer convergence was achieved with smaller elastic timesteps.

dansand:
@naliboff here are the results for a x0.1 reduction in both the elastic time step (0.01) and the numerical timestep (0.0025). 

I'll add the prm file to this pull request now. 


simple_shear_dte_comparison

gassmoeller:
@naliboff: Ok, if you expect the Stokes solution to not change over the first timesteps then it is fine that the solver converges immediately. Can you still modify the test to just do some more timesteps (until the solution changes and it actually solves something)? Otherwise this might be the right behavior, but it is not testable behavior (i.e. we can not distinguish between the solver doing the right thing, and the solver doing nothing, because nothing for this model is the right thing :-)).

naliboff:
@gassmoeller - I updated the test in question and added a new test to show when the behavior in question (0 Stokes iterations) occurs. The test parameter files also mention this in the documentation at the top of the file. The benchmark @dansand provided (thanks!) is also included. If you think further investigation of the Stokes solver behavior is warranted, lets certainly dig a bit deeper before merging.

gassmoeller:
Thanks. I think this is good to merge, and I also think I finally understand why the Stokes solver does not need any iterations. The shear velocity is constant and the elastic stress buildup is also more or less constant over the first time steps, resulting in a constant velocity field. Only when things change (e.g. the elastic stress is reaching steady-state, or plasticity sets in, or the strain rate is changed) does the solver need to do anything. Thanks for testing all of this!

PR #3293, opened 2019-11-15
Add missing benchmark to check script

gassmoeller:
I noticed this while reviewing #3292. This benchmark was not in the check script and somehow we forgot to update its parameter file. This should fix it. Edit: ~~The tests will fail until #3292 is merged.~~


tjhei:
This doesn't do anything on the testers, so this is safe to do.

PR #3294, opened 2019-11-19
Move ownership and responiblity of particle world to simulator

MFraters:
Currently the particle world is owned by the particle post-processor. This may have made sense in the beginning, but since particles can be active now it seems to make more sense to me for them to be part of the simulator. Furthermore, I am developing a plugin which needs access to up to date particle information. Currently this is only the case if the user provides that post processor after the particle post processor because the particle post-processor also moves and updates the particles.

Although further reorganization might we wanted in the future, this pull request is limited to just letting the simulator own the particle world and update the particles, so that the order given by the user of the post-processors do not matter any more.

I am still looking at a bug (there seems to be a subscriptor which is not deleted at the end of the simulation) and some of the code can probably be improved, but I thought it might be good to make this pull request now so that changes can be discussed early on.

And thanks @gassmoeller for the help and advice already giving :)

### For all pull requests:

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -28,6 +28,7 @@
 #include <aspect/stokes_matrix_free.h>
 #include <aspect/mesh_deformation/interface.h>
 #include <aspect/citation_info.h>
+#include <aspect/postprocess/particles.h>
 
 #ifdef ASPECT_USE_WORLD_BUILDER
 #  include <world_builder/world.h>
@@ -190,7 +191,7 @@ namespace aspect
                    nullptr),
 #endif
     boundary_heat_flux (BoundaryHeatFlux::create_boundary_heat_flux<dim>(prm)),
-
+    particle_world(nullptr),
     time (numbers::signaling_nan<double>()),
     time_step (numbers::signaling_nan<double>()),
     old_time_step (numbers::signaling_nan<double>()),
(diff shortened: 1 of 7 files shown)


MFraters:
The tests work again and the code is ready for review.

source/simulator/core.cc
```diff
@@ -396,6 +397,16 @@ namespace aspect
     postprocess_manager.initialize_simulator (*this);
     postprocess_manager.parse_parameters (prm);
 
+    if (postprocess_manager.template has_matching_postprocessor<Postprocess::Particles<dim> >())
```
gassmoeller: Very nice use of this function!

source/simulator/simulator_access.cc
```diff
@@ -720,6 +720,25 @@ namespace aspect
...
     return simulator->postprocess_manager;
   }
+
+
+  template <int dim>
+  const Particle::World<dim> &
+  SimulatorAccess<dim>::get_particle_world() const
+  {
+    Assert (simulator->particle_world.get() != nullptr,
+            ExcMessage("You can not call this function if no such model is actually available."));
```
gassmoeller:
```suggestion
            ExcMessage("You can not call this function if there is no particle world."));
```

source/simulator/simulator_access.cc
```diff
@@ -720,6 +720,25 @@ namespace aspect
...
+    return *simulator->particle_world.get();
+  }
+
+  template <int dim>
+  Particle::World<dim> &
+  SimulatorAccess<dim>::get_particle_world()
+  {
+    Assert (simulator->particle_world.get() != nullptr,
+            ExcMessage("You can not call this function if no such model is actually available."));
```
gassmoeller:
```suggestion
            ExcMessage("You can not call this function if there is no particle world."));
```

gassmoeller:
Oh and could you squash your commits? There are some that would not be useful to have in master.

MFraters:
Thanks for the quick review! I have addressed all your comments.

PR #3297, opened 2019-11-20
Read files from url using libdap

kneumiller:

We added a function in utilities to allow Aspect to read files from a url. If the user has installed libdap and has set "read libdap" to ON when they run CMake then anytime a url is passed in as the filename Aspect will attempt to use libdap to read the requested file.
Adiabatic_boundary was modified to read files similar to how other functions do.

* [ X ] I have read the guidelines in our [CONTRIBUTING.md](../CONTRIBUTING.md) document.

### For all pull requests:

  **We created a .prm file to test the reading from a url.**


gassmoeller:
Hi @kneumiller,
Thanks for suggesting to merge this into ASPECT! This looks like a more complex addition and it will likely take some time until somebody can fully review it, but I have some high-level question you could help me with / could address in the meantime:
- Could you give me an idea on the background on this work? I.e. which data will be hosted somewhere, and how can we access it? How many people are already using this feature? I am just trying to get a sense for the scope of the application of this. I feel like we discussed this on the mailing list or forum a while ago, but could you point me to that discussion again?
- Please avoid adding new directories to the top-level directory of ASPECT that are not absolutely necessary. Smaller additions like scripts and libraries can go into `contrib/`
- Also there are some files in subdirectories that should likely not be in our main version (e.g. source/INSTALL that is very specific to your addition. If you need installation instructions for your module, please add them to the manual in `doc/manual/manual.tex`).
- You have a bunch of `.DS_store` files that are added by the MacOS file explorer into every directory. Please remove those.
- You indented the code, but either with a wrong version of astyle or not with our common indentation rules. Please make sure you are using astyle 1.04, and use `make indent` in the build directory to do the indentation.
- It looks like you modified the `initial_temperature/adiabatic_boundary` plugin quite heavily. Is the functionality still the same?

I will try to come back to this as soon as possible, but in the worst case it might have to wait until after AGU. Maybe someone else can help out with the review?

tjhei:
Thanks for working on extending ASPECT in this direction. To add to what Rene said:
1. Where can I learn about "libdap", and what does it more than just performing a http get?
2. What kind of servers exist and why is a any webserver not good enough?
3. You will need to clean up many of the files you added. For example, you must have included a .so file by accident.
4. I think I mentioned that before: I don't like adding a unit test that does a http request every time we run the tests.

tjhei:
> This looks like a more complex addition and it will likely take some time until somebody can fully review it

Maybe we can also break this down into smaller pieces that are easier to review?

tjhei:
I tried configuring and I am getting the following output:
```
$ cmake -D LIBDAP=ON .
[...]
-- Could NOT find LibXml2 (missing: LIBXML2_LIBRARY LIBXML2_INCLUDE_DIR) 
-- Found CURL: /usr/lib/x86_64-linux-gnu/libcurl.so (found version "7.47.0")  
-- libdap found at LIBDAP_CLIENT_LIBRARY-NOTFOUND;LIBDAP_LIBRARY-NOTFOUND
```
This means the find cmake script incorrectly always returns success. We will need to fix that.

tjhei:
I also think having to find libxml separately is a problem (it should be discovered from the libdap installation). Anyways, now I am getting the following compile error:
```

In file included from ../source/utilities.cc:26:
In file included from /ssd/spack/opt/spack/linux-ubuntu16.04-broadwell/gcc-5.4.0/libdap4-3.20.4-2dzjmq6onqavz3cc5gbdeh3jgppkv7lu/include/libdap/Connect.h:62:
In file included from /ssd/spack/opt/spack/linux-ubuntu16.04-broadwell/gcc-5.4.0/libdap4-3.20.4-2dzjmq6onqavz3cc5gbdeh3jgppkv7lu/include/libdap/DAS.h:46:
In file included from /ssd/spack/opt/spack/linux-ubuntu16.04-broadwell/gcc-5.4.0/libdap4-3.20.4-2dzjmq6onqavz3cc5gbdeh3jgppkv7lu/include/libdap/AttrTable.h:54:
/ssd/spack/opt/spack/linux-ubuntu16.04-broadwell/gcc-5.4.0/libdap4-3.20.4-2dzjmq6onqavz3cc5gbdeh3jgppkv7lu/include/libdap/XMLWriter.h:33:10: fatal error: 'libxml/xmlwriter.h' file not found
```
It looks like the libxml include path is not added correctly. Somehow, my include files are in ``include/libxml2/libxml/xmlwriter.h``. This fixes it for me:
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 6c18b2767..364d1a9b5 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,7 +479,7 @@ IF(LIBDAP)
        IF(${LIBDAP_FOUND})
          FIND_PACKAGE(LibXml2)
          FIND_PACKAGE(CURL)
-         INCLUDE_DIRECTORIES(${LIBDAP_INCLUDE_DIR} ${LIBXML2_INCLUDE_DIR} ${CURL_INCLUDE_DIR})
+         INCLUDE_DIRECTORIES(${LIBDAP_INCLUDE_DIR} ${LIBXML2_INCLUDE_DIR} ${LIBXML2_INCLUDE_DIR}/libxml2 ${CURL_INCLUDE_DIR})
          TARGET_LINK_LIBRARIES(${TARGET} ${LIBDAP_LIBRARIES} ${LIBXML2_LIBRARIES} ${CURL_LIBRARIES})
          ADD_DEFINITIONS(-DHAVE_LIBDAP=TRUE)
            MESSAGE(STATUS "libdap found at ${LIBDAP_LIBRARIES}")
```

cmake/modules/FindLIBDAP.cmake
```diff
@@ -0,0 +1,42 @@
...
+UNSET (LIBDAP_CLIENT_LIBRARY CACHE)
+
+
+FIND_PATH(LIBDAP_INCLUDE_DIR 
+  NAMES Connect.h
+  HINTS ${LIBDAP_INC} ${LIBDAP_DIR}/include/libdap 
+  )
+
+FIND_LIBRARY(LIBDAP_LIBRARY 
+  NAMES libdap.a
```
tjhei: We should link to the dynamic library instead

cmake/modules/FindLIBDAP.cmake
```diff
@@ -0,0 +1,42 @@
...
+  NAMES libdap.a 
+  HINTS ${LIBDAP_LIB} ${LIBDAP_DIR}/lib
+  )
+  
+FIND_LIBRARY(LIBDAP_CLIENT_LIBRARY
+  NAMES libdapclient.a
+  HINTS ${LIBDAP_LIB} ${LIBDAP_DIR}/lib
+  )
+
+SET(LIBDAP_FOUND TRUE)
```
tjhei: You will need to check if the FIND_* calls succeed.

tests/url_reader.prm
```diff
@@ -0,0 +1,102 @@
...
+  set Number of fields = 1
+end
+
+# We use the ascii data plugin to read in the file created with geomIO.
+subsection Initial composition model
+  set Model name = ascii data
+
+  subsection Ascii data model
+    set Data directory       = 
+    set Data file name       = http://localhost:8080/opendap/data/csv/geomIO_urlTest.csv
```
tjhei: how is this going to work without a local http server?

jgallagher59701:
Rene, Timo,
I'll break these down into a set of actions and we'll work on them. To break this down into several PRs, how does this sound:

1. A PR that modifies the cmake scripts to optionally look for libdap and its libxml2, libcurl and bison/flex dependencies.
2. A PR that combines documentation and the modifications to the initial_temperature/adiabatic_boundary plugin, including a way to test the remote access functionality without actually using a remote server (the latter will free the unit test from using HTTP).

If this seems reasonable, I'll take the comments from 20 and 21 Nov and turn them into tickets. We can then close this PR and issue two new ones when the work is done.

Since those things all seem reasonable - and thanks very much for the time needed to read over what we have done and make those - we'll start on them now. Let me know if the two PR breakdown seems like enough or if would prefer something else.

Thanks and Happy Holidays,
James

cmake/modules/FindLIBDAP.cmake
jgallagher59701: Yup, fix in the queue

cmake/modules/FindLIBDAP.cmake
jgallagher59701: ditto

tests/url_reader.prm
jgallagher59701: We can use a file:// URL to local test data.

tjhei: perfect. You can use ``$ASPECT_SOURCE_DIR``, which will be expanded, see https://github.com/geodynamics/aspect/blob/master/tests/ascii_data_adiabatic_conditions.prm#L13 for example

tjhei:
> Since those things all seem reasonable - and thanks very much for the time needed to read over what we have done and make those - we'll start on them now. Let me know if the two PR breakdown seems like enough or if would prefer something else.

sounds good, thank you! :+1:

jgallagher59701:
FYI. The ticket is here - I made 'story' with a bunch of sub-tasks. I'm pretty sure it's accessible to the world. https://opendap.atlassian.net/browse/HK-556

gassmoeller:
Am I correct in assuming we merged all the changes in this PR in the other PRs? Can this one be closed?

tjhei:
I think so. Closing.

PR #3298, opened 2019-11-21
Unify the output of Stokes and Advection solver exceptions

gassmoeller:
I found this very useful while trying to debug a crash of the advection solver. It unifies the error messages and the output of the convergence history between Stokes and Advection solver, and additionally adds to the output the residual which would be considered converged (previously you would not know if the solver nearly converged, or was still orders of magnitude off).


gassmoeller:
@MFraters asked for a test, so I refactored the error message into a separate function. The Stokes error message is already tested in `n_expensive_stokes_solver_steps` and the advection error message is now tested in `advection_solver_fail`.

PR #3300, opened 2019-11-25
Add spack installation file

gassmoeller:
This adds a 'spack' configuration file into `contrib/`, which allows for an easy installation on systems that already contain spack. `spack install` executed in that folder will install ASPECT and all its dependencies, `spack install --only dependencies` will install only dependencies (dealii and the necessary packages). The benefit of this file over doing `spack install aspect` is that I can explicitly disable a lot of options in our dependencies, which greatly reduces the time for compiling, and the likelihood of compiler problems (e.g. version 5.3.0 of suite-sparse does not seem to work with the other packages so I had to manually specify a version). For now I would keep this hidden, but perspectively, I think spack is the more sustainable installation option compared to candi (but that is a discussion for a different day).

### DIFF
--- a/contrib/spack/spack.yaml
+++ b/contrib/spack/spack.yaml
@@ -1,0 +1,12 @@
+# This file can be used to install either ASPECT, or only its dependencies
+# using the 'spack' package manager. Usage: `spack install` in this directory
+# will install the development version of ASPECT with minimal dependencies.
+# `spack install --only dependencies` will only install ASPECT's dependencies,
+# after which ASPECT can be compiled from this folder as usual.
+
+spack:
+  specs:
+  - aspect@develop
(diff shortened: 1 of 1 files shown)


tjhei:
I don't mind merging this and I agree that spack is neat, but I am consistently frustrated by random compile failures (because some package gets updated similar to what you saw), bloat (compiling way too much stuff), mpi configuration on clusters, etc.. So I don't think I am switching any time soon. ;-)

gassmoeller:
I know what you mean. I think spack has some conflicts between usage for system administrators (who want to compile all optional features so that they only need a single package for all users) vs. users (who want to compile the minimally necessary dependencies to minimize compile time and incompatibilities). What I find neat about this .yaml file approach is that we can manually fix versions and features in all of the dependencies, so hopefully this would be more useful than the general aspect package.

PR #3304, opened 2019-11-30
Replace dynamic_casts by proper function calls

gassmoeller:
This is a follow-up to #3173 that replaces some more dynamic casts by the functions in the `Plugins` namespace. I also moved some geometry model checks into more appropriate functions (only check once per model run instead of every call).

### DIFF
--- a/source/boundary_composition/box.cc
+++ b/source/boundary_composition/box.cc
@@ -39,12 +39,6 @@ namespace aspect
                           const Point<dim> &/*position*/,
                           const unsigned int compositional_field) const
     {
-      // verify that the geometry is a box since only for this geometry
-      // do we know for sure what boundary indicators it uses and what they mean
-      Assert (Plugins::plugin_type_matches<const GeometryModel::Box<dim>>(this->get_geometry_model()),
-              ExcMessage ("This boundary model is only implemented if the geometry is "
-                          "a box."));
-
       Assert (boundary_indicator<2*dim, ExcMessage ("The given boundary indicator needs to be less than 2*dimension.."));
       return composition_values[boundary_indicator][compositional_field];
     }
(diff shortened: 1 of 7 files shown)


source/boundary_composition/two_merged_boxes.cc
```diff
@@ -167,6 +160,12 @@ namespace aspect
     void
     TwoMergedBoxes<dim>::initialize()
     {
+      // verify that the geometry is a box since only for this geometry
+      // do we know for sure what boundary indicators it uses and what they mean
+      AssertThrow (Plugins::plugin_type_matches<const GeometryModel::TwoMergedBoxes<dim>>(this->get_geometry_model()),
```
MFraters: nice change, moving it to init and making it AssertThrow :)

source/boundary_temperature/box.cc
```diff
@@ -149,6 +142,13 @@ namespace aspect
               default:
                 Assert (false, ExcNotImplemented());
             }
+
+          // verify that the geometry is a box since only for this geometry
+          // do we know for sure what boundary indicators it uses and what they mean
+          AssertThrow (Plugins::plugin_type_matches<const GeometryModel::Box<dim>>(this->get_geometry_model()),
```
MFraters: It might be nice to make sure that this check is in the same place for all plugins. I think the init function is the best place to put this since it is called earlier. The downside of that would be that extra functions need to be added to the plugins. What do you think?

source/boundary_composition/box.cc
```diff
@@ -41,8 +41,7 @@ namespace aspect
     {
       // verify that the geometry is a box since only for this geometry
       // do we know for sure what boundary indicators it uses and what they mean
-      Assert (dynamic_cast<const GeometryModel::Box<dim>*>(&this->get_geometry_model())
-              != nullptr,
+      Assert (Plugins::plugin_type_matches<const GeometryModel::Box<dim>>(this->get_geometry_model()),
```
tjhei: this could also become an AssertThrow and only be done once.

source/boundary_temperature/box.cc
tjhei: parse_parameters() is also okay with me as all objects are already constructed. Both are better than during the actual usage.

gassmoeller:
I addressed @tjhei's one comment. 
About the appropriate place for these asserts: I feel like we are overengineering things a bit. It would be nice to have a consistent place for the asserts (initialize() or parse_parameters()), but in the end it does not really matter since they are equivalent and each plugin could handle it like it wants to. I just went with what fit better into the existing plugin structure.

MFraters:
Yes, you are right. I am fine with merging this when the tests pass.

PR #3306, opened 2019-12-02
Add viscoelasticity to the visco plastic material model

naliboff:
This PR adds the option to use viscoelasticity in the visco plastic material model. Either before the PR is merged or afterward, the viscoelastic plastic material can be removed as all of its functionality will be covered in the visco plastic model. 

At the moment, this is a work in progress and I am opening a PR only for initial review and testing. Note that the documentation in the compute_isostrain_viscosities function has been rewritten with the hope that the order of operations is clearer.


naliboff:
Note that this PR is derived from an early version of #3292 (see first commit) and will be updated once that PR is merged.

naliboff:
@tjhei - I think this is ready for a review if you have time to take a look. My preference would be to merge this PR and then rebase #3119 so we can test it with strain-softening and elasticity.

source/material_model/visco_plastic.cc
```diff
@@ -140,19 +140,58 @@ namespace aspect
...
+                             std::max(std::sqrt(std::fabs(second_invariant(2. * (deviator(strain_rate)) + stress_old / (elastic_shear_moduli[j] * dte)))), min_strain_rate) );
+            }
+
+          // Step 1: viscous behavior
+
+          // Step 1a: compute viscosity from diffusion creep law
           const double viscosity_diffusion = diffusion_creep.compute_viscosity(pressure, temperature_for_viscosity, j);
 
-          // Compute visocisty from dislocation creep law
+          // Step 1b: Compute visocisty from dislocation creep law
```
tjhei:
```suggestion
          // Step 1b: Compute viscosity from dislocation creep law
```

source/material_model/rheology/elasticity.cc
```diff
@@ -180,10 +180,10 @@ namespace aspect
 
         // Functionality to average the additional RHS terms over the cell is not implemented.
         // This enforces that the variable 'Material averaging' is set to 'none'.
-        AssertThrow(this->get_parameters().material_averaging == MaterialModel::MaterialAveraging::none,
-                    ExcMessage("Material models with elasticity cannot be used with "
-                               "material averaging. The variable 'Material averaging' "
-                               "in the 'Material model' subsection must be set to 'none'."));
+        //AssertThrow(this->get_parameters().material_averaging == MaterialModel::MaterialAveraging::none,
```
tjhei: what do we want to do about this? Can we remove this change and tackle this separately?

source/material_model/visco_plastic.cc
```diff
@@ -178,49 +217,81 @@ namespace aspect
...
-          // If the viscous stress is greater than the yield stress, indicate we are in the yielding regime.
-          const double viscous_stress = 2. * viscosity_pre_yield * edot_ii;
-          if (viscous_stress >= yield_stress)
-            composition_yielding[j] = true;
+          // Step 4b: Select the strain rate and stress invariants value to use for rescaling viscosity the plastic yield stress
+          double current_edot_ii;
+          double current_stress;
+          if (use_elasticity == false)
+            {
+              current_edot_ii = edot_ii;
```
tjhei: I would assign these values unconditionally (two lines up)

source/material_model/visco_plastic.cc
```diff
@@ -178,49 +217,81 @@ namespace aspect
...
+              current_stress = viscous_stress;
+            }
+          else
```
tjhei: and only keep this as ``if (use_elasticity)``

source/material_model/visco_plastic.cc
```diff
@@ -593,6 +693,12 @@ namespace aspect
                              "Using a pressure gradient of 32436 Pa/m, then a value of "
                              "0.3 $K/km$ = 0.0003 $K/m$ = 9.24e-09 $K/Pa$ gives an earth-like adiabat."
                              "Units: $K/Pa$");
+
+          prm.declare_entry ("Include viscoelasticity","false",
```
tjhei:
```suggestion
          prm.declare_entry ("Include viscoelasticity", "false",
```

source/material_model/visco_plastic.cc
```diff
@@ -593,6 +693,12 @@ namespace aspect
...
+                             Patterns::Bool (),
+                             "Whether to inlcude viscoelasticity in the rheological formulation. ");
```
tjhei:
```suggestion
                             "Whether to include viscoelasticity in the rheological formulation.");
```

source/material_model/visco_plastic.cc
```diff
@@ -815,6 +929,107 @@ namespace aspect
                                    "standard viscosity rescaling method (concretely, values $n_y>20$ "
                                    "are large enough)."
                                    "\n\n "
+                                   "The visco-plastic rheology described above may also be modified to include "
+                                   "viscoelastic deformation, thus producing a viscoelastic plastic constitutive "
+                                   "relationship. "
+                                   "\n\n "
+                                   "The viscoelastic rheology behavior takes into account the elastic shear "
```
tjhei: Independent of this PR, but can I get you to write a section in the manual with these formulas? Something similar to section 2.12 or 2.13. This might be a better location for references and this lengthy description.

source/material_model/visco_plastic.cc
```diff
@@ -140,19 +140,58 @@ namespace aspect
...
       std::vector<bool> composition_yielding(volume_fractions.size());
 
+      std::vector<double> composition_viscosities(volume_fractions.size());
+
+      std::vector<double> viscosities_ve(volume_fractions.size());
+      std::vector<double> viscosities_vep(volume_fractions.size());
+
+      std::vector<double> elastic_shear_moduli(volume_fractions.size());
+
+      std::vector<double> stresses_ve(volume_fractions.size());
```
tjhei: Can you fill these variables with signaling NaNs?

tjhei:
> My preference would be to merge this PR and then rebase #3119 so we can test it with strain-softening and elasticity.

I think we can tackle these independently as well: I will prepare a simple version of the prescribed terms (without the material model updates).

source/material_model/rheology/elasticity.cc
naliboff: Yes, lets tackle this separately. None of the tests use cell averaging, so nothing should break by re-introducing this exception.

source/material_model/visco_plastic.cc
naliboff: Yes, will do. I've opened an issue (#3365) and assigned myself.

naliboff:
@tjhei - Thanks for the review and I believe I've addressed all of the comments.

Regarding the correctness of the implementation relative to other material models, all of the new test results were copied directly from existing tests that either used the `viscoelastic_plastic` or `viscoelastic` material model. 

Only two of six test results changed (see `update test results` commit) and the changes are fairly minor. The ordering of how compositional fields are averaged is slightly different between material models, and I'm actually surprised most of the tests passed without modification.

@dansand - are you satisfied with the current structure and are the new benchmarks behaving roughly as expected? 

Reviews from any other frequent users of `visco_plastic` would be helpful, even if only to comment on the revised documentation.

naliboff:
@gassmoeller - I'm in the process of updating some of the benchmarks to use `visco_plastic` instead of `viscoelastic_plastic`. In doing this, it will actually break some of the` viscoelastic_plastic` tests tied to the benchmarks.

Two options for how to proceed:

1.  Remove the viscoelastic_plastic material model and associated tests in the same commit.
2. Update the viscoelastic_plastic test parameter files and then remove everything in a follow-up PR.

Preferences for one of these two options or some other solution? No strong preference on my side.

gassmoeller:
What about option 3: Update the test results according to the new benchmark results? I would like to keep the tests, and I assume that while you update the benchmarks to the visco_plastic model you check that the new (slightly changed) results are still correct?

naliboff:
@gassmoeller - that seems like a good solution.

> I assume that while you update the benchmarks to the visco_plastic model you check that the new (slightly changed) results are still correct?

For the most part I've been running all the existing benchmarks with every new commit that changes source code, but I just checked on one of the benchmarks again last night and found some spurious behavior.

My famous last words from above
> Only two of six test results changed (see update test results commit) and the changes are fairly minor
were clearly ill-informed ;)

I'll check back in when I'm satisfied with all the benchmarks and the test suites have been updated accordingly.

naliboff:
Good news, the spurious benchmark result was the result unphysical artifacts in the viscoelastic stress field arising from fixed compositional boundary conditions. Removing these boundary conditions produces a perfect fit to the benchmark results. The order of averaging in `viscoelastic_plastic` somehow manages to prevent propagation of the edge effects observed when using` visco_plastic` for the same problem.

@gassmoeller @tjhei - I'm going to prepare a series pull requests that update tests and benchmarks for `viscoelastic` and `viscoelastic_plastic`. These changes will help simplify this PR and make it much clearer what test results change between material models.

naliboff:
FYI, I re-ran one of the full VEP benchmarks earlier today and am unfortunately getting a deviation from the analytical results. I've traced this back to a rebase, but I'm going to need to spend a bit of time figuring out exactly what the issue is. In any case, this will be a good opportunity to go back and redo how the new tests are added, which was not done in an ideal manner in the last few commits.

naliboff:
Resolved the issue and the aforementioned benchmarks are now passing. See line 260-263 in `visco_plastic.cc` for an explanation of the issue, which was related to differences between how viscous and viscoelastic-plastic stresses are calculated and treated in the `Drucker Prager` rheology module.

In addition to the fix, I've also significantly cleaned up the commit history and added more documentation. @dansand and I are discussing one additional vep benchmark, but the current structure should be by and large complete. 

@tjhei - As soon as this is merged, I think we can safely move on with the associated plasticity testing and PRs. Sorry that this has held up those PRs and related testing!

naliboff:
@gassmoeller - This is now ready for review, as all the relevant tests and benchmarks have been modified or added. In principle, we should now be able to delete `viscoelastic_plastic`.

naliboff:
@gassmoeller - One point I forgot to add above is that I am not happy with how unruly and extensive the documentation for this material model now is. I think quite a bit can be removed or moved into the relevant rheology module sections, and more in-depth explanations added in the manual. #3365 is a reminder for this task.

source/material_model/visco_plastic.cc
```diff
@@ -178,49 +216,80 @@ namespace aspect
...
-          // Select if yield viscosity is based on Drucker Prager or stress limiter rheology
+          // Step 4c: select if yield viscosity is based on Drucker Prager or stress limiter rheology
           double viscosity_yield = viscosity_pre_yield;
           switch (yield_type)
             {
               case stress_limiter:
               {
+                //Step 4c-1: always rescale the viscosity back to the yield surface
                 const double viscosity_limiter = yield_stress / (2.0 * ref_strain_rate)
                                                  * std::pow((edot_ii/ref_strain_rate), 1./exponents_stress_limiter[j] - 1.0);
```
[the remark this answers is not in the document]
naliboff: @gassmoeller - hmmm, ood question. I have not tested the stress limiter with elasticity and I'm not sure there is a benchmark I can compare with. @cedrict and I tested this for simple visco-plastic shear band problems originally, and perhaps the best thing to do is enforce that the VEP rheology is only used in combination with the `Drucker Prager` rheology, module for which we have benchmarks.

source/material_model/visco_plastic.cc
```diff
@@ -178,49 +216,80 @@ namespace aspect
...
-            composition_yielding[j] = true;
+          // Step 4b: Select the strain rate and stress invariants value to use for rescaling viscosity the plastic yield stress
+          double current_edot_ii = edot_ii;
+          double current_stress = viscous_stress;
+          if (use_elasticity)
+            {
+              // The viscoelastic strain rate is divided by 2 here as the Drucker Prager
+              // viscosity calculation assumes stress = 2 * viscosity * strain_rate_invariant,
+              // whereas the combined the viscoelastic + viscous stresses do not include the
+              // 2x factor.
```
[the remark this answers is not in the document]
naliboff: @gassmoeller - We absolutely could do `stresses_ve[j] = 2. * viscosities_ve[j] * edot_ii_ve` and not modify the strain rate, but then this affects the stress in line 284 (prior to your commit):
  `if (current_stress >= yield_stress)`

So, I think adding the factor of 2 in at line 33 would means are stresses would be 2x higher than they are currently defined here and in `viscoelastic_plastic`. However, I may be wrong about this and did not just read through the code logic correctly (definitely easier to read after your commit 👍  ).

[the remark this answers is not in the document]
naliboff: Regarding the stress limiter, per the comment above I think we should just restrict that yielding option to only work with a visco plastic rheology (i.e., without elasticity).

naliboff:
@gassmoeller - thanks for the review and commit! See the responses above in regards to the remaining issues.

gassmoeller:
Sorry, I broke a line with my PR last week. I fixed it and also resolved the remaining issues and rebased to master. This should be good to go, but please take another look. I asserted that the stress limiter is not combined with VEP for now. I also now understand why the strain rate is handled differently (the x2 only applies to the viscous stress part, but not the elastic stress part, which is correctly considered in the visco-elastic part, but not the drucker prager plasticity, which is why the strain rate has to be modified to compute the correct plastic behavior, just documenting here and in the code for future reference).

naliboff:
@gassmoeller - Just had another look through and agree everything looks to go. Thank you for the review and last two commits! I'll let you merge when ready.

naliboff:
@gassmoeller - Once this is merged, is there any reason to wait on removing `viscoelastic_plastic` and all associated files? The benchmarks using `viscoelastic_plastic` have all been updated here to use `visco_plastic`.

gassmoeller:
Ok, I will merge this.

> Once this is merged, is there any reason to wait on removing viscoelastic_plastic and all associated files? 

No, please remove viscoelastic_plastic. Removing duplication is always good.

naliboff:
Follow-up: #3439 removes the viscoelastic plastic material model and test suite.

PR #3307, opened 2019-12-03
Fix particle exclude properties

MFraters:
When excluding we currently only test with gnuplot output. I somehow missed that when turning on vtu output, an assert is thrown: 
```
An error occurred in line <6002> of file </home/fraters/Documents/post-doc/2019-02-01-magali/code/dealii/2019-10-03-dealii-manual/dealii/source/base/data_out_base.cc> in function
    void dealii::DataOutBase::write_pvtu_record(std::ostream&, const std::vector<std::__cxx11::basic_string<char> >&, const std::vector<std::__cxx11::basic_string<char> >&, const std::vector<std::tuple<unsigned int, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, dealii::DataComponentInterpretation::DataComponentInterpretation> >&)
The violated condition was: 
    last_component < n_data_sets
Additional information: 
    Index 14 is not in the half-open range [0,13).
```

The problem can be reproduced by changing `gnuplot` to `vtu` in the `tests/particle_exclude_one_property.prm`.

I solved the problem by limiting the push back to dataset_names. I have added several tests to show that it works now for several cases of excluding different amount of properties and different types of properties. 

### For all pull requests:

* [x] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../tests/) directory.


source/postprocess/particles.cc
```diff
@@ -97,16 +97,16 @@ namespace aspect
               }
 
             // If the property has dim components, we treat it as vector
-            if (n_components == dim)
+            if (n_components == dim && property_index_to_output_index[field_position] > 0)
```
gassmoeller: I do not see why this should fix the error. If `property_index_to_output_index[field_position] == 0` we either have the first field (in which case we *do* want to add the field name to the list of `vector_datasets`), or we have a field that should be excluded (in which case we never get here, because line 84 is triggered). To me it looks like there is an off-by-one error somewhere else, and you fix it by omitting something else here. Have you checked in the vtu output if the name of a particle property is correct for the data that is written?

MFraters:
I have addressed the comments as discussed.

PR #3308, opened 2019-12-03
Replace some more dynamic_casts

gassmoeller:
Follow-up to #3304. Replace some more dynamic_casts and simplify some checks. I replaced checks for specific geometry models with `natural_coordinate_system` wherever I was sure the functionality would work with any geometry model of that coordinate system someone could come up with. When I was unsure if that was true I left checks for specific geometry model plugins (like the initial topography).

### DIFF
--- a/include/aspect/boundary_velocity/gplates.h
+++ b/include/aspect/boundary_velocity/gplates.h
@@ -24,7 +24,6 @@
 
 #include <aspect/boundary_velocity/interface.h>
 #include <aspect/simulator_access.h>
-#include <aspect/compat.h>
 
 #include <array>
 #include <deal.II/base/function_lib.h>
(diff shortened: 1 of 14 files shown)


gassmoeller:
I have addressed your comment in the gplates plugin.

> are you intentionally excluding ellipsiodal coordinates here? Would ellipsoidal need a different case?

Yes, I think it was a mistake to allow this option for the ellipsoidal coordinates before (we probably did not think about it when it was introduced). The function that converts the velocity unit vectors from curved to cartesian always uses `Utilities::Coordinates::spherical_to_cartesian_vector`. Therefore if you would prescribe velocities that are purely in longitude direction, with this function  the computed velocities would have a radial component for the ellipsoid (material would follow the spherical surface, not the ellipsoid surface). We would need a new utility function to convert between ellipsoidal unit vectors and cartesian unit vectors before we allow this.

PR #3312, opened 2019-12-10
Add adiabatic surface temperature to S20RTS cookbook

gassmoeller:
This is not exactly a bug for this cookbook, but if you modify it for applications that rely on the reference profile we want to set the correct adiabatic surface temperature. I feel like having the wrong adiabatic profile is a classical bug in many of our models, we need to find a better way than this PR, but this is a start.

### DIFF
--- a/cookbooks/S20RTS.prm
+++ b/cookbooks/S20RTS.prm
@@ -15,9 +15,12 @@ set Output directory                       = output-S20RTS
 
 # The following variables describe how the pressure should
 # be normalized. Here, we choose a zero average pressure
-# at the surface of the domain 
+# at the surface of the domain. The 'Surface pressure' and
+# 'Adiabatic surface temperature' are used to compute
+# the adiabatic reference profile.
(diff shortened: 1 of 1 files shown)


cookbooks/S20RTS.prm
```diff
@@ -15,9 +15,12 @@ set Output directory                       = output-S20RTS
 
 # The following variables describe how the pressure should
 # be normalized. Here, we choose a zero average pressure
-# at the surface of the domain 
+# at the surface of the domain. The surface pressure and
+# Adiabatic surface temperature are also used to compute
```
naliboff:
```suggestion
# adiabatic surface temperature are also used to compute
```

naliboff: Sorry, missed one thing: Consistent capitalization of `Adiabatic`, or capitalize `Surface pressure` if you intended to match parameter syntax.

gassmoeller:
I fixed the capitalization and clarified that I use the parameter names.

PR #3315, opened 2019-12-13
Add ref, remove unused parameter

anne-glerum:
- added reference
- removed unused parameter
- moved AssertThrow to before looping over checked parameter

### DIFF
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -620,7 +620,6 @@ namespace aspect
           strain_rheology.parse_parameters(prm);
 
           // Reference and minimum/maximum values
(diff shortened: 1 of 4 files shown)


source/material_model/visco_plastic.cc
```diff
@@ -620,7 +620,6 @@ namespace aspect
           strain_rheology.parse_parameters(prm);
 
           // Reference and minimum/maximum values
-          reference_T = prm.get_double("Reference temperature");
```
gassmoeller: Is there still a declaration of this variable in visco_plastic.h? If so, please remove as well.

anne-glerum: Good point. Removed.

PR #3316, opened 2019-12-17
Replace most dynamic_casts

gassmoeller:
Follow up to #3173, #3304 and #3308. I had some time in the plane :smile:. All of this should be purely refactoring/replacements with two exceptions: A few code movements (moving asserts from often called functions to initializations), and in the AsciiData classes I now allow more geometry models (see utilities.cc). The class was written when we did not have the new geometries, but there is no reason it should not work for them.


tjhei:
Looks reasonable but 54 tests fail with "Not a valid geometry model for the initial conditions modelharmonic perturbation." (notice the missing space in the error message?)

gassmoeller:
I fixed the tests (some logic bugs about if vs else-if). The new test results are caused by the replacement of `3.14159` with `numbers::PI` in  source/initial_temperature/spherical_shell.cc.

Issue #3322, opened 2020-01-04
unaccounted timing information


tjhei:
We are missing a large amount of compute time that is not accounted for in some runs. Investigate what is missing and add some timing sections.
Example:
```
+----------------------------------------------+------------+------------+
| Total wallclock time elapsed since start     |       181s |            |
|                                              |            |            |
| Section                          | no. calls |  wall time | % of total |
+----------------------------------+-----------+------------+------------+
| Assemble Stokes system rhs       |         2 |     0.371s |       0.2% |
| Assemble temperature system      |         2 |      20.9s |        12% |
| Build temperature preconditioner |         2 |    0.0779s |         0% |
| Initialization                   |         1 |      1.38s |      0.76% |
| Postprocessing                   |         2 |      73.8s |        41% |
| Setup dof systems                |         1 |     0.325s |      0.18% |
| Setup initial conditions         |         1 |     0.555s |      0.31% |
| Setup matrices                   |         1 |     0.173s |         0% |
| Solve Stokes system              |         2 |        43s |        24% |
| Solve temperature system         |         2 |   0.00334s |         0% |
+----------------------------------+-----------+------------+------------+
```
(23% unaccounted)

PR #3323, opened 2020-01-04
report prescribed advection field timer

tjhei:
part of #3322, now down to 12% unaccounted.

### DIFF
--- a/source/simulator/solver_schemes.cc
+++ b/source/simulator/solver_schemes.cc
@@ -155,6 +155,7 @@ namespace aspect
             }
 
           const double current_residual = solve_advection(AdvectionField::temperature());
+
           current_linearization_point.block(introspection.block_indices.temperature)
             = solution.block(introspection.block_indices.temperature);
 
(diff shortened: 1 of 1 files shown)


gassmoeller:
Could also be shortened to "Prescribe temperature" "Prescribe composition", and "Prescribe Stokes solution" or "Interpolate temperature", "Interpolate composition", "Interpolate Stokes solution"

tjhei:
I changed it to your first suggestion.

tjhei:
I don't really like the ordering of the timing sections (done alphabetically). I have a few suggestions to improve it:
1. Rename them as "Stokes: Assembly", "Stokes: build preconditioner", etc. to have things grouped reasonably well.
2. Add a feature to deal.II to not reorder things but keep them in the order they are first executed (might not be super useful).
3. Define a custom ordering and teach deal.II about that.
4. Rename things, for example using "1. xxx" to force a good numbering.
Thoughts?

gassmoeller:
I am not sure renaming would be a good idea, at least I have some scripts that will break (though if we decide that is the best way I am ok with it). Would it be possible to add an optional parameter to the timer class that defines the position of a section? Or to hand over a map when printing the sections that links section name to position?

PR #3327, opened 2020-01-06
www: update database.js tag to force loading new file

tjhei:
We will need to manually modify this url, otherwise the browser uses an
old cached version.

### DIFF
--- a/citing.html
+++ b/citing.html
@@ -70,7 +70,7 @@
       <div class="row">
         <div class="well">
 
-          <script src="database.js"></script>
+          <script src="database.js?t=2020-01-06"></script>
 
           <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


bangerth:
I might have missed some context, but can you explain why this is necessary? Does the `.jl` script actually parse the date argument? And *whose* browser would use an old cached version?

tjhei:
After I updated the .js file on the server, I could not make my browser reload the file as it is cached on disk (even ctrl+f5 etc). The argument just gets dropped on the server side but chrome won't cache the js permanently.

gassmoeller:
I do not understand the details here, but since it was necessary for Timo I do not see harm in merging it. Any objections?

PR #3330, opened 2020-01-15
do not include SUPG parameter in heat flux calculations

maxrudolph:
When calculating the heat flow across boundaries, ASPECT currently uses the maximum of the thermal conductivity and the stabilization parameter used for the energy equation. For the entropy viscosity scheme, this makes some sense but for SUPG, this parameter doesn't have the physical units of a thermal conductivity so to include in this way is an error. This code removes the stabilization parameter if the advection stabilization method is SUPG. I have tested this with one 3D spherical geometry model under Earth-like conditions and the boundary heat flux appears to be closer to what is computed using CitcomS - (~10^13 W instead of ~10^19 W).


tjhei:
awesome, yes. That makes a lot of sense. I was going to push a commit to fix indentation but you are not allowing this. Can you please indent?

maxrudolph:
Sorry about that - I fixed the indentation. So odd - I tried to do this previously but it's as if the commit never appeared.

tjhei:
one test ("supg") failed, can you update the test results please? You can download the diff from https://jenkins.geodynamics.org/blue/organizations/jenkins/aspect/detail/PR-3330/2/artifacts and apply it using ``patch -p1 changes-test-results.diff``.

tjhei:
thanks, looks good.

PR #3333, opened 2020-01-16
Libdap

kneumiller:
Part of the pull requests for OPeNDAP:
These new additions will search for the libdap libraries through cmake. Libdap is necessary for reading data from a url, which is what the "url reader" pull request adds.

* [ X ] I have read the guidelines in our [CONTRIBUTING.md](../CONTRIBUTING.md) document.

### For all pull requests:

* [ X ] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../tests/) directory.


cmake/modules/FindLIBDAP.cmake
```diff
@@ -0,0 +1,44 @@
...
+UNSET (LIBDAP_CLIENT_LIBRARY CACHE)
+
+
+FIND_PATH(LIBDAP_INCLUDE_DIR
+        NAMES Connect.h
+        HINTS ${LIBDAP_INC} ${LIBDAP_DIR}/include/libdap
+        )
+
+FIND_LIBRARY(LIBDAP_LIBRARY
+        NAMES libdap.dylib
```
tjhei: This seems to be OSX specific. Can you try if this works when using "NAMES libdap" ?

doc/modules/changes/20200116_kneumiller
```diff
@@ -0,0 +1,4 @@
+Added: A module in the cmake directory that looks for, and adds the path of, the libdap libraries.
+A lookup was also added in CMakeLists for the libdap, libxml, and curl libraries.
+<br>
+(Kodi Neumiller, 2020/01/16)
```
tjhei: please add a newline at the end.

CMakeLists.txt
```diff
@@ -479,6 +479,25 @@ ELSE()
 ENDIF()
 
 
+# Find the libdap package so that Aspect can connect to the OPeNDAP servers
+# Author: Kodi Neumiller
+# Added: 10/16/18
+IF(LIBDAP)
+  LIST(APPEND CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}/cmake/modules")
+  MESSAGE(STATUS "cmake module path = ${CMAKE_MODULE_PATH}")
```
tjhei: remove this

CMakeLists.txt
```diff
@@ -479,6 +479,25 @@ ELSE()
...
+  FIND_PACKAGE(LIBDAP)
+  IF(${LIBDAP_FOUND})
+    FIND_PACKAGE(LibXml2)
+    FIND_PACKAGE(CURL)
```
tjhei: Same here. Also, what if one of these checks fails?

CMakeLists.txt
```diff
@@ -479,6 +479,25 @@ ELSE()
...
 
+# Find the libdap package so that Aspect can connect to the OPeNDAP servers
+# Author: Kodi Neumiller
+# Added: 10/16/18
+IF(LIBDAP)
+  LIST(APPEND CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}/cmake/modules")
+  MESSAGE(STATUS "cmake module path = ${CMAKE_MODULE_PATH}")
+  FIND_PACKAGE(LIBDAP)
+  IF(${LIBDAP_FOUND})
+    FIND_PACKAGE(LibXml2)
```
tjhei: What if this check find a different libxml than libdap was configured with?

CMakeLists.txt
```diff
@@ -479,6 +479,25 @@ ELSE()
 ENDIF()
 
 
+# Find the libdap package so that Aspect can connect to the OPeNDAP servers
+# Author: Kodi Neumiller
+# Added: 10/16/18
+IF(LIBDAP)
```
tjhei: what does this IF do?

CMakeLists.txt
```diff
@@ -479,6 +479,25 @@ ELSE()
...
+  LIST(APPEND CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}/cmake/modules")
+  MESSAGE(STATUS "cmake module path = ${CMAKE_MODULE_PATH}")
+  FIND_PACKAGE(LIBDAP)
+  IF(${LIBDAP_FOUND})
+    FIND_PACKAGE(LibXml2)
+    FIND_PACKAGE(CURL)
+    INCLUDE_DIRECTORIES(${LIBDAP_INCLUDE_DIR} ${LIBXML2_INCLUDE_DIR} ${LIBXML2_INCLUDE_DIR}/libxml2 ${CURL_INCLUDE_DIR})
+    TARGET_LINK_LIBRARIES(${TARGET} ${LIBDAP_LIBRARIES} ${LIBXML2_LIBRARIES} ${CURL_LIBRARIES})
+    ADD_DEFINITIONS(-DHAVE_LIBDAP=TRUE)
```
tjhei: can you use something like https://github.com/geodynamics/aspect/blob/master/CMakeLists.txt#L441 instead please?

cmake/modules/FindLIBDAP.cmake
kneumiller: I couldn't get it to work with just "NAMES libdap", but I can list different options so that it searches for our library with different extensions in the order given: `NAMES libdap.so libdap.dylib libdap.la`

doc/modules/changes/20200116_kneumiller
kneumiller: Do you mean another newline after my name/date?

CMakeLists.txt
kneumiller: We put this in there in case the user wants to compile aspect without libdap. We added an option for when cmake is run: 'LIBDAP=ON'. The default is set to OFF, but if the user wishes to search for and use the libdap package they will need to run cmake with the option '-DLIBDAP=ON'.

This is also used by the code we added in utilities.cc in the url_reader pull request.

CMakeLists.txt
kneumiller: If I remove this then CMakeLists appears to not be able to run FindLIBDAP.cmake which is in the cmake/modules/ directory. If I remove the `LIST(APPEND CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}/cmake/modules")` then I have to put the FindLIBDAP.cmake file in the aspect directory, unless there is another way to do this.

kneumiller: I can remove the `MESSAGE(STATUS "cmake module path = ${CMAKE_MODULE_PATH}")` if you want though, that was just there so the user knew where CMakeLists.txt was calling its other, external 'Find<package>.cmake' files.

tjhei: Yes, sorry, I meant the MESSAGE. I don't think it is particularly useful for the user to know.

CMakeLists.txt
tjhei: oh, okay. Can you use something like ASPECT_WITH_LIBDAP and create a cache variable like https://github.com/geodynamics/aspect/blob/master/CMakeLists.txt#L116-L118 ? This way it can be toggled inside "ccmake".

doc/modules/changes/20200116_kneumiller
tjhei: correct.

cmake/modules/FindLIBDAP.cmake
tjhei: at least .so and .dylib. I am not sure about ".la".

CMakeLists.txt
kneumiller: That looks like a much better idea, I'll make those changes.

CMakeLists.txt
kneumiller: Right now if the FIND_PACKAGE fails it just returns: "Could not find a package configuration file provided by CURL/LibXml2".

However it looks like it still finds the libdap libraries if they are installed. Without CURL or libxml2 then the code we added to utilities.cc in url_reader won't be able to read data from a url. I could add a warning or an error if one or both of the CURL/libxml2 packages aren't found.

tjhei: Then it is probably best to find the packages first and only enable it if you find all three. I still confused though, shouldn't libdab provide you with all information including what libraries we need to link to? This can be handled inside FindLIBDAP if needed.
I just checked my installation of libdab and found ``./bin/dap-config --libs``, which returns the complete link line. That would be one way to find the necessary information.

tjhei: not exactly the same as curl doesn't have other libs as dependencies, but this might help: https://github.com/Kitware/CMake/blob/master/Modules/FindCURL.cmake#L136

kneumiller: I've updated the code to not build if one or both CURL and libxml2 are not found. In your latest comment are you saying I could add the dependent library lookups to FindLIBDAP?

Thanks for all of the help by the way

tjhei:
> I've updated the code to not build if one or both CURL and libxml2 are not found. If you think there's more to be done please let me know. In your latest comment are you saying I could add the dependent library lookups to FindLIBDAP?

This is a good start, but I think that you need to discover from libdap where its dependencies are. For example, my system has different versions of libcurl installed and the one linked to by libdap is not the one in /usr/lib/.
Is there documentation available how libdap likes people to link to it?

kneumiller:
> but I think that you need to discover from libdap where its dependencies are. 

I don't think we support that in libdap itself. Unless I'm misunderstanding and there is a way to do it in CMakeLists or FindLibdap.

> Is there documentation available how libdap likes people to link to it?

there is dap-config which is a script that takes a number of switches like --libs and --cflags that tell poeple how to link and compile code that will use libdap.

tjhei:
> there is dap-config which is a script that takes a number of switches like --libs and --cflags that tell poeple how to link and compile code that will use libdap.

To get correct results, we probably need to parse exactly that information. Let me explain with my machine as an example:
```
$ cmake -D ASPECT_WITH_LIBDAP=ON -D LIBDAP_DIR=/ssd/spack/opt/spack/linux-ubuntu16.04-broadwell/gcc-5.4.0/libdap4-3.20.4-2dzjmq6onqavz3cc5gbdeh3jgppkv7lu  ..
-- Found CURL: /usr/lib/x86_64-linux-gnu/libcurl.so (found version "7.47.0")  
[...]
```
but
```
$ ldd $DAP_PATH/libdapclient.so | grep curl
        libcurl.so.4 => /ssd/spack/opt/spack/linux-ubuntu16.04-broadwell/gcc-5.4.0/curl-7.63.0-776fukubmmxkthgfnaohu2ntw36ukseq/lib/libcurl.so.4 (0x00007f0a38a94000
[...]
```
which means dap is compiled with a different version of libcurl than the one you detect. It is even worse for libxml2: the script can not find it, as there is no system installation (I have all 3 installed using spack).

Let me know if you want me to help with making this work.

kneumiller:
I would greatly appreciate the help, what are you thinking?

tjhei:
I have a draft in this branch: https://github.com/tjhei/aspect/tree/libdap_dep
Can you please take a look and test if this works for you on OSX?

kneumiller:
Looks good. I will push the changes up. Do we want to keep all of the output of libs and flags? It looks like a lot of information being displayed at once.

tjhei:
> Do we want to keep all of the output of libs and flags? It looks like a lot of information being displayed at once.

It helped me with debugging the output, but I don't feel strongly about it. You can remove the output inside the loop if you prefer.

.gitignore
```diff
@@ -61,3 +61,4 @@
 
 .idea/
 cmake-build-debug/
+/.autotools
```
tjhei: is this still needed?

doc/modules/changes/20200116_kneumiller
```diff
@@ -0,0 +1,5 @@
...
+<br>
```
tjhei: can you remove this <br> line please?

tjhei:
> If so then I think it should be ready, correct?

with the changes above, I think we are finally good to go.

PR #3334, opened 2020-01-16
Url reader

kneumiller:
Part of the pull requests for OPeNDAP:
'read_and_distribute_file_content' has been modified to check for a url in place of filenames. If the user requests data from a libdap server (given that the libdap libraries are installed) then the data will be pulled from the url server through libdap and converted to the proper format for use within aspect.

### For all pull requests:

* [ X ] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../tests/) directory.


source/utilities.cc
```diff
@@ -21,6 +21,13 @@
 #include <aspect/utilities.h>
 #include <aspect/simulator_access.h>
 
+#ifdef HAVE_LIBDAP
+#include <D4Connect.h>
```
tjhei: I am surprised that libdap does not put its header files into a folder. A header like Array.h has the potential for conflicts.

source/utilities.cc
```diff
@@ -906,11 +913,14 @@ namespace aspect
...
     }
 
-
-    std::pair<double,double> real_spherical_harmonic( const unsigned int l,
-                                                      const unsigned int m,
-                                                      const double theta,
-                                                      const double phi)
+    //Evaluate the cosine and sine terms of a real spherical harmonic.
+    //This is a fully normalized harmonic, that is to say, inner products
+    //of these functions should integrate to a kronecker delta over
```
tjhei: this change is unrelated to your PR (and we also only keep the comments in the header file). Can you undo the change, please?

source/utilities.cc
```diff
@@ -938,12 +948,166 @@ namespace aspect
...
+      std::stringstream urlString;
 
       if (Utilities::MPI::this_mpi_process(comm) == 0)
         {
           // set file size to an invalid size (signaling an error if we can not read it)
           unsigned int filesize = numbers::invalid_unsigned_int;
 
+          /**
+           * Lookup and use data from the provided url
+           * Author: Kodi Neumiller
```
tjhei: please remove this whole block, we typically don't put authorship into code blocks

source/utilities.cc
```diff
@@ -938,12 +948,166 @@ namespace aspect
...
+           * Added: 10/9/18
+           **/
+          //----Only run if the user wishes to use the libdap packages----//
```
tjhei: please remove this comment line

source/utilities.cc
```diff
@@ -938,12 +948,166 @@ namespace aspect
...
+#ifdef HAVE_LIBDAP
+          //Check to see if the prm file will be reading data from the disk or
+          // from a provided URL
+          if (filename.find("http://") == 0 || filename.find("https://") == 0 || filename.find("file://") == 0)
+            {
+              libdap::Connect *url = 0;
+              url = new libdap::Connect(filename);
```
tjhei: 1. Move declaration and assignment into a single line.
2. You never seem to delete this object again, meaning you are leaking memory. Why do you not do ``libdap::Connect url(filename);``?

source/utilities.cc
```diff
@@ -938,12 +948,166 @@ namespace aspect
...
+              libdap::BaseTypeFactory factory;
+              libdap::DataDDS dds(&factory);
+              libdap::DAS das;
+
+              url->request_data(dds, "");
+              url->request_das(das);
+
+
+              //Array to store the url data
+              libdap::Array *urlArray;
```
tjhei: can you move this declaration inside the loop where you use it? We typically declare variables in the narrowest scope possible.

source/utilities.cc
```diff
@@ -938,12 +948,166 @@ namespace aspect
...
+                                               " Maybe it was not of the correct type?"));
+                    }
+                }
+
+
+              //Add the POINTS data that is required and found at the top of the data file.
+              // The POINTS values are set as attributes inside a table.
+              // Loop through the Attribute table to locate the points values within
+              std::vector<std::string> points;
+              libdap::AttrTable *table;
```
tjhei: move declaration inside loop

source/utilities.cc
```diff
@@ -938,12 +948,166 @@ namespace aspect
...
+              for (libdap::AttrTable::Attr_iter i = das.var_begin(); i != das.var_end(); i++)
+                {
+                  table = das.get_table(i);
+                  if (table->get_attr("points") != "")
+                    points.push_back(table->get_attr("points"));
+                  else
+                    break;
```
tjhei: why break here?

source/utilities.cc
```diff
@@ -938,12 +948,166 @@ namespace aspect
...
+                  urlString << "\n";
+                }
+
+              data_string = urlString.str();
+              filesize = data_string.size();
+            }
+
+          else
+            {
+              std::ifstream filestream(filename.c_str());
```
tjhei: is this just a copy&paste from the block below? I think you can reorder your ``ifdef`` and your ``if (starts with http)`` to avoid this copy.

source/utilities.cc
```diff
@@ -938,12 +948,166 @@ namespace aspect
                                      const MPI_Comm &comm)
     {
       std::string data_string;
+      std::stringstream urlString;
```
tjhei: move this declaration into the scope and as close to the first place you use it.

doc/modules/changes/20200116_kneumiller_2
```diff
@@ -0,0 +1,6 @@
+New: The 'read_and_distribute_file_content' now checks if the requested file lookup is a url.
+If the user has the libdap libraries installed and is requesting data from a libdap server then the
+data is pulled from the server, converted to the proper format, and passed in.
+<br>
+(Kodi Neumiller, 2020/01/16)
+<br>
```
tjhei: please remove the last line

source/utilities.cc
kneumiller: Did you want me to remove the comment line "//----Only run if the user wishes to use the libdap packages----//"?

source/utilities.cc
kneumiller: Fixed, thanks for pointing that out.

source/utilities.cc
kneumiller: I think that may have been leftover code. Removed it.

source/utilities.cc
kneumiller: I'm not sure if I can/know how. The reason it's copied is in case the user has libdap turned on but is trying to read a local file. It's currently setup so that the `if (starts with http)` is triggered in the #else (meaning libdap is turned off) then it throws an error. Is there a way to setup the ifdef so that it can read a local file if libdap is on or off?

source/utilities.cc
tjhei: yes

tjhei:
You can restructure the code in the following way:
```
          if (filename.find("http://") == 0 || filename.find("https://") == 0 || filename.find("file://") == 0)
            {
#ifdef HAVE_LIBDAP
[your code to read using libdap]
#else
AssertThrow(false, ExcMessage(...))
#endif
  }
else
  {
    the normal code
  }
}
```

kneumiller:
Oh man, thank you, can't believe I didn't see that.

Also I just realized, I may need to revert some of the declarations because many of them are used outside of the loop they're instantiated in.

kneumiller:
Let me know if this looks okay. 

Thanks for the help!

doc/modules/changes/20200116_kneumiller_2
```diff
@@ -0,0 +1,5 @@
+New: The 'read_and_distribute_file_content' now checks if the requested file lookup is a url.
```
tjhei:
```suggestion
New: The 'read_and_distribute_file_content' function now checks if the requested file lookup is a URL.
```

doc/modules/changes/20200116_kneumiller_2
```diff
@@ -0,0 +1,5 @@
...
+If the user has the libdap libraries installed and is requesting data from a libdap server then the
+data is pulled from the server, converted to the proper format, and passed in.
+<br>
+(Kodi Neumiller, 2020/01/16)
```
tjhei: please hit return at the end of the last line.

tjhei:
What files are available at test.opendap.org, will the currently used files stay around, and would one go about adding new ones?

tjhei:
@gassmoeller can you take a look at this PR as well, please? How do you feel about the files under contrib/? Should we move them somewhere else? cookbooks/ maybe?

kneumiller:
Thanks for all of the suggestions, I'll look them over and push up the changes.

contrib/opendap/OPeNDAP.md
```diff
@@ -0,0 +1,157 @@
+
+# The OPeNDAP extension to ASPECT
+
+The OPeNDAP client extension to ASPECT has been tested on OSX (10.13,
```
bangerth: I think it would be useful if you started this section with a sentence or two what opendap actually is. People may read this file trying to find out.

source/utilities.cc
```diff
@@ -944,38 +951,158 @@ namespace aspect
           // set file size to an invalid size (signaling an error if we can not read it)
           unsigned int filesize = numbers::invalid_unsigned_int;
 
-          std::ifstream filestream(filename.c_str());
-
-          if (!filestream)
+          //Check to see if the prm file will be reading data from the disk or
+          // from a provided URL
+          if (filename.find("http://") == 0 || filename.find("https://") == 0 || filename.find("file://") == 0)
```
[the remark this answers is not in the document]
kneumiller: Yeah that sounds like a good alternative

kneumiller:
> What files are available at test.opendap.org, will the currently used files stay around, and would one go about adding new ones?

We use test.opendap.org to store data used for testing. The data should remain there until it is manually removed. You would have to talk to James about adding new data, but it is a fairly easy process.

source/utilities.cc
```diff
@@ -944,38 +951,158 @@ namespace aspect
...
+              urlString << "# POINTS:";
+              for (unsigned int i = 0; i < points.size(); i++)
+                {
+                  urlString << " " << points[i];
+                }
+              urlString << "\n";
+
+              //Add the values from the arrays into the stringstream. The values are passed in
+              // per row with a character return added at the end of each row.
+              // TODO: Add a check to make sure that each column is the same size before writing
```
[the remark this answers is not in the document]
kneumiller: Still needs work. An error occurs when reading one of the columns.

source/utilities.cc
```diff
@@ -944,34 +961,157 @@ namespace aspect
...
-              AssertThrow (false,
-                           ExcMessage (std::string("Reading of file ") + filename + " finished " +
-                                       "before the end of file was reached. Is the file corrupted or"
-                                       "too large for the input buffer?"));
-              return data_string; // never reached
+              MPI_Bcast(&filesize, 1, MPI_UNSIGNED, 0, comm);
+              AssertThrow(false,
+                          ExcMessage(std::string("Reading of file ") + filename + " failed. " +
+                                     "Make sure you have the dependencies for reading a url " +
+                                     "(when running cmake make sure -DLIBDAP_ON=ON)"));
```
tjhei:
```suggestion
                                     "(run cmake with -D ASPECT_WITH_LIBDAP=ON)"));
```

source/utilities.cc
```diff
@@ -944,34 +961,157 @@ namespace aspect
...
+
+              std::ifstream filestream(filename.c_str());
```
tjhei: why are you opening a file here? Isn't this the case where it is a URL and we don't have libdap? I think you can delete this whole block of code.

kneumiller: I think you're right, fixed that.

tjhei:
thank you, @kneumiller . This is okay from my end. Would you mind squashing your commits into a single one?

kneumiller:
I can squash more commits if needed. Thanks for all of the help, glad we could get this pushed up

gassmoeller:
Looks good from my point of view except for the unanswered comment from Wolfgang about providing a sentence or two at the beginning of the Readme about what this extension actually does. Maybe start the readme with a sentence like the following:
```
This is the OPeNDAP client extension for ASPECT. OPeNDAP is ... (1 sentence). 
Using the OPeNDAP extension ASPECT can query data from remote servers and 
include it in local computations. ... (continue with the existing text)
```
Feel free to modify as appropriate.

dsarahstamps:
This needs to say something about the EarthCube funded BALTO project. It’s the BALTO plugin, not OPeNDAP.

tjhei:
Looks good, thank you everyone!

PR #3337, opened 2020-01-18
Add appropriate #include file.

bangerth:
When using PETSc for linear algebra, we use a function from namespace
SparsityTools. Make sure it is available.


bangerth:
Part of #3338.

PR #3343, opened 2020-01-23
Remove restriction to 3D from S40RTS initial condition plugin.

gassmoeller:
This allows to use the S40RTS initial temperature plugin for 2D models.


source/initial_temperature/S40RTS_perturbation.cc
```diff
@@ -242,7 +221,18 @@ namespace aspect
         {
           for (unsigned int order_m = 0; order_m < degree_l+1; ++order_m)
             {
-              const std::pair<double,double> sph_harm_vals = Utilities::real_spherical_harmonic(degree_l, order_m, scoord[2], scoord[1]);
+              std::pair<double,double> sph_harm_vals;
+              if (dim == 3)
+                {
+                  sph_harm_vals = Utilities::real_spherical_harmonic(degree_l, order_m, scoord[2], scoord[1]);
```
tjhei: you could move this line outside the if and only assign ``tmp=scoord[2]`` or ``tmp=PI_2`` inside the if.

PR #3349, opened 2020-01-24
Fix probability density mpi problem

MFraters:
Found a problem where for the probability density function particle generator
1. setting 1 particle in input, while having 3 mpi processes, no particles are created;
2. setting 2 particles in input, while having 3 mpi processes, creates 3 particles.

This should fix the problem and I added some tests (2 of which would have failed without this code). The problem was in how the number of particles per mpi process was computed, which is fixed now with the help of @gassmoeller. I tested all the particle generator tests locally and they all still pass.

I did check why the 3 mpi 3 particle case had two particles on process 0, zero on processor 1 and one on processor 2, and it seem to have to do with that processor 0 has a bigger domain (probably 2 cells, and the other ones just 1 cell each).

```
01 Rank: 0, local_weight_integral = 8, local_start_weight = 8
02 Rank: 0, local_weight_integral = 8, local_start_weight = 0
Rank: 0, n_particles = 3, start_id = 0, n_local_particles = 2, particles_per_cell[0] = 0
01 Rank: 1, local_weight_integral = 4, local_start_weight = 12
02 Rank: 1, local_weight_integral = 4, local_start_weight = 8
Rank: 1, n_particles = 3, start_id = 2, n_local_particles = 0, particles_per_cell[0] = 0
01 Rank: 2, local_weight_integral = 4, local_start_weight = 16
02 Rank: 2, local_weight_integral = 4, local_start_weight = 12
Rank: 2, n_particles = 3, start_id = 2, n_local_particles = 1, particles_per_cell[0] = 0
```

Should I also add a changelog for this?

### For all pull requests:


masterleinad:
Thanks!

source/particle/generator/probability_density_function.cc
```diff
@@ -237,7 +239,8 @@ namespace aspect
                                    "runs. Change to get a different distribution. In parallel "
                                    "computations the seed is further modified on each process "
                                    "to ensure different particle patterns on different "
-                                   "processes.");
+                                   "processes. Note that the amount of particles per processor "
```
tjhei:
```suggestion
                                   "processes. Note that the number of particles per processor "
```

doc/modules/changes/20200124_mfraters
```diff
@@ -0,0 +1,4 @@
+Fix: Fixed a bug in the particle generator probability density function when 
+using less particles then mpi processes.
```
tjhei:
```suggestion
using fewer particles than mpi processes.
```

source/particle/generator/probability_density_function.cc
```diff
@@ -68,11 +68,13 @@ namespace aspect
         // the weights of all processes with a lower rank
         double local_start_weight = 0.0;
         MPI_Scan(&local_weight_integral, &local_start_weight, 1, MPI_DOUBLE, MPI_SUM, this->get_mpi_communicator());
```
bangerth: Just as a follow-up: In deal.II, this call has been replaced by `MPI_Exscan`. If you felt ambitious, feel free to change it here as well :-)

MFraters: Yes, Rene and I discussed that a bit after I made the pull request, but I ended up not changing it. I could make a new pull request for that, should not be that much work to bring it in line with deal.ii.

MFraters: done #3353 :)

PR #3355, opened 2020-01-28
followup to #3333: libdap configuration

tjhei:
- fix list of libraries linked to
- add ASPECT header
- fix changelog and add myself
- fix include path detection

@kneumiller please take a look and let me know if this is okay with you.


kneumiller:
Looks good!

Issue #3335, opened 2020-01-16
SUPG Convergence Issues


maxrudolph:
I have been encountering failures to converge when using SUPG in 3D. We actually haven't been successful in running a single 3D model to completion with SUPG. This includes both 3D spherical geometry models and 3D cartesian models, both with and without AMR. I have tried to come up with the simplest possible testcase, which I have attached. This crashes after about 40 timesteps due to failure to converge. It takes less than a minute to run on 32 cores.

[solver_history.txt](https://github.com/geodynamics/aspect/files/4073697/solver_history.txt)
[supg_testcase.prm.txt](https://github.com/geodynamics/aspect/files/4073696/supg_testcase.prm.txt)
[log.txt](https://github.com/geodynamics/aspect/files/4073700/log.txt)

I haven't tried using SUPG in 2D so I'm not sure if these issues are 3D-specific. In 2D it is generally cheap enough to run at a high enough refinement level to reduce the entropy viscosity to values below the thermal conductivity.

tjhei:
Thanks for this test case. :+1: 
I have not used SUPG for any larger or more realistic computations, so I am not too surprised that you are encountering problems. I have some ideas on how to fix this. Will report back.

tjhei:
This is the error I get in debug mode after 62 timesteps:
```

*** Timestep 62:  t=2.3924e-05 seconds
   Solving temperature system... 30 iterations.
   Rebuilding Stokes preconditioner...
   Solving Stokes system... 83+0 iterations.

   Postprocessing:
     RMS, max velocity:                  9.45e+04 m/s, 5.35e+05 m/s
     Temperature min/avg/max:            -1.013 K, 0.4936 K, 1.603 K
     Heat fluxes through boundary parts: 6906 W, -6906 W, -3992 W, 3992 W, -2883 W, 1645 W
     Writing graphical output:           output_supg_testcase/solution/solution-00062


--------------------------------------------------------
An error occurred in line <605> of file <../source/simulator/helper_functions.cc> in function
    double aspect::Simulator<3>::compute_time_step() const [dim = 3]
The violated condition was:
    rho * c_p > 0
Additional information:
    The product of density and c_P needs to be a non-negative quantity.
```
A couple of things about this:
- We should rethink if this should also be detected in release mode @gassmoeller ?
- Should the material model maybe make sure that the density stays positive?
- Your setup is severely underresolved, which probably explains the situation. Maybe we need to tune the SUPG parameters (in 3d)?

[...]

PR #3356, opened 2020-01-29
use reference density for conduction timestep

tjhei:
If using a formulation like BA, we should use adiabatic density instead
of the real density to compute the conduction timestep.

part of #3335

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -591,6 +591,22 @@ namespace aspect
 
               material_model->evaluate(in, out);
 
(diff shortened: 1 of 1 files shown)


tjhei:
> Anyway, ready to merge once the tester is happy.

well, it reduces the timestep size significantly so it is not crashing for me (at least in the first 100 steps).

gassmoeller:
Oh, interesting, smaller timesteps will of course help. But I do not understand why this happens, I will post in the issue about it. This PR is good anyway.

PR #3357, opened 2020-01-30
Simplify step limiting logic and add error message for negative timesteps

gassmoeller:
After looking at #3356 I noticed two things: 
1. We can potentially allow negative time step sizes if the material model hands over negative densities, specific heat capacities or thermal conductivities and the conduction time step is used. This is addressed in this PR by adding a new AssertThrow that also trigger in release mode.
2. The logic for limiting time steps at the end of this function was pretty hard to understand. I simplified it here. All of the functionality should be preserved (every condition change should be equivalent to the conditions before).

Let me know what you think.

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -631,24 +631,26 @@ namespace aspect
     double new_time_step = std::min(min_convection_timestep,
                                     min_conduction_timestep);
 
-    if (new_time_step == std::numeric_limits<double>::max())
-      {
-        // In some models the velocity is zero, either because that is the prescribed
-        // Stokes solution, or just because there is no buoyancy and nothing is moving.
-        // If this is the case, and if we either do not compute the conduction time
-        // step or do not have any conduction, it is somewhat arbitrary what time step
-        // we should choose. In that case, set the time step to the 'Maximum time step'.
-        new_time_step = parameters.maximum_time_step;
-      }
+    AssertThrow (new_time_step > 0,
+                 ExcMessage("The time step length for the each time step needs to be positive, "
+                            "but the computed step length was: " + std::to_string(new_time_step) + ". "
+                            "Please check for non-positive material properties."));
 
-    // make sure that the timestep doesn't increase too fast
+    // Make sure we do not exceed the maximum time step length. This can happen
+    // if velocities get too small or even zero in models that are stably stratified
+    // or use prescribed velocities.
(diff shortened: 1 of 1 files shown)


source/simulator/helper_functions.cc
```diff
@@ -631,24 +631,26 @@ namespace aspect
...
-      }
+    AssertThrow (new_time_step > 0,
+                 ExcMessage("The time step length for the each time step needs to be positive, "
+                            "but the computed step length was: " + std::to_string(new_time_step) + ". "
+                            "Please check for non-positive material properties."));
 
-    // make sure that the timestep doesn't increase too fast
-    if (time_step != 0)
+    // Make sure that the time step doesn't increase too fast
+    if ((time_step != 0) && (new_time_step < parameters.maximum_time_step))
```
tjhei: why did you add this second conditional? I would think this limiting should also apply if the new time step is larger than maximum_time_step.

gassmoeller: The main reason is that previously we had the check for `new_time_step < std::..::max()` in line 634 above, and now that is implied in the limiting to `maximum_time_step` in line 650 below. I do not know what happens if I multiply `std::max()` with a number larger than one (is there a floating point exception?) that is why I prevented it here. You are right we now skip the time step increase for the case where `maximum_time_step < new_time_step < std::max()`, however this line can only increase the new time step size, and it will be limited to maximum_time_step afterwards anyway, so this change should not have an effect. Would you prefer me to change the line to the following?
```
     if ((time_step != 0) && (new_time_step != std::numeric_limits<double>::max())) 
```

tjhei: I was referring to the situation ``time_step==1`` and ``new_time_step == 2*parameters.maximum_time_step``. The old code would go into this ``if`` and restrict the time step increase while the new code does not (limit will go to ``maximum_time_step`` at the bottom but not be set to something potentially smaller).
I am okay how you handle the max case.

tjhei: I think you can just unconditionally set ``new_time_step = std::min(new_time_step, parameters.maximum_time_step)`` as the first choice. Then you don't have to worry about the ``<double::max`` case.

gassmoeller: Correct, I fixed the order.

tjhei:
Much cleaner than before, awesome work! :+1:

PR #3364, opened 2020-02-04
Fix periodicity direction in box with lithosphere boundary indicators

anne-glerum:
ASPECT currently crashes when periodicity is prescribed on the additional boundary indicators, as noticed by @EstherHeck, in `dealii::DoFTools::make_periodicity_constraints`. This PR sets the correct direction for periodicity when prescribing periodic boundary conditions on any of the additional boundary indicators. Only the directions [0,dim) are allowed, but the plugin currently continues counting due to the additional boundary indicators.


source/geometry_model/two_merged_boxes.cc
```diff
@@ -201,7 +201,11 @@ namespace aspect
       std::set< std::pair< std::pair<types::boundary_id, types::boundary_id>, unsigned int> > periodic_boundaries;
       for ( unsigned int i=0; i<dim+dim-1; ++i)
         if (periodic[i])
-          periodic_boundaries.insert( std::make_pair( std::pair<types::boundary_id, types::boundary_id>(2*i, 2*i+1), i) );
+          {
+            // The direction of periodicity (0,dim-1)
+            const unsigned int tmp_dim = i>=dim ? i-dim : i;
```
tjhei: This is the direction,  can you rename the variable to ``direction`` please? Then the comment is not needed anymore.

doc/modules/changes/20200402_glerum
```diff
@@ -0,0 +1,5 @@
+Fixed: There was a bug in the 'box with lithosphere' geometry model plugin
+that lead to a crash when using periodic boundary conditions on any of th
```
tjhei:
```suggestion
that lead to a crash when using periodic boundary conditions on any of the
```

Issue #2244, opened 2018-05-14
Post-2.0 cleanups


bangerth:
I can think of the following things we ought to do:
- [ ] Require deal.II 9.0 in `CMakeLists.txt`
- [ ] Document this where appropriate
- [ ] Switch the testers to use deal.II 9.0
- [ ]  Remove `#if DEAL_II_VERSION_GTE(9,0)` statements.

This means we are now allowed to use C++11 -- yay!

bangerth:
We can also remove `#if DEAL_II_VERSION_GTE(9,0)` statements.

tjhei:
Does this make sense to do before the hackathon?
Pro: 
- we can hand out starter projects to do c++11 related cleanups.
- we don't have to worry about 8.5 and 9.0 compatibility
Cons:
- Updating the test infrastructure (and test output) is a lot of work
- This forces everybody to install 9.0 on their machines.

tjhei:
As a compromise: we could require c++11 (enforce in cmakelists.txt) and start using c++11 stuff today! This would give some nice starter projects...

bangerth:
You mean continue to allow deal.II 8.5, but require C++11 anyway? That sounds fine with me.

bangerth:
We are tracking the C++11 uses in #2375.

bangerth:
We agreed to not require deal.II 9.0 at this time, so let's close this.

PR #3366, opened 2020-02-11
Specify thermal conductivities in visco_plastic material model

naliboff:
This pull request adds the option to specify thermal conductivities in the `visco_plastic` material, rather than have the values calculated using thermal diffusivities. This is necessary when using sticky-air that contains significantly different thermodynamic properties in combination with cell averaging.

@gassmoeller - I agree that at some point we should break backward compatibility and make thermal conductivity the default user-defined variable.


naliboff:
@gassmoeller @jdannberg - A recap of our conversation regarding the options for this PR:

1. Merge the pull request as is, but before the next release break backwards compatibility and remove the option to specify thermal diffusivity.
2. Merge the pull request as is, but before the next release switch the default value of` Define thermal conductivities` from false to true. This breaks backward compatability, but we should be able to address this change in an update script.
3. Modify the current PR to calculate thermal conductivity from thermal diffusivity when the parameters are parsed, but still keep the new option to directly specify thermal conductivities. This will require updating the test suite due to the ordering of averaging.
4. Other option I forgot?

My preference would be for option 1 as it removes code and eliminates any ambiguity about how the thermal conductivity is calculated.

include/aspect/material_model/visco_plastic.h
```diff
@@ -179,6 +179,13 @@ namespace aspect
 
         std::vector<double> thermal_diffusivities;
 
+        /**
+         * Whether to use user-defined thermal conductivites instead of thermal diffusivities.
+         */
+        bool define_conductivities;
```
gassmoeller: Remove this parameter.

source/material_model/visco_plastic.cc
```diff
@@ -425,6 +425,11 @@ namespace aspect
           else
             out.thermal_conductivities[i] = thermal_diffusivity * out.specific_heat[i] * out.densities[i];
 
+          // Replace thermal conductivities with values directly specified in the parameter, if this
+          // option was selected.
+          if (define_conductivities == true)
```
gassmoeller: Remove this `if` but keep the line about thermal conductivities below. Remove everything related to thermal diffusivity above.

source/material_model/visco_plastic.cc
```diff
@@ -548,6 +553,16 @@ namespace aspect
                              "List of thermal diffusivities, for background material and compositional fields, "
                              "for a total of N+1 values, where N is the number of compositional fields. "
                              "If only one value is given, then all use the same value.  Units: $m^2/s$");
+          prm.declare_entry ("Define thermal conductivities","false",
```
gassmoeller: Remove this parameter.

source/material_model/visco_plastic.cc
```diff
@@ -630,6 +645,12 @@ namespace aspect
                                                                           n_fields,
                                                                           "Thermal diffusivities");
 
+          define_conductivities = prm.get_bool ("Define thermal conductivities");
```
gassmoeller: Remove this parameter.

include/aspect/material_model/visco_plastic.h
gassmoeller: Also remove the `thermal_diffusivities` parameter above.

source/material_model/visco_plastic.cc
```diff
@@ -630,6 +645,12 @@ namespace aspect
...
+
```
gassmoeller: Here add the following:
1. An Assert that makes sure at most one of thermal diffusivities or thermal conductivities was changed from the default value.
2. If both are at default values or if only thermal conductivities are set, execute the line below.
3. If only thermal diffusivities are set differently from the default value, use line 644 to read thermal diffusivities in a temporary vector, then convert and store them using density and cp into `thermal conductivities`.

This should take care of every case existing parameter files and new files, without introducing a temporary input parameter that we would need to remove later.

naliboff: A bit confused here - one of the comments requests `thermal_diffusivities` be removed from `visco_plastic.h` ? Was that correct or did you mean remove `thermal_conductivites`?

Maybe the simplest thing is to just switch to solely using thermal conductivity now as that is what we will do in the long run regardless?

gassmoeller: Yes, I think we agreed to remove `thermal_diffusivities` as a member variable from visco_plastic.h, but keep the input parameter for now (and store it in a temporary vector inside of `parse_parameters` to then convert it to thermal conductivities and store it in `thermal_conductivities`). This way we keep the code simpler and maintain parameter file compatibility for now.

naliboff: Thanks and glad you had a proper recollection of the discussion. This will likely change the test results due to the order of averaging thermodynamic properties, but I'm fine with that if you are as well.

I'm still a bit confused about the proposed Assert in your step 1 above - it seems like this requires parsing the thermal conductivities first? Or would this be an Assert that checks whether the parameter is simply declared in the parameter file?

gassmoeller: I would write it like this, which would check that at least one of the two parameters is set to the default value:
```
AssertThrow (prm.get("Thermal conductivities") == "3.0" || prm.get("Thermal diffusivities") == "0.8e-6", ExcMessage("Both thermal conductivities and thermal diffusivities have been set. This is not possible. Please only specify thermal conductivities as the input parameter 'thermal diffusivities' will be removed in a future version."));
```

naliboff:
@gassmoeller - I've rebased and implemented most of the suggestions above, but have gotten stuck on how to best calculate the thermal conductivities from thermal diffusivities in the parse parameter section.

The issues is as follows: While the thermal diffusivities are first parsed in the normal manner (extend vector to n_fields), we still need to get the densities and heat capacities from the equation of state module and make some assumptions about what phase we are in.

gassmoeller:
You are right, that is a serious complication. I do not think your workaround would be a good solution, it would prevent us from having different thermal diffusivities for different phases, and it would also modify the results of existing models. I thought about actually computing EoS parameters inside `parse_parameters`, but we do not know the reference temperature outside of the EoS module. Maybe that means we should go with your initial version of this PR and allow the user to explicitly choose (with the intent of deprecating and removing the thermal diffusivities option in a future version). I took another look at the current state of this PR and think we could probably merge nearly as is. The only modification I would like to see is that in line 411-431 the thermal diffusivity is only used if it is actually selected, so if that section would be like this:
```
        if (define_conductivities == false)
        {
          double thermal_diffusivity = 0.0;
            thermal_diffusivity += volume_fractions[j] * thermal_diffusivities[j];
          // Thermal conductivity at the given positions. If the temperature equation uses
          // the reference density profile formulation, use the reference density to
          // calculate thermal conductivity. Otherwise, use the real density. If the adiabatic
          // conditions are not yet initialized, the real density will still be used.
          if (this->get_parameters().formulation_temperature_equation ==
              Parameters<dim>::Formulation::TemperatureEquation::reference_density_profile &&
              this->get_adiabatic_conditions().is_initialized())
            out.thermal_conductivities[i] = thermal_diffusivity * out.specific_heat[i] *
                                            this->get_adiabatic_conditions().density(in.position[i]);
          else
            out.thermal_conductivities[i] = thermal_diffusivity * out.specific_heat[i] * out.densities[i];
       }
       else
       {
          // Use thermal conductivity values specified in the parameter file, if this
          // option was selected.
            out.thermal_conductivities[i] = MaterialUtilities::average_value (volume_fractions, thermal_conductivities, MaterialUtilities::arithmetic);
}
```

naliboff:
@gassmoeller - I agree the most straightforward path forward is to just use the current PR structure and then deprecate the option to specify thermal diffusivity before the next release. I've added your suggestions for lines 411-431 and will update when the tester is finished.

Is this a good time to start a release checklist?

gassmoeller:
> Is this a good time to start a release checklist?

Any time is a good time to start a release checklist :-)

But what would would want to do to deprecate the parameter? I would suggest to keep everything as in this PR for the release so that both version work. After the release we can remove the thermal diffusivities code and let the code issue a nice error message for when they are used in the parameter file. Maybe an issue like this would help us keep track of things: https://github.com/geodynamics/aspect/issues/2244

naliboff:
> Maybe an issue like this would help us keep track of things: #2244
Deprecating the parameter after the next release and adding an error message sounds like a good plan to me. Is the next planned release 2.2?

naliboff:
@gassmoeller - I think this is good to go unless you would like any additional changes.

gassmoeller:
> Is the next planned release 2.2

I think so. Even if we decide to change the name, we can always edit the issue title.

PR #3367, opened 2020-02-11
[WIP] Enable additional RHS terms and elasticity in Newton solver

naliboff:
This PR enables additional RHS terms, including elasticity, in the Newton solver assembly. 

As the `viscoelastic` and `viscoelastic_plastic material` model currently do not setup the Newton derivatives, this PR should wait until #3306 is merged and this is rebased on top of it. 

The single added test using the Newton solver is a cheat in that only one nonlinear iteration is performed and thus no call to assemble derivatives is required.

@MFraters - opening this so you can take a look.

### DIFF
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -249,8 +249,8 @@ namespace aspect
                  internal::Assembly::CopyData::CopyDataBase<dim> &data_base) const override;
 
         /**
-         * Create additional material models outputs for assembly of derivatives or adding additional 
-         * terms to the right hand side of the Stokes equations. The latter could include viscoelastic 
+         * Create additional material models outputs for assembly of derivatives or adding additional
+         * terms to the right hand side of the Stokes equations. The latter could include viscoelastic
          * forces or other user-defined values calculated within the material model.
          */
         void create_additional_material_model_outputs(MaterialModel::MaterialModelOutputs<dim> &outputs) const override;
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -249,7 +249,9 @@ namespace aspect
                  internal::Assembly::CopyData::CopyDataBase<dim> &data_base) const override;
 
         /**
-         * Create AdditionalMaterialOutputsStokesRHS if we need to do so.
+         * Create additional material models outputs for assembly of derivatives or adding additional 
+         * terms to the right hand side of the Stokes equations. The latter could include viscoelastic 
+         * forces or other user-defined values calculated within the material model.
          */
         void create_additional_material_model_outputs(MaterialModel::MaterialModelOutputs<dim> &outputs) const override;
     };
--- a/include/aspect/newton.h
+++ b/include/aspect/newton.h
@@ -247,6 +247,11 @@ namespace aspect
         void
         execute (internal::Assembly::Scratch::ScratchBase<dim>  &scratch_base,
                  internal::Assembly::CopyData::CopyDataBase<dim> &data_base) const override;
+
+        /**
+         * Create AdditionalMaterialOutputsStokesRHS if we need to do so.
+         */
+        void create_additional_material_model_outputs(MaterialModel::MaterialModelOutputs<dim> &outputs) const override;
     };
 
     /**
--- a/source/simulator/assemblers/newton_stokes.cc
+++ b/source/simulator/assemblers/newton_stokes.cc
@@ -485,6 +485,11 @@ namespace aspect
              ||
              outputs.template get_additional_output<MaterialModel::ElasticOutputs<dim> >()->elastic_force.size()
              == n_points, ExcInternalError());
+
+      if (this->get_newton_handler().parameters.newton_derivative_scaling_factor == 0)
+        return;
+
+      NewtonHandler<dim>::create_material_model_outputs(outputs);
     }
 
 
--- a/source/simulator/assemblers/newton_stokes.cc
+++ b/source/simulator/assemblers/newton_stokes.cc
@@ -268,6 +268,12 @@ namespace aspect
       const unsigned int n_q_points    = scratch.finite_element_values.n_quadrature_points;
       const double derivative_scaling_factor = this->get_newton_handler().parameters.newton_derivative_scaling_factor;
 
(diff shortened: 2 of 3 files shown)


tjhei:
This looks good to me. The following tests are failing:
```
	425 - nonlinear_channel_flow_tractions_Newton_Stokes (Failed)
	426 - nonlinear_channel_flow_tractions_Newton_Stokes_no_deviator (Failed)
	428 - nonlinear_channel_flow_velocities_Newton_Stokes (Failed)
	431 - nonlinear_channel_flow_velocities_Newton_Stokes_no_deviator (Failed)
	433 - nonlinear_channel_flow_velocities_single_advection_Newton_Stokes (Failed)
	641 - spiegelman_fail_test (Failed)
	720 - viscoelastic_newton (Failed)
```

@naliboff Is this change only necessary for the Newton solver or was the Picard version for problems with elasticity also incorrect?

naliboff:
@tjhei - yes, this is only necessary for the Newton assembly. The terms were already assembled correctly in `stokes.cc`. 

Adding the additional RHS interface seems to have broken how the derivatives are assembled from the material model via the additional material model outputs mechanism. Should be a straightforward fix.

tjhei:
> yes, this is only necessary for the Newton assembly.

to make sure that I understand:
1. iterated Stokes: before: correct, after: correct
2. iterated Newton Stokes Picard only: before: incorrect, after: correct
3. iterated Newton Stokes Newton: before: incorrect, after: still incorrect as derivatives are missing.

correct?

naliboff:
> 1. iterated Stokes: before: correct, after: correct
> 2. iterated Newton Stokes Picard only: before: incorrect, after: correct
> 3. iterated Newton Stokes Newton: before: incorrect, after: still incorrect as derivatives are missing.

Yes, this is my understanding as well. However, perhaps we don't need additional derivatives as the elasticity is incorporated by modifying the viscosity and adding forces to momentum right hand side? Would we need derivatives of the elastic stress w.r.t. strain rate and pressure?

include/aspect/newton.h
```diff
@@ -247,6 +247,11 @@ namespace aspect
         void
         execute (internal::Assembly::Scratch::ScratchBase<dim>  &scratch_base,
                  internal::Assembly::CopyData::CopyDataBase<dim> &data_base) const override;
+
+        /**
+         * Create AdditionalMaterialOutputsStokesRHS if we need to do so.
```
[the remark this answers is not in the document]
naliboff: The documentation should probably actually be modified a bit further now that the derivates are handled with the same function. How does the following sound?:
` Create additional material models outputs for assembly of derivatives or adding additional terms to the right hand side of the Stokes equations. The latter could include viscoelastic forces or other user-defined values calculated within the material model`

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -438,6 +455,43 @@ namespace aspect
...
+
+      Assert(!this->get_parameters().enable_elasticity
+             ||
+             outputs.template get_additional_output<MaterialModel::ElasticOutputs<dim> >()->elastic_force.size()
+             == n_points, ExcInternalError());
+
+      if (this->get_newton_handler().parameters.newton_derivative_scaling_factor == 0)
+        return;
+
+      NewtonHandler<dim>::create_material_model_outputs(outputs);
```
MFraters: Looking again at this code, I think it can be simplified into the following:

```suggestion
      if (this->get_newton_handler().parameters.newton_derivative_scaling_factor != 0)
        NewtonHandler<dim>::create_material_model_outputs(outputs);
```

naliboff: 👍Aside from this change, are you satisfied with the current structure? There may be a way to get around some of the code duplication in the last commit, but with the change above it is just two lines being duplicated.

include/aspect/newton.h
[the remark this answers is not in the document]
MFraters: looks good to me. Because it look kind of weird to me that it is in the `NewtonStokesIncompressibleTerms` class, while the derivative are in principle also for the compressible case, I looked around a bit more in the code. 

The central part of this code is in the `NewtonHandler` `set_assemblers` function in `newton.cc` line 51. There it always inserts the `NewtonStokesIncompressibleTerms` assembler. The reason the test failed is because before you created the specific `create_additional_material_model_outputs` function, it used the default implantation of the `NewtonInterface`, and now that is overwritten. 

Since the `NewtonStokesIncompressibleTerms` is always used, even in the compressible case, I do not see a problem keeping it here.

source/simulator/assemblers/newton_stokes.cc
MFraters: The code in the default implementation in `NewtonInterface` can be deleted now.

The structure look correct to me. Although I think it is alright, I am just doubting whether this is the correct place to add the newton additional material model outputs. Any opinions? @gassmoeller @tjhei @jdannberg

naliboff: @MFraters - are operations other than `create_additional_material_model_outputs` performed in the `NewtonInterface` class? If not, can that class be removed? I agree there might be a better place for all of this code, although this is now consistent with the structure in the `stokes.cc` assembly.

MFraters: I guess you are right that it can be completely removed. Just replace the `:public NewtonInterface<dim>` with `: public aspect::Assemblers::Interface<dim>, public SimulatorAccess<dim>` where needed. 

Yes, you are right that in this way it conforms to `stokes.cc`, so I think the new structure is good.

naliboff: I was wrong, sorry. The `NewtonStokesPreconditioner` still needs to access the derivatives through the `NewtonInterface` (local tests failed). I vote to just keep the structure in the most recent commit unless someone has strong feelings about reorganization. I'll revisit this when #3306 is merged and we can fully test to make sure the elastic terms are assembled properly.

MFraters: hmm, that is interesting. Yes, let's just keep it as it is :)

PR #3369, opened 2020-02-13
add picard iteration gmg

tcclevenger:
One can now use GMG with Picard iteration. There is an `Assert(newton_derivative_scaling_factor==0)` in `solve_iterated_advection_and_newton_stokes()` in `solver_schemes.cc` so that the computation is aborted if the code switches to Newton solver (still not implemented).

I've created a test based on `test/nonlinear_channel_flow_velocities_Newton_Stokes`, except the `.prm` has been changed to use harmonic averaging and to never switch to Newton. I verified the output match the same type of run with AMG.

I'll make a few comments after creating the pull request to clarify some rearranging I had to do for the StokesMatrixFree class.


include/aspect/stokes_matrix_free.h
```diff
@@ -306,8 +306,21 @@ namespace aspect
       virtual void setup_dofs()=0;
 
       /**
```
tcclevenger: I had to move `evaluate_material_model()` and `correct_stokes_rhs()` to be public functions since I now need to call them separate from the `build_preconditioner()` in `solver_schemes.cc` and `assembly.cc`. Also, now `build_preconditioner()` is the same as the old `compute_A_block_diagonals()`.

source/simulator/assembly.cc
```diff
@@ -814,6 +817,13 @@ namespace aspect
       }
```
tcclevenger: It seems natural to me to update to matrix-free right-hand side right after assembling. Maybe the `evaluate_material_model()` can move, however. Where are the material model values set and reset for each nonlinear iteration? In `core.cc`? Maybe we don't need to do this every time we assemble.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1322,7 +1322,6 @@ namespace aspect
                                sim.pressure_scaling,
                                sim.triangulation,
                                dof_handler_projection);
```
tcclevenger: the mass matrix diagonal is needed for the preconditioner. It seems more appropriate in the `build_preconditioner()` function.

include/aspect/stokes_matrix_free.h
tjhei: I like the new naming.

source/simulator/stokes_matrix_free.cc
```diff
@@ -2021,21 +2020,15 @@ namespace aspect
   }
 
 
-
```
gassmoeller: Keep the third empty line, that is our usual style

doc/modules/changes/20200214_clevenger
```diff
@@ -0,0 +1,3 @@
+New: Geometric Multigrid now supports computations with Picard iteration.
```
gassmoeller:
```suggestion
New: The geometric multigrid preconditioner for the Stokes system now supports Picard iterations for the defect correction form of the Stokes equations.
```

The original message was somewhat misleading, because as far as I can tell Picard iterations for the original form of the Stokes equations were already supported beforehand.

tcclevenger: Yes, this is correct!

tcclevenger:
> Looks good, just a few minor comments.

Thanks, @gassmoeller! I've addressed the comments. I'll make sure the tests pass (last time I think one failed? We will see.)

gassmoeller:
Thanks. 

> last time I think one failed?

You can also check old test results by clicking on the little red x next to the commit hash in the pull request conversation (right above your comment). Saves you some waiting for the tester.

tcclevenger:
As with all the other GMG tests, the test fails because of the dependence on the machine architecture. I tested and verified on an AVX machine (vectorization level 2) and the Jenkins tester is a SSE2 machine (vectorization level 1). This can slightly alter iteration counts (as I see from the test output). Is there a way for me to access the output from Jenkins to change my test output file?

tjhei:
Details, artifacts in the top right.

tcclevenger:
Issues with the test have been resolved. Everything passed. Let me know if you have anymore comments!

PR #3370, opened 2020-02-14
Fix dynamic topography postprocessor coordinate output to respect manifold

gassmoeller:
As described in the changelog entry this fixes a minor bug in the dynamic topography postprocessor. First reported by @mnkajan. Also fix some random spaces in the code.

The test data has been updated and validated. 
Old radius of surface point: 5921515.0978 m. 
New radius of surface point: 6335999.9998 m. 
Prescribed radius in input file: 6336000 m.


naliboff:
Looks good, thanks for doing additional cleanup as well! It looks one additional set of test results need to be updated (`geoid`), but other than that good to go.

PR #3372, opened 2020-02-14
update vep test parameter file

naliboff:
This PR is part of the preparation for the update to the visco_plastic material model in #3306. 

The test parameter file updated here is derived from the `kaus_2010_extension benchmark`, which uses the viscoelastic_plastic material model. In #3306 this benchmark will be modified to use the visco_plastic material, but for a short period we are going to keep the viscoelastic_plastic material for testing purposes. This change ensures continuity of the test suite and allows for easier comparisons of test results in #3306. 

### DIFF
--- a/tests/viscoelastic_plastic_brick_extension.prm
+++ b/tests/viscoelastic_plastic_brick_extension.prm
@@ -35,6 +35,31 @@ subsection Initial composition model
   end
 end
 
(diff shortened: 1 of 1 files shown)


naliboff:
@tjhei - If you are still looking at PR's, this one should ready to merge (just an update to test prm file with no change to test results).

PR #3374, opened 2020-02-15
Update benchmarks

naliboff:
This PR is on top of the single commit in #3372 and is part of the preparation for #3306. 

The changes to the benchmarks, and one corresponding test, either remove boundary effects in the compositional fields for viscoelastic stresses or remove viscoelastic effects in purely viscous material. 

This should produce a change in the test results, but the benchmarks have already been rerun to confirm improved results.


tests/viscoelastic_plastic_brick_extension.prm
```diff
@@ -35,6 +35,31 @@ subsection Initial composition model
   end
 end
 
+# Use viscoelastic plastic material model
+subsection Material model
```
tjhei: When you later update the base (included prm), we should probably remove the material model section here.

tjhei:
Excellent, looks good.

tests/viscoelastic_plastic_brick_extension.prm
naliboff: That is the plan 👍. The `viscoelastic_plastic` test suite will keep the `Material model` section as the corresponding benchmarks will soon use `visco_plastic`, but the intention is to completely remove `viscoelastic_plastic` (source code, tests) soon thereafter.

naliboff: Thanks for handling all of the reviews and merges today!

Issue #3359, opened 2020-01-30
Chunk+free surface unwanted deformation


fionaclerc:
When running the free surface and 2-D 90-degree chunk together, it seems as if the free surface is moving out from the bottom (east) boundary, by ~kms. This is surprising as density/temperature are constant such that no deformation should occur. I'm attaching a plot of the topography at different time steps, as well as a snapshot of paraview showing the pressure field, and the input parameter file. I did play around with different aspect ratios which didn't resolve this, as well as increasing the radial depth which generated instabilities. Thanks for your help ! -Fiona

-----------------------------------
[play_traction6B.prm.txt](https://github.com/geodynamics/aspect/files/4135966/play_traction6B.prm.txt)

![pressure_chunk_free_surface](https://user-images.githubusercontent.com/37747437/73484303-0e62ac80-436f-11ea-8623-e9779759aab2.png)

[chunk_free_surface_topo.pdf](https://github.com/geodynamics/aspect/files/4135944/chunk_free_surface_topo.pdf)

gassmoeller:
Hi Fiona,
Weird, I do not see anything particular off about this model setup, but I would check the following things:
- Does this also happen if you use a spherical shell model with the same setup (of course the elements will have a bad aspect ratio, but maybe it still runs)
- Was there a reason you switched to first order temperature elements for this model? Does anything change with 2nd order?
- Is the nonadiabatic pressure more or less constant this way, or are there oscillations? If so do they affect the small pressure spikes at the top and bottom boundary or does the whole pressure anomaly oscillate between left and right boundary?
- Does this go away if you place a small temperature anomaly somewhere in the center of the domain (say 50 km radius, 50 K anomaly)? Currently there is no preferred movement for the material so the velocity field might not make much sense.

fionaclerc:
Hi Rene,
Thanks for your reply and suggestions. I did try them:

> * Does this also happen if you use a spherical shell model with the same setup (of course the elements will have a bad aspect ratio, but maybe it still runs)

This leads to small oscillations in the velocity field 
![spherical_shell_velY](https://user-images.githubusercontent.com/37747437/74687200-b49d1780-51a1-11ea-8be8-23fade0925eb.png)
[...]

PR #3375, opened 2020-02-19
Add mapping as argument to some boundary value calls

gassmoeller:
While looking into #3359 I came across these function calls that omit the mapping when interpolating boundary values. It does not fix the problems there, and it may not be a problem at all as we use a MappingQ1Eulerian for free surface at the moment, but it might be one in the future if we change that, or it might already be one if the mesh is already deformed? In the past we had some hard to find bugs because of forgotten mapping arguments, so I thought I better add these. Not sure if the tester will find a difference, lets see.


bangerth:
Nice find!

PR #1533, opened 2017-05-09
For dealii 9, use chunk manifold only, otherwise set correct boundary objects

anne-glerum:
See discussion in #1361. @gassmoeller Could you try it with dealii version 9?

PR #3377, opened 2020-02-21
Workaround for issue 3359

gassmoeller:
This is a workaround for #3359 and sits on top of PR #3376. The real changes are in the second commit: https://github.com/geodynamics/aspect/commit/f2cccf26a37a227f0265a21712ba3ba1113bbb44 . In short the general computation of the normal vector for the chunk manifold is too inaccurate for boundaries that are not aligned with cartesian coordinate directions. This leads to spurious strain rates and wrong velocity solutions (presumably because the normal vector is not aligned with the gravity leading the whole model to *slip* along the lower boundary as described by @fionaclerc in #3359. I copied the solution from the `SphericalManifold` class and verified that this reduces the magnitude of the residual velocities by around 7 orders of magnitude. I also checked that the behavior is identical to the `PolarManifold` class (in other words the polar manifold in deal.II currently has the same problem, and it is fixed if I apply the same workaround, I will do a PR for that later). 

There are two reasons this is only a workaround and can not be merged as is:
1. I do not understand where the problem originally comes from. All the pull_back and push_forward and push_forward_gradient functions are analytical expressions that I have checked for correctness and the computation of tangent and normal vector is pure algebra. Does that mean the whole residual comes from floating-point inaccuracies? Also the current change only fixes top and bottom boundaries (i.e. radius constant), however I suspect that the same problem might be responsible for #1671 and the residual velocity in 3D at constant latitude boundaries. For those boundaries I am not aware of a simple analytical expression for the normal vector.
2. This does currently not work if there are prescribed initial topographies. The Chunk Manifold has a complicated nested manifold for topography and it is not obvious to me how to write a normal_vector function that takes that into account (most likely I would need to combine a part of ChartManifold::normal_vector and ChunkGeometry::push_forward_gradient to combine the spherical and topography parts of the manifold).


tjhei:
I have to admit that I also don't understand the intricacies in Manifold, but what you are doing looks reasonable. @bangerth , can you take a look please?

bangerth:
> reduces the magnitude of the residual velocities by around 7 orders of magnitude

:-)

source/geometry_model/chunk.cc
```diff
@@ -249,6 +249,47 @@ namespace aspect
...
+      Tensor<1, dim>
+      ChunkGeometry<dim>::
+      normal_vector(const typename Triangulation<dim>::face_iterator &face,
+                    const Point<dim> &p) const
+      {
+        // if the maximum deviation for the distance from the vertices to the center
+        // is less than 1.e-5 of the minimum distance to the first vertex, assume we
+        // can simply return p-center. otherwise, we compute the normal using
+        // get_normal_vector
+        constexpr unsigned int n_vertices = GeometryInfo<dim>::vertices_per_face;
```
bangerth: If you don't have this function in place, what function would be called instead? The one that computes the normal based on pull-backs and push-forwards?

The issue in question is not that of the computation of the normal vector, but what the point `p` really is. In essence, we are given a point `p` in real coordinates of the cell. We need to find an accurate representation of the point in the reference coordinate system to use the pull-back/push-forward machinery. I think that what you are doing here makes sense to me -- it just computes a more accurate representation of the normal vector.

You might want to copy the documentation improvements from https://github.com/dealii/dealii/pull/9559.

bangerth:
Can you put a piece of code into the new function that compares the computed normal vector with the one the function in the base class would produce? I agree that in 2d, where pull-back and push-forward functions are explicit, the two should result in the same normal vector...

gassmoeller:
That makes a lot of sense, I can also confirm that the issue is fixed if we use the correct positions, instead of the ones provided by the the `MappingQ1Eulerian` for the call to `compute_no_normal_flux_constraints` (by handing over a `MappingManifold` object instead of the actual mapping). Would this be generally a good idea? All of our geometry models use manifolds that describe the geometry adequately, and the limitations of the `MappingManifold` class are not an issue inside `compute_no_normal_flux_constraints` because we only need `update_quadrature_points | update_normal_vectors`). A separate discussion maybe.

I also convinced myself that this should work for initial topography as well (my point 2. above). Because the initial topography actually deforms the mesh vertices they should no longer lie on the same radius and this shortcut should not trigger.

So in short, I think this fixes the issue and should be safe to merge, I added a simplified version of @fionaclerc's testcase as a new test.

gassmoeller:
> Can you put a piece of code into the new function that compares the computed normal vector with the one the function in the base class would produce? 

For testing? Or for the final version?

>I agree that in 2d, where pull-back and push-forward functions are explicit, the two should result in the same normal vector...
In fact for the chunk manifold all push-forward and pull-back functions are explicit for 3D as well. I added a corresponding test for the `PolarManifold` in the deal.II test. Here it is harder, because I have no access to a SphericalManifold inside that class.

anne-glerum:
@gassmoeller does this mean the issues in #1671 are fixed too? That's awesome!

anne-glerum:
> Also the current change only fixes top and bottom boundaries (i.e. radius constant), however I suspect that the same problem might be responsible for #1671 and the residual velocity in 3D at constant latitude boundaries. For those boundaries I am not aware of a simple analytical expression for the normal vector.

Maybe this is not what you mean, but in #1533 I added normal vectors for the conical latitude boundaries.

gassmoeller:
Ok, I think this is ready, I updated the tests and took over all changes from deal.II

> @gassmoeller does this mean the issues in #1671 are fixed too? That's awesome!

I am not a 100% sure, but I can not longer produce the problems we saw, so maybe?

> Maybe this is not what you mean, but in #1533 I added normal vectors for the conical constant-latitude boundaries.

Ah great that was what I was looking for. Let's first see if the problems resurface before implementing another special case though.

anne-glerum:
@gassmoeller With a branch from December as well as today's master and the input file given in #1671 I get the following velocity field:
Screenshot 2020-03-16 at 15 43 22
I.e. maximum velocity is very low, on the order of 3e-16 m/s (there are no driving forces). However, the higher velocities are only at the conical sides and abruptly end on the left and right boundaries, even though all vertical boundaries have the same, tangential, velocity boundary conditions. There is also an increase in velocity with depth, as before in the figure in #1533. Moreover, if I scale the inner and outer radius (now 1 and 2) with a factor 10, the min and max velocity are increased by a factor 100. 
So without your changes, the problem is still the same as in #1533.

Not sure if these very small gravity variations are problematic, but still quite funky:
Screenshot 2020-03-16 at 16 02 11

gassmoeller:
I see the same, so it looks like #1671 is not solved by this PR. 

Since #3359 depends on this we should still merge it, and think about the other issue later.

gassmoeller:
Could someone take another look at this PR? It is essentially a copy of dealii/dealii#9560 and fixes a significant bug (although for a special case).

PR #3378, opened 2020-02-22
add depth offset parameter for the world builder.

MFraters:
In response to issue https://github.com/GeodynamicWorldBuilder/WorldBuilder/issues/152 I looked for a way to set a free surface initial conditions in aspect through the world builder. The problem here is that the world builder defines everything with respect to the surface (which is used in computing temperatures for example), while in aspect depth 0 seem to be always the top of the model. 

I doubted whether it would be better to address this inside the world builder itself or withing aspect. I decided on putting the offset in aspect because this is not a change in the model itself, but how it is used in aspect. In other words, I think it is fundamentally still the same model since it is defined in the world with respect to the surface, so I don't think the offset should be defined in the world builder file, but in aspect.

Let me know what you think :)


source/simulator/parameters.cc
```diff
@@ -311,6 +311,12 @@ namespace aspect
                        Patterns::FileName(),
                        "Name of the world builder file. If empty, the world builder is not initialized.");
 
+    prm.declare_entry ("World builder depth offset", "0",
+                       Patterns::Double (),
+                       "This value offsets the depth given to the world builer. This can be "
+                       "usefull when setting a free surface for example since the world builder "
```
tjhei:
```suggestion
                       "useful when using a sticky air layer since the world builder "
```

gassmoeller:
Is the problem only that you need a reference surface instead of the (potentially variable) surface of the model domain? Have you considered if the `height_above_reference_surface` of the geometry model would be sufficient?

MFraters:
Thanks for the comments Timo and Rene! 

I did not conciser the `height_above_reference_surface` geometry model function. I tested it, and if you use it in the world builder plugin, instead of the depth, together with the initial topography plugin it works. What this solution changes is that the initial topography doesn't influence the initial conditions of the world builder anymore. When think of it, that might actually be a good thing since it aligns more with how the world builder itself works (everything is a depth with respect to a reference surface).

I will update the code and add some tests.

MFraters:
I have updated the pull request. I think this is the best solution for now, although I think that in the future we might want to have a offset parameter for the reference height in the geometry model subsection.

I have only added a test for cartesian box, because the chunk doesn't support the function initial topography. It might be worth it to at some point add a constant topography or add a offset to the zero topography function if someone would need it, but that would be a different issue.

I think it is ready to be reviewed again.

PR #3379, opened 2020-02-23
Fix hdf5 output after restart

maxrudolph:
This pull requests addresses a problem where subdomains belonging to different MPI processes are swapped in HDF5 visualization output following restart from a checkpoint. The proposed fix writes the h5 mesh file after ASPECT finishes resuming from a checkpoint.

### For all pull requests:

* [X ] I have followed the [instructions for indenting my code](../blob/master/CONTRIBUTING.md#making-aspect-better).

PR #3381, opened 2020-02-24
fix hdf5 output after restart

tjhei:
alternative to #3379

### DIFF
--- a/source/postprocess/visualization.cc
+++ b/source/postprocess/visualization.cc
@@ -246,7 +246,12 @@ namespace aspect
       // constructor. This will result in a new mesh file the first time we
       // create visualization output after resuming from a snapshot. Otherwise
       // we might get corrupted graphical output, because the ordering of
-      // vertices can be different after resuming.
+      // vertices can be different after resuming. (This is because when
+      // deal.II builds a triangulation, it tears down and re-creates its mesh every time
+      // p4est changes the partitioning; this process introduces a history where
+      // cells and vertices may be numbered differently if the triangulation
+      // had previous content than when -- as is done after a restart -- the
+      // triangulation is empty.)
     }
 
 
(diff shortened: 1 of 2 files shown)


tjhei:
(I did not test if this works, but we incorrectly serialized mesh_changed)

gassmoeller:
This makes sense as well. @maxrudolph could you check if this solves your problem as well? Then we do not need to connect another function to the signal as in #3379.
In whatever PR we move forward with we still need a changelog entry.

source/postprocess/visualization.cc
```diff
@@ -235,12 +235,18 @@ namespace aspect
...
       & times_and_pvtu_names
       & output_file_names_by_timestep
       & xdmf_entries
       ;
+
+      // We do not serialize mesh_changed but use the default (true) from our
+      // constructor. This will result in a new mesh file the first time we
+      // create visualization output after resuming from a snapshot. Otherwise
+      // we might get corrupted graphical output, because the ordering of
+      // vertices can be different after resuming.
```
bangerth: I had to think about this for a good long time. Can you add something like
```suggestion
      // vertices can be different after resuming. (This is because when
      // deal.II builds a triangulation, it tears down and re-creates its mesh every time
      // p4est changes the partitioning; this process introduces a history where
      // cells and vertices may be numbered differently if the triangulation
      // had previous content than when -- as is done after a restart -- the
      // triangulation is empty.)
```

maxrudolph:
I will test this shortly.

maxrudolph:
This fixes the problem. Thank you.

gassmoeller:
Awesome, Timo can you add a changelog entry like the one I proposed in #3379? After that feel free to merge yourself.

tjhei:
thank you @maxrudolph !

PR #3382, opened 2020-02-24
gmres for gmg cheap

tcclevenger:
Change the Stokes solve for GMG with the cheap preconditioner to use GMRES instead of FGMRES (save about 1/2 of the memory required for temporary vectors, also should be faster). This is accomplished by using a GMG v-cycle in place of the mass matrix solve (just like for the A-block cheap preconditioner). Like the A-block, when switching to the expensive preconditioner, we will switch to a CG solve of the mass matrix with the GMG v-vcycle as a preconditioner. 

Here is an example of the difference in runtimes and iteration counts between the old and new method. The iterations for any problem should not differ by more than a couple of iterations, and for larger meshes the new cheap preconditioner should be faster. 
iters-time

Also included is a small optimization for the `BlockSchurGMGPreconditioner` class where we now do not reconstruct the 2 temporary vectors needed everytime (only on the first vmult()). This save should become apparent when communication is significant in runtime (lower DoFs/core). I include it here since we are making other changes to the same class and vmult function.


gassmoeller:
Can you comment on whether the new S block approximation is more or less accurate than the old one? I am worried that if the new S block is worse it might deteriorate convergence of the outer loop in hard problems. Are the example numbers above from the sinker benchmark?

tcclevenger:
The outer iteration counts should be around the same for the CG solver vs. GMG vcycle, but if I would have to pick, I would expect the CG solve to give fewer outer iterations the majority of the time, but that isn't always the case and the iterations should be very similar. 

Here is an example which gives the outer iterations of the sinker benchmark. I consider DR_mu=1e4,1e6,1e8 with N-sinkers=4,8,12. These problems pass the point (200 cheap iterations) where the ASPECT default switches to the expensive solver (which returns to the CG solve for mass matrix in my new implementation). For the old way, mass matrix CG has the default 1e-6 S block tolerance. The iteration counts are almost always identical or differ very slightly.
iters-time

In my experience, the issue with tougher problems is more about the mass matrix as a poor approximation for the Schur complement more so then slightly less accurate mass matrix solve. (Also in the literature, many use only the diagonal of the mass matrix as a further approximation, which we could also do here, instead of GMG vcycle). This is only my experience from the few benchmarks I've worked with, though. If you have a more realistic setup I would be willing to implement something similar with AMG to test this.

But, it is still my goal to implement the BFBT stuff, which would make this obsolete. But I also realize it has been my goal for a long time.. ;)

gassmoeller:
> In my experience, the issue with tougher problems is more about the mass matrix as a poor approximation for the Schur complement more so then slightly less accurate mass matrix solve.

Yes that sounds reasonable. Ok, thanks for the additional testing. I will take a look at the code later today or tomorrow.

include/aspect/stokes_matrix_free.h
```diff
@@ -152,7 +152,8 @@ namespace aspect
         void fill_cell_data (const dealii::LinearAlgebra::distributed::Vector<number> &viscosity_values,
                              const double pressure_scaling,
                              const Triangulation<dim> &tria,
-                             const DoFHandler<dim> &dof_handler_for_projection);
+                             const DoFHandler<dim> &dof_handler_for_projection,
+                             const bool for_mg);
```
gassmoeller: The meaning of this parameter is not clear from its name. What does `for_mg` mean? The whole operator is used in a multigrid cycle. Please rename or explain in the documentation.

source/simulator/stokes_matrix_free.cc
```diff
@@ -450,14 +450,15 @@ namespace aspect
...
          */
-        BlockSchurGMGPreconditioner (const StokesMatrixType  &S,
-                                     const ABlockMatrixType  &A,
-                                     const MassMatrixType  &Mass,
-                                     const PreconditionerMp                     &Mppreconditioner,
-                                     const PreconditionerA                      &Apreconditioner,
-                                     const bool                                  do_solve_A,
-                                     const double                                A_block_tolerance,
-                                     const double                                S_block_tolerance);
+        BlockSchurGMGPreconditioner (const StokesMatrixType &S,
```
gassmoeller: The parameter names for this function are somewhat confusing (not only since your addition). Can we maybe reorganize so that `S = Stokes` is distinguishable from `S = SchurComplement`. Also we currently implicitly assume that the Schur complement is a mass matrix, but that might not stay like this. Would the following parameter be more reasonable (also asking @tjhei)?
```
        BlockSchurGMGPreconditioner (const StokesMatrixType &Stokes_matrix,
                                     const ABlockMatrixType &A_block,
                                     const MassMatrixType   &Schur_complement_block,
                                     const PreconditionerA  &A_block_preconditioner,
                                     const PreconditionerMp &Schur_complement_preconditioner,
                                     const bool              do_solve_A_block,
                                     const bool              do_solve_Schur_complement,
                                     const double            A_block_tolerance,
                                     const double            Schur_complement_tolerance);

File: source/simulator/stokes_matrix_free.cc [hunk 3]
```diff
@@ -475,42 +476,47 @@ namespace aspect
          */
         const StokesMatrixType &stokes_matrix;
         const ABlockMatrixType &velocity_matrix;
-        const MassMatrixType &mass_matrix;
-        const PreconditionerMp                    &mp_preconditioner;
-        const PreconditionerA                     &a_preconditioner;
+        const MassMatrixType   &mass_matrix;
```
gassmoeller: Same here, at the moment we implicitly assume that the reader knows that the `mass_matrix` is the `Schur complement`.

source/simulator/stokes_matrix_free.cc
tjhei: Looks reasonable to me. We can also change the template name from ``MassMatrixType`` to ``SchurComplementMatrixType`` or something. Similarly, ``PreconditionerMp``.

include/aspect/stokes_matrix_free.h
tcclevenger: This variable was needed since for both the `mass_matrix` and `velocity_matrix` needed both active and level versions. We used this variable to construct the correct cell_iterator when filling the viscosity data. 

There is a function in the deal.II `MatrixFree` class that tells us this information. I've removed the input parameter `for_mg` and call to the deal.II function inside the implementation of `fill_cell_data()`.

source/simulator/stokes_matrix_free.cc
tcclevenger: I agree with the renaming as well. I've changed many names and the documentation of the  `BlockSchurGMGPreconditioner` class, as well as a few names in the `StokesMatrixFreeHandlerImplementation` class  (e.g., `_M` in most places becomes `_Schur`).

I still keep the naming `MassMatrixOperator` for the matrix-free class that describes the pressure mass matrix.

tjhei:
please fix indentation, @tcclevenger

source/simulator/stokes_matrix_free.cc
```diff
@@ -429,160 +429,189 @@ namespace aspect
...
-                                                       ". It reported the following error:\n\n")
-                                           +
-                                           exc.what()))
-                  else
-                    throw QuietException();
-              }
-          }
-        dst.block(1) *= -1.0;
-      }
+      // either solve with the Schur complement matrix (if do_solve_M==true)
```
gassmoeller:
```suggestion
      // either solve with the Schur complement matrix (if do_solve_Schur_complement==true)
```

source/simulator/stokes_matrix_free.cc
```diff
@@ -1420,69 +1465,111 @@ namespace aspect
...
-    MassPreconditioner prec_S;
-    typename MassPreconditioner::AdditionalData prec_S_data;
-    prec_S_data.smoothing_range = 1e-3;
-    prec_S_data.degree = numbers::invalid_unsigned_int;
-    prec_S_data.eig_cg_n_iterations = mass_matrix.m();
-    prec_S_data.preconditioner = mass_matrix.get_matrix_diagonal_inverse();
-    prec_S.initialize(mass_matrix,prec_S_data);
+    // Schur complement matrix GMG
+    typedef PreconditionMG<dim, vector_t, MGTransferMatrixFree<dim,double> > SchurPreconditioner;
+    APreconditioner prec_Schur(dof_handler_p, mg_Schur, mg_transfer_Schur_complement);
```
gassmoeller:
```suggestion
    SchurPreconditioner prec_Schur(dof_handler_p, mg_Schur, mg_transfer_Schur_complement);
```

source/simulator/stokes_matrix_free.cc
```diff
@@ -1861,34 +1951,43 @@ namespace aspect
...
 
-    mg_constrained_dofs.clear();
-    mg_constrained_dofs.initialize(dof_handler_v);
+      mg_constrained_dofs_A_block.clear();
+      mg_constrained_dofs_A_block.initialize(dof_handler_v);
+
+      std::set<types::boundary_id> dirichlet_boundary = sim.boundary_velocity_manager.get_zero_boundary_velocity_indicators();
+      for (auto it: sim.boundary_velocity_manager.get_active_boundary_velocity_names())
+        {
+          int bdryid = it.first;
```
gassmoeller: Isnt there a `types::boundary_id` for this type?

source/simulator/stokes_matrix_free.cc
tcclevenger: This is a good catch. Now we only technically need one `typedef` since both preconditioners are of type `PreconditionMG<dim,dealii::LinearAlgebra::distributed::Vector<double>,MGTransferMatrixFree<dim,double>`. I just have `typedef PreconditionMG<...> GMGPreconditioner` for both now. Let me know if you like this fix, or still prefer to have the two different names (I can see a reason for both).

tjhei:
> but isn't the v-cycle also a different preconditioner in every iteration?

It is not a question if the preconditioner is "changing", but if it is a linear operator (which of course also implies that you get the same answer for the same residual). If you do things correctly, then the GMG vcycle (or the AMG vcycle) is a linear operator: conceptually, smoother and transfer are typically matrix-vector products (linear operators) and the coarse solver is as well (direct solver or fixed number of smoother steps). The combination of linear operators is again a linear operator.

You can of course break this by doing an iterative coarse solve, or weird smoothers. :-)

include/aspect/stokes_matrix_free.h
tcclevenger: It turns out that the function in the deal.II `MatrixFree` class that I mention is only in the development version of the code (which is why the last tester failed). I've readded the `for_mg` parameter under the name `is_mg_level_data`, and added a short explanation in the documentation.

tcclevenger:
I expect these tests to fail, and I will go and fix them now. Question: I see with Jenkins that I can look at the diff of the failed tests. Is there a way to look at the whole output file from the tester for a single test (so that I can copy and paste the entire screen output)?

tjhei:
> (so that I can copy and paste the entire screen output)?

just go to artifacts and download changes-test-results.diff. You can than apply it to your repo using ``patch -p1 <changes-test-results.diff``. There is no way to see/download the whole test output directly.

tjhei:
> There is no way to see/download the whole test output directly.

Well, except that you can run the tests locally inside a docker container and look at the results. See 6.8 in the pdf manual.

gassmoeller:
> Let me know if you like this fix, or still prefer to have the two different names

I am fine either way. The code looks good now.

> It is not a question if the preconditioner is "changing", but if it is a linear operator

Interesting, I never thought about it that way. On the other hand, all my knowledge about GMRES comes from "Iterative Methods for Sparse Linear Systems" (Saad), which essentially says "to see how to get from GMRES to FGMRES assume that the preconditioner matrix changes every iteration, know lets see how the algorithm changes". 

> You can of course break this by doing an iterative coarse solve, or weird smoothers. :-)

Ah, so using GMRES is only possible with the specific selection of smoothers and coarse solver we use at the moment. Is that possible to check or assert or at least worth documenting?

tjhei:
> Is that possible to check or assert or at least worth documenting?

I don't know if there is a good way to check this. All reasonable smoothers available in deal.II right now are okay. I am also not sure if the coarse solver matters enough to break GMRES (maybe?).
@tcclevenger can you leave a note in the code where you pick GMRES that FGMRES is not needed as the preconditioner (transfer/smoother/coarse solver) are linear operators?

tcclevenger:
> I don't know if there is a good way to check this. All reasonable smoothers available in deal.II right now are okay. I am also not sure if the coarse solver matters enough to break GMRES (maybe?).
> @tcclevenger can you leave a note in the code where you pick GMRES that FGMRES is not needed as the preconditioner (transfer/smoother/coarse solver) are linear operators?

Done. Also, I don't believe using a iterative coarse solver (with appropriate tolerance) would greatly affect the GMRES convergence, but I have not tested that.

doc/modules/changes/20200228_clevenger
```diff
@@ -0,0 +1,6 @@
+New: The geometric multigrid preconditioner for the Stokes system now uses
+a non-flexible version of the  preconditioner for the cheap iterations,
+alowing for the use of GMRES, which requires roughly half of the memory of
```
tjhei:
```suggestion
allowing for the use of GMRES instead of FGMRES, which requires roughly half of the memory of
```

tcclevenger:
All tests past. @gassmoeller, let us know if you have any other questions.

gassmoeller:
Looks good :+1:

gassmoeller:
Hi Conrad, just a quick update on this. I did some testing for a somewhat extreme model. Model setup: instantaneous, 3D, high viscosity jumps, only cheap iterations, restart length of 400 (I need a lot for this model to converge), stokes iterations: 280 (old), 264 (new). 

Timing output old:
```
| Total wallclock time elapsed since start     |       124s |            |
| Assemble Stokes system rhs       |         1 |     0.331s |      0.27% |
| Assemble composition system      |         1 |      0.41s |      0.33% |
| Assemble temperature system      |         1 |     0.564s |      0.46% |
| Build Stokes preconditioner      |         1 |     0.718s |      0.58% |
| Build composition preconditioner |         1 |    0.0976s |         0% |
| Build temperature preconditioner |         1 |     0.101s |         0% |
| Initialization                   |         1 |     0.472s |      0.38% |
| Postprocessing                   |         1 |      17.3s |        14% |
| Setup dof systems                |         1 |     0.704s |      0.57% |
| Setup initial conditions         |         1 |      10.2s |       8.3% |
| Setup matrices                   |         1 |     0.234s |      0.19% |
| Solve Stokes system              |         1 |      92.3s |        75% |
| Solve composition system         |         1 |     0.014s |         0% |
| Solve temperature system         |         1 |    0.0155s |         0% |
```

Timing output new:

```
| Total wallclock time elapsed since start     |      64.8s |            |
| Assemble Stokes system rhs       |         1 |     0.294s |      0.45% |
| Assemble composition system      |         1 |     0.417s |      0.64% |
| Assemble temperature system      |         1 |     0.554s |      0.85% |
| Build Stokes preconditioner      |         1 |     0.716s |       1.1% |
| Build composition preconditioner |         1 |     0.101s |      0.16% |
| Build temperature preconditioner |         1 |    0.0998s |      0.15% |
| Initialization                   |         1 |     0.439s |      0.68% |
| Postprocessing                   |         1 |        17s |        26% |
| Setup dof systems                |         1 |     0.676s |         1% |
| Setup initial conditions         |         1 |      10.6s |        16% |
| Setup matrices                   |         1 |     0.218s |      0.34% |
| Solve Stokes system              |         1 |      33.4s |        52% |
| Solve composition system         |         1 |    0.0139s |         0% |
| Solve temperature system         |         1 |    0.0168s |         0% |
```

This is a further speedup of the Stokes solver by a factor of 3! Pretty amazing :-)

tjhei:
wait, this is a 3x speedup because of this PR? How many Dofs/processes?

gassmoeller:
> wait, this is a 3x speedup because of this PR?

Yes as far as I can tell :-). I used git commits 1a4a501b3 (pre this PR) and 9a84bd335 (post this PR).

> How many Dofs/processes?

2 million dofs on 64 processes -> ~30 kdofs/core

I checked a few more numbers and attach the log and statistics files. This is the default setup of the matrix free constant grain size model of our FRES repo (S20RTS.prm), so you can repeat the model yourself.

I can think of 2 reasons for the improvement, but I do not see how they add up to such a big difference:
1. The number of S iterations drops, although it was not high before (around 2 S iterations per outer GMRES iteration, down to now 1 v-cycle per iteration).
2. The reduction in outer iterations cuts the last iterations, which are the most expensive ones (I guess that at 260 iterations the cost of the orthogonalization of the Krylov basis may be more expensive than the v-cycle for the A and S blocks).

Other reasons I can think of:

- I thought maybe the solution vectors just happen to fit into the cache. The hardware I use has 160 Mb of cache and according to the memory statistics postprocessor the solution vectors could use 180 MB (260*0.7Mb) for gmres and 400 Mb (280*0.7Mb * 2) for fgmres. However I repeated the run for a smaller model (one refinement less, 4k dofs/core) and I see exactly the same speed-up (this time a factor of 3.2 instead of 2.7 as for the larger model).
- Maybe there is a difference in the implementation of the GMRES and FGMRES solvers that explains the difference?

[statistics_fgmres.txt](https://github.com/geodynamics/aspect/files/4299707/statistics_fgmres.txt)
[statistics_gmres.txt](https://github.com/geodynamics/aspect/files/4299715/statistics_gmres.txt)

[log_fgmres.txt](https://github.com/geodynamics/aspect/files/4299706/log_fgmres.txt)
[log_gmres.txt](https://github.com/geodynamics/aspect/files/4299716/log_gmres.txt)

tjhei:
Wow, this is impressive. @tcclevenger what do you think? 
What I see as possible reasons:
1. "Also included is a small optimization for the BlockSchurGMGPreconditioner class where we now do not reconstruct the 2 temporary vectors needed everytime" - I assume this makes a bigger difference than @tcclevenger thinks.
2. S iterations are cut in half (267 instead of 564)
3. inner solve for S used in FGMRES also requires additional mat-vecs (one in each step plus to compute the residual)
4. GMRES uses half as many temporary vectors, so this would speed up computation if we are memory bound

Maybe it is worth testing 1) and 4) separately from 2&3. @tcclevenger , do you want to test this out? A comparison with AMG would also be very interesting. :-D

tcclevenger:
@gassmoeller this is great!

I have a strong suspicion that the reason that we see such a large gap in the times is that we no longer perform a CG solve on the active level to solve for the eigenvalues for the Chebyshev preconditioner of the mass matrix. Even if we did not switch to the vcycle, this needed to be fixed (and it is fixed for the expensive solver). Previously, there was the following line:
`prec_S_data.eig_cg_n_iterations = mass_matrix.m();`
meaning that an full CG solve on the pressure space with only a Jacobi preconditioner is used.

This would be easy to test by simply changing line 1483 in source/simulator/stokes_matrix_free.cc from the previous version (https://github.com/geodynamics/aspect/commit/1a4a501b3351296f09e99c37bcc08815a3512b3e) to be:
`prec_S_data.eig_cg_n_iterations = 100;`

I can test this if you like? Where can I find the FRES repo? Is it private?

gassmoeller:
I will send you the model and the output I have so far as email. The repo is private and the model is in a very preliminary state so there is not much point in posting it here.

tcclevenger:
The improvement I suggested actually did not make a large difference. On 28 cores with the same setup I also see around a factor of 3x faster for the new GMRES solve over the FGMRES one (improved with the 
`prec_S_data.eig_cg_n_iterations = 100;`
fix). From Timo's suggestion, I think it would be very interesting to test 1. I'll talk to Timo, but we can rerun some large NSinker runs on Palmetto and directly compare the 2.

tcclevenger:
I've found the increase in time. I'm testing with 32 cores.

Looking at the timing each component in BlockSchurGMGPreconditioner. Timings represent seconds/iterations: (278 for FGMRES, 270 for GMRES)
FGMRES:  A vcycle: 0.292371,         S^-1: 0.00943429,   B^T: 0.00431276
GMRES:    A vcycle: 0.166182,   S vcycle: 0.00992095,   B^T: 0.0042778

Interestingly enough, the ABlock GMG vcycle _by itself_ is half the time on master as before the latest changes. Looking at what changed for the Ablock, it was similar to what I mentioned above:
`smoother_data_A[0].eig_cg_n_iterations = mg_matrices_A_block[0].m();`
Basically, during our coarse solve we were doing a full CG solve for the largest eigenvalue during the first iteration. This did not show up in our timings over the last year since we were (unintentionally) initializing the smoother before the solve. 

In the new master version, we set 
`smoother_data_A[0].eig_cg_n_iterations = 100;`
 and this (interestingly enough) results in fewer iterations, even though the coarse solve gives a cruder approximation of the eigenvalues in the smoother.

Using this information, and changing a few more things around with a quick trial-and-error testing with other coarse solver parameters, I can get the following 
GMRES (221 iterations):    A vcycle: 0.0366724,   S vcycle: 0.00914907,   B^T: 0.0043631 
If I then compare the old and new version, we see:

old, FGMRES, no updates to smoother
fgmres

new, GMRES, updates to smoother
gmres

That's an over 6x speedup for the solve. It is not obvious to me why these parameters for the smoother can decrease the iterations so drastically, and so we need to do more testing on exactly what we should do for the coarse solver for problems with larger coarse meshes (I need to also consider switching to AMG). 

Any opinions on what we should change for master? I can test these updated parameters on different problems to see if we see the same improvements.

tjhei:
> It is not obvious to me why these parameters for the smoother can decrease the iterations so drastically,

The first time the preconditioner is applied, the eigenvalue solve of the CG is run. As the coarse mesh is big, this will take a lot of time. This is not at all surprising to me. 
My thoughts:
1. Things seem to be quite good as is.
2. We could move the initialization outside of the solve so we can time it separately so the user will know if the coarse mesh is the problem. Needs a feature inside deal.II.
3. Do we even need 100 iterations? It might be worth looking at the computed eigenvalues and how sensitive they are with respect to this.
4. Decide to switch to AMG on the coarse mesh. It might be easier to do this than spending too much time on the other items above.

gassmoeller:
Which additional changes did you make? I am happy about any further speedup.

> Do we even need 100 iterations? It might be worth looking at the computed eigenvalues and how sensitive they are with respect to this.
> Decide to switch to AMG on the coarse mesh. It might be easier to do this than spending too much time on the other items above.

Both sounds promising to me, but I know too little to recommend one over the other.

tcclevenger:
> The first time the preconditioner is applied, the eigenvalue solve of the CG is run. As the coarse mesh is big, this will take a lot of time. This is not at all surprising to me.

It is not surprising to me that the runtime is much faster, but only that making the coarse solver "weaker" we go from 278 -> 270 -> 221 iterations (see the parameters below).

@gassmoeller Here are the new parameters. I'll list 
[old version -> master version -> new (unmerged) version].
The old version took 278 iterations, master took 270 iterations, newest took only 221 iterations.

Lines 1491-1493 and lines 1517-1519:
`smoother_data_*[0].smoothing_range = 1e-3 -> 1e-3 -> 15;`
`smoother_data_*[0].degree = numbers::invalid_unsigned_int -> numbers::invalid_unsigned_int -> 8;`
`smoother_data_*[0].eig_cg_n_iterations = mg_matrices_A[0].m() -> 100 -> 10;`

> 1. Things seem to be quite good as is.

I agree, I'll create a new pull request changing the parameters.

> 2. We could move the initialization outside of the solve so we can time it separately so the user will know if the coarse mesh is the problem. Needs a feature inside deal.II.

Agree, I'll address this sometime soon. 

> 3. Do we even need 100 iterations? It might be worth looking at the computed eigenvalues and how sensitive they are with respect to this.

No, the new one above only has 10 (and is better).

> 4. Decide to switch to AMG on the coarse mesh. It might be easier to do this than spending too much time on the other items above.

Agree as well. Interested to see if we have the same low iteration counts (I'm guessing we will).

tjhei:
> Agree as well. Interested to see if we have the same low iteration counts (I'm guessing we will).

For small problems, AMG will be a direct solver (so yes). For bigger problems, I would assume that it is still quite good, as viscosity is smooth on the coarsest level. :-)

PR #3383, opened 2020-02-24
Add particle plugin existence and order checks

MFraters:
### For all pull requests:

### DIFF
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -1,5 +1,5 @@
 /*
- Copyright (C) 2015 - 2019 by the authors of the ASPECT code.
+ Copyright (C) 2015 - 2020 by the authors of the ASPECT code.
 
  This file is part of ASPECT.
 
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -537,6 +537,24 @@ namespace aspect
           get_needed_update_flags () const;
 
           /**
+           * Checks if the particle plugin specified by @p name exists
+           * in this model.
+           */
+          bool
+          pluginname_exists(const std::string &name) const;
+
+          /**
+           * Checks if the particle plugin specified by @p name acts
+           * before an other particle plugin in this model.
+           *
+           * Throws an assert when one of the plugin names does not
+           * exist. You can use the function pluginname_exists() to
+           * check in advance wheter a plugin exists
+           */
+          bool
+          check_plugin_order(const std::string &first, const std::string &second) const;
+
+          /**
            * Get the number of components required to represent this particle's
            * properties.
            *
@@ -631,6 +649,11 @@ namespace aspect
           write_plugin_graph (std::ostream &output_stream);
 
         private:
+          /**
+           * Stores the names of the plugins which are present
+           * in the order they are executed.
+           */
+          std::vector<std::string> plugin_names;
 
           /**
            * A list of property objects that have been requested in the
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -555,6 +555,11 @@ namespace aspect
           check_plugin_order(const std::string &first, const std::string &second) const;
 
           /**
+           * Get the plugin index of the particle plugin specified by @p name.
+           */
+          unsigned int get_plugin_index_by_name(const std::string &name) const;
+
+          /**
            * Get the number of components required to represent this particle's
            * properties.
            *
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -541,14 +541,15 @@ namespace aspect
            * in this model.
            */
           bool
-          pluginname_exists(const std::string &name) const;
+          plugin_name_exists(const std::string &name) const;
 
           /**
-           * Checks if the particle plugin specified by @p name acts
-           * before an other particle plugin in this model.
+           * Checks if the particle property plugin specified by @p first
+           * is executed before another particle property plugin specified
+           * by @p second.
            *
            * Throws an assert when one of the plugin names does not
-           * exist. You can use the function pluginname_exists() to
+           * exist. You can use the function plugin_name_exists() to
            * check in advance wheter a plugin exists
            */
           bool
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -550,7 +550,7 @@ namespace aspect
            *
            * Throws an assert when one of the plugin names does not
            * exist. You can use the function plugin_name_exists() to
-           * check in advance wheter a plugin exists
+           * check in advance whether a plugin exists
            */
           bool
           check_plugin_order(const std::string &first, const std::string &second) const;
@@ -699,4 +699,3 @@ namespace aspect
 }
 
 #endif
-
--- a/source/particle/property/interface.cc
+++ b/source/particle/property/interface.cc
@@ -393,6 +393,8 @@ namespace aspect
         return (update & (update_default | update_values | update_gradients));
       }
 
+
+
       template <int dim>
       bool
       Manager<dim>::plugin_name_exists(const std::string &name) const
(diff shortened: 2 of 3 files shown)


MFraters:
I noticed I needed an other function (get_plugin_index_by_name), so I have added that as well to this pull request.

include/aspect/particle/property/interface.h
```diff
@@ -536,6 +536,29 @@ namespace aspect
...
 
+          /**
+           * Checks if the particle plugin specified by @p name exists
+           * in this model.
+           */
+          bool
+          pluginname_exists(const std::string &name) const;
+
+          /**
+           * Checks if the particle plugin specified by @p name acts
```
gassmoeller:
```suggestion
           * Checks if the property plugin specified by @p first is executed
```

include/aspect/particle/property/interface.h
```diff
@@ -536,6 +536,29 @@ namespace aspect
...
+           * before an other particle plugin in this model.
+           *
+           * Throws an assert when one of the plugin names does not
+           * exist. You can use the function pluginname_exists() to
```
gassmoeller:
```suggestion
           * exist. You can use the function plugin_name_exists() to
```

source/particle/property/interface.cc
```diff
@@ -516,6 +550,7 @@ namespace aspect
 
             property_list.back()->parse_parameters (prm);
           }
+        plugin_names = prop_names;
```
gassmoeller: Why not remove `prop_names` and read into `plugin_names` instead? More code changes, but more intuitive.

unit_tests/particles.cc
```diff
@@ -0,0 +1,54 @@
...
+
+  // existing and listed pluring
+  REQUIRE(manager.pluginname_exists("composition") == true);
+  REQUIRE(manager.pluginname_exists("position") == true);
+
+  // existing but not listed plugin
+  REQUIRE(manager.pluginname_exists("pT path") == false);
+
+  // non-existed plugin
+  REQUIRE(manager.pluginname_exists("non-existend plugin name") == false);
```
gassmoeller:
```suggestion
  REQUIRE(manager.pluginname_exists("non-existent plugin name") == false);
```

unit_tests/particles.cc
```diff
@@ -0,0 +1,54 @@
+/*
+  Copyright (C) 2018 by the authors of the ASPECT code.
```
gassmoeller:
```suggestion
  Copyright (C) 2020 by the authors of the ASPECT code.
```

include/aspect/particle/property/interface.h
```diff
@@ -536,6 +536,29 @@ namespace aspect
           UpdateFlags
           get_needed_update_flags () const;
 
+          /**
+           * Checks if the particle plugin specified by @p name exists
+           * in this model.
+           */
+          bool
+          pluginname_exists(const std::string &name) const;
```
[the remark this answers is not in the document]
MFraters: ah, yes. I based this of the `fieldyname_exists` plugin thinking it might be some grammer thing. `plugin_name_exists` seems better to me too. I will change it.

MFraters:
Thanks! I have grown to like these kind of unit tests in the world builder because they are very cheap and show exactly what you are testing :)

I did think about the `required_other_postprocessors` mechanism, and although I like the idea, I decided not doing it now because 
1. it looked like quite a bit of work  
2. with what I am doing, the required plugin (most likely) needs to be setup by the user anyway, so if the user did not do that, I think for now it makes more sense to throw an assert and try to guess what the user exactly wants ;)

include/aspect/particle/property/interface.h
```diff
@@ -536,6 +536,30 @@ namespace aspect
...
+          plugin_name_exists(const std::string &name) const;
+
+          /**
+           * Checks if the particle property plugin specified by @p first
+           * is executed before another particle property plugin specified
+           * by @p second.
+           *
+           * Throws an assert when one of the plugin names does not
+           * exist. You can use the function plugin_name_exists() to
+           * check in advance wheter a plugin exists
```
bangerth:
```suggestion
           * check in advance whether a plugin exists
```

source/particle/property/interface.cc
```diff
@@ -393,6 +393,40 @@ namespace aspect
...
       }
 
+      template <int dim>
+      bool
+      Manager<dim>::plugin_name_exists(const std::string &name) const
+      {
+        return (std::find(plugin_names.begin(),plugin_names.end(),name) != plugin_names.end());
+      }
+
+      template <int dim>
```
bangerth:
```suggestion


      template <int dim>
```

source/particle/property/interface.cc
```diff
@@ -393,6 +393,40 @@ namespace aspect
...
+      bool
+      Manager<dim>::check_plugin_order(const std::string &first, const std::string &second) const
+      {
+        AssertThrow(plugin_name_exists(first),
+                    ExcMessage("Could not find a plugin with the name " + first + "."));
```
bangerth: It's often easier to read if you enclose information that's not part of the regular sentence in some sort of bracket:
```suggestion
                    ExcMessage("Could not find a plugin with the name <" + first + ">."));
```
Think of the case where the plugin name has multiple words with spaces in between.

source/particle/property/interface.cc
```diff
@@ -393,6 +393,40 @@ namespace aspect
...
+        AssertThrow(plugin_name_exists(second),
+                    ExcMessage("Could not find a plugin with the name " + second + "."));
```
bangerth:
```suggestion
                    ExcMessage("Could not find a plugin with the name <" + second + ">."));
```

source/particle/property/interface.cc
```diff
@@ -393,6 +393,40 @@ namespace aspect
...
+
+        return (std::find(plugin_names.begin(),plugin_names.end(),first) < std::find(plugin_names.begin(),plugin_names.end(),second));
```
bangerth:
```suggestion
        return (std::find(plugin_names.begin(),plugin_names.end(),first) 
                   < std::find(plugin_names.begin(),plugin_names.end(),second));
```

bangerth: What happens if `first == second`? Shouldn't we rule that out?

source/particle/property/interface.cc
```diff
@@ -393,6 +393,40 @@ namespace aspect
...
+      }
+
+      template <int dim>
```
bangerth:
```suggestion


      template <int dim>
```

source/particle/property/interface.cc
```diff
@@ -393,6 +393,40 @@ namespace aspect
...
+      unsigned int
+      Manager<dim>::get_plugin_index_by_name(const std::string &name) const
+      {
+        const std::vector<std::string>::const_iterator plugin = std::find(plugin_names.begin(),
+                                                                          plugin_names.end(),
+                                                                          name);
+
+        AssertThrow(plugin != plugin_names.end(),
+                    ExcMessage("The particle property manager was asked for a plugin "
+                               "with the name <" + name + ">, but no such plugin could "
```
bangerth: like here :-)

source/particle/property/interface.cc
```diff
@@ -393,6 +393,40 @@ namespace aspect
...
+                               "be found."));
+        return std::distance(plugin_names.begin(),plugin);
+      }
+
       template <int dim>
```
bangerth:
```suggestion


      template <int dim>
```

source/particle/property/interface.cc
MFraters: hmm, `1 < 1 = false`, I agree an assert might be more appropriate.

source/particle/property/interface.cc
MFraters: I thought the convention was one new line, but I now see that the most of this file is not consistent with the amount of new lines being used. Is two lines the convention?

source/particle/property/interface.cc
MFraters: That is the power of copying code ;)

source/particle/property/interface.cc
bangerth: Three empty lines between function definitions.

MFraters: So should I also add 3 lines between all the other function definitions in this file, or am I missing something?

MFraters:
I have addressed the comments.

source/particle/property/interface.cc
bangerth: In principle yes, but if it's for lines that you didn't touch in this patch, then let's not worry about it here. (But if you wanted to fix existing code in a separate PR, that is always welcome.)

MFraters: makes sense. I will make another pull request after this one to fix that.

Issue #3360, opened 2020-01-30
deprecated-copy warning


tjhei:
With gcc 9 I am getting
```
/work/02578/theister/frontera/aspect-git/include/aspect/simulator/assemblers/interface.h:383:16: warning: implicitly-declared ‘aspect::internal::Assembly::CopyData::StokesPreconditioner<2>& aspect::internal::Assembly::CopyData::StokesPreconditioner<2>::operator=(const aspect::internal::Assembly::CopyData::StokesPreconditioner<2>&)’ is deprecated [-Wdeprecated-copy]
  383 |         struct StokesSystem : public StokesPreconditioner<dim>
      |                ^~~~~~~~~~~~
/work/02578/theister/frontera/aspect-git/include/aspect/simulator/assemblers/interface.h:360:11: note: because ‘aspect::internal::Assembly::CopyData::StokesPreconditioner<2>’ has user-provided ‘aspect::internal::Assembly::CopyData::StokesPreconditioner<dim>::StokesPreconditioner(const aspect::internal::Assembly::CopyData::StokesPreconditioner<dim>&) [with int dim = 2]’
  360 |           StokesPreconditioner (const StokesPreconditioner &data);
      |           ^~~~~~~~~~~~~~~~~~~~
In file included from /opt/apps/gcc/9.1.0/include/c++/9.1.0/bits/specfun.h:45,
                 from /opt/apps/gcc/9.1.0/include/c++/9.1.0/cmath:1927,
                 from /work/02578/theister/frontera/deal-git/include/deal.II/base/numbers.h:28,
                 from /work/02578/theister/frontera/env8/deal-64/include/deal.II/base/config.h:427,
                 from /work/02578/theister/frontera/deal-git/include/deal.II/base/timer.h:19,
                 from /work/02578/theister/frontera/aspect-git/include/aspect/simulator.h:25,
                 from /work/02578/theister/frontera/aspect-git/source/simulator/assembly.cc:22:
/opt/apps/gcc/9.1.0/include/c++/9.1.0/bits/stl_algobase.h:703:11: note: synthesized method ‘aspect::internal::Assembly::CopyData::StokesSystem<2>& aspect::internal::Assembly::CopyData::StokesSystem<2>::operator=(const aspect::internal::Assembly::CopyData::StokesSystem<2>&)’ first required here
  703 |  *__first = __value;
      |  ~~~~~~~~~^~~~~~~~~

```

[...]

PR #3384, opened 2020-02-24
Fix deprecation warnings

gassmoeller:
This PR fixes a bunch of deprecation warnings with deal.II master and gcc 9.2.1.
I tested that everything runs with the most recent deal.II and the tester will take care of 9.1.1.


tjhei:
closes #3360

PR #3385, opened 2020-02-25
Avoid using deprecated functions.

bangerth:
These are all functions recently deprecated in deal.II.


bangerth:
Well, actually hold on: The patch that deprecates these functions has not even been submitted to deal.II :-) It's just been sitting on my hard drive...

bangerth:
The patch in question is this one here: https://github.com/dealii/dealii/pull/9566

bangerth:
We never actually merged that patch and decided to go a different route. So closing.

PR #3386, opened 2020-02-25
Simplify the logic of do_output_statistics.

bangerth:
First step towards addressing the issue mentioned yesterday on the forum by @maxrudolph . This just simplifies the logic a bit and will allow me, in the next step, access more member variables of the main class in the place where we actually do the writing.

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -189,43 +189,6 @@ namespace aspect
   }
 
 
-  namespace
-  {
-    /**
-     * A function that writes the statistics object into a file.
-     *
-     * @param stat_file_name The name of the file into which the result
-     * should go
-     * @param copy_of_table A copy of the table that we're to write. Since
-     * this function is called in the background on a separate thread,
-     * the actual table might be modified while we are about to write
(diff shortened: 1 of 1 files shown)


source/simulator/helper_functions.cc
```diff
@@ -189,43 +189,6 @@ namespace aspect
...
-     * of this function.
-     */
-    // We need to pass the arguments by value, as this function can be called on a separate thread:
-    void do_output_statistics (const std::string stat_file_name, //NOLINT(performance-unnecessary-value-param)
-                               const TableHandler *copy_of_table)
-    {
-      // write into a temporary file for now so that we don't
-      // interrupt anyone who might want to look at the real
-      // statistics file while the program is still running
-      const std::string tmp_file_name = stat_file_name + " tmp";
```
bangerth: Not sure if this was intentional, but it should probably have been `".tmp"` (with a period) all along. Fixed in the patch while there.

tjhei: looks like it, yes.

tjhei: looks like it, yes.

PR #3389, opened 2020-02-25
Only update the statistics file if possible, rather than writing it anew.

bangerth:
Following yesterday's discussion on the forum regarding the problem @maxrudolph saw.

/rebuild

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1750,6 +1750,23 @@ namespace aspect
        */
       TableHandler                        statistics;
 
+      /**
+       * The following two variables keep track which parts of the statistics
+       * object have already been written. This is because the TableHandler
+       * class has no way to keep track what it has already written, and so
+       * we can not just append the last row of the table to the output
+       * file. Rather, we keep track how many bytes we already wrote,
+       * and a has of what they contained, and if these so-many bytes have
+       * not changed between the previous and current write operation, then
+       * we only open the file in 'append' mode to add the new bytes from the
+       * last row. If what we would write now has changed from what we wrote
+       * back then in the first so-many bytes (e.g., because column widths of
+       * the table have changed), then we just replace the previous file by
+       * the current table contents in their entirety.
+       */
+      std::size_t                         statistics_last_write_size;
+      std::size_t                         statistics_last_hash;
+
       mutable TimerOutput                 computing_timer;
 
       /**
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1756,7 +1756,7 @@ namespace aspect
        * class has no way to keep track what it has already written, and so
        * we can not just append the last row of the table to the output
        * file. Rather, we keep track how many bytes we already wrote,
-       * and a has of what they contained, and if these so-many bytes have
+       * and a hash of what they contained, and if these so-many bytes have
        * not changed between the previous and current write operation, then
        * we only open the file in 'append' mode to add the new bytes from the
        * last row. If what we would write now has changed from what we wrote
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -289,19 +289,66 @@ namespace aspect
     auto write_statistics
       = [statistics_copy_ptr,this]()
     {
-      // Write into a temporary file for now so that we don't
-      // interrupt anyone who might want to look at the real
-      // statistics file while the program is still running
-      const std::string stat_file_name = parameters.output_directory + "statistics";
-      const std::string tmp_file_name = stat_file_name + ".tmp";
-
-      std::ofstream stat_file (tmp_file_name.c_str());
-      statistics_copy_ptr->write_text (stat_file,
+      // First write everything into a string in memory
+      std::ostringstream stream;
+      statistics_copy_ptr->write_text (stream,
                                        TableHandler::table_with_separate_column_description);
-      stat_file.close();
+      stream.flush();
 
-      // Now move the temporary file into place
-      std::rename(tmp_file_name.c_str(), stat_file_name.c_str());
+      const std::string statistics_contents = stream.str();
+
+      // Next find out whether we need to write everything into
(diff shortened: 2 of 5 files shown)


bangerth:
I checked on a couple of input files that the output is only written from scratch for the first couple of time steps. After that, we only get the update operation that appends a few bytes.

tjhei:
I am not sure I understand the logic. It looks like you never actually look at the existing statistics file but infer everything from internal data. 
What if we write a checkpoint, run two more timesteps, exit, then resume from the last checkpoint? Or did I misunderstand?

bangerth:
On 2/25/20 8:46 PM, Timo Heister wrote:
> I am not sure I understand the logic. It looks like you never actually look at 
> the existing statistics file but infer everything from internal data.

Correct. Ingenious, eh? ;-)


> What if we write a checkpoint, run two more timesteps, exit, then resume from 
> the last checkpoint? Or did I misunderstand?

As long as you don't delete the statistics file or change it in any other way, 
everything should work -- but you'll get the two new lines twice. I see the 
problem. We should just always write the whole file upon restart.

I know how to fix this, will do tomorrow morning.

tjhei:
> Correct. Ingenious, eh? ;-)

well... :-)

> We should just always write the whole file upon restart.

Yes. You can do this by not serializing the variables. Even better: check the file size to make sure it is identical to the last serialized size.

tjhei:
I have two more comments:
1. Serializing potentially 100s of MB of statistics data and formatting it into a string every step is probably not ideal and fast. Not sure how to work around that if you want to be robust with checkpointing.
2. Do we need to avoid opening/closing the file every step?

maxrudolph:
Why not just always output the entire checkpoint file after resume from
checkpoint by attaching to the appropriate signal? Then you can assume that
the statistics file is up to date and just print out the accumulated
values. You could put it within a maybe_write_statistics() function in the
same style as the other maybe_do_thisorthat() functions in core.cc and
helper_functions.cc.

bangerth:
On 2/26/20 10:56 AM, Timo Heister wrote:
> Yes. You can do this by not serializing the variables.

That's exactly my plan. Then the variables are initialized in the 
constructor to their default values, which leads a complete write-out as 
desired.

I'll get to it this pm when back home where I have the code on my laptop.


> Even better: 
> check the file size to make sure it is identical to the last serialized 
> size.

But that requires file system operations. I think that computing the 
hash is cheaper and achieves the same result.

bangerth:
>  1. Serializing potentially 100s of MB of statistics data and formatting
>     it into a string every step is probably not ideal and fast. Not sure
>     how to work around that if you want to be robust with checkpointing.

I suspect that it's not quite that much. @maxrudolph reports 281 bytes 
per time step, which seems reasonable. Even if you ran 100,000 time 
steps, you'd still only get to 28MB -- which is approximately the size 
of the 32MB he reports. I would expect that the overall output is much 
larger. We don't seem to see checkpointing in our profiles very much, so 
I don't think that it's an issue with run times.

It is a different issue whether we want to keep all of that data in 
memory all the time. But there, too, I don't see a lot of people running 
out of memory.

(As for the speed of formatting all of this: That's why this is running 
on a background thread.)


>  2. Do we need to avoid opening/closing the file every step?

Well, you sort of have to if you want to write data, don't you? The 
original approach tried to avoid that by just dumping everything into a 
file and then moving that file in place. But apparently that's a bad 
idea as well...

maxrudolph:
> >   2. Do we need to avoid opening/closing the file every step?
> Well, you sort of have to if you want to write data, don't you? The
> original approach tried to avoid that by just dumping everything into a
> file and then moving that file in place. But apparently that's a bad
> idea as well...

Why not just open the statistics file once when the computation starts (or resumes) and keep it open until the program exits, allowing the writes to be buffered?

bangerth:
Here is the updated patch, doing a complete rewrite after a restart.

include/aspect/simulator.h
```diff
@@ -1750,6 +1750,23 @@ namespace aspect
        */
       TableHandler                        statistics;
 
+      /**
+       * The following two variables keep track which parts of the statistics
+       * object have already been written. This is because the TableHandler
+       * class has no way to keep track what it has already written, and so
+       * we can not just append the last row of the table to the output
+       * file. Rather, we keep track how many bytes we already wrote,
+       * and a has of what they contained, and if these so-many bytes have
```
tjhei:
```suggestion
       * and a hash of the contents, and if these so-many bytes have
```

source/simulator/helper_functions.cc
```diff
@@ -289,19 +289,86 @@ namespace aspect
...
+            // case happens if the statistics_last_write_size is at the
+            // value initialized by the Simulator::Simulator()
+            // constructor, and this can happen in two situations:
+            // (i) At the end of the first time step; and (ii) at the
+            // end of the first time step after restart since the
+            // variable we query here is not serialized. It is clear that
+            // at the end of the first time step, we want to write the
+            // entire contents of the statistics object. For the second
+            // case, this is also appropriate since someone may have
+            // previously restarted from a checkpoint, run a couple of
```
tjhei: This also happens without a "previous restart" if the simulation continued after writing a checkpoint (very likely).

tjhei: the situation is much simpler: if you write a checkpoint and continue the simulation, more data is written into the statistics file (which will be duplicated if you resume from the checkpoint).

bangerth:
Excellent points. I've updated the documentation in these two places!

tjhei:
Notice that our checkpoint tests were failing in your first implementation because of duplicated lines:
https://jenkins.tjhei.info/blue/organizations/jenkins/aspect/detail/PR-3389/2/pipeline

Another win for good test coverage. Yay! :-)

bangerth:
I didn't notice -- but yes, it's good to know!

bangerth:
Want to merge?

PR #3391, opened 2020-02-27
Add particle property post initialize function

MFraters:
I need to initialize a member variable based on the particle property info, which is filled  in the initialize function. Therefore I some kind of post initialize function, which I have implemented here.

### For all pull requests:

### DIFF
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -314,6 +314,15 @@ namespace aspect
           initialize ();
 
           /**
+           * Initialization function. This function is called once directly
+           * after the initialize function when the property_information is
+           * set.
+           */
+          virtual
+          void
+          post_initialize ();
+
+          /**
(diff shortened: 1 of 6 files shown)


MFraters:
Yes, I can imagine that :)

The problem is that for the particle property plugins which process lpo particle property data need to know where the lpo data is stored in the data array and I want to be able to store this in the class so that I do not have to look it up for every particle and particle property:
```
const Particle::Property::Manager<dim> &manager = this->get_particle_world().get_property_manager();
lpo_data_position = manager.get_data_info().get_position_by_plugin_index(manager.get_plugin_index_by_name("lpo"));
```

So I need to be able to insert plugin specific code just after the line containing `property_information = ParticlePropertyInformation(info);` in the `manager::initialize()` function.

I have thought a bit about it and I think that I have found a less intrusive solution by reordering stuff in the initialize function as you suggested. What do you think?

PR #3393, opened 2020-02-27
Fix a doxygen flag.

bangerth:
The flag in question determines the doxygen output directory if one
generates LaTeX output -- which is not something that we probably want
to do, but it's still worth giving the flag a meaningful directory
name. (The previous value, 'me-latex', refers to the 'multiple experiment
inverse problem' code I wrote during my PhD thesis and from which this
options.dox file apparently derives.)


tjhei:
> he previous value, 'me-latex', refers to the 'multiple experiment
> inverse problem' code I wrote during my PhD thesis

fascinating!

PR #3395, opened 2020-02-27
Ensure three lines between function defintions in particle interface files

MFraters:
As title states. Follow up from discussion in #3383


MFraters:
It was about 10 minutes of work and since I am probably going to work a bit more on this part of the code I rather do it now than be annoyed half the time ;)

PR #3401, opened 2020-03-03
Three new features for the gravity_point_values postprocessor 

LudovicJnnt:
I updated the gravity_point_values postprocessor with 3 new features:

- profile sampling scheme to calculate gravity and its components radially along a profile from a minimum radius to maximum radius (only work for 1 point long/lat coordinate).

- new output file for some statistics; it includes the number of cells in the domain, the polynomial degree used and increased when called the postprocessor, average gravity acceleration, potential and gradients.

- replace density at quadrature points by PREM; PREM densities are given as a function of the radius. This feature has no other function than for a benchmark.

### For all pull requests:


gassmoeller:
Hi Ludo, thanks for proposing these changes. Could you please split this PR up into PRs that propose the changes individually? We usually try to keep PRs small and focussed on one change at a time, and the reason is that we could run into situations like with this PR, where I think some of the changes are great and could go in as is, while I would like to suggest some significant changes to others. Keeping them in one PR makes the review process more complicated and slows things down significantly (because the easy changes have to wait for the complicated ones, and the reviewer always has to look through all of them).

My general comments to the individual changes:
- I think the profile sampling scheme is great and can essentially go in as is (need to do a proper review on the final PR, but I like the concept and idea)
- Is there an important reason for starting a new statistics file? All postprocessors can add more quantities to our statistics file already (take a look at how material_statistics.cc does this for example), so that would be the usual place to add these quantities. We just fixed performance problems with our original statistics file, and if we start a new one here we could reintroduce those problems (see #3389). Some of your quantities are already in the existing statistics file (like the number of cells), others could be computed by other postprocessors (like the total volume, that could go into `source/postprocess/material_statistics.cc`), and the remaining ones could indeed be added by the gravity postprocessor. One additional question I have is if the averaging of individual components of gravity or gravity gradient is of any actual use? Wouldnt you only want to average magnitudes? Or do you want to have an average direction?
- Replacing the density by a hard-coded prem density is something that I see pretty critical. I can see the value of comparing against PREM, but I feel that defining a density and overwriting the data from the material data if a parameter is set in the postprocessor is not a clean way to do this (especially if it is only for one particular benchmark). I would suggest to instead make a new benchmark directory and create a material model in there that returns the PREM density. Alternatively, we could also think about a utility function that returns the PREM density (so essentially moving your code out of the postprocessor into a separate function). That might be useful in other places as well (I have one of my applications in mind).
- Adding a random perturbation to the coordinates is something that I see as valuable during testing and benchmarking, but do you think it will be of use to other users once testing has finished? Just adding features that might not be used again complicates maintaining the code for the future.

Let me know if you have any questions.

LudovicJnnt:
Hi Rene, 

Thanks for your valuable comments! 
Now you have made comments, I still would like to discuss them but maybe here is not the right place to do anymore if we close this PR? 
I will submit a new PR now starting with some small changes regarding the profile sampling scheme.

gassmoeller:
@LudovicJnnt: Can this pull request be closed? Did we add all the features you want in the other pull requests or is there something left to do?

LudovicJnnt:
@gassmoeller : One feature remain which is the PREM density model, but as you suggested, I will certainly create a material model that return the PREM density. Except by doing so, will the density also be evaluated at the nodes points of the finite element? so that if there is a sharp jump of density (upper crust - water), how this can be properly interpolated?

Apart from that I think you can close that PR.

gassmoeller:
> Except by doing so, will the density also be evaluated at the nodes points of the finite element? so that if there is a sharp jump of density (upper crust - water), how this can be properly interpolated?

The material model will evaluate the density at whatever position it is given in `in.position`. That usually means whenever you want to integrate over a volume it will be given the quadrature points (oftentimes gauss quadrature points that are in the inside of the cell). But you could also use a trapezoidal quadrature (`QTrapez` instead of `QGauss`), which has evaluation points at the vertices. There are a few places in the code already that do this, if you search for QTrapez you will find a few examples. I will close this PR, feel free to open a new one for the PREM density when you have something you want me to take a look at.

PR #3404, opened 2020-03-07
Improve allocation of matrix blocks

gassmoeller:
While testing some things for #3382 I noticed that in some cases we still allocate matrix blocks for temperature and composition even if not necessary. In particular that was the case for the `no Advection, single Stokes` solver scheme (but not the `first timestep only, single Stokes`) which was not tested for.
Additionally it was the case if any `no Advection` solver scheme was selected, but the compositional field method was left at a field method (the coupling would only check the compositional field method and decide that it needs a matrix block, even though the solver scheme would never actually solve the composition). Since the whole creation of the coupling table was quite messy I decided to refactor it into a separate function and fix the special cases. 
I also extended the check for unnecessary constrained matrix entries after creating the sparsity pattern. So far we only removed entries for the empty composition blocks, but still created entries for the velocity and pressure blocks even for the matrix free solver. The new method checks all matrix blocks and removes all entries from the sparsity pattern if these block do not couple with anything.


source/simulator/core.cc
```diff
@@ -870,125 +870,157 @@ namespace aspect
...
+        // necessary entries.
+        if (parameters.use_equal_order_interpolation_for_stokes == true)
+          coupling[x.pressure][x.pressure] = DoFTools::always;
+      }
 
-      for (unsigned int c=0; c<dim; ++c)
-        for (unsigned int d=0; d<dim; ++d)
-          coupling[x.velocities[c]][x.velocities[d]] = DoFTools::always;
+    // Now determine the coupling for the temperature block
+    coupling[x.temperature][x.temperature] = DoFTools::always;
```
tjhei: can you move remove this line and put it inside the if below by inverting the logic for the if?

source/simulator/core.cc
```diff
@@ -1011,8 +1043,7 @@ namespace aspect
         if (parameters.use_discontinuous_temperature_discretization)
           face_coupling[x.temperature][x.temperature] = DoFTools::always;
 
-        // Only allocate composition 0 matrix if needed. Same as the non-DG case (see above)
-        if (parameters.use_discontinuous_composition_discretization && have_fem_compositional_field)
+        if (parameters.use_discontinuous_composition_discretization)
```
tjhei: does this work correctly with your change here? What if you have 0 compositional fields?

source/simulator/core.cc
```diff
@@ -1048,20 +1079,24 @@ namespace aspect
...
+      for (unsigned int j=0; j<introspection.n_components; ++j)
+        {
+          if (coupling[i][j] == DoFTools::none)
+            {
+              const unsigned int block_i = introspection.block_index_for_component(i);
+              const unsigned int block_j = introspection.block_index_for_component(j);
+
+              // TODO: using clear() would be nice here but clear() also resets the
+              // size, so just reinit():
+              sp.block(block_i,block_j).reinit(sp.block(block_i,block_j).locally_owned_range_indices(),
```
tjhei: By the way, does this do anything for blocks not on the diagonal?

source/simulator/introspection.cc
```diff
@@ -328,6 +328,29 @@ namespace aspect
...
+        return block_indices.velocities;
+
+    if (component_index == component_indices.temperature)
+      return block_indices.temperature;
+
+    for (unsigned int i=0; i<n_compositional_fields; ++i)
+      if (component_index == component_indices.compositional_fields[i])
+        return block_indices.compositional_fields[i];
+
+    AssertThrow(false, ExcMessage("The requested component index does not exist."));
```
tjhei: I would use Assert() here.

source/simulator/core.cc
```diff
@@ -870,125 +870,157 @@ namespace aspect
...
-    bool have_fem_compositional_field = false;
-    for (unsigned int c=0; c<introspection.n_compositional_fields; ++c)
+    // Start by assuming nothing couples
+    coupling.fill (DoFTools::none);
+
+    const typename Introspection<dim>::ComponentIndices &x
+      = introspection.component_indices;
+
+    // The matrix-free solver does not work with melt transport
+    Assert(!(parameters.include_melt_transport && parameters.include_melt_transport), ExcNotImplemented());
```
tjhei: :-)

gassmoeller: Yes that was not quite the condition I had in mind :-) ... is fixed now.

gassmoeller:

> By the way, does this do anything for blocks not on the diagonal?

It does, although the difference is small. Here is the difference, 'matrix statistics' output with GMG on master:
```
Total system matrix memory consumption: 252.27 MB.
Total system matrix nnz: 99,245
system matrix nnz by block:
      98,310          64          64          64
          64          59          64          64
          64          64          59          59
          64          64          59          59

Total system preconditioner matrix memory consumption: 0.03 MB.
Total system preconditioner matrix nnz: 0
```

and with GMG and this branch:
```
Total system matrix memory consumption: 251.13 MB.
Total system matrix nnz: 0
system matrix nnz by block:
           0           0           0           0
           0           0           0           0
           0           0           0           0
           0           0           0           0

Total system preconditioner matrix memory consumption: 0.03 MB.
Total system preconditioner matrix nnz: 0
``` 

For comparison AMG with this branch:
```
Total system matrix memory consumption: 24131.30 MB.
Total system matrix nnz: 2,086,680,912
system matrix nnz by block: 
1,792,381,394 147,149,759           0           0
 147,149,759           0           0           0
           0           0           0           0
           0           0           0           0

Total system preconditioner matrix memory consumption: 7402.26 MB.
Total system preconditioner matrix nnz: 624,875,182
 614,146,806          64          64          64
          64  10,727,618          64          64
          64          64           0          59
          64          64          59           0
```

The main difference is of course the Stokes block that was wrongly allocated before, but there are entries in the off-diagonal blocks. I am somewhat surprised that the total system matrix memory still stands at 251.13 MB (for 16 million dofs), do you have an feeling on what might cause this? (the preconditioner is at 0.03 Mb). Still it might not be worth caring about the small numbers compared to the massive savings from AMG.

gassmoeller:
Ok, I think I need some help here. I found that the number of weird entries in off-diagonal blocks is exactly the number of MPI processes (64 in the example above). In other words, manually clearing the sparsity pattern blocks leads to problems (I still have failing tests after refinements happen), and would likely not be necessary if we figure out why there are exactly n_MPI_processes allocated entries in the off-diagonal blocks where there should be none. @tjhei do you have an idea where these entries come from? Or should I just drop this change from the PR as it is only a small effect anyway?

tjhei:
I think the way I am clearing out the matrices might be problematic. 

> do you have an idea where these entries come from?

Yes, this is a dummy entry we allocate to make things work. Let me try to find it in the code...

tjhei:
Here is the dummy element inserted on each processor:
https://github.com/dealii/dealii/blob/master/source/lac/trilinos_sparsity_pattern.cc#L797-L801

I am not sure if this is still needed, but I would not worry too much. Can you try only removing diagonal blocks? If that works, we can merge this PR.

source/simulator/core.cc
```diff
@@ -1048,19 +1099,29 @@ namespace aspect
...
+            // sparsity pattern in that case.
+            if ((i == introspection.component_indices.pressure) &&
+                coupling[introspection.component_indices.velocities[0]][introspection.component_indices.velocities[0]] != DoFTools::none)
+              continue;
+
+            const unsigned int block = introspection.get_components_to_blocks()[i];
+
+            // TODO: using clear() would be nice here but clear() also resets the
+            // size, so just reinit():
+            sp.block(block,block).reinit(sp.block(block,block).locally_owned_range_indices(),
```
tjhei: this looks good to me now! :+1:

tjhei:
```
The following tests FAILED:
	374 - melt_compressible_advection (Failed)
	384 - melt_property_visualization (Failed)
	390 - melt_transport_compressible (Failed)
```
with tiny differences in the output. Can this be related to you cleaning out the melt pressure block?

gassmoeller:
No, I had accidentally copied over the coupling of the preconditioner S block (the one between fluid and compaction pressure) into the main matrix coupling. Not sure why that caused the differences, but it is not necessary and I reverted that change. Also I squashed the commits.

tjhei:
Looks good to me. :+1:

tjhei:
Please merge yourself if tests pass.

PR #3406, opened 2020-03-09
update chebyshev parameters

tcclevenger:
Updates the Chebyshev parameters for the Coarse GMG solver. Previously, the coarse solver contained a full CG solve of eigenvalues. This is not an issue on meshes with small coarse levels, but is quite slow otherwise. 

An example of the time saved (over 4x faster) is discussed in https://github.com/geodynamics/aspect/pull/3382.

All current GMG tests should have the same output (or slightly lower iteration counts). 

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1523,6 +1523,18 @@ namespace aspect
       mg_smoother_Schur.initialize(mg_matrices_Schur_complement, smoother_data_Schur);
     }
 
+    // Estimate the eigenvalues for the Chebyshev smoothers.
+    for (unsigned int level = 0; level<sim.triangulation.n_global_levels(); ++level)
+      {
+        vector_t temp_velocity;
+        vector_t temp_pressure;
+        mg_matrices_A_block[level].initialize_dof_vector(temp_velocity);
+        mg_matrices_Schur_complement[level].initialize_dof_vector(temp_pressure);
+
+        mg_smoother_A[level].estimate_eigenvalues(temp_velocity);
+        mg_smoother_Schur[level].estimate_eigenvalues(temp_pressure);
+      }
+
     // Coarse Solver is just an application of the Chebyshev smoother setup
     // in such a way to be a solver
     //ABlock GMG
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1488,9 +1488,9 @@ namespace aspect
             }
           else
             {
-              smoother_data_A[0].smoothing_range = 15.;
+              smoother_data_A[0].smoothing_range = 1e-3;
               smoother_data_A[0].degree = 8;
-              smoother_data_A[0].eig_cg_n_iterations = 10;
+              smoother_data_A[0].eig_cg_n_iterations = 100;
             }
           smoother_data_A[level].preconditioner = mg_matrices_A_block[level].get_matrix_diagonal_inverse();
         }
@@ -1514,9 +1514,9 @@ namespace aspect
             }
           else
             {
-              smoother_data_Schur[0].smoothing_range = 15.;
+              smoother_data_Schur[0].smoothing_range = 1e-3;
               smoother_data_Schur[0].degree = 8;
-              smoother_data_Schur[0].eig_cg_n_iterations = 10;
+              smoother_data_Schur[0].eig_cg_n_iterations = 100;
             }
           smoother_data_Schur[level].preconditioner = mg_matrices_Schur_complement[level].get_matrix_diagonal_inverse();
         }
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1524,7 +1524,13 @@ namespace aspect
     }
 
 #if DEAL_II_VERSION_GTE(9,2,0)
-    // Estimate the eigenvalues for the Chebyshev smoothers.
+    // Estimate the eigenvalues for the Chebyshev smoothers. If not running with
+    // deal.II 9.2.0, the eigenvalue estimate will be performed during the first
+    // application of the Chebyshev smoother during the solve.
+
+    //TODO: The setup for the smoother (as well as the entire GMG setup) should
+    //       be moved to an assembly timing block instead of the Stokes solve
+    //       timing block (as is currently the case).
     for (unsigned int level = 0; level<sim.triangulation.n_global_levels(); ++level)
       {
         vector_t temp_velocity;
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1523,6 +1523,7 @@ namespace aspect
       mg_smoother_Schur.initialize(mg_matrices_Schur_complement, smoother_data_Schur);
     }
 
+#if DEAL_II_VERSION_GTE(9,2,0)
     // Estimate the eigenvalues for the Chebyshev smoothers.
     for (unsigned int level = 0; level<sim.triangulation.n_global_levels(); ++level)
       {
@@ -1534,6 +1535,7 @@ namespace aspect
         mg_smoother_A[level].estimate_eigenvalues(temp_velocity);
         mg_smoother_Schur[level].estimate_eigenvalues(temp_pressure);
       }
+#endif
 
     // Coarse Solver is just an application of the Chebyshev smoother setup
     // in such a way to be a solver
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1488,9 +1488,9 @@ namespace aspect
             }
           else
             {
-              smoother_data_A[0].smoothing_range = 1e-3;
-              smoother_data_A[0].degree = numbers::invalid_unsigned_int;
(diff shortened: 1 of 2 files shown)


gassmoeller:
Looks like this fails for `sol_cx_2_gmg`. Maybe we need to do more testing on the optimal parameters? Or test the AMG option.

tcclevenger:
@gassmoeller Looking at the test output, all iteration counts are identical, but the residual values are all slightly different (this is to be expected since we are very slightly changing the preconditioner). But effectively speaking, the results are identical, which makes since sense all tests use very small coarse meshes (all 1 cell, I think). Later today I will update the tests.

gassmoeller:
Most tests do what you say, but `sol_cx_2_gmg` gives the following:
```
An error occurred in line <406> of file </home/dealii/dealii-v9.0.1/include/deal.II/lac/solver_cg.h> in function
void dealii::SolverCG<VectorType>::solve(const MatrixType&, VectorType&, const VectorType&, const PreconditionerType&) [with MatrixType = aspect::MatrixFreeStokesOperators::MassMatrixOperator<2, 1, double>; PreconditionerType = dealii::PreconditionMG<2, dealii::LinearAlgebra::distributed::Vector<double>, dealii::MGTransferMatrixFree<2, double> >; VectorType = dealii::LinearAlgebra::distributed::Vector<double>]
The violated condition was:
beta != 0.
Additional information:
A piece of code is attempting a division by zero. This is likely going to lead to results that make no sense.
```

tjhei:
I am not quite sure why the solve fails. On my machine with the latest deal.II master, the test sol_cx_2_gmg passes.
Conrad, here are the eigenvalues estimates with your settings for the test:
```
*** Timestep 0:  t=0 seconds
   Solving Stokes system... DEAL:GMRES::Check 0 0.0536849
DEAL:GMRES::Starting value 0.0536849
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.323227 max=2.43640 cg-iters=10
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.333333 max=1.80000 cg-iters=5
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=1.00000 max=1.00000 cg-iters=1
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.0974690 max=2.12743 cg-iters=10
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.242839 max=1.87836 cg-iters=10
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.735615 max=1.67964 cg-iters=5
```
and for ../tests/shell_3d_gmg.prm.dealii_9.2:
```
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.199692 max=3.88570 cg-iters=10
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.182478 max=4.06419 cg-iters=10
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.116467 max=3.51202 cg-iters=10
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.0782969 max=3.02494 cg-iters=10
```
same as above except setting the iterations to 100:
```
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.199692 max=3.88570 cg-iters=10
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.122320 max=4.07926 cg-iters=39
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.116467 max=3.51202 cg-iters=10
DEAL:GMRES::PreconditionChebyshev: eigenvalue estimate min=0.0171089 max=3.07877 cg-iters=100
```
This means we don't converge, even with 100 iterations here. Results are slightly different as well.

Regarding your changes:
- I think you want a smoothing range <1 if you want to use it as a solver.
- We should check that your results are robust when doing the maximum number of iterations. 10 is not enough for the coarse solve as you can see.

tjhei:
> * I think you want a smoothing range <1 if you want to use it as a solver.

The eigenvalues are about 0.017 to 3.07 for the spherical shell coarse mesh, so a smoothing range of [max_ev/smoothing_range, max_ev]=[3.07/15,3.07] is not enough to cover [0.017,3.07]. See documentation of PreconditionChebyshev.

tcclevenger:
Hi guys,sorry for the delay. I've updated the smoothing range to be less than 1 for the coarse Chebyshev, and I have increased the coarse iterations to 100. I believe this should fix the problem.

> I think you want a smoothing range <1 if you want to use it as a solver.

I think this is the key change from the output of the eigenvalues. The eigenvalue solve should not have to be completely accurate (I don't think) as long as the range is correct.

> We should check that your results are robust when doing the maximum number of iterations. 10 is not enough for the coarse solve as you can see.

From the constant_grain_size problem, if we use the max number of iterations then the coarse solve is much too expensive, so if this is the case I think we will have to implement the AMG coarse solver. The question would be what we do in the meantime? Assuming the tests pass with the current version I have just pushed, should we keep that setup or return to the previous way of exact eigen-solves on the coarsest level and prioritize coarseAMG?

tjhei:
> if we use the max number of iterations then the coarse solve is much too expensive

You mean the setup ,right? How expensive is the setup for 100 iterations compared to the whole v-cycle?

> Assuming the tests pass with the current version I have just pushed, should we keep that setup or return to the previous way of exact eigen-solves on the coarsest level and prioritize coarseAMG?

If 100 iterations gives a working coarse solve and is not too expensive compared to the solve itself, I would suggest using that for now.

Now you are only changing the degree. What difference does it make?

tcclevenger:
> You mean the setup ,right? How expensive is the setup for 100 iterations compared to the whole v-cycle?

I have pulled the `estimate_eigenvalues()` out in the setup. Here is a test where I put separate timing blocks around the eig estimate and the GMRES solve:

**eig_cg_n_iterations=10**
Iterations _____ Cheby eigenvalue time ______ Solve time
224 __________ 0.029s  ___________________ 6.6s

**eig_cg_n_iterations=100**
Iterations _____ Cheby eigenvalue time ______ Solve time
238  __________  0.038s  __________________ 7.2s

**eig_cg_n_iterations=mg_matrices[0].m()**
Iterations _____ Cheby eigenvalue time ______ Solve time
238 ____________ 0.050s _________________  7.2s

It turns out that the eigenvalue estimate was not the problem in the time increase from the old setup, but it actually came from the degree being set to `numbers::invalid_unsigned_int` (see below).

> If 100 iterations gives a working coarse solve and is not too expensive compared to the solve itself, I would suggest using that for now.

I agree. 
 
> Now you are only changing the degree. What difference does it make?

So it turns out that the degree being set to `numbers::invalid_unsigned_int` was the cause of the vast majority of the time increase. It was calculating the degree necessary to be quite large, causing the smoother to be very expensive. Here is an example:

**degree=8, eig_cg_n_iterations=100**
Iterations _____ Cheby eigenvalue time ______ Solve time
238 ___________ 0.04s  ___________________ 7.1s

**degree=numbers::invalid_unsigned_int, eig_cg_n_iterations=100**
Iterations _____ Cheby eigenvalue time ______ Solve time
275 __________ 0.04s ____________________ 42.0s

I set degree=8 based on trial and error from the constant_grain_size problem. All the tests have such small coarse grid that this shouldn't matter as long as the smoothing range is properly set, which it is now. 

New: `degree=8` and `smoothing_range<1` and `eig_cg_n_iterations=100` and `estimate_eigenvalues()` called after creating smoother.

tjhei:
> All the tests have such small coarse grid that this shouldn't matter as long as the smoothing range is properly set, which it is now.

Well, the number of iteration increases from 238 to 275, so the coarse solve *does* matter for that problem.

Anyways, I think this is an improvement and should be merged as is. I will let @gassmoeller take a look as well, though.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1515,14 +1515,28 @@ namespace aspect
...
+              smoother_data_Schur[0].degree = 8;
+              smoother_data_Schur[0].eig_cg_n_iterations = 100;
             }
           smoother_data_Schur[level].preconditioner = mg_matrices_Schur_complement[level].get_matrix_diagonal_inverse();
         }
       mg_smoother_Schur.initialize(mg_matrices_Schur_complement, smoother_data_Schur);
     }
 
+#if DEAL_II_VERSION_GTE(9,2,0)
+    // Estimate the eigenvalues for the Chebyshev smoothers.
```
gassmoeller: Please extend the comment to say why you do this (is it faster? is it for error checking?). Even after our conversation I do not understand, but maybe that is just me :-)

gassmoeller:
I am happy to merge this after you extend on the one comment I marked up. Thanks for doing this, excited about the improvements :+1:

source/simulator/stokes_matrix_free.cc
tcclevenger: Eventually we would like to move the setup of the GMG preconditioner into an assembly timing block (probably "Build Stokes preconditioner") instead of begin contained within the "Solve Stokes system" timing block, and so this new feature in deal.II (being able to estimate the Chebyshev eigenvalues outside the smoother application) gets us closer to this. But there is no time save here.

tjhei:
Excellent, thanks @tcclevenger. Put the timing blocks on your todo list, ok?

PR #3407, opened 2020-03-10
Diffusion of surface boundary topography

anne-glerum:
This PR takes the topography of a mesh boundary (in most cases the surface boundary), which includes any initial topography plus later mesh deformation, and diffuses this topography by solving a dim-1 diffusion equation in the dim mesh_deformation FE space. Zero gradient boundary conditions are assumed. The plugin returns a boundary velocity by computing the difference in topography divided by the timestep.

I have added a 2D benchmark setup (/benchmarks/1_sine_zero_flux.prm) in which an initial hill with zero topography on the left and right boundaries is diffused away. At every timestep, the postprocessor 'analytical topography' outputs both the actual topography and the analytical solution of the topography.

The results of the above benchmark look promising (Fig. 1&2) as the maximum topography decreases while the height of the surface at the model sides increases accordingly. However, at t1, there is a unexpected shift of the mesh vertices in the top of the domain to the right (so horizontal motion). After this first shift, motion is mostly vertical as expected.
Topo_benchmark_AMR2
_Fig. 1 Computed (lines) and analytical (symbols) surface topography over time for low resolution, showing the initial shift in surface mesh vertices to the right._
Topo_benchmark_AMR5
_Fig. 2 Computed (lines) and analytical (symbols) surface topography over time for a higher resolution, showing the initial shift in surface mesh vertices to the right._

The horizontal shift to the right is independent of resolution and timestep size for as far as I could see. The shift leads to distorted elements in the right top domain corner, and for some settings the diffusion solve then fails. Stricter values of solver tolerance or higher diffusivities also cause the model to fail, but I don't know whether this is because of a new problem, or because of the initial horizontal motion. 

mesh_t0-t4
_Fig. 3 Mesh at time t0, t1, t2 and t3 to show the deformation of the elements and the changing aspect ratio towards the right boundary._

In other words, I'd appreciate input on the cause of this shift and any other comments on the implementation. I also haven't thought about fixed topography boundary conditions.

### For all pull requests:


anne-glerum:
Hi @gassmoeller , thanks for your suggestions. 

> As I wrote in the comment, the right hand side of the diffusion equation looks somewhat suspicious to me, I do not understand why the j dof index appears in there and it is inside the inner loop. (see source/simulator/assemblers/advection.cc:363 for a classical diffusion rhs). Again maybe this is correct because it is solving on a surface of a volume, but something to check.

Yes. So at first I didn't have the j dof index and assembled the rhs in the outer loop as usual. This also gave the shift and didn't fit the analytical solution, and after another look at the equations to solve, I thought this was needed. The fit with the analytical solution is better, but the shift is still there. Now I'm confused and should have another look, but in *both* cases the shift to the right occurs and depending on the boundary conditions the shift is larger for the one or the other. See graphs below. 
Topo_benchmark_AMR4_t1_both

Increasing the global resolution does not affect the size of the shift, but does lead to spurious displacements (black lines).
Topo_benchmark_AMRx_t5


> What happens if you run this with the no Advection, no Stokes solver scheme? This should only trigger the surface deformation (Stokes velocity of zero). Do you still see the shift?

Changing the solver scheme does not change the results. 
Additionally, setting all Stokes BC to no slip does not change the shift, but fixes the corner points of the surface to 0 topography.
Setting just the top B to no slip does not change anything.

> I could imagine there is some weird interaction between the tangengial mesh velocity constraints and the diffusion plugin constraints going on, do you see the same x-shift going on if you remove the tangential side mesh velocities? Not adding these constraints to the diffusion constraints will then likely lead to strongly deformed boundary cells, but at least we would see if it is symmetric

This does not change the shift. Actually at t1 the topography is exactly the same, with vertical displacement along the side boundaries, because the sides are free slip. Setting them to zero slip, just fixes the side nodes.

> This reminds me of the symmetry problems you had reported for free-surface models, maybe the two have a common cause? Am I paranoid about this? On the upside if we could find the problem here maybe we fix it for both cases.

Maybe. None of the setups I tried show a shift to left, so the positive x-direction seems to be preferred. Could the time stepping scheme (first order backward) have anything to do with it? It seems so weird the shift is only at t1.

gassmoeller:
I think I found the problem. Your diffusion problem is currently set up as a vector-valued problem for x- and y- mesh displacements. In that form you can diffuse the y-deformation into the x-deformation (which is why in the first timestep a lot of y-displacement converts into x-displacement; when both are the same both decay at the same rate). What you probably really want is to only diffuse the y- displacements and have no lateral movement of mesh nodes because of this diffusion. I modified the assembly loop to only assemble the diffusion problem for the y-component, which gives visually correct results:
```
// Loop over the shape functions
                    for (unsigned int i=0; i<dofs_per_cell; ++i)
                      {
                        if (mesh_deformation_dof_handler.get_fe().system_to_component_index(i).first != dim-1)
                          continue;

                        // Assemble the RHS
                        // RHS = M*H_old
                        cell_vector(i) += displacement *
                                          fs_fe_face_values.shape_value (i, point) *
                                          fs_fe_face_values.JxW(point);

                        for (unsigned int j=0; j<dofs_per_cell; ++j)
                          {
                            if (mesh_deformation_dof_handler.get_fe().system_to_component_index(j).first != dim-1)
                              continue;

                            // Assemble the matrix, for backward first order time discretization:
                            // Matrix := (M+dt*K) = (M+dt*B^T*kappa*B)
                            cell_matrix(i,j) +=
                              (
                                this->get_timestep() * diffusivity *
                                (projection * fs_fe_face_values.shape_grad(i, point)) * (projection * fs_fe_face_values.shape_grad(j,point)) +
                                fs_fe_face_values.shape_value (i, point) * fs_fe_face_values.shape_value (j, point)
                              )
                              * fs_fe_face_values.JxW(point);
                          }
                      }
```

This seems to work for your box, however I am not sure how to do this best for curved geometries, where we can not simply check the vector component of displacement. Maybe you would need to add constraints to the `mass_matrix_constraints` that only allow movement along the direction of gravity? Or the normal of the surface? Maybe the function `compute_normal_flux_constraints` from deal.II can help?

anne-glerum:
ah @gassmoeller awesome, thank you! I will test it somewhere in the coming days and update you!

anne-glerum:
Hi @gassmoeller just to give an update, I reran the benchmark with your suggestions and the shift in the x-direction is indeed gone. However, the diffusion solver still crashes sooner or later, depending on resolution, diffusivity and tolerance. Comparing the computed and analytical topography over time shows that the difference is largest at the boundaries and from there grows inwards over time. I'm looking into that some more. 
Topo_benchmark_AMR6_t1
Topo_benchmark_AMR6_t14

gassmoeller:
Ok, good. Let me know if you want some more input. Is #3480 related to this PR?

anne-glerum:
Hi Rene, I looked a bit closer at the solutions of the diffusion problem and noticed very large anomalies:
Screenshot 2020-05-26 at 17 11 11
Looking at the settings of the Jacobi Preconditioner, I then set a small number to avoid division by zero for elements on the matrix diagonal. This gives a smooth solution that is close to the analytical solution:
Screenshot 2020-05-26 at 17 13 23
For example, here I tested the difference with the analytical solution over time for different timesteps:
Topo_benchmark_1_convergence_timestep
For a benchmark with fixed topography at the top corner points, the error is smoother (here varying solver tolerance):
Topo_benchmark_2_convergence_tolerance

anne-glerum:
What do you think about the change to the preconditioner?

These benchmarks were run with the `no Advection, no Stokes` solver, I'll now setup a test with some internal dynamics.

gassmoeller:
Yes this makes a lot of sense. Glad to see the results are better. Did it work for a dynamic model too?

anne-glerum:
Yes, it works for dynamic models too. 

I applied to the `continental_extension` test with a free surface. Below a graph of the topography in the center of the domain. The black lines represent the reference model with only a free surface. Dark grey model applies diffusion, the light grey one only every 2nd timestep. The coral model increased the diffusivity with a factor 10. Smoothing of topography as expected. 

Note that these models only ran for 2 My on level 6 resolution, so I will test this setup on a high resolution model that actually creates steep gradients. Also, at the moment the plugin applies one diffusion timestep of the same length as the regular timestep. I think it would be good to set a maximum timestep for diffusion and loop over the diffusion to reach the regular timestep. I'll test the impact/necessity with FastScape first. 

Topo_continental_extension_zoom

anne-glerum:
I think this is almost ready for a first review. When running it in parallel however, I found that there is some mismatch in the size of the distributed Vectors. Reshuffling the order of addition in lines 364 and down to 
   
       LinearAlgebra::Vector d_displacement(mesh_locally_owned, this->get_mpi_communicator());
       d_displacement = displacements;
       d_displacement *= -1.;
       d_displacement += solution;
       d_displacement -= initial_topography;

helps, but still fails at subtracting `initial_topography`. I had a look at the bit in `mesh_deformation/interface.cc` where I set the initial topography, but I couldn't spot why it would go wrong there. Do you have an idea? Also, do you agree with the way I set the initial topography? I could also extract it from the mesh itself instead of going through the topography plugin. 


> Technically the backward euler should be stable no matter the timestep, but it might be overly diffusive. One question would be if the flow is ever able to create gradients that are steeper than can be accurately diffused away (i.e. if our advection timestep is already enough). There is already some code for computing diffusion timesteps in `helper_functions.cc::compute_time_step()`, that should be adaptable for topography diffusion.

Good point. I've added in a check that fails if the diffusion timestep is smaller than the current timestep.

> On a separate note (and beyond the scope of this PR) I was always a bit concerned that we use a first-order scheme for the free surface advection (and now also for diffusion), while using BDF2 for the temperature equation (and the Stokes equation if using the projected density approximation). At some point it may be worth checking if that affects the overall accuracy of the solution, and if we can unify our time-stepping schemes. Let me know in case you are interested in investigating that further.

Yes, I'm definitely up to investigate that further.

gassmoeller:
> I had a look at the bit in mesh_deformation/interface.cc where I set the initial topography, but I couldn't spot why it would go wrong there. Do you have an idea? 

I am not 100% sure, but it seems the vectors are not distributed in the same way across processes. This is even more likely if you replace your lines by this:
```
      d_displacement.add(-1.0, displacements);
      d_displacement.add(-1.0, initial_topography);
```
Which produces the following message:
```
An error occurred in line <1958> of file </home/rene/software/deal.II-v9.0.1/include/deal.II/lac/trilinos_vector.h> in function
    void dealii::TrilinosWrappers::MPI::Vector::add(dealii::TrilinosScalar, const dealii::TrilinosWrappers::MPI::Vector&)
The violated condition was: 
    local_size() == v.local_size()
Additional information: 
    Dimension 30 not equal to 40.
```
It looks like they should be created identically, but maybe one of them is created before the initial mesh refinement and the other afterwards? And does the `initial_topography` get redistributed during mesh refinements? Maybe it needs to be regenerated after every refinement? Just ideas, not sure which one is the best path forward.

> I could also extract it from the mesh itself instead of going through the topography plugin.

If you extract it from the mesh: How would you know which part of the mesh deformation is initial topography and which part is accumulated deformation? Or do you mean before any movement?

anne-glerum:
> It looks like they should be created identically, but maybe one of them is created before the initial mesh refinement and the other afterwards? And does the initial_topography get redistributed during mesh refinements? Maybe it needs to be regenerated after every refinement?

The initial topography is treated the same as the mesh displacements upon mesh refinement, so it should be redistributed.

I thinking it's due to mixing ghosted and non-ghosted vectors to get to the displacement difference. The `initial_topography` and `mesh_displacements` vectors have the same local size and so do the `solution` and `d_displacement`, but a different size. The former include `mesh_locally_relevant`, the latter don't.

> If you extract it from the mesh: How would you know which part of the mesh deformation is initial topography and which part is accumulated deformation? Or do you mean before any movement?

I thought instead of calling for the topography plugin to get the topography once before movement, I could store the height of the mesh node. This would include the initial unperturbed location + applied topography, but that wouldn't matter for the diffusion, it's just an upward shift.

anne-glerum:
This plot shows the topography at 4 timesteps on the left side of the domain of the continental extension cookbook where I changed the free surface velocity projection to normal instead of vertical and increased the extension velocity to make it fail due to element distortion. With a bit of diffusion, the model runs to completion :)

Topo_continental_extension_diffusion_HRLN_zoom_left

naliboff:
> With a bit of diffusion, the model runs to completion :)

@anne-glerum fantastic news!

anne-glerum:
@gassmoeller I've added benchmarks and tests, I think it's ready for review now :)

gassmoeller:
Thanks, was a bit busy this week (moving), will take a look asap next week.

bangerth:
I'm going to look at this once back from vacation in ten days. But that shouldn't stop a merge if someone else wanted to take the initiative before me.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+      cg.solve (mass_matrix, solution, system_rhs, preconditioner_mass);
+
+      // Distribute constraints on mass matrix
+      mass_matrix_constraints.distribute (solution);
+
+      // The solution contains the new displacements, but we need to return a velocity.
+      // Therefore, we compute v=d_displacement/d_t.
+      // d_displacement are the new mesh node locations
+      // minus the old locations, which are initial_topography + displacements.
+      LinearAlgebra::Vector d_displacement(mesh_locally_owned, mesh_locally_relevant, this->get_mpi_communicator());
```
bangerth, line 368: call the variable `velocity`, because that's what you're putting into it.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+          start_timestep = this->get_timestep_number();
+        }
+
+      // Set the current time and timestep
+      current_time = (this->get_time());
+      const unsigned int current_timestep_number = this->get_timestep_number();
+
+      // Determine whether we need to apply diffusion based
+      // on the time or timestep interval between applications.
+      if (current_timestep_number != 0)
```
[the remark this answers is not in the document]
anne-glerum: Mesh velocity is solved for first thing at the beginning of each time step. At t0, mesh velocity is zero, because Stokes velocity is zero. So I didn't want to apply this mesh deformation when free surface itself isn't applied either. Also, when setting initial topography at t0, would you want it to diffuse at t0 already, or would you expect to diffuse in the following timestep?

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+                  }
+              }
+
+      // Get the global minimum timestep
+      const double min_conduction_timestep = - Utilities::MPI::max (-min_local_conduction_timestep, this->get_mpi_communicator());
+
+      AssertThrow (min_conduction_timestep > 0.,
+                   ExcMessage("The time step length for diffusion of the surface needs to be positive, "
+                              "but the computed step length was: " + std::to_string(min_conduction_timestep) + ". "
+                              "Please check for non-positive diffusivity."));
```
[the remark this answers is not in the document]
anne-glerum: Yep. And I don't allow negative diffusivities in declare_parameters, so even the Assert would be unnecessary.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+
+      double conduction_timestep = min_conduction_timestep;
+      if (this->convert_output_to_years())
+        conduction_timestep /= year_in_seconds;
```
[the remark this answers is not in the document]
anne-glerum: This function computes the conductive timestep and compares it to Aspect's advective timestep and fails the Assert when it's smaller. In some of our rift setups it turned out that the timestep actually was larger than the conductive timestep. Although the timestepping scheme should be stable, I kept it in anyway (as I haven't tested different timesteps extensively).

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+
+    template <int dim>
+    void Diffusion<dim>::diffuse_boundary(const DoFHandler<dim> &mesh_deformation_dof_handler,
+                                          const IndexSet &mesh_locally_owned,
+                                          const IndexSet &mesh_locally_relevant,
+                                          LinearAlgebra::Vector &output,
+                                          const std::set<types::boundary_id> boundary_ids) const
+    {
+      // Check that the current timestep does not exceed the diffusion timestep
+      compute_time_step(mesh_deformation_dof_handler, boundary_ids);
```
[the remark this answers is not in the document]
anne-glerum: See my previous answer. It checks that the conductive timestep is >= than Aspect's advective timestep.

benchmarks/diffusion_of_hill/README
```diff
@@ -0,0 +1,30 @@
+This directory contains to benchmarks of a 2D hill diffusion away
```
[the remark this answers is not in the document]
anne-glerum: Thanks! I switched to QtCreator recently, and haven't found a working spell check unfortunately.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+    void Diffusion<dim>::parse_parameters(ParameterHandler &prm)
+    {
+      // The list of tangential Stokes velocity boundary indicators.
+      tangential_boundary_velocity_indicators = this->get_boundary_velocity_manager().get_tangential_boundary_velocity_indicators();
+      // The list of zero Stokes velocity boundary indicators.
+      zero_boundary_velocity_indicators = this->get_boundary_velocity_manager().get_zero_boundary_velocity_indicators();
+      // The list of prescribed Stokes velocity boundary indicators.
+      const std::map<types::boundary_id, std::pair<std::string,std::vector<std::string> > > active_boundary_velocity_indicators =
+        this->get_boundary_velocity_manager().get_active_boundary_velocity_names();
+      for (std::map<types::boundary_id, std::pair<std::string, std::vector<std::string> > >::const_iterator p = active_boundary_velocity_indicators.begin();
```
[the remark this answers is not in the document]
anne-glerum: I went with `auto` instead of `std::map<types::boundary_id, std::pair<std::string, std::vector<std::string> > >::const_iterator`, is that ok too?

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+      mass_matrix.compress(VectorOperation::add);
+
+      // Jacobi seems to be fine here.  Other preconditioners (ILU, IC) run into trouble
+      // because the matrix is mostly empty, since we don't touch internal vertices.
+      LinearAlgebra::PreconditionJacobi preconditioner_mass;
+      LinearAlgebra::PreconditionJacobi::AdditionalData preconditioner_control(1,1e-16,1);
+      preconditioner_mass.initialize(mass_matrix, preconditioner_control);
+
+      this->get_pcout() << "   Solving mesh surface diffusion" << std::endl;
+      SolverControl solver_control(50*system_rhs.size(), this->get_parameters().linear_stokes_solver_tolerance*system_rhs.l2_norm());
```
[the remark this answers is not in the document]
anne-glerum: Mmm yes, this is a leftover from trying to make the solver work before I added the precondition_control line. I put it back to 5, which is what's used for the free surface.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+      SolverCG<LinearAlgebra::Vector> cg(solver_control);
+      cg.solve (mass_matrix, solution, system_rhs, preconditioner_mass);
```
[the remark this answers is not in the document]
anne-glerum: I think I do the same as in the free surface implementation, which I copy here below for ease:
```
      // Jacobi seems to be fine here.  Other preconditioners (ILU, IC) run into troubles
      // because the matrix is mostly empty, since we don't touch internal vertices.
      LinearAlgebra::PreconditionJacobi preconditioner_mass;
      preconditioner_mass.initialize(mass_matrix);

      SolverControl solver_control(5*rhs.size(), this->get_parameters().linear_stokes_solver_tolerance*rhs.l2_norm());
      SolverCG<LinearAlgebra::Vector> cg(solver_control);
      cg.solve (mass_matrix, dist_solution, rhs, preconditioner_mass);
```

However, this gave me very unstable results, with only certain solver settings leading to solver convergence. This was fixed when I added
```
      LinearAlgebra::PreconditionJacobi::AdditionalData preconditioner_control(1,1e-16,1);
```
which takes care of the zero entries in the matrix. See the discussion above your comments.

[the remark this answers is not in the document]
anne-glerum: I remember that last hack we looked at the free surface implementation with Timo as well, and decided it could be better (see issue #3110), but works for now.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+              }
+
+      system_rhs.compress (VectorOperation::add);
+      mass_matrix.compress(VectorOperation::add);
+
+      // Jacobi seems to be fine here.  Other preconditioners (ILU, IC) run into trouble
+      // because the matrix is mostly empty, since we don't touch internal vertices.
+      LinearAlgebra::PreconditionJacobi preconditioner_mass;
+      LinearAlgebra::PreconditionJacobi::AdditionalData preconditioner_control(1,1e-16,1);
+      preconditioner_mass.initialize(mass_matrix, preconditioner_control);
```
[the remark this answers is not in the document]
anne-glerum: Yes, I haven't tried any other preconditioners, because of the comment (taken from the free surface implementation) that they won't like the zero matrix entries. But I could look into that.

include/aspect/mesh_deformation/interface.h
```diff
@@ -314,6 +321,12 @@ namespace aspect
          */
         void compute_mesh_displacements ();
 
+        /**
+         * Set up the Vector with initial displacments of the mesh
```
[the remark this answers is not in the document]
anne-glerum: Thanks! I switched to QtCreator recently, and haven't found a working spell checker unfortunately.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,583 @@
...
+                for (unsigned int point=0; point<n_fs_face_q_points; ++point)
+                  {
+                    // Get the gravity vector to compute the outward direction of displacement
+                    Tensor<1,dim> direction = -(this->get_gravity_model().gravity_vector(fs_fe_face_values.quadrature_point(point)));
+                    // Normalize direction vector
+                    if (direction.norm() > 0.0)
+                      direction *= 1./direction.norm();
+                    // TODO this is only correct for box geometries
+                    else
+                      direction[dim-1] = 1.;
```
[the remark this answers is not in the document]
anne-glerum: I agree a gravity of zero wouldn't make sense, but you could still diffuse a hill with the equation we solve.

anne-glerum:
@bangerth Ready for another look and discussion :)

source/mesh_deformation/diffusion.cc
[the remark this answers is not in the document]
bangerth: Ah, it's executed at the beginning of the time step. Can you add your explanation to the comment?

source/mesh_deformation/diffusion.cc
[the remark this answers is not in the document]
bangerth: Yes, definitely!

source/mesh_deformation/diffusion.cc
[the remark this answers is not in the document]
bangerth: Right, but then why does it make sense to assume that the vertical direction is `(0,0,1)`? I think at some point we will probably also want to enable this for spherical geometries, and then this isn't right any more. I'd just error out if that's the case -- there are probably other problems as well if gravity is ever zero.

include/aspect/mesh_deformation/diffusion.h
```diff
@@ -0,0 +1,185 @@
...
+         * used to determine whether diffusion should
+         * be applied.
+         */
+        double start_time;
+
+        /**
+         * The current time used to determine whether diffusion should
+         * be applied.
+         */
+        double current_time;
```
bangerth, line 121: Is this and the previous value equal to what the `Simulator` class stores? I would very much prefer if we could query it from there, rather than duplicating information -- this kind of thing has a tendency to go out of sync, with very difficult-to-find bugs.

include/aspect/mesh_deformation/diffusion.h
```diff
@@ -0,0 +1,185 @@
...
+         * to the boundary with the given mesh deformation boundary indicator.
+         */
+        double time_between_diffusion;
+
+        /**
+         * The start timestep of the model run
+         * used to determine whether diffusion should
+         * be applied.
+         */
+        unsigned int start_timestep;
```
bangerth, line 141: same here

include/aspect/mesh_deformation/diffusion.h
```diff
@@ -0,0 +1,185 @@
...
+        /**
+         * Whether or not in the current timestep, diffusion
+         * should be applied.
+         */
+        bool apply_diffusion;
+
+        /**
+         * Boundaries along which the Stokes velocity is set to tangential.
+         */
+        //std::set<types::boundary_id> tangential_boundary_velocity_indicators;
```
bangerth, line 163: just remove it if you don't need it

include/aspect/mesh_deformation/diffusion.h
```diff
@@ -0,0 +1,185 @@
...
+        /**
+         * Boundaries along which the mesh is allowed to move tangentially
+         * despite of the Stokes velocity boundary conditions.
+         */
+        std::set<types::boundary_id> additional_tangential_mesh_boundary_indicators;
+
+        /**
+         * Boundaries along which the Stokes velocity is set to zero.
+         */
+        //std::set<types::boundary_id> zero_boundary_velocity_indicators;
```
bangerth, line 174: same for this and the next variable

include/aspect/mesh_deformation/free_surface.h
```diff
@@ -130,7 +130,7 @@ namespace aspect
          */
         struct SurfaceAdvection
         {
-          enum Direction { normal, vertical };
+          enum Direction { normal, vertical, geometric_normal };
```
bangerth, line 133: Can you document what these flags mean?

include/aspect/mesh_deformation/interface.h
```diff
@@ -237,6 +237,13 @@ namespace aspect
         const std::set<types::boundary_id> &
         get_free_surface_boundary_indicators () const;
 
+        /**
+         * Return the initial topography stored on
+         * the Q1 finite element that describes the mesh geometry.
+         */
+        const LinearAlgebra::Vector &
+        get_initial_topography () const;
```
bangerth, line 245: Is this a vector for which only those nodes on the boundary have meaning? If so, can you say so?

include/aspect/mesh_deformation/interface.h
```diff
@@ -314,6 +321,20 @@ namespace aspect
          */
         void compute_mesh_displacements ();
 
+        /**
+         * Set up the vector with initial displacements of the mesh
+         * due to the initial topography, as supplied by the initial
+         * topography plugin. We set all entries to the initial topography
+         * based on its surface coordinates, i.e. the initial topography
+         * is not corrected for depth from the surface as it is
+         * for the initial mesh deformation. TODO this is ok for now,
```
bangerth, line 330: is this something we need to worry about now?

include/aspect/mesh_deformation/interface.h
```diff
@@ -411,6 +442,13 @@ namespace aspect
          */
         std::set<types::boundary_id> free_surface_boundary_ids;
 
+        /**
+         * A pointer to the initial topography model.
+         */
+        InitialTopographyModel::Interface<dim> *topo_model;
```
bangerth, line 448: Rather than storing a plain pointer, is there a way how we can access this via `SimulatorAccess`?

source/mesh_deformation/free_surface.cc
```diff
@@ -307,15 +307,27 @@ namespace aspect
 
                 cell_vector = 0;
                 cell_matrix = 0;
+                Tensor<1,dim> direction;
+                if (advection_direction == SurfaceAdvection::geometric_normal)
+                  // Only for 2D
+                  {
```
bangerth, line 313: Add an assertion if it only works for a certain case.

source/mesh_deformation/free_surface.cc
```diff
@@ -307,15 +307,27 @@ namespace aspect
...
+                    Tensor<1,dim> tangent = (fs_fe_face_values.quadrature_point(n_face_q_points-1) - fs_fe_face_values.quadrature_point(0));
+                    direction[0] = -tangent[1];
+                    direction[1] = tangent[0];
+                    // Check that normal points upwards
+                    if (direction[1] < 0)
+                      direction *= -1.;
```
bangerth, line 319: This seems very specific to the box geometry. If that's what you want, you should probably also add an assertion for that.

source/mesh_deformation/free_surface.cc
```diff
@@ -307,15 +307,27 @@ namespace aspect
...
-                    Tensor<1,dim> direction;
                     if ( advection_direction == SurfaceAdvection::normal ) // project onto normal vector
-                      direction = fs_fe_face_values.normal_vector(point);
+                      {
+                        direction = fs_fe_face_values.normal_vector(point);
+                      }
                     else if ( advection_direction == SurfaceAdvection::vertical ) // project onto local gravity
                       direction = this->get_gravity_model().gravity_vector(fs_fe_face_values.quadrature_point(point));
-                    else
+                    else if (advection_direction != SurfaceAdvection::geometric_normal)
```
bangerth, line 330: now that you have 3 cases, it would probably be worthwhile making this into a `switch...case` thing

include/aspect/mesh_deformation/free_surface.h
bangerth: Separately, why are the existing options not good enough?

source/mesh_deformation/interface.cc
```diff
@@ -93,6 +97,24 @@ namespace aspect
     }
 
 
+    template <int dim>
+    void
+    MeshDeformationHandler<dim>::initialize ()
+    {
+      // Get pointer to initial topography model
+      topo_model = const_cast<InitialTopographyModel::Interface<dim>*>(&this->get_initial_topography_model());
```
bangerth, line 105: Ah, that's really unfriendly. Why do you need to cast away the `const`ness? This suggests some kind of design failure.

source/mesh_deformation/interface.cc
```diff
@@ -93,6 +97,24 @@ namespace aspect
...
+      // In case we prescribed initial topography, we should take this into
+      // account. However, it is not included in the mesh displacements,
+      // so we need to fetch it separately.
```
bangerth, line 108: This is something that will go away with @gassmoeller recent patch

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+
+    template <int dim>
+    void
+    Diffusion<dim>::update ()
+    {
+      // Initialize the start time and timestep
+      if (std::isnan(start_time))
+        {
+          start_time = (this->get_time());
+          start_timestep = this->get_timestep_number();
```
bangerth, line 77: as mentioned above, I'd like to avoid storing duplicate information.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+        }
+
+      // Set the current time and timestep
+      current_time = (this->get_time());
+      const unsigned int current_timestep_number = this->get_timestep_number();
```
bangerth, line 82: same here -- though I could accept the local `const` variable

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+      // zero mesh velocities, which means the zero/prescribed Stokes velocity
+      // boundaries minus those that are listed as Additional
+      // tangential boundary indicators.
+
+      // The list of prescribed Stokes velocity boundary indicators.
+      const std::map<types::boundary_id, std::pair<std::string,std::vector<std::string> > > active_boundary_velocity_indicators =
+        this->get_boundary_velocity_manager().get_active_boundary_velocity_names();
+      std::set<types::boundary_id> prescribed_boundary_velocity_indicators;
+      //for (std::map<types::boundary_id, std::pair<std::string, std::vector<std::string> > >::const_iterator p = active_boundary_velocity_indicators.begin();
+      //     p != active_boundary_velocity_indicators.end(); ++p)
```
bangerth, line 142: just remove it if you don't need it

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+      // no_normal_flux_constraints, which means all
+      // minus the diffusion mesh deformation boundary indicators
+      // and minus the zero/prescribed Stokes velocity boundary indicators.
+      std::set<types::boundary_id> x_no_flux_boundary_indicators = this->get_boundary_velocity_manager().get_tangential_boundary_velocity_indicators();
+      x_no_flux_boundary_indicators.insert(additional_tangential_mesh_boundary_indicators.begin(),additional_tangential_mesh_boundary_indicators.end());
+      for (std::set<types::boundary_id>::const_iterator p = x_no_flux_boundary_indicators.begin();
+           p != x_no_flux_boundary_indicators.end(); ++p)
+        if (boundary_ids.find(*p) != boundary_ids.end())
+          {
+            x_no_flux_boundary_indicators.erase(*p);
```
bangerth, line 170: I think you can do this more easily with `std::set_minus`. If you grep for it in the code base, @gassmoeller used this somewhere recently.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+      sp.compress();
+      matrix.reinit (sp);
+#endif
+
+      LinearAlgebra::Vector system_rhs, solution;
+      system_rhs.reinit(mesh_locally_owned, this->get_mpi_communicator());
+      solution.reinit(mesh_locally_owned, this->get_mpi_communicator());
+
+      // Initialize Gauss-Legendre quadrature for degree+1 quadrature points of the surface faces
+      QGauss<dim-1> face_quadrature(mesh_deformation_dof_handler.get_fe().degree+1);
```
bangerth, line 215:
```suggestion
      const QGauss<dim-1> face_quadrature(mesh_deformation_dof_handler.get_fe().degree+1);
```

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+      FEValuesExtractors::Vector extract_vertical_displacements(0);
+
+      // An extractor for the dim-valued initial topography vectors
+      // Later on we will compute the gravity-parallel displacement
+      FEValuesExtractors::Vector extract_initial_topography(0);
+
+      // Cell iterator over the MeshDeformation FE
+      typename DoFHandler<dim>::active_cell_iterator
+      fscell = mesh_deformation_dof_handler.begin_active(),
+      fsendc= mesh_deformation_dof_handler.end();
```
bangerth, line 265:
```suggestion
      fsendc = mesh_deformation_dof_handler.end();
```

source/mesh_deformation/diffusion.cc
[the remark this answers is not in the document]
bangerth: This still seems relevant.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+                    const Tensor<2, dim, double> projection = unit_symmetric_tensor<dim>() -
+                                                              outer_product(fs_fe_face_values.normal_vector(point), fs_fe_face_values.normal_vector(point));
+
+                    // Loop over the shape functions
+                    for (unsigned int i=0; i<dofs_per_cell; ++i)
+                      {
+                        // Make sure we only assemble for the y-component
+                        // TODO this is only correct for box geometries
+                        if (mesh_deformation_dof_handler.get_fe().system_to_component_index(i).first != dim-1)
+                          continue;
```
bangerth, line 326: same here -- add an assertion of some sort

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+
+                        // Assemble the RHS
+                        // RHS = M*H_old
+                        cell_vector(i) += displacement *
+                                          fs_fe_face_values.shape_value (i, point) *
```
bangerth, line 331: Just for stylistic reasons: We try to multiply with phi_i from the left. So write it (completely equivalently) as 
```
  fs_fe_face_values.shape_value (i, point) * displacement * ...
```

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+                                          fs_fe_face_values.JxW(point);
+
+
+                        for (unsigned int j=0; j<dofs_per_cell; ++j)
+                          {
+                            // Make sure we only assemble for the y-component
+                            // TODO this is only correct for box geometries
+                            if (mesh_deformation_dof_handler.get_fe().system_to_component_index(j).first != dim-1)
+                              continue;
```
bangerth, line 340: both here and above, I had to think for 10 seconds. I think this would be easier to read if you had
```
  for (unsigned int j=...)
   if (mesh_deformation_dof_handler.get_fe().system_to_component_index(j).first == dim-1)
    {
      ...
```

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+                            // Assemble the matrix, for backward first order time discretization:
+                            // Matrix := (M+dt*K) = (M+dt*B^T*kappa*B)
+                            cell_matrix(i,j) +=
+                              (
+                                this->get_timestep() * diffusivity *
+                                (projection * fs_fe_face_values.shape_grad(i, point)) *
+                                (projection * fs_fe_face_values.shape_grad(j, point)) +
+                                fs_fe_face_values.shape_value (i, point) * fs_fe_face_values.shape_value (j, point)
```
bangerth, line 348: Since you have `M+dt*...`, move this line to the front of the formula to match the comment

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+      LinearAlgebra::Vector velocity(mesh_locally_owned, mesh_locally_relevant, this->get_mpi_communicator());
+      velocity = solution;
+      velocity -= initial_topography;
+      velocity -= displacements;
+
+      // The velocity
+      if (this->get_timestep() > 0.)
+        velocity /= this->get_timestep();
+      else
+        velocity = 0.;
```
bangerth, line 389: Can this ever happen? I think it would be safer to put an assert here

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+      const double min_conduction_timestep = Utilities::MPI::min (min_local_conduction_timestep, this->get_mpi_communicator());
+
+      double conduction_timestep = min_conduction_timestep;
+      if (this->convert_output_to_years())
+        conduction_timestep /= year_in_seconds;
+
+      AssertThrow (this->get_timestep() <= min_conduction_timestep,
+                   ExcMessage("The numerical timestep is too large for diffusion of the surface. Although the "
+                              "diffusion scheme is stable, note that the error increases linearly with the timestep. "
+                              "The diffusion timestep is: " + std::to_string(conduction_timestep) + ". "));
```
bangerth, line 438: maybe also output the value of `this->get_timestep()`?

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+          prm.declare_entry("Time between diffusion", "1e20",
+                            Patterns::Double(0,std::numeric_limits<double>::max()),
+                            "The time between each application of diffusion. "
+                            "Units: years if the "
+                            "'Use years in output instead of seconds' parameter is set; "
+                            "seconds otherwise.");
+          prm.declare_entry("Time steps between diffusion", "1",
+                            Patterns::Integer(0,std::numeric_limits<int>::max()),
+                            "The number of time steps between each application of "
+                            "diffusion.");
```
bangerth, line 513: Can you document what happens if one specifies both `Time between diffusion` and `Time steps between diffusion`? 

Maybe separately: Why do you actually need both? Is this so expensive that you wouldn't want to do it in every time step? And if so, why not just use it at fixed intervals in time (rather than in time steps)?

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+                          {
+                            // Make sure we only assemble for the y-component
+                            // TODO this is only correct for box geometries
+                            if (mesh_deformation_dof_handler.get_fe().system_to_component_index(j).first != dim-1)
+                              continue;
+                            // Assemble the matrix, for backward first order time discretization:
+                            // Matrix := (M+dt*K) = (M+dt*B^T*kappa*B)
+                            cell_matrix(i,j) +=
+                              (
+                                this->get_timestep() * diffusivity *
```
bangerth, line 345: I think this is wrong. If you don't run this function in every time step, then you will end up with a dt that corresponds to one time step, but not to the interval between two solves.

My suggestion would be to run this every time step. That avoids having to think what dt you want to use here, and avoids some difficulty with the run-time parameters

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,591 @@
...
+                                           "height of a point along the top boundary with respect to  the surface of the unperturbed domain. "
+                                           "\n\n"
+                                           "Using this definition, the plugin then solves for one time step, i.e., "
+                                           "using as initial condition $h(t_{n-1})$ the current surface elevation, "
+                                           "and computing $h(t_n)$ from it by solving the equation above over "
+                                           "the time interval $t_n-t_{n-1}$. From this, one can then compute "
+                                           "a surface velocity $v = \\frac{h(t_n)-h(t_{n-1})}{t_n-t_{n-1}}$. "
+                                           "\n\n"
+                                           "This surface velocity is used to deform the surface and as a boundary condition "
+                                           "for solving the Laplace equation to determing the mesh velocity in the "
```
bangerth, line 584:
```suggestion
                                           "for solving the Laplace equation to determining the mesh velocity in the "
```

source/mesh_deformation/interface.cc
```diff
@@ -581,6 +603,67 @@ namespace aspect
 
 
 
+    template <int dim>
+    void MeshDeformationHandler<dim>::set_initial_topography()
```
bangerth, line 607: This function might become obsolete with @gassmoeller recent patch. We'll have to figure out who gets to merge first :-)

source/mesh_deformation/diffusion.cc
anne-glerum: There is an AssertThrow for Box geometries in the `initialize` function, isn't that enough?

source/mesh_deformation/interface.cc
anne-glerum: Yes, that will be great! There are also some places in core.cc that will have to be reverted back then.

source/mesh_deformation/diffusion.cc
anne-glerum: I don't understand why this is wrong. We always take the topography of the previous timestep to diffuse to compute the current topography. We do not store the topography for the next time we diffuse it, so why does it matter that the timestep is not the interval between to diffusion solves?

source/mesh_deformation/diffusion.cc
bangerth: Once someone goes around and fixes the first place, these secondary places still have to be fixed, but one might not notice if the first assertion is removed.

I see assertions as basically a code expression of what I assume to be true for each piece of code. If you make the same assumption in 7 different functions, then there should be 7 assertions that check that the assumption is true. It will make fixing this at a later time so much easier, because you will have already marked up all of the places where someone will have to go check.

source/mesh_deformation/free_surface.cc
anne-glerum: I'm sorry I've left these changes to `free_surface.cc` in. They were for testing whether a different normal would get rid of surface topography asymmetry, and they have nothing to do with this PR.

source/mesh_deformation/diffusion.cc
bangerth: Oh, I see, you think of the "every once in a while" thing not as "using a larger time step for the diffusion than for the flow", but really only as "I'm fixing up the boundary a little bit every once in a while". That corresponds to having no diffusion for the majority of time steps, and only occasionally having a little bit of diffusion -- a rain burst in the desert.

But is that then really what you want? If, for example, you only do the diffusion step once every million years, then the amount of diffusion you get per million years is proportional to the length of the time step you happen to be in during that one time step when you hit the next time threshold. That means in particular that if you refine the mesh, your time step goes down, and the amount of diffusion goes down. 

In different words: You're not going to *converge* with this scheme as you refine the mesh because the amount of diffusion in the model depends on the length of the time step.

include/aspect/mesh_deformation/interface.h
anne-glerum: I don't think so, because this will become obsolete once Rene's initial mesh deformation functionality is merged.

include/aspect/mesh_deformation/interface.h
anne-glerum: Yes, but this will become obsolete once Rene's initial mesh deformation functionality is merged, so then we could remove it all together.

anne-glerum: Changed it to going through the `SimulatorAccess` when we need the topography, so we don't need to store the pointer anymore.

source/mesh_deformation/diffusion.cc
anne-glerum: Thanks for the offline talk. I'll remove the `time_between_diffusion` parameter.

bangerth:
The indentation check fails, and I suspect that you probably also want to squash the 43 commits into a smaller number. Other than that, this is ok with me.

anne-glerum:
Great thanks! I will squash the commits. 2 things:
1. Not indent but building the documentation fails, the pdf could not be built. Is there a quick way to see what's wrong?
2. Rene greatly simplified the boundary conditions for mesh deformation in #3751. I will have a look at that PR to see if I can simplify the logic in this plugin as well. Or should I leave that for a separate PR?

anne-glerum:
@gassmoeller Manual fixed, squashed, this is ready :)

naliboff:
@anne-glerum - It looks like two of the tests still need to be updated. @bangerth @gassmoeller - After the tests are updated, do you want to do a final look through or is it ok to go ahead and merge?

anne-glerum:
I've updated the test results with the changes from #3855.

include/aspect/mesh_deformation/diffusion.h
```diff
@@ -0,0 +1,132 @@
+/*
+  Copyright (C) 2011 - 2018 by the authors of the ASPECT code.
```
gassmoeller, line 2: 2020?

include/aspect/mesh_deformation/diffusion.h
```diff
@@ -0,0 +1,132 @@
...
+                                        const std::set<types::boundary_id> &boundary_ids) const;
+
+        /**
+         * The hillslope transport coefficient or diffusivity [m2/s]
+         * used in the hillslope diffusion of the deformed
+         * surface.
+         */
+        double diffusivity;
+
+
```
gassmoeller, line 110: remove one empty line

include/aspect/mesh_deformation/free_surface.h
gassmoeller: I think this is still not resolved?

include/aspect/mesh_deformation/interface.h
gassmoeller: Does this need to be changed now that #3751 is merged?

include/aspect/mesh_deformation/interface.h
gassmoeller: But the pointer is still in the code? Can it be removed? Also it may be a while before we can completely remove the old initial topography implementation due to backward compatibility issues.

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,587 @@
...
+                // Get the boundary indicator of current cell face
+                const types::boundary_id boundary_indicator
+                  = fscell->face(face_no)->boundary_id();
+
+                // Only consider the requested boundaries
+                if (boundary_ids.find(boundary_indicator) == boundary_ids.end())
+                  continue;
+
+                // Reinitialize update flags for current cell face
+                fs_fe_face_values.reinit (fscell, face_no);
```
gassmoeller, line 423: I do not see where you use `fs_fe_face_values`, can it be removed completetly?

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,587 @@
...
+      AssertThrow (this->get_timestep() <= min_conduction_timestep,
+                   ExcMessage("The numerical timestep is too large for diffusion of the surface. Although the "
+                              "diffusion scheme is stable, note that the error increases linearly with the timestep. "
+                              "The diffusion timestep is: " + std::to_string(conduction_timestep) + ", while "
+                              "the advection timestep is: " + std::to_string (this->get_timestep ())));
+    }
+
+
+    /**
+     * A function that creates constraints for the velocity of certain mesh
```
gassmoeller, line 447: This comment should not be here, it should be in the header (but I think it is already there?)

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,587 @@
...
+    {
+      prm.enter_subsection("Mesh deformation");
+      {
+        prm.enter_subsection("Diffusion");
+        {
+          prm.declare_entry("Hillslope transport coefficient", "1e-6",
+                            Patterns::Double(0),
+                            "The hillslope transport coefficient $\\kappa$ used to "
+                            "diffuse the free surface, either as a  "
+                            "stabilization step or a to mimic erosional "
```
gassmoeller, line 506:
```suggestion
                            "stabilization step or to mimic erosional "
```

source/mesh_deformation/diffusion.cc
gassmoeller: I think this question is not yet resolved? If you only want to diffuse every n timesteps, because you do not want to smooth too much, could you not instead reduce the diffusion coefficient?

source/mesh_deformation/diffusion.cc
```diff
@@ -0,0 +1,587 @@
...
+                              {
+                                // Make sure we only assemble for the y-component
+                                // TODO this is only correct for box geometries
+                                if (mesh_deformation_dof_handler.get_fe().system_to_component_index(j).first == dim-1)
+                                  {
+                                    // Assemble the matrix, for backward first order time discretization:
+                                    // Matrix := (M+dt*K) = (M+dt*B^T*kappa*B)
+                                    cell_matrix(i,j) +=
+                                      (
+                                        fs_fe_face_values.shape_value (i, point) * fs_fe_face_values.shape_value (j, point) +
```
gassmoeller, line 346: Are you concerned about the speed of this function, i.e. did you see this assembly taking a lot of time? If not, do not bother, but if yes, there is a lot of optimization potential in this assembly. For example you can precompute `phi[i] = fs_fe_face_values.shape_value (i, point)` for all i at the beginning of the loop over `point`, and then save a lot of calls to `shape_value` in line 333 and here (for i and j). Even more helpful will be precomputing `proj_grad_phi[i] = projection * fs_fe_face_values.shape_grad(i, point)` for all i, because it reduces the calls in the inner loop to the gradient calculation and the tensor product of projection and the gradient.

gassmoeller (commented):
This looks very close to done. I have a few small comments, mostly minor comments and if there is anything that needs updating now that #3751 is merged. Unfortunately I caused a conflict when merging that so you need to rebase. Also please check if you addressed all of @bangerth's comments, I found a few that looked like they were not addressed (rebasing sometimes marks comments as outdated or resolved and then you need to click on them to see what they were about).

gassmoeller:
Some of my comments were replies to conversations that happened in the 'Files changed' panel, but they do not show up here in the 'conversations' panel. When going through them, it is probably easiest to go through the 'Files changed' panel and check for the comments I was referring to, sorry for the extra work.

source/mesh_deformation/diffusion.cc
anne-glerum: I'm not sure if I did entirely what you intended, but I precompute `phi[i]` and `projected_grad_phi[i]` now.

include/aspect/mesh_deformation/free_surface.h
anne-glerum: Ah, I had removed all other instances of `geometric_normal`, except for this one. It's gone now. I have not added documentation, as I don't think that should be part of this PR, which is only about the diffusion plugin.

include/aspect/mesh_deformation/interface.h
anne-glerum: I have added a few sentences to explain this.

include/aspect/mesh_deformation/interface.h
anne-glerum: I don't think it can be changed yet. As long as initial topography can still be prescribed through geometry_model/initial_topography, we need these functions.

include/aspect/mesh_deformation/interface.h
```diff
@@ -314,6 +321,12 @@ namespace aspect
          */
         void compute_mesh_displacements ();
 
+        /**
+         * Set up the Vector with initial displacments of the mesh
+         * due to the initial topography.
+         */
+        void set_initial_topography ();
```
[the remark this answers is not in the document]
anne-glerum: Yep, added to the first sentence.

include/aspect/mesh_deformation/interface.h
anne-glerum: Yes, another left-over, unused parameter, sorry.

source/mesh_deformation/diffusion.cc
anne-glerum: yes

source/mesh_deformation/diffusion.cc
anne-glerum: yes, removed the comment.

source/mesh_deformation/diffusion.cc
anne-glerum: Wolfgang's question is resolved after we talked about it in person: I don't need both and removed `Time between diffusion`.

anne-glerum:
Hi @gassmoeller, I went through the resolved comments and think I have addressed everything. My answers also only show up in the Files changed tab. It's a shame, but as long as the initial topography can still be applied the old way, the additions to core have to stay. Once everything is done through initial_mesh_deformation, these and several functions in interface.cc can be removed.

gassmoeller (approved):
This looks great. Let's merge it as is and if we find minor things to improve later we can always do it (I will open a follow up PR with minor improvements, please take a look). This is a really cool new feature!

PR #3409, opened 2020-03-10
Jenkins: ignore coveralls failure

tjhei:
The coveralls step is not useful to run if tests failed, so only run on
success. This fixes the problem that clicking on details of a failed run
no longer shows thousands of output lines from coveralls, that are
difficult to hide to see the actual error.


tjhei:
See https://jenkins.geodynamics.org/blue/organizations/jenkins/aspect/detail/PR-3406/2/pipeline/ for an example. The coveralls step fails with "exit code 8" because of the previous failure in the earlier step.

PR #3410, opened 2020-03-10
S40RTS/SAVANI: avoid data copies and cleanup

tjhei:
- avoid copying std::vectors containing the coefficient data at each
evaluation point
- minor cleanup

### DIFF
--- a/source/initial_temperature/S40RTS_perturbation.cc
+++ b/source/initial_temperature/S40RTS_perturbation.cc
@@ -194,17 +194,17 @@ namespace aspect
       const unsigned int num_spline_knots = 21;
 
       // get the spherical harmonics coefficients
-      const std::vector<double> a_lm = spherical_harmonics_lookup->cos_coeffs();
-      const std::vector<double> b_lm = spherical_harmonics_lookup->sin_coeffs();
+      const std::vector<double> &a_lm = spherical_harmonics_lookup->cos_coeffs();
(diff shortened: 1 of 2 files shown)


source/initial_temperature/S40RTS_perturbation.cc
```diff
@@ -194,17 +194,17 @@ namespace aspect
       const unsigned int num_spline_knots = 21;
 
       // get the spherical harmonics coefficients
-      const std::vector<double> a_lm = spherical_harmonics_lookup->cos_coeffs();
-      const std::vector<double> b_lm = spherical_harmonics_lookup->sin_coeffs();
+      const std::vector<double> a_lm & = spherical_harmonics_lookup->cos_coeffs();
```
gassmoeller: Nice find. I am more used to `int &a = ...` than `int a & = ...`, is there any difference in meaning? The tester doesnt seem to like it either.

tjhei: Oops. That is me not compiling things. :-)

tjhei:
> Is that something that only works with new compiler versions (or clang vs gcc)?

no, I am just storing the returned reference instead of copying the vector.

gassmoeller:
> no, I am just storing the returned reference instead of copying the vector.
I was wondering if `&a` vs `a&` is a compiler version thing, but you clarified that above. Ready to go.

PR #3411, opened 2020-03-10
Improve preconditioner coupling and assembly

gassmoeller:
This is a follow up to #3404, which makes similar improvements to the coupling of the preconditioner matrix. It turns out we allocate the A block of the preconditioner matrix even if we use the system matrix for the preconditioner in the end.  Additionally, we also assemble the preconditioner A block for free surface models and for melt models, even though we use the system matrix block for the preconditioner. This PR should save some memory and some time in the assembly for all melt and free surface models.

### DIFF
--- a/source/simulator/assemblers/stokes.cc
+++ b/source/simulator/assemblers/stokes.cc
@@ -42,8 +42,6 @@ namespace aspect
       const unsigned int stokes_dofs_per_cell = data.local_dof_indices.size();
       const unsigned int n_q_points           = scratch.finite_element_values.n_quadrature_points;
       const double pressure_scaling = this->get_pressure_scaling();
-      const bool assemble_A_approximation = this->get_parameters().use_full_A_block_preconditioner == false &&
-                                            this->get_parameters().mesh_deformation_enabled == false;
 
       // First loop over all dofs and find those that are in the Stokes system
       // save the component (pressure and dim velocities) each belongs to.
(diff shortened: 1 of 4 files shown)


tjhei:
can you rebase to master please? This is hard to read.

gassmoeller:
I rebased to master. The change in `setup_system_matrix()` is hard to read, because git is confused. I really only added one outer if-condition around the whole code, everything inside remains the same.

source/simulator/assemblers/stokes.cc
```diff
@@ -196,6 +197,12 @@ namespace aspect
       Assert (this->get_parameters().use_equal_order_interpolation_for_stokes == false,
               ExcNotImplemented());
 
+      const bool assemble_A_approximation = this->get_parameters().use_full_A_block_preconditioner == false &&
+                                            this->get_parameters().mesh_deformation_enabled == false;
+
+      if (assemble_A_approximation == false)
```
tjhei: I would think this should this be an assert instead (and this assembler should never be called in the first place). Do you agree?

source/simulator/core.cc
```diff
@@ -351,6 +351,10 @@ namespace aspect
     // Initialize the mesh deformation handler
     if (parameters.mesh_deformation_enabled)
       {
+        // Models with deformed boundaries require
+        // the full A block to converge consistently
+        parameters.use_full_A_block_preconditioner = true;
```
tjhei: great idea! :+1:

tjhei:
excellent, thank you!

gassmoeller:
Thanks, I thought that having a unified parameter (which was already around) was more convenient that checking each individual reason why we might need A or not.

PR #3412, opened 2020-03-11
use switch for solver-type check

tjhei:
Introduce a switch for the newly introduced helper functions. This makes
the code longer, but will be easier to maintain as it will trigger an
exception when a new solver type is implemented.


tjhei:
@gassmoeller do you agree?

PR #3413, opened 2020-03-11
Reuse system matrix for melt solver

gassmoeller:
This PR removes the reconstruction of a matrix block whenever the fluid velocity is solved, and instead uses the existing system matrix (as for all other solution variables we have). This means the memory is allocated permanently than just while solving for the fluid velocity, but in some sense that is more consistent with the other blocks (and also does not change the maximum amount of memory allocated during the whole run).

### DIFF
--- a/source/simulator/melt.cc
+++ b/source/simulator/melt.cc
@@ -1053,7 +1053,9 @@ namespace aspect
   template <int dim>
   void
   MeltHandler<dim>::
-  compute_melt_variables(LinearAlgebra::BlockVector &solution)
+  compute_melt_variables(LinearAlgebra::BlockSparseMatrix &system_matrix,
+                         LinearAlgebra::BlockVector &solution,
+                         LinearAlgebra::BlockVector &system_rhs)
   {
     if (!this->include_melt_transport())
       return;
@@ -1061,8 +1063,8 @@ namespace aspect
     Assert (this->include_melt_transport(), ExcMessage ("'Include melt transport' has to be on to "
                                                         "compute melt variables"));
 
-    LinearAlgebra::BlockVector distributed_vector (this->introspection().index_sets.system_partitioning,
-                                                   this->get_mpi_communicator());
+    LinearAlgebra::BlockVector distributed_solution (this->introspection().index_sets.system_partitioning,
+                                                     this->get_mpi_communicator());
 
     const unsigned int por_idx = this->introspection().compositional_index_for_name("porosity");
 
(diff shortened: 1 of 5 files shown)


source/simulator/melt.cc
```diff
@@ -1071,53 +1073,6 @@ namespace aspect
       // u_f =  u_s - K_D (nabla p_f - rho_f g) / phi  or = 0
       // by solving a mass matrix problem
 
-      // TODO: lots of cleanup/optimization opportunities here:
```
tjhei: I am missing the code to zero out the matrix and RHS or is this hidden somewhere?

source/simulator/solver.cc
```diff
@@ -936,7 +936,7 @@ namespace aspect
 
     // convert melt pressures:
     if (parameters.include_melt_transport)
-      melt_handler->compute_melt_variables(solution);
+      melt_handler->compute_melt_variables(system_matrix,solution,system_rhs);
```
tjhei: When you do this, we need to make sure that the system matrix will be reasssembled in the next step, right? Is this guaranteed to work? If this is already the case, we should leave a comment here.

source/simulator/melt.cc
gassmoeller: No I think it was missing. I added it now. I think it did not make a difference, because whenever we rebuild the Stokes matrix we zero out the whole system matrix in `assembly.cc:assemble_Stokes_system()`. Not sure that is how it should be (the advection system only zeroes out its own block).

tjhei: I think each assemble routine should zero out exactly the blocks we are writing to.

source/simulator/solver.cc
gassmoeller: I do not quite understand your question. Why do we need to reassemble the matrix in the next step? We only use the fluid velocity blocks, all other parts of the matrix remain untouched.

source/simulator/melt.cc
gassmoeller: I can open another PR for the Stokes assembly.

source/simulator/solver.cc
tjhei: Oh, I thought we reuse the normal velocity block and we overwrite its content. Nevermind.

gassmoeller:
I think I fixed all comments.

> Oh, I thought we reuse the normal velocity block and we overwrite its content. Nevermind.

That is an interesting idea to save memory, but I would leave it for future optimization.

PR #3416, opened 2020-03-16
Output degrees instead of radians

anne-glerum:
The PlasticAdditionalOutputs output the internal angle of friction in radians. This PR outputs them in degrees, which is more intuitive. I don't think it will break any tests, because I didn't see any that combines a nonzero friction angle with named additional output and compares the gnuplot output. 

### For all pull requests:

### DIFF
--- a/doc/modules/changes/20200319_glerum
+++ b/doc/modules/changes/20200319_glerum
@@ -1,0 +1,5 @@
+Changed: The named additional outputs of the visco plastic material model
+now output the current internal angle of friction in degrees instead
+of radians.
+<br>
+(Anne Glerum, 2020/03/19)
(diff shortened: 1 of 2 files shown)


gassmoeller:
Hmm, the change looks good, but I have a hard time deciding what is 'more intuitive'. @naliboff do you have an opinion? If we change it we should at least create a changelog entry.

naliboff:
I agree outputting in degrees is more intuitive than radians, given that the values are entered in degrees within the parameter file. Thanks for doing this change @anne-glerum!

After adding a changelog entry, this should be good to go :+1:

gassmoeller:
Thanks!

tjhei:
Can I talk @naliboff or @anne-glerum into creating a small test for this output?

naliboff:
@tjhei - sure thing, it is on my list for today :+1:

naliboff:
The aforementioned test is added in #3437.

PR #3420, opened 2020-03-18
Add average- minimun- and maximum gravity/potential in Gravity Postprocessor

LudovicJnnt:
This PR relates to second point of PR #3401 
It adds the average- minimum- and maximum gravity acceleration and potential to the statistics file.
I also added the Stokes velocity polynomial degree and quadrature degree increase; those are needed for my current work when calculating the cost of computation (Number of elements * Number of Gauss-Legendre Quadrature). It is better to have them in the statistic file because I perform many run in a loop and combine all statistic files into one. 

* [x] I have [created a testcase]


source/postprocess/gravity_point_values.cc
```diff
@@ -414,6 +432,47 @@ namespace aspect
             }
         }
 
+      // write quantities in the statistic file
+      const std::string name0("Stokes velocity polynomial degree");
```
gassmoeller: I would prefer this not to be written into the statistics file, I will expand on this in the main reply.

source/postprocess/gravity_point_values.cc
```diff
@@ -414,6 +432,47 @@ namespace aspect
...
+      statistics.add_value (name0, degree-quadrature_degree_increase);
+      statistics.set_precision (name0, 0);
+      statistics.set_scientific (name0, false);
+
+      const std::string name1("Quadrature degree increase for gravity calculation");
```
gassmoeller: Same here, this is not a model result.

source/postprocess/gravity_point_values.cc
```diff
@@ -414,6 +432,47 @@ namespace aspect
...
+      statistics.add_value (name1, quadrature_degree_increase);
+      statistics.set_precision (name1, 0);
+      statistics.set_scientific (name1, false);
+
+      const std::string name2("Average gravity (m/s^2)");
```
gassmoeller:
```suggestion
      const std::string name2("Average gravity acceleration (m/s^2)");
```

source/postprocess/gravity_point_values.cc
```diff
@@ -414,6 +432,47 @@ namespace aspect
...
+      statistics.add_value (name2, sum_g/n_satellites);
+      statistics.set_precision (name2, 18);
+      statistics.set_scientific (name2, true);
+
+      const std::string name3("Minimum gravity (m/s^2)");
```
gassmoeller:
```suggestion
      const std::string name3("Minimum gravity acceleration (m/s^2)");
```

source/postprocess/gravity_point_values.cc
```diff
@@ -414,6 +432,47 @@ namespace aspect
...
+      statistics.add_value (name3, min_g);
+      statistics.set_precision (name3, 18);
+      statistics.set_scientific (name3, true);
+
+      const std::string name4("Maximum gravity (m/s^2)");
```
gassmoeller:
```suggestion
      const std::string name4("Maximum gravity acceleration (m/s^2)");
```

source/postprocess/gravity_point_values.cc
```diff
@@ -414,6 +432,47 @@ namespace aspect
...
+      statistics.add_value (name4, max_g);
+      statistics.set_precision (name4, 18);
+      statistics.set_scientific (name4, true);
+
+      const std::string name5("Average potential (m^2/s^2)");
```
gassmoeller:
```suggestion
      const std::string name5("Average gravity potential (m^2/s^2)");
```

and the same for the two properties below.

source/postprocess/gravity_point_values.cc
```diff
@@ -414,6 +432,47 @@ namespace aspect
...
+      statistics.set_scientific (name0, false);
+
+      const std::string name1("Quadrature degree increase for gravity calculation");
+      statistics.add_value (name1, quadrature_degree_increase);
+      statistics.set_precision (name1, 0);
+      statistics.set_scientific (name1, false);
+
+      const std::string name2("Average gravity (m/s^2)");
+      statistics.add_value (name2, sum_g/n_satellites);
+      statistics.set_precision (name2, 18);
```
gassmoeller: Do you really need 18 digits of precision? First of all `double` will only accurately represent 15 digits anyway, and that is only in case there is no loss of precision due to calculations and no model (iterative solvers, limited resolution) error at all. Our usual accuracy is 8 digits. Some quantities require higher precision and they usually use 12 digits. I realize gravity variations can be very small compared to the full gravity, but have you tested that you need more than 12 digits?

LudovicJnnt: I chose 18 digits based on early tests I made for the graph below. I remember I picked 8, 12, 14 then jumped to 18 digits, unaware that the double was limited to 15. It was useful for a benchmark, but I guess for users it is pointless (even perhaps for benchmarks) to have that high accuracy/precision since there is a clear convergence of the solution way below the milligal threshold. 
image

LudovicJnnt: In the output of the gravity postprocessor, should I also change the precision to 12 digits?

LudovicJnnt:
@gassmoeller Thanks for the comments. Also thanks for the tip to write a shell script for the benchmark. I thus removed the Stokes degree and quadrature degree increase from the statistics file.

For the 18 digits comment, see my reply above. For now, I replaced all precision digits > 12 to 12 digits.

doc/modules/changes/20200318_jeanniot
```diff
@@ -0,0 +1,5 @@
+Added: The Stokes velocity polynomial degree, quadrature degree increase,
```
gassmoeller: Please update the changelog file.

LudovicJnnt:
@gassmoeller I updated the changelog file and set precision in the output file back to 18 (for next PR) as suggested. Cheers

gassmoeller:
Thanks! I reworded the changelog slightly, but once the tester is done this is ready to merge.

PR #3421, opened 2020-03-18
Fix capitalization of Aspect -> ASPECT

gassmoeller:
I think we usually use all caps for ASPECT whenever we use it. This is currently not fully consistent in our code base (and website, e.g. the changelog at https://aspect.geodynamics.org/doc/doxygen/changes_current.html). This PR fixes most occurences of 'Aspect'. @tjhei: I did not modify the CMakeLists.txt for now, however in there it is pretty inconsistent as well. I remember that you and Wolfgang changed the capitalization in there at a hackathon long ago. Do you remember if there was a reason why it was not consistently capitalized everywhere? Should I have a look to see what can be changed without breaking backward compatibility?

### DIFF
--- a/doc/update_all_files.sh
+++ b/doc/update_all_files.sh
@@ -9,8 +9,8 @@ DOC_DIR=`dirname $0`
 BASE_DIR=`pwd $DOC_DIR/..`
 
 # Update source files
-SOURCE_FILES=`find $BASE_DIR -type f \( -name *.cc -or -name *.h \) -and -not -name *.bak | grep -v doc`
-bash ${DOC_DIR}/update_source_files.sh $SOURCE_FILES
+#SOURCE_FILES=`find $BASE_DIR -type f \( -name *.cc -or -name *.h \) -and -not -name *.bak | grep -v doc`
+#bash ${DOC_DIR}/update_source_files.sh $SOURCE_FILES
 
 # Update prm files
 PRM_FILES=`find $BASE_DIR -type f -name *.prm* -not -name *update_script* -not -name *.bak`
@@ -19,4 +19,4 @@ bash ${DOC_DIR}/update_prm_files.sh $PRM_FILES
 # To remove the backup files that are created you will likely want to use the
 # following command. It is commented out by default, because you should think
 # carefully before automatically removing files.
-# find ${BASE_DIR} -name *.bak -delete
+find ${BASE_DIR} -name *.bak -delete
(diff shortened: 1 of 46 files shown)


gassmoeller:
Some simple scripting (`sed -i 's/Aspect/ASPECT/g' *.h` or similar) and some manual tuning, because some cases are not easy to script. I thought about adding an update script, but then it is not really changing any functionality, and I would prefer to not accidentally break things.

tjhei:
> I think we usually use all caps for ASPECT whenever we use it

excellent, I like all caps for the name.

> I did not modify the CMakeLists.txt for now, however in there it is pretty inconsistent as well. I remember that you and Wolfgang changed the capitalization in there at a hackathon long ago. Do you remember if there was a reason why it was not consistently capitalized everywhere?

It is a bit complicated and I am not sure I remember 100%. The default is that module names are not ALL_CAPS (see https://cmake.org/cmake/help/git-master/manual/cmake-developer.7.html#a-sample-find-module for example) and the files are named ``AspectConfig.cmake``. The package name argument of FIND_PACKAGE() is (was?) also case sensitive. We then tried to be consistent with ``-D Aspect_DIR`` and the calls to ``FIND_PACKAGE(Aspect ...)`` etc..
Today I would probably make all of this ASPECT, but I don't think this can be changed in a backwards compatible way.

Also see:
https://cmake.org/cmake/help/v3.3/command/find_package.html
https://gitlab.kitware.com/cmake/cmake/issues/17258

doc/update_all_files.sh
```diff
@@ -9,8 +9,8 @@ DOC_DIR=`dirname $0`
 BASE_DIR=`pwd $DOC_DIR/..`
 
 # Update source files
-SOURCE_FILES=`find $BASE_DIR -type f \( -name *.cc -or -name *.h \) -and -not -name *.bak | grep -v doc`
-bash ${DOC_DIR}/update_source_files.sh $SOURCE_FILES
+#SOURCE_FILES=`find $BASE_DIR -type f \( -name *.cc -or -name *.h \) -and -not -name *.bak | grep -v doc`
```
tjhei: what is the reason for this change?

gassmoeller: That was a leftover of my updates, I fixed it in a second commit.

gassmoeller:
> Today I would probably make all of this ASPECT, but I don't think this can be changed in a backwards compatible way.

Ok, then lets keep it for now. I updated 2 occurences in the CMakeLists.txt comments that should have been all-caps, but skipped all file names and internal cmake variables.

I made all the changes I wanted to make, so feel free to merge if the tester is ok.

PR #3422, opened 2020-03-18
Move convection-box manual files into cookbooks directory

gassmoeller:
During the last user meetings we had discussed how to move the documentation of cookbooks closer to the corresponding parameter files, also in preparation for a possible switch to markdown documentation. This PR is a suggestion for how such a move could look like with one of the cookbooks (convection-box). I created a new subdirectory in `cookbooks/convection-box/` that now contains all files associated with that cookbook (parameter file, latex documentation, all figures and parameter file parts used in the documentation). I tried to keep changes in file structure to a minimum, which works nicely if we replace the currect directories with cookbook specific documentations (doc/manual/cookbooks/...) with symbolic links to the actual directories in (cookbooks/...). One benefit of this move is immediately visible in this PR, we had a duplicated file box.prm in the manual directory that was already out of sync with its original in cookbooks/convection-box.prm. Now the same file is used for running the cookbooks and generating the manual, reducing duplication. No functional changes were made (although both the text and .prm file could use some updates).

I am open to comments if this looks reasonable, or what other options we should consider.


doc/manual/manual.tex
```diff
@@ -4680,555 +4680,7 @@ \subsection{How to set up computations}
...
-
-As a final remark, to test our hypothesis that it was indeed the temperature
-polynomial degree that was the limiting factor, we can increase the Stokes
-polynomial degree to 3 while leaving the temperature polynomial degree at 2. A
-quick computation shows that in that case we get a heat flux of 4.747 -- almost 
-the same value as we got initially with the lower order Stokes element. In other
-words, at least for this testcase, it really was the temperature variable that
-limits the accuracy.
-
+\input{cookbooks/convection-box/convection-box.tex}
```
tjhei: shouldn't this be ``cookbooks/convection-box/manual/convection-box.tex``?

gassmoeller:
I agree, I have renamed manual -> doc and put the .tex inside the doc directory.

> how does this work with the .prm.out files? I don't like if they end up in the ./cookbooks/ directory.

I can make that happen (see changes to the Makefile), but it will also apply to all existing non-converted cookbooks (so the .prm.out of every non .part.prm file inside doc/manual/cookbooks/ will also be placed in a subfolder /doc/). This make one search-replace for all cookbooks in the manual necessary to point to the new location of the .prm.out. Are you ok with this solution, or do you see something better (I spent an hour looking through Makefile documentation but did not see anything).

tjhei:
> I can make that happen (see changes to the Makefile), but it will also apply to all existing non-converted cookbooks

Yes, that seems like a sensible solution. Just create a separate directory for the .out files and include those instead (something like ``doc/manual/generated/*``). Would it make sense to do this in a separate PR first? I am not sure I understand the logic in your changes to the Makefile (where is the different folder chosen?).

gassmoeller:
At the moment the generated files would all be in `doc/manual/cookbooks/cookbook_name/doc/` would that be ok with you? It is the simplest to implement and also prevents name clashes that could occur if all the generated files were in a single directory `generated`. I can certainly make a first PR that modifies the current cookbook files to follow the new pattern, this will actually simplify the Makefile logic.
The main new logic piece in the Makefile is that there are two rules to generate .prm.out files. One triggers if the according .prm file is in the same folder (for all .part. files that are already in the `/doc/`subfolder) and another rule that triggers if the corresponding .prm file is in the parent directory (for the main cookbook files, so that the .prm.out file is generated inside the `doc/` directory).

tjhei:
> At the moment the generated files would all be in `doc/manual/cookbooks/cookbook_name/doc/` would that be ok with you?

It is not ideal, but I think we can live with this. :+1:

tjhei:
```
generating 'manual/cookbooks/platelike-boundary/doc/platelike.prm.out' from 'Makefile'...
```
looks wrong.

doc/Makefile
```diff
@@ -34,6 +43,12 @@ all: aspect.tag manual.pdf
 	@perl manual/cookbooks/annotate.pl $< >$@
 	@for i in `seq 1 10`; do sed -i.bak 's/{\([^!]*\)!\([^!]*\)!\([^!]*\)!\([^}]*\)}/{\1!\2!\3\/\4}/' $@; rm $@.bak; done
 
+%.prm.out: $(realpath ../%.prm) Makefile
```
tjhei: I am not convinced that this works as you have two identical rules here.

tjhei:
Let me know if you want me to take a look at the Makefile. Otherwise we go with what you have.

gassmoeller:
I created #3428 as a first step. It takes what we discussed here and applies it to all existing documentation folders (without moving them to cookbooks like I did here). I also documented the Makefile changes more extensively there. Lets first handle #3428, afterwards this will be a super simple change.

gassmoeller:
I updated the PR to the state after #3428.

PR #3426, opened 2020-03-21
Prescribed plastic dilation

tjhei:
This PR supersedes #3119. Written with help from @naliboff and @cedrict.

This implements a new material model additional output to add a "forced plastic dilation" (basically prescribing a compressibility R with div u = R).

### DIFF
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -970,6 +970,46 @@ namespace aspect
         std::vector<double> rhs_melt_pc;
     };
 
+
+
+    /**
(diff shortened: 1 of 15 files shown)


tjhei:
I extended the comment and added the necessary terms (and a test) for Newton.

> 2\. Should an `AssertThrow` be added somewhere to enforce what solvers and material models this is used with? 

I don't see a reason why we need to enforce what material model to use (and any user-written material needs to be allowed). You will get a NaN in the RHS of the Stokes system if the material model does not fill this additional output.

PR #3430, opened 2020-03-25
fix path of convection-box.prm as mentioned in manual

tjhei:


gassmoeller:
Good catch. Ready once the tester is done.

PR #3431, opened 2020-03-25
Fix use of invalid iterator

gassmoeller:
Found while looking into #3407. If we apply tangential mesh constraints to the same boundary that has some other prescribed mesh deformation (e.g. tangential movement) we try to increment an iterator after it was invalidated by erasing its target. This application is very rare.

### DIFF
--- a/source/mesh_deformation/interface.cc
+++ b/source/mesh_deformation/interface.cc
@@ -386,7 +386,7 @@ namespace aspect
            p != x_no_flux_boundary_indicators.end(); ++p)
         if (mesh_deformation_boundary_indicators_set.find(*p) != mesh_deformation_boundary_indicators_set.end())
           {
-            x_no_flux_boundary_indicators.erase(*p);
+            p = x_no_flux_boundary_indicators.erase(p);
           }
 
       sim.signals.pre_compute_no_normal_flux_constraints(sim.triangulation);
--- a/source/mesh_deformation/interface.cc
+++ b/source/mesh_deformation/interface.cc
@@ -381,22 +381,20 @@ namespace aspect
       // The list of tangential boundary indicators for which we need to set
       // no_normal_flux_constraints, which means the tangential boundary indicators
       // minus the mesh deformation boundary indicators.
-      std::set< types::boundary_id > x_no_flux_boundary_indicators = tangential_mesh_boundary_indicators;
-      for (std::set<types::boundary_id>::const_iterator p = x_no_flux_boundary_indicators.begin();
-           p != x_no_flux_boundary_indicators.end();)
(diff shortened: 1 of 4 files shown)


tests/free_surface_tangential_mesh_velocity_2.prm
```diff
@@ -0,0 +1,100 @@
...
+
+subsection Boundary temperature model
+  set Fixed temperature boundary indicators   = bottom, top
+end
+
+subsection Boundary velocity model
+  set Tangential velocity boundary indicators = bottom, left, right
+end
+
+subsection Boundary velocity model
```
MFraters, line 35: why is this subsection here?

gassmoeller:
I simplified the parameter file and removed unnecessary subsections and outdated comments.

source/mesh_deformation/interface.cc
```diff
@@ -386,7 +386,7 @@ namespace aspect
            p != x_no_flux_boundary_indicators.end(); ++p)
         if (mesh_deformation_boundary_indicators_set.find(*p) != mesh_deformation_boundary_indicators_set.end())
           {
-            x_no_flux_boundary_indicators.erase(*p);
+            p = x_no_flux_boundary_indicators.erase(p);
```
tjhei, line 389: That is not correct, because erase() returns the element after p, which you then increment ``++p``. This means you skip one entry. Instead you need to do
```
for (std::set<types::boundary_id>::const_iterator p = x_no_flux_boundary_indicators.begin();
           p != x_no_flux_boundary_indicators.end();)
if (bla)
   p = x_no_flux_boundary_indicators.erase(p);
else
  ++p;

```

bangerth: This is so hard to read (and remember) that I'd suggest using `std::set_difference` instead: https://en.cppreference.com/w/cpp/algorithm/set_difference

It would be nice to use `std::set::erase_if`, but that's only C++20 :-(

bangerth: Specifically, what you need here is
```
  std::set_difference (tangential_mesh_boundary_indicators.begin(),
                                 tangential_mesh_boundary_indicators.end(),
                                 mesh_deformation_boundary_indicators_set.begin(),
                                 mesh_deformation_boundary_indicators_set.end(),
                                 std::inserter (x_no_flux_boundary_indicators, x_no_flux_boundary_indicators.begin()));
```

gassmoeller:
Thanks, this is indeed simpler to read. I updated the code accordingly.

gassmoeller:
Since this was marked ready to merge and the tests pass I am going to merge.

Issue #2767, opened 2019-01-13
Geometric Multigrid for faster linear solvers


tjhei:
issue to track progress for GMG that @tcclevenger is working on:

- [x] simple benchmark problems
- [x] free slip
- [x] spherical geometry and free slip (with deal.II master)
- [x] relax averaging requirement
- [x] compressibility
- [ ] periodic boundary conditions
- [x] free surface
- [ ] BFBT Schur complement

tcclevenger:
An update: 
Currently we have a matrix-free Stokes block/solver for prescribed velocity boundary (all meshes) and free-slip boundary (box only). We use GMG for velocity block. _Must use cell-wise averaging_.

Still needed:
1. Free-slip boundary for sphere/spherical shell. (current: works in 2D serial, should be simple fix for parallel, will require dealii-developer version, 3D should be logical extension)
2. Relax the requirement for cell-wise averaging. Options: store all quadrature points (not ideal) or project with higher order finite element.
3. Compressibility (should be easy for basic case, already implemented in part, but not tested)
4. Free surface
5. BFBT for schur complement
6. Melt transport

maxrudolph:
Would it be  difficult to extend the GMG to work for periodic boundary conditions? I would be interested in using it for a 3D cartesian model with periodic x- and y- boundary conditions.

tjhei:
> Would it be difficult to extend the GMG to work for periodic boundary conditions? I would be interested in using it for a 3D cartesian model with periodic x- and y- boundary conditions.

I was thinking about that when you posted your test case the other day. :-)
It is likely not too difficult to make it work. Conrad said he would try to do it at some time this spring. He will talk about this (and other plans) at the virtual hackathon tomorrow.

tcclevenger:
Yes, I’ll mention this tomorrow. All the required stuff should already be in DEALII so it will hopefully not be too much work to add the functionality to aspect. I’ll keep you updated.

tcclevenger:
@gassmoeller and @tjhei 
Here are some initial results for the viscosity projection:

-  I tested 3 different benchmark problems which have reference solutions and smooth viscosity: solkz, hollow sphere, burstedde. 

- My viscosity strategy: take quadrature defined, active mesh viscosity (given by the benchmark), project this to a `DGQ(k)` vector on the active mesh, transfer this vector to GMG level vectors, evaluate vectors at quadrature points on GMG levels. 

[...]

PR #3433, opened 2020-03-26
Only average viscosity for gmg

gassmoeller:
Is there a reason we are currently averaging all material properties and not just the viscosity for GMG? For models with constant viscosity (or at least cell-wise constant viscosity, like the solcx benchmark) this limits our convergence rate. I have an application that would benefit from only averaging viscosity (long-term statistics at different Rayleigh numbers with constant viscosity each). I have checked that with this PR sol_cx reaches the expected convergence rate for velocity (h^3) with GMG while previously it only reach h^2. I converted some tests to not use averaging, because that would be the more common case (there are some left that still use averaging).


tjhei:
yes, @tcclevenger has been playing with exactly this before. I think it does make sense.
An argument against this is that it is somewhat inconsistent to average only some of the properties. Not sure how people feel about that. I don't know if there are situations where this will produce worse results, for example.
The way I wanted @tcclevenger to implement this is an new averaging scheme "only viscosity harmonic" instead, but your approach is interesting as well.
@tcclevenger can you remind me what the status of your experiments is and take a look at what @gassmoeller is proposing?

tcclevenger:
Yes, this is something I've used a few times in my own code and I have found the same things as @gassmoeller (annulus and solcx benchmarks recover the correct h^3 rates). 

Similar to this pull request, if `averaging==None`, then I would always force viscosity to be averaged and nothing else. If this is the case, maybe a warning should be thrown onto the screen so that the user whose doesn't know the GMG code won't be confused by a loss in convergence with benchmarks having continuous viscosity? 

I think that I like @tjhei option slightly better for the reason that it would force the user to recognize that viscosity must be averaged. Maybe in the `Assert(averaging!=None)` message we can point the user to the new averaging scheme? But this also could confuse less debug-savy users, so maybe an automatic averaging like proposed in this pull request is better.

I've gotten distracted from the Q1 viscosity projection, but from https://github.com/geodynamics/aspect/issues/2767 I should be able to move forward on that (the 2 issues, storage and under/over-shoots, should not be too complicated). Once that is the case, it could be reasonable to have Q1 viscosity as the default when `averaging==None` since the user should never see a loss in convergence with smooth viscosity. Thoughts?

gassmoeller:
> If this is the case, maybe a warning should be thrown onto the screen so that the user whose doesn't know the GMG code won't be confused by a loss in convergence with benchmarks having continuous viscosity?

I am not sure, we do not often write warnings to the screen. Maybe documenting this behavior is enough? I added a commit with a suggestion for documentation. I feel that if someone is savvy enough to run benchmarks and determine convergence rates they should also check the documentation of the solver they are using (in particular if it is not the default solver). We currently also do not warn about the loss of convergence rates when enabling the material model averaging.

> Once that is the case, it could be reasonable to have Q1 viscosity as the default when averaging==None since the user should never see a loss in convergence with smooth viscosity. 

Yes absolutely. Once that is the case a sentence in the documentation about this behavior is definitely enough.

tcclevenger:
> I am not sure, we do not often write warnings to the screen. Maybe documenting this behavior is enough? I added a commit with a suggestion for documentation. I feel that if someone is savvy enough to run benchmarks and determine convergence rates they should also check the documentation of the solver they are using (in particular if it is not the default solver). We currently also do not warn about the loss of convergence rates when enabling the material model averaging.

Good points. In m opinion then, I think this would be enough and I'll get back to work on Q1 viscosity.

gassmoeller:
@tjhei: Would this be closer to what you thought about? It adds another averaging scheme that only averages for viscosity. I think this is more secure and it allows the same use I had in mind, so I am fine with this version as well.

tjhei:
@tcclevenger FYI.

PR #3435, opened 2020-03-27
clang 10 fixes

tjhei:
fix various issues reported by clang 10.


tjhei:
> I suppose every place where you removed an implementation it was the same as the default anyway

precisely. Test seem to pass, so this is correct.

> places where you added move assignments would have disabled move assignments otherwise

These are new clang warnings. I split the commits and added reasoning to the commit messages.

Issue #3390, opened 2020-02-26
I Think Our Solve for the Coefficients in the Particle Bilinear Interpolation is Stable or Well-Advised


egpuckett:
In the bilinear.cc we solve the so-called  'Normal Equations':  A^TAx=A^Tb (Sorry, r just doesn't make sense to me notationally, since we are trying to find the best approximation to A x = b. :-)

        // Matrix A can be rank deficient if it does not have full rank, therefore singular.
        // To circumvent this issue, we solve A^TAx=A^Tr by using singular value
        // decomposition (SVD).
        A.Tmmult(B, A, false);
        A.Tvmult(c_ATr,r);`

This is well known to be unstable.  It has absolutely nothing to do with whether the matrix A is of full rank or not.  See the attached document:

[DO_NOT_TO_SOLVE_THE_NORMAL_EQUATIONS.pdf](https://github.com/geodynamics/aspect/files/4253036/DO_NOT_TO_SOLVE_THE_NORMAL_EQUATIONS.pdf)

Instead, one can use a (stable version of) QR to solve the least squares problem.  See Section 5.2 of the FIRST EDITION of 

> Matrix Methods in Data Mining and Pattern Recognition by Lars Elden SIAM, Jul 12, 2007  (FIRST EDITION)

There is now a second edition of this book.  Also, I am  pretty sure there is a standard what to use the SVD to solve the Least Squares problem stably, but I have to do a little bit of work to find it.

bangerth:
It's not wrong that we could use a QR factorization. But do we have any evidence that the problem is, indeed, poorly conditioned?

egpuckett:
Well, how does one tell?  If it is unstable then it will lose N digits of accuracy, worst case 16 of them.  That will cause problems with convergence, which we showed happens in our latest GJI paper.

I am always in favor of 'Better safe than sorry."  

Is there a large difference in computational cost using a QR solve versus solving the Normal Equations with some solver (i.e., inverting A^TA) as we are currently doing?  I have known for quite a while that we were doing this but somehow it didn't 'click' until recently, so to speak.

I don't see any harm in changing it on a 'test' branch and seeing if the convergence rates from our GJI paper change for the better.  

Do you have any ideas for checking the accuracy of a QR solve versus what we are currently doing?

Do you have a reason for objecting to implementing what most, if not all, 'data scientists', or however you want to characterize these people, would argue is the 'right way' to solve for the least squares coefficients?

bangerth:
[...]

PR #3224, opened 2019-09-17
Bilinear in 3d

sac-bsa:
Pull Request Checklist. Please read and check each box with an X. Delete any part not applicable. Ask on the [forum](https://community.geodynamics.org/c/aspect) if you need help with any step.

*Describe what you did in this PR and why you did it.*
I have been working with @egpuckett on making the bi-linear least squares particle interpolator work in for three dimensions. @gassmoeller has seen a draft of this code and recommended that we make a PR. 

### For all pull requests:

* [X] I have followed the [instructions for indenting my code](../CONTRIBUTING.md#making-aspect-better).

### For new features/models or changes of existing features:

* [X] I have tested my new feature locally to ensure it is correct.
* [X] I have created a testcase; see #3208
(http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../tests/) directory.

* [ ] I have added a changelog entry in the [doc/modules/changes](../doc/modules/changes) directory that will inform other users of my change.

Do I have to make a changelog entry for this change? And if so, please instruct.

PR #3436, opened 2020-03-28
Qr bilinear interpolation

sac-bsa:
I changed the bilinear least squares particle interpolator so that it doesn't solve the normal equations as referenced in #3390 by @egpuckett but instead uses deal.ii's QR decomposition as suggested. Also removed the comment about bilinear only working in 2D, as it has now been implemented in 3D, since #3224.

### For all pull requests:

### DIFF
--- a/source/particle/interpolator/bilinear_least_squares.cc
+++ b/source/particle/interpolator/bilinear_least_squares.cc
@@ -103,41 +103,41 @@ namespace aspect
           if (selected_properties[property_index])
             b[property_index] = 0;
 
-        unsigned int positions_index = 0;
+        unsigned int particle_index = 0;
         const double cell_diameter = found_cell->diameter();
         for (typename ParticleHandler<dim>::particle_iterator particle = particle_range.begin();
-             particle != particle_range.end(); ++particle, ++positions_index)
+             particle != particle_range.end(); ++particle, ++particle_index)
           {
             const auto &particle_property_value = particle->get_properties();
             for (unsigned int property_index = 0; property_index < n_particle_properties; ++property_index)
               if (selected_properties[property_index])
-                b[property_index][positions_index] = particle_property_value[property_index];
+                b[property_index][particle_index] = particle_property_value[property_index];
             const Tensor<1, dim, double> relative_particle_position = (particle->get_location() - approximated_cell_midpoint) / cell_diameter;
             // A is accessed by A[column][row] here since we need to append
             // columns into the qr matrix.
-            A[0][positions_index] = 1;
-            A[1][positions_index] = relative_particle_position[0];
-            A[2][positions_index] = relative_particle_position[1];
+            A[0][particle_index] = 1;
+            A[1][particle_index] = relative_particle_position[0];
+            A[2][particle_index] = relative_particle_position[1];
             if (dim == 2)
               {
-                A[3][positions_index] = relative_particle_position[0] * relative_particle_position[1];
+                A[3][particle_index] = relative_particle_position[0] * relative_particle_position[1];
               }
             else
               {
-                A[3][positions_index] = relative_particle_position[2];
-                A[4][positions_index] = relative_particle_position[0] * relative_particle_position[1];
-                A[5][positions_index] = relative_particle_position[0] * relative_particle_position[2];
-                A[6][positions_index] = relative_particle_position[1] * relative_particle_position[2];
-                A[7][positions_index] = relative_particle_position[0] * relative_particle_position[1] * relative_particle_position[2];
+                A[3][particle_index] = relative_particle_position[2];
+                A[4][particle_index] = relative_particle_position[0] * relative_particle_position[1];
+                A[5][particle_index] = relative_particle_position[0] * relative_particle_position[2];
+                A[6][particle_index] = relative_particle_position[1] * relative_particle_position[2];
+                A[7][particle_index] = relative_particle_position[0] * relative_particle_position[1] * relative_particle_position[2];
               }
           }
 
         for (unsigned int column_index = 0; column_index < n_matrix_columns; ++column_index)
           qr.append_column(A[column_index]);
-        AssertThrow(qr.size() == n_matrix_columns,
+       // If A is rank deficent, qr.append_column will not append the column.
+       // We check that all columns were added through this assertion
+       AssertThrow(qr.size() == n_matrix_columns,
                     ExcMessage("The matrix A was rank deficent during bilinear least squares interpolation."));
-        // If A is rank deficent, qr.append_column will not append the column.
-        // We check that all columns were added through this assertion
 
         for (unsigned int property_index = 0; property_index < n_particle_properties; ++property_index)
           {
@@ -280,8 +280,9 @@ namespace aspect
       ASPECT_REGISTER_PARTICLE_INTERPOLATOR(BilinearLeastSquares,
                                             "bilinear least squares",
                                             "Interpolates particle properties onto a vector of points using a "
-                                            "bilinear least squares method. Note that deal.II must be configured "
-                                            "with BLAS/LAPACK.")
+                                            "bilinear least squares method. "
+                                            "Note that deal.II must be configured BLAS and LAPACK to "
+                                            "support this operation.")
     }
   }
 }
--- a/source/particle/interpolator/bilinear_least_squares.cc
+++ b/source/particle/interpolator/bilinear_least_squares.cc
@@ -24,7 +24,7 @@
 
 #include <deal.II/grid/grid_tools.h>
 #include <deal.II/base/signaling_nan.h>
-#include <deal.II/lac/lapack_full_matrix.h>
+#include <deal.II/lac/qr.h>
 
 #include <boost/lexical_cast.hpp>
 
@@ -89,14 +89,17 @@ namespace aspect
 
         // Noticed that the size of matrix A is n_particles x matrix_dimension
         // which usually is not a square matrix. Therefore, we find the
-        // least squares solution of Ax=r by solving the "normal" equations
-        // (A^TA) x = A^Tr.
+        // least squares solution of Ax=r by solving reduced QR factorization
+        // Ax = QRc = b -> Q^TQRc = Rc =Q^Tb
+        dealii::ImplicitQR<dealii::Vector<double>> qr;
         const unsigned int matrix_dimension = (dim == 2) ? 4 : 8;
-        dealii::LAPACKFullMatrix<double> A(n_particles, matrix_dimension);
-        std::vector<Vector<double>> r(n_particle_properties, Vector<double>(n_particles));
+        std::vector<dealii::Vector<double>> A(matrix_dimension, dealii::Vector<double>(n_particles));
+        std::vector<Vector<double>> b(n_particle_properties, Vector<double>(n_particles));
+        std::vector<Vector<double>> QTb(n_particle_properties, Vector<double>(matrix_dimension));
+        std::vector<Vector<double>> c(n_particle_properties, Vector<double>(matrix_dimension));
         for (unsigned int property_index = 0; property_index < n_particle_properties; ++property_index)
           if (selected_properties[property_index])
-            r[property_index] = 0;
+            b[property_index] = 0;
 
         unsigned int positions_index = 0;
         const double cell_diameter = found_cell->diameter();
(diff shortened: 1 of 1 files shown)


source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -25,6 +25,7 @@
 #include <deal.II/grid/grid_tools.h>
 #include <deal.II/base/signaling_nan.h>
 #include <deal.II/lac/full_matrix.templates.h>
```
gassmoeller: Do we still need the fullmatrix include?

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -89,14 +90,17 @@ namespace aspect
 
         // Noticed that the size of matrix A is n_particles x matrix_dimension
         // which usually is not a square matrix. Therefore, we find the
-        // least squares solution of Ax=r by solving the "normal" equations
-        // (A^TA) x = A^Tr.
+        // least squares solution of Ax=r by solving reduced QR factorization
```
gassmoeller:
```suggestion
        // least squares solution of Ax=r by solving the reduced QR factorization
```

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -278,8 +279,8 @@ namespace aspect
       ASPECT_REGISTER_PARTICLE_INTERPOLATOR(BilinearLeastSquares,
                                             "bilinear least squares",
                                             "Interpolates particle properties onto a vector of points using a "
-                                            "bilinear least squares method. Currently only 2D models are "
-                                            "supported. Note that deal.II must be configured with BLAS/LAPACK.")
+                                            "bilinear least squares method. Note that deal.II must be configured"
```
gassmoeller: Now that you removed the LAPACKFullMatrix and the SVD call it should be possible to use this without LAPACK do I understand that correctly? If so please remove the note about BLAS/LAPACK here.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -89,14 +90,17 @@ namespace aspect
 
         // Noticed that the size of matrix A is n_particles x matrix_dimension
         // which usually is not a square matrix. Therefore, we find the
-        // least squares solution of Ax=r by solving the "normal" equations
-        // (A^TA) x = A^Tr.
+        // least squares solution of Ax=r by solving reduced QR factorization
+        // Ax = QRc = b -> Q^TQRc = Rc =Q^Tb
```
gassmoeller: In the line above you reference the problem as `Ax=r` while here you use `Ax = b`. Is `b` different from `r` or do you just use a different name to avoid confusion with the `r` matrix? If so, please change the line above into `b` just to be consistent.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -89,14 +90,17 @@ namespace aspect
...
+        dealii::ImplicitQR<dealii::Vector<double>> qr;
         const unsigned int matrix_dimension = (dim == 2) ? 4 : 8;
-        dealii::LAPACKFullMatrix<double> A(n_particles, matrix_dimension);
-        std::vector<Vector<double>> r(n_particle_properties, Vector<double>(n_particles));
+        std::vector<dealii::Vector<double>> A(matrix_dimension, dealii::Vector<double>(n_particles));
```
gassmoeller: Please leave a comment on this line why you store the matrix A as a vector of vectors instead of a matrix (I suppose it is because the ImplicitQR requires you to add it column by column to the QR factorization?).

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -89,14 +90,17 @@ namespace aspect
 
         // Noticed that the size of matrix A is n_particles x matrix_dimension
         // which usually is not a square matrix. Therefore, we find the
-        // least squares solution of Ax=r by solving the "normal" equations
-        // (A^TA) x = A^Tr.
+        // least squares solution of Ax=r by solving reduced QR factorization
+        // Ax = QRc = b -> Q^TQRc = Rc =Q^Tb
+        dealii::ImplicitQR<dealii::Vector<double>> qr;
         const unsigned int matrix_dimension = (dim == 2) ? 4 : 8;
```
gassmoeller: Not your fault, but now that I think about it `matrix_dimension` is not really a helpful variable name for a matrix that is not a square matrix. Is there a better name? What about `n_matrix_columns`?

egpuckett: Yes, we know how many columns A will have.

In one dimension A has two columns;, namely the number of unknowns we are solving for.  Also, the first column is all 1's, but that might not be relevant here.  (However, it is relevant to a proof that A can't be rank deficient in 1 dimension, which of course we don't use, but it may be possible to extend this fact to two and three dimensions to prove that A can't be rank deficient for bilinear LS.  
However, it's going to take me some time to follow up on this. I gave my first online (Zoom) lecture of the Spring Quarter to 90 students yesterday and it was a disaster. :-(  Feel free to work on this problem if you like.)

In two dimensions the number of columns in A is again the number of unknowns, say q.  This integer q will depend on whether we are keeping the xy term in bilinear LS or not. ( I don't think I need to write the bilinear approximation out for @gassmoeller or @sac-bsa.)

Again, same in 3D.  I'm kind of rushed, so I'll think about a good variable name, but perhaps one of you will come up with one first?

Also note that A^TA will be a q x q matrix.  (Why doesn't GitHub have a LaTeX editor?)

source/particle/interpolator/bilinear_least_squares.cc
egpuckett: @gassmoeller Someone removed this comment from the earlier version of bilinear_least_squares.cc :

>         // Matrix A can be rank deficient if it does not have full rank, therefore singular.
        // To circumvent this issue, we solve A^TAx=A^Tr by using singular value
        // decomposition (SVD).

Did you do this?  

@sac-bsa and I really want to put in the SVD algorithm, since it supposedly can handle a rank deficient matrix and would therefore make this part of the algorithm 'bullet proof'.   However we wouldn't solve the 'Normal Equations'  A^TAx=A^Tr , since that is the part that is wildly unstable. We would feed it A x = b (or r) and then the SVD would factor A = U \Sigma V^T, etc.

The current QR implementation should be much better / more accurate, which numerical linear algebra people refer to as more stable, by, in some cases many, many digits.  I can post or email an example to you. Let me know.

However, in the long run, I would like to implement the SVD algorithm for solving LS problems when A is rank deficient.  @sac-bsa and I both understand the theory, but we have to figure out whether that particular call to the SVD would return the minimizer in the L2 norm even when A is rank deficient.  The time factor depends on how much time Mack and my classes take out of our time for research budget.

source/particle/interpolator/bilinear_least_squares.cc
egpuckett: I believe that is correct.  @sac-bsa will know for sure.

gassmoeller:
Ok, so here is a complication. The QR routines were added after deal.II 9.0, which is the oldest version of deal.II we still support. In other words this change would change our minimum requirements for deal.II to 9.1. We have planned to increase the minimum deal.II version anyway, but after our next release (in a few months, before the hackathon in July). So here are our options:
1. Implement preprocessor conditions and depending on the deal.II version use the old or the new algorithm (see source/postprocess/particles.cc:102 for an example, though for this change we would need multiple conditions and it would be a bit more complicated)
2. Backport the content of `include/dealii/lac/qr.h` from deal.II into `include/aspect/compat.h`, we have done that for a few other things already. Each change gets its own `#if !DEAL_II_VERSION_GTE(9,1,0)` section.
3. Wait with merging this PR until we require deal.II 9.1 or newer.

I would leave the decision about this up to @sac-bsa and @egpuckett. Option 3 is the least work, but will delay this PR by a few months. Option 1 is the cleanest if you want to get the change in now (but a bit complicated to get right), option 2 is fine too.

source/particle/interpolator/bilinear_least_squares.cc
sac-bsa: Internally the QR class has a LAPACKFullMatrix in which it stores R. From what I can tell, it uses BLAS as well for its multiplication routines, because of this, I believe that deal.ii would still need to be configured with BLAS/LAPACK.

egpuckett:
@sac-bsa & @gassmoeller  I am really fine with any of the above. Waiting a few months for Option 3 is fine with me if Mack is OK with it.  The less work for everyone the better.

Also note my comment above about our desire to ultimately implement the rank deficient SVD Least Squares algorithm, so I'm not sure a couple of months will interfere with anyone's work, except possibly @naliboff 's. 

I am wondering if the QR solver will change the results in our GJI paper, but of course we don't need to merge our QR with the main branch of ASPECT to look into this.

source/particle/interpolator/bilinear_least_squares.cc
sac-bsa: Yes, that reasoning is correct, I will add a comment here.

source/particle/interpolator/bilinear_least_squares.cc
gassmoeller: > Internally the QR class has a LAPACKFullMatrix in which it stores R. From what I can tell, it uses BLAS as well for its multiplication routines, because of this, I believe that deal.ii would still need to be configured with BLAS/LAPACK.

Ah, ok then leave the comment in. Still since the FullMatrix is not longer used directly in this file you can remove the fullmatrix include at the top of the file.

> Someone removed this comment from the earlier version of bilinear_least_squares.cc. Did you do this?

I dont think so (dont quite remember), but I approved the change because the sentence was worded a bit repetitively. I thought we kept the note that we use SVD because A might be rank deficient, but I do not see it in the current code. If you want to add a note that A might be rank deficient and what that means for the current algorithms I would be fine with it.

gassmoeller:
> Waiting a few months for Option 3 is fine with me if Mack is OK with it. The less work for everyone the better.

Ok, then that would be the easiest option, in particular since you plan on working on the SVD part in the meantime.

sac-bsa:
@gassmoeller I believe that I have fixed all of the suggestions. It appears that the testers both failed due to qr.h only being in deal.II 9.1 as you said, but otherwise locally with deal.II 9.1, the 5 tests involving bilinear least squares pass without being changed.
As to stability, QR should be more stable than calculating it through the normal equations as we are now, however will still fail if A is rank deficient. This makes @egpuckett and I want to look into using SVD instead here. Implementing bilinear least squares using SVD would require something like option 2 or that we wait until dealii/dealii#8739 becomes part of a deal.II release that is required for ASPECT.

gassmoeller:
Thanks

> Implementing bilinear least squares using SVD would require something like option 2 or that we wait until dealii/dealii#8739 becomes part of a deal.II release that is required for ASPECT.

Yes I agree. What about you go ahead with working on the SVD and let me know when you are ready. At that point we can decide how long we would have to wait and make a decision whether to back-port the functionality or wait for the new release.

egpuckett:
@gassmoeller & @bangerth I believe that now deal.ii version 9.2.0 is the standard this is ready to merge.  @sac-bsa is this correct?

source/particle/interpolator/bilinear_least_squares.cc
bangerth: I think it would be easier to read if you used `compositional_field.get_fe().dofs_per_cell` or whatever the correct call is. That way, it's clear that the dimension of the matrix equals the number of unknowns you're trying to determine. That *happens* to be equal to 4 and 8, but makes it *clear* what it is semantically.

bangerth:
Can you rebase this to the current master branch?

bangerth:
You'll need to fix the history of this PR. It currently has 691 commits and changes 280,000 lines :-)

sac-bsa:
@bangerth I also sent the following to your email
I am rather certain that I didn't fix the history of the PR correctly. Here is what my bash_history remembers of what I did so far:
I pulled the master branch, and checked out my qr bilinear interpolator branch and then ran
git rebase -i master
git rebase --edit-todo
git rebase --continue
git stash push doc/manual/burnman doc/manual/convection-box
git status
git commit --amend
git rebase --continue
What would you recommend that I do to fix it? How can I tell that I fixed the history correctly?

Thank you for your help,
Mack Gregory

bangerth:
I think the following should work:
```
git stash                                                     # save everything you may have been working on at the moment
git checkout qr_bilinear_interpolation
git checkout master
git pull upstream master                            # assuming you've set 'upstream' to the https://github.com/geodynamics/aspect repo
git checkout qr_bilinear_interpolation
git rebase master
```
If each of these steps goes through without error, then do
```
git push -f origin qr_bilinear_interpolation
```
If you get an error at some point, stop what you're doing and, if it's in the last step, do `git rebase --abort`. Report back what the error is.

sac-bsa:
@bangerth and @gassmoeller, I believe that this is ready for another look.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -87,66 +87,67 @@ namespace aspect
...
+        std::vector<dealii::Vector<double>> A(n_matrix_columns, dealii::Vector<double>(n_particles));
+        std::vector<Vector<double>> b(n_particle_properties, Vector<double>(n_particles));
+        std::vector<Vector<double>> QTb(n_particle_properties, Vector<double>(n_matrix_columns));
+        std::vector<Vector<double>> c(n_particle_properties, Vector<double>(n_matrix_columns));
         for (unsigned int property_index = 0; property_index < n_particle_properties; ++property_index)
           if (selected_properties[property_index])
-            r[property_index] = 0;
+            b[property_index] = 0;
 
         unsigned int positions_index = 0;
```
gassmoeller, line 106: Similar to the change in #3671 can you rename this `particle_index`

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -87,66 +87,67 @@ namespace aspect
...
-        // Form the matrix B=A^TA and right hand side A^Tr of the normal equation.
-        A.Tmmult(B, A, false);
-        dealii::LAPACKFullMatrix<double> B_inverse(B);
-        B_inverse.compute_inverse_svd(threshold);
+        for (unsigned int column_index = 0; column_index < n_matrix_columns; ++column_index)
+          qr.append_column(A[column_index]);
+        AssertThrow(qr.size() == n_matrix_columns,
+                    ExcMessage("The matrix A was rank deficent during bilinear least squares interpolation."));
+        // If A is rank deficent, qr.append_column will not append the column.
+        // We check that all columns were added through this assertion
```
gassmoeller, line 140: Move this above the line with `AssertThrow`

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -279,9 +280,8 @@ namespace aspect
       ASPECT_REGISTER_PARTICLE_INTERPOLATOR(BilinearLeastSquares,
                                             "bilinear least squares",
                                             "Interpolates particle properties onto a vector of points using a "
-                                            "bilinear least squares method. "
-                                            "Note that deal.II must be configured with BLAS and LAPACK to "
-                                            "support this operation.")
+                                            "bilinear least squares method. Note that deal.II must be configured "
+                                            "with BLAS/LAPACK.")
```
gassmoeller, line 284: I do not see the value in this change, can you remove the change (revert to the state before)?

sac-bsa:
`visco_plastic_yield_plastic_strain_weakening_particles.prm` with many particles (100k).
With QR bilinear interpolation (this PR):
0.131s | 1.8%, 0.155s | 2.2%, 0.166s |  2.3%
With bilinear interpolation as it currently is:
0.154s | 2.1%, 0.114s | 2%, 0.119s | 2.1%
With nearest neighbor:
0.431s | 6.1%, 0.522s | 7.2%, 0.346s | 4.8%
With cell average:
0.0777s | 1.3%, 0.0803s | 1.2%, 0.121s | 1.6%

egpuckett:
@sac-bsa This looks great!  Would you please edit your last comment to include a header.  I assume the first column is run time, but what are the others?

bangerth:
There's a merge conflict related to header files. Can you please rebase to master and fix the conflict? Thanks!

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -280,7 +281,7 @@ namespace aspect
                                             "bilinear least squares",
                                             "Interpolates particle properties onto a vector of points using a "
                                             "bilinear least squares method. "
-                                            "Note that deal.II must be configured with BLAS and LAPACK to "
+                                            "Note that deal.II must be configured BLAS and LAPACK to "
```
gassmoeller, line 284: what happened here? please keep the `with`.

PR #3437, opened 2020-03-29
Add test for visco plastic additional outputs (internal angle of friction)

naliboff:
This is a follow-up to #3416, which changes the output units of the angle of internal friction in visco_plastic from radians to degrees. This PR adds a test to ensure that we check for the units of the internal friction angle. 

It turns out at least two other tests already output the internal friction angle in a solution file via the "plastic additional outputs", but the value was 0 (hence not affected by units change). While having an internal friction angle of 0 makes it easy to quickly determine the yield stress and associated output values, it is a good to have an additional test with a non-zero value.

 

tjhei:
Looks good. Can you update the test results please?

naliboff:
@tjhei - I updated the test result and the CIG tester just finished. However, it looks like your tester may be offline? In any case, the test change was purely formatting related in `screen-output` and not numerical in nature.

tjhei:
Yeah, I have some issues with my machines. Thanks for the update.

PR #3438, opened 2020-03-30
Set the precision of gravity to 12 in gravity_point_values postprocessor

LudovicJnnt:
This PR just changes the precision of gravity written in the output of the gravity_point_values postprocessor to 12 instead of 18. 
The tests were not altered by this change. 

* [x] I have followed the [instructions for indenting my code]

### DIFF
--- a/doc/modules/changes/20200330_jeanniot
+++ b/doc/modules/changes/20200330_jeanniot
@@ -1,0 +1,6 @@
+Added: the precision of gravity acceleration, potential and gradients written in the output
+and statistics files of the "gravity point values" postprocessor may be changed by the user
+via the parameter file under the name "set Precision in gravity output". 
+The default value is 12.
+<br>
+(Ludovic Jeanniot, 2020/03/30)
--- a/source/postprocess/gravity_point_values.cc
+++ b/source/postprocess/gravity_point_values.cc
@@ -409,13 +409,14 @@ namespace aspect
                      << position_satellite[0] << ' '
                      << position_satellite[1] << ' '
                      << position_satellite[2] << ' '
-                     << std::setprecision(18) << g << ' '
-                     << std::setprecision(18) << g.norm() << ' '
-                     << std::setprecision(18) << g_theory << ' '
-                     << std::setprecision(9) << g_potential << ' '
-                     << std::setprecision(9) << g_potential_theory << ' '
-                     << std::setprecision(9) << g_anomaly << ' '
-                     << std::setprecision(9) << g_anomaly.norm() << ' '
+                     << std::setprecision(precision)
+                     << g << ' '
+                     << g.norm() << ' '
+                     << g_theory << ' '
+                     << g_potential << ' '
+                     << g_potential_theory << ' '
+                     << g_anomaly << ' '
+                     << g_anomaly.norm() << ' '
                      << g_gradient[0][0] *1e9 << ' '
                      << g_gradient[1][1] *1e9 << ' '
                      << g_gradient[2][2] *1e9 << ' '
@@ -435,32 +436,32 @@ namespace aspect
       // write quantities in the statistic file
       const std::string name2("Average gravity acceleration (m/s^2)");
       statistics.add_value (name2, sum_g/n_satellites);
-      statistics.set_precision (name2, 12);
+      statistics.set_precision (name2, precision);
       statistics.set_scientific (name2, true);
 
       const std::string name3("Minimum gravity acceleration (m/s^2)");
       statistics.add_value (name3, min_g);
-      statistics.set_precision (name3, 12);
+      statistics.set_precision (name3, precision);
       statistics.set_scientific (name3, true);
 
       const std::string name4("Maximum gravity acceleration (m/s^2)");
       statistics.add_value (name4, max_g);
-      statistics.set_precision (name4, 12);
+      statistics.set_precision (name4, precision);
       statistics.set_scientific (name4, true);
 
       const std::string name5("Average gravity potential (m^2/s^2)");
       statistics.add_value (name5, sum_g_potential/n_satellites);
-      statistics.set_precision (name5, 12);
+      statistics.set_precision (name5, precision);
       statistics.set_scientific (name5, true);
 
       const std::string name6("Minimum gravity potential (m^2/s^2)");
       statistics.add_value (name6, min_g_potential);
-      statistics.set_precision (name6, 12);
+      statistics.set_precision (name6, precision);
       statistics.set_scientific (name6, true);
 
       const std::string name7("Maximum gravity potential (m^2/s^2)");
       statistics.add_value (name7, max_g_potential);
-      statistics.set_precision (name7, 12);
+      statistics.set_precision (name7, precision);
       statistics.set_scientific (name7, true);
 
       // up the next time we need output:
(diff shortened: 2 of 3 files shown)


LudovicJnnt:
Hi John, this PR follows a discussion I had with @gassmoeller. 
First of all, the precision which I had initially set to 18 (just for 3 calculated quantities) goes over the maximum precision of double number which is 15. Technically, 15 could then be used but if someone uses the postprocessor, I am not sure that a precision that high is very necessary. We then decided to use 12 instead.
At the moment, it is for me necessary to have a high precision for gravity, but only for a benchmark purpose. And in that case, I can myself locally change the precision on a separate branch. 

It is just a tiny change, and there are 1 or 2 other PR still in prep.

gassmoeller:
Hi Ludo, thanks for the change. I will let John handle this PR since he commented first ;-).

naliboff:
@LudovicJnnt - Thanks for the explanation! Do you think the benchmarking results that require 15 digits could end up in a paper or elsewhere in the ASPECT repository? If so, would it be worth setting the precision to 15 or allowing the user to define the precision via an input parameter? Seems to cumbersome to have a separate branch for benchmarking if nothing is lost using the higher value as the default.

LudovicJnnt:
@naliboff yes the benchmarking results will definitely end up in a paper. Also, for the last PR about gravity statistics (min, max and average gravity acceleration and potential) in the statistics file, we set the precision to 12 while it will also be used for the benchmarking at precision 15. We basically ran models (hollow sphere with constant density) from GMR 1 to 4 and for GMR 4, we needed a precision >12. We do not need to use GMR >4. 

If we change the precision to 15 for the output, then should the precision in the statistics file also bet set to 15?

naliboff:
@LudovicJnnt - My feeling is that if 15 digits are needed to accurately reproduce benchmarks, then this should be the default value used in the code. 

> If we change the precision to 15 for the output, then should the precision in the statistics file also bet set to 15?
If a precision of 15 in the statistics file is also needed for benchmarking, then I think it is worth changing it is as well.

gassmoeller:
@naliboff: The problem is that to have an accuracy of 15 digits you need to change some other variables in the code too and need to rely on the fact that you have a problem that can be solved up to that accuracy. With the default solvers in ASPECT there will always be a small residual, and if we tune the default output precision of this postprocessor to 15 digits this will suggests an accuracy that is not there for nearly all models. That is why I asked Ludo to reduce the digits to 12 (which we use for some other outputs that require high precision already).

LudovicJnnt:
@gassmoeller @naliboff , just an idea but maybe would it worth to create an input variable in the prm file (in postprocessor gravity subsection) for the precision ? with 12 as the default but possibility to have 15? or less?

gassmoeller:
Yes absolutely, if that is useful for your benchmark.

naliboff:
@gassmoeller - Ah, sorry, I didn't realize changing the precision was connected to another set of variables as well. I also support having an input parameter that allows the user to set the precision for the relevant gravity postprocessors.

source/postprocess/gravity_point_values.cc
```diff
@@ -409,13 +409,13 @@ namespace aspect
...
-                     << std::setprecision(9) << g_anomaly << ' '
-                     << std::setprecision(9) << g_anomaly.norm() << ' '
+                     << std::setprecision(precision) << g << ' '
+                     << std::setprecision(precision) << g.norm() << ' '
+                     << std::setprecision(precision) << g_theory << ' '
+                     << std::setprecision(precision) << g_potential << ' '
+                     << std::setprecision(precision) << g_potential_theory << ' '
+                     << std::setprecision(precision) << g_anomaly << ' '
+                     << std::setprecision(precision) << g_anomaly.norm() << ' '
                      << g_gradient[0][0] *1e9 << ' '
```
LudovicJnnt, line 419: gradients below will have the same precision as stated in line 418. In fact, since the precision will now be the same for every quantities, should I just not delete the std::setprecision at lines 413-418 and just keep one? Or should I just leave it like this?

LudovicJnnt:
@naliboff I have added a variable in the prm file to set the precision in the output and statistics file. Default is 12. 
It does not change the tests output for some reason (probably due to initial setting of the model), but I have tested this precision variable with another model initial condition and it does the job. 

In the review I added a comment/question at line 418.

doc/modules/changes/20200330_jeanniot
```diff
@@ -0,0 +1,5 @@
+Added: the precision of gravity acceleration and potential written in the output and 
+statistics files of the "gravity point values" postprocessor may be changed by the user
+via the parameter file under the name "set Precision in gravity output". Default is 12.
```
naliboff, line 3:
```suggestion
via the parameter file under the name "set Precision in gravity output". The default value is 12.
```

source/postprocess/gravity_point_values.cc
```diff
@@ -545,6 +545,10 @@ namespace aspect
                              Patterns::Double (0.0),
                              "Gravity anomalies may be computed using density "
                              "anomalies relative to a reference density.");
+          prm.declare_entry ("Precision in gravity output", "12",
+                             Patterns::Integer (0.0),
+                             "Set the precision of gravity acceleration in the gravity "
```
naliboff, line 550:
```suggestion
                             "Set the precision of gravity acceleration and potential in the gravity "
```

LudovicJnnt:
@naliboff , I resolved your comments. Could you also address my question at line 418 please?

source/postprocess/gravity_point_values.cc
naliboff: @LudovicJnnt - I'm not sure I fully understand the question. Is the precision for this file now set globally above? If not, is the standard out precision sufficient? Sorry, I'm likely overlooking something from the previous PR.

naliboff:
@LudovicJnnt - Thanks for the changes! Before merging, we just need to resolve the question from line 418 and can you also squash the commits down to a single commit? The commit name could be something like `adjust gravity output precision`, etc.

source/postprocess/gravity_point_values.cc
LudovicJnnt: @naliboff , sorry for the confusion, I hope this is better explained:
when specified `std::precision(N) <<  quantities_1 << quantity_2 << ... ` all quantities (1, 2, ...) following have the same precision N.
This is true until a new specification std::precision(M) is stated:
`std::precision(N) <<  quantities_1 << quantity_2 << ... << std::precision(M) << quantity_3`

In my case, all the gradient quantities at lines 419 to 430 will have the precision stated at line 418. But in the previous lines, I basically reset the precision (for the same precision value) which is therefore completely pointless. I should only set the precision once (at line 413)

Now my question is if I include the same precision for gravity and potential also for gradients, or I set a new precision value just for the gradient fixed to 9 only.

naliboff: @LudovicJnnt - Gotcha, thanks for the explanation. The only reason to reset the precision for every value in lines 419-430 is if you think it will be useful down the line to specify a different precision for each value. If not, I would agree that is best to just set the precision once at line 413.

> Now my question is if I include the same precision for gravity and potential also for gradients, or I set a new precision value just for the gradient fixed to 9 only.

To avoid confusion I would use the same precision for the gradients, but perhaps add a comment that the high precision is really only needed for the gravity and potential values.

tjhei:
A few comments:
- There is no reason to repeatedly set the precision, just set them once before the sequence of outputs.
- There is no reason to set the precision higher than 16 for doubles.
- What is the problem with always outputting 16 digits for the columns where you care about accuracy? Is there a concern about the amount of data?

LudovicJnnt:
@naliboff  I made the changes for all the output quantities with only setting the precision once. 

@tjhei  For your last point, I followed the advice of @gassmoeller. Basically the concern was more for the precision set in the statistics file. For a benchmark or under rare specific cases, the precision needs to be high but in any other common case, it should not be.

gassmoeller:
> What is the problem with always outputting 16 digits for the columns where you care about accuracy? Is there a concern about the amount of data?

It is more that I was worried it would suggest an accuracy that might not be there for most models (unless you increase solver tolerances).

From my perspective this looks ready, are there any remaining questions?

LudovicJnnt:
Thanks @gassmoeller for re-reclarifying :) 
If it is ready, then I think there are no remaining questions.

PR #3439, opened 2020-03-31
Remove viscoelastic plastic material model and test suite

naliboff:
This PR removes the` Viscoelastic Plastic` material model, whose functionality is now included in the `Visco Plastic` material model via #3306.


naliboff:
@gassmoeller - good catch regarding the tests. There are actually a few from viscoelastic_plastic that only test viscoelastic behavior (i.e, no yielding) that I did not add to the visco_plastic suite. I'll readd and modify those tests to use `visco_plastic`.

naliboff:
@gassmoeller - I've renamed and modified all the relevant tests to use the visco plastic material model. The tester is happy and please let me know if you would like any further changes.

PR #3440, opened 2020-03-31
tests: normalize/remove TimerOutput text frame

tjhei:
For tests, we always removed TimerOutput text from the screen-output but the frames

+---------------------------------------------+------------+-----------+

remained. A recent change in the deal.II development version changes the
width and as such the number of dashes in these lines. To avoid this
problem, filter in normalize.pl and delete all of them.


tjhei:
(a few restart tests probably still need adjustment)

PR #3442, opened 2020-04-01
Refactor viscoelastic computations

gassmoeller:
Seen while looking into #3184. This deduplicates some code and moves a bit of code inside `elastic.cc` so that it is available for all material models using elastic rheologies.


gassmoeller:
> Do you think it would be useful to make a material utility or rheology module for calculating the standard viscous strain rate invariant?

Yes, probably best in MaterialModel::Utiltiies?

PR #3445, opened 2020-04-07
Patterns::Double and Integer arguments

tjhei:
- Patterns::Double(min, max): use doubles for min and max
- default value: use a "0." to denote doubles
- Patterns::Integer(min, max): use ints for min and max


tjhei:
fixed.

PR #3447, opened 2020-04-07
Add a spiral sampling method in gravity postprocessor

LudovicJnnt:
I added a sampling method parameter entry in the gravity_point_values postprocessor which allows to select either equiangular or spiral (equidistant) sampling. The reason behind this PR is that when calculating the average global gravity using the equiangular sampling method, there is an overshoot of sampling points at the poles which alter the average gravity.

This required to rename the sampling scheme "uniform distribution" by "domain of points" simply because the initially so-called uniform distribution was not uniform at all, but already equiangular. (I could have created a separate PR for renaming the sampling scheme, but here at least it explicitly better explains why I want to rename it again - let me know if it necessary to separate this change).

The domain of points spiral sampling method is based on the Fibonacci suit and required a number of points where gravity is calculated ("Number points spiral sampling method" in the input file) rather than the number points of longitude and latitude.
The spiral sampling method can only be used for global sampling. 

The figure below shows the points coordinates for 400 points: 
spiral

* [x] I have followed the [instructions for indenting my code]
* [x] I have [created a testcase]


gassmoeller:
Hi Ludo,
sorry for leaving this so long, I was a bit confused by the new options. I have a few question before moving forward with this:
- Is there evidence that the spiral sampling scheme is a uniform distribution on the surface of a sphere (i.e. uniform point density per area)? I googled a bit and think you are implementing the Fibonacci spiral method (e.g. here https://stackoverflow.com/a/26127012), which if I understand correctly creates a regular distribution in the limit of infinitely many points? This is one way to generate regular distributions of points, but there are others. One of the problems with our implementation is that it is not clear what "uniform" should mean. Maybe we should call this method "Fibonacci spiral"/`fibonacci_spiral` or something else that uniquely says what it does?
- The number of options to generate points keeps increasing, and the internal structure is getting too complex for my taste. We now have a structure `sampling_scheme` with 4 options (2 of which are never used), but if one of the options is selected there is another structure `sampling_method` that has two more suboptions. I would like to keep it at one list of options `sampling_scheme` and rather give each a unique name as mentioned above. So what about the following list of options:
```
        /**
         * Specify the sampling scheme determining if gravity calculation is performed
         * for a uniform distribution of points or a list of points.
         * for a domain of points or a list of points.
         */
        enum SamplingScheme
        {
          fibonacci_spiral,
          map,
          list_of_points
        } sampling_scheme;

// REMOVE EVERYTHING BELOW

        /**
         * Specify the domain of points sampling method determining how gravity calculation
         * is performed: equiangular or spiral equidistant.
         */
        enum SamplingMethod
        {
          equiangular,
          spiral
        } sampling_method;
```
We can also safely get rid of the `domain of points` and `list` options, because they are not used anymore. They are allowed as input parameters, but if `list` is selected in the input file, the `parse_parameters` function will always map it to the `list_of_points` option.

I will do a review for the code soon (hopefully today)

include/aspect/postprocess/gravity_point_values.h
```diff
@@ -232,16 +237,26 @@ namespace aspect
...
+        /**
+         * Specify the domain of points sampling method determining how gravity calculation
+         * is performed: equiangular or spiral equidistant.
+         */
+        enum SamplingMethod
+        {
+          equiangular,
+          spiral
+        } sampling_method;
+
```
gassmoeller, line 259:
```suggestion
```

include/aspect/postprocess/gravity_point_values.h
```diff
@@ -232,16 +237,26 @@ namespace aspect
...
-         * for a uniform distribution of points or a list of points.
+         * for a domain of points or a list of points.
          */
         enum SamplingScheme
         {
           map,
           list,
-          uniform_distribution,
+          domain_of_points,
           list_of_points
```
gassmoeller, line 247: I think this should be:
```
map,
fibonacci_spiral,
list_of_points

File: source/postprocess/gravity_point_values.cc [hunk 3]
```diff
@@ -249,6 +254,25 @@ namespace aspect
                 }
             }
         }
+      else if ( (sampling_scheme == domain_of_points) && (sampling_method == spiral) )
+        {
+          double golden_ratio = (1. + std::sqrt(5.))/2.;
```
gassmoeller, line 259: `const`

source/postprocess/gravity_point_values.cc
```diff
@@ -249,6 +254,25 @@ namespace aspect
...
+          double golden_angle = 2. * numbers::PI * (1. - 1./golden_ratio);
```
gassmoeller, line 260: `const`

LudovicJnnt:
Hi Rene, and thanks!

Indeed, the method I implemented is called Fibonacci spiral, and it is the same as the two links you posted. My implementation is taken from an add-on of MATLAB (a sampling toolbox), and they refer to this: https://blog.wolfram.com/2011/07/28/how-i-made-wine-glasses-from-sunflowers/

I still read carefully what you posted to try to spot the differences: 
In https://stackoverflow.com/a/26127012, they combined the angle ratio with the golden ratio, and they get the points in cartesian coordinates. I have a strong preference to keep the golden ratio and golden angle separately so that it is clear that this implementation is based on fibonacci. However I did not checked the conversion from cartesian to spherical coordinates.

In https://arxiv.org/pdf/0912.4540.pdf, they use the golden ratio without pre-calculating the golden angle although it is implicitly used the same way (eq. 4). Then the points are slightly differently set using asin and not acos, but if I understood their setting, asin goes -90 < lat < 90 while acos goes 0 < lat < 180. Therefore how it was implemented in the postprocessor seems adequate. 

I resolved all the comments, and rename accordingly. I think the way you suggested is best and simpler. 
I also removed every occurrence of the "uniform distribution" sampling scheme name (which was a replacement for the map sampling scheme) as it is not appropriate anymore. The map sampling scheme thus remain the map sampling scheme. In comments/documentation it is explain that the map sampling scheme is equiangled while the fibonacci spiral sampling scheme produced a uniformly distributed map. 

As usual, let me now where I am missing clarity in the description.

include/aspect/postprocess/gravity_point_values.h
```diff
@@ -157,65 +157,70 @@ namespace aspect
...
+         * Parameter for the map sampling scheme:
          * Gravity may be calculated for a sets of points along the radius (e.g. depth
          * profile) between a minimum and maximum radius. Number of points along the radius
          * is specified with n_points_radius.
          */
         unsigned int n_points_radius;
 
         /**
-         * Parameter for the uniform distribution sampling scheme:
+         * Parameter for the map and fibonacci spiral sampling scheme:
```
gassmoeller, line 173: where is this used in the fibonacci spiral? I think this should to to the `n_points_radius` variable.

include/aspect/postprocess/gravity_point_values.h
```diff
@@ -232,13 +237,12 @@ namespace aspect
 
         /**
          * Specify the sampling scheme determining if gravity calculation is performed
-         * for a uniform distribution of points or a list of points.
+         * for a equianfgled map, uniformly distributed map or a list of points..
```
gassmoeller, line 240: I would simply remove the options (they are listed in the possible values). That way you do not need to change the description every time a new option is added.

`Specify the sampling scheme determining where gravity calculation is performed.` should be sufficient.

source/postprocess/gravity_point_values.cc
```diff
@@ -479,13 +500,21 @@ namespace aspect
...
+                             "radius, longitude, and latitude. A list of points contains "
+                             "the specific coordinates of the satellites. The fibonacci "
+                             "spiral sampling scheme produces a uniformly distributed map "
+                             "on the surface of sphere defined by a minimum and/or "
+                             "maximum radius.");
+          prm.declare_entry ("Number points fibonacci spiral", "200",
+                             Patterns::Integer (0),
+                             "Parameter for the fibonacci spiral sampling scheme: "
+                             "This specifies the desired number of satellites per number "
+                             "points radius. The default value is 200. Note that sampling "
```
gassmoeller, line 516:
```suggestion
                             "This specifies the desired number of satellites per radius "
                             "layer. The default value is 200. Note that sampling "
```

source/postprocess/gravity_point_values.cc
```diff
@@ -495,50 +524,50 @@ namespace aspect
...
           prm.declare_entry ("Maximum radius", "0.",
                              Patterns::Double (0.0),
-                             "Parameter for the uniform distribution sampling scheme: "
+                             "Parameter for the map sampling scheme: "
                              "Maximum radius can be defined in or outside the model.");
-          prm.declare_entry ("Minimum longitude", "-180.",
-                             Patterns::Double (-180.0, 180.0),
-                             "Parameter for the uniform distribution sampling scheme: "
+          prm.declare_entry ("Minimum longitude", "-180",
+                             Patterns::Double (-180.0,180.0),
```
gassmoeller, line 554: Keep these two lines as they were before (with dots and spaces). Same for the parameters below.

source/postprocess/gravity_point_values.cc
```diff
@@ -603,12 +632,15 @@ namespace aspect
...
-            sampling_scheme = uniform_distribution;
+            sampling_scheme = map;
           else if ( (prm.get ("Sampling scheme") == "list of points") || (prm.get ("Sampling scheme") == "list") )
             sampling_scheme = list_of_points;
+          else if (prm.get ("Sampling scheme") == "fibonacci spiral")
+            sampling_scheme = fibonacci_spiral;
           else
             AssertThrow (false, ExcMessage ("Not a valid sampling scheme."));
-          quadrature_degree_increase = prm.get_integer ("Quadrature degree increase");
+          quadrature_degree_increase = prm.get_double ("Quadrature degree increase");
```
gassmoeller, line 642: Keep this line in its old state, "Quadrature degree increase" is an integer, not a double.

tests/gravity_point_values_spiral.prm
```diff
@@ -0,0 +1,71 @@
+# A simple setup for testing the gravity postprocessor using the 
+# sample scheme 'map'.
```
gassmoeller, line 2: 'fibonacci spiral'?

LudovicJnnt:
Thanks Rene. I think I messed up somewhere in git so which led to some the minor changes (double instead of integer, and the one with the dots and space in the parse_paramater). 
Anyway, it is all done now.

LudovicJnnt:
It is perfect, thanks!

PR #3449, opened 2020-04-09
fix GMG for compressible computations - correct_stokes_rhs()

tcclevenger:
Two fixes are required for `StokesMatrixFree::correct_stokes_rhs()` function. 

1. When the variable `Simulator::do_pressure_rhs_compatibility_modification = true`, `make_pressure_rhs_compatible()` must  be called before adding the rhs correction to `system_rhs`. 
2. For compressible computations, the `2*eta/3*div` term needs to be included in the residual computation for the rhs correction. This was an oversight for this function and did not show up since I only had compressible tests with Zero or Tangential boundary (`correct_stokes_rhs()` is not necessary in that case).

I added two tests, prescribed_dilation_gmg and simple_compresssibility_iterated_Stokes_gmg. Both require fix 1. Fix 2. is required by simple_compresssibility_iterated_Stokes_gmg as it is a compressible computation with non-zero prescribed boundary. I have tested the solution norms for each test vs. AMG solver and they are correct.

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -2206,7 +2206,8 @@ namespace aspect
 
     const bool is_compressible = sim.material_model->is_compressible();
 
-    // GMG diagonals
+    // Assemble and store the diagonal of the GMG level matrices derived from:
+    // 2*eta*(symgrad u, symgrad v) - (if compressible) 2*eta/3*(div u, div v)
     for (unsigned int level=0; level < sim.triangulation.n_global_levels(); ++level)
       {
         mg_matrices_Schur_complement[level].compute_diagonal();
@@ -2276,7 +2277,9 @@ namespace aspect
                       for (unsigned int k=0; k<dofs_per_cell; ++k)
                         {
                           symgrad_phi_u[k] = fe_values[velocities].symmetric_gradient (k, q);
-                          div_phi_u[k] = fe_values[velocities].divergence (k, q);
+
+                          if (is_compressible)
+                            div_phi_u[k] = fe_values[velocities].divergence (k, q);
                         }
 
                       const double JxW = fe_values.JxW(q);
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1405,6 +1405,8 @@ namespace aspect
   template <int dim, int velocity_degree>
   void StokesMatrixFreeHandlerImplementation<dim, velocity_degree>::correct_stokes_rhs()
   {
+    const bool is_compressible = sim.material_model->is_compressible();
+
     dealii::LinearAlgebra::distributed::BlockVector<double> rhs_correction(2);
     dealii::LinearAlgebra::distributed::BlockVector<double> u0(2);
 
@@ -1440,13 +1442,17 @@ namespace aspect
             SymmetricTensor<2,dim,VectorizedArray<double>> sym_grad_u =
                                                           velocity.get_symmetric_gradient (q);
             VectorizedArray<double> pres = pressure.get_value(q);
-            VectorizedArray<double> div = -trace(sym_grad_u);
-            pressure.submit_value   (-1.0*sim.pressure_scaling*div, q);
+            VectorizedArray<double> div = trace(sym_grad_u);
+            pressure.submit_value   (sim.pressure_scaling*div, q);
 
             sym_grad_u *= cell_viscosity_x_2;
 
             for (unsigned int d=0; d<dim; ++d)
               sym_grad_u[d][d] -= sim.pressure_scaling*pres;
+
+            if (is_compressible)
+              for (unsigned int d=0; d<dim; ++d)
+                sym_grad_u[d][d] -= cell_viscosity_x_2/3.0*div;
 
             velocity.submit_symmetric_gradient(-1.0*sym_grad_u, q);
           }
@@ -1460,6 +1466,7 @@ namespace aspect
 
     LinearAlgebra::BlockVector stokes_rhs_correction (sim.introspection.index_sets.stokes_partitioning, sim.mpi_communicator);
     internal::ChangeVectorTypes::copy(stokes_rhs_correction,rhs_correction);
+
     sim.system_rhs.block(0) += stokes_rhs_correction.block(0);
     sim.system_rhs.block(1) += stokes_rhs_correction.block(1);
   }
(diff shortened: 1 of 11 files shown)


source/simulator/stokes_matrix_free.cc
```diff
@@ -2208,6 +2216,10 @@ namespace aspect
             &&
             sim.geometry_model->has_curved_elements())
           {
+            // TODO: add in 2/3*div term, should be simple
```
tcclevenger, line 2219: Currently I'm not including the `2/3*div` term when computing the A-block diagonal for non-box meshes with tangential boundaries. This can be easily fixed as part of this pull request. @gassmoeller or @tjhei, do you know of an example of a simple compressible computation on sphere/shell mesh with tangential boundary conditions I can use to test changes?

source/simulator/stokes_matrix_free.cc
```diff
@@ -1440,14 +1440,18 @@ namespace aspect
...
-            pressure.submit_value   (-1.0*sim.pressure_scaling*div, q);
+            VectorizedArray<double> div = trace(sym_grad_u);
+            pressure.submit_value   (sim.pressure_scaling*div, q);
 
             sym_grad_u *= cell_viscosity_x_2;
 
             for (unsigned int d=0; d<dim; ++d)
               sym_grad_u[d][d] -= sim.pressure_scaling*pres;
 
+            if (sim.material_model->is_compressible())
```
tjhei, line 1451: I wonder if this will result in a virtual function call in the inner-most loop. Can you create a ``bool`` and move this call outside the loops?

tests/prescribed_dilation_gmg/screen-output
```diff
@@ -0,0 +1,50 @@
+[1586368725.360759] [node0567:1223 :0]    ucp_context.c:671  UCX  WARN  network device 'mlx5_0:1' is not available, please use one or more of: 'eth0'(tcp), 'ib0'(tcp), 'mlx4_0:1'(ib)
```
tjhei, line 1: remove this line.

tests/simple_compressibility_iterated_stokes_gmg/screen-output
```diff
@@ -0,0 +1,41 @@
+[1586399160.943582] [node0567:50437:0]    ucp_context.c:671  UCX  WARN  network device 'mlx5_0:1' is not available, please use one or more of: 'eth0'(tcp), 'ib0'(tcp), 'mlx4_0:1'(ib)
```
tjhei, line 1: same here

source/simulator/stokes_matrix_free.cc
tjhei: we can do this in a separate PR, I guess.

tjhei:
awesome, great work in finding the solution!

source/simulator/stokes_matrix_free.cc
gassmoeller: You should be able to use `tests/steinberger_viscosity.prm` that one is an octant of a 3D spherical with some tangential velocity boundary conditions. It does not have an analytical solution though (but I guess you are only interested in the iteration count?).

source/simulator/stokes_matrix_free.cc
```diff
@@ -1460,6 +1466,10 @@ namespace aspect
 
     LinearAlgebra::BlockVector stokes_rhs_correction (sim.introspection.index_sets.stokes_partitioning, sim.mpi_communicator);
     internal::ChangeVectorTypes::copy(stokes_rhs_correction,rhs_correction);
+
+    if (sim.do_pressure_rhs_compatibility_modification)
+      sim.make_pressure_rhs_compatible(stokes_rhs_correction);
```
gassmoeller, line 1471: I do not quite understand this line. Shouldnt you first add the correction to the system_rhs and then make the new system_rhs compatible? Or does the order of operations not matter, because if system_rhs is already compatible, and you add a compatible correction the results is still compatible? I do not exactly remember how the correction looks like.

tjhei: > Or does the order of operations not matter, because if system_rhs is already compatible, and you add a compatible correction the results is still compatible?

I think that is correct. @tcclevenger please explain and put it into a comment above this line of code.

source/simulator/stokes_matrix_free.cc
tcclevenger: Yes, that example is exactly what I need. I've added the fix to the `build_preconditioner()` function and a test (steinberger_compressible_gmg).

source/simulator/stokes_matrix_free.cc
tcclevenger: I've changed this up so that we find the rhs correction and add it to the `system_rhs` in  `StokesMatrixFree`, and then do the compatibility function call in assembly.cc _only after_ calling `correct_stokes_rhs()`. This removes the need to do this 2x.

tcclevenger:
I removed the new test. I forget that curved boundaries with tangential BCs require deal.II-9.1.

tjhei:
Please add the test but rename the .prm to .prm.dealii9.2

tjhei:
can you please squash your commits?

source/simulator/stokes_matrix_free.cc
```diff
@@ -2264,13 +2274,22 @@ namespace aspect
                   for (unsigned int q=0; q<n_q_points; ++q)
                     {
                       for (unsigned int k=0; k<dofs_per_cell; ++k)
-                        symgrad_phi_u[k] = fe_values[velocities].symmetric_gradient (k, q);
+                        {
+                          symgrad_phi_u[k] = fe_values[velocities].symmetric_gradient (k, q);
+                          div_phi_u[k] = fe_values[velocities].divergence (k, q);
```
tjhei, line 2279: you only need this if compressible, right? So maybe put an if here?

source/simulator/stokes_matrix_free.cc
```diff
@@ -2264,13 +2274,22 @@ namespace aspect
...
                       const double JxW = fe_values.JxW(q);
                       for (unsigned int i=0; i<dofs_per_cell; ++i)
                         for (unsigned int j=0; j<dofs_per_cell; ++j)
-                          cell_matrix(i,j) += 2. * viscosity * (symgrad_phi_u[i]*symgrad_phi_u[j])
-                                              * JxW;
+                          {
+                            cell_matrix(i,j) += 2. * viscosity * (symgrad_phi_u[i]*symgrad_phi_u[j])
+                                                * JxW;
+
+                            if (is_compressible)
```
tjhei, line 2289: Can you please put a comment block at the beginning of this function that contains the formula of what you are assembling?

PR #3453, opened 2020-04-13
Use diagonal strengthing in advection solve

tjhei:
Especially with SUPG, the advection/temperature solve can fail to
converge (maybe the matrix is not diagonally dominant without entropy
viscosity). Now, if the solver fails to converge, set up the ILU
preconditioner with diagonal strengthening and try again. Don't do
this by default as it increases iteration counts in normal
situations..


tjhei:
I updated the test results.

gassmoeller:
Hmm, the code makes sense, but can we be sure the new fail-safe preconditioner is actually better? The failing test now ends with a higher residual than before.

gassmoeller:
With the information in #3335 it looks like this is an improvement. :+1:

PR #3457, opened 2020-04-16
Allow requests for material properties

gassmoeller:
Follow up to #3456, which allows to tell a material model which properties to compute, and for a material model to find out which properties are requested. This is implemented so that no current functionality should break. In later PRs I would suggest the following:

- [ ] make all calling sides set the request_properties variable correctly when creating their MaterialModelInputs object (modifying / creating new constructors for MaterialModelInputs)
- [ ] always input strain rate, because we now have material models that use them for reaction terms as well as viscosity (and who knows what else) and the calling side has no way of knowing what the material model might need.
- [ ] let the expensive material models respect the requested properties (in particular grain_size and diffusion_dislocation, which contain internal iterations for strain rate, but also the melt models)
- [ ] remove the two TODOs in this PR
- [ ] deprecate the MaterialModelInputs constructors that take the `compute_strainrate` input argument

### DIFF
--- a/benchmarks/shear_bands/shear_bands.cc
+++ b/benchmarks/shear_bands/shear_bands.cc
@@ -124,7 +124,7 @@ namespace aspect
             {
               double porosity = std::max(in.composition[i][porosity_idx],1e-4);
               out.viscosities[i] = eta_0 * std::exp(alpha*(porosity - background_porosity));
-              if (in.strain_rate.size())
+              if (in.requests_property(MaterialProperties::viscosity))
                 {
                   const SymmetricTensor<2,dim> shear_strain_rate = in.strain_rate[i]
                                                                    - 1./dim * trace(in.strain_rate[i]) * unit_symmetric_tensor<dim>();
--- a/benchmarks/shear_bands/shear_bands.cc
+++ b/benchmarks/shear_bands/shear_bands.cc
@@ -124,7 +124,7 @@ namespace aspect
             {
               double porosity = std::max(in.composition[i][porosity_idx],1e-4);
               out.viscosities[i] = eta_0 * std::exp(alpha*(porosity - background_porosity));
-              if (in.requests_property(MaterialProperties::viscosity))
+              if (in.requests_property(MaterialModel::MaterialProperties::viscosity))
                 {
                   const SymmetricTensor<2,dim> shear_strain_rate = in.strain_rate[i]
                                                                    - 1./dim * trace(in.strain_rate[i]) * unit_symmetric_tensor<dim>();
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -221,8 +221,7 @@ namespace aspect
       inline Property operator |= (Property &d1,
                                    const Property d2)
       {
-        d1 = (d1 | d2);
-        return d1;
+        return (d1 | d2);
       }
     }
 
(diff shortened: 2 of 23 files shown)


gassmoeller:
Note: Most of the change in include/aspect/material_model/interface.h is just changed indentation and can be removed by excluding whitespace in the github diff settings.

source/material_model/interface.cc
```diff
@@ -407,6 +411,25 @@ namespace aspect
...
+      //TODO: Remove this once all callers set requested_properties correctly
+      if ((property & MaterialProperties::Property::viscosity) != 0)
+        return (strain_rate.size() != 0);
+
+      //TODO: Remove this once all callers set requested_properties correctly
+      if ((property & MaterialProperties::Property::reaction_terms) != 0)
+        return (strain_rate.size() != 0);
+
+      // Note that this means requested_properties can have other properties than
+      // property, but at least property.
```
naliboff, line 427: was this the intended wording?
```suggestion
      // property, but at least one property.
```

gassmoeller:
I think I addressed all comments and extended and reworded the documentation a bit.

> is the main purpose of the proposed change to improve model run time by having the material perform specific operations based on the input property, or are there other additional advantages related to future PRs? I may have missed this in an earlier discussion.

Yes the main benefit will be improved runtime by avoiding calculating viscosity and reaction terms if only one of them is necessary. There may be other benefits I do not see yet. It also detangles the meaning of `if strain_rate.size() == 0`. At the moment it is not clear if this refers to viscosity or reaction terms or both. And if a material model does not know that sometimes strain_rate will not be provided it may access invalid memory..

> No objection, but curious as to why this approach is optimal.

I do not know if it is optimal. There are other approaches that are simpler to understand (a structure of several `bool` variables). But I copied the approach of `NonlinearDependence` and of the deal.II `UpdateFlags` structure (used for the `FEValues` class).

tjhei:
> Would you like an additional reviewer before merging

I have a few comments, please wait with merging.

include/aspect/material_model/interface.h
```diff
@@ -167,9 +167,66 @@ namespace aspect
...
+      inline Property operator | (const Property d1,
+                                  const Property d2)
+      {
+        return Property(static_cast<int>(d1) | static_cast<int>(d2));
+      }
+
+      inline Property operator |= (Property &d1,
+                                   const Property d2)
+      {
+        d1 = (d1 | d2);
```
tjhei, line 224: why not ``return (d1 | d2);``?

include/aspect/material_model/interface.h
```diff
@@ -180,201 +237,216 @@ namespace aspect
...
+         * Constant version of get_additional_input() returning a const pointer.
+         */
+        template <class AdditionalInputType>
+        const AdditionalInputType *get_additional_input() const;
+
+      private:
+        /**
+         * A member variable that stores which properties the material model should compute.
+         */
+        MaterialProperties::Property requested_properties;
```
tjhei, line 449: Maybe we should not make this private? All other properties in this class are public (I see it as a struct).

source/material_model/interface.cc
```diff
@@ -407,6 +411,25 @@ namespace aspect
...
+      if ((property & MaterialProperties::Property::viscosity) != 0)
+        return (strain_rate.size() != 0);
+
+      //TODO: Remove this once all callers set requested_properties correctly
+      if ((property & MaterialProperties::Property::reaction_terms) != 0)
+        return (strain_rate.size() != 0);
+
+      // Note that this means 'requested_properties' can include other properties than
+      // just 'property', but in any case it at least requests 'property'.
+      return (requested_properties & property) == property;
```
tjhei, line 428: This is a bit limiting because you can not use this to see if a or b is needed: ``in.requests_property(a|b)``. If we make the member public, we won't even need this function...

tjhei:
I really like the idea!

include/aspect/material_model/interface.h
gassmoeller: Copy-paste from line 110. I changed it.

include/aspect/material_model/interface.h
gassmoeller: I am fine with making it public. I made it private because I thought all use-cases are handled by the request_property function and I find the internal data structure complicated.

source/material_model/interface.cc
gassmoeller: True I had not thought of this case. Still I would like to keep the function, because using `requested_properties` directly requires you to understand bitwise operators, which are a pain. Also I can nicely handle the backwards compatibility issues inside the function (because the implementation is in one place instead of 30 material models). It is not important to me to keep it inside the struct though. If we make the variable public it might as well be a MaterialModel::Utilities function.

gassmoeller:
What do you think of this version?

source/material_model/interface.cc
tjhei: >  Still I would like to keep the function

true, I agree. Keeping it in the class here makes sense as well.

tjhei: what about making the last line ``!=0`` instead of ``== property``? Then one can pass ``a|b`` as well.

benchmarks/shear_bands/shear_bands.cc
```diff
@@ -124,7 +124,7 @@ namespace aspect
             {
               double porosity = std::max(in.composition[i][porosity_idx],1e-4);
               out.viscosities[i] = eta_0 * std::exp(alpha*(porosity - background_porosity));
-              if (in.strain_rate.size())
+              if (in.requests_property(MaterialModel::MaterialProperties::viscosity))
```
tjhei, line 127: it is a bit unfortunate that we have to specify ``MaterialModel::`` here because this class is not inside the namespace. I am not sure if that is worth changing.

gassmoeller: I agree this line is not so nice. But this benchmark has a number of different plugins in the general `aspect` namespace that all reference each other (see instantiations below), so putting them all in the usual namespaces will require some more changes to the file. I think the current way is better. Is that ok with you? Otherwise I can also open a separate PR that moves the classes into their namespaces.

```
// explicit instantiations
namespace aspect
{
  namespace ShearBands
  {
    ASPECT_REGISTER_MATERIAL_MODEL(ShearBandsMaterial,
                                   "shear bands material",
                                   "A material model that corresponds to the setup to"
                                   "generate magmatic shear bands described in Katz et al., "
                                   "Nature, 2006.")

    ASPECT_REGISTER_POSTPROCESSOR(ShearBandsPostprocessor,
                                  "shear bands statistics",
                                  "A postprocessor that writes output for the porosity field, "
                                  "which can be used to calculate the angle of the shear "
                                  "bands in the model.")

    ASPECT_REGISTER_POSTPROCESSOR(ShearBandsGrowthRate,
                                  "shear bands growth rate",
                                  "A postprocessor computes the growth rate of melt bands as "
                                  "presicted from linear stability analysis (Spiegelman, 2003), "
                                  "and compares modelled and analytical solution.")

    ASPECT_REGISTER_INITIAL_COMPOSITION_MODEL(ShearBandsInitialCondition,
                                              "shear bands initial condition",
                                              "Composition is set to background porosity plus "
                                              "white noise.")

    ASPECT_REGISTER_INITIAL_COMPOSITION_MODEL(PlaneWaveMeltBandsInitialCondition,
                                              "plane wave melt bands initial condition",
                                              "Composition is set to background porosity plus "
                                              "plane wave.")
  }
}

```

tjhei: I agree that the current status is good enough.

Issue #2347, opened 2018-06-20
update_prm_files.sh doesn't work correctly on Mac


anne-glerum:
The update_prm_files_to_2.0.0_model_settings.sed file does not update input files correctly (see fig for result) on macOS High Sierra 10.13.4.

![screen shot 2018-06-20 at 14 27 37](https://user-images.githubusercontent.com/7631624/41685748-63f79c36-7496-11e8-92da-834c6bb81aba.png)

anne-glerum:
Just ran into an update script error again for the update_prm_files_to_2.0.0_model_settings script:

`sed: 203: ./update_scripts/prm_files/update_prm_files_to_2.0.0_model_settings.sed: command i expects \ followed by text`

Running on MacOS Mojave 10.14.5

tjhei:
can you try to edit
```
  sed -i.tmp -f $script "$@"
```
to say
```
  sed -i .tmp -f $script "$@"
```
? (add a space before ``.tmp``)

anne-glerum:
I changed the line you mentioned in `update_prm_files.sh`, but get the same error. The error complains about the underlying file `update_scripts/prm_files/update_prm_files_to_2.0.0_model_settings.sed`.

tjhei:
ah, got it. Can you please try to change
```
i # The parameters below this comment were created by the update script\
```
to
```
i\
# The parameters below this comment were created by the update script\
```
(``i text`` is a GNU extension that doesn't work under osx): https://www.gnu.org/software/sed/manual/sed.html#Other-Commands

anne-glerum:
mmm, there is no error any more, but there are too many `end`s inserted now, see the attached diff and snapshot.
![Screenshot 2020-04-14 at 17 43 15](https://user-images.githubusercontent.com/7631624/79244660-73f70d00-7e77-11ea-8245-cb93a401c831.png)
[diff_resume_free_surface.txt](https://github.com/geodynamics/aspect/files/4476296/diff_resume_free_surface.txt)

Sorry that I don't really have time to investigate it myself, sed is also not my strong suite.

tjhei:
> Sorry that I don't really have time to investigate it myself, sed is also not my strong suite.

same here and I don't have access to my OSX machine right now.

@gassmoeller do you have a way to test this right now?

gassmoeller:
[...]

PR #3458, opened 2020-04-17
Make parameter file update script work on Mac

gassmoeller:
The script to update parameter files to more recent versions did use GNU sed extensions that are not available on MacOS leading to broken files (see #2347). This PR fixes some of the issues.

### DIFF
--- a/benchmarks/entropy_adiabat/entropy_adiabat.prm
+++ b/benchmarks/entropy_adiabat/entropy_adiabat.prm
@@ -113,7 +113,7 @@ end
 # be different from the initial entropy in the model.
 subsection Boundary composition model
   set Fixed composition boundary indicators = top
-  set Model name = function
+  set List of model names = function
   subsection Function
     # Only the first component (entropy) matters, the other equations
     # are not solved but computed from entropy.
--- a/doc/update_scripts/prm_files/update_prm_files_to_2.0.0.sed
+++ b/doc/update_scripts/prm_files/update_prm_files_to_2.0.0.sed
@@ -16,11 +16,11 @@ s/set Model name = initial profile/set Model name = compute profile/g
 # when we find it, check if 'Magnitude at bottom' is inside, and
 # insert it, if not.
 :jump_before_gravity
-/subsection Radial linear/,/^[[:space:]]*end[[:space:]]*/ {
+/subsection Radial linear/,/[[:space:]]*end[[:space:]]*/ {
 # Print the line 'subsection ...' and read next line
 n
 # If next line was already 'end' jump out of this part
-/^ *end/b jump_before_gravity
+/^[[:space:]]*end/b jump_before_gravity
 
 :jump_in_gravity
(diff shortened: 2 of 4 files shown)


gassmoeller:
Sure, I was thinking of leaving it open and let @anne-glerum have a try so I can update if there are remaining issues. But it is a step forward, and I can always make a follow-up later.

tjhei:
on OSX I get:
```
cd cookbooks
% ~/aspect/doc/update_prm_files.sh
sed: -i may not be used with stdin
sed: -i may not be used with stdin
sed: 209: ~/aspect/doc/update_scripts/prm_files/update_prm_files_to_2.0.0_model_settings.sed: unterminated substitute pattern
sed: -i may not be used with stdin
sed: -i may not be used with stdin
```

gassmoeller:
Can you try that on an actual file (`~/aspect/doc/update_prm_files.sh ~/aspect/cookbooks/gplates-2d.prm`)? I suspect the stdin error should go away and I found at least one discussion which had the unterminated substitute error go away as well.

tjhei:
I get
```
sed: 209: update_prm_files_to_2.0.0_model_settings.sed: unterminated substitute pattern
```
which I can get rid of by going back to ``s/\n *$//``

With it, the "Model settings" look correct, but the lines after that are replaced by nonsense (repeated ``# The parameters below this comment were created`` ...).

gassmoeller:
This should now fix everything reported in #2347. I found another non POSIX compliant command. I also reran the script on all the parameter files in the repo and updated a few.
I can still not reproduce the `sed: 209: update_prm_files_to_2.0.0_model_settings.sed: unterminated substitute pattern` @tjhei reported. Running the script with gnu sed or gnu sed with `--posix` both works for me. Timo could you try again, and if the problem is still there send me a sed and MacOs version? Not sure how to test this without access to a BSD or Mac machine.

tjhei:
```
$ ./doc/update_prm_files.sh cookbooks/gplates/gplates_2d.prm 
sed: 209: ./doc/update_scripts/prm_files/update_prm_files_to_2.0.0_model_settings.sed: unterminated substitute pattern
$ sw_vers -productVersion
10.13.6
```

``sed --version`` doesn't exist, see https://stackoverflow.com/questions/37639496/how-can-i-check-the-version-of-sed-in-os-x

we could a) require installing GNU sed or b) move away to python or something else?

tjhei:
Anything you want to do here before the release?

gassmoeller:
I reverted the change to the command that did not work for you. Everything that is left should be strictly improvements.

tjhei:
Looks good. It does sensible things with cookbooks/gplates/gplates_2d.prm (no change)  and updates tests/update_script.prm

gassmoeller:
Great, then this is ready to go from my side. @anne-glerum if you have a minute next week could you check if this resolves all remaining issues for you?

tjhei:
sounds good. I will merge to have this in the release...

PR #3459, opened 2020-04-17
Update parameter files

gassmoeller:
This PR applies our parameter file update script to all the parameter files we currently have in our repository. Due to overlaps in PRs with changes of parameter names we merged some files that use outdated parameter names, this should update most of them.
I will open a separate issue or PR for benchmarks/newton_solver_benchmark_set/tosi_et_al_2015 that seems to be broken, but this PR already updates the obvious lines.

PR #3460, opened 2020-04-17
Fix newton tosi benchmark

gassmoeller:
This PR sits on top of #3459 (first commit) and fixes some of the remaining issues of the benchmarks/newton_solver_benchmark_set/tosi_et_al_2015 setup (second and third commit). The parameter file now loads without error, but only runs for a few timesteps and crashes with a floating point exception the first time it switches to the Newton solver. @Mfraters could you have a look at my changes and maybe give the model a try as well? I had to comment one of the parameters to the material model that talked about the analytical derivatives, because the material model of the other tosi benchmark that is used did not know the parameter. Did you maybe run the benchmark with a modified version of this material model?

Log file
[log.txt](https://github.com/geodynamics/aspect/files/4494307/log.txt)

Error message:
```
[rene-laptop:16768] *** Process received signal ***
[rene-laptop:16768] Signal: Floating point exception (8)
[rene-laptop:16768] Signal code: Invalid floating point operation (7)
[rene-laptop:16768] Failing at address: 0x557906d93046
[rene-laptop:16768] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x15540)[0x7f4610e41540]
[rene-laptop:16768] [ 1] aspect(_ZN6dealii7numbers12NumberTraitsIdE10abs_squareIdEENSt9enable_ifIXaasrSt7is_sameIT_dE5valuesrNS0_18is_cuda_compatibleIS6_vEE5valueEdE4typeERKd+0x1c)[0x557906d93046]
[rene-laptop:16768] [ 2] aspect(_ZN6dealii8internal12compute_normILi2EdEENS_7numbers12NumberTraitsIT0_E9real_typeERKNS0_24SymmetricTensorAccessors11StorageTypeILi2EXT_ES4_E16base_tensor_typeE+0x51)[0x5579072139dc]
[rene-laptop:16768] [ 3] aspect(_ZNK6dealii15SymmetricTensorILi2ELi2EdE4normEv+0x1c)[0x557907213676]
[rene-laptop:16768] [ 4] aspect(_ZN6aspect9Utilities18compute_spd_factorILi2EEEddRKN6dealii15SymmetricTensorILi2EXT_EdEES6_d+0x6e)[0x557907e204be]
[rene-laptop:16768] [ 5] aspect(_ZNK6aspect10Assemblers31NewtonStokesIncompressibleTermsILi2EE7executeERNS_8internal8Assembly7Scratch11ScratchBaseILi2EEERNS4_8CopyData12CopyDataBaseILi2EEE+0x248c)[0x557907914d04]
[rene-laptop:16768] [ 6] aspect(_ZN6aspect9SimulatorILi2EE28local_assemble_stokes_systemERKN6dealii18TriaActiveIteratorINS2_15DoFCellAccessorINS2_10DoFHandlerILi2ELi2EEELb0EEEEERNS_8internal8Assembly7Scratch12StokesSystemILi2EEERNSC_8CopyData12StokesSystemILi2EEE+0x4cd)[0x557907941215]
[rene-laptop:16768] [ 7] aspect(_ZZN6aspect9SimulatorILi2EE22assemble_stokes_systemEvENKUlRKN6dealii18TriaActiveIteratorINS2_15DoFCellAccessorINS2_10DoFHandlerILi2ELi2EEELb0EEEEERNS_8internal8Assembly7Scratch12StokesSystemILi2EEERNSC_8CopyData12StokesSystemILi2EEEE_clESA_SG_SK_+0x37)[0x55790794a3cb]
[rene-laptop:16768] [ 8] aspect(_ZN6dealii10WorkStream3runIZN6aspect9SimulatorILi2EE22assemble_stokes_systemEvEUlRKNS_18TriaActiveIteratorINS_15DoFCellAccessorINS_10DoFHandlerILi2ELi2EEELb0EEEEERNS2_8internal8Assembly7Scratch12StokesSystemILi2EEERNSE_8CopyData12StokesSystemILi2EEEE_ZNS4_22assemble_stokes_systemEvEUlRKSL_E0_NS_16FilteredIteratorISA_EESH_SL_EEvRKT1_RKNS_8identityIST_E4typeET_T0_RKT2_RKT3_jj+0x26f)[0x55790794d466]
[rene-laptop:16768] [ 9] aspect(_ZN6aspect9SimulatorILi2EE22assemble_stokes_systemEv+0x7e4)[0x557907941d6a]
[rene-laptop:16768] [10] aspect(_ZN6aspect9SimulatorILi2EE43assemble_and_solve_defect_correction_StokesERNS_25DefectCorrectionResidualsEb+0x677)[0x557907b8ab95]
[rene-laptop:16768] [11] aspect(_ZN6aspect9SimulatorILi2EE42solve_iterated_advection_and_newton_stokesEv+0x3b1)[0x557907b86507]
[rene-laptop:16768] [12] aspect(_ZN6aspect9SimulatorILi2EE14solve_timestepEv+0x12f)[0x5579079e59f7]
[rene-laptop:16768] [13] aspect(_ZN6aspect9SimulatorILi2EE3runEv+0x521)[0x5579079e2b73]
[rene-laptop:16768] [14] aspect(_Z13run_simulatorILi2EEvRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES7_bbb+0x369)[0x55790714bc08]
[rene-laptop:16768] [15] aspect(main+0x64c)[0x5579071250a2]
[rene-laptop:16768] [16] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3)[0x7f46109251e3]
[rene-laptop:16768] [17] aspect(_start+0x2e)[0x557906d48c4e]
[rene-laptop:16768] *** End of error message ***
Floating point exception (core dumped)
```


benchmarks/check.sh
```diff
@@ -151,6 +148,11 @@ wait # newton_solver_benchmark_set/nonlinear_channel_flow depends on nonlinear_c
 
 ( (cd tosi_et_al_2015_gcubed/ && make_lib && run_all_prms ) || { echo "FAILED"; exit 1; } ) &
 
+# TODO: broken
+# is down here, because it requires the library from the
+# tosi_et_al_2015_gcubed/ model above
+# ( (cd newton_solver_benchmark_set/tosi_et_al_2015/ && run_all_prms ) || { echo "FAILED"; exit 1; } ) &
```
tjhei, line 154: does this run with your changes now?

tjhei:
I will let @MFraters take a look but this PR is a step forward.

MFraters:
Sorry for my late reply. The changes look good (and sensible) to me, but will test them today as you suggest. I am very happy to see that you correct my path mistake :)

MFraters:
So, it took me a bit longer to look into. First I thought that it was just the input file which was the problem, because I only used it as a basis for the bash scripts. Then I thought it might have been the change in the tosi.cc file or a change in the code. In the end it was that the benchmark used a material model called tosi (which seems to be like @gassmoeller suggested a modified version of the tosi benchmark file). 

Should I add this in a different pull request, or make a pull request to your branch Rene?

gassmoeller:
Make a pull request to my branch that way we can fix it here and merge everything together. Thanks for checking.

benchmarks/check.sh
```diff
@@ -151,6 +148,11 @@ wait # newton_solver_benchmark_set/nonlinear_channel_flow depends on nonlinear_c
 
 ( (cd tosi_et_al_2015_gcubed/ && make_lib && run_all_prms ) || { echo "FAILED"; exit 1; } ) &
 
+# TODO: broken
```
MFraters, line 151: I think this is fixed now

gassmoeller:
Ok, I merged @MFraters material model, but since it was a copy of the one in the original Tosi benchmark with an extension for the Newton solver, I merged the two material models into one. In the process I realized that there are no tests for either benchmark (Tosi benchmark with/without Newton solver), so I added two tests. I think this is in a reasonable state right now, but it probably needs another review.

gassmoeller:
I think there were two bugs in the finite difference derivative calculation of this benchmark. @MFraters could you take a look at my last commit? 
1. When computing the finite difference derivative we divided by a different delta than we modified the strain rate by (slightly different).
2. For the xy component we multiplied the delta strain rate term by 0.5, and the comment said because of symmetry of the tensor. But I do not see where this would happen, to me it looks like we do the following: `(eta(x+0.5d) - eta(x)) / d`, which is not an accurate derivative.

I checked the results of the base input file in the newton benchmark folder and the convergence behavior is slightly different (a few more nonlinear iterations with my new version, so not a bug? Or just a harder problem, because it is more accurately solved?).

benchmarks/tosi_et_al_2015_gcubed/tosi.cc
```diff
@@ -149,19 +149,18 @@ namespace aspect
...
-                        deta[0][0] /= std::fabs(dstrain_rate[i][0][0]) * finite_difference_accuracy;
+                        deta[0][0] /= std::fabs(in.strain_rate[i][0][0]) * finite_difference_accuracy;
                       else
                         deta[0][0] = 0;
 
                       // derivative in xy direction
                       dstrain_rate = in.strain_rate[i];
-                      // 0.5 because symmetry doubles this entry
-                      dstrain_rate[1][0]  += 0.5 * std::fabs(in.strain_rate[i][1][0]) * finite_difference_accuracy; 
+                      dstrain_rate[1][0]  += std::fabs(in.strain_rate[i][1][0]) * finite_difference_accuracy;
```
MFraters, line 158: unfortunately, this is not correct. I don't remember the details of why it is happening (long time ago and it is late now), but I remember that this needs to be there. Just to be sure I quickly ran a test to compare it against the analytic solution (just removing the if statement so that they both run and output both to screen), assuming the analytic solution is correct of course ;)

Without 0.5: 
```
analytical derivative = 1.43213e-10 -1.33969e-12 -1.33969e-12 -1.44939e-10
finite difference derivative = 1.43212e-10 -2.76208e-12 -2.76208e-12 -1.4494e-10
```
with the 0.5:
```
analytical derivative = 1.43213e-10 -1.33969e-12 -1.33969e-12 -1.44939e-10
finite difference derivative = 1.43212e-10 -1.38104e-12 -1.38104e-12 -1.4494e-10
```

I also tested the change from dstrain_rate to in.strain_rate and that didn't matter at all:
```
analytical derivative = 1.43213e-10 -1.33969e-12 -1.33969e-12 -1.44939e-10
finite difference derivative = 1.43212e-10 -1.38104e-12 -1.38104e-12 -1.4494e-10
```

gassmoeller: Thanks for checking, I think I figured out the reason. When we modify the yx (1,0) component of the strain rate we implicitly also modify the xy (0,1) component, because it is a symmetric tensor. Therefore the delta in strain rate when computing the derivative for the off-diagonal terms is actually bigger than the ` 1e-7 * value` unless we account for the symmetry. I removed the change and added documentation.

> I also tested the change from dstrain_rate to in.strain_rate and that didn't matter at all:

Well it may be too small to make a difference here, but do you agree that we need to divide by the actual delta we modify the strain rate by, and not the modified value of the strain rate times the relative difference? The difference is likely around 1e-7 times the derivative.

MFraters: Ah, yes, that makes sense. We can't check them independently due to symmetry. Thanks for expanding the documentation!

I was already a bit worried when I looked again at the viscoplastic derivatives, but this makes me more worried, so I will need to take a closer look at that. (worried in the sense that it is not optimal. If that is the case we might actually improve the convergence rate)

> Well it may be too small to make a difference here, but do you agree that we need to divide by the actual delta we modify the strain rate by, and not the modified value of the strain rate times the relative difference? The difference is likely around 1e-7 times the derivative.

Yes, I agree that that is more correct.

PR #3464, opened 2020-04-20
Make error message clearer.

bangerth:
Related to Felipe's problem on the mailing list.

/rebuild


naliboff:
@bangerth - Thanks! This will indeed hopefully prevent some confusion down the line. Once the test results are updated this will be good to go.

tjhei:
I went ahead and pushed the update to the test results.

PR #3466, opened 2020-04-21
parameter_view: add run.sh script

tjhei:
Chrome can not open .xls from file:// urls locally for security reasons.
Add a simple shell script to launch the browser.


gassmoeller:
Oh nice, I knew the problem but did not figure out a nice solution when I wrote the .xls and used Firefox for local files instead.

PR #3468, opened 2020-04-21
Fix renaming variable checkpoint data

### LINKED ISSUE #3463: resume computation fails with particles
I get the error described here https://github.com/geodynamics/aspect/issues/3253
when I want to restart a model that uses _particles_ in the _list of postprocessors_. 

The first run where the snapshot is created works fine, but when restarting the model, the error message is: "Cannot open snapshot mesh file or read the triangulation stored there." 
When I outcomment the _particles_ in the _list of postprocessors_, the model restarts as  it should. The error appears both, when restarting from a snapshot from time 0 and from later times. 

I atttach an input file that produces the error: 
[particle_generator_random_uniform_restart.txt](https://github.com/geodynamics/aspect/files/4503420/particle_generator_random_uniform_restart.txt)
It is the particle_generator_random_uniform test where I added the lines to restart it. I also tried it with several other tests and the same error appears when adding _particles_ in the _list of postprocessors_.

I use: 
--     . version 2.2.0-pre (master, 767fb667f)
--     . using deal.II 9.1.1
--     . using p4est 2.2.0

gassmoeller:
This should fix #3463. In #3181 I forgot to also rename the variable data file that is written since deal.II 9.1 (and since our tester runs on 9.0 that did not show until now).


source/simulator/checkpoint_restart.cc
```diff
@@ -392,6 +392,16 @@ namespace aspect
 #if DEAL_II_VERSION_GTE(9,1,0)
             move_file (parameters.output_directory + "restart.mesh_fixed.data",
                        parameters.output_directory + "restart.mesh_fixed.data.old");
+
+            std::fstream check_variable_file((parameters.output_directory + "restart.mesh_variable.data").c_str());
+            const bool variable_data_file_exists = check_variable_file.is_open();
```
tjhei, line 397: can you use ``Utilities::fexists()`` instead?

gassmoeller: Absolutely, forgot about that function.

gassmoeller:
@EstherHeck reported in #3463 that her models work now. So good to go.

PR #3472, opened 2020-04-22
averaged strain rate postprocessor

tjhei:
This PR adds a new visualization postprocessor "averaged strain rate" that computes an average strain rate per cell. This is convenient to use as the real strain rate seems to be quite noisy and contains over/undershoots.


gassmoeller:
Hmm, I agree that the current strain rate is noisy and we should have a better output. But I worry that having a `strain rate`, `strain rate tensor`, `volumetric strain rate`, and `average strain rate` postprocessor is getting confusing. Do you think it would make sense to replace the current output of `strain rate` with a smoothed version instead (e.g. projected into a Q1 space)? Or give `strain rate` an input parameter that allows to average the nodal values? I realize that would be more work, let me know if you prefer to stick with the current approach and I will review this one instead.

gassmoeller:
Superseded by #3478?

tjhei:
yes, thanks!

PR #3473, opened 2020-04-22
Add plastic damper to plasticity

naliboff:
This PR is a work in progress for adding a stabilization term to plasticity to produce mesh-independent shear band widths and improved convergence behavior. The methodology follows that of Duretz et al., 2020, GRL: 

Opening this now to provide a place for discussion and sharing of test results. So far, no improvement to convergence behavior for the VEP or VP cases: https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2019GL086027


naliboff:
Some preliminary results for visco-viscoplastic (1 time step) and viscoelastic-viscoplastic models (1.3 Myr) following the setup described in Duretz et al.

naliboff:
@jdannberg - @cedrict and I spoke with Thibault and there a number of changes we need to make before adding the stabilization term. I'll send an email summarizing the discussion and strategies for proceeding.

bobmyhill (changes requested):
@naliboff, I know this is a work in progress, but I've made a few very minor suggestions. A simple test is still needed to make sure the damper is working as expected (the Peierls one was great!).

Are the benchmarks likely to be a WIP for a long time? If so, might it be good to split this PR into 1) viscosity-damper + test and 2) benchmarks.

naliboff:
@bobmyhill - thanks for the review! I think this is actually pretty close to a final state, and aside from addressing reviews I just need to rebase once #3825 is merged. I'm going to remove all the other benchmarks and just the one from #3825 to demonstrate the damper is working properly. I'll try and do that over the next day or two as it would be great to have plasticity integrated soon into the composite rheology formulations.

naliboff:
@bobmyhill - Thanks again the review. I addressed your comments about source/header files, and then squashed everything into a single commit during a rebase. I removed all the current benchmarks, and will just use a modified version of the parameter file from #3825 to make sure the damper is working properly. 

In going through the rebase, I realize the current structure should work with a visco plastic rheology, but a few more adjustments need to be made when viscoelasticity is included.

EstherHeck:
hi, great that this is progressing, looking forward to it! While going through the files, I saw a typo in line 335 in visco_plastic.cc :
// than the lithostatic prssure.

naliboff:
@EstherHeck - Thanks! will correct that with other cleanup items after a review.

naliboff:
Ok, this should now be ready for review, assuming the tests are still passing after the last series of commits. I've done a comparison with @cedrict and we are getting effectively (within expected bounds) the same result for the benchmark case included in this PR.

bobmyhill:
Hi @naliboff,

If I understand correctly, pre-yield viscosity is a constant viscosity element in parallel with the (optionally damped) plastic element. May I suggest that this element gets added to a composite rheology model, rather than the damped Drucker-Prager? Otherwise ASPECT will end up with multiple "max viscosity" elements, each of which will influence results.

bobmyhill (changes requested):
Nice! Just a couple of minor edits, plus suggested removal of the pre-yield viscosity term.

naliboff:
@bobmyhill - Thanks for the review, going through the comments now! 

I am a bit confused about the suggestion for defining the pre_yield_viscosity in the composite rheology model, rather than here. In `Visco Plastic`, the pre_yield_viscosity is simply the viscosity arising from the creep laws +/- viscoelasticity (i.e., not set by user). I specified this as a constant value in the other two material models (`Drucker Prager`, `Dynamic Friction`) as it is enforced that the damper is never used in these models.

I may be missing the key point here. Regardless, it might be good to discuss how this will/should work with the composite rheology model and simplifying things as much as possible. Probably easiest to do offline? I'll go through the composite rheology later today!

include/aspect/material_model/rheology/drucker_prager.h
```diff
@@ -96,7 +107,10 @@ namespace aspect
                              const double angle_internal_friction,
                              const double pressure,
                              const double effective_strain_rate,
-                             const double max_yield_stress) const;
+                             const double max_yield_stress,
+                             const bool use_plastic_damper,
+                             const double damper_viscosity,
+                             const double pre_yield_viscosity) const;
```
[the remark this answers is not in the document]
bobmyhill: Ok, after out chat I agree that pre-yield viscosity is fine, but I think that it should be given a default value, i.e.
```suggestion
                             const double pre_yield_viscosity=std::numeric_limits<double>::infinity()) const;
```
Then the function calls from the material models don't need to include this value (the two places where they read 1.0 at the moment).

source/material_model/rheology/drucker_prager.cc
```diff
@@ -57,13 +58,33 @@ namespace aspect
...
+                                        ( 1 + ( damper_viscosity / pre_yield_viscosity ) );
+
+            const double pre_yield_strain_rate = total_stress / ( 2. * pre_yield_viscosity);
+
+            const double plastic_strain_rate = effective_strain_rate - pre_yield_strain_rate;
+
+            plastic_viscosity = yield_stress / (2.*plastic_strain_rate) + damper_viscosity;
+
+            // Effective viscosity
+            plastic_viscosity = 1. / (1./plastic_viscosity + 1./pre_yield_viscosity);
```
bobmyhill, line 84: Thinking some more on our conversation, I think pre_yield_viscosity should have some effect whether or not the plastic damper is used. In other words, this function should produce identical output for the inputs:
`use_plastic_damper = false`
`damper_viscosity = 1e8`
`pre_yield_viscosity = 1e10` (for example)
or
`use_plastic_damper = true`
`damper_viscosity = 0.`
`pre_yield_viscosity = 1e10`

naliboff: @bobmyhill - that was the original intent, but unfortunately I don't see a way for this to work without breaking backwards compatibility. 

When `use_plastic_damper = true`, the plastic viscosity uses the plastic strain rate (`plastic_strain_rate`) rather than the total strain rate (standard method) to calculate the plastic viscosity.

Long-term, I would be in favor of only using the damper method where a unique plastic strain is calculated, but that will be a major switch unless there is someway around it I'm not seeing.

include/aspect/material_model/rheology/drucker_prager.h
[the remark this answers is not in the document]
naliboff: In order to ensure the function call still has all the same visible inputs between material models, would you be fine with replacing `1.0` with `std::numeric_limits<double>::infinity()` in the function calls themselves. 

As we discussed, I'm supportive of restructuring much of `visco plastic` and `drucker prager` in the near future.

source/material_model/rheology/drucker_prager.cc
bobmyhill: I'm not sure I understand. You've added new variables to the function, so I think backwards compatibility is already broken.

My suggestion is just to move 
`plastic_viscosity = 1. / (1./plastic_viscosity + 1./pre_yield_viscosity);` outside of the if clause, and have pre_yield_viscosity = std::numeric_limits<double>::infinity() as a default value. Is there some reason this wouldn't work?

include/aspect/material_model/rheology/drucker_prager.h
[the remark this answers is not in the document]
bobmyhill: Yes, I'd be fine with that :)

source/material_model/rheology/drucker_prager.cc
bobmyhill: Oh, I think I understand now. Is it true that pre_yield_viscosity is only useful when the viscous damper is active? If that's true, then the current code design makes sense.

bobmyhill:
Thanks for making those changes and explaining the code, @naliboff! I'm happy with the code design, especially as we aren't yet sure what constraints the as-yet-unwritten higher level code will place on this plugin.

 I'll leave this to one of the maintainers for final suggestions / approval.

source/material_model/rheology/drucker_prager.cc
naliboff: @bobmyhill - sorry for the confusion through all that! Yes, ` pre_yield_viscosity` is only used when the damper is active. They are not intrinsically related to other, but the averaging with         `pre_yield_viscosity` is used by Duretz et al. We may want to rethink about the the default assumptions later on and break backwards compatibility in multiple places in one fell swoop.

naliboff:
@bobmyhill - Thanks again for the thorough review and discussion. developers/maintainers - this should be in pretty good shape. As soon as it is merged, @bobmyhill can proceed with #3884.

gassmoeller:
@naliboff: I had a look at this PR and I would like to propose some small changes (overall it is ok). I will open a PR to your branch, everything else seems ok.

gassmoeller:
See comments in https://github.com/naliboff/aspect/pull/7.

bobmyhill (approved):

PR #3474, opened 2020-04-23
Add citation for latest paper

gassmoeller:
Adding a reference to our recently published paper in GJI. Did I understand correctly what I have to do to add it to the list of citable papers? Follow-up PR for the www branch to come in a minute.

PR #3475, opened 2020-04-23
Add website checkbox for projected density approximation

gassmoeller:
Companion PR to #3474. Are these the only changes necessary for the www branch?


tjhei:
no, but I will do the update if #3474 is merged.

PR #3476, opened 2020-04-24
Make Assert AssertThrow in vp 

anne-glerum:
To get an informative error message for the combination
 `set Strategy                                 = minimum refinement function, viscosity `
 `set Skip solvers on initial refinement       = true`
 `set Skip setup initial conditions on initial refinement = true`
which leads to zero temperatures in the pre-refinement phase, which are not allowed when computing the viscosity in the visco_plastic material model. Viscosity is however requested by the refinement strategy. In parallel release mode, the model just hangs at the moment.

### DIFF
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -147,14 +147,14 @@ namespace aspect
           // gradient used when calculating the viscosity. This allows the same activation volume
           // to be used in incompressible and compressible models.
           const double temperature_for_viscosity = temperature + adiabatic_temperature_gradient_for_viscosity*pressure;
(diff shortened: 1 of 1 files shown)


anne-glerum:
@naliboff Yes, actually both `set Skip solvers on initial refinement` and `set Skip setup initial conditions on initial refinement` should probably be set to false to get a reasonable viscoplastic viscosity for refinement. But you can limit the number of nonlinear iterations in prerefinement, so it's not too bad.

But I think there's a chance others end up with this combination of parameters too, so I think it should be an AssertThrow and not an Assert.

naliboff:
I agree it definitely should be an `AssertThrow` given the scenario you ran into. Thanks again for the change!

PR #3478, opened 2020-04-24
Visualization: add averaged stress/strain option

tjhei:
Introduce the setting "Point-wise stress and strain" to average
variables related to stress and strain in visualization.

Note: as discussed in the biweekly ASPECT meeting, this changes the default to average the values.

alternative to #3472


tjhei:
I added a second version for each test that depended on point-wise output to have both versions around.

include/aspect/postprocess/visualization.h
```diff
@@ -368,6 +387,13 @@ namespace aspect
         void
         write_plugin_graph (std::ostream &output_stream);
 
+
```
gassmoeller, line 390: remove additional empty line

include/aspect/postprocess/visualization.h
```diff
@@ -454,6 +480,14 @@ namespace aspect
          */
         bool filter_output;
 
+
```
gassmoeller, line 483: same here

include/aspect/postprocess/visualization/averaged_strain_rate.h
```diff
@@ -0,0 +1,59 @@
+/*
```
gassmoeller, line 1: Is this file still needed, I thought we would use the existing postprocessor?

source/postprocess/visualization/averaged_strain_rate.cc
```diff
@@ -0,0 +1,118 @@
+/*
```
gassmoeller, line 1: same here, is this file needed?

tjhei:
thanks for the review. Fixed!

PR #3480, opened 2020-04-24
Allow tangential mesh velocities for zero Stokes velocity boundaries

anne-glerum:
This PR applies the same logic to zero Stokes velocity boundaries that are marked as `Additional tangential mesh deformation boundary indicators` as prescribed Stokes velocity boundaries. At the moment, zero Stokes velocity boundaries have zero mesh velocity even though they can be marked as `Additional tangential mesh deformation boundary indicators` which should allow them to move. 

I didn't see any tests with this combination of input parameters. Should we add one?

### For all pull requests:


gassmoeller:
Oh, and yes a test would be good.

anne-glerum:
> Oh, and yes a test would be good.

Alright, I've added a test. A 2d box with a free surface has an outward flow on the right boundary. The left boundary is set to zero velocity, but the mesh along both the left and right boundary is allowed to move. 

Checking the test results, the solved-for mesh displacement seems less than ideal. While the mesh moves along the left boundary, the top left few elements seem to lag behind. This creates a peak in the mesh that eventually leads to a crash (see figs below at t0, t4 and t10).

surface_and_velocity_t0
surface_and_velocity_t4
surface_and_velocity_t10

gassmoeller:
Hmm, I have a question about your test: If you prescribe the velocity to zero at the left boundary, and you only use the free surface mesh deformation plugin, would you not expect the left boundary to stay unchanged? After all the free surface mesh deformation plugin deforms the mesh exactly according to the Stokes velocity. Shouldnt we only expect tangential mesh movement if either: 1. using it on a boundary with Stokes velocities different from zero, or 2. using a mesh deformation plugin like your diffusion (or the function plugin) that changes the mesh independently from the Stokes velocity?

anne-glerum:
> Hmm, I have a question about your test: If you prescribe the velocity to zero at the left boundary, and you only use the free surface mesh deformation plugin, would you not expect the left boundary to stay unchanged? After all the free surface mesh deformation plugin deforms the mesh exactly according to the Stokes velocity. Shouldnt we only expect tangential mesh movement if either: 1. using it on a boundary with Stokes velocities different from zero, or 2. using a mesh deformation plugin like your diffusion (or the function plugin) that changes the mesh independently from the Stokes velocity?

Implementation-wise, the Stokes velocity is only projected onto the free surface boundary. The free surface plugin is not concerned with the other boundaries. In the mesh deformation interface, the mesh velocity of the rest of the domain is solved for, in this case with the tangential boundary condition on the left boundary.

The left top mesh corner point in this test is actually moving, although much slower than the rest of the left boundary. Maybe projection of the zero Stokes velocity set on the left boundary to the free surface does not lead to zero mesh velocity of the corner point?

In general, I agree that a test setup with a function prescribing the surface motion i.c.w. zero Stokes velocity and free tangential mesh movement along the left boundary would make more sense. There also doesn't seem to be a test using the function. I'll update the test.

anne-glerum:
@gassmoeller I've updated the test.

PR #3482, opened 2020-04-24
Extend particle property update interface

gassmoeller:
This follows a discussion with @naliboff. The current interface for updating particle properties does not give information about which cell the current particle is in, which e.g. prevents particle properties from calling some material models to figure out properties. I think we need to extend our current interface for that, and there is no reason particle properties should not know all information about the particle they are currently updating. 
In theory we could use the same interface as postprocessors (handing over a `DataPostprocessorInputs::Vector<dim>` input argument) or as the material model (handing over a `MaterialModelInputs` argument), but I think changing the interface as little as possible with just the small addition of handing over the particle iterator -- instead of the particle location and the particle properties -- works just as well. I have not yet updated all existing plugins (the change is backward compatible), but if we think this is useful and the right change I will do that.

@naliboff: You should now be able to create your particle property roughly the following way:
```
ElasticProperty<dim>::update_particle_property(const unsigned int data_position,
                                    const Vector<double> &solution,
                                    const std::vector<Tensor<1,dim> > &gradients,
                                    typename ParticleHandler<dim>::particle_iterator &particle)
{
MaterialModelInputs<dim> in(1, n_compositions);
in.temperature = solution[this->introspection().component_indices.temperature];
in.pressure, in.strain_rate ...
in.cell = particle->get_surrounding_cell(this->get_triangulation());

// create your additional outputs here

this->get_material_model().evaluate (in,out);
// get your additional outputs from out
additional_output = out. ...

particle->get_properties()[data_position] = additional_output[...];
}
```

### DIFF
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -344,6 +344,37 @@ namespace aspect
            * denotes the first component of this property, all other components
            * fill consecutive entries in the @p particle_properties vector.
            *
+           * @param [in] solution The values of the solution variables at the
+           * current particle position.
+           *
+           * @param [in] gradients The gradients of the solution variables at
+           * the current particle position.
+           *
+           * @param [in,out] particle The particle that is updated within
+           * the call of this function.
+           */
+          virtual
+          void
+          update_particle_property (const unsigned int data_position,
+                                    const Vector<double> &solution,
+                                    const std::vector<Tensor<1,dim> > &gradients,
+                                    typename ParticleHandler<dim>::particle_iterator &particle) const;
+
+          /**
+           * Update function. This function is called every time an update is
+           * request by need_update() for every particle for every property.
+           * It is obvious that
+           * this function is called a lot, so its code should be efficient.
+           * The interface provides a default implementation that does nothing,
+           * therefore derived plugins that do not require an update do not
+           * need to implement this function.
+           *
+           * @param [in] data_position An unsigned integer that denotes which
+           * component of the particle property vector is associated with the
+           * current property. For properties that own several components it
+           * denotes the first component of this property, all other components
+           * fill consecutive entries in the @p particle_properties vector.
+           *
            * @param [in] position The current particle position.
            *
            * @param [in] solution The values of the solution variables at the
@@ -354,7 +385,11 @@ namespace aspect
            *
            * @param [in,out] particle_properties The properties of the particle
            * that is updated within the call of this function.
+           *
+           * @deprecated Use the other overload of update_one_particle_property()
+           * instead.
            */
+          DEAL_II_DEPRECATED
           virtual
           void
           update_one_particle_property (const unsigned int data_position,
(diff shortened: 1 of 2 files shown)


include/aspect/particle/property/interface.h
```diff
@@ -329,6 +329,37 @@ namespace aspect
...
+          /**
+           * Update function. This function is called every time an update is
+           * request by need_update() for every particle for every property.
+           * It is obvious that
+           * this function is called a lot, so its code should be efficient.
+           * The interface provides a default implementation that does nothing,
+           * therefore derived plugins that do not require an update do not
+           * need to implement this function.
+           *
+           * @param [in] data_position An unsigned integer that denotes which
```
tjhei, line 341: can you mention that the location can be accessed using ``particle->get_location()`` and its properties with ``particle->get_properties()``?

naliboff:
@gassmoeller - Sorry for the delay and thanks for this! I agree that this seems like a sensible solution. I'm working through the elasticity on particles now and update once that is done or if issues arise.

naliboff:
@gassmoeller - A few questions, perhaps easier to discuss offline?

1. In order to set the `pressure_gradient` for the `MaterialModelInputs` we need access to `solution_gradients` (currently we just have `solution`). The particle property interface describes `gradients` as the gradients of the solution variables, but as this variable is currently used it appears just to contain the velocity gradients (ex: see `integrated_strain` particle property). I am missing something with regards to `gradients` or do we in fact need access to `solution_gradients`? 

2. The method proposed above for setting `in.cell` is giving a compiler error that I'm not sure how to address:

> error: cannot convert ‘dealii::Triangulation<2, 2>::cell_iterator {aka dealii::TriaIterator<dealii::CellAccessor<2, 2> >}’ to ‘const active_cell_iterator* {aka const dealii::TriaActiveIterator<dealii::DoFCellAccessor<dealii::DoFHandler<2, 2>, false> >*}’

naliboff:
@gassmoeller - A few updates below:

1. Per our discussion, I checked the new interface with the particle property `pT_path` and everything worked as expected. I'm still not clear as to where the error was in my original implementation, but I reverse engineered the .cc file from the updated `pT_path` and things are working now (with one caveat).
2. In examining the compositional field values accessed by the particles, I realized that by the time the particle properties are updated the compositional fields have already been updated to contain the correct stress values for that time step. While the stress values are initially computed on the reaction terms, the compositional fields are updated to contain the exact values.
3. As such, it turns out the new interface was not needed, as I just needed to access the compositional field data from the solution vector. However, the new interface is returning the correct viscosity values from the material model outputs, and this functionality may be useful down the line.
4. While the correct viscosity values are returned by the `MaterialModel Outputs`, the reaction terms are returned as zero. I'm not sure why this is the case, but perhaps I did not set up something correctly in the `MaterialModel Inputs`.

Question regarding how to proceed: Should we find a different use case for this PR (ex: tracking viscosity on particles) and use the old interface for tracking the elastic stresses as it does not need the `MaterialModel Outputs`?

PR #3484, opened 2020-04-25
www: update readme

tjhei:


tjhei:
hopefully, that makes the convoluted process a bit more transparent @gassmoeller

PR #3487, opened 2020-04-27
reduce gmg storage

tcclevenger:
Reduces the storage requirement for the `block GMG` Stokes solver. Now, only 1 active level coefficient table needs to by stored instead of 3 (was 1 for each StokesOperator, ABlockOperator, and MassMatrixOperator), and only 1 multievel set of coefficient tables needs to be stored instead of 2 (was 1 for each ABlock GMG and SchurComplement GMG).

This is not too consequential for the current setup (cellwise constant viscosity), but will be necessary for the initial version of DGQ1 viscosity. The pull request for the DG1 projected viscosity will be able to be submitted once this change is incorporated.

### DIFF
--- a/include/aspect/stokes_matrix_free.h
+++ b/include/aspect/stokes_matrix_free.h
@@ -75,17 +75,9 @@ namespace aspect
          * Fills in the viscosity table, sets the value for the pressure scaling constant,
          * and gives information regarding compressibility.
          */
-        void fill_cell_data(const dealii::LinearAlgebra::distributed::Vector<number> &viscosity_values,
-                            const double pressure_scaling,
-                            const Triangulation<dim> &tria,
-                            const DoFHandler<dim> &dof_handler_for_projection,
-                            const bool is_compressible);
-
-        /**
-         * Returns the viscosity table.
-         */
-        const Table<1, VectorizedArray<number> > &
-        get_viscosity_x_2_table();
+        void fill_cell_data (const Table<1, VectorizedArray<number>> &viscosity_table,
+                             const double pressure_scaling,
+                             const bool is_compressible);
 
         /**
          * Computes the diagonal of the matrix. Since matrix-free operators have not access
@@ -114,7 +106,7 @@ namespace aspect
         /**
          * Table which stores a viscosity value for each cell.
          */
-        Table<1, VectorizedArray<number> > viscosity_x_2;
+        const Table<1, VectorizedArray<number>> *viscosity;
 
         /**
          * Pressure scaling constant.
@@ -151,10 +143,7 @@ namespace aspect
          * @p is_mg_level_data describes whether the viscosity values are defined for a multigrid level
          * matrix or for the active level matrix.
          */
-        void fill_cell_data (const dealii::LinearAlgebra::distributed::Vector<number> &viscosity_values,
-                             const Triangulation<dim> &tria,
-                             const DoFHandler<dim> &dof_handler_for_projection,
-                             const bool is_mg_level_data,
+        void fill_cell_data (const Table<1, VectorizedArray<number>> &viscosity_table,
                              const double pressure_scaling);
 
 
@@ -194,7 +183,7 @@ namespace aspect
         /**
          * Table which stores a viscosity value for each cell.
          */
-        Table<1, VectorizedArray<number> > one_over_viscosity;
+        const Table<1, VectorizedArray<number>> *viscosity;
 
         /**
          * Pressure scaling constant.
@@ -227,11 +216,8 @@ namespace aspect
          * @p is_mg_level_data describes whether the viscosity values are defined for a multigrid level
          * matrix or for the active level matrix.
          */
-        void fill_cell_data(const dealii::LinearAlgebra::distributed::Vector<number> &viscosity_values,
-                            const Triangulation<dim> &tria,
-                            const DoFHandler<dim> &dof_handler_for_projection,
-                            const bool is_mg_level_data,
-                            const bool is_compressible);
+        void fill_cell_data (const Table<1, VectorizedArray<number>> &viscosity_table,
+                             const bool is_compressible);
 
         /**
          * Computes the diagonal of the matrix. Since matrix-free operators have not access
@@ -275,7 +261,7 @@ namespace aspect
         /**
          * Table which stores a viscosity value for each cell.
          */
-        Table<1, VectorizedArray<number> > viscosity_x_2;
+        const Table<1, VectorizedArray<number>> *viscosity;
 
         /**
           * Information on the compressibility of the flow.
@@ -419,10 +405,16 @@ namespace aspect
       DoFHandler<dim> dof_handler_p;
       DoFHandler<dim> dof_handler_projection;
 
-      FESystem<dim> stokes_fe;
       FESystem<dim> fe_v;
       FESystem<dim> fe_p;
       FESystem<dim> fe_projection;
+
+      Table<1, VectorizedArray<double>> active_viscosity_table;
+      MGLevelObject<Table<1, VectorizedArray<double>>> level_viscosity_tables;
+
+      // This variable is needed only in the setup in both evaluate_material_model()
+      // and build_preconditioner(). It will be deleted after the last use.
+      MGLevelObject<dealii::LinearAlgebra::distributed::Vector<double> > level_viscosity_vector;
 
       typedef MatrixFreeStokesOperators::StokesOperator<dim,velocity_degree,double> StokesMatrixType;
       typedef MatrixFreeStokesOperators::MassMatrixOperator<dim,velocity_degree-1,double> SchurComplementMatrixType;
@@ -434,7 +426,6 @@ namespace aspect
 
       ConstraintMatrix constraints_v;
       ConstraintMatrix constraints_p;
-      ConstraintMatrix constraints_projection;
 
       MGLevelObject<ABlockMatrixType> mg_matrices_A_block;
       MGLevelObject<SchurComplementMatrixType> mg_matrices_Schur_complement;
@@ -442,9 +433,6 @@ namespace aspect
       MGConstrainedDoFs mg_constrained_dofs_A_block;
       MGConstrainedDoFs mg_constrained_dofs_Schur_complement;
       MGConstrainedDoFs mg_constrained_dofs_projection;
-
-      dealii::LinearAlgebra::distributed::Vector<double> active_coef_dof_vec;
-      MGLevelObject<dealii::LinearAlgebra::distributed::Vector<double> > level_coef_dof_vec;
 
       MGTransferMatrixFree<dim,double> mg_transfer_A_block;
       MGTransferMatrixFree<dim,double> mg_transfer_Schur_complement;
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -797,11 +797,13 @@ namespace aspect
 
     for (unsigned int cell=cell_range.first; cell<cell_range.second; ++cell)
       {
+        VectorizedArray<number> one_over_cell_viscosity = (*viscosity)(cell);
+
         // The /= operator for VectorizedArray results in a foating point operation
         // (divide by 0) since the (*viscosity)(cell) array is not completely filled.
         // Therefore, we need to divide each entry manually.
-        VectorizedArray<number> one_over_cell_viscosity = (*viscosity)(cell);
-        for (unsigned int c=0; c<this->get_matrix_free()->n_components_filled(cell); ++c)
+        const unsigned int n_components_filled = this->get_matrix_free()->n_components_filled(cell);
+        for (unsigned int c=0; c<n_components_filled; ++c)
           one_over_cell_viscosity[c] = pressure_scaling*pressure_scaling/one_over_cell_viscosity[c];
 
         pressure.reinit (cell);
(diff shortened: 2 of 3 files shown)


include/aspect/stokes_matrix_free.h
```diff
@@ -419,7 +405,6 @@ namespace aspect
       DoFHandler<dim> dof_handler_p;
       DoFHandler<dim> dof_handler_projection;
 
-      FESystem<dim> stokes_fe;
```
tcclevenger, line 422: Unused variable from an older version of the solver.

include/aspect/stokes_matrix_free.h
```diff
@@ -434,7 +419,6 @@ namespace aspect
 
       ConstraintMatrix constraints_v;
       ConstraintMatrix constraints_p;
-      ConstraintMatrix constraints_projection;
```
tcclevenger, line 437: Unused variable from an older version of the solver.

include/aspect/stokes_matrix_free.h
```diff
@@ -75,17 +75,9 @@ namespace aspect
...
-                            const Triangulation<dim> &tria,
-                            const DoFHandler<dim> &dof_handler_for_projection,
-                            const bool is_compressible);
-
-        /**
-         * Returns the viscosity table.
-         */
-        const Table<1, VectorizedArray<number> > &
-        get_viscosity_x_2_table();
+        void fill_cell_data (const Table<1,VectorizedArray<number>> *viscosity_table,
```
tjhei, line 78: why don't you pass this in as a reference?

include/aspect/stokes_matrix_free.h
```diff
@@ -151,10 +143,7 @@ namespace aspect
          * @p is_mg_level_data describes whether the viscosity values are defined for a multigrid level
          * matrix or for the active level matrix.
          */
-        void fill_cell_data (const dealii::LinearAlgebra::distributed::Vector<number> &viscosity_values,
-                             const Triangulation<dim> &tria,
-                             const DoFHandler<dim> &dof_handler_for_projection,
-                             const bool is_mg_level_data,
+        void fill_cell_data (const Table<1,VectorizedArray<number>> *viscosity_table,
```
tjhei, line 146: same here

source/simulator/stokes_matrix_free.cc
```diff
@@ -857,13 +795,18 @@ namespace aspect
 
     for (unsigned int cell=cell_range.first; cell<cell_range.second; ++cell)
       {
-        const VectorizedArray<number> &cell_one_over_viscosity = one_over_viscosity(cell);
+        // The /= operator for VectorizedArray results in a foating point operation
+        // (divide by 0) since the (*viscosity)(cell) array is not completely filled.
+        // Therefore, we need to divide each entry manually.
+        VectorizedArray<number> &one_over_cell_viscosity = (*viscosity)(cell);
```
tjhei, line 801: you are taking a reference to the vectorized array, doesn't this modify the content in there?

source/simulator/stokes_matrix_free.cc
```diff
@@ -926,7 +869,12 @@ namespace aspect
     FEEvaluation<dim,degree_p,degree_p+2,1,number> pressure (data, 0);
     for (unsigned int cell=cell_range.first; cell<cell_range.second; ++cell)
       {
-        const VectorizedArray<number> &cell_one_over_viscosity = one_over_viscosity(cell);
+        // The /= operator for VectorizedArray results in a foating point operation
+        // (divide by 0) since the (*viscosity)(cell) array is not completely filled.
+        // Therefore, we need to divide each entry manually.
+        VectorizedArray<number> &one_over_cell_viscosity = (*viscosity)(cell);
```
tjhei, line 875: same here

include/aspect/stokes_matrix_free.h
```diff
@@ -443,11 +427,12 @@ namespace aspect
...
 
-      dealii::LinearAlgebra::distributed::Vector<double> active_coef_dof_vec;
-      MGLevelObject<dealii::LinearAlgebra::distributed::Vector<double> > level_coef_dof_vec;
-
       MGTransferMatrixFree<dim,double> mg_transfer_A_block;
       MGTransferMatrixFree<dim,double> mg_transfer_Schur_complement;
+
+      Table<1, VectorizedArray<double>> active_viscosity_table;
+      MGLevelObject<Table<1, VectorizedArray<double>>> level_viscosity_tables;
+      MGLevelObject<dealii::LinearAlgebra::distributed::Vector<double> > level_coef_dof_vec;
```
tjhei, line 435: Do you need to keep both (table and vector) around? Maybe leave a comment here why?

tcclevenger:
> Looks reasonable to me! Do you report memory consumption of all the matrix-free objects? This would be great to have.

Yes, for that I had to implement a lot of new SimulatorAccess functions and functions into the StokesMatrixFree and StokesMatrixFreeImplementation and so I am making that a separate pull request (once this gets merged).

include/aspect/stokes_matrix_free.h
tcclevenger: Yes, I need to keep the level viscosities around because I use them in both the `evaluate_material_model()` (storing the matrix-free table) and `build_preconditioner()` (when I have to compute the diagonal for the shell meshes with tangential boundary) functions.

Another option is to recompute the active and level viscosities again at the beginning of `build_preconditioner()`. We would save memory at the cost of an extra loop over cells (to project viscosity to DGQ0 or DGQ1) and an MGTransfer. 

I'll add a comment now, but if you don't think the extra computation is too expensive, I think the other way is a ok too.

source/simulator/stokes_matrix_free.cc
tcclevenger: You're right, it should. Not sure why this wasn't messing up the tests.. Maybe I was lucky with the order of operations.. I've changed this.

include/aspect/stokes_matrix_free.h
tjhei: I let you decide what makes more sense. Keeping the vector around during setup and then deleting it is probably a decent approach?

tjhei:
```
#0  /home/dealii/dealii-v9.0.1/lib/libdeal_II.g.so.9.0.1: dealii::Subscriptor::check_no_subscribers() const

#1  /home/dealii/dealii-v9.0.1/lib/libdeal_II.g.so.9.0.1: dealii::Subscriptor::~Subscriptor()

#2  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: dealii::Table<1, dealii::VectorizedArray<double> >::~Table()

#3  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: std::_Sp_counted_ptr_inplace<dealii::Table<1, dealii::VectorizedArray<double> >, std::allocator<dealii::Table<1, dealii::VectorizedArray<double> > >, (__gnu_cxx::_Lock_policy)2>::_M_dispose()

#4  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release()

#5  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: dealii::MGLevelObject<dealii::Table<1, dealii::VectorizedArray<double> > >::resize(unsigned int, unsigned int)

#6  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: aspect::StokesMatrixFreeHandlerImplementation<2, 2>::evaluate_material_model()
```

tcclevenger:
> ```
> #0  /home/dealii/dealii-v9.0.1/lib/libdeal_II.g.so.9.0.1: dealii::Subscriptor::check_no_subscribers() const
> 
> #1  /home/dealii/dealii-v9.0.1/lib/libdeal_II.g.so.9.0.1: dealii::Subscriptor::~Subscriptor()
> 
> #2  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: dealii::Table<1, dealii::VectorizedArray<double> >::~Table()
> 
> #3  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: std::_Sp_counted_ptr_inplace<dealii::Table<1, dealii::VectorizedArray<double> >, std::allocator<dealii::Table<1, dealii::VectorizedArray<double> > >, (__gnu_cxx::_Lock_policy)2>::_M_dispose()
> 
> #4  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release()
> 
> #5  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: dealii::MGLevelObject<dealii::Table<1, dealii::VectorizedArray<double> > >::resize(unsigned int, unsigned int)
> 
> #6  /jenkins/workspace/aspect_PR-3487/build-gcc-fast/aspect: aspect::StokesMatrixFreeHandlerImplementation<2, 2>::evaluate_material_model()
> ```

Yeah, its complaining that I am trying to resize the MGLevelObjects<Table> variable but it still has a subscriptor. It happens during iterated Stokes (but not when running on the workstation with newest dealii). Trying to fix now.

include/aspect/stokes_matrix_free.h
tcclevenger: Ok, we'll keep it as is.

source/simulator/stokes_matrix_free.cc
```diff
@@ -914,7 +912,6 @@ namespace aspect
   void
   MatrixFreeStokesOperators::ABlockOperator<dim,degree_v,number>::clear ()
   {
-    viscosity.clear();
```
tjhei: does it make sense to do ``viscosity = nullptr;`` to get rid of the reference?

tcclevenger:
I can trigger the error now on the workstation so I'll commit when I've figured it out.

source/simulator/stokes_matrix_free.cc
```diff
@@ -938,7 +888,7 @@ namespace aspect
 
             pressure.evaluate (true,false,false);
             for (unsigned int q=0; q<pressure.n_q_points; ++q)
-              pressure.submit_value(cell_one_over_viscosity*pressure_scaling*pressure_scaling*
+              pressure.submit_value(one_over_cell_viscosity*pressure_scaling*pressure_scaling*
```
tjhei, line 891: would it make sense to move the ``*pressure_scaling*pressure_scaling`` into the ``VectorizedArray``?

tcclevenger: Yes, it would be cleaner.

tcclevenger:
I have a solution using `std::shared_ptr`. @tjhei can you take a look? It may be overkill. 

The failing test is the nonlinear example (a few values are sightly off with changing the pressure scaling). I should be able to fix it easily.

tjhei:
> I have a solution using `std::shared_ptr`. @tjhei can you take a look? It may be overkill.

yeah, I think this is more complicated than it needs to be. I will take a look.

tjhei:
What do you think of https://github.com/tjhei/aspect/commits/pr3487 ? I think we can also go back to a ``SmartPointer`` with the order of the members switched, but this seems good enough?

tcclevenger:
Yeah, I think it's definitely good enough. Makes it much easier to look at. (At east I learned a lot about pointers along the way :) ).

So the failed test (I think) comes from a slight difference in the floating point operations when pressure_scaling is moved out of the submit_value call. I've moved it back in to test this. It makes sense to me since the only test that is effected is the non-linear one which is more sensitive to small changes (and the difference in the test is very minimal). Should I just update the test output?

tjhei:
> Should I just update the test output?

yes.

tcclevenger:
Ok, changed the test and pressure_scaling. Do you want to merge your changes?

tjhei:
> Ok, changed the test and pressure_scaling.

It looks like you still need to fix nonlinear_channel_flow_velocities_Newton_Stokes_GMG:
https://jenkins.tjhei.info/blue/organizations/jenkins/aspect/detail/PR-3487/13/artifacts

> Do you want to merge your changes?

just cherry-pick my commit and push.

> Then does it make sense to squash the comment

Yes, please squash everything.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1345,59 +1262,88 @@ namespace aspect
...
+    }
+
+    // Create active viscosity table
+    {
+      const unsigned int n_cells = stokes_matrix.get_matrix_free()->n_macro_cells();
+      active_viscosity_table.reinit(TableIndices<1>(n_cells));
+
+      std::vector<types::global_dof_index> local_dof_indices(dof_handler_projection.get_fe().dofs_per_cell);
+      for (unsigned int cell=0; cell<n_cells; ++cell)
+        for (unsigned int i=0; i<stokes_matrix.get_matrix_free()->n_components_filled(cell); ++i)
```
gassmoeller, line 1277: and here

source/simulator/stokes_matrix_free.cc
```diff
@@ -1345,59 +1262,88 @@ namespace aspect
...
-                                                           /*is_mg_level_data*/true,
-                                                           sim.pressure_scaling);
+        // Create level viscosity table
+        {
+          const unsigned int n_cells = mg_matrices_A_block[level].get_matrix_free()->n_macro_cells();
+          level_viscosity_tables[level].reinit(TableIndices<1>(n_cells));
+
+          std::vector<types::global_dof_index> local_dof_indices(dof_handler_projection.get_fe().dofs_per_cell);
+          for (unsigned int cell=0; cell<n_cells; ++cell)
+            for (unsigned int i=0; i<mg_matrices_A_block[level].get_matrix_free()->n_components_filled(cell); ++i)
```
gassmoeller, line 1328: and here

source/simulator/stokes_matrix_free.cc
```diff
@@ -857,13 +797,18 @@ namespace aspect
 
     for (unsigned int cell=cell_range.first; cell<cell_range.second; ++cell)
       {
-        const VectorizedArray<number> &cell_one_over_viscosity = one_over_viscosity(cell);
+        // The /= operator for VectorizedArray results in a foating point operation
+        // (divide by 0) since the (*viscosity)(cell) array is not completely filled.
+        // Therefore, we need to divide each entry manually.
+        VectorizedArray<number> one_over_cell_viscosity = (*viscosity)(cell);
+        for (unsigned int c=0; c<this->get_matrix_free()->n_components_filled(cell); ++c)
```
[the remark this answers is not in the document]
tcclevenger: The function should be cheap, but it is called for every cell, for every vmult. I will just change it to be safe.

tjhei:
excellent, thank you!

tjhei:
I will go ahead and merge this as you addressed my and @gassmoeller s comments.

PR #3488, opened 2020-04-29
separate two merged boxes from box geometry model

MFraters:
Separate two merged boxes from box geometry model and define a independent height_above_reference_surface function.

I found that the world builder didn't work for a model anymore where I did work before, and it turned out that due to pull request #3378, now the wrong height above the surface was passed (always 0) because the two merged boxes relied on the function from the box model, which relied on a point variable which was never filled in the two merged boxes model. To prevent these kind of problems in the future I separated them.  

I have not added a test yet, but will do that later.

### DIFF
--- a/source/geometry_model/initial_topography_model/function.cc
+++ b/source/geometry_model/initial_topography_model/function.cc
@@ -47,7 +47,7 @@ namespace aspect
     value (const Point<dim-1> &surface_point) const
     {
       Point<dim> global_point;
-      if (Plugins::plugin_type_matches<GeometryModel::Box<dim> >(this->get_geometry_model())||
+      if (Plugins::plugin_type_matches<GeometryModel::Box<dim> >(this->get_geometry_model()) ||
           Plugins::plugin_type_matches<const GeometryModel::TwoMergedBoxes<dim>> (this->get_geometry_model()))
         {
           // No need to set the vertical coordinate correctly,
--- a/source/geometry_model/initial_topography_model/function.cc
+++ b/source/geometry_model/initial_topography_model/function.cc
@@ -22,6 +22,7 @@
 #include <aspect/geometry_model/initial_topography_model/function.h>
 #include <aspect/geometry_model/interface.h>
 #include <aspect/geometry_model/box.h>
+#include <aspect/geometry_model/two_merged_boxes.h>
 #include <aspect/geometry_model/sphere.h>
 #include <aspect/geometry_model/spherical_shell.h>
 #include <aspect/geometry_model/chunk.h>
@@ -46,7 +47,8 @@ namespace aspect
     value (const Point<dim-1> &surface_point) const
     {
       Point<dim> global_point;
-      if (Plugins::plugin_type_matches<GeometryModel::Box<dim> >(this->get_geometry_model()))
+      if (Plugins::plugin_type_matches<GeometryModel::Box<dim> >(this->get_geometry_model())||
+          Plugins::plugin_type_matches<const GeometryModel::TwoMergedBoxes<dim>> (this->get_geometry_model()))
         {
           // No need to set the vertical coordinate correctly,
           // because it will be thrown away in get_data_component anyway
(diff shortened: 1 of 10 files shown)


anne-glerum:
Originally TwoMergedBoxes was derived from Box to avoid having to test both for Box and for TwoMergedBoxes when checking whether a model is spherical or a box, it saves a conditional statement. Is there a way to keep this?

MFraters:
That depends on what exactly you want to test. If you just want to know what the coordinate system is, you can just use the function `natural_coordinate_system`, but if you want to know whether a specific function is available, you will need to test separately. I think this is better because I also noticed that this way initial topography was allowed, but the depth function did not account for it in the same way the box function does.

gassmoeller:
> it saves a conditional statement. Is there a way to keep this?

I am not sure we should try to keep this class hierarchy only for that reason. I had a quick look through the files that check for `GeometryModel::Box` and some already accurately check for `Box || TwoMergedBoxes`. Others only check for `Box`, but also work for `TwoMergedBoxes`. However, there are also some that just check for `Box`, but would clearly not work or crash for `TwoMergedBoxes`, e.g. boundary_temperature/box.cc and boundary_composition/box.cc. So having TwoMergedBoxes derived from Box not only makes it vulnerable to changes in Box, it also allows for misunderstandings in other plugins (which think they deal with a Box, when in reality they deal with TwoMergedBoxes)

I think there are two and a half reasonable ways forward:
1. Keeping TwoMergedBoxes derived from Box. This would mean only fixing the topography problems Menno found, and going through all plugins that currently only check for Box, but do not work for TwoMergedBoxes and fixing them to work for TwoMergedBoxes as well.
2. Going forward with this PR. This case would mean going through all checks for `GeometryModel::Box` (~15) and either changing them to `Box || TwoMergedBoxes`, or if they do not work for TwoMergedBoxes fix them or just leave them as is (=disabling these plugins for TwoMergedBoxes).
(3.) If we want to keep the ability to use box functions for both models and check for both models at the same time, we can introduce an abstract `BoxBase` class, and derive both Box and TwoMergedBoxes from that. It is not clear to me if this is worth the effort and the added complication, but maybe there are places where this would help?

What do you two think?

anne-glerum:
> I had a quick look through the files that check for GeometryModel::Box and some already accurately check for Box || TwoMergedBoxes. Others only check for Box, but also work for TwoMergedBoxes. However, there are also some that just check for Box, but would clearly not work or crash for TwoMergedBoxes, e.g. boundary_temperature/box.cc and boundary_composition/box.cc. 

Yes let's unify that. Perhaps method 2 is the most clear for future plugins that test for geometry plugins, as it forces you to explicitly think about whether it would work for TwoMergedBoxes or not.

tjhei:
> I am not sure we should try to keep this class hierarchy only for that reason.

agreed. I would go forward with this PR.

MFraters:
Oke, I will update the pull request and go though the different checks to update them

MFraters:
I updated the GeometryModel::Box where I though it was correct. 

The only place I am doubting about is in the plugin's named box (e.g. initial temperature/box.cc). There is no correctness reason why it should not be possible, and it might be convenient to have it in the same place, but it is not a box geometry model any more. I am fine either way.

I didn't update the shearbands benchmark because I don't know enough about it and it is not one of the main plugins.

source/geometry_model/initial_topography_model/function.cc
```diff
@@ -46,7 +47,8 @@ namespace aspect
     value (const Point<dim-1> &surface_point) const
     {
       Point<dim> global_point;
-      if (Plugins::plugin_type_matches<GeometryModel::Box<dim> >(this->get_geometry_model()))
+      if (Plugins::plugin_type_matches<GeometryModel::Box<dim> >(this->get_geometry_model())||
```
gassmoeller, line 50:
```suggestion
      if (Plugins::plugin_type_matches<GeometryModel::Box<dim> >(this->get_geometry_model()) ||
```

source/geometry_model/two_merged_boxes.cc
```diff
@@ -34,6 +34,7 @@ namespace aspect
   namespace GeometryModel
   {
 
+
```
gassmoeller, line 37:
```suggestion
```

PR #3489, opened 2020-04-30
Fix visco plastic finite difference

MFraters:
in pull request #3460 a discussion came up about the 0.5 in the finite difference terms. This discussion was concluded that the 0.5 should be there. During that discussion I looked at the visco plastic material model for reference, and noticed that it wasn't there while it should be. 

I have done some testing to convince myself that this is actually better, and that was harder then expected. For the simple "compex" cases it converges very vast anyway or not at all anyway. The nonlinear channel flow was actually a nice in between case, where I could study the actual impact of adding the 0.5. The conclusion is that sometimes it is better, sometimes it is worse, but without the 0.5 the SPDness is lost. Without the 0.5 every now then the assert

```
The violated condition was: 
    data.local_matrix.matrix_norm_square(tmp)/(tmp*tmp) >= -1e-12*data.local_matrix.frobenius_norm()
Additional information: 
    The top left block of the local Newton-Stokes matrix is not positive definite but has an eigenvalue less than -2.0913049911280974e+21 < 0. This should not happen.

```
was triggers (not always, because the assert generates a few random tensors to test the SPDness). This never happened to me when the 0.5 was there. This convince me that the 0.5 should be there. 

I added that test and update the relevant tests.

### DIFF
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -349,6 +349,9 @@ namespace aspect
             {
               const TableIndices<2> strain_rate_indices = SymmetricTensor<2,dim>::unrolled_to_component_indices (component);
 
+               // components that are not on the diagonal are multiplied by 0.5, because the symmetric tensor
+               // is modified by 0.5 in both symmetric directions (xy/yx) simultaneously and we compute the combined
+               // derivative
               const SymmetricTensor<2,dim> strain_rate_difference = in.strain_rate[i]
                                                                     + std::max(std::fabs(in.strain_rate[i][strain_rate_indices]), min_strain_rate)
                                                                     * (component > dim-1 ? 0.5 : 1 )
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -349,9 +349,9 @@ namespace aspect
             {
               const TableIndices<2> strain_rate_indices = SymmetricTensor<2,dim>::unrolled_to_component_indices (component);
 
-               // components that are not on the diagonal are multiplied by 0.5, because the symmetric tensor
-               // is modified by 0.5 in both symmetric directions (xy/yx) simultaneously and we compute the combined
-               // derivative
+              // components that are not on the diagonal are multiplied by 0.5, because the symmetric tensor
+              // is modified by 0.5 in both symmetric directions (xy/yx) simultaneously and we compute the combined
+              // derivative
               const SymmetricTensor<2,dim> strain_rate_difference = in.strain_rate[i]
                                                                     + std::max(std::fabs(in.strain_rate[i][strain_rate_indices]), min_strain_rate)
                                                                     * (component > dim-1 ? 0.5 : 1 )
--- a/source/material_model/visco_plastic.cc
+++ b/source/material_model/visco_plastic.cc
@@ -351,6 +351,7 @@ namespace aspect
 
               const SymmetricTensor<2,dim> strain_rate_difference = in.strain_rate[i]
                                                                     + std::max(std::fabs(in.strain_rate[i][strain_rate_indices]), min_strain_rate)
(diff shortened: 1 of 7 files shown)


source/material_model/visco_plastic.cc
```diff
@@ -351,6 +351,7 @@ namespace aspect
 
               const SymmetricTensor<2,dim> strain_rate_difference = in.strain_rate[i]
                                                                     + std::max(std::fabs(in.strain_rate[i][strain_rate_indices]), min_strain_rate)
+                                                                    * (component > dim-1 ? 0.5 : 1 )
                                                                     * finite_difference_accuracy
                                                                     * Utilities::nth_basis_for_symmetric_tensors<dim>(component);
```
gassmoeller, line 356:
```suggestion
               // components that are not on the diagonal are multiplied by 0.5, because the symmetric tensor
               // is modified by 0.5 in both symmetric directions (xy/yx) simultaneously and we compute the combined
               // derivative
              const SymmetricTensor<2,dim> strain_rate_difference = in.strain_rate[i]
                                                                    + std::max(std::fabs(in.strain_rate[i][strain_rate_indices]), min_strain_rate)
                                                                    * (component > dim-1 ? 0.5 : 1 )
                                                                    * finite_difference_accuracy
                                                                    * Utilities::nth_basis_for_symmetric_tensors<dim>(component);
```

tests/visco_plastic_derivatives_2d.cc
```diff
@@ -173,6 +173,7 @@ void f(const aspect::SimulatorAccess<dim> &simulator_access,
         {
           in_dviscositydstrainrate.strain_rate[i] = in_base.strain_rate[i]
                                                     + std::fabs(in_base.strain_rate[i][strain_rate_indices])
+                                                    * (component > dim-1 ? 0.5 : 1 )
                                                     * finite_difference_accuracy
                                                     * aspect::Utilities::nth_basis_for_symmetric_tensors<dim>(component);
```
gassmoeller, line 178:
```suggestion
          // components that are not on the diagonal are multiplied by 0.5, because the symmetric tensor
          // is modified by 0.5 in both symmetric directions (xy/yx) simultaneously and we compute the combined
          // derivative
          in_dviscositydstrainrate.strain_rate[i] = in_base.strain_rate[i]
                                                    + std::fabs(in_base.strain_rate[i][strain_rate_indices])
                                                    * (component > dim-1 ? 0.5 : 1 )
                                                    * finite_difference_accuracy
                                                    * aspect::Utilities::nth_basis_for_symmetric_tensors<dim>(component);
```

tests/visco_plastic_derivatives_3d.cc
```diff
@@ -188,6 +188,7 @@ void f(const aspect::SimulatorAccess<dim> &simulator_access,
         {
           in_dviscositydstrainrate.strain_rate[i] = in_base.strain_rate[i]
                                                     + std::fabs(in_base.strain_rate[i][strain_rate_indices])
+                                                    * (component > dim-1 ? 0.5 : 1 )
                                                     * finite_difference_accuracy
                                                     * aspect::Utilities::nth_basis_for_symmetric_tensors<dim>(component);
```
gassmoeller, line 193:
```suggestion
          // components that are not on the diagonal are multiplied by 0.5, because the symmetric tensor
          // is modified by 0.5 in both symmetric directions (xy/yx) simultaneously and we compute the combined
          // derivative
          in_dviscositydstrainrate.strain_rate[i] = in_base.strain_rate[i]
                                                    + std::fabs(in_base.strain_rate[i][strain_rate_indices])
                                                    * (component > dim-1 ? 0.5 : 1 )
                                                    * finite_difference_accuracy
                                                    * aspect::Utilities::nth_basis_for_symmetric_tensors<dim>(component);
```

MFraters:
Thanks for the review, those are good suggestions! I have applied and indented them.

PR #3490, opened 2020-05-01
add gmg objects to memory stats

tcclevenger:
This adds gmg objects to the postprocess/memory_statistics.cc. A few things:

1. This added a lot of functions. I may be making this more difficult. 

- `simulator_access.h`: For each object I had to add a `get_*()` function.

- `stokes_matrix_free.h`: For each object, for both classes `StokesMatrixFreeHandler` and `StokesMatrixFreeHandlerImplementation`, I had to add the same `get_*()` function. This is because the `stokes_matrix_free` member object in `simulator_access.h` cannot access the member variables (it is of BaseClass type, not ImplementationClass.) In the `stokes_matrix_free.cc`, we only need one actual implementation for each `get_*()`.

2. I chose to have only one `DoFHandler` output in `memory_statistics.cc`. I added the memory consumption from the 3 matrix-free DoFHandlers to the system DoFHandler instead of having separate entries. Same with `ConstraintMatrix` and `MGTransfer` objects.

3. A question: Should we include the `system_rhs` or any of the timestep vectors to this output? If so, I can add the diagonal vectors for the matrixfree operators all into a category called "Vector memory consumption (MB)", instead of having "Solution vector memory consumption (MB)".

Let me know thoughts. There may be an implementation that requires less code.


include/aspect/simulator_access.h
```diff
@@ -901,6 +902,74 @@ namespace aspect
       Particle::World<dim> &
       get_particle_world();
 
+      /**
+       *  Return true if using the block GMG Stokes solver.
+       */
+      bool is_matrix_free();
```
tjhei, line 908: we currently do not solve temperature equations matrix-free but we might at some point, so this should contain ``stokes`` somewhere in the name.

tcclevenger:
> I agree that this is a lot of exposed functionality that would otherwise not be needed. One alternative would be to just create one or two `memory_consumption()` functions (or a function that returns a `std::map<name,memory>`). Not sure if this is better than the current approach though. I am fine with keeping it as is.

It is ok with me then. I may try and clean it up later.

tcclevenger:
I am going to run into the issue that I've changed "current_constraints memory consumption (MB)" to "ConstraintMatrix memory consumption (MB)". So virtually every test will fail.

Should I change all the tests to have the new output text? Or just change the code back to "current_constraints"? I think I prefer this new name, but I'm curious what others think.

tcclevenger:
> I am going to run into the issue that I've changed "current_constraints memory consumption (MB)" to "ConstraintMatrix memory consumption (MB)". So virtually every test will fail.
> 
Never mind, I did not realize that only 4 tests output memory statistics. I will just change those tests.

include/aspect/stokes_matrix_free.h
```diff
@@ -392,6 +455,69 @@ namespace aspect
       static
       void declare_parameters (ParameterHandler &prm);
 
+      /**
+       * Return a reference to the DoFHandler that is used for velocity in
+       * the block GMG solver.
+       */
+      const DoFHandler<dim> &
+      get_dof_handler_v () const override;
```
gassmoeller, line 463: I am late, but in the current form you add a lot of very specific functions to SimulatorAccess. For all other cases we usually expose the owning object through `SimulatorAccess` (e.g. `SimulatorAccess::get_newton_handler`) and let  the returned object decide what to expose. Could we do the same here: `SimulatorAccess::get_stokes_matrix_free_handler`? It saves us some functions in SimulatorAccess and reduces the number of places we would have to change if something changes about the implementation.

tcclevenger: Yes, I see that now. I was able to remove all other matrix-free functions in SimulatorAccess with the exception of `is_stokes_matrix_free()`.

include/aspect/simulator_access.h
```diff
@@ -34,6 +34,7 @@
 #include <deal.II/dofs/dof_handler.h>
 #include <deal.II/fe/fe.h>
 #include <deal.II/fe/mapping_q.h>
+#include <deal.II/multigrid/mg_transfer_matrix_free.h>
```
tjhei, line 37: is this needed?

tcclevenger: No, just left over from an older version. I'll remove it.

include/aspect/simulator_access.h
```diff
@@ -901,6 +902,19 @@ namespace aspect
       Particle::World<dim> &
       get_particle_world();
 
+
```
tjhei, line 905: one empty line only

source/postprocess/memory_statistics.cc
```diff
@@ -37,13 +37,39 @@ namespace aspect
...
-      statistics.add_value ("DoFHandler memory consumption (MB) ", this->get_dof_handler().memory_consumption()/mb);
-      statistics.add_value ("current_constraints memory consumption (MB) ", this->get_current_constraints().memory_consumption()/mb);
+
+      double dof_handler_mem = this->get_dof_handler().memory_consumption();
+      double constraints_mem = this->get_current_constraints().memory_consumption();
+      if (this->is_stokes_matrix_free())
+        {
+          dof_handler_mem += this->get_stokes_matrix_free().get_dof_handler_v().memory_consumption()
+                             + this->get_stokes_matrix_free().get_dof_handler_p().memory_consumption();
+#if DEAL_II_VERSION_GTE(9,2,0)
```
tjhei, line 51: what is the reason for this ifdef?

tcclevenger: `memory_consumption()` had not yet been implemented for DG DoFHandlers on the version of dealii the tester is run on.

tjhei: huh, interesting. Can you add this as a comment here, then?

PR #3491, opened 2020-05-01
Maintenance May 2020

tjhei:


tjhei:
> It looks like you deleted doc/manual/plugin_graph.pdf was that intended? Is that not needed for the manual?

huh, there must have been an error that I did not catch.

tjhei:
This looks like a bug on github. The file still exists and looks good, but is a couple of bytes smaller (after all, the viscoplastic box is deleted).
If you click on view file you will get
https://github.com/geodynamics/aspect/blob/5c85005efdb6e2452411d81d729b517b1fcdbeec/doc/manual/plugin_graph.pdf
which looks reasonable.

gassmoeller:
Thanks, the changed files tab looked like the file would be deleted. Good to go then.

PR #3492, opened 2020-05-01
Move dynamic_friction functionality to an existing rheology module

naliboff:
This PR modifies the `drucker_prager` rheology module to include a function and input parameters for calculating the dynamic coefficient of friction. The option to use this functionality is added to the `visco_plastic` material model.

The benefits of this PR is that it would allow removing the `dynamic_friction` material model and allow further testing of rate-state friction type plastic rheology models in the `visco_plastic` material model.

The downside to this PR is that the test case only contained the screen output and we do not have a benchmark for the material models. Perhaps it would be worth expanding the `dynamic_friction` test case first, but then again it is a relatively straightforward change.


include/aspect/material_model/rheology/drucker_prager.h
```diff
@@ -79,6 +84,17 @@ namespace aspect
           parse_parameters (const unsigned int n_fields,
                             ParameterHandler &prm);
 
+          /**
+           * Compute the steady-state angle of internal friction using the standard (static)
+           * angle of internal friction, dynamic angle of internal friction, effective strain
+           * rate, and reference strain rate.
+           */
+          double
+          compute_steady_state_friction (const double angle_internal_friction,
```
gassmoeller, line 93: As far as I can see this function is not really related to the drucker_prager module? All input parameters are provided from the outside. Would it make sense to move it into a separate module called `steady_state_friction_angle`? Or is the computation inside only valid for drucker prager plasticity?
In any case I would suggest renaming to `compute_steady_state_friction_angle` for clarity.

source/material_model/rheology/drucker_prager.cc
```diff
@@ -29,6 +29,29 @@ namespace aspect
...
+      template <int dim>
+      double
+      DruckerPrager<dim>::compute_steady_state_friction (const double angle_internal_friction,
+                                                         const double angle_internal_friction_dynamic,
+                                                         const double effective_strain_rate,
+                                                         const double reference_strain_rate) const
+      {
+
+        // Calculate effective steady-state friction coefficient (mu). The formula below is equivalent to the
+        // equation 13 in van Dinther et al., (2013, JGR). Although here the dynamic friction coefficient
```
gassmoeller, line 41: Please copy the sentence about the publication into the header as well. That is something that is interesting even for readers who are not interested in how it is implemented exactly (=for ASPECT), but only want to understand what this function does in general.

source/material_model/visco_plastic.cc
```diff
@@ -707,6 +716,12 @@ namespace aspect
           // Drucker Prager plasticity parameters
           Rheology::DruckerPrager<dim>::declare_parameters(prm);
 
+          //
```
gassmoeller, line 719: Remove this comment line.

source/material_model/visco_plastic.cc
```diff
@@ -240,9 +240,18 @@ namespace aspect
 
           // Step 3b: calculate weakened friction, cohesion, and pre-yield viscosity
           const double current_cohesion = drucker_prager_parameters.cohesions[j] * weakening_factors[0];
-          const double current_friction = drucker_prager_parameters.angles_internal_friction[j] * weakening_factors[1];
+          double current_friction = drucker_prager_parameters.angles_internal_friction[j] * weakening_factors[1];
           viscosity_pre_yield *= weakening_factors[2];
 
+          // Step 3d: calculate steady-state coefficient of internal friction
+          if (use_steady_state_friction_angle == true)
```
gassmoeller, line 247: What happens if you remove this if-loop completely? As far as I can see the return value of `compute_steady_state_friction` for `drucker_prager_parameters.angles_internal_friction_dynamic[j] == 0` should be exactly `current_friction` and all tests should continue to work. If there are problems due to floating-point inaccuracies, we could introduce a bypass into `compute_steady_state_friction` (e.g. `if (angle_internal_friction_dynamic == 0.0) return angle_internal_friction;`.
This would allow you to remove the additional variable for visco-plastic (which is in essence just a short-hand for `angle_internal_friction_dynamic != 0.0` anyway).

naliboff:
@anne-glerum @EstherHeck - Is this PR worth continuing? If you have prefer a different methodology for moving the dynamic friction to a rheology module or would like to submit a new PR, just let me know :)

naliboff:
Closing this as it will be superseded by work from @EstherHeck and @anne-glerum.

PR #3112, opened 2019-05-31
Create dc picard solver schemes

MFraters:
Depended on #3109. I added one example of what this could look like. I would like to get some opinions on whether this is the correct approach and right naming

PR #3493, opened 2020-05-02
add more tests for reference and start optimizing.

MFraters:
This pull request is to see if the Newton/defect correction solver can be optimized in some ways.  Although there was a discussion in #3112, with my recent tests (with deal.ii master) it already seems faster when doing the same test, but I think there still is room for improvment. This is a work in progress.

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -2419,8 +2419,10 @@ namespace aspect
                           current_linearization_point);
 
     // rebuild the whole system to compute the rhs.
-    rebuild_stokes_matrix = assemble_newton_stokes_system = assemble_newton_stokes_matrix = true;
+    assemble_newton_stokes_system = true;
     rebuild_stokes_preconditioner = false;
+    rebuild_stokes_matrix = boundary_velocity_manager.get_active_boundary_velocity_conditions().size()!=0;
+    assemble_newton_stokes_matrix = boundary_velocity_manager.get_active_boundary_velocity_conditions().size()!=0;
 
     compute_current_constraints ();
 
--- a/tests/nonlinear_channel_flow_tractions_Newton_Stokes_no_line_search.prm
+++ b/tests/nonlinear_channel_flow_tractions_Newton_Stokes_no_line_search.prm
@@ -1,0 +1,7 @@
+include $ASPECT_SOURCE_DIR/tests/nonlinear_channel_flow_tractions_Newton_Stokes.prm
+
+subsection Solver parameters
+  subsection Newton solver parameters
+    set Max Newton line search iterations = 0
+  end
+end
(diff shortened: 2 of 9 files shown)


MFraters:
This pull request is ready for review. I set out to optimize a part of the code I had seen being very slow due to it using a function in deal.ii which made it slow. I wanted to optimize by making sure that part of the code was called as few times as possible. When I tried it again with the newest deal.ii I found that it wasn't slow any more, but I did find an other part of the code which could be optimized, which is done here.

I might try again in the future to find exactly why it was slow back then, but I think this pull request is valuable on it's own to be in the next release.

source/simulator/helper_functions.cc
```diff
@@ -2419,8 +2419,9 @@ namespace aspect
                           current_linearization_point);
 
     // rebuild the whole system to compute the rhs.
-    rebuild_stokes_matrix = assemble_newton_stokes_system = assemble_newton_stokes_matrix = true;
+    rebuild_stokes_matrix = assemble_newton_stokes_system = true;
     rebuild_stokes_preconditioner = false;
+    rebuild_stokes_matrix = assemble_newton_stokes_matrix = boundary_velocity_manager.get_active_boundary_velocity_conditions().size()!=0;
```
gassmoeller, line 2424: Could you avoid these double assignments, in particular because the variables are named so similar (the current form is hard to read). Would the following work?
```
rebuild_stokes_preconditioner = false;
rebuild_stokes_matrix         = boundary_velocity_manager.get_active_boundary_velocity_conditions().size() != 0;
assemble_newton_stokes_matrix = boundary_velocity_manager.get_active_boundary_velocity_conditions().size() != 0;
```

tests/nonlinear_channel_flow_tractions_Newton_Stokes_no_line_search.prm
```diff
@@ -0,0 +1,148 @@
...
+set Maximum time step                      = 1
+set End time                               = 4
+set Start time                             = 0
+set Adiabatic surface temperature          = 0
+set Surface pressure                       = 0
+set Use years in output instead of seconds = false  # default: true
+set Nonlinear solver scheme = iterated Advection and Newton Stokes
+set Max nonlinear iterations = 45 
+set Nonlinear solver tolerance = 2.5e-10
+set Additional shared libraries = ../benchmarks/nonlinear_channel_flow/libsimple_nonlinear.so
```
gassmoeller, line 16: Remove this line, it is confusing (and not used) in a test. If this is only a slightly modified copy of a benchmark or another test, consider using a `include $ASPECT_SOURCE_DIR/tests/visualization_strain_rate_tensor.prm` statement, as long as you only need to add a few lines afterwards.

tests/nonlinear_channel_flow_velocities_Newton_Stokes_no_line_search.prm
```diff
@@ -0,0 +1,145 @@
...
+set Maximum time step                      = 1
+set End time                               = 4
+set Start time                             = 0
+set Adiabatic surface temperature          = 0
+set Surface pressure                       = 0
+set Use years in output instead of seconds = false  # default: true
+set Nonlinear solver scheme = iterated Advection and Newton Stokes
+set Max nonlinear iterations = 30
+set Nonlinear solver tolerance = 1e-14
+set Additional shared libraries = ../benchmarks/nonlinear_channel_flow/libsimple_nonlinear.so
```
gassmoeller, line 16: same comment as for the other test

MFraters:
Thanks for the review! I have addressed the comments.

PR #3495, opened 2020-05-06
add Q1 viscosity projection for GMG

tcclevenger:
It's finally here. This pull request adds the ability to use "project to Q1" averaging for GMG. It also adds a type of averaging called "project to Q1 only viscosity" which keeps all other averaging to "none" (except viscosity). I've tested convergence rates with solcx, solkx, hollow_sphere, and burstedde benchmarks and we match the convergence rates for both the velocity and pressure of the the AMG solve with no averaging. (Let me know and I can send anyone a spreadsheet with all the values).

For this code, we needed a cellwise Q1 projection. @tjhei noticed that in source/simulator/initial_conditions.cc we had a similar projection for discontinuous pressure. I abstracted that function and moved it to utilities.h/.cc. and call it in both initial_conditions.cc (for pressure) and stokes_matrix_free.cc (for viscosity projection). 

I've also added a single test which uses this projection for the hollow_sphere benchmark.

As far as storage is concerned, we now need to store a viscosity value per cell, per quadrature point per level, which is much more expensive than just per cell (as with cellwise averages). It was hypothesized that this would be quite a burden for Q1 viscosity, but actually the storage is about equivalent to a single Stokes vector and therefore not overly expensive. We have the following memory (in MB) for a hollow_sphere run:
mem
Table DG0 is the viscosity storage for cellwise constant averaging. Table DG1 is the viscosity storage for project to Q1 viscosity. I also output the system matrix for the AMG solver for comparison. And for the Tables, I give the values for Q3/Q2 Stokes element.


include/aspect/utilities.h
```diff
@@ -1366,6 +1366,36 @@ namespace aspect
...
+
+    /**
+     * Compute the cellwise projection of @p function to the finite element space
+     * described by @p dof_handler.
+     *
+     * @param[in] mapping The mapping object to use.
+     * @param[in] dof_handler The DoFHandler the describes the finite element space to
+     * project into and that corresponds to @p vec_result.
+     * @param[in] component_index The component index of the @p dof_handler for which
+     * the projection is being performed. This component should be desribed by
```
tjhei, line 1379:
```suggestion
     * the projection is being performed. This component should be described by
```

include/aspect/utilities.h
```diff
@@ -1366,6 +1366,36 @@ namespace aspect
...
+     * @param[in] function The function to project into the finite element space.
+     * This function should store the value of the function at the points described
+     * by the 2nd argument into the 3rd argument as an std::vector<double>
+     * of size quadrature.size().
+     * @param[out] vec_result The output vector where the projected function will be
+     * stored in.
+     */
+    template <int dim, typename VectorType>
+    void
+    project_cellwise(const Mapping<dim>                                        &mapping,
```
tjhei, line 1390: how difficult would it be to break this part into a separate PR?

source/simulator/initial_conditions.cc
```diff
@@ -375,84 +375,30 @@ namespace aspect
...
-        Vector<double> cell_vector (dofs_per_cell);
-        Vector<double> local_projection (dofs_per_cell);
-        FullMatrix<double> local_mass_matrix (dofs_per_cell, dofs_per_cell);
-
-        std::vector<double> rhs_values(n_q_points);
-
-        ScalarFunctionFromFunctionObject<dim>
-        adiabatic_pressure (
-          [&](const Point<dim> &p) -> double
+        Utilities::project_cellwise<dim,LinearAlgebra::BlockVector>(*mapping,
```
tjhei, line 389: very nice!

source/simulator/initial_conditions.cc
```diff
@@ -375,84 +375,30 @@ namespace aspect
...
-              local_mass_matrix.vmult (local_projection, cell_vector);
-
-              // then set the global solution vector to the values just computed
-              cell->set_dof_values (local_projection, system_tmp);
-            }
+          for (unsigned int i=0; i<values.size(); ++i)
+            values[i] = adiabatic_conditions->pressure(q_points[i]);
+          return;
+        },
+        system_tmp);
```
tjhei, line 401: why is this not ``system_tmp.block(...)``?

source/simulator/stokes_matrix_free.cc
```diff
@@ -723,20 +721,26 @@ namespace aspect
 
         for (unsigned int q=0; q<velocity.n_q_points; ++q)
           {
+            VectorizedArray<number> viscosity_x_2;
```
tjhei, line 724: Does it make sense to move this variable outside of this loop and initialize with ``(cell, 0)``?  Then you only need the ``else`` here.

source/simulator/stokes_matrix_free.cc
```diff
@@ -797,21 +801,27 @@ namespace aspect
...
-          one_over_cell_viscosity[c] = pressure_scaling*pressure_scaling/one_over_cell_viscosity[c];
-
         pressure.reinit (cell);
         pressure.read_dof_values(src);
         pressure.evaluate (true, false);
         for (unsigned int q=0; q<pressure.n_q_points; ++q)
-          pressure.submit_value(one_over_cell_viscosity*
-                                pressure.get_value(q),q);
+          {
+            VectorizedArray<number> one_over_viscosity;
```
tjhei, line 809: same here

source/simulator/stokes_matrix_free.cc
```diff
@@ -892,8 +893,23 @@ namespace aspect
 
             pressure.evaluate (true,false,false);
             for (unsigned int q=0; q<pressure.n_q_points; ++q)
-              pressure.submit_value(one_over_cell_viscosity*
-                                    pressure.get_value(q),q);
+              {
+                VectorizedArray<number> one_over_viscosity;
```
tjhei, line 897: same here

source/simulator/stokes_matrix_free.cc
```diff
@@ -944,16 +960,20 @@ namespace aspect
...
     for (unsigned int cell=cell_range.first; cell<cell_range.second; ++cell)
       {
-        const VectorizedArray<number> cell_viscosity_x_2 = 2.0*(*viscosity)(cell);
-
         velocity.reinit (cell);
         velocity.read_dof_values(src);
         velocity.evaluate (false, true, false);
         for (unsigned int q=0; q<velocity.n_q_points; ++q)
           {
+            VectorizedArray<number> viscosity_x_2;
```
tjhei, line 968: same here

source/simulator/stokes_matrix_free.cc
```diff
@@ -1028,10 +1046,16 @@ namespace aspect
             velocity.evaluate (false,true,false);
             for (unsigned int q=0; q<velocity.n_q_points; ++q)
               {
+                VectorizedArray<number> viscosity_x_2;
```
tjhei, line 1049: and here. :-)

source/simulator/stokes_matrix_free.cc
```diff
@@ -1396,20 +1501,26 @@ namespace aspect
 
         for (unsigned int q=0; q<velocity.n_q_points; ++q)
           {
+            VectorizedArray<double> viscosity_x_2;
```
tjhei, line 1504: same here

source/utilities.cc
```diff
@@ -3410,6 +3410,79 @@ namespace aspect
...
+                    cell_vector(i) +=
+                      rhs_values[point] *
+                      fe_values[extractor].value(i,point) *
+                      fe_values.JxW(point);
+
+                  for (unsigned int j=0; j<dofs_per_cell; ++j)
+                    if ((dof_handler.get_fe().system_to_component_index(i).first ==
+                         component_index)
+                        &&
+
```
tjhei, line 3468: remove empty line

source/utilities.cc
```diff
@@ -3410,6 +3410,79 @@ namespace aspect
...
+                        (dof_handler.get_fe().system_to_component_index(j).first ==
+                         component_index))
+                      local_mass_matrix(j,i) += (fe_values[extractor].value(i,point) * fe_values[extractor].value(j,point) *
+                                                 fe_values.JxW(point));
+                    else if (i == j)
+                      local_mass_matrix(i,j) = 1;
```
tjhei, line 3474: ``1.``

source/simulator/initial_conditions.cc
tcclevenger: The input Vector  needs to match the DoFHandler for `set_dof_value()` to correctly fill the vector.

source/simulator/stokes_matrix_free.cc
tcclevenger: Yes, that definitely makes sense.

tcclevenger:
> excellent! Overall, this looks very good. I left a bunch of comments.
> Can you check if it is easy to split this into 3 PRs:
> 
> * new averaging scheme
> * introduce project()
> * the new GMG
>   ?

Yes, I will go ahead and create the project_cellwise and the averaging scheme pull request. Then I'll do the GMG projection once the project_cellwise is complete?

tjhei:
> Then I'll do the GMG projection once the project_cellwise is complete?

you can just rebase this PR after we merge the other two (which should be quick and easy).

tcclevenger:
@tjhei and @gassmoeller I am creating the pull request for `project_to_q1_only_viscosity` and I found interesting behavior. 

It looks like the GMG project to Q1 method (computed in stokes_matrix_free.cc `evaluate_material_model()`) is more accurate than the current approach in material_model/interface.cc (using `compute_projection_matrix()` and `average()`). 

Here is a demonstration using solkz with the default parameters in benchmarks/solkz/solkz.prm. I calculate the `u_l2` error for a few global refinements. The rate should be 8, but using AMG with project to Q1 only viscosity does not get this. Here

errors

I previously thought that the approaches were identical, however, there are 2 differences that I see, both related to forcing the values to be between a min and max:

1. In stokes_matrix_free.cc, per cell
project quad values to Q1 -> expand back to quad values -> force values to be in the [min,max]
In material_model/interface.cc, per cell
project quad values to Q1 -> force values to be in the [min,max] -> expand back to quad values.
Changing this alone (to the stokes_matrix_free.cc way) does not give the correct error rates for AMG, but does decrease the error slightly. 

2. In stokes_matrix_free.cc, I find a global min/max for the viscosity, but in material_model/interface.cc it is a cellwise min/max. I haven't tested this one yet.

I will start to investigate again on Tuesday, namely if changing both 1. and 2. gives us the correct rate. Or perhaps there is yet another difference.

tjhei:
If you force the Q1 values in the support points to be within [a,b], then the expanded quadrature points will also be between [a,b], so that should not make a difference.

solkz has viscosity ``exp(c y)``. I would think that the optimal fit gives support points outside the min,max range of the quadrature points, so we will always clamp the values which destroys the convergence order.
Using a global min/max makes sense more sense but might be more expensive and makes the operation no longer a local one. Hmmm.

tcclevenger:
Ok, so here are some new results. Before doing any assembly, I calculated the global min/max viscosity values and past them to the `MaterialModel::average()` function each time it is called. This complexity is just a loop over cells, then a loop over quadrature, once before assembly; the `MaterialModel::average()` is still local (it just has knowledge of the global values). Here are the rates for solkz. Both correct using a global min/max, but differ when they correct.

Correcting the values in Q1 (before expanding to quadrature)
error-before

Correcting the values in quadrature 
error-after

So, it looks like it does matter when we do the correction. This makes sense to me since the Q1 projection naturally contains over and under shoots and that is accounted for in the approximation quality (2nd order for Q1), and my guess is that we destroy that when we change the values in Q1. My guess too is that we are affecting more quadrature points second hand by changing in Q1 than we do by just correcting in quadrature. I also imagine changing the quadrature points could affect the rates for some examples, but that is just a price of this method. 

Here is the commit with my changes: https://github.com/tcclevenger/aspect/commit/f31a7c1ae6f513aefc06e87008aca2f7949cd854
I would need to find a cleaner implementation, but if you are curious it's there. 

Opinions for proceeding: 
1. In GMG I create a viscosity DGQ1 vector using the project_cellwise() function. Would it make sense to do the same as a shared variable (viscosity_q1_vector) in simulator.h? 
2. Or, like I am doing, does it makes sense to just calculate the min/max values and pass to material_model (along with others: density, specific_heat, etc.)? 
3. Or should GMG just be slightly more accurate since it is more necessary to have a Q1 projection than AMG?

Open to suggestions!

tjhei:
I would suggest we proceed with the better method for GMG separately for now and think about the MaterialAveraging separately.

One problem is that we use the averaging in many places in the code and we can not expect to know global min/max values at all times. Adiabatic conditions, visualization, etc.. come to mind.

tcclevenger:
All other pull requests have been merged and now this one is ready for review.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1130,7 +1166,14 @@ namespace aspect
 
       fe_v (FE_Q<dim>(sim.parameters.stokes_velocity_degree), dim),
       fe_p (FE_Q<dim>(sim.parameters.stokes_velocity_degree-1),1),
-      fe_projection(FE_DGQ<dim>(0),1)
+      fe_projection(FE_DGQ<dim>(sim.parameters.material_averaging
```
tjhei, line 1169: can you add a comment here to explain this logic?

source/simulator/stokes_matrix_free.cc
```diff
@@ -2226,10 +2353,25 @@ namespace aspect
...
+                    {
+                      fe_values_projection.reinit(DG_cell);
+                      fe_values_projection.get_function_values(level_viscosity_vector[level],
+                                                               dg_dof_indices,
+                                                               visc_on_quad);
+                    }
 
                   for (unsigned int q=0; q<n_q_points; ++q)
                     {
+                      double viscosity = 1.0;
```
tjhei, line 2369: use ``const double viscosity = (bla ) ? bla : bla``

source/simulator/stokes_matrix_free.cc
```diff
@@ -1325,30 +1417,57 @@ namespace aspect
...
+                    fe_values_projection.reinit(DG_cell);
+                    fe_values_projection.get_function_values(level_viscosity_vector[level],
+                                                             local_dof_indices,
+                                                             values_on_quad);
+
+                    for (unsigned int q=0; q<n_q_points; ++q)
+                      {
+                        // Do not allow viscosity to be greater than or less than the limits
+                        // of the evaluated viscosity on the active level.
+                        if (values_on_quad[q] > max_el)
```
tjhei, line 1461: can you use ``std::min(std::max(x,min),max)`` ?

tcclevenger:
Currently, we are correcting min/max values at quadrature level to keep the correct rates. Should there be a non-negativity check at support points as well?

tjhei:
> Currently, we are correcting min/max values at quadrature level to keep the correct rates.

Can you summarize your findings and our conversation today? I don't think we understand why limiting in support points is incorrect or do I misremember?

tcclevenger:
Yes, I'll summarize: 

**Using project to Q1, in material_model/interface.cc for active viscosity**
This is what is used to define viscosity for AMG computations. It has the following steps:
Given any cell:
- project to Q1 element on cell,
- force Q1 values to be within local cellwise `[min, max]`,
- then project back to quadrature values.  

**GMG project to Q1 for active viscosity**
In `evaluate_material_model()` in stokes_matrix_free.cc, I create the viscosity vector in the following steps:
- For each cell, project from quadrature to Q1 element and update global min/max viscosity values.
- For each cell, project back to quadrature values and then force values (at quadrature level) to by within the global  `[min, max]`.
- Note: Level viscosity irrelevant for solution accuracy

Both methods compute _local_ projections, but the one for GMG does bookkeeping to find global min and max values. 

**Differences**
1. AMG enforces min/max bound at Q1 level, GMG enforces at quadrature level
2. AMG uses cellwise min/max, GMG uses global min/max

**Test**
Using SolKz benchmark, I'll show the l2 error for velocity and convergence rate. I'll consider 3 things: AMG as is, GMG as is (in this pull request), and GMG but enforcing the min/max bound at Q1 (i.e. 1. matches AMG). I'll use the "project to Q1 only viscosity" for all runs.

errors

**Conclusion**
Enforcing the bound in support points reduces the convergence rate. Also, while I don't have the computation here, I can also show that, for AMG, enforcing the bound in quadrature but still using a local cellwise min/max also has a reduces convergence rate. Therefore, to get the correct rates we need differences 1. and 2. for GMG.

I believe @bangerth explanation was that enforcing the bound in the support points could create the situation where the accuracy of the Q1 projection degrades to O(h) instead of O(h^3). That makes sense to me; what doesn't make sense, however, is that changing the quadrature doesn't also reduce convergence order.

gassmoeller:
Oh and can you take a look at what is happening with the tests?

tcclevenger:
@gassmoeller I made the requested changes and fixed the new Q1 test.

The other GMG test that is breaking is the one which uses the Picard iteration (and does not rely on this new code). The output (nonlinear and Stokes iterations) are very different, but it appears that we are getting to the same solution. Has code relating to Picard changed in between GMG pull requests?

gassmoeller:
Thanks for the changes.

No there was no change to the Picard iterations, also the tester runs against a merged version of your PR with master, so any change in master should already be considered. Is it possible that with your new implementation the behavior has changed if no Q1 projection is done? Previously we would do a cell-wise average (harmonic average for the failing test), and now it seems you project to a DGQ0 space without averaging. You probably need to include a call to `MaterialModel::MaterialAveraging::average` inside the function for `project_cellwise`, but only if `dof_handler_projection.get_fe().degree == 0` or something like that.

tjhei:
1. You will need to update hollow_sphere_gmg_project_q1
2. nonlinear_channel_flow_velocities_Newton_Stokes_GMG still gives slightly different results, but it looks much closer. I am okay with applying the changes.

tcclevenger:
> 1. You will need to update hollow_sphere_gmg_project_q1
> 2. nonlinear_channel_flow_velocities_Newton_Stokes_GMG still gives slightly different results, but it looks much closer. I am okay with applying the changes.

I actually think I've just ruined the convergence order.. I know the solution though! I'll let y'all know when I get this fixed.

gassmoeller:
Did you have time to look at the issue? I think it should be sufficient to put `if (dof_handler_projection.get_fe().degree == 0)` before the call to `average`, because otherwise you call the broken Q1 projection operation inside the GMG projection function.

tjhei:
@tcclevenger is/was offline for a couple days. He will get to it next week, I would think.

gassmoeller:
Ah sure, no rush. I was just posting because his last post sounded like he was getting to it right away and I worried there were complications.

tcclevenger:
Ok, all the tests now pass (I had to slightly alter the nonlinear test, but that is to be expected since I slightly change the viscosity calculation).

Let me know if you guys have any other question or suggestions.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1222,12 +1290,21 @@ namespace aspect
...
                                                                                sim.triangulation.get_communicator());
-    {
-      const QGauss<dim> quadrature_formula (sim.parameters.stokes_velocity_degree+1);
 
+    const QGauss<dim> quadrature_formula (sim.parameters.stokes_velocity_degree+1);
+    FEValues<dim> fe_values_projection (*(sim.mapping),
+                                        fe_projection,
+                                        quadrature_formula,
+                                        update_values);
+    std::vector<double> values_on_quad;
```
tjhei, line 1301: can you move this down to where it is used?

source/simulator/stokes_matrix_free.cc
```diff
@@ -1325,30 +1440,51 @@ namespace aspect
 
     for (unsigned int level=0; level<n_levels; ++level)
       {
-        // Create level viscosity table
-        {
-          const unsigned int n_cells = mg_matrices_A_block[level].get_matrix_free()->n_macro_cells();
-          level_viscosity_tables[level].reinit(TableIndices<1>(n_cells));
+        // Create multilevel viscosity tables. For DGQ0, this is one value per cell,
+        // for DGQ1 this is n_q_points values per cell.      {
```
tjhei, line 1444: what is this opening ``{``?

tests/hollow_sphere_gmg_project_q1.prm
```diff
@@ -0,0 +1,31 @@
...
+
+subsection Material model
+  set Model name = HollowSphereMaterial
+  set Material averaging = project to Q1 only viscosity
+end
+
+
+subsection Postprocess
+  set List of postprocessors = HollowSpherePostprocessor
+end
```
tjhei, line 31: missing newline

tjhei:
Can we get a changelog entry as well, please?

source/simulator/stokes_matrix_free.cc
tcclevenger: This is left over from before I removed excess brackets. Just a typo.

source/simulator/stokes_matrix_free.cc
tcclevenger: I end up needing these for the multilevel computations as well, which is why I need to define them here. I can move the `values_on_quad` vector and just redefine for multilevel. Should I do the same with QGauss and FEValues?

tjhei: try to move things down where they are used (these are only needed after the project* call, right?).

tjhei:
What is the status here, @tcclevenger ? Are you planning on moving code around still?

source/simulator/stokes_matrix_free.cc
tcclevenger: I end up using `fe_values_projection` in lines 1403 and 1479 and similarly for `quadrature_formula` and `values_on_quad`. I can redefine them in the various different blocks if that looks nicer?

tjhei: I was just confused that the projection right below these lines does not use it. If you think this is still sensible this way, we can keep it as is.

tcclevenger: Oh, I didn't realize it wasn't used anymore.. Yes, you are right. I've moved two of the variables down to the next block.

tcclevenger:
> What is the status here, @tcclevenger ? Are you planning on moving code around still?

If all tests pass, it will be good to merge

tcclevenger:
All tests pass. Let me know if I should rebase again.

gassmoeller:
Same here. :tada:

tcclevenger:
Thanks! Glad to finally get this in! @tjhei @gassmoeller

PR #3497, opened 2020-05-07
Added stabilization interface for advection assemblers

gassmoeller:
This is an interface that allows assemblers for the advection equation to scale the stabilization terms of the entropy viscosity method. The current implementation keeps all current functionality, however it looks a bit awkward for the following reason: I think it makes sense to eventually use this for entropy viscosity and SUPG, however the two methods use different algorithms to determine the scaling of the equations:
SUPG computes the cell-wise maximum of the point-wise quantity rho\*cp
EV computes the cell-wise maximum of rho and cp and then compute rho_max \* cp_max.
EV will always compute a bigger quantity than SUPG, although SUPG is likely closer to the real maximum of rho\*cp in the cell.
Should I:
1. Keep the interface as I implemented here, returning a vector of values? 
2. Simplify the interface to only return a single cell-wise value?
In both cases: Should the default implementation follow the current algorithm for EV or SUPG?

I tend towards 1. and using the implementation of SUPG, but that potentially changes a lot of tests. It would slightly decrease the stabilization parameter for EV.

### DIFF
--- a/include/aspect/melt.h
+++ b/include/aspect/melt.h
@@ -261,7 +261,7 @@ namespace aspect
      * the case of melt migration on a single cell.
      */
     template <int dim>
-    class MeltAdvectionSystem : public MeltInterface<dim>
+    class MeltAdvectionSystem : public MeltInterface<dim>, public Assemblers::AdvectionStabilizationInterface<dim>
     {
       public:
         void
--- a/include/aspect/simulator/assemblers/advection.h
+++ b/include/aspect/simulator/assemblers/advection.h
@@ -34,7 +34,7 @@ namespace aspect
      * equation for the current cell.
      */
     template <int dim>
-    class AdvectionSystem : public Assemblers::Interface<dim>,
+    class AdvectionSystem : public Assemblers::Interface<dim>, public Assemblers::AdvectionStabilizationInterface<dim>,
       public SimulatorAccess<dim>
     {
       public:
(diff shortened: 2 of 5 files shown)


tjhei:
> SUPG computes the cell-wise maximum of the point-wise quantity rho*cp
> EV computes the cell-wise maximum of rho and cp and then compute rho_max * cp_max.

huh, interesting. I would think that only the product rho*cp appears, so EV seems to be incorrect.

> I tend towards 1. and using the implementation of SUPG, but that potentially changes a lot of tests. It would slightly decrease the stabilization parameter for EV.

This might require tuning the stabilization parameters for EV again... :-(

gassmoeller:
I have updated the implementation to use the cell wise max of rho*cp. Only 8 tests are affected and most of them just in the number of iterations, or properties that are close to zero. The differences are likely small, because the variations in rho and cp within a cell are usually small (at most a few percent).  This will also only affect the maximum viscosity, not the residual based viscosity. I feel relatively confident that we do not need to recalibrate the parameter. What do you think?

tjhei:
> What do you think?

I agree that we should make that change.

gassmoeller:
I updated the documentation. No further changes from my side.

gassmoeller:
Lets postpone it. I think it is progress, but it changes behaviour in subtle and not completely predictable ways. I also just realized there is at least one problem I need to address: If there is more than one advection assembler (never happening at the moment), which assembler determines the prefactor scaling? We should probably make sure only one assembler is derived from the interface, or we need to check inside the assemblers if the assembler is relevant for the current advection field.

PR #3498, opened 2020-05-07
add project_cellwise function

tcclevenger:
From https://github.com/geodynamics/aspect/pull/3495. This `project_cellwise` function computes a DGQ1 projection of a set of input values given in the std::function into a given vector. 

We will use this in the Q1 viscosity for GMG, and it is used once now in the code (DGQ1 pressure). I did not write a test, since every test which has `set Use locally conservative discretization = true` will be reliant on this code and will pass now if this is implemented correctly.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1366,6 +1366,36 @@ namespace aspect
          */
         std::array<double,dim> coordinates;
     };
+
+
+    /**
+     * Compute the cellwise projection of @p function to the finite element space
+     * described by @p dof_handler.
+     *
+     * @param[in] mapping The mapping object to use.
+     * @param[in] dof_handler The DoFHandler the describes the finite element space to
+     * project into and that corresponds to @p vec_result.
+     * @param[in] component_index The component index of the @p dof_handler for which
+     * the projection is being performed. This component should be described by
+     * a DG finite element.
+     * @param[in] function The function to project into the finite element space.
+     * This function should store the value of the function at the points described
+     * by the 2nd argument into the 3rd argument as an std::vector<double>
+     * of size quadrature.size().
+     * @param[out] vec_result The output vector where the projected function will be
+     * stored in.
+     */
+    template <int dim, typename VectorType>
+    void
+    project_cellwise(const Mapping<dim>                                        &mapping,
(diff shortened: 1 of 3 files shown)


include/aspect/utilities.h
```diff
@@ -1366,6 +1366,36 @@ namespace aspect
          */
         std::array<double,dim> coordinates;
     };
+
+
+    /**
+     * Compute the cellwise projection of @p function to the finite element space
```
tjhei, line 1372:
```suggestion
     * Compute the cellwise projection of component @p component_index of @p function to the finite element space
```

tjhei:
I am okay with this part but I would like to hear from @bangerth if this is also worth putting into deal.II in this shape.

tjhei:
@bangerth do you have time to take a look? Otherwise I would think we should proceed.

tcclevenger:
Thanks @gassmoeller! 

Can you send a link to an intro particle paper so that I can get an overview? I'm currently not that familiar, but I imagine that a particle version of this function would not be too difficult.

tjhei:
> I'm currently not that familiar, but I imagine that a particle version of this function would not be too difficult.

The fact that it is particles doesn't really matter here. Imagine the user supplies a set of points with coordinates and the function values in these points on each cell (instead of point values at given quadrature points). The number of these points can vary as well.

@gassmoeller It might be easiest and fastest to go from particles to quadrature points and then call this function. Otherwise you might have an underdetermined system?

gassmoeller:
> Otherwise you might have an underdetermined system?

This will be the main problem. "Going" from particles to quadrature points will create the same challenge unfortunately. For starters would it be enough to ensure there are at least n_dofs particles per cell? They could of course be unfortunately positioned, but assuming they are randomly distributed would a sufficient number guarantee that the system is not underdetermined?

tjhei:
> For starters would it be enough to ensure there are at least n_dofs particles per cell? They could of course be unfortunately positioned, but assuming they are randomly distributed would a sufficient number guarantee that the system is not underdetermined?

I am not quite sure what a reasonable requirement for sensible results is. We normally have way more than n_dof quadrature points for example. You can also pose a least square problem on each cell like we already do for particle interpolation (right?).
What do you want to use this method for?

PR #3504, opened 2020-05-19
Allow disable EV for compositions at boundaries

gassmoeller:
Another follow-up optimization related to #3499. We currently disable EV stabilization at Dirichlet boundaries for temperature, but when solving an entropy equation we want to do the same. In this PR I add a new input parameter that allows to specify the names of compositional fields for which EV should be disabled at Dirichlet boundaries (using the same procedure as for temperature).

The changes are actually pretty small if you hide whitespace changes in the github settings (a lot of code is now differently indented).


tjhei:
Could this influence accuracy for Zhong spherical benchmarks?

PR #3507, opened 2020-05-22
Add info unit boundary velocity

anne-glerum:
I added a sentence to the description of the Boundary function mesh deformation plugin to explain that the velocity should always be given in m/s, even if Use years in output instead of seconds = true. I didn't see a quick fix as the function is directly given to the interpolate_boundary_values function. 

### For all pull requests:


gassmoeller:
Thanks for the added documentation. I think the way to fix this would be to wrap the stored function inside another function object (not a parsed function) that scales it correctly to years or seconds, similar to what the `VectorFunctionFromVelocityFunctionObject` inside core.cc does. Could you open an issue about this that reminds us of the missing conversion?

anne-glerum:
Yep, opened an issue.

PR #3509, opened 2020-05-22
Bypass topography in coordinate conversion

anne-glerum:
This PR avoids getting stuck into a loop when reading ascii data, which requires converting points to natural coordinates, which requires knowing the topography which is being read from the ascii data. Same change as was made to the chunk geometry. 

Actually using AsciiDataBoundary for the EllipsoidalChunk will require some changes to utilities.cc

### For all pull requests:


anne-glerum:
ah sorry, it builds now.

PR #3510, opened 2020-05-22
add IDR(s) Stokes Kylov solver

tcclevenger:
This pull request adds the IDR(s) solver discussed in Chapter 5 of my [PhD thesis](https://tigerprints.clemson.edu/all_dissertations/2523/) and section 3.4 of the [GMG Stokes paper](https://arxiv.org/pdf/1907.06696.pdf) @tjhei and I wrote (under review). Here is a summary chart from the preprint for IDR(2) compared with GMRES, as well as a scaling plot of IDR(2) (I think these are the largest runs done with ASPECT, even if they aren't that interesting scientifically).

compare-idr

scale

Summary: Around the same cost as GMRES, but much less memory required (roughly 1/3 in the example above). There is no convergence guarantee like GMRES, but various papers suggest the behavior is similar to GMRES as `s` is increased.

### DIFF
--- a/doc/modules/changes/20200525_clevenger
+++ b/doc/modules/changes/20200525_clevenger
@@ -1,0 +1,6 @@
+New: The IDR(s) Krylov method can now be used for solving the Stokes
+system, only when using the matrix-free GMG solver. The parameter
+s can be chosen by the user. This method has a short term recurrence
+and can greatly reduce the memory requirement as compared to GMRES.
+<br>
+(Thomas C. Clevenger, 2020/05/25)
--- a/source/simulator/parameters.cc
+++ b/source/simulator/parameters.cc
@@ -357,6 +357,28 @@ namespace aspect
                            "complement solver is used. The direct solver is only efficient "
                            "for small problems.");
 
+        prm.declare_entry ("Krylov method for cheap solver steps", "GMRES",
+                           Patterns::Selection(StokesKrylovType::pattern()),
+                           "This is the Krylov method used to solve the Stokes system. Both options, GMRES "
+                           "and IDR(s), solve non-symmetric, indefinite systems. GMRES "
+                           "guarantees the residual will be reduced in each iteration while IDR(s) has "
+                           "no such property. On the other hand, the vector storage requirement for GMRES is "
+                           "dependent on the restart length and can be quite restrictive (since, for "
+                           "the matrix-free GMG solver, memory is dominated by these vectors) whereas "
+                           "IDR(s) has a short term recurrence. "
+                           "Note that the IDR(s) Krylov method is not available for the AMG solver since "
+                           "it is not a flexible method, i.e., it cannot handle a preconditioner which "
+                           "may change in each iteration (the AMG-based preconditioner contains a CG solve "
+                           "in the pressure space which may have different number of iterations each step).");
+
+        prm.declare_entry ("IDR(s) parameter", "2",
+                           Patterns::Integer(1),
+                           "This is the sole parameter for the IDR(s) Krylov solver and will dictate the "
+                           "number of matrix-vector products and preconditioner applications per iteration (s+1) "
+                           "and the total number of temporary vectors required (5+3*s). For s=1, this method is "
+                           "analogous to BiCGStab. As s is increased this method is expected to converge to "
+                           "GMRES in terms of matrix-vector/preconditioner applications to solution.");
+
         prm.declare_entry ("Linear solver tolerance", "1e-7",
                            Patterns::Double(0., 1.),
                            "A relative tolerance up to which the linear Stokes systems in each "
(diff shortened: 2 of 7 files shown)


source/simulator/parameters.cc
```diff
@@ -451,6 +451,27 @@ namespace aspect
                            "\\cite{KHB12}.");
       }
       prm.leave_subsection ();
+      prm.enter_subsection ("GMG parameters");
```
tjhei, line 454: Would it make sense to put this setting outside of GMG because it is also feasible to implement IDR if we are using the matrix-based solver? This doesn't need to be implemented right now of course.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1751,17 +1752,39 @@ namespace aspect
...
+                          rhs_copy,
+                          preconditioner_cheap);
+          }
+        else if (sim.parameters.stokes_krylov_type == Parameters<dim>::StokesKrylovType::idr_s)
+          {
+#if !DEAL_II_VERSION_GTE(9,2,0)
+            Assert(false, ExcNotImplemented());
+#endif
+
+            SolverIDR<dealii::LinearAlgebra::distributed::BlockVector<double> >
```
tjhei, line 1776: shouldn't this be inside ``#else``?

tjhei:
Would it make sense to use IDR as a default? Or are we not sure it is always the better choice?

source/simulator/parameters.cc
gassmoeller: I agree, this should be in the general Stokes solver section, even if at the moment it only works for GMG.

tcclevenger: Should I add an Assert if this is called without GMG? This issue with not using this for AMG is that preconditioner requires a flexible Krylov method.

source/simulator/stokes_matrix_free.cc
tcclevenger: yes, it should..

tcclevenger:
> Would it make sense to use IDR as a default? Or are we not sure it is always the better choice?

I _think_ it would be safe to use as the default, but we've only tested it on the Nsinker benchmark. That being said, IDR(2) seemed to be more robust than GMRES(50) increasing DR(mu): (from my thesis)
iters

I'd be ok to make it the default. It could be confusing though since AMG can't use it, but the user could assume it was being used if they just looked at the parameter declaration.

gassmoeller:
> It could be confusing though since AMG can't use it, but the user could assume it was being used if they just looked at the parameter declaration.

This is a problem, yes. Since it is currently mostly useful for GMG (since the memory savings are not as useful if you have a matrix) maybe we do not make it a default until we also make GMG the default preconditioner?

tcclevenger:
> This is a problem, yes. Since it is currently mostly useful for GMG (since the memory savings are not as useful if you have a matrix) maybe we do not make it a default until we also make GMG the default preconditioner?

I agree. I've updated the parameter documentation and there I make a note that the IDR(s) krylov method is unavailable for AMG-based preconditioner

source/simulator/stokes_matrix_free.cc
```diff
@@ -37,6 +37,10 @@
 #include <deal.II/lac/solver_gmres.h>
 #include <deal.II/lac/read_write_vector.templates.h>
 
+#if DEAL_II_VERSION_GTE(9,2,0)
+#include <deal.II/lac/solver_idr.h>
+#else
```
tjhei, line 42: endif

doc/modules/changes/20200525_clevenger
```diff
@@ -0,0 +1,6 @@
+New: The cheap GMG preconditioner for the Stokes system can now use the
+IDR(s) Krylov method. The parameter s can be chosen by the user. This
+method has a short term recurrence and can greatly reduce the memory
+requirement as compared to GMRES.
```
tjhei, line 4: mention that this works with matrix-free only.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1751,17 +1755,39 @@ namespace aspect
...
+            solver(solver_control_cheap, mem,
+                   SolverIDR<dealii::LinearAlgebra::distributed::BlockVector<double> >::
+                   AdditionalData(sim.parameters.idr_s_parameter));
+
+            solver.solve (stokes_matrix,
+                          solution_copy,
+                          rhs_copy,
+                          preconditioner_cheap);
+#else
+            Assert(false, ExcNotImplemented());
```
tjhei, line 1786:
```suggestion
            Assert(false, ExcMessage("The IDR(s) solver requires deal.II 9.2.0 or newer."));
```

tcclevenger:
Thank everyone! I've incorporated @tjhei changes and squashed.

Issue #3512, opened 2020-05-26
No normal flux GMG parallel issues


tcclevenger:
I have found a bug  using no normal flux boundary conditions with spherical shell mesh and block GMG Stokes solver in parallel. If we take the test `tests/shell_3d_gmg.prm`, using 1 processor, we have the computation on 28K DoFs requires 36 iterations (compared to 113 for AMG). But switching to 2 processors requires 106 iterations (112 for AMG) and switching to 4 processors requires 100 iterations.

The implementation should be equivalent for any number of processors for GMG (I expect differing by only a few iterations). Obviously something is going wrong. This bug could be in a few places:
- Creating the no normal flux constraints in `stokes_matrix_free.cc` function `internal::TangentialBoundaryFunctions::compute_no_normal_flux_constraints_shell()`
- Creating the diagonal of the A block GMG matrices for preconditioning (computed in `build_preconditioner()`)
- deal.II bug: incorrect use of constraints in `MGTransferMatrixFree`

It is clearly a preconditioning problem since the solutions between AMG and GMG match for all runs/all processor counts. I am surprised we wouldn't have found during testing. I'll continue looking, but let me know if anyone else has seen unusually high iterations counts for GMG with spherical shells.

tcclevenger:
Simple fix (see https://github.com/geodynamics/aspect/pull/3514).

gassmoeller:
Fixed in #3514

PR #3514, opened 2020-05-27
Fix tangential BCs for GMG

tcclevenger:
From issue https://github.com/geodynamics/aspect/issues/3512, the iteration counts for Stokes would increase when running in parallel on spherical mesh with Tangential boundary. Simple fix: ~~when~~ after constructing the diagonal of the A block operator (used in GMG smoothers), ~~we needed to loop over all owned and ghost cells (previously just looped over owned cells)~~, we need to call `compress()` to exchange contribution from ghosted cells (previously only contributions from owned cells were accounted for).

There is actually already a test (`shell_3d_gmg.prm.dealii_9.2`), but I believe that since it is only on version dealii9.2 it wasn't being run and the high iteration counts were overlooked.

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -2215,9 +2215,7 @@ namespace aspect
             typename DoFHandler<dim>::level_cell_iterator cell = dof_handler_v.begin(level),
                                                           endc = dof_handler_v.end(level);
             for (; cell!=endc; ++cell)
-              if (cell->level_subdomain_id() != numbers::artificial_subdomain_id
-                  &&
-                  cell->level_subdomain_id() != numbers::invalid_subdomain_id)
+              if (cell->level_subdomain_id() == sim.triangulation.locally_owned_subdomain())
                 {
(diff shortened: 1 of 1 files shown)


gassmoeller:
I believe you that this fixes the issue, but can you explain quickly why we need the contributions from ghost cells? As far as I know for AMG we only assemble locally owned cells and that is enough. Is this because the matrix entries for AMG are communicated between neighbors so that the contributions end up in the correct places, and GMG does some local operations without communication?

tcclevenger:
> As far as I know for AMG we only assemble locally owned cells and that is enough. 

Good call, that also holds here. So I rephrase the issue as, I was missing a call to `compress()` at the end of assembly (which communicates the ghost values). So now the fix is just to add `compress()` and keep assembly to only owned cells (which should be cheaper than every processor assembling on ghost cells). Thanks, Rene!

tjhei:
> So now the fix is just to add `compress()`

This means that ghost values had incorrect entries because of the missing call to compress? It might be worth investigating putting NaN into the ghost region when a vector is created in debug mode.

Issue #3517, opened 2020-06-02
Add new ASPECT Publication


ljhwang:
Add new publication to ASPECT website.

@article{doi:10.1146/annurev-earth-082517-010225,
author = {Mitrovica, J.X. and Austermann, J. and Coulson, S. and Creveling, J.R. and Hoggard, M.J. and Jarvis, G.T. and Richards, F.D.},
title = {Dynamic Topography and Ice Age Paleoclimate},
journal = {Annual Review of Earth and Planetary Sciences},
volume = {48},
number = {1},
pages = {585-621},
year = {2020},
doi = {10.1146/annurev-earth-082517-010225},

URL = {
        https://doi.org/10.1146/annurev-earth-082517-010225

},
eprint = {
        https://doi.org/10.1146/annurev-earth-082517-010225

}
,
    abstract = { The connection between the geological record and dynamic topography driven by mantle convective flow has been established over widely varying temporal and spatial scales. As observations of the process have increased and numerical modeling of thermochemical convection has improved, a burgeoning direction of research targeting outstanding issues in ice age paleoclimate has emerged. This review focuses on studies of the Plio-Pleistocene ice age, including investigations of the stability of ice sheets during ice age warm periods and the inception of Northern Hemisphere glaciation. However, studies that have revealed nuanced connections of dynamic topography to biodiversity, ecology, ocean chemistry, and circulation since the start of the current ice-house world are also considered. In some cases, a recognition of the importance of dynamic topography resolves enigmatic events and in others it confounds already complex, unanswered questions. All such studies highlight the role of solid Earth geophysics in paleoclimate research and undermine a common assumption, beyond the field of glacial isostatic adjustment, that the solid Earth remains a rigid, passive substrate during the evolution of the ice age climate system. ▪  Dynamic topography is the large-scale, vertical deflection of Earth's crust driven by mantle convective flow. ▪  This review highlights recent research exploring the implications of the process on key issues in ice age paleoclimate. ▪  This research includes studies of ice sheet stability and inception as well as inferences of peak sea levels during periods of relative ice age warmth. ▪  This review also includes studies on longer timescales, continental-scale ecology and biodiversity, the long-term carbon cycle, and water flux across oceanic gateways. }
}

PR #3520, opened 2020-06-03
add publications from 2019 and 2020

jdannberg:
This addresses #3517 and adds a few more publications. 
This list is probably not complete, I only checked for the last two years, going through the list of papers referencing our 2017 paper to check if they were using ASPECT.


tjhei:
awesome! :-)

Issue #3521, opened 2020-06-05
various plugins that write files do so on each rank


tjhei:
Several plugins write to a file and currently do so from each MPI rank. This is redundant and will lead to bad performance and possibly corrupted data. Offenders I could find:
- [x] point value
- [ ] gravity point value

PR #3522, opened 2020-06-05
point values postprocessor only on rank 0

tjhei:
The point values postprocessor wrote the same file from all MPI ranks
potentially overwriting the results. Fix this.

part of #3521


tjhei:
It happens a couple lines above (MPI::sum).

gassmoeller:
Ah I see. I was confused by the sum. Seems inefficient to communicate and store the data on all processes, but at least its consistent (also probably does not matter for individual values). Good to go.

PR #3524, opened 2020-06-10
Track viscoelastic stresses on particles

naliboff:
This PR enables tracking viscoelastic stresses on particles. Some additional cleanup and documentation is needed before a review, but a number of people (@dansand) may be interested in doing some preliminary testing with it. I've only tested this with the `viscoelastic_stress_build` benchmark so far, but that seems to be working.

This PR builds on #3482, but we ended up not needing to use the MaterialModel inputs and outputs in this instance. However, no issues with the new interface :+1: 


benchmarks/viscoelastic_stress_build-up/viscoelastic_stress_build-up_particles.prm
```diff
@@ -0,0 +1,203 @@
+# This parameter file reproduces a 2D analytical benchmark for the build-up
```
gassmoeller, line 1: This looks a lot like a modified copy of the `viscoelastic_stress_build-up.prm` file in that folder. Could you use the `include` directive (as in `tests/rigid_shear_instantaneous.prm`) and then only keep the modified commands in this file? That would also make it simpler to keep up to date.

include/aspect/particle/property/elastic_stress.h
```diff
@@ -0,0 +1,126 @@
...
+ <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _aspect_particle_property_elastic_stress_h
+#define _aspect_particle_property_elastic_stress_h
+
+#include <aspect/particle/property/interface.h>
+#include <aspect/simulator_access.h>
+#include <aspect/material_model/visco_plastic.h>
+#include <aspect/material_model/viscoelastic.h>
```
gassmoeller, line 27: I dont think these two includes are needed in the header file. Please remove (decreases compile time).

include/aspect/particle/property/elastic_stress.h
```diff
@@ -0,0 +1,126 @@
...
+                 * Initialization function. This function is called once at the
+                 * creation of every particle for every property to initialize its
+                 * value.
+                 *
+                 * @param [in] position The current particle position.
+                 * @param [in,out] particle_properties The properties of the particle
+                 * that is initialized within the call of this function. The purpose
+                 * of this function should be to extend this vector by a number of
+                 * properties.
+                 */
```
gassmoeller, line 65: Please indent manually

include/aspect/particle/property/elastic_stress.h
```diff
@@ -0,0 +1,126 @@
...
+           * @param [in] position The current particle position.
+           *
+           * @param [in] solution The values of the solution variables at the
+           * current particle position.
+           *
+           * @param [in] gradients The gradients of the solution variables at
+           * the current particle position.
+           *
+           * @param [in,out] particle_properties The properties of the particle
+           * that is updated within the call of this function.
```
gassmoeller, line 89: This documentation is out of date, please use the updated one, or use `@copydoc aspect::Particle::Property::update_particle_property()`

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,171 @@
...
+namespace aspect
+{
+  namespace Particle
+  {
+    namespace Property
+    {
+      template <int dim>
+      ElasticStress<dim>::ElasticStress ()
+      {}
+
```
gassmoeller, line 35:
```suggestion



```

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,171 @@
...
+        data.push_back(this->get_initial_composition_manager().initial_composition(position,this->introspection().compositional_index_for_name("stress_xx")));
+
+        data.push_back(this->get_initial_composition_manager().initial_composition(position,this->introspection().compositional_index_for_name("stress_yy")));
+
+        data.push_back(this->get_initial_composition_manager().initial_composition(position,this->introspection().compositional_index_for_name("stress_xy")));
+
+        if (dim == 3)
+          {
+            data.push_back(this->get_initial_composition_manager().initial_composition(position,this->introspection().compositional_index_for_name("stress_zz")));
+
```
gassmoeller, line 65: I would like to keep the order always consistent with the deal.II tensor components, it might come back to bite us if we change the order just to make this code a bit simpler.

```suggestion
        if (dim == 2)
        {
          data.push_back(this->get_initial_composition_manager().initial_composition(position,this->introspection().compositional_index_for_name("stress_xy")));
}
        else if (dim == 3)
          {
            data.push_back(this->get_initial_composition_manager().initial_composition(position,this->introspection().compositional_index_for_name("stress_zz")));

data.push_back(this->get_initial_composition_manager().initial_composition(position,this->introspection().compositional_index_for_name("stress_xy")));
```

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,171 @@
...
+      template <int dim>
+      void
+      ElasticStress<dim>::update_particle_property(const unsigned int data_position,
+                                                   const Vector<double> &solution,
+                                                   const std::vector<Tensor<1,dim> > &gradients,
+                                                   typename ParticleHandler<dim>::particle_iterator &particle) const
+      {
+        MaterialModel::MaterialModelInputs<dim> in(1, this->n_compositional_fields());
+
```
gassmoeller, line 81: Why not also set `in.cell` now that it is simple to do? Who knows which parts of the material model may depend on the cell?

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,171 @@
...
+      ElasticStress<dim>::need_update() const
+      {
+        return update_output_step;
+      }
+
+      template <int dim>
+      UpdateFlags
+      ElasticStress<dim>::get_needed_update_flags () const
+      {
+        return update_values;
```
gassmoeller, line 117:
```suggestion
        return update_values | update_gradients;
```

Otherwise you can not ask for the strain rate in the update function.

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,171 @@
...
+          {
+            particle->get_properties()[data_position + i] = solution[this->introspection().component_indices.compositional_fields[i]];
+          }
+      }
+
+      template <int dim>
+      UpdateTimeFlags
+      ElasticStress<dim>::need_update() const
+      {
+        return update_output_step;
```
gassmoeller, line 110: So at the moment this is purely an output tracer? This means you could not use this particle for actually carrying the stress instead of a compositional field, is that what you want? Otherwise this would need to be `update_time_step` (I think).

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,171 @@
...
+        if (this->introspection().compositional_name_exists("stress_xy"))
+          property_information.emplace_back("stress_xy",1);
+
+        if (dim == 3)
+          {
+            if (this->introspection().compositional_name_exists("stress_zz"))
+              property_information.emplace_back("stress_zz",1);
+
+            if (this->introspection().compositional_name_exists("stress_xz"))
+              property_information.emplace_back("stress_xz",1);
```
gassmoeller, line 142: This would need to be adjusted as above.

naliboff:
@gassmoeller - Thanks for the review, but unfortunately still some debugging left to be done. It turns out we will need the `MaterialModelOutputs` reaction terms accessed through the interface, as of course this is all just passive tracking if the stresses are specifically labeled in the parameter file as being tracked through particles. Currently trying to figure out why the reaction terms are still being returned as 0. Sorry, should have mentioned this here yesterday when I realized the issue.

source/particle/property/elastic_stress.cc
gassmoeller: `in.cell = particle->get_surrounding_cell(this->get_triangulation());`

naliboff: @tjhei - @gassmoeller and I had a question regarding obtain a cell iterator for `in.cell`. The line above returns a cell iterator for the triangulation, but what we actually need is a DoF cell iterator for in.cell. Do you know of a way to convert from a triangulation to DoF cell iterator? 

We tried a few permutations (see example below), but couldn't find the right structure. Example:

gassmoeller: I had another look and the following seems to work for me:
```
        in.current_cell = typename DoFHandler<dim>::active_cell_iterator(*particle->get_surrounding_cell(this->get_triangulation()), 
          &(this->get_dof_handler()));
```
Let me know if that works for you.
I will check if I can add a convenience function to deal.II to avoid this awkward conversion (in future versions).

gassmoeller: Oh, and you also have to set `in.position` to the particle positions, otherwise the material model will create floating point exceptions (that confused me when testing this).

naliboff: @gassmoeller - this did the trick, thanks! I'll update the PR again once I've finished addressing the remaining comments, tested all the relevant benchmarks, and added a few tests.

Fingers crossed that using the particles to smooth out oscillations in the elastic stresses will improve the VEP non-linear convergence behavior.

include/aspect/particle/property/elastic_stress.h
naliboff: @gassmoeller - these were needed for the following checks in `elastic_stress.cc`:
```
        AssertThrow((Plugins::plugin_type_matches<const MaterialModel::ViscoPlastic<dim>>(this->get_material_model())
                     ||
                     Plugins::plugin_type_matches<const MaterialModel::Viscoelastic<dim>>(this->get_material_model())),
                    ExcMessage("This particle property only makes sense in combination with the viscoelastic or visco_plastic material model."));
```

naliboff:
@gassmoeller - I think this is ready for another review. The three tests I added are all passing and everything should be in fairly good shape.

source/particle/property/elastic_stress.cc
gassmoeller: I think this still needs 2 more empty lines

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,183 @@
...
+        AssertThrow((Plugins::plugin_type_matches<const MaterialModel::ViscoPlastic<dim>>(this->get_material_model())
+                     ||
+                     Plugins::plugin_type_matches<const MaterialModel::Viscoelastic<dim>>(this->get_material_model())),
+                    ExcMessage("This particle property only makes sense in combination with the viscoelastic or visco_plastic material model."));
+
+        AssertThrow(this->get_parameters().enable_elasticity == true,
+                    ExcMessage ("This particle property should only be used if 'Enable elasticity' is set to true"));
+
+      }
+
```
gassmoeller, line 49: same here and all other functions

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,183 @@
...
+        Tensor<2,dim> grad_u;
+        for (unsigned int d=0; d<dim; ++d)
+          grad_u[d] = gradients[d];
+        in.strain_rate[0] = symmetrize (grad_u);
+
+        MaterialModel::MaterialModelOutputs<dim> out(1, this->n_compositional_fields());
+        this->get_material_model().evaluate (in,out);
+
+        for (unsigned int i = 0; i < SymmetricTensor<2,dim>::n_independent_components ; ++i)
+          particle->get_properties()[data_position + i] += out.reaction_terms[0][i];
```
gassmoeller, line 110: Is it guaranteed that the stress terms are the first terms in these material models?

source/particle/property/elastic_stress.cc
```diff
@@ -0,0 +1,183 @@
...
+      }
+
+      template <int dim>
+      void
+      ElasticStress<dim>::update_particle_property(const unsigned int data_position,
+                                                   const Vector<double> &solution,
+                                                   const std::vector<Tensor<1,dim> > &gradients,
+                                                   typename ParticleHandler<dim>::particle_iterator &particle) const
+      {
+        MaterialModel::MaterialModelInputs<dim> in(1, this->n_compositional_fields());
```
gassmoeller, line 84: You could make this a mutable class member like I did in Property::ViscoPlasticStrainInvariant. This would save some time allocating these vectors for every particle. Same for the MaterialModelOutputs object. But I guess that since you need to call the whole expensive evaluate function it will not matter much, this property is going to be expensive either way.

source/particle/property/elastic_stress.cc
naliboff: Yes, that is enforced in the elastic rheology module.

source/particle/property/elastic_stress.cc
naliboff: I gave this a shot and we'll see if I did it correctly. However, in the process I did have to add an `override` to `update_particle_property` in `include/aspect/particle/property/composition.h` to suppress a compiler warning. I'm not sure exactly why that came up now, but will revert if this was not the correct fix or I implemented something incorrectly on my end.

naliboff:
@gassmoeller - Thanks for another review and great to hear this is close to being ready!

egpuckett:
@sac-bsa has been computing the bending beam with pure viscosity.  This is great! Go for it!

naliboff:
@gassmoeller - The tests are still passing after making the material model inputs and outputs objects mutable 👍.

source/particle/property/elastic_stress.cc
```diff
@@ -74,56 +85,64 @@ namespace aspect
...
+
       template <int dim>
       void
       ElasticStress<dim>::update_particle_property(const unsigned int data_position,
                                                    const Vector<double> &solution,
                                                    const std::vector<Tensor<1,dim> > &gradients,
                                                    typename ParticleHandler<dim>::particle_iterator &particle) const
       {
-        MaterialModel::MaterialModelInputs<dim> in(1, this->n_compositional_fields());
+        //material_inputs = MaterialModel::MaterialModelInputs<dim>(1, this->n_compositional_fields());
```
gassmoeller, line 97: remove the commented line

source/particle/property/elastic_stress.cc
```diff
@@ -74,56 +85,64 @@ namespace aspect
...
 
         Tensor<2,dim> grad_u;
         for (unsigned int d=0; d<dim; ++d)
           grad_u[d] = gradients[d];
-        in.strain_rate[0] = symmetrize (grad_u);
+        material_inputs.strain_rate[0] = symmetrize (grad_u);
 
-        MaterialModel::MaterialModelOutputs<dim> out(1, this->n_compositional_fields());
-        this->get_material_model().evaluate (in,out);
+        //material_outputs = MaterialModel::MaterialModelOutputs<dim>(1, this->n_compositional_fields());
```
gassmoeller, line 119: remove commented line

source/particle/property/elastic_stress.cc
naliboff: @gassmoeller - done and same with the line below. I squashed the commit in the last one.

PR #3530, opened 2020-06-15
fix benchmarks and cookbooks problems

tjhei:
fix various warnings and errors in cookbooks and benchmarks

### DIFF
--- a/benchmarks/buiter_et_al_2016_jsg/exp_1.prm
+++ b/benchmarks/buiter_et_al_2016_jsg/exp_1.prm
@@ -138,7 +138,7 @@ subsection Material model
 #                                                     [bkd, sand,  bound, block, sticky_air, total_strain]
     set Reference strain rate                       = 1.e-5
     set Minimum strain rate                         = 1.e-10
-    set Strain weakening mechanism                  = finite strain tensor
+    set Strain weakening mechanism                  = total strain
 
     set Start plasticity strain weakening intervals = 0,    0.50,  0.5,   0,     0,          1e5
     set End plasticity strain weakening intervals   = 0.1,  1.00,  1,     0.1,   0.1,        1e4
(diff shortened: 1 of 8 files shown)


tjhei:
@naliboff can you please take a look at the necessary changes related to visco-plastic? I have no idea what the correct settings should be.

benchmarks/buiter_et_al_2016_jsg/exp_1.prm
```diff
@@ -138,7 +138,8 @@ subsection Material model
 #                                                     [bkd, sand,  bound, block, sticky_air, total_strain]
     set Reference strain rate                       = 1.e-5
     set Minimum strain rate                         = 1.e-10
-    set Use strain weakening                        = true
+    set Strain weakening mechanism                  = finite strain tensor
```
naliboff, line 141:
```suggestion
    set Strain weakening mechanism                  = total strain
```

benchmarks/yamauchi_takei_2016_anelasticity/anelasticity_temperature.cc
```diff
@@ -298,15 +298,15 @@ namespace aspect
...
+      // (peak_amplitude*std::exp(-1*(std::pow(std::log(peak_period/normalised_period),2)/
+      // (2*std::pow(peak_width,2))))))+(unrelaxed_compliance*normalised_period);
       // calculate anharmonic Vs
-      anharmonic_Vs=1/(std::sqrt(density*unrelaxed_compliance)*1e3);
+      // anharmonic_Vs=1/(std::sqrt(density*unrelaxed_compliance)*1e3);
       // calculate Vs
       anelastic_Vs=1/(std::sqrt(density*storage_compliance)*1e3);
       // calculate attenuation
-      attenuation=loss_compliance/storage_compliance;
+      // attenuation=loss_compliance/storage_compliance;
```
naliboff, line 309: Should the commented lines above just be removed?

tjhei: I am not sure if it makes sense to preserve the logic for computing all these quantities. I can remove all of it, if you prefer.

naliboff: I'm not familiar with this benchmark either, so maybe best to presence the logic for the time being.

PR #3531, opened 2020-06-15
fix geoid warning

tjhei:
introduced in #3527


Shangxin-Liu:
Hi Timo, @tjhei 

Thanks! This warning did not come up in my test so I did not notice it.

I'm still a little bit confused on how this citation call can be linked to the paper citation suggestion on ASPECT features in the website aspect.geodynamics.org. Do we still need to add the paper there manually and link to this citation call somehow? I'm just not sure how this citation call works.

tjhei:
> I'm still a little bit confused on how this citation call can be linked to the paper citation suggestion on ASPECT features in the website aspect.geodynamics.org

I will do this in the next few days, don't worry.

PR #3532, opened 2020-06-15
use single precision for gmg

tcclevenger:
Based on the discussion in section 4.2 of https://dl.acm.org/doi/pdf/10.1145/3322813. 

This pull request changes the the GMG solver to be setup and run in single precision. It will not change the accuracy of the solution (since all active level computations are done in double precision), but improves runtime.

- Test with Nsinker, 7 global refinements (6.8M DoFs), run with 8 cores:
**Single precision GMG**: 28 iterations, 10.5s Stokes solve time
**Double precision GMG**: 28 iterations, 14.9s Stokes solve time
This gives a roughly 30% decrease in runtime which is similar to the results in the paper. 

The changes are basically just changing a few `double` to `float`. I define a type `gmg_number` to `float` at the top of stokes_matrix_free.h. This can be changed to `double` to go back to double precision gmg. I'll comment the only other significant change.

### DIFF
--- a/include/aspect/stokes_matrix_free.h
+++ b/include/aspect/stokes_matrix_free.h
@@ -43,6 +43,8 @@
 #include <deal.II/lac/la_parallel_vector.h>
 #include <deal.II/lac/la_parallel_block_vector.h>
 
+using gmg_number = float;
+
 namespace aspect
 {
   using namespace dealii;
@@ -360,14 +362,14 @@ namespace aspect
        * Return a pointer to the MGTransfer object used for the A block
        * of the block GMG Stokes solver.
        */
-      virtual const MGTransferMatrixFree<dim,double> &
+      virtual const MGTransferMatrixFree<dim,gmg_number> &
       get_mg_transfer_A () const = 0;
 
       /**
        * Return a pointer to the MGTransfer object used for the Schur
        * complement block of the block GMG Stokes solver.
        */
-      virtual const MGTransferMatrixFree<dim,double> &
+      virtual const MGTransferMatrixFree<dim,gmg_number> &
       get_mg_transfer_S () const = 0;
 
       /**
@@ -381,7 +383,7 @@ namespace aspect
        * Return a pointer to the Tables containing the viscosities on
        * the multigrid levels used in the block GMG Stokes solver.
        */
-      virtual const MGLevelObject<Table<1, VectorizedArray<double>>> &
+      virtual const MGLevelObject<Table<1, VectorizedArray<gmg_number>>> &
       get_level_viscosity_tables() const = 0;
   };
 
@@ -494,14 +496,14 @@ namespace aspect
        * Return a pointer to the MGTransfer object used for the A block
        * of the block GMG Stokes solver.
        */
-      const MGTransferMatrixFree<dim,double> &
+      const MGTransferMatrixFree<dim,gmg_number> &
       get_mg_transfer_A () const override;
 
       /**
        * Return a pointer to the MGTransfer object used for the Schur
        * complement block of the block GMG Stokes solver.
        */
-      const MGTransferMatrixFree<dim,double> &
+      const MGTransferMatrixFree<dim,gmg_number> &
       get_mg_transfer_S () const override;
 
       /**
@@ -515,7 +517,7 @@ namespace aspect
        * Return a pointer to the Tables containing the viscosities on
        * the multigrid levels used in the block GMG Stokes solver.
        */
-      const MGLevelObject<Table<1, VectorizedArray<double>>> &
+      const MGLevelObject<Table<1, VectorizedArray<gmg_number>>> &
       get_level_viscosity_tables() const override;
 
     private:
@@ -536,15 +538,18 @@ namespace aspect
       FESystem<dim> fe_projection;
 
       Table<1, VectorizedArray<double>> active_viscosity_table;
-      MGLevelObject<Table<1, VectorizedArray<double>>> level_viscosity_tables;
+      MGLevelObject<Table<1, VectorizedArray<gmg_number>>> level_viscosity_tables;
 
       // This variable is needed only in the setup in both evaluate_material_model()
       // and build_preconditioner(). It will be deleted after the last use.
-      MGLevelObject<dealii::LinearAlgebra::distributed::Vector<double> > level_viscosity_vector;
+      MGLevelObject<dealii::LinearAlgebra::distributed::Vector<gmg_number> > level_viscosity_vector;
 
-      typedef MatrixFreeStokesOperators::StokesOperator<dim,velocity_degree,double> StokesMatrixType;
-      typedef MatrixFreeStokesOperators::MassMatrixOperator<dim,velocity_degree-1,double> SchurComplementMatrixType;
-      typedef MatrixFreeStokesOperators::ABlockOperator<dim,velocity_degree,double> ABlockMatrixType;
+      using StokesMatrixType = MatrixFreeStokesOperators::StokesOperator<dim,velocity_degree,double>;
+      using SchurComplementMatrixType = MatrixFreeStokesOperators::MassMatrixOperator<dim,velocity_degree-1,double>;
+      using ABlockMatrixType = MatrixFreeStokesOperators::ABlockOperator<dim,velocity_degree,double>;
+
+      using GMGSchurComplementMatrixType = MatrixFreeStokesOperators::MassMatrixOperator<dim,velocity_degree-1,gmg_number>;
+      using GMGABlockMatrixType = MatrixFreeStokesOperators::ABlockOperator<dim,velocity_degree,gmg_number>;
 
       StokesMatrixType stokes_matrix;
       ABlockMatrixType A_block_matrix;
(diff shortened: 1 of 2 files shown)


source/simulator/stokes_matrix_free.cc
```diff
@@ -2285,7 +2285,10 @@ namespace aspect
                 }
 
             diagonal_matrix.compress(VectorOperation::add);
-            mg_matrices_A_block[level].set_diagonal(diagonal_matrix.get_vector());
+
+            dealii::LinearAlgebra::distributed::Vector<gmg_number> diagonal;
+            diagonal = diagonal_matrix.get_vector();
+            mg_matrices_A_block[level].set_diagonal(diagonal);
```
tcclevenger, line 2291: This is a bit ugly, but for reasons of instantiations inside the dealii ConstraintMatrix classes, we must compute the diagonal of the GMG matrices (for curved boundary, tangential constraints) in double precision. Thus, we need to create a single precision diagonal vector and copy the contents after computing the diagonal and before setting them in the MatrixFree operators. This should not cause a significant increase in runtime.

include/aspect/stokes_matrix_free.h
```diff
@@ -43,6 +43,8 @@
 #include <deal.II/lac/la_parallel_vector.h>
 #include <deal.II/lac/la_parallel_block_vector.h>
 
+using gmg_number = float;
```
tjhei, line 46:
```suggestion
/**
 * Typedef for the number type for the multigrid operators. Can be either float or double.
 */
using GMGNumberType = float;
```

source/simulator/stokes_matrix_free.cc
tjhei: please leave a comment before the assignment that this does the double->float conversion.

gassmoeller:
Merging #3495 created some conflicts, could you rebase? Otherwise good to go.

tjhei:
and test results are somehow broken again.

tcclevenger:
> and test results are somehow broken again.

I will have to investigate. But the iteration counts seem to have risen drastically for only some of the tests.

tjhei:
you will need to rebase and fix the single precision issues as well. Let me know if you need help with finding the bug.

tcclevenger:
> you will need to rebase and fix the single precision issues as well. Let me know if you need help with finding the bug.

I fixed the bug. I'll focus on testing the failing benchmarks tomorrow.

tcclevenger:
The issue seems to show up when using `no Advection, iterated Stokes`, but not on the first iteration (I only tested with `first timestep only, single Stokes`, which gives basically identical output for `float` and `double`). I'm still investigating.

Outside of that, thought, there could be an issue with the scale of values needed. I believe `float` type is bounded above by 3.40282e+38. Is it the case where values of viscosity can exceed this number?

gassmoeller:
> Outside of that, thought, there could be an issue with the scale of values needed. I believe float type is bounded above by 3.40282e+38. Is it the case where values of viscosity can exceed this number?

Realistic values for viscosity will lie between 1e16 (worst case) to even lower (in some melt models, and 1 for nondimensional models) and 1e25. 1e38 is outside of realistic ranges, but may occur in some unrealistic tests. If that turns out to be the problem I think it is ok to throw an exception.

naliboff:
> Realistic values for viscosity will lie between 1e16 (worst case) to even lower (in some melt models, and 1 for nondimensional models) and 1e25
It likely would not make much of a difference here, but I often have 1e26 or 1e28 Pa s as the upper viscosity limit in many models. The minimum viscosity in these models would be typically be 1e18 or 1e20 Pa s.

tjhei:
> I'm still investigating.

could this be a missing update_ghost_values() on the diagonal vector?

tcclevenger:
I'm continuing to investigate. It looks like the `initial_nonlinear_residual` (first return value of `solve_stokes()`) is very different on the 2+ iterated Stokes solves between `float` and `double` GMG. I'm a little confused since computing this value does not involve GMG (except that the previous solution is used). Perhaps the previous solution is slightly less accurate? But this doesn't make sense since it is solved to the same tolerance no matter the GMGNumberType.

gassmoeller:
Any new insights? Let us know if you want input.

tjhei:
> Any new insights? Let us know if you want input.

thanks, Rene. We are working on things and we do have a lead. Conrad will get to this in the next 2 weeks, I hope.

tcclevenger:
@gassmoeller sorry for the absence. I’ll be getting back to this this week. @tjhei and I talked last week and there are a few simple fixes that could work, so I’m I should be able to wrap this up. I’ll keep you guys updated.

tcclevenger:
A closer look, I believe the new code is implemented correctly. Here is a comparison of the differences in the performance with a few benchmarks:

- **Nonlinear channel flow, 4 timesteps (0-4), 214K DoFs, 1 processor**
__Single precision GMG__:   63 total Stokes solves, 75.5s solve time
__Double precision GMG__: 65 total Stokes solves, 79.9s solve time

- **SolCx, no advection iterated Stokes, 3K DoFs, 1 processor**
__Single precision GMG__:   First Stokes: 12 iterations, 2nd Stokes: 3 iteration
__Double precision GMG__: First Stokes: 12 iterations, 2nd Stokes: 0 iteration

- **Nsinker, no advection iterated Stokes, 150K DoFs, 1 processor**
__Single precision GMG__:   First Stokes: 23 iterations, 2nd Stokes: 1 iteration
__Double precision GMG__: First Stokes: 23 iterations, 2nd Stokes: 1 iteration

- **Nsinker, first timestep only, single Stokes, 6.8M DoFs, 8 processor**
__Single precision GMG__:  28 iterations,  10.5s solve time
__Double precision GMG__: 28 iterations, 14.9s solve time


There were 2 types of failing tests before:
1. Tests which were failing due to slight differences in the solver performance (the preconditioner is slightly less effective so we expect higher iteration counts. Nonlinear channel flow test is in this category.) or different memory statistics since the viscosity tables in single precision will require less memory (matrix_nonzeros_7).
2. Tests which were failing due major differences in solver performance, specifically when using "no advection, iterated Stokes". All these tests were on very small problems (<1000 unknowns) and so we are not worried about the differences. I tested the same benchmarks with larger number of DoFs and we get expected output. For all of these (except `prescribed_dilation_gmg` since it's .prm is based on a different test) I decided to change to "first timestep only, single Stokes` for simplicity.

Other tests will fail now with the changes and so I will be sure to correct these when the tester is finished.

tjhei:
Thanks, Conrad. I will take a look at the tests in more detail. Would you mind squashing your commits down and rebasing to the current master? That would make it a lot easier to inspect the changes in the test results.

tjhei:
@gassmoeller can you please take another look?

gassmoeller:
I am ok with the slight increase in linear iterations (expected due to lower precision of preconditioner), but can you explain your reasoning here again:

> 2. Tests which were failing due major differences in solver performance, specifically when using "no advection, iterated Stokes". All these tests were on very small problems (<1000 unknowns) and so we are not worried about the differences. I tested the same benchmarks with larger number of DoFs and we get expected output. For all of these (except prescribed_dilation_gmg since it's .prm is based on a different test) I decided to change to "first timestep only, single Stokes` for simplicity.

I ran the sol_cx_2_gmg test with double and single precision with the iterated Stokes scheme and like you found significant differences in nonlinear solver behavior. The old version converges to 1e-8 in the second iteration (expected, because this is actually a linear problem, viscosity and density are independent of the solution), while in single precision I get this:

```
  Solving Stokes system... 16+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 1: 1

   Solving Stokes system... 14+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 2: 4.08645

   Solving Stokes system... 14+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 3: 2.29006

   Solving Stokes system... 16+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 4: 1.01736

   Solving Stokes system... 14+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 5: 0.644649

   Solving Stokes system... 14+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 6: 0.132881

   Solving Stokes system... 12+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 7: 0.0387798

   Solving Stokes system... 9+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 8: 0.00600291

   Solving Stokes system... 8+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 9: 0.00220304

   Solving Stokes system... 8+0 iterations.
      Relative nonlinear residual (Stokes system) after nonlinear iteration 10: 0.000221702
```

Also when I decrease the linear solver tolerance (from default 1e-7 to 1e-9 or 1e-11) the nonlinear solver behavior becomes worse (higher residual after 10 nonlinear iterations).

This suggests the solution in the first linear solve is actually worse than before (i.e. either the tolerance computation is different, or the solver stops although the tolerance is not reached, maybe because some residual computation is done in single precision?), and additionally it does not converge in the expected number of nonlinear iterations. Your changed test does not show this anymore because it uses a single Stokes solve, but that does not solve or explain the problem. Why do you think it is ok, because this is a small test?

tjhei:
This is a good point, @gassmoeller . I think we need to look at this in more detail. Conrad somehow convinced me that the change is acceptable in our last conversation.

> This suggests the solution in the first linear solve is actually worse than before (i.e. either the tolerance computation is different, or the solver stops although the tolerance is not reached, maybe because some residual computation is done in single precision?),

Tolerance and residual computation is done as before in double precision (only the preconditioner is different), so I am not sure what could be different. I will need to take a look as well, I think.

bangerth:
What happens if you put the new typedef to `double`? You should be getting the exact same results as are currently in the testsuite. If you don't, then that's a lead worth pursuing.

tjhei:
> What happens if you put the new typedef to `double`?

After rebase to current master and setting things back to double, all tests pass without changes to test output.

bangerth:
So then at least you haven't introduced a bug :-)

tjhei:
I made some progress with my research, but I will have to talk to Conrad. It looks like the resulting constant pressure is vastly different between the float and double preconditioner.

gassmoeller:
Are we running into the issue that the dynamic pressure is usually 3-7 orders of magnitude smaller than the static pressure and the dynamic pressure is the only part that affects the velocities? This is why many other codes solve for dynamic pressure only.

PR #3537, opened 2020-06-16
[CI] test cookbooks/benchmarks

tjhei:
- generate makefiles to configure/compile/validate all files under cookbooks/ and benchmarks/
- optionally also execute the .prm files
- add the validation to jenkins


tjhei:
this is ready for review

tjhei:
this adds about 1 minute to the test times as it only validates (and not runs).

tjhei:
> I would think we can fix the todos in follow ups, since they are not introduced by this PR, right?

yes, I agree.

PR #3544, opened 2020-06-18
phase transition: average phase value

lhy11009:

This is the second step of implementing phase transitions to the material model. I added a phase_average_value function to utilities.cc calculate parameters of compositions from the parameters of phases using the gamma function. On top of that, I changed the implementation of function phase_average_equation_of_state_outputs in interfaces.cc.
As @gassmoeller has addressed before, we should not call this phase_average_value function if there is no phase on a composition. That's what I have taken care of in this PR. Although I haven't included the viscosity part of it, which I intend to deal with in the third part. 
Thanks, @gassmoeller for review the last one, this one and another one in the near future -v-

### For all pull requests:

### DIFF
--- a/doc/modules/changes/20200701_haoyuanli
+++ b/doc/modules/changes/20200701_haoyuanli
@@ -1,8 +1,4 @@
-New: Added function phase_average_value, namespace PhaseUtilities to namespace MaterialModel::MaterialUtilities.
-This new function averages parameters(e.g density, viscosity) across all phases for a composition. It
-takes parameters of phases and values of gamma functions as inputs, applys an arithmetic or 
-logarithmic averaging and return the averaged value. A enum type PhaseUtilities::PhaseAveragingOperation 
-is defined to determine the type of operation(e.g arithmetic, logarithmic) to perform in the phase_average_value
-function.
+New: Added a new function PhaseUtilities::phase_average_value to the namespace MaterialModel::MaterialUtilities.
+This new function averages parameters(e.g density, viscosity) across all phases for a composition.
 <br>
 (Haoyuan Li, 2020/07/01)
--- a/doc/modules/changes/20200701_haoyuanli
+++ b/doc/modules/changes/20200701_haoyuanli
@@ -1,0 +1,8 @@
+New: Added function phase_average_value, namespace PhaseUtilities to namespace MaterialModel::MaterialUtilities.
+This new function averages parameters(e.g density, viscosity) across all phases for a composition. It
+takes parameters of phases and values of gamma functions as inputs, applys an arithmetic or 
+logarithmic averaging and return the averaged value. A enum type PhaseUtilities::PhaseAveragingOperation 
+is defined to determine the type of operation(e.g arithmetic, logarithmic) to perform in the phase_average_value
+function.
+<br>
+(Haoyuan Li, 2020/07/01)
--- a/include/aspect/material_model/equation_of_state/interface.h
+++ b/include/aspect/material_model/equation_of_state/interface.h
@@ -113,8 +113,8 @@ namespace aspect
     template <int dim>
     void
     phase_average_equation_of_state_outputs(const EquationOfStateOutputs<dim> &eos_outputs_all_phases,
-                                            const MaterialUtilities::PhaseFunction<dim> &phase_function,
-                                            MaterialUtilities::PhaseFunctionInputs<dim> &phase_in,
+                                            const std::vector<double> &phase_function_values,
+                                            const std::vector<unsigned int> &n_phases_per_composition,
                                             EquationOfStateOutputs<dim> &eos_outputs);
   }
 }
--- a/include/aspect/material_model/utilities.h
+++ b/include/aspect/material_model/utilities.h
@@ -268,6 +268,46 @@ namespace aspect
                             const std::vector<double> &parameter_values,
                             const CompositionalAveragingOperation &average_type);
 
+      /**
+       * Utilities for material models with multiple phases
+       */
+      namespace PhaseUtilities
+      {
+        /**
+         * Enumeration for selecting which averaging scheme to use when
+         * averaging the properties of different phases.
+         * Select between arithmetic and logarithmic.
+        */
+        enum PhaseAveragingOperation
+        {
+          arithmetic,
+          logarithmic
+        };
+      }
+
+      /**
+      * Material models compute output quantities such as the viscosity, the
+      * density, etc. For some models, these values may depend on the phase in
+      * addition to the composition, and more than one phase field might have
+      * nonzero values at a given quadrature point. This means that properties
+      * for each composition have to be averaged based on the fractions of each
+      * phase field present. This function performs this type of averaging.
+      * The averaging is based on the choice in @p operation. Averaging is conducted
+      * over the phase functions given in @phase_function_values, with
+      * @parameter_values containing values of all individual phases. Unlike the average_value
+      * function defined for compositions, averaging in this function is calculated based
+      * on phase functions and the change of variables on the trajectory of phase boundaries.
+      * Thus on a single phase boundary, values of variables change gradually from one phase
+      * to the other. The values of the phase function used to average the properties varies
+      * between 0 and 1.
+      */
+      double phase_average_value (const std::vector<double> &phase_function_values,
+                                  const std::vector<unsigned int> &n_phases_per_composition,
+                                  const std::vector<double> &parameter_values,
+                                  const unsigned int composition,
+                                  const PhaseUtilities::PhaseAveragingOperation operation = PhaseUtilities::arithmetic);
+
+
 
       /**
        * A data structure with all inputs for the
--- a/source/material_model/equation_of_state/interface.cc
+++ b/source/material_model/equation_of_state/interface.cc
@@ -43,47 +43,24 @@ namespace aspect
     template <int dim>
     void
     phase_average_equation_of_state_outputs(const EquationOfStateOutputs<dim> &eos_outputs_all_phases,
-                                            const MaterialUtilities::PhaseFunction<dim> &phase_function,
-                                            MaterialUtilities::PhaseFunctionInputs<dim> &phase_in,
+                                            const std::vector<double> &phase_function_values,
+                                            const std::vector<unsigned int> &n_phases_per_composition,
                                             EquationOfStateOutputs<dim> &eos_outputs)
     {
-      const std::vector<unsigned int> &n_phases_per_composition =
-        phase_function.n_phase_transitions_for_each_composition();
-
-      unsigned int j=0;
       for (unsigned int c=0; c<eos_outputs.densities.size(); ++c)
         {
-          Assert(j<eos_outputs_all_phases.densities.size(),
-                 ExcInternalError());
-
-          eos_outputs.densities[c] = eos_outputs_all_phases.densities[j];
-          eos_outputs.thermal_expansion_coefficients[c] = eos_outputs_all_phases.thermal_expansion_coefficients[j];
-          eos_outputs.specific_heat_capacities[c] = eos_outputs_all_phases.specific_heat_capacities[j];
-          eos_outputs.compressibilities[c] = eos_outputs_all_phases.compressibilities[j];
-          eos_outputs.entropy_derivative_pressure[c] = eos_outputs_all_phases.entropy_derivative_pressure[j];
-          eos_outputs.entropy_derivative_temperature[c] = eos_outputs_all_phases.entropy_derivative_temperature[j];
-
-          ++j;
-
-          for (unsigned int p=0; p<n_phases_per_composition[c]; ++p, ++j)
-            {
-              Assert(j<eos_outputs_all_phases.densities.size(),
-                     ExcInternalError());
-
-              // Compute which phase index the current phase of the current composition has
-              // in the list of all phase functions. This is the number of times we visited
-              // this line j minus one per composition c (the original state of each composition
-              // does not have a phase function).
-              phase_in.phase_index = j-(c+1);
-              const double gamma = phase_function.compute_value(phase_in);
-
-              eos_outputs.densities[c] += gamma * (eos_outputs_all_phases.densities[j]-eos_outputs_all_phases.densities[j-1]);
-              eos_outputs.thermal_expansion_coefficients[c] += gamma * (eos_outputs_all_phases.thermal_expansion_coefficients[j]-eos_outputs_all_phases.thermal_expansion_coefficients[j-1]);
-              eos_outputs.specific_heat_capacities[c] += gamma * (eos_outputs_all_phases.specific_heat_capacities[j]-eos_outputs_all_phases.specific_heat_capacities[j-1]);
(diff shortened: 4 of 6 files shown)


include/aspect/material_model/equation_of_state/interface.h
```diff
@@ -113,8 +113,7 @@ namespace aspect
     template <int dim>
     void
     phase_average_equation_of_state_outputs(const EquationOfStateOutputs<dim> &eos_outputs_all_phases,
-                                            const MaterialUtilities::PhaseFunction<dim> &phase_function,
-                                            MaterialUtilities::PhaseFunctionInputs<dim> &phase_in,
+                                            const std::pair<std::vector<double>, const std::vector<unsigned int>> &gamma_inputs,
```
gassmoeller, line 116: You are right that this interface is not ideal, because it would compute the phase function values several times (once for the equation of state, once for the rheology, and so on). But your alternative is somewhat confusing, because `gamma_inputs` does not explain what it is. What about the following interface:

```
    phase_average_equation_of_state_outputs(const EquationOfStateOutputs<dim> &eos_outputs_all_phases,
                                            const std::vector<unsigned int>> &n_phases_per_composition,
                                            EquationOfStateOutputs<dim> &eos_outputs);
```

Please also update the documentation test above this function.

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,29 @@ namespace aspect
...
+      * function defined for compositions, averaging in this function is calculated base
+      * on gamma functions and the change of variables on the trajectory of phase boundaries.
+      * Thus on a single phase boundary, values of variables change stepwise from one phase
+      * to the other. And this is achieved by the definition of gamma function, which varies
+      * between 0 and 1.
+      */
+      double phase_average_value (const std::pair<std::vector<double>, const std::vector<unsigned int>> &gamma_inputs,
+                                  const std::vector<double> &parameter_values,
+                                  const unsigned int composition,
+                                  bool by_log=false);
```
gassmoeller, line 291: Same comment as above, I would suggest a different interface to make things easier to understand. What about the following:

```suggestion
      double phase_average_value (const std::vector<double> &parameter_values,
                                  const std::vector<double> &phase_function_values, 
                                  const std::vector<unsigned int> &n_phases_per_composition,
                                  const unsigned int composition_index,
                                  const MaterialAveraging::AveragingOperation operation = MaterialAveraging::arithmetic_average);
```

See the comment inside the function about how to use the `MaterialAveraging::AveragingOperation`.

source/material_model/equation_of_state/interface.cc
```diff
@@ -43,47 +43,23 @@ namespace aspect
...
-              const double gamma = phase_function.compute_value(phase_in);
-
-              eos_outputs.densities[c] += gamma * (eos_outputs_all_phases.densities[j]-eos_outputs_all_phases.densities[j-1]);
-              eos_outputs.thermal_expansion_coefficients[c] += gamma * (eos_outputs_all_phases.thermal_expansion_coefficients[j]-eos_outputs_all_phases.thermal_expansion_coefficients[j-1]);
-              eos_outputs.specific_heat_capacities[c] += gamma * (eos_outputs_all_phases.specific_heat_capacities[j]-eos_outputs_all_phases.specific_heat_capacities[j-1]);
-              eos_outputs.compressibilities[c] = gamma * (eos_outputs_all_phases.compressibilities[j]-eos_outputs_all_phases.compressibilities[j-1]);
-              eos_outputs.entropy_derivative_pressure[c] = gamma * (eos_outputs_all_phases.entropy_derivative_pressure[j]-eos_outputs_all_phases.entropy_derivative_pressure[j-1]);
-              eos_outputs.entropy_derivative_temperature[c] = gamma * (eos_outputs_all_phases.entropy_derivative_temperature[j]-eos_outputs_all_phases.entropy_derivative_temperature[j-1]);
-            }
+          eos_outputs.densities[c] = \
```
gassmoeller, line 51: Yes I like this. Much simpler than before. Please remove the `\` though, C++ only ends statements when it encounters an `;` anyway.

source/material_model/utilities.cc
```diff
@@ -699,6 +699,43 @@ namespace aspect
       }
 
 
+      // Calculate a parameter of single composition at single point by averaging phases.
+      // Either linear(by_log is 0) or using log form(by_log is 1), default is linear.
```
gassmoeller, line 703: Remove the comment above. Documentation about the general purpose of this function belongs in the header file.

source/material_model/utilities.cc
```diff
@@ -699,6 +699,43 @@ namespace aspect
...
+      double phase_average_value (const std::pair<std::vector<double>, const std::vector<unsigned int>> &gamma_inputs,
+                                  const std::vector<double> &parameter_values,
+                                  const unsigned int composition,
+                                  bool by_log)
+      {
+        const std::vector<double> &gamma_values = gamma_inputs.first;
+        const std::vector<unsigned int> &n_phases_per_composition = gamma_inputs.second;
```
gassmoeller, line 710: Since you take this parameter apart immediately when you enter the function you can as well hand it over as two separate parameters, which is easier to read.

source/material_model/utilities.cc
```diff
@@ -699,6 +699,43 @@ namespace aspect
...
+        // Calculate base index and assign base value
+        unsigned int base = 0;
+        for (unsigned int i=0; i<composition; ++i)
+          base += n_phases_per_composition[i] + 1;
+        double averaged_parameter = parameter_values[base];
+        if (n_phases_per_composition[composition] > 0)
+          {
+            // Do averaging when there are multiple phases
+            if (by_log)
```
gassmoeller, line 719:
```suggestion
            if (operation == MaterialAveraging::log_average)
```

source/material_model/utilities.cc
```diff
@@ -699,6 +699,43 @@ namespace aspect
...
+              averaged_parameter = log(averaged_parameter);
+            for (unsigned int i=0; i<n_phases_per_composition[composition]; ++i)
```
gassmoeller, line 721: add a newline before the `for` to make clear that line 720 is the only line inside the `if (log)`.

source/material_model/utilities.cc
```diff
@@ -699,6 +699,43 @@ namespace aspect
...
+              {
+                Assert(base+i+1<parameter_values.size(), ExcInternalError());
+                if (by_log)
```
gassmoeller, line 724: same as above

source/material_model/utilities.cc
```diff
@@ -699,6 +699,43 @@ namespace aspect
...
+                  {
+                    // First average by log values and then take the exponential.
+                    // This is used for averaging prefactors in flow laws.
+                    averaged_parameter += gamma_values[base-composition+i] * log(parameter_values[base+i+1] / parameter_values[base+i]);
+                  }
+                else
```
gassmoeller, line 730:
```suggestion
                else if (operation == MaterialAveraging::arithmetic_average)
                   averaged_parameter += gamma_values[base-composition+i] * (parameter_values[base+i+1] - parameter_values[base+i]);
                else
                  AssertThrow(false, ExcInternalError());
```

source/material_model/visco_plastic.cc
```diff
@@ -464,6 +464,13 @@ namespace aspect
 
       std::vector<double> average_elastic_shear_moduli (in.n_evaluation_points());
 
+      // Store value of gamma functions, and number of gamma function for each compostion  # todo
+      // Construct a gamma_inputs structure for passing that information into functions
+      // While number of gamma function is set from the start, value of gamma functions keep updating on every point
+      std::vector<double> gamma_values(phase_function.n_phase_transitions(), 0.0);
+      std::pair<std::vector<double>, const std::vector<unsigned int>> gamma_inputs =
+                                                                     std::make_pair(gamma_values, phase_function.n_phase_transitions_for_each_composition());
```
gassmoeller, line 472:
```suggestion
      // Store value of phase function for each phase and composition
      // While the number of phases is fixed, the value of the phase function is updated for every point
      std::vector<double> phase_function_values(phase_function.n_phase_transitions(), 0.0);
```

source/material_model/visco_plastic.cc
```diff
@@ -485,9 +492,17 @@ namespace aspect
...
+            {
+              phase_inputs.phase_index = j;
+              gamma_inputs.first[j] = phase_function.compute_value(phase_inputs);
+            }
+
+          // Average by value of gamma function to get value of compositions
           phase_average_equation_of_state_outputs(eos_outputs_all_phases,
-                                                  phase_function,
-                                                  phase_inputs,
+                                                  gamma_inputs,
```
gassmoeller, line 505:
```suggestion
                                                  phase_function_values,
                                                  phase_function.n_phase_transitions(),
```

gassmoeller:
Yes I see the problem. `MaterialAveraging` is defined inside `interface.h`, but that is not included in `utilities.h`, because `interface.h` already includes `utilities.h`. So we have a circular dependency here, mostly caused by the fact that we were not careful with the relation between interface and utilities. Solving this is outside the scope of this PR, so lets go a different route: We should probably create a different averaging operation for phases anyway, because it behaves differently than `MaterialAveraging`. Please add the following above `phase_average_value` inside `utilities.h` (this is a modified copy of the `CompositionalAveragingOperation` we already have):
```
      /**
       * For material models with multiple phases:
       * Enumeration for selecting which averaging scheme to use when
       * averaging the properties of different phases.
       * Select between arithmetic and logarithmic.
       */
      enum PhaseAveragingOperation
      {
        arithmetic,
        logarithmic
      };
```

Then you can use `PhaseAveragingOperation` as the type of your `operation` parameter, and `PhaseAveragingOperation::logarithmic`/`PhaseAveragingOperation::arithmetic` inside the function.

source/material_model/utilities.cc
lhy11009: @gassmoeller Hi, Rene, I think I have addressed all of your comments, with expect that I have to name the variable ar phase_transition_arithmetic in PhaseAveragingOperation in order to not conflict with the one defined in CompositionalAveragingOperation, is there a better solution to that?

lhy11009: And right now, I see the first checking failed with the cookbook and some issues with make which I don't quite understand.

lhy11009:
Hi, I commented a week ago. But maybe I should put it in a separate comment instead of putting it under one of the previous comments
@gassmoeller Hi, Rene, I think I have addressed all of your comments, except that I have to name the variable ar phase_transition_arithmetic in PhaseAveragingOperation in order to not conflict with the one defined in CompositionalAveragingOperation, is there a better solution to that?
@gassmoeller, And right now, I see the first checking failed with the cookbook and some issues with make which I don't quite understand.

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,42 @@ namespace aspect
...
+       * Select between arithmetic and logarithmic.
+       */
+      enum PhaseAveragingOperation
+      {
+        phase_average_arithmetic,
+        phase_average_logarithmic
+      };
+
+      /**
+      * For multicomponent material models:
```
gassmoeller, line 284:
```suggestion
```

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,42 @@ namespace aspect
...
+      * Material models compute output quantities such as the viscosity, the
+      * density, etc. For some models, these values may depend on the phase in
+      * addition to the composition, and more than one phase field might have
+      * nonzero values at a given quadrature point. This means that properties
+      * for each composition have to be averaged based on the fractions of each
+      * phase field present. This function performs this type of averaging.
+      * The averaging is based on the choice in @p by_log. Averaging is conducted
```
gassmoeller, line 291:
```suggestion
      * The averaging is based on the choice in @p operation. Averaging is conducted
```

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,42 @@ namespace aspect
...
+      * over the gamma functions of phases given in @gamma_inputs.second, with
```
gassmoeller, line 292:
```suggestion
      * over the phase functions given in @phase_function_values, with
```

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,42 @@ namespace aspect
...
+      * parameter_values stand for values of all the phases. Unlike the average_value
```
gassmoeller, line 293:
```suggestion
      * @parameter_values containing values of all individual phases. Unlike the average_value
```

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,42 @@ namespace aspect
...
+      * function defined for compositions, averaging in this function is calculated base
```
gassmoeller, line 294:
```suggestion
      * function defined for compositions, averaging in this function is calculated based
```

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,42 @@ namespace aspect
...
+      * on gamma functions and the change of variables on the trajectory of phase boundaries.
```
gassmoeller, line 295:
```suggestion
      * on phase functions and the change of variables on the trajectory of phase boundaries.
```

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,42 @@ namespace aspect
...
+      * Thus on a single phase boundary, values of variables change stepwise from one phase
```
gassmoeller, line 296:
```suggestion
      * Thus on a single phase boundary, values of variables change gradually from one phase
```

include/aspect/material_model/utilities.h
```diff
@@ -268,6 +268,42 @@ namespace aspect
...
+      * to the other. And this is achieved by the definition of gamma function, which varies
+      * between 0 and 1.
```
gassmoeller, line 298:
```suggestion
      * to the other. The values of the phase function used to average the properties varies
      * between 0 and 1.
```

lhy11009:
That's okay. Happy movement! It's an event to be celebrated in my culture. I have addressed all your comments. The solution for the enum type does make sense, so there should be one namespace for a type. I have also created a change_log while I am not sure if it is needed.

doc/modules/changes/20200701_haoyuanli
```diff
@@ -0,0 +1,8 @@
+New: Added function phase_average_value, namespace PhaseUtilities to namespace MaterialModel::MaterialUtilities.
+This new function averages parameters(e.g density, viscosity) across all phases for a composition. It
+takes parameters of phases and values of gamma functions as inputs, applys an arithmetic or 
+logarithmic averaging and return the averaged value. A enum type PhaseUtilities::PhaseAveragingOperation 
+is defined to determine the type of operation(e.g arithmetic, logarithmic) to perform in the phase_average_value
+function.
```
gassmoeller, line 6: Thanks for the changelog. We usually try to keep the description here as short as possible, the exact usage of the function can be seen in the documentation of the header file. This is more a note to users that this new functionality exists. Here is how I would change it:
```suggestion
New: Added a new function PhaseUtilities::phase_average_value to the namespace MaterialModel::MaterialUtilities.
This new function averages parameters(e.g density, viscosity) across all phases for a composition.
```

lhy11009:
Yes, that makes sense. I have committed the change to the change_log

PR #3546, opened 2020-06-19
Update world builder version to v0.3.0

MFraters:
This pull request updates the world builder version from 0.2.0 to 0.3.0 (which was just released today!). It also updates version number in the world builder files used in the tests.


MFraters:
Timo's tester give a fast positive error about a maybe initialized value. I can fix it easily, but I am just a bit amazed by it fact that I am currently testing the world builder with gcc version 4.9, 5.5, 6.5, 7.5, 8.1, 9.3 with this flag (Werror=maybe-uninitialized), clang 3.5 (with -Werror=uninitialized) and AppleClang 10.0 (-Werror=uninitialized), and none give warnings for the library, but gcc version 5.4 fails... 

Also, why does Timo's tester fail, even though the CIG tester passes? They both say they have gcc 5.4.

tjhei:
I am not sure why, but could you try setting ``sourceIndex = 0`` or something a few lines above?

MFraters:
I pushed a fix (what you suggest for this parameter and the one below) and I am also testing it with the world builder test suite, but everything seems to compile for me locally. Hopefully this fixes the problem.

MFraters:
It looks like it build fine. 

For the review I would like to point some attentions to two licensing matters. The first one is the inclusion of a small modified subset of the glm library (https://glm.g-truc.net/0.9.9/index.html, code is in contrib/world_builder/include/glm/glm.h) which is under the The Happy Bunny License and MIT License. The second one is an agorithm used for the creating uniform random rotation matrices (see line 128 of for example contrib/world_builder/source/features/continental_plate_models/grains/random_uniform_distribution.cc). This seems to have two licenses, one stating that it is in the public domain and the other only limiting that you can't say that you write the algorithm, but you can use it for anything without grantees. I am sure that these licences allow for the use in the world builder and aspect, but I wanted to make you aware of it anyway.

contrib/world_builder/include/visualization/main.h
```diff
@@ -0,0 +1,32 @@
...
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+void project_on_sphere(double, double &, double &, double &);
```
tjhei, line 20: do you not use header guards?

tjhei:
> I am sure that these licences allow for the use in the world builder and aspect, but I wanted to make you aware of it anyway.

yes, that sounds reasonable to me.

contrib/world_builder/CMakeLists.txt
```diff
@@ -15,51 +23,81 @@ set (WORLD_BUILDER_SOURCE_DIR ${PROJECT_SOURCE_DIR})
...
-elseif(APPLE)
-   SET(CMAKE_CXX_FLAGS "-std=c++11 -Xpreprocessor -fopenmp ${OpenMP_CXX_FLAGS} -pedantic -fPIC -Wall -Wextra -Wpointer-arith -Wwrite-strings -Wsynth -Wsign-compare -Woverloaded-virtual -Wno-literal-range -Wno-parentheses -Wno-unused-local-typedefs -Wcast-qual -fstrict-aliasing -stdlib=libc++ -Wuninitialized -Werror=uninitialized")
-   SET(CMAKE_EXE_LINKER_FLAGS "-lc++${OpenMP_EXE_LINKER_FLAGS}") #-stdlib=libc++")
 else()
-    SET(CMAKE_CXX_FLAGS "/W3")
+  FIND_PACKAGE(Python COMPONENTS Interpreter Development)
+  if(Python_FOUND)
+    include_directories(${Python_INCLUDE_DIRS})
+  else()
+    message(STATUS "Python was not found, disableling python wrapper compilation.")
```
tjhei, line 75:
```suggestion
    message(STATUS "Python was not found, disabling python wrapper compilation.")
```

contrib/world_builder/CMakeLists.txt
```diff
@@ -15,51 +23,81 @@ set (WORLD_BUILDER_SOURCE_DIR ${PROJECT_SOURCE_DIR})
...
 ENDIF(NOT CMAKE_BUILD_TYPE)
 
-IF ( CMAKE_BUILD_TYPE STREQUAL Coverage )
-  SET( CMAKE_VERBOSE_MAKEFILE on )
-  SET(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS} ")
-  SET(CMAKE_CXX_FLAGS_COVERAGE "${CMAKE_CXX_FLAGS_DEBUG} --coverage -fprofile-arcs -ftest-coverage ")
-  SET(CMAKE_Fortran_FLAGS_COVERAGE --coverage)
-  SET(CMAKE_EXE_LINKER_FLAGS_COVERAGE "${CMAKE_EXE_LINKER_FLAGS_DEBUG} --coverage -fprofile-arcs -ftest-coverage ")
-  SET(CMAKE_SHARED_LINKER_FLAGS_COVERAGE "${CMAKE_SHARED_LINKER_FLAGS_DEBUG} --coverage -fprofile-arcs -ftest-coverage ")
+IF(NOT CMAKE_RUN_APP_TESTS)
```
tjhei, line 94: does this need to be enabled inside ASPECT?

contrib/world_builder/include/visualization/main.h
MFraters: I normally do. Since this is a very simple program contained to one file, it didn't have a headerfile to begin with. When I made a headerfile I apparently forgotten to add header guards. This won't influence aspect in any way, since aspect currently doesn't use this file (it is only used when the project itself is compiled in the directory, aspect currently just uses the core library files directly. In the future I would like to make it a bit more integrated through cmake). I will fix it in the next version of the world builder.

MFraters:
Thanks for the comments Timo, they are useful and I will fix them in the next release of the world builder. Aspect currently doesn't use the cmake file of the world builder, but just uses the source and header files directly. It is something I would like to change in the future, to let the aspect cmake call the world builder cmake, and the changes to the cmake files contain some first steps in this direction. The advantage of this integration would be that the tools for easy visualization would be automatically build (if the user wants it) and that the wb tester suit can be automatically run from aspect. But it would require as you suggest more flexibility and more testing of the world builder cmake file. 

It will currently only affect users who go into the directory and run cmake there, but that should again not influence aspect in any way, even if a user does that.

MFraters:
> Looks good to me. This is a gigantic amount of code. :-)

yes, it adds a whole new interface, so instead of just temperature and composition, you can now also ask for grains (sizes and orientations).

This last one is still experimental, but I think really cool in combination with the particle LPO plugin :)

tjhei:
> Aspect currently doesn't use the cmake file of the world builder, but just uses the source and header files directly.

excellent, I guess it doesn't matter than.

PR #3551, opened 2020-06-23
New logo

gassmoeller:
This PR contains a suggestion for a new ASPECT logo, including all files necessary to reproduce the model that creates the logo. The model should probably go into cookbooks instead and we likely only want to merge this after the release to have time to discuss the changes and if we agree to modify all occurences of the logo. I mostly open this PR so that we can reference it when necessary.


gassmoeller:
After our discussion last week I created some new variants of the mesh figure. Here are my suggestions: Version 1 is the one in this PR. Version 3 has the symmetric cutoff (showing both cold and hot features without mesh) and a straight border on the right. Version 4 has the symmetric cutoff and an organic border on the right. 

I favor version 4, what do you think?

Version 1:

logo-white-mesh

Version 3:

logo-white-mesh-3

Version 4:

logo-white-mesh-4

tjhei:
I played around with it and I have some alternative suggestions as I find the mesh too busy in your design:
1. white wireframe mesh overlay (left half):
logo-fine-white
2. colored wireframe (left half):
logo-fine-color
3. coarse mesh (like 1):
logo-coarse-white
4. coarse mesh (like 2):
logo-coarse-color

I did not know about the "feature edges" option in Paraview. Neat.

ljhwang:
I am not feeling the white background on Timo 2 and 4 but do think a coarser mesh is more reasonable at the size this will be printed at.

How would a coarser mesh look on Rene 1 and 3? (what happened to 2?)

Best,
-Lorraine

*****************************
Lorraine Hwang, Ph.D. - Director
UC Davis
530.752.3656
geodynamics.org

jdannberg:
I like Rene's version 1 and 4 best. I just like the gradual transition between mesh and no mesh much better than having a sharp cut. Timo's version 1 and 3 also wouldn't work with a dark background because the mesh is already black. 
But I agree that a coarser mesh is something we may want to consider (that depends on the size of the logo, and for smaller versions we would probably use it without the mesh anyway).

tjhei:
@gassmoeller this might be helpful for reproducing what I have:
notes

bangerth:
I really like the gradient that makes the mesh fade away in the last comment above!

jaustermann:
I like that a lot, too! Sorry I missed the logo discussion today.

gassmoeller:
Wolfgang and Jacky you summarized our discussion excellently without attending! We agreed that we want to have a fading mesh, instead of any hard cutoff or organic cutoff like the earlier variants. @jdannberg and I will try to create 2 or 3 variants of this over the weekend, and we agreed to have another meeting next week to see if we can agree on one of them.

gassmoeller:
I present our new choices for the discussion tomorrow (top row centered mesh, bottom row left-right gradient):

comparison

gassmoeller:
Ok, our discussion this morning confirmed the bottom row is our new logo. I will prepare the pull request next week.

gassmoeller:
I updated the pull request with the final version of our new logo and replaced the use of the old logo in all places in the repo I could find (website still needs to be updated separately). The front page of the manual may need a redesign, because the new logo without ASPECT text has a different aspect ratio, but I would like to postpone that as a follow-up. The current version looks good enough in my opinion. I had to change the color of the top bar of the manual, because light blue does not look good with our new logo. I chose a red that resembles the one in the logo, which is also more in line with the figure CIG uses for the ASPECT software: see https://geodynamics.org/cig/software/aspect/. I attach a screenshot of the new titlepage.

This is ready from my side, but happy to hear opinions.

manual_titlepage

ljhwang:
Looks awesome. 

Did you try the red of the new mantle convection icon?  I do not know we want to go this route for our manuals (matching their domain identity) but it may be worth a try.  I can get the hex from Janice.

Best,
-Lorraine

*****************************
Lorraine Hwang, Ph.D. - Director
UC Davis
530.752.3656
geodynamics.org

gassmoeller:
> Did you try the red of the new mantle convection icon?

No, I just used the one from the logo itself. If you send me a color code I can try it.

gassmoeller:
Here are the two versions. I have a slight preference for the bottom version, because of the stronger contrast, but they are very similar. If you think it is useful to have ASPECT use the CIG mantle convection color, I am ok with that as well.

CIG mantle convection red:

manual_cover_cig

ASPECT logo red:

manual_cover_aspect

bangerth:
I think I prefer the bottom one as well. @ljhwang ?

bangerth:
@gassmoeller -- which version is which?

ljhwang:
Either looks good to me. I would agree with @bangerth on overall aesthetic within the logo itself but still do not know about the design concept of matching the new color palette.

bangerth:
Let's merge the design so that at least that is out of the way. The color would be easy to adjust if we want to.

gassmoeller:
Awesome, happy to have this out of the way! The top one was the CIG color Lorraine sent, the bottom one the color I extracted from the new logo. And yes, changing the color is possible at any point and very easy. When I find a minute this week I will update the website.

PR #3552, opened 2020-06-23
Fix opendap example

### LINKED ISSUE #3418: libdap example is broken
I tried to run the examples introduced by @kneumiller in #3334 but this doesn't work:
```
subsection Initial temperature model
   set Model name = adiabatic boundary
 subsection Ascii data model 
    set Data directory    = 
 set Data file name       = http://test.opendap.org/opendap/BALTO/lithospheric_thickness.csv
end 
```
reports
```
An error occurred in line <136> of file <../source/initial_temperature/adiabatic_boundary.cc> in function
    virtual void aspect::InitialTemperature::AdiabaticBoundary<3>::parse_parameters(dealii::ParameterHandler &) [dim = 3]
The violated condition was: 
    Plugins::plugin_type_matches<const GeometryModel::EllipsoidalChunk<dim>> (this->get_geometry_model())
Additional information: 
    This initial condition can only be used if the geometry is an ellipsoidal chunk.
```

So, "adiabatic boundary" can only be used with "EllipsoidalChunk" but here we use "Chunk". Note that "adiabatic boundary" doesn't even use the "Ascii data model" parameters set above.

I am not sure what is intended here. @kneumiller @dsarahstamps ?

tjhei:
- fix incorrect subsection in .prms
- remove unused data column in data file (this causes the parsing to fail as exactly one data column is expected)

This will also require updating the data file on http://test.opendap.org/

fixes #3418

FYI: @kneumiller @dsarahstamps


tjhei:
@jdannberg is this what you tried as well?

This is the solution I get:
image

jdannberg:
Yes, this is also what I tried (I didn't look at the solution though, and I only changed the txt file). 
I remember last time we discussed which column we should remove (the last or the second-to-last), because we weren't sure which of the two actually was the lithosphere thickness. So it would be great to get some clarification on that from @kneumiller.

gassmoeller:
I opened #3560 to remind us of the issue. I think we should merge this as a running example with wrong data is better than a not running example. Whoever runs the model will want to replace the data anyway.

tjhei:
Sounds good. I wrote to @kneumiller but this can be fixed after the release.

kneumiller:
Updated the prm with a new link to an updated csv file

Issue #3262, opened 2019-10-16
Initial global refinement and maximum/minimum refinement function varies with number of cores


Djneu:
I ran into the following error while trying to run periodic boundary conditions with a maximum/minimum refinement function. 

```
An error occurred in line <405> of file </home/bbpdneu1/software/deal_libraries/tmp/unpack/deal.II-v9.1.1/include/deal.II/lac/affine_constraints.templates.h> in function
    void dealii::AffineConstraints<number>::close() [with number = double]
The violated condition was: 
    dof_index != line.index
Additional information: 
    Cycle in constraints detected!

Stacktrace:
-----------
#0  /home/bbpdneu1/software/deal_libraries/deal.II-v9.1.1/lib/libdeal_II.g.so.9.1.1: dealii::AffineConstraints<double>::close()
#1  /home/bbpdneu1/software/builds/debug_fast/aspect: aspect::Simulator<3>::setup_dofs()
#2  /home/bbpdneu1/software/builds/debug_fast/aspect: aspect::Simulator<3>::run()
#3  /home/bbpdneu1/software/builds/debug_fast/aspect: void run_simulator<3>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, bool)
#4  /home/bbpdneu1/software/builds/debug_fast/aspect: main
```
@anne-glerum and I looked into it briefly and found that the maximum and minimum refinement function works on the initial global refinement even without any adaptive refinements. We also noticed that, depending on the number of cores, this can lead to variable meshes on each side that don't always fit the given constraints. We ran a simple model with all tangential boundaries
 using ASPECT 2.2.0-pre (a856e796c) and deal.II 9.1.1 with the following mesh constraints at 1, 2, 4, and 8 mpi processes. 

```
subsection Mesh refinement
  set Initial global refinement                = 3
  set Initial adaptive refinement              = 0
  set Time steps between mesh refinement       = 0
  set Refinement fraction = 0.5
  set Strategy                                 =  minimum refinement function, maximum refinement function
  
  subsection Maximum refinement function
      set Coordinate system   = cartesian
      set Variable names      = x,y,z
      set Function expression = if(z>=150e3, if(z>=160e3, if(z>=170e3, 3, 2), 1), 0)
  end
  
   subsection Minimum refinement function
      set Coordinate system   = cartesian
[...]

PR #3556, opened 2020-06-25
[WIP] disable mesh smoothing on refine/coarsen

tjhei:
The mesh smoothing algorithms in deal.II change coarsening/refinement
flags but they are inconsistent in parallel leading to inconsistent
refinement that depends on the number of processors used (especially
with very few cells per processor).
Disable this setting until something better is realized in deal.II.

related: #3262

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -197,14 +197,9 @@ namespace aspect
     triangulation (mpi_communicator,
                    (parameters.stokes_solver_type == Parameters<dim>::StokesSolverType::block_gmg
                     ?
-                    typename Triangulation<dim>::MeshSmoothing
-                    (Triangulation<dim>::smoothing_on_refinement |
-                     Triangulation<dim>::smoothing_on_coarsening |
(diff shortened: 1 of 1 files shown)


tjhei:
This certainly needs a test, but I want to see if this breaks anything at this point.

tjhei:
The motivation is the prm in #3262 that generates the following mesh with 10 ranks:
image

Notice the one cell that should be refined (and is if you run with <10 ranks).

tjhei:
This breaks a lot of tests as the refinement is slightly different. I would still want to proceed with this change. The question is if we should delay this until after 2.2?

gassmoeller:
I would prefer to postpone changes to behaviour until after the release unless they are clear bug fixes.

tjhei:
> unless they are clear bug fixes

It is a clear bug (minimum refinement asks for refinement everywhere but it won't happen if one of the processors has only one cell), but I agree that this is quite a disruptive change.

Okay, will delay this and #3262 then.

tjhei:
We should go with #3595 instead.

PR #3557, opened 2020-06-25
add GWB and Newton to dynamic citation page

tjhei:


tjhei:
FYI @MFraters

MFraters:
Thanks! The newton solver entry is a duplicate of entry FBTGS18 (line 11499 in manual.bib), but since the new one is more complete I think it is better to delete that one.

tjhei:
I took the entry from citing_aspect.bib. I will clean that up.

tjhei:
maybe like this?

MFraters:
Yes, I think that is a good way to make it consistent. Just one thing, why did you use "FBTGS18" instead of "FBTGS19"?

tjhei:
> why did you use "FBTGS18" instead of "FBTGS19"?

Because this is what is already used in the manual. I did not create that entry.

tjhei:
I renamed the citation.

PR #3568, opened 2020-06-29
adding the two missing files of the viscosity_grooves benchmark

cedrict:
Pull Request Checklist. Please read and check each box with an X. Delete any part not applicable. Ask on the [forum](https://community.geodynamics.org/c/aspect) if you need help with any step.

re added two missing files that mysteriously disappeared off the original PR. 

### Before your first pull request:

* [ ] I have read the guidelines in our [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) document.

### For all pull requests:

* [ ] I have followed the [instructions for indenting my code](../blob/master/CONTRIBUTING.md#making-aspect-better).

### For new features/models or changes of existing features:

* [x] I have tested my new feature locally to ensure it is correct.
* [ ] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/master/tests/) directory.
* [ ] I have added a changelog entry in the [doc/modules/changes](../blob/master/doc/modules/changes) directory that will inform other users of my change.

PR #3569, opened 2020-06-29
update viscosity_grooves benchmark

tjhei:
- use double numbers
- update copyright years
- use compute_global_error
- enable in automatic check

followup to #3568


benchmarks/check.mk
```diff
@@ -173,6 +173,3 @@ compressibility_formulations/: dummy
 	@$(def); run_prm $@/lateral_pipe_transient lateral_pipe.prm
 	@$(def); run_prm $@/vertical_pipe vertical_pipe.prm
 
-#https://github.com/geodynamics/aspect/issues/3534
```
gassmoeller, line 176: Do you need to replace this with something that runs the benchmark?

tjhei: No, because the default rule will compile and run all prms.

Issue #3585, opened 2020-07-03
modernize unity builds


tjhei:
The current way to do unity builds using COTIRE is only supported for cmake version 3.8 and older. This makes the feature a lot less useful and more difficult to test (the 9.2 tester has a newer cmake version, see #3583 ).
I would suggest:
a) remove cotire
b) investigate cmake unity builds [1], [2]

[1] https://cmake.org/cmake/help/v3.18/prop_tgt/UNITY_BUILD.html
[2] https://onqtam.com/programming/2019-12-20-pch-unity-cmake-3-16/

I looked into enabling ``CMAKE_UNITY_BUILD``, but this doesn't have any effect...

gassmoeller:
I agree that we need to replace cotire (it has reached end of support anyway, https://github.com/sakra/cotire#cotire). Ideally replacing by native cmake precompile headers and unity build sounds best, but it has this some limitations (https://cmake.org/cmake/help/v3.16/prop_tgt/UNITY_BUILD.html):

```
Any source file that has a non-empty COMPILE_OPTIONS, COMPILE_DEFINITIONS, COMPILE_FLAGS, or INCLUDE_DIRECTORIES source property will not be combined into a unity source.
```
As far as I can tell we use these a lot for our files (maybe I misunderstand how we use them vs what this limitation is). It seems the fix would be to create a configuration file that defines the necessary macros and is created by cmake during configuration? Like we do for revision.h.in in include/aspect or in world_builder_config.cc.

At least precompiling headers is already possible and gives some real benefit:

Without precompile: make -j12

real    7m54.763s
user    85m49.481s
sys     4m25.993s

With precompile global.h: make -j12
CMakeLists.txt:470: `TARGET_PRECOMPILE_HEADERS(aspect PRIVATE <aspect/global.h>)`

real    5m26.267s
user    54m37.578s
sys     3m43.557s

With precompile simulator_access.h: make -j12
CMakeLists.txt:470: `TARGET_PRECOMPILE_HEADERS(aspect PRIVATE <aspect/simulator_access.h>)`

real	3m2.929s
user	24m40.395s
sys	2m21.270s

Unfortunately precompiling simulator_access.h breaks if `ASPECT_WITH_WORLD_BUILDER` is on:
[...]

PR #3583, opened 2020-07-02
set required deal.II version to 9.2.0, update testers

tjhei:
- docker image in contrib/ci/ changed to 9.2.0
- require 9.2.0 in CMakeLists

part of #3582

PR #3587, opened 2020-07-03
Use native cmake to precompile headers

gassmoeller:
See the discussion in #3585. We could also enable this unconditionally as long cmake supports it. Or we keep it like this for now and enable it by default later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -513,6 +513,8 @@ IF(CMAKE_VERSION VERSION_LESS 3.9)
   SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "Precompile external header files using cotire to speedup compile time. Currently only supported for cmake versions older than 3.9 or newer than 3.15.")
 ELSEIF(CMAKE_VERSION VERSION_GREATER 3.15)
   SET(ASPECT_PRECOMPILE_HEADERS ON CACHE BOOL "Precompile external header files to speedup compile time. Currently only supported for cmake versions older than 3.9 or newer than 3.9.")
+ELSE()
+  SET(ASPECT_PRECOMPILE_HEADERS OFF)
 ENDIF()
 
 IF (ASPECT_PRECOMPILE_HEADERS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -423,19 +423,6 @@ IF (ASPECT_HAVE_LINK_H)
   MESSAGE(STATUS "Enabling checking of compatible deal.II library when loading plugins")
 ENDIF()
 
-# Check if we want to precompile header files. This speeds up compile time,
-# but can fail on some machines with old cmake, therefore it is disabled by
-# default. Also cotire currently has issues with cmake generator expressions
-# used in deal.II for cmake version 3.9 and above. Disable cotire for cmake >=
-# 3.9. Starting with cmake 3.15 there is native support inside cmake and we can
-# use that.
-IF(CMAKE_VERSION VERSION_LESS 3.9)
-  SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "Precompile external header files using cotire to speedup compile time. Currently only supported for cmake versions older than 3.9 or newer than 3.15.")
-ELSEIF(CMAKE_VERSION VERSION_GREATER 3.15)
-  SET(ASPECT_PRECOMPILE_HEADERS ON CACHE BOOL "Precompile external header files to speedup compile time. Currently only supported for cmake versions older than 3.9 or newer than 3.9.")
-ENDIF()
-
-
 ADD_EXECUTABLE(${TARGET} ${TARGET_SRC})
 DEAL_II_SETUP_TARGET(${TARGET})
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -467,6 +467,12 @@ ENDIF()
 ADD_EXECUTABLE(${TARGET} ${TARGET_SRC})
 DEAL_II_SETUP_TARGET(${TARGET})
 
+IF (ASPECT_PRECOMPILE_HEADERS)
+  IF(CMAKE_VERSION VERSION_GREATER_EQUAL 3.16)
```
tjhei, line 471: I am not sure I understand. Wouldn't you run into the FATAL_ERROR 6 lines above?

CMakeLists.txt
```diff
@@ -467,6 +467,12 @@ ENDIF()
...
+    TARGET_PRECOMPILE_HEADERS(aspect PRIVATE <aspect/global.h> <aspect/introspection.h> <aspect/parameters.h> <deal.II/particles/particle_handler.h> <deal.II/base/table_handler.h> <deal.II/base/timer.h> <deal.II/base/conditional_ostream.h> <deal.II/distributed/tria.h> <deal.II/dofs/dof_handler.h> <deal.II/fe/fe.h> <deal.II/fe/mapping_q.h>)
```
tjhei, line 472: is particle_handler useful to have in here?

gassmoeller:
You are right, I should test the final version before pushing (it also didnt work with old cmakes, e.g. 3.5 like on the tester), should be fixed now (I tested with cmake 3.5.1, 3.8.2 and 3.16.3).

Having the particle handler header in there saves about 20 s (it is usually included in `simulator_access.h` via `aspect/particle/world.h`, but that one includes all kind of stuff that should not be in the precompiled headers). 
Additionally, excluding the world builder files from including the precompiled headers saves about 15 s.

CMakeLists.txt
```diff
@@ -458,6 +460,8 @@ IF (ASPECT_PRECOMPILE_HEADERS)
     FIND_FILE(HELPER_PATH helper_functions.cc HINTS "${CMAKE_SOURCE_DIR}/source/simulator")
     LIST(REMOVE_ITEM TARGET_SRC ${HELPER_PATH})
     LIST(INSERT TARGET_SRC 0 ${HELPER_PATH})
+  ELSEIF(CMAKE_VERSION VERSION_GREATER 3.15)
+    # Do nothing, this case is handled below
   ELSE()
     MESSAGE(FATAL_ERROR "ASPECT_PRECOMPILE_HEADERS is currently only supported for CMake version 3.8 or less.")
```
tjhei, line 466: can you update this error message please?

CMakeLists.txt
```diff
@@ -144,13 +144,15 @@ if(ASPECT_WITH_WORLD_BUILDER)
   # set ASPECT_ENABLE_WORLD_BUILDER define for all source files:
   FOREACH(_source_file ${wb_files})
     SET_PROPERTY(SOURCE ${_source_file} PROPERTY COTIRE_EXCLUDED TRUE ) 
+    SET_PROPERTY(SOURCE ${_source_file} PROPERTY SKIP_PRECOMPILE_HEADERS TRUE )
```
tjhei, line 147: why do we need to skip this? Can you leave a comment?

CMakeLists.txt
```diff
@@ -467,6 +471,10 @@ ENDIF()
 ADD_EXECUTABLE(${TARGET} ${TARGET_SRC})
 DEAL_II_SETUP_TARGET(${TARGET})
 
+IF (ASPECT_PRECOMPILE_HEADERS AND CMAKE_VERSION VERSION_GREATER 3.15)
+  TARGET_PRECOMPILE_HEADERS(aspect PRIVATE <aspect/global.h> <aspect/introspection.h> <aspect/parameters.h> <deal.II/particles/particle_handler.h> <deal.II/base/table_handler.h> <deal.II/base/timer.h> <deal.II/base/conditional_ostream.h> <deal.II/distributed/tria.h> <deal.II/dofs/dof_handler.h> <deal.II/fe/fe.h> <deal.II/fe/mapping_q.h>)
```
tjhei, line 475: would you mind breaking this into one file per line? Maybe add a comment with the reasoning what to include and what to leave out?

gassmoeller:
I think I addressed all issues and also move the declaration of the cmake variable to the place where it is used (had to be earlier in the file before we decided to unconditionally reorder helper_functions.cc in #3591). After the results discussed in #3585 I also decided to make precompiling the headers the default for cmake >3.15, do you agree? Ready from my side.

CMakeLists.txt
```diff
@@ -513,24 +503,53 @@ CONFIGURE_FILE(
...
+# but can fail on some machines with old cmake, therefore it is disabled by
+# default. Also cotire currently has issues with cmake generator expressions
+# used in deal.II for cmake version 3.9 and above. Disable cotire for cmake >=
+# 3.9. Starting with cmake 3.15 there is native support inside cmake and we can
+# use that.
+IF(CMAKE_VERSION VERSION_LESS 3.9)
+  SET(ASPECT_PRECOMPILE_HEADERS OFF CACHE BOOL "Precompile external header files using cotire to speedup compile time. Currently only supported for cmake versions older than 3.9 or newer than 3.15.")
+ELSEIF(CMAKE_VERSION VERSION_GREATER 3.15)
+  SET(ASPECT_PRECOMPILE_HEADERS ON CACHE BOOL "Precompile external header files to speedup compile time. Currently only supported for cmake versions older than 3.9 or newer than 3.9.")
+ENDIF()
```
tjhei, line 516: We should declare this variable in all 3 cases. Can you add an else with ``SET(... OFF)``?

CMakeLists.txt
```diff
@@ -513,24 +503,53 @@ CONFIGURE_FILE(
...
-  # files that would otherwise interfere with each other.
-  SET_TARGET_PROPERTIES (aspect PROPERTIES
-      COTIRE_PREFIX_HEADER_IGNORE_PATH
-      "/usr/include/;${CMAKE_SOURCE_DIR};${CMAKE_BINARY_DIR}")
-
-  # This line activates the cotire module for the aspect target, 
-  # and therefore the whole precompilation
-  cotire(aspect)
+  IF (CMAKE_VERSION VERSION_LESS 3.9)
+    MESSAGE(STATUS "Enabling cotire to precompile external header files.")
```
tjhei, line 520: I would suggest we get rid of cotire entirely. It is too complicated to make sure all 3 variants work (and I don't see how we can test this on jenkins). We can delay this until we bump to deal.II 9.2 if you prefer.

tjhei:
> decided to make precompiling the headers the default

sounds good.

We need to (separately from this PR) figure out how we want to test using Jenkins. Maybe disable precompiled headers there?

gassmoeller:
> I would suggest we get rid of cotire entirely. It is too complicated to make sure all 3 variants work (and I don't see how we can test this on jenkins). We can delay this until we bump to deal.II 9.2 if you prefer.

I agree, but lets do that as follow-up to #3583, when we update the testers. At the moment it is still using cotire.

> We need to (separately from this PR) figure out how we want to test using Jenkins. Maybe disable precompiled headers there?

Currently we run a precompiled header build and a unity build. I would suggest afterwards we run one unity build with precompiled headers (which hopefully catches most/all errors for unity builds and precompiled headers individually), and one build without both. We could split this into your tester vs CIG testers? Your tester is typically faster, so maybe you compile without unity/precompile? For a full test we would need to check all 4 variations.

tjhei:
> At the moment it is still using cotire.

okay, I updated  #3583 to use the new unity build because cotire won't work there (newer cmake).

We can delete cotire later.

PR #3591, opened 2020-07-05
unity build fixes

tjhei:
Necessary to make CMAKE_UNITY_BUILD work. I am not quite sure why explicit instantiations for ``Simulator`` and ``AdvectionField`` need to be in the same file, but this is the only way I could make it work with and without unity build.

related to #3585


tjhei:
It seems to work.

PR #3595, opened 2020-07-07
communicate refinement flags on ghost cells

tjhei:
alternative to #3556 to make refinement independent of number of
processors as suggested by @bangerth .
Related to  #3262


gassmoeller:
Do you prefer to move forward with this PR or #3556? The code seems fine to me.

tjhei:
> Do you prefer to move forward with this PR

I think this PR is the better choice as it has smaller impact (smaller number of tests change).

I fear that there are other situations, especially with periodic boundaries, that might not be fixed with this, but this is a big step forward.

PR #3596, opened 2020-07-09
[Discussion] Include tracking of configuration data.

gassmoeller:
This is an implementation of something we talked about in January at the ASPECT virtual user meeting and that I originally saw when reviewing the Underworld2 code (https://github.com/underworldcode/underworld2). 

It would be of great use for us (as a community) to have some idea about the number of ASPECT users and the versions of ASPECT and Deal.II in use. This would also help us (or anyone writing proposals using ASPECT actually) to prove ASPECTs impact and sustainability. The easiest way to achieve this is to track some limited usage information via Google Analytics. We are all aware of the privacy issues connected to this, but if I recall our discussion correctly, users would generally be fine providing this information as long as we limit it to what is strictly necessary, and avoid tracking anything that could be considered sensitive. Below is my shot at implementing this including options to opt-out and a policy explicitly stating what we do. I attach an example for the type of report we could get from this data (created using this branch of ASPECT on my laptop, the data is somewhat inconsistent because I successively added features). I think this would be very useful for annual reporting, writing proposals, and discussing dropping support for older versions and would be in favour of moving forward with it, but I am open for discussions and counter arguments.

We should not merge this before we had another discussion about it in the forum, during the online user meetings, and maybe during the hackathon, but I would be happy to get some feedback already. Good idea? Bad idea? Horrible idea? :-)

I would propose to make all reports we get from this publicly available (e.g. during our release on our website) to be transparent about what information was collected (and for promotional purposes :-)).

[Analytics All Web Site Data ASPECT use 20200707-20200709.pdf](https://github.com/geodynamics/aspect/files/4899337/Analytics.All.Web.Site.Data.ASPECT.use.20200707-20200709.pdf)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,6 +143,11 @@ if(ASPECT_WITH_WORLD_BUILDER)
   SET_PROPERTY(SOURCE "${CMAKE_BINARY_DIR}/world_builder_config.cc" PROPERTY SKIP_PRECOMPILE_HEADERS TRUE )
 endif()
 
+SET(ASPECT_TRACK_USAGE_DATA ON CACHE BOOL "Whether to allow ASPECT to transmit version information and basic usage data. See PRIVACY.md for detailed information on the collected data.")
+
+IF(ASPECT_TRACK_USAGE_DATA)
+  INCLUDE_DIRECTORIES("${CMAKE_SOURCE_DIR}/contrib/cpp-httplib/include/")
+ENDIF()
 
 # load in version info and export it
 FILE(STRINGS "${CMAKE_SOURCE_DIR}/VERSION" ASPECT_PACKAGE_VERSION LIMIT_COUNT 1)
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,6 +82,7 @@ pipeline {
         -D ASPECT_USE_PETSC='OFF' \
         -D ASPECT_RUN_ALL_TESTS='ON' \
         -D ASPECT_PRECOMPILE_HEADERS='ON' \
+        -D ASPECT_TRACK_USAGE_DATA='OFF' \
         ..
         '''
 
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -76,6 +76,7 @@ pipeline {
         cmake \
         -D ASPECT_USE_PETSC='OFF' \
         -D ASPECT_PRECOMPILE_HEADERS='ON' \
+        -D ASPECT_TRACK_USAGE_DATA='OFF' \
         -D COTIRE_MAXIMUM_NUMBER_OF_UNITY_INCLUDES='100' \
         ..
         '''
@@ -105,6 +106,7 @@ pipeline {
         -D ASPECT_USE_PETSC='OFF' \
         -D ASPECT_RUN_ALL_TESTS='ON' \
         -D ASPECT_PRECOMPILE_HEADERS='ON' \
+        -D ASPECT_TRACK_USAGE_DATA='OFF' \
         ..
         '''
 
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -1,0 +1,52 @@
+Privacy Policy
+==============
+
+We as the developers of the ASPECT project collect basic information about the
+use of ASPECT using Google Analytics. This information is of great help for us
+to further develop and improve ASPECT, e.g. to make decisions about versions in
+use and backwards compatibility. It also allows us to justify ASPECT's impact
+to our funding agencies, e.g. by estimating the size of our user community,
+which is notoriously difficult for open-source community projects like ASPECT.
+We are aware of the justified concerns about user privacy and therefore adopt
+the following policies: 
+
+We collect **only** the following information, which is also displayed on
+screen when ASPECT starts:
+
+* ASPECT version number (excluding git branch or commit hash),
+* Version number of the main dependencies (deal.II, trilinos, p4est),
+* Limited compile-time configurations (build type and vectorization level),
+* Number of MPI processes used.
+
+All source code collecting this information can be found in 'source/main.cc' in
+the function 'track_usage()'.
+
+Google Analytics also automatically collects an anonymized version of your IP
+for geolocational purposes. This IP can not be traced back to you, but does
+allow the identification of your country and potentially your city. While this
+is not our intent, we cannot prevent the collection of this data.
+
+
+Participation
+-------------
+
+You may opt out of this tracking by setting the cmake variable
+'ASPECT_TRACK_USAGE_DATA=OFF' during configuration, which will disable all
+tracking code. We ask that you only do so if you have serious concerns about
+the collected data (in which case please also send us an email to discuss your
+concerns, we are open to improvements). Providing this data is a valuable
+contribution that supports ASPECT's development.
+
+
+Policy
+------
+
+1. We respect the privacy of our users and will never collect information
+   that can be used to identify individual users or that can be used to
+   reconstruct the intent of your use of ASPECT.
+
+2. We collect information only where this is necessary for the above specified
+   purpose.
+
+3. We will only use this information for the purpose for which it was 
+   collected.
--- a/contrib/cpp-httplib/LICENSE
+++ b/contrib/cpp-httplib/LICENSE
@@ -1,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) 2017 yhirose
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
--- a/contrib/cpp-httplib/include/cpp-httplib/httplib.h
+++ b/contrib/cpp-httplib/include/cpp-httplib/httplib.h
@@ -1,0 +1,5823 @@
+//
+//  httplib.h
+//
+//  Copyright (c) 2020 Yuji Hirose. All rights reserved.
+//  MIT License
+//
+
+#ifndef CPPHTTPLIB_HTTPLIB_H
+#define CPPHTTPLIB_HTTPLIB_H
+
+/*
+ * Configuration
+ */
+
+#ifndef CPPHTTPLIB_KEEPALIVE_TIMEOUT_SECOND
+#define CPPHTTPLIB_KEEPALIVE_TIMEOUT_SECOND 5
+#endif
+
+#ifndef CPPHTTPLIB_KEEPALIVE_TIMEOUT_USECOND
+#define CPPHTTPLIB_KEEPALIVE_TIMEOUT_USECOND 0
+#endif
+
+#ifndef CPPHTTPLIB_KEEPALIVE_MAX_COUNT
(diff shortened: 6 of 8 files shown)


tjhei:
I am not sure yet how I feel about this, but let's have a conversation about it.
1. This will run on every invocation of ASPECT, which generates a large number of events. While neat, this is more then what we actually want (to see how many users we have). More importantly, we won't get any measure to know how many users we have because there is no identifying information or am I wrong? How does underworld do it? Do they have any of there results available online?
2. Is this using http or https?
3. What happens if networking is disabled or ips are blocked, does it fail silently?
4. Do you accidentally run this on every MPI rank right now? Good luck running this with 100,000 MPI ranks.

gassmoeller:
> I am not sure yet how I feel about this, but let's have a conversation about it.

Sure I am open to discussions.

> This will run on every invocation of ASPECT, which generates a large number of events. While neat, this is more then what we actually want (to see how many users we have). More importantly, we won't get any measure to know how many users we have because there is no identifying information or am I wrong? How does underworld do it? Do they have any of there results available online?

I dont think Google cares about the number of events, and I thought for us having the total number of runs might actually be more useful than the number of unique users. Underworld generates an anonymous but unique identifier when compiling the code and then uses that to track individual installations (also tracking every time the library is loaded). I felt that that invades the privacy more than collecting every single run without identifying features. Changing from one to the other is relatively simple (just need a way to come up with the unique identifier and store that in config.h during cmake configuration).

> Is this using http or https?

Not sure actually, but the library I included can do both. I can look into it.

> What happens if networking is disabled or ips are blocked, does it fail silently?

Yes, if there is no network nothing happens and the model runs as normal. I have not checked what happens on clusters behind a firewall yet.

> Do you accidentally run this on every MPI rank right now? Good luck running this with 100,000 MPI ranks.

It is inside ` if (i_am_proc_0)`, so no it is only run once.

tjhei:
> It is inside ` if (i_am_proc_0)`, so no it is only run once.

oh, sorry. I missed that somehow.

MFraters:
I think it is very cool and useful, but I do have a few worries and comments, which I would like to give before the discussion we will probably have during the hackaton.

On the collecting side, I think the compiler version would be useful for deciding how fast (or slow) we can upgrade to a new cpp version and what compilers are generally used. 

On the privacy side, I am worried about using google. If I understand correctly, we basically want it to fill in an excel sheet based on a http request which does not have to return any data. Although I don't know how many models are run per day, but even if it are a hundred or even a thousand, I am quite sure this will be less intensive  and less  less data then a typical single website request would be. If this is what we want it could be for example as easy as a php script similar to this (with the database value in a different file of course, located outside the www directory of the server):

``` php
<?php
// check if values match with hash of values and if request if valid
if(strlen($_POST["hash"]) > 8 && $_POST["hash"]=="$_POST["year"]+$_POST["month"]+$_POST["day"]+...)
{
// we can do more checking on individual values.
try {
  $conn = new PDO("mysql:host=$servername;dbname=$dbname", $username, $password);
  // set the PDO error mode to exception
  $conn->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);

  // prepare sql and bind parameters. This is sanitizing the data so no sql injection can occur, 
  //see for example: http://wiki.hashphp.org/PDO_Tutorial_for_MySQL_Developers
  $stmt = $conn->prepare("INSERT INTO MyGuests (year,month,day, aspect_version_major,  aspect_version_minor,  aspect_version_fix, compiler, cpp_version etc..)
  VALUES (:year,:month,:day, :aspect_version_major, :aspect_version, :compiler, :cpp_version)");
  $stmt->bindParam(':year', $year);
  $stmt->bindParam(':month', $month);
  $stmt->bindParam(':day', $day);
  $stmt->bindParam(':aspect_version_major', $aspect_version_major);
  $stmt->bindParam(':aspect_version_minor', $aspect_version_minor);
  $stmt->bindParam(':aspect_version_fix', $aspect_version_fix);
  $stmt->bindParam(':compiler', $compiler);
  $stmt->bindParam(':cpp_version', $cpp_version);
// etc.

  // turn data into non-string data like numbers for years,months,days, version number, etc, and enums for for example compilers.
  // there is no need to store stings, so just turn everything into int's.
  $year = (int) $_POST["year"];
  $month = (int) $_POST["month"];
  $day = (int) $_POST["day"];
  $aspect_version_major = (int) $_POST["aspect_version_major"];
  $aspect_version_minor = (int) $_POST["aspect_version_minor"];
  $aspect_version_fix = (int) $_POST["aspect_version_fix"];
  switch ($_POST["compiler"]) { // this case can also be done within aspect already
    case "gcc": 
      $compiler = 1;
      break;
    case "clang": 
      $compiler = 2;
    default:
      $compiler = 0;
}
  $cpp_version = (int) $_POST["cpp_version"]; // or this could also be a switch statment.
  // etc.

  $stmt->execute();

  //echo "New records created successfully";
} catch(PDOException $e) {
  //echo "Error: " . $e->getMessage(); // We could also (try to) create a separate record for errors if we want to.
}
$conn = null;
}
?> 
```

I am **not** saying that we should literally use this code, or even php as a language. I know that there are downsides to this approach. It would take more time and we would need to host it ourselves. I do think it could be done safely on a vps or maybe in a similar docker container as jenkins (see for example https://www.digitalocean.com/community/tutorials/how-to-install-linux-nginx-mysql-php-lemp-stack-ubuntu-18-04, plus adding a fail2ban, this is also what I have been doing for my own webserver, but without mysql and php). Also, maybe there are frameworks which can already do this.  But the advantages of it are that all the involved code can be open source both client and server side (expect the passwords), only what we thing is relevant is stored and is controlled by us (CIG?), and we could publicly show the database (with a day granularity). This would provide the openness I think is needed in this case.

The list of entries I would think are useful to collect for the purpose of knowing what minimum version we can require are:

1. year of run (int)
2. month of run (int)
3. day of month of run (int), although we could even do without this.
4. aspect version (3 ints)
5. deal.ii version (3 int)
6. trilinos version (3 ints)
7. p4st version (3 ints)
8. cpp version (1 int?)
9. cpp compiler (1 enum int, with 0 is other)
10. a flag for features of aspect which are enabled/used (like for the references)
11. OS and OS verison?

I deliberately didn't include a timestap or IP adress/geolocation. I don't think we need them for the specified purpose, and then we would be storing data which would potentially identifiable to the user. A server would temporarily store ip addresses in their logs, but these are rotating files, so even in the worst case only the most recent ip addresses could be possibly compromised, and the data in the database which is already public and doesn't contain personal identifiable data. Furthermore, it is easy in nginx to set how long to keep the logs (https://stackoverflow.com/questions/44180125/keep-the-nginx-logs-of-the-last-30-days).

It has become a bit longer post then expected, but I wanted to add this to the discussion, so that we can consider the option.

MFraters:
I coudn't help myself from thinking a bit more about this problem overnight, and I came up with some solutions to potential problems like someone trying to fill the database with a huge amount of nonsence data (unlikely case, but still). _I would like to stress again that I recognise that this might not be what to do in the end, and even if we would want to go this route it would require further thinking, but I would like the option to be on the table._ 

Thinking about what we want to record is (I think, but this can be discussed) only when a user on a given day uses aspect and in what configuration. So if a user runs the same model a hundred times, we ideally would only have one entry. But if the user runs it with for example a different compiler that day, I think it would still be useful to record that. This could be a potential privacy risk, but I don't think it has to be. 

The solution to this problem would be to have (at least) two tables in the database. One table contains a index, a [salted](https://en.wikipedia.org/wiki/Salt_(cryptography)) ip adress, the salt value, and an index of the entry it relates to. We won't need a date or time in this table, as I will show later. The second table would contain an index, date (server side for consistency, and no time) and all the run information we would like to store.

In short, a hash in this case is a one way function, which maps a input value (ip address) to an other (output) value in a non-unique way, so that input values close to each other generate very different output values. With non-unique I mean that every input value would generate a unique output hash, but the same output hash can theoretically be generated by multiple different input values. Adding a unique salt value basically prevents the creation of useful [rainbow tables](https://en.wikipedia.org/wiki/Rainbow_table). So there would be no way of relating an ip address to a specific user. This method is mainly used for safe password storage, but I think it will also help with this problem. On top of that the table with the salted ip addresses is automatically emptied every night, because we don't need that data any more for the purpose we wanted to use it for.

It is then very easy to check whether the hash of the IP already exists (if not, add a new entry if the data is valid), or if it exists, whether the values are the same as the entries stored before (that day) through the ip hash. If the values are different a new entry is created.

P.S.
If we wanted to be really secure about the ip adresses, we would even add [pepper](https://en.wikipedia.org/wiki/Pepper_(cryptography)) to the salt ;)

gassmoeller:
Thanks for all the thought Menno, I think all of these options should factor into our discussion during the hackathon. I would like to point out that the main reason I was using Google Analytics (besides being widespread and easy to implement) is that it is safe, both in terms of cybersecurity and legally, i.e. if we use the anonymized IP and transmit no information that can be traced to a single user we only need a very simple privacy policy and we do not need to worry much about privacy laws of different countries. I am very concerned that any form of self-written solution is actually much worse in terms of effort, maintenance, security, and legal complications. If someone claims our privacy policy for Google analytics is not good enough we can always update it. But if someone claims using Google Analytics is against their countries' law we can easily refute that claim, while we can not do so with a homegrown solution. We may feel that a homegrown solution is safer because *we* control it, but I am much more comfortable saying Google knows what it is doing in terms of cyber security and privacy laws than of my own understanding of it. I share a diffuse dislike of Google's privacy practices, but in terms of concrete legal implementations I trust their lawyers over my gut feelings.

As for the type of data collected: I think the compiler type and version would be a fair addition. I originally debated including the citation string, but felt that it comes too close to what most of our users would be sensitive about, i.e. the type of model they are running with ASPECT. I would prefer to only include compile-time configuration data, except for the MPI ranks (which do not say much about the type of model run). Of course we can debate if we need every run-time event, or if we could get away with sending information during compile time (removing the MPI rank information). I am open for discussions.

MFraters:
Yes, I agree that there are large advantages to using google analytics, mainly with ease of use an maintainability and it being free to use for us. 

> I would like to point out that the main reason I was using Google Analytics (besides being widespread and easy to implement) is that it is safe, both in terms of cybersecurity and legally, i.e. if we use the anonymized IP and transmit no information that can be traced to a single user we only need a very simple privacy policy and we do not need to worry much about privacy laws of different countries.

I don't think that is a correct assumption, but please correct me if I am wrong here. Doing a quick search it look to me like what google lawyers have doing is simply shift the responsibility to the user of their products (us). See for example [this site](https://www.privacypolicies.com/blog/privacy-policy-google-analytics/) for some explication or [this site](https://termly.io/resources/articles/google-analytics-privacy-policy/), which allows you to build a privacy policy for google analytics.  So we would still need to create a full privacy policy, which in the case of using google analytics would be a lot more complicated. So I doubt the current proposed privacy is compliant with for example EU or Californian privacy laws, but then again, I could very well be wrong. What did you use to make the privacy policy?

I agree that strictly speaking security wise google is much better than we could ever be, but I haven't delved into their privacy policy enough to know what they are collecting (even if we aren't collecting it on purpose, or are able to see it). So it **could** be that we would be giving a lot a of data to a third party, including identifiable data. So that would be something we would need to figure out and put into the privacy policy before we could use this. But even then, if we would go ahead with this approach, I would be much more comfortable with an opt in, while an option which doesn't require a third party I would be fine with an opt out.

Just an other note on security and maintainability, we also run our own jenkins and discourse servers, which also requires some maintenance. If we use some kind of sandbox with only this webserver and database I think the the maintenance burden and security risks are comparable, if not lower, then for those severs.

If we would want something which is not homegrown, we could also take a look at [open source google analytics alternatives](https://opensource.com/article/18/1/top-4-open-source-analytics-tools). This might be the best of both worlds, but would probably still require an extensive privacy policy since they would probably (try to) collect a lot more data since they are made for web analytics.

> As for the type of data collected: I think the compiler type and version would be a fair addition. I originally debated including the citation string, but felt that it comes too close to what most of our users would be sensitive about, i.e. the type of model they are running with ASPECT. I would prefer to only include compile-time configuration data, except for the MPI ranks (which do not say much about the type of model run). Of course we can debate if we need every run-time event, or if we could get away with sending information during compile time (removing the MPI rank information). I am open for discussions.

hmm, I do not have a strong opinion on this. I think it is useful to have information on MPI and what features are being used, and I don't think that just knowing that a feature is being used really tell anything about the model being run. We could make it a separate opt out? But I would be fine with it being in there.

gassmoeller:
Let's move that discussion to an in person discussion at some time (it takes much time), but in short I think most online guides do not apply to us, because they discuss using Google Analytics on *websites*. On websites google analytics automatically assigns unique user IDs to visitors and cross-references the browser with known accounts, and uses cookies on the users computers, which require all of the extended privacy policies (and a cookie policy). In contrast for applications that use the https requests they only get exactly the information we send plus the anonymized IP address. As far as I understand this does not even constitute monitoring behavior under the GDPR, because it is impossible for both Google and ourselves to connect the transmitted data to a specific user, something that would be possible if we transmitted a unique user ID. Taken together we would only need the most basic requirements of a privacy policy, which I took from here: https://www.iubenda.com/en/help/6187-what-should-be-in-a-privacy-policy.

MFraters:
> Let's move that discussion to an in person discussion at some time (it takes much time)

Yes, that makes sense. I think we have a reasonable basis for the hackaton discussion now :)

> which I took from here: https://www.iubenda.com/en/help/6187-what-should-be-in-a-privacy-policy.

Thanks for the link!

PR #3598, opened 2020-07-11
Calculate viscosity on phases

lhy11009:

Hi Rene,

@gassmoeller 
I have finished most of this pull request, to calculate viscosity on phases. I have changed the parse function and the compute_viscosity function in the two 'creep.cc' file. And I also changed the interface in the 'calculate_isostrain_viscosities' function in order to have 'phase_function_values' as an entry. I have added a new test, 'visco_plastic_phases_viscosity', where I used this new pull request.
I have also fixed the related part in ViscoPlasticStrainInvariant<dim>::update_particle_property with the interface there. However, in order to do that, I have to define three public functions in ViscoPlastic class. It does look a little bit messy to me to fix this small part. I expert that we may find better solutions.

### For all pull requests:

### DIFF
--- a/source/material_model/rheology/diffusion_creep.cc
+++ b/source/material_model/rheology/diffusion_creep.cc
@@ -20,6 +20,7 @@
 
 
 #include <aspect/material_model/rheology/diffusion_creep.h>
+#include <aspect/material_model/utilities.h>
 #include <aspect/utilities.h>
 
 #include <deal.II/base/signaling_nan.h>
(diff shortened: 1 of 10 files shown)


lhy11009:
Hi Rene

I have addressed all of your suggestions. And what do you think of the 'passing variables to calculate_isostrain_viscosity' issue that I mentioned at the start? In short,  Particle::Property::ViscoPlasticStrainInvariant::update_one_particle_property() uses this function out of the namespace, which makes it hard to pass in phase function values.

gassmoeller:
Ah, I did not pay attention, I will take a closer look later today.

gassmoeller:
After a closer inspection here are my thoughts:

> One is in the 'evaluate' function, where there is already a 'phase_function_values" variable, so this is straightforward.
A second one is in the "compute_viscosity_derivatives" function. For that, I also change the interface for the "derivatives" function in order to pass the 'phase_function_values' variable into this function. Is that what we want to do?

Yes this seems like the best approach. This is a private function that is only callable from inside ViscoPlastic, and it is only called from `evaluate` at the moment, so we know that these values will always be available.

> A third one is in the 'is_yield' function and that one is called by Particle::Property::ViscoPlasticStrainInvariant::update_one_particle_property(), which is out of the MaterialModel namespace. In order to pass into that function, we may need to compute phase function values inside that function. And I also noticed that there is an instantiation of the ViscoPlastic class inside that function. So does this object know the information of the phase function? I haven't change the implementation for this part yet.

Like you mentioned this part is more complicated. You are right, it is simplest to recompute the values of the phase function inside `is_yielding`, but the current interface of the function does not get all the necessary information for that (in particular it does not get the position). I am working on a fix for the interface, but that will take until tomorrow (#3612 is a first step in that direction).

lhy11009:
Right, for that part, I think we'll need to wait until the change in interface is done. Would I finish that in this PR or another one?

gassmoeller:
I opened #3612 and #3613 that together should allow you to copy the necessary computations from `evaluate` to `is_yielding` (you then have access to things like the `position` which is currently not available inside `is_yielding`, but is needed for the phase function). Since I made these other PRs we will need someone else to review them, maybe we can ask @naliboff (although I know he is very busy at the moment) or @tjhei?

tjhei:
> or @tjhei?

also busy, but I took a quick look. :-)

gassmoeller:
Thanks a lot! I was not implying you're bored, but I know John is leading a conference this week so I was not counting on him being able to take a look :-).

gassmoeller:
@lhy11009: The other two PRs are merged. Could you update your master branch, rebase this branch onto master and copy the necessary steps to use the phase transitions from `ViscoPlastic::evaluate` to `ViscoPlastic::is_yielding` (the new version of the function)? Afterwards this should be ready.

gassmoeller:
I just wanted to check in if my suggestions were good enough. Did you have time to look into them?

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -37,23 +38,48 @@ namespace aspect
       {}
 
 
-
```
gassmoeller, line 40: Please keep this line

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -107,11 +109,36 @@ namespace aspect
...
+            reference_density = eos_outputs_all_phases.densities[0];
+          }
+
+        MaterialModel::MaterialUtilities::PhaseFunctionInputs<dim> phase_inputs(material_inputs.temperature[0],
+                                                                                material_inputs.pressure[0],
+                                                                                this->get_geometry_model().depth(material_inputs.position[0]),
+                                                                                gravity_norm*reference_density,
+                                                                                numbers::invalid_unsigned_int);
+
+        std::vector<double> phase_function_values = viscoplastic.phase_function_compute_value(phase_inputs);
```
gassmoeller, line 139: As just discussed: If you move this code into `ViscoPlastic::is_yielding` you do not need the new functions inside `ViscoPlastic` and can save some code.

lhy11009: Thanks, I have modified the codes according to your suggestions. Now please take a look again and see what to do next.

source/material_model/visco_plastic.cc
```diff
@@ -78,11 +78,44 @@ namespace aspect
...
+                                                               gravity_norm*reference_density,
+                                                               numbers::invalid_unsigned_int);
+
+      std::vector<double> phase_function_values(phase_function.n_phase_transitions(), 0.0);
+
+      for (unsigned int j=0; j < phase_function.n_phase_transitions(); j++)
+        {
+          phase_inputs.phase_index = j;
+          phase_function_values[j] = phase_function.compute_value(phase_inputs);
+        }
```
gassmoeller, line 111: Yes exactly!
Can you guard this to only execute if there are phase transitions? E.g. like this:
```
std::vector<double> phase_function_values(phase_function.n_phase_transitions(), 0.0);
if (phase_function.n_phase_transitions() > 0)
  {
   ... here goes all the new code that sets values for phase_function_values
  }
here it continues as before
```

source/material_model/visco_plastic.cc
```diff
@@ -91,7 +124,6 @@ namespace aspect
     }
 
 
-
```
gassmoeller, line 94: Please add this line again, I think you removed it by accident.

lhy11009: I have added that. So it seems 'make indent' doesn't take care of that?

source/material_model/visco_plastic.cc
lhy11009: This is also addressed

PR #3601, opened 2020-07-14
Fix cmake version check in CMakeLists.txt

gassmoeller:
This was reported by Kiran Chotalia for cmake version 3.15.6. Our current cmake check tries to precompile headers for this version, but the feature is only available starting with 3.16. Ideally we would like to use `CMAKE_VERSION VERSION_GREATER_EQUAL 3.16`, but `VERSION_GREATER_EQUAL` is only available since cmake version 3.7 (and we allow versions since 2.8.12). This workaround fixes the problem (as long as cmake never releases a version 3.15.100).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -510,9 +510,9 @@ CONFIGURE_FILE(
 # 3.9. Starting with cmake 3.16 there is native support inside cmake and we can
 # use that.
 IF(CMAKE_VERSION VERSION_LESS 3.9)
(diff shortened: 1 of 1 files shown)


MFraters:
Looks good to me, but can't you just use `CMAKE_VERSION VERSION_LESS 3.16`. That should be available and prevents the 100 problem.

tjhei:
> but can't you just use `CMAKE_VERSION VERSION_LESS 3.16`

Yes, you can even use ``NOT (CMAKE_VERSION VERSION_LESS 3.16)`` and keep the logic the same way.

PR #3607, opened 2020-07-19
disable part of mesh smoothing in refine/coarse

### LINKED ISSUE #3604: periodic adaptive refinement can cause infinite loop in deal.II 9.2.0
observed in deal.II 9.2.0 and ASPECT 2.2 and current master (also see #3583).

To reproduce, run ``tests/periodic_box2.prm`` (even with 1 processor) and observe that we hang in time step 10 (when adaptive refinement happens). The test passes with older deal.II versions.

It looks like we end up in an infinite loop in the following code section:
https://github.com/dealii/dealii/blob/f29b68a2829ecdccb1c2753c7e290505901b02ad/source/distributed/tria.cc#L3442-L3450

tjhei:
The mesh smoothing algorithms in deal.II change coarsening/refinement
flags but lead to a hang/crash when combined with periodic boundaries.

As a work-around, disable one particular smoothing flag
eliminate_refined_boundary_islands. This flag is not important to have
regardless. Strictly speaking, this work-around is not necessary with
9.3.0 onward.

Finally, eliminate the logic for limit_level_difference_at_vertices. It
turns out that this is contained in smoothing_on_refinement already.

fixes #3604


tjhei:
This is a more conservative version of #3556

PR #3611, opened 2020-07-22
Enable unity build by default if possible

gassmoeller:
This is the last item I would like to do to address #3585. I think if this works it will speed up peoples builds significantly, and if we run into too many problems during the hackathon or later, we can still decide to revert this before the next release (but so far I have not encountered issues). I decided to go with `ASPECT_UNITY_BUILD` instead of `CMAKE_UNITY_BUILD` because the latter seems to be only usable from the command line (not from CMakeLists), and it fits better into our other options.


tjhei:
This is okay with me, but:
1. Could you change the CI to disable it by default (and enable it in the CIG "unity build" explicitly)?
2. How does this interact with specifying ``CMAKE_UNITY_BUILD=on``? Not a big problem, but I would assume that either option enable it? Should we maybe set ``UNITY_BUILD`` property to OFF when ASPECT_UNITY_BUILD if OFF?

gassmoeller:
Like this? In this configuration CMAKE_UNITY_BUILD has no effect at all, all that matters is ASPECT_UNITY_BUILD.

PR #3612, opened 2020-07-23
Use new particle property interface

gassmoeller:
This PR updates the particle property plugins to use the new interface function introduced in #3482, no functional changes are expected. The old function is already deprecated. This will be necessary to address a question raised in #3598. I will open a follow-up PR in which the connection will become more clear.


tjhei:
Indeed, this looks simple enough.

PR #3613, opened 2020-07-24
Update visco plastic is yielding

gassmoeller:
This PR sits on top of #3612 and allows to input more properties into the `ViscoPlastic::is_yielding` function as is necessary for #3598. I have checked that the version of this function in this PR is as fast as the one in master (this is important because this function is called once per particle per timestep and can become very expensive).

### DIFF
--- a/include/aspect/material_model/visco_plastic.h
+++ b/include/aspect/material_model/visco_plastic.h
@@ -254,11 +254,19 @@ namespace aspect
          * A function that returns whether the material is plastically yielding at
          * the given pressure, temperature, composition, and strain rate.
          */
+        DEAL_II_DEPRECATED
         bool
         is_yielding ( const double &pressure,
                       const double &temperature,
                       const std::vector<double> &composition,
                       const SymmetricTensor<2,dim> &strain_rate) const;
+
+        /**
+        * A function that returns whether the material is plastically yielding at
+        * the given pressure, temperature, composition, and strain rate.
+        */
+        bool
+        is_yielding (const MaterialModelInputs<dim> &in) const;
 
       private:
 
--- a/include/aspect/material_model/visco_plastic.h
+++ b/include/aspect/material_model/visco_plastic.h
@@ -253,6 +253,9 @@ namespace aspect
         /**
          * A function that returns whether the material is plastically yielding at
          * the given pressure, temperature, composition, and strain rate.
+         *
+         * @deprecated: Use the other function with this name instead, which allows
+         * to pass in more general input variables.
          */
         DEAL_II_DEPRECATED
         bool
@@ -263,7 +266,8 @@ namespace aspect
 
         /**
         * A function that returns whether the material is plastically yielding at
-        * the given pressure, temperature, composition, and strain rate.
+        * the given input variables (pressure, temperature, composition, strain rate,
+        * and so on).
         */
         bool
         is_yielding (const MaterialModelInputs<dim> &in) const;
--- a/include/aspect/particle/property/viscoplastic_strain_invariants.h
+++ b/include/aspect/particle/property/viscoplastic_strain_invariants.h
@@ -91,6 +91,12 @@ namespace aspect
         private:
           unsigned int n_components;
 
+          /**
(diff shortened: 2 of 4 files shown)


include/aspect/material_model/visco_plastic.h
```diff
@@ -254,12 +254,20 @@ namespace aspect
          * A function that returns whether the material is plastically yielding at
          * the given pressure, temperature, composition, and strain rate.
          */
+        DEAL_II_DEPRECATED
```
tjhei, line 257: can you add a comment ``@deprecated ...`` to use the other function instead?

include/aspect/material_model/visco_plastic.h
```diff
@@ -254,12 +254,20 @@ namespace aspect
...
         bool
         is_yielding ( const double &pressure,
                       const double &temperature,
                       const std::vector<double> &composition,
                       const SymmetricTensor<2,dim> &strain_rate) const;
 
+        /**
+        * A function that returns whether the material is plastically yielding at
+        * the given pressure, temperature, composition, and strain rate.
```
tjhei, line 266: Maybe update the comment?

include/aspect/particle/property/viscoplastic_strain_invariants.h
```diff
@@ -91,6 +91,12 @@ namespace aspect
         private:
           unsigned int n_components;
 
+          /**
+           * An object that is used to compute the particle property. Since the
+           * object is expensive to create and is needed often it is kept as a
+           * member variable.
+           */
+          mutable MaterialModel::MaterialModelInputs<dim> material_inputs;
```
tjhei, line 99: Is it guaranteed that we are not doing any updates from multiple threads? I think this is not called from assembly right now, but somebody might want to do particle updates from multiple threads. Not sure how important that is right now, but it is worth thinking about.

source/material_model/visco_plastic.cc
```diff
@@ -67,6 +67,33 @@ namespace aspect
...
+    bool
+    ViscoPlastic<dim>::
+    is_yielding(const MaterialModelInputs<dim> &in) const
+    {
+      Assert(in.n_evaluation_points() == 1, ExcInternalError());
+
+      /* The following returns whether or not the material is plastically yielding
+       * as documented in evaluate.
+       */
+      bool plastic_yielding = false;
```
tjhei, line 82: why don't you move the declaration down to the assignment? (I know that this is copy&paste).

include/aspect/material_model/visco_plastic.h
naliboff: Yes, it would be good to indicate in the comments how the two versions of `is_yielding` differ and where they will be used.

include/aspect/particle/property/viscoplastic_strain_invariants.h
naliboff: @gassmoeller - could you give a quick summary of why the object needs to be defined as `mutable`?

gassmoeller:
I think I addressed all comments and answered the questions in the added documentation. @tjhei: It is currently not used from a workstream or other multi-threaded context, but if I wanted to make this thread safe would I place a std::mutex around the section using `material_inputs` or do I need to do something different for cpp-taskflow?

naliboff:
@gassmoeller - this looks good to go from my end, but I'll let @tjhei do another check before merging.

tjhei:
> It is currently not used from a workstream or other multi-threaded contex

Good. I am sorry for this vague comment. I thought about this some more and I can not think of an easy way to detect or protect the code. Maybe treat this more as a reminder to always think about threading when you add a mutable variable. :-)

> but if I wanted to make this thread safe would I place a std::mutex around the section using `material_inputs` or do I need to do something different for cpp-taskflow?

If it is used from multiple threads (TBB, taskflow, background threads), then a mutex around the whole material_input section would be one option. Alternatively, you would pass the material_inputs to the function itself and let the outer function take care of it.

Anyways. I am not sure we can do something sensible here (and just adding a mutex would be overkill).

Issue #3582, opened 2020-07-02
deal.II version bump to 9.2


tjhei:
We decided yesterday that we want to move from deal.II 9.0.0 to 9.2.0 as the minimum version. We don't want to require 9.2 immediately, but we should make the switch before August. Here is what we need to do:

- [x] update testers to run 9.2.0
- [ ] update docker images in ./docker/
- [ ] check manual / installation instructions if they are up to date (OSX, versions)  [STARTER PROJECT]
- [x] set required deal.II version to 9.2.0 in CMakeLists.txt
- [x] enable currently disabled tests (see tests folder) #3614
- [ ] remove ``#ifdefs`` for older versions from code  [STARTER PROJECT]

PR #3614, opened 2020-07-25
enable tests that require deal.II 9.2

tjhei:
part of #3582


tjhei:
this will interfere with the changes in #3532, so we need to be careful.

gassmoeller:
Do you want me to wait with this until #3532 is resolved?

tjhei:
Either way. We just need to make sure that we rebase. Maybe we wait with this for Conrad.

gassmoeller:
Since #3532 seems to be needing another pass lets merge this to get it out of the way.

PR #3616, opened 2020-07-27
add netcdf cmake module

tjhei:
This adds cmake detection for the NetCDF c library. It is not used at
the moment.


tjhei:
Yes, that makes sense.

PR #3617, opened 2020-07-27
Enable precompiling headers on CIG testers

gassmoeller:
This was switched on before #3583 and it turns out some of the older CIG test agents can not finish compiling in 30 minutes without the setting (e.g. seen in #3598). Since the 'OFF'/'OFF' (precompiling headers/unity build) option is already tested on jenkins.tjhei.info, we can probably enable it here. The alternative is to increase the walltime limit and wait for a long time for some PRs.


tjhei:
I am fine with that change. The means we need for both of them to finish though, right?

PR #3618, opened 2020-07-28
enforce test timeout limit in CI

### LINKED ISSUE #3605: CI: test timeout not respected
discovered in #3583.
We are prebuiilding and running tests using ``ninja -k 0 tests || true`` but this will ignore the ctest timeout options of 600s for each test. The whole thing will abort after 2.5 hours, but this way you can not tell which test is broken.
We should fix that somehow. Maybe have a shorter timeout for the prebuild so we at least see the hanging tests when we go to the second phase?

tjhei:
We precompile the tests bypassing "ctest". This means that the normal
test timeout is being ignored. As a work-around, run using the posix
"timeout" function.

fixes #3605


gassmoeller:
I am going to merge this although the CIG tester has not finished, because it has passed on jenkins.tjhei.info, and because it will help with other PRs.

gassmoeller:
Actually, let me rethink this. Are you sure this is going to work? I see in the output of the tester that individual tests fail due to timeout, but that the script does not cancel, likely because it is run as `ninja -k 0 tests || true` from the Jenkinsfile.

tjhei:
> from the Jenkinsfile.

Yeah, this is kind of subtle and not perfect:
prebuild ignores errors during and after the run (because of -k and the ``true``). If there is a timeout, the next step should run those tests again (and then timeout again).
Is that acceptable?

gassmoeller:
Oh, I see. Yes that works. It is still a stopgap solution, because we have to wait twice for the tests to timeout, but better than the current state. But then I want to wait to see if the timeout actually happens before merging.

tjhei:
can we merge this, @gassmoeller ? I think this will help with some of the slowdowns.

gassmoeller:
Yes, I was just waiting for the testers, but they have finished now.

PR #3619, opened 2020-07-29
mark test to require world builder

tjhei:


tjhei:
> Just out of curiosity, what would happen is the world builder is not present

before: the test would fail. Now: the test won't be enabled, see https://github.com/geodynamics/aspect/blob/69ebaeac5fe28928ceeaf13586f8938298eb02b3/tests/CMakeLists.txt#L106-L109

tjhei:
@gassmoeller please take a look if you are happy with the tester situation.

gassmoeller:
I ran it through one of the fast CIG testers and it worked. Ready to go.

PR #3620, opened 2020-07-29
indent script: return failure

tjhei:
not very important, but useful when run in a CI environment


gassmoeller:
This only affected the indent phase and that already passed on the tester. It looks like jenkins.tjhei.info was not triggered by this PR?

tjhei:
> This only affected the indent phase and that already passed on the tester. It looks like jenkins.tjhei.info was not triggered by this PR?

my machine is behind with the current list of jobs. :-)

gassmoeller:
This looks good. The CIG tester fails are unrelated (see #3627).

PR #3621, opened 2020-07-29
CI: Jenkinsfile make the same

tjhei:
- increase build time to 45min
- make CIG and Timo's tester do the same stuff


tjhei:
As prep for the hackathon as discussed with @gassmoeller today.

PR #3622, opened 2020-07-29
fix error handling in update_parameters.sh

tjhei:
Round parenthesis in bash open a subshell, so the outer script doesn't
exit. Replace by curly to make that work.


gassmoeller:
Is it possible that this PR broke the Github actions? See: https://github.com/geodynamics/aspect/pull/3627/checks?check_run_id=928603909

tjhei:
yes, that is likely. Let me think about this.

tjhei:
Yes, the error was ignored before: https://github.com/geodynamics/aspect/runs/928167749#step:6:12

tjhei:
I am working on this in #3629

PR #3624, opened 2020-07-29
Limit number of MPI ranks for tests

gassmoeller:
This is just a test to see how much time this limitation saves on the slower jenkins agents. Obviously these test do not work any more with the new settings (they specifically test problems with many MPI ranks).

### DIFF
--- a/tests/parallel_refine_crash.prm
+++ b/tests/parallel_refine_crash.prm
@@ -1,4 +1,4 @@
-# MPI: 7
+# MPI: 4
 
 # Test case that showed two problems:
 #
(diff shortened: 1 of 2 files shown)


tjhei:
wait, why does my Jenkins build succeed here?!

gassmoeller:
I think because the tests are checking that they succeed with many MPI ranks. Succeeding with less MPI ranks was never a problem. It is interesting that the partition does not seem to matter though.

But did you see that the change also sped up your tester by 10 min (from 57 to 46 min for the tests, assuming your testers are all the same)?

tjhei:
If we gain 10 minutes, then this means that oversubscription is indeed a problem.

PR #3625, opened 2020-07-30
jenkins experiments

tjhei:
please do not merge, I am playing with different options to speed up the testing.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -105,7 +105,6 @@ pipeline {
         cmake \
         -G 'Ninja' \
         -D CMAKE_CXX_FLAGS='-Werror' \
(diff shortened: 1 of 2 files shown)


tjhei:
@gassmoeller why testing is relatively fast (20 min), "generating reference output" now takes 6m because make is so much slower than ninja. I don't think that is a good option.
We could refactor the "generating reference output" thing to not run ctest (we could always generate the files and then just copy them over), but this is a bit worrying before the hackathon.

bangerth:
Is this still current? Do you want to have it merged, or closed?

tjhei:
I don't want to change anything about the testers during the hackathon, but I need to have another chat with @gassmoeller if we should remove prebuilding tests as proposed here.

PR #3631, opened 2020-07-30
fix conflicting lib detection for deal.II ubuntu package

tjhei:
The deal.II ubuntu package is
/lib/x86_64-linux-gnu/libdeal.ii.g.so.9.2.0
so the detection for library mismatch between plugin and binary fails.
Fix this.


tjhei:
part of #3582

PR #3636, opened 2020-07-31
github-actions: test no-unity build and unittests

tjhei:
- add second github actions to do a build without unity/precompiled
header
- also run the unittests

as discussed with @gassmoeller


tjhei:
> Looks good. If the no-unity job succeeds, do you want to remove the no-unity build from the jenkinsfiles in this same PR?

I was going to do that separately so we can see for a bit, if this thing works reliably.

gassmoeller:
This is independent of the jenkins tester so I am merging.

Issue #3628, opened 2020-07-30
Need to speed up slow tests


tjhei:
We have two sequential tests:
```
        599 - rising_melt_blob_freezing (Timeout)                                                                                           
        667 - statistics_output (Timeout)   
```
that seem to take a very, very long time since we switched to deal.II 9.2 (``statistics_output`` takes 45min+ for me).

I am not sure what is going on here...

tjhei:
rising_melt_blob_freezing:
```
| Total wallclock time elapsed since start     |       815s |            |
|                                              |            |            |
| Section                          | no. calls |  wall time | % of total |
+----------------------------------+-----------+------------+------------+
| Assemble Stokes system           |        96 |       137s |        17% |
| Assemble composition system      |       192 |       265s |        33% |
| Assemble temperature system      |        96 |       139s |        17% |
| Build Stokes preconditioner      |        96 |      79.3s |       9.7% |

```

tjhei:
this should be fixed with #3618

tjhei:
new times with #3618:
```
1/1 Test #599: rising_melt_blob_freezing ........   Passed  229.85 sec
1/1 Test #667: statistics_output ................   Passed  575.50 sec
```
So they are much better but still likely responsible in the order of 8% of the total test time (assuming 13 minutes of real time with 12 cores).

tjhei:
new statistics_output timing:
```
+----------------------------------------------+------------+------------+
| Total wallclock time elapsed since start     |       570s |            |
|                                              |            |            |
| Section                          | no. calls |  wall time | % of total |
+----------------------------------+-----------+------------+------------+
| Assemble Stokes system           |        16 |      75.9s |        13% |
| Assemble composition system      |        32 |       190s |        33% |
| Assemble temperature system      |        16 |      99.9s |        18% |
| Build Stokes preconditioner      |        16 |      49.5s |       8.7% |
```

tjhei:
Other slow tests (prebuild timed out probably due to parallel load), see https://jenkins.geodynamics.org/blue/organizations/jenkins/aspect/detail/PR-3631/1/pipeline
Run sequentially in phase 2 they take:
```
181/802 Test #181: continental_extension ..............................................   Passed  345.64 sec
603/802 Test #603: segregation_heating ................................................   Passed  291.77 sec
667/802 Test #667: statistics_output ..................................................   Passed  536.98 sec
[...]

PR #3639, opened 2020-07-31
Speedup some unnecessarily slow tests

gassmoeller:
This tries to address #3628. On my system (limited a docker container to 4 threads) this bring some substantial benefits:

before: 5116.58
after:  3951.27

Lets see what the tester say (I will start it on purpose on one of the slow testers).


tests/discontinuous_temperature_blankenbach/screen-output
```diff
@@ -139,10242 +139,18 @@ Number of degrees of freedom: 1,235 (578+81+576)
...
-     Heating rate (average/total):       0 W/kg, 0 W
-
-*** Timestep 587:  t=0.102547 seconds
-   Solving temperature system... 4 iterations.
-   Rebuilding Stokes preconditioner...
-   Solving Stokes system... 3+0 iterations.
-
-   Postprocessing:
-     RMS, max velocity:                  895 m/s, 1.73e+03 m/s
-     Temperature min/avg/max:            -0.006475 K, 0.5 K, 1.
```
tjhei, line 10366: We were doing 939 timesteps? Wow.

tjhei:
This looks great. One of the tests is failing though.

How did you pick the list of tests? Would it be worth sorting all tests by time and take a look at the slowest?

gassmoeller:
> One of the tests is failing though.

That is on me, I forgot to update after I made a final change. I updated the results.

> How did you pick the list of tests? Would it be worth sorting all tests by time and take a look at the slowest?

That is essentially what I did. I took a look at tests that took more than roughly 100-150 seconds and selected the ones where I could figure out the meaning of the test and where I thought I could make them faster without removing the purpose of the test. I think we now have no tests left that take more than 200 seconds, and most between 100-200 are related to melt solvers or 3D models where the test might not converge with lower resolution or make no sense with shorter end time. There are a bunch of tests around 100 seconds for the Newton solver that I think could be simplified, but the solver settings are a bit tricky so I left them alone.

gassmoeller:
No luck, while the tests are faster the slow testers still timeout (just barely, they timeout after finishing all tests and running through cmake). I restarted the run on a fast tester to at least get this passed and merged.

tjhei:
This looks great, so I will merge.

Issue #3630, opened 2020-07-30
slow CI since switching to deal.II 9.2


tjhei:
It looks like our testing time increased substantially with the switch to the newer docker image. 
For example in #3425 
![image](https://user-images.githubusercontent.com/1531285/88964293-34fa9500-d277-11ea-9668-d2a40ab0087c.png)
We go from 16 to 70 minutes

This might be related to:
- [ ] MPI oversubscription and tests using too many cores, but this should have been a problem beforehand or not? 
- [x] See #3624 for reducing the number of ranks for 2 tests
- [x] Two sequential tests now take way too long (20 and 45minutes) and currently slow down the prebuild, see #3618 for a fix
- Note that we now do a unity and nonunity build, so the build is okay to be slower. See #3621

gassmoeller:
I had another idea for why tests under 9.2 are so much slower. Between 9.0 and 9.1 deal.II disabled -Og for gcc and fell back to -O0 for debug mode (dealii/dealii#7207). I liked the change because I often debug under gcc and often had to set -O0 manually to debug, but this may cause the tester to be slower (there is even a discussion there that the same happened to the deal.II testers). Should we aim to create a tester with -Og, because we do not care about debugability?

tjhei:
Wow. I totally forgot about that. This is worth looking into and explains the slower assembly. Thank you!

tjhei:
update: we are now down to 35min (that still includes 6min for the no-unity build, that will be gone).

bangerth:
We don't actually care about debugging symbols and similar on the CI machines. Why not use `-O3` and similar flags even for debug mode? All that's necessary is to run in debug mode, but that doesn't mean that we have to use flags that actually make debugging possible.

gassmoeller:
Either are very good options. I was not aware that the asserts depend on the build type and not on the flags, although it makes sense of course. Learned something here :-). @tjhei do you have time to create a docker image with faster flags? If not I can give it a try.

gassmoeller:
Oh, I only saw #3643 after my previous post. But #3643 will only compile ASPECT with different option. I would guess the expensive part is actually happening inside deal.II?

tjhei:
> But #3643 will only compile ASPECT with different option.

I am not sure how much of a difference it makes to only optimize ASPECT, but I assume it makes some difference regarding inlining (everything that can be inlined from the deal.II headers will be inlined even if deal.II is compiled with -O0). So I wanted to try that first.

[...]

PR #3640, opened 2020-08-01
jenkins: remove unity build

tjhei:
We added the unity build to github actions in #3636, so we can remove it
from Jenkins to improve CI times:
- remove unity build from both jenkins
- take over some changes from cig->tjhei
- rename build directory

part of #3630

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -66,56 +66,30 @@ pipeline {
       }
     }
 
-    stage('Unity Build') {
+    stage('Build') {
       options {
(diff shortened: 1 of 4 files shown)


tjhei:
maybe we should wait a few days before merging to make sure that things are working correctly.

tjhei:
Huh, and now I understand why "zero_matrix.prm" always fails on my machine. Let me fix this. :-)

tjhei:
As discussed with @gassmoeller . Rene, should be merge before Monday?

PR #3641, opened 2020-08-01
github actions: enable -Werror

tjhei:
I realized that we should error out when we find warnings.


gassmoeller:
This only affects the github actions and they pass.

PR #3643, opened 2020-08-02
cmake: allow additional cxx flags

tjhei:
CMAKE_CXX_FLAGS take action before DEAL_II_CXX_FLAGS that are loaded
from deal.II. That means one can not overwrite settings like -O0 from
deal.II. Provide a new option instead.

related #3630


gassmoeller:
All builds and your tester have finished, so no need to wait.

PR #3646, opened 2020-08-02
Jenkins: run testers with -O1

tjhei:
part of #3630

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,7 +80,7 @@ pipeline {
         cmake \
         -G 'Ninja' \
         -D CMAKE_CXX_FLAGS='-Werror' \
-        -D ASPECT_ADDITIONAL_CXX_FLAGS='-O1' \
+        -D ASPECT_ADDITIONAL_CXX_FLAGS='-O3' \
         -D ASPECT_TEST_GENERATOR='Ninja' \
         -D ASPECT_PRECOMPILE_HEADERS=ON \
         -D ASPECT_UNITY_BUILD=ON \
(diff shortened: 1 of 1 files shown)


tjhei:
Let's see if this makes a difference.

tjhei:
It looks like it only makes a small difference (18min vs 20mins for the testing). Should we still merge it and also apply it to the CIG Jenkins and Github Actions?

gassmoeller:
I think ultimately we should be consistent with whatever we decide for the deal.II base image (-Og or even -O3 if we decide to use that, I would be open for it). It would be confusing to compile deal.II with one set, but aspect with another. If you want we can merge this nevertheless for now, it does not seem to do harm.

bangerth:
If you can go to `-O1`, you might as well go all the way to `-O3`.

deal.II is designed in such a way that the flags don't need to be consistent -- we make sure that the ABI is compatible between debug and release version. If for whatever reason, something doesn't work, then that's a bug and finding that out via the CI infrastructure is a valuable service :-)

source/geometry_model/chunk.cc
```diff
@@ -934,6 +934,15 @@ namespace aspect
 {
   namespace GeometryModel
   {
+    namespace internal
+    {
+#define INSTANTIATE(dim) \
+  template class ChunkGeometry<dim>;
+      ASPECT_INSTANTIATE(INSTANTIATE)
+
+#undef INSTANTIATE
```
bangerth, line 943: If this was missing, make that a separate PR.

tjhei:
O3 is a tiny bit faster (16min for tests). Do we want to merge this as is? I would think it is a strict improvement but we might throw the changes away if we change the docker image.

gassmoeller:
Let's merge it as is once the testers are happy.

PR #3649, opened 2020-08-02
add .mailmap file to better associate authors

tjhei:


bangerth:
Are you going to use it in a CI step next?

tjhei:
> Are you going to use it in a CI step next?

I just used it to see who the new contributors are. How do you propose we used it in CI?

Something I thought about at some point: require every commit to come from a user who is listed in the authors file. Every new user would add themselves there.

bangerth:
On 8/3/20 6:29 AM, Timo Heister wrote:
> I just used it to see who the new contributors are. How do you propose we used 
> it in CI?

Like in deal.II, where a PR is rejected if the user is not either listed in 
the file, or has a "firstname lastname <...>" email address. At least that's 
what I *thought* you're doing in deal.II :-)


> Something I thought about at some point: require every commit to come from a 
> user who is listed in the authors file. Every new user would add themselves there.

So that would be everyone's first PR. But I don't see why that is necessary -- 
why not just copy the workflow we have in deal.II?

tjhei:
> But I don't see why that is necessary -- why not just copy the workflow we have in deal.II?

I see a large number of garbage email addresses used in ASPECT (check this PR). That would avoid that problem.

Issue #3633, opened 2020-07-31
clang compile errors with unity builds


tjhei:
As reported on the mailing list, we have errors with clang9 and clang10 with enabled unity builds like:
```
/home/heister/aspect-git/source/volume_of_fluid/handler.cc:374:22: error: duplicate explicit instantiation of 'copy_local_to_global_volume_o
f_fluid_system'                                                                                                                               ASPECT_INSTANTIATE(INSTANTIATE)                                    
                     ^                                                                                                                      
/home/heister/aspect-git/source/volume_of_fluid/assembly.cc:235:21: note: previous explicit instantiation is here
                    copy_local_to_global_volume_of_fluid_system,  
```
and
```
/Users/billen/Box Sync/Aspect-Src/aspect/source/volume_of_fluid/utilities.cc:32:14: error:
explicit specialization of ‘compute_fluid_fraction<2>’ after instantiation
double compute_fluid_fraction<2> (const Tensor<1, 2> normal,
^
/Users/billen/Box Sync/Aspect-Src/aspect/source/volume_of_fluid/setup_initial_conditions.cc:210:76: note:
implicit instantiation first required here
```

I am also seeing some warnings that I will tackle:
```
../include/aspect/particle/property/integrated_strain.h:66:11: warning: 'update_particle_property' overrides a member function but is not ma
rked 'override' [-Winconsistent-missing-override]                                                                                                     update_particle_property (const unsigned int data_position,
 ```

**Comment by bangerth** [comment]
I believe this is now fixed -- can we close it?

**Comment by tjhei** [comment]
> I believe this is now fixed -- can we close it?

No, I see several other errors and warnings:
```
/home/heister/aspect-git/source/volume_of_fluid/reconstruct.cc:31:33: error: explicit specialization of 'update_volume_of_fluid_normals' after instantiation
  void VolumeOfFluidHandler<2>::update_volume_of_fluid_normals (const VolumeOfFluidField<2> &field,
                                ^
/home/heister/aspect-git/source/volume_of_fluid/handler.cc:527:9: note: implicit instantiation first required here
        update_volume_of_fluid_normals (volume_of_fluid_field, sim.solution);

/home/heister/aspect-git/source/volume_of_fluid/reconstruct.cc:437:33: error: explicit specialization of 'update_volume_of_fluid_composition' after instantiation
  void VolumeOfFluidHandler<2>::update_volume_of_fluid_composition (const typename Simulator<2>::AdvectionField &composition_field,
[...]

PR #3650, opened 2020-08-03
move VoFHandler implementation into single .cc

tjhei:
part of #3633


tjhei:
I am not a big fan of moving everything into a single .cc but I don't know of a good way around it: putting them into a header that is then ``#include``'d, doesn't seem to be the better choice.

bangerth:
The unity tester fails, but I think that's because of a connectivity problem. If you're confident that that's spurious, then go about and merge yourself.

tjhei:
Let's not go this way and look into #3653 instead.

bangerth:
So just out of curiosity, why *not* combine the file? Is it getting so large that it's going to dominate build time?

tjhei:
> So just out of curiosity, why _not_ combine the file?

I think they were split up to separate code by topic (and this is unrelated to code size). I don't think this is very important when using an IDE these days.

Do you want to combine all instances into single files (core.cc/helper_functions.cc/etc.)? If not, then we need the work-around for unity builds. Then I don't see a need to merge this PR.

PR #3653, opened 2020-08-03
unity: move some files to the end

### LINKED ISSUE #3651: class instantiations with implementations in different .ccs?
We have various issues with unity builds because we have implementations and instantiations split into separate files (#3637, #3650, ...). I don't like that solution and we have this also in core.cc, assembly.cc, helper_functions.cc, etc..
Questions:
1. Is this even correct what we do without unity builds? We list function instantiations at the bottom for file that do not instantiate the main class. For example https://github.com/geodynamics/aspect/blob/master/source/simulator/assembly.cc#L1254
But we are not consistent about it.
2. What do we do to avoid this?
a. Exclude from unity build or explicitly reorder as in https://github.com/geodynamics/aspect/blob/d84f68a39a22c37a0cd9789126840f9c76bfe05a/CMakeLists.txt#L31-L35 ?
b. Move everything into one .cc
c. Turn the helper implementations into a ``xyz.template.h`` and include in the main file.

tjhei:
This should fix instantiation warnings/errors

part of #3633
fixes #3651 
closes #3650 (better alternative)
related to #3637

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,11 +28,21 @@ SET(TARGET "aspect")
 FILE(GLOB_RECURSE TARGET_SRC
      "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
 
-# For the unity build we need to compile helper_functions.cc before core.cc
-# ensure this by reordering the files
-FIND_FILE(HELPER_PATH helper_functions.cc HINTS "${CMAKE_SOURCE_DIR}/source/simulator")
-LIST(REMOVE_ITEM TARGET_SRC ${HELPER_PATH})
-LIST(INSERT TARGET_SRC 0 ${HELPER_PATH})
+# special treatment of some files for unity builds:
+LIST(APPEND UNITY_LAST_FILES
+	"source/simulator/core.cc;source/volume_of_fluid/handler.cc")
+
+# Fine the files listed above and move them to the end of the list:
+FOREACH(_source_file ${UNITY_LAST_FILES})
+  SET(_full_name "${CMAKE_SOURCE_DIR}/${_source_file}")
+  LIST(FIND TARGET_SRC ${_full_name} _index)
+  IF(_index EQUAL -1)
+    MESSAGE(FATAL_ERROR "could not find ${_full_name}.")
+  ENDIF()
+
+  LIST(REMOVE_ITEM TARGET_SRC ${_full_name})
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -28,11 +28,21 @@ SET(TARGET "aspect")
...
      "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
 
-# For the unity build we need to compile helper_functions.cc before core.cc
-# ensure this by reordering the files
-FIND_FILE(HELPER_PATH helper_functions.cc HINTS "${CMAKE_SOURCE_DIR}/source/simulator")
-LIST(REMOVE_ITEM TARGET_SRC ${HELPER_PATH})
-LIST(INSERT TARGET_SRC 0 ${HELPER_PATH})
+# special treatment of some files for unity builds:
+LIST(APPEND UNITY_LAST_FILES
+	"source/simulator/core.cc;source/volume_of_fluid/handler.cc")
```
bangerth, line 33: so helper_functions.cc is no longer needed?

CMakeLists.txt
```diff
@@ -28,11 +28,21 @@ SET(TARGET "aspect")
...
+
+# Fine the files listed above and move them to the end of the list:
```
bangerth, line 35:
```suggestion
# Find the files listed above and move them to the end of the list:
```

CMakeLists.txt
```diff
@@ -28,11 +28,21 @@ SET(TARGET "aspect")
 FILE(GLOB_RECURSE TARGET_SRC
      "source/*.cc" "unit_tests/*.cc" "include/*.h" "contrib/catch/catch.hpp")
 
-# For the unity build we need to compile helper_functions.cc before core.cc
-# ensure this by reordering the files
-FIND_FILE(HELPER_PATH helper_functions.cc HINTS "${CMAKE_SOURCE_DIR}/source/simulator")
-LIST(REMOVE_ITEM TARGET_SRC ${HELPER_PATH})
-LIST(INSERT TARGET_SRC 0 ${HELPER_PATH})
+# special treatment of some files for unity builds:
```
bangerth, line 31: Can you elaborate what "special" means and why it is necessary?

CMakeLists.txt
tjhei: helper_functions was moved to the front to be in front of core.cc, Now I move things to the back instead.

tjhei:
updated.

PR #3654, opened 2020-08-03
remove stray preprocessor command

tjhei:
fixes #3647

### DIFF
--- a/source/material_model/interface.cc
+++ b/source/material_model/interface.cc
@@ -351,8 +351,6 @@ namespace aspect
     }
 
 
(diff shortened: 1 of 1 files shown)


tjhei:
@bangerth typing "rebuild" is not required for ASPECT (we build automatically for "trusted" users). Right now, this just adds additional load to the testers. We could change that, though. Thoughts?

bangerth:
Ah, goes to show that I can't keep the procedures of two projects in my head at the same time :-) 

I think the policy is good as is. I wished we had that for deal.II as well.

PR #3664, opened 2020-08-03
removed a now unnecessary compatibility fix for DEAL-II version

mibillen:
A pull request created during git tutorial. Removes an outdated compatibility fix.

### DIFF
--- a/source/geometry_model/spherical_shell.cc
+++ b/source/geometry_model/spherical_shell.cc
@@ -165,25 +165,7 @@ namespace aspect
 
                       this_cell.material_id = cell->material_id();
 
-#if !DEAL_II_VERSION_GTE(9,1,0)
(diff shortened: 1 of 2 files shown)


mibillen:
For the second commit (e0705ba), the if statement is asking "if not DEAL-II version 9.1" then do something special. The comment suggests that this was taking care of behavior prior to version 9.1, so that everything inside the if statement should be removed. I hope I understood this correctly.  I was also surprised that this commit got added to pull request we made together... so do I need to make a different branch for a new pull request to be made? It makes sense for these to be together, I just want to make sure I understand.

tjhei:
> so do I need to make a different branch for a new pull request to be made? It makes sense for these to be together, I just want to make sure I understand.

This is fine to have in a single PR and your changes look good to me. :+1:

gassmoeller:
All tests have passed and the PR has an approving review, so I am going to merge.

Issue #3599, opened 2020-07-13
Improve GPlates parameter documentation


gassmoeller:
'Point one' and 'Point two' parameters are not used in 3D models, but this is not documented, I will open a PR to fix this.

alarshi:
I will work on it and implement it in the documentation.

PR #3668, opened 2020-08-03
Added documentation for points in 3D case.

alarshi:
This pull request aims to address the issue  #3599 .

### For all pull requests:

### DIFF
--- a/source/boundary_velocity/gplates.cc
+++ b/source/boundary_velocity/gplates.cc
@@ -866,12 +866,12 @@ namespace aspect
                              "plate reconstruction.");
           prm.declare_entry ("Point one", "1.570796,0.0",
                              Patterns::Anything (),
-                             "Point that determines the plane in which a 2D model lies in. Has to be in the format `a,b' where a and b are theta (polar angle)  and phi in radians."
-                             "This value is not utilized in the 3D geometry models, and could therefore be set to the default or any user-defined quantity.");
+                             "Point that determines the plane in which a 2D model lies in. Has to be in the format `a,b' where a and b are theta (polar angle) and "
+                             "phi in radians. This value is not utilized in 3D geometries, and can therefore be set to the default or any user-defined quantity.");
           prm.declare_entry ("Point two", "1.570796,1.570796",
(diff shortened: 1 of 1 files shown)


source/boundary_velocity/gplates.cc
```diff
@@ -866,10 +866,12 @@ namespace aspect
                              "plate reconstruction.");
           prm.declare_entry ("Point one", "1.570796,0.0",
                              Patterns::Anything (),
-                             "Point that determines the plane in which a 2D model lies in. Has to be in the format `a,b' where a and b are theta (polar angle)  and phi in radians.");
+                             "Point that determines the plane in which a 2D model lies in. Has to be in the format `a,b' where a and b are theta (polar angle)  and phi in radians."
+                             "This value is not utilized in the 3D geometry models, and could therefore be set to the default or any user-defined quantity.");
```
gassmoeller, line 870:
```suggestion
                             "Point that determines the plane in which a 2D model lies in. Has to be in the format `a,b' where a and b are theta (polar angle)  and phi in radians. "
                             "This value is not utilized in 3D geometries, and can therefore be set to the default or any user-defined quantity.");
```

source/boundary_velocity/gplates.cc
```diff
@@ -866,10 +866,12 @@ namespace aspect
...
           prm.declare_entry ("Point two", "1.570796,1.570796",
                              Patterns::Anything (),
-                             "Point that determines the plane in which a 2D model lies in. Has to be in the format `a,b' where a and b are theta (polar angle)  and phi in radians.");
+                             "Point that determines the plane in which a 2D model lies in. Has to be in the format `a,b' where a and b are theta (polar angle)  and phi in radians."
+                             "This value is not utilized in the 3D geometry models, and could therefore be set to the default or any user-defined quantity.");
```
gassmoeller, line 874: same changes as above

source/boundary_velocity/gplates.cc
alarshi: Got it, thank you! :)

PR #3671, opened 2020-08-04
Quadratic least squares particle interpolation

sac-bsa:
The Quadratic Least Squares interpolation algorithm requires only a small, fixed number of particles per cell to achieve optimal accuracy as the cell size h goes to zero, unlike the bilinear least squares interpolation algorithm. See the [attached file](https://github.com/geodynamics/aspect/files/5019414/MJG_and_EGP_2020_AGU_Abstract.pdf) that @egpuckett and I made for a short explanation of this algorithm and why it's potentially useful.
@gassmoeller @tjhei @bangerth 
### For all pull requests:


tjhei:
Sorry for the accidental click to close it. We are a busy with the hackathon, so the review might take a bit longer than normal.

tjhei:
You can already take a look at the error message: https://github.com/geodynamics/aspect/pull/3671/checks?check_run_id=942864983#step:4:250

egpuckett:
@tjhei No Problem.  My understanding of the Hackathon was that the goal was to have as many PR's merged as possible.  Just trying to help meet that goal! :-)  Take Care and let's have fun!

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,294 @@
...
+            const Tensor<1, dim, double> relative_support_point_location = (*itr - approximated_cell_midpoint) / cell_diameter;
+            for (unsigned int property_index = 0; property_index < n_particle_properties; ++property_index)
+              {
+                double interpolated_value = c[property_index][0] +
+                                            c[property_index][1] * relative_support_point_location[0] +
+                                            c[property_index][2] * relative_support_point_location[1];
+                if (dim == 2)
+                  {
+                    interpolated_value += c[property_index][3] * relative_support_point_location[0] * relative_support_point_location[1] +
+                                          c[property_index][4] * std::pow(relative_support_point_location[0], 2) +
```
gassmoeller, line 166: `std::pow` is convenient, but it is often much slower than `relative_support_point_location[0]*relative_support_point_location[0]` (because it can calculate arbitrary powers). Maybe there are specializations for the power of 2, but I would not count on that. Please replace occurences of std::pow by actual multiplications.
```suggestion
                                          c[property_index][4] * std::pow(relative_support_point_location[0], 2) +
```

tests/particle_interpolator_quadratic_least_squares_2d_add_particles.prm
```diff
@@ -0,0 +1,127 @@
+# A test based on the bilinear SolKz benchmark that ensures that additional particles
+# are added to the model during the computation with multiple properties on 
+# the particles.
```
gassmoeller, line 3: Please update this text to say that this test is based on the bilinear SolKz benchmark, but uses the quadratic least squares interpolator.

tests/particle_interpolator_quadratic_least_squares_2d_solkz.prm
```diff
@@ -0,0 +1,125 @@
+# A test based on the SolKz benchmark that performs quadratic
+# least squares interpolation from particle properties to
+# compositional fields that are flagged as 'particle' advected fields.
+
```
gassmoeller, line 4: What is the difference between this test and the one you added above? The original difference for the linear least squares case was that one tests the addition of new particles, and the other the accurace of the interpolation. But for the quadratic least squares only the interpolation has changed, which would suggest we would only need one additional test. Or do you test something else with this test (or the other one)? If so, please add to the comment at the beginning of the files.

tests/particle_interpolator_quadratic_least_squares_2d_add_particles.prm
egpuckett: What do you mean by the **bilinear** SolKz benchmark?  As far as I know, SolKz is just a set of initial conditions that produce a solution to steady Stokes flow.  Would it referred to as the **bilinear SolKz benchmark** if it were used as a test for the compositional fields, or particle averages?

tests/particle_interpolator_quadratic_least_squares_2d_solkz.prm
egpuckett: Good catch! @sac-bsa  and I discussed this issue while he was creating the tests before making the PR.  He made the comment that the test without the addition of particles was unnecessary, since it was a duplication of the 'add particles', without testing the addition of particles. There may be something similar in the 3D bilinear interpolation tests.  We need to check. Mack mentioned that one of the current overarching goals is to make the tests finish sooner, or run faster.  Please accept my apologies for not removing the test that doesn't add particles, thereby making more work for you.

tests/particle_interpolator_quadratic_least_squares_2d_add_particles.prm
gassmoeller: Oh I was just quoting the current description in that part. I think what "bilinear SolKz benchmark" means is "the SolKz benchmark using particles that use the bilinear least squares interpolation". Feel free to clarify that in the test description.

bangerth:
@sac-bsa Let us know once you've addressed @gassmoeller 's comments and we can take another look!

sac-bsa:
@bangerth I believe that I have fixed all of @gassmoeller 's comments.

egpuckett:
@bangerth If you are OK with our changes, this is ready to be merged.

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,302 @@
...
+        std::vector<std::vector<double> > cell_properties(positions.size(),
+                                                          std::vector<double>(n_particle_properties,
+                                                                              numbers::signaling_nan<double>()));
+
+        const unsigned int n_particles = std::distance(particle_range.begin(), particle_range.end());
+
+        const unsigned int n_matrix_columns = (dim == 2) ? 6 : 10;
+        AssertThrow(n_particles >= n_matrix_columns,
+                    ExcMessage("At least one cell didn't contain enough particles to interpolate a unique solution for the cell."
+                               " The 'quadratic' interpolation scheme does not support this case."));
```
bangerth, line 88: That's something one could address in a later step: if there are not enough particles, one can always back down to a linear scheme or, in the worst case, a constant scheme. That would be in the spirit of a pseudoinverse to an operator.

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,302 @@
...
+
+
+        // Notice that the size of matrix A is n_particles x n_matrix_columns
+        // which usually is not a square matrix. Therefore, we find the
+        // least squares solution of Ac=r by solving the reduced QR factorization
+        // Ac = QRc = b -> Q^TQRc = Rc =Q^Tb
+        dealii::ImplicitQR<dealii::Vector<double>> qr;
```
bangerth, line 95:
```suggestion
        ImplicitQR<dealii::Vector<double>> qr;
```

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,302 @@
...
+                A[5][positions_index] = relative_particle_position[0] * relative_particle_position[2];
+                A[6][positions_index] = relative_particle_position[1] * relative_particle_position[2];
+                A[7][positions_index] = relative_particle_position[0] * relative_particle_position[0];
+                A[8][positions_index] = relative_particle_position[1] * relative_particle_position[1];
+                A[9][positions_index] = relative_particle_position[2] * relative_particle_position[2];
+              }
+          }
+
+        for (unsigned int column_index = 0; column_index < n_matrix_columns; ++column_index)
+          qr.append_column(A[column_index]);
```
bangerth, line 146: This is the first place where you actually use the `qr` variable. Move the declaration of it right above this loop.

source/particle/interpolator/quadratic_least_squares.cc
bangerth: You don't use that variable for a good long while. Move its declaration right above the place where you first use it.

Same, by the way, for `QTb` and `c`.

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,302 @@
...
+        dealii::ImplicitQR<dealii::Vector<double>> qr;
+        // A is a std::vector of Vectors(which are it's columns) so that we create what the ImplicitQR
+        // class needs.
+        std::vector<dealii::Vector<double>> A(n_matrix_columns, dealii::Vector<double>(n_particles));
+        std::vector<Vector<double>> b(n_particle_properties, Vector<double>(n_particles));
+        std::vector<Vector<double>> QTb(n_particle_properties, Vector<double>(n_matrix_columns));
+        std::vector<Vector<double>> c(n_particle_properties, Vector<double>(n_matrix_columns));
+        for (unsigned int property_index = 0; property_index < n_particle_properties; ++property_index)
+          if (selected_properties[property_index])
+            b[property_index] = 0;
```
bangerth, line 104: no need for these three lines -- the way you create `b` above already has the component vectors zeroed out.

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,302 @@
...
+
+        unsigned int positions_index = 0;
```
bangerth, line 106: can you rename this variable to `particle_index`? I think that more accurately represents what it stands for.

source/particle/interpolator/quadratic_least_squares.cc
bangerth: You could also write this loop in a more elegant way as
```
  for (const auto &column : A)
    qr.append_column(column);
```

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,302 @@
...
+          }
+
+        for (unsigned int column_index = 0; column_index < n_matrix_columns; ++column_index)
+          qr.append_column(A[column_index]);
+        AssertThrow(qr.size() == n_matrix_columns,
+                    ExcMessage("The matrix A was rank deficent during quadratic least squares interpolation."));
+        // If A is rank deficent then qr.append_column will not append
+        // the first column that can be written as a linear combination of
+        // other columns. We check that all columns were added through
+        // this one assertion
```
bangerth, line 152: move the explanation above the assertion

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,302 @@
...
+{
+  namespace Particle
+  {
+    namespace Interpolator
+    {
+      ASPECT_REGISTER_PARTICLE_INTERPOLATOR(QuadraticLeastSquares,
+                                            "quadratic least squares",
+                                            "Interpolates particle properties onto a vector of points using a "
+                                            "quadratic least squares method. Note that deal.II must be configured"
+                                            " with BLAS/LAPACK.")
```
bangerth, line 299: move the space to the end of the previous line

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -0,0 +1,302 @@
...
+            const auto &particle_property_value = particle->get_properties();
+            for (unsigned int property_index = 0; property_index < n_particle_properties; ++property_index)
+              if (selected_properties[property_index])
+                b[property_index][positions_index] = particle_property_value[property_index];
+            const Tensor<1, dim, double> relative_particle_position = (particle->get_location() - approximated_cell_midpoint) / cell_diameter;
+            // A is accessed by A[column][row] here since we need to append
+            // columns into the qr matrix.
+
+            // There is a potential that in the future we may need to add more
+            // 'mixed' terms to the interpolation for some problems. We for now
```
bangerth, line 120: I don't get this comment. What terms are you thinking of? You've covered all quadratic mixed terms...

sac-bsa: This comment was meant to answer @gassmoeller 's suggestion to include in 2D (x^2)y x(y^2) and (xy)^2 terms. The corresponding change to 3D would cause it to have 27 total terms(all combinations of a,b,c in {0,1,2} generating (x^a)(y^b)(z^c)). @egpuckett and I haven't found a case so far that has needed these terms (or a number of particles per cell that increases with mesh refinement) to get the expected convergence, but after Rene's comment I wanted to acknowledge that there may be a case where those terms would provide additional accuracy and after finding such case we would want to add those terms. The drawback of adding those terms is the decreased efficiency of using more particles per cell.

bangerth: Oh, I see -- you're projecting onto the space $P_2$, not $Q_2$. I guess that's legitimate.

Can you incorporate a version of your answer here into the comment in question?

egpuckett:
I think this is ready to merge unless there are more requested changes by the reviewers.

bangerth:
Please squash into one and add an entry to `doc/modules/changes/` in the style of the other ones.

sac-bsa:
@bangerth and @gassmoeller, I believe that this is ready for another look.

PR #3675, opened 2020-08-04
Fix .gitignore.

bangerth:
We really do just one `*`, `**` is for recursive matching of directories.


tjhei:
No need to run testers...

PR #3676, opened 2020-08-04
Add the ability to write data on faces only, not in the domain.

bangerth:
Here is something I had started (with #3037) at last year's hackathon and, despite promises to @anne-glerum never finished :-( In any case, better late than never I hope, so here is the finished product: We can now write visualization postprocessors that only generate output on the surface of the domain.

I'm including two such visualization postprocessors: For the stress, and for the dynamic topography. From the perspective of visualization postprocessor writers, it's actually really easy: All you have to do is derive from one additional base class, and everything else happens under the hood.

@anne-glerum -- if you tell me what current viz postprocessor you want to also have in a surface variant, then I will gladly oblige and write it for you!

@cedrict -- as it turns out, Slayer albums also work very well to get extreme programming done :-)


bangerth:
Seems to work now!

anne-glerum:
Hi @bangerth that is great, thanks a lot!! 
Besides stress, strain rate, composition and velocity would be great to have on the surface mesh. Maybe there could be a surface postprocessor that outputs all the solution variables on the surface? Similar to the way that all the solution variables are by default outputted on the whole domain?

include/aspect/postprocess/visualization/surface_dynamic_topography.h
```diff
@@ -0,0 +1,72 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_postprocess_visualization_dynamic_topography_h
+#define _aspect_postprocess_visualization_dynamic_topography_h
```
tjhei, line 23: update include guard?

include/aspect/postprocess/visualization.h
```diff
@@ -269,6 +269,26 @@ namespace aspect
...
+
+
+      /**
+       * This class is a tag class: If a visualization postprocessor is derived
+       * from it, then this is interpreted as saying that the class will only
+       * be used to generate graphical output on the surface of the model,
+       * rather than for the entire domain.
+       */
+      template <int dim>
+      class SurfaceOnlyVisualization
```
tjhei, line 281: with "surface" you mean the boundary of the mesh, not necessarily the surface in the earth science sense, right? Worth clarifying?

source/postprocess/visualization.cc
```diff
@@ -392,11 +397,15 @@ namespace aspect
                      std::is_same<DataOutType,DataOutFaces<dim>>::value,
                      "The only allowed template types of this function are "
                      "DataOut and DataOutFaces.");
+      const bool is_cell_data_output = (typeid(DataOutType) == typeid(DataOut<dim>));
```
tjhei, line 400: same here

source/postprocess/visualization/dynamic_topography.cc
```diff
@@ -139,7 +143,12 @@ namespace aspect
                                                   "within which we produce data for visualization. You probably "
                                                   "only want to visualize whatever data this postprocessor generates "
                                                   "at the surface of your domain and simply ignore the rest of the "
-                                                  "data generated.")
+                                                  "data generated."
+                                                  "\n\n"
+                                                  "Alternatively,  consider using the "
```
tjhei, line 148:
```suggestion
                                                  "Alternatively, consider using the "
```

source/postprocess/visualization/surface_dynamic_topography.cc
```diff
@@ -0,0 +1,145 @@
...
+
+            for (unsigned int q=0; q<quadrature_formula.size(); ++q)
+              computed_quantities[q](0) = dynamic_topography_values[q];
+          }
+      }
+
+
+
+      /**
+       * Register the other postprocessor that we need: SurfaceDynamicTopography
```
tjhei, line 95: typo?

bangerth:
> Besides stress, strain rate, composition and velocity would be great to have 
> on the surface mesh. Maybe there could be a surface postprocessor that outputs 
> all the solution variables on the surface? Similar to the way that all the 
> solution variables are by default outputted on the whole domain?

I was thinking of this as a way to output quantities that genuinely are only 
defined on the boundary. Dynamic topography is a good example: We currently 
extend it by zero into the interior, which is completely arbitrary and hard to 
visualize.

I had no intention to do this for the primary solution variables. (And maybe 
the stress was not a good example either -- I should have output the *normal* 
stress instead.) Why would you want the solution field on the boundary? You 
can get that by just visualizing it as a volume quantity without opacity.

include/aspect/postprocess/visualization.h
bangerth: Ah, that's a good question. It really is the entire "boundary" right now.

I was about to change this to `BoundaryOnly`, but I've just changed my mind -- let me leave it as is right now, and then I'll re-work it in a follow-up to allow selecting boundary indicators on which to actually create output. There is probably no need to have much information output on the lateral boundaries.

In other words: let's leave it as is for now, and I'll get back to that later!

bangerth:
Updated as requested.

bangerth:
Tests correctly now :-)

anne-glerum:
> I was thinking of this as a way to output quantities that genuinely are only defined on the boundary. Dynamic topography is a good example: We currently extend it by zero into the interior, which is completely arbitrary and hard to visualize. 

> I had no intention to do this for the primary solution variables. (And maybe the stress was not a good example either -- I should have output the *normal* stress instead.) Why would you want the solution field on the boundary? You can get that by just visualizing it as a volume quantity without opacity.
One reason would be that with mesh deformation, it is not easy to extract data on the surface (e.g. through paraview) as the surface does not correspond to a certain radius/depth. At the moment, we therefore have to extract data for plotting with e.g. GMT at some arbitrary plane/radius below the minimum value of the surface in paraview. 

Another use-case would be my colleagues that want to feed aspect output (strain rate, strain, topography and maximum horizontal stress) at the surface to a fault extraction algorithm for analysis of fault growth, network formation etc. At the moment, they extract interpolated data at some shallow isotherm (see issue above) or they would have to loop over the whole model domain the find the data at the surface exactly.

anne-glerum:
The HeatFluxMap would also be useful, and this one only makes sense along the top and bottom boundary at the moment :)

tjhei:
I am okay with merging if you follow-up to clarify "boundary" vs "surface" in several places.

bangerth:
Yes, sure. I actually have a good idea of where I want to go with this.

tjhei:
We have unity build errors in this PR. So we should wait and rebase after #3756

anne-glerum:
@bangerth Did I by chance convince you of the use of outputting solution variables (and derivations of those)?

bangerth:
@anne-glerum -- yes, once this is merged I've got plans to restructure things in a way that allows you to output the solutions as well! But I'm an incrementalist -- one step at a time!

bangerth:
@tjhei -- rebased and tests are ok.

PR #3681, opened 2020-08-04
Modify a function to compute reference viscosity

gassmoeller:
This computes the reference viscosity. For now it only tests that the reference viscosity is not too different from the material model reference viscosity.


source/simulator/helper_functions.cc
```diff
@@ -385,9 +385,64 @@ namespace aspect
...
+              local_integrated_viscosity += fe_values.JxW(q)/out.viscosities[q];
+              local_volume += fe_values.JxW(q);
+            }
+        }
+
+    const double integrated_viscosity
+      = Utilities::MPI::sum (local_integrated_viscosity, mpi_communicator);
+
+    const double global_volume
+      = Utilities::MPI::sum (local_volume, mpi_communicator);
```
bangerth, line 439: It doesn't really matter, but you can put the two variables together in an array, sum them in one step and with one communication, and then take them apart again. We do this in most other places where we compute an average.

source/simulator/helper_functions.cc
```diff
@@ -385,9 +385,64 @@ namespace aspect
...
+
+          // Evaluate viscosity at each quadrature point and
+          // calculate the harmonic average
+          for (unsigned int q=0; q<n_q_points; ++q)
+            {
+              Assert(out.viscosities[q] > 0,
+                     ExcMessage ("The viscosity needs to be a "
+                                 "non-negative quantity."));
+
+              local_integrated_viscosity += fe_values.JxW(q)/out.viscosities[q];
```
bangerth, line 430: Your name is wrong -- it's the inverse viscosity

source/simulator/helper_functions.cc
```diff
@@ -385,9 +385,64 @@ namespace aspect
...
+              local_volume += fe_values.JxW(q);
+            }
+        }
+
+    const double integrated_viscosity
+      = Utilities::MPI::sum (local_integrated_viscosity, mpi_communicator);
```
bangerth, line 436: you need one-over here

bangerth:
By the way, in a never-finished PR, I had this comment at the top of the function:
```
+    // Determine how to treat the pressure. We have to scale it for the solver
+    // to make velocities and pressures of roughly the same (numerical) size,
+    // and we may have to fix up the right hand side vector before solving for
+    // compressible models if there are no in-/outflow boundaries
+    //
+    // We do this by scaling the divergence equation by a constant factor that
+    // is equal to a reference viscosity divided by a length scale.
+    // We get the latter from the geometry model, and the former
+    // by looping over all cells and averaging the "order of magnitude"
+    // of the viscosity. The order of magnitude is the logarithm of
+    // the viscosity, so
+    //
+    //   \eta_{ref} = exp ( 1/N * (log(eta_1) + log(eta_2) + ... + log(eta_N))
+    //
+    // where the \eta_i are typical viscosities on the cells of the mesh.
+    // For this, we just take the viscosity at the cell center.
+    //
+    // The formula above computes the exponential of the average of the
+    // logarithms. It is easy to verify that this is equivalent to
+    // computing the *geometric* mean of the viscosities, but the
+    // formula above is numerically stable.
```
Maybe you can incorporate something like this into what you have.

gassmoeller:
All very good suggestions. I had used the harmonic average because I remembered that was our preferred way for intra-cell averaging, but you are correct for the global model an average of the order of magnitude probably makes more sense. I took over your comment, and also made the function a bit more intuitive to use by returning the computed value instead of setting some hidden member variable. I will go through the tests tomorrow and see which ones fail spectacularly :-).

include/aspect/simulator.h
```diff
@@ -484,16 +484,16 @@ namespace aspect
       /**
        * Compute the factor by which we scale the second of
        * the Stokes equations (the "pressure scaling factor").
-       * We do this for the current time step by taking some kind
-       * of average of the viscosities we find on the cells in this domain.
+       * We do this for the current time step by taking the average
+       * of the logarithm of the viscosities we find on the cells in this domain.
```
bangerth, line 488: That's not strictly true -- but maybe call it "logarithmic average" or something similar.

source/simulator/helper_functions.cc
```diff
@@ -385,9 +385,64 @@ namespace aspect
...
+          fe_values.reinit (cell);
+          in.reinit(fe_values,
+                    cell,
+                    introspection,
+                    solution);
+
+          // We do not average the output, because  it is not
+          // necessary for a single quadrature point of QMidpoint
+          // (and it fails for Q1 averaging)
+          material_model->evaluate(in, out);
```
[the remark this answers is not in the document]
bangerth: Ah, I see now -- it's also funny because there is really nothing to average if you have only one quadrature point ;-)

source/simulator/helper_functions.cc
```diff
@@ -381,13 +381,87 @@ namespace aspect
...
+              const double JxW = fe_values.JxW(q);
+              local_integrated_viscosity_logarithm += std::log(out.viscosities[q]) * JxW;
+              local_volume += JxW;
+            }
+        }
+
+    // vector for packing local values before MPI summing them
+    double values[2] = {local_integrated_viscosity_logarithm, local_volume};
+
+    Utilities::MPI::sum<double, 2>(values, mpi_communicator, values);
```
bangerth, line 460: Do you actually need the explicit template arguments?

gassmoeller:
I went through the test results a few days ago and only saw significant differences in models where the reference viscosity so far was significantly wrong. Lots of minor iteration count changes (and following changes in residuals). I was planning to update the test results again next week when things have cooled down a bit and go through the results a bit more carefully, but then this should be ready.

bangerth:
OK once you can convince yourself that the results look reasonable.

anne-glerum:
This PR is so useful, thanks @gassmoeller! We've also had earthquake cycle models that fail during the event, but when restarted with a reference viscosity more appropriate for the viscosities during the event, do run through. I'm hoping this PR will prevent the fail in the first place!

anne-glerum:
@gassmoeller I had a look and the following tests failed for these reasons:
401 - melt_transport_compressible → timeout                       
445 - nonlinear_channel_flow_velocities_Newton_Stokes_GMG → aborted with non-convergence Stokes solve Timestep 0 
446 - nonlinear_channel_flow_velocities_Newton_Stokes_no_derivative → aborted with non-convergence Stokes solve Timestep 0
448 - nonlinear_channel_flow_velocities_Newton_Stokes_no_deviator → aborted with non-convergence Stokes solve Timestep 1
449 - nonlinear_channel_flow_velocities_Newton_Stokes_no_line_search → aborted with non-convergence Stokes solve Timestep 4
644 - smoothing → aborted with negative viscosity error
694 - steinberger_viscosity_adiabatic → aborted with non-convergence 1st Stokes solve
700 - supg → aborted with negative viscosity error
754 - visco_plastic_peierls_mei_exact → less Stokes solves iterations (14 ipv 20) in log.txt and statistics
828 - zero_matrix (Failed) → is_finite() = false

I'll rebase the PR and fix the tests.

anne-glerum:
I've undone your latest commit to update the test results, checkout a copy of your branch and rebased to master. After this, the following tests no longer fail to converge:

	448 - nonlinear_channel_flow_velocities_Newton_Stokes_no_deviator → no longer aborted
	449 - nonlinear_channel_flow_velocities_Newton_Stokes_no_line_search → no longer aborted

The following tests don't reach the required nonlinear tolerance before not converging in the inner iterations, but do reach a good residual, ~3e-10:
	445 - nonlinear_channel_flow_velocities_Newton_Stokes_GMG
	446 - nonlinear_channel_flow_velocities_Newton_Stokes_no_derivative
They can be fixed by loosening the linear or nonlinear solver tolerance, or increasing the Maximum expensive solver steps. Also, the computed reference viscosity is not a good guess, because the viscosity is strain rate dependent, but the minimum strain rate is left at the default 1.96e-40, which is not a good guess. Hence the `computed_reference_viscosity`is the maximum allowed viscosity (1e24 instead of ~5e21). A minimum strain rate of 1e-16 gives a better viscosity.

These two tests have a viscosity set to 0 and a reference viscosity set to 1e23. Although Stokes is not solved, the pressure scaling factor is computed and this requires a positive reference viscosity. Setting the viscosity to 1 lets the test run.
	644 - smoothing → viscosity needs to be set to something other than 0 (before, the default reference viscosity was taken, now the average of the viscosity, which is set to 0)
	700 - supg → set Viscosity to 1

This test I'm not sure about:
	694 - steinberger_viscosity_adiabatic (Failed) → aborted with non-convergence 1st Stokes solve → the `computed_reference_viscosity` 5.8e21 is not that far off from the old `reference_viscosity` of 1e21, but this does decide convergence or not. The original test does also require a lot of inner iterations (200+300). At higher resolution (6ipv3), the current test does converge.

This test gets the expected fail, but the error output in screen-output is different
	828 - zero_matrix → fail expected

This test just needs to be updated
	754 - visco_plastic_peierls_mei_exact (Failed) → less Stokes solves iterations (14 ipv 20) in log.txt and statistics

gassmoeller:
I rebased the PR and fixed the two tests `nonlinear_channel_flow_velocities_Newton_Stokes_GMG` (by loosening the linear solver tolerance it converges, and it does not affect the nonlinear convergence), and `steinberger_viscosity_adiabatic` (by optimizing linear solver properties towards convergence over speed). 

I have lost track a bit if #3916 already takes care of all the remaining tests that needed fixing, @anne-glerum could you take a look at what I did in here, and if this is now ready?

anne-glerum:
Hi @gassmoeller, the tests that were failing prior to #3916 and rebase have all been updated here or in #3916:
	401 - melt_transport_compressible → updated here, but many more nonlinear iterations, don't know if that's a problem.             
	445 - nonlinear_channel_flow_velocities_Newton_Stokes_GMG → #3916 & here
	446 - nonlinear_channel_flow_velocities_Newton_Stokes_no_derivative → #3916 + updated here
	448 - nonlinear_channel_flow_velocities_Newton_Stokes_no_deviator → #3916 + updated here
	449 - nonlinear_channel_flow_velocities_Newton_Stokes_no_line_search → #3916 + updated here
	644 - smoothing → #3916 
	694 - steinberger_viscosity_adiabatic → here
	700 - supg → #3916 
	754 - visco_plastic_peierls_mei_exact → here
	828 - zero_matrix →no longer an issue

gassmoeller:
Ok, I fixed the remaining tests for this PR. Should be ready to merge from my side.

gassmoeller:
Could we merge this PR in the coming days? A wrong reference viscosity often leads to poor solver performance, and we keep forgetting to finish this PR.

tests/ascii_data_boundary_velocity_2d_box_time_non_equidistant/screen-output
```diff
@@ -39,7 +39,7 @@ Number of degrees of freedom: 1,212 (738+105+369)
      Heat fluxes through boundary parts: -1.492e+08 W, 1.261e+08 W, 4.946e+05 W, 3.942e+05 W
 
 *** Timestep 3:  t=203329 years, dt=79206.9 years
-   Solving temperature system... 32 iterations.
+   Solving temperature system... 48 iterations.
```
anne-glerum, line 42: Why would just the temperature iterations change?

tests/ascii_data_initial_composition_2d_box/screen-output
```diff
@@ -115,7 +115,7 @@ Number of degrees of freedom: 349 (162+25+81+81)
 
 *** Timestep 9:  t=732284 years, dt=80478.7 years
    Solving temperature system... 20 iterations.
-   Solving C_1 system ... 70 iterations.
+   Solving C_1 system ... 58 iterations.
```
anne-glerum, line 118: Why would C_1 iterations change?

anne-glerum (commented):
Hi @gassmoeller , I just had a quick look and apart from a comment that could be clarified, I think this is ready to go, as Wolfgang and Timo also mentioned. I was wondering whether you went through the test changes though. I randomly looked at a few and commented on some changes, but if you can explain all the diffs, then I think this can be merged.

gassmoeller:
Hi @anne-glerum thanks for the review! I rebased to the latest master and clarified the comment. I also updated the analytic_artificial_viscosity test to terminate after 2 timesteps (like before), and fixed the melt_transport_compressible test which did not reach its nonlinear tolerance anymore (just barely, I think previously it just barely converged).

I can not explain why we have a very few tests that now need a different number of advection iterations, while everything else including the results remains unchanged. I could imagine the stokes solution is minimally different because of the new pressure scaling, and this slight difference changes the advection equation, but it seems unlikely. Anyway, since the results remain unchanged and all other tests behave as expected I do not think I can do anything about that. Do you have another idea?

If there are no other comments I would really like to get this into 2.3, because it helps massively with a lot of models.

anne-glerum (approved):
Hi Rene, looks good! I did not have another idea to explain the changes to the advection solves. I think this can be merged once the tester is happy.

tjhei (approved):

gassmoeller:
Tester looks good so I am going to merge. Thanks all!

PR #3684, opened 2020-08-05
Fixes and updates to citing_aspect.bib

ljhwang:
This has a numerous miscellaneous corrections discovered while parsing the citing_aspect.bib file. 

Changes include:
- adding missing commas in multiple places
- adding 'school' attribute to Robey Thesis
- changing Bangerth2017a from type techreport to misc to be consistent with figshare (note consistent with manual.bib)
- updated Muluneh 2020

### DIFF
--- a/doc/manual/citing_aspect.bib
+++ b/doc/manual/citing_aspect.bib
@@ -233,7 +233,7 @@ year = {2017}
   publisher={Nature Publishing Group}
 }
 @article{ONeill2018inception,
-  doi={10.1098/rsta.2017.0414}
+  doi={10.1098/rsta.2017.0414},
   title={The inception of plate tectonics: a record of failure},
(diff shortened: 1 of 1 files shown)


tjhei:
Awesome! I just realized that the script generating https://aspect.geodynamics.org/publications.html was broken since May because of these missing entries. Sigh.

tjhei:
I would think it makes sense to update the aspectmanual entry to be identical between the two bib files. Any reason you don't want to make that change?

ljhwang:
@tjhei - dinner. I will make a separate PR to align the 2 now that this is clear.

bangerth:
Let's get these PRs out of the way and merged.

@ljhwang -- if you wanted to update something else to match, please open a separate PR!

tjhei:
I can confirm that https://aspect.geodynamics.org/publications.html is updating again. Thanks, @ljhwang ! :+1:

PR #3686, opened 2020-08-05
AsciiDataLookup: allow directly initializing

tjhei:
Allow specifying the data in the AsciiDataLookup directly via a call to
reinit() instead of going through parsing a .txt file. This will allow
us to offer different ways to load data (binary, etc.) and doesn't
require the opendap reader to generate a text document.

- extract the initialization from load_file into new reinit()
- let load_file use the new function
- check coordinate data more rigorously, fix data files 
- add unit tests
- add a second argument to the private compute_table_indices function as
the dimensions are not set when it is used in load_file.


tjhei:
@kneumiller FYI.
@gassmoeller Would it make sense to change the interface of reinit() to just require a single list of coordinate values per coordinate axis (instead of dim times table<dim>)?

gassmoeller:
I think that depends a bit on the format of the Netcdf data we want to put in, but I forgot how that is structured. What is the most natural way to take that data and put into this function?

tjhei:
Fixed.

include/aspect/utilities.h
```diff
@@ -604,6 +604,25 @@ namespace aspect
...
+         *
+         * The data consists of @p n_components (implicitly given by the size of @p column_names)
+         * specified at @p n_points_per_direction in each of the dim coordinate directions.
+         *
+         * The data in @p raw_data consists of dim+n_components number of tables, where the
+         * first dim tables are the coordinate values (raw_data[0] is the table of x
+         * coordinates, raw_data[1] is the table of y coordinates, etc..) and the remaining
+         * columns are the data columns.
+         */
+        void reinit(const TableIndices<dim> &n_points_per_direction,
```
bangerth, line 621: do you actually need this parameter? I'd think you can get that as the size of one of the data tables.

include/aspect/utilities.h
```diff
@@ -604,6 +604,25 @@ namespace aspect
...
+                    const std::vector<std::string> &column_names,
+                    const std::vector<Table<dim,double> > &raw_data
```
bangerth, line 623: I don't understand your choice of data format.

If I understand this right, then you have a 3d collection of data points, at each of which you store `n_components` data values. But these 3d points are not arbitrarily distributed: they're either uniformly or non-uniformly spaced along each axis, but they're tensor products of x-, y-, and z-spaced points. (That's necessary because the classes that interpolated are either InterpolatedUniformGridData or InterpolatedTensorProductGridData).

So I don't understand why you would want to store the x, y, and z-coordinates of all points as 3d tables: That's highly redundant. All you need to provide are `dim` vectors with abscissae, and that's also what needs to be input into the two classes mentioned above.

source/utilities.cc
```diff
@@ -1724,11 +1724,127 @@ namespace aspect
...
+      // We also check the requirement that the coordinates are
+      // strictly ascending.
+
+      // The number of intervals in each direction
+      std::array<unsigned int,dim> table_intervals;
+
+      // Whether or not the grid is equidistant
+      coordinate_values_are_equidistant = true;
+
+      for (unsigned int i = 0; i < dim; i++)
```
bangerth, line 1771: You can save yourself all of this if you used a better input format.

bangerth: Separately, it would be easier to read if you had chosen `d` as loop index over dimensions...

source/utilities.cc
```diff
@@ -1724,11 +1724,127 @@ namespace aspect
...
+        {
+          table_intervals[i] = table_points[i] - 1;
+
+          TableIndices<dim> idx;
+          double temp_coord = raw_data[i](idx);
```
bangerth, line 1776:
```suggestion
          double temp_coord = raw_data[i](TableIndices<dim>());
```

source/utilities.cc
```diff
@@ -1724,11 +1724,127 @@ namespace aspect
...
+      std::array<unsigned int,dim> table_intervals;
+
+      // Whether or not the grid is equidistant
+      coordinate_values_are_equidistant = true;
+
+      for (unsigned int i = 0; i < dim; i++)
+        {
+          table_intervals[i] = table_points[i] - 1;
+
+          TableIndices<dim> idx;
```
bangerth, line 1775:
```suggestion
```

source/utilities.cc
```diff
@@ -1724,11 +1724,127 @@ namespace aspect
...
+          coordinate_values[i].clear();
+          coordinate_values[i].push_back(temp_coord);
+
+          // The grid spacing
+          double grid_spacing = numbers::signaling_nan<double>();
+
+          // Loop over the rest of the coordinate points
+          for (unsigned int n = 1; n < table_points[i]; n++)
+            {
+              idx[i] = n;
```
bangerth, line 1792:
```suggestion
              TableIndices<dim> idx;
              idx[i] = n;
```

source/utilities.cc
```diff
@@ -1724,11 +1724,127 @@ namespace aspect
...
+
+              // Set the coordinate value
+              coordinate_values[i].push_back(new_temp_coord);
+
+              temp_coord = new_temp_coord;
+            }
+
+          // The maximum coordinate
+          grid_extent[i].second = temp_coord;
+        }
```
bangerth, line 1819: The code you have up there doesn't actually check whether the data is tensor product. It really only walks down the first row, column, pillar, and never checks any of the other (hopefully redundant) coordinate values. I think that's ok, but it illustrates that the input data format is not well chosen: It allows for redundancies, which you never check, and make implicit assumptions about.

include/aspect/utilities.h
tjhei: > So I don't understand why you would want to store the x, y, and z-coordinates of all points as 3d tables

I did not come up with this. This is extracted from ``load_file`` without changes. I agree that this not great (and commented about this at the top of this PR.

source/utilities.cc
tjhei: agreed. Again, just moved from load_file.

include/aspect/utilities.h
bangerth: Hm, what do others think? I'm a bit torn between saying "that's how it always was" and saying "if you're already rewriting it, let's do this right".

tjhei:
Okay, reworked everything based on your comments. I now also check that the coordinate values are consistent in the txt files.

tjhei:
> and never checks any of the other (hopefully redundant) coordinate values. I think that's ok, but it illustrates that the input data format is not well chosen:

... and of course some of our files were in fact broken. :-)

bangerth:
On 8/7/20 5:43 PM, Timo Heister wrote:
> and never checks any of the other (hopefully redundant) coordinate values. 
> I think that's ok, but it illustrates that the input data format is not 
> well chosen:
> 
> ... and of course some of our files were in fact broken. :-)

As I like to quip: "It's not that stuff that isn't tested *may* be buggy. 
Stuff that isn't test *is* buggy."

source/utilities.cc
```diff
@@ -1724,11 +1724,103 @@ namespace aspect
...
+            {
+              const double current_grid_spacing = coordinate_values[d][n] - coordinate_values_[d][n-1];
+
+              AssertThrow(current_grid_spacing > 0.0,
+                          ExcMessage ("Coordinates in dimension "
+                                      + Utilities::int_to_string(d)
+                                      + " are not strictly ascending. "));
+
+              if (std::abs(current_grid_spacing - grid_spacing) > 0.005*(current_grid_spacing+grid_spacing))
+                coordinate_values_are_equidistant = false;
```
bangerth, line 1791: You can `break` here -- it's never going back to `true` :-)

source/utilities.cc
```diff
@@ -1724,11 +1724,103 @@ namespace aspect
...
+        {
+          table_intervals[d] = table_points[d]-1;
+
+          // The minimum and maximum coordinate values:
+          grid_extent[d].first = coordinate_values[d][0];
+          grid_extent[d].second = coordinate_values[d][table_points[d]-1];
+
+          const double grid_spacing = coordinate_values[d][1] - coordinate_values[d][0];
+
+          for (unsigned int n = 1; n < table_points[d]; n++)
```
bangerth, line 1781:
```suggestion
          for (unsigned int n = 1; n < table_points[d]; ++n)
```

source/utilities.cc
```diff
@@ -1724,11 +1724,103 @@ namespace aspect
...
+            {
+              const double current_grid_spacing = coordinate_values[d][n] - coordinate_values_[d][n-1];
+
+              AssertThrow(current_grid_spacing > 0.0,
+                          ExcMessage ("Coordinates in dimension "
+                                      + Utilities::int_to_string(d)
+                                      + " are not strictly ascending. "));
+
+              if (std::abs(current_grid_spacing - grid_spacing) > 0.005*(current_grid_spacing+grid_spacing))
```
bangerth, line 1790: Is there a good reason for the 0.005? I guess that's good enough if you give things to three digits.

source/utilities.cc
```diff
@@ -1724,11 +1724,103 @@ namespace aspect
...
+                coordinate_values_are_equidistant = false;
+            }
+        }
```
bangerth, line 1793: And if the flag is false, you can also `break` out of the outer loop.

source/utilities.cc
```diff
@@ -1724,11 +1724,103 @@ namespace aspect
...
+
+
     template <int dim>
     void
     AsciiDataLookup<dim>::load_file(const std::string &filename,
                                     const MPI_Comm &comm)
     {
+      // grab the values already stored in this class (if they exist), this way we can
+      // check if somebody changes the size of the table over time and error out (see below)
+      TableIndices<dim> table_points = this->table_points;
```
bangerth, line 1821: Uh, that's going to confuse someone at some point. Just call the new variable `new_table_points` or something like that.

source/utilities.cc
```diff
@@ -1812,43 +1904,65 @@ namespace aspect
...
 
+      if (column_names.size()==0)
+        {
+          // set default column names:
+          for (unsigned int c=0; c<components; ++c)
+            column_names.push_back("column " + Utilities::int_to_string(c,2));
+        }
 
-      // Read data lines
+      // Finaylly read data lines
```
bangerth, line 1937:
```suggestion
      // Finally read data lines
```

source/utilities.cc
```diff
@@ -1812,43 +1904,65 @@ namespace aspect
...
+              // This is a coordinate. Store (and check that they are consistent)
+              const double old_value = coordinate_values[column_num][idx[column_num]];
 
-          data_tables[column_num](compute_table_indices(read_data_entries)) = temp_data;
+              AssertThrow(old_value == 0. ||
+                          (std::abs(old_value-temp_data) < 1e-8*std::abs(old_value)),
+                          ExcMessage("Invalid coordinate "
+                                     + Utilities::int_to_string(column_num) + " in row "
+                                     + Utilities::int_to_string(row_num)
+                                     + " in file " + filename));
```
bangerth, line 1956: Can you clarify in this error message that the problem is that the new position has moved compared to the old position?

bangerth:
Getting closer!

source/utilities.cc
tjhei: Yes, but I want to run through and do the AssertThrow check.

source/utilities.cc
tjhei: No, grid_extent and the AssertThrow should run through everything.

source/utilities.cc
tjhei: ask @gassmoeller ?

source/utilities.cc
tjhei: That is a good point. Fixed.

tjhei:
updated.

source/utilities.cc
bangerth: Ah, I see. Can you add a comment here then why you're not `break`ing right away?

bangerth:
OK to merge if you add the one comment.

tjhei:
sure, done.

tjhei:
Yay!

PR #3687, opened 2020-08-05
Fixed default viscosity file for depth dependent model

mnkajan:
I noticed that the depth dependent material model has a default viscosity depth file that doesn't exist. I switched it to the test data from the `depth_dependent_box_file_simple.prm` test which I think is more helpful for users.

Also, do we typically store sample data under the tests folder like this? I left things as is for now, but I was wondering if the data should be stored somewhere under `$ASPECT_SOURCE_DIR/data/material-model/` instead.


mnkajan:
I was thinking the same thing, so it's no problem at all. 🙂 I'll make these changes this afternoon!

mnkajan:
@gassmoeller I switched the pull request to a draft while I was still working on things (there's an option to do so in the sidebar of an open pull request) and then clicked "ready for review" in the main conversation section here once I was finished cleaning up the commits on my own branch. Definitely useful!

PR #3691, opened 2020-08-05
Add post nonlinear solver signal

gassmoeller:
This is mostly here for testing purposes and for showing this to @jdannberg and @tjhei for the time stepping discussion. I will update the PR for the other solver schemes and see if any tests change (they should not).

### DIFF
--- a/include/aspect/simulator_signals.h
+++ b/include/aspect/simulator_signals.h
@@ -251,6 +251,17 @@ namespace aspect
                                   const SolverControl &solver_control)> post_advection_solver;
 
     /**
+     * A signal that is fired when the nonlinear solver scheme is done.
+     * The signal parameter is an object that contains information
+     * about the final state (failure/success), number of
+     * iterations and history of residuals of the nonlinear solver.
+     * If there is no nonlinear solver (only a single solve), the
+     * SolverControl object will report a successful state, a single iteration
+     * and a remaining residual of zero.
+     */
+    boost::signals2::signal<void (const SolverControl &)> post_nonlinear_solver;
+
+    /**
      * A signal that is fired at the end of the set_assemblers() function that
      * allows modification of the assembly objects active in this simulation.
      */
--- a/source/simulator/solver_schemes.cc
+++ b/source/simulator/solver_schemes.cc
@@ -678,6 +678,12 @@ namespace aspect
 
     if (parameters.run_postprocessors_on_nonlinear_iterations)
       postprocess ();
+
+    // Setup a nonlinear solver control that only allows a single iteration
+    SolverControl nonlinear_solver_control(1,1.0);
+    // Announce that we did a single iteration, and assume we have converged
+    nonlinear_solver_control.check(1,0.0);
+    signals.post_nonlinear_solver(nonlinear_solver_control);
   }
 
 
@@ -694,25 +700,30 @@ namespace aspect
                parameters.max_nonlinear_iterations_in_prerefinement)
       :
       parameters.max_nonlinear_iterations;
+
+    SolverControl nonlinear_solver_control(max_nonlinear_iterations,
+                                           parameters.nonlinear_tolerance);
+
+    double relative_residual = std::numeric_limits<double>::max();
+    nonlinear_iteration = 0;
     do
       {
-        const double relative_nonlinear_stokes_residual =
+        relative_residual =
           assemble_and_solve_stokes(nonlinear_iteration == 0, &initial_stokes_residual);
 
         pcout << "      Relative nonlinear residual (Stokes system) after nonlinear iteration " << nonlinear_iteration+1
-              << ": " << relative_nonlinear_stokes_residual
+              << ": " << relative_residual
               << std::endl
               << std::endl;
 
         if (parameters.run_postprocessors_on_nonlinear_iterations)
           postprocess ();
 
-        if (relative_nonlinear_stokes_residual < parameters.nonlinear_tolerance)
-          break;
-
         ++nonlinear_iteration;
       }
-    while (nonlinear_iteration < max_nonlinear_iterations);
+    while (nonlinear_solver_control.check(nonlinear_iteration, relative_residual) == SolverControl::iterate);
+
+    signals.post_nonlinear_solver(nonlinear_solver_control);
   }
 
 
@@ -723,6 +734,12 @@ namespace aspect
 
     if (parameters.run_postprocessors_on_nonlinear_iterations)
       postprocess ();
+
+    // Setup a nonlinear solver control that only allows a single iteration
+    SolverControl nonlinear_solver_control(1,1.0);
+    // Announce that we did a single iteration, and assume we have converged
+    nonlinear_solver_control.check(1,0.0);
+    signals.post_nonlinear_solver(nonlinear_solver_control);
   }
 
   template <int dim>
@@ -730,6 +747,12 @@ namespace aspect
   {
     if (timestep_number == 0)
       assemble_and_solve_stokes();
+
+    // Setup a nonlinear solver control that only allows a single iteration
+    SolverControl nonlinear_solver_control(1,1.0);
+    // Announce that we did a single iteration, and assume we have converged
+    nonlinear_solver_control.check(1,0.0);
+    signals.post_nonlinear_solver(nonlinear_solver_control);
   }
 
 
@@ -749,6 +772,11 @@ namespace aspect
       :
       parameters.max_nonlinear_iterations;
 
+    SolverControl nonlinear_solver_control(max_nonlinear_iterations,
+                                           parameters.nonlinear_tolerance);
+
+    double relative_residual = std::numeric_limits<double>::max();
+    nonlinear_iteration = 0;
     do
       {
         const double relative_temperature_residual =
@@ -767,7 +795,8 @@ namespace aspect
         pcout << ", " << relative_nonlinear_stokes_residual;
         pcout << std::endl;
 
-        double max = 0.0;
+        // Find the maximum residual of the individual equations
+        relative_residual = relative_temperature_residual;
         for (unsigned int c=0; c<introspection.n_compositional_fields; ++c)
           {
             // in models with melt migration the melt advection equation includes the divergence of the velocity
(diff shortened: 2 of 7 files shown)


source/simulator/solver_schemes.cc
```diff
@@ -672,12 +672,22 @@ namespace aspect
   template <int dim>
   void Simulator<dim>::solve_single_advection_single_stokes ()
   {
+    // Setup a nonlinear solver control that only allows a single iteration
+    SolverControl nonlinear_solver(1,
```
tjhei, line 676: Using a SolverControl is a good idea!

include/aspect/simulator_signals.h
```diff
@@ -250,6 +250,16 @@ namespace aspect
...
                                   const SolverControl &solver_control)> post_advection_solver;
 
+    /**
+    * A signal that is fired when the nonlinear solver scheme is done.
+    * Parameters are a reference to the SimulatorAccess, and an 
+    * information object that contains information
+    * about the final state (failure/success), number of 
+    * iterations and history of residuals.
+    */
+    boost::signals2::signal<void (const SimulatorAccess<dim> &,
```
tjhei, line 260: We normally don't pass SimulatorAccess to the signals because there are other ways to get access to it. Maybe remove?

gassmoeller:
I added the solver control and signal to all solver schemes, which also allowed me to unify the nonlinear iteration loop in the schemes. I think the only thing left to do (tomorrow) is to add a test where the nonlinear solver fails and check that the state and residual are reported correctly.

source/simulator/solver_schemes.cc
```diff
@@ -694,25 +700,30 @@ namespace aspect
                parameters.max_nonlinear_iterations_in_prerefinement)
       :
       parameters.max_nonlinear_iterations;
-    do
+
+    SolverControl nonlinear_solver_control(max_nonlinear_iterations,
+                                           parameters.nonlinear_tolerance,
+                                           true,
+                                           true);
```
tjhei, line 707: I would remove the two bool args, the defaults are fine (here and in all other places).

source/simulator/solver_schemes.cc
```diff
@@ -908,24 +952,25 @@ namespace aspect
...
-        if (dcr.residual/dcr.initial_residual < parameters.nonlinear_tolerance)
-          break;
+        postprocess ();
       }
 
-    // Reset the Newton stabilization at the end of the timestep.
-    newton_handler->parameters.preconditioner_stabilization = starting_preconditioner_stabilization;
-    newton_handler->parameters.velocity_block_stabilization = starting_velocity_block_stabilization;
+  // Reset the Newton stabilization at the end of the timestep.
+  newton_handler->parameters.preconditioner_stabilization = starting_preconditioner_stabilization;
```
tjhei, line 962: what happened with the indentation here?

include/aspect/simulator_signals.h
```diff
@@ -250,6 +250,14 @@ namespace aspect
                                   const unsigned int compositional_index,
                                   const SolverControl &solver_control)> post_advection_solver;
 
+    /**
+    * A signal that is fired when the nonlinear solver scheme is done.
```
bangerth, line 254: Might want to align the asterisks.

include/aspect/simulator_signals.h
```diff
@@ -250,6 +250,14 @@ namespace aspect
...
+    * The parameter is an object that contains information
+    * about the final state (failure/success), number of
+    * iterations and history of residuals of the nonlinear solver.
+    */
+    boost::signals2::signal<void (const SolverControl &)> post_nonlinear_solver;
```
bangerth, line 259: I think you want to document that this signal is also called for the linear schemes that do not actually iterate, and in that case will simply record a single iteration with a success outcome.

source/simulator/solver_schemes.cc
```diff
@@ -694,25 +700,30 @@ namespace aspect
...
+
+    SolverControl nonlinear_solver_control(max_nonlinear_iterations,
+                                           parameters.nonlinear_tolerance,
+                                           true,
+                                           true);
+
+    double relative_residual = std::numeric_limits<double>::max();
+    for (nonlinear_iteration = 0;
+         nonlinear_solver_control.check(nonlinear_iteration, relative_residual) == SolverControl::iterate;
+         ++nonlinear_iteration)
```
bangerth, line 712: I think this would be easier to read as a `while` loop:
```
  nonlinear_iteration = 0;
  while (control.check(...) == ...)
  {
    ...
    ++nonlinear_iteration;
  }
```

bangerth: Or maybe even easier, via the `do ... while(...)` construct.

source/simulator/solver_schemes.cc
```diff
@@ -749,7 +772,15 @@ namespace aspect
...
-    do
+    SolverControl nonlinear_solver_control(max_nonlinear_iterations,
+                                           parameters.nonlinear_tolerance,
+                                           true,
+                                           true);
+
+    double relative_residual = std::numeric_limits<double>::max();
+    for (nonlinear_iteration = 0;
+         nonlinear_solver_control.check(nonlinear_iteration, relative_residual) == SolverControl::iterate;
+         ++nonlinear_iteration)
```
bangerth, line 783: same here

source/simulator/solver_schemes.cc
```diff
@@ -822,26 +849,29 @@ namespace aspect
...
```
bangerth, line 860: and here

source/simulator/solver_schemes.cc
```diff
@@ -882,13 +912,27 @@ namespace aspect
...
+
+    SolverControl nonlinear_solver_control_picard(newton_handler->parameters.max_pre_newton_nonlinear_iterations,
+                                                  newton_handler->parameters.nonlinear_switch_tolerance,
+                                                  true,
+                                                  true);
+
+    double relative_residual = std::numeric_limits<double>::max();
+    for (nonlinear_iteration = 0;
+         nonlinear_solver_control.check(nonlinear_iteration, relative_residual) == SolverControl::iterate;
+         ++nonlinear_iteration)
```
bangerth, line 929: and here

source/simulator/solver_schemes.cc
bangerth: I think you lost a closing scope brace?

source/simulator/solver_schemes.cc
```diff
@@ -966,13 +1011,28 @@ namespace aspect
...
+
     // Now iterate out the nonlinearities.
     dcr.stokes_residuals = std::pair<double,double>  (numbers::signaling_nan<double>(),
                                                       numbers::signaling_nan<double>());
-    for (nonlinear_iteration = 0; nonlinear_iteration < max_nonlinear_iterations; ++nonlinear_iteration)
+
+    double relative_residual = std::numeric_limits<double>::max();
+    for (nonlinear_iteration = 0;
+         nonlinear_solver_control.check(nonlinear_iteration, relative_residual) == SolverControl::iterate;
+         ++nonlinear_iteration)
```
bangerth, line 1031: and here

gassmoeller:
I think I addressed all comments. To be consistent with our current test results I had to create one `while` loop and all other solver schemes as `do ... while` loops, because some tests rely on the fact that no iteration is done if `max_nonlinear_iterations == 0`.

source/simulator/solver_schemes.cc
```diff
@@ -882,13 +912,23 @@ namespace aspect
...
+    SolverControl nonlinear_solver_control(max_nonlinear_iterations,
+                                           parameters.nonlinear_tolerance);
+
+    SolverControl nonlinear_solver_control_picard(newton_handler->parameters.max_pre_newton_nonlinear_iterations,
+                                                  newton_handler->parameters.nonlinear_switch_tolerance);
+
+    double relative_residual = std::numeric_limits<double>::max();
+    nonlinear_iteration = 0;
+
+    while (nonlinear_solver_control.check(nonlinear_iteration, relative_residual) == SolverControl::iterate)
```
tjhei, line 925: why is this one a while loop while you use do ... while in all other places?

gassmoeller: Because a) it was that way before as well (it was a for loop before), and b) we have some tests that assume this is a while loop (they set 0 max nonlinear iterations and expect to never go into the loop) and other tests that require all other schemes to not be a while loop (they require that the check is only executed after the nonlinear solve, and the number of nonlinear residuals recorded is 1 after 1 solve). I could make it work with consistent do .. while loops if I added a special condition that skipped the loop if max_nonlinear_iterations == 0, would you prefer that instead to keep the loops consistent?

tjhei: I would think setting the nonlinear iterations to 0 is a bug and should fail. What do you think?

source/simulator/solver_schemes.cc
```diff
@@ -678,6 +678,12 @@ namespace aspect
 
     if (parameters.run_postprocessors_on_nonlinear_iterations)
       postprocess ();
+
+    // Setup a nonlinear solver control that only allows a single iteration
+    SolverControl nonlinear_solver_control(1,1.0);
+    // Announce that we did a single iteration, and assume we have converged
+    nonlinear_solver_control.check(0,0.0);
```
bangerth, line 685: I think you want
```suggestion
    nonlinear_solver_control.check(1, 0.0);
```
to indicate that *one* iteration was actually done.

source/simulator/solver_schemes.cc
bangerth: Is that the total number of nonlinear steps (Picard + Newton)? That seems wrong -- what is the solution vector supposed  to be if that happens?

source/simulator/solver_schemes.cc
```diff
@@ -1059,6 +1115,12 @@ namespace aspect
...
```
bangerth, line 1122: same here

source/simulator/solver_schemes.cc
```diff
@@ -1068,6 +1130,12 @@ namespace aspect
   {
     if (parameters.run_postprocessors_on_nonlinear_iterations)
       postprocess ();
+
+    // Setup a nonlinear solver control that only allows a single iteration
+    SolverControl nonlinear_solver_control(1,1.0);
+    // Announce that we did a single iteration, and assume we have converged
+    nonlinear_solver_control.check(0,0.0);
```
bangerth, line 1137: and here

tjhei:
I will give @bangerth a chance to comment before merging.

bangerth:
Looks good, thanks!

Issue #3549, opened 2020-06-22
More control on timestepping logic


anne-glerum:
As discussed in the biweekly ASPECT meeting, it would be great to have a plugin structure for timestepping control. An example would be @tjhei 's implementation to redo a timestep if the mesh resolution in the shear bands is not the same throughout the bands. Another application would be to assess the solution for a specific timestep size and recompute the timestep for a smaller timestep size if a certain threshold was exceeded.

tjhei:
Thanks for the reminder Anne.
I put things on hold to work on the release for now, but I should put my ideas down before I forget:

- related is #2744 (redo timestep if nonlinear solver fails)
- scenario: if a check determines that refinement is needed (run every time step?), refine and redo timestep
- scenario: if the solution (for example stresses exceed threshold) redo using smaller time step

PR #3692, opened 2020-08-05
Introduce TimeStepping::Manager

tjhei:
This is step 1 of N to introduce a new facility to handle the task to
    determinate time step sizes and allow redoing time steps for various
    reasons.
    This PR creates a new TimeStepping::Manager class without changing
    functionality:
    
    - Introduce TimeStepping::Manager that contains the TerminationManager
    - move do_checkpoint_on_terminate to it
    - simplify core.cc termination handling

part of #3549

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -61,7 +61,7 @@ DEAL_II_ENABLE_EXTRA_DIAGNOSTICS
 #include <aspect/boundary_fluid_pressure/interface.h>
 #include <aspect/boundary_traction/interface.h>
 #include <aspect/mesh_refinement/interface.h>
-#include <aspect/termination_criteria/interface.h>
+#include <aspect/time_stepping/interface.h>
 #include <aspect/postprocess/interface.h>
 #include <aspect/adiabatic_conditions/interface.h>
 #include <aspect/particle/world.h>
@@ -1467,14 +1467,14 @@ namespace aspect
       void maybe_write_timing_output () const;
 
       /**
-       * Check if a checkpoint should be written in this timestep. Is so create
+       * Check if a checkpoint should be written in this timestep. If so create
        * one. Returns whether a checkpoint was written.
        *
        * This function is implemented in
        * <code>source/simulator/helper_functions.cc</code>.
        */
       bool maybe_write_checkpoint (const time_t last_checkpoint_time,
-                                   const std::pair<bool,bool> termination_output);
+                                   const bool force_writing_checkpoint);
 
       /**
        * Check if we should do an initial refinement cycle in this timestep.
@@ -1504,18 +1504,6 @@ namespace aspect
        */
       void maybe_refine_mesh (const double new_time_step,
                               unsigned int &max_refinement_level);
-
-      /**
-       * Compute the size of the next time step from the mesh size and the
-       * velocity on each cell. The computed time step has to satisfy the CFL
-       * number chosen in the input parameter file on each cell of the mesh.
-       * If specified in the parameter file, the time step will be the minimum
-       * of the convection *and* conduction time steps.
-       *
-       * This function is implemented in
-       * <code>source/simulator/helper_functions.cc</code>.
-       */
-      double compute_time_step () const;
 
       /**
        * Advance the current time by the given @p step_size and update the
@@ -1835,10 +1823,10 @@ namespace aspect
        */
 
       /**
-       * @name Variables related to simulation termination
+       * @name Variables related to simulation time stepping
        * @{
        */
-      TerminationCriteria::Manager<dim>                         termination_manager;
+      TimeStepping::Manager<dim>                                time_stepping_manager;
       /**
        * @}
        */
--- a/include/aspect/time_stepping/interface.h
+++ b/include/aspect/time_stepping/interface.h
@@ -1,0 +1,101 @@
+/*
+  Copyright (C) 2019 - 2020 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+ */
+
+
+#ifndef _aspect_time_stepping_interface_h
+#define _aspect_time_stepping_interface_h
+
+#include <aspect/global.h>
+#include <aspect/simulator_access.h>
+#include <aspect/termination_criteria/interface.h>
+
+namespace aspect
+{
+  using namespace dealii;
+
+  namespace TimeStepping
+  {
+
+    /**
+     * A class to handle computation of the next time step (as desired by the user) and
+     * checking if the simulation is finished.
+     */
+    template <int dim>
+    class Manager : public SimulatorAccess<dim>
+    {
+      public:
+        /**
+         * Override initialize_simulator() so that we can also initialize the contained
+         * termination_manager.
+         */
+        virtual void initialize_simulator (const Simulator<dim> &simulator_object) override;
+
+        /**
+         * Compute the size of the next time step potentially taking into account
+         * the current solution (convection time step, conduction time step), settings
+         * from parameters, and termination criteria (to hit the end time exactly).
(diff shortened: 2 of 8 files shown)


tjhei:
@jdannberg @gassmoeller FYI

tjhei:
Oh, and if you have different suggestions for naming, let me know.

source/simulator/core.cc
```diff
@@ -2099,14 +2101,18 @@ namespace aspect
...
-        const bool checkpoint_written = maybe_write_checkpoint(last_checkpoint_time, termination);
         if (checkpoint_written)
           last_checkpoint_time = std::time(nullptr);
 
-        // see if we want to terminate
-        if (termination.first)
+        if (should_terminate)
           break;
       }
     while (true);
```
gassmoeller, line 2118: Isnt this equivalent to
```
}
while (should_terminate == false);
```

source/time_handler.cc
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+
+
+  template <int dim>
+  bool
+  TimeHandler<dim>::
+  need_checkpoint_on_terminate()
+  {
+
```
gassmoeller, line 47: remove empty line

source/time_handler.cc
```diff
@@ -0,0 +1,121 @@
...
+  {
+    SimulatorAccess<dim>::initialize_simulator(simulator_object);
+    termination_manager.initialize_simulator(simulator_object);
+  }
+
+
+
+  template <int dim>
+  void
+  TimeHandler<dim>:: declare_parameters (ParameterHandler &prm)
```
gassmoeller, line 74: remove stray space

include/aspect/time_handler.h
```diff
@@ -0,0 +1,90 @@
...
+#include <aspect/global.h>
+#include <aspect/simulator_access.h>
+#include <aspect/termination_criteria/interface.h>
+
+namespace aspect
+{
+  using namespace dealii;
+
+  /**
+   * A class ...
```
gassmoeller, line 34: Missing documentation.

include/aspect/simulator.h
```diff
@@ -2014,6 +2014,7 @@ namespace aspect
       template <int dimension, int velocity_degree>
       friend class StokesMatrixFreeHandlerImplementation;
       friend struct Parameters<dim>;
+      friend class TimeHandler<dim>;
```
gassmoeller, line 2017: Eww, another friend class? ;-) Can we avoid this if you move the `compute_time_step` function into the `TimeHandler` as well? This seems to be the only thing needing the friend access, and it is the logical place for it anyway. It may need some changes inside the function, but with `SimulatorAccess` everything in `compute_time_step` should be possible.

include/aspect/time_handler.h
```diff
@@ -0,0 +1,90 @@
...
+   */
+  template <int dim>
+  class TimeHandler : public SimulatorAccess<dim>
+  {
+    public:
+
+
+      virtual void initialize_simulator (const Simulator<dim> &simulator_object) override;
```
gassmoeller, line 42: Add some documentation here on why you override.

include/aspect/time_handler.h
```diff
@@ -0,0 +1,90 @@
...
+      double compute_time_step();
+
+      /**
+       * If true, the simulator should perform a checkpoint before terminating.
+       */
+      bool need_checkpoint_on_terminate();
+
+      /**
+       * Check if the simulation is ready to terminate sucessfully.
+       **/
```
gassmoeller, line 58: double *

include/aspect/time_handler.h
```diff
@@ -0,0 +1,90 @@
...
+
+    private:
+
+      /**
+       * Whether to do a final checkpoint before termination. This is
+       * specified in the parameters.
+       */
+      bool do_checkpoint_on_terminate;
+
+      TerminationCriteria::Manager<dim> termination_manager;
```
gassmoeller, line 85: Add some documentation

source/simulator/core.cc
```diff
@@ -2099,14 +2101,18 @@ namespace aspect
         // Check whether to terminate the simulation. The first part of the
         // pair indicates whether to terminate the execution; the second
         // indicates whether to do one more checkpoint:
-        const std::pair<bool,bool> termination = termination_manager.execute();
```
gassmoeller, line 2102: Update the comment above

tjhei:
> The one thing that is a bit unintuitive is that the TimeHandler owns the `TerminationCriteria::Manager`, but then again it makes sense that the object responsible for time advancement should own the information on when to stop.

Yes, I decided to just move it into the class because it made sense conceptually (and the simulator then has to deal with one fewer object). We can get rid of the Manager in the future and put the termination criteria into this class directly. This would break user plugins. Thoughts?

> I would prefer avoiding another friend class of Simulator though, maybe you can check if that is possible?

I was going to do that in a follow-up including splitting up ``compute_time_step()`` into smaller pieces that just compute convective/conductive  (so that this PR doesn't get too big). I will see if I can just move it for now.

source/simulator/core.cc
tjhei: Yes, but then I need to move should_terminate outside the large while loop. :-(

tjhei:
> I will see if I can just move it for now.

The last commit removes the friend and moves the implementation (without other refactoring or changes for now) into the new class. Is this a good start?

source/simulator/core.cc
gassmoeller: Oh, ok, leave it then.

include/aspect/time_handler.h
```diff
@@ -0,0 +1,96 @@
...
+       * termination_manager.
+       */
+      virtual void initialize_simulator (const Simulator<dim> &simulator_object) override;
+
+      /**
+       * Compute the size of the next time step potentially taking into account
+       * the current solution (convection time step, conduction time step), settings
+       * from parameters, and termination criteria (to hit the end time exactly).
+       */
+      double compute_time_step();
```
bangerth, line 51:
```suggestion
      double compute_time_step_size();
```

bangerth: But I believe that we actually spell it "timestep_size" elsewhere.

source/simulator/core.cc
bangerth: I've long wished that the C++ gods had made the trailing `while (...)` clause part of the enclosed scope :-(

source/simulator/helper_functions.cc
```diff
@@ -545,7 +545,7 @@ namespace aspect
 
     // Do a checkpoint if this is the end of simulation,
     // and the termination criteria say to checkpoint at the end.
-    if (termination_output.first && termination_output.second)
+    if (force_writing_checkpoint)
       write_checkpoint = true;
```
bangerth, line 549: I think this would be easier to read if you did
```
  bool write_checkpoint = force_writing_checkpoint;
  if (write_checkpoint == false && condition 1)
    ...
  if (write_checkpoint == false && condition 2)
    ...
```
This way you save a bit of time by not executing some checks, and conceptually you start with the "force" operation.

include/aspect/time_handler.h
```diff
@@ -0,0 +1,96 @@
...
+#include <aspect/global.h>
+#include <aspect/simulator_access.h>
+#include <aspect/termination_criteria/interface.h>
+
+namespace aspect
+{
+  using namespace dealii;
+
+  /**
+   * A class to handle computation of the next time step (as desired by the user).
```
bangerth, line 34: But the class actually does much more than just this.

tjhei:
> It's called `TimeHandler`, but it doesn't actually keep time

In light of the planned plugins, I am renaming it to ``TimeStepping::Manager`` now...

include/aspect/time_stepping/interface.h
```diff
@@ -0,0 +1,101 @@
...
+         * Compute the size of the next time step potentially taking into account
+         * the current solution (convection time step, conduction time step), settings
+         * from parameters, and termination criteria (to hit the end time exactly).
+         */
+        double compute_time_step_size();
+
+        /**
+         * If true, the simulator should perform a checkpoint before terminating.
+         */
+        bool need_checkpoint_on_terminate();
```
jdannberg, line 60: Should this be const?

include/aspect/time_stepping/interface.h
```diff
@@ -0,0 +1,101 @@
...
+
+        /**
+         * Check if the simulation is ready to terminate sucessfully.
+         */
+        bool should_simulation_terminate_now();
```
jdannberg, line 65: The name suggests that this function just answers the questions if the simulation should terminate, so could this be const too?

tjhei:
All good points. Fixed.

(I will rebase and squash after @bangerth is okay with the changes, separate commits are messy but might make review easier)

bangerth:
You'll have to rebase.

gassmoeller:
Ok, I think by now all maintainers have reviewed this and are basically ok with it. I am going to merge it :-)

PR #3693, opened 2020-08-05
fix temperature dependence of viscosity in latent heat models

jdannberg:
Currently, the latent heat material models use a constant reference temperature when computing the viscosity, even if the model uses adiabatic heating. This is inconsistent, for example with the computation of the density, which uses a constant reference temperature for models without adiabatic heating, and the adiabatic temperature for models with adiabatic heating. I believe that in almost every case, the user would actually want the latter type of behaviour for both the density and the viscosity (and in any case, both should be consistent), and I've changed the material models accordingly. 

This will probably break some tests (so I am waiting to see which ones, and and if it doesn't break tests, I will add a new one). 

I also still need to add an entry in [doc/modules/changes](../blob/master/doc/modules/changes) if we agree this is the behaviour we want. 

In a way, this type of rheology is also equivalent to the Frank-Kamenetskii rheology; so once we have a pull request for that, I want to make sure that it is consistent with the latent heat models, so that I can make a follow-up pull request that uses the new Frank-Kamenetskii rheology module (which then shouldn't change any functionality, just the implementation). 

### For all pull requests:

* [x ] I have followed the [instructions for indenting my code](../blob/master/CONTRIBUTING.md#making-aspect-better).


jdannberg:
Alright, I added a test and a changelog entry. Could you have another look?
I'll have a look at #3698.

tests/latent_heat_viscosity_adiabatic.prm
```diff
@@ -0,0 +1,114 @@
+# A modification of the latent heat viscosity test, but with adiabatic heating. 
+# It tests that the viscosity of th different layers is constant.
```
gassmoeller, line 2:
```suggestion
# It tests that the viscosity of the different layers is constant. 
```

doc/modules/changes/20200802_jdannberg
```diff
@@ -0,0 +1,6 @@
+<li> Fixed: The latent heat and latent heat melt material models now 
+use the adiabatic temperature as the reference temperature for the 
+viscosity if adiabatic heating is switched on (and hence, the 
+geotherm is an adiabat rather than a constant temperature). 
+<br>
+(Juliane Dannberg, 2020/08/05)
```
gassmoeller, line 6: Please make this data the same as the file name.

jdannberg:
Fixed!

PR #3699, opened 2020-08-05
Search for correct version of astyle, in case multiple versions exist

jperryhouts:
When I went to indent my code, I remembered why I had been using astyle directly, rather than the included indent script (this came up in the hackathon slack channel yesterday). I have the up-to-date version of astyle installed along side version 2.04, but `which astyle` refers to the more recent version.

This PR adds a function to search the user's `$PATH` for executables with "astyle" in their name. It checks each one until it either finds version 2.04, or runs out of paths to check.

### For all pull requests:


contrib/utilities/indent
```diff
@@ -21,6 +21,20 @@
...
+find_astyle() {
+  IFS=':' read -r -a PATHS <<< "$PATH"
+
+  astyle=""
+  while IFS= read -r -d '' exe; do
+    astyle="${exe}"
+    if test "$("${astyle}" --version 2>&1)" == "Artistic Style Version 2.04" ; then
+      break
+    fi
+  done < <(find "${PATHS[@]}" -iname '*astyle*' -print0 2>/dev/null)
```
bangerth, line 33: Interesting construct. I think it would be easier to read if you did
```
  find ... | (while ... ; do ... ; done)
```

bangerth: Or is that not the same? I think the style most people would use is
```
  ASTYLE_PATHS=`find ...`
  for i in $ASTYLE_PATHS ; do ... ; done
```

jperryhouts: There's a bunch of ways to do this. The first approach `(find... | while ...)` works, but if I remember correctly it has some weird scope thing about variables inside the while loop. I might be wrong about that. The second one `(ASTYLE_PATHS=...; for i in $ASTYLE_PATHS...)` is dangerous if there are spaces in any of the path names. That can be fixed with arrays, and careful use of quote marks I think.

On the one hand, only a maniac would put spaces in their executable binary file names. But on the other hand, if you have a bunch of non-standard PATH entries I could see it happening.

tl;dr: Whatever you think is most clear. Let me know what you prefer.

bangerth: Fair points. I'm just not (yet!) familiar with the `while ... done < <(find ...)` construct :-) I learn something new every day!

@tjhei @gassmoeller -- any thoughts?

contrib/utilities/indent
```diff
@@ -31,22 +45,26 @@ if test ! -f contrib/utilities/astyle.rc ; then
...
-if test -z "`which astyle`" ; then
+astyle="$(find_astyle)"
+
+if test -z "${astyle}"; then
   echo "*** No astyle program found."
   exit 1
 fi
 
-if test "`astyle --version 2>&1`" != "Artistic Style Version 2.04" ; then
+if test "$("${astyle}" --version 2>&1)" != "Artistic Style Version 2.04" ; then
```
tjhei, line 55:
```suggestion
if test "$(${astyle} --version 2>&1)" != "Artistic Style Version 2.04" ; then
```

contrib/utilities/indent
```diff
@@ -21,6 +21,20 @@
 ##
 ## ---------------------------------------------------------------------
 
+find_astyle() {
+  IFS=':' read -r -a PATHS <<< "$PATH"
```
tjhei, line 25: I am worried that this might not work on everybody machine (shell, OSX, etc.). Can we use this logic only if the default ``astyle`` doesn't work?

contrib/utilities/indent
```diff
@@ -31,22 +45,26 @@ if test ! -f contrib/utilities/astyle.rc ; then
...
 
+echo "--- Found astyle 2.04 at ${astyle}"
+
 
 # collect all header and source files and process them in batches of 50 files
 # with up to 10 in parallel
 echo "--- Indenting all ASPECT header and source files"
 
-find tests include source benchmarks cookbooks unit_tests \( -name '*.cc' -o -name '*.h' -o -name '*.h.in' \) -print | xargs -n 50 -P 10 astyle --options=contrib/utilities/astyle.rc
+find tests include source benchmarks cookbooks unit_tests \( -name '*.cc' -o -name '*.h' -o -name '*.h.in' \) -print | xargs -n 50 -P 10 "${astyle}" --options=contrib/utilities/astyle.rc
```
tjhei, line 67: do you need the quotes here?

contrib/utilities/indent
jperryhouts: I think the `#!/bin/bash` should make this compatible, but I added a fallback in https://github.com/geodynamics/aspect/pull/3699/commits/1fbe28dc90afdba449819299668b72347f87370f that should return the output of `which astyle` in case the other parts of this function fail.

contrib/utilities/indent
jperryhouts: Without the quotes this would fail if there are spaces in the executable's path.

contrib/utilities/indent
jperryhouts: Same as below. This would fail without the extra quotes if there happens to be spaces in `$astyle`.

jperryhouts:
I think this is good to go. Still open to changing the structure of that search function, if there are opinions one way or another. Let me know.

contrib/utilities/indent
gassmoeller: I have to say I do not understand the `while ... done < <` construct (and I would guess most readers of the file wouldnt, on the other hand there will not be many readers of such a file). I am just not very much into bash scripting, I know how to do simple loops, pipes and functions. Does this pipe the results of the `find` into the `done` command, which magically means it is available inside the loop? And why the two separate `<` marks? If I had written it I would go with `ASTYLE_PATHS=; for i in $ASTYLE_PATHS` and use an array or enclose each argument in quotations (dont know exactly how complicated that would be though, I know I have done it before long ago).

tjhei:
This is okay with me, but let's wait until after the hackathon. I also need to test this on MacOS.

contrib/utilities/indent
bangerth: @gassmoeller I believe that the `while ... done < <(...)` really just follows the `command < file` structure that pipes the contents of a file into a command. The only part that is not commonly used is (i) that the command is here the `while ... done` loop, that `<(...)` isn't a real file but uses the output of another file -- see https://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html#Process-Substitution . I had no idea that either works.

The problem with `ASTYLE_PATHS` in your example is if any of these paths contain spaces or, god forbid, `\0` characters which are apparently valid parts of file names. But I think that the `find ... | (while ... done)` syntax might be easier to read for sure, principally because it puts that which has to happen first (finding file names) *before* that which has to happen next (checking these file names).

jperryhouts: @gassmoeller @bangerth Problem with `find ... | while ...` is that for whatever reason it puts the loop block in its own background process, such that it cannot modify variables in the outer scope. Consider the following, for example:

    x='foo'
    find /bin -name 'bash' | while read exe; do x="$exe"; done
    echo "$x"
    >> foo

Apparently that can be avoided by first running `shopt -s lastpipe`, which instructs BASH to execute the loop contents within the current process, but I don't know whether that command is a standard part of BASH, or a linux specific thing. The process substitution version looks weird, but as Wolfgang noted, it's made of standard shell operations, and so should be as portable as possible.

Check out this stackoverflow thread for details: https://stackoverflow.com/a/23357277

On the other hand, I stand by my comment before that only a maniac would put whitespace characters in their PATH, so all of this seems pretty over the top anyway.

bangerth: Oh, then just do
```
  executable="`find ... | while ... echo $exe ... done`"
```

jperryhouts: Of course! I guess I got carried away. I'll update this after morning rounds.

jperryhouts: Someone in morning rounds mentioned that OSX apparently defaults to zsh now? The shebang in this file should force it to use bash if available, but while I was changing this, I added a commit (https://github.com/geodynamics/aspect/pull/3699/commits/5af4424b4f715944355c9ad31a8fad62f3676567) which makes this script also compatible with zsh (at least on linux). Tricky part was that zsh doesn't have an `export -f` equivalent, but if we're using the `find ... | while ...` construct here anyway, might as well use it for the dos2unix and tab replacement parts as well.

I can't believe I'm spending so much time generalizing something I could have fixed locally with a symlink! I'm gonna hold off on thinking about this until after the hackathon :)

tjhei:
I can confirm that this PR works on my old OSX laptop with bash.

contrib/utilities/indent
```diff
@@ -21,6 +21,21 @@
 ##
 ## ---------------------------------------------------------------------
 
+find_astyle() {
+  # Split $PATH variable into an array
+  setopt shwordsplit 2>/dev/null # zsh compatibility
```
bangerth, line 26: You don't actually need this -- the first line of the script determines that we run this file with bash, whether or not that is the default on a system or not.

contrib/utilities/indent
```diff
@@ -31,49 +46,49 @@ if test ! -f contrib/utilities/astyle.rc ; then
...
+astyle="$(find_astyle)"
+astyle="${astyle:-`which astyle`}" # Fallback to `which astyle`
+
+if test -z "${astyle}"; then
   echo "*** No astyle program found."
   exit 1
 fi
 
-if test "`astyle --version 2>&1`" != "Artistic Style Version 2.04" ; then
+if [[ "$("${astyle}" --version 2>&1)" != "Artistic Style Version 2.04" ]]; then
```
bangerth, line 57: is `[[` any different from `[`?

contrib/utilities/indent
```diff
@@ -31,49 +46,49 @@ if test ! -f contrib/utilities/astyle.rc ; then
...
-{
-    f=$1
-    tr -d '\015' <$f >$f.tmp
-    diff -q $f $f.tmp >/dev/null || mv $f.tmp $f
-    rm -f $f.tmp
-}
-export -f dos_to_unix
-find tests include source unit_tests cookbooks benchmarks \( -name '*.cc' -o -name '*.h' -o -name '*.prm' -o -name '*.h.in' \) -print | xargs -n 1 -P 10 -I {} bash -c 'dos_to_unix "$@"' _ {} 
+find tests include source unit_tests cookbooks benchmarks \( -name '*.cc' -o -name '*.h' -o -name '*.prm' -o -name '*.h.in' \) -print0 2>/dev/null \
+| while IFS=  read -r -d $'\0' src; do
```
bangerth, line 78: This change is unfortunate because it removes the `xargs -P 10` part and so treats file sequentially.

contrib/utilities/indent
```diff
@@ -31,49 +46,49 @@ if test ! -f contrib/utilities/astyle.rc ; then
...
-    # awkward tab replacement because of OSX sed, do not change unless you test it on OSX
-    TAB=$'\t'
-    sed -e "s/$TAB/  /g" $f >$f.tmp
-    diff -q $f $f.tmp >/dev/null || mv $f.tmp $f
-    rm -f $f.tmp
-}
-export -f tab_to_space
-find tests include source unit_tests cookbooks benchmarks \( -name '*.cc' -o -name '*.h' -o -name '*.prm' -o -name '*.h.in' -o -name 'CMakeLists.txt' -o -name '*.cmake' \) -print | xargs -n 1 -P 10 -I {} bash -c 'tab_to_space "$@"' _ {} 
+find tests include source unit_tests cookbooks benchmarks \( -name '*.cc' -o -name '*.h' -o -name '*.prm' -o -name '*.h.in' -o -name 'CMakeLists.txt' -o -name '*.cmake' \) -print0 2>/dev/null \
+| while IFS=  read -r -d $'\0' src; do
```
bangerth, line 86: same here

contrib/utilities/indent
jperryhouts: Updated loop contents to execute all in parallel. I'm not sure if limiting the number of parallel processes to 10 is important here. It doesn't seem to use a significant amount of memory on my computer, but parallelizing the whole loop did speed the script up from 20 seconds (fully serial) to 3 seconds (fully parallel).

contrib/utilities/indent
jperryhouts: True. There's no reason that macOS would stop shipping bash alongside zsh. If zsh isn't important, we can toss the last two commits in this PR, since they both make no difference in bash. Should I do that?

contrib/utilities/indent
jperryhouts: Yes, [here](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#index-_005b_005b) and [here](https://unix.stackexchange.com/a/255483) have some info on the difference. `[[` behaves more predictably than `[` in a lot of cases, but I think it's not POSIX compliant. I used it here because for some reason zsh doesn't like '==' as a comparison operator in an ordinary `[` environment. As above though, not necessary if there's no need to support zsh.

contrib/utilities/indent
bangerth: Yes. The script is run in bash, and we have no intention of changing that.

contrib/utilities/indent
bangerth: Let's stick with `test` then -- that's what's used everywhere else.

jperryhouts:
ok. I think all concerns are addressed. @tjhei I reverted some changes since you tested on your mac, so you may want to try that again just in case.

PR #3701, opened 2020-08-06
Improve performance dcpicard and newton solver

MFraters:
This pull request is a split off from some work done on pull request #3112. Aligning the iterated dc picard nonlinear tolerance with the stokes one turned out to be a worst case scenario for the dc picard method because because of the change the amount of non-linear iterations was determined by the compositional residual instead of the stokes residual, and the extra assembly time was in this case dominating the performance gains of the stokes solve. This pull request changes to code to speed up the assembly enough to be slightly faster in this particular case than the stokes system. The averaged time over at least 25 run of the iterated advection dc picard stokes solver for that case went down from 21.02 seconds to 15.21 seconds, compared to 16.95 seconds for the normal iterated advection and stokes solver.

This pull request consists of two parts. The first part has a very minimal but measurable performance gain, but I think the code is more clean. If others have a significant other opinion on it, I am fine with reverting it. The second part contains the most significant performance gain and is the most important part of this pull request. The second part gains performance by not rebuilding the preconditioner and part of the matrix when it is not needed. This add the same logic of rebuilding which is used by the default stokes solver.

### DIFF
--- a/source/simulator/assemblers/newton_stokes.cc
+++ b/source/simulator/assemblers/newton_stokes.cc
@@ -376,12 +376,12 @@ namespace aspect
                                                 eta * 2.0 * (scratch.grads_phi_u[i] * scratch.grads_phi_u[j])
                                                 // assemble \nabla p as -(p, div v):
                                                 - (pressure_scaling *
-                                                   (scratch.div_phi_u[i] * scratch.phi_p[j]))
+                                                   scratch.div_phi_u[i] * scratch.phi_p[j])
                                                 // assemble the term -div(u) as -(div u, q).
                                                 // Note the negative sign to make this
                                                 // operator adjoint to the grad p term:
                                                 - (pressure_scaling *
-                                                   (scratch.phi_p[i] * scratch.div_phi_u[j])))
+                                                   scratch.phi_p[i] * scratch.div_phi_u[j]))
                                               * JxW;
                   }
 
@@ -424,7 +424,7 @@ namespace aspect
                           {
                             data.local_matrix(i,j) += ( derivative_scaling_factor * alpha * (scratch.grads_phi_u[i] * deta_deps_times_eps_times_phi[j]
                                                                                              + scratch.grads_phi_u[j] * deta_deps_times_eps_times_phi[i])
-                                                        + derivative_scaling_factor * pressure_scaling * 2.0 * viscosity_derivative_wrt_pressure * (scratch.phi_p[j] * scratch.grads_phi_u[i]) * strain_rate )
+                                                        + derivative_scaling_factor * pressure_scaling * 2.0 * viscosity_derivative_wrt_pressure * scratch.phi_p[j] * scratch.grads_phi_u[i] * strain_rate )
                                                       * JxW;
 
                             Assert(dealii::numbers::is_finite(data.local_matrix(i,j)),
@@ -439,7 +439,7 @@ namespace aspect
                         for (unsigned int j=0; j<stokes_dofs_per_cell; ++j)
                           {
                             data.local_matrix(i,j) += ( derivative_scaling_factor * alpha * 2.0 * (scratch.grads_phi_u[i] * deta_deps_times_eps_times_phi[j])
-                                                        + derivative_scaling_factor * pressure_scaling * 2.0 * viscosity_derivative_wrt_pressure * (scratch.phi_p[j] * scratch.grads_phi_u[i]) * strain_rate )
+                                                        + derivative_scaling_factor * pressure_scaling * 2.0 * viscosity_derivative_wrt_pressure * scratch.phi_p[j] * scratch.grads_phi_u[i] * strain_rate )
                                                       * JxW;
 
                             Assert(dealii::numbers::is_finite(data.local_matrix(i,j)),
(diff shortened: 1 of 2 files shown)


source/simulator/assemblers/newton_stokes.cc
```diff
@@ -328,12 +335,12 @@ namespace aspect
                                     -(density * gravity * scratch.phi_u[i]))
                                    * JxW;
 
-              if (force != nullptr && this->get_parameters().enable_additional_stokes_rhs)
+              if (enable_additional_stokes_rhs && force != nullptr)
```
tjhei, line 338: Now you only need one or the other check but not both

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -268,11 +268,18 @@ namespace aspect
...
 
-      const MaterialModel::AdditionalMaterialOutputsStokesRHS<dim>
-      *force = scratch.material_model_outputs.template get_additional_output<MaterialModel::AdditionalMaterialOutputsStokesRHS<dim> >();
 
-      const MaterialModel::ElasticOutputs<dim>
-      *elastic_outputs = scratch.material_model_outputs.template get_additional_output<MaterialModel::ElasticOutputs<dim> >();
+      const bool enable_additional_stokes_rhs = this->get_parameters().enable_additional_stokes_rhs;
+
+      const MaterialModel::AdditionalMaterialOutputsStokesRHS<dim> *force = nullptr;
+      if (enable_additional_stokes_rhs == true)
```
tjhei, line 275:
```suggestion
      if (enable_additional_stokes_rhs)
```

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -268,11 +268,18 @@ namespace aspect
...
+        force = scratch.material_model_outputs.template get_additional_output<MaterialModel::AdditionalMaterialOutputsStokesRHS<dim> >();
+
+      const bool enable_elasticity = this->get_parameters().enable_elasticity;
+
+      const MaterialModel::ElasticOutputs<dim> *elastic_outputs = nullptr;
+      if (enable_elasticity == true)
```
tjhei, line 281: Not sure this is more readable, I would probably got for
```
 ... = (enable_elasticity) ? bla : nullptr;
```

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -367,12 +374,12 @@ namespace aspect
                                                 eta * 2.0 * (scratch.grads_phi_u[i] * scratch.grads_phi_u[j])
                                                 // assemble \nabla p as -(p, div v):
                                                 - (pressure_scaling *
-                                                   scratch.div_phi_u[i] * scratch.phi_p[j])
+                                                   (scratch.div_phi_u[i] * scratch.phi_p[j]))
```
tjhei, line 377: yes, very good!

source/simulator/solver_schemes.cc
```diff
@@ -495,9 +502,20 @@ namespace aspect
             pcout << "Solve failed and catched, try again with stabilisation" << std::endl;
             newton_handler->parameters.preconditioner_stabilization = Newton::Parameters::Stabilization::SPD;
             newton_handler->parameters.velocity_block_stabilization = Newton::Parameters::Stabilization::SPD;
-            rebuild_stokes_matrix = rebuild_stokes_preconditioner = assemble_newton_stokes_matrix = true;
+
+            // velocity boundary conditions, we need to re-assemble the system matrix
```
tjhei, line 506: Missing first line?

MFraters: ah, yes...

source/simulator/assemblers/newton_stokes.cc
MFraters: ok, will fix it. I wasn't sure any more what convention aspect uses :)

source/simulator/solver_schemes.cc
```diff
@@ -431,12 +431,19 @@ namespace aspect
...
-    // need to update the Stokes preconditioner.
-    rebuild_stokes_matrix = rebuild_stokes_preconditioner = assemble_newton_stokes_matrix = true;
+    // If the Stokes matrix depends on the solution, or we have active
+    // velocity boundary conditions, we need to re-assemble the system matrix
+    // (and preconditioner) every time. If we have active boundary conditions,
+    // they could a) depend on the solution, or b) be inhomogeneous. In both
+    // cases, just assembling the RHS will be incorrect.  If no active
+    // boundaries exist, we only have no-slip or free slip conditions, so we
+    // don't need to force assembly of the matrix.
+    if ((nonlinear_iteration == 0 && time_step == 0)
```
tjhei, line 441: Are you sure you need the first check here? I would think that the flags are true when the simulator starts? Also, isn't ``time_step`` the timestep size? Did you want to use ``time_step_number``?

MFraters: Yes, it should have been `time_step_number`. I had to add it to make it work, but I will test whether that is still the case.

MFraters:
The test failed with small differences. I tested in the last commit whether that was due to differences in rounding due to the brackets and it is (at least in some cases it becomes slightly worse). I am not sure what option is the best in this case. Opinions?

source/simulator/solver_schemes.cc
MFraters: It works without it, so I removed it.

bangerth:
Is this superseded by the other PRs? If yes, close this one. If not, mark it as WIP until the other ones are merged and this one can be rebased.

MFraters:
Yes, I will close it.

PR #3703, opened 2020-08-06
Output the stress as a tensor, rather than individual scalar components.

bangerth:
More specifically, use the deal.II machinery for tensors by deriving the stress viz postprocessor from `DataPostprocessorTensor`.

But it's not quite so easy:
* On the negative side, output files become a bit larger: They store the stress as tensors, rather than symmetric tensors, and that entails more components.
* Also on the negative side, neither Visit nor Paraview by default read what we write as tensors. One has to define tensor fields by composing them from the scalar fields that are currently output. That is no different than what one has to do right now.
* But, on the plus side, we're using the deal.II machinery that will, hopefully, allow us to one day write real tensors into VTU files. For some other output formats, that may already be possible. In other words, we will benefit from work done elsewhere.

If this patch finds favor, I will do the same for the strain rate and the shear stress.


tests/visualization_stress/solution/solution-00000.0000.gnuplot
```diff
@@ -1,455 +1,455 @@
...
+# Time =  21:36:16
 #
 # For a description of the GNUPLOT format see the GNUPLOT manual.
 #
-# <x> <y> <velocity> <velocity> <p> <T> <stress_xx> <stress_yy> <stress_xy> 
-0 0 0 0 9.99962 1 9.99962 9.99962 42 
-0.125 0 0 0 10 0 10 9.99997 41.9999 
+# <x> <y> <velocity> <velocity> <p> <T> <stress_xx> <stress_xy> <stress_yx> <stress_yy> 
+0 0 0 0 9.99962 1 9.99962 42 42 9.99962 
+0.125 0 0 0 10 0 10 41.9999 41.9999 9.99997
```
bangerth, line 9: The last 3 components of the old file were the xx, yy, and xy components. The last four components of the new file are the xx, xy, yx, yy components of the tensor.

tjhei:
I was also thinking about this when I saw your last PR touching that file. :-)

> * neither Visit nor Paraview by default read what we write as tensors

Then it doesn't make much sense, in my opinion. It looks like we don't write them as tensors in the .vtu files, but we could:
https://lorensen.github.io/VTKExamples/site/VTKFileFormats/#serial-xml-file-formats
(search for "active tensors")
It looks like we don't even write out vectors correctly but the tools seem to detect them automatically, see
https://github.com/dealii/dealii/blob/b0f67085059c9d4e5677f284d2cf6df03b485616/source/base/data_out_base.cc#L5455


> If this patch finds favor, I will do the same for the strain rate and the shear stress.

I would suggest we only do this if it has an actual advantage. I can look into tensor support in vtu at some point (but not this week).

bangerth:
I think you were just making the case that we *should* actually merge this, because deal.II could easily gain the functionality to make this patch useful :-)

bangerth:
It turns out that the problem isn't in the fact that deal.II doesn't support writing tensor data, but that there was a bug in `DataPostprocessor`: https://github.com/dealii/dealii/pull/10803

With the patch to deal.II applied, this PR really produces the expected tensor-valued output. So it now actually makes sense to merge it for sure :-)

bangerth:
See also dealii/dealii#10806

bangerth:
Now also converted the strain rate and shear stress viz postprocessors so that they output everything as tensors.

tjhei:
What is the status of this PR? While I think this goes in the right direction, what do we do when this is used with 9.2 that still has these bugs?

bangerth:
The patch is ready. With deal.II 9.2, the components of the tensor are just output as individual elements, just as they are right now. With 9.3, they will be output as tensors. in other words, it's no regression with 9.2 and a progression with 9.3. I see no reason not to merge.

tjhei:
> I see no reason not to merge.

Ah, I was thinking 9.2 would output this as a variable named ``strain_xx__strain_xy__strain_yx__strain_yy`` in vtu as seen in https://github.com/dealii/dealii/pull/10803. I guess this is only with your other patch applied.

tjhei:
Please rebase to master and fix the following tests:
```
The following tests FAILED:

	288 - gravity_function_time (Failed)

	775 - visualization_shear_stress (Failed)

	776 - visualization_shear_stress_pointwise (Failed)

	779 - visualization_strain_rate_tensor (Failed)

	780 - visualization_strain_rate_tensor_pointwise (Failed)
```

bangerth:
Right, the funny output only happens with https://github.com/dealii/dealii/pull/10803 and is again suppressed by https://github.com/dealii/dealii/pull/10806.

I've updated the patch here. There is still the issue that we get slightly different output with deal.II 9.3, but I suspect that this isn't the only place where 9.2 and 9.3 will produce different outputs.

bangerth:
OK, passes the tester now.

bangerth:
Changelog entry added.

gassmoeller:
I think you need to update the test results.

bangerth:
Works now!

bangerth:
Yay!

PR #3705, opened 2020-08-06
Newton Stokes failing for a linear problem

tjhei:
This currently failing test is a linear problem, so the second nonlinear
iteration should do 0 iterations (and does when you use "single
Advection, iterated Stokes") but it crashes with the Newton solver because the Stokes solver is not
converging. It looks like we try to reduce the residual to 1e-20, which
won't work.
The test passes if you remove "set Linear solver tolerance".

### DIFF
--- a/tests/sol_cx_newton_crash.cc
+++ b/tests/sol_cx_newton_crash.cc
@@ -1,0 +1,2 @@
+#include "../benchmarks/solcx/solcx.cc"
+
--- a/tests/sol_cx_newton_crash.prm
+++ b/tests/sol_cx_newton_crash.prm
@@ -1,0 +1,122 @@
+# A linear problem (copied from sol_cx_4) that crashes when "iterated Newton Stokes" is
+# used an the linear tolerance is reduced.
+
+set Dimension = 2
+
+
+set CFL number                             = 1.0
+
+set End time                               = 0
+
+
+set Resume computation                     = false
+
+set Start time                             = 0
+
+set Adiabatic surface temperature          = 0
+
+set Surface pressure                       = 0
+
+set Use years in output instead of seconds = false  # default: true
+
+set Nonlinear solver scheme                = single Advection, iterated Newton Stokes
+
+
+subsection Boundary temperature model
+  set List of model names = box
+
+end
+
+
+subsection Discretization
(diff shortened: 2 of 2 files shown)


tjhei:
@MFraters what do you think is happening here?

MFraters:
I have some idea's which can explain it, but I need to take a closer look at what the code exactly does. Will come back to this later today.

In general, you want to use a much looser tolerance for the Newton (or defect correction Stokes) solver then you would use for the full stokes solver.

MFraters:
So the relevant lines are in solver.cc around 781. I don't remember entirely why it is different, but I think it comes down to that a defection correction system is setup differently and therefore requires a different (less strict) tolerance. One important difference could be the initial guess used by the linear solver. The Newton solver simply sets the guess solution vector to zero, while the normal linear solver guess is a bit more complicated.

bangerth:
I think the criterion should be: If the nonlinear residual in iteration 2 is vastly smaller than in iteration 1 (e.g., because the problem is linear), then skip the solve.

MFraters:
> I think the criterion should be: If the nonlinear residual in iteration 2 is vastly smaller than in iteration 1 (e.g., because the problem is linear), then skip the solve.

I don't really understand what you are proposing here. The linear solver is crashing in nonlinear iteration 2, so we never know what the residual is from nonlinear iteration 2.

bangerth:
I was talking about the nonlinear residual, so the norm of the right hand side. If the norm of the rhs in iteration 2 is vastly smaller than the norm of the rhs in iteration 1, then the model is linear and we should not run the linear solver after assembly (or only run it with a very loose tolerance).

MFraters:
ah, I now understand. We would need to check that in the `assemble_and_solve_defect_correction_Stokes` function at about line 442 in solver_schemes.cc. Most of the information is already computed for nonlinear iteration > 1, so it would be a small change. I could make a seperate pull request but I am also fine with @tjhei changing it if he wants to.

bangerth:
Yes, right: The check needs to be right after `assemble_stokes_system()`. If the norm of the rhs is much smaller than it was in the previous iteration, then you can skip the solution of the linear systems, set the update to zero, and skip the line search stuff.

tjhei:
This has been merged as part of #3787

PR #3706, opened 2020-08-06
Set mesh displacement to zero

anne-glerum:
Originally, this was only triggered in case initial adaptive refinement is applied, not when only initial global refinement occurs.  From the documentation, I gather that some versions of LinearAlgebra::Vector::reinit set the vector entries to zero by default, but this particular version does not. Which also makes me wonder why this works for non adaptive initial meshes. 
 
With the new condition, the mesh_displacements are set to zero once, before solving for the mesh velocity on the coarse mesh. With `sim.time==0`, they would be set to zero before solving on the coarse mesh and again before solving on the initially adaptively refined mesh.

With just the free surface, it doesn't really matter when during the first timestep the displacements are set to zero, because the Stokes velocities are zero and thus the mesh velocities are zero on both the coarse and the adaptively refined mesh. But it would be nice if this does what we think it should do.

### For all pull requests:

### DIFF
--- a/source/mesh_deformation/interface.cc
+++ b/source/mesh_deformation/interface.cc
@@ -661,12 +661,10 @@ namespace aspect
       DoFTools::extract_locally_relevant_dofs (mesh_deformation_dof_handler,
                                                mesh_locally_relevant);
 
+      // This will initialize the mesh displacement and free surface 
+      // mesh velocity vectors with zero-valued entries.
       mesh_displacements.reinit(mesh_locally_owned, mesh_locally_relevant, sim.mpi_communicator);
       fs_mesh_velocity.reinit(mesh_locally_owned, mesh_locally_relevant, sim.mpi_communicator);
-
-      // if we are just starting, we need to initialize the mesh displacement vector.
-      if (!this->simulator_is_past_initialization())
-        mesh_displacements = 0.;
 
       // We would like to make sure that the mesh stays conforming upon
       // redistribution, so we construct mesh_vertex_constraints, which
--- a/source/mesh_deformation/interface.cc
+++ b/source/mesh_deformation/interface.cc
@@ -665,7 +665,7 @@ namespace aspect
       fs_mesh_velocity.reinit(mesh_locally_owned, mesh_locally_relevant, sim.mpi_communicator);
 
       // if we are just starting, we need to initialize the mesh displacement vector.
-      if (this->get_timestep_number() == 0)
+      if (!sim.simulator_is_past_initialization)
         mesh_displacements = 0.;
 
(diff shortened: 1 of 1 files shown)


source/mesh_deformation/interface.cc
```diff
@@ -665,7 +665,7 @@ namespace aspect
       fs_mesh_velocity.reinit(mesh_locally_owned, mesh_locally_relevant, sim.mpi_communicator);
 
       // if we are just starting, we need to initialize the mesh displacement vector.
-      if (this->get_timestep_number() == 0)
+      if (!this->simulator_is_past_initialization())
         mesh_displacements = 0.;
```
bangerth, line 669: What happens upon restart?

source/mesh_deformation/interface.cc
```diff
@@ -665,7 +665,7 @@ namespace aspect
       fs_mesh_velocity.reinit(mesh_locally_owned, mesh_locally_relevant, sim.mpi_communicator);
```
bangerth, line 665: Specifically, in line 664 (two lines up, but I can't annotate that line :-) ) `mesh_displacements` is set to zero. In other words, the call you are modifying should not actually have any discernible consequence. If you're wondering, you could just put
```
  Assert (mesh_displacement.l2_norm() == 0, ExcInternalError());`
```
here and verify that that is always satisfied. I believe that you can completely remove the two lines below.

anne-glerum:
> Separately: `reinit(size)` always sets the vector type to zero: it is not necessary to first resize and then explicitly set to zero. Only if you call `reinit(size,true)` does it not do that -- but that's a dangerous step. Can you find where the vector is resized and check what is being called?

So is it just the documentation that threw me off then? For this particular reinit, the `omit_zeroing_entries = false` is not mentioned (the 3/6 reinit here https://www.dealii.org/current/doxygen/deal.II/classLinearAlgebra_1_1distributed_1_1Vector.html#a48fec65453ceae3cf75f4e6f169ebcc1)

bangerth:
On 8/6/20 11:40 AM, anne-glerum wrote:
> So is it just the documentation that threw me off then? For this particular 
> reinit, the |omit_zeroing_entries = false| is not mentioned

Correct -- in which case the intention is that the vector really is zeroed 
out. I just double checked that indeed in the implementation of the reinit() 
function, the vector is zeroed out.

bangerth:
In other words, you can remove the explicit zeroing out -- give that a try and 
see what the tests say.

anne-glerum:
Yep, already done. There is just the Jenkins tester that fails, but the error does not look like something I'm familiar with.

gassmoeller:
This looks like a tester hickup, I will restart them.

PR #3707, opened 2020-08-06
PerpleX lookup now uses column labels to assign properties

bobmyhill:
The PerpleX lookup table previously assumed a certain ordering of columns, and of units. This PR instead reads the column labels, and assigns properties according to those labels.

This PR is a first step toward adding additional columns to the lookup, such as individual phase volumes.

### DIFF
--- a/source/material_model/utilities.cc
+++ b/source/material_model/utilities.cc
@@ -473,9 +473,29 @@ namespace aspect
           std::istringstream in(Utilities::read_and_distribute_file_content(filename, comm));
 
           // The following lines read in a PerpleX tab file in standard format
+          // The first 13 lines are a header in the format:
+          // |<perplex version>
+          // <table filename>
+          // <grid dim>
+          // <grid variable 1> (usually T(K) or P(bar))
+          // <min grid variable 1>
+          // <delta grid variable 1>
+          // <n steps grid variable 1>
+          // <grid variable 2> (usually T(K) or P(bar))
+          // <min grid variable 2>
+          // <delta grid variable 2>
+          // <n steps grid variable 2>
+          // Number of property columns in the table
+          // Column labels
+
           // First line is the Perplex version number
           std::getline(in, temp); // get next line, table file name
+
           std::getline(in, temp); // get next line, dimension of table
+          unsigned int n_variables;
+          in >> n_variables;
+          AssertThrow (n_variables==2, ExcMessage("The PerpleX file must be two dimensional (P(bar)-T(K))."));
+
           std::getline(in, temp); // get next line, either T(K) or P(bar)
 
           for (unsigned int i=0; i<2; i++)
@@ -483,7 +503,7 @@ namespace aspect
               std::string natural_variable;
               in >> natural_variable;
 
-              if (natural_variable[0] == 'T')
+              if (natural_variable == "T(K)")
                 {
                   std::getline(in, temp);
                   in >> min_temp;
@@ -493,7 +513,7 @@ namespace aspect
                   in >> n_temperature;
                   std::getline(in, temp); // get next line, either T(K), P(bar) or number of columns
                 }
-              else if (natural_variable[0] == 'P')
+              else if (natural_variable == "P(bar)")
                 {
                   std::getline(in, temp);
                   in >> min_press;
@@ -523,17 +543,17 @@ namespace aspect
             {
               std::string label;
               in >> label;
-              if (label.compare("rho,kg/m3") == 0)
+              if (label == "rho,kg/m3")
                 prp_indices[0] = n;
-              else if (label.compare("alpha,1/K") == 0)
+              else if (label == "alpha,1/K")
                 prp_indices[1] = n;
-              else if (label.compare("cp,J/K/kg") == 0)
+              else if (label == "cp,J/K/kg")
                 prp_indices[2] = n;
-              else if (label.compare("vp,km/s") == 0)
+              else if (label == "vp,km/s")
                 prp_indices[3] = n;
-              else if (label.compare("vs,km/s") == 0)
+              else if (label == "vs,km/s")
                 prp_indices[4] = n;
-              else if (label.compare("h,J/kg") == 0)
+              else if (label == "h,J/kg")
                 prp_indices[5] = n;
             }
           AssertThrow(std::all_of(prp_indices.begin(), prp_indices.end(), [](int i)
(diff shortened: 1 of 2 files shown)


source/material_model/utilities.cc
```diff
@@ -472,28 +472,78 @@ namespace aspect
...
+          std::getline(in, temp); // get next line, table file name
+          std::getline(in, temp); // get next line, dimension of table
+          std::getline(in, temp); // get next line, either T(K) or P(bar)
+
+          for (unsigned int i=0; i<2; i++)
+            {
+              std::string natural_variable;
+              in >> natural_variable;
+
+              if (natural_variable[0] == 'T')
```
bangerth, line 486: I think it would greatly help to understand the order of variables if you could copy-paste a typical file header here. This way, we would see which lines you're ignoring in the various `std::getline(in, temp);` below.

source/material_model/utilities.cc
```diff
@@ -472,28 +472,78 @@ namespace aspect
...
+                  std::getline(in, temp); // get next line, either T(K), P(bar) or number of columns
+                }
+              else if (natural_variable[0] == 'P')
+                {
+                  std::getline(in, temp);
+                  in >> min_press;
+                  min_press *= 1e5;  // conversion from [bar] to [Pa]
+                  std::getline(in, temp);
+                  in >> delta_press;
+                  delta_press *= 1e5; // conversion from [bar] to [Pa]
```
bangerth, line 503: Isn't that factor 1.014e5?

source/material_model/utilities.cc
```diff
@@ -472,28 +472,78 @@ namespace aspect
...
+          std::getline(in, temp); // get next line, either T(K) or P(bar)
+
+          for (unsigned int i=0; i<2; i++)
+            {
+              std::string natural_variable;
+              in >> natural_variable;
+
+              if (natural_variable[0] == 'T')
+                {
+                  std::getline(in, temp);
```
bangerth, line 488: For each of these lines you're ignoring here, are these comment lines? Or are you skipping the end of a line? In either case, if you know something about how what you're skipping should look like, an `AssertThrow` might make ignoring it safer.

source/material_model/utilities.cc
```diff
@@ -514,58 +564,39 @@ namespace aspect
...
-                  in.clear();
-                  h = enthalpy_values[(i-1)%n_temperature][(i-1)/n_temperature];
+                  in >> row_values[n];
+                  // PerpleX P-T grids sometimes contain rows filled with NaNs
+                  // in the corners of the grid.
+                  // These P-T regions are unimportant, and we don't want ASPECT
+                  // to crash just because it sees a NaN.
+                  // Here we choose to replace invalid doubles
+                  // with the most recent valid double.
+                  if (in.fail())
```
bangerth, line 582: Is that what you really get? In other words, when you have NaN in the file, do you get `in.fail() == true` or is `row_values[n]` simply assigned the `nan` value?

source/material_model/utilities.cc
```diff
@@ -514,58 +564,39 @@ namespace aspect
...
+                    {
+                      in.clear();
+                      row_values[n] = previous_row_values[n];
```
bangerth, line 585: You're basically padding out the values with valid values. Wouldn't it make more sense to also set these values as `nan` and hope that we never actually query them in a simulation? (Because if we ever did, we can't claim that the values we return with this scheme are actually "correct".)

source/material_model/utilities.cc
bangerth: Turns out that wikipedia says that 1 bar = 10^5 Pa indeed, and so slight less than one standard atmosphere (which is 1.014 bar).

bobmyhill: Some of the old low pressure papers report their pressures in mmHg... it's a fun world!!

source/material_model/utilities.cc
bobmyhill: Ok, I'll add a couple more assert throws where I can.

source/material_model/utilities.cc
bobmyhill: I'm a bit concerned that this might break existing prms. In addition to the corner issue, PerpleX has a habit of producing single P-T pixels of NaN (for a reason I genuinely don't understand). In these cases, selecting the previous row is a reasonable approach. Of course, it would be better to preprocess the tables so that they don't contain these dodgy NaNs, but at this point, the PerpleX lookup has been in ASPECT since 2015...

For now, I'll extend this comment to include the above, but if you'd like me to change the functionality, then I can.

source/material_model/utilities.cc
bobmyhill: In this case, in.fail() == true and row_values[n] is assigned a 0.

source/material_model/utilities.cc
bobmyhill: Done.

source/material_model/utilities.cc
bangerth: No, that sounds reasonable if you think you have good reasons. Let's go with what you have then.

Is this documented in the description of the plugin?

source/material_model/utilities.cc
```diff
@@ -472,28 +472,98 @@ namespace aspect
...
+          AssertThrow (n_variables==2, ExcMessage("The PerpleX file must be two dimensional (P(bar)-T(K))."));
+
+          std::getline(in, temp); // get next line, either T(K) or P(bar)
+
+          for (unsigned int i=0; i<2; i++)
+            {
+              std::string natural_variable;
+              in >> natural_variable;
+
+              if (natural_variable.compare("T(K)") == 0)
```
bangerth, line 506: You've been hanging out with other languages too much that treat strings differently :-)
```suggestion
              if (natural_variable -- "T(K)")
```

source/material_model/utilities.cc
```diff
@@ -472,28 +472,98 @@ namespace aspect
...
+                {
+                  std::getline(in, temp);
+                  in >> min_temp;
+                  std::getline(in, temp);
+                  in >> delta_temp;
+                  std::getline(in, temp);
+                  in >> n_temperature;
+                  std::getline(in, temp); // get next line, either T(K), P(bar) or number of columns
+                }
+              else if (natural_variable.compare("P(bar)") == 0)
```
bangerth, line 516: same here

source/material_model/utilities.cc
bobmyhill: Guilty :)

bobmyhill:
Ok, this has been partially squished. Should be ready to go.

PR #3711, opened 2020-08-06
add brackets to newton solver assembly.

MFraters:
splitting up pull request #3701 part 3. This add some brackets to the assembly of the newton solver. I think I placed the brackets in the wrong location before, which may have caused the problem. This might be the correct place.


source/simulator/assemblers/newton_stokes.cc
```diff
@@ -367,12 +367,12 @@ namespace aspect
                                                 eta * 2.0 * (scratch.grads_phi_u[i] * scratch.grads_phi_u[j])
                                                 // assemble \nabla p as -(p, div v):
                                                 - (pressure_scaling *
-                                                   scratch.div_phi_u[i] * scratch.phi_p[j])
+                                                   (scratch.div_phi_u[i] * scratch.phi_p[j]))
```
bangerth, line 370: It should make no difference -- both quantities are scalars, so at least from a mathematical perspective, this will have no effect (but is also not wrong).

source/simulator/assemblers/newton_stokes.cc
```diff
@@ -414,8 +414,8 @@ namespace aspect
                         for (unsigned int j=0; j<stokes_dofs_per_cell; ++j)
                           {
                             data.local_matrix(i,j) += ( derivative_scaling_factor * alpha * (scratch.grads_phi_u[i] * deta_deps_times_eps_times_phi[j]
-                                                                                             + scratch.grads_phi_u[j] * deta_deps_times_eps_times_phi[i])
-                                                        + derivative_scaling_factor * pressure_scaling * 2.0 * viscosity_derivative_wrt_pressure * scratch.phi_p[j] * scratch.grads_phi_u[i] * strain_rate )
+                                                                                             + (scratch.grads_phi_u[j] * deta_deps_times_eps_times_phi[i]))
+                                                        + derivative_scaling_factor * pressure_scaling * 2.0 * viscosity_derivative_wrt_pressure * scratch.phi_p[j] * (scratch.grads_phi_u[i] * strain_rate) )
```
bangerth, line 418: This, too, should make no difference because multiplication is distributive. But, it's quite possible that doing the dot product between matrices first,  and then multiplying with scalars (in the second of these lines) is substantially faster than multiplying a tensor repeatedly by scalars, and only then doing the dot product.

tjhei: ... but this can lead to floating point differences.

tjhei:
the output should be safe to update, but we need to double-check: https://jenkins.tjhei.info/blue/organizations/jenkins/aspect/detail/PR-3711/1/tests

source/simulator/assemblers/newton_stokes.cc
MFraters: ah, yes, you are right. I will remove it.

MFraters:
Yes, I agree that it should not matter and be safe, but I am a bit worried that in some of our test cases it actually adds several more iterations (see for example the first test in https://jenkins.tjhei.info/job/aspect/job/PR-3711/1/artifact/changes-test-results.diff), an in general the accuracy seems to be worse. So I am not sure whether we should continue with this...

source/simulator/assemblers/newton_stokes.cc
bangerth: I don't think there's harm in keeping it. Do as you think is easiest to read.

bangerth:
This only seems to happen once the relative nonlinear residual is already at the level of 3e-10 or smaller. Is it even reasonable to ask for it to be so low?

bangerth:
Also, why does it look like the linear solver tolerance is stuck at 0.1 in that test case?

MFraters:
> This only seems to happen once the relative nonlinear residual is already at the level of 3e-10 or smaller. Is it even reasonable to ask for it to be so low?

hmm, that is true. I might have pushed the tests a little bit too much. I could set the tolerance at 5-10 or 1e-9

bangerth:
Yes, maybe adjust the test in a separate PR to set something reasonable (might also be just 1e-8) and then rebase this PR once the other one is merged.

MFraters:
> Yes, maybe adjust the test in a separate PR to set something reasonable (might also be just 1e-8) and then rebase this PR once the other one is merged.

ok, will do that.

PR #3713, opened 2020-08-06
Adjoint equations

jaustermann:
This is a pull request for the implementation of the stokes adjoint equations. This is not meant to be merged into the developer branch but just serves to share the code to check the current implementation and think about how to reorganize this (i.e. what to keep, what to adjust, and what to move to a plugin). I've been talking with Wolfgang, Rene, and Menno about this. 

I rebased this to the current master and resolved conflicts. The code compiles but has a few warnings (e.g. some things are deprecated) that I did *not* fix. It also runs, however, there are some updates to the parameter file that I'm now working through to check. While I'm looking through this I think this doesn't affect the core of the implementation so wanted to share this as is. 

Happy to look through this together to think about how to implement / split this! Also if you want me to clean up parts before going through this further let me know.


gassmoeller:
Let us know when you want us to take another look.

jaustermann:
Yes, would be great to talk again about the structure. Items for discussion / restructuring:
(1) Currently, there is a function 'compute_parameter_updates' in assembly.cc This function calculates the sensitivity kernels and updates the compositional field so that in a next iteration the density and viscosity have different values that should yield a better fit to the data (minimize objective functional). Even if we don't want to iterate, this function is useful because it updates the compositional field (sensitivity kernel), which can then be plotted / exported without having to write a separate postprocessor for the kernels. On the downside, using that function to calculate the kernels requires using a specific material model & compositional field structure. This is also where a conjugate gradient method (or other gradient optimization scheme) would be linked in. So the question is - where should this live? Should this be implemented differently? Etc. 
(2) The parameter update iteration is currently done in core.cc and if we want to keep the adjoint solver as separate as possible that's probably not ideal(?). I think the iteration could also be done within the solver scheme and I forget now why I moved it to the core. Would be good to double-check the best place for this. 
(3) The rest of the added files are postprocessing and material model. Most of this is already separate, but some might need separating out.

source/simulator/core.cc
```diff
@@ -2035,52 +2050,57 @@ namespace aspect
     simulator_is_past_initialization = true;
     do
       {
-        // Only solve if we are not in pre-refinement, or we do not want to skip
-        // solving in pre-refinement.
-        if (! (parameters.skip_solvers_on_initial_refinement
-               && pre_refinement_step < parameters.initial_adaptive_refinement))
```
gassmoeller, line 2041: I think you accidentally deleted this while rebasing?

source/simulator/core.cc
```diff
@@ -2035,52 +2050,57 @@ namespace aspect
...
+
+        start_timestep ();
+
+        // since the default for num_it_adjoint i 0, this won't do anything if this isn't run in adjoint mode
+        for (unsigned int i=0; i<parameters.num_it_adjoint; ++i)
           {
-            start_timestep ();
+
+            if (parameters.nonlinear_solver == NonlinearSolver::Stokes_adjoint)
+              pcout << " ^^ Adjoint iteration number " << i  << std::endl;
```
gassmoeller, line 2061: Can this output go into your solver scheme?

source/simulator/parameters.cc
```diff
@@ -1266,6 +1267,32 @@ namespace aspect
     // declare the VolumeOfFluid parameters
     VolumeOfFluidHandler<dim>::declare_parameters(prm);
 
+
+    prm.enter_subsection("Adjoint problem");
+    {
+      prm.declare_entry ("Input filename for ajoint points", "/Users/jackyaustermann/Desktop/Aspect_code/aspect/data/adjoint-observations/dynamic_topography_observations.txt",
+                         Patterns::Anything(),
+                         "");
+      prm.declare_entry ("Read points in from file", "false",
```
gassmoeller, line 1276:
```suggestion
      prm.declare_entry ("Read points from file", "false",
```

source/simulator/parameters.cc
```diff
@@ -1266,6 +1267,32 @@ namespace aspect
...
+                         Patterns::Bool (),
+                         "");
```
gassmoeller, line 1278: All parameters are missing documentation.

benchmarks/adjointkernel_degree4.prm
```diff
@@ -0,0 +1,146 @@
+# A simple setup for convection in a quarter of a 2d shell. See the
```
gassmoeller, line 1: Please move this into its own directory

benchmarks/adjointkernel_degree4.prm
```diff
@@ -0,0 +1,146 @@
...
+set Dimension                              = 2
+set Use years in output instead of seconds = true
+set End time                               = 0 #1.5e9
+set Output directory                       = output_4
+set Pressure normalization                = surface
+set Surface pressure                      = 0
+set Nonlinear solver scheme = Stokes adjoint
+
+subsection Adjoint problem
+  set Input filename for ajoint points      = /Users/jackyaustermann/Sync/Aspect_code/aspect_laptop/data/adjoint_observations/plume_observations.txt
```
gassmoeller, line 14:
```suggestion
  set Input filename for ajoint points      = $ASPECT_SOURCE_DIR/data/adjoint_observations/plume_observations.txt
```

include/aspect/postprocess/adjoint_kernels.h
```diff
@@ -0,0 +1,54 @@
...
+#include <aspect/simulator_access.h>
+
+
+namespace aspect
+{
+  namespace Postprocess
+  {
+
+    /**
+     * A postprocessor that computes dynamic topography at the surface.
```
gassmoeller, line 35: update description

include/aspect/postprocess/visualization/adjoint_velocity.h
```diff
@@ -0,0 +1,62 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_postprocess_visualization_gravity_h
```
gassmoeller, line 22: adjoint_velocity

include/aspect/simulator.h
```diff
@@ -354,6 +354,11 @@ namespace aspect
         unsigned int polynomial_degree(const Introspection<dim> &introspection) const;
       };
 
+      virtual
+      std::list<std::string>
+      required_other_postprocessors() const;
+
```
gassmoeller, line 360: does this do something?

source/postprocess/visualization.cc
```diff
@@ -572,12 +656,18 @@ namespace aspect
...
 
       DataOut<dim> data_out;
       data_out.attach_dof_handler (this->get_dof_handler());
       data_out.add_data_vector (this->get_solution(),
                                 base_variables);
 
+      if (this->get_adjoint_problem() == true)
+        data_out.add_data_vector (this->get_current_adjoint_solution(),
+                                  adjoint_base_variables);
+
```
gassmoeller, line 670: Couldnt you always add this if the nonlinear solver scheme is `stokes_adjoint`? After every timestep (when this function is called) there should be a normal solution, and an adjoint solution to output, so lets just output everything in a single operation. Or am I missing something and not all information is still available at the end of the timestep (e.g. is something overwritten in the adjoint phase)?

source/simulator/assemblers/stokes.cc
```diff
@@ -372,8 +373,10 @@ namespace aspect
 
           for (unsigned int i=0; i<stokes_dofs_per_cell; ++i)
             {
-              data.local_rhs(i) += (density * gravity * scratch.phi_u[i])
-                                   * JxW;
+              // if we're solving the adjoint problem use the adjoint RHS instead of the normal body force
+              data.local_rhs(i) += (this->get_adjoint_problem() ? 0 :
+                                    (density * gravity * scratch.phi_u[i])
+                                    * JxW);
```
gassmoeller, line 379: Now that you have the adjoint assembler above is this still necessary?

source/simulator/core.cc
```diff
@@ -1851,6 +1857,9 @@ namespace aspect
   Simulator<dim>::
   solve_timestep ()
   {
+    // set adjoint problem to false as default
+    adjoint_problem = false;
+
```
gassmoeller, line 1862: not longer needed?

source/simulator/simulator_access.cc
```diff
@@ -750,6 +750,20 @@ namespace aspect
             ExcMessage("You can not call this function if the matrix-free Stokes solver is not used."));
     return *(simulator->stokes_matrix_free);
   }
+
+
+  template <int dim>
+  bool SimulatorAccess<dim>::get_adjoint_problem () const
+  {
+    return simulator->adjoint_problem;
+  }
```
gassmoeller, line 759: this is probably also no longer needed?

source/simulator/solver_schemes.cc
```diff
@@ -1069,6 +1070,124 @@ namespace aspect
...
+
+
+
+  template <int dim>
+  void Simulator<dim>::solve_stokes_adjoint ()
+  {
+    // TODO iterate
+    // todo make number of interations a parameter and change Do iterations for inversion parameter
+//    for (unsigned int i=0; i<parameters.num_it_adjoint; ++i)
+//      {
```
gassmoeller, line 1082: can you just use the `max_nonlinear_iterations` parameter that is already around? How would such a nonlinear scheme ook like? Would you just repeat the whole process (forward + adjoint) n times? Or do you need different number of iterations for the forward problem and the overall problem? I

source/simulator/solver_schemes.cc
```diff
@@ -1069,6 +1070,124 @@ namespace aspect
...
+    pcout << "   Adjoint problem ... " << std::endl;
+
+    // only do adjoint if final (refined) mesh is reached
+    // I think at this point this flag only affects postprocessing
+    adjoint_problem = true;
+
+    // clear all RHS assemblers and only do pressure rhs compatibility and assemble the Adjoint RHS;
+    assemblers->stokes_system.clear();
+    assemblers->stokes_system_on_boundary_face.clear();
+
```
gassmoeller, line 1126: You mentioned you need the pressure rhs modification? If so you can add it like this:
```
    // add the terms necessary to normalize the pressure
    if (do_pressure_rhs_compatibility_modification)
      assemblers->stokes_system.push_back(
        std_cxx14::make_unique<aspect::Assemblers::StokesPressureRHSCompatibilityModification<dim> >());

```
probably also needs SimulatorAccess like your assembler below

gassmoeller (commented):
I had a first look, and in the current form I think this fits nicely, without too many conflicts. I marked up places you will want to change, but I did not see anything that seemed impossible to do in our exiting structure. We could even make the adjoint scheme an option available for all solver schemes, and only call the second half of your solver scheme function after normal solver schemes (for that to work the adjoint solver has to take care to restore the assembler structure after it has finished, instead of waiting until the next forward solve to restore the normal assemblers). But I would suggest to first clean things up with the solver scheme, and then revisit the topic. Let me know if you want help with any of the changes.

gassmoeller:
To your questions: 
(1) yes it's current spot is not ideal. However, I would think that moving it to `helper_functions.cc`, naming it `compute_adjoint_parameter_update` and calling it from the solver scheme just like you do at the moment seems fine to me. I see that it depends on the `DynamicTopography` postprocessor, so somewhere it should asserted that this postprocessor is active (probably when reading in the nonlinear solver scheme). But I do not see where you require a specific material model for this function. You evaluate the material like normal with `in`, `out`, and then you generate an additional `in_adjoint`, but that is never passed into the material model, you seem to only use it to store some values (extract them from `adjoint_solution`), so this should actually work with any material model? Of course some other places may require your specific material model.
(2) Where is this precisely? I see some code in assembly.cc and solver_schemes.cc, but I do not see where the update is done in core.cc.
(3) Yes I think the rest is nicely separate. Is the material model useful without the solver scheme? If not, you can assert that the solver scheme is actually selected and throw an exception if not.

source/simulator/core.cc
```diff
@@ -898,6 +901,7 @@ namespace aspect
           case Parameters<dim>::NonlinearSolver::Kind::iterated_Advection_and_Newton_Stokes:
           case Parameters<dim>::NonlinearSolver::Kind::single_Advection_iterated_Newton_Stokes:
           case Parameters<dim>::NonlinearSolver::Kind::single_Advection_no_Stokes:
+          case Parameters<dim>::NonlinearSolver::Kind::Stokes_adjoint: // is that right?
```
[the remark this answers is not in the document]
jaustermann: I don't think so, but if put it in the group below I get the following runtime error:

An error occurred in line <1683> of file </var/folders/8z/hlb6vc015qjggytkxn84m6_c0000gn/T/heltai/spack-stage/spack-stage-dealii-9.2.0-yv7ehpigjxtbrhqi7y4menbu532z6rxi/spack-src/source/lac/trilinos_sparse_matrix.cc> in function
    void dealii::TrilinosWrappers::SparseMatrix::add(const dealii::TrilinosWrappers::SparseMatrix::size_type, const dealii::TrilinosWrappers::SparseMatrix::size_type, const dealii::TrilinosWrappers::SparseMatrix::size_type *, const dealii::TrilinosScalar *, const bool, const bool)
The violated condition was: 
    ierr == 0
Additional information: 
    An error with error number 2 occurred while calling a Trilinos function

source/postprocess/visualization.cc
jaustermann: Yes and no - I think this is a good suggestion and allows to further remove instances of where I call get_adjoint_problem. I think one thing I need to change for this though is that after the forward run I do postprocess() so that the dynamic topography is calculated. I don't actually need the other postprocessors at this point and it would be better to postprocess everything else after the adjoint equations are solved as well because then there's actually something here to plot. Can I run just the DT postprocessor ... or how do I make sure that that's accessible after the forward solution is done?

source/simulator/core.cc
jaustermann: This is still used in postprocessing, namely in global statistics and in visualization. I need to adjoint these files first before removing that.

source/simulator/solver_schemes.cc
jaustermann: Maybe I could use that parameter. The scheme would just repeat the whole process n times. The aspect I'm still trying to figure out how to do best is to save the input from several iterations. At the moment the iteration is done in core.cc (for (unsigned int i=0; i<=parameters.num_it_adjoint; ++i) and then the iterations are saved as different timesteps but this iteration should of course happen here. Do you have any pointers for where to look regarding saving iterative output?

jaustermann:
(1) Great, I'll move that to the helper_functions. And I think I sort of misspoke. It doesn't require a specific material model, but it does use two different compositional fields into which it writes the sensitivity kernels. In the material model these are then used to update the material properties. So yes, it doesn't need a specific material model, but it needs the compositional fields and if you want to use the kernels in an iteration it needs the material model. I'll set some descriptive exception for this. 
(2) It is in core.cc for convenience because it allowed me to save different iterative steps as different timesteps. I moved this into the solver scheme now and need to figure out how to best do the postprocessing. 
(3) Great! I'm not sure the material model is useful elsewhere so I'll put in exceptions that make sure the right setup is chosen. 

I resolved the comments that I addressed and will push an updated version with those. Next on my list is moving the compute_update_parameters to helper_functions, figure out how to best store the output, set exceptions to make sure the right material model - compositional field combo is chosen, and add some brief documentation.

bobmyhill (commented):
I wanted to understand this PR, and in reading through I found a few trivial things you might want to change :)

source/material_model/additive.cc
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+
+    template <int dim>
+    void
+    Additive<dim>::evaluate(const typename Interface<dim>::MaterialModelInputs &in,
+                            typename Interface<dim>::MaterialModelOutputs &out) const
+    {
+      const unsigned int density_idx = this->introspection().compositional_index_for_name("density_term");
+      const unsigned int viscosity_idx = this->introspection().compositional_index_for_name("viscosity_factor");
```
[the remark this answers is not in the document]
jaustermann: Good question - as it's implemented the new density is density+density_term while the new viscosity is viscosity*viscosity_factor. At least that's what I think I implemented. That's why they have a slightly different name.

include/aspect/material_model/additive.h
```diff
@@ -0,0 +1,108 @@
...
+#include <aspect/simulator_access.h>
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    using namespace dealii;
+
+    /**
+     * A material model that applies an average of the quadrature points in a cell  to
```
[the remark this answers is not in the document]
jaustermann: It doesn't - I've been bad about updating comments from files that I've copied from. I need to go through all of the new files to check and update that. Thanks for pointing it out!

source/material_model/additive.cc
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// explicit instantiations
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    ASPECT_REGISTER_MATERIAL_MODEL(Additive,
+                                   "additive",
+                                   " Explanation ")
```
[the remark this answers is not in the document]
jaustermann: Haha ... yes exactly.

jaustermann:
These are updates to address the comments above. One thing that is still missing is (better) documentation. Apart from that - any feedback is welcome, especially on how the new code fits within aspect. I actually would also be interested to see if it breaks any tests. We've talked about moving things into a separate plugin and I'm still happy to work on that if we think that would be best.

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,155 @@ namespace aspect
...
+        // -------------------------------------------------------------
+
+        // calculate the sensitivity kernels and use them to update the initial guess of density and viscosity
+        // through simple steepest descent. Note: this means the initial density and viscosity is overwritten
+        // for the visualization
+        compute_parameter_update();
+
+        // postprocess unless this is the last iteration because then it will be postprocessed outside of this solver scheme
+        if (i<max_nonlinear_iterations-1)
+          postprocess ();
```
gassmoeller, line 1602: What do you do if the solver converges before max_nonlinear_iterations? Then you will postprocess the last iteration twice. Could you instead use the same approach as the other solver schemes, or does this not work for adjoint? `parameters.run_postprocessors_on_nonlinear_iterations` disables the postprocessing after the nonlinear solver finishes, so every nonlinear iterations is only postprocessed once.

```suggestion
        if (parameters.run_postprocessors_on_nonlinear_iterations)
          postprocess ();
```

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+  Simulator<dim>::compute_parameter_update ()
+  {
+
+    // interior term
+    system_matrix = 0.;
+    system_rhs = 0.;
+
+    const unsigned int quadrature_degree = introspection.polynomial_degree.compositional_fields + 1;
+
+    // Gauss quadrature in the interior for best accuracy.
```
gassmoeller, line 2421: You dont use a different quadrature, and it is pretty standard in ASPECT to use QGauss, so no need for this comment.

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+    FEFaceValues<dim> fe_face_values (*mapping,
+                                      finite_element,
+                                      quadrature_formula_face,
+                                      update_values |
+                                      update_gradients |
+                                      update_quadrature_points |
+                                      update_JxW_values);
+
+    // Storage for shape function values for the current solution.
+    // Used for constructing the known side of the CBF system.
```
gassmoeller, line 2454: update comment

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+
+
+          // SURFACE TERM
+          // TODO: the solution at the surface does not agree with the benchmark.
+          // this could be because of the calculation here or the assembly of the RHS in assemblers/adjoint.cc
+
+          // see if the cell is at the *top* boundary and if so start surface term calculation
+          unsigned int top_face_idx = numbers::invalid_unsigned_int;
+          for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
+            if (cell->at_boundary(f) && geometry_model->depth (cell->face(f)->center()) < cell->face(f)->minimum_vertex_distance()/3)
```
gassmoeller, line 2549: Same comment about how to check for a surface boundary face.

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+
+    FEValues<dim> fe_values (*mapping,
+                             finite_element,
+                             quadrature_formula,
+                             update_values |
+                             update_gradients |
+                             update_quadrature_points |
+                             update_JxW_values);
+
+    FEValues<dim> fe_values_adjoint (*mapping,
```
gassmoeller, line 2437: You dont use `fe_values_adjoint`, but even if you did, it is set up exactly like `fe_values` so you may as well use that one instead. If you use `fe_values` with the adjoint solution vector it will work just like `fe_values_adjoint` (a single FEValues object can be used on multiple solution vectors, as long as the solution vectors have the same structure).

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+          MaterialModel::MaterialModelOutputs<dim> out(fe_values.n_quadrature_points, introspection.n_compositional_fields);
+          material_model->evaluate(in, out);
+
+          MaterialModel::MaterialModelInputs<dim> in_adjoint(fe_values, cell, introspection, current_adjoint_solution);
+
+          local_rhs = 0.;
+          local_mass_matrix = 0.;
+
+
+          // VOLUME TERM
```
gassmoeller, line 2498: Capital letters seem to shout at me while reading the source code ;-). Also remove at least one of the empty lines, it is not that important that a new block that starts here.

```suggestion
          // Assemble the volume term
```

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+
+              // Get parameters for the viscosity kernel
+              const double eta = out.viscosities[q];
+              const SymmetricTensor<2,dim> strain_rate_forward = in.strain_rate[q] - 1./3 * trace(in.strain_rate[q]) * unit_symmetric_tensor<dim>();
+              const SymmetricTensor<2,dim> strain_rate_adjoint = in_adjoint.strain_rate[q] - 1./3 * trace(in_adjoint.strain_rate[q]) * unit_symmetric_tensor<dim>();
+
+              for (unsigned int k=0; k<dofs_per_cell; ++k)
+                {
+                  phi_rho[k] = fe_values[introspection.extractors.compositional_fields[0]].value(k,q);
+                  phi_eta[k] = fe_values[introspection.extractors.compositional_fields[1]].value(k,q);
```
gassmoeller, line 2514: Here you also want to replace 0 and 1 by the correct compositional indices (compute and store them at the beginning of this function to avoid looking them up for every quadrature point and dof).

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+                }
+
+              for (unsigned int i = 0; i<dofs_per_cell; ++i)
+                {
+                  // Calculate the density kernel
+                  local_rhs(i) += -(gravity * velocity_adjoint) * phi_rho[i] * fe_values.JxW(q);
```
gassmoeller, line 2520: Precompute `fe_value.JxW(q)` somewhere around line 2510, it only depends on `q` not on `i` so precomputing it saves time (and you can reuse it in the line below and for the matrix).

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+                    {
+                      // Assemble Mass Matrix for the first compositional field. the mass
+                      // matrix for the second compositional field would look exactly the
+                      // same, so we can avoid building it for efficiency. (in fact we
+                      // also would have to deal with the fact that no sparsity pattern
+                      // and memory is allocated for the second compositional field; but
+                      // because we don't build the matrix, we don't care here.)
+                      local_mass_matrix(i,j) += fe_values[introspection.extractors.compositional_fields[0]].value(i,q) *
+                                                fe_values[introspection.extractors.compositional_fields[0]].value(j,q) *
+                                                fe_values.JxW(q);
```
gassmoeller, line 2536: Isnt this the same as:

```suggestion
                      local_mass_matrix(i,j) += phi_rho[i] *
                                                phi_rho[j] *
                                                JxW;
```

This would be much faster, because the three quantities would not need to be computed n_dof * n_dof times.

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+                    }
+                }
+            }
+
+
+          // SURFACE TERM
```
gassmoeller, line 2542:
```suggestion
          // Assemble the surface term
```

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+    for (; cell!=endc; ++cell)
+      if (cell->is_locally_owned())
+        {
+          fe_values.reinit (cell);
+          fe_values_adjoint.reinit (cell);
+
+          // Evaluate the material properties and the solution within the cell
+          MaterialModel::MaterialModelInputs<dim> in(fe_values, cell, introspection, solution);
+          MaterialModel::MaterialModelOutputs<dim> out(fe_values.n_quadrature_points, introspection.n_compositional_fields);
+          material_model->evaluate(in, out);
```
gassmoeller, line 2490: Here, and below for the face term you may want to include the material model averaging (your assembler includes it as well, because it gets a precomputed material model inputs object):
```
    MaterialModel::MaterialAveraging::average (parameters.material_averaging,
                                               cell,
                                               fe_values.get_quadrature(),
                                               fe_values.get_mapping(),
                                               out);
```

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+              }
+
+          if (top_face_idx != numbers::invalid_unsigned_int)
+            {
+              fe_face_values.reinit (cell,top_face_idx);
+
+              // Evaluate the material properties and the solution at the cell boundary
+              MaterialModel::MaterialModelInputs<dim> in_face(fe_face_values, cell, introspection, solution);
+              MaterialModel::MaterialModelOutputs<dim> out_face(fe_face_values.n_quadrature_points, introspection.n_compositional_fields);
+              material_model->evaluate(in_face, out_face);
```
gassmoeller, line 2562: average material model if necessary

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+
+              fe_face_values[introspection.extractors.temperature].get_function_values(topo_vector, topo_values);
+
+              for (unsigned int q=0; q<n_q_face_points; ++q)
+                {
+                  Point<dim> location = fe_face_values.quadrature_point(q);
```
gassmoeller, line 2568:
```suggestion
                  const Point<dim> location = fe_face_values.quadrature_point(q);
```

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+                  const double eta       = out_face.viscosities[q];
+
+                  const SymmetricTensor<2,dim> strain_rate = in_face.strain_rate[q] - 1./3 * trace(in_face.strain_rate[q]) * unit_symmetric_tensor<dim>();
+                  const Tensor<1,dim> gravity = gravity_model->gravity_vector(location);
+                  const Tensor<1,dim> n_hat = - gravity/gravity.norm();
+
+                  for (unsigned int k=0; k<dofs_per_cell; ++k)
+                    {
+                      phi_rho[k] = fe_values[introspection.extractors.compositional_fields[0]].value(k,q);
+                      phi_eta[k] = fe_values[introspection.extractors.compositional_fields[1]].value(k,q);
```
gassmoeller, line 2581: replace 0 and 1 by precomputed composition indices

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+                    }
+
+                  for (unsigned int i = 0; i<dofs_per_cell; ++i)
+                    {
+                      // calculate the density term
+                      local_rhs(i) += - topo_values[q] * topo_values[q] / (density - density_above) * phi_eta[i] * fe_face_values.JxW(q);
```
gassmoeller, line 2587: precompute JxW outside of this loop for here and use it here and in the line below

gassmoeller: What happens if `density == density_above`? Or does this never happen because density_above is either 0 (air) or 1000 (water)?

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,235 @@ namespace aspect
...
+                                                          system_rhs);
+
+        }   // loop over all cells
+
+    system_rhs.compress(VectorOperation::add);
+    system_matrix.compress(VectorOperation::add);
+
+    // solve linear system
+    const unsigned int rho_comp_block = AdvectionField::composition(0).block_index(introspection);
+    const unsigned int eta_comp_block = AdvectionField::composition(1).block_index(introspection);
```
gassmoeller, line 2616: replace 0 and 1 by indices

gassmoeller (commented):
Very nice, I think this integrates well into the general structure and we do not need to create another manager class. I am not sure I caught everything, but I left a number of comments. Most are straightforward, but let me know if you get stuck with some of the trickier ones.

Main point: It would be good if for the next review the documentation is updated that would help with the review process (and reviewing documentation is also important to make sure a non expert understands it). The easiest way to do this is for you to do a review of your own code, i.e. go through the pull request on the 'Files changed' tab (https://github.com/geodynamics/aspect/pull/3713/files) from top to bottom and immediately fix any comment you see that is outdated. Let us know when you want more feedback.

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,155 @@ namespace aspect
...
+
+
+  template <int dim>
+  void Simulator<dim>::solve_stokes_adjoint ()
+  {
+
+    // This solver scheme only works with a specific material model and number of compositional fiels. Check that
+    // both is chosen correctly. TODO add Assert for material model
+    Assert(introspection.n_compositional_fields == 2,
+           ExcMessage ("This solver scheme requires two compositional fields for the density and viscosity kernels."));
```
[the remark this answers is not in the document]
jaustermann: I'm checking this now at the beginning in the parameter file rather than within each individual file (e.g. the material model, the adjoint kernel postprocessor, and here). So I removed this here.

source/simulator/solver_schemes.cc
jaustermann: That works really well!

source/simulator/helper_functions.cc
jaustermann: I've added this but it doesn't compile and says it can't find a matching function for call to 'average'. Since this is a function in simulator I would have thought that this is already defined. Do have you an idea why it doesn't find this?

bangerth (commented):
Hi @jaustermann , I read everything down to `visualization.cc`, but it's late and I have to postpone the rest for tomorrow. I'm sorry that there will be a bit of work left. I think the design is all there and correct, it's really just small cosmetic stuff to work through.

More tomorrow!

include/aspect/simulator.h
```diff
@@ -2104,6 +2104,13 @@ namespace aspect
       template <int dimension, int velocity_degree>
       friend class StokesMatrixFreeHandlerImplementation;
       friend struct Parameters<dim>;
+
+      /** Parameters for the adjoint problem
+       */
+      LinearAlgebra::BlockVector                                current_adjoint_solution;
+      void solve_stokes_adjoint ();
+      void compute_parameter_update ();
```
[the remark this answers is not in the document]
jaustermann: I've renamed this to 'compute_sensitivity_kernel' because that's really what it does and then stores that in the compositional field. The material model then updates the parameters based on that. I've also improved the documentation here. Let me know if you think this name is okay or should be changed.

source/postprocess/adjoint_kernels.cc
```diff
@@ -0,0 +1,203 @@
...
+    AdjointKernels<dim>::execute (TableHandler &)
+    {
+      if (this->n_compositional_fields() == 0)
+        return std::pair<std::string,std::string>();
+
+      // create a quadrature formula based on the compositional element alone.
+      // be defensive about determining that a compositional field actually exists
+      AssertThrow (this->introspection().base_elements.compositional_fields
+                   != numbers::invalid_unsigned_int,
+                   ExcMessage("This postprocessor cannot be used without compositional fields."));
```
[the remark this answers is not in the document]
jaustermann: It seems like I repeated this assert statement in different files so thought it would be better to check in parameters.cc that if the adjoint stokes solver scheme was chosen that it needs at least 2 compositional fields and that two of them have to have the specific names. Then in this file and in the material model I simply check that they are only called if the solver scheme adjoint stokes is chosen and I also left a note that this then already checks for compositional fields. Let me know if that makes sense or should be changed.

source/postprocess/adjoint_kernels.cc
```diff
@@ -0,0 +1,203 @@
...
+            fe_values.reinit (cell);
+
+            double kernel_density = 0;
+            double kernel_viscosity = 0;
+            double volume = 0;
+            const Point<dim> midpoint_of_cell = cell->center();
+
+            fe_values[this->introspection().extractors.compositional_fields[0]].get_function_values (this->get_solution(),
+                compositional_values_density);
+            fe_values[this->introspection().extractors.compositional_fields[1]].get_function_values (this->get_solution(),
```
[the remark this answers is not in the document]
jaustermann: Yep, I understand that and already fixed it.

jaustermann:
Thanks so much for the comments Wolfgang! I've addressed the majority of Rene's comments but haven't pushed them yet (because I didn't quite get through all). I'll go through yours now. I'm not sure if it's better for you if I push my recent changes so that you see the most recent code of if I don't push them so that you can continue to work on the same code you started with. So ... let me know if you want me to push.

benchmarks/adjoint/adjointkernel_degree4.prm
```diff
@@ -0,0 +1,141 @@
...
+
+
+subsection Material model
+  # additive means after each iteration we add the value stored in the 
+  # composition field to the existing density / multiply the viscosity 
+  # with the new composition field
+  set Model name = additive
+
+  subsection Additive model
+     set Base model additive = simple
```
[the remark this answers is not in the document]
jaustermann: Yes I was wondering about this as well. I *think* the reason for choosing 'Base model additive' was that you can still combine this with another composite model and then you would have two base models. So for example you could choose the depth dependent viscosity material model and then the simple model as the base model for that. Maybe this is the same for other composite models though and we still always call it 'base model' even if it shows up multiple times. Let me know what the preference is to make it most consistent with other code.

benchmarks/adjoint/adjointkernel_degree4.prm
```diff
@@ -0,0 +1,141 @@
...
+
+  subsection Radial constant
+    set Magnitude = 10
+  end
+
+end
+
+
+subsection Mesh refinement
+  set Initial global refinement          = 5
```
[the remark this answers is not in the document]
jaustermann: Good point - the parameter file is in 2D so it does run quickly. However, the actual benchmark would be in 3D. Should I leave it in 2D and make a note of that or should I change it to 3D and add information on the computational cost?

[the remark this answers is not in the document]
jaustermann: What else should be saved in the benchmark folder? Should it include the output? I can have a look through the other folders.

include/aspect/parameters.h
```diff
@@ -702,7 +703,14 @@ namespace aspect
      * @}
      */
 
-
+    /**
+     * Adjoint problem
+     */
+    std::string        adjoint_input_file;
+    bool               read_in_points;
```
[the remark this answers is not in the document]
jaustermann: I ended up removing these for this first pull request because including specific data locations will need some more code adjustment. I thought maybe it's better to keep this first pull request to a minimum and then extend it later with more options.

source/postprocess/visualization.cc
```diff
@@ -127,6 +127,82 @@ namespace aspect
...
+
+          virtual
+          std::vector<DataComponentInterpretation::DataComponentInterpretation>
+          get_data_component_interpretation () const
+          {
+            std::vector<DataComponentInterpretation::DataComponentInterpretation>
+            interpretation (dim,
+                            DataComponentInterpretation::component_is_part_of_vector);
+            if (this->include_melt_transport())
```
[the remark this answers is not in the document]
jaustermann: Are you suggesting that the lines 162:167 should also stay in? Following Rene's suggestion, I now check in the parameter file whether the calculation includes melt while also using the adjoint solver and if so to throw an exception. I'm not sure if it would break things if you include melt, but it might?

source/postprocess/visualization.cc
```diff
@@ -127,6 +127,82 @@ namespace aspect
...
+                      (this->include_melt_transport()
+                       && this->introspection().variable("fluid velocity").component_mask[i]))
+                    computed_quantities[q][i]=input_data.solution_values[q][i] * velocity_scaling_factor;
+                  else
+                    computed_quantities[q][i]=input_data.solution_values[q][i];
+                }
+          }
+
+          virtual std::vector<std::string> get_names () const
+          {
```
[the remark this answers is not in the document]
jaustermann: I implemented this and it does work, but it then also adds two additional variables to the visual output, which are the adjoint compositional fields. Those don't really have any meaning and are just 0. If I keep the code here the same but pull from BaseVariablePostprocessor in the other functions I of course get an inconsistent number of variables (4 from the get_names and 6 from the others). Can you think of a fix for this?

source/simulator/helper_functions.cc
jaustermann: For any physically reasonable problem, the two densities should never be the same. Should I add an assert that checks that? We currently use the same equations in the DT postprocessor and it doesn't check that there either.

bangerth:
No need to review something that's not up to date :-) Let me know once you're through addressing all of @gassmoeller 's comments and I will take a look at the rest!

benchmarks/adjoint/adjointkernel_degree4.prm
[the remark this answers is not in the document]
bangerth: I think it's ok if there are parameters with the same name but in different sections. That makes it easy to remember how they are called and avoids having to look up the exact name of a parameter. So I think I'm still in favor of "Base model" :-)

benchmarks/adjoint/adjointkernel_degree4.prm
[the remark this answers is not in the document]
bangerth: Ah, I missed that that was in 2d. It's fine to leave it as is, and say somewhere in the text of the manual that the .prm runs in 2d, and then show 3d results where you make it clear that one has to change the dimension in the .prm file to reproduce those. That's better than having a super expensive computation in the .prm file because people will try to run it and it will lock up their machine :-)

In the folder, only store what is necessary to run the benchmark, none of the output files other than what you need to show pictures and tables in the manual.

source/postprocess/visualization.cc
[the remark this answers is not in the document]
bangerth: I think a general principle is that if you extend code like you do, and there is an easy way to support a complex scenario in one place, then you do that even though maybe you don't (and assert) in another place. Here, you can just import the data component interpretation flags from the primal solution like I showed for the names. I would do that, and that will work with or without melt transport. That doesn't mean that you have to support melt transport in all other places you change, and you should add asserts in those places that melt transport isn't enabled. But here, it's easy and it is one place less that someone would have to change later on when adding melt to adjoints.

Now, I don't think anyone will ever come around and combine melt and adjoints into anything that is actually useful. So the question is academic for this particular case, but I thought I'd explain my thinking because you're going to run into the same kind of question in other places and contexts as well.

source/postprocess/visualization.cc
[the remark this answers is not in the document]
bangerth: Hm, I see. The adjoint solution vector you store has these two additional compositional vector components, but what you are trying to do here is to just not output them at all. (In fact, I think you're dropping *all* compositional fields, right?) I had not noticed that. I guess that your previous approach makes sense in that case since you're just not using the adjoint compositional fields at all and never will because you're only solving the adjoint *Stokes* equations.

So go with your original approach, then. Sorry for the churn :-(

benchmarks/adjoint/adjointkernel_degree4.prm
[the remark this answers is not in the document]
jaustermann: That makes sense to me! Done.

source/simulator/helper_functions.cc
jaustermann: This currently doesn't work (see TODO in assembly). I've added a note about this in the code.

jaustermann:
Wolfgang and Rene, I've pushed new commits that I hope address your comments so far. This is ready to be looked at again. I want to test a couple of things but then I'll also do my own review as Rene suggested. One item that I still want to add is a test. The other item that I'm not sure about is the benchmark file. Currently, the benchmark works within the model domain but not at the surface. There's still something wrong that needs fixing. Do I keep the parameter file in the benchmark folder for now but wait with further documentation in the manual until this fully works? Or should I remove the parameter file until it works?

gassmoeller (commented):
Only part way through, but I need to leave it for now. Getting closer.

bangerth:
On 7/13/21 2:03 PM, Jacky Austermann wrote:
> The other item that I'm not sure about is the benchmark file. Currently, the 
> benchmark works within the model domain but not at the surface. There's still 
> something wrong that needs fixing. Do I keep the parameter file in the 
> benchmark folder for now but wait with further documentation in the manual 
> until this fully works? Or should I remove the parameter file until it works?

My preference is to only have things in the repository for which we know that 
they actually work.

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,149 @@ namespace aspect
     nonlinear_solver_control.check(1,0.0);
     signals.post_nonlinear_solver(nonlinear_solver_control);
   }
+
+
+
+  template <int dim>
+  void Simulator<dim>::solve_stokes_adjoint ()
+  {
+    // TODO: set an assert here or elsewhere that the material model additive is needed if the max_nonlinear_iteration is more than 1
```
bangerth, line 1464:
```suggestion
    // TODO: set an assert here or elsewhere that the material 'additive' material model is
    // needed if the max_nonlinear_iteration is more than 1
```

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,149 @@ namespace aspect
...
+    // What is the best way of getting the material model name?
```
bangerth, line 1465: That's typically done by doing something along the lines of 
```
  Assert (dynamic_cast<const Additive<dim>*>(&get_material_model()) != nullptr,
```
that is, you don't ask for the symbolic name but just see whether you can cast the material model object to the desired type.

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,149 @@ namespace aspect
...
+
+    // This calculation requires that the DT postprocessor is selected
+    AssertThrow(postprocess_manager.template has_matching_postprocessor<const Postprocess::DynamicTopography<dim>> (),
+                ExcMessage("The dynamic topography postprocessor has to be active for the current adjoint computations."));
+
+    // This calculation only works for instantaneous calculations
+    // Should there be an assert that checks for that? I don't see one in the other 'no advection' solvers.
```
bangerth, line 1472: always. feel free to add one to other solvers :-)

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,149 @@ namespace aspect
...
+
+    // determine how often to iterate over the forward & adjoint stokes equations
```
bangerth, line 1474:
```suggestion
    // determine how often to iterate over the forward & adjoint Stokes equations
```

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,149 @@ namespace aspect
...
+    nonlinear_iteration = 0;
+    do
+      {
+        pcout << "      Number of current adjoint iteration: " << nonlinear_iteration+1  << std::endl;
+        pcout << "      Solving the forward problem ... " << std::endl;
+
+        // set the assemblers for the RHS
+        set_assemblers();
+
+        // build the stokes system
```
bangerth, line 1501:
```suggestion
        // build the Stokes system
```

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,149 @@ namespace aspect
...
+        current_linearization_point.block(introspection.block_indices.velocities)
+          = solution.block(introspection.block_indices.velocities);
+        if (introspection.block_indices.velocities != introspection.block_indices.pressure)
+          current_linearization_point.block(introspection.block_indices.pressure)
+            = solution.block(introspection.block_indices.pressure);
+
+        // calculate the dynamic topography based on the forward solution,
+        // which will be needed for the new RHS assembly of the adjoint system
+        Postprocess::DynamicTopography<dim> &dynamic_topography = const_cast<Postprocess::DynamicTopography<dim> &> (
+                                                                    postprocess_manager.template get_matching_postprocessor<Postprocess::DynamicTopography<dim> >());
```
bangerth, line 1521: Just break before the `=` sign. Same a few lines down.

source/simulator/solver_schemes.cc
```diff
@@ -1452,6 +1455,149 @@ namespace aspect
...
+
+        dynamic_topography.execute(statistics);
+
+        // save the global statics output of the forward solution
+        Postprocess::GlobalStatistics<dim> &global_statistics = const_cast<Postprocess::GlobalStatistics<dim> &> (
+                                                                  postprocess_manager.template get_matching_postprocessor<Postprocess::GlobalStatistics<dim> >());
+
+        global_statistics.execute(statistics);
```
bangerth, line 1529: I don't understand the order of things here. You first save the state and then compute statistics. What do you actually do with the saved statistics?

bangerth (commented):
I've read through the rest, and I have primarily one general comment that I'd like us to address:

The adjoint equations are driven by a right hand side function that is typically the derivative of the objective function, i.e., misfit. In `adjoint.cc` and in the place where you compute the adjoint kernel, you make a very specific choice of how this functional looks like, namely the difference between some reference dynamic topography values and what your model predicts, but that is a case that very likely not a lot of people will actually be interested in. The way one would typically address this is by introducing a plugin system whereby people can implement different objective functions, and these may also include other kinds of observations. Have you given this some thought? I'm not saying that that needs to happen for this particular patch, but the code that is currently in the PR is oddly specific to the very particular case you are considering.

jaustermann:
Thank you both for your comments! I’ll work through those today. 

Thanks Wolfgang for your question about making this code more versatile and flexible by allowing for different RHS plugins. At the moment it really only does the bare bones. As you noted at a different point, I’m not even reading in data yet to calculate a misfit but just calculate the sensitivity to the produced topography. This is what is used in the benchmark. I had additional code in there to read in data but then thought it might be best to start with the simplest implementation and then add to it in a next step (next pull request) rather than add at this stage. If you want me to add this now though let me know.

It would be great to talk about how this could be made flexible to add other observations as plugins. Currently, I was mostly thinking of adding this in the same execute function (adjoint.cc) and choosing which observation to include based on parameter input, but I see that that’s probably not the most flexible way. Since I would like to add geoid observations as well it would be good to chat about how this should be done best. I’m thinking of saving the actual implementation of the geoid observations for a separate pull request but setting up the right structure already now could be good. Note that changing the observation may also affect the kernel calculation (this is in helper_functions -> compute_sensitivity_kernel).

I also wanted to respond to your comment on the benchmark because that’s a more general point that the code doesn’t yet reproduce the benchmark at the surface. I totally agree that only code that fully works (to the best of our knowledge) should be in the repository. I do think that means for this pull request that we still have to figure out how to properly deal with the surface term, but this should only affect a very small part of the code. So maybe the goal for the hackathon can be to get everything else finalized but then wait till later once the surface term aspect is figured out to merge. That would work for me and my plan is to continue working on this this summer and fall until it's done (hopefully with help from you Wolfgang :).

source/postprocess/visualization.cc
```diff
@@ -635,6 +707,15 @@ namespace aspect
       })
       != postprocessors.end());
 
+      // If we want to visualize the solution of the adjoint problem
+      internal::BaseAdjointVariablePostprocessor<dim> adjoint_base_variables;
```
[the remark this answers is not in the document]
jaustermann: If I move it into the if condition I get the following runtime error: 

An error occurred in line <381> of file </var/folders/8z/hlb6vc015qjggytkxn84m6_c0000gn/T/heltai/spack-stage/spack-stage-dealii-9.3.0-zy7k3uwnakcqjvrajvacy5l4jrl7eaex/spack-src/include/deal.II/base/smartpointer.h> in function
    T *dealii::SmartPointer<const dealii::DataPostprocessor<2>, void>::operator->() const [T = const dealii::DataPostprocessor<2>, P = void]
The violated condition was: 
    pointed_to_object_is_alive
Additional information: 
    The object pointed to is not valid anymore.

source/simulator/assemblers/adjoint.cc
```diff
@@ -0,0 +1,121 @@
...
+              const Tensor<1,dim> n_hat = scratch.face_finite_element_values.normal_vector(q);
+              const Tensor<1,dim> gravity = this->get_gravity_model().gravity_vector (scratch.face_finite_element_values.quadrature_point(q));
+              const double density = scratch.material_model_outputs.densities[q];
+              const double eta = scratch.material_model_outputs.viscosities[q];
+              const double JxW = scratch.face_finite_element_values.JxW(q);
+
+              for (unsigned int i=0; i<stokes_dofs_per_cell; ++i)
+                {
+                  // The new RHS is the derivative of the objective functional with respect to pressure and velocity
+                  data.local_rhs(i) += topo_values[q] * (2.0*eta *(n_hat * (scratch.grads_phi_u[i] * n_hat))
```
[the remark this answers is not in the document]
jaustermann: Yes that's right. Currently, it calculates the sensitivity to the current topography or in other words it assumes the observed topography is zero. There are benchmarks for this scenario and it's sort of the easiest case. I would suggest leaving this as is for this first pull request but you're right that including observations is the logical next step (and I have done this already on a different branch to test it out).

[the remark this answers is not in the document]
jaustermann: Yes that's right. Currently, it calculates the sensitivity to the current topography or in other words it assumes the observed topography is zero. There are benchmarks for this scenario and it's sort of the easiest case. I would suggest leaving this as is for this first pull request but you're right that including observations is the logical next step (and I have done this already on a different branch to test it out).

source/simulator/core.cc
```diff
@@ -468,10 +468,13 @@ namespace aspect
...
     // compressible or compatible (in the case of melt transport), and
     // there is no open boundary to balance the pressure.
+    // The same correction has to be applied for the adjoint equation
     do_pressure_rhs_compatibility_modification = ((material_model->is_compressible() && !parameters.include_melt_transport)
                                                   ||
                                                   (parameters.include_melt_transport && !material_model->is_compressible())
-                                                  || parameters.enable_prescribed_dilation)
+                                                  || parameters.enable_prescribed_dilation
+                                                  ||
+                                                  parameters.nonlinear_solver == NonlinearSolver::no_Advection_adjoint_Stokes)
```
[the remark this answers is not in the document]
jaustermann: Yes I think that's right. Is it wrong to do the modification for an incompressible case or is it just unnecessary?

source/simulator/core.cc
```diff
@@ -796,6 +799,7 @@ namespace aspect
           case Parameters<dim>::NonlinearSolver::Kind::iterated_Advection_and_Newton_Stokes:
           case Parameters<dim>::NonlinearSolver::Kind::single_Advection_iterated_Newton_Stokes:
           case Parameters<dim>::NonlinearSolver::Kind::single_Advection_no_Stokes:
+          case Parameters<dim>::NonlinearSolver::Kind::no_Advection_adjoint_Stokes:
             return true;
```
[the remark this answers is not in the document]
jaustermann: Agreed that this seems wrong, but if I don't set this to true I get a runtime error during the parameter_update:

An error occurred in line <1684> of file </var/folders/8z/hlb6vc015qjggytkxn84m6_c0000gn/T/heltai/spack-stage/spack-stage-dealii-9.3.0-zy7k3uwnakcqjvrajvacy5l4jrl7eaex/spack-src/source/lac/trilinos_sparse_matrix.cc> in function
    void dealii::TrilinosWrappers::SparseMatrix::add(const dealii::TrilinosWrappers::SparseMatrix::size_type, const dealii::TrilinosWrappers::SparseMatrix::size_type, const dealii::TrilinosWrappers::SparseMatrix::size_type *, const dealii::TrilinosScalar *, const bool, const bool)
The violated condition was: 
    ierr == 0
Additional information: 
    An error with error number 2 occurred while calling a Trilinos
    function

Looking through the core.cc file I noticed that composition coupling is only enabled if this is true and wondered if this is required in my parameter update. However, enabling these directly for when the adjoint_stokes solver scheme was selected didn't work. I'm not sure what the issue is here because I can't imagine that it's temperature-related. Any ideas?

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,260 @@ namespace aspect
...
+              const double eta = out.viscosities[q];
+              const SymmetricTensor<2,dim> strain_rate_forward = in.strain_rate[q] - 1./3 * trace(in.strain_rate[q]) * unit_symmetric_tensor<dim>();
+              const SymmetricTensor<2,dim> strain_rate_adjoint = in_adjoint.strain_rate[q] - 1./3 * trace(in_adjoint.strain_rate[q]) * unit_symmetric_tensor<dim>();
+
+              for (unsigned int k=0; k<dofs_per_cell; ++k)
+                {
+                  phi_rho[k] = fe_values[introspection.extractors.compositional_fields[density_idx]].value(k,q);
+                  phi_eta[k] = fe_values[introspection.extractors.compositional_fields[viscosity_idx]].value(k,q);
+                  // we will need the values of the 0th compositional field for the mass matrix assembly below.
+                  phi_0[k] = fe_values[introspection.extractors.compositional_fields[0]].value(k,q);
```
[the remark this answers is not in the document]
jaustermann: I believe I need the first compositional field to assemble the mass matrix in line 2537. I'm not exactly sure why but if it's not the first field it doesn't work. I'm also doing something similar in line 2641 with the zero_comp_block variable. Since I'm not sure why it needs to be the first field I'm not sure what a good symbolic name is.

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,260 @@ namespace aspect
...
+
+              // Evaluate the material properties and the solution at the cell boundary
+              MaterialModel::MaterialModelInputs<dim> in_face(fe_face_values, cell, introspection, solution);
+              MaterialModel::MaterialModelOutputs<dim> out_face(fe_face_values.n_quadrature_points, introspection.n_compositional_fields);
+              material_model->evaluate(in_face, out_face);
+
+              // TODO: Include model averaging. The assembler includes this as well because
+              // it gets a precomputed material model inputs object. However, this doesn't currently
+              // compile because the get_quadrature() call returns a dim-1 dimensional quadrature.
+              // This is also noted in assembly.cc
```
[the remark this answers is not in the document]
jaustermann: Do you mean I should set an assert that if people use an averaging scheme other than none that it stops here?

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,260 @@ namespace aspect
...
+    const unsigned int rho_comp_block = AdvectionField::composition(density_idx).block_index(introspection);
+    const unsigned int eta_comp_block = AdvectionField::composition(viscosity_idx).block_index(introspection);
+
+    // we will need the values of the 0th compositional field to solve the linear system below
+    const unsigned int zero_comp_block = AdvectionField::composition(0).block_index(introspection);
+
+    LinearAlgebra::BlockVector delta (system_rhs);
+    SolverControl control(1000, 1e-6);
+    SolverCG<LinearAlgebra::Vector> solver_cg (control);
+    solver_cg.solve (system_matrix.block(zero_comp_block, zero_comp_block),
```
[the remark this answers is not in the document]
jaustermann: I would agree but it does require the 0th compositional block. If I use rho_comp_block and that's the 0th compositional field in my parameter file then it works. But if I use viscosity as my 0th compositional but keep the rho_comp_block in this line then it gives me a runtime error:

An error occurred in line <421> of file </Applications/deal.II.app/Contents/Resources/Libraries/include/deal.II/lac/solver_cg.h> in function
    void dealii::SolverCG<dealii::TrilinosWrappers::MPI::Vector>::solve(const MatrixType &, VectorType &, const VectorType &, const PreconditionerType &) [VectorType = dealii::TrilinosWrappers::MPI::Vector, MatrixType = dealii::TrilinosWrappers::SparseMatrix, PreconditionerType = dealii::PreconditionIdentity]
The violated condition was: 
    std::abs(alpha) != 0.
Additional information: 
    A piece of code is attempting a division by zero. This is likely going
    to lead to results that make no sense.

As above, I'm not fully sure why it needs the 0th compositional field (block).

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,260 @@ namespace aspect
       }
     return new_linear_stokes_solver_tolerance;
   }
+
+
+  template <int dim>
+  void
+  Simulator<dim>::compute_sensitivity_kernel ()
```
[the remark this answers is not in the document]
jaustermann: I'm happy to rename it to what you suggest, but I'm pretty sure based on how sensitivity kernels are defined in the geodynamic literature (including the benchmark) that that's exactly what is calculated here. I actually think calling it the density..._update is less accurate because the update doesn't technically happen here, it happens in the material model, but this is of course directly linked to the compositional fields that are calculated here, so I'm fine with calling it adjoint_update_density_and_viscosity.

bangerth:
@jaustermann and I talked this through earlier today and the plan is to merge the PR in essentially the form it is right now (subject to addressing the comments other than the ones that pertain to the future generalization), given that the PR is big enough as it is now. We'd then come back to the correctness and extensibility question in future PRs.

source/simulator/core.cc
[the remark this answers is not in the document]
bangerth: I would very much like it if the forward solve happens *exactly* like it does if there were no dual problem to be solved at a later time. You might have to search for where this variable is used, and do something like
```
  if (do_pressure_rhs_compatibility_modification || (currently doing adjoint solve))
```

source/simulator/core.cc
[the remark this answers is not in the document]
bangerth: This looks like the error you'd get if you're trying to write into the matrix and the sparsity pattern has not allocated an entry for the position you're trying to write into. Is this function used in determining what the sparsity pattern of the matrix should be? If so, maybe do
```
  if (solver_scheme_solves_advection_equations() || (solve == adjoint solver))
```

source/simulator/solver_schemes.cc
bangerth: Yes. Things that are not supported should just result in an `abort()`, which is what a failed assertion triggers.

source/simulator/helper_functions.cc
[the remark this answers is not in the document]
bangerth: What happens when you pick a different compositional field? Is it the case that we simply didn't set up the sparsity pattern for this block, as discussed in a different comment? That would be the place to fix it.

source/simulator/helper_functions.cc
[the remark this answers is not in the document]
bangerth: Maybe `adjoint_update_compositional_fields()` then? Or `adjoint_update_fields_with_gradient()` [of the objective function]? You do an update at the end of the function, after all :-)

source/simulator/helper_functions.cc
```diff
@@ -2403,6 +2404,260 @@ namespace aspect
...
+                }
+            }
+
+
+          // Assemble the surface term
+
+          // Note that the solution at the surface does currently not agree with the benchmark.
+          // This could be because of an error here or in the assembly of the RHS in assemblers/adjoint.cc
+          // Given that the volume kernel already spikes at the surface, the issue is likely not (just) the surface
+          // term that is assembled here.
```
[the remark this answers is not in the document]
jaustermann: Haha I know, frowny face indeed. Currently, the issue definitely originates in the RHS assembly, because even if I don't add the surface term here, I get adjoint strainrates that are too high in the surface cell and hence a viscosity and density update that is too large in the surface cell. Of course, that doesn't exclude that the issue is here, because maybe the surface term - if calculated correctly - corrects for this. The surface term is currently very small and doesn't really affect the viscosity / density update.

source/simulator/solver_schemes.cc
jaustermann: I thought it could be useful to have the statistics for the forward run, but maybe that's not necessary (and more confusing than helpful). I'll remove this.

source/simulator/helper_functions.cc
jaustermann: This is now being checked in the DT postprocessor. Pull request #4241 by Cedric.

source/simulator/assemblers/adjoint.cc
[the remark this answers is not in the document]
jaustermann: I left a comment here and in the helper_function to indicate that it currently calculates the gradient with respect fo the squared topography (instead of a misfit).

source/simulator/helper_functions.cc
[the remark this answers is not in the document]
jaustermann: Thank you for the suggestion. I like adjoint_update_compositional_fields! I changed it to that.

source/simulator/core.cc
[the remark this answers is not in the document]
jaustermann: I'm not sure where the sparsity pattern of the matrix is determined. I can change things such that at every time it checks for solver_scheme_solves_advection_equations it also checks || (solve == adjoint solver) and that works. Is that what you were suggesting?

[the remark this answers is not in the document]
bangerth: Yes.

As for the sparsity pattern, let's let that one go. We've already made the decision to use the (comp0,comp0) block to solve for compositional fields. Though I would prefer if you took the correct block of the right hand side vector at least.

source/simulator/helper_functions.cc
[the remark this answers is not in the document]
jaustermann: If I change this to picking the 1st (instead of the 0th compositional field) I get the following run time error

An error occurred in line <1684> of file </var/folders/8z/hlb6vc015qjggytkxn84m6_c0000gn/T/heltai/spack-stage/spack-stage-dealii-9.3.0-zy7k3uwnakcqjvrajvacy5l4jrl7eaex/spack-src/source/lac/trilinos_sparse_matrix.cc> in function
    void dealii::TrilinosWrappers::SparseMatrix::add(const dealii::TrilinosWrappers::SparseMatrix::size_type, const dealii::TrilinosWrappers::SparseMatrix::size_type, const dealii::TrilinosWrappers::SparseMatrix::size_type *, const dealii::TrilinosScalar *, const bool, const bool)
The violated condition was: 
    ierr == 0
Additional information: 
    An error with error number 2 occurred while calling a Trilinos
    function

Yes maybe there is an error with the sparsity pattern for this block. Maybe we could talk about that?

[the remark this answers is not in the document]
bangerth: As mentioned in the other comment, let that go for now. Just use the (comp0,comp0) block of the matrix.

jaustermann:
Thank you so much Wolfgang and Rene for your continued help with this! I addressed all the comments and added a test. I also combined some commits (but not all). Lastly, I rebased to master. I expect my added test to fail and I plan to update it with the test output. However, it seems like other things are failing as well (the manual? and it also cannot be built?). Any thoughts about what's going wrong? It does compile and run correctly on my desktop.

jaustermann:
I noticed something weird in the iterations: The (forward) velocity field isn't changing from one iteration to the next even if I choose variable updates that lead to significantly different density or viscosity (and hence should lead to a different flow field). I checked that by only postprocessing the solution after the forward iteration - the density field changes but the velocity field doesn't. It's almost like it takes the initial condition of the compositional fields rather than the current compositional field. What is further weird though is that the # of stokes iterations (in the forward calculation) changes from one iteration to the next so it does seem to solve a different system. I'll continue to think through this but if anything comes to mind why the velocity wouldn't change even though the density does, let me know.

source/simulator/helper_functions.cc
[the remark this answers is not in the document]
gassmoeller: I am not sure if you have talked about this already, but here is the reason you can only use the 0th block at the moment:

- We want to reduce the amount of memory allocated for compositional matrix blocks, so we keep only a single one around (the 0th). All compositional fields have the same degree and boundary condition type (though could have different values), so the sparsity pattern will be the same.
- In order to use reasonable indices for the blocks (and maybe another reason linked to constraints I dont remember) we reinit the block we are going to solve using the sparsity pattern of block0 (see source/simulator/assembly.cc:1140-1147). After solving we clear the block again (solver_schemes.cc:250-252). You need to do the same before your assembly and after your solve. Doing that should simplify the code a bit and helps maybe finding the reason for the wrong results.

gassmoeller (commented):
I think I found your bug, and left some comments on how to circumvent the block0 workaround. I didnt go through all of the code, because I thought you want to try it first.

bobmyhill:
With the updated formulation in #7120, let's close this one.

PR #3716, opened 2020-08-06
Make global constants 'constexpr' variables.

bangerth:
This ought to make ASPECT so much faster, with all of the conversions between seconds and years and all :-)


tjhei:
Twice as fast for sure!

bangerth:
At least, man, at least!

PR #3721, opened 2020-08-06
slightly loosen the tolerance for the Newton tests.

MFraters:
A step needed for pull request #3711. I loosens the tolerance for the Newton solver test to be slightly more reasonable.


MFraters:
Btw, pull request #3711 will still change some results, but that will only be minor (if this solves the problem, I will test it before this gets merged).

MFraters:
So the reason I did set it that strict was that otherwise it isn't doing much work, and therefore not really testing. I think I can find a compromise by setting it a bit stricter.

MFraters:
Please do not merge until it is clear that the problem in #3711 is resolved.

MFraters:
Ok, the brackets pull request rebased on this one looks good. There are only minor changes in the testing result now. I think this can be merged.

bangerth:
Good. Let me know once you've got the other PR rebased!

PR #3405, opened 2020-03-09
Ability to interpolate multiple properties

sac-bsa:
*Describe what you did in this PR and why you did it.*
#3371 showed that the bilinear particle interpolator would fail an assertion if a particle was added during the computation since it would be called to interpolate all properties to the new particle. This PR aims to fix that. 
### For all pull requests:

* [X] I have followed the [instructions for indenting my code](../blob/master/CONTRIBUTING.md#making-aspect-better).

### For new features/models or changes of existing features:
* [X] I have tested my new feature locally to ensure it is correct.
* [X] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/master/tests/) directory.

PR #3722, opened 2020-08-06
Deleted redundant test for Bilinear Least Squares Particle Interpolat…

egpuckett:
…ion.

@gassmoeller I am deleting the redundant test for the Bilinear Least Squares Particle Interpolation algorithm that I mentioned in response to your comment about a redundant test in #3671.

### For all pull requests:


bangerth:
Is the test you're removing *identical* to an existing one, and if so which one?

If it tests the same general approach with a different test case, I think there is nothing wrong with leaving it in place and having some redundancy.

bangerth:
Ping?

sac-bsa:
I guess that we wouldn't be able to call the tests identical. We had made this with the idea that the particle_interpolator_bilinear_least_squares_solkz test and the particle_interpolator_bilinear_add_particles were very similar. The difference between the two is that the add_particles test additionally tests for the functionality added by #3405 (interpolating multiple properties in one call). Both tests otherwise are copies of the SolKz benchmark found in solkz_particles.prm. We were thinking with this, that since the add_particles test will also go through the same interpolation steps as the least_squares_solkz test that it could be removed without any reduction in coverage.

bangerth:
If one test checks a subset of another test, then it's of course possible to delete the first one. But we often keep it nonetheless, because that means that the first one is generally simpler and if something does break, it's often easier to start debugging with the simpler one. 

So my suggestion would be to just keep both around -- more tests is generally more better.

gassmoeller:
I think this may have been a misunderstanding. When I commented in #3671 that one of the tests is not necessary I meant one of the new tests in #3671. As Wolfgang mentioned it is useful to have one test for the bilinear least squares and another one for the bilinear least squares with particle addition during the model run (to test the particle addition algorithm), but it is not especially useful to add another two tests (with/without particle addition) for the quadratic least squares interpolator, because the particle addition is separated from the interpolator, and so if particle addition works for the bilinear case, it will most likely also work for the quadratic case. One new test in #3671 seems sufficient.

egpuckett:
OK I did misunderstand.  Shall I close the PR, or do you want to do that?

gassmoeller:
It does not really matter, I will go ahead and close it. I will take a look at #3671 once I have some time.

PR #3730, opened 2020-08-07
CMake: clean up unity file handling

tjhei:
follow-up to #3724

- comment things
- move world builder unity logic into the IF()
- clarify a couple of namespaces (not strictly necessary right now, but this should make things more robust)


tjhei:
ready for review.

CMakeLists.txt
```diff
@@ -144,9 +144,23 @@ if(ASPECT_WITH_WORLD_BUILDER)
   CONFIGURE_FILE("${WORLD_BUILDER_SOURCE_DIR}/source/config.cc.in" "${CMAKE_BINARY_DIR}/world_builder_config.cc" @ONLY)
   LIST(INSERT TARGET_SRC 0 "${CMAKE_BINARY_DIR}/world_builder_config.cc")
 
+  # Move some file to the end for unity builds to make sure other file come
```
MFraters, line 147: Just as a clarification for me, this text is just to explain what happens, e.g. the code has not changed in functionality, or has it? As far as I can see you only moved the foreach code up, and accounted for that the source file contains the full name. Is that correct?

tjhei: Correct.

MFraters:
Looks good to me. I does need a rebuild, but I am not sure if I can do that ( /rebuild )

PR #3732, opened 2020-08-07
fix postprocessing nonlinear iterations in Newton schemes

jdannberg:
This pull request fixes postprocessing nonlinear iterations with the Newton solver. Before, the postprocessing would be performed on the solution vector that still contained the update to the solution, rather than the solution itself. 

(If this breaks some of the Newton tests, I will probably have to copy the update back into the solution vector after postprocessing, so I wanted to see the tester output on this). 

### For all pull requests:


gassmoeller:
This makes sense and the testers seem ok with it.

PR #3733, opened 2020-08-07
Perplex lookup volume fractions

bobmyhill:
New functionality: The PerpleX lookup now allows the user to add columns to the lookup which correspond to the volume fractions of phases in the assemblages. These can then be read by material models. 

This PR also provides an example of the use of the new functionality via a modification to the Steinberger material model. The material model calculates the phase volume proportions and passes them to a NamedAdditionalMaterialOutputs object.

Finally, a new test is provided which checks the new functionality.

This PR builds on PR #3707 (now merged).


include/aspect/material_model/interface.h
```diff
@@ -903,6 +903,31 @@ namespace aspect
     };
 
 
+    /**
+     * Additional output fields for phase properties to be added to
+     * the MaterialModel::MaterialModelOutputs structure and filled in the
+     * MaterialModel::Interface::evaluate() function.
+     */
+    template <int dim>
+    class PhasePropertyOutputs : public NamedAdditionalMaterialOutputs<dim>
```
gassmoeller, line 912: Actually I think this is general enough to be an addition to the base class `NamedAdditionalMaterialOutputs` and should be an independent pull request. The only modification to the base class is:
- Add a constructor to the base class that takes `property_names` and `n_points` and is implemented exactly like you do here
- Add a ` std::vector<std::vector<double> > property_values;` member to the base class
- Add a `get_nth_output` implementation to the base class that does exactly what you do here, except it also contains an `Assert` that the `property_values.size` is actually set to something > 0 (to make sure it was properly initialized).

Then you can simply use an object of type `NamedAdditionalMaterialOutputs` instead of `PhasePropertyOutputs`. And in the future the other derived classes could use the new functionality as well (they can keep their names if they want, but use the functionality you implemented here). I think this simplifies the implementation.

source/material_model/steinberger.cc
```diff
@@ -383,6 +409,56 @@ namespace aspect
...
+      if (material_lookup.size() == 1)
+        {
+          for (unsigned int j = 0; j < unique_phase_names.size(); j++)
+            {
+              if (unique_phase_indices[0][j] != -1)
+                volume_fractions[j] += material_lookup[0]->phase_volume_fraction(unique_phase_indices[0][j],temperature,pressure);
+            }
+
+        }
+      else if (material_lookup.size() == compositional_fields.size() + 1)
```
jdannberg, line 431: Have you tested that this works with more than one compositional field, and different phase names for the different compositions, with just some overlap?

data/material-model/steinberger/pyr_with_volume_fractions.txt
```diff
@@ -0,0 +1,1613 @@
...
+P(bar)
+   1.0000000000000000
+   6410.2307692307695
+          40
+T(K)
+   300.00000000000000
+   56.410256410256409
+          40
+          30
+P(bar)         T(K)           rho,kg/m3      alpha,1/K      cp,J/K/kg      vp,km/s        vs,km/s        h,J/kg        	vol_Pl1 vol_Pl2 vol_O vol_Cpx1 vol_Cpx2 vol_ky vol_q vol_Opx vol_Gt1 vol_Gt2 vol_Sp vol_C2overc vol_Ring vol_Wad vol_st vol_Wus1 vol_Wus2 vol_Aki1 vol_Aki2 vol_capv vol_CF vol_Pv
```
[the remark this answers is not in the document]
bobmyhill: Actually, this isn't the raw PerpleX file; I had to run the output through a (bash) script to get everything in the correct format. I guess I should add the script to contrib? Or is there a better place for it?

The names are those that come out of PerpleX; capitalized phases signify that the phase is a solution of different endmembers, lower case are pure endmembers. I can change the columns by hand if you want?

source/material_model/interface.cc
```diff
@@ -914,6 +914,32 @@ namespace aspect
...
+
+
+    template<int dim>
+    std::vector<double>
+    PhasePropertyOutputs<dim>::get_nth_output(const unsigned int idx) const
+    {
+      // we have to extract the nth phase property output
+      // but the vector in the material model outputs is sorted so that the
+      // number of evaluation points (and not the number of phase properties)
+      // is the outer vector
```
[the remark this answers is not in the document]
bobmyhill: At the moment, the Steinberger model loops over each evaluation point, so it was easier to pass in.temperature[i], in.pressure[i], in.composition[i] ... to the new phase_volume_fraction function. 

But I could equally pass the function a reference to `in` and do both the inner and outer loops within the function. This will mean a few more lines of code in Steinberger, but interface.cc will be neater (and a bit faster).

include/aspect/material_model/interface.h
bobmyhill: Resolved by #3772

include/aspect/material_model/steinberger.h
```diff
@@ -305,6 +310,12 @@ namespace aspect
          */
         std::vector<std::unique_ptr<MaterialModel::MaterialUtilities::Lookup::PerplexReader> > material_lookup;
 
+        /**
+        * List of strings containing the names of the unique phases in all the material lookups
+        */
+        std::vector<std::string> unique_phase_names;
+        std::vector<std::vector<int>> unique_phase_indices;
```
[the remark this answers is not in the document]
bobmyhill: Now an unsigned int.
These have to be in the Steinberger model because unique_phase_indices is potentially a function of all of the compositional fields read into Steinberger.

include/aspect/material_model/interface.h
```diff
@@ -903,6 +903,31 @@ namespace aspect
...
+                              const std::vector<std::string> phase_names);
+
+        std::vector<double> get_nth_output(const unsigned int idx) const override;
+
+        /**
+         * Phase properties for all phases at the evaluation points
+         * that are passed to the instance of MaterialModel::Interface::evaluate()
+         * that fills the current object.
+         * phase_property_values[q][c] is the phase property at the evaluation point q
+         * for the compositional field with the index c.
```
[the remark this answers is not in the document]
bobmyhill: Now outdated, after removing PhasePropertyOutputs :)

source/material_model/interface.cc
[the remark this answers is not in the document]
bobmyhill: Now implemented the other way round :)

source/material_model/steinberger.cc
```diff
@@ -132,16 +132,42 @@ namespace aspect
     void
     Steinberger<dim>::initialize()
     {
+      std::set<std::string> set_phase_names;
```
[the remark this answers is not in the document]
bobmyhill: This is used to make a set of all of the phase names from all of the lookups. Some phase names will be repeated in all/some of the lookups, others will already be unique.

Making a unique_phase_names set and then making an array which maps the correct index in each lookup to the index in unique_phase_names saves on looking things up again and again.

I'll add some comments here making all of this more explicit.

source/material_model/steinberger.cc
```diff
@@ -132,16 +132,42 @@ namespace aspect
...
+        }
+
       lateral_viscosity_lookup
         = std_cxx14::make_unique<internal::LateralViscosityLookup>(data_directory+lateral_viscosity_file_name,
                                                                    this->get_mpi_communicator());
       radial_viscosity_lookup
         = std_cxx14::make_unique<internal::RadialViscosityLookup>(data_directory+radial_viscosity_file_name,
                                                                   this->get_mpi_communicator());
       avg_temp.resize(n_lateral_slices);
+
```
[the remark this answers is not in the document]
bobmyhill: Yes :)

include/aspect/material_model/utilities.h
```diff
@@ -156,6 +167,10 @@ namespace aspect
             dealii::Table<2,double> vs_values;
             dealii::Table<2,double> enthalpy_values;
 
+            std::vector<std::string> phase_names;
+            std::map<std::string, int> phase_name_index;
+            std::vector<dealii::Table<2,double>> phase_volume_fraction_values;
```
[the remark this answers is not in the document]
bobmyhill: Good idea.

source/material_model/steinberger.cc
bobmyhill: Now I have :)

include/aspect/material_model/utilities.h
```diff
@@ -156,6 +167,10 @@ namespace aspect
             dealii::Table<2,double> vs_values;
             dealii::Table<2,double> enthalpy_values;
 
+            std::vector<std::string> phase_names;
+            std::map<std::string, int> phase_name_index;
```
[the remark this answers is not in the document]
bobmyhill: Ooops, missed this one. I agree that we only need either a pair of vectors or a map, but having both is faster, I think. 
- Converting all the keys into a vector every time `phase_volume_column_names()` is called is a bit pointless.
- Running a `find_if` on `phase_column_names` to get the index of a string, to then use to get an int from another vector is also complicated when compared with using a map.

Am I missing a third way?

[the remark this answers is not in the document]
bobmyhill: I've changed the name of the map to phase_name_index_map anyway, whatever we end up choosing.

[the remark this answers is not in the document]
bobmyhill: I've removed the map and changed the logic of the phase column lookup in steinberger.cc.

include/aspect/material_model/utilities.h
```diff
@@ -118,6 +118,17 @@ namespace aspect
             dRhodp (const double temperature,
                     const double pressure) const;
 
+            std::vector<std::string>
+            phase_volume_phase_names() const;
```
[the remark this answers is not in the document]
bobmyhill: Done!

include/aspect/material_model/steinberger.h
```diff
@@ -183,6 +183,11 @@ namespace aspect
                                    const std::vector<double> &compositional_fields,
                                    const Point<dim> &position) const;
 
+        virtual std::vector<double> phase_volume_fractions (const double      temperature,
+                                                            const double      pressure,
+                                                            const std::vector<double> &compositional_fields,
+                                                            const Point<dim> &position) const;
```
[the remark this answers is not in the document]
bobmyhill: The updated NamedAdditionalMaterialOutputs now requires that we pass a reference to in, so this has now changed.

source/material_model/steinberger.cc
```diff
@@ -383,6 +409,56 @@ namespace aspect
...
+              volume_fractions[j] = background_volume_fraction;
+              for (unsigned int i = 0; i < compositional_fields.size(); ++i)
+                {
+                  if (unique_phase_indices[i+1][j] != -1)
+                    volume_fractions[j] += compositional_fields[i] * (material_lookup[i+1]->phase_volume_fraction(unique_phase_indices[i+1][j],temperature,pressure) - background_volume_fraction);
+                }
+            }
+        }
+      else
```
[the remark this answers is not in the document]
bobmyhill: Good catch.

source/material_model/utilities.cc
```diff
@@ -538,29 +552,76 @@ namespace aspect
...
+                                    phase_column_names.end(),
+                                    column_name) != phase_column_names.end())
+                        {
+                          AssertThrow(false,
+                                      ExcMessage("The PerpleX lookup file " + filename + " must have unique column names. "
+                                                 "Sometimes, the same phase is stable with >1 composition at the same "
+                                                 "pressure and temperature, so you may see several columns with the same name. "
+                                                 "Either combine columns with the same name, or change the names."));
+                        }
+                      else
```
jdannberg, line 609: You  don't have to put this into an else statement. If the assert triggers, we never get here anyway.

source/material_model/utilities.cc
```diff
@@ -538,29 +552,76 @@ namespace aspect
...
+                        {
+                          // Populate phase_column_names with the column name
+                          // and phase_column_indices with the column index in the current lookup file
+                          // This makes populating the phase_volume_fractions object easier.
```
jdannberg, line 613: Can you explain this a bit more? (Why? Easier than what?)

source/material_model/utilities.cc
```diff
@@ -618,13 +683,34 @@ namespace aspect
...
+                  vp_values[i%n_temperature][i/n_temperature]=row_values[prp_indices[3]];
+                  vs_values[i%n_temperature][i/n_temperature]=row_values[prp_indices[4]];
+                  enthalpy_values[i%n_temperature][i/n_temperature]=row_values[prp_indices[5]];
+
+                  for (unsigned int n=0; n<phase_volume_fractions.size(); n++)
+                    {
+                      phase_volume_fractions[n][i%n_temperature][i/n_temperature]=row_values[phase_column_indices[n]];
+                    }
+                }
+              else // first_natural_variable == "P(bar)"
```
jdannberg, line 700: Wait, was this here before? What does it do? Is that change related to the other changes in this pull request?

tests/steinberger_phase_fractions_background.prm
```diff
@@ -0,0 +1,131 @@
+# This is a copy of Steinberger compressible test,
+# but it uses a pyrolite material property table instead
+# of the test data, so that reasonable seismic velocities are
+# computed. The property table also contains volume fractions of the
+# stable phases in the rock, which can be output using
+# the "named additional outputs" in the list of named outputs for
+# the Visualization postprocessor.
```
jdannberg, line 7: Can you update the description to make it clear how this test is different from the one above?

include/aspect/material_model/utilities.h
```diff
@@ -118,6 +118,23 @@ namespace aspect
...
+             * Returns a vector of all the column names in the lookup file
+             * that start with the character string vol_fraction_
+             */
+            std::vector<std::string>
+            phase_volume_column_names() const;
+
+            /**
+             * Returns the volume fraction of the phase_idth phase
+             * at a given temperature and pressure, where the phase_column_name
+             * corresponds to that given in phase_volume_column_names()[phase_id].
```
jdannberg, line 131: Wait, but this doesn't really have anything to do with what phase_volume_column_names() does, phase_volume_column_names() could be changed, but this function would still return the volume fraction of the phase with the column index phase_id at a given temperature and pressure. I think the thing you're saying here really belongs into the documentation of std::vector<std::string> phase_column_names and std::vector<dealii::Table<2,double>> phase_volume_fractions (namely, that these belong together and are in the same order).

contrib/python/create_aspect_readable_perplex_table_with_phase_volumes.py
```diff
@@ -0,0 +1,180 @@
...
+
+After this has been done, call this python script from the command line.
+The following arguments are accepted by the file:
+--werami_path path [path ...]
+--project project name [project name ...]
+--outfile output file
+--n_pressures N_PRESSURES
+--n_temperatures N_TEMPERATURES
+[--pressure_range LOW_PRESSURE_LIMIT HIGH_PRESSURE_LIMIT]
+[--temperature_range LOW_TEMPERATURE_LIMIT HIGH_TEMPERATURE_LIMIT]
```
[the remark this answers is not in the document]
bobmyhill: This is what comes out of the python argument parser <shrugs>

source/material_model/steinberger.cc
```diff
@@ -383,6 +426,74 @@ namespace aspect
...
+      std::vector<std::vector<double>> volume_fractions(unique_phase_volume_column_names.size(),
+                                                        std::vector<double>(in.n_evaluation_points(), 0.));
+
+      if (material_lookup.size() == 1)
+        {
+          for (unsigned int i = 0; i < unique_phase_volume_column_names.size(); i++)
+            {
+              for (unsigned int j = 0; j < in.n_evaluation_points(); j++)
+                {
+                  if (unique_phase_volume_column_indices[0][i] != -1)
```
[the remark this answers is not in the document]
bobmyhill: I had the realisation this morning that the vector of vector<double> doesn't need to look like a regular table, and that therefore I could reverse the logic of the lookup, removing all of the -1s and speeding up the lookups...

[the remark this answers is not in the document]
bobmyhill: So I've now done that. Sorry for making you look through another iteration of this, but I think this is closer to what you would like and is cleaner.

data/material-model/steinberger/morb_G13_with_volume_fractions_lo_res.dat
```diff
@@ -0,0 +1,2529 @@
+|6.6.6
```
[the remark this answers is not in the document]
bobmyhill: Yes, and also for pyrolite. I've put the references and compositions into the prm files.

include/aspect/material_model/steinberger.h
```diff
@@ -305,6 +307,12 @@ namespace aspect
          */
         std::vector<std::unique_ptr<MaterialModel::MaterialUtilities::Lookup::PerplexReader> > material_lookup;
 
+        /**
+        * List of strings containing the names of the unique phases in all the material lookups
+        */
+        std::vector<std::string> unique_phase_volume_column_names;
+        std::vector<std::vector<int>> unique_phase_volume_column_indices;
```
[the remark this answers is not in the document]
bobmyhill: Done, and done.

include/aspect/material_model/steinberger.h
```diff
@@ -183,6 +183,8 @@ namespace aspect
                                    const std::vector<double> &compositional_fields,
                                    const Point<dim> &position) const;
 
+        virtual std::vector<std::vector<double>> phase_volume_fractions (const MaterialModel::MaterialModelInputs<dim> &in) const;
```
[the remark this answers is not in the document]
bobmyhill: Done.

source/material_model/utilities.cc
bobmyhill: :D

source/material_model/utilities.cc
bobmyhill: I've removed this line; it used to make sense when I had the map, but not any more!

source/material_model/utilities.cc
bobmyhill: No, it wasn't here before - this was the headache I found on Wednesday night.

The ordering of the first two columns in the PerpleX table files dictates whether the inner loop is over temperature or pressure. The first column is always the inner loop.

I've now added some documentation to describe this. The existing tests cover the possible range of possibilities.

tests/steinberger_phase_fractions_background.prm
bobmyhill: Done.

include/aspect/material_model/utilities.h
bobmyhill: Good point. I've removed this sentence and added documentation to phase_column_names and phase_volume_fractions.

bobmyhill:
Ok, I'm not sure why the new tests are failing. I'll investigate after morning rounds.

jdannberg:
Okay, let me know when you want me to look at this again!

source/material_model/steinberger.cc
```diff
@@ -489,14 +600,20 @@ namespace aspect
...
             {
               seismic_out->vp[i] = seismic_Vp(in.temperature[i], in.pressure[i], in.composition[i], in.position[i]);
               seismic_out->vs[i] = seismic_Vs(in.temperature[i], in.pressure[i], in.composition[i], in.position[i]);
             }
         }
 
+      // fill phase volume outputs if they exist
+      if (NamedAdditionalMaterialOutputs<dim> *phase_volume_fractions_out = out.template get_additional_output<NamedAdditionalMaterialOutputs<dim> >())
+        {
+          phase_volume_fractions_out->output_values = phase_volume_fractions(in);
```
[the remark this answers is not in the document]
jdannberg: What about this? Didn't work, or you just missed my comment?

source/material_model/steinberger.cc
```diff
@@ -132,9 +132,52 @@ namespace aspect
...
+          for (unsigned int j = 0; j < phase_volume_column_names.size(); j++)
+            {
+              std::vector<std::string>::iterator it = std::find(unique_phase_volume_column_names.begin(),
+                                                                unique_phase_volume_column_names.end(),
+                                                                phase_volume_column_names[j]);
+              unsigned int i_unique = std::distance(unique_phase_volume_column_names.begin(), it);
+              unique_phase_volume_column_indices[i][i_unique] = j;
+            }
+        }
+
```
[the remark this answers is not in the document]
bobmyhill: Github had this comment, and I didn't see it. Thank goodness I had the same idea :)

source/material_model/steinberger.cc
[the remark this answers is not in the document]
bobmyhill: Missed this one, but also my attempts over the past hour and a bit didn't work. Can we raise this as an issue for later?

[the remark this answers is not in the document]
bobmyhill: I should elaborate: I changed the function, gave it std::vector<std::vector<double>> &volume_fractions as a new argument, and passed phase_volume_fractions_out->output_values to the function.

The vector of vectors filled fine within the function, and I could retrieve the values within the if {} clause, but then maybe the values got lost and ASPECT seg-faulted during postprocessing.

source/material_model/steinberger.cc
```diff
@@ -383,6 +426,74 @@ namespace aspect
...
+            {
+              for (unsigned int j = 0; j < in.n_evaluation_points(); j++)
+                {
+                  if (unique_phase_volume_column_indices[0][i] != -1)
+                    volume_fractions[i][j] += material_lookup[0]->phase_volume_fraction(unique_phase_volume_column_indices[0][i],in.temperature[j],in.pressure[j]);
+                }
+            }
+        }
+      else if (material_lookup.size() == in.composition[0].size() + 1) // background field
+        {
```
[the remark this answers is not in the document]
jdannberg: I still think it would be easier to read without the brackets :grin:

source/material_model/steinberger.cc
```diff
@@ -383,6 +427,77 @@ namespace aspect
...
+          // We first look up the volume fractions corresponding to this background field
+          // (assuming the domain is filled 100% with this single composition)
+          std::vector<std::vector<double>> background_volume_fractions(unique_phase_names.size(),
+                                                                       std::vector<double>(in.n_evaluation_points(), 0.));
+          for (unsigned int j = 0; j < unique_phase_indices[0].size(); j++)
+            {
+              for (unsigned int k = 0; k < in.n_evaluation_points(); k++)
+                {
+                  background_volume_fractions[unique_phase_indices[0][j]][k] += material_lookup[0]->phase_volume_fraction(j,in.temperature[k],in.pressure[k]);
+                  volume_fractions[unique_phase_indices[0][j]][k] += background_volume_fractions[unique_phase_indices[0][j]][k];
```
jdannberg, line 464: I think you can avoid looping over j and k twice, if you loop from i=0 below and then just make this the spacial case for i=0.

source/material_model/steinberger.cc
```diff
@@ -132,9 +132,53 @@ namespace aspect
...
+            }
+        }
+
+      // We now fill the unique_phase_indices object
+      // with the unique_phase_names index for each of the phases
+      // stored in each material lookup.
+
+      // Initialise the column indices
+      unique_phase_indices.resize(material_file_names.size(),
+                                  std::vector<unsigned int>());
```
jdannberg, line 167: I would put that on one line. Also: you can resize this before you start your first loop in line 144, and then you only need one loop over all i.

source/material_model/steinberger.cc
```diff
@@ -132,9 +132,53 @@ namespace aspect
...
+
+      for (unsigned int i = 0; i < material_file_names.size(); i++)
+        {
+          std::vector<std::string> phase_volume_column_names = material_lookup[i]->phase_volume_column_names();
+          for (unsigned int j = 0; j < phase_volume_column_names.size(); j++)
```
jdannberg, line 172: Same here: you now loop over all j twice, but you can combine this with the loop over the strings above. (At least I am thinking this will make the code shorter, if that just makes it more complicated, it's not worth changing).

source/material_model/steinberger.cc
```diff
@@ -132,9 +132,53 @@ namespace aspect
...
+            {
+              std::vector<std::string>::iterator it = std::find(unique_phase_names.begin(),
+                                                                unique_phase_names.end(),
+                                                                phase_volume_column_names[j]);
```
jdannberg, line 176: The finding you also only need to do once

source/material_model/utilities.cc
jdannberg: I am just saying: this is independent of the phase stuff, right? Before, we just didn't allow having pressure as the inner loop? We could also just make sure temperature is always the first column. Is there a reason why one would want to have pressure first?

jdannberg: But I guess since it's already part of this pull request, and you've made a test for it, if there's a reason why someone would want to have pressure first, let's just leave this as it is.

source/material_model/steinberger.cc
```diff
@@ -383,6 +427,77 @@ namespace aspect
...
 
 
+    template <int dim>
+    std::vector<std::vector<double>>
+                                  Steinberger<dim>::
+                                  phase_volume_fractions (const MaterialModel::MaterialModelInputs<dim> &in) const
+    {
+      // In the following function,
+      // the index i corresponds to the ith compositional field
+      // the index j corresponds to the jth unique phase
```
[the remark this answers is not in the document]
bobmyhill: Right. Good catch.

source/material_model/steinberger.cc
bobmyhill: Ok!

source/material_model/steinberger.cc
bobmyhill: I don't think this is true. The second loop over i requires `unique_phase_names`, which is only complete at the very end of the first loop over i.

bobmyhill: The call to resize is now on one line.

source/material_model/steinberger.cc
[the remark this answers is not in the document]
bobmyhill: I missed your comment originally, and then had problems last night doing what you suggested. It works this morning though :)

source/material_model/steinberger.cc
bobmyhill: Same problem as above. The two loops need to be separated, because the second loop requires a completed `unique_phase_names` object from the first.

source/material_model/steinberger.cc
bobmyhill: Oh, wait, now I see what you mean. When this involved a set, I needed the two loops, but now I don't. Very good catch!

source/material_model/steinberger.cc
bobmyhill: Not a problem!

source/material_model/steinberger.cc
bobmyhill: Yup, now changed. Thanks!

bobmyhill:
So it turns out that trying to fix `things` late at night is not a productive use of time. Who knew?
I think I've addressed all your comments. The code looks a lot neater now, thank you!

source/material_model/utilities.cc
bobmyhill: Thanks; yes, I should have split this into a separate PR, but now that it's in there...

The reason why I think this is needed is that I genuinely don't understand how PerpleX chooses how to output the table. The ordering of P-first or T-first seems to have changed a few months ago, or maybe there's a flag that I don't know about.

Either way, if I don't know about how PerpleX makes these choices, I'm sure there are other users who would be similarly PerpleXed if their completely-reasonable-looking input file produced garbage. It also makes the code somewhat more future-proof.

include/aspect/material_model/steinberger.h
```diff
@@ -183,6 +183,14 @@ namespace aspect
                                    const std::vector<double> &compositional_fields,
                                    const Point<dim> &position) const;
 
+        /**
+        * Returns the volume fractions of all the phases over all of the current
+        * evaluation points. The outer loop is over the phases, and the inner
+        * loop is over all the evaluation points.
```
jdannberg, line 189: You still need to update the documentation. (the function doesn't return the volume fractions any more). 
I would also phrase the second part in a different way: You want to say what this new input object is and how it is filled. Something like: 
This function uses the MaterialModelInputs &in to fill the output_values of the phase_volume_fractions_out output object with [...]. This output_values object is a vector of vectors; the outer vector is expected to have a size that equals the number of unique phases, the inner vector is expected to have a size that equals the number of evaluation points.

jdannberg:
Great, thank you!

PR #3735, opened 2020-08-08
fix deprecation warnings in tests

tjhei:
we do ignore them, but this will break at some point if we don't fix
them.


tests/signals_connect.cc
```diff
@@ -8,7 +8,7 @@ using namespace aspect;
 
 template <int dim>
 void post_constraints_creation (const SimulatorAccess<dim> &,
-                                ConstraintMatrix &)
+                                aspect::ConstraintMatrix &)
```
bangerth, line 11: Can you revert this change and the corresponding changes in the other two files? I'm removing all of these uses of the class in #3738.

tjhei:
done

PR #3739, opened 2020-08-10
Modified calculate_isostrain_viscosities function inputs.

alarshi:
This PR modifies the existing calculate_isostrain_viscosities function such that it takes in a MaterialModelInputs structure instead of pressure, temperature, strain_rate, and composition. 
This is based on the suggestions by @gassmoeller  and @jdannberg, after their reviews on the diffusion branch.

### For all pull requests:


source/material_model/visco_plastic.cc
```diff
@@ -56,10 +56,18 @@ namespace aspect
        */
       bool plastic_yielding = false;
 
+      MaterialModel::MaterialModelInputs <dim> in_new (in_new.n_evaluation_points(), this->n_compositional_fields());
```
gassmoeller, line 59: Nice fix for the deprecated version of the function! I would still prefer to call it `in` not `in_new`. There is no other `in` here, and additions like `_new` or `_2` are almost always bad ideas, because they carry no meaning.

source/material_model/visco_plastic.cc
```diff
@@ -411,6 +418,9 @@ namespace aspect
 
           const double finite_difference_accuracy = 1e-7;
 
+          // A new material model inputs variable that uses the modified strain rate and pressure.
+          MaterialModel::MaterialModelInputs<dim> in_new = in;
```
gassmoeller, line 422: Here you need a new name, but try to give it a name that explains the meaning. E.g. `in_derivatives`

source/material_model/visco_plastic.cc
alarshi: Thanks, Rene for reviewing! I have modified the variable names in the latest commit.

gassmoeller:
Somewhere there must be a bug in the derivative calculation. The test `visco_plastic_derivatives_2d` reports different derivatives and the newton solver tests converge slower, which also indicates a somewhat wrong derivative. Can you take a look? Let me know if you need help.

alarshi:
@gassmoeller Thank you for reviewing the PR, Rene! I pushed the fix for the tests failing in my latest commit. It doesn't look very neat way to address the error, so please let me know how can I address the change better, whenever you have time :)

alarshi:
@gassmoeller Thank you for reviewing the PR, Rene! I pushed the fix for the tests failing in my latest commit. It doesn't look very neat way to address the error, so please let me know how can I address the change better, whenever you have time :)

gassmoeller:
This looks perfect, there is no better way to fix this! Ready to merge,

PR #3740, opened 2020-08-10
fix deal.II 9.2 instantiation bug in qprojector.h

### LINKED ISSUE #3728: potential compile error with unity build and deal.II 9.2.0
When changing the number of batched files with unity build, I am running into a deal.II bug that has been fixed after 9.2:
```
/Unity/unity_2_cxx.cxx.o.d -o CMakeFiles/aspect.dir/Unity/unity_2_cxx.cxx.o -c CMakeFiles/aspect.dir/Unity/unity_2_cxx.cxx
In file included from /hdscratch/shared/libs-candi-9.2.0-r1/deal.II-v9.2.0/include/deal.II/fe/mapping_cartesian.h:22:0,
                 from /home/heister/aspect-git/source/simulator/core.cc:53,
                 from CMakeFiles/aspect.dir/Unity/unity_2_cxx.cxx:97:
/hdscratch/shared/libs-candi-9.2.0-r1/deal.II-v9.2.0/include/deal.II/base/qprojector.h:414:54: error: specialization of ‘dealii::QIterated<dim>::QIterated(const dealii::Quadrature<1>&, unsigned int) [with int dim = 1]’ after instantiation
                         const unsigned int   n_copies);
```
see https://github.com/dealii/dealii/commit/1e5d09f19a26eb85cbf3a360586592283e358c90#diff-cababb69c891317230d592c968e1c64f for the bugfix.
We will compile cleanly as long as no header that includes qiterated.h is batched together with core.cc. This is not the case with default settings, but it might change randomly if a user adds another .cc file.
What to do: exclude core.cc from unity build (for now)?
Ignore this until it breaks something?
Any other suggestions? (include qprojector in every file before qiterated?)

@bangerth do you have an idea?

tjhei:
Fixes #3728, which could generate the following error
```
In file included from /hdscratch/shared/libs-
candi-9.2.0-r1/deal.II-v9.2.0/include/deal.II/fe/mapping_cartesian.h:22:0,
                 from /home/heister/aspect-
git/source/simulator/core.cc:53,
                 from CMakeFiles/aspect.dir/Unity/unity_2_cxx.cxx:97:
/hdscratch/shared/libs-
candi-9.2.0-r1/deal.II-v9.2.0/include/deal.II/base/qprojector.h:414:54:
error: specialization of ‘dealii::QIterated<dim>::QIterated(const
dealii::Quadrature<1>&, unsigned int) [with int dim = 1]’ after
instantiation
                         const unsigned int   n_copies);
```

This is due to deal.II 9.2. having an incorrect specialization of
QIterated inside quadrature.h.

### DIFF
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -26,7 +26,17 @@
 #include <aspect/material_model/utilities.h>
 
 #include <deal.II/base/point.h>
+
+// Work around an incorrect instantiation in qprojector.h of deal.II 9.2.0,
+// which requires including qprojector.h before quadrature.h (and not
+// after). This file doesn't actually need qprojector.h, so the include can be
+// removed when we require 9.3.. For more info see
+// https://github.com/geodynamics/aspect/issues/3728
+#if !DEAL_II_VERSION_GTE(9,3,0)
+#include <deal.II/base/qprojector.h>
+#endif
 #include <deal.II/base/quadrature.h>
+
 #include <deal.II/base/symmetric_tensor.h>
 #include <deal.II/base/parameter_handler.h>
 #include <deal.II/dofs/dof_handler.h>


include/aspect/material_model/interface.h
```diff
@@ -26,7 +26,17 @@
 #include <aspect/material_model/utilities.h>
 
 #include <deal.II/base/point.h>
+
+// Work around an incorrect instantiation in qprojector.h of deal.II 9.2.0,
```
bangerth, line 30: Very strictly speaking, it's not an instantiation that is in that file (which would be preceded by `template ...`) but a declaration of a specialization (preceded by `template <> ...`), but who cares. This is going to go away in a year or two anyway :-)

PR #3742, opened 2020-08-10
Require at least one nonlinear iteration

gassmoeller:
In response to #3691 this sets the minimum number of nonlinear iterations to 1. If models require no solve performed at all they should choose the nonlinear solver scheme `no Advection, no Stokes` instead of setting the number of iterations to 0.


tjhei:
Looks good, but there are two more broken tests that need updating.

gassmoeller:
Since this is approved and the tests pass I am going to merge to make progress with #3691.

PR #3743, opened 2020-08-10
Have time as a postprocessor particle property

EstherHeck:

I added a postprocessor that writes the current model time into the files of particle properties.

### For all pull requests:

### DIFF
--- a/include/aspect/particle/property/time_particle.h
+++ b/include/aspect/particle/property/time_particle.h
@@ -1,0 +1,112 @@
+/*
+ Copyright (C) 2015 - 2017 by the authors of the ASPECT code.
+
+ This file is part of ASPECT.
+
+ ASPECT is free software; you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation; either version 2, or (at your option)
+ any later version.
+
+ ASPECT is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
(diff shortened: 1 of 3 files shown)


gassmoeller:
Hi Esther, can you explain the purpose for which you want to use this? It feels like a lot of memory to store the same model time as a property for every particle, and each particle output file is already associated with a time (that is how paraview knows the time for each file). Do you need this for further calculations?

EstherHeck:
Hi, I need this when I output particle properties to ascii files and then plot them directly from there without paraview. Like in my case velocity over time. Then I have it all in the same set of files, which makes it easier.

bangerth:
Oh, I didn't even read @EstherHeck 's comment above and asked the same question that @gassmoeller already did.

I think what we're both saying is that there ought to be a better way to get at that information at the place where you're plotting it. When you're opening a file with particle information, you should be able to infer the corresponding time by looking at the statistics file what time that file corresponds to. It's then just a constant number you would use in gnuplot.

EstherHeck:
ok. I found it useful, so I thought others might find it too. But if you don't think so, I will just use it for myself as a library. Should I then close this PR?

bangerth:
But then we also don't want to discouraging! I think what we're both concerned about is people using this when they have a billion particles.

Do you know how to create your visualizations without this data in gnuplot? How about we close this PR and we'll help you with finding an alternative way to do what you want to do if you haven't found one already?

EstherHeck:
Don't worry, I am not discouraged at all. And yes, you are right, when using a billion particles that is not helpful. I used 10 as a sort of analog for a GPS station. If you have a good idea on how to do that differently, that would be nice. Especially because I am thinking to switch to having active particles to track the composition

bangerth:
Can you show me how you're currently using gnuplot to visualize your data?

Issue #3657, opened 2020-08-03
Move cookbooks and their documentation into folders


tjhei:
- Like done in #3422, move each cookbook into a folder under ``cookbooks/`` like ``convection-box``
- Move the latex including pictures/etc. from the manual into a doc/ folder 
- Also pay attention to use underscores instead of dashes #3255

alarshi:
I will work on this if noone else has already started it.

tjhei:
> I will work on this if noone else has already started it.

Sure, please start with a single .prm and get it reviewed. After that, it also make sense to work in smaller batches to make it easier to review.

jdannberg:
There's one more thing I noticed: We also have to update the section in the manual that describes how to contribute a cookbook. It should explain the new directory structure, and how to create the symbolic link in the doc/manual/cookbooks folder. 

- [ ] update section in the manual

bangerth:
I believe that this is finished.

PR #3744, opened 2020-08-10
Cookbooks files movearound

alarshi:
This is a PR to address issue #3657 for the Burnman cookbook.

* [x ] I have read the guidelines in our [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) document.

### For all pull requests:


tjhei:
The documentation doesn't compile but fails with 
```make: *** No rule to make target 'manual/cookbooks/burnman/doc/adiabatic_conditions.part.prm.out', needed by 'manual.pdf'.  Stop.
```
I think your symbolic link should be to the parent directory (and not to /doc/).

cookbooks/burnman/doc/burnman.tex
```diff
@@ -15,25 +16,28 @@ \subsubsection{2D compressible convection with a reference profile and material
 to generate the reference profile has the advantage that all the material property data
 are consistent, for example, the gravity profile is computed using the reference density.
 \begin{figure}
-  \includegraphics[width=\textwidth]{./reference_profile.pdf}
+  \includegraphics[width=\textwidth]{../../cookbooks/burnman/doc/reference_profile.pdf}
```
tjhei, line 19: I think you can also write this and all other places as 
```suggestion
  \includegraphics[width=\textwidth]{cookbooks/burnman/doc/reference_profile.pdf}
```

Can you try?

tjhei:
This looks good (and compiles). I will let you check if the simpler path also works. Otherwise, this is good to go.

cookbooks/burnman/doc/burnman.tex
alarshi: Yes, I will try that and also remove some past commits. I was just wondering that the path is relative to the manual folder, so if we remove the linked list, this path would not work, or am I understanding it incorrectly?

gassmoeller:
Tests have finished. Thanks!

PR #3745, opened 2020-08-10
add particle advection substep update call.

MFraters:
This change allows for particle property plugin to ask for an update every advection step so that they can internally carry out the integration scheme to update the particle properties while the particles are advected. It would be the responsibility of the particle property plugins themselves to ask which integration scheme is being used, store all the substeps, and combine them at the correct update call.

### DIFF
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -259,7 +259,17 @@ namespace aspect
          * integration of solution properties or time varying particle
          * properties are used while solving the model problem.
          */
-        update_time_step
+        update_time_step,
+        /**
+         * Update the particles at every particle advection substep. The substeps are
+         * defined by the particle integrator which include an euler, a RK2 and a RK4
+         * scheme. This flags make sure that the update function is called at every
+         * integration substep and at the end (which is the same time as update_time_step).
+         * This flag is only necessary if the particle properties need to update the data
+         * based on the integration substeps, for example a tensor which needs to be
+         * rotated by the advection.
+         */
+        update_time_substep
       };
 
       /**
--- a/source/particle/world.cc
+++ b/source/particle/world.cc
@@ -731,6 +731,8 @@ namespace aspect
       do
         {
           advect_particles();
+          if (property_manager->need_update() == Property::update_time_substep)
+            update_particles();
         }
       // Keep calling the integrator until it indicates it is finished
       while (integrator->new_integration_step());
@@ -738,7 +740,7 @@ namespace aspect
       apply_particle_per_cell_bounds();
 
       // Update particle properties
-      if (property_manager->need_update() == Property::update_time_step)
+      if (property_manager->need_update() == Property::update_time_step || property_manager->need_update() == Property::update_time_substep)
         update_particles();
 
       // Now that all particle information was updated, exchange the new


include/aspect/particle/property/interface.h
```diff
@@ -259,7 +259,13 @@ namespace aspect
          * integration of solution properties or time varying particle
          * properties are used while solving the model problem.
          */
-        update_time_step
+        update_time_step,
+        /**
+         * Update the particles at every advection substep. This is only necessary
```
bangerth, line 264: Can you explain what a "substep" is? Do you mean "after every solve that updates one of the compositional fields", rather than "after the last compositional field has been updated"?

bangerth: Oh, or are you thinking of a multistep RK scheme? In any case, I think that would be worth outlining.

MFraters:
@bangerth I have expanded the explanation in the function documentation. Does it make more sense now?


> I think I don't quite understand why you need this. Can you outline a case where that is necessary, and maybe come up with a test case that illustrates this?
> 
> If you're thinking of this as the sub-steps of an RK scheme, presumably you want to update things based on the current state -- but how would the particle know which integrator is used, how each sub-step is to be intrepreted, etc?

Yes, that is exactly what I have in mind. The application is for the LPO particle plugin, where the plugin needs to be called at each substep of the integration. The idea is that if the particle property plugin returns this flag, that it checks itself with the integrator what type it is and how many substeps there are and act accordingly. I have discussed several options for this with @gassmoeller, and this is what we thought was the least intrusive way if implementing it. If you have other suggestions I will be happy to hear them.

source/particle/world.cc
```diff
@@ -731,14 +731,16 @@ namespace aspect
...
+            update_particles();
         }
       // Keep calling the integrator until it indicates it is finished
       while (integrator->new_integration_step());
 
       apply_particle_per_cell_bounds();
 
       // Update particle properties
-      if (property_manager->need_update() == Property::update_time_step)
+      if (property_manager->need_update() == Property::update_time_step || property_manager->need_update() == Property::update_time_substep)
```
gassmoeller, line 743: Do you need the `||` here? You already called `update_particles` in line 735 above, and nothing has changed since then (a few particles may have been removed or new added in line 740, but that is unrelated).

include/aspect/particle/property/interface.h
```diff
@@ -259,7 +259,17 @@ namespace aspect
          * integration of solution properties or time varying particle
          * properties are used while solving the model problem.
          */
-        update_time_step
+        update_time_step,
+        /**
+         * Update the particles at every particle advection substep. The substeps are
+         * defined by the particle integrator which include an euler, a RK2 and a RK4
```
gassmoeller, line 265: R is one of those weird characters in English that is a consonant, but is treated like a vowel for the purpose of a/an (because it is pronounced Arrrrr, or something like that :-)).

```suggestion
         * defined by the particle integrator which include an Euler, an RK2 and an RK4
```

include/aspect/particle/property/interface.h
```diff
@@ -259,7 +259,17 @@ namespace aspect
...
+         * scheme. This flags make sure that the update function is called at every
+         * integration substep and at the end (which is the same time as update_time_step).
```
gassmoeller, line 267: I dont think you need the extra `at the end`. It is called after every integration substep, that should be enough.

source/particle/world.cc
bangerth: Yes, I think you're actually calling it twice.

gassmoeller:
> I suspect that they all actually look like this. Note that they update the property with the full time step. That's going to be wrong now: We don't actually want this updating to happen 4 times per time step, but that's exactly what's going to happen if there is just one property that requests update_time_substep. We just introduced really difficult to find bugs because the patch is silently introducing a bug into every existing particle property function.

Ah, yes that is a problem. I think Menno wants to use this update scheme for particles that have a single very specific property that requires and supports this scheme. But nevertheless we need to ensure that if this update scheme is selected, all selected particle properties have to use the same scheme (or do not update at all). @MFraters can you add an assert for this?

bangerth:
I missed that discussion, but maybe you can explain what you really wanted to do and see if we can find a different solution? What is the property you need to update in each sub-step?

source/particle/world.cc
MFraters: I modelled it after how the integrators work internally: call supbsteps and then have a seperate call to combine them, but I now realize that that would be quite expensive for because a update is just expensive. Since it is not needed, I will remove this line change.

MFraters:
After some discussion with @bangerth and @gassmoeller, I think this is indeed not the best way forward for the problem I want to solve, so I will close this pull request.

PR #3746, opened 2020-08-10
fix mesh refinement level in if-else statement

mibillen:
Changed the minimum refinement set in the example of using an if-else function expression to be the same as the level set with initial global refinement (3). The else part had been set to 0, which leads to coarsening below the level set by the initial global refinement. Added text to describe what this values does when set to 3 versus what would happen if it was set to 0.  
- new prm file with changed parameter and updated text (ran this and it works)
- updated the text in the manual
- replaced the figure showing temperature on the refined mesh (this changed because of the changed parameter)
- changed the code snippet used in the directory to match that in the updated prm
- removed an un-used png-figure from the manual cookbook directory


cookbooks/muparser_temperature_example/muparser-temperature-example.prm
```diff
@@ -61,12 +61,12 @@ end
 
 # Refine the upper 150 km of the mesh so lithosphere structure is resolved.
 # Function expression asks if the depth is less than the lithosphere depth and
-# then refinement level = 7, else refinement level = 0. 
-# Note, the minimum refinement level for the lower 850 km of the mesh would be 
-# coarser than the initial one level of global refinement. This would then lead
+# then refinement level = 7, else refinement level = 3. 
+# Note, if the minimum refinement level for the lower 850 km of the mesh would be 
+# less than the initial global refinement level, then this would lead
```
gassmoeller, line 66: Did you actually observe this case? It seems like a bug that the minimum refinement function would overwrite the general parameter `Minimum refinement level`. I was not aware this is happening.

gassmoeller:
Then again, maybe I do not quite understand the loop in core.cc correctly (maybe it continues on lower levels). Still a first question would be: Do you have a model that shows this behavior? It seems like a bug to me.

tjhei:
> maybe I do not quite understand the loop in core.cc correctly

Line 1731? This clears the "coarsen" flag for all cells with level <= min_grid_level. Meaning, no cell can be coarsened if it is at or below the minimum. That looks correct to me.

gassmoeller:
So if we start on level 0 `++cell` will move to the next higher level? Ok, then I misunderstood the loop. But then I still do not understand how Magali can get cells that are coarser than the global minimum refinement level.

tjhei:
> So if we start on level 0 `++cell` will move to the next higher level?

Yes. You go through all cells (sorted by level).

> But then I still do not understand how Magali can get cells that are coarser than the global minimum refinement level.

She uses "minimum refinement function", right? That is different than "Mesh refinement"/"Minimum refinement level".

gassmoeller:
Yes, but the input file that is posted in this PR has
```
set Minimum refinement level = 3

  set Function expression = (something) ? 7 : 0
end
```
And I do not understand why the `0` should circumvent the global minimum refinement level of 3 as the text suggests.

tjhei:
> And I do not understand why the `0` should circumvent the global minimum refinement level of 3 as the text suggests.

Same, I don't understand that either.

mibillen:
I will try to look at this again tonight.

mibillen:
I will try to look at this again tonight.

mibillen:
I'm sorry I was able to check this earlier to save you all time.  The reason I started this was because of this PR https://github.com/geodynamics/aspect/pull/2973. The explanation given there (last comments) states that using a level of 0 in the else statement will lead to one less level of refinement in the bottom of the box. I have now tested this carefully (which I should have done before). This does occur, but only if the Minimum refinement level is not set.   In this case, if I have  set Function expression = ((ymax-y<=lith) ? 7 : 0),
level0
 the bottom of the box indeed has one level lower refinement then if I have  set Function expression = ((ymax-y<=lith) ? 7 : 3).
level3

So, I think what I need to do is revise the explanation to explain that if the Minimum refinement level is not set, then using "else 0" will lead to less refinement than using "else 3" which matches the initial refinement level.   It is still not clear to me why its only one level lower refinement. So, I don't understand what '"else 0" means in the case when the minimum refinement level is not set... its not 0 refinement, or maybe it is if the box is deeper and the mesh is allowed to coarsen more?

gassmoeller:
> It is still not clear to me why its only one level lower refinement. So, I don't understand what '"else 0" means in the case when the minimum refinement level is not set... 

It means there is *in theory* no hard limit on how coarse the mesh can get. However, *in practice* there could be other constraints on the mesh, although in this example I can not tell you which of them exactly prevent it from being coarsened. There always is the 2:1 refinement rule (meaning neighbouring cells can at most differ by one level in refinement), but that should work here. Deal.II has additional mesh smoothing flags that prevent some weird mesh configurations (such as small islands of refined/coarsened cells), but I think most of them are switched off for recent ASPECT version. Also on periodic boundary conditions the mesh has to match as far as  know, but that is not the case here either. So maybe the algorithm is confused, because you only prescribe the minimum refinement function, which provides hard upper and lower limits on the mesh level, but does not actually compute an error indicator per cell. So the algorithms may not know which cells to refine or coarsen, because all of them have an error of 0. What happens if you add a second refinement strategy, e.g. temperature?

mibillen:
Adding the temperature strategy does change things. However, I did test the effect of the initial adaptive refinement (IAR). If the IAR is such that the total refinement level (repetitions+initial global +IAR) is less than the level I set in the Set minimum refinement function, then I get the lower value. So if I Rep+initial-global + IAR = 6, but I set ((ymax-y<=lith) ? 7 : 0) in the minimum refinement function, then the actual refinement is only at 6, not 7 (useful to know).   Finally, I found the answer to what 'else 0" means. I found if I made the box deeper, then the refinement level became coarser by one more level in the bottom of the box. And, if I make it even deeper, it becomes even coarser again. So, I think that when the else is set to '0' the bottom level is constrained geometrically, it can literally coarsen to level 0 (or maybe 1, since I have repetitions on).   I think I know how to edit the text now. One should use an else value that matches the minimum refinement level you want in the rest of the box. That may or may not match your global refinement. The minimum level of refinement that appears  if else is less than the rep+iGR+iAR depends on geometry. 
Here's a picture of the deeper box. 
Screen Shot 2020-08-13 at 11 46 24 PM

mibillen:
The screen-shot I just posted had iAR = 3, the previous pics had iAR = 4, that 's why the finest level is not as small. But the rest of the mesh is the same when I make iAR = 4. 
Screen Shot 2020-08-13 at 11 54 14 PM

PR #3731, opened 2020-08-07
Implement a time update() function for the gravity function plugin

jaustermann:
Pull Request Checklist. Please read and check each box with an X. Delete any part not applicable. Ask on the [forum](https://community.geodynamics.org/c/aspect) if you need help with any step.

*Describe what you did in this PR and why you did it.*

I added a time update() function to the gravity function. I did that following the implementation in other function plugins. I did not test this addition because it's not immediately obvious to me how it would best be tested but am happy to do so if you have suggestions on how this would be done. This addresses issue #3341. 

### For all pull requests:

* [X] I have followed the [instructions for indenting my code](../blob/master/CONTRIBUTING.md#making-aspect-better).

### For new features/models or changes of existing features:

* [x] I have tested my new feature locally to ensure it is correct.
* [x] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/master/tests/) directory.
* [x] I have added a changelog entry in the [doc/modules/changes](../blob/master/doc/modules/changes) directory that will inform other users of my change.

PR #3749, opened 2020-08-10
fix gravity test.

MFraters:
The tester in #3745 fails because of a difference in the tester output in an unrelated file/test. This fixes the test reference output.


tjhei:
oh no, I think I broke master by merging #3731 before the tester finished. :-(

MFraters:
> oh no, I think I broke master by merging #3731 before the tester finished. :-(

ah, I guess that's why two of the testers are now set as required? :P

tjhei:
> ah, I guess that's why two of the testers are now set as required? :P

yes, a mistake that is easy to make...

gassmoeller:
Tests have finished.

PR #3750, opened 2020-08-11
update melt force vector test

jdannberg:
This pull request updates the melt force vector test so that all errors in the screen output now converge. This is achieved by using the correct reference solution in the .cc file. 

This is useful because this is one of the few tests that uses additional material model outputs to add terms to the RHS of the equations, and this update makes it more clear how to use this feature. 

### DIFF
--- a/tests/melt_force_vector.cc
+++ b/tests/melt_force_vector.cc
@@ -80,13 +80,11 @@ namespace aspect
                             typename MaterialModel::Interface<dim>::MaterialModelOutputs &out) const
       {
         const unsigned int porosity_idx = this->introspection().compositional_index_for_name("porosity");
-        double c = 1.0;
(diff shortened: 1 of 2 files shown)


jdannberg:
Yes, the reference solution for the fluid pressure and the fluid velocity weren't set to the correct expressions (for example the fluid velocity was just set to 1). So I suspect it's not that someone made a mistake, but more like we didn't bother to fill in the correct solutions because the solid pressure and fluid velocity are not solution variables (so we didn't care about their convergence rate).

jdannberg:
Do you want me to rebase because of the WorldBuilder issues in the tests?

tjhei:
> Do you want me to rebase because of the WorldBuilder issues in the tests?

not necessary.

Issue #3399, opened 2020-03-02
Should initial topography be part of the mesh deformation module?


anne-glerum:
Following the discussion between @gassmoeller, @tjhei and me during the last user meeting, I'd like to open up the following discussion:

1. **Should the `initial topography` functionality become part of the `mesh deformation` modules, instead of being its own module?** In the end, both `initial topography` and `mesh deformation` want to modify the mesh. The former is only restricted to a specific point in time. 
  Some context: To diffuse the free surface, I need access to the current height of the surface points. Mesh displacements are stored and can thus be queried for any topography gained over time. However, they do not include the initial topography. My lengthy workaround was 1) to store the initial topography similarly to the mesh displacements; or 2) to apply a mesh displacement at t1 instead of initial topography at t0. Rene however suggested making the whole `initial topography` module part of `mesh deformation`.

2. **If so, should we change/expand the interface of the `mesh deformation` plugins?** Currently, one has to specify a boundary velocity, which is internally translated to a mesh displacement based on the timestep. However, specifying an initial topography through a boundary velocity (with the timestep unknown) is cumbersome. One would like to directly prescribe the displacement.

tjhei:
1. Yes, I think that makes a lot of sense to me.
2. I find it more natural if the mesh deformation returns a displacement, not a velocity. Going from velocity to displacement is easy enough.

gassmoeller:
1. I agree. Reducing the number of plugin systems while keeping the functionality sounds great and is more intuitive as well (any kind of mesh deformation is a mesh deformation plugin then :-)).
2. Displacement is more intuitive, but I think there was a reason why it prescribes velocity at the moment (possibly, because for the free surface we need to correct for mesh movements in the advection equation). For the initial topography that is not relevant, so maybe we can offer both functions in the interface and do something like: `total_displacement = plugin_displacement + plugin_velocity * timestep` in the calling function? Not sure that will work, but I would want to at least check.

anne-glerum:
> 2. Displacement is more intuitive, but I think there was a reason why it prescribes velocity at the moment (possibly, because for the free surface we need to correct for mesh movements in the advection equation). For the initial topography that is not relevant, so maybe we can offer both functions in the interface and do something like: total_displacement = plugin_displacement + plugin_velocity * timestep in the calling function? Not sure that will work, but I would want to at least check.

[...]

PR #3751, opened 2020-08-11
Extend initial mesh deformation interface

gassmoeller:
This is a first step towards addressing #3399. This PR extends the interface for mesh deformation plugins to also prescribe initial deformations (instead of just deformations later during the model run). Plugins have both functions and can choose to implement either one (or both). This currently comes without a tests case (I will add that later) to allow @ricitron to test his model with this new interface. 
@ricitron: In order to make an initial_topography plugin work as a mesh deformation plugin the following steps are needed:
1. Make a copy of the initial topography plugin (.cc and .h file) and place it in the corresponding mesh_deformation folders
2. Adjust names of header files, input parameters according to the new plugin system (initial topography -> mesh deformation and so on).
3. rework the `value` function from the following old function declaration:
```
template <int dim>
    double
    AsciiData<dim>::
    value (const Point<dim-1> &surface_point) const
```
to the new function:
```
    template <int dim>
    Tensor<1,dim>
    AsciiData<dim>::
    compute_initial_deformation_on_boundary(const types::boundary_id boundary_indicator,
                                                                         const Point<dim> &position) const
```

The last step is the tricky part, you need to convert an amount of topography (`double`) into a displacement vector (`Tensor<1,dim>`). You can start with the following simplistic implementation and then try something more complicated afterwards:
```
    template <int dim>
    Tensor<1,dim>
    AsciiData<dim>::
    compute_initial_deformation_on_boundary(const types::boundary_id boundary_indicator,
                                                                         const Point<dim> &position) const
{
      Tensor<1,dim> initial_deformation;
      initial_deformation[dim-1] = position[0];
      return initial_deformation;
}
```


include/aspect/coordinate_systems.h
```diff
@@ -45,6 +45,38 @@ namespace aspect
...
+            case CoordinateSystem::cartesian:
+            {
+              depth_direction[dim-2] = -1.0;
+
+              break;
+            }
+            case CoordinateSystem::spherical:
+            case CoordinateSystem::ellipsoidal:
+            {
+              depth_direction = position / position.norm();
```
bangerth, line 70: I think I don't quite get what the *coordinate system* has to do with it? You seem to be thinking of the coordinate system as a proxy for the kind of geometry you're working in, but that's not how `CoordinateSystem` was intended to be used, I believe, but rather as a way to expressing how a tuple of values should be interpreted as coordinates for a point.

include/aspect/coordinate_systems.h
```diff
@@ -45,6 +45,38 @@ namespace aspect
         ellipsoidal,
         invalid
       };
+
+      template<int dim>
+      Tensor<1, dim>
+      get_depth_direction(Point<dim> &position, CoordinateSystem coordinate_system)
```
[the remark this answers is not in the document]
bangerth: By the way, in other places we determine what "vertical" is by projecting onto the gravity direction.

include/aspect/mesh_deformation/interface.h
```diff
@@ -87,17 +87,30 @@ namespace aspect
          */
         virtual void update();
 
+        /**
+        * A function that returns the initial deformation of certain mesh
+        * vertices (e.g. the surface vertices). @p position is the original
+        * position of each vertex and this function is expected to return the
+        * displacement vector of this position. The default implementation does
+        * return a zero displacement (=no initial deformation).
```
bangerth, line 95: ...returns a zero...

include/aspect/mesh_deformation/interface.h
```diff
@@ -157,6 +170,11 @@ namespace aspect
          */
         void update();
 
+        /**
+         * Compute the initial mesh deformation from plugins that implement it.
+         */
+        void execute_initial_deformation();
```
bangerth, line 176: Does it *compute* it, or does it actually *execute* the deformation? Function name and documentation disagree :-)

(Separately, maybe "perform" is a better word than "execute".)

include/aspect/mesh_deformation/interface.h
```diff
@@ -295,6 +313,23 @@ namespace aspect
...
+        * problem, which computes the initial displacements of the internal
+        * vertices so that the mesh does not become too distorted due to
+        * motion of the surface. Displacements of vertices on the
+        * deforming surface are fixed according to the selected deformation
+        * plugins.
+        */
+        AffineConstraints<double> make_initial_constraints ();
+
+        /**
+         * Solve vector Laplacian equation for initial mesh displacements
```
bangerth, line 327:
```suggestion
         * Solve a vector Laplacian equation for initial mesh displacements
```

include/aspect/mesh_deformation/interface.h
```diff
@@ -295,6 +313,23 @@ namespace aspect
...
+         * given the @p initial_mesh_displacement_constraints provided
+         * as argument.
+         */
+        void compute_initial_mesh_displacements (const AffineConstraints<double> &initial_mesh_displacement_constraints);
```
bangerth, line 331: Since the function computes something, can you say where it stores it?

source/mesh_deformation/interface.cc
```diff
@@ -308,6 +333,18 @@ namespace aspect
...
 
 
+    template <int dim>
+    void MeshDeformationHandler<dim>::execute_initial_deformation()
+    {
+      TimerOutput::Scope timer (sim.computing_timer, "Mesh deformation initialize");
+      const AffineConstraints<double> initial_deformation_constraints = make_initial_constraints();
+
+      compute_initial_mesh_displacements(initial_deformation_constraints);
+
```
bangerth, line 343:
```suggestion
```

bangerth: Why not let `compute_initial_displacements()` call `make_initial_constraints()`? This would seem more natural, and then you don't need the `execute_...()` function at all.

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +501,116 @@ namespace aspect
...
+      AssertThrow(this->get_parameters().mesh_deformation_enabled, ExcInternalError());
+
+      AffineConstraints<double> initial_deformation_constraints;
+
+      // Now construct the mesh displacement constraints
+      initial_deformation_constraints.reinit(mesh_locally_relevant);
+
+      // mesh_velocity_constraints can use the same hanging node
+      // information that was used for mesh_vertex constraints.
+      initial_deformation_constraints.merge(mesh_vertex_constraints);
```
bangerth, line 516: could you just copy? the lhs object is empty here

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +501,116 @@ namespace aspect
...
+                     mesh_deformation_boundary_indicators_set.begin(),mesh_deformation_boundary_indicators_set.end(),
+                     std::inserter(all_deformable_boundaries, all_deformable_boundaries.end()));
+
+      std::set<types::boundary_id> all_fixed_boundaries;
+      std::set_difference(all_boundaries.begin(),all_boundaries.end(),
+                          all_deformable_boundaries.begin(),all_deformable_boundaries.end(),
+                          std::inserter(all_fixed_boundaries, all_fixed_boundaries.end()));
+
+      // Zero out the displacement for the fixed boundaries
+      for (const types::boundary_id &boundary_id: all_fixed_boundaries)
```
bangerth, line 540:
```suggestion
      for (const types::boundary_id &boundary_id : all_fixed_boundaries)
```

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +501,116 @@ namespace aspect
...
+                                                    Functions::ZeroFunction<dim>(dim),
+                                                    initial_deformation_constraints);
+        }
+
+      // Make tangential deformation constraints for tangential boundaries
+      VectorTools::compute_no_normal_flux_constraints (mesh_deformation_dof_handler,
+                                                       /* first_vector_component= */
+                                                       0,
+                                                       tangential_mesh_boundary_indicators,
+                                                       initial_deformation_constraints, this->get_mapping());
```
bangerth, line 554:
```suggestion
                                                       initial_deformation_constraints, 
                                                       this->get_mapping());
```

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +501,116 @@ namespace aspect
...
+                        {
+                          const double inhomogeneity = plugin_constraints.get_inhomogeneity(*index);
+                          plugin_constraints.set_inhomogeneity(*index, current_plugin_constraints.get_inhomogeneity(*index) + inhomogeneity);
+                        }
+                    }
+                }
+            }
+        }
+
+      initial_deformation_constraints.merge(plugin_constraints,AffineConstraints<double>::left_object_wins);
```
bangerth, line 606:
```suggestion
      initial_deformation_constraints.merge(plugin_constraints, AffineConstraints<double>::left_object_wins);
```

include/aspect/coordinate_systems.h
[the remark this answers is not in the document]
gassmoeller: Ah, I forgot about that convention. I was confused we had no function for it, and did not want to implement it in every geometry model. You are right, we should stick to the definition that gravity defines *down*, although I am sure this is going to bite us once we have a gravity solver included ^^ ... but let's worry about that when it actually happens. I will remove the function.

gassmoeller:
I addressed the comments. Testcase still missing.

source/mesh_deformation/interface.cc
gassmoeller: If I just replace `.merge` by `.copy_from` I get the following error:
```
00: --------------------------------------------------------
100: An error occurred in line <1949> of file </home/rene/software/deal.II-dev/include/deal.II/lac/affine_constraints.h> in function
100:     void dealii::AffineConstraints<number>::add_line(dealii::AffineConstraints<number>::size_type) [with number = double; dealii::AffineConstraints<number>::size_type = unsigned int]
100: The violated condition was: 
100:     sorted == false
100: Additional information: 
100:     (none)
100: 
100: Stacktrace:
100: -----------
100: #0  /home/rene/software/aspect-debug/aspect: dealii::AffineConstraints<double>::add_line(unsigned int)
100: #1  /home/rene/software/deal.II-dev/lib/libdeal_II.g.so.9.3.0-pre: void dealii::VectorTools::interpolate_boundary_values<2, 2, dealii::DoFHandler, double>(dealii::Mapping<2, 2> const&, dealii::DoFHandler<2, 2> const&, std::map<unsigned int, dealii::Function<2, double> const*, std::less<unsigned int>, std::allocator<std::pair<unsigned int const, dealii::Function<2, double> const*> > > const&, dealii::AffineConstraints<double>&, dealii::ComponentMask const&)
100: #2  /home/rene/software/deal.II-dev/lib/libdeal_II.g.so.9.3.0-pre: void dealii::VectorTools::interpolate_boundary_values<2, 2, dealii::DoFHandler, double>(dealii::Mapping<2, 2> const&, dealii::DoFHandler<2, 2> const&, unsigned int, dealii::Function<2, double> const&, dealii::AffineConstraints<double>&, dealii::ComponentMask const&)
100: #3  /home/rene/software/aspect-debug/aspect: aspect::MeshDeformation::MeshDeformationHandler<2>::make_initial_constraints()
```

And I get the same error if I use the copy-constructor.

gassmoeller:
Tests are fixed.

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +489,114 @@ namespace aspect
...
+      const std::set<types::boundary_id> all_boundaries = this->get_geometry_model().get_used_boundary_indicators();
+      std::set<types::boundary_id> all_deformable_boundaries;
+      std::set_union(tangential_mesh_boundary_indicators.begin(),tangential_mesh_boundary_indicators.end(),
+                     mesh_deformation_boundary_indicators_set.begin(),mesh_deformation_boundary_indicators_set.end(),
+                     std::inserter(all_deformable_boundaries, all_deformable_boundaries.end()));
+
+      std::set<types::boundary_id> all_fixed_boundaries;
+      std::set_difference(all_boundaries.begin(),all_boundaries.end(),
+                          all_deformable_boundaries.begin(),all_deformable_boundaries.end(),
+                          std::inserter(all_fixed_boundaries, all_fixed_boundaries.end()));
```
anne-glerum, line 521: What about the periodic boundaries? Aren't they counted twice?

gassmoeller:
Very good point. I have fixed the handling of periodic boundaries. They are treated like tangential mesh boundaries for the time evolution, so I did the same here. I also added a test that confirms they are now identical to tangential mesh boundaries.

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +489,127 @@ namespace aspect
...
+                                                 (*p).first.second,
+                                                 (*p).second,
+                                                 initial_deformation_constraints);
+        }
+
+      const std::set<types::boundary_id> all_boundaries = this->get_geometry_model().get_used_boundary_indicators();
+
+      // Periodic boundaries are also considered tangential mesh movement boundaries
+      std::set<types::boundary_id> all_tangential_boundaries;
+      std::set_union(tangential_mesh_boundary_indicators.begin(),tangential_mesh_boundary_indicators.end(),
```
anne-glerum, line 522: These tangential_mesh_boundary_indicators include `Additional tangential mesh boundary indicators` and `Tangential velocity boundary indicators`. A mesh deformation boundary can also have a tangential Stokes velocity boundary condition set on it. So these should be taken out, otherwise the mesh def BI will have no normal flux BC set.

gassmoeller:
Ok, next try. Let's see if I got it right this time. Instead of fixing the problem after we have this confusing list of boundaries I decided to fix the problem at its source in `parse_parameters`. This also simplifies the other make_constraints function. @anne-glerum could you take another look?

include/aspect/mesh_deformation/interface.h
```diff
@@ -87,17 +87,30 @@ namespace aspect
          */
         virtual void update();
 
+        /**
+        * A function that returns the initial deformation of certain mesh
+        * vertices (e.g. the surface vertices). @p position is the original
```
tjhei, line 92: We currently use a MappingQ1Eulerian meaning the positions are indeed the vertices, but this is not necessarily the case in the future. Does it make sense to make this comment more general and just state that this will be evaluated on various points on the boundary?

source/mesh_deformation/interface.cc
```diff
@@ -282,6 +310,20 @@ namespace aspect
       }
       prm.leave_subsection ();
 
+      // Remove all prescribed mesh deformation boundary indicators from the
```
tjhei, line 313: why "prescribed"?

source/mesh_deformation/interface.cc
```diff
@@ -282,6 +310,20 @@ namespace aspect
...
+      // set of tangential mesh deformation boundary indicators
+      for (const auto &boundary_id : mesh_deformation_boundary_indicators_set)
+        tangential_mesh_boundary_indicators.erase(boundary_id);
+
+      // Add all periodic boundaries to the set of tangential mesh deformation
+      // boundary indicators
```
tjhei, line 319: The comment says what you are doing, but not why. Would you be willing to add a sentence (here and above)?

source/mesh_deformation/interface.cc
tjhei: a call to compress() should fix this error (but I think you can leave things as is).

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +483,121 @@ namespace aspect
...
+      // initial_deformation_constraints can use the same hanging node
+      // information that was used for mesh_vertex constraints.
+      AffineConstraints<double> initial_deformation_constraints(mesh_locally_relevant);
+      initial_deformation_constraints.merge(mesh_vertex_constraints);
+
+      // Add the vanilla periodic boundary constraints
+      std::set< types::boundary_id > periodic_boundaries;
+      using periodic_boundary_pairs = std::set< std::pair< std::pair<types::boundary_id, types::boundary_id>, unsigned int> >;
+      periodic_boundary_pairs pbp = this->get_geometry_model().get_periodic_boundary_pairs();
+      for (periodic_boundary_pairs::iterator p = pbp.begin(); p != pbp.end(); ++p)
```
tjhei, line 500: Maybe just use ``for (const auto& p : this->get_geometry_model().get_periodic_boundary_pairs())``?

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +483,121 @@ namespace aspect
...
+      for (const types::boundary_id &boundary_id : all_fixed_boundaries)
+        {
+          VectorTools::interpolate_boundary_values (this->get_mapping(),
+                                                    mesh_deformation_dof_handler,
+                                                    boundary_id,
+                                                    Functions::ZeroFunction<dim>(dim),
+                                                    initial_deformation_constraints);
+        }
+
+      // Make tangential deformation constraints for tangential boundaries
```
tjhei, line 534: I am lost with all these different sets of boundaries. Can we have a comment somewhere that untangles all this?

source/mesh_deformation/interface.cc
```diff
@@ -464,6 +483,121 @@ namespace aspect
...
+
+              VectorTools::interpolate_boundary_values (this->get_mapping(),
+                                                        mesh_deformation_dof_handler,
+                                                        boundary_id->first,
+                                                        vel,
+                                                        current_plugin_constraints,
+                                                        ComponentMask());
+
+              const IndexSet local_lines = current_plugin_constraints.get_local_lines();
+              for (auto index = local_lines.begin(); index != local_lines.end(); ++index)
```
tjhei, line 573: what is the difference between this and merge()? Maybe leave a comment.

source/mesh_deformation/interface.cc
```diff
@@ -585,6 +719,123 @@ namespace aspect
...
+      // carry out the solution
+      FEValuesExtractors::Vector extract_vel(0);
+
+      LinearAlgebra::Vector rhs, deformation_solution;
+      rhs.reinit(mesh_locally_owned, sim.mpi_communicator);
+      deformation_solution.reinit(mesh_locally_owned, sim.mpi_communicator);
+
+      typename DoFHandler<dim>::active_cell_iterator cell = mesh_deformation_dof_handler.begin_active(),
+                                                     endc= mesh_deformation_dof_handler.end();
+      for (; cell!=endc; ++cell)
```
tjhei, line 784: ``for (const auto& cell : mesh_deformation_dof_handler.active_cell_iterators())``

source/mesh_deformation/interface.cc
gassmoeller: I specifically chose not to in this case, because the type is so complicated that an `auto` is very confusing (it is near impossible to tell what the type is without going digging into the return type of the geometry model function). This is an aspect of the `auto` keyword that I think is not optimal.

gassmoeller:
I wanted to address some of your comments in a follow-up PR, but you are right the current situation is just too confusing. I added the refactoring to this PR instead. Now there are three mutually exclusive sets `prescribed_mesh_deformation_boundary_indicators`, `tangential_mesh_deformation_boundary_indicators`, and `zero_mesh_deformation_boundary_indicators`, which are set in `parse_parameters` and can be used by all other functions without further modifications. Additionally `free_surface_boundary_indicators` is that subset of `prescribed_mesh_deformation_boundary_indicators` which has an assigned free surface plugin. I addressed all other comments as well.

gassmoeller:
I rebased and squashed everything into two commits. The first contains all changes up to before @tjhei 's review. The second one contains the renaming and documentation changes. @anne-glerum: can you take a look at the second commit? The changes to `parse_parameters` look confusing because I had to reorder things, but the purpose becomes clear if you start from the bottom and look at how much easier the `make_constraints` and `make_initial_constraints` functions are now.

gassmoeller:
Ok, it turns out that the new way to write things also exposed something like a bug or at least unexpected behavior in the main branch. If you had no Stokes velocity boundary condition on a boundary that was not declared a free surface boundary there would be no constraints on the mesh movement as well, i.e. vertices would be moved in whatever way laterally or vertically that would satisfy the Laplace equation. This was happening in the topo_chunk test, which (unintentionally) has material flowing out at the side boundaries. The new default is to fix the mesh at these boundaries, which seems more reasonable to me.

anne-glerum:
Hi @gassmoeller, thanks for this extension of the interface and the simplification of the boundary conditions logic! I left a few comments/questions, otherwise ready!

gassmoeller:
Thanks for the review! I addressed all points. About your questions:

> For the cosine topography prescribed, shouldn't the max topography be close to 2e4 m? Max topo in the periodic test is 2e4. Because the left and right boundaries are set to zero Stokes velocity, so no mesh deformation of the left and right boundary, the corner points are not allowed to move, but internal nodes should have positive topography as well right?

That was an artifact of the test resolution. I had only 4 cells and the vertices were at the boundary (h=0 because of BC), and at  x/4 and 3/4 x (h=0 because of cos(2 pi x)) and at x/2 (h=-2e4). I increased the resolution to 8 cells and now have positive topography as well.

> This function takes 1 boundary id, but the compute_velocity_constraints_on_boundary function takes a set of boundary ids, which in practice is called with just 1 boundary in the set. Should we change compute_velocity_constraints_on_boundary to also just take 1 boundary id instead of a set (in a separate PR)?

Yes I should probably update that interface at some point (it also has a `const` input parameter after a non-const output parameter), but I would like to postpone that.

Should hopefully be good to go now.

source/mesh_deformation/interface.cc
```diff
@@ -46,11 +46,36 @@ namespace aspect
...
     void
     Interface<dim>::update ()
     {}
 
+
+
+    template <int dim>
+    Tensor<1,dim>
+    Interface<dim>::
+    compute_initial_deformation_on_boundary(const types::boundary_id /*boundary_indicator*/,
```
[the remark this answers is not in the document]
tjhei: I don't think so, as this is called for an individual vertex on a boundary face, while the other function is called once.

include/aspect/mesh_deformation/interface.h
```diff
@@ -384,32 +410,44 @@ namespace aspect
...
          */
-        std::set<types::boundary_id> tangential_mesh_boundary_indicators;
+        std::map<types::boundary_id,std::vector<std::unique_ptr<Interface<dim> > > > mesh_deformation_objects;
 
         /**
          * Map from boundary id to a vector of names representing
          * mesh deformation objects.
          */
-        std::map<types::boundary_id, std::vector<std::string> > mesh_deformation_boundary_indicators_map;
+        std::map<types::boundary_id, std::vector<std::string> > mesh_deformation_boundary_object_names;
```
tjhei, line 422: Why is this not called mesh_deformation_object_names as the variable above?

source/mesh_deformation/interface.cc
```diff
@@ -338,89 +371,152 @@ namespace aspect
...
-      // Zero out the displacement for the prescribed velocity boundaries
-      // if the boundary is not in the set of tangential mesh boundaries and not in the set of mesh deformation boundary indicators
-      for (const auto &p : sim.boundary_velocity_manager.get_active_boundary_velocity_names())
+      this->get_signals().pre_compute_no_normal_flux_constraints(sim.triangulation);
+      // Make the no flux boundary constraints
+      VectorTools::compute_no_normal_flux_constraints (mesh_deformation_dof_handler,
+                                                       /* first_vector_component= */
+                                                       0,
+                                                       tangential_mesh_deformation_boundary_indicators,
+                                                       mesh_velocity_constraints, this->get_mapping());
```
tjhei, line 399:
```suggestion
                                                       mesh_velocity_constraints, 
                                                       this->get_mapping());
```

source/mesh_deformation/interface.cc
```diff
@@ -338,89 +371,152 @@ namespace aspect
...
+              {
+                return deformation_object->compute_initial_deformation_on_boundary(boundary_id_and_deformation_objects.first, x);
+              });
+
+              VectorTools::interpolate_boundary_values (this->get_mapping(),
+                                                        mesh_deformation_dof_handler,
+                                                        boundary_id_and_deformation_objects.first,
+                                                        vel,
+                                                        current_plugin_constraints,
+                                                        ComponentMask());
```
tjhei, line 519: I think you can remove the last argument as this is the default value.

tjhei (commented):
Just a few comments.

include/aspect/mesh_deformation/interface.h
gassmoeller: Copy-paste error probably. fixed now.

gassmoeller:
Addressed the comments.

tjhei (approved):
I am okay with the current state. Thanks, Rene.

source/mesh_deformation/interface.cc
```diff
@@ -664,6 +877,11 @@ namespace aspect
 
       // We can safely close this now
       mesh_vertex_constraints.close();
+
+      // if we are just starting, we need to initialize the mesh displacement vector.
+      if (this->simulator_is_past_initialization() == false ||
```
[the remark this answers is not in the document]
anne-glerum: Hi @gassmoeller, I'm just curious what your thoughts are on this :)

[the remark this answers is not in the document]
gassmoeller: I think we need to run this several times in case of initial adaptive refinement at the boundary (each time the deformation is initialized with 0), because we want the final deformation for the final resolution to conform as closely as possible to the prescribed initial topography. Do you think there are situations where this would be wrong?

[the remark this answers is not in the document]
anne-glerum: No that makes total sense. Don't know why I was doubting, maybe I thought that it could in principle be done once at the end of initial mesh refinement unless you wanted to refine based on topography. Anyway, this is good to go from my side as well!

gassmoeller:
I updated the tests (because of #3855) and since Timo and Anne agreed I am going to merge.

PR #3723, opened 2020-08-07
add Peierls creep rheology

naliboff:
This PR add a rheology module for a Peierls creep flow law, which is used in the `visco plastic` material model. The current formulation is an approximate solution for the Peierls creep rheology following a derivation from @mibillen.

While the current implementation returns the correct viscosity, a number of things need to be done before a full review:

1. Rechecking the implemented formulation against the analytical solution with @mibillen 
2. Add solution files for the provided test
3. Adding an option to use the full generalized Peierls creep flow law, which uses stress, rather than strain rate, as input. This would require a proper decomposition of the stress between different flow law mechanisms, as is done in `diffusion_dislocation`.
4. Add in the option to account different phases, following #3598.
 
I've opened this PR so @mibillen, @danieldouglas92, and others can take a look, add comments and make suggestions.

* [X] I have followed the [instructions for indenting my code](../blob/master/CONTRIBUTING.md#making-aspect-better).
* [X] I have tested my new feature locally to ensure it is correct.
* [x] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/master/tests/) directory.
* [x] I have added a changelog entry in the [doc/modules/changes](../blob/master/doc/modules/changes) directory that will inform other users of my change.

PR #3755, opened 2020-08-11
Compute edot ii in a function

EstherHeck:
I moved the computation of the current_edot_ii from visco_plastic into function inside utilities.

### For all pull requests:

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1258,8 +1258,9 @@ namespace aspect
                                     const double ref_strain_rate,
                                     const double min_strain_rate,
                                     const SymmetricTensor<2,dim> &strain_rate,
-                                    bool use_elasticity,
-                                    bool use_reference_strainrate);
+                                    const std::vector<double> &elastic_shear_moduli,
+                                    const bool use_elasticity,
+                                    const bool use_reference_strainrate);
 
     /**
      * Calculate the square root of the second moment invariant for the deviatoric
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1249,32 +1249,6 @@ namespace aspect
                               const double SPD_safety_factor);
 
     /**
-     * Computes the current_edot_ii and current_stress, with the option for
-     * viscous and viscoelastic contribution.
-     */
-    template <int dim>
-    double compute_current_edot_ii (const unsigned int j,  // the volume fraction
-                                    const std::vector<double> &composition,
-                                    const double ref_strain_rate,
-                                    const double min_strain_rate,
-                                    const SymmetricTensor<2,dim> &strain_rate,
-                                    const std::vector<double> &elastic_shear_moduli,
-                                    const bool use_elasticity,
-                                    const bool use_reference_strainrate,
-                                    const double dte);
-
-    /**
-     * Calculate the square root of the second moment invariant for the deviatoric
-     * strain rate tensor, including viscoelastic stresses.
-     */
-    template <int dim>
-    double
-    calculate_viscoelastic_strain_rate (const SymmetricTensor<2,dim> &strain_rate,
-                                        const SymmetricTensor<2,dim> &stress,
-                                        const double shear_modulus,
-                                        const double dte);
-
-    /**
      * Converts an array of size dim to a Point of size dim.
      */
     template <int dim>
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1249,6 +1249,19 @@ namespace aspect
                               const double SPD_safety_factor);
 
     /**
+     * Computes the current_edot_ii and current_stress, with the option for
+     * viscous and viscoelastic contribution.
+     */
+    template <int dim>
+    double compute_current_edot_ii (const unsigned int j,  // the volume fraction
+                                                       const std::vector<double> &composition,
+                                                       const double ref_strain_rate,
+                                                       const double min_strain_rate,
+                                                       const double strain_rate,
+                                                       bool use_elasticity,
+                                                       bool use_reference_strainrate);
+
+    /**
      * Converts an array of size dim to a Point of size dim.
      */
     template <int dim>
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1254,12 +1254,21 @@ namespace aspect
      */
     template <int dim>
     double compute_current_edot_ii (const unsigned int j,  // the volume fraction
-                                                       const std::vector<double> &composition,
-                                                       const double ref_strain_rate,
-                                                       const double min_strain_rate,
-                                                       const SymmetricTensor<2,dim> &strain_rate,
-                                                       bool use_elasticity,
-                                                       bool use_reference_strainrate);
+                                    const std::vector<double> &composition,
+                                    const double ref_strain_rate,
+                                    const double min_strain_rate,
+                                    const SymmetricTensor<2,dim> &strain_rate,
+                                    bool use_elasticity,
+                                    bool use_reference_strainrate);
+
+    /**
+     * Calculate the square root of the second moment invariant for the deviatoric
+     * strain rate tensor, including viscoelastic stresses.
+     */
+    double
+    calculate_viscoelastic_strain_rate (const SymmetricTensor<2,dim> &strain_rate,
+                                        const SymmetricTensor<2,dim> &stress,
+                                        const double shear_modulus) const;
 
     /**
      * Converts an array of size dim to a Point of size dim.
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -1271,7 +1271,8 @@ namespace aspect
     double
     calculate_viscoelastic_strain_rate (const SymmetricTensor<2,dim> &strain_rate,
                                         const SymmetricTensor<2,dim> &stress,
-                                        const double shear_modulus);
+                                        const double shear_modulus,
+                                        const double dte);
 
     /**
      * Converts an array of size dim to a Point of size dim.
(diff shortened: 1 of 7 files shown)


EstherHeck:
there is an error left, because the new function in utilities cannot access elastic_rheology. How can I resolve that? Including elastic_rheology in it? Or is there a simpler way to access it?

naliboff:
@EstherHeck - thanks working on this! I see the issue with accessing the elastic rheology module. I'm not sure if it makes sense to do use the elastic.cc function for calculating the viscoelastic strain-rate and we may want to rewrite it in this utility. Perhaps the utility could replace the function in `elastic.cc`? The issue of accessing specific elastic variables (shear modulus, etc), can be circumvented by accessing these values in visco_plastic and passing them to function in material utilities. @gassmoeller @jdannberg @anne-glerum - Thoughts on the above approach?

include/aspect/utilities.h
```diff
@@ -1232,6 +1232,19 @@ namespace aspect
                               const SymmetricTensor<2,dim> &dviscosities_dstrain_rate,
                               const double SPD_safety_factor);
 
+    /**
+     * Computes the current_edot_ii and current_stress, with the option for
+     * viscous and viscoelastic contribution.
+     */
+    template <int dim>
+    double compute_current_edot_ii (const unsigned int j,  // the volume fraction
```
bangerth, line 1240: I'm currently reading "Code Complete" by Steve McConnell with all of my students, which has a lot of good advice on function and argument names. So I'll comment on this function.

For `j`, if you need to document the name of an argument, the name is probably poorly chosen :-) If the argument really denoted the volume fraction, then `volume_fraction` would be a better argument name. But that can't be true, because it is an integer (and the volume fraction should be a `double`), so it must be something different -- I don't know what it is, but it would be useful to choose a more descriptive name than `j` :-)

include/aspect/utilities.h
```diff
@@ -1232,6 +1232,19 @@ namespace aspect
...
+                                                       const std::vector<double> &composition,
```
bangerth, line 1241: It's an array, so a plural variable name would be more appropriate. Maybe
```suggestion
                                                       const std::vector<double> &compositions,
```
?

include/aspect/utilities.h
```diff
@@ -1232,6 +1232,19 @@ namespace aspect
...
+                                                       const double ref_strain_rate,
+                                                       const double min_strain_rate,
+                                                       const SymmetricTensor<2,dim> &strain_rate,
+                                                       bool use_elasticity,
+                                                       bool use_reference_strainrate);
```
bangerth, line 1246: These two are input arguments, which you can indicate by marking them as `const` just like all of the other arguments.

include/aspect/utilities.h
EstherHeck: yes, I see my comment about j was misleading. So j comes from this for loop: 
which is way I made the little comment for myself that it is not from another loop through lets say n_evaluation_points

EstherHeck:
I got around my problem with functions from elastic_rheology as you suggested above. 
Now another compile error occurred, which I don't know how to fix: 
 
image

include/aspect/utilities.h
```diff
@@ -1232,6 +1232,29 @@ namespace aspect
...
+                                    const SymmetricTensor<2,dim> &strain_rate,
+                                    const std::vector<double> &elastic_shear_moduli,
+                                    const bool use_elasticity,
+                                    const bool use_reference_strainrate);
+
+    /**
+     * Calculate the square root of the second moment invariant for the deviatoric
+     * strain rate tensor, including viscoelastic stresses.
+     */
+    double
```
bangerth, line 1253: You are missing the `template <int dim>` before this line:
```suggestion
    template <int dim>
    double
```
This should make the error go away.

source/utilities.cc
```diff
@@ -3184,6 +3184,75 @@ namespace aspect
...
+          // whereas the combined viscoelastic + viscous stresses already include the
+          // 2x factor (see computation of edot inside elastic_rheology).
+          current_edot_ii /= 2.;
+        }
+
+      return current_edot_ii;
+    }
+
+    template <int dim>
+    double calculate_viscoelastic_strain_rate(const SymmetricTensor<2,dim> &strain_rate,
```
bangerth, line 3246: You've got it correct here.

include/aspect/utilities.h
EstherHeck: thanks!

EstherHeck:
I got this now as a message after make: 
image
don't know what to make out of that. It has linked an executable, but also got an error...

bangerth:
Turns out that I sent this last night to one of my students:


What's happening is that you need to put "explicit instantiations" into your .cc files.

https://stackoverflow.com/questions/2351148/explicit-template-instantiation-when-is-it-used

The problem is that in your .h file, you only say that these functions (e.g., the constructor you reference above) *exist*. In the .cc file, you provide a template, i.e., a way to generate that function *if you knew what the dim template argument actually is*, but you never tell the compiler what value for dim you actually want to use. Then, in main.cc, you do say so, but in that file you only #include the .h file, and so the compiler has no access to the template definition in the .cc file. So the compiler sees that function template when compiling your .cc file but doesn't know what to do with it, and the error message above is the consequence.

EstherHeck:
Hi, I am trying to do a rebase to the newest master version, but somehow can't make it work. I get the following message: 
image
which is weird as these are files I did not touch for this PR. I also tried to "add" them, but that did not help. Can someone help me out?

naliboff:
@EstherHeck -yes, I had this issue as well in another PR (#3723). What I believe ended up working was the following
1. checkout your master branch and pull the changes from upstream:` git pull upstream master`
2. checkout your branch with the conflict: `git checkout Compute_edot_ii_in_a_function`
3. rebase with respect to your local master: `git rebase master`

Can you give this a shot and see if it works?

For some reason, doing an interactive rebase against the upstream master did not work: `git rebase -i upstream/master`

Also deleted the aforementioned files/folders, resetting all changes, and rebasing against the upstream master did not work either.

I confess I don't fully understand why the procedure I used previously worked. Normally most merge conflicts can be resolved with an interactive rebase, but this was a bit more complicated to do files/folders being moved. Others can likely explain in much better detail :-)

EstherHeck:
Hi John, thanks for your answer! Yes, that did actually worked, in that the rebase finished. Earlier on I did steps 1 and 2 as described, but then tried an interactive rebase. Now without the "-i" it worked!
However, something must have changed in master, so that I get new compile errors now. Will work on it on monday

bangerth:
Post the error messages you get if you need help!

EstherHeck:
I just realized, that there is another utilities.cc file inside the material-model folder. Should I have put the function into that one? At the moment it is in the utilities.cc in the source folder.

naliboff:
> I just realized, that there is another utilities.cc file inside the material-model folder. Should I have put the function into that one? At the moment it is in the utilities.cc in the source folder.

@EstherHeck - I'm afraid so, `material_model/utilities.cc` is where this should probably be located. Sorry not catching that sooner!

EstherHeck:
ok, done

EstherHeck:
now, everything should be clean and in the right file

gassmoeller:
Hi Esther, sorry that I did not take a look at this during the hackathon already, but I have a few questions before we move forward with this. 
What is the purpose of moving this function outside of the material model? Is it to unify the computation of the strain rate invariant between different places in the code? The reason I ask is that at the moment it only replaces exactly the computation inside the visco plastic material model, and it contains quite a number of features that only this material model can know.
In general functions in the utilities namespace should be useful for a number of different places (otherwise they may as well stay inside the one place where they are used). But if I would want to call this function from some postprocessor to compute the strain rate, I would need to figure out things like the elastic time step, the reference strain rate, the minimum strain rate, or the elasticity module, which are only known inside visco plastic. That does not mean we should not have such a function, but maybe it needs to work differently and if you can explain what you want to do with this function we can figure out how to make it work that way.

EstherHeck:
HI @gassmoeller, yes, that is true, at the moment this is only used once in the code. However, for the rate-and-state friction I need it in several places, so it was more convenient to have it in a function and it also makes visco_plastic more readable I think. About where to put that function is a good question. @naliboff  suggested utilities for that. If you have a better place for it, I am happy to move it there.

EstherHeck:
for the friction things, I call this function from inside evaluate, compute_isostrain_viscosities and additionalPlasticOutputs

gassmoeller:
Ah, I see. Will the rate-and-state friction become a new separate rheology plugin, or be integrated into an existing one?

EstherHeck:
it is a new separate rheology plugin

gassmoeller:
Do you have a version of that plugin somewhere I could take a look at it? (a branch or a WIP PR, or just via email?) I feel like there should be a simpler way than to create these general utility functions that need a lot of internal information from the visco plastic material model or the elasticity rheology. For example: The rate-and-state-friction would likely be used by the visco plastic material model. Could the visco plastic material model compute the correct strain rate (just as it does in the old version of the code) and hand over the already computed strain rate to the rate and state friction module?
I see the utility of having a general function for computing the strain rate invariant as `std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i]))))` and it's compressible complications, because this is used all over the place, I am just not sure about the minimum strain rate, reference strain rate, and elasticity complications and would like to check if there is another way, before we commit to having such a utility function.

bangerth:
How do we want to move forward with this? 

I'm with @gassmoeller on the fact that it doesn't make sense to me to introduce functions in general namespaces that take a lot of internal information specific to what material models do. I think that this is more easily handled inside those material models themselves. I do believe that 
```
std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i]))))
```
could usefully be encapsulated into a utility function, though.

EstherHeck:
Hi @bangerth , good question. Personnally, I found another way to do the things I want to do without this function. So from my side we could just as well close this PR. Unless of course this appears useful in general

PR #3763, opened 2020-08-11
TimeStepping: introduce plugins

tjhei:
- introduce plugin architecture for time stepping
- move convection/conduction into their own plugins

Part 2 of N for #3549

### DIFF
--- a/include/aspect/time_stepping/conduction_time_step.h
+++ b/include/aspect/time_stepping/conduction_time_step.h
@@ -19,7 +19,7 @@
 */
 
 
-#ifndef _aspect_time_stepping_convection_time_step_h
+#ifndef _aspect_time_stepping_conduction_time_step_h
 #define _aspect_time_stepping_conduction_time_step_h
 
 #include <aspect/time_stepping/interface.h>
@@ -32,8 +32,8 @@ namespace aspect
     using namespace dealii;
 
     /**
-     * A class that implements a heating model based on a functional
-     * description provided in the input file.
+     * Compute the conduction time step based on the current solution and
+     * return this as the time step.
      *
      * @ingroup TimeStepping
      */
@@ -48,13 +48,12 @@ namespace aspect
 
 
         /**
-         * The main execute() function.
+         * @copydoc aspect::TimeStepping::Interface<dim>::execute()
          */
         virtual
         std::pair<Reaction, double>
         execute() override;
 
-      private:
     };
   }
 }
--- a/include/aspect/time_stepping/conduction_time_step.h
+++ b/include/aspect/time_stepping/conduction_time_step.h
@@ -1,0 +1,63 @@
+/*
+  Copyright (C) 2018 - 2020 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_time_stepping_convection_time_step_h
+#define _aspect_time_stepping_conduction_time_step_h
+
+#include <aspect/time_stepping/interface.h>
+
+
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    using namespace dealii;
+
+    /**
+     * A class that implements a heating model based on a functional
+     * description provided in the input file.
+     *
+     * @ingroup TimeStepping
+     */
+    template <int dim>
+    class ConductionTimeStep : public Interface<dim>, public SimulatorAccess<dim>
+    {
+      public:
+        /**
+         * Constructor.
+         */
+        ConductionTimeStep () = default;
+
+
+        /**
+         * The main execute() function.
+         */
+        virtual
+        std::pair<Reaction, double>
+        execute() override;
+
+      private:
+    };
+  }
+}
+
+
+#endif
--- a/include/aspect/time_stepping/convection_time_step.h
+++ b/include/aspect/time_stepping/convection_time_step.h
@@ -1,0 +1,62 @@
+/*
+  Copyright (C) 2014 - 2020 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_time_stepping_convection_time_step_h
+#define _aspect_time_stepping_convection_time_step_h
+
+#include <aspect/time_stepping/interface.h>
+
+
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    using namespace dealii;
+
+    /**
+     *
+     *
+     * @ingroup TimeStepping
+     */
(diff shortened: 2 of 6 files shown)


MFraters:
Is this ready for review? I still see some todo's in the code. I can do a first review round, but since I was not involved in the discussion on the timestepping it would be useful if someone who was involved also takes a look.

tjhei:
Good point about the TODO. Removed.

This is ready to be reviewed/merged.

include/aspect/time_stepping/convection_time_step.h
```diff
@@ -0,0 +1,62 @@
...
+
+
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    using namespace dealii;
+
+    /**
+     *
```
MFraters, line 35: Could you add some documentation here of what this class is doing?

include/aspect/time_stepping/convection_time_step.h
```diff
@@ -0,0 +1,62 @@
...
+    {
+      public:
+        /**
+         * Constructor.
+         */
+        ConvectionTimeStep () = default;
+
+
+        /**
+         * The main execute() function.
```
MFraters, line 50: Personally I would find it useful to have here similair (or the same) information which you have in the interface class for this function.

source/time_stepping/conduction_time_step.cc
```diff
@@ -0,0 +1,128 @@
...
+          {
+            fe_values.reinit (cell);
+            in.reinit(fe_values,
+                      cell,
+                      this->introspection(),
+                      this->get_solution());
+
+            this->get_material_model().evaluate(in, out);
+
+            if (this->get_parameters().formulation_temperature_equation
```
MFraters, line 64: why did you decide to go for if and else statements and not a switch and case system? The switch case systems seems to me to describe better what you are doing here, but that might be a personal preference.

source/time_stepping/conduction_time_step.cc
```diff
@@ -0,0 +1,128 @@
...
+// explicit instantiations
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    ASPECT_REGISTER_TIME_STEPPING_MODEL(ConductionTimeStep,
+                                        "conduction time step",
+                                        "This model computes the conduction time step as the minimum "
+                                        "over all cells of $ CFL h^2 \\cdot \\rho C_p / k$, "
+                                        "where k is the thermal conductivity.")
```
MFraters, line 126: It might be good to document the type of reaction (advance) which is returned by this model.

source/time_stepping/convection_time_step.cc
```diff
@@ -0,0 +1,121 @@
...
+// explicit instantiations
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    ASPECT_REGISTER_TIME_STEPPING_MODEL(ConvectionTimeStep,
+                                        "convection time step",
+                                        "This model computes the convection time step as the minimum "
+                                        "over all cells of $ CFL h^2 \\cdot \\rho C_p / k$, "
+                                        "where k is the thermal conductivity.")
```
MFraters, line 119: same here (returns reaction advance).

source/time_stepping/interface.cc
```diff
@@ -26,137 +26,73 @@ namespace aspect
...
+    }
 
-      double min_convection_timestep = std::numeric_limits<double>::max();
-      double min_conduction_timestep = std::numeric_limits<double>::max();
 
-      if (max_global_speed_over_meshsize != 0.0)
-        min_convection_timestep = this->get_parameters().CFL_number / (this->get_parameters().temperature_degree * max_global_speed_over_meshsize);
 
-      if (this->get_parameters().use_conduction_timestep)
-        min_conduction_timestep = - Utilities::MPI::max (-min_local_conduction_timestep, this->get_mpi_communicator());
```
MFraters, line 151: You cancelled out the sign, which honestly make a lot more sense to me, but I am wondering why a minus sign was used here in the first place.

include/aspect/time_stepping/conduction_time_step.h
```diff
@@ -0,0 +1,63 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_time_stepping_convection_time_step_h
+#define _aspect_time_stepping_conduction_time_step_h
```
gassmoeller, line 23: conduction

include/aspect/time_stepping/conduction_time_step.h
```diff
@@ -0,0 +1,63 @@
...
+
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    using namespace dealii;
+
+    /**
+     * A class that implements a heating model based on a functional
+     * description provided in the input file.
```
gassmoeller, line 36: update

include/aspect/time_stepping/conduction_time_step.h
```diff
@@ -0,0 +1,63 @@
...
+
+
+        /**
+         * The main execute() function.
+         */
+        virtual
+        std::pair<Reaction, double>
+        execute() override;
+
+      private:
```
gassmoeller, line 57: not necessary

include/aspect/time_stepping/convection_time_step.h
gassmoeller, line 35: Documentation.

source/time_stepping/conduction_time_step.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#include <aspect/global.h>
+#include <aspect/time_stepping/conduction_time_step.h>
+#include <aspect/simulator.h>
```
gassmoeller, line 24: why do you need simulator.h?

source/time_stepping/convection_time_step.cc
```diff
@@ -0,0 +1,121 @@
...
+      std::vector<Tensor<1,dim> > velocity_values(n_q_points);
+      std::vector<Tensor<1,dim> > fluid_velocity_values(n_q_points);
+      std::vector<std::vector<double> > composition_values (this->introspection().n_compositional_fields,std::vector<double> (n_q_points));
+
+      double max_local_speed_over_meshsize = 0;
+
+      MaterialModel::MaterialModelInputs<dim> in(n_q_points,
+                                                 this->introspection().n_compositional_fields);
+      MaterialModel::MaterialModelOutputs<dim> out(n_q_points,
+                                                   this->introspection().n_compositional_fields);
```
gassmoeller, line 59: I dont think you use these at the moment?

source/time_stepping/convection_time_step.cc
```diff
@@ -0,0 +1,121 @@
...
+                                ?
+                                update_quadrature_points
+                                :
+                                update_default));
+
+      const unsigned int n_q_points = quadrature_formula.size();
+
+      std::vector<Tensor<1,dim> > velocity_values(n_q_points);
+      std::vector<Tensor<1,dim> > fluid_velocity_values(n_q_points);
+      std::vector<std::vector<double> > composition_values (this->introspection().n_compositional_fields,std::vector<double> (n_q_points));
```
gassmoeller, line 52: not used?

source/time_stepping/convection_time_step.cc
```diff
@@ -0,0 +1,121 @@
...
+                                               this->get_parameters().stokes_velocity_degree);
+
+      FEValues<dim> fe_values (this->get_mapping(),
+                               this->get_fe(),
+                               quadrature_formula,
+                               update_values |
+                               update_gradients |
+                               ((this->get_parameters().include_melt_transport)
+                                ?
+                                update_quadrature_points
```
gassmoeller, line 44: I dont think you need the quadrature points

MFraters:
@gassmoeller, well you beat me with the review with 1 minute difference :P

gassmoeller:
Yeah, I was afraid of that :-) ... but I wanted to take a look as well anyway.

source/time_stepping/interface.cc
tjhei: We computed ``x = min(x[i])`` as ``x = - max(-x[i])`` because min() did not exist (but that was a long time ago).

tjhei:
> well you beat me with the review with 1 minute difference

But you found different things. More reviews are always good. :-)

source/time_stepping/conduction_time_step.cc
tjhei: No specific reason (this code has just been moved around), but a quick search says that we do it exactly in the same way in all other 5+ places in the code.
A switch is probably a few more lines, but I don't mind making that change (but better in a a separate PR?).

tjhei:
I think I addressed the comment. Thanks, guys.

source/time_stepping/interface.cc
MFraters: ah, wow I never though a max function would exists without a min function :s

PR #3765, opened 2020-08-11
Incorporating strain healing in strain dependent rheology

erinheilman:

Adding calculation of strain healing into the strain dependent rheology.

### For all pull requests:


include/aspect/material_model/rheology/strain_dependent.h
```diff
@@ -54,6 +54,17 @@ namespace aspect
         plastic_weakening_with_plastic_strain_and_viscous_weakening_with_viscous_strain,
         viscous_weakening_with_viscous_strain_only
       };
+      /**
+       * Enumeration for selecting which type of healing mechanism to use.
+       * For none, no strain healing occurs.
```
naliboff, line 59:
```suggestion
       * For no_healing, no strain healing occurs.
```

include/aspect/material_model/rheology/strain_dependent.h
```diff
@@ -54,6 +54,17 @@ namespace aspect
...
+       * Otherwise, the material is healed based on a temperature dependent
```
naliboff, line 60:
```suggestion
       * Otherwise, the strain is healed (reduced) as a function of temperature,
       * a user defined healing time scale and additional parameters.
```

include/aspect/material_model/rheology/strain_dependent.h
```diff
@@ -54,6 +54,17 @@ namespace aspect
...
+       * strain healing calculation
```
naliboff, line 61:
```suggestion
       * Future models could consider strain healing formulations that are a function
       * of time, deformation rate, or other parameters.
```

include/aspect/material_model/rheology/strain_dependent.h
```diff
@@ -83,6 +94,14 @@ namespace aspect
           compute_strain_weakening_factors(const unsigned int j,
                                            const std::vector<double> &composition) const;
 
+          /**
+           * A function that computes by how much the strain is reduced
```
naliboff, line 98:
```suggestion
           * A function that computes by strain healing (reduction).
```

include/aspect/material_model/rheology/strain_dependent.h
```diff
@@ -178,6 +204,15 @@ namespace aspect
            */
           std::vector<double> viscous_strain_weakening_factors;
 
+          /**
+           * The factor of healing time used in the strain healing calculation.
```
naliboff, line 208: The wording here seems off. How about the "The healing time scale used in the temperature dependent" strain healing model.

include/aspect/material_model/rheology/strain_dependent.h
```diff
@@ -178,6 +204,15 @@ namespace aspect
...
+           */ 
+          double strain_healing_recovery_timescale;
+	
+          /**
+           * A prefactor of viscosity used in the strain healing calculation.
```
naliboff, line 213: Does this quantity actually have units of Pa s?

include/aspect/material_model/rheology/strain_dependent.h
naliboff: Given that the temperature dependent model effectively follows the Frank Kamenetskii viscosity formulation, I'm wondering if we should rename `temperature_dependent` to something like `frank_kamenetskii_temperature_dependent`?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -147,6 +147,24 @@ namespace aspect
                            "for background material and compositional fields, "
                            "for a total of N+1 values, where N is the number of compositional fields. "
                            "If only one value is given, then all use the same value.  Units: None.");
+
+        prm.declare_entry ("Strain healing mechanism", "no healing",
+                           Patterns::Selection("none|temperature dependent"),
+                           "Whether to apply strain healing to viscosity, and if yes, which method to "
```
naliboff, line 153: The method is applied to strain terms for plastic yielding and not just viscosity, yes?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -147,6 +147,24 @@ namespace aspect
...
+                           "use. The following methods are available:"
+                           "\n\n"
+                           "\\item ``no healing'': No strain healing is applied. "
+                           "\n\n"
+                           "\\item ``temperature dependent'': Purely temperature dependent "
+                           "strain healing applied to viscosity from Fuchs and Becker, 2019");
```
naliboff, line 159: See above?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -283,6 +301,24 @@ namespace aspect
...
+          healing_mechanism = temperature_dependent;
+        else
+          AssertThrow(false, ExcMessage("Not a valid Strain healing mechanism!"));
+        
+        // Currently this functionality only works in field composition
+        if (healing_mechanism != no_healing)
+        {
+          AssertThrow(this->introspection().compositional_name_exists("particles"),
+                          ExcMessage("This healing mechanism currently does not work with particles!"));
+        }
```
naliboff: I think the line above is checking wether there is compositional field named `particles` and not whether the particles are being used as the "compositional field method". We can work on this after addressing other items.

source/material_model/rheology/strain_dependent.cc
```diff
@@ -360,6 +396,37 @@ namespace aspect
...
+                               const double e_ii,
+                               const unsigned int j) const
+      {
+        const double reference_temperature = this->get_adiabatic_surface_temperature();
+        double healed_strain = e_ii; // e'_II  * dt
+	      /** - strain * A * exp(-E * 0.5 * (1.0 - T/T_0))) * dt
+        * original equation from plugin for reference
+        * strain_add -= in.composition[i][out_c_ind] * strain_healing_A *
+		    *std::exp(-strain_healing_E*(0.5- in.temperature[i]/reference_T/2))*this->get_timestep();
+		    */
```
naliboff: Go ahead and remove the above lines (comments w/ equations).

source/material_model/rheology/strain_dependent.cc
```diff
@@ -415,7 +482,9 @@ namespace aspect
         if  (this->simulator_is_past_initialization() && this->get_timestep_number() > 0 && in.requests_property(MaterialProperties::reaction_terms))
           {
             const double edot_ii = std::max(sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),min_strain_rate);
-            const double e_ii = edot_ii*this->get_timestep();
+            double e_ii = edot_ii*this->get_timestep(); //changed from const to double so can be modified
```
naliboff, line 484:
```suggestion
            double e_ii = edot_ii*this->get_timestep();
```

source/material_model/rheology/strain_dependent.cc
```diff
@@ -415,7 +482,9 @@ namespace aspect
...
+            if (healing_mechanism != no_healing)
+              e_ii = calculate_strain_healing(in,e_ii,i); //not sure if this is how you call correctly
```
naliboff, line 486: Is this line producing the correct results?

include/aspect/material_model/rheology/strain_dependent.h
```diff
@@ -54,6 +54,19 @@ namespace aspect
         plastic_weakening_with_plastic_strain_and_viscous_weakening_with_viscous_strain,
         viscous_weakening_with_viscous_strain_only
       };
+      /**
+       * Enumeration for selecting which type of healing mechanism to use.
+       * For no healing, no strain healing occurs.
```
naliboff, line 59:
```suggestion
       * For the case no healing, no strain healing occurs.
```

include/aspect/material_model/rheology/strain_dependent.h
```diff
@@ -178,6 +206,15 @@ namespace aspect
            */
           std::vector<double> viscous_strain_weakening_factors;
 
+          /**
+           * The healing time scale used in the temperature dependent strain healing model.
```
naliboff, line 210: Reading further down, this variable has units of 1/s. Is this correct?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -147,6 +147,25 @@ namespace aspect
...
+        prm.declare_entry ("Strain healing mechanism", "no healing",
+                           Patterns::Selection("no healing|temperature dependent"),
+                           "Whether to apply strain healing to plastic yielding and viscosity terms, "
+                           "and if yes, which method to use. The following methods are available:"
+                           "\n\n"
+                           "\\item ``no healing'': No strain healing is applied. "
+                           "\n\n"
+                           "\\item ``temperature dependent'': Purely temperature dependent "
+                           "strain healing applied to plastic yielding and viscosity terms, "
+                           "as done in Fuchs and Becker, 2019");
```
naliboff, line 160: Can you add a line stating that the formulation closely follows that of the Frank Kemenetskii viscosity model, or something to that effect?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -283,6 +302,23 @@ namespace aspect
...
+
+        if (prm.get ("Strain healing mechanism") == "no healing")
+          healing_mechanism = no_healing;
+        else if (prm.get ("Strain healing mechanism") == "temperature dependent")
+          healing_mechanism = temperature_dependent;
+        else
+          AssertThrow(false, ExcMessage("Not a valid Strain healing mechanism!"));
+
+        // Currently this functionality only works in field composition
+        if (healing_mechanism != no_healing && this->introspection().compositional_name_exists("particles"))
```
naliboff, line 314: I'll do some testing tomorrow to see how this should be restructured.

source/material_model/rheology/strain_dependent.cc
```diff
@@ -360,6 +396,32 @@ namespace aspect
...
+          {
+            case no_healing:
+            {
+              break;
+            }
+            case temperature_dependent:
+            {
+              healed_strain -= in.composition[j][0] * strain_healing_recovery_timescale *
+                               std::exp(-strain_healing_viscosity_prefactor * 0.5 * (1.0 - in.temperature[j]/reference_temperature))
+                               * this->get_timestep();
```
naliboff, line 419: Are you getting unrealistic changes in strain when `in.temperature[j]` >> `reference_temperature`?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -415,7 +477,9 @@ namespace aspect
         if  (this->simulator_is_past_initialization() && this->get_timestep_number() > 0 && in.requests_property(MaterialProperties::reaction_terms))
           {
             const double edot_ii = std::max(sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),min_strain_rate);
-            const double e_ii = edot_ii*this->get_timestep();
+            double e_ii = edot_ii*this->get_timestep();
+            if (healing_mechanism != no_healing)
+              e_ii = calculate_strain_healing(in,e_ii,i);
```
naliboff, line 482: Add a space below here.

source/material_model/rheology/strain_dependent.cc
```diff
@@ -415,7 +477,9 @@ namespace aspect
         if  (this->simulator_is_past_initialization() && this->get_timestep_number() > 0 && in.requests_property(MaterialProperties::reaction_terms))
           {
             const double edot_ii = std::max(sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),min_strain_rate);
-            const double e_ii = edot_ii*this->get_timestep();
+            double e_ii = edot_ii*this->get_timestep();
+            if (healing_mechanism != no_healing)
```
naliboff, line 481: Can you add a short comment above this line describing what is happening? (i.e., adjusting the strain values to account for strain healing).

source/material_model/rheology/strain_dependent.cc
```diff
@@ -360,6 +396,32 @@ namespace aspect
...
       }
 
+      template <int dim>
+      double
+      StrainDependent<dim>::
+      calculate_strain_healing(const MaterialModel::MaterialModelInputs<dim> &in,
+                               const double e_ii,
+                               const unsigned int j) const
+      {
+        const double reference_temperature = this->get_adiabatic_surface_temperature();
```
naliboff, line 406: We need a check somewhere to make sure that `get_adiabatic_surface_temperature()` is not zero, as you will have divide by zero scenario below.

tests/visco_plastic_yield_plastic_viscous_strain_healing.prm
```diff
@@ -0,0 +1,139 @@
+# Global parameters
```
naliboff, line 1: Since the reference Adiabatic surface temperature is not prescribed below, you are dividing by zero in the strain healing equation.

gassmoeller:
I would also like to take a look at this, I will try to get to it tomorrow.

source/material_model/rheology/strain_dependent.cc
```diff
@@ -306,7 +308,13 @@ namespace aspect
         if (prm.get ("Strain healing mechanism") == "no healing")
           healing_mechanism = no_healing;
         else if (prm.get ("Strain healing mechanism") == "temperature dependent")
-          healing_mechanism = temperature_dependent;
+          {
+            healing_mechanism = temperature_dependent;
+            AssertThrow (this->get_adiabatic_surface_temperature() > 0.0,
+                         ExcMessage("The temperature dependent strain healing can only be used when the adiabatic "
+                                    "surface temperature (reference_temperature in equation for viscosity) "
```
naliboff, line 315:
```suggestion
                                    "surface temperature (reference_temperature in equation for strain healing) "
```

naliboff:
```suggestion
                                    "surface temperature (reference_temperature in equation for strain healing) "
```

source/material_model/rheology/strain_dependent.cc
```diff
@@ -478,8 +486,10 @@ namespace aspect
           {
             const double edot_ii = std::max(sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),min_strain_rate);
             double e_ii = edot_ii*this->get_timestep();
+            // Adjusting strain values to account for strain healing without exceeding an unreasonable range
```
naliboff, line 489: Add a space above this line

include/aspect/material_model/rheology/strain_dependent.h
naliboff: Checking back in on this. Since the units are 1/s, isn't the value more akin to a "rate" rather than a "timescale"?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -147,6 +147,27 @@ namespace aspect
...
+                           Patterns::Selection("no healing|temperature dependent"),
+                           "Whether to apply strain healing to plastic yielding and viscosity terms, "
+                           "and if yes, which method to use. The following methods are available:"
+                           "\n\n"
+                           "\\item ``no healing'': No strain healing is applied. "
+                           "\n\n"
+                           "\\item ``temperature dependent'': Purely temperature dependent "
+                           "strain healing applied to plastic yielding and viscosity terms, similar "
+                           "to the temperature-dependent Frank Kamenetskii formulation, computes "
+                           "strain healing as removing strain as a function of temperature and time"
```
naliboff, line 161:
```suggestion
                           "strain healing as removing strain as a function of temperature, time, and a 
                            user-defined healing rate and prefactor."
```

source/material_model/rheology/strain_dependent.cc
```diff
@@ -147,6 +147,27 @@ namespace aspect
...
+                           "as done in Fuchs and Becker, 2019, for mantle convection");
+
+        prm.declare_entry ("Strain healing recovery timescale", "1.e-15", Patterns::Double(0),
+                           "Recovery timescale prefactor for temperature dependent "
+                           "strain healing. Units: $1/s$");
+
+        prm.declare_entry ("Strain healing viscosity prefactor", "15.", Patterns::Double(0),
```
naliboff, line 168:
```suggestion
        prm.declare_entry ("Strain healing temperature dependent prefactor", "15.", Patterns::Double(0),
```

source/material_model/rheology/strain_dependent.cc
```diff
@@ -147,6 +147,27 @@ namespace aspect
...
+                           "\n\n"
+                           "\\item ``no healing'': No strain healing is applied. "
+                           "\n\n"
+                           "\\item ``temperature dependent'': Purely temperature dependent "
+                           "strain healing applied to plastic yielding and viscosity terms, similar "
+                           "to the temperature-dependent Frank Kamenetskii formulation, computes "
+                           "strain healing as removing strain as a function of temperature and time"
+                           "as done in Fuchs and Becker, 2019, for mantle convection");
+
+        prm.declare_entry ("Strain healing recovery timescale", "1.e-15", Patterns::Double(0),
```
naliboff, line 164:
```suggestion
        prm.declare_entry ("Strain healing temperature dependent recovery timescale", "1.e-15", Patterns::Double(0),
```

naliboff: Adding in `temperature dependent` to the above parameters to avoid confusion when additional strain healing mechanisms are added.

naliboff: Also, see my above suggestion for calling this variable a `recovery rate`.

source/material_model/rheology/strain_dependent.cc
```diff
@@ -283,6 +304,29 @@ namespace aspect
         friction_strain_weakening_factors = Utilities::possibly_extend_from_1_to_N (Utilities::string_to_double(Utilities::split_string_list(prm.get("Friction strain weakening factors"))),
                                                                                     n_fields,
                                                                                     "Friction strain weakening factors");
+
+        if (prm.get ("Strain healing mechanism") == "no healing")
+          healing_mechanism = no_healing;
+        else if (prm.get ("Strain healing mechanism") == "temperature dependent")
+          {
+            healing_mechanism = temperature_dependent;
+            AssertThrow (this->get_adiabatic_surface_temperature() > 0.0,
```
naliboff, line 313: Stylistically, it might be better to move this `AssertThrow` beneath this series of if-else if-else, but I'll let other reviewers comment on that.

source/material_model/rheology/strain_dependent.cc
```diff
@@ -360,6 +404,32 @@ namespace aspect
...
+
+        switch (healing_mechanism)
+          {
+            case no_healing:
+            {
+              break;
+            }
+            case temperature_dependent:
+            {
+              healed_strain -= in.composition[j][0] * strain_healing_recovery_timescale *
```
naliboff, line 425: There is an error in this line. `in.composition[j][0]` is just accessing the first compositional field, which may or may not be a strain value. I'm actually not sure why you need the composition term in this equation?

erinheilman: I was thinking about that, I think its just an artifact of how it was written in the older version of the code

naliboff: In that case can you remove `MaterialModel::MaterialModelInputs<dim> &in` and `j` as inputs to `calculate_strain_healing`?

erinheilman: Wouldn't I still need those to access the temperature at the point?

naliboff: yes, of course. No need to modify the interface.

erinheilman:
@gassmoeller could you give this a final review so it can be merged? Thanks

source/material_model/rheology/strain_dependent.cc
```diff
@@ -415,7 +489,12 @@ namespace aspect
         if  (this->simulator_is_past_initialization() && this->get_timestep_number() > 0 && in.requests_property(MaterialProperties::reaction_terms))
           {
             const double edot_ii = std::max(sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),min_strain_rate);
-            const double e_ii = edot_ii*this->get_timestep();
+            double e_ii = edot_ii*this->get_timestep();
+
+            // Adjusting strain values to account for strain healing without exceeding an unreasonable range
+            if (healing_mechanism != no_healing)
+              e_ii = std::min(std::max(calculate_strain_healing(in,edot_ii,i),0.0),1.e5);
```
naliboff, line 496: This line may actually break a few tests. While I agree limiting the strain to have a minimum value of 0 is a good idea, very small negative strain can arise due to advection under/over shoots. We will see if any of the tests actually contain a negative value :-).

source/material_model/rheology/strain_dependent.cc
```diff
@@ -283,6 +305,32 @@ namespace aspect
...
+
+        if (prm.get ("Strain healing mechanism") == "no healing")
+          healing_mechanism = no_healing;
+        else if (prm.get ("Strain healing mechanism") == "temperature dependent")
+          healing_mechanism = temperature_dependent;
+        else
+          AssertThrow(false, ExcMessage("Not a valid Strain healing mechanism!"));
+
+        // Currently this functionality only works in field composition
+        if (healing_mechanism != no_healing && this->introspection().compositional_name_exists("particles"))
```
naliboff, line 317: This line is still a problem. I think the check we want is whether the list of particle properties includes `viscoplastic strain invariants`, rather than `this->introspection().compositional_name_exists("particles")`.

@gassmoeller - Is there a way straightforward way to access the list of particle properties here?

naliboff: Long-term, we can avoid this by having the viscoelastic strain invariants particle property obtain values from the rheology module, rather than recalculating the values.

naliboff (commented):
@erinheilman - sorry it took me a while to get back to this! At this point, I only have one major comment left, which deals with how to enforce not using this option in combination with certain particle properties. I asked for @gassmoeller advice on this front. Other than that, I think this is in good shape. @gassmoeller wanted to do a review as well and I would like to do a second sweep through, but I think most of the code is in good shape.

source/material_model/rheology/strain_dependent.cc
egpuckett: Hi @erinheilman & @naliboff  John, Erin and I haven't met. I don't think that she is aware of the limiters that @sac-bsa (Mack) and I are working on for the various particle interpolation algorithms.  I don't know what you are working on in this PR, but it looks like you are trying to preserve the positivity of the strain, which is the sort of thing our limiters for particle interpolation are designed to do.  Our work is based on ideas that come from preventing overshoot and undershoot when using finite volume codes to model discontinuities in gas dynamics; i.e., shock waves.  Such methods go by many names, but one is positivity preserving; e.g., the density can _never_ be <= 0. Anyway Mack and I have implemented some limiters that will do this in other problems; e.g., the viscoelastic bending beam benchmark for the value of the variable 'beam', which is one of the particle properties.  John N. is working with us..

John, I am not clear what you meant by this comment

> While I agree limiting the strain to have a minimum value of 0 is a good idea, very small negative strain can arise due to advection under/over shoots.

Were you simply saying that this happens, or are you saying it is OK if this happens?  I agree with the former statement, but not with the latter.

naliboff: @egpuckett - The intent was to state that negative stress do arise, but not that it is ok or physical. The strain does not have an upper bound, but fantastic if I a limiter can be used with just a lower bound.

tjhei:
Can you update the following tests, please?
```
*** "Test visco_plastic_yield_plastic_viscous_strain_healing failed": ***
    Line <112> of file <input string>: No entry with name <Strain healing recovery timescale> was declared in the current subsection.
*** "Test visco_plastic_yield_total_strain_healing failed": ***
    Line <112> of file <input string>: No entry with name <Strain healing recovery timescale> was declared in the current subsection.
```

see https://jenkins.tjhei.info/blue/organizations/jenkins/aspect/detail/PR-3765/27/tests

doc/modules/changes/20200812_erinheilman
```diff
@@ -0,0 +1,4 @@
+<li> New: Added calculation for temperature-dependent strain healing in the strain 
+dependent rheology module.
+<br>
+(Erin Heilman, 2020/08/12)
```
tjhei, line 4: please add a newline at the end of the file

source/material_model/rheology/strain_dependent.cc
```diff
@@ -147,6 +147,28 @@ namespace aspect
...
+                           "Whether to apply strain healing to plastic yielding and viscosity terms, "
+                           "and if yes, which method to use. The following methods are available:"
+                           "\n\n"
+                           "\\item ``no healing'': No strain healing is applied. "
+                           "\n\n"
+                           "\\item ``temperature dependent'': Purely temperature dependent "
+                           "strain healing applied to plastic yielding and viscosity terms, similar "
+                           "to the temperature-dependent Frank Kamenetskii formulation, computes "
+                           "strain healing as removing strain as a function of temperature, time, "
+                           "and a user-defined healing rate and prefactor"
```
tjhei, line 162: missing space at the end of the line

source/material_model/rheology/strain_dependent.cc
```diff
@@ -360,6 +408,32 @@ namespace aspect
...
+
+        switch (healing_mechanism)
+          {
+            case no_healing:
+            {
+              break;
+            }
+            case temperature_dependent:
+            {
+              healed_strain -= strain_healing_temperature_dependent_recovery_rate *
```
gassmoeller, line 429: Following my last comment this should then probably be `healed_strain = ... `

source/material_model/rheology/strain_dependent.cc
gassmoeller, line 496: Following my comment from above this line should probably be something like:

```suggestion
{
    // Never heal more strain than exists.
     const double healed_strain = std::min(calculate_strain_healing(in,edot_ii,i), e_ii);
      e_ii -= healed_strain;
}
```
From how I understood your strain healing model the returned value should not be negative (i.e. it is always healing, never damaging), so we should not need the upper limiter and only need the lower limit to ensure e_ii stays positive or 0. Is that correct?

source/material_model/rheology/strain_dependent.cc
gassmoeller: I think this should work approximately as follows:
```
if (healing_mechanism != no_healing)
{
    const Postprocess::Particles<dim> &particle_postprocessor =
      postprocess_manager.template get_matching_postprocessor<Postprocess::Particles<dim> >();
    const Particle::Property::Manager<dim> *particle_property_manager = &particle_postprocessor.get_particle_world().get_property_manager();

             AssertThrow(particle_property_manager.plugin_name_exists("viscoplastic strain invariants") == false, ExcMessage("This healing mechanism currently does not work if the strain is tracked on particles."));
}
```
For this to work you will need to at least add the following headers to the top of the file:
```
#include <aspect/postprocess/particles.h>
#include <aspect/particle/property/interface.h>
```
and maybe also `aspect/simulator.h` (not sure at the moment, try it if it does not compile without).

gassmoeller (commented):
Sorry for the long delay. I think this is in a good state, but would suggest some small changes to how the healing is calculated (only where the existing strain is calculated) and some other minor things. The only critical part left is to get the assert for particles to work, and I think what I suggested should work, let me know how that goes.

source/material_model/rheology/strain_dependent.cc
erinheilman: Yes, I don't want to return a negative value. The upper limiter is there because if the recovery rate is very fast, the strain healing returned becomes physically unreasonable and this stops it.

erinheilman:
@gassmoeller For the Assert for particles, I added the headers (and the simulator.h) but this line is giving me errors when compiling.
const Postprocess::Particles<dim> &particle_postprocessor = postprocess_manager.template get_matching_postprocessor<Postprocess::Particles<dim> >();
The error is saying postprocess_manager isn't declared in the scope?

gassmoeller:
> @gassmoeller For the Assert for particles, I added the headers (and the simulator.h) but this line is giving me errors when compiling.
> const Postprocess::Particles &particle_postprocessor =
> postprocess_manager.template get_matching_postprocessor<Postprocess::Particles >();
> The error is saying postprocess_manager isn't declared in the scope?

Oh, yes, replace that by `this->get_postprocess_manager().template get_matching_postprocessor...` ... my bad. I copied that from inside the simulator class, which has direct access to the postprocess manager, but this plugin has to go through the simulator access function `get_postprocess_manager()`.

source/material_model/rheology/strain_dependent.cc
gassmoeller: But would that not mean you need to limit this only in one direction (preventing too fast healing)? If so limiting the amount of healing to at most the amount of currently stored strain should be enough.

erinheilman: Ah I see, yeah putting e_ii as the other option makes sense.

erinheilman:
@gassmoeller From what I understand, the tests are failing because particles aren't being called as a postprocessor in the input file?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -283,6 +308,35 @@ namespace aspect
...
+
+        if (prm.get ("Strain healing mechanism") == "no healing")
+          healing_mechanism = no_healing;
+        else if (prm.get ("Strain healing mechanism") == "temperature dependent")
+          healing_mechanism = temperature_dependent;
+        else
+          AssertThrow(false, ExcMessage("Not a valid Strain healing mechanism!"));
+
+        // Currently this functionality only works in field composition
+        if (healing_mechanism != no_healing)
```
gassmoeller, line 320: You are right, we first need to check if there are particles. This condition needs to be extended in the following way:

```suggestion
        if (healing_mechanism != no_healing && this->get_postprocess_manager().template has_matching_postprocessor<Postprocess::Particles<dim> >())
```

erinheilman:
@gassmoeller Is this ready to merge now?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -415,7 +494,15 @@ namespace aspect
...
-            const double e_ii = edot_ii*this->get_timestep();
+            double e_ii = edot_ii*this->get_timestep();
+
+            // Adjusting strain values to account for strain healing without exceeding an unreasonable range
+            if (healing_mechanism != no_healing)
+              {
+                // Never heal more strain than exists
+                const double healed_strain = std::min(calculate_strain_healing(in,i), e_ii);
+                e_ii -= healed_strain;
+              }
```
gassmoeller, line 505: Ah, wait a second. I was confused by the name `e_ii`. It is not the accumulated strain, but the amount of strain that gets added in this timestep (`edot_ii * dt`). So it should be named `delta_e_ii`, and the amount of healed strain should not be limited to `delta_e_ii`, but to the total amount of accumulated strain plus the newly added strain (`e_ii + delta_e_ii`). Unfortunately we do not know the value of the accumulated strain here, because it depends on which strain is used for strain weakening (the if conditions below). So we need to unconditionally add the healing term here, and then check below that it does not exceed the value of the respective accumulated strain. Did I understand this correctly?

```suggestion
            double delta_e_ii = edot_ii*this->get_timestep();

            if (healing_mechanism != no_healing)
                delta_e_ii -= calculate_strain_healing(in,i);
```

source/material_model/rheology/strain_dependent.cc
```diff
@@ -415,7 +494,15 @@ namespace aspect
...
             if (weakening_mechanism == plastic_weakening_with_plastic_strain_only && plastic_yielding == true)
               out.reaction_terms[i][this->introspection().compositional_index_for_name("plastic_strain")] = e_ii;
```
gassmoeller, line 507: Each of the lines down here would then need to be changed the following way:

```suggestion
              out.reaction_terms[i][this->introspection().compositional_index_for_name("plastic_strain")] = 
              std::max(delta_e_ii, -in.composition[i][this->introspection().compositional_index_for_name("plastic_strain")]);
```

gassmoeller (commented):
The changes to the assert seem to work, but I think I found an error in the strain healing application. Can you check if my new suggestion is correct? Is there a model for which we can see that the amount of healed strain is actually what we expect (i.e. the resulting strain is never negative, and we heal by the expected amount)?

source/material_model/rheology/strain_dependent.cc
erinheilman: Yes, that would be right

naliboff (approved):
@erinheilman - sorry to take so long to get back to this. Thumbs up from my end. 

@gassmoeller @tjhei - Do you want to do a final check? If not, I think this can be merged.

gassmoeller (approved):
Looks good, thanks!

Issue #3762, opened 2020-08-11
Unusual floating point exception in frank-kamenetskii viscosities.


alarshi:
I noticed a very strange behavior while working on the diffusion strain rates. The new viscosities in the visco_plastic.cc computed for frank_kamenetskii cases throws a floating-point exception because it computes the term ''temperature/reference_temperature" and by default the reference temperature = this->get_adiabatic_surface_temperature(), is zero. All the existing tests throw the same error. But, when I discussed this issue with @erinheilman  and @naliboff , the tests pass even with the default adiabatic surface temperature = 0 on her machine. I am not sure what's going on because the function to compute frank_kamenetskii is always called and the reference_temperature = 0, by default.  Therefore, we should get the floating-point exception but it passed all the tests.

naliboff:
I'm confused about this as well. The lines in question from frank_kamenetskii.cc are:
```
        const double reference_temperature = this->get_adiabatic_surface_temperature();

        const double viscosity_frank_kamenetskii = prefactors_frank_kamenetskii[composition] * std::exp(viscosity_ratios_frank_kamenetskii[composition] * 0.5 * (1.0-temperature/reference_temperature));
```
By default, the adiabatic surface temperature is set to 0, which should have thrown multiple errors for all the default visco_plastic tests.

tjhei:
Well, it looks like the current tester does not have floating point exceptions enabled:
```
-- Performing Test HAVE_FP_EXCEPTIONS - Failed
-- No support for feenableexcept(), disabling runtime floating point exception checks.
```
I can look into that, but that explains why the official tester is not complaining. It depends on the machine/compiler/etc. if we have support for this check.

That said, we should avoid the division by zero.

gassmoeller:
I think there are separate problems here:
- The FK viscosity should not be computed if it is not needed (I did not notice that when reviewing #3698), this is a simple fix.
- The FK should have an `AssertThrow` that makes sure the reference temperature is not 0, since FK divides by the reference temperature there is no way around it, to use FK the adiabatic surface temperature (=reference temperature) has to be non-zero.
- Our default for adiabatic surface temperature of 0 is poorly chosen, but this can hardly be changed, because so much else depends on it.

naliboff:
```
- The FK viscosity should not be computed if it is not needed (I did not notice that when reviewing #3698), this is a simple fix.
[...]

PR #3766, opened 2020-08-11
Fix frank kamenetskii rheology

naliboff:
Following the discussion in #3762, This pull request add a few small fixes to the Frank Kamenetskii rheology module and its implementation in visco_plastic:

1. Ensure the Frank Kamenetskii viscosity is only computed if it is selected.

2. Add an AssertThrow that makes sure the adiabatic surface temperature is not zero when the Frank Kamenetskii rheology is used.

The only downside to what I have implemented is we now need to do the following when going through the `viscous_type` cases:
```
case frank_kamenetskii:
{
  break;
}
```
The Frank Kamenetskii is computed in the next step with: 
```
if (viscous_type == frank_kamenetskii)
viscosity_pre_yield = frank_kamenetskii_rheology.compute_viscosity(temperature, j);
```

We could get around this by removing `frank_kamenetskii` as one of the option for `viscous_type` and having a new input parameter called `Use Frank Kamenetskii rheology`.


source/material_model/visco_plastic.cc
```diff
@@ -263,6 +259,10 @@ namespace aspect
               }
             }
 
+          // Step 1c-1: If selected, replace the current pre-yield viscosity with the Frank Kamenetskii viscosity
+          if (viscous_type == frank_kamenetskii)
+            viscosity_pre_yield = frank_kamenetskii_rheology.compute_viscosity(temperature, j);
```
gassmoeller, line 264: Can you put this in line 250?

naliboff: Yep, I definitely did not need to add these lines.

naliboff: Updated with the last commit.

source/material_model/visco_plastic.cc
```diff
@@ -925,9 +922,12 @@ namespace aspect
...
           dislocation_creep.parse_parameters(prm, std::make_shared<std::vector<unsigned int>>(n_phase_transitions_for_each_composition));
 
-          // FK viscosity parameters
-          frank_kamenetskii_rheology.initialize_simulator (this->get_simulator());
-          frank_kamenetskii_rheology.parse_parameters(prm);
+          // Frank Kamenetskii viscosity parameters
+          if (viscous_flow_law == frank_kamenetskii)
+            {
+              frank_kamenetskii_rheology.initialize_simulator (this->get_simulator());
+              frank_kamenetskii_rheology.parse_parameters(prm);
```
bangerth, line 929: The changes above this block are obviously correct, but I don't understand this. It should always be ok to call `initialize_simulator()` and `parse_parameters()` assuming that `frank_kamenetskii` is a valid object. Why do you need to guard these calls?

tests/visco_plastic_phases_viscosity.prm
```diff
@@ -120,4 +120,4 @@ subsection Solver parameters
   subsection Stokes solver parameters
     set Number of cheap Stokes solver steps = 0
   end
-end
\ No newline at end of file
+end
```
bangerth, line 123: :-)

source/material_model/visco_plastic.cc
bangerth: Or are you saying "we only need this object if the user selected FK rheology, and if we don't need it, there is no reason to initialize this object?

If that were the case, then one should probably also (i) do the same change for the previous `diffusion_creep` and `dislocation_creep` objects or, (i) maybe more elegantly, make these objects `std::unique_ptr` objects that really only point to anything if they are used -- initialization would then happen at the place where you create the objects.

Regardless, this is a question separate from moving the one line of code above, so maybe we could decouple these into separate PRs.

source/material_model/visco_plastic.cc
```diff
@@ -228,9 +228,6 @@ namespace aspect
                                                                                    phase_function_values,
                                                                                    phase_function.n_phase_transitions_for_each_composition());
```
bangerth, line 229: You could also move these statements into their respective `switch` clauses below so that they don't get executed if you don't actually need these factors.

source/material_model/visco_plastic.cc
gassmoeller: In order to not delay this PR (we currently have tests that run into floating point exceptions, although the tester does not pick them up) @naliboff: can you just revert this to the previous state? `frank_kamenetskii` is a valid member of this class, and as @bangerth points out there is no harm in calling these functions, even if the object is never used.

tests/visco_plastic_phases_viscosity.prm
naliboff: sorry, that must have been from `make indent`

gassmoeller:
Looks good to go from my side. We could make further improvements in a follow-up.

source/material_model/visco_plastic.cc
naliboff: Reverted :) I had guarded other calls elsewhere in visco_plastic, but now I don't recall why I did that. I any case, it was clearly unnecessary.

tests/visco_plastic_phases_viscosity.prm
bangerth: No, adding the endline is the right thing to do -- your patch did what should have happened when that file was first added :-)

bangerth:
Thanks -- feel free to guard the initialization of these members in a separate PR: if you don't need a member because of parameter settings, then making it a pointer that is never set to anything other than `nullptr` is the right thing.

source/material_model/visco_plastic.cc
gassmoeller: Ah, I see now why you guarded the call. Now the tests fail because the rheology checks for the reference temperature being larger than 0. Can you go with the proper fix that Wolfgang proposed and make `frank_kamenetskii` a `std::unique_ptr` that only gets initialized if selected?

naliboff: 👍 Working on this now. To be fair, I did not have that Assert in mind when guarding the calls :)

naliboff: Updated, let's see if this works. Thanks @gassmoeller for the explanation on why making the `frank_kamenetskii` a  `unique_ptr` is better practice.

naliboff:
@bangerth @gassmoeller - Two of the testers are passing, but one tester is having the visco plastic derivatives randomly fail? I recall that this was an issue in another PR?

bangerth:
Let's finally get this merged. If you wanted to work on a follow-up, the other variables just preceding the FK object could probably be dealt with in a similar way. 

/rebuild

naliboff:
Rebasing to matter to see if that fixes the issue with the randomly failing newton derivative tests.

gassmoeller:
The tester was not happy, because the `visco_plastic_derivatives` tests unintentionally created copies of the material model, but copying a `std::unique_ptr` is not allowed. I think the copy was a bug anyway, because it was only used to call a static function of the class, so I updated the tests. Lets see if the tester is happy now.

tests/visco_plastic_derivatives_2d.cc
```diff
@@ -105,8 +105,8 @@ void f(const aspect::SimulatorAccess<dim> &simulator_access,
   // initialize the material we want to test.
   aspect::ParameterHandler prm;
 
-  const aspect::MaterialModel::ViscoPlastic<dim> const_material_model = dynamic_cast<const aspect::MaterialModel::ViscoPlastic<dim> &>(simulator_access.get_material_model());
-  aspect::MaterialModel::ViscoPlastic<dim> material_model = const_cast<aspect::MaterialModel::ViscoPlastic<dim> &>(const_material_model);
+  const aspect::MaterialModel::ViscoPlastic<dim> &const_material_model = dynamic_cast<const aspect::MaterialModel::ViscoPlastic<dim> &>(simulator_access.get_material_model());
+  aspect::MaterialModel::ViscoPlastic<dim> &material_model = const_cast<aspect::MaterialModel::ViscoPlastic<dim> &>(const_material_model);
```
bangerth, line 109: Oh, yes, that was not the intent :-)

PR #3767, opened 2020-08-11
C++11: Avoid 'typedef' in favor of 'using' declarations.

bangerth:


bangerth:
No advantages, but `using` is idiomatic C++ these days with `typedef` being out of fashion because it makes it so totally awkward to define pointers to functions:
```
  typedef int (*ptr)(void (*)(void));
```
(That's the type of the `atexit` function. Try also `man on_exit` for amusement.)

I see `using` vs `typedef` in the same way as range-based `for` loops vs old-style `for` loops: The former is what people will primarily use in the future, and so this patch is simply a "modernization" of our code base.

PR #3770, opened 2020-08-12
Added Stabilization Factor for Elasticity

danieldouglas92:

* [ ] I have read the guidelines in our [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) document.

### For all pull requests:


source/material_model/rheology/elasticity.cc
```diff
@@ -325,18 +334,16 @@ namespace aspect
...
                              ?
-                             this->get_timestep()
+                             this->get_timestep() * stabilization_scale_factor
                              :
                              fixed_elastic_time_step);
         return dte;
       }
 
-
-
```
gassmoeller, line 339: please restore these empty lines. our convention is to have three empty lines between function in .cc files, and one empty line between functions in .h files.

source/material_model/rheology/elasticity.cc
```diff
@@ -325,18 +334,16 @@ namespace aspect
         //
         // We also use this parameter when we are still *before* the first time step,
         // i.e., if the time step number is numbers::invalid_unsigned_int.
-        const double dte = ( ( this->get_timestep_number() > 0 &&
+        double dte = ( ( this->get_timestep_number() > 0 &&
```
gassmoeller, line 337: this can still be `const double`

include/aspect/material_model/rheology/elasticity.h
```diff
@@ -160,6 +160,12 @@ namespace aspect
            * Double for fixed elastic time step value, read from parameter file
            */
           double fixed_elastic_time_step;
+
+          /**
+           * Double for amount of stabilization applied to elastic stresses
+           */
+
```
gassmoeller, line 167: please remove empty line

include/aspect/material_model/rheology/elasticity.h
```diff
@@ -160,6 +160,12 @@ namespace aspect
            * Double for fixed elastic time step value, read from parameter file
            */
           double fixed_elastic_time_step;
+
+          /**
+           * Double for amount of stabilization applied to elastic stresses
```
gassmoeller, line 165: I would rephrase this the following way: `A stabilization factor for the elastic stresses that influences how fast elastic stresses adjust to deformation. 1.0 is equivalent to no stabilization, and infinity is equivalent to not applying elastic stresses at all. The factor is multiplied with the computational time step to create a time scale.`

Writing useful documentation is a bit unintuitive, but it is essentially not at all about telling what this **is** (users can see that in the code), but what this is **for**.

include/aspect/material_model/rheology/elasticity.h
```diff
@@ -160,6 +160,12 @@ namespace aspect
...
+           */
+
+          double stabilization_scale_factor;
```
gassmoeller, line 168: maybe `stabilization_time_scale_factor`

source/material_model/rheology/elasticity.cc
```diff
@@ -96,6 +96,10 @@ namespace aspect
                            Patterns::Bool (),
                            "Whether to apply a stress averaging scheme to account for differences "
                            "between the fixed elastic time step and numerical time step. ");
+        prm.declare_entry ("Stabilization scale factor", "1.",
+                           Patterns::Double (1., std::numeric_limits<double>::max()),
+                           "Scale Factor which controls the degree of stabilization applied to elastic "
+                           "stresses.");
```
gassmoeller, line 102: Copy the documentation I suggested above.

source/material_model/rheology/elasticity.cc
```diff
@@ -96,6 +96,10 @@ namespace aspect
                            Patterns::Bool (),
                            "Whether to apply a stress averaging scheme to account for differences "
                            "between the fixed elastic time step and numerical time step. ");
+        prm.declare_entry ("Stabilization scale factor", "1.",
```
gassmoeller, line 99: Stabilization time scale factor

source/material_model/rheology/elasticity.cc
```diff
@@ -123,6 +127,11 @@ namespace aspect
           AssertThrow(use_fixed_elastic_time_step == true,
                       ExcMessage("Stress averaging can only be used if 'Use fixed elastic time step' is set to true'"));
 
+        stabilization_scale_factor = prm.get_double ("Stabilization scale factor");
+        if (stabilization_scale_factor < 1)
+          AssertThrow(stabilization_scale_factor < 1,
+                      ExcMessage("Stabilization scale factor must be greater than one"));
```
gassmoeller, line 133: You can simplify this in the following way:
```suggestion
          AssertThrow(stabilization_scale_factor >= 1.,
                      ExcMessage("Stabilization time scale factor must be greater than or equal to one."));
```

Technically this will never happen, because you declared the minimum allowed value in the `declare_parameters` function already (you can check that by trying to set it lower and commenting out this line). However it does not hurt to keep it to make this point extra obvious.

danieldouglas92:
I think I've taken care of all the commits! Thanks for the tips about commenting and the blank line format I'll try to remember that for future commits

source/material_model/rheology/elasticity.cc
```diff
@@ -127,10 +129,9 @@ namespace aspect
           AssertThrow(use_fixed_elastic_time_step == true,
                       ExcMessage("Stress averaging can only be used if 'Use fixed elastic time step' is set to true'"));
 
-        stabilization_scale_factor = prm.get_double ("Stabilization scale factor");
-        if (stabilization_scale_factor < 1)
-          AssertThrow(stabilization_scale_factor < 1,
-                      ExcMessage("Stabilization scale factor must be greater than one"));
+        stabilization_time_scale_factor = prm.get_double ("Stabilization time scale factor");
+        AssertThrow(stabilization_time_scale_factor >= 1,
```
gassmoeller, line 133: Please make the following minor change to signal this is a `double` not an integer.

```suggestion
        AssertThrow(stabilization_time_scale_factor >= 1.,
```

source/material_model/rheology/elasticity.cc
```diff
@@ -96,6 +96,12 @@ namespace aspect
                            Patterns::Bool (),
                            "Whether to apply a stress averaging scheme to account for differences "
                            "between the fixed elastic time step and numerical time step. ");
+        prm.declare_entry ("Stabilization time scale factor", "1.",
+                           Patterns::Double (1., std::numeric_limits<double>::max()),
```
tjhei, line 100:
```suggestion
                           Patterns::Double (1.),
```

source/material_model/rheology/elasticity.cc
tjhei: This assert is not necessary as the get_double should do the check, but I guess it doesn't hurt.

danieldouglas92:
I hope that adding the entire output directory and test input file is what I was supposed to commit for the test case. I checked that there was a difference in the output when the stabilization factor was 2 vs 1!

doc/modules/changes/20200813_danieldouglas92
```diff
@@ -0,0 +1,6 @@
+New: Added a new double stabilization_time_scale_factor to Rheology::Elasticity.
+
+This variable is a stabilization factor for the elastic stresses that influences how fast elastic stresses adjust to deformation.
```
gassmoeller, line 3:
```suggestion
New: Added a new input parameter stabilization_time_scale_factor to Rheology::Elasticity.
This variable is a stabilization factor for the elastic stresses that influences how fast elastic stresses adjust to deformation.
```

tests/elastic_stabilization_time_scale_factor/log.txt
```diff
@@ -0,0 +1,80 @@
+-----------------------------------------------------------------------------
```
gassmoeller, line 1: Please rename this file to `screen-output` (this has historical reasons)

tests/elastic_stabilization_time_scale_factor/original.prm
```diff
@@ -0,0 +1,40 @@
+# This test checks whether the viscoelastic finite-length elastic plate
```
gassmoeller, line 1: please remove this file and all other parameters files. The idea is to only have files in the directory that show some model result that lets us determine if the test output has changed (typical examples are `screen-output` and `statistics`)

danieldouglas92:
Should I rebase this pull request?

gassmoeller:
No need for rebasing, but you need to fix the indentation (`make indent` or download the artifact from jenkins.tjhei.info and do `git apply` on the patch file), and afterwards probably need to update the test output with the output from the tester. Let me know if you need help with that.

danieldouglas92:
Hello again! Sorry for taking so long to get back to this indent issue, I was trying to configure astyle and I was having troubles getting it working (I still don't have it working so hopefully the manual indent fix I did worked) and at some point I must have forgotten about this pull request

doc/modules/changes/20200813_danieldouglas92
```diff
@@ -0,0 +1,7 @@
+New: Added a new input parameter stabilization_time_scale_factor to Rheology::Elasticity.
```
tjhei, line 1:
```suggestion
New: Added a new input parameter "Stabilization time scale factor" to Rheology::Elasticity.
```

EstherHeck:
Hi, I have started testing this with the free-surface-tractions benchmark and noticed, that now the parameter "set Use fixed elastic time step" becomes somewhat misleading. I need to set it to true, so stress averaging can be used. But I find the naming confusing now. As it is not a fixed amount of time anymore, but a fixed multiple of the computational timestep. What do you think about it?

EstherHeck:
Hi all, 
thanks for tackling this. I have run a few test models with the free-surface-tractions/viscoelastic/ 2D loading benchmark and noticed, that the new scaling of the elastic time step only works if "set Use fixed elastic time step = false". However, I can only set "set Use stress averaging = true" if "set Use fixed elastic time step = true". That means, that I cannot use the new scaling together with the stress averaging. 
Looking at the results from the loading benchmark, I noticed, that the fit to the analytical solution does increase with the new scaling factor in terms of abruptness of change in minimum surface topography. 
But without the stress averaging, I do not fit the magnitude. 

Here is a plot of minimum topography over time that shows what I mean:
red: analytical solution
green: uses a fixed elastic timestep and stress averaging
orange: uses a scaling factor of 10, but no stress averaging
(Forget about the blue in the legend ;) )
image

gassmoeller:
Thanks for the feedback!
@danieldouglas92: If you are having trouble installing astyle (we need version 2.04, most conveniently installed via candi https://github.com/dealii/candi using `candi.sh --packages="astyle" -p INSTALL_PATH)` then you can always download the correctly indented file from the tester by clicking on 'Details' next to the test results of continuous-integration/jenkins/pr-merge. Go to the 'Artifacts' tab and download the 'changes-astyle.diff' file. Put it into the aspect directory and run `git apply changes-astyle.diff`. This will correctly indent your files, and you can then commit and push them.

@EstherHeck: Ah, I did not realize this. Yes, the assert in line 128-130 now makes no sense, because we can also have stress averaging without a fixed timestep. Also, because of the if-condition in line 308 the stress averaging is only done if a fixed elastic time step is used. @danieldouglas92: Could you remove the assert in lines 128-130 allowing to use stress averaging without a fixed timestep, and remove the `use_fixed_elastic_time_step == true` part of the if-condition in line 308? Esther could you make the same changes and rerun your benchmark?

EstherHeck:
Hi @gassmoeller and @danieldouglas92, 
the changes you proposed Rene, seem to work pretty well, thanks for that! Here are the results from the loading benchmark. scaling factor is set to 10 in all models: 

the orange curve remained the same, using the new elastic time step scaling factor without stress averaging. Amplitude underestimated
the green curve uses it with stress averaging. Amplitude overestimated
blue curve is the same as green, but setting the maximum relative increase from PR #3842 to 2. The instantaneous change in surface topography is very well represented, the entire curve is just a bit offset upwards compared to the analytical solution. 
the red curve uses the maximum relative increase with a fixed elastic time step. Offset is similar as for the blue one, but the kink is less sharp, so this can be attributed to the new elastic scaling factor
purple is without using any of the new functionalities. It does better represent the amplitude of the analytical solution, but the kink is more round and it took longer to compute, as it uses dt-el=5 and dt-comp=0.5 years

image

EstherHeck:
and maybe just to let you know, the model crashes as it does not converge when I set the scaling factor to 2

gassmoeller:
> and maybe just to let you know, the model crashes as it does not converge when I set the scaling factor to 2

Well that would be good to avoid, wouldnt it :-). Just to clarify: It works ok with a factor of 10, but crashes with 2? Does it also crash with 1?

I am a bit confused by your description, because there are too many features combined for me too follow it. Could you let me know about the following setups and if my summary sounds correct (all with stress averaging):

1. The old way you did things was the purple curve which severely limited the time step size and fixed the elastic time step and was slow to compute, but gave approximately correct results.
2. If you exactly repeat model one (purple line), but instead of a fixed elastic time step you use the new factor of 10, and keep the normal time step size limited the same way, you should get exactly the same results, did you try this setup?
3. If you run several models and gradually decrease the elastic factor from 10 to 1 (e.g. at 5, 3, 2, 1) you should see the kinks becoming more pronounced/less rounded, but also gradually more oscillations until the model does not converge any more. could you try this and report back at which point the model starts oscillating or not converging? Somewhere above that point will be the optimal value for the scaling factor.
4. If you gradually relax the limitation on the normal time step size (with a fixed elastic factor of either 10, or any other stable constant number) you see a transition from the purple line to the green curve if I understood you description correctly. Is that what happens?
5. Finally, without limiting the time step, and using a factor of 10 and #3842 allows to recover nearly the old behavior of the purple line, but without needing to prescribe a fixed elastic time step, or artificially limit the actual time step, is this correct? How large are the actual time steps in this model, compared to the purple line model? Are they significantly larger, or about the same size?

gassmoeller:
Oh, I should also add I am open to merging this without all of the models I have described above, but I need at least a clear indication if the state you described works for you. The steps I described are the ones I would take to determine the optimal number for the elastic time step.

EstherHeck:
HI @gassmoeller the models are running. Am away for a week and will post the results afterwards. Thought they would be done by now, but unfortunately they are not

EstherHeck:
Hi Rene, 
here are the results from the test runs: 

>     1. The old way you did things was the purple curve which severely limited the time step size and fixed the elastic time step and was slow to compute, but gave approximately correct results.

Yes, that is correct. 
 
>     2. If you exactly repeat model one (purple line), but instead of a fixed elastic time step you use the new factor of 10, and keep the normal time step size limited the same way, you should get exactly the same results, did you try this setup?

this gives the following plot. I do not get the same results, though it should have been the same from my understanding

EstherHeck:
blue is the old way
orange the new way 
image

EstherHeck:
> Well that would be good to avoid, wouldnt it :-). Just to clarify: It works ok with a factor of 10, but crashes with 2? Does it also crash with 1?

>     3. If you run several models and gradually decrease the elastic factor from 10 to 1 (e.g. at 5, 3, 2, 1) you should see the kinks becoming more pronounced/less rounded, but also gradually more oscillations until the model does not converge any more. could you try this and report back at which point the model starts oscillating or not converging? Somewhere above that point will be the optimal value for the scaling factor.

I tried elastic factor of 1,2,3,4,5,7,10
Except for 1 and 10 all models failed because iterative solver did not converge. 

Models 1 and 10 look the same, no difference in roundness. I will try another set with a higher CFL number.

image

EstherHeck:
>     4. If you gradually relax the limitation on the normal time step size (with a fixed elastic factor of either 10, or any other stable constant number) you see a transition from the purple line to the green curve if I understood you description correctly. Is that what happens?

This is what I would have thought too, but results are different. In the plot, the models with the same maximum time step size have the same line style and those models with the same CFL number have the same color. 
Those models without the new elastic scaling factor, but with the old fixed elastic dt are grey.

First thing to notice is, that the max dt size still controls the sharpness of the kink. dt=10 is most rounded (dashed), then comes dt=5 (dotted), dt=2 (solid), dt=0.5 (dashdotted)

For the CFL number, we only see two (+1) colours. The other curves are hidden behind these two colours. So CFL = 0.1 and 0.01 and 0.001 are hidden behind the magenta lines which is CFL = 0.5
CFL = 0.00001 failed
So either we set CFL pretty low, to 0.0001 and get very different amplitudes of the surface topography or we set CFL higher, then it does not make a difference anymore, and amplitudes stay closer to the  analytical solution. 
(Repeating time steps was not allowed here and the elastic scaling factor was set to 10)
image

EstherHeck:
>     5. Finally, without limiting the time step, and using a factor of 10 and #3842 allows to recover nearly the old behavior of the purple line, but without needing to prescribe a fixed elastic time step, or artificially limit the actual time step, is this correct? How large are the actual time steps in this model, compared to the purple line model? Are they significantly larger, or about the same size?

Here I did not prescribe the maximum time step size, allowed the model  to repeat time steps  and varied the CFL number. 
CFL = 0.1 and 0.5 both dont do much.
CFL=0.01 also not suitable
CFL = 0.001 does have some topography change after 1000 years
CFL = 0.0001 gets the kink, but a bit too late. It does only repeat at 1400 something but not before 1000.
CFL = 0.00001 failed as it was repeating the first time step over and over again

image

time step sizes are:
image

or leaving high CFL numbers out:
image

EstherHeck:
> > Well that would be good to avoid, wouldnt it :-). Just to clarify: It works ok with a factor of 10, but crashes with 2? Does it also crash with 1?
> 
> > ```
> > 3. If you run several models and gradually decrease the elastic factor from 10 to 1 (e.g. at 5, 3, 2, 1) you should see the kinks becoming more pronounced/less rounded, but also gradually more oscillations until the model does not converge any more. could you try this and report back at which point the model starts oscillating or not converging? Somewhere above that point will be the optimal value for the scaling factor.
> > ```
> 
> I tried elastic factor of 1,2,3,4,5,7,10
> Except for 1 and 10 all models failed because iterative solver did not converge.
> 
> Models 1 and 10 look the same, no difference in roundness. I will try another set with a higher CFL number.
> 
This time it worked. No models failed, and there is a transition from oscillating to kinks to rounded.  CFL number = 0.01 for these models, while it was 0.0001 for the ones above

image

bobmyhill:
Hi @danieldouglas92, @erinheilman, @gassmoeller 
I hope you won't mind me jumping in to the party so late. This PR implements a firmoviscous (Kelvin) rheology, so I'm wondering whether it might be good to make this explicit by replacing "Stabilization time scale factor" (`f`) with "Elastic damper viscosity" (`eta_ed`)?

The relationship between the two parameters is 
`eta_ed = shear_modulus*elastic_timestep*(f - 1)` or 
`f = 1 + eta_ed/(shear_modulus*elastic_timestep)`. 

Making `eta_ed` constant might also help ensure stabiility throughout a simulation.

bobmyhill:
I should add that I'm not necessarily against having the stabilization kept as it is. I was just looking ahead to what @naliboff and I will have to do to make the new composite VEP model, and realised that this PR almost does what I had in mind.

bobmyhill:
@naliboff has kindly taught me a bit more about how the elastic model works. If I follow everything correctly, you have *exactly* implemented a firmoviscous model :) 

The only changes that would be necessary to make this explicit and to make the damper viscosity constant (again, if I understand everything correctly, and again, if you want to), would be to
- make the new parameter `damper_viscosity`
- make `elastic_timestep()` dependent on the shear modulus (so the four calls to this function would need to be changed and moved into for loops where necessary)
- Add the following line to that function: 
`stabilization_time_scale_factor = 1 + damper_viscosity/(shear_modulus*elastic_timestep);`.

We can discuss this if you want? Either way, I think this PR is cool :)

naliboff:
@danieldouglas92 @gassmoeller @EstherHeck @bobmyhill - Jumping into the conversation as well :-)

I just read through the reviews and independent of the tests Esther conducted, are these the only remaining items to complete before it is ready to merge?:

bobmyhill:
Hi @naliboff
I've gone through the derivation this afternoon (and have had to modify my previous comment accordingly). Hopefully I can now answer your questions accurately.
1. When we enforce isostrain conditions on two rheological elements (here elastic and viscous), the corresponding properties are intimately linked. Just like the damped Drucker-Prager, we can't split the calculations.   
2. ASPECT treat distinct compositions separately at the moment (isostrain for each material). So if the rheology is physically reasonable for a single material, it is also reasonable for multiple materials. As for whether an isostrain condition is reasonable, that's a question for another time (and one that I know we've been discussing since at least 2015 :) ).
3. A constant damper viscosity will affect the materials with the smallest shear modulus most severely, which *I think* might be most useful in terms of encouraging solver stability in materials with >1 composition. But it would also make sense to allow the user to assign the damper viscosity separately for each material. Some users may want a constant ratio of shear modulus to damper viscosity (the same characteristic timescale for all materials). I'm ambivalent about the choice made, but I do think that it's valuable for the user to visualise what they are doing when they modify the stabilization factor.

EstherHeck:
> @EstherHeck @danieldouglas92 - From your testing, does the optimal factor vary quite a bit between different model setups? Are their guidelines that we can provide in the documentation?

Well, that is a difficult question. from the plots above you can see, that a factor of 2 is optimal in terms of reproducing sudden changes. However, in other, more complex, models I got into troubles of convergence with a factor of 2 and use 5 at the moment. So as a guideline I would say: make it as low as possible (best would be 2), but as soon as there are oscillating motions, increase the factor.  If it is very high, the elastic response will dissolve.

naliboff:
@EstherHeck - Thanks for the summary! I think roughly what you wrote above should be added to the parameter description.

naliboff:
@bobmyhill and I discussed the issue of the plastic damper further, and concluded that we can wait on those decisions (i.e., defining 'factor' versus 'damper') can wait until after this PR is merged.

EstherHeck:
@naliboff with the changes of #3923 I get the following graph and want to add to the description of the parameter, that larger values for the factor are needed, when the option to repeat time steps is used, e.g. = 10

grafik

bobmyhill:
@EstherHeck These new plots look great! Can you explain why the damped behaviour has gone away?

EstherHeck:
before, the kinkyness or roundness of the surface topography was dominated by the size of the elastic timestep relative to the computational timestep and hence by the elastic scaling factor. These new models use the option to repeat timesteps, so both, the computational and the elastic timestep become very small when velocities go up, which is why on this scale, the curves do not look round anymore. However, I just realized, that the green curve did not make use of the option to repeat a timestep, only yellow and blue did.

bobmyhill:
@EstherHeck Thanks! So am I correct in thinking that the stabilization factor also controls the duration of the recalculated timestep? That's probably ok, although it could do with some documentation. I'll be interested to see how this performs in a more complex model.

EstherHeck:
@bobmyhill yes I guess this is right, as the elastic scaling factor is the only thing that varies between the models above. Two models repeated but the third one did not, which means that the elastic scaling factor of 10 made velocities increase too slowly for the repeat timestep statement to take place

naliboff:
@EstherHeck - Following our discussion at the ASPECT user meeting, would it be possible to redo the above checks with the `viscoelastic_stress_build-up` benchmarks? That benchmark does not use a free surface and also has a simple analytical benchmark. 

Another option would be the `viscoelastic_stress_build-up_yield` (same benchmark + fixed yield stress) or the `viscoelastic_plastic_simple_shear` benchmark.

EstherHeck:
Hi @naliboff , yes I can have a look at it, probably at the end of this week

bobmyhill:
Hi @EstherHeck, hi @danieldouglas92, any update on this one?

EstherHeck:
yes, kind of. Testing it with another benchmark, but have no satisfying answer yet, hopefully tomorrow, its high on my list

EstherHeck:
Hi @naliboff I did run those models and as discussed this is what I get:

legend:
y: means this is a modified input file from viscoelastic_stress_build_up_YIELD
-r: timesteps are allowed to be repeated
-f: using a fixed elastic timestep
-a: using stress averaging only indicated in legend for models using fixed elastic dt. Not indicated but used in all models with elastic timescale factor
-SOMENUMBER: the value of the elastic time scale factor

As in the other model setup it is obvious, that the smaller the elastic time step that is used, the closer we get to the analytical solution.  Though even with elastic scaling factor equaling 1 we never truely get it. However, when using repeating time steps we always perfectly match the analytical stress build up (until yielding occurs), regardless of the elastic time scale factor
[compare_stress_-onlydtel_zoom.pdf](https://github.com/geodynamics/aspect/files/6085366/compare_stress_-onlydtel_zoom.pdf)

This compilation cleary shows that the fixed elastic time step should NEVER be used without averaging (the two outliers at the left) and that for the timescales looked at here, the fixed elastic time step (fixed to dtel = 2 dtcomp) matches the elastic time scale factor of 2 (no wonder, its the same: dtel = 2 dtcomp). However, when using the repeating timesteps the fixed elastic timestep will never match the analytical solution while the factor does! (compare dashed pink againstplain pink whihc is hidden below moss green)
[compare_stress_-_zoom.pdf](https://github.com/geodynamics/aspect/files/6085373/compare_stress_-_zoom.pdf)

I dont know if the factor sould be 1, though this plot suggests it. I would say it should be 2, because setting it to 1 results in velocity oscillations for the other models...

I guess my recommendation from this really would be, choose whatever factor, but use repeating timesteps. just to illustrate that again, even the viscoelastic (nonplastic model) fits the analytical solution so much better with repeating timesteps
[compare_stress_.pdf](https://github.com/geodynamics/aspect/files/6085412/compare_stress_.pdf)

naliboff:
@EstherHeck - fantastic, thank you for all this testing! I largely agree with your summary, although I wonder if we should keep the default scaling factor at 1 given that it provides the best fit for a key benchmarks? We can always carefully document in the parameter description and manual that one should thoroughly test the effect of this parameter for the problem at hand.

Once that decision is made and the PR is updated to match the changes on your branch, I think we should be ready to merge.

@gassmoeller, @bobmyhill  and others - Should these results be discussed at the user meeting or elsewhere before the final merge?

Thanks again for all the work @EstherHeck!

bobmyhill:
If @EstherHeck or @danieldouglas92 are willing to discuss this PR at the user meeting, I think others would be interested. I don't think that should hold up updating/merging this PR though. 

@EstherHeck, @danieldouglas92, in case you need advice on how to update the PR, one option would be for Esther to make a pull request from her branch into danieldouglas92:elasticity_stabilization_parameter (rather than into geodynamics:master), and then for Daniel to merge that PR into his branch and make any necessary updates from there.

EstherHeck:
Hi all, yes, I can join the next user meeting. 

For the changes, I don't have a clean branch with them, it is all within my rate-and-state friction branch. So it is the same amount of work if I make  a new branch with those changes or @danieldouglas92 makes them. What do you prefer, @danieldouglas92 ?

EstherHeck:
as @danieldouglas92 does not answer. Should I make a new PR with those changes from my branch?

danieldouglas92:
Apologies for dropping the ball on this pull request. If it's not too much trouble for you @EstherHeck I think you've done a lot more leg work in terms of testing the stabilization parameter so I think it would be great if you took over this PR

EstherHeck:
Hi all, yes I can do that, I only need some help how to take over. Make a new PR altogether and just link this one as a comment? 
and @danieldouglas92 thanks for "giving me the credits"

bobmyhill:
@EstherHeck, there are a lot of different ways to take over. 

If it isn't too much work to build off the branch of @danieldouglas92 you could do that. You might already be doing this, but if you aren't, the steps are:
- Add Daniel's remote to your local repo:
`git remote add danieldouglas92 git@ssh.github.com:danieldouglas92/aspect.git`
- `git checkout -b elasticity_stabilization_parameter danieldouglas92/elasticity_stabilization_parameter`

This way, you preserve Daniel's commits, but you can then add more changes and open a new PR.

In the changelog, you should add your own name in addition to Daniel's, and also add @gassmoeller, as I think this stabilisation scheme was originally his idea.

EstherHeck:
Thanks @bobmyhill for your git-help! I included all the changes and suggestions I could find in all those comments above in the new elastic stabilization parameter PR #4011

bobmyhill:
Should this PR have been merged? I guess it's fine if it didn't break anything, but I thought that #4011 should have been sufficient.

bobmyhill:
Oh, wait, all the old commits were preserved in the new PR. Ignore me!

PR #3772, opened 2020-08-12
added constructor for NamedAdditionalMaterialOutputs

bobmyhill:
This PR adds a constructor to NamedAdditionalMaterialOutputs so that the user can pass user-defined output names and a number of points for which those outputs should be stored.

This PR was suggested by @gassmoeller based on comments in #3733.

### DIFF
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -838,7 +838,7 @@ namespace aspect
     {
       public:
         /**
-         * Constructor.
+         * Base constructor.
          *
          * @param output_names A list of names for the additional output variables
          *   this object will store. The length of the list also indicates
@@ -846,6 +846,19 @@ namespace aspect
          *   will store.
          */
         NamedAdditionalMaterialOutputs(const std::vector<std::string> &output_names);
+
+        /**
+         * Constructor for case where outputs are stored for a number of points.
+         *
+         * @param output_names A list of names for the additional output variables
+         *   this object will store. The length of the list also indicates
+         *   how many additional output variables objects of derived classes
+         *   will store.
+         * @param n_points The number of points for which to store each of the
+         *   output variables.
+         */
+        NamedAdditionalMaterialOutputs(const std::vector<std::string> &output_names,
+                                       const unsigned int n_points);
 
         /**
          * Destructor.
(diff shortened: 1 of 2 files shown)


bobmyhill:
@gassmoeller, I've tentatively chosen the loop hierarchy suggested by @jdannberg in #3733 (output_values[i][j] corresponds to the ith output for the jth evaluation point). This simplifies `get_nth_output`, but it does mean that filling the output_values object will require a few more lines in the material models in which it is required. Are you happy with this way round?

source/material_model/interface.cc
```diff
@@ -852,6 +863,19 @@ namespace aspect
 
 
 
+    template<int dim>
+    std::vector<double>
+    NamedAdditionalMaterialOutputs<dim>::get_nth_output(const unsigned int idx) const
+    {
+      // In this function we extract the values for the nth output
+      // The number of outputs is the outer vector
+      Assert (output_values[idx].size() > 0,
```
gassmoeller, line 872: Yes, I think you need another Assert for `output_values.size() > idx` just to be sure.

include/aspect/material_model/interface.h
```diff
@@ -869,6 +882,13 @@ namespace aspect
                       const FullMatrix<double>  &/*expansion_matrix*/) override
         {}
 
+
```
gassmoeller, line 885: up in line 878 (cannot comment there) you can now remove the `= 0` from the `get_nth_output` function, because we now have a default implementation.

bobmyhill:
Thanks @gassmoeller. Currently I can compile on my own computer, but the testers fail. I do get a warning:

`In file included from /home/bob/software/aspect/build/CMakeFiles/aspect.dir/Unity/unity_12_cxx.cxx:10:
/home/bob/software/aspect/source/material_model/interface.cc: In instantiation of ‘aspect::MaterialModel::NamedAdditionalMaterialOutputs<dim>::NamedAdditionalMaterialOutputs(const std::vector<std::__cxx11::basic_string<char> >&, unsigned int) [with int dim = 2]’:
/home/bob/software/aspect/source/material_model/interface.cc:1119:5:   required from here
/home/bob/software/aspect/source/material_model/interface.cc:840:5: warning:   when initialized here [-Wreorder]
  840 |     NamedAdditionalMaterialOutputs<dim>::`

Do you know what I'm doing wrong?

source/material_model/interface.cc
```diff
@@ -836,6 +836,17 @@ namespace aspect
 
 
 
+    template <int dim>
+    NamedAdditionalMaterialOutputs<dim>::
+    NamedAdditionalMaterialOutputs(const std::vector<std::string> &output_names,
+                                   const unsigned int n_points)
+      :
+      names(output_names),
+      output_values(output_names.size(), std::vector<double>(n_points, numbers::signaling_nan<double>()))
```
gassmoeller, line 845: You need to switch these two lines so that the order is the same as in the header file, I think this is causing the warning.

bobmyhill: Thanks for the quick reply :) ... fingers crossed!

PR #3773, opened 2020-08-12
Sea level postprocessor: finds and writes elevation of any desired location.

mfmweerdesteijn:
…red location

Adding sealevel.cc and sealevel.h
Function which can find the elevation of any desired location, so not dependent on topography postprocessor.
For now the execute function asks for the Guassian integration points
Rene: Im not sure if I base the quadrature degree on the right thing for QGauss. Can you have a look at that?


gassmoeller:
As discussed we will return to this PR once #3776 has progressed (just as a note for other reviewers).

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+#include <aspect/simulator_access.h>
+
+
+namespace aspect
+{
+  namespace Postprocess
+  {
+
+    /**
+     * A postprocessor that outputs the sea level to file.
```
gassmoeller, line 35: This will need more documentation. How is the sea level computed? Which other quantities are used (geoid? topography? ice?).

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_postprocess_sealevel_h
+#define _aspect_postprocess_sealevel_h
```
gassmoeller, line 23: general question: Would you rather write this a sealevel equation, or sea level equation? If the second, then maybe the postprocessor should be called `sea_level` instead of `sealevel`? I don't know about the naming conventions for this, but compared to most of our other postprocessors I would favor `sea_level` (as file name and here, `SeaLevel` can remain the class name).

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+    class SeaLevel : public Interface<dim>, public ::aspect::SimulatorAccess<dim>
+    {
+      public:
+        /**
+         * Sets up structured data lookup for topography and ice height input data.
+         */
+        void initialize() override;
+
+        /**
+         * Output non-uniform sea level change [m] to file.
```
gassmoeller, line 49: As far as I understood there is a uniform and non-uniform part to the sea level output. Will this function only write the non-uniform part or the sum between uniform and non-uniform part? Update the comment accordingly.

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+         */
+        std::pair<std::string,std::string> 
+        execute (TableHandler &statistics) override;
+
+        /**
+         * Evaluate the total surface pressure from ice and water loading at 
+         * a point.
```
gassmoeller, line 56:
```suggestion
         * Evaluate the total surface pressure from ice and water loading at 
         * a point at the top surface of the domain. While the function will
         * return pressure values for any location, it will always assume the
         * volume between the point and the sea level is filled with water or ice,
         * therefore calling this function for positions inside the domain will result
         * in unrealistic pressure values.
```

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+         */
+        double
+        compute_total_surface_pressure(const Point<dim> &) const;
```
gassmoeller, line 59:
```suggestion
        compute_total_surface_pressure(const Point<dim> &position) const;
```

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+
+      private:
+        /**
+         * Function to determine non-uniform sea level change for any given location.
+         */
+        double
+        compute_nonuniform_sealevel_change(const Point<dim> &position) const;
+        
+        /**
+         * Function to determine the sea level offset.
```
gassmoeller, line 89:
```suggestion
         * Function to determine the global sea level offset.
```

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+
+        /**
+         * Read the parameters this class declares from the parameter file.
+         */
+        void
+        parse_parameters (ParameterHandler &prm) override;
+
+      private:
+        /**
+         * Function to determine non-uniform sea level change for any given location.
```
gassmoeller, line 83:
```suggestion
         * Function to determine non-uniform sea level change for any given location.
         * This is a local computation that makes use of the computed sea level offset,
         * so make sure to call compute_sealevel_offset() before calling this function.
```

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+         * Information about the location of topography data files.
+         */
+        std::string data_directory_topography;
+        std::string data_file_name_topography;
+
+        /**
+         * Information about the location of ice height data files.
+         */        
+        std::string data_directory_iceheight;
+        std::string data_file_name_iceheight;
```
gassmoeller, line 104:
```suggestion
        std::string data_directory_ice_height;
        std::string data_file_name_ice_height;
```

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+
+        /**
+         * Pointer to the StructuredDataLookup object that holds the topography  and ice height data.
+         */
+        std::unique_ptr<Utilities::StructuredDataLookup<2>> topography_lookup;
+        std::unique_ptr<Utilities::StructuredDataLookup<2>> iceheight_lookup;
```
gassmoeller, line 110:
```suggestion
        std::unique_ptr<Utilities::StructuredDataLookup<2>> ice_height_lookup;
```

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+
+        /**
+         * description
```
gassmoeller, line 113: todo

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+         */
+        double density_water;
+
+        /**
+         * description
```
gassmoeller, line 118: todo

include/aspect/postprocess/sealevel.h
```diff
@@ -0,0 +1,150 @@
...
+         */
+        double density_ice;
+
+        /**
+         * The sealevel offset.
+         */
+        double sealevel_offset;
```
gassmoeller, line 125:
```suggestion
        double sea_level_offset;
```

source/postprocess/sealevel.cc
```diff
@@ -0,0 +1,655 @@
...
+      // surface, evaluate the topography/geoid displacement/ocean mask/ice height there.
+      for (const auto &cell : this->get_dof_handler().active_cell_iterators())
+        if (cell->is_locally_owned() && cell->at_boundary())
+          {
+            unsigned int face_idx = numbers::invalid_unsigned_int;
+            bool at_upper_surface = false;
+            {
+              for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
+                {
+                  if (cell->at_boundary(f) && cell->face(f)->boundary_id() == this->get_geometry_model().translate_symbolic_boundary_name_to_id("top"))
```
gassmoeller, line 146: Please make the same change here that I did for the geoid in #4223. It does not change the function, but it makes it faster.

source/postprocess/sealevel.cc
```diff
@@ -0,0 +1,655 @@
...
+                                                                         oceanmask_stored_values[i].second.second
+                                                                        });
+
+          // Theta, phi, spherical infinitesimal, and ice height
+          iceheight_spherical_function.emplace_back(std::vector<double> {scoord[2],
+                                                                         scoord[1],
+                                                                         infinitesimal,
+                                                                         iceheight_stored_values[i].second.second
+                                                                        });
+        }
```
gassmoeller, line 243: It looks like you are not using these lines. Maybe remove? Or where do you plan to use them?

gassmoeller (commented):
I am aware this is not finished yet, but I will be away for two weeks and I thought I give some feedback now already. The general structure goes in the right direction, I have one major comment and that is that it seems to me `sea_level` would be a better name for the file (because the name of the plugin in the input file is `sea level` and we usually replace spaces with underscores). The name of the class can remain `SeaLevel` that is what we do in general. Apart from that it would help me with the review if you could extend the documentation of the functions in the header file. I.e. first precisely write down what each function should do. Afterwards you can think about how to exactly implement that.

source/postprocess/sealevel.cc
mfmweerdesteijn: You're right! I do want to write these variables, but I'll figure that out in the execute function

include/aspect/postprocess/sealevel.h
mfmweerdesteijn: Sea_level is better indeed! Because it's sea level, and not sealevel

mfmweerdesteijn:
Thank you for the comments Rene! No need to review the last commit (wip), but I pushed it just in case things would get messy with changing filenames (sealevel to sea_level). I hope you'll have a very nice holiday!:)

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+      statistics.add_value ("Maximum non-uniform sea_level change (m)",
+                            max_nonuniform_sea_level_change);
+      const char *columns[] = { "Minimum non-uniform sea_level change (m)",
+                                "Maximum non-uniform sea_level change (m)"
+                              };
+      for (unsigned int i=0; i<sizeof(columns)/sizeof(columns[0]); ++i)
+        {
+          statistics.set_precision (columns[i], 8);
+          statistics.set_scientific (columns[i], true);
+        }
```
bangerth, line 356: We've got a new style:
```suggestion
      for (const auto &column : columns)
        {
          statistics.set_precision (column, 8);
          statistics.set_scientific (column, true);
        }
```

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+      if (std::isnan(last_output_time))
+        {
+          last_output_time = this->get_time() - output_interval;
+        }
+
+      // Just return stats if text output is not required at all or not needed at this time.
+      if (!write_to_file || ((this->get_time() < last_output_time + output_interval)
+                             && (this->get_timestep_number() != 0)))
+        return std::pair<std::string, std::string> ("Non-uniform sea level change min/max:",
+                                                    output_stats.str());
```
bangerth, line 376:
```suggestion
        return {"Non-uniform sea level change min/max:", output_stats.str()};
```

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+
+      std::string filename = this->get_output_directory() +
+                             "nonuniform_sea_level_change." +
+                             Utilities::int_to_string(this->get_timestep_number(), 5);
+      if (this->get_parameters().run_postprocessors_on_nonlinear_iterations)
+        filename.append("." + Utilities::int_to_string (this->get_nonlinear_iteration(), 4));
```
bangerth, line 382: I *think* (but haven't checked) that you only need the filename in the `if (Utilities::MPI::this_mpi_process(this->get_mpi_communicator()) == 0)` branch below. Move these lines there then.

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+        {
+          // We need to find the last time output was supposed to be written.
+          // This is the last_output_time plus the largest positive multiple
+          // of output_intervals that passed since then. We need to handle the
+          // edge case where last_output_time+output_interval==current_time,
+          // we did an output and std::floor sadly rounds to zero. This is done
+          // by forcing std::floor to round 1.0-eps to 1.0.
+          const double magic = 1.0+2.0*std::numeric_limits<double>::epsilon();
+          last_output_time = last_output_time + std::floor((this->get_time()-last_output_time)/output_interval*magic) * output_interval/magic;
+        }
```
bangerth, line 452: I made a comment above about some other plugin that had broken this piece of functionality into a separate function. Feel free to do as suggested there, or ignore my suggestion.

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+
+      return std::pair<std::string, std::string> ("Non-uniform sea level change min/max:",
+                                                  output_stats.str());
```
bangerth, line 455:
```suggestion
      return {"Non-uniform sea level change min/max:", output_stats.str()};
```

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+                             "located when ASPECT was compiled. This interpretation allows, "
+                             "for example, to reference files located in the `data/' subdirectory "
+                             "of ASPECT.");
+          prm.declare_entry ("Data file name topography", "shell_3d_outer.0.txt",
+                             Patterns::Anything (),
+                             "The file name of the topography ascii data. For the ascii data, "
+                             "provide file in the same format as described in "
+                             "'ascii data' initial composition plugin." );
+
+// The ice height data is not constant over time. It has to be the input data from the current timestep. How to retrieve data from correct timestep with filenumber placeholder? Does it work if you specify e.g. ice_heights_outer.%d.txt? Just like you would in the boundary traction subsection.
```
bangerth, line 501: Did you resolve this question? If not, let's talk about it separately.

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+      const Postprocess::Geoid<dim> &geoid =
+        this->get_postprocess_manager().template get_matching_postprocessor<Postprocess::Geoid<dim>>();
+
+      const double geoid_displacement = geoid.evaluate(position); // (check sign of geoid_displacement)
+      const double topography = this->get_geometry_model().height_above_reference_surface(position);
+
+      Point<dim-1> internal_position;
+      const std::array<double,dim> spherical_position = aspect::Utilities::Coordinates::cartesian_to_spherical_coordinates(position);
+      for (unsigned int d = 1; d < dim; ++d)
+        internal_position[d-1] = spherical_position[d];
```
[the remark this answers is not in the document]
mfmweerdesteijn: Okay, I replaced internal_position with long_colat

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+      const double topography_init = topography_lookup->get_data(internal_position,0);
+      int ocean_mask = 0;
+      if (topography_init > 0.0)
+        ocean_mask = 1;
+      else
+        ocean_mask = 0;
+
+      const double nonuniform_sea_level_change = (geoid_displacement-topography)*ocean_mask;
+
+      return nonuniform_sea_level_change;
```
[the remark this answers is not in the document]
mfmweerdesteijn: I understand this is shorter and does the same. But I find the initial implementation easier to read/understand, because it's more descriptive, and easier to identify which part of the sea level equation this is.

source/postprocess/sea_level.cc
```diff
@@ -0,0 +1,584 @@
...
+    }
+
+
+    template <int dim>
+    std::list<std::string>
+    SeaLevel<dim>::required_other_postprocessors() const
+    {
+      std::list<std::string> deps;
+      deps.emplace_back("geoid");
+      return deps;
```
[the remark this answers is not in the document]
mfmweerdesteijn: Does the required_other_postprocessors function here check that the geoid postprocessor is on?

include/aspect/postprocess/sea_level.h
```diff
@@ -0,0 +1,162 @@
...
+         * is read from the input file and consequently is not part of the
+         * state that needs to be saved and restored.
+         */
+        double output_interval;
+
+        /**
+         * A time (in seconds) at which the last text output was supposed
+         * to be produced. Used to check for the next necessary output time.
+         */
+        double last_output_time;
```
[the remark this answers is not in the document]
mfmweerdesteijn: density_water, density_ice, write_to_file, output_interval are set in parameter file.
sea_level_offset and last_output_time are computed in the sea level postprocessor. so they need to be read again then for checkpointing

source/postprocess/sea_level.cc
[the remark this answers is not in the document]
bangerth: OK, that's fine then.

source/postprocess/sea_level.cc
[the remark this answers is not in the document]
bangerth: No, not this function itself. It just says what other postprocessors this one needs, and then the place that asks for this information will make sure of it.

include/aspect/postprocess/sea_level.h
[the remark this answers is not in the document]
bangerth: OK. The you need essentially the code I copied above that saves/restores these two variables.

mfmweerdesteijn:
@bangerth I think it is done :)

bangerth:
It's still all 27 commits up here on github, though. Did you forget to push them from your local machine?

mfmweerdesteijn:
@bangerth do you mean I should all squash them?

tests/sea_level_postprocessor/screen-output
```diff
@@ -0,0 +1,37 @@
...
+*** Timestep 0:  t=0 years, dt=0 years
+   Solving mesh displacement system... 0 iterations.
+
+   Postprocessing:
+     Compositions min/max/mass:            0/0/0 // 0/0/0 // 0/0/0 // 0/0/0 // 0/0/0 // 0/0/0
+     Temperature min/avg/max:              293 K, 293 K, 293 K
+     RMS, max velocity:                    0 m/year, 0 m/year
+     Topography min/max:                   -2.794e-09 m, 0 m
+     Density at top/bottom of domain:      3300 kg/m^3, 3300 kg/m^3
+     Writing geoid anomaly:                output-sea_level_postprocessor/geoid_anomaly.00000
```
bangerth, line 20: One last request: The postprocessor writes this file to disk, but it isn't part of your test. As a consequence, *what* the postprocessor computes is not actually tested, just that it runs. Could you add the file to `tests/sea_level_postprocessor/` so that this is also tested?

bangerth (commented):
I have one last request -- my apologies for not seeing this any earlier.

bangerth (approved):

PR #3776, opened 2020-08-12
geoid postprocessor working with free surface

mfmweerdesteijn:
Added a function for free_surface_topography_contribution, which is similar to dynamic_topography_contribution, but gets the topography from the geometry model, and doesn't compute CMB geoid contribution.
Also made changes in execute function based on whether dynamic or free surface topography is desired.
Added if statement to postprocessor dependency for when mesh_deformation_enabled == false (dynamic topography postprocessor output required)
For the free surface topo contribution, the topography postprocessor is not required, as the code directly accesses the geometry model.

Now the setup is: you can choose if you want topography contributions to the geoid included. The code then figures out from mesh_deformation_enabled if it needs to take into account the dynamic topography or the free surface topography.

@gassmoeller @jaustermann 
Change of things from Jacky's and Rene's comments:
- To include the free surface CMB contribution, I figured it would make more sense to have one topography_contribution function, because then there is only a small difference if we base it on dynamic topography postprocessor (PP) or the actual mesh in case of a free surface. Only the part is different where we use output from the dynamic topography postprocessor or the actual mesh in case of a free surface. This is evaluated now by this->get_parameters().mesh_deformation_enabled. (needs to be adjusted to check is either top or bottom boundary is active free surface as something similar to suggested by Rene)

- The naming in the code is now very general 'topography'. I added to the topography_contribution function in geoid.h that it is either based on the dyna topo PP or the free surface mesh. So now the input parameter is 'Include the contribution from topography'. I don't see another way to get around the geophysical naming convention, then specifying what 2 options topography can be based on.

- Before I add the bools for if the top or bottom boundary is a free surface as suggested by Rene the following question: if the user wants to include the dynamic topography, it will always include dyna topo from both bottom and top surface (when no mesh deformation of top or bottom surface). Do we want to separate this even further, that you can have 1 contributing surface instead of both? Same for the actual mesh/free surface. But for the free surface, if either top or bottom is the free surface, it will compute the contribution for both surfaces, but one will then be zeroes. So I don't see a reason to separate between those. What do you think?


include/aspect/postprocess/geoid.h
```diff
@@ -94,7 +94,7 @@ namespace aspect
         /**
          * A parameter to control whether to output the spherical harmonic coefficients of the surface dynamic topography contribution
```
jaustermann, line 95: Seems like you almost caught all the 'dynamic's - here's the only one I found that slipped through.

source/postprocess/geoid.cc
```diff
@@ -934,9 +1071,9 @@ namespace aspect
       {
         prm.enter_subsection("Geoid");
         {
-          prm.declare_entry("Include the contributon from dynamic topography", "true",
+          prm.declare_entry("Include the contributon from topography", "true",
```
gassmoeller, line 1074: Oh, changing input parameters will break every parameter file that used this parameter. This is something we try to avoid as much as possible. However, there is a typo in this parameter (contributon), which we have to fix anyway. Could you fix that as well, since you change the line anyway?

source/postprocess/geoid.cc
```diff
@@ -259,71 +249,155 @@ namespace aspect
...
+                        for (unsigned int q=0; q<fe_face_values.n_quadrature_points; ++q)
+                          surface_stored_values.emplace_back (std::make_pair(fe_face_values.quadrature_point(q), std::make_pair(fe_face_values.JxW(q), topo_values[q])));
+                      }
+                  }
+              }
+            else
               {
-                for (unsigned int q=0; q<fe_face_values.n_quadrature_points; ++q)
-                  surface_stored_values.emplace_back (std::make_pair(fe_face_values.quadrature_point(q), std::make_pair(fe_face_values.JxW(q), topo_values[q])));
+                // If the cell is at the top boundary, add its contributions to the topography surface storage vector.
```
jaustermann, line 291: Can you adjust this comment to reflect what you're doing in the next couple of lines?

source/postprocess/geoid.cc
```diff
@@ -259,71 +249,155 @@ namespace aspect
...
+                if (face_idx != numbers::invalid_unsigned_int && at_upper_surface)
```
jaustermann, line 292: I believe if(at_upper_surface) is sufficient.

source/postprocess/geoid.cc
```diff
@@ -259,71 +249,155 @@ namespace aspect
...
+                  {
+                    for (unsigned int q=0; q<fe_face_values.n_quadrature_points; ++q)
+                      {
+                        double force_zero = 0.0;
+                        surface_stored_values.emplace_back (std::make_pair(fe_face_values.quadrature_point(q), std::make_pair(fe_face_values.JxW(q), force_zero)));
```
jaustermann, line 297: You can just write this in directly without making a new variable. 

surface_stored_values.emplace_back (std::make_pair(fe_face_values.quadrature_point(q), std::make_pair(fe_face_values.JxW(q), 0.0)));

source/postprocess/geoid.cc
```diff
@@ -259,71 +249,155 @@ namespace aspect
...
+                      }
+                  }
               }
 
-            // if the cell at bottom boundary, add its contributions dynamic topography storage vector
-            if (face_idx != numbers::invalid_unsigned_int && !at_upper_surface)
+            if (include_CMB_topo_contribution == true)
```
jaustermann, line 302: The same comments as for the surface contribution.

source/postprocess/geoid.cc
```diff
@@ -332,52 +406,69 @@ namespace aspect
...
       // Get the spherical harmonic coefficients of the density contribution.
       std::pair<std::vector<double>,std::vector<double> > SH_density_coes = density_contribution(outer_radius);
 
-      std::pair<double, std::pair<std::vector<double>,std::vector<double> > > SH_surface_dyna_topo_coes;
-      std::pair<double, std::pair<std::vector<double>,std::vector<double> > > SH_CMB_dyna_topo_coes;
-      // Initialize the surface and CMB density contrasts with NaNs becasue they may be unused in case of no dynamic topography contribution.
+      std::pair<double, std::pair<std::vector<double>,std::vector<double> > > SH_surface_topo_coes;
+      std::pair<double, std::pair<std::vector<double>,std::vector<double> > > SH_CMB_topo_coes;
+
+      // Initialize the surface and CMB density contrasts with NaNs becasue they may be unused in case of no topography contribution.
```
jaustermann, line 427: typo in 'because' :)

source/postprocess/geoid.cc
```diff
@@ -889,11 +980,11 @@ namespace aspect
     Geoid<dim>::required_other_postprocessors() const
     {
       std::list<std::string> deps;
-      if (include_dynamic_topo_contribution == true)
-        {
-          deps.emplace_back("dynamic topography");
-          deps.emplace_back("boundary densities");
-        }
+      if ( (include_surface_topo_contribution == true && use_real_surface_topography == false) || (include_CMB_topo_contribution == true && use_real_CMB_topography == false) )
```
jaustermann, line 983: I'm not sure about this one, but could imagine that required_other_postprocessor is called before execute, in which case you don't know yet if use_real_surface_topography and use_real_CMB_topography are true or false. I will defer this one to @gassmoeller.

source/postprocess/geoid.cc
```diff
@@ -934,9 +1025,15 @@ namespace aspect
       {
         prm.enter_subsection("Geoid");
         {
-          prm.declare_entry("Include the contributon from dynamic topography", "true",
+          prm.declare_entry("Include the contribution from topography", "true",
                             Patterns::Bool(),
-                            "Option to include the contribution from dynamic topography on geoid. The default is true.");
+                            "Option to include the contribution from topography on geoid. The default is true.");
```
jaustermann, line 1030: You're not actually using this parameter anymore so you can remove it.

source/postprocess/geoid.cc
```diff
@@ -992,27 +1089,16 @@ namespace aspect
       {
         prm.enter_subsection("Geoid");
         {
-          include_dynamic_topo_contribution = prm.get_bool ("Include the contributon from dynamic topography");
+          include_topo_contribution = prm.get_bool ("Include the contribution from topography");
```
jaustermann, line 1092: Same here, remove this.

include/aspect/postprocess/geoid.h
```diff
@@ -118,9 +118,29 @@ namespace aspect
         double density_below;
 
         /**
-         * A parameter to control whether to include the dynamic topography contribution on geoid
+         * A parameter to control whether to include the topography contribution on geoid
          */
-        bool include_dynamic_topo_contribution;
+        bool include_topo_contribution;
```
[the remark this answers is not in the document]
mfmweerdesteijn: I think we wanted to keep this so that we can throw an error message when someone uses the old parameter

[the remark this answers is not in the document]
jaustermann: Ah yes, got it. In that case you probably also don't want to correct the typo in the parameter name (where you parse it at the end). Maybe add a comment in here about this planned change?

source/postprocess/geoid.cc
mfmweerdesteijn: You are right. I can't use use_real_surface_topography and use_real_CMB_topography. Not sure how to solve this differently

source/postprocess/geoid.cc
```diff
@@ -259,20 +246,99 @@ namespace aspect
...
-            fe_face_values[this->introspection().extractors.temperature].get_function_values(topo_vector, topo_values);
 
-            // if the cell at top boundary, add its contributions dynamic topography storage vector
-            if (face_idx != numbers::invalid_unsigned_int && at_upper_surface)
+            if (include_surface_topo_contribution == true)
               {
-                for (unsigned int q=0; q<fe_face_values.n_quadrature_points; ++q)
-                  surface_stored_values.emplace_back (std::make_pair(fe_face_values.quadrature_point(q), std::make_pair(fe_face_values.JxW(q), topo_values[q])));
+                // If the cell is at the top boundary, add its contributions to the topography surface storage vector.
+                if (at_upper_surface)
```
gassmoeller, line 253: I think this function would be easier if you reverted the order of these if-conditions. Currently you have the following:
```
if (include_contribution){
  if (at_upper_surface){
    if (use_real_surface)
    else
  }
}
else
  if (at_upper_surface){
}
```
This would be simpler if you would do:
```
if (at_upper_surface){
  if (include_contribution)
    if (use_real_surface)
    else
  else
}
```

And the same for the CMB block below.

source/postprocess/geoid.cc
```diff
@@ -524,116 +609,114 @@ namespace aspect
...
+          // Have a stream into which we write the SH coefficients data from surface topography contribution.
+          // The text stream is then later sent to processor 0.
+          std::ostringstream output_surface_topo_contribution_SH_coes;
 
-              // Prepare the output SH coefficients data from surface dynamic topography contribution.
-              unsigned int SH_coes_ind = 0;
-              for (unsigned int ideg =  min_degree; ideg < max_degree+1; ideg++)
+          // Prepare the output SH coefficients data from surface topography contribution.
+          unsigned int SH_coes_ind = 0;
+          for (unsigned int ideg =  min_degree; ideg < max_degree+1; ideg++)
```
gassmoeller, line 626: please change `ideg++` to `++ideg`. The details of this change are not relevant here (but if you want you can look them up as pre-increment and post-increment operators), but since you changed the line, please fix this as well here and in the line below.

source/postprocess/geoid.cc
```diff
@@ -524,116 +609,114 @@ namespace aspect
...
+          std::ostringstream output_CMB_topo_contribution_SH_coes;
 
-              // Prepare the output SH coefficients data from CMB dynamic topography contribution.
-              unsigned int SH_coes_ind = 0;
-              for (unsigned int ideg =  min_degree; ideg < max_degree+1; ideg++)
+          // Prepare the output SH coefficients data from CMB topography contribution.
+          unsigned int SH_coes_ind = 0;
+          for (unsigned int ideg =  min_degree; ideg < max_degree+1; ideg++)
+            {
+              for (unsigned int iord = 0; iord < ideg+1; iord++)
```
gassmoeller, line 676: same comment about changing `x++` into `++x`

source/postprocess/geoid.cc
gassmoeller: My suggestion above will fix this. Both `parse_parameters` and `initialize` are called before `required_other_postprocessors`.

source/postprocess/geoid.cc
```diff
@@ -931,9 +1015,16 @@ namespace aspect
       {
         prm.enter_subsection("Geoid");
         {
+          // This parameter doesn't exist anymore (incl. spelling error). TODO: throw error message when used.
```
gassmoeller, line 1018: No need to have this comment here. This line only tells ASPECT that it is ok to find this parameter in the input file.

source/postprocess/geoid.cc
```diff
@@ -931,9 +1015,16 @@ namespace aspect
...
           prm.declare_entry("Include the contributon from dynamic topography", "true",
                             Patterns::Bool(),
                             "Option to include the contribution from dynamic topography on geoid. The default is true.");
+          prm.declare_entry("Include the contribution from surface topography", "true",
+                            Patterns::Bool(),
+                            "Option to include the contribution from surface topography on geoid. The default is true.");
+          prm.declare_entry("Include the contribution from CMB topography", "true",
+                            Patterns::Bool(),
+                            "Option to include the contribution from CMB topography on geoid. The default is true.");
```
gassmoeller, line 1027: This is our chance to improve the naming scheme of the input variables. Here is my suggestion: `Include surface topography contribution` and `Include CMB topography contribution`.

source/postprocess/geoid.cc
```diff
@@ -957,13 +1048,13 @@ namespace aspect
...
+          prm.declare_entry("Also output the spherical harmonic coefficients of surface topography contribution", "false",
                             Patterns::Bool(),
-                            "Option to also output the spherical harmonic coefficients of the surface dynamic topography contribution "
+                            "Option to also output the spherical harmonic coefficients of the surface topography contribution "
                             "to the maximum degree. The default is false. ");
-          prm.declare_entry("Also output the spherical harmonic coefficients of CMB dynamic topography contribution", "false",
+          prm.declare_entry("Also output the spherical harmonic coefficients of CMB topography contribution", "false",
                             Patterns::Bool(),
-                            "Option to also output the spherical harmonic coefficients of the CMB dynamic topography contribution "
+                            "Option to also output the spherical harmonic coefficients of the CMB topography contribution "
```
gassmoeller, line 1057: Changing the names of input parameters will break every input file that uses them, which is why we do this very rarely. Here are two options: 
(1) Leave them unchanged and accept the `dynamic` in the parameter. 
(2) Rename them, and create an alias for the old parameter (`prm.declare_alias(new_parameter_name, old_parameter_name);`). If you decide to go this route, we should come up with better names. E.g. `Output geoid anomaly coefficients`, `Output surface topography contribution coefficients`, `Output CMB topography contribution coefficients`, `Output density anomaly contribution coefficients`, and `Output gravity anomaly coefficients`.
I leave the choice between (1) and (2) up to you.

source/postprocess/geoid.cc
```diff
@@ -989,27 +1080,17 @@ namespace aspect
       {
         prm.enter_subsection("Geoid");
         {
-          include_dynamic_topo_contribution = prm.get_bool ("Include the contributon from dynamic topography");
+          include_topo_contribution = prm.get_bool ("Include the contributon from dynamic topography"); // This parameter doesn't exist anymore (incl. spelling error). TODO: throw error message when used.
```
gassmoeller, line 1083: If you remove the declaration of this parameter from the header file you have to declare it here: `const bool include_topo_contribution = ...`.

Add the following line to address the comment (you may need to wrap that over several lines):
```
AssertThrow (include_topo_contribution == true, 
                      ExcMessage("The parameter 'Include the contributon from dynamic topography' has been replaced by the two parameters 'Include the contribution from surface topography' and 'Include the contribution from CMB topography', please use them instead.);
```

source/postprocess/geoid.cc
mfmweerdesteijn: should all ideg++ be changed to ++ideg?

source/postprocess/geoid.cc
mfmweerdesteijn: Im going for 2 :)

mfmweerdesteijn:
@gassmoeller  Hi rene, I could really use your help. I addressed the comments, and aspect compiles on my machine (well, virtual machine), and then I pushed it to here. But there is a conflict. However, when I followed the rebase procedure, I could resolve the conflict, but the code outside of the conflict was not my most recent version. This is what I did (first git add and git commit):
1. git checkout master
2. git pull upstream master
3. git checkout mybranch
4. git rebase master
5. open qtcreator to address conflict (but then problem as described above)
(6. git rebase --abort)

Another try:
1. git merge-base master mybranch
2. git rebase -i 'commit ID'
3. change 2nd commit (address comments) pick to squash
4. git rebase master
5. open qtcreator to address conflict: but again problem as described above)
(6. git rebase --abort)

Also if I check the conflict here on github at 'Resolve conflicts', it shows the 3 conflicting lines, but everything outside of the conflict is not my code version which I want be pushed.

mfmweerdesteijn:
If code approved like this, will now work on test case and adding changelog entry

source/postprocess/geoid.cc
```diff
@@ -524,116 +591,114 @@ namespace aspect
...
+          std::ostringstream output_surface_topo_contribution_SH_coes;
 
-              // Prepare the output SH coefficients data from surface dynamic topography contribution.
-              unsigned int SH_coes_ind = 0;
-              for (unsigned int ideg =  min_degree; ideg < max_degree+1; ideg++)
+          // Prepare the output SH coefficients data from surface topography contribution.
+          unsigned int SH_coes_ind = 0;
+          for (unsigned int ideg =  min_degree; ideg < max_degree+1; ++ideg)
+            {
+              for (unsigned int iord = 0; iord < ideg+1; iord++)
```
gassmoeller, line 610:
```suggestion
              for (unsigned int iord = 0; iord < ideg+1; ++iord)
```

source/postprocess/geoid.cc
gassmoeller: yes all places

source/postprocess/geoid.cc
```diff
@@ -201,52 +195,46 @@ namespace aspect
...
       for (const auto &cell : this->get_dof_handler().active_cell_iterators())
         if (cell->is_locally_owned() && cell->at_boundary())
           {
-            // see if the cell is at the *top* boundary or CMB, not just any boundary
             unsigned int face_idx = numbers::invalid_unsigned_int;
             bool at_upper_surface = false;
             {
               for (unsigned int f=0; f<GeometryInfo<3>::faces_per_cell; ++f)
                 {
                   if (cell->at_boundary(f) && this->get_geometry_model().depth (cell->face(f)->center()) < cell->face(f)->minimum_vertex_distance()/3)
```
gassmoeller, line 212: I just realized that this condition will no longer work once you have a real free surface. Currently the postprocessor checks if the midpoint of the current cell face is close to a depth of 0 (for top) or the maximal_depth (CMB) to check if it is the top or bottom surface. This condition will need to be replaced by the following (which is better anyway):

```suggestion
                  if (cell->at_boundary(f) && cell->face(f)->boundary_id() == this->get_geometry_model().translate_symbolic_boundary_name_to_id("top"))
```

or the following for the CMB check: 


```
                  if (cell->at_boundary(f) && cell->face(f)->boundary_id() == this->get_geometry_model().translate_symbolic_boundary_name_to_id("bottom"))
```

Please have a look through the postprocessor and replace the old conditions with the new ones (I think there are 4 places that need to be replaced). Otherwise you may see wrong results from the postprocessor with a free surface.

mfmweerdesteijn: Thanks for your quick reviewing!

mfmweerdesteijn:
@gassmoeller  I did something wrong with "git pull upstream master" today (oct 28). I did it accidentally on this branch and not on my master branch. Does it have consequences?

gassmoeller:
Merging master into your branch can create problems later. As long as you have no uncommitted changes on your branch and you have made no commits after the merge you can undo it by resetting your branch to the last commit you made before the merge. Assuming you do not have local commits that are not already on github (shown in this PR) the following would do the trick:
```
git reset --hard 5ebd2b6
```
If you have local commits before the merge: Find the newest commit ID before the merge and replace 5ebd2b6 with that ID.
If you have local commits after the merge: Let me know, in this case we need to shuffle commits around with an interactive rebase.

bobmyhill:
Hi @mfmweerdesteijn, did you need more help with this one? 
If 47369ba is the dodgy merge commit (and as long as you didn't make any more local commits that we can't see on github), the easiest solution is probably:
- `git reset --hard 8691633`
- Re-remove "coefficients" from the parameter name "Output gravity anomaly coefficients" in source/postprocess/geoid.cc (redoing the three-line change you made in 5ebd2b6).
- `git add -u`
- `git commit -m "Changing 1 parameter name: actual gravity anomaly, not SH coefficients"`

That avoids an interactive rebase. You will still have to rebase to the new master, as geoid.cc has changed in the last few months.

mfmweerdesteijn:
@bobmyhill Hey Bob, thank you so much for these instructions! I performed your bullet points. This new commit now has three lines changed in 1 file (geoid.cc). I didn't make other changes on my local branch since october, so that's good. 
"You will still have to rebase to the new master, as geoid.cc has changed in the last few months." It's been a while I've done anything with git commands and I'm a bit lost. Does this mean I have to do the following? :
- git checkout master
- git pull upstream master
- git checkout -b 'my local branch' (I named it geoid_postprocessor_with_free_surface)
- git push origin geoid_postprocessor_with_free_surface
Or should I have done the 'git pull upstream master' before I created this new commit as in your bullet points?

bobmyhill:
Hi @mfmweerdesteijn,

Sure, no problem! To rebase interactively, your first three commands were correct, but before you push, you need to run the rebase:
- git checkout master
- git pull upstream master
- git checkout -b geoid_posprocessor_with_free_surface
- *git rebase -i master*

During the rebase, you'll need to manually edit geoid.cc. git modifies your file to include both the changes in master and the changes from geoid_posprocessor_with_free_surface. You'll get one or more sections that look like this:

`<<<<<<< geoid_posprocessor_with_free_surface`
`                <changes to the file in geoid_posprocessor_with_free_surface>`
` =======`
`                 <changes to the file in master>`
` >>>>>>> master`

Sometimes you can just choose one of the snippets inside and delete the rest (or some IDEs allow you to click on the section you want to keep). You might have to combine changes from both branches.

Once you've made the changes, you can type
- git add -u
- git rebase --continue

You may have to repeat the process for more commits until git tells you the rebase is complete. Finally:
- git push origin geoid_posprocessor_with_free_surface --force
will finish the process and push the new files up to github.

mfmweerdesteijn:
Hey @bobmyhill, very clear explanation, thanks! I think I did it correctly now:) I changed things manually during the rebase, made sure that past changes from the master are included here (only a few lines and minor changes). 

I left a few lines starting with // in with the old code lines (the lines on my branch that I had to adjust for the past changes made on master branch). These can be deleted of course, but since I'm not very proficient with git and version control etc. it felt like a safe thing to do for now;)

Thank again! Now it feels like less of an obstacle to continue with this again

bobmyhill:
Hey @mfmweerdesteijn, great! It looks like the rebase was successful. It was sensible to leave those commented lines in, at least until you've tested the code on your own machine. Sometimes rebasing can be fiddly.

Let us know when you want another review.

gassmoeller (commented):
I think this PR is still in pretty good shape, we just need a test case. Let me know when you have a model that uses a free surface and this postprocessor and I can take another look. I also found some superfluous lines.

include/aspect/postprocess/geoid.h
```diff
@@ -118,9 +123,24 @@ namespace aspect
...
+
+        /**
+         * A parameter to control whether to include the CMB topography contribution on geoid
+         */
+        bool include_CMB_topo_contribution;
+
+        /**
+         * A parameter to specify if the top boundary is an active free surface
+         */
+        bool use_real_surface_topography;
```
jaustermann, line 138: I would suggest changing this variable name to 'use_free_surface_topography' and below 'use_free_CMB_topography'. "Real" sounds to me like it's Earth-like, which it doesn't have to be so maybe 'free' is slightly more descriptive.

include/aspect/postprocess/geoid.h
```diff
@@ -118,9 +123,24 @@ namespace aspect
...
+
+        /**
+         * A parameter to control whether to include the CMB topography contribution on geoid
+         */
+        bool include_CMB_topo_contribution;
+
+        /**
+         * A parameter to specify if the top boundary is an active free surface
+         */
+        bool use_freesurface_surface_topography;
```
gassmoeller, line 138: Please use @jaustermann 's suggested name instead:
```suggestion
        bool use_free_surface_topography;
```

include/aspect/postprocess/geoid.h
```diff
@@ -118,9 +123,24 @@ namespace aspect
...
+
+        /**
+         * A parameter to specify if the bottom boundary is an active free surface
          */
-        bool include_dynamic_topo_contribution;
+        bool use_freesurface_CMB_topography;
```
gassmoeller, line 143:
```suggestion
        bool use_free_CMB_topography;
```

include/aspect/postprocess/geoid.h
mfmweerdesteijn: Sure? I think this naming 'use_free_surface_topography' is not so intuitive. Because it looks like ' use free surface' but that's only for the surface

mfmweerdesteijn:
@gassmoeller It didn't fail the geoid_freesurface test, so there's no 'diff' files to apply locally. Isn't it strange that it didn't fail? Also, do you want me to 'fixup' the last commits to make the PR look clean (Wolfgang explained to me how to do that:) )

gassmoeller:
Ready to go, thanks for the fixes and sticking with us through all the long review process! Just waiting for the last tester to finish.

@mfmweerdesteijn: That is alright, it is not necessary that the test fails, sometimes a test is just not sensitive to differences between different systems (or all library versions on your system are sufficiently similar to the reference tester that is produces the same results). I can see that the test was executed and passed, so no need to worry.

@bobmyhill and @bangerth: FYI this PR reverts the typo fix of #3808, because it introduces two new parameters that replace the parameter with the typo. We kept the old one in for backwards compatibility of prm files, but we assert that it is set to the default value, i.e. it is not and cannot longer be actively used. We can then remove the old parameter in a future release.

PR #3778, opened 2020-08-12
Add syntax highlighting for prm files in vim editor

jperryhouts:
Added a directory for syntax highlighting at `contrib/utilities/syntax_highlighting`, which could be expanded with highlighting macros for other editors.

Note: I sometimes setup batch runs by preprocessing a parameter file with a tool called pyexpander, which allows embedding arbitrary python code. This syntax file contains handlers for those embedded python environments, which presumably is not necessary for most (any?) other users. But that also doesn't impact the regular functionality. I left them in here, because I don't see an obvious benefit to removing them. Let me know if you object to that.

### For all pull requests:


tjhei:
@bangerth should we put this into deal.II contrib/styles/ instead?

bangerth:
That would be my preference. If you want to make it easier for ASPECT users, we could leave a stub directory that in a README.md file just says to look <elsewhere>.

jperryhouts:
:+1:

Issue #3774, opened 2020-08-12
C++11: Use range-based for loops


bangerth:
We have a lot of places of the form
```
      typename DoFHandler<dim>::active_cell_iterator cell = mesh_deformation_dof_handler.begin_active(),
                                                     endc= mesh_deformation_dof_handler.end();
      for (; cell!=endc; ++cell)
        ...
```
But using C++11 features, we can more easily write this as
```
for (const auto &cell : this->get_dof_handler().active_cell_iterators())
```

Similarly, there are plenty of places of the form (or in many variations -- searching for `for... ::const_iterator` will yield many)
```
      for (typename std::list<std::unique_ptr<Interface<dim> > >::const_iterator
           p = mesh_refinement_objects.begin();
           p != mesh_refinement_objects.end(); ++p)
```
that could now be written as
```
  for (const auto p : mesh_refinement_objects)
```

All of these might make the code simpler to read.

pmbremner:
@bangerth I see you have two merged pull requests related to this. Has this issue been addressed?

bangerth:
It seems to me like the majority of places has been converted. But there are a few places one could still touch:
```
> egrep -n -r 'for.*typename' source/
source/postprocess/heating_statistics.cc:100:            for (typename std::list<std::unique_ptr<HeatingModel::Interface<dim>>>::const_iterator
source/postprocess/heating_statistics.cc:122:      for (typename std::list<std::unique_ptr<HeatingModel::Interface<dim>>>::const_iterator
source/postprocess/visualization.cc:1176:              for (typename std::list<typename aspect::internal::Plugins::PluginList<VisualizationPostprocessors::Interface<dim>>::PluginInfo>::const_iterator
source/postprocess/visualization/heating.cc:153:        for (typename std::list<std::unique_ptr<HeatingModel::Interface<dim>>>::const_iterator
source/postprocess/interface.cc:234:          for (typename std::list<typename aspect::internal::Plugins::PluginList<Interface<dim>>::PluginInfo>::const_iterator
source/simulator/core.cc:1626:            for (typename Triangulation<dim>::active_cell_iterator
source/simulator/core.cc:1641:          for (typename Triangulation<dim>::active_cell_iterator
source/termination_criteria/interface.cc:87:      for (typename std::list<std::unique_ptr<Interface<dim>>>::const_iterator
[...]

PR #3782, opened 2020-08-13
Use range-based for loops in a variety of places.

bangerth:
Addresses a number of those places that could benefit from #3774 . Required some wicked perl magic + more manual labor than I wanted this to be :-)

I've put thought into whether to capture the variables by-value or by-reference. I've generally chosen references if it's not a scalar, but by value if it's an integer of some type.

### DIFF
--- a/source/mesh_refinement/interface.cc
+++ b/source/mesh_refinement/interface.cc
@@ -91,13 +91,11 @@ namespace aspect
       Assert (mesh_refinement_objects.size() > 0, ExcInternalError());
 
       // call the update() functions of all refinement plugins.
-      for (typename std::list<std::unique_ptr<Interface<dim> > >::const_iterator
-           p = mesh_refinement_objects.begin();
-           p != mesh_refinement_objects.end(); ++p)
+      for (const auto &p : mesh_refinement_objects)
         {
           try
             {
-              (*p)->update ();
+              p->update ();
             }
 
           // plugins that throw exceptions usually do not result in
@@ -114,7 +112,7 @@ namespace aspect
               std::cerr << "Exception on MPI process <"
                         << Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)
                         << "> while running mesh refinement plugin <"
-                        << typeid(**p).name()
+                        << typeid(*p).name()
                         << ">: " << std::endl
                         << exc.what() << std::endl
                         << "Aborting!" << std::endl
@@ -132,7 +130,7 @@ namespace aspect
               std::cerr << "Exception on MPI process <"
                         << Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)
                         << "> while running mesh refinement plugin <"
-                        << typeid(**p).name()
+                        << typeid(*p).name()
                         << ">: " << std::endl;
               std::cerr << "Unknown exception!" << std::endl
                         << "Aborting!" << std::endl
@@ -265,13 +263,11 @@ namespace aspect
 
       // call the tag_additional_cells() functions of all
       // plugins we have here in turns.
-      for (typename std::list<std::unique_ptr<Interface<dim> > >::const_iterator
-           p = mesh_refinement_objects.begin();
-           p != mesh_refinement_objects.end(); ++p)
+      for (const auto &p : mesh_refinement_objects)
         {
           try
             {
-              (*p)->tag_additional_cells ();
+              p->tag_additional_cells ();
             }
 
           // plugins that throw exceptions usually do not result in
@@ -288,7 +284,7 @@ namespace aspect
               std::cerr << "Exception on MPI process <"
                         << Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)
                         << "> while running mesh refinement plugin <"
-                        << typeid(**p).name()
+                        << typeid(*p).name()
                         << ">: " << std::endl
                         << exc.what() << std::endl
                         << "Aborting!" << std::endl
@@ -306,7 +302,7 @@ namespace aspect
               std::cerr << "Exception on MPI process <"
                         << Utilities::MPI::this_mpi_process(MPI_COMM_WORLD)
                         << "> while running mesh refinement plugin <"
-                        << typeid(**p).name()
+                        << typeid(*p).name()
                         << ">: " << std::endl;
               std::cerr << "Unknown exception!" << std::endl
                         << "Aborting!" << std::endl
--- a/source/particle/property/interface.cc
+++ b/source/particle/property/interface.cc
@@ -306,18 +306,16 @@ namespace aspect
         std::vector<std::vector<std::pair<std::string, unsigned int> > > info;
 
         // Get the property information of the selected plugins
-        for (typename std::list<std::unique_ptr<Interface<dim> > >::const_iterator
-             p = property_list.begin(); p!=property_list.end(); ++p)
+        for (const auto &p : property_list)
           {
-            info.push_back((*p)->get_property_information());
+            info.push_back(p->get_property_information());
           }
 
         // Initialize our property information
         property_information = ParticlePropertyInformation(info);
-        for (typename std::list<std::unique_ptr<Interface<dim> > >::const_iterator
-             p = property_list.begin(); p!=property_list.end(); ++p)
+        for (const auto &p : property_list)
           {
-            (*p)->initialize();
+            p->initialize();
           }
       }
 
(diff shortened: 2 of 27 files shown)


bangerth:
Oh, it breaks 260 tests... This will have to wait for a few days, I'm afraid :-(

tjhei:
> Oh, it breaks 260 tests...

candidate for the bucket award? :-)

tjhei:
Fascinating. How on earth can range-based for loops break the tests?

tjhei:
This looks like you introduced a bug:
```

##21      <==      Heat fluxes through boundary parts: 1.176e+06 W, -4.157e+06 W, 0 W, 0 W
##21      ==>      Heat fluxes through boundary parts: 1.176e+06 W, 1.176e+06 W, 1.176e+06 W, 1.176e+06 W, 

```

source/postprocess/heat_flux_statistics.cc
```diff
@@ -61,37 +61,31 @@ namespace aspect
...
         Utilities::MPI::sum (local_values, this->get_mpi_communicator(), global_values);
 
         // and now take them apart into the global map again
         unsigned int index = 0;
-        for (std::set<types::boundary_id>::const_iterator
-             p = boundary_indicators.begin();
-             p != boundary_indicators.end(); ++p, ++index)
-          global_boundary_fluxes[*p] = global_values[index];
+        for (const auto p : boundary_indicators)
+          global_boundary_fluxes[p] = global_values[index];
```
tjhei, line 74: and this seems to be a bug.

source/postprocess/heat_flux_statistics.cc
```diff
@@ -61,37 +61,31 @@ namespace aspect
...
       }
 
       // now add all of the computed heat fluxes to the statistics object
       // and create a single string that can be output to the screen
       std::ostringstream screen_text;
       unsigned int index = 0;
-      for (std::map<types::boundary_id, double>::const_iterator
-           p = global_boundary_fluxes.begin();
-           p != global_boundary_fluxes.end(); ++p, ++index)
```
tjhei, line 87: same here

bangerth:
On 8/13/20 8:39 AM, Timo Heister wrote:
> candidate for the bucket award? :-)

Yep, I'm afraid so :-(

source/mesh_refinement/interface.cc
```diff
@@ -159,13 +157,11 @@ namespace aspect
       std::vector<Vector<float> > all_error_indicators (mesh_refinement_objects.size(),
                                                         Vector<float>(error_indicators.size()));
       unsigned int index = 0;
-      for (typename std::list<std::unique_ptr<Interface<dim> > >::const_iterator
-           p = mesh_refinement_objects.begin();
-           p != mesh_refinement_objects.end(); ++p, ++index)
```
tjhei, line 164: this is incorrect

source/particle/property/interface.cc
```diff
@@ -365,10 +362,9 @@ namespace aspect
         particle_properties.reserve(property_information.n_components());
 
         unsigned int property_index = 0;
-        for (typename std::list<std::unique_ptr<Interface<dim> > >::const_iterator
-             p = property_list.begin(); p!=property_list.end(); ++p, ++property_index)
```
tjhei, line 369: this is incorrect

source/postprocess/heat_flux_densities.cc
```diff
@@ -93,16 +93,14 @@ namespace aspect
       // and create a single string that can be output to the screen
       std::ostringstream screen_text;
       unsigned int index = 0;
-      for (std::map<types::boundary_id, double>::const_iterator
-           p = global_boundary_flux_densities.begin();
-           p != global_boundary_flux_densities.end(); ++p, ++index)
```
tjhei, line 98: this is incorrect

source/postprocess/heating_statistics.cc
```diff
@@ -100,11 +100,9 @@ namespace aspect
               local_mass += out.densities[q] * fe_values.JxW(q);
 
             unsigned int index = 0;
-            for (typename std::list<std::unique_ptr<HeatingModel::Interface<dim> > >::const_iterator
-                 heating_model = heating_model_objects.begin();
-                 heating_model != heating_model_objects.end(); ++heating_model, ++index)
```
tjhei, line 105: this is incorrect

source/postprocess/mass_flux_statistics.cc
```diff
@@ -104,37 +104,31 @@ namespace aspect
...
         // then collect contributions from all processors
         std::vector<double> global_values (local_values.size());
         Utilities::MPI::sum (local_values, this->get_mpi_communicator(), global_values);
 
         // and now take them apart into the global map again
         unsigned int index = 0;
-        for (std::set<types::boundary_id>::const_iterator
-             p = boundary_indicators.begin();
-             p != boundary_indicators.end(); ++p, ++index)
-          global_boundary_fluxes[*p] = global_values[index];
```
tjhei, line 121: this is incorrect

source/postprocess/mass_flux_statistics.cc
```diff
@@ -104,37 +104,31 @@ namespace aspect
...
       }
 
       // now add all of the computed mass fluxes to the statistics object
       // and create a single string that can be output to the screen
       std::ostringstream screen_text;
       unsigned int index = 0;
-      for (std::map<types::boundary_id, double>::const_iterator
-           p = global_boundary_fluxes.begin();
-           p != global_boundary_fluxes.end(); ++p, ++index)
+      for (const auto &p : global_boundary_fluxes)
```
tjhei, line 124: this is incorrect

source/postprocess/velocity_boundary_statistics.cc
```diff
@@ -143,13 +139,11 @@ namespace aspect
         // and now take them apart into the global map again
         // At the same time, calculate the rms velocity for each boundary
         unsigned int index = 0;
-        for (std::set<types::boundary_id>::const_iterator
-             p = boundary_indicators.begin();
-             p != boundary_indicators.end(); ++p, ++index)
+        for (const auto p : boundary_indicators)
```
tjhei, line 142: this is incorrect

source/postprocess/visualization/heating.cc
```diff
@@ -144,11 +144,9 @@ namespace aspect
           AssertThrow(false, ExcNotImplemented());
 
         unsigned int index = 0;
-        for (typename std::list<std::unique_ptr<HeatingModel::Interface<dim> > >::const_iterator
-             heating_model = heating_model_objects.begin();
-             heating_model != heating_model_objects.end(); ++heating_model, ++index)
+        for (const auto &heating_model : heating_model_objects)
```
tjhei, line 147: here you can delete the index.

bangerth:
Thanks for your help with finding all of these places. Let's see how this is going to shake out now. If I had been a bit smarter, I wouldn't have deleted everything after the `for (...variable = abc.begin(); ...` and instead checked whether there is also an `index` variable somewhere...

bangerth:
Only 31 tests failing now :-)

bangerth:
Down to 19 failures, out of 811. Surely that's good enough to merge, right? I mean, that's a 97.66% success rate!

source/termination_criteria/interface.cc
```diff
@@ -90,13 +88,11 @@ namespace aspect
       // call the execute() functions of all plugins we have
       // here in turns.
       std::list<std::string>::const_iterator  itn = termination_obj_names.begin();
-      for (typename std::list<std::unique_ptr<Interface<dim> > >::const_iterator
-           p = termination_objects.begin();
-           p != termination_objects.end(); ++p,++itn)
```
tjhei, line 95: here is another bug.

bangerth: Thanks. You don't have to find these bugs for me, though I very much appreciate that you do :-)

bangerth:
Oh, now, look at this: It actually works! Although I will 'fess up to the fact that that took far more work than I actually wanted to put into this patch...

tjhei:
So, did you just exclude the places where you made mistakes before?

bangerth:
On 8/15/20 1:36 PM, Timo Heister wrote:
> So, did you just exclude the places where you made mistakes before?

Yes. There is really not very much one can do about the issue of iterating 
over a collection while also incrementing an index: This can't easily be fit 
into a range-based for loop unless it's a random-access collection for which 
one could use |operator-|. (There is boost's zip-iterator, but that seemed 
like a step backward in readability.)

PR #3787, opened 2020-08-13
Newton fail test fix

MFraters:
This is a proposed fix for the problem found in #3705. I made a new pull request out of it to allow me to easily change the code.

In short, this introduces a test of whether the rhs has dropped more than 1e-8 (I could make this a parameter if desired).

While fixing this, I noticed that to get it working I had to enable rebuilding of the Newton stokes matrix every 0th nonlinear iteration. It doesn't seem to change anything for the other tests (as far as I could see locally) because they already did that due to their boundary conditions.

### DIFF
--- a/source/simulator/solver_schemes.cc
+++ b/source/simulator/solver_schemes.cc
@@ -409,6 +409,7 @@ namespace aspect
       {
         dcr.initial_residual = compute_initial_newton_residual(linearized_stokes_initial_guess);
         dcr.switch_initial_residual = dcr.initial_residual;
+        dcr.residual_old = dcr.initial_residual;
         dcr.residual = dcr.initial_residual;
       }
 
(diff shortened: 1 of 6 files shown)


source/simulator/solver_schemes.cc
```diff
@@ -440,7 +441,7 @@ namespace aspect
     // don't need to force assembly of the matrix.
     if (stokes_matrix_depends_on_solution()
         ||
-        (nonlinear_iteration == 0 && boundary_velocity_manager.get_active_boundary_velocity_conditions().size() > 0))
+        (nonlinear_iteration == 0 ))
```
bangerth, line 444: Is this an accident? I think that the original condition should have had an `||` instead of `&&` -- that strikes me as a bug in the original code. Or do I misunderstand the situation?

source/simulator/solver_schemes.cc
```diff
@@ -449,15 +450,22 @@ namespace aspect
...
 
+    // recompute rhs
+    dcr.velocity_residual = system_rhs.block(introspection.block_indices.velocities).l2_norm();
+    dcr.pressure_residual = system_rhs.block(introspection.block_indices.pressure).l2_norm();
+    dcr.residual = std::sqrt(dcr.velocity_residual * dcr.velocity_residual + dcr.pressure_residual * dcr.pressure_residual);
+
+    // Test whether the rhs has dropped so much that we can assume that the iteration is done.
+    if (dcr.residual < dcr.residual_old * 1e-8)
+      {
+        return;
```
bangerth, line 461: Could you output some status message here? Say
```suggestion
        pcout << "    Nonlinear residual is small; skipping Stokes solve" << std::endl;
        return;
```

MFraters:
Thanks for the review, I think that I have addressed your comments.

MFraters:
I have changed the output a little bit to be a bit more descriptive and I updated the test results.

source/simulator/solver_schemes.cc
bangerth: I don't think that your new message text reflects what is happening well. You are saying that "it has been very large", but that's not the point -- it's just been the regular nonlinear residual. The point is that it is vastly smaller this time around, and that that's likely because the model is (more or less) a linear model that is adequately solved in one iteration, and you're now in the next iteration. I think that my text reflected that better: it's about being small this time; it may or may not have been large last time around, we really neither know nor care.

bangerth: Related, what you print in parentheses is not actually the old residual that you reference when you say "has been very large"), but the ratio. It's confusing to see a statement "has been very large (1.234e-13)" I think :-)

MFraters: hmm, isn't the point that the reduction in the residual has been very large? Do you have a better way of phrasing it? Should I just state that the reduction in the residual has been very large, and/or do you want me to show a different number?

bangerth: I think I misread the text. I though you had said "Nonlinear residual has been very large", but you're saying "Nonlinear residual *reduction* has been very large. Yes, now it makes sense to me.

bangerth:
OK to merge if it tests ok.

PR #3789, opened 2020-08-13
Fix a whitespace problem.

bangerth:
I was too quick with merging #3786 and only realized that I had missed a missing whitespace once I had hit the merge button. Fixed hereby -- @MFraters, FYI.


bangerth:
While there, why not also mark up formulas properly.

PR #3709, opened 2020-08-06
Optimize the newton solver scheme rebuild flags.

MFraters:
splitting up pull request #3701 part 2. This prevents rebuilding matrices when not needed.

PR #3790, opened 2020-08-14
Always rebuild Newton stokes matrix on first nonlinear iteration

MFraters:
In both pull request #3112 and #3787 I ran into the issue that the optimalizations in #3709 may have set some flags too strict. This pull request makes sure that the matrix is always rebuild on the first nonlinear iteration. We may be able to set it more strict again after #3112 and #3787 are merged and have a better idea of when exactly a rebuild (and of what) is needed.

### DIFF
--- a/source/simulator/solver_schemes.cc
+++ b/source/simulator/solver_schemes.cc
@@ -440,7 +440,7 @@ namespace aspect
     // don't need to force assembly of the matrix.
     if (stokes_matrix_depends_on_solution()
         ||
-        (nonlinear_iteration == 0 && boundary_velocity_manager.get_active_boundary_velocity_conditions().size() > 0))
+        nonlinear_iteration == 0)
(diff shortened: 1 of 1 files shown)


source/simulator/solver_schemes.cc
```diff
@@ -440,7 +440,7 @@ namespace aspect
     // don't need to force assembly of the matrix.
     if (stokes_matrix_depends_on_solution()
         ||
-        (nonlinear_iteration == 0 && boundary_velocity_manager.get_active_boundary_velocity_conditions().size() > 0))
+        nonlinear_iteration == 0)
```
bangerth, line 443: Is this an accident? I think that the original condition should have had an || instead of && -- that strikes me as a bug in the original code. Or do I misunderstand the situation?

bangerth: Hm, on second thought, I think you're right.

PR #3794, opened 2020-08-14
Update class documentation.

bangerth:
The text is a bit terse :-)


bangerth:
On second thought (and some code digging), it turns out that this class actually supports 3d. So I removed the two notes that it only supports 2d.

PR #3759, opened 2020-08-11
added new strain rate and derivative functions to diffusion and dislocation creep

bobmyhill:
This PR adds `strain rate(P, T, C, stress)` and `d(strain rate)/d(stress) (P, T, C, stress)` functions to diffusion and dislocation creep. These stress-dependent functions are needed for correct compositing (via a Newton-Raphson iteration). 

At the moment, diffusion_dislocation.cc contains these functions, but more composite rheologies are now being considered, and each of these will need the same functions.

Opened for comment, especially on the use of std::map<std::string, double> to store the creep parameters. The storage avoids repeated calculation of the creep parameters in compute_strain_rate_and_derivative(), but may slow down the code. If there's a more clean and efficient way to do this, I'll make a change.

PR #3800, opened 2020-08-14
Added compute_strain_rate_and_derivative to ConstantViscosityPrefactors, with unit test

bobmyhill:
This PR adds a compute_strain_rate_and_derivative function to the ConstantViscosityPrefactors rheology module.
It also adds a new test, rheology_checker.prm, that is essentially a unit tester for otherwise hard-to-reach functions within the rheology module. I intend to add to this test as I add compute_strain_rate_and_derivative to more rheology plugins.

This PR is a continuation of the quest started in #3759.


bobmyhill:
Maybe the test should actually be a unit test?

bobmyhill:
That was the original intention, but in coding up a generalisation of diff-disl yesterday, I realised that I didn't actually use the constant viscosity strain rate or derivatives directly. I jumped the gun a bit, sorry about that. So for now I think this PR can be closed.

gassmoeller:
Thanks, no need to be sorry. Keep the branch around in case we need it in the future :+1:

PR #3802, opened 2020-08-14
Remove some unnecessary 'typename's.

bangerth:
Not all `typename`s are actually necessary: It's only necessary when the references name is a type inside a template-argument dependent (outer) type. Remove a number of cases where that is not actually necessary.


bangerth:
Looks like I'm making good progress this evening adding to my quote of PRs for the hackthon :-)

PR #3805, opened 2020-08-15
create and use is_defect_correction helper function.

MFraters:
follow up to #3112. This creates a helper function to centralize the question whether a type of nonlinear solver is a defect correction solver. 

I placed it close to the enum definition in parameters.h, but I am fine with putting it at an other location if there are strong opinions about it.

### DIFF
--- a/include/aspect/parameters.h
+++ b/include/aspect/parameters.h
@@ -80,6 +80,21 @@ namespace aspect
       };
     };
 
+    static
+    bool
+    is_defect_correction(const typename NonlinearSolver::Kind &input)
+    {
+      return input == NonlinearSolver::iterated_Advection_and_Newton_Stokes ||
+             input == NonlinearSolver::single_Advection_iterated_Newton_Stokes ||
+             input == NonlinearSolver::no_Advection_iterated_defect_correction_Stokes ||
+             input == NonlinearSolver::single_Advection_iterated_defect_correction_Stokes ||
+             input == NonlinearSolver::iterated_Advection_and_defect_correction_Stokes
+             ?
+             true
+             :
+             false;
+    }
+
     /**
      * @brief The NullspaceRemoval struct
      */
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -142,11 +142,7 @@ namespace aspect
     melt_handler (parameters.include_melt_transport ?
                   std_cxx14::make_unique<MeltHandler<dim>>(prm) :
                   nullptr),
-    newton_handler ((parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_Newton_Stokes ||
-                     parameters.nonlinear_solver == NonlinearSolver::single_Advection_iterated_Newton_Stokes ||
-                     parameters.nonlinear_solver == NonlinearSolver::no_Advection_iterated_defect_correction_Stokes ||
-                     parameters.nonlinear_solver == NonlinearSolver::single_Advection_iterated_defect_correction_Stokes ||
-                     parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_defect_correction_Stokes) ?
+    newton_handler (Parameters<dim>::is_defect_correction(parameters.nonlinear_solver) ?
                     std_cxx14::make_unique<NewtonHandler<dim>>() :
                     nullptr),
     post_signal_creation(
@@ -233,11 +229,7 @@ namespace aspect
 
     rebuild_stokes_matrix (true),
     assemble_newton_stokes_matrix (true),
-    assemble_newton_stokes_system ((parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_Newton_Stokes ||
-                                    parameters.nonlinear_solver == NonlinearSolver::single_Advection_iterated_Newton_Stokes ||
-                                    parameters.nonlinear_solver == NonlinearSolver::no_Advection_iterated_defect_correction_Stokes ||
-                                    parameters.nonlinear_solver == NonlinearSolver::single_Advection_iterated_defect_correction_Stokes ||
-                                    parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_defect_correction_Stokes)
+    assemble_newton_stokes_system (Parameters<dim>::is_defect_correction(parameters.nonlinear_solver)
                                    ?
                                    true
                                    :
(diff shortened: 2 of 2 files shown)


include/aspect/parameters.h
```diff
@@ -80,6 +80,21 @@ namespace aspect
...
 
+    static
+    bool
+    is_defect_correction(const typename NonlinearSolver::Kind &input)
+    {
+      return input == NonlinearSolver::iterated_Advection_and_Newton_Stokes ||
+             input == NonlinearSolver::single_Advection_iterated_Newton_Stokes ||
+             input == NonlinearSolver::no_Advection_iterated_defect_correction_Stokes ||
+             input == NonlinearSolver::single_Advection_iterated_defect_correction_Stokes ||
+             input == NonlinearSolver::iterated_Advection_and_defect_correction_Stokes
```
bangerth, line 91: I think the function's name doesn't actually represent what it is doing. That's because a "defect correction" is not the same as a first-order Picard-type iteration. More precisely, "defect" is an older word for "residual", and "correction" means that we compute an *update* from the defect. So DC-Picard definitely falls into this category, but `iterated_advection_and_Newton_Stokes` and `single_Advection_iterated_Newton_Stokes` don't actually fall into the "defect correction" category.

Start with the use cases: Why do the places where you use this function group them together? Maybe we can derive a better name that is based on the *consequences*, not a commonality of the schemes you're considering.

include/aspect/parameters.h
```diff
@@ -80,6 +80,21 @@ namespace aspect
...
+             ?
+             true
+             :
+             false;
```
bangerth, line 95: The statement
```
  return (cond ? true : false);
```
is the same as
```
  return cond;
```
:-)

include/aspect/parameters.h
```diff
@@ -80,6 +80,21 @@ namespace aspect
       };
     };
 
+    static
```
bangerth, line 83: Can you move the implementation of the function into the `.cc` file?

bangerth:
@ping?

include/aspect/parameters.h
MFraters: hmm, I always considered the Newton solver as a defect correction scheme, where we potentially adjust the update a bit based on the derivatives. The reason for that is that is how it is implemented, and they all follow exactly the same logic and limitations with for example how the boundary conditions are set up. How I view it is that the consequence is that instead of `Ax=b` we now solve `Jdx = Ax-b` where the dx is a update of x and J can be A or the actual Jacobian (or something in between). Would it be more correct to just call it a correction scheme instead of a defect correction scheme, since in both cases we compute a update?

include/aspect/parameters.h
MFraters: yes, I will fix that as soon as we have figure out the name ;)

MFraters:
@bangerth Sorry for leaving this out for so long. Since I considered mostly a stylistic change and I really had to think about what you said about the naming I kept putting it off in favour of other work...

include/aspect/parameters.h
bangerth: I think the question is still why `single_Advection_...` is part of the function, because that `single_Advection` isn't computing an "update" `dx` but computes `x` right away.

As I mentioned before, I think the better approach is to start with the place where you *need* this function and ask what the *property* is that you require.

source/simulator/core.cc
```diff
@@ -142,11 +142,7 @@ namespace aspect
     melt_handler (parameters.include_melt_transport ?
                   std_cxx14::make_unique<MeltHandler<dim>>(prm) :
                   nullptr),
-    newton_handler ((parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_Newton_Stokes ||
-                     parameters.nonlinear_solver == NonlinearSolver::single_Advection_iterated_Newton_Stokes ||
-                     parameters.nonlinear_solver == NonlinearSolver::no_Advection_iterated_defect_correction_Stokes ||
-                     parameters.nonlinear_solver == NonlinearSolver::single_Advection_iterated_defect_correction_Stokes ||
-                     parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_defect_correction_Stokes) ?
+    newton_handler (Parameters<dim>::is_defect_correction(parameters.nonlinear_solver) ?
```
bangerth, line 145: Here, the property seems to be "requires NewtonHandler". That would suggest calling the function `requires_newton_handler()`.

That makes this suspicious, though -- why would `no_Advection_iterated_defect_correction_Stokes` require a `NewtonHandler`?

source/simulator/core.cc
```diff
@@ -233,11 +229,7 @@ namespace aspect
 
     rebuild_stokes_matrix (true),
     assemble_newton_stokes_matrix (true),
-    assemble_newton_stokes_system ((parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_Newton_Stokes ||
-                                    parameters.nonlinear_solver == NonlinearSolver::single_Advection_iterated_Newton_Stokes ||
-                                    parameters.nonlinear_solver == NonlinearSolver::no_Advection_iterated_defect_correction_Stokes ||
-                                    parameters.nonlinear_solver == NonlinearSolver::single_Advection_iterated_defect_correction_Stokes ||
-                                    parameters.nonlinear_solver == NonlinearSolver::iterated_Advection_and_defect_correction_Stokes)
+    assemble_newton_stokes_system (Parameters<dim>::is_defect_correction(parameters.nonlinear_solver)
```
bangerth, line 232: Here it seems to be "requires assembly of the Newton Stokes system".

source/simulator/core.cc
MFraters: Why does that make you suspicious? They are separate things. A model with no advection can still be very nonlinear (for the first timestep).

include/aspect/parameters.h
MFraters: > I think the question is still why `single_Advection_...` is part of the function, because that `single_Advection` isn't computing an "update" `dx` but computes `x` right away.

I would think that the Stokes part can still be nonlinear, so it may have to iterated in a defect correction way for the first timestep. The defect correction is only about the Stokes part.

source/simulator/core.cc
MFraters: The naming of the these functions is because it started out as that. If I would rename it I would probably name it `assemble_defect_correction_Stokes_system`, except that I still don't really understand what your objection to that name is since I do not see a fundamental difference between the DC-Picard and Newton methods. Maybe it is easier to have a quick talk about this?

source/simulator/core.cc
bangerth: Right. What I meant to say is this: there is a disconnect somewhere if we initialize a *Newton* handler also in case where we don't actually do a Newton iteration (e.g., `no_Advection_and_defect_correction_Stokes`). It may be that the `NewtonHandler` class also provides services for other solvers, in which case it is probably that that class is misnamed and `is_defect_correction()` is the correct name. All I meant to suggest is that one or the other is misnamed -- that the juxtaposition of the two names is dissonant and suggests that one of the two is misnamed.

bangerth (approved):
Let's not agonize too much about names :-)

MFraters:
ow, I didn't fix the other two issues (return condition and moving implementation to `.cc` file) yet because I was waiting for the decision on the name. I can make a new pull request for that, or did you change your mind?

source/simulator/core.cc
MFraters: Ah, then I totally agree. The name of the NewtonHandler currently does not reflect anymore what services it actually provides.

PR #3816, opened 2020-08-17
List Menno, Anne, John as maintainers in the manual.

bangerth:
This required playing with some spacing on the front page, but I think the latex code easier to read this way as well.

@MFraters @naliboff @anne-glerum I've used the email addressed I usually use for you in the manual. If you want something different, please speak up!

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -134,33 +134,49 @@ COMPUTATIONAL INFRASTRUCTURE FOR GEODYNAMICS (CIG)
 {\Large Advanced Solver for Problems in Earth's ConvecTion}
 \end{center}
 
-%MAIN PICTURE%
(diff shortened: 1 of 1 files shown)


bangerth:
On 8/18/20 11:34 AM, Menno Fraters wrote:
> @bangerth <https://github.com/bangerth>, thanks! That email is fine. My only 
> worry is that we might need to change things around when the new logo arrives, 
> but since you already did this work I would be happy with this being merged :)

Yes, one issue per commit//pull request :-)

> Oke, one other thing that is kinda bothering me is that my name length breaks 
> the flow. Maybe I should change my last name to start with Ge instead of Fr or 
> something like that, or just add more letters to my first name...

Don't you have several middle initials? ;-)

gassmoeller:
I will merge this tomorrow unless someone asks for a different email address.

PR #3818, opened 2020-08-18
Introduce copy_particle_handler_function to particle world.

gassmoeller:
This is the function I promised @anne-glerum for #3760 to work. The test case does something very similar to what is needed there, it creates a new particle handler as a copy of the existing one and afterwards copies the new particle handler back into the existing one (thus overwriting the particles in the existing handler). @tjhei: I decided to register all signals to the copy as well, i.e. the particles of the copied handler will be shipped around in case the triangulation is refined or repartitioned. I did not attach the the signal that actually modifies the set of stored particles (the ones that enforce the particle per cell limit). Is this what you need as well?

### DIFF
--- a/include/aspect/particle/world.h
+++ b/include/aspect/particle/world.h
@@ -112,6 +112,18 @@ namespace aspect
         get_particle_handler() const;
 
         /**
+         * Copy the state of particle handler @p from_particle_handler into the
+         * particle handler @p to_particle_handler. This will copy
+         * all particles and properties and leave @p to_particle_handler
+         * as an identical copy of @p from_particle_handler, assuming it
+         * was set up by this particle world class. This means we assume
+         * @p from_particle_handler uses the same triangulation and
+         * particle properties as are used in this model.
+         */
+        void copy_particle_handler (const Particles::ParticleHandler<dim> &from_particle_handler,
+                                    Particles::ParticleHandler<dim> &to_particle_handler) const;
+
+        /**
          * Do initial logic for handling pre-refinement steps
          */
         void setup_initial_state ();
@@ -368,6 +380,14 @@ namespace aspect
         local_advect_particles(const typename DoFHandler<dim>::active_cell_iterator &cell,
                                const typename ParticleHandler<dim>::particle_iterator &begin_particle,
                                const typename ParticleHandler<dim>::particle_iterator &end_particle);
+
+        /**
+         * This function registers the necessary functions to the
+         * @p signals that the @p particle_handler needs to know about.
+         */
+        void
+        connect_particle_handler_signals(aspect::SimulatorSignals<dim> &signals,
+                                         ParticleHandler<dim> &particle_handler) const;
     };
 
     /* -------------------------- inline and template functions ---------------------- */
--- a/include/aspect/particle/world.h
+++ b/include/aspect/particle/world.h
@@ -118,7 +118,16 @@ namespace aspect
          * as an identical copy of @p from_particle_handler, assuming it
          * was set up by this particle world class. This means we assume
          * @p from_particle_handler uses the same triangulation and
-         * particle properties as are used in this model.
+         * particle properties as are used in this model. Existing
+         * particles in @p to_particle_handler are deleted.
+         *
+         * This function is expensive as it has to duplicate all data
+         * in @p from_particle_handler, and insert it into @p to_particle_handler,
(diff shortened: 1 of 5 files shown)


include/aspect/particle/world.h
```diff
@@ -111,6 +111,18 @@ namespace aspect
...
 
+        /**
+         * Copy the state of particle handler @p from_particle_handler into the
+         * particle handler @p to_particle_handler. This will copy
+         * all particles and properties and leave @p to_particle_handler
+         * as an identical copy of @p from_particle_handler, assuming it
+         * was set up by this particle world class. This means we assume
+         * @p from_particle_handler uses the same triangulation and
+         * particle properties as are used in this model.
+         */
```
tjhei, line 122: can you comment about what happens with existing particles in ``to_particle_handler``?

include/aspect/particle/world.h
```diff
@@ -111,6 +111,18 @@ namespace aspect
...
+        void copy_particle_handler (const Particles::ParticleHandler<dim> &from_particle_handler,
+                                    Particles::ParticleHandler<dim> &to_particle_handler) const;
```
tjhei, line 124: At least in deal.II we list the destination first (vmult for example). I don't insist that you change it that way, but it might be more intuitive.

bangerth: No, we have input first, output last. vmult really is a historical mistake.

bangerth:
Can you add to the documentation (i) that this is a very expensive operation, and (ii) an example of why one might want to use this function? 

In deal.II, several of the expensive functions have a `copy_from()` function. It might be useful to add such a function to `ParticleHandler` and, if available, use it in your current function. `copy_from()` would likely be far more efficient than what you are doing by hand here.

source/particle/world.cc
```diff
@@ -108,6 +108,59 @@ namespace aspect
...
+        connect_particle_handler_signals(this->get_signals(),to_particle_handler);
+
+        std::multimap<typename Triangulation<dim>::active_cell_iterator, Particles::Particle<dim> > new_particles;
+
+        for (const auto &particle : from_particle_handler)
+          {
+            Particles::Particle<dim> new_particle (particle.get_location(),
+                                                   particle.get_reference_location(),
+                                                   particle.get_id());
+            new_particle.set_property_pool(to_particle_handler.get_property_pool());
```
bangerth, line 136: Doesn't `ParticleHandler::insert()` set the property pool of the particles being inserted?

include/aspect/particle/world.h
tjhei: Oh great. I always look up vmult() to know the correct order. :-)

gassmoeller:
I think I have adressed all the comments. I have also rewoked the code slightly to make the copy operation faster by about 20% (removed one unnecessary command that already happens inside `insert_particles`, and avoided one copy operation of properties by copying the properties after inserting the particles into their final places). It is still about half as expensive as an actual particle advection step, so maybe we do not always do it for now, but only if necessary. If the `copy_from` function inside deal.II is significantly faster we can think about keeping an `old_particle_handler` around.

> Doesn't ParticleHandler::insert() set the property pool of the particles being inserted?

`ParticleHandler::insert_particle()` does set the property pool, but the collective `ParticleHandler::insert_particles()` functions do not. I think this is a historical accident, because the property pool is assigned separately in our `initialize_particles` function. This should already happen in the ParticleHandler class. Will open a PR for deal.II for that.

gassmoeller:
It seems setting the property pool is already fixed in deal.II since dealii/dealii#10589. I have limited the according line to deal.II 9.2.0.

include/aspect/particle/world.h
```diff
@@ -111,6 +111,27 @@ namespace aspect
...
+         * as an identical copy of @p from_particle_handler, assuming it
+         * was set up by this particle world class. This means we assume
+         * @p from_particle_handler uses the same triangulation and
+         * particle properties as are used in this model. Existing
+         * particles in @p to_particle_handler are deleted.
+         *
+         * This function is expensive as it has to duplicate all data
+         * in @p from_particle_handler, and insert it into @p to_particle_handler,
+         * which may be a significant amount of data. However, it can
+         * be useful for example to save the state of a particle
```
tjhei, line 127:
```suggestion
         * be useful to save the state of a particle
```

gassmoeller:
Done, also fixed the indentation.

PR #3820, opened 2020-08-19
New: Add Peierls exact formulation

bobmyhill:
This PR adds the exact formulation for Peierls creep to the Peierls creep rheology module, along with the associated strain rate and derivative functions.
It also adds a new test using the analytical calculations by @mibillen and @naliboff.

Based on #3723.

### DIFF
--- a/include/aspect/material_model/rheology/peierls_creep.h
+++ b/include/aspect/material_model/rheology/peierls_creep.h
@@ -108,6 +108,8 @@ namespace aspect
 
           /**
            * Compute the viscosity based on the selected Peierls creep flow law.
+           * This function uses either the compute_approximate_viscosity
+           * or the compute_exact_viscosity function.
            */
           double
           compute_viscosity (const double strain_rate,
@@ -138,6 +140,9 @@ namespace aspect
           /**
            * Compute the strain rate and first stress derivative
            * as a function of stress based on the selected Peierls creep law.
+           * This function uses either the
+           * compute_approximate_strain_rate_and_derivative
+           * or the compute_exact_strain_rate_and_derivative function.
            */
           std::pair<double, double>
           compute_strain_rate_and_derivative (const double stress,
--- a/include/aspect/material_model/rheology/peierls_creep.h
+++ b/include/aspect/material_model/rheology/peierls_creep.h
@@ -33,6 +33,26 @@ namespace aspect
     namespace Rheology
     {
       /**
+       * Data structure for Peierls creep parameters.
+       */
+      struct PeierlsCreepParameters
+      {
+        /**
+         * The Peierls creep prefactor, stress exponent, activation energy,
+         * activation volume, Peierls stress, glide parameters p and q
+         * and fitting parameter,
+         */
+        double prefactor;
+        double stress_exponent;
+        double activation_energy;
+        double activation_volume;
+        double peierls_stress;
+        double glide_parameter_p;
+        double glide_parameter_q;
+        double fitting_parameter;
+      };
+
+      /**
        * Peierls creep is a low temperature, high stress viscous deformation mechanism.
        * Crystal deformation occurs through dislocation glide, which requires high stresses.
        * It is considered an important deformation mechanism in the lithosphere and subducting
@@ -50,6 +70,12 @@ namespace aspect
           PeierlsCreep();
 
           /**
+           * Compute the creep parameters for the Peierls creep law.
+           */
+          const PeierlsCreepParameters
+          compute_creep_parameters (const unsigned int composition) const;
+
+          /**
            * Declare the parameters this function takes through input files.
            */
           static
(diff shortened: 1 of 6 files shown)


bobmyhill:
Hi @naliboff, I think this can be checked over now that your PR is merged :)

source/material_model/rheology/peierls_creep.cc
```diff
@@ -107,15 +199,124 @@ namespace aspect
...
       void
       PeierlsCreep<dim>::declare_parameters (ParameterHandler &prm)
       {
         prm.declare_entry ("Peierls creep flow law", "viscosity approximation",
-                           Patterns::Selection("viscosity approximation"),
+                           Patterns::Selection("viscosity approximation|exact"),
                            "Select what type of Peierls creep flow law to use. Currently, the "
-                           "only available option is an approximation to Peierls creep, which "
-                           "uses the strain rate invariant, rather than stress, as an input. ");
+                           "available options are 'exact', which uses a Newton-Raphson iteration "
```
naliboff, line 307: maybe `Newton-Raphson iterative method`?

source/material_model/rheology/peierls_creep.cc
```diff
@@ -107,15 +199,124 @@ namespace aspect
...
+                           "to find the stress and then compute viscosity, and 'viscosity approximation', "
+                           "in which viscosity is an explicit function of the strain rate invariant, "
+                           "rather than stress. ");
+
+        // Viscosity iteration parameters
+        prm.declare_entry ("Peierls strain rate residual tolerance", "1e-22", Patterns::Double(0.),
+                           "Tolerance for correct Peierls creep strain rate.");
```
naliboff, line 314:
```suggestion
                           "Tolerance for the iterative solve to find the correct Peierls creep strain rate.");
```

include/aspect/material_model/rheology/peierls_creep.h
```diff
@@ -71,11 +115,43 @@ namespace aspect
...
+           */
+          std::pair<double, double>
+          compute_exact_strain_rate_and_derivative (const double stress,
+                                                    const double pressure,
+                                                    const double temperature,
+                                                    const PeierlsCreepParameters creep_parameters) const;
+
+          /**
+           * Compute the strain rate and first stress derivative
+           * as a function of stress based on the selected Peierls creep law.
```
naliboff, line 140: At first reading I was a bit confused about the difference between this function and the two above, but this became immediately clear after reading on. Perhaps worth adding a note that this function utilizes the two functions above?  If this seems excessive feel free to disregard.

source/material_model/rheology/peierls_creep.cc
```diff
@@ -38,61 +38,153 @@ namespace aspect
...
+      PeierlsCreep<dim>::compute_exact_viscosity (const double strain_rate,
+                                                  const double pressure,
+                                                  const double temperature,
+                                                  const unsigned int composition) const
+      {
+        /**
+         * A generalised Peierls creep formulation. The Peierls creep expression
+         * for the strain rate has multiple stress-dependent terms, and cannot be
+         * directly inverted to find an expression for viscosity in terms of
+         * strain rate. For this reason, a Newton Raphson iteration is required,
```
naliboff, line 127:
```suggestion
         * strain rate. For this reason, a Newton-Raphson iteration is required,
```

source/material_model/rheology/peierls_creep.cc
```diff
@@ -38,61 +38,153 @@ namespace aspect
...
+
+        // Create a starting guess for the stress using
+        // the approximate form of the viscosity expression
+        double viscosity = compute_approximate_viscosity(strain_rate, pressure, temperature, composition);
+        double stress_ii = 2.*viscosity*strain_rate;
+        double strain_rate_residual = 2.*strain_rate_residual_threshold;
+
+        double strain_rate_deriv = 0;
+        unsigned int stress_iteration = 0;
+        while (std::abs(strain_rate_residual) > strain_rate_residual_threshold
```
naliboff, line 147: In diffusion_dislocation a fixpoint iteration is used if the Newton iteration does not succeed. Are we likely to run into the same issues here? If so, should we add this option in now? I have no practical experience with this type of rheology, and I will defer to your judgement. Worst case scenario, we add in the fixpoint iterative scheme if we find the Newton method frequently fails. 

However, right now if `stress_max_iteration_number` is reached there is no message to indicate the strain rate residual threshold was not reached.

tests/visco_plastic_peierls_mei_exact.prm
```diff
@@ -0,0 +1,115 @@
...
+    set Include Peierls creep                     = true
+    set Peierls creep flow law                    = exact
+    set Prefactors for Peierls creep              = 1.4e-19
+    set Activation energies for Peierls creep     = 320.e3
+    set Activation volumes for Peierls creep      = 0.
+    set Peierls stresses                          = 5.9e9
+    set Stress exponents for Peierls creep        = 2.0
+    set Peierls fitting parameters                = 0.15
+    set Peierls glide parameters p                = 0.5
+    set Peierls glide parameters q                = 1.0
```
naliboff, line 100: Pleased add in the following lines as a reminder for those looking at the test that these are additional options with the exact method:
`set Peierls strain rate residual tolerance = 1e-22`
`set Maximum Peierls strain rate iterations = 40`

bobmyhill:
Hi @naliboff,
I've addressed your comments, and added an AssertThrow to make sure that ASPECT terminates if the Newton iterative scheme fails.

There are a few changes that will probably need to be made to this rheology when it starts being used in models. Yes, we may need a backup iterative scheme (although I'm hopeful that we won't need this as we're using Magali's approximate scheme as a starting point). There should also probably be a limiter on the min and max viscosities for both the approximate and exact flow laws. 

I'd argue that we should save these modifications until we know we have a problem. I suspect that the problems will make themselves apparent very quickly in "realistic" models using only Peierls creep.

PR #3823, opened 2020-08-27
New: Composite viscous creep rheology

bobmyhill:
This PR adds a new Composite viscous creep rheology where diffusion, dislocation and Peierls creep rheologies are all active. At very low and high stresses, the viscosity asymptotically approaches user-defined minimum and maximum limits. 

The model can be visualised in 1D form as 3 nonlinear viscous creep dashpots connected in series, a linear viscous dashpot in parallel (where eta_lim = eta_min*eta_max/(eta_max - eta_min), and a final linear viscous dashpot in series with the entire system:
image

Building on this PR, a damped Drucker Prager element is added in #3884.


source/material_model/rheology/composite_viscous_creep.cc
```diff
@@ -0,0 +1,238 @@
...
+                                                                      const DiffusionCreepParameters diffusion_creep_parameters,
+                                                                      const DislocationCreepParameters dislocation_creep_parameters,
+                                                                      const PeierlsCreepParameters peierls_creep_parameters,
+                                                                      const double min_viscosity,
+                                                                      const double max_viscosity) const
+      {
+        const std::pair<double, double> diff_edot_and_deriv = diffusion_creep.compute_strain_rate_and_derivative(creep_stress, pressure, temperature, diffusion_creep_parameters);
+        const std::pair<double, double> disl_edot_and_deriv = dislocation_creep.compute_strain_rate_and_derivative(creep_stress, pressure, temperature, dislocation_creep_parameters);
+        const std::pair<double, double> prls_edot_and_deriv = peierls_creep.compute_strain_rate_and_derivative(creep_stress, pressure, temperature, peierls_creep_parameters);
```
naliboff, line 128: I think it would be really useful to output these three strain rates as an additional material model output. If we output the plastic and elastic strain rate as well, then we can effectively see what deformation mechanism is dominant at each point. Would it make sense to return these three values in this function, or write a separate function to do this?

source/material_model/rheology/composite_viscous_creep.cc
```diff
@@ -0,0 +1,238 @@
...
+      }
+
+
+
+      template <int dim>
+      void
+      CompositeViscousCreep<dim>::declare_parameters (ParameterHandler &prm)
+      {
+
+        prm.declare_entry ("Minimum strain rate", "1.4e-20", Patterns::Double(0.),
```
naliboff, line 143: There is already a parameter called Minimum strain rate in the `visco_plastic` (and `diffusion_dislocation`?) material model. To avoid confusion, should this be called the `Minimum creep strain rate`? Perhaps a minimum strain rate should be defined not in the specific material model (or rheology) subjection, but rather as a main material model parameter? It seems like the value specified here and the one used by the material model should be consistent.

source/material_model/rheology/composite_viscous_creep.cc
```diff
@@ -0,0 +1,238 @@
...
+                           "Stabilizes strain dependent viscosity. Units: \\si{\\per\\second}.");
+
+        // Viscosity iteration parameters
+        prm.declare_entry ("Strain rate residual tolerance", "1e-22", Patterns::Double(0.),
```
naliboff, line 147: Similar to above, I wonder if there is way we can declare these parameters once, and then use them for all of the relevant rheology modules?

source/material_model/rheology/composite_viscous_creep.cc
```diff
@@ -0,0 +1,238 @@
...
+                           "Tolerance for correct diffusion/dislocation strain rate ratio.");
+        prm.declare_entry ("Maximum strain rate ratio iterations", "40", Patterns::Integer(0),
+                           "Maximum number of iterations to find the correct "
+                           "diffusion/dislocation strain rate ratio.");
+
+        // Strain rate and stress limiting parameters
+        prm.declare_entry ("Minimum viscosities", "1.e17",
```
naliboff, line 154: Same here, a number of material models declare min and max viscosities. It would be nice to have all these min/max values defined only once.

source/material_model/rheology/composite_viscous_creep.cc
```diff
@@ -0,0 +1,238 @@
...
+      CompositeViscousCreep<dim>::parse_parameters (ParameterHandler &prm,
+                                                    const std::shared_ptr<std::vector<unsigned int>> &expected_n_phases_per_composition)
+      {
+        // Retrieve the list of composition names
+        const std::vector<std::string> list_of_composition_names = this->introspection().get_composition_names();
+
+        // Establish that a background field is required here
+        const bool has_background_field = true;
+
+
```
naliboff, line 185: Remove extra space.

source/material_model/rheology/composite_viscous_creep.cc
naliboff: I wonder if having the same (identical) names will actually cause issues during parsing. It seems like it should? If so, the name may need to be altered to something like `Minimum creep viscosity`

include/aspect/material_model/rheology/composite_viscous_creep.h
```diff
@@ -0,0 +1,120 @@
...
+
+        private:
+          /**
+           * Objects for computing viscous creep viscosities.
+           */
+          Rheology::DiffusionCreep<dim> diffusion_creep;
+          Rheology::DislocationCreep<dim> dislocation_creep;
+          Rheology::PeierlsCreep<dim> peierls_creep;
+
+          std::vector<double> min_viscosities;
```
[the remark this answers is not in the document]
bobmyhill: I don't know of a case where we would want composition-dependent values - it was just easy to implement this generalization, so I did :).

I'll keep this in until our conversation next week, but am happy to remove it.

source/material_model/rheology/composite_viscous_creep.cc
bobmyhill: I think this parameter should be shared by both the material model and the rheology model. This is how the current code does it (because we don't enter different subsections for the rheology models). Because this material model uses Minimum strain rate, it should be declared here, but it can share the parameter given to it by the user with the rheology model.

I've added a comment saying that this parameter is shared with the rheology models.

source/material_model/rheology/composite_viscous_creep.cc
bobmyhill: I don't like the idea of having implicit declarations in the material models. Redeclaring the same entry just overwrites the previous default, so there shouldn't be a problem.

I've changed the ordering of the declarations so that the default given here overrides whatever is given in the rheology modules.

source/material_model/rheology/composite_viscous_creep.cc
bobmyhill: See above; it isn't a problem. I have added a comment describing how the deal.II parameter handler works, and what is happening here.

source/material_model/rheology/composite_viscous_creep.cc
bobmyhill: I've tentatively chosen to pass a strain-rate vector by reference to the compute_viscosity function, which gets filled with these strain rates and the strain rate of the max_limiter (the four strain rates add up to the total strain rate).

bobmyhill:
@naliboff Thanks for the review! I've made all but one of the changes you requested ready for our meeting.

naliboff:
@bobmyhill - Thanks for addressing all of the points and looking forward to discussing this more at the next meeting!

gassmoeller:
This seems to be in a very good state. Do you want me to do a full review or did you wanted to just hear a general opinion? You still need a test and a changelog entry. Can this replace some part of the visco-plastic material model, or are there too many complications in that model (like the strain dependence)?

bobmyhill:
Hi @gassmoeller. At the moment, there is no material model which uses this rheology, so I wanted to ask what you think I should do. This PR is nice and simple, but hard to test, impossible to use and the additional named outputs produce a defined-but-not-used warning at compile time). If I add a new material model, the PR becomes more testable and useful, but also more complex. There's not a perfect place for this rheology in existing material models. Diffusion-dislocation is very similar, but the fixed-point iteration and hard viscosity limits are two notable differences.

> Can this replace some part of the visco-plastic material model, or are there too many complications in that model (like the strain dependence)?

I don't think this will replace the current V-P material model (at least not yet), because the approximations made in that model avoid the (potentially) expensive internal Newton solve. In one of the planned V(E)P papers, there will be a comparison of the two methods. I'd favour keeping these as separate models, rather than having yet another switch. 

Am I correct in thinking that the overhaul of the material model plugin system is going to be a while in coming? If so, I should probably make a new material model now (something similar to diffusion-dislocation would be my preference). But I wanted to check with a maintainer before I add more code.

gassmoeller:
> Am I correct in thinking that the overhaul of the material model plugin system is going to be a while in coming? If so, I should probably make a new material model now (something similar to diffusion-dislocation would be my preference). But I wanted to check with a maintainer before I add more code.

Then I would suggest to add a test that only creates the simplest possible version of your rheology model and probe it with analytic input values and check if the output makes sense. There is a template in tests/visco_plastic_derivatives_2d.cc, except your case could be even simpler (you do not need to run the model multiple times for different averaging schemes). The idea is to create the plugin outside of any material model and then feed it known inputs and post the outputs to the screen to check for correctness. The signal mechanism is a bit tricky, let me know if you have questions about that.

bobmyhill:
I think @alarshi also recommended that I do this.
Are you ok with defined-but-not-used warnings in main branch? If so, great! If not, should I comment out the additional named outputs for the time being?

tjhei:
> Are you ok with defined-but-not-used warnings in main branch?

That won't work because the tester will fail. You will need to address those.

bobmyhill:
Hi @gassmoeller, hi @naliboff 

This is now ready for a review.

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -0,0 +1,372 @@
...
+        // Diffusion creep parameters
+        Rheology::DiffusionCreep<dim>::declare_parameters(prm);
+
+        // Dislocation creep parameters
+        Rheology::DislocationCreep<dim>::declare_parameters(prm);
+
+        // Dislocation creep parameters
+        Rheology::PeierlsCreep<dim>::declare_parameters(prm);
+
+        // Some of the parameters below are shared with the rheology models,
```
naliboff, line 260:
```suggestion
        // Some of the parameters below are shared with the material models,
```
or perhaps I'm misreading? Will the values declared here override the values in the material models?

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -0,0 +1,372 @@
...
+        // and will already have been declared. This is fine, the deal.II
+        // parameter handler allows repeated declarations. The default values
+        // below will override any defaults given in the rheology models.
```
naliboff, line 263: Again, I'm not sure I understand ... the values below are in a rheology model. Should this read `... defaults given in the material models`?

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -0,0 +1,372 @@
...
+                           "Stabilizes strain dependent viscosity. Units: \\si{\\per\\second}.");
+
+        // Viscosity iteration parameters
+        prm.declare_entry ("Strain rate residual tolerance", "1e-22", Patterns::Double(0.),
+                           "Tolerance for correct diffusion/dislocation strain rate ratio.");
+        prm.declare_entry ("Maximum creep strain rate iterations", "40", Patterns::Integer(0),
+                           "Maximum number of iterations to find the correct "
+                           "creep strain rate.");
+
+        // Strain rate and stress limiting parameters
```
naliboff, line 276: Another questions about the minimum and maximum viscosities. Here, there is a unique value for each compositional field, whereas in visco plastic and diffusion dislocation they are single values. Could this cause any unforeseen issues?

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -0,0 +1,372 @@
...
+#include <deal.II/base/parameter_handler.h>
+
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    namespace Rheology
+    {
+      // The following function is not yet used by ASPECT, so
```
naliboff, line 36: At some point a comment should be added here (or elsewhere) that even a specific flow law mechanism is not used (ex: Peierls), there will still be a strain rate value output for that mechanism as the default flow law viscosity are the user-defined maximum viscosities.

naliboff (commented):
@bobmyhill - This looks great! I have a few minor questions and comments, some of which we may have already covered and I just forgot. Aside from that, this looks good to go from end. 

@gassmoeller - Would you like to do another check before merging?

source/material_model/rheology/composite_visco_plastic.cc
bobmyhill: No, the values declared here override only the values in the subordinate rheology models (diffusion, dislocation, ...). 
Most of the time, this will not matter - if the user declares a value for a parameter in the prm file, then that value will be used. These comments are only for people reading the code - the user shouldn't notice any odd behaviour.

I've extended the comment to be more explicit.

source/material_model/rheology/composite_visco_plastic.cc
bobmyhill: As above.

source/material_model/rheology/composite_visco_plastic.cc
bobmyhill: I don't think so (or at least, if there are, they are unforeseen for both of us!). Single values are also accepted for multiple compositional fields.

As an aside, I don't think diffusion-dislocation will end up using this composite, because of the hard limited viscosity in that model. We can discuss this and other minor issues when we start *the big tidy* that @gassmoeller dreams about :)

source/material_model/rheology/composite_visco_plastic.cc
bobmyhill: That isn't quite true. If a specific flow law mechanism is not used the strain rate for that mechanism will be exactly zero. To avoid unnecessary complexity, I've fixed the length and order of the outputs, so a vector may contain multiple outputs. I've extended the comment to make this clear.

bobmyhill:
Thanks @naliboff!

source/material_model/rheology/composite_visco_plastic.cc
naliboff: 👍

source/material_model/rheology/composite_visco_plastic.cc
naliboff: Thanks for extending the comment and really slick design!

naliboff (approved):
@bobmyhill - Looks great, the tester just needs to rerun and then this is good to go 👍

naliboff:
sorry, forgot to check if @gassmoeller wants to take another look.

naliboff (approved):
@bobmyhill - Just had a final read through this, look good to go. Thank you again for putting this together and for your patience with the long review process! Excited to try incorporating this into existing models!

PR #3828, opened 2020-08-28
New: Make thermodynamic table lookup into an EquationOfState

bobmyhill:
This PR makes a new EquationOfState plugin called ThermodynamicTableLookup. This can be used by both the Steinberger and GrainSize Material models, removing two almost identical chunks of code from those models. It will also make it easier for new material models to use the table lookup.

To make this PR as easy to read as possible, I will modify the GrainSize material model in a separate PR.

### DIFF
--- a/include/aspect/material_model/steinberger.h
+++ b/include/aspect/material_model/steinberger.h
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2011 - 2020 by the authors of the ASPECT code.
+  Copyright (C) 2011 - 2021 by the authors of the ASPECT code.
 
   This file is part of ASPECT.
 
--- a/include/aspect/material_model/steinberger.h
+++ b/include/aspect/material_model/steinberger.h
@@ -260,7 +260,6 @@ namespace aspect
          * Information about the location of data files.
          */
         std::string data_directory;
-        std::vector<std::string> material_file_names;
         std::string radial_viscosity_file_name;
         std::string lateral_viscosity_file_name;
 
--- a/include/aspect/material_model/steinberger.h
+++ b/include/aspect/material_model/steinberger.h
@@ -219,14 +219,9 @@ namespace aspect
         EquationOfState::ThermodynamicTableLookup<dim> equation_of_state;
 
         /**
-         * Boolean describing whether to calculate material properties
-         * including latent heat effects.
-         */
-        bool latent_heat;
-
-
-        /**
-         * Boolean describing whether to use the lateral average temperature.
+         * Boolean describing whether to use the lateral average temperature
+         * for computing the viscosity, rather than the temperature
+         * on the reference adiabat.
          */
         bool use_lateral_average_temperature;
 
--- a/include/aspect/material_model/steinberger.h
+++ b/include/aspect/material_model/steinberger.h
@@ -22,6 +22,7 @@
 #define _aspect_material_model_steinberger_h
 
 #include <aspect/material_model/interface.h>
+#include <aspect/material_model/equation_of_state/thermodynamic_table_lookup.h>
 
 #include <aspect/simulator_access.h>
 
@@ -147,44 +148,6 @@ namespace aspect
                                   const std::vector<double>    &compositional_fields,
                                   const SymmetricTensor<2,dim> &strain_rate,
                                   const Point<dim>             &position) const;
-
-        void fill_mass_and_volume_fractions (const MaterialModel::MaterialModelInputs<dim> &in,
-                                             std::vector<std::vector<double>> &mass_fractions,
-                                             std::vector<std::vector<double>> &volume_fractions) const;
-
-        void fill_seismic_velocities (const MaterialModel::MaterialModelInputs<dim> &in,
-                                      const std::vector<double> &composite_densities,
-                                      const std::vector<std::vector<double>> &volume_fractions,
-                                      SeismicAdditionalOutputs<dim> *seismic_out) const;
-
-        /**
-        * This function uses the MaterialModelInputs &in to fill the output_values
-        * of the phase_volume_fractions_out output object with the volume
-        * fractions of each of the unique phases at each of the evaluation points.
-        * These volume fractions are obtained from the PerpleX-derived
-        * pressure-temperature lookup tables.
-        * The filled output_values object is a vector of vector<double>;
-        * the outer vector is expected to have a size that equals the number
-        * of unique phases, the inner vector is expected to have a size that
-        * equals the number of evaluation points.
-        */
-        void fill_phase_volume_fractions (const MaterialModel::MaterialModelInputs<dim> &in,
-                                          const std::vector<std::vector<double>> &volume_fractions,
-                                          NamedAdditionalMaterialOutputs<dim> *phase_volume_fractions_out) const;
-
-        /**
-         * Returns the cell-wise averaged enthalpy derivatives for the evaluate
-         * function and postprocessors. The function returns two pairs, the
-         * first one represents the temperature derivative, the second one the
-         * pressure derivative. The first member of each pair is the derivative,
-         * the second one the number of vertex combinations the function could
-         * use to compute the derivative. The second member is useful to handle
-         * the case no suitable combination of vertices could be found (e.g.
-         * if the temperature and pressure on all vertices of the current
-         * cell is identical.
-         */
-        std::array<std::pair<double, unsigned int>,2>
-        enthalpy_derivatives (const typename Interface<dim>::MaterialModelInputs &in) const;
         /**
          * @}
          */
@@ -250,11 +213,21 @@ namespace aspect
 
 
       private:
-        bool has_background;
-        unsigned int first_composition_index;
+        /**
+         * The thermodynamic lookup equation of state.
+         */
+        EquationOfState::ThermodynamicTableLookup<dim> equation_of_state;
 
-        bool interpolation;
+        /**
+         * Boolean describing whether to calculate material properties
+         * including latent heat effects.
+         */
         bool latent_heat;
+
+
+        /**
+         * Boolean describing whether to use the lateral average temperature.
+         */
         bool use_lateral_average_temperature;
 
         /**
@@ -290,26 +263,6 @@ namespace aspect
         std::vector<std::string> material_file_names;
         std::string radial_viscosity_file_name;
         std::string lateral_viscosity_file_name;
-
-        /**
-         * List of pointers to objects that read and process data we get from
-         * Perplex files.
-         */
-        std::vector<std::unique_ptr<MaterialModel::MaterialUtilities::Lookup::PerplexReader> > material_lookup;
-
-        /**
-         * Vector of strings containing the names of the unique phases in all
-         * the material lookups.
-         */
-        std::vector<std::string> unique_phase_names;
-
-        /**
-         * Vector of vector of unsigned ints which constitutes mappings
-         * between lookup phase name vectors and unique_phase_names. The
-         * element unique_phase_indices[i][j] contains the index of phase name
-         * j from lookup i as it is found in unique_phase_names.
-         */
-        std::vector<std::vector<unsigned int>> unique_phase_indices;
 
         /**
          * Pointer to an object that reads and processes data for the lateral
--- a/source/material_model/equation_of_state/thermodynamic_table_lookup.cc
+++ b/source/material_model/equation_of_state/thermodynamic_table_lookup.cc
@@ -1,0 +1,712 @@
+/*
+  Copyright (C) 2011 - 2020 by the authors of the ASPECT code.
+
+  This file is part of ASPECT.
+
+  ASPECT is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2, or (at your option)
+  any later version.
+
+  ASPECT is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#include <aspect/material_model/equation_of_state/thermodynamic_table_lookup.h>
+#include <aspect/utilities.h>
+
+
+namespace aspect
+{
+  namespace MaterialModel
+  {
+    namespace EquationOfState
+    {
+      template <int dim>
+      void
+      ThermodynamicTableLookup<dim>::initialize()
+      {
+        // This model allows the user to provide several PerpleX or HeFESTo
+        // P-T lookup files, each of which corresponds to a different material.
+        // The thermodynamic properties and stable mineral phases within each material
+        // will in general be different, and must be averaged in a reasonable way.
+
+        // In the following, we populate the material_lookups.
+        // We also populate the unique_phase_names vector, which is used to
+        // calculate the phase volumes if they are provided by the lookup.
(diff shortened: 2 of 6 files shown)


bobmyhill:
Could someone take a look at this?
@cedrict and I were discussing some of his student projects, and this would have been useful for them.

source/material_model/steinberger.cc
```diff
@@ -697,21 +407,6 @@ namespace aspect
                                 "prescribed " + Utilities::int_to_string(material_file_names.size()) + " material data files, but there are " +
                                 Utilities::int_to_string(this->n_compositional_fields()) + " compositional fields."));
 
-        // The Steinberger material model currently assumes that all the
-        // compositional fields correspond to materials with lookup tables.
-        // Therefore the first composition index is hard-coded as zero.
-        // If more compositional fields are needed, this parameter allows
-        // the user to declare the first compositional index which
-        // corresponds to a lookup.
-        first_composition_index = 0;
```
jdannberg, line 706: So it looks like this variable wasn't used before (and also isn't used in the grain size model). In this case, I think it makes sense to just remove it rather than to keep all the lines of code in the material model to make this a potential option in the future (but still hardcode it to zero). Ideally, we probably want to implement this in the same way it is done in some other models, that it's possible to provide the names of phases that are mass/volume fractions and should be used for material averaging, rather than just providing one index.

source/material_model/steinberger.cc
```diff
@@ -675,14 +368,31 @@ namespace aspect
...
+          // PerpleX lookup tables.
+          // Therefore the first composition index is hard-coded as zero and
+          // a background field exists if the number of files is one greater
+          // then the number of compositional fields.
+          prm.set("Material file format", "perplex");
+          prm.set("Index of first mass fraction compositional field", "0");
+          if (material_file_names.size() == this->n_compositional_fields() + 1)
+            prm.set("Background material", "true");
+          else
+            prm.set("Background material", "false");
```
jdannberg, line 392: Is there a reason you set this here rather than in the equation of state? 
With these prm.set functions, what happens if a user sees that option in the parameter viewer, and decides they want to use it? Is this then overwritten here? That's probably not great for debugging. Wouldn't it be better to throw an Assert instead if someone tries to use this material model with the HeFesTo reader? The default is perplex anyway.

source/material_model/equation_of_state/thermodynamic_table_lookup.cc
```diff
@@ -0,0 +1,536 @@
...
+      ThermodynamicTableLookup<dim>::
+      evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
+               const unsigned int q,
+               MaterialModel::EquationOfStateOutputs<dim> &out) const
+      {
+        const double pressure = in.pressure[q];
+        const double temperature = in.temperature[q];
+
+        // The following lines take the appropriate averages of the
+        // thermodynamic material properties
```
jdannberg, line 313: This comment is outdated, please remove it.

source/material_model/steinberger.cc
```diff
@@ -697,21 +407,6 @@ namespace aspect
                                 "prescribed " + Utilities::int_to_string(material_file_names.size()) + " material data files, but there are " +
                                 Utilities::int_to_string(this->n_compositional_fields()) + " compositional fields."));
```
jdannberg, line 408: This whole block should also go into the Lookup equation of state.

jdannberg (changes requested):
Thank you for taking a stab at this! There are a few things I would change to make this as little code as possible within the material model itself (this is the goal after all, if someone wants to use this in a different material model). Otherwise it looks good!

source/material_model/steinberger.cc
bobmyhill: Yes, I agree. I was chatting with @MFraters a while ago about introducing a new labelling option for compositional fields, as the logic system we have at the moment is not very extensible.

source/material_model/steinberger.cc
bobmyhill: Oh, good point. Steinberger should be able to use either Perple_X/HeFESTo tables, so I'll remove that prm.set(). The others I'll change to AssertThrows.

source/material_model/steinberger.cc
```diff
@@ -392,166 +219,54 @@ namespace aspect
...
-                  out.specific_heat[i] = dH[0].first;
-                }
-              else
-                {
-                  out.thermal_expansion_coefficients[i] = (1 - out.densities[i] * material_lookup[0]->dHdp(in.temperature[i],pressure)) / in.temperature[i];
-                  out.specific_heat[i] = material_lookup[0]->dHdT(in.temperature[i],pressure);
-                }
-            }
-        }
+        equation_of_state.evaluate_using_enthalpy_derivatives(in, out);
```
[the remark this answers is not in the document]
bobmyhill: Thanks! When I started this PR, I couldn't decide whether to go down this route or try to keep the changes minimal. I chose the latter, but I think I agree with you that migrating more code now will make everything easier to maintain in the future.

source/material_model/steinberger.cc
```diff
@@ -392,166 +219,54 @@ namespace aspect
...
 
           // The density and isothermal compressibility are both volume-averaged
-          out.densities[i] = MaterialUtilities::average_value(volume_fractions[i], densities, MaterialUtilities::arithmetic);
-          out.compressibilities[i] = MaterialUtilities::average_value(volume_fractions[i], compressibilities, MaterialUtilities::arithmetic);
+          out.densities[i] = MaterialUtilities::average_value (volume_fractions[i], eos_outputs.densities, MaterialUtilities::arithmetic);
+          out.compressibilities[i] = MaterialUtilities::average_value (volume_fractions[i], eos_outputs.compressibilities, MaterialUtilities::arithmetic);
+          out.entropy_derivative_pressure[i] = MaterialUtilities::average_value (mass_fractions[i], eos_outputs.entropy_derivative_pressure, MaterialUtilities::arithmetic);
+          out.entropy_derivative_temperature[i] = MaterialUtilities::average_value (mass_fractions[i], eos_outputs.entropy_derivative_temperature, MaterialUtilities::arithmetic);
 
           if (!latent_heat)
```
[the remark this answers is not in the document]
bobmyhill: The if statement was to avoid unnecessary calculations. I've now migrated all of this code block to the thermodynamic lookup, so it won't be a problem any more.

source/material_model/steinberger.cc
```diff
@@ -392,166 +219,54 @@ namespace aspect
...
-                  out.specific_heat[i] = material_lookup[0]->dHdT(in.temperature[i],pressure);
-                }
-            }
-        }
+        equation_of_state.evaluate_using_enthalpy_derivatives(in, out);
 
       // fill seismic velocity outputs if they exist
       if (SeismicAdditionalOutputs<dim> *seismic_out = out.template get_additional_output<SeismicAdditionalOutputs<dim> >())
-        fill_seismic_velocities(in, out.densities, volume_fractions, seismic_out);
+        equation_of_state.fill_seismic_velocities(in, out.densities, volume_fractions, seismic_out);
```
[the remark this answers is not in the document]
bobmyhill: It just needs to be done after the material averaging, and I was uncertain whether we wanted to fix the material averaging (i.e. the mass / volume averaging for each property). The grain size does the averaging differently, for example (density is mass-averaged, for example).

I've now migrated everything over, but we should discuss what we want to do with the grain size model.

source/material_model/equation_of_state/thermodynamic_table_lookup.cc
```diff
@@ -0,0 +1,536 @@
...
+          }
+      }
+
+
+
+      template <int dim>
+      void
+      ThermodynamicTableLookup<dim>::declare_parameters (ParameterHandler &prm)
+      {
+        prm.declare_entry ("Data directory", "$ASPECT_SOURCE_DIR/data/material-model/steinberger/",
```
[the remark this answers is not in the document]
bobmyhill: No, me neither.

jdannberg (commented):
As discussed, first half of the review so that you can see what I just talked about.

include/aspect/material_model/steinberger.h
```diff
@@ -250,11 +213,21 @@ namespace aspect
...
+        /**
+         * Boolean describing whether to calculate material properties
+         * including latent heat effects.
+         */
         bool latent_heat;
+
+
+        /**
+         * Boolean describing whether to use the lateral average temperature.
+         */
```
jdannberg, line 230: If a comment basically just says what you can already figure out from the variable name, it doesn't add any new information. So I would either make the comment more useful ("Boolean describing whether to use the lateral average temperature for computing the viscosity, as opposed to the temperature on the reference adiabat."), or in this case, since this is already documented in the declare parameters function, I might just leave this without additional documentation.

source/material_model/equation_of_state/thermodynamic_table_lookup.cc
```diff
@@ -0,0 +1,611 @@
...
+        // Do some error checking
+
+        // First, do a fairly lax check on the number of material files.
+        // If the following comparison evaluates to be an equality,
+        // it implies that all compositional fields after
+        // "first_composition_index" correspond to mass fractions
+
+        if (has_background)
+          {
+            AssertThrow ((material_file_names.size() <= this->n_compositional_fields() - first_composition_index + 1),
```
jdannberg, line 534: As mentioned in the previous review, I think I would just remove the first_composition_index here and everywhere and assume it is zero. The option was not available before (one could have changed the hard-coded number, but it wasn't an option in the input file), so we can easily remove it and instead later on add an option to define the compositional fields that represent mass fractions.

source/material_model/equation_of_state/thermodynamic_table_lookup.cc
```diff
@@ -0,0 +1,611 @@
...
+        if (latent_heat)
+          evaluate_thermal_enthalpy_derivatives(in, out);
+
+        // fill seismic velocity outputs if they exist
+        if (SeismicAdditionalOutputs<dim> *seismic_out = out.template get_additional_output<SeismicAdditionalOutputs<dim> >())
+          fill_seismic_velocities(in, out.densities, volume_fractions, seismic_out);
+
+        // fill phase volume outputs if they exist
+        if (NamedAdditionalMaterialOutputs<dim> *phase_volume_fractions_out = out.template get_additional_output<NamedAdditionalMaterialOutputs<dim> >())
+          fill_phase_volume_fractions(in, volume_fractions, phase_volume_fractions_out);
```
jdannberg, line 337: I just noticed that you can't simply move those lines into the material model, since you create the additional outputs in the equation of state. But you could write a new function as part of the equation of state and then call that from the material model at the end of the evaluate function as `equation_of_state.fill_additional_outputs(in, volume_fractions, out)` or something like that. That should work.

jdannberg (commented):
Okay, this looks good apart from the points we discussed!

source/material_model/steinberger.cc
```diff
@@ -362,51 +334,46 @@ namespace aspect
...
+
           prm.leave_subsection();
         }
         prm.leave_subsection();
 
-        // Do some error checking
-        AssertThrow ((material_file_names.size() == 1) ||
-                     (material_file_names.size() == this->n_compositional_fields()) ||
-                     (material_file_names.size() == this->n_compositional_fields() + 1),
-                     ExcMessage("This material model expects either one material data file, or as many files as compositional fields, "
```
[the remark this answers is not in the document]
bobmyhill: > Now having only one data file is not allowed any more
Sure it is! It's allowed if the number of compositional fields is zero (background field) or one (no background field).

source/material_model/steinberger.cc
```diff
@@ -224,11 +224,11 @@ namespace aspect
...
                                MaterialModel::MaterialModelOutputs<dim> &out) const
     {
-      EquationOfStateOutputs<dim> eos_outputs (equation_of_state.number_of_lookups());
-
       std::vector<std::vector<double>> mass_fractions;
       std::vector<std::vector<double>> volume_fractions;
-      equation_of_state.fill_mass_and_volume_fractions (in, mass_fractions, volume_fractions);
+
+      // Evaluate the equation of state properties over all evaluation points
+      equation_of_state.evaluate(in, out, mass_fractions, volume_fractions);
```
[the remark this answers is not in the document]
bobmyhill: That works fine! Thanks!

source/material_model/steinberger.cc
```diff
@@ -362,51 +334,46 @@ namespace aspect
...
+              AssertThrow ((equation_of_state.number_of_lookups() == this->n_compositional_fields()),
+                           ExcMessage("The Steinberger material model assumes that all compositional "
+                                      "fields correspond to mass fractions of materials. There must either be "
+                                      "the same number of material lookup files as compositional fields, "
+                                      "or one additional file (if a background field is used)."));
+
+              prm.set("Background material", "false");
+            }
+
+          AssertThrow (prm.get_integer ("Index of first mass fraction compositional field") == 0,
```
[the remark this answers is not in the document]
bobmyhill: To summarise our slack chat, we'll need this for the grain size model so it's easier to include it now.

source/material_model/equation_of_state/thermodynamic_table_lookup.cc
bobmyhill: Hmmm, ok, you make a good point. Maybe this is a good opportunity to implement the compositional type labelling...

source/material_model/steinberger.cc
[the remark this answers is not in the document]
bobmyhill: Now removed as requested :)

source/material_model/steinberger.cc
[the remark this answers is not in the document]
jdannberg: Yes, but before it was allowed for any number of compositional fields. Basically, if you only have one table, it's clear that you want to use that, so you can use as many compositional fields as you want for whatever else you want. That now doesn't work any more. So for backwards compatibility, I would change the Assert above so that it also allows for having one table independently of the number of compositional fields.

source/material_model/equation_of_state/thermodynamic_table_lookup.cc
```diff
@@ -0,0 +1,577 @@
...
+                                                              MaterialModel::MaterialModelOutputs<dim> &out) const
+      {
+        // fill seismic velocity outputs if they exist
+        if (SeismicAdditionalOutputs<dim> *seismic_out = out.template get_additional_output<SeismicAdditionalOutputs<dim> >())
+          fill_seismic_velocities(in, out.densities, volume_fractions, seismic_out);
+
+        // fill phase volume outputs if they exist
+        if (NamedAdditionalMaterialOutputs<dim> *phase_volume_fractions_out = out.template get_additional_output<NamedAdditionalMaterialOutputs<dim> >())
+          fill_phase_volume_fractions(in, volume_fractions, phase_volume_fractions_out);
+      }
```
jdannberg, line 535: Can you move that implementation above the one of the declare parameters function?

source/material_model/steinberger.cc
```diff
@@ -442,116 +224,37 @@ namespace aspect
     Steinberger<dim>::evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
                                MaterialModel::MaterialModelOutputs<dim> &out) const
     {
+
```
jdannberg, line 227: Please remove this empty line.

jdannberg (approved):
Okay, looks good now. There are just three tiny changes I would suggest, then we can merge this.

source/material_model/steinberger.cc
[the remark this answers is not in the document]
bobmyhill: Oh, holy squirrels, you're right. Sorry, I had misunderstood/forgotten that a single file could be used for multiple material fields. I've relaxed the assert.

jdannberg:
As discussed, there is already an existing function for filling mass/volume fractions in the material model utilities that should be used instead, and this is also where the information about the background field belongs, but that can be fixed in a follow-up pull request.

PR #3829, opened 2020-08-31
Depth dependent viscosity input from an ascii data file.

alarshi:
This PR assigns depth-dependent rheology from an input ASCII data file, which can then be utilized into material models.

### For all pull requests:


jdannberg:
For adding a test (as discussed earlier today) one good starting point could be the [compressibility.cc](https://github.com/geodynamics/aspect/blob/master/tests/compressibility.cc) file. It makes a new material model derived from simple, and then mostly changes the evaluate function, which is the same thing you would want to do (you don't need the is_compressible() part).

alarshi:
@gassmoeller Hi Rene, I pushed the test case and the modified material model. I have two questions/comments on it: 
(1) ```make setup_tests``` gave an error on this test ```test ascii_depth_profile_rheology.prm is missing a folder with reference data ```. Since the test is not running, I created an empty statistics and screen-output files in the corresponding test directory. Is there another way to avoid the error message.
(2) I get a different error message on the test material model (I checked the structure of the test material model with the modified simple model, but couldn't find the difference. I will paste the errors here:

[test case](https://www.dropbox.com/s/dbxqpksxfiidw45/error_test.png?dl=0)
[modified simple](https://www.dropbox.com/s/z7w0t81ukwbm6l0/image.png?dl=0)

I am guessing that the additional input parameter that determines if we want to use viscosity from an ascii file or not, is not correctly declared or parsed in the test material model. Could you please look into it, and tell me how to resolve this error(s)? 

Thank you!!

gassmoeller:
About 1) we have a script that checks for the existence of a reference output folder. It seems that this script does not recognize empty files, try copying a screen-output and statistics file from an existing test. This will fail of course, but should allow you to run the test.
About 2) The first error message says that it could not find the shared library. Did you compile the test .cc file into a shared library? This happens automatically, when you run ctest (as described here: http://www.math.clemson.edu/~heister/manual.pdf#sec%3Arunning_tests), but not when you run the test manually (as in `./aspect tests/ascii_depth_profile_rheology.prm`). Try running the test through ctest and see if the error message changes. You may need to run it from your build directory as `ctest -R ascii_depth_profile_rheology -V` to see the error message. But in addition to that: You need to switch the order of lines 92-94 in the test .cc file. Whenever you initialize a plugin inside another plugin it needs to happen in the following order:
1. Create the plugin object (line 91)
2. Initialize the SimulatorAccess base class if necessary (currently line 93), so that all calls like `this->get_mpi_communicator()` work.
3. Read in input parameters (currently line 94)
4, Initialize the plugin itself (currently line 92).

This should get rid of the first error message.

alarshi:
@gassmoeller : Hello Rene. Thank you for explaining the error messages! 
In the latest commit, I modified the depth-dependent material model so that it uses the rheology plugin when ```depth_dependence == File ```. Currently, I have commented the original version because I am not sure if this is what we want. Could you please give me your suggestions on it, whenever you could?

include/aspect/material_model/rheology/ascii_depth_profile.h
```diff
@@ -0,0 +1,94 @@
...
+           */
+          void initialize ();
+
+          // avoid -Woverloaded-virtual:
+          using Utilities::AsciiDataProfile<dim>::initialize;
+
+          /**
+          * Return the viscosity at a given point of the domain.
+          */
+          double get_viscosity (const double depth) const;
```
gassmoeller, line 67: To be consistent with most of the existing rheology modules please call this `compute_viscosity`

source/material_model/depth_dependent.cc
```diff
@@ -57,57 +57,58 @@ namespace aspect
...
-      Table<1,double> viscosity_table( depth_table[0].size() );
-
-      for ( unsigned int i=0; i<visc_vec.size(); i++)
-        {
-          viscosity_table[i] = visc_vec[i];
-        }
-      viscosity_file_function
-        = std_cxx14::make_unique<Functions::InterpolatedTensorProductGridData<1>>(depth_table, viscosity_table);
-    }
+    // template <int dim>
```
gassmoeller, line 60: Please remove all commented code. We do not need it, because if you want to go back to an earlier version you can always do `git revert COMMIT` to make a new commit that exactly undoes what you did in commit id COMMIT, or `git reset COMMIT` (optional: `git reset --hard COMMIT`) if you want to change your branch (with --hard: and checked out files) to exactly the state they were at commit id COMMIT, losing all later commits unless you have a different branch that still has them.

source/material_model/depth_dependent.cc
```diff
@@ -57,57 +57,58 @@ namespace aspect
...
 
     template <int dim>
     double
     DepthDependent<dim>::viscosity_from_file(const double &depth) const
     {
-      const Point<1> dpoint(depth);
-      return viscosity_file_function->value( dpoint );
+      // const Point<1> dpoint(depth);
+      // return viscosity_file_function->value( dpoint );
+      return depth_dependent_rheology->get_viscosity(depth);
```
gassmoeller, line 111: Much simpler now :-)

source/material_model/depth_dependent.cc
```diff
@@ -197,16 +198,16 @@ namespace aspect
...
+          // prm.declare_entry ("Data directory", "$ASPECT_SOURCE_DIR/data/material-model/depth-dependent/",
+          //                    Patterns::DirectoryName (),
+          //                    "The path to the model data. The path may also include the special "
+          //                    "text `$ASPECT_SOURCE_DIR' which will be interpreted as the path "
+          //                    "in which the ASPECT source files were located when ASPECT was "
+          //                    "compiled. This interpretation allows, for example, to reference "
+          //                    "files located in the `data/' subdirectory of ASPECT.");
+          // prm.declare_entry("Viscosity depth file", "visc_depth.txt",
+          //                   Patterns::Anything (),
+          //                   "The name of the file containing depth-dependent viscosity data.");
```
gassmoeller, line 210: Also remove commented lines. But please add `prm.declare_alias ("Data file name","Viscosity depth file");`. This will allow to use the old parameter file name.

tests/ascii_depth_profile_rheology.prm
```diff
@@ -0,0 +1,110 @@
+# This is a relatively realistic incompressible model setup.
```
gassmoeller, line 1: I am not sure you wanted to add this test? At the moment it does not use your new rheology plugin. But now that you use the new plugin inside the depth_dependent material model you also do not need an extra test, because it is already used in the `depth_dependent_box_file_simple.prm` test.

gassmoeller (commented):
Yes, this is exactly how I imagined this to be used. I allowed running the tester so next time you update the PR we will see if the tests for the `depth_dependent` material model will still succeed. I left some comments, and please remove all the commented code. Also you are missing a changelog entry.

gassmoeller (approved):
This looks ready. Existing tests use the new model and pass, meaning no functionality has been changed. Thanks!

PR #3833, opened 2020-09-01
doxygen: fix image path for logo

tjhei:


tjhei:
I should have posted the error message: my doxygen might be older, but it fails with comma.

PR #3835, opened 2020-09-01
Use -O3 for tester

gassmoeller:
This is an experiment to address #3630. This will likely take a long time on the testers the first time it is run, because they have to compile deal.II to build the image. I hope this falls outside the timeout limit, but lets see. Otherwise we could change the deal.II base image (or derive another image from the deal.II base image and precompile there outside the testers).

### DIFF
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -4,7 +4,7 @@ pipeline {
   agent {
     // first build a docker image for running the tests from the repo
     docker {
-      image 'geodynamics/aspect-tester:optimized'
+      image 'geodynamics/aspect-tester:o1'
       // We mount /repos into the docker image. This allows us to cache
       // the git repo by setting "advanced clone behaviors". If the
       // directory does not exist, this will be ignored.
--- a/Jenkinsfile.cig
+++ b/Jenkinsfile.cig
@@ -3,8 +3,8 @@
 pipeline {
   agent {
     // first build a docker image for running the tests from the repo
-    dockerfile {
-      dir 'contrib/ci'
+    docker {
+      image 'geodynamics/aspect-tester:optimized'
       // We mount /repos into the docker image. This allows us to cache
       // the git repo by setting "advanced clone behaviors". If the
       // directory does not exist, this will be ignored.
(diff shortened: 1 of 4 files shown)


tjhei:
> because they have to compile deal.II to build the image

Yes, I am not too fond of this design but I guess it is okay to test things. I might have to clean out temporary images as they will just accumulate over time.
How much faster is ASPECT testing with this setting?

gassmoeller:
I switched it to use a prebuild docker image instead, and only use it for the CIG tester so far. On a quad-core the time for the test stage goes down from 47 min to 21 min (build time goes up from 2min 15s to 3min 50s). I think this is useful for speeding up the tests (probably on both testers), what do you think?

tjhei:
> what do you think?

Yes, that is certainly useful. It certainly reverts the slowdown we saw from 9.1 to 9.2 where deal.II switched from -O0 to -Og.

Did you look at any other optimization settings or did you go straight to -O3? I am asking because I need to pick an optimization setting for the docker container, but I am not sure if we really need to go to O3. I would probably pick -O1 but maybe we need to test things...

gassmoeller:
Lets find out, I build another image with O1. Will have a look once the tester is through.

bangerth:
I have a suspicion that it won't matter very much -- the CI time is probably dominated by running the tests, not the build, right? If that's true, then `-O3` would be the best choice. If the build dominates, then `-O1` would be it.

gassmoeller:
Here are the timings. O1 is not quite comparable (different machine, but also 4 cores), I could rerun on the same machine, but the numbers are consistent as they are, and the results are always identical (at least the tester passes for all 3). So I would think using O3 is the best choice, it just saves a lot of time on the testers.

| Level | Build | Run tests |
| ------- | --------| ------------ |
| O0 | 2:15 | ~47 min
| O1 | 3:30 |  ~30 min
| O3 | 3:50 |  ~21 min

bangerth:
Nice comparison!

bangerth:
Amazing to see that just by compiler optimizations, we can save a factor of more than 2x! Now I'm curious whether it's worth using `-O3 -march=native -mtune-native`?

tjhei:
I am updating the docker image to use -O3 (but I am against ``march``) as these images are used on different machines.

gassmoeller:
Superseded by #3848.

PR #3837, opened 2020-09-01
Fix newton solver tests

gassmoeller:
Contains the changes from #3836, and working towards #3681, these tests set an unrealistic nonlinear solver tolerance and a wrong reference viscosity, which becomes a problem when computing the reference viscosity. I made this PR to separate the changes to the tests from the changes that are caused by computing the reference viscosity. #3836 should be merged before this one.


naliboff (approved):
Looks good 👍

PR #3842, opened 2020-09-07
Timestepping: allow repeating of time steps

tjhei:
This is the continuation of the time stepping functionality and adds the implementation of the "Reaction" introduced in #3763.

- first determine the collective time step size, then check for the
reaction (based on current time). Adjust interface.
- implement different reactions
- add "repeat on cutback" scheme
- add a test

This probably needs:
1. Additional tests (I am only testing 2 out of the 4 reactions)
2. More thinking about other features that might break when repeating a time step (Assert for particles is already there)


tjhei:
@EstherHeck Take a look at the included test. It might already what you want. To avoid the situation that every time step will repeated once (large time step: repeat with small time step, advance to next, again too large, etc.), I am using 
``set Maximum relative increase in time step     = 2``
This is how it looks like (y axis is time step size and the test introduces a large velocity at t=0.75):
image

Without the new feature, the event at t=0.75 is not resolved:
image

tjhei:
Well, moving postprocessing is not an option it seems (judging from the number of tests I broke). This is not ideal and needs some extra work.

source/simulator/core.cc
```diff
@@ -1969,18 +1969,58 @@ namespace aspect
...
-        // if we postprocess nonlinear iterations, this function is called within
-        // solve_timestep () in the individual solver schemes
-        if (!parameters.run_postprocessors_on_nonlinear_iterations)
-          postprocess ();
+              // fall-through into the next switch case
+              DEAL_II_FALLTHROUGH;
+            }
+            case TimeStepping::Reaction::repeat_step:
+            {
+              pcout << "Repeating the current time step based on the time stepping manager ..." << std::endl;
```
anne-glerum, line 1988: The wording here and below is a bit confusing to me. We repeat the `time step number`, but with a different `time step size` (so earlier in time). Yet the parameter is called `new_time_step`, with which is meant the time step size.

source/time_stepping/interface.cc
```diff
@@ -76,23 +86,18 @@ namespace aspect
...
+    compute_time_step_size(Reaction &reaction) const
     {
       double new_time_step = std::numeric_limits<double>::max();
-      Reaction reaction = Reaction::advance;
 
       for (const auto &plugin : active_plugins)
         {
-          std::pair<Reaction, double> answer
-            = plugin->execute();
+          double this_time = plugin->execute();
```
anne-glerum, line 95: I'd rename this `this_time_step` for clarity.

source/time_stepping/repeat_on_cutback.cc
```diff
@@ -0,0 +1,96 @@
...
+    template <int dim>
+    Reaction
+    RepeatOnCutback<dim>::determine_reaction (const double time_step_size)
+    {
+      std::cout
+          << " old:" << this->get_old_timestep()
+          << " last: " << this->get_timestep()
+          << " now: " << time_step_size << std::endl;
+
+
```
anne-glerum, line 47: rm empty line

source/time_stepping/repeat_on_cutback.cc
```diff
@@ -0,0 +1,96 @@
...
+
+    template <int dim>
+    void
+    RepeatOnCutback<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Time stepping");
+      {
+        prm.enter_subsection("Function");
```
anne-glerum, line 62: Rename `Function` to `Repeat on cutback`. The `0.2` factor could be made into a input parameter right?

source/time_stepping/repeat_on_cutback.cc
```diff
@@ -0,0 +1,96 @@
...
+
+        prm.leave_subsection();
+      }
+      prm.leave_subsection();
+    }
+
+
```
anne-glerum, line 69: Add empty line.

source/time_stepping/repeat_on_cutback.cc
```diff
@@ -0,0 +1,96 @@
...
+    template <int dim>
+    void
+    RepeatOnCutback<dim>::parse_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Time stepping");
+      {
+        prm.enter_subsection("Function");
```
anne-glerum, line 76: Rename `Function`.

source/time_stepping/repeat_on_cutback.cc
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// explicit instantiations
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    ASPECT_REGISTER_TIME_STEPPING_MODEL(RepeatOnCutback,
+                                        "repeat on cutback",
+                                        "This model ")
```
anne-glerum, line 94: Add to model description.

tests/time_stepping_repeat.prm
```diff
@@ -0,0 +1,120 @@
...
+  end
+end
+
+
+subsection Boundary velocity model
+  subsection Function
+    set Variable names      = x,z,t
+    set Function constants  = pi=3.1415926
+    set Function expression = if(t<0.75, if(x<0.5,-1,1), 100); 0
+#    if(x>1+sin(0.5*pi*t), 1, -1); 0
```
anne-glerum, line 56: rm line

source/time_stepping/repeat_on_cutback.cc
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+
+
+    template <int dim>
+    Reaction
+    RepeatOnCutback<dim>::determine_reaction (const double time_step_size)
+    {
+      std::cout
+          << " old:" << this->get_old_timestep()
```
anne-glerum, line 43: Not sure whether you want to keep this in or not, but if so, I'd make it more descriptive (like `Previous time step size`, `Current time step size`) and add a space after `old:`.

anne-glerum (commented):
This is really great, @tjhei! I know it's not done yet, but going through it for an internal discussion, I just left some comments as I went along.

gassmoeller (commented):
I like the new structure because it solves the information problem for the individual plugins. I have a few high-level comments. I have not quite decided which of them work best, but would like to hear your thoughts:
- I like the idea of a `reaction` that is not a one-dimensional enum, but either a number of flags or a struct with members. That way we avoid duplication in the switch-statement, but simply saying: if `enforce refine` then `refine`; if `advance` then `advance` else if `repeat` then `repeat` else if `abort` then abort.
- Related: I would like the switch statement around the reaction to move out of the `run` function. We could have a `maybe_advance_time` function or `prepare_next_time_step` function that handles everything from line 1972 to 2029 (excluding the timing_output function). We could also keep the `reaction` out of the simulator class if the time stepping manager had functions similar to `should_simulation_terminate_now` (e.g. `should_simulation_refine_now`, `should_simulation_repeat_time_step`). That way we keep the data structure completely out of simulator and could change it later.
- I think the `refine` cases should be handled inside `maybe_refine_mesh` if that is possible. That function already checks for several conditions on when to refine, so it should also be responsible to query the time stepping manager if refinement is necessary. If the `refine` case can be separated into that function the remaining options are essentially `repeat`, `advance`, `terminate` (already existing). We could split hairs and add an `abort` reaction, where `terminate` signals that one of the termination criteria was reached, while `abort` signals that something went wrong.

What do you think? Which of the ideas are useful?

tjhei:
> * but either a number of flags or a struct with members.

I decided to keep the enum so I can do an MPI::min(), but I do not expose it anymore (instead, we now have several bools).

> * I would like the switch statement around the reaction to move out of the `run` function

I left it there right now. We can't completelty remove it though (because of the ``continue;``) and I feel like having the logic in one place makes sense. If you insist, we can move stuff like mesh refinement into the helper functions.

I am still unhappy with the current design of the plugins (and I still can not control the repeated time step size)...

tjhei:
@gassmoeller this is ready for another look, I think.

include/aspect/time_stepping/repeat_on_cutback.h
```diff
@@ -0,0 +1,78 @@
...
+        void
+        declare_parameters (ParameterHandler &prm);
+
+        virtual
+        void
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+        double cut_back_amount;
+        double repeat_threshold;
```
gassmoeller, line 72: The members need some documentation, at least if they are upper bounds or prefactors.

source/time_stepping/interface.cc
```diff
@@ -108,13 +113,96 @@ namespace aspect
...
+        {
+          const std::pair<Reaction, double> answer = plugin->determine_reaction(info);
+          reaction = static_cast<Reaction>(std::min(reaction, answer.first));
+          repeat_step_size = std::min(repeat_step_size, answer.second);
+        }
+
+      reaction = static_cast<Reaction>(Utilities::MPI::min(static_cast<int>(reaction), this->get_mpi_communicator()));
+      repeat_step_size = Utilities::MPI::min(repeat_step_size, this->get_mpi_communicator());
+
+      this->current_reaction = reaction;
```
gassmoeller, line 146: Hmm, not sure why you use `this->` here and in the lines below. To make clear that they are class members?

source/time_stepping/repeat_on_cutback.cc
```diff
@@ -0,0 +1,122 @@
...
+    template <int dim>
+    std::pair<Reaction, double>
+    RepeatOnCutback<dim>::determine_reaction (const TimeStepInfo &info)
+    {
+      std::cout
+          << "RepeatOnCutback old:" << this->get_old_timestep()
+          << " last: " << this->get_timestep()
+          << " next: " << info.next_time_step_size
+          << " last? " << info.reduced_by_termination_plugin
+          << std::endl;
```
gassmoeller, line 47: Is this debug output? If so remove. If not it would need a bit more explanation, like `computed timestep:`, `modified timestep:`...

source/time_stepping/repeat_on_cutback.cc
```diff
@@ -0,0 +1,122 @@
...
+
+// explicit instantiations
+namespace aspect
+{
+  namespace TimeStepping
+  {
+    ASPECT_REGISTER_TIME_STEPPING_MODEL(RepeatOnCutback,
+                                        "repeat on cutback",
+                                        "This time stepping plugin will detect a situation where the computed time step "
+                                        "shrinks by more than a user-controlled factor. In that situation, the time step "
```
gassmoeller, line 115: I would clarify this:
```suggestion
                                        "shrinks by more than a user-controlled factor. In that situation, the previous time step "
```

gassmoeller (changes requested):
I like it! I only have a few minor comments and I think there are a few remaining comments from Anne, but if you fix them this looks good.

source/time_stepping/interface.cc
tjhei: I find it more readable while working on the code, but I think it is not our usual style, so I will delete it.

tjhei:
updated.

source/time_stepping/interface.cc
gassmoeller: see Anne's comment: rename to `this_time_step`

gassmoeller (approved):
Looks good except for the one comment from Anne that was not addressed (probably hidden as outdated). I commented on that to point it out. If you fix that, add a changelog entry, and the tester passes this is ready to merge.

tjhei:
Good point. Updated.

PR #3844, opened 2020-09-09
Cookbook: Prescribed velocity ascii data

bobmyhill:
This PR adds a new cookbook showing how to prescribe velocities based on ASCII data input.
The required ASCII files are built from a png image with a python file. Both the image and the python file can be found in the cookbook directory.


bangerth:
@bobmyhill I realize that we have a cookbook here that's nearly a year old. What's the status? Should we review this and get it merged, or was there something that still needed to be done?

bobmyhill:
Hi @bangerth ... I think this was ready to be reviewed. 

It would be nice at some point to make this work with AMR, but that'll be another project for another day (and ?person).

cookbooks/prescribed_velocity_ascii_data/make_ascii_files_from_png.py
```diff
@@ -0,0 +1,65 @@
+from PIL import Image
```
bangerth, line 1: This is an awesome little script. I've often wished we had something like this for any number of purposes. Can you add a comment to the top what it does and how to call it?

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+                       "are constrained."
+                      );
+
+    prm.enter_subsection ("Prescribed velocities");
+    {
+      {
+        if (dim == 2)
+          {
+            prescribed_velocity_field_params_2d.declare_parameters(prm, "blob", "blob", "Ascii data model");
+            //prescribed_velocity_field_data_2d.initialize(1);
```
bangerth, line 73: remove the commented out line if you don't need it.

The last 3 arguments of the `declare_parameters` call are funny. Is there a better choice one could pass here?

Separately, the function is (should be) `static`, so maybe `Functions::ParsedFunction<2>::declare_parameters(...);` would be easier to read.

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+          }
+        else
+          {
+            prescribed_velocity_field_params_3d.declare_parameters(prm, "blob", "blob", "Ascii data model");
+            //prescribed_velocity_field_data_3d.initialize(1);
```
bangerth, line 78: same comments here, obviously.

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+          }
+      }
+
+      prm.enter_subsection ("Velocity function");
+      {
+        if (dim == 2)
+          Functions::ParsedFunction<2>::declare_parameters (prm, 2);
```
bangerth, line 85: Looks like you already do that here :-)

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+   * The DGP element used when 'set Use locally conservative discretization = true' does not
+   * have support points. If these elements are in use, a fictitious support point at the cell
+   * center is returned for each shape function that corresponds to the pressure variable,
+   * whereas the support points for the velocity are correct; the fictitious points don't matter
+   * because we only use this function when interpolating the velocity variable, and ignore the
+   * evaluation at the pressure support points.
+   */
+  template <int dim>
+  std::vector< Point<dim> >
+  get_unit_support_points_for_velocity(const SimulatorAccess<dim> &simulator_access)
```
bangerth, line 150: This is a tool function copied somewhere from within ASPECT, right? If so, maybe we could move that function into namespace `Utilities` so that it doesn't have to be duplicated.

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+      }
+  }
+
+  /**
+   * A set of helper functions that either return the point passed to it (if
+   * the current dimension is the same) or return a dummy value (otherwise).
+   */
+  namespace
+  {
+    const Point<2> as_2d(const Point<3> &/*p*/)
```
bangerth, line 193:
```suggestion
    Point<2> as_2d(const Point<3> &/*p*/)
```

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+    {
+      return Point<2>();
+    }
+
+    const Point<2> &as_2d(const Point<2> &p)
```
bangerth, line 198: Not wrong, but no need for the reference either
```suggestion
    Point<2> as_2d(const Point<2> &p)
```

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+
+  /**
+   * A set of helper functions that either return the point passed to it (if
+   * the current dimension is the same) or return a dummy value (otherwise).
+   */
+  namespace
+  {
+    const Point<2> as_2d(const Point<3> &/*p*/)
+    {
+      return Point<2>();
```
bangerth, line 195: Is this function supposed to actually do something useful? Or do you just need it to make compilation work? If the latter, put an `Assert (false, ExcInternalError())` into it to make sure that anyone who *actually* calls it gets an exception.

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    const Point<2> &as_2d(const Point<2> &p)
+    {
+      return p;
+    }
+
+    const Point<3> as_3d(const Point<2> &/*p*/)
```
bangerth, line 203: same idea for these functions

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+      {
+        const std::vector< Point<dim> > points = get_unit_support_points_for_velocity(simulator_access);
+        const Quadrature<dim> quadrature (points);
+        FEValues<dim> fe_values (simulator_access.get_fe(), quadrature, update_quadrature_points);
+        typename DoFHandler<dim>::active_cell_iterator cell;
+
+        // Loop over all cells
+        for (cell = simulator_access.get_dof_handler().begin_active();
+             cell != simulator_access.get_dof_handler().end();
+             ++cell)
```
bangerth, line 254: Now works with this:
```
  for (const auto &cell : dof_handler.active_cell_iterators())
```

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+          if (! cell->is_artificial())
+            {
+              fe_values.reinit (cell);
+              std::vector<types::global_dof_index> local_dof_indices(simulator_access.get_fe().dofs_per_cell);
+              cell->get_dof_indices (local_dof_indices);
+
+              for (unsigned int q=0; q<quadrature.size(); q++)
```
bangerth, line 261:
```suggestion
              for (unsigned int q=0; q<quadrature.size(); ++q)
```

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
```diff
@@ -0,0 +1,350 @@
...
+                         simulator_access.introspection().component_indices.velocities[dim-1]))
+                      {
+                        // Which velocity component is this DOF associated with?
+                        const unsigned int component_direction
+                          = (c_idx
+                             - simulator_access.introspection().component_indices.velocities[0]);
+
+                        // we get time passed as seconds (always) but may want
+                        // to reinterpret it in years
+                        double time = simulator_access.get_time();
```
bangerth, line 284: this line and the following several should probably be moved out of the loop over all cells

doc/manual/manual.tex
```diff
@@ -5569,6 +5569,48 @@ \subsubsection{Prescribed internal velocity constraints}
...
 \end{figure}
 
+\subsubsection{Prescribing internal velocity constraints with ASCII files}
+\label{sec:prescribed-velocities-ascii-data}
+\textit{This section was contributed by Bob Myhill}
+
+Building on \ref{sec:prescribed-velocities}, the \url{cookbooks/prescribed_velocity_ascii_data}
+directory contains a plugin which uses an ASCII data file to specify where to prescribe
+internal velocities. Velocities are prescribed wherever the field value indicated by the ASCII data file
+is less than 0.5. As before, the plugin is loaded in parameter files as an additional shared library:
```
bangerth, line 5579: I *think* that in the code you actually use "greater than 0.5".

doc/manual/manual.tex
```diff
@@ -5569,6 +5569,48 @@ \subsubsection{Prescribed internal velocity constraints}
...
+An example parameter file using this plugin can be found at
+\url{cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.prm}.
+In this file, the velocities are constrained to be zero within the letters ``ASPECT'' (Figure~\ref{fig:prescribed-velocity-ascii-data-init}).
+The part of this file which provides the location of the ASCII file and the
+prescribed velocity field function is:
+\lstinputlisting[language=prmfile]{cookbooks/prescribed_velocity_ascii_data/doc/prescribed_velocity_ascii_data.prm.1.out}
+A temperature gradient is applied within the letters, while the temperature field outside the letters is set to be constant.
+This initial temperature field is specified by another ASCII data file:
+\lstinputlisting[language=prmfile]{cookbooks/prescribed_velocity_ascii_data/doc/prescribed_velocity_ascii_data.prm.2.out}
+These two ASCII data files are generated from aspect\_name.png by the python file make\_ascii\_files\_from\_png.py,
```
bangerth, line 5591:
```suggestion
These two ASCII data files are generated from \texttt{aspect\_name.png} by the python file \texttt{make\_ascii\_files\_from\_png.py},
```

doc/manual/manual.tex
```diff
@@ -5569,6 +5569,48 @@ \subsubsection{Prescribed internal velocity constraints}
...
+\end{figure}
+
+During the simulation, excess heat diffuses out from the tops of the letters, and
+into the bases of the letters. The temperature gradients in the unconstrained
+part of the domain then generate convective flow. Figure~\ref{fig:prescribed-velocity-ascii-data}
+illustrates the resulting flow field.
+\begin{figure}
+    \centering
+    \includegraphics[width=0.8\textwidth]{cookbooks/prescribed_velocity_ascii_data/doc/prescribed_velocity_ascii_data.png}
+    \caption{Convective flow around the letters ASPECT, within which velocities are prescribed to be zero, as described in Section \ref{sec:prescribed-velocities-ascii-data}.}
```
bangerth, line 5609:
```suggestion
    \caption{\it Convective flow around the letters ASPECT, within which velocities are prescribed to be zero, as described in Section \ref{sec:prescribed-velocities-ascii-data}.}
```

bangerth (changes requested):
That's actually quite clever and neat! Nice work!

I have only minor issues.

doc/manual/manual.tex
bobmyhill: Good catch!

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
bobmyhill: This object is only used to read in the location of the data file (and optionally a scale parameter). The middle two arguments are default file directory and filename. I've changed those to be a bit more sensible.

cookbooks/prescribed_velocity_ascii_data/prescribed_velocity_ascii_data.cc
bobmyhill: It's copied from @jperryhouts' cookbook and from the associated test, so it isn't used anywhere in ASPECT's source directory. 

Do you want it moved, @bangerth? I'm happy to do that if you think it'll be more widely used.

bangerth:
Let's get this merged and, if you wouldn't mind, merge the code duplication into a `Utilities` function as a follow-up.

Can you squash or fix up your commits into a smaller number?

bobmyhill:
Done! Yes, I'm happy to migrate the duplicated function after this is merged.

bangerth (approved):

PR #3845, opened 2020-09-10
Add to the documentation of a function.

bangerth:
Just looking around at stuff.

### DIFF
--- a/benchmarks/nsinker/nsinker.cc
+++ b/benchmarks/nsinker/nsinker.cc
@@ -34,8 +34,6 @@ namespace aspect
          */
         NSinkerMaterial ();
 
-        double value (const Point<dim> &/*p*/) const;
-
         /**
          * @name Physical parameters used in the basic equations
          * @{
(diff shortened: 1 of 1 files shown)


benchmarks/nsinker/nsinker.cc
```diff
@@ -34,7 +34,10 @@ namespace aspect
          */
         NSinkerMaterial ();
 
-        double value (const Point<dim> &/*p*/) const;
+        /**
+         * Return the viscosity value at the given point `p`.
+         */
+        double value (const Point<dim> &p) const;
```
tjhei, line 40: I don't think this is correct. It returns the chi, which used to compute the viscosity.

bangerth:
Good point. I misread the code. I renamed the function now to express in the name what it is doing.

tjhei (approved):

tjhei:
The tester says:
```
/jenkins/workspace/aspect_PR-3845/benchmarks/nsinker/nsinker.cc:293:51: error: no member function ‘value’ declared in ‘aspect::NSinkerBenchmark::NSinkerMaterial<3>’
     NSinkerMaterial<3>::value (const Point<3> &p) const
                                                   ^~~~~
```

tjhei (approved):

PR #3847, opened 2020-09-11
fix a minus sign in GravityPointValues postprocessor

LudovicJnnt:
This PR just fixes a minus that got mistakenly introduced from a previous update. 

Somehow, the tests did not see the minus to one of the output value? 
There are 3 tests for the gravity post-processor:
* gravity_point_values_map  --> +value
* gravity_point_values_list     --> +value
* gravity_point_values_spiral --> **-value**

the spiral test have the negative value because it was created just after the mistake was introduced. 
Do I still need to update the tests? 


tjhei (approved):

tjhei:
> Do I still need to update the tests?

I would hope that this breaks one of the tests, we'll see. If yes, please update the test outputs.

tjhei:
Do you happen to know why this does not break any tests? Is this specific plugin not tested?

LudovicJnnt:
This plugin changes the gravity-00000, column 11. For clarity:
T=Test output (M=map, L=List, S=Spiral) 
F=new compared output 

When running the tests TM, TL, TS against F: 
They all passed. Because I was also surprised that they all passed, I compared the results of the values of column 11 in TM, TL,, TS and F. And they were surprisingly the same! EXCEPT that the precision set in the gravity output for all values in TM & TL were different (it used to be 18 (TM, TL) before I changed them to default 12 (TS, F) in a previous PR). The reason I did not updated TM & TL in the previous PR is probably because the tests already did not failed at that time so I did not think to update them. 

Then I also tested with reintroducing the minus, as how it was before. All tests passed, although the values of column 11 in N have a minus sign, while the test TM, TL, TS do not have the signs. Because all tests passes, I did not see there was a mistake as I did not plotted this specific value. 

Finally, to be sure that the gravity output was being compared, I changed the PRM file in TM and run the test, so that column 1 of the gravity output was different, and then the tests did fail.  

Now the reason why all tests may have passed: when testing, does the compared values have a precision condition (e.g. 1e-6 magnitude) so that if I have a compared values 4e-7 and -4e-7, the tests will not see that as a fail?

LudovicJnnt:
Tests not seeing failure: 
gravity_point_values_map
gravity_point_values_list

In the output_garvity/gravity-00000 column 11

tjhei:
Yeah, the values are smaller than 1e-6, which is the threshold we are using when comparing test results.

LudovicJnnt:
@tjhei I updated the 3 tests output of gravity_point_values (map, list and spiral) as it will be required after #3853

LudovicJnnt:
I updated the postprocessor adding another minus in the calculation of gravity gx, gy, gz so that plotting the vectors g(x,y,z) points towards the center of the sphere. The tests output are therefore also updated (this times they all failed because the diff was higher than the threshold values)

tjhei (commented):
Looks good. Please rebase.

LudovicJnnt:
@tjhei I rebased, tested again and all seems to work.

tjhei (approved):

tjhei:
great, thank you!

PR #3850, opened 2020-09-17
[WIP] Benchmark gravity postprocessor with a prem density material model pluggin

LudovicJnnt:
The benchmark consists of calculating gravity and its derivatives of a spherical domain filled with either PREM densities or a constant density, and compare the numerical predictions with the analytical solution. 
 
Is included a material plugin prem.cc which just set the density at depth following the equations given in the description of the preliminary reference Earth model (PREM) by Dziewonski & Anderson (1981). A test of the plugin is on the way (it does works and output the right density at the right place), but I would appreciate if someone can have a look and give some comments.

There is two .prm files for the benchmark of both model setups:
 - gravity_prem.prm
 - gravity_constant.prm

Also, @cedrict and I have started to write some text with figures for the manual. This benchmark will serve the purpose of a cookbook for the gravity postprocessor, and will also include the setup of a custom spherical shell. 


LudovicJnnt:
I have created a test-case for the prem material model plugin, but because it is a plugin, ASPECT cannot identify the library.
Do benchmark plugins require a test?

LudovicJnnt:
@gassmoeller, @naliboff : I think there is everything required for the benchmark in this PR. Before getting into coding details, @cedrict and I would appreciate a first overview of the text and figures and if we employed the right methodology.

LudovicJnnt:
After the PR merge that fixes a minus in the gravity calculation postprocessor, I had to rebase the benchmark gravity prem.
before that I had to install a newer version of candi/dealii otherwise I couldn't compile ASPECT.

Then when done, compile including the prem_density plugin but i encountered the following error which I do not understand - I have no idea what I have to do to fix this:

In file included from /home/ludovic/aspect/aspect_20-03-10/aspect/include/aspect/material_model/equation_of_state/linearized_incompressible.h:25:0,
                 from /home/ludovic/aspect/aspect_20-03-10/aspect/include/aspect/material_model/simple.h:25,
                 from /home/ludovic/aspect/aspect_20-03-10/aspect/benchmarks/gravity_prem/prem.cc:20:
/home/ludovic/aspect/aspect_20-03-10/aspect/include/aspect/simulator_access.h:780:13: error: ‘AffineConstraints’ does not name a type; did you mean ‘BVInsertConstraints’?
       const AffineConstraints<double> &
             ^~~~~~~~~~~~~~~~~
             BVInsertConstraints
CMakeFiles/gravity_prem.dir/build.make:62: recipe for target 'CMakeFiles/gravity_prem.dir/prem.cc.o' failed
make[2]: *** [CMakeFiles/gravity_prem.dir/prem.cc.o] Error 1
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/gravity_prem.dir/all' failed
make[1]: *** [CMakeFiles/gravity_prem.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2

gassmoeller:
Hey Ludo, sorry this took so long to come back to. I think your approach looks good and the benchmark seems to be in good shape. I can do a full review when you think it is ready. Have you figured out the problems with the test yet? The way you include the plugin looks ok, and the test on the reference tester fails, but not because it cant find the library. Instead it seems to have some problems with the mesh (0 or negative cell volume), does it work when you run it locally for you?

LudovicJnnt:
Hey Rene, thanks for looking into it. It is ready for review. 

The last commit was after a rebase and changing the test slightly to make it more rapid, but locally it always passed. 
I have rebased today again to be sure, retested locally and again it passes, so I do not know what the problem is.

benchmarks/gravity_prem/doc/gravity_prem.tex
```diff
@@ -0,0 +1,295 @@
...
+  subsection Gravity calculation
+    set Sampling scheme             = list of points 
+    set List of radius              = 6596000
+    set List of longitude           = 38.7578, 174.7633, -157.8583,   85.324, -77.0428,   2.3522, -21.9426, -122.4194, 139.6917,  16.3738
+    set List of latitude            =  8.9806, -36.8485,   21.3069,  27.7172, -12.0464,  48.8566,  64.1466,   37.7749,  35.6895,  48.2082 
+  end
+end
+\end{lstlisting}
+
+Results are shown in Fig.~\ref{fig:gravitycities}. Measurements agree well with the expected analytical value but the presence of under- and overshoot is the unavoidable consequence of the type of mesh used and quadrature order.
```
gassmoeller, line 148:
```suggestion
Results are shown in Fig.~\ref{fig:gravitycities}. Measurements agree well with the expected analytical value. The remaining anomalies are the consequence of the mesh resolution and the quadrature order and disappear with higher resolution.
```

benchmarks/gravity_prem/doc/gravity_prem.tex
```diff
@@ -0,0 +1,295 @@
...
+\includegraphics[width=0.48\linewidth]{../../benchmarks/gravity_prem/doc/profile_gravity_const_U.pdf}
+\caption{Gravity acceleration $|{\mathbf g}|$ and potential $U$ along a radial profile for a hollow sphere of constant density of 3000 \si{\kilogram\per\cubic\metre}. The grey area symbolises the hollow sphere interior where density is non zero. The analytical solution is also shown in black.}
+\label{fig:gravityline}
+\end{figure}
+
+ \begin{figure}[h!]
+\centering
+\includegraphics[width=0.48\linewidth]{../../benchmarks/gravity_prem/doc/cities_gravity_const_g.pdf}
+\includegraphics[width=0.48\linewidth]{../../benchmarks/gravity_prem/doc/cities_gravity_const_U.pdf}
+\caption{Gravity acceleration $|{\mathbf g}|$ and potential $U$ at GOCE height  (225 \si{km}) above 10 cities query points for a hollow sphere of constant density of 3000 \si{\kilogram\per\cubic\metre} counting 61440 elements. We varied the increased quadrature point degree from 2 to 4 to illustrate the rapid convergence of the gravity and potential toward the analytical solution shown in black.}
```
gassmoeller, line 184: In this figure it looks like the optimal quadrature degree is nq+3 (I guess that means the degree of the finite element plus 3?). It is slightly surprising that nq+4 is worse, I would expect it to get better with increasing quadrature degree until at some point it does not make a difference. Can you check that the captions and the plotted lines are consistent?

benchmarks/gravity_prem/doc/gravity_prem.tex
```diff
@@ -0,0 +1,295 @@
...
+transition zone 2 & $5771<r<5971$\si{km} & $\rho(r)=11.2494-8.0298x$\\
+transition zone 3 & $5971<r<6151$\si{km} & $\rho(r)=7.1089-3.8045x$\\
+Low velocity zone &$6151<r<6291$\si{km} & $\rho(r)=2.6910+0.6924x$\\
+LID  &$6291<r<6346.6$\si{km} & $\rho(r)=2.6910+0.6924x$\\
+Lower Crust &$6346.6<r<6356$\si{km} & $\rho(r)=2.9$\\
+Upper Crust &$6356<r<6368$\si{km} & $\rho(r)=2.6$\\
+Ocean &$6368<r<6371$\si{km} & $\rho(r)=1.020$\\
+\hline
+\end{tabular}
+\caption{List of ten layers making the PREM model and their corresponding polynomial expression for the density as a function of $x=r/R_{Earth}$.
```
gassmoeller, line 213:
```suggestion
\caption{List of ten layers defining the PREM model and their corresponding polynomial expression for the density as a function of $x=r/R_{Earth}$.
```

benchmarks/gravity_prem/doc/gravity_prem.tex
```diff
@@ -0,0 +1,295 @@
...
+Note that the returned densities should be multiplied by 1000 to obtain
+units of \si{\kilogram\per\cubic\metre}.}
+\label{tab:premdensity}
+\end{table}
+
+As in the first benchmark, gravity is computed on a line and results are shown in Fig.~\ref{fig:gravityline2}. The PREM article also specifies the gravity inside the planet and we see that the results agree with the PREM profile. Also, we recover a value of 9.81 \si{\metre\per\square\second} at the surface of the Earth as expected.
```
gassmoeller, line 219:
```suggestion
As in the first benchmark, gravity is computed on a line and results are shown in Fig.~\ref{fig:gravityline2}. The PREM publication also specifies the gravity inside the planet and we show that the benchmark results agree with the published PREM profile. Also, we recover a value of 9.81 \si{\metre\per\square\second} at the surface of the Earth as expected. 
```

benchmarks/gravity_prem/doc/gravity_prem.tex
```diff
@@ -0,0 +1,295 @@
...
+units of \si{\kilogram\per\cubic\metre}.}
+\label{tab:premdensity}
+\end{table}
+
+As in the first benchmark, gravity is computed on a line and results are shown in Fig.~\ref{fig:gravityline2}. The PREM article also specifies the gravity inside the planet and we see that the results agree with the PREM profile. Also, we recover a value of 9.81 \si{\metre\per\square\second} at the surface of the Earth as expected. 
+
+\begin{figure}[h!]
+\centering
+\includegraphics[scale=0.30]{../../benchmarks/gravity_prem/doc/default_shell_full.png}
+\caption{Whole sphere mesh filled using the default mesh generation method. The sphere counts 28672 elements and is filled with the PREM densities.}
```
gassmoeller, line 224:
```suggestion
\caption{Full spherical mesh generated using the default mesh generation method. The sphere counts 28672 elements and its density is defined using the PREM densities.}
```

gassmoeller (commented):
So far for review part 1. I did not look into the python conversion scripts yet, but if you want to routinely output the gravity values in different formas it would be relatively straightforward to use the deal.II output functionality instead of hand-writing the output files. This would be similar to what we do in source/postprocess/particles.cc, except simpler, because we always know which properties each of your satellite points has. Not something that has to happen right away, but could be a future extension.

One comment on your lateral output of the different cities: We can not see any signal there except for the noise, because you only use a layered PREM density model, is that correct? In that case it may make sense to quantify the error in the text, i.e. how much of a deviation do you see for which resolution and quadrature degree and is that sufficient for your purpose?

benchmarks/gravity_prem/doc/gravity_prem.tex
LudovicJnnt: Well spotted! I missed that and for this case, there has been a swap between nq+3 and nq+4. That is because I ran the models individually and an error occurred. It is indeed expected at a specific height, increasing the quadrature degree results in a convergence. as shown in this figure: 




However, I was not initially surprised because I did see cases with other experiments that I am performing now, where the nq+4 is worse than nq+3. Those new experiments are ran in a shellscript using varied GMR from 0 to 4 and increase in nq+ from 0 to 7. Then the gnuplot is also from a loop. 
Here are a series of figures showing gravity acceleration difference between gravity measured and the gravity theoretic along a radial profile for a sphere with PREM densities. You can see that increasing nq+X does not predict a smooth convergence. 

B2_profile_g_gmr_nq0_diff
B2_profile_g_gmr_nq7_diff
B2_profile_g_nq_gmr1_diff
B2_profile_g_nq_gmr4_diff

LudovicJnnt:
@gassmoeller, I pushed the changes following your review on part 1 (except for the quantification error, yet, as I am still thinking of it).

gassmoeller:
@LudovicJnnt: Thanks, I will take a look at your files next, but it seems you somehow got some changes from other branches on master into your PR. Could you please rebase this branch to a recent version of master to make the review easier?

LudovicJnnt:
Hi @gassmoeller 
Thanks for looking at the PR, been a while I did not work on it so I had no idea what happened. 
I rebased, fix manually the conflicts and force-pushed which I hope was what I was meant to do.. 
During the rebase process I got warnings for whitespace errors which I never had before so I didn't do anything about it. 

For the tests, locally the benchmark_prem_density test still passes, however, the zero_tempearture test failed like it did here.

benchmarks/gravity_prem/gnuplot_script_PREM_density_profile
```diff
@@ -0,0 +1,67 @@
+# -------------------------------------------------------------------------------------
```
gassmoeller, line 1: We usually use the extension `plt` for gnuplot scripts. Please rename all gnuplot script files, e.g. this one into `PREM_density_profile.plt`. While this is not a universal standard it makes finding files easier for us.

benchmarks/gravity_prem/gravity_PREM_density.prm
```diff
@@ -0,0 +1,143 @@
...
+# polynomial degree as the base for the number of quadrature
+# points on which the density evaluation is performed. 
+subsection Discretization
+  set Composition polynomial degree     = 2
+  set Stokes velocity polynomial degree = 2
+  set Temperature polynomial degree     = 1
+  set Use locally conservative discretization = true
+end
+
+# A gravity model always have to be specified. This does not
```
gassmoeller, line 57:
```suggestion
# A gravity model always has to be specified. This does not
```

benchmarks/gravity_prem/gravity_PREM_density.prm
```diff
@@ -0,0 +1,143 @@
...
+end
+
+# Call the premDensity material model pluggin written specifically
+# for this benchmark.
+subsection Material model
+  set Model name = premDensity
+end
+
+# We output the density, gravity and some statistics.
+# Are shown here 2 subsections of the gravity calculation:
```
gassmoeller, line 73:
```suggestion
# Here shown are 2 subsections using different patterns of gravity calculation:
```

benchmarks/gravity_prem/gravity_PREM_density.prm
```diff
@@ -0,0 +1,143 @@
...
+  end
+
+# subsection A: calculate gravity along the radial profile 
+# at the 'random' spherical coordinate [R,115,87]
+  subsection Gravity calculation
+    set Sampling scheme             = map
+   
+    # The density is evaluated at the quadrature points. You
+    # may increase the base number of quadrature points:
+    # e.g. if the base number is 2 (Q2) then an increase of
```
gassmoeller, line 94:
```suggestion
    # e.g. if the base number is 2 (Q2 velocity element) then an increase of
```

benchmarks/gravity_prem/gravity_PREM_density.prm
```diff
@@ -0,0 +1,143 @@
...
+
+# subsection B: calculate gravity on a uniformly distributed 
+# map at GOCE height (225km above Earth surface) using the
+# fibonacci spiral sampling scheme.
+  #subsection Gravity calculation
+  #  set Sampling scheme             = fibonacci spiral #map
+ 
+    # The density is evaluated at the quadrature points. You
+    # may increase the base number of quadrature points:
+    # e.g. if the base number is 2 (Q2) then an increase of
```
gassmoeller, line 125:
```suggestion
    # e.g. if the base number is 2 (Q2 velocity element) then an increase of
```

benchmarks/gravity_prem/gravity_constant_density.prm
```diff
@@ -0,0 +1,177 @@
+# This prm file is for the branchmark gravity_prem only:
+# it describes the model setup to calculate gravity of a
+# custom spheriral shell filled with a constant density.
```
gassmoeller, line 3:
```suggestion
# custom spherical shell filled with a constant density. 
```

benchmarks/gravity_prem/prem.cc
```diff
@@ -0,0 +1,176 @@
...
+         * @name Physical parameters used in the basic equations
+         * @{
+         */
+        virtual void evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
+                              MaterialModel::MaterialModelOutputs<dim> &out) const
+        {
+          for (unsigned int i=0; i < in.n_evaluation_points(); ++i)
+            {
+              const std::array<double,dim> position_point_sph = Utilities::Coordinates::cartesian_to_spherical_coordinates<dim>(in.position[i]);
+              const double x = position_point_sph[0] / 6371e3;
```
gassmoeller, line 61: easier to understand:
```suggestion
              const double normalized_r = position_point_sph[0] / 6371e3;
```

benchmarks/gravity_prem/prem.cc
```diff
@@ -0,0 +1,176 @@
...
+         * @}
+         */
+
+    };
+
+
+    template <int dim>
+    double
+    PremDensity<dim>::
+    reference_viscosity () const
```
gassmoeller, line 146: Since you already have the implementation of the `evaluate` function in the class declaration above, you may as well move this function and the `is_compressible` function in there as well to be consistent. Saves some lines.

benchmarks/gravity_prem/prem.cc
```diff
@@ -0,0 +1,176 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+#include <aspect/simulator.h>
+#include <aspect/material_model/simple.h>
+#include <aspect/boundary_velocity/interface.h>
+#include <aspect/simulator_access.h>
+#include <aspect/global.h>
```
gassmoeller, line 24: I dont think you need any of these lines. The only one you need is:
```suggestion
 #include <aspect/material_model/interface.h>
```

benchmarks/gravity_prem/prem.cc
```diff
@@ -0,0 +1,176 @@
...
+#include <aspect/utilities.h>
+
+#include <deal.II/dofs/dof_tools.h>
+#include <deal.II/numerics/data_out.h>
+#include <deal.II/base/quadrature_lib.h>
+#include <deal.II/base/function_lib.h>
+#include <deal.II/numerics/error_estimator.h>
+#include <deal.II/numerics/vector_tools.h>
```
gassmoeller, line 32: All of these can go too.

```suggestion
```

doc/modules/changes/20201006_jeanniot
```diff
@@ -0,0 +1,12 @@
...
+gravity acceleration and potential for a hollow sphere filled with
+a single material of constant density, compared with the analytical
+solution. The second benchmark calculates gravity acceleration, 
+potential and gradients for a full sphere filled with PREM density.
+Both benchmarks also serve as a tutorial to use the functionality of 
+the postprocessor gravity_point_values, including a list of points, 
+a radial profile, an equiangular or equidistant (uniform) field of 
+points, the latter using the Fibonacci spiral sampling scheme.
+<br>
+(L. Jeanniot & C. Thieulot, 2020/10/06)
```
gassmoeller, line 12: We are not consistent, but try to avoid special characters here (this will go into html code). Use either comma or `and`
```suggestion
(L. Jeanniot, C. Thieulot, 2020/10/06)
```

gassmoeller (commented):
Nice job, and yes the rebase fixed the issues. I only have some minor comments, renaming and stuff. Otherwise this looks good. Let me know when you have done the adjustments, afterwards this is ready.

benchmarks/gravity_prem/prem.cc
LudovicJnnt: @gassmoeller I am not sure where exactly I have to move those blocks. And if I move them, do I need to change the format of the blocks? I'll need help on this one.

gassmoeller: You can take a look at how the `evaluate` function is implemented. There is only one definition of that function (not two like for `reference_viscosity` in line 135 and 146) and the one definition is further up inside the `class PremDensity{ };` block. If you move this function to line 135, and make the function look like the `reference_viscosity` function there (remove the class name and the template line), then it should work just the same as now. The same for the `is_compressible` function in line 155 and 111 (move it from 155 to 111). Does that help?

LudovicJnnt: @gassmoeller thanks! I have moved those lines.

PR #3851, opened 2020-09-21
Boundary velocity residual

alarshi:
This PR adds the statistics postprocessor for calculating the statistics of the boundary velocity residual visualization postprocessor. I have tested the visualization, but need a review if the input position for data velocity, (GPlates or Ascii) (ln 88 in boundary_velocity_residual_postprocessor.cc) is correct.

### For all pull requests:


include/aspect/postprocess/boundary_velocity_residual_statistics.h
```diff
@@ -0,0 +1,128 @@
+/*
+  Copyright (C) 2014 - 2019 by the authors of the ASPECT code.
```
gassmoeller, line 2: 2020

include/aspect/postprocess/boundary_velocity_residual_statistics.h
```diff
@@ -0,0 +1,128 @@
...
+#include <aspect/boundary_velocity/gplates.h>
+
+
+namespace aspect
+{
+  namespace Postprocess
+  {
+
+    /**
+     * A postprocessor that computes some statistics about the velocity at
```
gassmoeller, line 37: Please update the description to say something about the data the velocity is compared against.

include/aspect/postprocess/boundary_velocity_residual_statistics.h
```diff
@@ -0,0 +1,128 @@
...
+        /**
+         * Evaluate the boundary velocity residual solution at a point.
+         * The evaluation point must be at the top of the model domain, this function
+         * is called after the execute()
+         */
+        Tensor<1,dim>
+        evaluate (const Point<dim> &p) const;
+
+        /**
+           * Declare the parameters this class takes through input files.
```
gassmoeller, line 67: fix the indentation here ... astyle can not automatically do this for doxygen comments.

include/aspect/postprocess/boundary_velocity_residual_statistics.h
```diff
@@ -0,0 +1,128 @@
...
+        execute (TableHandler &statistics) override;
+
+        /**
+          * This function reads the specified input velocity data files, i.e., either an ascii data file or
+          * a file from the GPlates model.
+          */
+        void initialize () override;
+
+        /**
+         * Evaluate the boundary velocity residual solution at a point.
```
gassmoeller, line 59: I think this explanation needs to be a bit more clear. The function only gets a position, therefore it can not compute a residual (or the solution), all it can do is evaluate the boundary velocity data that is read from the data files.

include/aspect/postprocess/boundary_velocity_residual_statistics.h
```diff
@@ -0,0 +1,128 @@
...
+         * The evaluation point must be at the top of the model domain, this function
+         * is called after the execute()
```
gassmoeller, line 61: I do not think this is strictly true, it is called *from* the execute() function. And what happens if the point p is not at the top of the model domain?

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
+/*
+  Copyright (C) 2011 - 2019 by the authors of the ASPECT code.
```
gassmoeller, line 2: 2020

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+      std::set<types::boundary_id> boundary_indicators;
+      boundary_indicators.insert(this->get_geometry_model().translate_symbolic_boundary_name_to_id("top"));
+
+      for (const auto p : boundary_indicators)
+        {
+          local_max_vel[p] = -std::numeric_limits<double>::max();
+          local_min_vel[p] = std::numeric_limits<double>::max();
+        }
+
+      // for every surface face that is part of a geometry boundary
```
gassmoeller, line 65: I would update that this only happens for every *selected* geometry boundary (which happens to be only "top" at the moment).

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+        if (cell->is_locally_owned())
+          for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
+            if (cell->face(f)->at_boundary() && cell->face(f)->boundary_id() ==
+                this->get_geometry_model().translate_symbolic_boundary_name_to_id("top"))
+              {
+                fe_face_values.reinit (cell, f);
+                // extract velocities
+                fe_face_values[this->introspection().extractors.velocities].get_function_values (this->get_solution(),
+                    velocities);
+                // determine the max, min, and squared velocity on the face
```
gassmoeller, line 79: velocity difference?

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+                                                                                         .translate_id_to_symbol_name (a->first))
+                                           + " (m/yr)";
+              statistics.add_value (name_min, a->second*year_in_seconds);
+              const std::string name_rms = "RMS residual velocity on boundary with indicator "
+                                           + Utilities::int_to_string(rms->first)
+                                           + aspect::Utilities::parenthesize_if_nonempty(this->get_geometry_model()
+                                                                                         .translate_id_to_symbol_name (rms->first))
+                                           + " (m/yr)";
+              statistics.add_value (name_rms, rms->second*year_in_seconds);
+              // also make sure that the other columns filled by the this object
```
gassmoeller, line 189: If you copied this from somewhere else, could you fix it there in a separate PR as well?

```suggestion
              // also make sure that the other columns filled by this object
```

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+
+      return std::pair<std::string, std::string> ("Max, min, and rms residual velocity along boundary parts:",
+                                                  screen_text.str());
+    }
+
+
+
+    template <int dim>
+    void
+    BoundaryVelocityResidualStatistics<dim>::initialize ()
```
gassmoeller, line 255: Please move this function to the beginning of the file. Our usual convention (not strictly followed, but a good convention) is the following order of functions in a plugin:

1. constructor (if any)
2. destructor (if any)
3. initialize (if present)
4. any plugin specific function (`evaluate` in this case)
5. the main function of the plugin (`execute` in this case)
6. declare_parameters (if present)
7. parse_parameters (if present)

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+
+
+    template <int dim>
+    Tensor<1,dim>
+    BoundaryVelocityResidualStatistics<dim>::evaluate (const Point<dim> &p) const
+    {
+
+      Tensor<1,dim> value;
```
gassmoeller, line 290: I think you can remove `value` and only work with `data_velocity` directly

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+
+          gplates_lookup->load_file(data_directory + data_file_name, this->get_mpi_communicator());
+        }
+    }
+
+
+
+    template <int dim>
+    Tensor<1,dim>
+    BoundaryVelocityResidualStatistics<dim>::evaluate (const Point<dim> &p) const
```
gassmoeller, line 287: `evaluate` is a bit ambiguous. What about `get_data_velocity`?

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+      else
+        {
+          data_velocity =  gplates_lookup->surface_velocity(p);
+        }
+
+      value = data_velocity ;
+
+      return value;
+    }
+
```
gassmoeller, line 321: add an empty line

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+namespace aspect
+{
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(BoundaryVelocityResidualStatistics,
+                                  "boundary velocity residual statistics",
+                                  "A postprocessor that computes some statistics about "
+                                  "the velocity along the boundaries. For each boundary "
+                                  "indicator (see your geometry description for which boundary "
+                                  "indicators are used), the min and max velocity magnitude is computed.")
```
gassmoeller, line 417: Update the description.

source/postprocess/visualization/boundary_velocity_residual.cc
```diff
@@ -95,129 +66,12 @@ namespace aspect
...
-            }
-            prm.leave_subsection();
-          }
-          prm.leave_subsection();
-        }
-        prm.leave_subsection();
+          for (unsigned int q=0; q<computed_quantities.size(); ++q)
+            for (unsigned int d = 0; d < dim; ++d)
+              computed_quantities[q](d) = boundary_velocity_residual_statistics.evaluate(input_data.evaluation_points[q])[d] -
+                                          input_data.solution_values[q][d] * velocity_scaling_factor;
```
gassmoeller, line 72: This is correct, but consider what is happening here: You call the `evaluate` function oncer per dimension, but then only use the `d`'th component and ignore the rest, then repeat this computation two more times. It likely does not matter here, but for other cases you want to optimize this in the following way. You can do this because the call to `evaluate` does not depend on `d`.

```suggestion
{
            const Tensor<1,dim> data_velocity = boundary_velocity_residual_statistics.evaluate(input_data.evaluation_points[q]);
            for (unsigned int d = 0; d < dim; ++d)
              computed_quantities[q](d) = data_velocity[d] -
                                          input_data.solution_values[q][d] * velocity_scaling_factor;
 }
```

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,419 @@
...
+
+      for (const auto p : boundary_indicators)
+        {
+          local_max_vel[p] = -std::numeric_limits<double>::max();
+          local_min_vel[p] = std::numeric_limits<double>::max();
+        }
+
+      // for every surface face that is part of a geometry boundary
+      // and that is owned by this processor,
+      // compute the maximum, minimum, and squared*area velocity magnitude,
```
gassmoeller, line 67: velocity *residual*?

gassmoeller (commented):
All of this looks correct to me. I left some smaller comments, mostly documentation. You will also need to add the function `required_other_postprocessors` to the visualization postprocessor to make sure the statistics postprocessor is active as well if only the visualization postprocessor was selected (you can copy the approach from the geoid postprocessor).

I will mark this ready for test, and when you addressed the comments and push the changes the tester should succeed as long as your change did not change the functionality of the visualization postprocessor. You could also add this postprocessor to the list of active postprocessors of that test.

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -0,0 +1,420 @@
...
+            {
+              screen_text << p->second << " m/s, "
+                          << a->second << " m/s, "
+                          << rms->second << " m/s"
+                          << (index == global_max_vel.size()-1 ? "" : ", ");
+            }
+
+        }
+
+      return std::pair<std::string, std::string> ("Max, min, and rms residual velocity along boundary parts:",
```
gassmoeller, line 315: Just to be consistent with the other uses of RMS:

```suggestion
      return std::pair<std::string, std::string> ("Max, min, and RMS residual velocity along boundary parts:",
```

gassmoeller (changes requested):
I have some small comments left, but overall very close.

alarshi:
@gassmoeller Thank you! I addressed the suggestions in the latest commit.

tjhei:
I just had a quick look and this looks very good!

gassmoeller:
I had a look into the remaining differences that are in the test (i.e. why is the RMS residual so large, although we compute the difference between a prescribed boundary velocity and itself). First observation: The difference shrinks with increasing resolution, pointing to the fact that it is an interpolation issue, not a systematic error in the equations or data. I then tried a different quadrature formula for the calculation (QTrapez), which made the difference disappear. This suggests the problem is that the boundary conditions are only strictly enforced at the support points of the finite element, while the residual is computed at the quadrature points (which are currently different). Because the velocity solution is that of a second order finite element, but the data is interpolated linearly that leads to differences in all points, except the support points of the finite element. We could use `QTrapez`, which only computes the differences at the vertices of each face, but that is not very accurate (and reducing the accuracy of the final calculation just to satisfy our test seems silly), so I would suggest to change the quadrature formula to a `QGaussLobatto` formula of degree `velocity_degree + 1`. This quadrature rule will (incidentally :-)) use exactly the support points of the finite element we use for the computation. Let me know if that works for you. After changing the quadrature and updating the test results this should be ready.

tjhei:
> This quadrature rule will (incidentally :-)) use exactly the support points of the finite element we use for the computation.

Is that the right thing to do, though? I guess it depends on what you want to use the plugin for, but a more accurate quadrature gives you the real residual on the boundary (which includes the approximation error due to a coarse mesh).

gassmoeller (approved):
I guess you are right, we really want the most accurate quadrature. If we imagine a test where the reference data came from an analytical function instead of a linearly interpolated ascii data file then the `QGauss` quadrature would be the most accurate and for the final models we also want to consider errors due to geometry approximations. Ok, good to go then.

PR #3853, opened 2020-09-23
tests: reduce absolute tolerance for diffs

tjhei:
As an experiment, see how many tests break when I lower the tolerance. Based on the discussion during the weekly meetings today.


tjhei:
This change breaks around 70 tests, but it might be useful to make this change (reasoning is the test in #3847 not failing as expected).
We rely on the blessed tester for the test output, so maybe setting a stricter tolerance is not an issue? Of course this makes it less likely for any other machine to test any of the tests (unlikely anyways).

LudovicJnnt:
If you rely on the tester, maybe it would be useful to have a **note** or a sentence in the test section in the manual that explains the output value threshold and thus to be careful if the output values are < threshold value?

gassmoeller (approved):
I am ok with changing this. We already state that the reference tester is the official output and most other systems will fail. Are you ok with the current state of this PR?

tjhei:
> Are you ok with the current state of this PR?

I think this is an improvement to the current state, so yes. One can argue whether ``-a 1e-10 -r 1e-8`` is a good choice, but it seems reasonable to me.

PR #3855, opened 2020-09-28
print dt to screen output

tjhei:
This adds the current time step size (or dt) to the screen output.

This might be a bit controversial as this touches every single test, but I think it is quite useful. I realized today that I wanted this for a very long time.


tjhei:
If will update the test output once I get at least one :+1:

MFraters (approved):
I can see that it can be useful and I have no objections against it. Also, a nice clean up of some of the code :) 

I do think that it would be good to get more opinions before merging.

gassmoeller (approved):
I am fine with the change, and the differences in test output look ok, the small changes in residuals etc are things that were not flagged as failures (but are now updated). @MFraters feel free to merge when the tester is done.

tjhei:
> the small changes in residuals etc are things that were not flagged as failures (but are now updated).

I can remove them from the diff if you want, but I would think it is better to be closer to the blessed output.

tjhei:
I forgot to warn you: all other open PRs now need to update test results. Sorry about that.

PR #3859, opened 2020-10-05
Improve placement of parentheses.

bangerth:
While there, also improve spacing of an asterisk.

Follow-up to #3855.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -550,11 +550,11 @@ namespace aspect
     // first produce some output for the screen to show where we are
     {
       const char *unit = (parameters.convert_to_years ? "years" : "seconds");
-      const double factor = (parameters.convert_to_years ? 1./year_in_seconds : 1.0);
+      const double multiplier = (parameters.convert_to_years ? 1./year_in_seconds : 1.0);
 
       pcout << "*** Timestep " << timestep_number
-            << ":  t=" << time *factor << ' ' << unit
-            << ", dt=" << time_step *factor << ' ' << unit
+            << ":  t=" << (time * multiplier) << ' ' << unit
+            << ", dt=" << (time_step * multiplier) << ' ' << unit
             << std::endl;
     }
 
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -549,8 +549,8 @@ namespace aspect
   {
     // first produce some output for the screen to show where we are
     {
-      const char *unit = (parameters.convert_to_years) ? "years" : "seconds";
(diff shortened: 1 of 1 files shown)


tjhei:
> Improve placement of parentheses.

Not that this is important, but I am a bit confused how this is supposed to improve readability.


> While there, also improve spacing of an asterisk.

Where?

source/simulator/core.cc
```diff
@@ -549,8 +549,8 @@ namespace aspect
   {
     // first produce some output for the screen to show where we are
     {
-      const char *unit = (parameters.convert_to_years) ? "years" : "seconds";
-      const double factor = (parameters.convert_to_years) ? 1./year_in_seconds : 1.0;
+      const char *unit = (parameters.convert_to_years ? "years" : "seconds");
```
bangerth, line 552: Whether it's an improvement might be subjective: I thought that putting parentheses around just the variable itself was probably not what you had in mind -- but putting them around the entire expression might make thinking about operator precedence easier. We often have parentheses around `?:` expressions.

source/simulator/core.cc
```diff
@@ -549,8 +549,8 @@ namespace aspect
...
+      const double factor = (parameters.convert_to_years ? 1./year_in_seconds : 1.0);
 
       pcout << "*** Timestep " << timestep_number
             << ":  t=" << time *factor << ' ' << unit
```
bangerth, line 556: Huh, I had put a space to the right of the asterisk here:
```suggestion
            << ":  t=" << time * factor << ' ' << unit
```
Bizarrely, the indentation script removes it again. Same for the following line. I have no idea why...

tjhei: Interesting. astyle must be confused here.

source/simulator/core.cc
tjhei: Interesting. I did have that in mind because I generally like to put the thing in front of ? in parentheses (imagine having ``!convert_to_years`` or ``convert_to_years==true``). Anyways, I don't really care either way.

source/simulator/core.cc
tjhei: do you want to try ``(time * factor)`` ?

source/simulator/core.cc
bangerth: Right, if the condition is an actual expression, I totally agree. It was more a question of what to do if the variable is only a single name.

source/simulator/core.cc
bangerth: Yes, that works.

bangerth:
Maybe now?

tjhei (approved):

tjhei:
yes, thank you.

PR #3860, opened 2020-10-05
TimeStepping: implement a minimum time step size

tjhei:


tjhei:
@EstherHeck FYI

EstherHeck:
Hi @tjhei 
I tested the minimum time step over the weekend and if I see it correctly, it did not work, or I did something wrong when declaring the parameters. I used these in the time stepping section: 

set CFL number                             = 0.0001
set Maximum time step                      = 5

subsection Time stepping
  set List of model names = convection time step, repeat on cutback
  set Minimum time step size = 0.000001268
  subsection Repeat on cutback
    set Relative repeat threshold = 0.1
  end
end


so I should not get time steps smaller than 1.27E-6. However, the statistics file shows time step sizes go down to as low as 5.461781845560e-12

EstherHeck:
Minimum time step works fine now. I first had the old version where there was no conversion to years, thats why it did not work before.

tjhei:
Great, this means this is ready for review.

gassmoeller (approved):
Looks great, ready to go.

PR #3863, opened 2020-10-07
GMG: template argument for number type

tjhei:
We still have problems switching to single precision in #3532 as some test results seem to break. As a first step, rebase that PR and switch the number type back to double.
This should have no effect on functionality and would be good to merge separately.

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1393,8 +1393,11 @@ namespace aspect
     // using MG transfer objects. This transfer is based on the same linear operator used to
     // transfer data inside a v-cycle.
     MGTransferMatrixFree<dim,GMGNumberType> transfer;
-
     transfer.build(dof_handler_projection);
+
+    // Explicitly pick the version with template argument double to convert
+    // double-valued active_viscosity_vector to GMGNumberType-valued
+    // level_viscosity_vector:
     transfer.template interpolate_to_mg<double>(dof_handler_projection,
                                                 level_viscosity_vector,
                                                 active_viscosity_vector);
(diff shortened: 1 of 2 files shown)


source/simulator/stokes_matrix_free.cc
```diff
@@ -1391,20 +1392,22 @@ namespace aspect
...
     // using MG transfer objects. This transfer is based on the same linear operator used to
     // transfer data inside a v-cycle.
-    MGTransferMatrixFree<dim,double> transfer;
+    MGTransferMatrixFree<dim,GMGNumberType> transfer;
+
     transfer.build(dof_handler_projection);
-    transfer.interpolate_to_mg(dof_handler_projection,
-                               level_viscosity_vector,
-                               active_viscosity_vector);
+    transfer.template interpolate_to_mg<double>(dof_handler_projection,
```
bangerth, line 1398: Is this correct then, or shouldn't this use the number type?

source/simulator/stokes_matrix_free.cc
```diff
@@ -1445,7 +1448,7 @@ namespace aspect
                     // of the evaluated viscosity on the active level.
                     for (unsigned int q=0; q<n_q_points; ++q)
                       level_viscosity_tables[level](cell,q)[i]
-                        = std::min(std::max(values_on_quad[q], min_el), max_el);
+                        = std::min(std::max(values_on_quad[q], (GMGNumberType)min_el), (GMGNumberType)max_el);
```
bangerth, line 1451: please use `static_cast` instead of the C-style cast.

bangerth (commented):
Nearly there.

source/simulator/stokes_matrix_free.cc
tjhei: Yes, it is. I left a comment.

tjhei:
updated.

PR #3865, opened 2020-10-08
Minor improvements to surface diffusion

gassmoeller:
Follow up to #3407 with a few cosmetic improvements.

### DIFF
--- a/benchmarks/diffusion_of_hill/analytical_topography.cc
+++ b/benchmarks/diffusion_of_hill/analytical_topography.cc
@@ -69,7 +69,11 @@ namespace aspect
       // loop over all of the surface cells and save the elevation to stored_value
       for (const auto &cell : this->get_triangulation().active_cell_iterators())
         if (cell->is_locally_owned() && cell->at_boundary())
+#if DEAL_II_VERSION_GTE(9,3,0)
+          for (const unsigned int face_no : cell->face_indices())
+#else
           for (unsigned int face_no = 0; face_no < GeometryInfo<dim>::faces_per_cell; ++face_no)
+#endif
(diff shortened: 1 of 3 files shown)


tjhei (approved):

benchmarks/diffusion_of_hill/analytical_topography.cc
```diff
@@ -69,7 +69,11 @@ namespace aspect
       // loop over all of the surface cells and save the elevation to stored_value
       for (const auto &cell : this->get_triangulation().active_cell_iterators())
         if (cell->is_locally_owned() && cell->at_boundary())
+#if DEAL_II_VERSION_GTE(9,2,0)
```
tjhei, line 72: isn't this only for 9.3?

gassmoeller: You are right. Fixed now.

source/mesh_deformation/interface.cc
```diff
@@ -1,5 +1,5 @@
 /*
-  Copyright (C) 2020 by the authors of the ASPECT code.
+  Copyright (C) 2014 - 2020 by the authors of the ASPECT code.
```
anne-glerum, line 2: Sorry for removing this! Originally it said `2011` instead of `2014`, is that important?

anne-glerum (approved):

source/mesh_deformation/interface.cc
gassmoeller: No big deal, I think the 2011 was wrong as well, this file was added in 2014 (I checked the git log). Ready from my side.

gassmoeller:
This seems to be approved.

PR #3868, opened 2020-10-19
fix melt migration with heat flux boundaries

jdannberg:
This fixes a bug described on the forum (https://community.geodynamics.org/t/melt-production-migration/1498/6?u=jdannberg). 
Basically, the melt migration routine calls its own set_assemblers function, so it didn't include the assembler for the heat flux boundary conditions. 

I think the same fix will have to be done for the Newton assembly as well. @MFraters, this wouldn't cause any problems for the assembler, since it is only an assembler for the temperature equation, correct?

* [x ] I have followed the [instructions for indenting my code](../blob/master/CONTRIBUTING.md#making-aspect-better).

### DIFF
--- a/source/simulator/melt.cc
+++ b/source/simulator/melt.cc
@@ -24,6 +24,7 @@
 #include <aspect/utilities.h>
 #include <aspect/citation_info.h>
 #include <aspect/mesh_deformation/interface.h>
+#include <aspect/simulator/assemblers/advection.h>
 #include <deal.II/base/signaling_nan.h>
 
 #include <deal.II/dofs/dof_tools.h>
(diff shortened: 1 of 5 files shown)


MFraters (approved):
Yes, this seems like the correct fix to me.  A fix also needs to be made for made for the newton solver indeed, thanks for pointing this out! 

I don't know why this did not end up in the Newton code since it was a copy of the same code with a few changes. I think it would be more appropriate to fix in the Newton case is fixing the underlying issue here for the newton solver directly. I think that the stokes part in set_default_assemblers should be separated (the first 2 lines) so that the code can be shared. The melt assembler looks like it shares way less code, so I am oke with keeping it separate, but we can discuss that in a separate pull request. 

Since there is someone waiting on a fix, I am fine with this being merged.

tjhei:
I would also think that the current setup is not great. Every time we add something, we would need to update several places. Instead, I would always run ``set_default_assemblers`` and let Newton/Melt _replace_ existing assemblers if necessary.

I agree with Menno, that this is independent of this fix and we can merge as is.

MFraters:
@jdannberg a bit late, but I think it would be useful if you would would also add a changelog entry, since it is an important fix. Can you make a pull request for that?

PR #3870, opened 2020-10-21
refactor set_default_assemblers to fix the Newton solver for heatflow boundaries

### LINKED ISSUE #3869: Solve problem with Newton solver with heat flux boundaries
This is a follow up issue to pull request #3868 (see that pull request for an initial discussion). The same needs to be fixed for the Newton solver, but it would make sense to do this in a way which is a bit more sustainable for future changes to the assembly. 

I don't think it should be too much work to just separate the stokes part and then for the normal and Newton assembles we can then just call `set_default_assemblers`. @tjhei suggested that we can also just set the assemblers and then replace them later. I am not sure how this would work since we push them somewhere in a assembly vector without knowing exactly where. @tjhei could you expand a little bit on that?

MFraters:
A slight refactoring of the code which should fix the problem with the heatflow boundaries and make it more future proof. 

followup to #3868. Fixes #3869.

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -1031,14 +1031,25 @@ namespace aspect
       /**
        * Determine, based on the run-time parameters of the current simulation,
        * which functions need to be called in order to assemble linear systems,
-       * matrices, and right hand side vectors. This function handles the
-       * default operation mode of ASPECT, i.e. without considering two-phase
-       * flow, or Newton solvers.
+       * matrices, and right hand side vectors for the advection. This function
+       * is used by both the default full Stokes solver and the Newton solvers,
+       * but not by the two-phase flow solver.
        *
        * This function is implemented in
        * <code>source/simulator/assembly.cc</code>.
        */
-      void set_default_assemblers ();
+      void set_advection_assemblers ();
+
+      /**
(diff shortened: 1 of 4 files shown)


include/aspect/simulator.h
```diff
@@ -1031,14 +1031,25 @@ namespace aspect
...
+      /**
+       * Determine, based on the run-time parameters of the current simulation,
+       * which functions need to be called in order to assemble linear systems,
+       * matrices, and right hand side vectors for the default full Stokes solver
+       * i.e. without considering two-phase flow, or Newton solvers.
+       *
+       * This function is implemented in
+       * <code>source/simulator/assembly.cc</code>.
+       */
+      void set_Stokes_assemblers ();
```
tjhei, line 1052:
```suggestion
      void set_stokes_assemblers ();
```
We typically keep "stokes" lower case (see ``solve_stokes``)

tjhei (commented):
Looks good (minus the two suggestions)

MFraters:
Thanks for the quick review! I have fixed your suggestions.

tjhei (approved):
Looks good to me.

gassmoeller (approved):

PR #3876, opened 2020-10-28
www: use new logo

tjhei:


tjhei:
I am not sure if the thumbnail will work like this, but there is no way for me to try this locally.

gassmoeller:
Thanks, I wanted to do this since the hackathon, but never got to it. Lets see if it works.

PR #3881, opened 2020-11-09
add depth averaging for compositional field mass

jdannberg:
This pull request adds a new property to the depth average postprocessor, called "composition mass". It computes the average of the product of compositional field and density. This is essential for computing mass balances of compositional fields in compressible models: Because the volume of a blob of compositionally distinct material changes as it moves up or down, the integral over just the composition will not stay the same (in other words, it will look like composition is not conserved). What is conserved is the mass of a compositional field, so this is the quantity one wants to look at in compressible models. 

### For all pull requests:

* [x] I modified an existing test for the new feature in the [tests/](../blob/master/tests/) directory.

Edit: I should also say that this changes the column names for the output of the "composition" property in the depth average files to the actual names of the compositional fields, rather than C_0, C_1, etc, which is apparent in the changes of the test results.


source/simulator/lateral_averaging.cc
```diff
@@ -308,6 +308,42 @@ namespace aspect
...
     };
+
+
+
+    template <int dim>
+    class FunctorDepthAverageFieldMass: public internal::FunctorBase<dim>
+    {
+      public:
+        FunctorDepthAverageFieldMass(const FEValuesExtractors::Scalar &field)
+          : field_(field)
```
gassmoeller, line 319: You would usually turn this around, call the stored member variable `field` and the function parameter `field_`. That way you can use the nicer variable name further down.

source/simulator/lateral_averaging.cc
```diff
@@ -679,10 +715,10 @@ namespace aspect
             functors.push_back(std_cxx14::make_unique<FunctorDepthAverageField<dim>>
                                (this->introspection().extractors.temperature));
           }
-        else if (property_names[property_index].substr(0,2) == "C_")
+        else if (this->introspection().compositional_name_exists(property_names[property_index]))
           {
             const unsigned int c =
-              Utilities::string_to_int(property_names[property_index].substr(2,std::string::npos));
+              this->introspection().compositional_index_for_name(property_names[property_index]);
```
gassmoeller, line 721: Nice improvement.

source/simulator/lateral_averaging.cc
```diff
@@ -725,6 +761,14 @@ namespace aspect
                                (this->introspection().extractors.velocities,
                                 &this->get_gravity_model()));
           }
+        else if (property_names[property_index].substr(property_names[property_index].size()-5) == "_mass")
```
gassmoeller, line 764: I am slightly concerned this creates conflicts with any other property called `something_mass`. Could you make this more secure by checking that the first part fits a compositional field and the last part is `_mass`? E.g.
```
else if (this->introspection().compositional_name_exists(property_names[property_index].substr(0, property_names[property_index].size()-5)) &&
property_names[property_index].substr(property_names[property_index].size()-5) == "_mass")
```

tests/adiabatic_conditions_composition_function/depth_average.gnuplot
```diff
@@ -4,74 +4,74 @@
 #
 # For a description of the GNUPLOT format see the GNUPLOT manual.
 #
-# <depth> <time> <temperature> <C_0> <adiabatic_temperature> <adiabatic_pressure> <adiabatic_density> <adiabatic_density_derivative> <velocity_magnitude> <sinking_velocity> <viscosity> <vertical_heat_flux> <vertical_mass_flux> 
-0 0 1613 0 1613 0 3300 0 7.06667e-16 7.80426e-17 1e+21 3.92932e-08 4.95593e-13 
-289000 0 1613 0 1613 0 3300 0 7.06667e-16 7.80426e-17 1e+21 3.92932e-08 4.95593e-13 
+# <depth> <time> <temperature> <C_1> <adiabatic_temperature> <adiabatic_pressure> <adiabatic_density> <adiabatic_density_derivative> <velocity_magnitude> <sinking_velocity> <viscosity> <vertical_heat_flux> <vertical_mass_flux> <C_1_mass>
```
gassmoeller, line 7: This is curious. Do we start our compositional field names with 1 internally? If so this new output is actually a fix.

source/postprocess/depth_average.cc
```diff
@@ -463,6 +463,10 @@ namespace aspect
 
             if ( output_all_variables || std::find( output_variables.begin(), output_variables.end(), "vertical mass flux") != output_variables.end() )
               variables.push_back("vertical_mass_flux");
+
+            if ( output_all_variables || std::find( output_variables.begin(), output_variables.end(), "composition mass") != output_variables.end() )
```
gassmoeller, line 467: I guess the position of these lines is a trade-off between keeping similar quantities together in the output (C_1 C_1_mass) and breaking compatibility of postprocessing scripts by changing the order of existing columns?

gassmoeller (commented):
This looks good, I just have some minor comments. Maybe it is time to think about not computing all available quantities in the depth average postprocessor as default but choose a reasonable subset (if we break compatibility for that we may as well group C_1 and C_1_mass together). Anyway probably not in this PR, but something we could look into in the future.

tests/adiabatic_conditions_composition_function/depth_average.gnuplot
jdannberg: Yes, this is correct. We start with C_1 for the names, so this is a fix.

source/postprocess/depth_average.cc
jdannberg: Yes.

source/simulator/lateral_averaging.cc
jdannberg: This is consistent with the implementation of all the other functors.

jdannberg:
Thank you for the review! I think I addressed your comments. Could you have another look?

gassmoeller (approved):

PR #3884, opened 2020-11-13
Addition: Composite viscoplastic now includes damped Drucker-Prager element

bobmyhill:
The CompositeViscoPlastic rheology now includes a damped Drucker-Prager element. The rheology model now looks like this:
image

This PR also renames compute_viscosity to compute_composition_viscosity (as it only computes the viscosity for a single compositional field), and adds a new compute_viscosity function that loops over all compositional fields to calculate the total effective viscosity.


include/aspect/material_model/rheology/composite_visco_plastic.h
```diff
@@ -94,30 +112,36 @@ namespace aspect
...
           std::unique_ptr<Rheology::PeierlsCreep<dim>> peierls_creep;
+          std::unique_ptr<Rheology::DruckerPrager<dim>> drucker_prager;
+
+          DruckerPragerParameters drucker_prager_parameters;
 
-          std::vector<double> min_viscosities;
-          std::vector<double> max_viscosities;
+          unsigned int number_of_compositions;
+          double min_viscosity;
+          double max_viscosity;
```
naliboff, line 144: Can you remind me the reason for switching to single min/max viscosities? I recall we discussed this, but cannot remember the details. In any event, I agree it makes sense to have a single min/max viscosity that applied to all deformation mechanisms.

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -70,11 +72,59 @@ namespace aspect
...
                                                      const double temperature,
-                                                     const unsigned int composition,
+                                                     const std::vector<double> &volume_fractions,
                                                      const SymmetricTensor<2,dim> &strain_rate,
                                                      std::vector<double> &partial_strain_rates,
                                                      const std::vector<double> &phase_function_values,
                                                      const std::vector<unsigned int> &n_phases_per_composition) const
+      {
+        double viscosity = 0.;
+        partial_strain_rates.resize(5, 0.);
```
naliboff, line 82: What is the initial size of this vector that the function receives? Is it only just a single value (background strain rate) that is used as the initial guess?

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -70,11 +72,59 @@ namespace aspect
...
+
+        // Isostrain averaging
+        double total_volume_fraction = 1.;
+        for (unsigned int composition=0; composition < number_of_compositions; ++composition)
+          {
+            // Only include the contribution to the viscosity
+            // from a given composition if the volume fraction exceeds
+            // a certain (small) fraction.
+            if (volume_fractions[composition] > 1.e-5)
```
naliboff, line 91: Why not set this value to 0?

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -256,6 +323,11 @@ namespace aspect
                            Patterns::Bool (),
                            "Whether to include Peierls creep in the rheological formulation.");
 
+        prm.declare_entry ("Include Drucker Prager plasticity", "true",
```
naliboff, line 326: In order to avoid confusion with other parameters in the drucker_prager rheology module, I wonder if she specify that the prm entries include some reference to "composite rheology"?

```suggestion
        prm.declare_entry ("Include Drucker Prager plasticity in composite rheology", "true",
```

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -256,6 +323,11 @@ namespace aspect
...
+                           Patterns::Bool (),
+                           "Whether to include Drucker-Prager plasticity in the rheological formulation.");
```
naliboff, line 328:
```suggestion
                           "Whether to include Drucker-Prager plasticity in the composite rheological formulation.");
```

Same suggestion as above ...

naliboff (commented):
@bobmyhill - Thanks for this contribution and sorry it took me so long to review it! A few small technical questions, comments, and suggestions, but otherwise the code is in great shape. Thank you again for all your work this!

include/aspect/material_model/rheology/composite_visco_plastic.h
bobmyhill: I think the argument was that as it's just a stabilising parameter, we shouldn't give the user freedom for no good reason. 
Although I was originally ambivalent about the choice, single min/max parameters will also make for tidier formulations when we implement different averaging schemes.

source/material_model/rheology/composite_visco_plastic.cc
bobmyhill: The vector could be any size, with values equal to zero. The resizing seemed like the safest option in case an empty vector was passed, rather than a vector of length 5. The alternative would be to check `(size == 5)`, but I'm don't think that would be significantly faster.

source/material_model/rheology/composite_visco_plastic.cc
bobmyhill: My initial thought was to speed up calculations where one or more volume fractions were small-ish, but you're right, that might introduce (small) unexpected variations. I've now set it to 2.*std::numeric_limits<double>::epsilon().

source/material_model/rheology/composite_visco_plastic.cc
bobmyhill: Sure, this is a good idea for all the `Include mechanism` parameters.

source/material_model/rheology/composite_visco_plastic.cc
bobmyhill: Makes sense. Done.

include/aspect/material_model/rheology/composite_visco_plastic.h
naliboff: 👍

naliboff (approved):
@bobmyhill - Thanks for the changes! As, there will likely be a series of adjustments done when we test this with the visco plastic model + realistic scenarios, I don't see the need for any additional minor changes.

PR #3886, opened 2020-11-14
Refactor viscoelastic viscosity

bobmyhill:
This PR ensures that the Elasticity<dim>:: calculate_viscoelastic_viscosity outputs the correct value even when the viscous viscosity is infinite.

### DIFF
--- a/source/material_model/rheology/elasticity.cc
+++ b/source/material_model/rheology/elasticity.cc
@@ -358,8 +358,8 @@ namespace aspect
       calculate_viscoelastic_viscosity (const double viscosity,
                                         const double elastic_shear_modulus) const
       {
-        const double dte = elastic_timestep();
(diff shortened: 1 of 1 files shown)


bobmyhill:
Hi @naliboff, would you mind flagging this as ready to test? The PR only rearranges one equation to stop infinities appearing in the numerator and denominator.

naliboff:
@bobmyhill - Is an infinite viscous viscosity a possible input for one of the new composite rheology models?

bobmyhill:
@naliboff - If there are no viscous elements, then the viscous viscosity is by definition infinite. 
In the composite models, we treat each deformation mechanism separately - each element sees a stress or strain rate, and deforms accordingly. So it doesn't make sense to define a viscous viscosity inside elasticity.cc (the elastic element knows nothing about the other elements; it just knows about the stress or strain rate).

naliboff (approved):
Looks good, thanks!

Issue #3887, opened 2020-11-14
Possible factor 2 error in elasticity.cc


bobmyhill:
On line 378 of rheology/elasticity.cc (in function calculate_viscoelastic_strain_rate), should the line 
`const SymmetricTensor<2,dim> edot = 2. * (deviator(strain_rate)) + stress / (shear_modulus * elastic_timestep());`
instead read
`const SymmetricTensor<2,dim> edot = deviator(strain_rate) + 0.5 * stress / (shear_modulus * elastic_timestep());`
?

bobmyhill:
At the moment this function is only used in visco_plastic.cc, which corrects the factor two error on line 319.

naliboff:
@bobmyhill - sorry, just saw this. For some reason I've recently only been getting selected GitHub notifications via email.

Short answer - @dansand and I had multiple discussions about how this should be defined a while, but off the top of my head I can't think of a reason to support one formulation over the other. While your proposed method is indeed consistent with how the strain rate is defined normally in ASPECT, the current formulation matches benchmarks. Let's see how the tests change with #3888 :-)

bobmyhill:
Resolved by #3888 (now merged).

PR #3888, opened 2020-11-14
removed factor two in elasticity strain rate calculation

bobmyhill:
The elasticity rheology module strain rate function now returns a strain rate invariant consistent with the rest of ASPECT. 

This PR addresses Issue #3887.


bobmyhill:
Only one test fails. I think this might be due to round-off error, but perhaps I'm not understanding the code correctly. 
- The original code has: 
   
   (where "*s*" is `edot` in `Elasticity::calculate_viscoelastic_strain_rate`, and this calculation is split across elasticity.cc and visco_plastic.cc)
 - This PR changes that to 
   
- All the terms in the second invariant involve multiplying two elements of the initial array, so that *should* mean that 
   

Could someone confirm that for me? 

The thought process behind this PR runs as follows:
- If we set the shear modulus equal to zero, all viscoelastic functions should behave identically to their viscous counterparts. 
- Viscous strain rate is equal to `std::sqrt(std::fabs(second_invariant(deviator(viscous_strain_rate)));` in many places in ASPECT.
- In contrast, `calculate_viscoelastic_strain_rate` in elasticity.cc returns `std::sqrt(std::fabs(second_invariant(2*deviator(viscous_strain_rate)));`
- That factor of two is then removed in visco_plastic.cc.
- I expect more material models to use elasticity.cc in the future, so it would be good to sort this out now.

bobmyhill:
I've found the source of the lingering test failure. On lines 296 and 297, there's an if statement for the first time step:
`if (use_reference_strainrate == true)`
`                current_edot_ii = ref_strain_rate;`

In the current version of master, this current_edot_ii is still divided by two after the end of the if;else block. That is unexpected behaviour, and I think it should be fixed. I'll update the output of the failing test accordingly.

source/material_model/visco_plastic.cc
```diff
@@ -310,13 +310,7 @@ namespace aspect
...
 
               // Step 2b: calculate current (viscous + elastic) stress magnitude
-              current_stress = viscosity_pre_yield * current_edot_ii;
-
-              // The viscoelastic strain rate is divided by 2 here as the Drucker Prager
-              // viscosity calculation below assumes stress = 2 * viscosity * strain_rate_invariant,
-              // whereas the combined viscoelastic + viscous stresses already include the
-              // 2x factor (see computation of edot inside elastic_rheology).
-              current_edot_ii /= 2.;
+              current_stress = 2. * viscosity_pre_yield * current_edot_ii;
```
naliboff, line 313: As `current_stress = 2. * viscosity_pre_yield * current_edot_ii` now holds true regardless if viscoelasticity is turned on or off, should we move this line (and the equivalent one above, line 290) out of the if-else statement?

source/material_model/rheology/elasticity.cc
```diff
@@ -371,7 +371,7 @@ namespace aspect
                                          const SymmetricTensor<2,dim> &stress,
                                          const double shear_modulus) const
       {
-        const SymmetricTensor<2,dim> edot = 2. * (deviator(strain_rate)) + stress /
+        const SymmetricTensor<2,dim> edot = deviator(strain_rate) + 0.5*stress /
```
naliboff, line 374: I'm actually surprised this did not alter more of the VEP tests. At first glance, the change above combined with the one in `visco_plastic`  (remove `current_edot_ii /= 2.`) should not give equivalent values, as you first take the square root of the second invariant (line 377 below). I must be missing something here :-)

naliboff (commented):
@bobmyhill - Thanks for this PR! I have one minor suggestion and a question regarding why this does not affect the test output, but otherwise this should be good to go. Indeed, great that this will allow removing the confusing factor of 2 division of the viscoelastic strain rate.

source/material_model/rheology/elasticity.cc
naliboff: @bobmyhill - Never mind, I just saw your updated documentation in the comment above. Reading through that now.

naliboff: > All the terms in the second invariant involve multiplying two elements of the initial tensor, so that should mean that
>    

Well that's convenient 😄

source/material_model/visco_plastic.cc
bobmyhill: Good point. Now changed.

source/material_model/visco_plastic.cc
```diff
@@ -308,17 +302,11 @@ namespace aspect
...
-              current_stress = viscosity_pre_yield * current_edot_ii;
-
-              // The viscoelastic strain rate is divided by 2 here as the Drucker Prager
-              // viscosity calculation below assumes stress = 2 * viscosity * strain_rate_invariant,
-              // whereas the combined viscoelastic + viscous stresses already include the
-              // 2x factor (see computation of edot inside elastic_rheology).
-              current_edot_ii /= 2.;
             }
 
+          // Step 2b: calculate current (viscous + elastic) stress magnitude
```
naliboff, line 307:
```suggestion
          // Step 2b: calculate current (viscous or viscous + elastic) stress magnitude
```

source/material_model/rheology/elasticity.cc
naliboff, line 374: @bobmyhill - I think we should add a comment here to indicate why multiplying by 0.5 is necessary. Feel free to significantly modify the suggestion below. After, that this should be good to go.
```suggestion
        // Multiplying the elastic stress below by a factor of 0.5 is necessary to be consistent with 
        // with ASPECT's definition of stress, where stress = 2 * viscosity * strain_rate_second_invariant.
        const SymmetricTensor<2,dim> edot = deviator(strain_rate) + 0.5*stress /
```

naliboff (approved):
👍 Thanks!

PR #3889, opened 2020-11-15
fix undefined behavior in visco_plastic

tjhei:


gassmoeller (approved):
:open_mouth: I am surprised this compiled at all.

tjhei:
> I am surprised this compiled at all.

It probably compiled to return 0 but then we would overwrite memory at runtime...

PR #3891, opened 2020-11-15
add missing instantiation

tjhei:
internal::GPlatesLookup is used in boundary_velocity_residual_statistics
and clang correctly complains that we don't explicitly instantiate the
template class. Fix this.


gassmoeller (approved):
Thanks for the fix!

PR #3892, opened 2020-11-16
remove virtual when using override

tjhei:
follow-up to #3890

### DIFF
--- a/include/aspect/time_stepping/function.h
+++ b/include/aspect/time_stepping/function.h
@@ -49,7 +49,6 @@ namespace aspect
         /**
          * The main execute() function.
          */
-        virtual
         double
         execute() override;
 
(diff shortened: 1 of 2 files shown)


bangerth:
We're not really consistent in whether or not we use `virtual` on overridden functions. Are you suggesting to get rid of it everywhere?

tjhei:
Not sure if we ever made a decision about it, but @gassmoeller asked about it here: https://github.com/geodynamics/aspect/pull/3890#issuecomment-728212841

See
https://github.com/geodynamics/aspect/issues/3258 and https://github.com/geodynamics/aspect/pull/3249 for related discussions.

bangerth:
Ah, I see. I might not agree with it (I think that `override` is like `const` to let the compiler help me -- an insurance against mistakes, in this case in spelling of function names and the types of arguments -- whereas the duplicated `virtual` is for human readers). I think it is a mental stretch to infer from seeing `override` that the function is `virtual` when one does not make that stretch in the base class.

But that's besides the point. I should have spoken up in #3258. Now we should strive for consistency, so this is ok.

PR #3893, opened 2020-11-16
Modified the plugin for depths more than the input viscosity file.

alarshi:
This PR addresses a minor change in the exisiting test plugin which scales the aveage rheology of the material model to an input ascii reference data, to use the last depth layer when the dephs are greater than the input ascii data rheology.

### For all pull requests:

### DIFF
--- a/tests/rheology_scaled_profile.cc
+++ b/tests/rheology_scaled_profile.cc
@@ -152,9 +152,15 @@ namespace aspect
                                           std::lower_bound(reference_viscosity_coordinates.begin(),
                                                            reference_viscosity_coordinates.end(),
                                                            depth));
-              if (depth_index > 0)
-                depth_index -= 1;
             }
+          // depth_index calculated above is the first layer boundary larger than depth,
+          // the correct layer index is then one less than this (except for depth == depth_bounds[0],
+          // in which case the depth_bounds index is also the layer_index, namely 0).
+          // For depths > depth_bounds[n-1], where, n is the size of the depth_bounds, we need to use
+          // the depth_bound one less than this, i.e. depth_bounds[n-2] to compute the averages of the
+          // last layer.
+          if (depth_index > 0)
+            depth_index -= 1;
 
           // When evaluating reference viscosity, evaluate at the next lower depth that is stored
           // in the reference profile instead of the actual depth. This makes the profile piecewise


gassmoeller (commented):
This is correct, but I think it is a bit confusing. Can you instead leave line 147 as it was and instead move lines 155 and 156 out of the `else` branch? That should be equivalent, but easier to understand.

tjhei:
> Can you instead leave line 147 as it was and instead move lines 155 and 156 out of the `else` branch? That should be equivalent, but easier to understand.

Are you sure? I would think the ``depth > back()`` case should not be adjusted by -1. Or am I missing why this change is done?

@alarshi can you leave a comment right above the new code to say why this needs to be done? I assume the reasoning is that you want to use the value at the top of the layer for a point (otherwise the last two layers have the same value while the value in the first layer is not used).

tjhei:
Thanks for the update. I am still confused why ``depth_index = reference_viscosity_coordinates.size() - 2;`` is the correct value to use if the depth is bigger than all values (this is caused by Rene's suggestion to move the if outside the else branch. Shouldn't ``-1`` be correct?

alarshi:
@tjhei: I think without using an effective ```depth_index = reference_viscosity_coordinates.size()  - 2 ```, which is equivalent to Rene's suggestion of moving the else branch outside, we will get ```average_viscosity``` of almost 0.  For example, a reference viscosity profile with  2 layers having depth bounds at 0, 1e6, and 2e6 and a geometry model in which maximum depth is 2.5e6, the ```depth_index``` would be 2 previously (```reference_viscosity_coordinates.size() = 3```) , but the function ```compute_lateral_average()``` in ```lateral_averaging.cc``` only computes average for ```depth_bounds.front() <  depth < depth_bounds.back(),``` (ln 506 to ln 516 in ```lateral_averaging.cc```)  otherwise the variable, ```values_all``` and the averaged value, ```values[property][i]``` (ln 533 in ```lateral_averaging.cc```) is 0. This makes the compute_viscosity_scaling in ```rheology_scaled_plugin.cc``` throw a FPE.

tjhei:
Oh, I get it now. Thank you for the detailed explanation.

tests/rheology_scaled_profile.cc
```diff
@@ -152,9 +152,15 @@ namespace aspect
                                           std::lower_bound(reference_viscosity_coordinates.begin(),
                                                            reference_viscosity_coordinates.end(),
                                                            depth));
-              if (depth_index > 0)
-                depth_index -= 1;
             }
+          // Depth index calculated above is the first layer boundary larger than depth,
```
gassmoeller, line 156:
```suggestion
          // depth_index calculated above is the first layer boundary larger than depth,
```

tests/rheology_scaled_profile.cc
```diff
@@ -152,9 +152,15 @@ namespace aspect
...
+          // the correct layer index is then one less than this (except for depth == depth_bounds[0],
+          // in which case the depth_bounds index is also the layer_index, namely 0).
+          // For depths > depth_bounds[n-1], where, n is the size of the depth_bounds, we need to use
+          // the depth_bound ones less than this, i.e. depth_bounds[n-2] to compute the averages of the
```
gassmoeller, line 160:
```suggestion
          // the depth_bound one less than this, i.e. depth_bounds[n-2] to compute the averages of the
```

gassmoeller (commented):
@alarshi Could you update the documentation with my suggestions? Afterwards I think this is ready to merge.

alarshi:
@gassmoeller: Thank you! I addressed your comments in the latest commit.

gassmoeller (approved):

PR #3895, opened 2020-11-18
clang-tidy update

tjhei:
- update clang-tidy rules (taken from deal.II)
- run with new clang-tidy 10
- update and fix contrib/utilities/run_clang_tidy.sh
- apply fixes:
1. replace bind() by lambdas
2. use range for loop
3. various smaller fixes


source/simulator/core.cc
```diff
@@ -150,99 +150,102 @@ namespace aspect
...
-    post_geometry_model_creation_action (std::bind (&Parameters<dim>::parse_geometry_dependent_parameters,
-                                                    std::ref(parameters),
-                                                    std::ref(prm),
-                                                    std::cref(*geometry_model))),
-    material_model (MaterialModel::create_material_model<dim>(prm)),
-    gravity_model (GravityModel::create_gravity_model<dim>(prm)),
-    prescribed_stokes_solution (PrescribedStokesSolution::create_prescribed_stokes_solution<dim>(prm)),
-    adiabatic_conditions (AdiabaticConditions::create_adiabatic_conditions<dim>(prm)),
+      [&]()
+  {
```
tjhei, line 154: astyle sadly gets confused with the indentation here. Should we leave the bind() call in here?

source/postprocess/heat_flux_statistics.cc
```diff
@@ -61,6 +61,7 @@ namespace aspect
         boundary_indicators
           = this->get_geometry_model().get_used_boundary_indicators ();
         std::vector<double> local_values;
+        local_values.reserve(boundary_indicators.size());
         for (const auto p : boundary_indicators)
           local_values.push_back (local_boundary_fluxes[p]);
```
MFraters, line 66: Why doesn't this use emplace back here, since push_back was replaced by emplace back in a few other files?

source/postprocess/visualization.cc
```diff
@@ -343,6 +341,7 @@ namespace aspect
         // the global .visit file needs the relative path because it sits a
         // directory above
         std::vector<std::string> filenames_with_path;
+        filenames_with_path.reserve(filenames.size());
         for (const auto &filename : filenames)
           {
             filenames_with_path.push_back("solution/" + filename);
```
MFraters, line 347: same question about push back vs emplace back here

source/simulator/core.cc
MFraters: Is there a real downside of using bind? If not I would personally prefer the readability for now. Also in the other places where bind has been replaced.

MFraters (commented):
Wow, those are a lot of changes. I really like to loop modernizations. Did clang-tidy autofix the code or did you have to fix it all by hand? 

I do have a few small comments/questions, but I am generally fine with it being merged.

source/material_model/steinberger.cc
```diff
@@ -195,8 +195,8 @@ namespace aspect
       if (use_lateral_average_temperature)
         {
           this->get_lateral_averaging().get_temperature_averages(avg_temp);
-          for (unsigned int i = 0; i < avg_temp.size(); ++i)
-            AssertThrow(numbers::is_finite(avg_temp[i]),
+          for (double i : avg_temp)
```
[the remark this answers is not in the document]
tjhei: avg_temp is a class member. Maybe we do this separately after this PR is merged?

source/postprocess/heat_flux_statistics.cc
tjhei: I think it might not trigger for type double as it doesn't matter there, but I will replace it as well.

tjhei:
> Is there a real downside of using bind?

Not really. I agree, let's keep it for now. I updated the patch with (most of) your suggestions.

tjhei:
> Did clang-tidy autofix the code or did you have to fix it all by hand?

This uses the ``-fix`` setting of clang-tidy with some manual cleanup (variable names).

MFraters (approved):
looks good to me. 

> avg_temp is a class member. Maybe we do this separately after this PR is merged?

Sure, and since I am the one nitpicking here, I will put in the effort to make the change ;)

PR #3896, opened 2020-11-18
Add option to use the adiabatic temperature in the multicomponent incompressible density equation

naliboff:
This PR modifies the `multicomponent_incompressible` equation of state module to use the adiabatic temperature for the reference temperature (density formulation), if adiabatic heating is turned on.

This modification will ensure that incompressible models with adiabatic and shear heating are consistent with the extended Boussinesq approximation.

A question for reviewers - should a similar change also be added to the `linearized_incompressible` EOS module?


source/material_model/equation_of_state/multicomponent_incompressible.cc
```diff
@@ -36,9 +37,18 @@ namespace aspect
                const unsigned int input_index,
                MaterialModel::EquationOfStateOutputs<dim> &out) const
       {
+
+        // If adiabatic heating is used, the reference temperature used to calculate density should be the adiabatic
+        // temperature at the current position. This definition is consistent with the Extended Boussinesq approximation.
```
anne-glerum, line 42: Capital A on approximation?

tests/visco_plastic_adiabatic_heating_density.prm
```diff
@@ -0,0 +1,113 @@
...
+# adiabatic heating is turned on. The formula specified
+# in this test follows the Extended Boussinesq approximation.
+# The initial (constant) temperature throughout the model
+# has the same value as the Adiabatic surface temperature.
+# As a result, the density at the top (surface) of the model
+# should  be equal to the background value specified in the 
+# material model (3300):
+#   density = background_density * ( 1 - thermal_expansivity * (temperature - reference_temperature)),
+# where the reference_temperature is the adiabatic temperature.
+# At the base of the model the density should be lower than at the surface
```
anne-glerum, line 13: Actually, since the test only does the 1st timestep, in which the temperature is prescribed, the density is constant.

tests/visco_plastic_adiabatic_heating_density.prm
```diff
@@ -0,0 +1,113 @@
...
+# due to the increase in adiabatic temperature with depth.
```
anne-glerum, line 14: I would add that normally we would use an adiabatic initial condition, and the density would then be constant with depth.

tests/visco_plastic_adiabatic_heating_density.prm
```diff
@@ -0,0 +1,113 @@
...
+subsection Gravity model
+  set Model name = vertical
+  subsection Vertical
+    set Magnitude = 10.0
+  end
+end
+
+# Post processing
+subsection Postprocess
+  set List of postprocessors = mass flux statistics, material statistics, temperature statistics, visualization
```
anne-glerum, line 106: Doesn't really matter, but the mass flux statistics won't tell us much as the boundaries are all free slip. Maybe remove?

anne-glerum (commented):
Hi John, thanks for fixing this! The test is also a very clear user example on how to set up EBA (I don't think that's discussed elsewhere). 

In previous conversations with @gassmoeller, we decided to remove the user-set reference temperature completely. I think it would have to be replaced with the Adiabatic surface temperature. Should we do that now altogether @gassmoeller ?

tests/visco_plastic_adiabatic_heating_density.prm
naliboff: I think the density should vary, given that the calculated density in the first time step does take into account the adiabatic profile. 

However, I incorrectly stated the density should be lower at the surface, when in fact it should be higher (temperature < adiabatic temperature at base). I confirmed the calculated density at the base of the model (solution file) is consistent with the difference between the temperature and adiabatic temperature. 

Does this resolve the issue? I also updated the parameter documentation to make this more clear and address your comment below.

naliboff:
@anne-glerum - Thanks for the review! I believe I've addressed all of the comments, aside from the issue regarding the removing the reference temperature parameter:

> In previous conversations with @gassmoeller, we decided to remove the user-set reference temperature completely. I think it would have to be replaced with the Adiabatic surface temperature. Should we do that now altogether @gassmoeller ?

My preference would be to do that in a different PR for the following reasons:

- It will break a huge number of tests
- It may break backward compatibility 
- We likely need to update a large number of parameter files 
- Given the above factors, it may take quite some time to do the PR start to finish

I agree that we should eventually remove the reference temperature, but it would be nice to have the feature in this PR implemented soon for a multitude of projects.

tests/visco_plastic_adiabatic_heating_density.prm
anne-glerum: Sorry, I was confused by the average density being reported as exactly the reference value 3300. I think that's just rounded off, as the gnuplot file does report the density increasing with depth to 3300.4. 

I only wonder about the `adiabatic density` that is reported. This is computed by querying the material model for the density using the adiabatic temperature (>= 1573) as the temperature, at a moment when the adiabatic profile is not initialized yet. Thus the `reference_temperature` is taken to be the `reference_T` (273), and the adiabatic density turns out to be something very different from the actual density, namely 3256.68. The difference with `rho_0` (43) is much larger than the change in density along the adiabat (0.4). This seems wrong to me, but I'm not sure how to fix it. Perhaps a third condition where `this->include_adiabatic_heating() && !this->get_adiabatic_conditions().is_initialized()` where the surface adiabatic temperature is taken as reference temperature (since you want to keep the `reference_T` for now)?

source/material_model/equation_of_state/multicomponent_incompressible.cc
anne-glerum: Could you capitalize it in the prm file as well?

tests/visco_plastic_adiabatic_heating_density.prm
naliboff: @anne-glerum - good catch, that is definitely an issue! I ran the test for more than one time step and as anticipated the adiabatic density does not change. 

I'm actually a bit confused about the order of operations here, which I've tried to write out below: 

1. The adiabatic density in this case is being calculated in `adiabatic_conditions/compute_profile.cc`
2. The density in `adiabatic_conditions/compute_profile.cc` is derived from the material model, which prior to initialization of the adiabatic profile (happening concurrently) is use `reference_T` (273) rather than the adiabatic surface temperature?
3. Once the adiabatic conditions have been initialized, the adiabatic profile is not updated afterwards to account for the adiabatic surface temperature.

Is the above an accurate summary of the issue?

I agree adding the following third condition to use the surface adiabatic temperature will likely fix the issue:
`this->include_adiabatic_heating() && !this->get_adiabatic_conditions().is_initialized()`

I'll give it a shot and fingers crossed! More reason to remove `reference_Temperature` sooner rather than later.

source/material_model/equation_of_state/multicomponent_incompressible.cc
```diff
@@ -36,9 +37,18 @@ namespace aspect
                const unsigned int input_index,
                MaterialModel::EquationOfStateOutputs<dim> &out) const
       {
+
+        // If adiabatic heating is used, the reference temperature used to calculate density should be the adiabatic
+        // temperature at the current position. This definition is consistent with the Extended Boussinesq Approximation.
+        const double reference_temperature = (this->include_adiabatic_heating() && this->get_adiabatic_conditions().is_initialized()
```
gassmoeller, line 43: Could you try to remove the second condition `is_initialized`? In the current state I am concerned that we still compute the wrong adiabatic profile (i.e. the adiabatic profile would still have a decreasing density with depth). The computation of the adiabatic profile is written in a way that even during initialization pressure and temperature at the current position are already known when evaluating the material model (you can just not ask for temperature or pressure at larger depths, or for density at the current depth). If it does not work it should trigger a signalling NaN.

gassmoeller (commented):
Thanks for fixing this and except for my one comment it looks good! I agree let's remove the reference parameter in a later PR, that will shake up a lot of models and tests (we may need to coordinate that with other breaking changes).

tests/visco_plastic_adiabatic_heating_density.prm
naliboff: Adding `this->include_adiabatic_heating() && !this->get_adiabatic_conditions().is_initialized()` did the trick 👍 - the adiabatic density now looks correct.

source/material_model/equation_of_state/multicomponent_incompressible.cc
naliboff: @gassmoeller - I tried removing `is_initialized?` and this gave a constant adiabatic density throughout the model domain. 

Per @anne-glerum suggestion, I added a third condition (last commit) where the `Adiabatic surface temperature` is used as the reference density prior to initialization of the adiabatic conditions. The resulting adiabatic densities are consistent with the adiabatic temperature and adiabatic surface temperature values. 

Is this a reasonable solution? 

> If it does not work it should trigger a signalling NaN.
Sorry, I didn't follow this fully. Should a signaling NaN in theory be triggered with the current code structure?

naliboff:
@anne-glerum @gassmoeller - Are there any other changes I should make here, or are you happy with the current state?

gassmoeller:
> I tried removing is_initialized? and this gave a constant adiabatic density throughout the model domain.

But this sounds correct. On the adiabatic profile this function should return a constant density, and if you remove the check for `is_initialized` and always use the following is this not correct?
```
        if (this->include_adiabatic_heating())
          reference_temperature = this->get_adiabatic_conditions().temperature(in.position[input_index]);
        else
          reference_temperature = reference_T;
```

> Sorry, I didn't follow this fully. Should a signaling NaN in theory be triggered with the current code structure?

I meant if the suggested code above is wrong, it should trigger a floating point exception. Since you did not see that when trying it, the code should be fine.

naliboff:
> But this sounds correct. On the adiabatic profile this function should return a constant density, and if you remove the check for `is_initialized` and always use the following is this not correct?
> 
> 
>         if (this->include_adiabatic_heating())
>           reference_temperature = this->get_adiabatic_conditions().temperature(in.position[input_index]);
>         else
>           reference_temperature = reference_T;

@gassmoeller - Thanks for the clarification. The above does not give a floating point exception, but does not produce the correct adiabatic density profile. 

The check for `is_initialized` is needed along with the use of the adiabatic surface temperature for the following case: `(this->include_adiabatic_heating() && !this->get_adiabatic_conditions().is_initialized())`

Checking for all three cases (current code structure) is giving the current density and adiabatic density profiles.

gassmoeller:
> Checking for all three cases (current code structure) is giving the current density and adiabatic density profiles.

But is the current adiabatic density profile correct? I thought the current code is bugged, because it uses a constant reference temperature to compute the adiabatic density profile (resulting in a decreasing density with depth).

naliboff:
> But is the current adiabatic density profile correct? 

The current adiabatic density profile (i.e., the one calculated in this PR) is indeed now correct. It was incorrect when only doing:
```
    if (this->include_adiabatic_heating())
      reference_temperature = this->get_adiabatic_conditions().temperature(in.position[input_index]);
    else
      reference_temperature = reference_T;
```
If by current code you mean the code prior to this PR, then yes I think the adiabatic profile was also wrong when using the incompressible formulation with adiabatic heating.

> I thought the current code is bugged, because it uses a constant reference temperature to compute the adiabatic density profile (resulting in a decreasing density with depth for the EBA).

The absolute density being calculated was also incorrect as well due to not using the adiabatic temperature.

For the current test, we have the following conditions:
- Background density of 3300 (defined in parameter file)
- Constant initial model temperature of 1573
- Adiabatic surface temperature of 1573
- Extended BA approximation 

As currently constructed, this gives the following reference temperatures:
- reference temperature = adiabatic surface temperature before the adiabatic conditions are initialized.
- reference temperature = adiabatic temperature after the adiabatic conditions are initialized

At the top of the model, the initial temperature, adiabatic temperature, and adiabatic surface temperature are all equal (1573), so both the adiabatic and real densities are exactly 3300.

At the base of the model, the model temperature is 1573, while the adiabatic temperature is 1585.63. 
This should give:
real density = 3300. * (1. - 1.e-5 * (1573. - 1585.63)) = 3300.41679
adiabatic density = 3300. * (1. - 1.e-5 * (1585.63 - 1573)) = 3299.58321

The density values above are consistent with the solution values. 

@gassmoeller - Do the above values/logic make sense or is something still off?

gassmoeller:
I think there is still something off. First: The adiabatic density at the bottom of the model should be exactly 3300, just as the adiabatic density at the surface (because it should be computed as `rho * (1-alpha (T-T_adiabatic(depth)))`. The different result you get (3299.58) is caused by the second if condition, which computes the adiabatic density as `rho * (1-alpha (T-T_adiabatic(0)))` instead. 

Second the model setup is somewhat weird (but I do not know if this is intentional): If you use the EBA then your initial model temperature should not be constant with depth. It needs to include the adiabatic gradient, because you have adiabatic heating enabled. Using a constant temperature would cause material that rises to be cooler than the surrounding material and resist this rise. But for a constant model temperature in an EBA model you would expect a real density that increases with depth just as you observe, so that tells us that while the model setup is somewhat weird the computation of the real density is correct.

naliboff:
@gassmoeller - Thanks for the follow-up. 

A question regarding this statement:

>  The adiabatic density at the bottom of the model should be exactly 3300, just as the adiabatic density at the surface (because it should be computed as rho * (1-alpha (T-T_adiabatic(depth))).

Initially I would have thought that the adiabatic density should only be 3300 at both the top and bottom of the model if `T = T_adiabatic(depth)`, which is not the case in the model presented here. 

In `source/adiabatic_conditions/compute_profile.cc`, when the density profile is constructed it looks like the adiabatic heating term is added to the temperature (if adiabatic heating is enabled). Is this why the adiabatic density should be constant with depth?

Indeed, if we only use the condition `if (this->include_adiabatic_heating())` (per your suggestion), the adiabatic density is in fact constant with depth (3300 at top and bottom). 

> Second the model setup is somewhat weird (but I do not know if this is intentional):

Yes, this was intentional but perhaps not wise :-) I wanted a case where the initial temperature differed from the adiabatic temperature profile to make sure I was getting the correct real density. 

>  If you use the EBA then your initial model temperature should not be constant with depth. It needs to include the adiabatic gradient, because you have adiabatic heating enabled.

As a case example, consider a model where one wants to have a conductive temperature profile in the lithosphere and an adiabatic profile in the asthenosphere. Is the correct approach to first calculate the conductive profile then add the adiabatic heating profile to the conductive profile?

anne-glerum:
> Initially I would have thought that the adiabatic density should only be 3300 at both the top and bottom of the model if `T = T_adiabatic(depth)`, which is not the case in the model presented here.

But in case of computing the adiabatic density, the temperature fed to the material model will be the adiabatic temperature, right? So `T = T_adiabatic(depth)` and `T-T_ref=0`.

naliboff:
> But in case of computing the adiabatic density, the temperature fed to the material model will be the adiabatic temperature, right? So T = T_adiabatic(depth) and T-T_ref=0.

Yes, this is indeed what is happening. The adiabatic temperature is added is to the temperature in `adiabatic_conditions/compute_profile.cc `, and this new temperature is passed to the material model. 

However, if one had an initial temperature field that includes adiabatic heating, would this not double count the adiabatic terms in the initial setup of the reference profile?

gassmoeller:
> But in case of computing the adiabatic density, the temperature fed to the material model will be the adiabatic temperature, right? So T = T_adiabatic(depth) and T-T_ref=0.

Yes, this is exactly what happens during the computation of the adiabatic profile. 

> However, if one had an initial temperature field that includes adiabatic heating, would this not double count the adiabatic terms in the initial setup of the reference profile?

The initial temperature does not matter for the adiabatic profile. The adiabatic profile starts at the adiabatic surface temperature and computes an adiabatic profile using the material model. I.e. it hands over the surface adiabatic p,T, computes material properties, computes the next adiabatic T,p and repeats.

> For clarity, should we just use the adiabatic initial temperature conditions?

Yes, whenever possible the `adiabatic` initial temperature plugin is the best way to have an initial condition consistent with the adiabatic reference profile. It also has an option to include a half-space cooling anomaly at the top. John: There are a few different ways to do the transition. Either like the `adiabatic` plugin as an anomaly compared to the adiabatic profile, or like `adiabatic boundary` plugin which separates the domain into a lithosphere and mantle part and uses different models in them. Your choice. The tricky part is to have a somewhat realistic and continuous transition.

gassmoeller:
Yes, this looks good. If you update the test results this seems ready to go?

naliboff:
@gassmoeller - I just updated the tests (squashed into the last commit). Thanks for the explanation above and apologies for not realizing how the adiabatic profile was being set up initially, which would have shortened the discussion quite a bit.

@anne-glerum - I have some ideas for combining an adiabatic profile with a conductive geotherm in the lithosphere. I'll follow up on this with a separate PR, perhaps with an update the continental extension cookbook (would be good to show a cookbook using EBA).

gassmoeller (approved):
No problem, thanks for the fix :+1:

doc/modules/changes/20201128_naliboff
```diff
@@ -0,0 +1,5 @@
+<li> Fixed: The Multicomponent Incompressible equation
```
gassmoeller, line 1: Oh, just because we just merged #3917: Please remove the `<li>` and then merge yourself.

naliboff: done, thanks again to you and Anne for the review and discussion!

PR #3878, opened 2020-10-30
Principal stress visulization output

gassmoeller:
As described in the changelog entry. This adds a useful new visualization postprocessor to compute the principal stress axis and values (useful to determine failure modes or maximum differential stress).

Below are two example outputs for simple shear (top) and pure shear (bottom). Largest principal stress directions are lines in black, smallest principal stress directions lines in white, and differential stress is background color. Results are as expected for these benchmarks, i.e. maximum compressive stress 45 degree rotated to simple shear direction in top figure, and aligned with the pure shear direction of compression / y-axis in bottom figure, respectively.

simple_shear
pure_shear

PR #3897, opened 2020-11-19
Fix stress

anne-glerum:
As first mentioned on the forum by Sungho, the stress should be computed as `2*eta*strain_rate-P*I` or `-2*eta*strain_rate+P*I`, instead of `2*eta*strain_rate+P*I` as is currently done in the visualization postprocessors that output stresses. Here I fixed the sign, sticking to the convention of positive compressive stresses, so `-2*eta*strain_rate+P*I`.

See also discussion in #3878 .

### For all pull requests:


naliboff (commented):
Looks good, thanks @anne-glerum! The only key remaining item would be to update the test output.

A general questions: Is there somewhere in the manual that discusses visualizing stress that should be updated?

Once merged, the forum post should also be updated.

source/postprocess/visualization/shear_stress.cc
```diff
@@ -73,7 +73,8 @@ namespace aspect
 
             const double eta = out.viscosities[q];
 
-            const SymmetricTensor<2,dim> shear_stress = 2*eta*compressible_strain_rate;
+            // Compressive stress is positive in geoscience applications
+            const SymmetricTensor<2,dim> shear_stress = -2.*eta*compressible_strain_rate;
```
bobmyhill, line 77: The "compressible_strain_rate" variable is actually the incompressible part of the strain rate. Ditto for the other two files. Suggest changing the name to incompressible_strain_rate?

This is my only quibble :)

anne-glerum: Thanks for having a look. It depends on what the material model returns for `(this->get_material_model().is_compressible()` whether the strain rate in `compressible_strain_rate` is in fact compressible or the incompressible part. So the name is not always correct. But that's not really related to this PR, so I'll leave that for now, OK?

anne-glerum:
* Alright, I've checked that the failing tests failed because of the stress postprocessors and updated the results.
* I couldn't find an instance in the manual about stress visualization that needs to be updated. 
* I'll update the forum post when this is merged.

source/postprocess/visualization/shear_stress.cc
naliboff: Hmm, the relevant code above is:

>   = (this->get_material_model().is_compressible()
>                  ?
>                  strain_rate - 1./3 * trace(strain_rate) * unit_symmetric_tensor<dim>()
>                  :
>                  strain_rate);

Even if the material is compressible, the pressure is removed (-1/3 * trace(strain_rate)) and thus you have the deviatoric strain rate. If an incompressible formulation is used, then should the strain rate by default be deviatoric?

If the above is correct, would `deviatoric_strain_rate` be the most accurate name?

No preference on whether this is resolved now or in a follow-up PR, but will leave that up to @bobmyhill.

bobmyhill: @naliboff Yes, that's what I was trying to say. Thanks for clarifying!
I'm fine for the rename to be a separate PR.

naliboff (approved):
@anne-glerum - This is good to go, thanks! I'll make a follow-up PR that renames `compressible_strain_rate` to 'deviatoric_strain_rate', per @bobmyhill suggestion.

PR #3901, opened 2020-11-22
Prevent a warning in release mode of an unused variable in principle stress visualization

MFraters:
Prevent a warning in release mode of an unused variable in principle stress visualization by removing the variable and using the computation directly in the assert.


naliboff (approved):
Nice catch, thanks for the fix!

PR #3902, opened 2020-11-22
Rename strain rate variable

naliboff:
Following the discussion in #3897, the deviatoric strain rate is represented with a variable named `compressible_strain_rate` in multiple locations. This PR changes those variable names to `deviatoric_strain_rate`. @bobmyhill 

### DIFF
--- a/source/postprocess/visualization/strain_rate.cc
+++ b/source/postprocess/visualization/strain_rate.cc
@@ -58,14 +58,14 @@ namespace aspect
               grad_u[d] = input_data.solution_gradients[q][d];
 
             const SymmetricTensor<2,dim> strain_rate = symmetrize (grad_u);
-            const SymmetricTensor<2,dim> compressible_strain_rate
+            const SymmetricTensor<2,dim> deviatoric_strain_rate
               = (this->get_material_model().is_compressible()
                  ?
                  strain_rate - 1./3 * trace(strain_rate) * unit_symmetric_tensor<dim>()
                  :
                  strain_rate);
-            computed_quantities[q](0) = std::sqrt(compressible_strain_rate *
-                                                  compressible_strain_rate);
+            computed_quantities[q](0) = std::sqrt(deviatoric_strain_rate *
+                                                  deviatoric_strain_rate);
           }
 
         // average the values if requested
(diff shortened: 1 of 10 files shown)


source/postprocess/visualization/strain_rate.cc
```diff
@@ -58,14 +58,14 @@ namespace aspect
...
-            const SymmetricTensor<2,dim> compressible_strain_rate
+            const SymmetricTensor<2,dim> deviatoric_strain_rate
               = (this->get_material_model().is_compressible()
                  ?
                  strain_rate - 1./3 * trace(strain_rate) * unit_symmetric_tensor<dim>()
                  :
                  strain_rate);
-            computed_quantities[q](0) = std::sqrt(compressible_strain_rate *
-                                                  compressible_strain_rate);
+            computed_quantities[q](0) = std::sqrt(deviatoric_strain_rate *
```
bobmyhill, line 67: Is there a reason why this isn't just std::abs?

bobmyhill (commented):
Thanks @naliboff! Two more files with this variable name:
- cookbooks/anisotropic_viscosity/av_material.cc
- tests/anisotropic_viscosity.cc

Otherwise, there's just that one suggested change from sqrt(x^2) to std::abs(x) ... but that could be another PR.

naliboff:
@bobmyhill - Thanks for taking a look so quickly and finding those additional files, which are updated in the last commit.

Regarding the `std:abs(value)` versus `std::sqrt(value*value)`. At first glance I agree these should give the same result, but I think with these operations the tensors are being reduced to scalars. In that case, maybe they aren't equivalent? I'll need to write this out if the tensors are being reduced to scalars.

anne-glerum:
@naliboff these changes look good, thanks. The `std::sqrt(value*value)` takes the square root of the sum of the squares, so indeed we go from tensor to scalar (double contraction). It's like a Frobenius norm.

anne-glerum (approved):
Great, thanks!

bobmyhill:
> @naliboff these changes look good, thanks. The `std::sqrt(value*value)` takes the square root of the sum of the squares, so indeed we go from tensor to scalar (double contraction). It's like a Frobenius norm.

Thanks @anne-glerum! All looks good :)

bobmyhill (approved):

Issue #3885, opened 2020-11-13
Tracking issue for adding CPO/LPO tracking and development to aspect


MFraters:
Lattice/Crystal Preferred Orientation (LPO/CPO, they are different names for the same) is a useful tool to be able to compare against seismic anisotropy data, and can give us a more direct link to mantle flow. It could also be used in the future to make anisotropic viscosities.  

It has been a bit of a process of getting this into aspect, but I think it is final near the level to start making pull requests (note that is may still take a bit for the first pull request is made). To make the review process a bit more easy, I will split up the pull requests in several parts. This issue is to keep track of them and discuss if a different cut up of the pull request is needed. Each part will contain their own tests, although there will be some overlap in some tests besides the dependencies given in each point, which will be address along the way, tracked by point 6.

1. The core of the LPO/CPO property plugin. It will contain the particle properties, the advection algorithms and a derivative computation which doesn't change the grain size, and rotates the CPO direction with the rotation of the particle. Besides unit tests, this will contain a test where a particle is rotated in a circle, and the direction should always stay the same relative to the rotation of the particle.
2. Add the D-Rex computation of the derivatives for the grain-sizes and rotation of the direction matrices. (dependent on 1)
3. Add a particle property property plugin which can compute the Bingham average of the directions of the direction matrices (dependent on 1)
4. Add a particle property plugin which can compute the elastic tensor from the CPO. (dependent on 1)
5. Add a particle property plugin which can decompose the elastic tensor into the anistropic parts (Isotropic, Hexagonal, Tetragonal, Orthorhombic, Monoclinic and Triclinic). (dependent on 4)
6. Add the overlapping tests.
7. Add documentation

bobmyhill:
Wow, this is ambitious! Looking forward to seeing this move forwards!

MFraters:
It took ~3.5 years, but the CPO plugins are now part of ASPECT and tested :tada:  

I still need to add documentation on how to use it and add some benchmarks, but @magmaxt was planning to help me with that so that hopefully won't take another 3.5 years ;)

magmaxt:
> It took ~3.5 years, but the CPO plugins are now part of ASPECT and tested 🎉
> 
> I still need to add documentation on how to use it and add some benchmarks, but @magmaxt was planning to help me with that so that hopefully won't take another 3.5 years ;)

I am very grateful about the mentorship from @MFraters @gassmoeller @bangerth @tjhei on how to get started at the ASPECT hackathon !!
[...]

PR #3903, opened 2020-11-23
Add particle property plugin CPO core 

MFraters:
This is part one of adding CPO to aspect (see issue #3885). I am not entirely sure what the best approach is to get this all merged, so I tried to cut out as much as possible and focus this pull request mainly on the structure and testing the storing and loading of the CPO data in to the particle data vector. 

I have kept the advection algorithm, but only the most simple derivatives algorithm. Since the D-Rex algorithm implementation is a significant amount of complicated code, I left it out of this pull request, but I did keep some of the reference to it in the code. Although the world builder initial conditions coupling is not that much code, I removed it so that I don't have to add the tests in this pull request. 

Although I could already make a test for the advection schemes in this pull request, it will be a lot easier to do that once the bingham average particle property plugin has been added. So I would like to wait for this part to be merged (all the other parts are dependent on this part).

If there are strong opinions on what should be included or excluded in this pull request, let me know.


source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+                                      + std::to_string(a_cosine_matrices_grains[mineral_i][grain_i][1][1]) + " "
+                                      + std::to_string(a_cosine_matrices_grains[mineral_i][grain_i][1][2]) + "\n"
+                                      + std::to_string(a_cosine_matrices_grains[mineral_i][grain_i][2][0]) + " "
+                                      + std::to_string(a_cosine_matrices_grains[mineral_i][grain_i][2][1]) + " "
+                                      + std::to_string(a_cosine_matrices_grains[mineral_i][grain_i][2][2])));
+              }
+
+            /**
+            * Now we have loaded all the data and can do the actual computation.
+            * The computation consitsts of two parts. The first part is computing
```
bobmyhill, line 513:
```suggestion
            * The computation consists of two parts. The first part is computing
```

source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+                        Assert(!std::isnan(a_cosine_matrices_grains[mineral_i][i][j][k]), ExcMessage(" a_cosine_matrices_grains is nan before orthoganalization."));
+                      }
+
+                  }
+
+              }
+
+            /**
+             * Correct direction cosine matrices numerical error (orthnormality) after integration
+             * Follows same method as in matlab version from Thissen of finding the nearest orthonormal
```
bobmyhill, line 590: more detailed reference?

bobmyhill (changes requested):
Nice! 
Here's some suggested changes to get the ball rolling. I find it hard to read through an 1100 line .cc in one sitting, so there's still some more checking to do.

bobmyhill:
You might find aspell useful for spell checking other big chunks of code - this is what I use once a year to catch all the new typos, and it's much less painful than reading line by line.

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,362 @@
...
+        SpinTensor, DRex2004
+      };
+
+      /**
+       * The plugin manages and computes the evolution of Lattice/Crystal Preffered Orientations (LPO/CPO)
+       * on particles. Each ASPECT particle can contain many grains. Each grains contain a size and a orientation
+       * matrix. This allows for LPO evolution tracking with Polycrysal kinematic CPO evolution models such
+       * as D-Rex (Kaminski and Ribe, 2001; ́Kaminski et al., 2004).
+       *
+       * The layout of the data vector per perticle is the following (note that for this plugin the following dim's are always 3):
```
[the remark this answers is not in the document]
bobmyhill: This needs more clarification. If I understand the code correctly, it isn't that the plugin only works for dim=3, but rather that if dim=2, the plugin assumes an infinitely thin 3D domain.

unit_tests/cpo.cc
```diff
@@ -0,0 +1,841 @@
...
+          prm.set("Random number seed","1");
+          prm.set("Number of grains per praticle","5");
+          prm.set("CPO derivatives algorithm","Spin tensor");
+          prm.set("Property advection method","Forward Euler");
+          prm.enter_subsection("Initial grains");
+          {
+            prm.set("Model name","Uniform grains and Random Uniform rotations");
+            prm.enter_subsection("Uniform grains and Random Uniform rotations");
+            {
+              prm.set("Minerals","Passive,Passive");
```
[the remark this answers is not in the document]
MFraters: With that I mean that the rotation matrices just rotate with the rotation of the particle, so passive in that sense. I am open to suggestions :)

unit_tests/cpo.cc
```diff
@@ -0,0 +1,841 @@
...
+  a_cosine_matrices_grains_derivatives_ref[0][2][0][2] = 74;
+  a_cosine_matrices_grains_derivatives_ref[0][2][1][0] = 75;
+  a_cosine_matrices_grains_derivatives_ref[0][2][1][1] = 76;
+  a_cosine_matrices_grains_derivatives_ref[0][2][1][2] = 77;
+  a_cosine_matrices_grains_derivatives_ref[0][2][2][0] = 78;
+  a_cosine_matrices_grains_derivatives_ref[0][2][2][1] = 79;
+  a_cosine_matrices_grains_derivatives_ref[0][2][2][2] = 80;
+
+  volume_fractions_grains_derivatives_ref[1][0] = 1.0;
+  volume_fractions_grains_derivatives_ref[1][1] = 1.1;
```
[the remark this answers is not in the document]
MFraters: Yes, it looks (and is) odd but this is just a synthetic test to see if the values are stored and loaded correctly. The first assert you suggest would be good, but I think it should be placed where the computation is actually done and not in these store and load functions. 
I have thought about the second assert (sum volume_fractions ~1) and currently I think it has value to allow a sum larger than 1. The reason for that is that the D-Rex handles all the mineral types completely separately and the only connection is the `volume fraction olivine` and the `volume fraction enstatite = 1-volume fraction olivine`. Not checking for the sum therefore allows to in one run compute for example a set of olivine grains deforming while being 100%,70% and 50% of the total grains. To be honest, it is not something I am planning on using myself, and I am fine with limiting it to the sum if there are strong opinions about it.

unit_tests/cpo.cc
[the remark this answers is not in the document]
bobmyhill: The name is fine! But it would be great if you could add a comment describing what Passive means in physical terms.

source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+#include <aspect/utilities.h>
+
+namespace aspect
+{
+  namespace Particle
+  {
+    namespace Property
+    {
+
+      template <int dim>
```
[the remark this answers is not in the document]
MFraters: The are static variables, so they need to be declared in the source file.

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,362 @@
...
+       *         2.2. a_cosine_matrix grain -> 9 (Tensor<2,dim>::n_independent_components) doubles, starts at:
+       *                                      => cpo_data_position + 3 + grain_i * 10 + mineral_i * (n_grains * 10 + 2)
+       *
+       * Last used data entry is cpo_data_position  + n_minerals * (n_grains * 10 + 2). In case the Crank-Nickelson
+       * advection scheme is used, the previous deriative data is stored behind this. It stores per grain one
+       * double for the volume fraction derivative and 9 doubles for the rotation matrix derivative. This results
+       * the last data entry in this case being:
+       * cpo_data_position + n_minerals * (n_grains * 10 + 2) + mineral_n * (n_grains * 10).
+       *
+       * An other note is that we store exactly the same amount of all minerals (e.g. olivine and enstatite
```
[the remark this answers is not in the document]
MFraters: It is just a rotation matrix (it is a direction cosine matrix, and the letter used was A in the equation I think, so it became a_cosine_matrix). I will change it to rotation matrix, since I think that is the most clear. .

source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+        //    2.2. N grains times:
+        //         2.1. volume fraction grain -> 1 double, at location:
+        //                                      => data_position + 2 + i_grain * 10 + mineral_i *(n_grains * 10 + 2), or
+        //                                      => data_position + 2 + i_grain * (2 * Tensor<2,3>::n_independent_components+ 2) + mineral_i * (n_grains * 10 + 2)
+        //         2.2. a_cosine_matrix grain -> 9 (Tensor<2,dim>::n_independent_components) doubles, starts at:
+        //                                      => data_position + 3 + i_grain * 10 + mineral_i * (n_grains * 10 + 2), or
+        //                                      => data_position + 3 + i_grain * (2 * Tensor<2,3>::n_independent_components+ 2) + mineral_i * (n_grains * 10 + 2)
+        //
+        // Note is that we store exactly the same amount of all minerals (e.g. olivine and enstatite
+        // grains), although the volume may not be the same. This has to do with that we need a minimum amount
```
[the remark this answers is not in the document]
MFraters: does that last amount also need to be number?

source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+        // of grains per tracer to perform reliable statistics on it. This miminum is the same for both olivine
```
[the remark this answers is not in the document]
MFraters: minerals?

source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+              prm.declare_entry ("Random number seed", "1",
+                                 Patterns::Integer (0),
+                                 "The seed used to generate random numbers. This will make sure that "
+                                 "results are reproducable as long as the problem is run with the "
+                                 "same amount of MPI processes. It is implemented as final seed = "
+                                 "user seed + MPI Rank. ");
+
+              prm.declare_entry ("Number of grains per praticle", "50",
+                                 Patterns::Integer (0),
+                                 "The number of grains of olivine and the number of grain of enstatite "
```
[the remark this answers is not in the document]
MFraters: grains for each different?

source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+                                     "This determines what minerals and fabrics or fabric selectors are used used for the LPO calculation. "
+                                     "The options are Olivine: Passive, A-fabric, Olivine: B-fabric, Olivine: C-fabric, Olivine: D-fabric, "
+                                     "Olivine: E-fabric, Olivine: Karato 2008 or Enstatite. Passive sets all RRSS entries to the maximum. The "
+                                     "Karato 2008 selector selects a fabric based on stress and water content as defined in "
+                                     "figure 4 of the Karato 2008 review paper (doi: 10.1146/annurev.earth.36.031207.124120).");
+
+
+                  prm.declare_entry ("Volume fractions minerals", "0.5, 0.5",
+                                     Patterns::List(Patterns::Double(0)),
+                                     "The volume fraction for the different minerals. "
```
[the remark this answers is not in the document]
MFraters: oke, I will add that.

source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+                  for (size_t mineral_i = 0; mineral_i < n_minerals; mineral_i++)
+                    {
+                      if (temp_deformation_type_selector[mineral_i] == "Passive")
+                        {
+                          deformation_type_selector[mineral_i] = DeformationTypeSelector::Passive;
+                        }
+                      else
+                        {
+                          AssertThrow(false,
+                                      ExcMessage("The  fabric needs to be one of the following: Olivine: Karato 2008, "
```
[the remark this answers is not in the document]
MFraters: This is the style in which plugins are currently listed, so that I the reason I used this style. I agree that since these use more special characters, the comma-delimited comment add clarity. I am not sure how much extra quotes would help in readability.

unit_tests/cpo.cc
```diff
@@ -0,0 +1,841 @@
...
+        for (size_t iii = 0; iii < 3; iii++)
+          {
+            for (size_t jjj = 0; jjj < 3; jjj++)
+              {
+                CHECK(full_solution[iii][jjj] == Approx(deriv_direction_full_solution[index][iii][jjj]));
+              }
+          }
+      }
+  }
+}
```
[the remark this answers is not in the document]
MFraters: if you insist :P

unit_tests/cpo.cc
```diff
@@ -0,0 +1,841 @@
...
+                CHECK(full_solution[iii][jjj] == Approx(deriv_direction_full_solution[index][iii][jjj]));
+              }
+          }
+      }
+  }
+
+  {
+    // secondly test initialization 3d.
+    // This should be exactly the same as the 2d version
+    const int dim3=3;
```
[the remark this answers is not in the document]
MFraters: This was easier during development when I was trying to figure out what to exactly test. I think you are right that I could just make a loop, but I will need to test that.

MFraters (commented):
Thanks for the thorough review! I had not heard of aspell before. Will check it out :)

include/aspect/particle/property/cpo.h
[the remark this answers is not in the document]
bobmyhill: Thanks :). Again, I don't mind the name particularly, but some in line comment would be nice to show how the matrix relates to rotation of the crystal lattice.

unit_tests/cpo.cc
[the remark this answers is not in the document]
bobmyhill: Eh, I only suggested it because github puts a red cross at the end of files that don't end with a newline :D

unit_tests/cpo.cc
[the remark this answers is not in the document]
bobmyhill: Do I understand you correctly that the D-Rex-calculated CPO of X is dependent on the fraction of X but not on any of the coexisting phases. Isn't that inconsistent behaviour? What would happen if you had two types of olivine that differed only in colour (for example)? Splitting them 0:100 and 50:50 should have no effect on the resulting change in CPO. 

I'm probably misunderstanding something...

source/particle/property/cpo.cc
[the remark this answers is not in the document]
bobmyhill: Ok, that's fine by me.

source/particle/property/cpo.cc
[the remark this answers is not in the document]
bobmyhill: Either "of" or "for" would be ok, depending on whether in this sentence we are treating a grain as the physical object (grains of) or as the computational object (grains for, or grains assigned to). I think. Maybe.

Sometimes I wish I had been born in a country with a more sensible language.

source/particle/property/cpo.cc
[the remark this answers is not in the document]
bobmyhill: I wonder whether they need to be static, but I suppose it makes sense that you would want a single class-wide value.

source/particle/property/cpo.cc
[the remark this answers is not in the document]
bobmyhill: Ah, good catch, yes.

source/particle/property/cpo.cc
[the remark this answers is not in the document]
bobmyhill: Sure, minerals is also good.

source/particle/property/cpo.cc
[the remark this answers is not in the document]
MFraters: part of the reason why they are static is because it is information I would like to directly use in particle property plugins and static particle property functions (such as the store and load functions, which are use by these other plugins) which directly builds on this information.

source/particle/property/cpo.cc
```diff
@@ -0,0 +1,1124 @@
...
+              }
+          }
+      }
+
+      template <int dim>
+      void
+      CPO<dim>::initialize_one_particle_property(const Point<dim> &,
+                                                 std::vector<double> &data) const
+      {
+        // the layout of the data vector per perticle is the following (note that for this plugin the following dim's are always 3):
```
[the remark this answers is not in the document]
MFraters: the 2d case just fills in zero's for the 3rth dimension of the strain-rate and velocity gradient tensors. To be honest, I havn't spend that much thought on 2D, and visualization of the results is also not trivial, because currently the aspect paraview output doesn't understand a 3D vector in a 2D model. I would be fine with just asserting the 2D case, since I see little value in 2D cpo, but I don't know how others think about that.

source/particle/property/cpo.cc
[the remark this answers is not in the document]
MFraters: haha, I will trust your judgment in this :)

unit_tests/cpo.cc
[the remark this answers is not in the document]
MFraters: In each computation step it only influences the derivative of the volume fraction. So final derivative grain x = volume fraction mineral 0 * derivative grain x. So it does not matter what the other mineral grains are doing, but this indirectly influences the cpo development. What the current setup allows is to use that property of the algorithm to simultaneously compute for example how the olivine cpo develops if you have 60%, 70% or 100% olivine in the mix (and you could also compute the corresponding 40%, 70% and 0% enstatite if you want. This does currently get tricky if you are interested in the elastic tensor resulting from this matrix (which is not in this pull request), because that currently assumes that every counts up to 1. 

Like I said before, I am not planning on using that, but I could see that it might be useful to others and I didn't see any harm, so that is why I didn't disallow it. This of course stops working if cpo algorithms where to be implemented in which the minerals directly interact which each other.

MFraters:
Thanks again for the review. I addressed all your comments. Based on your review, there are just two outstanding issues in my mind:
1. Should CPO in 2D be allowed?
2. And do we want to limit the sum of the volume fractions to 1.
Currently I am thinking no on both, but I don't feel to strongly about it.

bobmyhill:
Hi @MFraters, a 2D implementation would only make sense if 2D deformation in a(n incompressible) 3D geometry model results in 2D CPO development (i.e. rotations are in-plane). Is this the case for the models you're implementing? Your simulations are likely to be CPU- and memory-intensive, so I could potentially see the benefit of a 2D (incompressible) CPO model. But it also makes for more possible cases to test, so I'm fine with limiting the implementation to 3D.

From how you've described the model, each grain is independent. Therefore, it should be possible to postprocess the simulation to determine the CPO development based on different volume fractions (iff CPO developed passively in the model and if one wished to do that). I'd therefore favour asserting that the volume fractions sum to one, as this will be necessary for any models where CPO has an effect on the rheology.

MFraters:
@bobmyhill I have disabled both. It will be easier to loosen restrictions than to make it stricter. I have put the asserts in the parse parameters, which I think should be sufficient.

bobmyhill:
Great, thanks @MFraters! 
Could someone else take a look at this? This is a big chunk of code, and it would be good to get one of the senseis to approve :)

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+           * Loads particle data into variables. Intended for use by other parts of aspect.
+           */
+          static
+          void
+          load_particle_data(unsigned int cpo_index,
+                             const ArrayView<double> &data,
+                             std::vector<unsigned int> &deformation_type,
+                             std::vector<double> &volume_fraction_mineral,
+                             std::vector<std::vector<double>> &volume_fractions_mineral,
+                             std::vector<std::vector<Tensor<2,3> > > &a_cosine_matrices_mineral);
```
gassmoeller, line 190: What does the `a` in `a_cosine_matrices_mineral` stand for? Do you mean `arccosine_matrices_mineral`? if so, maybe change the name, I was looking for other meanings.

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+       * The layout of the data vector per particle is the following (note that for this plugin the following dims are always 3):
+       * 1. M mineral times
+       *    1.1  Mineral deformation type   -> 1 double, at location
+       *                                      => cpo_data_position + 0 + mineral_i * (n_grains * 10 + 2)
+       *    2.1. Mineral volume fraction    -> 1 double, at location
+       *                                      => cpo_data_position + 1 + mineral_i * (n_grains * 10 + 2)
+       *    2.2. N grains times:
+       *         2.1. volume fraction grain -> 1 double, at location:
+       *                                      => cpo_data_position + 2 + grain_i * 10 + mineral_i * (n_grains * 10 + 2)
+       *         2.2. a_cosine_matrix grain -> 9 (Tensor<2,dim>::n_independent_components) doubles, starts at:
```
gassmoeller, line 72: see my comment below about the name `a_cosine_matrix`

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+                             const ArrayView<double> &data,
+                             std::vector<unsigned int> &deformation_type,
+                             std::vector<double> &volume_fraction_mineral,
+                             std::vector<std::vector<double>> &volume_fractions_mineral,
+                             std::vector<std::vector<Tensor<2,3> > > &a_cosine_matrices_mineral);
+
+
+          /**
+           * Loads particle data into variables. An extention of load_particle_data which includes the grain
+           * derivatvies. Intended mostly for use in this plugin.
```
gassmoeller, line 195: If this is mostly for internal use, can it be private? If not, does it make sense to not name this `_extended` but instead give it the same name as the function above, simply making it an overloaded function with more parameters?

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+           *
+           * @return A vector that contains pairs of the property names and the
+           * number of components this property plugin defines.
+           */
+          virtual
+          std::vector<std::pair<std::string, unsigned int> >
+          get_property_information() const;
+
+          /**
+           * Loads particle data into variables. Intended for use by other parts of aspect.
```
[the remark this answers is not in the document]
gassmoeller: Also please add a sentence that mentions that the parameters are explained in the general class documentation.

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+                                      const ArrayView<double> &data,
+                                      std::vector<unsigned int> &deformation_type,
+                                      std::vector<double> &volume_fraction_mineral,
+                                      std::vector<std::vector<double>> &volume_fractions_mineral,
+                                      std::vector<std::vector<Tensor<2,3> > > &a_cosine_matrices_mineral,
+                                      std::vector<std::vector<double> > &volume_fractions_mineral_derivatives,
+                                      std::vector<std::vector<Tensor<2,3> > > &a_cosine_matrices_mineral_derivatives) const;
+
+          /**
+           * Stores information in variables into the data array. Intended for use by other parts of ASPECT.
```
gassmoeller, line 208: same comments as for `load_particle_data` above.

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+          store_particle_data(unsigned int cpo_data_position,
+                              const ArrayView<double> &data,
+                              std::vector<unsigned int> &deformation_type,
+                              std::vector<double> &volume_fraction_mineral,
+                              std::vector<std::vector<double>> &volume_fractions_mineral,
+                              std::vector<std::vector<Tensor<2,3> > > &a_cosine_matrices_mineral);
+
+
+          /**
+           * Stores information in variables into the data array. An extention of load_particle_data which includes the grain
```
gassmoeller, line 221:
```suggestion
           * Stores information in variables into the data array. An extension of load_particle_data which includes the grain
```

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+           * derivatvies. Intended mostly for use in this plugin.
```
gassmoeller, line 222:
```suggestion
           * derivatives. Intended mostly for use in this plugin.
```

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+                                       std::vector<std::vector<double> > &volume_fractions_mineral_derivatives,
+                                       std::vector<std::vector<Tensor<2,3> > > &a_cosine_matrices_mineral_derivatives) const;
+
+          /**
+           * Updates the volume fractions and rotation matrices with a Forward Euler scheme.
+           */
+          double
+          advect_forward_euler(std::vector<double> &volume_fractions,
+                               std::vector<Tensor<2,3> > &a_cosine_matrices,
+                               const std::pair<std::vector<double>, std::vector<Tensor<2,3> > > &derivatives,
```
gassmoeller, line 240: We usually place input parameters (`const` ones) before output parameters. Please reorder.

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+                               const double strain_rate_second_invariant,
+                               const double dt) const;
+          /**
+           * Updates the volume fractions and rotation matrices with a Backward Euler scheme.
+           */
+          double
+          advect_backward_euler(std::vector<double> &volume_fractions,
+                                std::vector<Tensor<2,3> > &a_cosine_matrices,
+                                const std::pair<std::vector<double>, std::vector<Tensor<2,3> > > &derivatives,
```
gassmoeller, line 249: same here

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+                                       const ArrayView<double> &data,
+                                       std::vector<unsigned int> &deformation_type,
+                                       std::vector<double> &volume_fraction_mineral,
+                                       std::vector<std::vector<double>> &volume_fractions_mineral,
+                                       std::vector<std::vector<Tensor<2,3> > > &a_cosine_matrices_mineral,
+                                       std::vector<std::vector<double> > &volume_fractions_mineral_derivatives,
+                                       std::vector<std::vector<Tensor<2,3> > > &a_cosine_matrices_mineral_derivatives) const;
+
+          /**
+           * Updates the volume fractions and rotation matrices with a Forward Euler scheme.
```
gassmoeller, line 235: Add documentation for the parameters, at least the ones not explained in the general class description. In particular also: What does this function return?

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+       *    2.1. Mineral volume fraction    -> 1 double, at location
+       *                                      => cpo_data_position + 1 + mineral_i * (n_grains * 10 + 2)
+       *    2.2. N grains times:
+       *         2.1. volume fraction grain -> 1 double, at location:
+       *                                      => cpo_data_position + 2 + grain_i * 10 + mineral_i * (n_grains * 10 + 2)
+       *         2.2. a_cosine_matrix grain -> 9 (Tensor<2,dim>::n_independent_components) doubles, starts at:
+       *                                      => cpo_data_position + 3 + grain_i * 10 + mineral_i * (n_grains * 10 + 2)
+       *
+       * Last used data entry is cpo_data_position  + n_minerals * (n_grains * 10 + 2). In case the Crank-Nicolson
+       * advection scheme is used, the previous deriative data is stored behind this. It stores per grain one
```
gassmoeller, line 76: Please explain in a sentence what derivatives you are talking about. The spatial derivatives of the properties? The time derivatives?

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+                                const double strain_rate_second_invariant,
+                                const double dt) const;
+
+          /**
+           * Updates the volume fractions and rotation matrices with a Crank-Nicolson scheme.
+           */
+          double
+          advect_Crank_Nicolson(std::vector<double> &volume_fractions,
+                                std::vector<Tensor<2,3> > &a_cosine_matrices,
+                                const std::pair<std::vector<double>, std::vector<Tensor<2,3> > > &derivatives,
```
gassmoeller, line 259: and here

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+                                std::vector<double> &previous_volume_fraction_derivatives,
+                                std::vector<Tensor<2,3> > &previous_a_cosine_matrices_derivatives,
+                                const double strain_rate_second_invariant,
+                                const double dt) const;
+
+
+          /**
+           * A interface function which delegates the computation the derivatives
```
gassmoeller, line 267: You explain *how* this function does something, but it is more important to explain *what* this function does. Which derivatives? The fact that it calls other functions does not really matter.

```suggestion
           * An interface function, which computes and returns the derivatives of ...
```

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+          std::pair<std::vector<double>, std::vector<Tensor<2,3> > >
+          compute_derivatives(const std::vector<double> &volume_fractions,
+                              const std::vector<Tensor<2,3> > &a_cosine_matrices,
+                              const SymmetricTensor<2,3> &strain_rate_nondimensional,
+                              const Tensor<2,3> &velocity_gradient_tensor_nondimensional,
+                              const double volume_fraction_mineral,
+                              const std::array<double,4> &ref_resolved_shear_stress) const;
+
+          /**
+           * Computes the update for the property rotating passively with the particle.
```
gassmoeller, line 279: Hmm, so it is the time derivative you are computing here? Please specify. And which property? According to the function name the `spin_tensor`? But which one is that, I do not think this was mentioned before. Also explain the parameters, at least the ones not explained in the general class documentation. And what does the function return (and why are those vectors)?

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+          /**
+           * Read the parameters this class declares from the parameter file.
+           */
+          void
+          parse_parameters (ParameterHandler &prm) override;
+
+          /**
+           * Return the number of grains per particle
+           */
+          static
```
gassmoeller, line 306: We try to avoid static functions whenever possible. Is this necessary? It may create problems once you want to change the number of grains or minerals based on input files.

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+          /**
+           * Return the number of minerals per particle
+           */
+          static
+          unsigned int
+          get_number_of_minerals();
+
+        private:
+          // helpful shortcuts
+          double rad_to_degree = 180.0/M_PI;
```
gassmoeller, line 319: I have not looked into the code, but I can not imagine having these shortcuts as member variables is significantly better than just defining them as `const double` in the function where you need them. Also we do not usually use `M_PI`, we use `numbers::PI` instead

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+          double degree_to_rad = M_PI/180.0;
+
+          /**
+           * Random number generator used for initalization of particles
+           */
+          mutable boost::lagged_fibonacci44497 random_number_generator;
```
gassmoeller, line 325: Is it important to use this generator?  We usually either use the following generator or call a distribution directly:
```
    boost::mt19937                     rng(seed);
    boost::random::uniform_real_distribution<> uniform_distribution(-1, 1);
```

include/aspect/particle/property/cpo.h
```diff
@@ -0,0 +1,361 @@
...
+           * Advection method for particle properties
+           */
+          AdvectionMethod advection_method;
+
+          /**
+           * What algorithm to use to compute the derivatives
+           */
+          CPODerivativeAlgorithm cpo_derivative_algorithm;
+
+          double property_advection_tolerance;
```
gassmoeller, line 348: Add documentation for these two parameters.

gassmoeller (commented):
I will need to come back for the implementation, but here are some comments on the header for now. This will be really cool, but I think it needs some work first.

include/aspect/particle/property/cpo.h
MFraters: I can make it private, but then I can't test it with the unit test. So I went with the overloading option.

include/aspect/particle/property/cpo.h
MFraters: long story, but I changed it to `rotation_matrices_mineral`. 

(It is a general rotation matrix usually constructed by a Direction Cosine Matrix (DCM) with Euler angles which had the letter a in the equation.)

include/aspect/particle/property/cpo.h
MFraters: > We try to avoid static functions whenever possible. Is this necessary?

I agree that it is not ideal, but part of the reason why they are static is because it is information I would like to directly use in particle property plugins and static particle property functions (such as the store and load functions, which are use by these other plugins) which directly builds on this information.

> It may create problems once you want to change the number of grains or minerals based on input files.

With the current code you can already set the number of minerals and grains per mineral in the input file, so I am not sure what you mean.

include/aspect/particle/property/cpo.h
MFraters: Not a strong reason as far as I can remember (I probably just looked up the options and thought bigger cycle is better). In terms of speed it probably won't matter too much since since it is only used for initialization of the particle orientations.

Do you feel strongly about it? Because I am a bit hesitant to change it because I will probably have to update all the values in the unit test :P

include/aspect/particle/property/cpo.h
MFraters: I just removed them for now.

MFraters:
@gassmoeller Thanks for the review! I have addressed your comments and left a few discussions unresolved. I also found that one of the variables in the advection scheme is not needed, because it can be dealt with directly in the computation of the derivatives.

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+            {
+              prm.declare_entry ("Random number seed", "1",
+                                 Patterns::Integer (0),
+                                 "The seed used to generate random numbers. This will make sure that "
+                                 "results are reproducible as long as the problem is run with the "
+                                 "same number of MPI processes. It is implemented as final seed = "
+                                 "user seed + MPI Rank. ");
+
+              prm.declare_entry ("Number of grains per particle", "50",
+                                 Patterns::Integer (0),
```
gassmoeller, line 979: You probably want at least 1?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                                 "This option allows for setting a tolerance. When the norm of tensor new - tensor old is "
+                                 "smaller than this tolerance, the iteration is stopped.");
+
+              prm.declare_entry ("Property advection max iterations", "100",
+                                 Patterns::Integer(0),
+                                 "The Backward Euler and Crank-Nicolson property advection methods involve internal iterations. "
+                                 "This option allows for setting the maximum number of iterations. Note that when the iteration "
+                                 "is ended by the max iteration amount an assert is thrown.");
+
+              prm.declare_entry ("CPO derivatives algorithm", "D-Rex 2004",
```
gassmoeller, line 999: Isnt spin tensor the only one currently working? Make that the default, and remove the mention of D-Rex

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                                 Patterns::List(Patterns::Anything()),
+                                 "Options: Spin tensor, D-Rex 2004 (not implemented yet)");
+
+              prm.enter_subsection("Initial grains");
+              {
+                prm.declare_entry("Model name","Uniform grains and Random Uniform rotations",
```
gassmoeller, line 1005:
```suggestion
                prm.declare_entry("Model name","Uniform grains and random uniform rotations",
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                                  Patterns::Anything(),
+                                  "The model used to initialize the CPO for all particles. Currently 'Uniform grains and Random Uniform rotations' is the only valid option.");
```
gassmoeller, line 1007: Break this line

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                prm.enter_subsection("Uniform grains and Random Uniform rotations");
```
gassmoeller, line 1008:
```suggestion
                prm.enter_subsection("Uniform grains and random uniform rotations");
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+        }
+        prm.leave_subsection ();
+      }
+
+
+      template <int dim>
+      void
+      CrystalPreferredOrientation<dim>::parse_parameters (ParameterHandler &prm)
+      {
+        AssertThrow(dim != 2, ExcMessage("CPO computations are currently only supported for 3D models. "
```
gassmoeller, line 1042:
```suggestion
        AssertThrow(dim == 3, ExcMessage("CPO computations are currently only supported for 3D models. "
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                    {
+                      if (temp_deformation_type_selector[mineral_i] == "Passive")
+                        {
+                          deformation_type_selector[mineral_i] = DeformationTypeSelector::passive;
+                        }
+                      else
+                        {
+                          AssertThrow(false,
+                                      ExcMessage("The fabric needs to be assigned one of the following comma-delimited values: Olivine: Karato 2008, "
+                                                 "Olivine: A-fabric,Olivine: B-fabric,Olivine: C-fabric,Olivine: D-fabric,"
```
gassmoeller, line 1116:
```suggestion
                                                 "Olivine: A-fabric, Olivine: B-fabric, Olivine: C-fabric, Olivine: D-fabric,"
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                                                 "Olivine: E-fabric, Enstatite and Passive."))
```
gassmoeller, line 1117: Just to make it obvious
```suggestion
                                                 "Olivine: E-fabric, Enstatite, Passive."))
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                        }
+                    }
+
+                  volume_fractions_minerals = Utilities::string_to_double(dealii::Utilities::split_string_list(prm.get("Volume fractions minerals")));
+                  double volume_fractions_minerals_sum = 0;
+                  for (auto &&fraction : volume_fractions_minerals)
```
gassmoeller, line 1123: No need for the && when iterating over a vector of doubles. Actually slightly surprised this was not iterating over the memory addresses of volume_fractions_minerals instead of the values. Wouldnt it be of type `double *` in the current form?

```suggestion
                  for (auto fraction : volume_fractions_minerals)
```

gassmoeller (commented):
Ok, I took a first pass. I will take another look when you have adressed the changes. In general this looks good, but I am slightly concerned about two things:
- Introducing lots of new options can make this plugin overly large very quickly. At the moment my only concern is that there are three advection types, when there is probably one that is best. Let's have a chat about that at some point.
- You currently copy all properties from its storage location into new vectors in `pack/unpack_data` only because you can access them more conveniently that way. The same could be done by having utility functions, e.g. `get_mineral_volume_fraction(i_mineral)` or `get_grain_volume_fraction(i_mineral, i_grain)`, etc.. That way only these utility functions would need to know about the data layout. Do you think it would be better to have such accessor functions and not create new vectors for all your properties?

adigitoleo:
I hope this is an appropriate place for a quick comment: The rotation matrices used for the unit test seem rather arbitrary, I wonder if you have also tested a simpler configuration e.g.

```
     cos(-45)    0         cos(-45)
a =   0          1          0
     -sin(-45)   0         cos(-45)
```

i.e. a rotation of -45 deg about the global Y-axis. I only ask because I'm struggling to see how eq. 8 in the paper produces any nonzero slip invariant in some cases, for example with the above initial rotation and a simiple shear strain rate:

```
0  0  1
0  0  0
1  0  0
```

adigitoleo:
Note that in the linked derivation, I swapped the `n^\nu` and `l^\nu` vector definitions from table 2 in the paper, because that is how the original fortran seems to have been implemented, and it also seems to be what you have done in the supplementary derivation (part 4). However, I'm now thinking that you were correct and we should in fact be using the transpose of the orientation matrix to extract the slip invariants `I` (for orientation matrices defined using the usual alibi convention). Due to the symmetry of the orientation matrix, this still produces the same results for `I` and gamma in the case I considered above.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -0,0 +1,414 @@
...
+           */
+          void
+          parse_parameters (ParameterHandler &prm) override;
+
+          /**
+           * Return the number of grains per particle
+           */
+          static
+          unsigned int
+          get_number_of_grains();
```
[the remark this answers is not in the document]
MFraters: The problem is that if I name the funciton n_grains than it conflicts with the variable named n_grains.

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+      }
+
+      template <int dim>
+      std::pair<std::vector<double>, std::vector<Tensor<2,3> > >
+      CrystalPreferredOrientation<dim>::compute_derivatives(const std::vector<double> &,
+                                                            const std::vector<Tensor<2,3> > &,
+                                                            const SymmetricTensor<2,3> &,
+                                                            const Tensor<2,3> &velocity_gradient_tensor,
+                                                            const double,
+                                                            const std::array<double,4> &) const
```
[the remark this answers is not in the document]
MFraters: It would make life easier to keep them, but I will comment the variable names.

MFraters:
@gassmoeller Thanks for the review. I have addressed you comments. 

> You currently copy all properties from its storage location into new vectors in pack/unpack_data only because you can access them more conveniently that way. The same could be done by having utility functions, e.g. get_mineral_volume_fraction(i_mineral) or get_grain_volume_fraction(i_mineral, i_grain), etc.. That way only these utility functions would need to know about the data layout. Do you think it would be better to have such accessor functions and not create new vectors for all your properties?

I would agree that that is the better option, but it would be a significant amount of extra work for making, integrating and testing the new functions. Since I already have test that this works and I don't think the gain will big enough to delay this pull request any further. It can be refactored later. 

@adigitoleo Thank you for your comments. There are many ways to put together Euler angles, and Euler angles also have some issues with for example gimbal lock. That is the reaons the code is written that is internally uses rotation matrices. I agree that when we allow Euler angles for input and output, it needs to be documented well, but that will happen in a follow up pull request. Let me know if I misunderstood you comments.

MFraters:
@gassmoeller I rebased as discussed. Let me know if you have any other comments.

@bobmyhill since your review status is still on changes requested, can I request a re-review? Thanks!

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -0,0 +1,387 @@
+/*
+ Copyright (C) 2020 by the authors of the ASPECT code.
```
bobmyhill, line 2:
```suggestion
 Copyright (C) 2022 by the authors of the ASPECT code.
```

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -0,0 +1,387 @@
...
+       *
+       * Last used data entry is n_minerals * (n_grains * 10 + 2). In case the Crank-Nicolson
+       * advection scheme is used, the derivative data is stored after the main data and each grain
+       * stores twice the amount of data. The last data entry in this case is:
+       * n_minerals * (n_grains * 10 + 2) + n_minerals * (n_grains * 10).
+       *
+       * We store the same number of grains for all minerals (e.g. olivine and enstatite
+       * grains), although their volume fractions may not be the same. This is because we need a minimum number
+       * of grains per tracer to perform reliable statistics on it. This minimum should be the same for both
+       * olivine and enstatite.
```
bobmyhill, line 86: mineral specific, such it be?

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -0,0 +1,387 @@
...
+
+          /**
+           * This implementation tells the particle manager that
+           * we need to update particle properties every time step.
+           */
+          UpdateTimeFlags
+          need_update () const;
+
+          /**
+           * The CPO of late particles is initialized by interpolating from existing pariticles.
```
bobmyhill, line 160:
```suggestion
           * The CPO of late particles is initialized by interpolating from existing particles.
```

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -0,0 +1,387 @@
...
+                               std::vector<double> &volume_fraction_mineral,
+                               std::vector<std::vector<double>> &volume_fractions_mineral,
+                               std::vector<std::vector<Tensor<2,3>>> &rotation_matrices_mineral,
+                               std::vector<std::vector<double>> &volume_fractions_mineral_derivatives = {},
+                               std::vector<std::vector<Tensor<2,3>>> &rotation_matrices_mineral_derivatives = {}) const;
+
+          /**
+           * Packs information from variables into the global particle data array. Intended for use by other parts of ASPECT.
+           * The parameters are explained in the general class documentation.
+           * The parameters are explained in the general class documentation.
```
bobmyhill, line 202: Duplicate line.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -0,0 +1,387 @@
...
+           */
+          mutable boost::lagged_fibonacci44497 random_number_generator;
+          unsigned int random_number_seed;
+
+          unsigned int n_grains;
+
+          unsigned int n_minerals;
+
+          /**
+           * A vector containing the deformatation type selectors provided by the user.
```
bobmyhill, line 343:
```suggestion
           * A vector containing the deformation type selectors provided by the user.
```

unit_tests/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,641 @@
+/*
+  Copyright (C) 2020 by the authors of the ASPECT code.
```
bobmyhill, line 2:
```suggestion
  Copyright (C) 2022 by the authors of the ASPECT code.
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
+/*
+  Copyright (C) 2020 by the authors of the ASPECT code.
```
bobmyhill, line 2:
```suggestion
  Copyright (C) 2022 by the authors of the ASPECT code.
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
...
+        //         2.2. rotation matrix grain -> 9 (Tensor<2,dim>::n_independent_components) doubles, starts at:
+        //                                      => data_position + 3 + i_grain * 10 + mineral_i * (n_grains * 10 + 2), or
+        //                                      => data_position + 3 + i_grain * (2 * Tensor<2,3>::n_independent_components+ 2) + mineral_i * (n_grains * 10 + 2)
+        //
+        // Note that we store exactly the same number of grains of all minerals (e.g. olivine and enstatite
+        // grains), although their volume fractions may not be the same. We need a minimum amount
+        // of grains per tracer to perform reliable statistics on it. This minimum is the same for all phases.
+        // and enstatite.
+        //
+        // Futhermore, for this plugin the following dim's are always 3. When using 2D an infinitely thin 3D domain is assumed.
```
bobmyhill, line 333:
```suggestion
        // Furthermore, for this plugin the following dims are always 3. When using 2D an infinitely thin 3D domain is assumed.
```

bobmyhill (commented):
Several of @gassmoeller's suggestions still appeared during code review; could you check these and reply to or resolve them if you've made the necessary changes?

Otherwise I just spotted a few typos.

bobmyhill (approved):

include/aspect/particle/property/crystal_preferred_orientation.h
MFraters: I will make this into `This minimum should be the same for all minerals.`

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
...
+        const SymmetricTensor<2,dim> strain_rate = symmetrize (velocity_gradient);
+
+        const double dt = this->get_timestep();
+
+        // even in 2d we need 3d strain-rates and velocity gradient tensors. So we make them 3d by
+        // adding an extra dimension which is zero.
+        SymmetricTensor<2,3> strain_rate_3d;
+        strain_rate_3d[0][0] = strain_rate[0][0];
+        strain_rate_3d[0][1] = strain_rate[0][1];
+        //sym: strain_rate_3d[0][0] = strain_rate[1][0];
```
gassmoeller, line 441:
```suggestion
        //sym: strain_rate_3d[1][0] = strain_rate[1][0];
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
...
+        strain_rate_3d[1][1] = strain_rate[1][1];
+
+        if (dim == 3)
+          {
+            strain_rate_3d[0][2] = strain_rate[0][2];
+            strain_rate_3d[1][2] = strain_rate[1][2];
+            //sym: strain_rate_3d[0][0] = strain_rate[2][0];
```
gassmoeller, line 448:
```suggestion
            //sym: strain_rate_3d[2][0] = strain_rate[2][0];
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
...
+            //sym: strain_rate_3d[0][1] = strain_rate[2][1];
```
gassmoeller, line 449:
```suggestion
            //sym: strain_rate_3d[2][1] = strain_rate[2][1];
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
...
+                       ExcMessage("volume_fractions_grains[mineral_i]" + std::to_string(grain_i) + "] is not finite: "
+                                  + std::to_string(volume_fractions_grains[mineral_i][grain_i]) + ", inv_sum_volume_mineral = "
+                                  + std::to_string(inv_sum_volume_mineral) + "."));
+
+
+                for (size_t j = 0; j < 3; j++)
+                  {
+                    for (size_t k = 0; k < 3; k++)
+                      {
+                        Assert(!std::isnan(rotation_matrices_grains[mineral_i][grain_i][j][k]), ExcMessage(" rotation_matrices_grains is nan before orthoganalization."));
```
gassmoeller, line 556:
```suggestion
                        Assert(!std::isnan(rotation_matrices_grains[mineral_i][grain_i][j][k]), ExcMessage(" rotation_matrices_grains is nan before orthogonalization."));
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
...
+              {
+                rotation_matrices_grains[mineral_i][grain_i] = dealii::project_onto_orthogonal_tensors(rotation_matrices_grains[mineral_i][grain_i]);
+                for (size_t i = 0; i < 3; i++)
+                  for (size_t j = 0; j < 3; j++)
+                    {
+                      // I don't think this should happen with the projection, but D-Rex
+                      // does not do the orthogonal projection, but just clamps the values
+                      // to 1 and -1.
+                      Assert(std::fabs(rotation_matrices_grains[mineral_i][grain_i][i][j]) <= 1.0,
+                             ExcMessage("The rotation_matrices_grains[mineral_i] has a entry asolute larger than 1."));
```
gassmoeller, line 578:
```suggestion
                             ExcMessage("The rotation_matrices_grains[mineral_i] has a entry larger than 1."));
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
...
+            for (unsigned int grain_i = 0; grain_i < n_grains; ++grain_i)
+              {
+
+                Assert(isfinite(volume_fractions_grains[mineral_i][grain_i]),
+                       ExcMessage("volume_fractions_grains[mineral_i][" + std::to_string(grain_i) + "] is not finite: "
+                                  + std::to_string(volume_fractions_grains[mineral_i][grain_i]) + ", inv_sum_volume_grains[mineral_i] = "
+                                  + std::to_string(inv_sum_volume_mineral) + "."));
+
+                for (size_t j = 0; j < 3; j++)
+                  for (size_t k = 0; k < 3; k++)
```
gassmoeller, line 591: There are several of these loops in this PR, please change the increment operator to a pre-increment (`++j` instead of `j++`). It doesnt matter for these application, just for consistency, because we do it that way everywhere if possible.

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                                                                       + std::to_string(grain_i) + ", volume_fractions[grain_i] = " + std::to_string(volume_fractions[grain_i])
+                                                                       + ", derivatives.first[grain_i] = " + std::to_string(derivatives.first[grain_i])));
+
+            sum_volume_fractions += volume_fractions[grain_i];
+          }
+
+
+        for (unsigned int grain_i = 0; grain_i < n_grains; ++grain_i)
+          {
+            rotation_matrices[grain_i] = rotation_matrices[grain_i] + dt * rotation_matrices[grain_i] * derivatives.second[grain_i];
```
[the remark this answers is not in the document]
gassmoeller: This is not yet resolved?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                                      + std::to_string(rotation_matrices_grains[mineral_i][grain_i][0][0]) + " " + std::to_string(rotation_matrices_grains[mineral_i][grain_i][0][1]) + " " + std::to_string(rotation_matrices_grains[mineral_i][grain_i][0][2]) + "\n"
+                                      + std::to_string(rotation_matrices_grains[mineral_i][grain_i][1][0]) + " " + std::to_string(rotation_matrices_grains[mineral_i][grain_i][1][1]) + " " + std::to_string(rotation_matrices_grains[mineral_i][grain_i][1][2]) + "\n"
+                                      + std::to_string(rotation_matrices_grains[mineral_i][grain_i][2][0]) + " " + std::to_string(rotation_matrices_grains[mineral_i][grain_i][2][1]) + " " + std::to_string(rotation_matrices_grains[mineral_i][grain_i][2][2])));
+              }
+
+            for (unsigned int grain_i = 0; grain_i < n_grains; ++grain_i)
+              {
+                Assert(isfinite(volume_fractions_grains[mineral_i][grain_i]),
```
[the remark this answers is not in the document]
gassmoeller: Unresolved?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+                cosine_new = rotation_matrices[grain_i] + dt * cosine_new * derivatives.second[grain_i];
+
+                if ((cosine_new-cosine_old).norm() < property_advection_tolerance)
+                  {
+                    break;
+                  }
+                cosine_old = cosine_new;
+              }
+
+            rotation_matrices[grain_i] = cosine_new;
```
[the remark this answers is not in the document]
gassmoeller: Unresolved?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+      template <int dim>
+      double
+      CrystalPreferredOrientation<dim>::advect_Crank_Nicolson(const double dt,
+                                                              const std::pair<std::vector<double>, std::vector<Tensor<2,3> > > &derivatives,
+                                                              std::vector<double> &volume_fractions,
+                                                              std::vector<Tensor<2,3> > &rotation_matrices,
+                                                              std::vector<double> &previous_volume_fraction_derivatives,
+                                                              std::vector<Tensor<2,3> > &previous_rotation_matrices_derivatives) const
+      {
+        double sum_volume_fractions = 0;
```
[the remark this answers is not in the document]
gassmoeller: Comments?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1102 @@
...
+                                 "results are reproducible as long as the problem is run with the "
+                                 "same number of MPI processes. It is implemented as final seed = "
+                                 "user seed + MPI Rank. ");
+
+              prm.declare_entry ("Number of grains per particle", "50",
+                                 Patterns::Integer (1),
+                                 "The number of grains of each different mineral "
+                                 "each particle contains.");
+
+              prm.declare_entry ("Property advection method", "Forward Euler",
```
gassmoeller, line 924: Is ¨Forward Euler" the *best* for this property (whatever that means, but whatever we choose as default will be the method everyone uses). Which method do you usually use?

gassmoeller (commented):
Pretty close, but like @bobmyhill said there are still comments from my last review that are not fixed (see here for a full list of comments: https://github.com/geodynamics/aspect/pull/3903/files), and I found a few more. For some reason github shows some comments as resolved or outdated that are not fixed, you will have to scroll through the code and check if the comments have been addressed (I recommented on the ones I saw, but I may have missed a few).

source/particle/property/crystal_preferred_orientation.cc
MFraters: I usually use Backward Euler, so I changed the default to that. I also removed Crank-Nicklson for now (will post about it when I push a new version which also removes the pack and unpack functions)

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -0,0 +1,414 @@
...
+           */
+          static
+          unsigned int
+          get_number_of_minerals();
+
+        private:
+          /**
+           * Random number generator used for initalization of particles
+           */
+          mutable boost::lagged_fibonacci44497 random_number_generator;
```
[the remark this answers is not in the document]
MFraters: the `lagged_fibonacci44497` is more random for a lot of numbers, but I guess mt19937 should be sufficient as well. I changed it.

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1160 @@
...
+        // Note is that we store exactly the same number of grains of all minerals (e.g. olivine and enstatite
+        // grains), although their volume fractions may not be the same. We need a minimum amount
+        // of grains per tracer to perform reliable statistics on it. This minimum is the same for all phases.
+        // and enstatite.
+        //
+        // Futhermore, for this plugin the following dim's are always 3. When using 2D an infinitely thin 3D domain is assumed.
+        //
+        // The rotation matrix is a direction cosine matrix, representing the orientation of the grain in the domain.
+
+        // fabric. This is determined in the computations, so set it to -1 for now.
```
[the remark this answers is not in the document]
MFraters: I don't think there is something missing, but I reworded it a bit.

source/particle/property/crystal_preferred_orientation.cc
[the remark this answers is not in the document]
MFraters: I haven't really settled on one yet, but I have removed Crank Nicolson, since it is significantly more expensive an adds a lot of complexity. I think that it is good to keep the infrastructure to allow for multiple options in this case, so that Crank Nicolson can be added back if the need arises.

MFraters:
I am not sure what happened, because I am sure I had work on a version in which I removed the pack and unpack functions (which I thought I put here), but I haven't been able to find it. I should have noticed that, so sorry for that. I removed them again in favor of function which directly access the data array. The only data which is stored locally now is the matrix, since it is a lot easier to to the calculations with that. But that is now only done where it is needed. 

I decided to remove the Crank-Nicolson since it adds quite a bit of complexity, while also being much more expensive without having much evidence that it is much more accurate in practice. If desired it can always be added back in later. 

I also change the random number generator to mt19937 as suggested and made sure to do it in a separate commit, which worked before (and after) the change.

I went through all the unresolved comments, and made sure that they are resolved and set to resolved. Let me know if there are any other comments.

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -495,40 +278,23 @@ namespace aspect
             * derivatives are used to advect the particle properties.
             */
             double sum_volume_mineral = 0;
-            std::pair<std::vector<double>, std::vector<Tensor<2,3>>> derivatives_grains = this->compute_derivatives(volume_fractions_grains[mineral_i],
-                                                                                            rotation_matrices_grains[mineral_i],
+            std::pair<std::vector<double>, std::vector<Tensor<2,3>>> derivatives_grains = this->compute_derivatives(data_position,data,mineral_i,
```
gassmoeller, line 281: please put parameters on separate lines

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -255,7 +222,93 @@ namespace aspect
           unsigned int
           get_number_of_minerals();
 
+          /**
+           * return a reference to the value in the data array representing the
+           * deformation type.
+           */
+          inline
+          double &ref_deformation_type(const unsigned int cpo_data_position,
```
gassmoeller, line 230: Already better, but returning a reference is a somewhat dangerous design, because it is easy to accidentally change a value that you just want to read (plus the naming with ref_ ... is awkward). I would prefer you split all of these into get_ set_ functions like you do below.
I can see two reasons why you did it this way: 1. less function. 2. the construct below where you do `ref_function(...) *= some number` which saves you one function call by having the reference. However, for both cases I would still prefer to have separate getter/setter functions like you do for the rotation matrix. The number of functions does not matter much, since it is just a single line per function. And the performance loss by having to call the get_ function before multiplying the value and then calling the set_ function is actually very small (compared to creating a lot of vectors like you did before). Consider the following:
- A multiplication or addition costs around one clock cycle. You do maybe 10 of those in the function to determine the index plus you inline the function so the function call is nearly free.
- Accessing a variable that is already in L1 cache of the processor costs at most a few cycles (<10). This happens when you call the set_ function after calling the get_ function.
- Creating a variable on the stack (like a double or a single Tensor) costs less than 5 cycles (probably more like 1).
- So the total overhead of having separate get_/set_ functions is likely less than 20 clock cycles.
- Creating a single variable on the heap (like a std::vector) requires calling a complex algorithm that finds the necessary space, and on average costs around 90 clock cycles. 

So unless you have measured that this is indeed a performance bottleneck I would prefer an interface of the following form for all the function below:
```
inline
double get_deformation_type (... like before) const;

inline
void set_deformation_type (... like before ..., const double deformation_type) const;
```

gassmoeller (commented):
Much better thank you! I have a suggestion for improving the interface of the new functions, but everything else looks ready.

MFraters:
@gassmoeller Thanks for the review. I have addressed your comments :)

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -308,19 +312,18 @@ namespace aspect
 
             for (unsigned int grain_i = 0; grain_i < n_grains; ++grain_i)
               {
-                ref_volume_fractions_grains(data_position,data,mineral_i,grain_i) *= inv_sum_volume_mineral;
-                Assert(isfinite(ref_volume_fractions_grains(data_position,data,mineral_i,grain_i)),
+                set_volume_fractions_grains(data_position,data,mineral_i,grain_i,get_volume_fractions_grains(data_position,data,mineral_i,grain_i*inv_sum_volume_mineral));
```
gassmoeller, line 315: I think you misplaced a parenthesis in the last call. Make sure this is identical to the earlier version. I would recommend:
```
const double volume_fraction = get_volume_fractions_grains(data_position,data,mineral_i,grain_i) * inv_sum_volume_mineral;
set_volume_fractions_grains(...)
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -287,13 +286,18 @@ namespace aspect
               {
                 case AdvectionMethod::forward_euler:
 
-                  sum_volume_mineral = this->advect_forward_euler(data_position,data,mineral_i,dt,
+                  sum_volume_mineral = this->advect_forward_euler(data_position,data,
```
gassmoeller, line 289: Put `data`on its own line as well.

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -260,13 +260,10 @@ namespace aspect
...
           }
 
-        std::vector<unsigned int> deformation_types;
-        std::vector<double> volume_fraction_mineral;
-
         for (unsigned int mineral_i = 0; mineral_i < n_minerals; ++mineral_i)
           {
 
-            ref_deformation_type(data_position,data,mineral_i) = (unsigned int)DeformationType::passive;
+            set_deformation_type(data_position,data,mineral_i,(unsigned int)DeformationType::passive);
```
gassmoeller, line 266: We try to avoid C-style casts, better `static_cast<unsigned int>(DeformationType::passive)`

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -424,16 +427,16 @@ namespace aspect
...
             // Do the volume fraction of the grain
-            Assert(std::isfinite(ref_volume_fractions_grains(cpo_index,data,mineral_i,grain_i)),ExcMessage("volume_fractions[grain_i] is not finite before it is set."));
-            ref_volume_fractions_grains(cpo_index,data,mineral_i,grain_i) = ref_volume_fractions_grains(cpo_index,data,mineral_i,grain_i) + dt * ref_volume_fractions_grains(cpo_index,data,mineral_i,grain_i) * derivatives.first[grain_i];
-            Assert(std::isfinite(ref_volume_fractions_grains(cpo_index,data,mineral_i,grain_i)),ExcMessage("volume_fractions[grain_i] is not finite. grain_i = "
-                   + std::to_string(grain_i) + ", volume_fractions[grain_i] = " + std::to_string(ref_volume_fractions_grains(cpo_index,data,mineral_i,grain_i))
+            Assert(std::isfinite(get_volume_fractions_grains(cpo_index,data,mineral_i,grain_i)),ExcMessage("volume_fractions[grain_i] is not finite before it is set."));
+            set_volume_fractions_grains(cpo_index,data,mineral_i,grain_i,get_volume_fractions_grains(cpo_index,data,mineral_i,grain_i) + dt * get_volume_fractions_grains(cpo_index,data,mineral_i,grain_i) * derivatives.first[grain_i]);
```
gassmoeller, line 431: Split this into two lines as well, one call to `get_...` one call to `set_...`

gassmoeller (commented):
Nearly there, just a few minor comments left.

MFraters:
Thanks for the quick review! I have addressed your comments :)

gassmoeller (approved):
Great, ready to go when the tester is done! Oh, changelog entry please!

MFraters:
Thanks again for the quick review! hmm, I am fine with adding a changelog for this if you think it is important, but this pull request just adds some core infrastructure, which is not meant to be used yet. Wouldn't it be better to wait until some actual functionality is ready?

MFraters:
:tada:

PR #3905, opened 2020-11-26
Fix update_all_files.sh script

### LINKED ISSUE #3904: Update scripts in /doc
About 4 months ago, the update scripts were moved from /doc to /contrib/utilities. However, there is the script `update_all_files.sh` left in /doc that now no longer works as it calls the files now in /contrib/utilities.

tjhei:
Fix location and functionality of the update_all_files.sh script.

fixes #3904


contrib/utilities/update_all_files.sh
```diff
@@ -5,16 +5,17 @@
 # effect, because all files should be on the current version, but if you change
 # the code, or add new files it can be handy to execute this script.
 
-DOC_DIR=`dirname $0`
-BASE_DIR=`pwd $DOC_DIR/..`
```
tjhei, line 9: I don't understand how this could have worked. ``pwd`` does not take any arguments.

tjhei:
@gassmoeller I know you have used this in the past: When I run this, I get a significant number of changes in .cc and .prm files. I think we should address these, right?

gassmoeller (approved):
Yes this script by now makes some unwanted changes. It worked at some point, but then things got changed that do not interact well with the string substitutions (like output directories that contain strings that have changed in the .prm files). I am not confident we can really fix this, except at some point we can remove very old update scripts. I think this was more a stopgap solution during the update to release 2.0. Not sure how to proceed.

But independently of that, this PR makes sense!

PR #3908, opened 2020-11-29
ViscoPlastic rheology migrated into separate rheology module

bobmyhill:
This PR splits the rheological parts of the ViscoPlastic material model into a new Rheology model.
The intention is to make it easier for users to use the viscoplastic rheology in their own material models while using a different equation of state. 

### DIFF
--- a/include/aspect/material_model/rheology/visco_plastic.h
+++ b/include/aspect/material_model/rheology/visco_plastic.h
@@ -108,16 +108,38 @@ namespace aspect
                             const std::shared_ptr<std::vector<unsigned int>> &expected_n_phases_per_composition =
                               std::shared_ptr<std::vector<unsigned int>>());
 
-          double min_strain_rate;
-          double ref_strain_rate;
-          double min_visc;
-          double max_visc;
+
+
           double ref_visc;
+          double min_strain_rate;
 
           /**
            * Enumeration for selecting which viscosity averaging scheme to use.
            */
           MaterialUtilities::CompositionalAveragingOperation viscosity_averaging;
+
+
+          Rheology::StrainDependent<dim> strain_rheology;
+          /*
+           * Input parameters for the drucker prager plasticity.
+           */
+          Rheology::DruckerPragerParameters drucker_prager_parameters;
+          /**
+           * Object for computing viscoelastic viscosities and stresses.
+           */
+          Rheology::Elasticity<dim> elastic_rheology;
+
+          /**
+           * Whether to include viscoelasticity in the constitutive formulation.
+           */
+          bool use_elasticity;
+
+
+        private:
+
+          double ref_strain_rate;
+          double min_visc;
+          double max_visc;
 
           /**
            * Enumeration for selecting which type of viscous flow law to use.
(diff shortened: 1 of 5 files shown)


include/aspect/material_model/visco_plastic.h
```diff
@@ -276,11 +235,10 @@ namespace aspect
 
       private:
 
-        double min_strain_rate;
-        double ref_strain_rate;
-        double min_visc;
-        double max_visc;
-        double ref_visc;
+        /**
+         * Object for computing viscosities.
```
naliboff, line 239: I think the documentation for this object should be expanded quite a bit. While one the functions does compute the viscosity, this object also points to functions for additional material model outputs, the Newton derivatives, etc. Perhaps a more accurate single sentence description would be 'Object for rheology operations'. However, my preference would be to have an expanded description.

naliboff (changes requested):
👏 👏 👏 👏  - A round of applause, this is fantastic! I have one comment regarding documentation, but that is it for now. 

I think one of the other developers should have a look as well before merging, but in the end this is just moving code and I'm having a hard time anticipating any issues that the tester would not catch. @anne-glerum @MFraters @jdannberg @gassmoeller - Does one of you have time to take a look over the next few days?

bobmyhill:
Thanks for the quick review! I've extended the documentation as requested.

naliboff (approved):
@bobmyhill - Thanks, the new documentation looks great! Good to go from my end, but lets wait for another review.

anne-glerum (approved):
Thanks @bobmyhill ! I've only got textual/formatting comments, and most of those are about things not introduced by this PR. Not sure whether those should be addressed in a separate PR or can just be added to this PR. What do you think @naliboff ?

naliboff:
@anne-glerum - Thanks for the second review! If @bobmyhill does not mind documenting the parameters mentioned above, maybe easiest to do it in this PR? Bob - if you are ok with this, I'll add some suggestions for the missing parameter documentation to speed things up.

bobmyhill:
Thanks @anne-glerum! The commenting is much more uniform now :) 
@naliboff, I've made the changes suggested. Is anything else needed, or is this ready to go now (once the tester confirms that it is still happy)?

include/aspect/material_model/rheology/visco_plastic.h
```diff
@@ -183,13 +191,22 @@ namespace aspect
 
 
         private:
+
+          /**
+           * Reference strain rate for the first time step.
```
naliboff, line 196: Suggestion: `Reference strain rate for the first non-linear iteration in the first time step`

include/aspect/material_model/rheology/visco_plastic.h
```diff
@@ -160,15 +159,24 @@ namespace aspect
                                      const MaterialModel::MaterialModelInputs<dim> &in,
                                      MaterialModel::MaterialModelOutputs<dim> &out) const;
 
+          /**
+           * Reference viscosity used by material models using this rheology.
+           */
           double ref_visc;
+
+          /**
+           * Minimum strain rate used to stabilize the strain dependent rheology.
```
naliboff, line 168: Should this be` strain rate dependent rheology`?

naliboff (commented):
@bobmyhill - Two more minor comments on my end (feel free to ignore them), but after that ready to merge. Thank you again for this major improvement!

bobmyhill:
Ok, great! I've made those changes, so this is now ready to go :)

PR #3911, opened 2020-12-02
Simpler material model: remove reference_T

tjhei:
The member "reference temperature" is unused inside Simpler (but it is
declared and used in the EquationOfState::LinearizedIncompressible EoS).
This means this change should be backwards-compatible.


gassmoeller (approved):
:+1:

PR #3914, opened 2020-12-04
Update current stress taking strain weakening into account

anne-glerum:
The current viscous/visco-elastic stress is computed using the `pre_yield_viscosity`. However, this `pre_yield_viscosity` is later updated to take viscous strain weakening into account. This PR updates the `current_stress` accordingly.

### For all pull requests:


bobmyhill:
This is going to make #3908 harder to review and merge :/ 
I'm happy to amend that PR, but it would be good to merge it sooner rather than later.

bobmyhill:
Yes, I think you're correct. This is needed to assess whether the weakened material should yield plastically or not. 

It isn't clear to me what happens in the event that elasticity is turned on, but this certainly looks correct for viscoplastic.

naliboff:
@anne-glerum - Thanks for this fix and everything looks good. I was surprised that this does not change the test with viscous strain weakening (`visco_plastic_yield_viscous_strain_weakening.prm`). However, as indicated by the name, yielding occurs throughout the model and still occurs even with the properly updated `current_stress` term.

Would it be worth adding a test that does not include yielding or updating the current test to have larger a reduction in the viscosity with strain-weakening such that yielding does not occur? This can also be done in a follow-up PR.

anne-glerum:
@naliboff I rebased to master after #3908 was merged, and I added a test that does not plastically yield everywhere due to some viscous strain. I checked that without the strain weakening added to the current stress, yielding does occur everywhere. I just have to replace the test output with that from the tester.

naliboff (commented):
@anne-glerum - thanks, this looks good! All that is left to do is fix the test and then add a changelog entry indicating that the stress is now recomputed after viscous strain weakening.

anne-glerum:
done!

naliboff (approved):
@anne-glerum - Thanks!

PR #3916, opened 2020-12-08
Change test viscosity from 0 to 1

anne-glerum:
This PR makes 2 changes to several test prms to deal with the changes introduced by #3681.

1) Change Viscosity of `supg.prm` and `smoothing.prm` from 0 to 1
The changes in #3681 make these two tests fail because the reference viscosity is then computed based on the average viscosity (which is zero), which has to be >0. Because the Stokes equations are not solved for these tests, I don't think the tester will complain.  

2) Set a minimum strain rate for the set of `nonlinear_channel_flow_velocities_Newton_Stokes` tests.
This gives a better initial guess of the strain rate, hence leads to a more representative reference viscosity. Will probably change the amount of inner iterations in the first (few) nonlinear iteration.

### For all pull requests:


gassmoeller:
The changes make sense. If you think the test changes are reasonable please update, afterwards this looks good to me. @MFraters: Do you also want to take a look at the tests?

bangerth (approved):

bangerth:
Let's see how this tests:

/rebuild

anne-glerum:
As expected, the changes to the viscosity in `supg.prm` and `smoothing.prm` did not change the test results, but setting the `Minimum strain rate` did for the `nonlinear_channel_flow_velocities_Newton_Stokes`tests. The latter tests run to completion, but the initial Newton Stokes residual changes as well as iterations counts and residuals. What do you think @MFraters ?

MFraters:
Sorry for the late reply, I had missed this in my email. I don't really understand why you want to change the minimum strain-rate, but it should not cause any issues. Looking at the test results it actually becomes an issue though... 

The problem here is seems to be that the `Initial Newton Stokes residual` now starts out more than an order of magnitude lower. This is caused by the strain-rate given to the material model being much closer to the solution of the problem, which gives a better viscosity value. This doesn't really matter in normal cases, but in this case I probably pushed the nonlinear residual and set the nonlinear tolerances way too strict (to make sure that it has enough nonlinear iterations to actually show something). You should be able to safely lower the tolerance one or two orders of magnitude without any issue. Could you try and lower the nonlinear value until test tests converge again and have a similar number of iterations?

anne-glerum:
> Sorry for the late reply, I had missed this in my email. 

No worries, thanks for having such a quick look now. 

>I don't really understand why you want to change the minimum strain-rate, but it should not cause any issues. 

At the moment, the reference viscosity is a user-set parameter. #3681 computes the reference viscosity based on the average viscosity in the domain, which for strain rate dependent rheologies can be quite off from the average viscosity after solving when the minimum strain rate is left to the very low default. These Newton tests fail in #3681, and setting the minimum strain rate is part of getting them to work. 

> 
> The problem here is seems to be that the `Initial Newton Stokes residual` now starts out more than an order of magnitude lower. This is caused by the strain-rate given to the material model being much closer to the solution of the problem, which gives a better viscosity value. This doesn't really matter in normal cases, but in this case I probably pushed the nonlinear residual and set the nonlinear tolerances way too strict (to make sure that it has enough nonlinear iterations to actually show something). You should be able to safely lower the tolerance one or two orders of magnitude without any issue. Could you try and lower the nonlinear value until test tests converge again and have a similar number of iterations?

To summarize:
* Most tests do converge within the allowed number of nonlinear iterations, some are cut off
* Some tests need less nonlinear iterations than before
* Some tests do 0 inner iterations in timesteps >0
* For example, test `nonlinear_channel_flow_velocities_Newton_Stokes_no_line_search` needs less (28 instead of 30) iterations at t0 to get to a much better nonlinear residual (e-13 vs e-9), but then only needs 1 nonlinear iteration instead of 9 in timestep 1.

Decreasing the nonlinear tolerance as you asked will increase the number of iterations needed in later timesteps, but it won't increase the amount of inner iterations right?

MFraters:
There are two thing going on here. 

The first thing is what I mentioned above and is mainly targetted at test which really fail now (`nonlinear_channel_flow_velocities_Newton_Stokes_GMG`) or show strange values in the statistics (for example `nonlinear_channel_flow_velocities_Newton_Stokes_no_derivative`). For the strange values in the statistics I now see that they are correlated with the 0 linear iterations, so there is probably something doing 0-1 on a unsigned int. Setting the linear tolerance a bit stricter should help against this. 

The second thing is that with the new strain the initial guess is apparently also better, and the thing with the Newton iteration is that the closer it is to the solution, the quicker it converges. It seems that in some cases this advantage overcomes the disadvantage of the lower initial residual. 

The main thing to look at when fixing these tests is that the norm of the rhs should be similair at the end. So for example for the `nonlinear_channel_flow_velocities_Newton_Stokes_GMG` test, the reference value at the end of timestep 1 (nonlinear iteration 6) is `norm of the rhs: 346183`.  In this case the norm of the rhs is exactly the same (since it is using dc picard) at iteration 6, but because of the different initial residual the relative residual is different and it keeps iterating to crash at a rhs norm of `44511`. So in this case you should set the tolerance to 1e-11 and it should work again.

So in summary: 
- Set nonlinear tolerance of GMG test to 1e-11
- If there are no problems with the new test result, you can leave them.
- In general, set the nonlinear tolerance to result in a similar rhs
- For the tests which have 0 linear iterations, set the linear solver tolerance stricter.

anne-glerum:
> There are two thing going on here.
> 
> The first thing is what I mentioned above and is mainly targetted at test which really fail now (`nonlinear_channel_flow_velocities_Newton_Stokes_GMG`) 

This test no longer fails when I make the nonlinear solver tolerance _less_ strict, because we never reach the failing solve.

> or show strange values in the statistics (for example `nonlinear_channel_flow_velocities_Newton_Stokes_no_derivative`). For the strange values in the statistics I now see that they are correlated with the 0 linear iterations, so there is probably something doing 0-1 on a unsigned int. Setting the linear tolerance a bit stricter should help against this.

Actually, there are several tests that already have these maximum integer numbers in the statistics (without the changes in this PR), e.g `nonlinear_channel_flow_velocities_Newton_Stokes_no_deviator.prm` and `nonlinear_channel_flow_velocities_single_Advection_Newton_Stokes.prm`. Did you want the tests to do 0+0 inner iterations for several timesteps, or should those be updated?

> Could you try and lower the nonlinear value until test tests converge again and have a similar number of iterations?

I was confused by this, because the nonlinear tolerance should be _less_ strict, so that the rhs norm is of the same size instead of being smaller than before, right? I've played with the tolerances for several tests (see new test results), but I need to know what to do with the 0+0 inner iterations first, I guess.

MFraters:
> This test no longer fails when I make the nonlinear solver tolerance less strict, because we never reach the failing solve.

That is fine in the sense that the old test also would have failed with a much stricter tolerance.

> Actually, there are several tests that already have these maximum integer numbers in the statistics (without the changes in this PR), e.g `nonlinear_channel_flow_velocities_Newton_Stokes_no_deviator.prm` and `nonlinear_channel_flow_velocities_single_Advection_Newton_Stokes.prm`. Did you want the tests to do 0+0 inner iterations for several timesteps, or should those be updated?

hmm, if there are others you can try to fix them, but that is probably not really something for this pull request. I am fine with leaving them in, even if they are new in this pull request, since it is a different issue. But if you want to try to fix them, that is also fine with me (see next point).

> I was confused by this, because the nonlinear tolerance should be less strict, so that the rhs norm is of the same size instead of being smaller than before, right? I've played with the tolerances for several tests (see new test results), but I need to know what to do with the 0+0 inner iterations first, I guess.

Ah, sorry, I always have sometimes a bit of trouble speaking about tolerances. Yes, I meant making the nonlinear tolerances less strict. To fix the 0+0, setting the linear tolerance (`Maximum linear Stokes solver tolerance` in the newton solver parameters section) stricter should help.

anne-glerum:
> hmm, if there are others you can try to fix them, but that is probably not really something for this pull request. I am fine with leaving them in, even if they are new in this pull request, since it is a different issue. But if you want to try to fix them, that is also fine with me (see next point).

Alright, I'll leave the existing ones in, I also think that can/should be a different PR. If they are new, I'll try to fix it with the `Maximum linear Stokes solver tolerance`.

anne-glerum:
Hi @MFraters , I adapted the tests as best as I could to get the norms of the rhs on the same order of magnitude.
The following tests already did zero iterations:
    `nonlinear_channel_flow_velocities_Newton_Stokes`
    `nonlinear_channel_flow_velocities_Newton_Stokes_no_deviator`
    `nonlinear_channel_flow_velocities_Newton_Stokes_no_line_search`

For this test, I cannot get the same evolution of the rhs norm over time:
    `nonlinear_channel_flow_velocities_Newton_Stokes_no_line_search`
Should I change something else or is it OK the way it is?

MFraters (approved):
Wow, the convergence for `nonlinear_channel_flow_velocities_Newton_Stokes_no_line_search` is really a lot better. The important thing for most these test is to have multiple nonlinear iterations after the first timestep, to because that is when the DC is actually used. In this case, the test is just there to make sure that there is no line search when you say there should be no line search. This test still shows that, so I am fine with leaving it as it is.  
If you fix the results so the tester passes again, I think it is ready to merge.

anne-glerum:
Great! I've updated the tests!

gassmoeller:
Awesome, thank you, and thanks @MFraters for the additional review. I am going to merge.

PR #3917, opened 2020-12-08
Fix changelog

gassmoeller:
We introduced a stray `<li>` in our changelog that got copied into many entries. See: https://aspect.geodynamics.org/doc/doxygen/changes_current.html.

This should fix the issue.


naliboff (approved):
👍 thanks for the fix @gassmoeller!

PR #3919, opened 2020-12-08
Use deal.II particle copy

gassmoeller:
Since dealii/dealii#10858 was merged we should now be using the deal.II functionality to copy particle handlers. All the functionality that was written in this function (including ghost particle exchange) happens inside deal.II now. I checked that all tests still run with the most recent deal.II.


tjhei (approved):

bangerth (approved):
Always good to remove some code (or allow it to happen in the future) :-)

PR #3920, opened 2020-12-08
Fix deprecation warnings

gassmoeller:
This fixes the deprecation warnings we get when compiling against the latest deal.II version.


bangerth (approved):
Lots of stuff hackathon participants can later remove!

PR #3923, opened 2020-12-12
implement timestep repeating with mesh deformation

tjhei:
We forgot to undo mesh deformation when we repeat a timestep. Fix that.

### DIFF
--- a/include/aspect/mesh_deformation/interface.h
+++ b/include/aspect/mesh_deformation/interface.h
@@ -412,6 +412,11 @@ namespace aspect
         LinearAlgebra::Vector mesh_displacements;
 
         /**
+         * mesh_displacements from the last time step.
+         */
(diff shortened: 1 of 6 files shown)


tjhei:
@EstherHeck it took me way too long, but this might work. Can you check?

EstherHeck:
Hi @tjhei, thanks for getting back to this! I tried it with the free surface tractions benchmark and indeed it works. When I set the maximum relative increase parameter, I do get the same magnitude of surface movement as in the analytical solution! See more detail in #3770

tjhei:
> and indeed it works

Awesome, thank you for checking!

tjhei:
I think this is ready to go.

gassmoeller (approved):
Looks good, thanks!

PR #3924, opened 2020-12-12
remove some unnecessary includes

tjhei:


bangerth (approved):

bangerth:
It passes 3 of the 4 tests. That seems good enough to me given that it doesn't actually affect any functionality.

PR #3925, opened 2020-12-16
add defect correction to list of allowed solvers for strain dependent

EstherHeck:
I added the solver scheme _'single advection, iterated defect correction Stokes'_ to the list of allowed solvers for strain dependent rheology. only solver schemes with a single advection should be used for strain dependent, which is the case for this one. 

Doing this I also noticed that I cannot find this solver scheme in the parameters.prm file, as it is not listed in _prm.declare_entry ("Nonlinear solver scheme"_ in parameters.cc. It would be nice if someone who knows more about how it works could write a little paragraph about it

### For all pull requests:


MFraters (changes requested):
I don't know much about the strain dependent part, but for the nonlinear solver part this looks good to me. Could you add a test and make sure that the results are similar enough to the other scheme?

You are correct that the description of the defect correction schemes have not been added to `parameters.cc`. That is my mistake, and I will fix it in a different pull request.

MFraters (approved):

naliboff (changes requested):
@EstherHeck - thanks for these changes! Can you also add `'single Advection, iterated defect correction Stokes'` to the list allowed solvers in `source/material_model/rheology/elasticity.cc` (lines 172-182)?

EstherHeck:
good point, forgot that I encountered the same issue foe elastic stresses...

naliboff:
@EstherHeck @MFraters - Forgot to add that this solver scheme is certainly ok to use in strain_dependent.cc. We just can't do multiple advection solves when using the compositional fields to track strain, elasticity, etc. 

@MFraters - Would be possible to enable using the `Use Eisenstat Walker method` for the defect-correction Picard outside of the Newton solver interface, or would this be a major overhaul?

MFraters:
> @MFraters - Would be possible to enable using the Use Eisenstat Walker method for the defect-correction Picard outside of the Newton solver interface, or would this be a major overhaul?

I am not sure I understand your question, so let me give a few answers. Technically it is easy to use where ever in the code you want since the function to compute the tolerance is in `simulator/helper_functions.cc`. If you mean you want to use the Eisenstat Walker method for the DC Picard, then that should already work. Just set the `set Use Eisenstat Walker method for Picard iterations = true` in the subsection `Solver parameters->Newton solver parameters`.

naliboff:
> Just set the set Use Eisenstat Walker method for Picard iterations = true in the subsection Solver parameters->Newton solver parameters.

@MFraters - Will this work if you specify the solver as `'single Advection, iterated defect correction Stokes'`? In other words, if you don't specifically use a Newton solver scheme will the relevant parameters for the Eisenstat Walker method in the `Newton solver parameters` subsection still be accessed?

MFraters:
@naliboff, I think we are going a bit off topic here, or is this question related to this pull request? 

To answer the question: I don't think I  specifically tested for it, but it should work for the defect correction schemes since they are just a limited shortcut for the Newton solver code (keeping the derivatives always zero). I am a bit curious, do you have experience that using the Eisenstat Walker method during the Picard part is useful for your cases?

tests/visco_plastic_yield_plastic_strain_weakening.prm
```diff
@@ -1,133 +1,5 @@
...
 
-# Velocity on boundaries characterized by functions
-subsection Boundary velocity model
-  subsection Function
-    set Variable names      = x,y
-    set Function constants  = m=0.0005, year=1
-    set Function expression = if (x<50e3 , -1*m/year, 1*m/year); if (y<50e3 , 1*m/year, -1*m/year);
-  end
-end
+include $ASPECT_SOURCE_DIR/tests/visco_plastic_yield_plastic_strain_weakening.prm
```
naliboff, line 2: Ok, I'm a bit confused here. The include here is referring to the file it is located in. I am just not reading the file names correctly?

naliboff (commented):
@EstherHeck - thanks for updating` elasticity.cc`! You will need to update the test output and I also have one question about the include added to `visco_plastic_yield_plastic_strain_weakening.prm`.

tests/visco_plastic_yield_plastic_strain_weakening.prm
EstherHeck: Sorry, thanks for noticing that! My fault, thought I had fixed that before pushing... I was not reading filenames correctly yesterday evening, all so similar

naliboff:
@EstherHeck - I think this PR is to be merged after the test output is updated. Do you have time to do that sometime this week? This will be a useful addition for many users!

EstherHeck:
@naliboff reminded me that this didn't build because of indentation. Make indent didnt do anything, but the check said something about "no newline at end of file" for the new test I added, so I added a new line there and hope it builds now

naliboff:
@EstherHeck - Thanks, the tester is passing now! Can you squash these two commits? Once that is done, this is ready to merge. Thanks again!

naliboff (approved):
@EstherHeck - Thanks you again for these changes, this is ready to go!

PR #3927, opened 2020-12-17
Add visco plastic composition mask

MFraters:
This pull request adds the option for the user to mask compositional fields from use in the volume fraction calculation in the visco plastic material/rheology model (setting the volume fraction to zero). This is useful for for example the cpo code, since it requires a water compositional field which should not be used by the visco plastic plugin.

I have added a test which shows that the result has not changed by adding the extra composition compared to `visco_plastic_complex` (only extra degree's of freedoms).

### DIFF
--- a/source/material_model/rheology/visco_plastic.cc
+++ b/source/material_model/rheology/visco_plastic.cc
@@ -429,6 +429,8 @@ namespace aspect
               composition_mask.set(i,false);
           }
 
+        composition_mask = composition_mask & user_component_masks;
+
         return composition_mask;
       }
 
@@ -497,6 +499,15 @@ namespace aspect
                            "dynamic stresses are much higher than the lithostatic pressure. "
                            "If false, the minimum pressure in the plasticity formulation will "
                            "be set to zero.");
+        prm.declare_entry ("List of compositional field names to use", "",
+                           Patterns::Anything (),
+                           "Indicate which compositional field are used by the visco plastic "
+                           "rheology model by listing the names of the fields. If left empty "
+                           "all compositional fields will be used. Compositional fields not "
+                           "listed with receive a volume fraction of 0. This means that you "
+                           "still need to provide resonable values for prefactors, activation "
+                           "energies, etc. for them.");
+
 
         // Diffusion creep parameters
         Rheology::DiffusionCreep<dim>::declare_parameters(prm);
@@ -646,6 +657,17 @@ namespace aspect
                        ExcMessage("If adiabatic heating is enabled you should not add another adiabatic gradient"
                                   "to the temperature for computing the viscosity, because the ambient"
                                   "temperature profile already includes the adiabatic gradient."));
+
+        std::vector<std::string> component_masks_string = Utilities::split_string_list(prm.get("List of compositional field names to use"));
+        user_component_masks = component_masks_string.size() == 0
+                               ?
+                               ComponentMask(this->n_compositional_fields(),true)
+                               :
+                               ComponentMask(this->n_compositional_fields(),false);
+        for (auto &&mask : component_masks_string)
+          {
+            user_component_masks.set(this->introspection().compositional_index_for_name(mask), true);
+          }
 
       }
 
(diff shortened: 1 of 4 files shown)


source/material_model/rheology/visco_plastic.cc
```diff
@@ -497,6 +499,11 @@ namespace aspect
                            "dynamic stresses are much higher than the lithostatic pressure. "
                            "If false, the minimum pressure in the plasticity formulation will "
                            "be set to zero.");
+        prm.declare_entry ("List of component masks", "",
```
tjhei, line 502: I am not able to understand what this does from the parameter name and description. Can you try to improve this?

I assume this is currently "list of compositional field names to ignore"? I would prefer to negate this as "list of names of compositional fields to use" (and treat the empty case as meaning "all").

MFraters:
Thanks for the quick review. I agree that a `list of names of compositional fields to use` is preferable. I have changed it and expended the explanation a bit. Is this more clear?

source/material_model/rheology/visco_plastic.cc
```diff
@@ -497,6 +499,15 @@ namespace aspect
                            "dynamic stresses are much higher than the lithostatic pressure. "
                            "If false, the minimum pressure in the plasticity formulation will "
                            "be set to zero.");
+        prm.declare_entry ("List of compositional field names to use", "",
+                           Patterns::Anything (),
+                           "Indicate which compositional field are used by the visco plastic "
+                           "rheology model by listing the names of the fields. If left empty "
+                           "all compositional fields will be used. Compositional fields not "
+                           "listed with receive a volume fraction of 0. This means that you "
```
bobmyhill, line 507:
```suggestion
                           "listed will be assumed to not correspond to volume fractions. This means that you "
```

bobmyhill: Also suggest removing the last sentence, as it is confusing.

bobmyhill (commented):
Is this approach the best one to take? At the moment, the get_volumetric_composition_mask() function removes compositions that don't correspond to volume fractions by specific calls to other module functions, like: 
`strain_rheology.get_strain_composition_mask();` or `if use_elasticity`

Could you do the same thing for cpo? In pseudocode:
`if use_cpo_calculations and (some check that cpo expects water?)
composition_mask.set(this->introspection().get_composition_names().index('water'),false);

It would be good at some point to think about whether these compositional field assignments could be made more flexible/foolproof/consistent across material models. I might favour an extra parameter in the "Compositional fields" subsection, like "set Field type = material_fraction, material_fraction, cpo_variable, grainsize", that creates the masks before the first time step. Do you have any opinions/thoughts on this?

MFraters:
@bobmyhill Thanks for the review! I really like the idea of being able to set different field types. My guess would be that it would be a bit of work to make sure every part of the code accepts those directives. Since you are more aware of the code which uses it, what do you think? 

I hadn't thought of those field masks yet, so I will try them out and see if that can improve the code in the short term.

bobmyhill:
> I really like the idea of being able to set different field types. My guess would be that it would be a bit of work to make sure every part of the code accepts those directives. Since you are more aware of the code which uses it, what do you think?

It would certainly be a bit of work to sift through all the code. Adding the new parameter wouldn't be a breaking change, so the changes could be split into small PRs. The current hard-coding in some of the material models is confusing, so I'd quite like to see this feature implemented.

The prm files and source code could be modified in tandem so that users don't see changes in behaviour.

naliboff:
@MFraters @bobmyhill @anne-glerum @tjhei - should we return to this PR? Related to #4182 and #4246

anne-glerum:
Where needed rheologies already set masks for the fields they know to _not_ be used as material fractions, but those rheologies do control the evolution of those fields. If there are fields for which the evolution is not controlled by the used rheology, I think the rheology should not decide on their masks. Masking of those fields should happen on a higher level, like in Compositional fields as Bob suggested.

gassmoeller:
@MFraters: I think with #5029 and the field types added this PR is no longer needed? I will close for now, but feel free to reopen if it is still needed.

PR #3931, opened 2020-12-18
add defect correction solver schemes to parameters.cc parameter description

MFraters:
@EstherHeck noticed in #3925 that the defect correction schemes are not described in the declare entry of the Nonlinear solver schemes. This pull request adds that description.


gassmoeller (approved):
Looks good, thanks.

PR #3932, opened 2020-12-18
Particles plus repeated timestepping

anne-glerum:
This PR allows for the use of particles icw repeated timesteps, by copying the particle handler at the beginning of each timestep and restoring it when the timestep needs to be repeated. 

Reviewing the changed files, I should probably include another condition that ensures the copies are only made when the `repeat on cutback` timestepping plugin is used, not always when particles are used. 

I can also make the first copy in `run` instead of `start_timestep` and make the copy a local variable, but in #3760 I also need the copy to restore the particles after each nonlinear iteration. 

The below figure shows the results of the added test but with higher mesh and particle resolution. The left column is the result of this PR, the right column of mainline. The top row shows the field with the particle method + the particles + contours of both the field-field and the particle-field. The bottom row shows the field method field. Both compositional fields start out with the same initial condition prescribed (although slightly differently represented by the field and the particle method) and with this PR end up very similar.

Screenshot 2020-12-18 at 11 41 32

### For all pull requests:


tjhei:
Thanks for working on this! This seems to make sense, but I have a couple of comments:
1. I would prefer if we put the copy inside ``World`` instead of in the Simulator.
2. It is not clear to me if 100% of the state is in the ParticleHandler and not in the ``World`` right now (@gassmoeller ?). 
3. Avoiding the copy is not easy as we can not know if a plugin might ask for a repeat, unless we introduce a new function ``do_you_ever_plan_on_repeating_a_timestep`` or something to each plugin.

gassmoeller:
All about the particle state is stored inside the `ParticleHandler` so this should work. The only things stored in `World` are the input parameters for things like particle limits and chosen integrator, generator, and so on. Still, I agree it would be nice to store the copy inside World, maybe by adding functions `backup_particles` and `restore_particles`? This PR likely needs to be rebased once #3760 is merged?

anne-glerum:
Alright, I will store the ParticleHandler copy in World. The same thing will have to be done in #3760, where copies are made in solver_schemes.cc.

anne-glerum:
I've updated this PR after #3760 got merged and stored the Handler copy in world. The tests still fail however, and this is only due to the copy of the Handler stored in start_timestep(). It fails in `      triangulation.execute_coarsening_and_refinement ();` of `refine_mesh()` in the simulator for the last initial adaptive refinement. There is no helpful debug output, just std::exception. @gassmoeller do you have an idea why it would fail here?

gassmoeller:
I did not quite understand your message, the current state of this PR still has the copy in Simulator, not in World. Is the failing test output the same as for the current state of the PR? The one problematic point I could see is if there is adaptive refinement happening between storing and restoring the particle handler. Currently the particle handler copy is not connected to the triangulation signals that trigger the redistribution of particles during refinement. Is this the problem? If so, we need to modify the copy_particle_handler function to also connect the signals to the copy (meaning the refinement will also be more expensive because it has to redistribute the original as well as the copied particles).

anne-glerum:
Sorry, I'd forgotten to actually commit moving the PH copy to World :disappointed:. 

In `start_timestep` the PH copy is made after each initial adaptive refinement cycle. Also the PH copy signals are almost the same as the PH ones, only `setup_initial_state()` and `apply_particle_per_cell_bounds()` are not coupled for the former. The latter is not used when I set the `Load balancing strategy` to `none`, but the test still fails. What triangulation signals do you mean are not connected? I thought `register_store_callback_function` and `register_load_callback_function` are the ones to call before and after refinement.

gassmoeller:
I pushed two commits that solved most of the issues. We have a choice to make here. Because the backup/restore cycle includes a possible refinement, we need to connect the backup particle handler to the correct triangulation signals (but only once! this was a bug in the previous implementation of copy_particle_handler). However this also connects it to the checkpointing signals. Do we want to store the back particle handler in checkpoints? Currently this is not necessary, but we do something similar for the solution (we store solution, old_solution, and old_old_solution in the checkpoint). Is there value in having the old particle locations and properties upon restart?

anne-glerum:
Hi Rene, thanks for the fix! The test model I was using now runs as normal. 

Wouldn't the backup particle handler almost directly be overwritten by the current particle handler upon restart in `start_timestep`?  I don't see where could we use it before it's overwritten.

gassmoeller:
Ah, yes that is a good point. But that means we need to remove the backup particle handler again from world.h:serialize(), and we need to find a way to not connect the backup particle handler to the `pre_checkpoint_store_user_data` and `post_resume_load_user_data` signals inside world.cc::`connect_particle_handler_signals`. Maybe the function needs a new argument (bool connect_checkpoint_restore_signals), which is true when called with particle_handler, but false when called with particle_handler_backup?

anne-glerum:
Alright, I implemented the new argument and updated the test. What do you think @gassmoeller ?

include/aspect/particle/world.h
```diff
@@ -420,7 +434,8 @@ namespace aspect
          */
         void
         connect_particle_handler_signals(aspect::SimulatorSignals<dim> &signals,
-                                         ParticleHandler<dim> &particle_handler) const;
+                                         ParticleHandler<dim> &particle_handler,
+                                         const bool = true) const;
```
gassmoeller, line 438: This parameter is lacking a name.

source/particle/world.cc
```diff
@@ -67,6 +67,11 @@ namespace aspect
                                                                       this->get_mapping(),
                                                                       property_manager->get_n_property_components());
 
+      // Particle handler backup will not be stored for checkpointing
```
gassmoeller, line 70: This comment should go down to line 236

source/particle/world.cc
```diff
@@ -250,17 +268,21 @@ namespace aspect
         particle_handler_.register_load_callback_function(false);
       });
 
-      signals.pre_checkpoint_store_user_data.connect(
-        [&] (typename parallel::distributed::Triangulation<dim> &)
-      {
-        particle_handler_.register_store_callback_function();
-      });
+      // do not connect backup particle handler
```
gassmoeller, line 271:
```suggestion
      // Only connect to checkpoint signals if requested
```

gassmoeller (commented):
Very nice, just a few minor comments. And can you replace the .vtu particle output of the test with .gnuplot output? That makes comparisons easier in the future.

anne-glerum:
@gassmoeller I've addressed your comments, could you have a look? Thanks!

gassmoeller:
Sorry for dropping the ball on this. Looks good.

PR #3935, opened 2020-12-23
Urlreader fix

kneumiller:

Quick pull request to change how data is handled (and checked) from the requested url

### For all pull requests:

### DIFF
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -1028,8 +1028,8 @@ namespace aspect
                   libdap::AttrTable *table;
 
                   table = das.get_table(i);
-                  if (table->get_attr("points") != "")
-                    points.push_back(table->get_attr("points"));
+                  if (table->get_attr("POINTS") != "")
+                    points.push_back(table->get_attr("POINTS"));
                 }
 
               std::stringstream urlString;
@@ -1059,7 +1059,6 @@ namespace aspect
 
               data_string = urlString.str();
               filesize = data_string.size();
-
               delete url;
 #else // ASPECT_WITH_LIBDAP
 
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -981,9 +981,9 @@ namespace aspect
 
 
               // Temporary vector that will hold the different arrays stored in urlArray
-              std::vector<std::string> tmp;
+              std::vector<libdap::dods_float32> tmp;
               // Vector that will hold the arrays (columns) and the values within those arrays
-              std::vector<std::vector<std::string>> columns;
+              std::vector<std::vector<libdap::dods_float32>> columns;
 
               // Check dds values to make sure the arrays are of the same length and of type string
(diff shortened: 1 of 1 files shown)


tjhei:
@kneumiller can you explain what this fix changes please? Did something change on the server side or is this an optimization?

kneumiller:
> @kneumiller can you explain what this fix changes please? Did something change on the server side or is this an optimization?

The fix was because we wanted to change how the csv data was stored on the server from string to floats so I modified the url reader to check for this instead. The main changes are with the data itself and how it's stored on the server but I needed to update the code in Aspect to reflect these changes.

tjhei:
> we wanted to change how the csv data was stored on the server

When is this change on the server going to happen? Does this mean that the old code stops working at that moment or can the server serve both?

kneumiller:
We pushed the changes up to the server. We decided to have the url reader extension look for float32 values instead of strings. The old code won't work because the old file (the one with string values) had the data mangled when it was uploaded.
I can probably modify it so that it can check for both, but we assumed just passing numeric values as floats instead of strings was better,

tjhei:
> We pushed the changes up to the server. We decided to have the url reader extension look for float32 values instead of strings. The old code won't work because the old file (the one with string values) had the data mangled when it was uploaded.

OK. When you say "mangled", are you saying that the old client/server was not producing correct data? I thought this was tested to work?

I know backwards compatibility is difficult and not easy for online components, but I am a bit sad that the server suddenly changes and libdap in ASPECT 2.2 is now broken. There are plenty of people who use ASPECT 2.2. Anyways, I am happing to move forward with this change now, but it would be great to be more careful about these things in the future. One could keep old datasets in the old format, for example.

tjhei:
Can you please rebase to the current ASPECT master to remove the merge commit?

kneumiller:
> Can you please rebase to the current ASPECT master to remove the merge commit?

Yup! It should just be the two commits now, sorry for the hassle.



> > We pushed the changes up to the server. We decided to have the url reader extension look for float32 values instead of strings. The old code won't work because the old file (the one with string values) had the data mangled when it was uploaded.
> 
> OK. When you say "mangled", are you saying that the old client/server was not producing correct data? I thought this was tested to work?
> 
> I know backwards compatibility is difficult and not easy for online components, but I am a bit sad that the server suddenly changes and libdap in ASPECT 2.2 is now broken. There are plenty of people who use ASPECT 2.2. Anyways, I am happing to move forward with this change now, but it would be great to be more careful about these things in the future. One could keep old datasets in the old format, for example.

Yeah, there was some miscommunication between Opendap and the Balto team on the data being uploaded to our servers. The data itself was the issue and hopefully this fix will prevent it from happening in the future.

tjhei (approved):

tjhei:
Thanks!

PR #3936, opened 2021-01-01
fix incorrect initial surface topography

tjhei:
I am not sure if this code is tested right now and if it works
correctly, but writing into invalid memory can't be good.

### DIFF
--- a/source/mesh_deformation/interface.cc
+++ b/source/mesh_deformation/interface.cc
@@ -843,7 +843,7 @@ namespace aspect
                     else
                       {
                         for (unsigned int d=1; d<dim; ++d)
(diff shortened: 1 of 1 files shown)


gassmoeller (approved):
This is clearly a fix. It seems we are missing a test for initial topography for spherical geometries using the new initial mesh deformation interface. I will open an issue about it.

anne-glerum:
> This is clearly a fix. It seems we are missing a test for initial topography for spherical geometries using the new initial mesh deformation interface. I will open an issue about it.

To add, this fixed function is used when the initial topography is specified through the old interface under Geometry model, which was not possible for spherical shells.

PR #3944, opened 2021-01-07
testsuite: do not require timeout command

tjhei:
It turns out the the GNU timeout command is not available on OSX by
default. Work around this by ignoring the time limit on OSX.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -311,13 +311,13 @@ FOREACH(_test ${_tests})
     IF("${_mpi_count}" STREQUAL "1")
       ADD_CUSTOM_COMMAND(
         OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/output-${_testname}/screen-output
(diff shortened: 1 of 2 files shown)


tjhei:
I have no way to test this on OSX, but it works when hard-coding the if to be false. As reported by @anne-glerum .

anne-glerum:
I just tested this on my machine (after uninstalling coreutils and checking that the tester fails), and it works for `ctest -R additional_outputs_02`.

gassmoeller (approved):
Looks good to me. Not sure we state anywhere that we require bash is installed (`shift` seems to be bash specific), but I guess it is not a strong limitation and we may require it already in many places.

bangerth:
Why do you even need this? CMake has a default mechanism for setting timeouts, and we already use it at the bottom of the file:
```
    ADD_TEST(NAME ${_testname}
      COMMAND ${CMAKE_COMMAND}
              -DBINARY_DIR=${CMAKE_BINARY_DIR}
              -DTESTNAME=tests.${_testname}
              -DERROR="Test ${_testname} failed"
              -P ${CMAKE_SOURCE_DIR}/run_test.cmake
      WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
    SET_TESTS_PROPERTIES(${_testname} PROPERTIES TIMEOUT ${TEST_TIME_LIMIT})
```

tjhei:
I should have documented this better (I will make a PR), but basically the ctest timeout is not respected when we "prebuild" tests on the tester: #3605

PR #3943, opened 2021-01-06
Refactor nullspace removal

gassmoeller:
In preparation for a new nullspace removal option (remove only surface net rotation), this PR refactors the existing function `remove_net_angular_momentum`. 
No changes in tests should happen.

PR #3947, opened 2021-01-11
Refactor rotation statistics postprocessor

gassmoeller:
Similar to #3943, but this unifies the code used in the rotation statistics postprocessor with the similar code in nullspace removal. There should be no changes in behavior.

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -134,6 +134,38 @@ namespace aspect
   };
 
   /**
+   * A data structure with all properties relevant to compute angular momentum and rotation.
+   */
+  template <int dim>
+  struct RotationProperties
+  {
+    RotationProperties()
+      :
+      scalar_moment_of_inertia(numbers::signaling_nan<double>()),
+      scalar_angular_momentum(numbers::signaling_nan<double>()),
+      scalar_rotation(numbers::signaling_nan<double>()),
+      tensor_moment_of_inertia(numbers::signaling_nan<SymmetricTensor<2,dim>>()),
+      tensor_angular_momentum(numbers::signaling_nan<Tensor<1,dim>>()),
+      tensor_rotation(numbers::signaling_nan<Tensor<1,dim>>())
+    {};
+
+    /**
+     * Scalar properties for the two-dimensional case
+     * with a fixed rotation axis (z).
+     */
+    double scalar_moment_of_inertia;
+    double scalar_angular_momentum;
+    double scalar_rotation;
+
+    /**
+     * Tensor properties for the three-dimensional case.
+     */
+    SymmetricTensor<2,dim> tensor_moment_of_inertia;
+    Tensor<1,dim> tensor_angular_momentum;
+    Tensor<1,dim> tensor_rotation;
+  };
+
+  /**
    * This is the main class of ASPECT. It implements the overall simulation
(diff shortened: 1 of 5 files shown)


MFraters (commented):
The changes seem reasonable and an improvement to me. The tests still pass while the code while the relevant part of the code is fully covered. I also like that some of the template specializations have been unified in one function. 

The only thing I am hesitant about is that there is a separate set of variables in the `RotationProperties` structure for 2D and 3D. I think it would be cleaner in this case to specialize this into a 2D and 3D structure. What is your opinion on this?

gassmoeller:
I agree, but the problem is that the `compute_net_angular_momentum` function returns these properties. If I would generate separate 2D and 3D properties I would need to split the function in two, which would propagate into different simulatoraccess functions. That is why I think the current approach is cleaner than the alternative.

MFraters:
I might be missing something here but why can't you do something like this? Simulator access shouldn't have to know about whether it is 2D or 3D. (made the tensors into int's to make it compile outside aspect)

```C++
#include <iostream>
template<int dim> 
struct RotationProperties
{
};
template<> 
struct RotationProperties<2>
{
  RotationProperties()
  : 
  moment_of_inertia(1),
  angular_momentum(2),
  rotation(3)
  {};
    
  double moment_of_inertia;
  double angular_momentum;
  double rotation;
};

template<> 
struct RotationProperties<3> 
{
  RotationProperties() 
  :
  moment_of_inertia(4),
  angular_momentum(5),
  rotation(6)
  {};

  int moment_of_inertia;
  int angular_momentum;
  int rotation;
};
int main()
{
    RotationProperties<2> rp2d = RotationProperties<2>();
    RotationProperties<3> rp3d = RotationProperties<3>();
    
    std::cout << "2d: " << rp2d.moment_of_inertia << " " << rp2d.angular_momentum << " " << rp2d.rotation << std::endl;
    
    
    std::cout << "3d: " << rp3d.moment_of_inertia << " " << rp3d.angular_momentum << " " << rp3d.rotation << std::endl;
}
```
https://coliru.stacked-crooked.com/view?id=1f1e57bec0da1aa1

I also merged the names in this case because it allows for further generalization, but I can see the advantage of keeping them separate as `scalar_` and `tensor_`.

gassmoeller:
Thanks for pointing that out, I had in fact not know that you can specialize member variables in template classes. However, it wont work in this case unless I then also specialize every function that uses them (in particular the `execute` function in `RotationStatistics`, but also the `compute_net_angular_momentum` function in nullspace.cc). The reason is that in conditions like `if (dim == 2)` both branches (2D/3D) are compiled for all template instantiations, therefore I constantly run into problems like assigning a double to a tensor or vice versa. I can not even circumvent this by giving the members different names, because then the compiler thinks I am trying to assign something to a member that does not exist, even if this case will never happen (for dim == 3 I will never end up in the dim == 2 branch of the if condition). Apart from specializing everything, I think the current solution is the best.

MFraters (approved):
Ah, of course, you are right. For future reference, I reproduced the problem in the example by extending the above code snippet with an extra structure, class function, explicit instantiation and mimicking the aspect structure (http://cpp.sh/9smco).

I agree that this is probably the best solution to the problem, so I will merge it.

PR #3949, opened 2021-01-13
document timeout in test system

tjhei:
document timeout based on @bangerth 's comment in  https://github.com/geodynamics/aspect/pull/3944#issuecomment-759497805

related to #3605

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -307,7 +307,13 @@ FOREACH(_test ${_tests})
       MESSAGE(FATAL_ERROR "test ${_testname}.prm is missing a folder with reference data")
     ENDIF()
 
-    # run command
+    # Create the run command.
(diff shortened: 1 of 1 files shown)


MFraters (commented):
I think this text is clear and contains the information needed. One small comment, can you change that it is not available by default on OSX? I am not sure whether there would we any reason for normal users to use the prebuild option on OSX, but if that would be the case, I think it would be useful to also add a reference to how to make it available on OSX with for example the link you provided before (https://stackoverflow.com/questions/3504945/timeout-command-on-mac-os-x)

bangerth (approved):

tjhei:
> I am not sure whether there would we any reason for normal users to use the prebuild option on OSX

No, and I don't want to explain how to do this as it is difficult to use correctly unless you start with a clean build. As such, I also don't think it makes sense to explain this here.

PR #3953, opened 2021-01-18
fix description of convection timestep

tjhei:


gassmoeller (commented):
Looks reasonable if you fix the escape characters to make it compile.

gassmoeller (approved):

PR #3956, opened 2021-01-25
Add missing #include file.

bangerth:
This doesn't currently compile for me without the `#include` statement.

/rebuild


gassmoeller (approved):
Interesting, must be dependent on compiler or deal.II version. No harm in adding it.

PR #3960, opened 2021-01-26
manual: add GMG section

tjhei:
as discussed during the Q&A today.

### DIFF
--- a/doc/manual/cookbooks/overview/doc/gmg-average.part.prm
+++ b/doc/manual/cookbooks/overview/doc/gmg-average.part.prm
@@ -1,0 +1,3 @@
+subsection Material model
+  set Material averaging = harmonic average
+end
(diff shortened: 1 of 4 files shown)


tjhei:
@naliboff @gassmoeller FYI.

gassmoeller (approved):
Yes, thank you!

It would be great if we could at some point go through our cookbooks and just enable the GMG for all of them that can use it, but that is a separate PR.

tjhei:
> It would be great if we could at some point go through our cookbooks and just enable the GMG for all of them that can use it, but that is a separate PR.

At least some of them (or changing the default to GMG at some point). I would be hesitant to convert all of them, as it is not obvious for a beginner how to fix the situation if a small modification does not work with GMG (boundary condition, etc.).

PR #3961, opened 2021-01-28
Proposed change to multicomponent incompressible

bobmyhill:
The multicomponent incompressible equation of state was originally designed to have the following temperature dependence on the density:
rho = reference_density * ( 1 - thermal_expansivity * (temperature - reference_temperature))

In #3896, this was modified so that when adiabatic heating was turned on, the reference_temperature was replaced with the temperature on the reference adiabat at the local pressure (T(Sref, p)). This ensured zero compressibility along the isentrope, but meant that the equation of state was cryptically dependent on the adiabatic surface temperature. 

This PR explicitly removes the "Reference temperature" parameter. The densities now either correspond to those at the Adiabatic surface temperature (for models where adiabatic heating is turned off), or correspond to those on the reference adiabat.

[The first commit in this PR follows my original proposal, i.e.,
This PR changes the reference temperature to be equal to:
reference_temperature* = reference_temperature + T(Sref, p)) - T(Sref, psurface))

The equation of state is now functionally the same whether adiabatic heating is turned on or off. It does assume (inconsistently) that other isentropes are parallel to the principal isentrope, but there's no way around that.]

I've updated the tests and the documentation in multicomponent_incompressible.h.


bobmyhill:
After discussing this PR with @gassmoeller and @jdannberg, we decided to instead remove the Reference Temperature parameter. This broke a bunch of tests, a cookbook and some benchmarks. I've now (hopefully) fixed these without having to change any of the output files.

bobmyhill:
Spoke too soon. Lots of problems. Will try to fix tomorrow.

bobmyhill:
Well, I can't get this to work, and I've been staring at the code for hours.

@gassmoeller, do you want to have a look? Given how cryptic the proposed change is, I would personally favour reverting to my original proposal, which doesn't break any tests and should be obvious to people who read the docs. But if you can see what I've missed, I'll be happy to make the changes to the 50 broken tests.

bobmyhill:
Firstly, sorry for the spam. Secondly, I figured out the problem. I had forgotten that with this EoS, alpha isn't actually constant, so it isn't possible to just shift the reference density. 

@gassmoeller, I think the only option is to keep all the prms the same, and change all the test output. Do you agree?

include/aspect/material_model/equation_of_state/multicomponent_incompressible.h
```diff
@@ -46,20 +46,20 @@ namespace aspect
...
-       * Now there is no isentropic pressure-dependence of the density along the reference isentrope
-       * (i.e. the isentropic compressibility is equal to zero).
-       *
-       * This equation of state is pleasingly simple but not fully self-consistent,
-       * because the thermal expansivity divided by the pressure has a temperature dependence
-       * while the isobaric heat capacity is constant.
+       * This equation of state is pleasingly simple but not thermodynamically self-consistent,
+       * because the form of the density equation implies a non-constant thermal expansivity,
+       * and the thermal expansivity divided by the pressure has a temperature dependence
+       * which implies a non-constant isobaric heat capacity.
```
naliboff, line 62: Does this statement apply to basically all current mantle convection and lithospheric dynamics codes using the BA/EBA approximation? My understanding from reading through this and the related PRs (+ additional conversations) is that all codes these approximations suffer from the same thermodynamic inconsistency. If so, would it be worth pointing out that this is not just a problem specific to ASPECT?

doc/modules/changes/20210128_bobmyhill
```diff
@@ -0,0 +1,16 @@
...
+equation of state is now set to either the adiabatic surface temperature
+(if adiabatic heating is turned off), or the
+temperature along the reference adiabat at the local static pressure
+(when adiabatic heating is turned on).
+If users have previously used this model with adiabatic heating turned off
+and a user-defined reference temperature that is not equal to the
+adiabatic surface temperature, they should change the reference densities
+in their model according to the expression
+new_densities[i] = old_densities[i] * (1 - thermal_expansivities[i] *
+(adiabatic_surface_temperature - old_reference_temperature)). The results
```
naliboff, line 11: This is not done in the updated parameter files, but from the discussion it looks like there may be a reason for this? My first reaction is that many of the changed cookbooks, benchmarks, and tests should have the reference densities updated to get as close as possible to the original density profiles, but I admit I don't fully understand all of the related thermodynamic arguments.

naliboff (commented):
@bobmyhill - Thank you for bravely taking the plunge to remove the `Reference temperature` parameter! I have a few general questions, but these will likely need to be resolved with further group discussion. Thank you again for taking on this sorely needed change!

include/aspect/material_model/equation_of_state/multicomponent_incompressible.h
bobmyhill: Yes, of course, I'll do that. We could potentially make this EoS more consistent, but there was common consensus when we discussed this a few years ago that the benefits would be pretty minor (the inconsistency is more an annoyance than a serious modelling concern, and this is an *incompressible* model, which is never going to be very realistic).

doc/modules/changes/20210128_bobmyhill
bobmyhill: The reason I haven't done this is because a) I don't think any of the tests use real densities measured at the reference temperature and b) the test results would still be different.

I looked through the cookbooks and benchmarks, and it wasn't clear to me whether the authors cared about the exact density. @gassmoeller and @dannberg told me that a lot of users didn't understand what the parameter was, so I've opted for keeping simple round numbers. Happy to change this if the cookbooks/benchmarks require a specific density on the reference isentrope.

bobmyhill: I've also changed this description so that it no longer contains the equation. Users could calculate the density and thermal expansivity by extrapolation or direct measurement.

naliboff: I think we will definitely want to adjust the reference densities for the following cases:

- continental extension cookbook 
- Buiter et al., 2008, 2016 benchmarks
- Crameri et al. benchmark
- Slab detachment benchmark

All of these should be fairly straightforward changes to the reference density, except the continental extension benchmark will also require updates to the manual. It has been on my list to update that benchmark to make a number of other changes to this benchmark - shall I submit a separate PR for this in the next few days and then this PR can be rebased?

bobmyhill: Hi @naliboff,

I guess there are two questions here. 
* To which temperature was the reference density originally meant to correspond?
* How much do we care about new results being close (i.e. not identical) to the originals?

I've noticed that in a lot of prms, the adiabatic surface temperature is not changed from the default value (0 K). I don't know whether this is intentional, but even for crustal modelling on Earth that seems like an odd choice. So if these models are to be changed, maybe this parameter could also be modified?

gassmoeller (changes requested):
Thanks for doing this, I know that this is a lot of digging through models and figuring out what happened there. I am in favor of proceeding with this PR, but in order to do that we need to check the benchmarks, cookbooks and tests that have changed and see how to restore their original results. For quite a lot of them that probably means setting the Adiabatic surface temperature to the former value of the reference temperature of the material model, but for some of them the new results may actually more correct (i.e. `compositing_named_outputs` looks more correct now). Could you take a look? In case you do not have the time let us know, we can also make this a group effort.

doc/modules/changes/20210128_bobmyhill
gassmoeller: Sorry, just saw this discussion after my comment below. Yes, for many models the fact that adiabatic surface temperature was not set highlights the problem. Feel free to modify that parameter to the value of the reference temperature from the material model. That should fix a lot of the changes.

bobmyhill:
> in order to do that we need to check the benchmarks, cookbooks and tests that have changed and see how to restore their original results.

I don't think we can restore their original results *and* merge the current PR. The reason is that if we change the reference temperature we must change both the reference density *and* the reference thermal expansivity to match the original density curve. This is inconsistent with the formulation of the thermal expansivity (globally constant). The options are:

1. Change just the reference density (current PR), which breaks rho(T)
2. Change the density and the thermal expansivity, which fixes rho(T) but breaks alpha(T)
3. Change the adiabatic surface temperature to the old reference temperature. However, the adiabatic surface temperature seems to be used by other bits of the code even in BA (I tested this approach and got odd results).
4. Use my original proposal from the first commit in this PR, which is imperfect but doesn't break any tests.

@gassmoeller Do you think (3) could be made to work without changing the test output? Otherwise, if keeping the test results the same is the most important factor, (4) is the only option.

bobmyhill:
Closing this PR as we still don't have a good solution to the problem. 
Incompressible models will always be somewhat inconsistent with nature, and I'm unconvinced that the improvement I suggested here is worth the effort.

PR #3963, opened 2021-01-29
remove trailing whitespace, ensure single newline from source files

bobmyhill:
This PR adds two functions to contrib/utilities/indent, and applies those functions to all source files. These functions remove trailing whitespace and ensure a single newline at the end of the file (as per the POSIX standard).

The motivation for this is that atom installs the whitespace plugin and does this as standard. This PR means that users of atom (or other similar IDEs) don't need to turn off what is otherwise a useful tool in order to make small changes to ASPECT.


contrib/utilities/indent
```diff
@@ -95,4 +95,35 @@ tab_to_space()
...
+
+# Ensure only a single newline at end of files
+ensure_single_trailing_newline()
+{
+  f=$1
+
+  # Remove newlines at end of file
+  # then add a newline if none exists
+  sed -e :a -e '/^\n*$/{$d;N;};/\n$/ba' $f >$f.tmpi
+  sed -e '$a\' $f.tmpi >$f.tmp
```
bangerth, line 119: Is this a mistake? I have no idea what these two lines actually do (maybe add a bit of commentary), but the second one with the `\'` that apparently does not escape the `'` seems particularly peculiar...

bangerth (commented):
I like this idea a lot! I have just one question.

contrib/utilities/indent
bobmyhill: It's not a mistake, but I agree it looks peculiar. I've now added several explanatory lines.

bangerth: What a bizarre statement. I wouldn't even know how this is interpreted by the shell. I guess using `'...'` instead of `"..."` means that the shell doesn't interpret things...

bangerth (approved):

PR #3965, opened 2021-02-01
Small clean-ups in the VoF postprocessor.

bangerth:
Specifically, avoid a loop over an array of one element :-) I suspect that that is copy-pasted from elsewhere.

/rebuild

### DIFF
--- a/source/postprocess/volume_of_fluid_statistics.cc
+++ b/source/postprocess/volume_of_fluid_statistics.cc
@@ -39,7 +39,8 @@ namespace aspect
       const QGauss<dim> quadrature_formula (1);
       const unsigned int n_q_points = quadrature_formula.size();
 
-      unsigned int n_volume_of_fluid_fields = this->get_volume_of_fluid_handler().get_n_fields();
+      const unsigned int n_volume_of_fluid_fields
+        = this->get_volume_of_fluid_handler().get_n_fields();
 
       FEValues<dim> fe_values (this->get_mapping(),
                                this->get_fe(),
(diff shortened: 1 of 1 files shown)


source/postprocess/volume_of_fluid_statistics.cc
```diff
@@ -73,40 +75,34 @@ namespace aspect
...
         }
-
-      return std::pair<std::string, std::string> ("Global volume of fluid volumes (m^3):",
-                                                  output.str());
     }
+
+    return {"Global volume of fluid volumes (m^3):", output.str()};
   }
 }
+}
```
bobmyhill, line 105: Nice! Just need to remove this pesky bracket.

source/postprocess/volume_of_fluid_statistics.cc
```diff
@@ -73,40 +75,34 @@ namespace aspect
...
+          // show up with sufficient accuracy and in scientific notation
+          statistics.set_precision (col_name, 8);
+          statistics.set_scientific (col_name, true);
         }
-
-      return std::pair<std::string, std::string> ("Global volume of fluid volumes (m^3):",
-                                                  output.str());
     }
+
+    return {"Global volume of fluid volumes (m^3):", output.str()};
```
bobmyhill, line 102: In other places in the code, we use std::make_pair() to return a pair.

source/postprocess/volume_of_fluid_statistics.cc
bangerth: Oops...

source/postprocess/volume_of_fluid_statistics.cc
bangerth: True. We needed to, because this syntax is C++11 and so we haven't been able to use it before. 

Do you want me to go back to the old syntax, or allow me this language upgrade?

bobmyhill: No, the upgrade is fine! 
The line itself should go before the close bracket though ;-)

PR #3966, opened 2021-02-01
Better describe the years vs seconds issue.

bangerth:
/rebuild


bobmyhill:
Excellent, thanks! This is definitely a useful thing to add.

PR #3975, opened 2021-02-17
allow discontinuous particle interpolation for periodic BC

jdannberg:
### For all pull requests:

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -208,11 +208,17 @@ namespace aspect
     if (parameters.use_discontinuous_temperature_discretization ||
         parameters.use_discontinuous_composition_discretization)
       {
+        const bool no_field_method = std::find(parameters.compositional_field_methods.begin(),
+                                               parameters.compositional_field_methods.end(),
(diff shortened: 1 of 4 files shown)


gassmoeller (approved):
This makes sense, thanks for the fix. To summarize what is happening: If we do not solve an advection equation, but instead interpolate properties from particles then it does not matter that the DG elements currently do not correctly account for periodic boundaries, and we should allow using discontinuous elements.

tjhei:
Fascinating. I did not know that DG interpolation of particle properties works. Is there a good reason to use it or is this just a matter of the current limitation that all compositions are either DG or CG?

gassmoeller:
> Fascinating. I did not know that DG interpolation of particle properties works. Is there a good reason to use it or is this just a matter of the current limitation that all compositions are either DG or CG?

In fact all of the interpolation algorithms we have at the moment should use DG fields, because all of them create cell-wise approximations without consideration for neighbor cells. So ideally we would enforce this with an Assert. Unfortunately, there are applications that need continuous fields (like melt) and since all fields currently have to be the same type we can not enforce DG with particles. So #3978 would also help us to do that.

PR #3981, opened 2021-02-24
Allow particle initialization with Dirichlet boundary condition

gassmoeller:
Particles are often used to replace our compositional fields. When generating new particles to satisfy a minimum number of particles per cell we usually interpolate from the surrounding particles. However, for some particle properties (used to mirror compositional fields), it makes more sense to initialize them with the value of the boundary condition. This case will only be relevant for very few models, because we currently do not consistently implement boundary conditions for existing particles. 

This change will most likely only affect models where some material is flowing in through a boundary and new particles are generated there, and there are prescribed compositional boundary conditions there. For those models however, this change drastically increases the accuracy, because the current interpolation means that if a compositional anomaly exists close to the boundary, the newly generated particles (representing inflowing material) will also carry that anomaly, i.e. the boundary condition changes based on what is currently in that cell. @jdannberg FYI this should fix the model we talked about.


include/aspect/particle/property/initial_composition.h
```diff
@@ -55,6 +55,9 @@ namespace aspect
           initialize_one_particle_property (const Point<dim> &position,
                                             std::vector<double> &particle_properties) const override;
 
+          InitializationModeForLateParticles
+          late_initialization_mode () const override;
```
jdannberg, line 59: Also, does this function need a documentation?

source/particle/property/interface.cc
```diff
@@ -402,6 +404,61 @@ namespace aspect
...
+                  else
+                    {
+                      Assert(property_information.get_components_by_plugin_index(property_index) == this->n_compositional_fields(),
+                             ExcInternalError());
+
+                      const types::boundary_id boundary_id = cell->face(boundary_face)->boundary_id();
+
+                      for (unsigned int c=0; c<this->n_compositional_fields(); ++c)
+                        {
+                          const double composition = manager.boundary_composition(boundary_id,particle_location,c);
```
jdannberg, line 454: In principle, a composition boundary plugin could return different boundary compositions for different boundary ids. This function would only handle the first that is found. I don't know if this is actually done in any of the plugins, but if there's an easy fix for this you can think of, it may be worth implementing.

jdannberg (commented):
I think this is a nice fix! I know it changes some model results, but I see this more as a bugfix than as a change of existing functionality. So even if this is not the perfect solution, it is already much better than not including the boundary conditions at all. But it would be good to open an issue to improve how the boundary conditions are applied/where the new particles are generated.

gassmoeller:
I think I addressed your comments.

> I guess this would be the place where you could make it an option for the user to either use this or interpolate. But from our discussion, I actually think if there is no inflow, the model would simply not have Dirichlet boundary conditions and then it does not matter.

Agreed. I see no case at the moment where there is a boundary condition, yet the user would still want to interpolate particles rather than using the boundary condition.

> If I understand it correctly, this already would allow a user to change the value the particle is initialized to in dependence of its distance from the boundary.

This is correct, but the problem is that you would not know what to change the value to. As we discussed the only way I would see is to blend the boundary condition with the interpolated result in dependence of the distance to the boundary, but that seems overly complicated for a minor gain in accuracy. The correct solution (also as discussed, just for documentation) would be to modify the particle generation algorithm to only generate new particles in the part of the cell that moved in through the boundary and assign the boundary value to them, but that is a more complex change that would delay this fix.

> In principle, a composition boundary plugin could return different boundary compositions for different boundary ids. This function would only handle the first that is found. I don't know if this is actually done in any of the plugins, but if there's an easy fix for this you can think of, it may be worth implementing.

Some plugins (like box and ascii_data) do that already while others dont (function). This is a valid point, and I have implemented the following: A particle now checks for it's closest boundary with prescribed composition (by checking the distance to the face center) and uses the boundary condition of that boundary. That gives nice results in the new test case (check the bottom left corner, gray are initial particles, blue and red particles that were generated later according to the boundary conditions):

testcase


Let me know if there is anything else that needs changing.

source/particle/property/interface.cc
```diff
@@ -423,12 +423,17 @@ namespace aspect
...
                   bool cell_at_fixed_boundary = false;
                   unsigned int boundary_face = numbers::invalid_unsigned_int;
+                  double minimum_face_distance = std::numeric_limits<double>::max();
                   for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
                     if (cell->at_boundary(f) && fixed_boundaries.count(cell->face(f)->boundary_id()) == 1)
                       {
-                        boundary_face = f;
-                        cell_at_fixed_boundary = true;
-                        break;
+                        const double face_center_distance = particle_location.distance_square(cell->face(f)->center(true));
```
jdannberg, line 430: This won't work well for cells that have an aspect ratio that's different from 1, but I don't really see a better solution.

jdannberg (approved):
Nice! Great that you could fix the part about the boundary indicators. Can you open an issue for the particle generation algorithm (to only generate new particles in the part of the cell that moved in through the boundary and assign the boundary value to them)?

PR #3982, opened 2021-02-25
[WIP] Extend the viscoplastic rheology module for melt compatibility

jperryhouts:
This is far from complete, but it's part of an ongoing effort to merge the `melt_visco_plastic_*` tests into the visco plastic material model (or at least allow them to share a lot of code). Ultimately this will result in less duplicated code, but for now it means making the two as similar as possible without breaking the test cases. So far everything has been merged (sloppily) except about a third of the `evaluate` function, which needs more work.

The only change to Aspect is that the viscoplastic rheology module needs to be at least partially melt-aware. In particular, the function for calculating isostrain viscosity needs to account for the fluid pressure when evaluating stress. That turns out to be not such a big deal, since it already inherits from SimulatorAccess, and therefore has access to almost everything it needs as is.

### DIFF
--- a/include/aspect/material_model/rheology/visco_plastic.h
+++ b/include/aspect/material_model/rheology/visco_plastic.h
@@ -114,6 +114,8 @@ namespace aspect
                                             const std::vector<double> &volume_fractions,
                                             const std::vector<double> &phase_function_values = std::vector<double>(),
                                             const std::vector<unsigned int> &n_phases_per_composition =
+                                              std::vector<unsigned int>(),
+                                            const std::vector<unsigned int> &fluid_pressures =
                                               std::vector<unsigned int>()) const;
 
           /**
--- a/source/material_model/rheology/visco_plastic.cc
+++ b/source/material_model/rheology/visco_plastic.cc
@@ -24,9 +24,12 @@
 #include <aspect/newton.h>
 #include <aspect/adiabatic_conditions/interface.h>
 #include <aspect/gravity_model/interface.h>
+#include <aspect/melt.h>
+
 #include <deal.II/base/signaling_nan.h>
 #include <deal.II/base/parameter_handler.h>
 #include <deal.II/fe/fe_values.h>
+
 
 namespace aspect
 {
@@ -98,7 +101,8 @@ namespace aspect
                                        const unsigned int i,
                                        const std::vector<double> &volume_fractions,
                                        const std::vector<double> &phase_function_values,
-                                       const std::vector<unsigned int> &n_phases_per_composition) const
+                                       const std::vector<unsigned int> &n_phases_per_composition,
+                                       const std::vector<unsigned int> &fluid_pressures) const
       {
         IsostrainViscosities output_parameters;
 
@@ -128,6 +132,19 @@ namespace aspect
           edot_ii = std::max(std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),
                              min_strain_rate);
 
+        // In case porosity lies above the melt transport threshold
+        // P_effective = P_bulk - P_f = (1-porosity) * P_s + porosity * P_f - P_f = (1-porosity) * (P_s - P_f)
+        // otherwise,
+        // P_effective = P_bulk, which equals P_solid (which is given by in.pressure[i])
+        double effective_pressure = in.pressure[i], porosity = 0;
+
+        if (this->include_melt_transport() && this->get_melt_handler().is_melt_cell(in.current_cell))
+          {
+            const unsigned int pf_idx = this->introspection().compositional_index_for_name("porosity");
+            porosity = std::min(1.0, std::max(in.composition[i][pf_idx],0.0));
+            effective_pressure = (1. - porosity) * (in.pressure[i] - fluid_pressures[i]);
+          }
+
         // Calculate viscosities for each of the individual compositional phases
         for (unsigned int j=0; j < volume_fractions.size(); ++j)
           {
@@ -136,7 +153,7 @@ namespace aspect
             // Choice of activation volume depends on whether there is an adiabatic temperature
             // gradient used when calculating the viscosity. This allows the same activation volume
             // to be used in incompressible and compressible models.
-            const double temperature_for_viscosity = in.temperature[i] + adiabatic_temperature_gradient_for_viscosity*in.pressure[i];
+            const double temperature_for_viscosity = in.temperature[i] + adiabatic_temperature_gradient_for_viscosity*effective_pressure;
             AssertThrow(temperature_for_viscosity != 0, ExcMessage(
                           "The temperature used in the calculation of the visco-plastic rheology is zero. "
                           "This is not allowed, because this value is used to divide through. It is probably "
@@ -147,12 +164,12 @@ namespace aspect
                           + Utilities::to_string(in.pressure[i]) + ")."));
 
             // Step 1a: compute viscosity from diffusion creep law
-            const double viscosity_diffusion = diffusion_creep.compute_viscosity(in.pressure[i], temperature_for_viscosity, j,
+            const double viscosity_diffusion = diffusion_creep.compute_viscosity(effective_pressure, temperature_for_viscosity, j,
                                                                                  phase_function_values,
                                                                                  n_phases_per_composition);
 
             // Step 1b: compute viscosity from dislocation creep law
-            const double viscosity_dislocation = dislocation_creep.compute_viscosity(edot_ii, in.pressure[i], temperature_for_viscosity, j,
+            const double viscosity_dislocation = dislocation_creep.compute_viscosity(edot_ii, effective_pressure, temperature_for_viscosity, j,
                                                                                      phase_function_values,
                                                                                      n_phases_per_composition);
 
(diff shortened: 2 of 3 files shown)


include/aspect/material_model/rheology/visco_plastic.h
```diff
@@ -114,6 +114,8 @@ namespace aspect
                                             const std::vector<double> &volume_fractions,
                                             const std::vector<double> &phase_function_values = std::vector<double>(),
                                             const std::vector<unsigned int> &n_phases_per_composition =
+                                              std::vector<unsigned int>(),
```
naliboff, line 117: Indentation off?

include/aspect/material_model/rheology/visco_plastic.h
```diff
@@ -114,6 +114,8 @@ namespace aspect
...
+                                            const std::vector<unsigned int> &fluid_pressures =
```
naliboff, line 118: Why a vector? Is there more than one type of fluid pressure we may want to pass over? Or will this variable also be used to store the porosity?

source/material_model/rheology/visco_plastic.cc
```diff
@@ -128,6 +132,19 @@ namespace aspect
           edot_ii = std::max(std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),
                              min_strain_rate);
 
+        // In case porosity lies above the melt transport threshold
+        // P_effective = P_bulk - P_f = (1-porosity) * P_s + porosity * P_f - P_f = (1-porosity) * (P_s - P_f)
+        // otherwise,
+        // P_effective = P_bulk, which equals P_solid (which is given by in.pressure[i])
+        double effective_pressure = in.pressure[i], porosity = 0;
```
naliboff, line 139: Equation needs to be corrected?

```suggestion
        double effective_pressure = in.pressure[i];
```

source/material_model/rheology/visco_plastic.cc
```diff
@@ -128,6 +132,19 @@ namespace aspect
           edot_ii = std::max(std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),
                              min_strain_rate);
 
+        // In case porosity lies above the melt transport threshold
```
naliboff, line 135: Eventually the documentation should be updated to indicate that the visco_plastic rheology model itself does not handle melting, but may be used by a material model that includes melt transport (or something to that effect).

source/material_model/rheology/visco_plastic.cc
```diff
@@ -136,7 +153,7 @@ namespace aspect
             // Choice of activation volume depends on whether there is an adiabatic temperature
             // gradient used when calculating the viscosity. This allows the same activation volume
             // to be used in incompressible and compressible models.
-            const double temperature_for_viscosity = in.temperature[i] + adiabatic_temperature_gradient_for_viscosity*in.pressure[i];
+            const double temperature_for_viscosity = in.temperature[i] + adiabatic_temperature_gradient_for_viscosity*effective_pressure;
```
naliboff, line 156: Hmm, I'm not if it is correct to use the effective pressure (i.e., when fluid is present) for this term or other terms below aside from the plasticity. @jdannberg - do the melt material models only use a the solid pressure for the viscous flow laws?

source/material_model/rheology/visco_plastic.cc
```diff
@@ -147,12 +164,12 @@ namespace aspect
                           + Utilities::to_string(in.pressure[i]) + ")."));
 
             // Step 1a: compute viscosity from diffusion creep law
-            const double viscosity_diffusion = diffusion_creep.compute_viscosity(in.pressure[i], temperature_for_viscosity, j,
+            const double viscosity_diffusion = diffusion_creep.compute_viscosity(effective_pressure, temperature_for_viscosity, j,
```
naliboff, line 167: Same comment above applies here and to other viscous flow law terms below.

tests/melt_visco_plastic.cc
```diff
@@ -227,11 +189,11 @@ namespace aspect
          * MaterialModelOutputs object that is handed over, if it exists.
          * Does nothing otherwise.
          */
-        void fill_plastic_outputs (const unsigned int point_index,
-                                   const std::vector<double> &volume_fractions,
-                                   const bool plastic_yielding,
-                                   const MaterialModel::MaterialModelInputs<dim> &in,
-                                   MaterialModel::MaterialModelOutputs<dim> &out) const;
+        // void fill_plastic_outputs (const unsigned int point_index,
```
naliboff, line 192: Is there a compiler issue here with having both melt outputs and the plastic outputs? I recall running into something similar before.

source/material_model/rheology/visco_plastic.cc
```diff
@@ -98,7 +101,8 @@ namespace aspect
                                        const unsigned int i,
                                        const std::vector<double> &volume_fractions,
                                        const std::vector<double> &phase_function_values,
-                                       const std::vector<unsigned int> &n_phases_per_composition) const
+                                       const std::vector<unsigned int> &n_phases_per_composition,
+                                       const std::vector<unsigned int> &fluid_pressures) const
```
naliboff, line 105: To start with and for testing purposes, should we just call this new term `additional_pressures`?

source/material_model/rheology/visco_plastic.cc
```diff
@@ -128,6 +132,19 @@ namespace aspect
           edot_ii = std::max(std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[i])))),
                              min_strain_rate);
 
+        // In case porosity lies above the melt transport threshold
+        // P_effective = P_bulk - P_f = (1-porosity) * P_s + porosity * P_f - P_f = (1-porosity) * (P_s - P_f)
+        // otherwise,
+        // P_effective = P_bulk, which equals P_solid (which is given by in.pressure[i])
+        double effective_pressure = in.pressure[i], porosity = 0;
+
+        if (this->include_melt_transport() && this->get_melt_handler().is_melt_cell(in.current_cell))
```
naliboff, line 141: I wonder for now if we want to actually make this new functionality dependent on whether melt transport is included? For example, could we start with a general case where an additional pressure (of any origin) is passed to the visco plastic rheology model? This would make testing easier. However, maybe it makes sense to begin with having melt accounted for as well?

naliboff (commented):
@jperryhouts - Thanks for taking an initial stab at this! A few preliminary comments throughout.

One question/suggestion: would it be possible to make a new simple test case that checks whether new pressure term is correctly passed and used to the visco_plastic rheology module? Could this be done without modifying the melt_visco_plastic material model or having any connection to a model with melt? 

I'm wondering if we should calculate the effective pressure outside of the visco_plastic rheology module, so that is has no dependence and internal logic based on whether melt is used.

Cheers,
John

jperryhouts:
I've removed the changes to the rheology module, and just left the material model as closely aligned with ViscoPlastic as possible. I think this is actually the best available option since we're not ready to change anything about the physics in the rheology module at this point.

Following up on conversations earlier with @naliboff  and @jdannberg , both about how best to align this test with the rheology module system:

I don't think there's much advantage to trying to reduce the code sharing between this test and `ViscoPlastic`. If anything, this is going to be the most robust way to ensure that this test is sensitive to any changes that would also affect `ViscoPlastic`. I've tried both approaches we discussed to make this test actually depend on the `ViscoPlastic` class, and neither one makes sense. Ultimately, the two classes just don't have as much overlap as it seems. The rheology module system abstracts away a lot of the important bits. These are what I've found tinkering with other approaches:

**Approach 1:** Inherit from ViscoPlastic

* Inheriting from ViscoPlastic creates compile-time ambiguity between `ViscoPlastic` and `MeltInterface`, both of which extend `MaterialModel::Interface`. It might be possible to resolve that ambiguity with some sort of `using ViscoPlastic ...` stuff, but that gets messy because it's still important that this class does in fact inherit from `MeltInterface` in most places.
* The functions that ViscoPlastic actually implement are almost all virtual functions, which means that `MeltViscoPlastic` will need to override those anyway.
* Inheritance would only really be useful to track a few state variables that are currently duplicated. Those variables are all `private` members in the `ViscoPlastic` class, so all of them would either need getter/setter functions, or they would need to be declared as `protected` rather than `private`.

**Approach 2:** Own a ViscoPlastic object

* There are only three functions that a `ViscoPlastic` object could help with: `is_compressible`, `reference_viscosity`, and `create_additional_named_outputs`. The first two are only one line of code, and the last one is three lines of code.
* If we used the object to store otherwise duplicate member variables, we'd have the same problem as above-- they're currently all `private`. We would need to implement getter/setters, or make them all `public` in this case.

-----------------

Takeaway is that there is no coherent way to tie this test case more closely to the `ViscoPlastic` class itself. Rather, replicating `ViscoPlastic` as closely as possible is going to be the most effective way to keep this material model compatible with Aspect. This test is currently sensitive to changes in any of the classes and interfaces that `ViscoPlastic` relies on, including its rheology module (only for its member variables at this point).

Since we're definitely not at a stage to make decisions about changing the physics in ViscoPlastic, it makes most sense to avoid using `Rheology::ViscoPlastic::calculate_isostrain_viscosities` altogether, but still utilize all the other parts of the rheology module. That way nothing needs to change in Aspect, and this PR only updates the test case.

PR #3984, opened 2021-02-26
Adding 2 publications Barrioneuvo and Bredow

ljhwang:
Add two publication to the citations list: Barrionuevo et al. 2021 and Bredow and Steinberger 2021.


gassmoeller (approved):
Thank you, looks good to me. I will wait for the tester and then merge.

PR #3988, opened 2021-03-04
[WIP] add composite rheology to visco plastic

naliboff:
This PR enables use of the recently added `composite_visco_plastic` rheology modules within the` visco_plastic `rheology module. 

The PR is very a much at the early stages (WIP), and has only been opened for the purposes of public discussion on various design and implementation choices.

A list of items to sort out before the PR is merged:
- [ ] Should we have a large if-else statement that decides whether the current or iterative composite viscosity calculation is used? At present, I have an if statement at the end of the viscosity computation that replaces the normally calculated viscosities with the iterative composite viscosities, if the iterative composite option is selected by the user.
- [ ] How to handle the strain rheology (partly ties into question above)
- [ ] Special pressure terms that go into plasticity
- [ ] What benchmark(s) and tests to use?
- [ ] Should we add text to the manual covering the composite rheology and all other viscoelastic-plastic rheology options before this is merged?
- [ ] Elasticity in the composite rheology formulation - this likely needs to be handled in a different PR.

PR Requirements:


source/material_model/rheology/visco_plastic.cc
```diff
@@ -294,6 +294,16 @@ namespace aspect
 
             // Step 5: limit the viscosity with specified minimum and maximum bounds
             output_parameters.composition_viscosities[j] = std::min(std::max(viscosity_yield, min_visc), max_visc);
+
+            // Step 6: If the composite rheological formulation is selected, recalculate the viscosity
+            std::vector<double> partial_strain_rates(5, 0.);
+            if (use_composite_rheology)
```
naliboff, line 300: @bobmyhill - Should we have a much larger if-else statement that skips over steps 1-5 if this option is enabled?

bobmyhill: Yes, I think the methods should be entirely separate; that way we can implement improvements/fix bugs in either one without affecting the other.

naliboff: @bobmyhill - After a bit of thought, one disadvantage to the latter approach (large if-else statement) is that the compositing could still be used for the creep mechanisms in combination with plasticity.

In this scenario, the iterative procedure could be used instead for steps 1a-c to calculate the bulk viscosity for all of the creep mechanisms. If the viscous stress is greater than the yield stress, then the viscosity is rescaled to the yield stress and the strain-rate is entirely taken up by plasticity.

However, maybe this is will lead excessive confusion about how the compositing is being used? It would also require quite a few internal checks for how parameters are being used in combination with each other. 

Perhaps best to keep things simple for now per your recommendation (i.e., one large if-else statement)? We can always add more complexity later on.

bobmyhill: I think keeping things as simple as possible is a good idea. I understand your point about using composite as composition creep only, but I don't think that will give the user any additional benefits, and as you say, it'll be harder to understand/explain how the compositing and various stabilization schemes work together.

source/material_model/rheology/composite_visco_plastic.cc
```diff
@@ -123,6 +125,8 @@ namespace aspect
                                                                  const unsigned int composition,
                                                                  const SymmetricTensor<2,dim> &strain_rate,
                                                                  std::vector<double> &partial_strain_rates,
+                                                                 const double current_cohesion,
```
anne-glerum, line 128: Now that the drucker_prager_parameters are included in composite_visco_plastic, there is no need to pass around current_cohesion and current_friction_angle.

anne-glerum: Here the 2 lines can be removed.

PR #3992, opened 2021-03-14
Initialize particles in one go.

bangerth:
We currently initialize particles one cell at a time, but because they are stored per cell, this amounts to initializing them all in the same order in which they are stored. We might as well initialize them all at once.

/rebuild

### DIFF
--- a/source/particle/world.cc
+++ b/source/particle/world.cc
@@ -723,6 +723,8 @@ namespace aspect
                                        this->get_timestep());
     }
 
+
+
     template <int dim>
     void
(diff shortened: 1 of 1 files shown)


MFraters (approved):
This simplifies the code, making it more readable. I don't see any downside to it and the tests still pass. But I think it might be good if @gassmoeller also takes a quick look at it.

bangerth:
Ping? @gassmoeller ?

gassmoeller (approved):
Yes, this makes sense and thanks for the cleanup. Sorry this took a while.

PR #3995, opened 2021-03-16
Change computation of strain rate invariant in visualization postprocessor.

MFraters:
This pull request changes the way the second invariant of the strain rate tensor is computed in the visualization. 

The current way of computing the strain-rate invariant (`sqrt(strain_rate_deviator*strain_rate_deviator)`) is different from how it is commonly computed in the rest of the code (`sqrt(fabs(second_invariant(deviator(strain_rate)))`). See for example:
https://github.com/geodynamics/aspect/blob/09587597bce79bc648f56c8fca04fb6092751617/source/material_model/drucker_prager.cc#L78  
https://github.com/geodynamics/aspect/blob/fed300b4b4ea048f73418f26e37b6f72373261d9/source/particle/property/viscoplastic_strain_invariants.cc#L130
https://github.com/geodynamics/aspect/blob/e509ffe927bac6c0fa3e2265cfcbb185a3a92092/source/material_model/rheology/visco_plastic.cc#L128
https://github.com/geodynamics/aspect/blob/4efd1bf4ddd2e9730896aca019693d29bf57799c/source/particle/property/integrated_strain_invariant.cc#L62
The issue is that the result is different. For example, if we use a strain-rate tensor of:

equation

The currently implementation returns sqrt(0.5), while the way it is implemented in the rest of the code returns 0.5, which seems to me the desired value for this tensor. 

An other option would be to use `sqrt(0.5*strain_rate_deviator*strain_rate_deviator)`, which is how I usually see it being computed in papers, but I think this approach makes sure the same methods  of computing the second invariant is used throughout aspect.


gassmoeller (approved):
I am in favor of having a function that does this consistently, maybe in material_model/utilities.cc, but for now this is already progress. 

The fix is appreciated! :+1:

PR #4002, opened 2021-03-17
renamed compositional stress fields

EstherHeck:
Following yesterdays aspect meeting, I changed all compositional stress fields from stress_xx etc to ve_stress_xx etc. We agreed that this makes sense, because at the moment the stress postprocessor uses the same field names as the compositional fields on which the elastic stresses are stored. This led to overwriting and misleading output. Like this it is hopefully more clear what output comes from the compositional stress fields and what output is generated by a postprocessor. 

### For all pull requests:


MFraters (approved):
This is indeed what I remember that has been discussed during the aspect user meeting. Since it has been open for a few days now, I will merge tomorrow if there are no further comments. 

Thanks for making the changes @EstherHeck!

PR #4006, opened 2021-03-19
Fix spherical gaussian perturbation initial temperature

anne-glerum:
Currently the `spherical gaussian perturbation` initial temperature plugin fails with the message that there are no fixed temperature boundary indicators set, even though they are. This is because the initial temperature parameters are parsed before those of the boundary temperature in core.cc. As there was no test with this initial temperature condition, this was not noticed. I moved the AssertThrow to the initial_temperature() function, where it's also placed for the `spherical hexagonal perturbation` plugin. I also added a test that uses the gaussian perturbation.

### For all pull requests:


gassmoeller (approved):
Thanks for the cleanup, the fix, and the test! :+1:

PR #4008, opened 2021-03-23
[WIP] log verbosity

tjhei:
Add a parameter "set Output verbosity" to specify how much information should be printed to the screen.

This is WIP (quiet doesn't do anything yet, more things need to be updated).

### DIFF
--- a/include/aspect/parameters.h
+++ b/include/aspect/parameters.h
@@ -373,6 +373,43 @@ namespace aspect
       }
     };
 
+    struct OutputVerbosity
+    {
+      enum Kind
+      {
+        quiet,
+        normal,
+        /**
+        detailed: also output TimerOutput data
+        */
+        detailed,
+        maximum
+      };
+
+      static const std::string pattern()
+      {
+        return "quiet|normal|detailed|maximum";
+      }
+
+      static Kind
+      parse(const std::string &input)
+      {
+        if (input == "quiet")
+          return quiet;
+        else if (input == "normal")
+          return normal;
+        else if (input == "detailed")
+          return detailed;
+        else if (input == "maximum")
+          return maximum;
+        else
+          AssertThrow(false, ExcNotImplemented());
+
+        return Kind();
+      }
+
+    };
+
     /**
      * Constructor. Fills the values of member functions from the given
      * parameter object.
@@ -456,6 +493,7 @@ namespace aspect
     double                         surface_pressure;
     double                         adiabatic_surface_temperature;
     unsigned int                   timing_output_frequency;
+    typename OutputVerbosity::Kind output_verbosity;
     unsigned int                   max_nonlinear_iterations;
     unsigned int                   max_nonlinear_iterations_in_prerefinement;
     bool                           use_operator_splitting;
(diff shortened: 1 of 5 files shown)


tjhei:
Looking for comments about this, before I take this further. Good or bad idea?

Motivation:
Sometimes you don't care about a lot of information (think convection_box.prm where tons of stuff scrolls by) when at other times you do (think large, instantaneous model).

bobmyhill:
I think it's a good idea. The only thing I wonder about is how you decide what to add at different OutputVerbosity levels, and how the user will know which they should choose. Is there a clear motivation for each? 

I could see the following levels being useful:
- Near silent (Who doesn't want a quiet life?)
- Vital checks (Is ASPECT slow at a certain task? Will this run finish before the sun becomes a red giant?)
- Debug (ASPECT, what exactly do you think you're doing? Don't make me send you to your room.)

gassmoeller:
Sorry for not reacting. I like the general idea. However, one thing I would not want is to prefix every normal output line with `if (parameters.output_verbosity >= Parameters<dim>::OutputVerbosity::normal)`. Do you plan that 'quiet' simply implies deactivating `pcout` to remove all output? Also I am wondering how many levels we realistically need. Would a single 'detailed' not be sufficient? We could add additional levels if we ever need them.

tjhei:
> I would not want is to prefix every normal output line with

Agreed. But I don't think we need to do that outside of 

> Do you plan that 'quiet' simply implies deactivating `pcout` to remove all output?

No, I was just thinking to reduce some of the information printed in each timestep. This is not important to me at this point, though. 

I have a counter proposal: adding a few additional bool flags.

bangerth:
I think I'm not quite sure where your starting point is. Do you think that we're currently outputting too much information and you'd like a way to reduce that? Or that we don't output enough and you'd like a way to show more?

Many years ago I had a program that had a logger class that looks like `pcout` but knew something about "priorities". You'd do something like
```
  pcout << set_priority(12);
```
at the top of the function and then whatever you wrote into `pcout` only ended up on the screen if `pcout`'s threshold was set to 12 or above. Everything over the current threshold was just silently dropped. This way, one didn't have to annotate every output statement individually with a priority (or an `if` statement). 

On the other hand, everything was always output into a log file, with each line prefixed by the priority.

tjhei:
> Do you think that we're currently outputting too much information and you'd like a way to reduce that? Or that we don't output enough and you'd like a way to show more?

I have the following situations in mind:
1. Print timing for each section once completed: This is useful for very slow instantaneous models where some additional text doesn't bother me. For an example, see https://github.com/geodynamics/aspect/pull/4008#issuecomment-804978761
2. Print additional GMG solver info: I would like to optionally print additional details that help in determining solver performance (GMG levels, coarse grid, etc.)
3. solver residuals: For very big and slow models it would be useful to see individual Stokes iterations with current residual (as written to the history file)
4. (this is less interesting for me, but came to my mind) a "quiet" mode: convection_box.prm is slowed down a lot by the amount of text scrolling by. Wouldn't it be neat if only a line or two is printed to the screen for each time step?

Maybe a verbosity level is not the right choice. I can also add individual bool flags like "set Output solver residuals = true" etc..  One could put them where they fit (solver sections for example) or make a global "Output details" section.
Is that better than a verbosity level? Maybe...

PR #4009, opened 2021-03-23
Make sure visualization names are unique

### LINKED ISSUE #3996: Check for duplication of visualization output variable names
Compositional fields can be named by the user. Currently, when a field name is identical to an already existing visualization output name, the output will be overwritten in release mode. In debug mode, an exception is thrown. It would be good if there is always a check for duplication and potentially the field names get a pre/post fix to avoid problems.

gassmoeller:
Fixes #3996. The task is slightly more complicated than I thought at first, because each postprocessor can rightly return multiple copies of the same name if it returns a vector or tensor. Thus each postprocessors names are first copied into a set to make them unique, and then this set is merged with the existing set of names, while checking for duplicates.

### DIFF
--- a/tests/composition_names_duplicates.prm
+++ b/tests/composition_names_duplicates.prm
@@ -1,0 +1,109 @@
+# A test to ensure the naming of compositional fields works as expected.
+
+# EXPECT FAILURE
+
+set Adiabatic surface temperature          = 1623
+set CFL number                             = 1.0
+set Dimension                              = 2
+set End time                               = 0
+
+set Pressure normalization                 = surface
+set Surface pressure                       = 0
+set Resume computation                     = false
+set Start time                             = 0
+
+set Use years in output instead of seconds = true
+
+subsection Compositional fields
+  set Number of fields = 2
+  set Names of fields = peridotite, density
+end
+
+
+subsection Boundary temperature model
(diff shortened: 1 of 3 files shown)


tests/composition_names_duplicates.prm
```diff
@@ -0,0 +1,109 @@
+# A test to ensure the naming of compositional fields works as expected.
+
+# EXPECT FAILURE
```
bobmyhill, line 3: You should make motivational posters ;-)

bobmyhill (approved):
Looks good to me. Is there anything else you want to do to this?

tjhei (approved):

tjhei:
Can you fix indentation?

tests/composition_names_duplicates.prm
gassmoeller: Assume the worst then you will always be positively surprised :-)
But in this case it is a macro that tells the tester to pass if the test fails, and fail if the test passes.

gassmoeller:
27 days for one `make indent`. But it is done :-).

tests/composition_names_duplicates.prm
tjhei: I take credit for this brilliant feature including the string. :-)

PR #4011, opened 2021-03-31
Elasticity stabilization parameter 

EstherHeck:
This is the new version of PR #3770. 

### For all pull requests:


EstherHeck:
should I rebase with master?
and squash all the commits?

tests/elastic_stabilization_time_scale_factor/screen-output
```diff
@@ -0,0 +1,80 @@
+-----------------------------------------------------------------------------
```
bobmyhill, line 1: Could you strip out the lines that start with `--`, `+` and `|` from this file?
Alternatively, if you have tested this prm using ctest, you can copy the file
 `aspect/build/tests/output-<test name>/screen-output.notime` into 
 `aspect/build/tests/<test name>/screen-output`.

bobmyhill (changes requested):
Excellent, thanks! Just one change to the test output, than I'll set the tester running. 
If you want to squash some of your commits to make the history a bit tidier, then go ahead, but I wont insist on it.

tests/elastic_stabilization_time_scale_factor/screen-output
EstherHeck: done :)

bobmyhill (approved):

gassmoeller (approved):
Ok, I think we discussed this issue sufficiently (sorry for missing the past 4 months of the discussion or so :-)), let's go forward with this. If necessary we can adjust the default later (or remove the fixed timestep option at some point? @EstherHeck what is your thought on that?).

Also thanks @EstherHeck  for the incredible amount of testing and patience with this change!

EstherHeck:
Great, that this is finally merged! And yes, I personally don't see a reason for the fixed time step option, but I am also running quite specific models...

PR #4015, opened 2021-04-11
Included constraints for interpolate output into advection field.

alarshi:
While running one of my models which sets the temperature to a prescribed ASCII data file, I noticed that the boundary conditions for the temperature were also overwritten by the prescribed temperatures. This PR ensures that the constraints are set in the solution while using the function, ```interpolate_material_output_into_advection_field.```

### For all pull requests:

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -1839,6 +1839,7 @@ namespace aspect
     // updating the ghost elements of the 'solution' vector.
     const unsigned int advection_block = adv_field.block_index(introspection);
     distributed_vector.block(advection_block).compress(VectorOperation::insert);
+    current_constraints.distribute (solution);
     solution.block(advection_block) = distributed_vector.block(advection_block);
   }
 


source/simulator/helper_functions.cc
```diff
@@ -1839,6 +1839,7 @@ namespace aspect
     // updating the ghost elements of the 'solution' vector.
     const unsigned int advection_block = adv_field.block_index(introspection);
     distributed_vector.block(advection_block).compress(VectorOperation::insert);
+    current_constraints.distribute (solution);
```
tjhei, line 1842: That doesn't seem right. Either apply to distributed_vector or do this after the last line.

alarshi: Thank you, Timo! I corrected ir in the latest commit.

gassmoeller:
@alarshi can you update the test results as we discussed by downloading the reference results?

alarshi:
Hi @gassmoeller! Thank you for the reminder. It had completely slipped my mind. I just pushed the changes in the test results.

PR #4017, opened 2021-04-13
delay exit with failure in parallel

tjhei:
When an exception (like solver failure) is triggered, all but rank 0
immeditately return in main() without printing anything. Some MPI
implementations detect abnormal termination and abort the whole job.
This could happen before rank 0 produces an error message.
Avoid this by sleeping before exiting.

### DIFF
--- a/source/main.cc
+++ b/source/main.cc
@@ -861,6 +861,11 @@ int main (int argc, char *argv[])
       // root when we already know that processor 0 will generate
       // an exception. We do this to avoid creating too much
       // (duplicate) screen output.
+
+      // Sleep a few seconds before aborting. This allows text output from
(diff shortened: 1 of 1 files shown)


tjhei:
@alarshi @gassmoeller do you want to test this before merging? I have not tried to reproduce this issue, but I would guess that this patch helps.

alarshi:
Hello @tjhei: I tried this branch with a model that would give an error in the solver convergence. However, I am still getting the MPI_ALLreduce errors. Hereis the complete error message:
[test.2047344.log](https://github.com/geodynamics/aspect/files/6324149/test.2047344.log)

bangerth (approved):
I think this is reasonable.

PR #4024, opened 2021-04-30
Remove unnecessary particle backups

### LINKED ISSUE #4019: ParticleHandler is  backed up twice
The ParticleHandler is backed up in start_timestep(), but when an iterative Advection scheme is used, they are again backed up at the start of the iterations, to be restored after each Advection solve. This second backup can be removed.

gassmoeller:
This should fix #4019. @anne-glerum do you have a test case? It seems we have no test at the moment for an iterated solver scheme and particles.


anne-glerum:
Great! I have a test case somewhere, shall I make a separate PR for it?

gassmoeller:
You should have permissions to checkout this branch and directly push to it (because you are maintainer and this is an open PR). So feel free to directly add it to this PR (or you can also make a PR against my branch).

bangerth (commented):
Given that it passes the tests, this looks reasonable to me.

gassmoeller:
I would like to wait for a test, because I do not think we are currently testing the iterated schemes with particles.

anne-glerum:
OK, I've added a test and checked that for higher resolution it gives the same results as for the single Advection, iterated Stokes scheme and as for a field instead of particles. 

I had to adapt the test a little bit as compared to the single Advection scheme test that is already there,  because the sudden high velocity prescribed threw off the temperature solver in the second nonlinear iteration.

 As a side-note, we have been using the iterated Advection with particles in large setups as well.

gassmoeller:
Excellent, thanks! Could you replace the output files of the test with gnupot output to make them readable? Otherwise ready.

gassmoeller:
Thanks! Can someone else approve and merge? Since I opened the PR I can not approve.

PR #4027, opened 2021-05-05
Fix probability density particle generator

gassmoeller:
As described in the changelog, this is a minor fix for an edge case of the random uniform and probability density function particle generators.


bangerth (approved):
Good catch!

PR #4032, opened 2021-05-12
Work around removed exception

gassmoeller:
This is required to make mainline compile again after the changes in dealii/dealii#11276.

### DIFF
--- a/source/particle/generator/interface.cc
+++ b/source/particle/generator/interface.cc
@@ -61,6 +61,20 @@ namespace aspect
         // Try to find the cell of the given position. If the position is not
         // in the domain on the local process, throw a ExcParticlePointNotInDomain
         // exception.
+#if DEAL_II_VERSION_GTE(9,3,0)
+        std::pair<const typename parallel::distributed::Triangulation<dim>::active_cell_iterator,
+            Point<dim> > it =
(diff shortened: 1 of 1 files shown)


source/particle/generator/interface.cc
```diff
@@ -61,6 +61,20 @@ namespace aspect
         // Try to find the cell of the given position. If the position is not
         // in the domain on the local process, throw a ExcParticlePointNotInDomain
         // exception.
+#if DEAL_II_VERSION_GTE(9,3,0)
+        std::pair<const typename parallel::distributed::Triangulation<dim>::active_cell_iterator,
+            Point<dim> > it =
+              GridTools::find_active_cell_around_point<> (this->get_mapping(), this->get_triangulation(), position);
```
tjhei, line 67: Is the cell guaranteed to be locally owned if valid?

gassmoeller: I do not know, that is why I kept the additional assert below. I guess the function could return a ghost cell.

tjhei: Sorry, I didn't scroll to the right in the assert.

tjhei (approved):

PR #4036, opened 2021-05-15
Add a fallback interpolator to bilinear least squares

gassmoeller:
@kchotalia: @jdannberg told me you had still problems with the bilinear least squares interpolator running out of particles to interpolate from. Please give this branch a try, it should make your problem nearly impossible to trigger. If this works this is a solution that I we should apply to other particle interpolators as well. I will update this PR if this works.


kchotalia:
@gassmoeller this is working nicely! Models are running again

bangerth (approved):

PR #4038, opened 2021-05-17
update inner_core_traction param file

Yann-Musseau:

The inner core traction param file was not really usable (3D too expensive to run; problem with rotation, etc.)
We propose a new param file, changing some parameters to have a set-up that can be run easily on a laptop and provide nice translation instability. The different modifications are listed below:
- We  switched the model from 3D to 2D.
- We prevented the system from rotation (null space removal)
- We increased the Rayleigh number in order to have translation.
- We also changed the mesh refinement (thermal energy density refinement added) and the end time of the run to agree with the Rayleigh number.

### For all pull requests:

### DIFF
--- a/cookbooks/inner_core_convection/inner_core_traction.prm
+++ b/cookbooks/inner_core_convection/inner_core_traction.prm
@@ -11,11 +11,14 @@
 
 set Additional shared libraries            = ./libinner_core_convection.so
 
-set Dimension                              = 3
+set Dimension                              = 2
 set Use years in output instead of seconds = false
-set End time                               = 1.5e9
-set Output directory                       = output-inner_core_traction
+set End time                               = 39.275
+set Output directory                       = output_inner_core_traction_Ra_1e2_P_1e-2
 
+subsection Nullspace removal
+  set Remove nullspace = net rotation
+end
 # The equations are non-dimensionalized, and all of the
 # material properties are constant and set to one, except
 # for the density, which scales with temperature.
@@ -45,7 +48,7 @@ subsection Material model
 # and the last variable will be the time
   subsection Inner core
     subsection Phase change resistance function
-      set Variable names      = x,y,z
+      set Variable names      = x,y
       set Function expression = 1e-2     # <-- P
     end
   end
@@ -103,16 +106,17 @@ subsection Gravity model
 
   subsection Radial linear
     set Magnitude at bottom = 0.0
-    set Magnitude at surface = 2     # <-- Ra
+    set Magnitude at surface = 1e2     # <-- Ra
   end
 end
 
 
 subsection Mesh refinement
   set Initial global refinement          = 3
-  set Initial adaptive refinement        = 0
-  set Strategy                           = minimum refinement function
-  set Time steps between mesh refinement = 0
+  set Initial adaptive refinement        = 2
+  set Strategy                           = minimum refinement function, thermal energy density
+
+  set Time steps between mesh refinement = 15
 
   subsection Minimum refinement function
     set Variable names = depth, phi, theta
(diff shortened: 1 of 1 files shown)


gassmoeller (commented):
Hi @Yann-Musseau and thanks for this contribution. Can you clarify if this will require an update to the section of the manual (I think it is 5.3.10 Inner core convection)? If this requires an update to the text or figures, can you make the necessary changes in `doc/manual/manual.tex`? Also I would like to ask @jdannberg to take a look at this as she wrote the original example.

MarineLasbleis:
Hi @gassmoeller , yes, indeed, we should have changed the manual, since it will (very slightly) change the example of parameter file in the manual. 
I was wondering if @jdannberg has an opinion about having a param file in 2D? 
(I think the null space removal is required anyway, since it would also mess with the results in 3D, but the other changes are all related to the switch from 3D to 2D). 
I think the 2D is easier: it allows for a a fast computation, while keeping the overall dynamics (translation and plume convection transitions are at the same Rayleigh number and P number, even though the computation is in "cylinder" and not a "sphere" in 2D). 
This change was motivated by the difficulty to run the original param file on a desktop computer by @Yann-Musseau :-) We thought a param file easier to run would make it more accessible for other people as well.

bobmyhill:
I'd like both the 2D and 3D versions in the codebase. I think there's value to showing both, and it seems wasteful to remove @jdannberg's work. If the 3D version needs rotation-removal, that would be a valuable addition too.

MarineLasbleis:
The change to 3D to 2D was only: change the Dimension and the variables for the phase change setup (x,y,z to x, y) and @Yann-Musseau choose a mesh that would be fine enough to solve the problem but not too much either to be calculated fast. So one option would be to  write comments on how to change the parameter file to compute 3D? (or either come back to 3D and add comments to switch to 2D, but I am guessing one may not want to go further into the param file if they think the calculation is too long? )
The main discussion I had with researchers on IC dynamics was that ASPECT would be too expensive numerically compared to the spectral code they usually use. So showing that a convection/translation setting can be done in 5 min on a couple of cores (albeit not exactly the proper geometry) was a good thing for me :-)

jdannberg:
I think it would be great to have an example in 2D that can be run on a laptop!

I also think it would be nice to still have the option to run the 3D model, so what I would suggest is the following: Adding a separate parameter file for the 2D version that is based on the original file: The way this works is that you can include another prm file in a prm file. To do that, you only need to add a line at the start saying 

`include $ASPECT_SOURCE_DIR/cookbooks/inner_core_convection/inner_core_traction.prm`

(An example for that is given in this test: https://github.com/geodynamics/aspect/blob/master/tests/continental_extension.prm). 
Then you only need to add the lines you want to change from the original input file. This way, whenever there is a change done to the original file (like a bugfix), it will automatically carry over to the new version, but the user can still decide to run either a 2D or a 3D version. 

I also agree it will be useful to update the manual accordingly, as @gassmoeller suggested. And if the Nullspace removal is a general fix that would also be needed in 3d, then I agree this would be useful to change in the original 3d parameter file rather than the copy.

tjhei:
Wouldn't it be easy enough to have a 2d and a 3d prm file in the folder? It is fine if the manual only shows the 2d one. That would be my preference.

Yann-Musseau:
Hi, thank you all for your comment and advice. 
I have created another file .prm for the 2D version of the model as @jdannberg proposed with the line at the beginning and only the parameters that change between 3D and 2D. I also add the nullspace removal for the 3D model and write a few lines paragraph to put in the manual in order to explain quickly the file 2D. 

Let me know if you prefer to create a clean pull request. It's my first one so I don't really know how to do it properly.

MarineLasbleis:
Question for ASPECT developers: do you now why, even while "including" the main parameter file, we still need to specify the additional library? 
set Additional shared libraries            = ./libinner_core_convection.so

It does not work without this line in the (2D) parameter file.

Thanks
Marine

tjhei:
Marine, yes. This is not documented anywhere, but dimension and shared libraries are special and need to be in the main file.
This is because this information is needed very early and we parse it directly without honoring includes.

PR #4039, opened 2021-05-17
Add Dockerfile for TACC systems Stampede Frontera

gassmoeller:
The TACC systems Stampede2 and Frontera support running singularity containers. This is a preliminary Docker image that uses their base images. They run successfully in parallel on Frontera across nodes. I need to do some more testing regarding performance and extend the documentation, but I think this is something we may want to keep around to make installation on Stampede and Frontera trivial (you can put the lines in instructions.txt in any slurm script and it will download and run ASPECT without local installation).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -86,7 +86,7 @@ jobs:
           cache-to: type=inline
           platforms: linux/amd64
           push: true
-          tags: geodynamics/aspect_tacc:latest
+          tags: geodynamics/aspect:latest-tacc
 
       - name: Build and push Docker image for release
         if: contains(github.event_name, 'release')
@@ -97,4 +97,4 @@ jobs:
           cache-to: type=inline
           platforms: linux/amd64
           push: true
-          tags: geodynamics/aspect_tacc:${{github.ref_name}}
+          tags: geodynamics/aspect:${{github.ref_name}}-tacc
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -70,6 +70,12 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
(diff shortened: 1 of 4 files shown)


gassmoeller:
This is now ready for a review. I have verified that the generated image runs on Frontera. I probably need to add more documentation if we want to advertise this more widely, but I thought I ask for feedback now in case someone wants to try it already.

.github/workflows/docker.yml
```diff
@@ -57,3 +57,50 @@ jobs:
...
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push Docker image for main
+        if: contains(github.event_name, 'push')
```
tjhei, line 87: What does this do? only react to pushes? what about merges?

tjhei: shouldn't this be only on the main branch?

tjhei (approved):

.github/workflows/docker.yml
gassmoeller: The 'push' event triggers whenever commits are pushed to a branch. A merge counts as a push. There is no separate merge event (not clearly specified, this is all the docs say: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push). We ensure only pushes to 'main' are considered by specifying a branch in the 'on:push' section at the top of the workflow.  This logic is the same as the one further up in the file that creates our regular (non-TACC) docker images.
The main reason I need this 'if' here is to separate pushes to main from releases to give different tags to the created images.

I am not 100% sure this workflow will work right away, because I dont know if the login to ghcr.io will work. However, I think the best way to test it is to try it. The worst things that can happen is that we have one workflow that fails (it will not show up on PRs, because it is not triggered by PRs).

tjhei: sure.

PR #4041, opened 2021-05-19
Limit spherical harmonic degree to stable range

gassmoeller:
The implementation of spherical harmonics in the boost library is limited to a degree of 120 or so after which the output becomes inaccurate. This is actually already documented for this utility function, but all the places that use it happily accept larger values. We should prevent this. Please do not merge just yet, I am waiting for feedback from @wanyingw on what degree is still working well, we should maybe be conservative and limit ourselves to that.

### DIFF
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -920,6 +920,14 @@ namespace aspect
                                                       const double theta,
                                                       const double phi)
     {
+      Assert (l<65, ExcMessage("ASPECT uses the implementation of spherical harmonics provided by the "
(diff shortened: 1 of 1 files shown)


bangerth (approved):

bangerth:
OK to merge if you can get confirmation. I suspect that you get into trouble substantially before degree 120.

gassmoeller:
I have not heard back about what a safe value is, so I have chosen a more conservative value for now that is still high enough to cover most global seismic tomography models. Ok to merge?

bangerth (approved):

PR #4042, opened 2021-05-20
fix lateral viscosity variations

jdannberg:
The viscosity profile in the Steinberger material model has a radial and a lateral component, both read in through files. The file for the lateral variations doesn't seem to agree with what is in the Steinberger & Calderwood (2006) paper. First, it has weird jumps around 660 km depth. Second, it seems to assume n=3 rather than n=3.5 for the upper mantle. Since the original data are included in the data folder, all I had to do to produce a new viscosity profile was to write a python script that converts the data from this original format to the one ASPECT uses, and I have included this script here too. The original script also must have interpolated the values somehow (from 1 km to 3 km resolution), but I kept the original resolution here. The figures below show the old and the new profiles for H/nR (this is what the temp-viscosity-prefactor.txt file contains). 

plot
plot2

This change will of course affect model results, but I would consider this a bugfix since it makes the model consistent with the original publication the viscosity is based on. 

For the change in n, I would like to hear from @gassmoeller if that was intended or happened by accident. 

Finally, do we need a changelog entry for this?

### For all pull requests:


tjhei (approved):
Can you add a changelog entry please?

jdannberg:
Alright, I updated the tests and added a changelog entry. 

@gassmoeller when we talked today you mentioned that you thought n=3 was appropriate, but I went through the paper, and as far as I can tell, they use n=3.5 (In section 2.1, at the end, they write "As reference case will use eq. (7) with n=3.5 above 660 km and n=1 below." From what I could tell, n is not a parameter they vary, so all models should use this reference value. Could you check and tell me if/why you still think it should be n=3?

gassmoeller (approved):
Looks good to me. Checking the history I have no clue where the 3.0 came from. I have a vague recollection that 3.0 is an approximate average of dislocation creep (3.5) and diffusion creep (1.0) weighted by their relative contribution to flow in the upper mantle. However, I have no source where this comes from, and I found an email from Bernhard in 2010 where he specifically said this profile is for n=3.5. Sorry for messing that up, and thanks for fixing bugs that are a decade old!

PR #4045, opened 2021-05-26
add dependency info to readme

### LINKED ISSUE #4031: cmake documentation?
I'm finding installation instructions using `candi` which I gather installs all sorts of stuff I already have. Is there documentation of the cmake installation directly?

tjhei:
closes #4031

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,11 @@ problems during the installation, please consult our
 problems or specific instructions for MacOS users, before asking your question
 on the mailing list.
 
-
+In short, ASPECT is configured using CMake and has the following requirements:
+- CMake 2.8.12
+- GCC, Clang, or Intel compiler with C++11 support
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,11 @@ problems during the installation, please consult our
 problems or specific instructions for MacOS users, before asking your question
 on the mailing list.
 
-
+In short, ASPECT is configured using CMake and has the following requirements:
+- CMake 2.8.12
+- GCC, Clang, or Intel compiler with C++11 support
+- [deal.II](https://github.com/dealii/dealii) 9.2 or newer configured with MPI, Trilinos, p4est (optionally HDF5)
```
gassmoeller, line 32: Maybe add DEAL_II_WITH_LAPACK as well? It is needed for the GMG and the bilinear least squares particle interpolation as far as I know and I had to explain to some users that they need to recompile deal.II because they compiled without it.

Apart from that looks good.

tjhei: good point. updated.

bangerth (approved):

PR #4047, opened 2021-06-01
Avoid the use of Threads::Thread.

bangerth:
Use std::thread instead since the deal.II class is deprecated.

/rebuild


MFraters (approved):
Looks good to me. This is also a good reminder for me that there should 3 empty lines after a function (https://www.dealii.org/developer/doxygen/deal.II/CodingConventions.html) :)

tjhei (approved):

PR #4050, opened 2021-06-01
Avoid the use of deprecated functions for iterating over local vector elements.

bangerth:
We use the same construct several times over to compute the inverse of the diagonal of a matrix in matrix-free contexts. Replace this by what I think is the equivalent construct.

/rebuild

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -885,16 +885,21 @@ namespace aspect
 
     this->set_constrained_entries_to_one(diagonal);
     inverse_diagonal = diagonal;
-    const unsigned int local_size = inverse_diagonal.local_size();
-    for (unsigned int i=0; i<local_size; ++i)
+
+    // Finally loop over all of the computed diagonal elements and invert them.
+    // The following loop relies on the fact that inverse_diagonal.begin()/end()
+    // iterates only over the *locally owned* elements of the vector in which
+    // we store inverse_diagonal.
+    for (auto &local_element : inverse_diagonal)
       {
-        Assert(inverse_diagonal.local_element(i) > 0.,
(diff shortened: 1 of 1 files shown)


bangerth:
Hm, this does not seem to work, but I can't see where I'm going wrong. 

@tjhei Do you see anything that might be to blame?

source/simulator/stokes_matrix_free.cc
```diff
@@ -885,17 +885,23 @@ namespace aspect
...
     this->set_constrained_entries_to_one(diagonal);
     inverse_diagonal = diagonal;
-    const unsigned int local_size = inverse_diagonal.local_size();
-    for (unsigned int i=0; i<local_size; ++i)
+
+    // Finally loop over all of the computed diagonal elements and invert them.
+    // The following loop relies on the fact that inverse_diagonal.begin()/end()
+    // iterates only over the *locally owned* elements of the vector in which
+    // we store inverse_diagonal.
+    for (auto local_element : inverse_diagonal)
```
gassmoeller, line 893: Even though C++11 is old by now, I am still unsure about how to use `auto` correctly in some situations. But does this need to be an `auto &local_element` to make the assignment below actually change anything?

bangerth: Good question. Whatever this does, it seems to result in an lvalue that I can assign to a couple of lines below. But it is questionable for sure, so let's try the thing with the reference :-)

bangerth:
That seems to have worked! Thanks, @gassmoeller , for the suggestion!

gassmoeller:
Glad it worked. I found this: https://stackoverflow.com/questions/8542873/c11-auto-semantics.
`auto` always creates a copy, `auto &` a reference, but there are some special cases (none of which apply here). Anyway, this works now.

PR #4051, opened 2021-06-01
Shrink StructuredData.

bangerth:
The `StructuredData` class stores a number of member variables that are duplicates of what is actually stored further down in the objects that do the interpolation. This patch removes them so that we don't have to store duplicates.

The most noteworthy part is that that also removes a couple of accessor functions that, as it turns out, are not actually used anywhere. As a consequence, I think that it's relatively harmless to get rid of these.

/rebuild

### DIFF
--- a/include/aspect/structured_data.h
+++ b/include/aspect/structured_data.h
@@ -138,28 +138,6 @@ namespace aspect
         get_column_names() const;
 
         /**
-         * Returns whether the stored coordinates are equidistant. If
-         * coordinates are equidistant the lookup is more efficient. Returns
-         * false if no coordinates are loaded at the moment.
-         */
-        bool
-        has_equidistant_coordinates() const;
-
-        /**
-         * Returns the coordinates at which data is stored. This function
-         * can be used to determine the number of data points, or to query
-         * data only at exactly the positions at which it is available (avoiding
-         * interpolation).
-         *
-         * @param dimension The spatial direction for which to return the data
-         * coordinates, e.g. 0 for x-direction, 1 for y-direction, or equivalent
-         * values if your data coordinates are other dimensions such as
-         * temperature, pressure.
-         */
-        const std::vector<double> &
-        get_coordinates(const unsigned int dimension) const;
-
-        /**
          * Returns the column index of a column with the given name
          * @p column_name. Throws an exception if no such
          * column exists or no names were provided in the file.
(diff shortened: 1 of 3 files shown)


bangerth:
Wouldn't you know -- there is always a test that uses something you want to remove.

I don't particularly see a reason why the function I'm removing should actually exist, so in the place where the removed function is used, I'm simply hard-coding what the function would have returned. In this particular case, the plugin in the test wanted to know the interpolation points of the depth profile, which I tend to think is not something we should expose: What matters is that we have a depth profile, but how exactly that is parameterized should simply not be of anyone's concern.

tjhei:
> but how exactly that is parameterized should simply not be of anyone's concern.

I think there was a specific reason for exposing that and it is used in a plugin as well. Maybe @gassmoeller remembers.

include/aspect/structured_data.h
```diff
@@ -621,15 +584,6 @@ namespace aspect
...
         get_column_names() const;
 
-        /**
-        * Returns the coordinates at which profile data is stored. This function
-        * can be used to determine the number of data points, or to query
-        * data only at exactly the positions at which it is available (avoiding
-        * interpolation).
-        */
-        const std::vector<double> &
-        get_coordinates() const;
```
bangerth, line 631: ...whereas this one, in `AsciiDataProfile<dim>` that happens to have the same name, doesn't actually take an argument. The documentation doesn't say so, but it actually returns the interpolation points *in coordinate direction zero*. That's just confusing, and rather than document the issue, I would have renamed the function. But because I see no reason to expose this information to begin with, I think removing it is the best avenue.

bangerth:
Ah, let's see what happens. Is this a plugin in someone else's code, or in a cookbook?

gassmoeller:
This is a function that we need for the following purpose: We read in a structured data profile from disk that contains a viscosity reference profile for the Earth (based on observations in some form). We then want to compute the laterally averaged viscosity in ASPECT (based on a different theoretical parametrization) and compare that to the read in reference profile, and if they differ we modify the ASPECT profile to *on average* match the reference profile in a certain depth range. However in order to avoid aliasing between the two profiles we want to make sure to compute the laterally averaged property in ASPECT in exactly the same depth bounds as the file we read in. 

I do not think it is an usual use case to ask an interpolatory function to ask at which locations the values are known exactly, this could also be of interest to someone interested in the resolution of any particular region. I agree the interface is a bit awkward, would it be better if the function returned an `std::array<std::vector<double>, dim>`?

bangerth:
Hm. In an ideal world, the `InterpolatedUniformGridData` and `InterpolatedTensorProductGridData` classes would make this sort of information available given that they store it internally. But they don't, and whatever I add to them now doesn't provide me with a backward compatible way that keeps your plugin functioning. I might just have to go back to duplicating the information.

Would it be ok to change the interface of the latter function I annotated above so that it takes an argument indicating the dimension, rather than assuming it is the first coordinate direction?

bangerth:
Separately, could I rename `get_coordinates(direction)` to `get_interpolation_points(direction)`? The function is not used anywhere in ASPECT itself, but apparently only in your own plugin. I don't want to gratuitously break your code, so defer to your judgment to how difficult it would be for you to deal with an incompatible change.

bangerth:
I think we agreed to close this.

gassmoeller:
Ah I never replied to your questions. I am fine with updating our plugin in whatever way is best, as long as we can still access this information. Whether that is through the deal.II function or duplicating the information in aspect does not matter for us.

PR #4052, opened 2021-06-01
[WIP] Add github test action

gassmoeller:
Github now allows more runtime on their hosted runners than before, so we may be able to use them for our tester. It still needs some fixes, because some tests timeout at the moment, I will take a look at them when I get back to this.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -68,3 +68,39 @@ jobs:
       with:
         name: manual.log
         path: doc/manual/manual.log
+  job:
+    #linux build including indent and documentation
+    name: tests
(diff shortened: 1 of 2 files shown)


MFraters:
Very cool! Seeing that the prebuild tests takes more than 2 hours, it might be a good idea to look if they can be cached (https://github.com/marketplace/actions/cache). Using `ctest -j 2 ...` might also help, although looking at the output, it looks like the tester is mainly slow because 8  tests are extremely slow, while the rest is are about as fast as on the Jenskins testers.

bangerth:
There is a merge conflict now.

Issue #2982, opened 2019-05-23
Extract and visualize the surface mesh


anne-glerum:
Extract the surface mesh from the volume mesh and output it to vtk for visualizing things like the heat flux on the free surface, as discussed with @bangerth

anne-glerum:
@bangerth Out of curiosity, have you given this any more thought?

anne-glerum:
After talking to @bangerth, it might be worth finishing these PRs:
* #4061
* #4054
The discussion in #4054 concerns the duplication of code, i.e., each volume mesh postprocessor will need to be copied for the corresponding surface mesh postprocessor. Instead of waiting on a restructuring of the visualization interface, we could merge the PRs in the mean time and remove them again after restructuring.

bangerth:
@anne-glerum Tag me in these PRs once you think they're ready for review.

anne-glerum:
As both PRs mentioned above are now merged (ie base variables, strain rate and stress can all be outputted on a surface mesh), I'm closing this issue.

PR #4054, opened 2021-06-04
Add surface strainrate postprocessor

anne-glerum:
This PR adds a postprocessor that outputs the strain rate tensor on the surface of the mesh. 

### For all pull requests:


anne-glerum:
I added a test in 3D, and there I noticed some differences between the surface solution and the full mesh solution. However, similar differences occur for the existing surface stress postprocessor. I'll open an issue for it. 
Screenshot 2021-06-07 at 10 43 20

gassmoeller:
I will reply separately to your issue, but about the postprocessor: I have not looked at the code yet, but it is probably fine, because it likely is just a copy from the strain rate postprocessor? However, this is also what concerns me. With our current way of deriving surface postprocessors from a different base class we will eventually end up duplicating a lot of our postprocessors. Maybe we instead should have introduced a new function into the visualization postprocessor interface named `bool restrict_output_to_surfaces()` or similar, default it to false, and let some postprocessors set it to true based on an input parameter? Then we would not need to duplicate a lot of code. The only downside would be that we could not out surface and volume stress at the same time and the name of the output would likely be the same ("stress" no matter if it is just surface or everywhere). I think @bangerth implemented the interface, do you remember if there was a reason we need a different base class?
At the moment we only have two surface postprocessors so maybe it is not too late to change the interface? @anne-glerum What do you think about it? If it is important to you to get this in first we could also do that and fix the interface later (it will just slightly increase the effort to do so).

anne-glerum:
I agree the code duplication is not ideal. On the other hand, in our use-case, we will need both the volume and surface strain rate for example, so it wouldn't do to restrict output to one or the other. 

I don't mind taking the time to change the interface. We're using the plugin as is for now, and updating the interface won't change the workflow.

gassmoeller:
How many other postprocessors do you think we need to duplicate? I saw the base variables one in #4061. I think originally we had expected to only need very few postprocessors on the surface, but it seems we may have been wrong?

anne-glerum:
For our current purposes, the base variables + strain rate + stress are enough. I don't know about other people's plans though.

anne-glerum:
@bangerth , as discussed in #2982, this pull request is now updated and ready for another look. I'll squash the commits when you and the testers are happy. 

I've checked whether the volume and surface mesh output agree, and they do for those components that are of significant magnitude, see the figures below. 

surface_stress_strainrate_3d_strainrate_0
surface_stress_strainrate_3d_strainrate_1

include/aspect/postprocess/visualization/surface_strain_rate_tensor.h
```diff
@@ -0,0 +1,69 @@
...
+    {
+      /**
+       * A class derived from DataPostprocessor that takes an output vector
+       * and computes a variable that represents the 4 or 9
+       * components (in 2d and 3d, respectively) of the strain rate tensor at every
+       * point of the surface of the domain.
+       * The strain rate is defined as $\varepsilon(\mathbf u)
+       * - \tfrac 13 \textrm{trace}\ \varepsilon(\mathbf u) \mathbf 1 =
+       * \varepsilon(\mathbf u) - \frac 13 (\nabla \cdot \mathbf u)
+       * \mathbf I$.  The second term is zero if the
```
bangerth, line 45: This comment is not quite correct (but I know that you copied it from the `strain_rate_tensor.h` file). The correct description is in the `.cc` file in the doc string at the bottom. 

Would you mind copying the description from the .cc file here (and while you're there, also making the same change in `strain_rate_tensor.h`)?

bangerth (commented):
Just one small comment. Then please squash and tell me again.

anne-glerum:
@bangerth Updated and squashed :)

bangerth (approved):
Thank you!

PR #4060, opened 2021-06-14
Particle evaluation optimizations

gassmoeller:
With the new particle abilities in deal.II there is a lot that can be optimized in ASPECT's particle implementation. This is a first stab. The implementation is far from clean right now, I mostly open this to remind myself to finish it during the hackathon.

Major new changes:
- integrators should use particle properties instead of storing their own cached properties, making transfer and storage much more efficient. But these are temporary properties we usually do not want to appear in the output. I need to find a way to hide certain particle properties
- advection and update can be much more efficient using the FEPointEvaluation class instead of manually evaluating shape functions, or using FEValues

Missing features at the moment:
- boxes without mesh deformation do not benefit from FEPointEvaluation, because MappingCartesian is not implemented for that, only MappingQGeneric; boxes with mesh deformation may benefit, because they use MappingEulerian (not checked)
- particles that should be advected with the melt velocity fall back to the old implementation at the moment (no technical problem, just not implemented to evaluate melt velocity right now)
- rk2 integrator is faster, but needs the 'position' particle property as first property to store its temperorary data in (currently hard-coded); euler and rk4 are unchanged

### DIFF
--- a/source/particle/world.cc
+++ b/source/particle/world.cc
@@ -651,17 +651,20 @@ namespace aspect
                            solution_values.begin(),
                            solution_values.end());
 
-      evaluators.reinit (cell, positions, {solution_values.data(),solution_values.size()}, update_flags);
+      if (((update_flags & update_values) == true) || ((update_values & update_gradients) == true))
+        evaluators.reinit(cell, positions, {solution_values.data(), solution_values.size()}, update_flags);
 
       auto particle = begin_particle;
       for (unsigned int i = 0; particle!=end_particle; ++particle,++i)
         {
+          // Evaluate the solution, but only if it is requested in the update_flags
           const Vector<double> solution = (update_flags & update_values)
                                           ?
                                           evaluators.get_solution(i)
                                           :
                                           Vector<double>();
 
+          // Evaluate the gradients, but only if they are requested in the update_flags
           const std::vector<Tensor<1,dim>> gradients = (update_flags & update_gradients)
                                                        ?
                                                        evaluators.get_gradients(i)
@@ -989,19 +992,38 @@ namespace aspect
 #if DEAL_II_VERSION_GTE(9,3,0)
     namespace internal
     {
+      // This class evaluates the solution vector at arbitrary positions inside a cell.
+      // It uses the deal.II class FEPointEvaluation to do this efficiently. Because
+      // FEPointEvaluation only supports a single finite element, but ASPECT uses a FESystem with
+      // many components, this class creates several FEPointEvaluation objects that are used for
+      // the individual finite elements of our solution (pressure, velocity, temperature, and
+      // all other optional variables).
       template <int dim>
       class SolutionEvaluators
       {
         public:
+          // Constructor. Create the member variables given a simulator and a set of
+          // update flags. The update flags control if only the solution or also the gradients
+          // should be evaluated.
           SolutionEvaluators(const SimulatorAccess<dim> &simulator,
                              const UpdateFlags update_flags)
             :
-            velocity(simulator.get_mapping(),simulator.get_fe(), update_flags,simulator.introspection().component_indices.velocities[0]),
-            pressure(simulator.get_mapping(),simulator.get_fe(), update_flags,simulator.introspection().component_indices.pressure),
-            temperature(simulator.get_mapping(),simulator.get_fe(), update_flags,simulator.introspection().component_indices.temperature),
+            velocity(simulator.get_mapping(),
+                     simulator.get_fe(),
+                     update_flags,
+                     simulator.introspection().component_indices.velocities[0]),
+            pressure(simulator.get_mapping(),
+                     simulator.get_fe(),
+                     update_flags,
+                     simulator.introspection().component_indices.pressure),
+            temperature(simulator.get_mapping(),
+                        simulator.get_fe(),
+                        update_flags,
+                        simulator.introspection().component_indices.temperature),
             melt_component_indices(),
             simulator_access(simulator)
           {
+            // Create the evaluators for all compositional fields that exist
             const unsigned int n_compositional_fields = simulator_access.n_compositional_fields();
             const auto &component_indices = simulator_access.introspection().component_indices.compositional_fields;
             for (unsigned int composition = 0; composition < n_compositional_fields; ++composition)
@@ -1010,8 +1032,10 @@ namespace aspect
                                                                   update_flags,
                                                                   component_indices[composition]));
 
+            // Create the melt evaluators, but only if we use melt transport in the model
             if (simulator_access.include_melt_transport())
               {
+                // Store the melt component indices to avoid repeated string lookups later on
                 melt_component_indices[0] = simulator_access.introspection().variable("fluid velocity").first_component_index;
                 melt_component_indices[1] = simulator_access.introspection().variable("fluid pressure").first_component_index;
                 melt_component_indices[2] = simulator_access.introspection().variable("compaction pressure").first_component_index;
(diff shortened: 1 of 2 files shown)


gassmoeller:
An idea about the difference in the best case (time_dependent_annulus benchmark in a shell with rk2, gradient evaluation for properties, single core, 12000 particles). On master:
```
| Particles: Advect                |        44 |     0.187s |       3.4% |
| Particles: Copy                  |        22 |   0.00775s |      0.14% |
| Particles: Generate              |         1 |   0.00698s |      0.13% |
| Particles: Initialization        |         1 |  9.26e-05s |         0% |
| Particles: Initialize properties |         1 |   0.00403s |         0% |
| Particles: Interpolate           |        22 |     0.211s |       3.8% |
| Particles: Sort                  |        44 |    0.0956s |       1.7% |
| Particles: Update properties     |        22 |     0.719s |        13% |
```
On this branch:
```
| Particles: Advect                |        44 |    0.0656s |       1.4% |
| Particles: Copy                  |        22 |   0.00771s |      0.17% |
| Particles: Generate              |         1 |   0.00697s |      0.15% |
| Particles: Initialization        |         1 |  9.47e-05s |         0% |
| Particles: Initialize properties |         1 |   0.00397s |         0% |
| Particles: Interpolate           |        22 |     0.214s |       4.6% |
| Particles: Sort                  |        44 |    0.0966s |       2.1% |
| Particles: Update properties     |        22 |     0.119s |       2.6% |
```

source/particle/world.cc
```diff
@@ -852,15 +970,199 @@ namespace aspect
...
+        public:
+          SolutionEvaluators(const SimulatorAccess<dim> &simulator,
+                             const UpdateFlags update_flags)
+            :
+            velocity(simulator.get_mapping(),simulator.get_fe(), update_flags,simulator.introspection().component_indices.velocities[0]),
+            pressure(simulator.get_mapping(),simulator.get_fe(), update_flags,simulator.introspection().component_indices.pressure),
+            temperature(simulator.get_mapping(),simulator.get_fe(), update_flags,simulator.introspection().component_indices.temperature),
+            melt_component_indices(),
+            simulator_access(simulator)
+          {
```
bangerth, line 1004: Can you move this and the other member functions of this class out of line as we usually do?

source/particle/world.cc
```diff
@@ -852,15 +970,199 @@ namespace aspect
...
         }
     }
 
 
 
+#if DEAL_II_VERSION_GTE(9,3,0)
+    namespace internal
+    {
+      template <int dim>
+      class SolutionEvaluators
```
bangerth, line 993: This class and its members have no documentation, and there is also no documentation within its member functions. Would you mind adding some?

source/particle/world.cc
```diff
@@ -852,15 +970,199 @@ namespace aspect
...
+
+            Assert ((update_flags & ~(update_gradients | update_values)) == false,
+                    ExcNotImplemented());
+
+            velocity.reinit (cell, positions);
+            pressure.reinit (cell, positions);
+            temperature.reinit (cell, positions);
+
+            for (auto &evaluator_composition: compositions)
+              evaluator_composition.reinit (cell, positions);
```
bangerth, line 1057: Are these expensive calls? If so, it might actually be nice to run all of these `reinit()` calls in parallel using `Threads::new_task()`, but maybe that's overkill. It also only helps with serial runs.

Not necessary to change, just trying to plant a seed :-)

source/particle/world.cc
```diff
@@ -852,15 +970,199 @@ namespace aspect
...
+            if (simulator_access.include_melt_transport())
+              {
+                fluid_velocity->evaluate (solution_values, evaluation_flags);
+                fluid_pressure->evaluate (solution_values, evaluation_flags);
+                compaction_pressure->evaluate (solution_values, evaluation_flags);
+              }
+          }
+
+          Vector<double>
+          get_solution(const unsigned int i)
```
bangerth, line 1082: can you give `i` a better name?

source/particle/world.cc
```diff
@@ -852,15 +970,199 @@ namespace aspect
...
+                values_at_point[melt_component_indices[2]] = compaction_pressure->get_value(i);
+              }
+
+            return values_at_point;
+          }
+
+
+
+          std::vector<Tensor<1,dim>>
+                                  get_gradients(const unsigned int i)
```
bangerth, line 1116: This is funny indentation. Just put it on one line.

source/particle/world.cc
bangerth: Separately, this class always evaluates *all* solution components. Do we have a way to state which components we really need? If not, that's ok, just wondering. (Might also not be relevant because the run time doesn't matter, but I don't really know.)

source/particle/world.cc
bangerth: Same comments about `i`.

source/particle/world.cc
bangerth: I think `evaluation_point` is probably reasonable

bangerth (commented):
Very nice work!

source/particle/world.cc
```diff
@@ -620,6 +625,55 @@ namespace aspect
...
+                                       const typename ParticleHandler<dim>::particle_iterator &begin_particle,
+                                       const typename ParticleHandler<dim>::particle_iterator &end_particle,
+                                       internal::SolutionEvaluators<dim> &evaluators)
+    {
+      const unsigned int n_particles_in_cell = std::distance(begin_particle,end_particle);
+
+      std::vector<Point<dim> > positions;
+      positions.reserve(n_particles_in_cell);
+
+      for (auto particle = begin_particle; particle!=end_particle; ++particle)
```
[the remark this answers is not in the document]
gassmoeller: Once we require deal.II 10.0 I can make this much nicer, by just handing over a particle_iterator_range, but the necessary function in 9.3 is too slow. So yes, this will happen.

source/particle/world.cc
gassmoeller: I am not sure what you mean by out of line. I ran indent and broke up the long lines of arguments. Can you take another look if this is ok?

source/particle/world.cc
gassmoeller: At the moment there is no way for particle properties to signal which components they need, but yes that is a good suggestion, and if I have some time during the rest of the week I will work on that. I left a TODO. At the moment the class keeps the FEPointEvaluation members public to allow the owning object to access them individually. It is not as nice as being able to hand over a componentmask, but it works for now (the particle advection only accesses the velocity evaluator, the update uses the member functions to evaluate all components).

source/particle/world.cc
gassmoeller: I tend to think of ASPECT as a purely MPI parallel code with a very occasional use of threads for writing files. I know that the assembly uses the workstream interface, but other than that I don't think I usually think about parallelizing things in threads. Maybe it would be an interesting project at some point to investigate the potential of more threading, but until then I prefer the simpler code and concepts to a potential speedup for a small set of models. :-)

gassmoeller:
Thanks for the review! I think I addressed all of your comments, please take a look if the documentation of the class is sufficient.

source/particle/world.cc
bangerth: I meant to separate the declaration from the implementation of the member functions of this class.

source/particle/world.cc
bangerth: Yes, that can be a long-term goal. Using `ComponentMask` is probably the right abstraction.

source/particle/world.cc
bangerth: Yes, and when we use MPI, we do actually disable threading as well. But I think that a lot of initial tests are run without MPI and where it's easy, it might actually be nice to use threading.

But it's not a priority any of us should have :-)

source/particle/world.cc
gassmoeller: Oh sure I can do that. I thought it saves some lines to keep it together since the class only lives inside this .cc file. Will do.

gassmoeller:
Ok, I separated the function implementations from the declarations in the class. Ready for another review.

source/particle/world.cc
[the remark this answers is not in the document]
bangerth: :-)

bangerth (approved):

bangerth:
Thanks, very nice!

PR #4062, opened 2021-06-16
Store integrator properties as regular particle properties

gassmoeller:
This PR properly implements the first half of #4060. By not longer storing an internal map for the integrators that hold temporary data we avoid a lot of data allocation and lookups into the maps as well as the complicated data transfer. In one case with nearly 1 million particles per core this saves about 33% of the particle advection time (in most cases it is likely less).

I tried to hide the new particle property by:
 - defining it internally in interface.cc instead of as a separate plugin
 - adding its properties to the end of the particle properties
 - automatically ignoring it for the output

However, it is of course still there, so if you output all particle properties you will see properties at the end that have no relation to the selected particle properties in the list of properties. Is this acceptable? The only reasonable alternative I would see is to assert that the user adds an entry 'integration properties' to the list of particle properties to make it explicitly visible, but that would break a lot of models.

### DIFF
--- a/include/aspect/particle/integrator/rk_2.h
+++ b/include/aspect/particle/integrator/rk_2.h
@@ -34,7 +34,7 @@ namespace aspect
     {
       /**
        * Runge Kutta second order integrator.
-       * This scheme requires storing the original location, and the read/write_data functions reflect this.
+       * This scheme requires storing the original location in the particle properties.
        *
        * @ingroup ParticleIntegrators
        */
@@ -43,6 +43,14 @@ namespace aspect
       {
         public:
           RK2();
+
+          /**
+           * Look up where the RK2 data is stored. Done once and cached to
+           * avoid repeated lookups.
+           */
+          virtual
+          void
+          initialize ();
 
           /**
            * Perform an integration step of moving the particles of one cell
@@ -80,34 +88,6 @@ namespace aspect
            */
           bool new_integration_step() override;
 
-          /**
-           * Return data length of the integration related data required for
-           * communication in terms of number of bytes. When data about
-           * particles is transported from one processor to another, or stored
-           * on disk for snapshots, integrators get the chance to store
-           * whatever information they need with each particle. This function
-           * returns how many pieces of additional information a concrete
-           * integrator class needs to store for each particle.
-           *
-           * @return The number of bytes required to store the relevant
-           * integrator data for one particle.
-           */
-          std::size_t get_data_size() const override;
-
-          /**
-           * @copydoc Interface::read_data()
-           */
-          const void *
-          read_data(const typename ParticleHandler<dim>::particle_iterator &particle,
-                    const void *data) override;
-
-          /**
-           * @copydoc Interface::write_data()
-           */
-          void *
-          write_data(const typename ParticleHandler<dim>::particle_iterator &particle,
-                     void *data) const override;
-
         private:
           /**
            * The current integration step, i.e for RK2 a number that is either
@@ -116,12 +96,9 @@ namespace aspect
           unsigned int integrator_substep;
 
           /**
-           * The particle location before the first integration step. This is
-           * used in the second step and transferred to another process if
-           * the particle leaves the domain during the first step.
+           * The location of the RK2 data that is stored in the particle properties.
            */
-          std::map<types::particle_index, Point<dim> >   loc0;
-
+          unsigned int property_location;
       };
 
     }
--- a/include/aspect/particle/integrator/rk_2.h
+++ b/include/aspect/particle/integrator/rk_2.h
@@ -88,6 +88,19 @@ namespace aspect
            */
           bool new_integration_step() override;
 
+          /**
+           * We need to tell the property manager how many intermediate properties this integrator requires,
+           * so that it can allocate sufficient space for each particle. However, the integrator is not
+           * created at the time the property manager is set up and we can not reverse the order of creation,
+           * because the integrator needs to know where to store its properties, which requires the property manager
+           * to be finished setting up properties. Luckily the number of properties is constant, so we can make it
+           * a static property of this class. Therefore, the property manager can access this variable even
+           * before any object is constructed.
+           *
+           * The Runge-Kutta 2 integrator requires a single point with dim components.
+           */
+          static const unsigned int n_integrator_properties = dim;
+
         private:
           /**
            * The current integration step, i.e for RK2 a number that is either
@@ -98,7 +111,7 @@ namespace aspect
           /**
            * The location of the RK2 data that is stored in the particle properties.
            */
-          unsigned int property_location;
+          unsigned int property_index_old_location;
       };
 
     }
--- a/include/aspect/particle/integrator/rk_4.h
+++ b/include/aspect/particle/integrator/rk_4.h
@@ -31,16 +31,24 @@ namespace aspect
     {
       /**
        * Runge Kutta fourth order integrator. This scheme requires
-       * storing the original location and intermediate k1, k2, k3, k4
-       * values, so the read/write_data functions reflect this.
+       * storing the original location and intermediate k1, k2, k3
+       * values in the particle properties.
        *
        * @ingroup ParticleIntegrators
        */
       template <int dim>
-      class RK4 : public Interface<dim>
+      class RK4 : public Interface<dim>, public SimulatorAccess<dim>
       {
         public:
           RK4();
+
+          /**
+          * Look up where the RK4 data is stored. Done once and cached to
+          * avoid repeated lookups.
+          */
+          virtual
+          void
+          initialize ();
 
           /**
            * Perform an integration step of moving the particles of one cell
@@ -79,34 +87,6 @@ namespace aspect
            */
           bool new_integration_step() override;
 
-          /**
-           * Return data length of the integration related data required for
-           * communication in terms of number of bytes. When data about
-           * particles is transported from one processor to another, or stored
-           * on disk for snapshots, integrators get the chance to store
-           * whatever information they need with each particle. This function
-           * returns how many pieces of additional information a concrete
-           * integrator class needs to store for each particle.
-           *
-           * @return The number of bytes required to store the relevant
-           * integrator data for one particle.
-           */
-          std::size_t get_data_size() const override;
-
-          /**
-           * @copydoc Interface::read_data()
-           */
-          const void *
-          read_data(const typename ParticleHandler<dim>::particle_iterator &particle,
-                    const void *data) override;
-
-          /**
-           * @copydoc Interface::write_data()
-           */
-          void *
-          write_data(const typename ParticleHandler<dim>::particle_iterator &particle,
-                     void *data) const override;
-
         private:
           /**
            * The current integration step, i.e. for RK4 a number between 0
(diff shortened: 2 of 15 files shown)


source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -251,13 +251,15 @@ namespace aspect
 
                     const Postprocess::Particles<dim> &particle_postprocessor =
                       this->get_postprocess_manager().template get_matching_postprocessor<const Postprocess::Particles<dim> >();
-                    const unsigned int n_property_components = particle_postprocessor.get_particle_world().get_property_manager().get_n_property_components();
+                    const auto &particle_property_information = particle_postprocessor.get_particle_world().get_property_manager().get_data_info();
+                    const unsigned int n_property_components = particle_property_information.n_components();
+                    const unsigned int n_internal_components = particle_property_information.get_components_by_field_name("integrator properties");
 
-                    AssertThrow(global_minimum_particle_properties.size() == n_property_components,
+                    AssertThrow(global_minimum_particle_properties.size() == n_property_components - n_internal_components,
```
bangerth, line 258: Can you explain why this is necessary? I don't see why the interpolator should have to care how many internal properties there should be...

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -259,13 +259,15 @@ namespace aspect
 
                     const Postprocess::Particles<dim> &particle_postprocessor =
                       this->get_postprocess_manager().template get_matching_postprocessor<const Postprocess::Particles<dim> >();
-                    const unsigned int n_property_components = particle_postprocessor.get_particle_world().get_property_manager().get_n_property_components();
+                    const auto &particle_property_information = particle_postprocessor.get_particle_world().get_property_manager().get_data_info();
+                    const unsigned int n_property_components = particle_property_information.n_components();
+                    const unsigned int n_internal_components = particle_property_information.get_components_by_field_name("integrator properties");
 
-                    AssertThrow(global_minimum_particle_properties.size() == n_property_components,
+                    AssertThrow(global_minimum_particle_properties.size() == n_property_components - n_internal_components,
```
bangerth, line 266: same question here of course

source/particle/property/interface.cc
```diff
@@ -286,6 +286,56 @@ namespace aspect
 
 
 
+      template <int dim>
+      void
+      IntegratorProperties<dim>::initialize_one_particle_property(const Point<dim> &/*position*/,
+                                                                  std::vector<double> &data) const
+      {
+        data.resize(data.size() + n_integrator_properties);
```
bangerth, line 294: Could we fill with `numbers::signaling_nan<double>()`?

source/particle/property/interface.cc
```diff
@@ -286,6 +286,56 @@ namespace aspect
...
+      }
+
+
+
+      template <int dim>
+      std::vector<std::pair<std::string, unsigned int> >
+      IntegratorProperties<dim>::get_property_information() const
+      {
+        const std::vector<std::pair<std::string,unsigned int> > property_information (1,
+                                                                                      std::make_pair("integrator properties",n_integrator_properties));
```
bangerth, line 304:
```suggestion
        const std::vector<std::pair<std::string,unsigned int> > property_information
        = {{"integrator properties",n_integrator_properties}};
```

source/particle/property/interface.cc
```diff
@@ -286,6 +286,56 @@ namespace aspect
...
+        return property_information;
```
bangerth, line 305: Or just directly return that curly-braced enclosed list here.

source/particle/property/interface.cc
```diff
@@ -286,6 +286,56 @@ namespace aspect
...
+          prm.enter_subsection ("Particles");
+          {
+            name = prm.get ("Integration scheme");
+
+            if (name == "rk2")
+              n_integrator_properties = dim;
+            else if (name == "rk4")
+              n_integrator_properties = 4*dim;
+            else if (name == "euler")
+              n_integrator_properties = 0;
```
bangerth, line 326: Do we need to know this at this point already, or could we let the derived classes figure this out themselves and pass to the constructor? It's a quantity that is known at object-creation time. We just need to create the integrators either here or in the function that calls this place. Point being that we shouldn't have to figure this out here, but the integrators should say themselves how much internal space they need.

source/postprocess/particles.cc
```diff
@@ -705,6 +705,9 @@ namespace aspect
             }
 
           exclude_output_properties = Utilities::split_string_list(prm.get("Exclude output properties"));
+
+          // Never output the integrator properties that are for internal use only
+          exclude_output_properties.push_back("integrator properties");
```
bangerth, line 710: This is where it would be useful to mark this field as "internal:" or some such.

bangerth (commented):
Nice work!

I didn't see the place where you are currently serializing these extra properties via the `read_data()` and `write_data()` functions that you are removing from the interface. Where is that place where these functions used to be called and are now no longer?

tjhei:
@gassmoeller can this wait until after 2.3 or do you want it in?

gassmoeller:
Let's postpone until after 2.3. It is a significant improvement, but I am a bit worried about hidden bugs.

source/particle/property/interface.cc
gassmoeller: I would agree, unfortunately they depend on each other. Integrators are not yet created at this point, because when they are created they need to know where the propertymanager class stores their internal properties. So one of them has to go first, and it is easier for the property manager to figure out how many properties are stored by the integrator, than it is for the integrator to figure out at which position the integrator properties are going to be stored. This is also why I cant make this property `const` at the moment.

source/particle/property/interface.cc
gassmoeller: Unfortunately no. There are situations where these properties are read and copied, e.g. when during the initial adaptive refinement the number of particles violates the minimum particles per cell limit and new particles are generated. The new particle properties need to be interpolated from the old ones (still containing the signaling nan).

gassmoeller:
I think I addressed all comments, except for the ones I commented on.

include/aspect/particle/integrator/rk_2.h
```diff
@@ -116,12 +96,9 @@ namespace aspect
...
 
           /**
-           * The particle location before the first integration step. This is
-           * used in the second step and transferred to another process if
-           * the particle leaves the domain during the first step.
+           * The location of the RK2 data that is stored in the particle properties.
            */
-          std::map<types::particle_index, Point<dim> >   loc0;
-
+          unsigned int property_location;
```
bangerth, line 101: I'm having second thoughts about the names of these variables. They represent indices, so maybe `stage_data_index_within_properties`? That's a longer name, but I think it better represents what this variable is about.

include/aspect/particle/integrator/rk_4.h
```diff
@@ -115,19 +95,9 @@ namespace aspect
...
-          std::map<types::particle_index, Point<dim> > loc0;
-
-          /**
-           * The intermediate values of the RK4 scheme. These are
-           * used in the following steps and transferred to another process if
-           * the particle leaves the domain during one of the steps.
-           */
-          std::map<types::particle_index, Tensor<1,dim> > k1, k2, k3;
-
+          std::array<unsigned int,4> property_k;
```
bangerth, line 100: and then here `stage_data_indices_within_properties`?

source/particle/interpolator/bilinear_least_squares.cc
bangerth: This is still open.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -251,13 +251,17 @@ namespace aspect
...
 
-                    AssertThrow(global_minimum_particle_properties.size() == n_property_components,
+                    // Check that if a global limiter is used, we were given the minimum and maximum value for each
+                    // particle property that is not an internal property.
+                    AssertThrow(global_minimum_particle_properties.size() == n_property_components - n_internal_components,
                                 ExcMessage("Make sure that the size of list 'Global minimum particle property' "
                                            "is equivalent to the number of particle properties."));
 
-                    AssertThrow(global_maximum_particle_properties.size() == n_property_components,
+                    AssertThrow(global_maximum_particle_properties.size() == n_property_components - n_internal_components,
```
bangerth, line 264: same here

source/particle/property/interface.cc
bangerth: I see. Can you at least initialize with an explicit zero as second argument to `resize`?

source/particle/property/interface.cc
bangerth: Ping?

source/particle/property/interface.cc
bangerth: That's a bummer. This place knows the integrator classes, though. Can we make the number of intermediate properties a `static const` member variable of these classes, and use symbolic names here? In other words, let the integrator classes provide this information so one isn't left wondering where these numbers come from.

include/aspect/particle/property/interface.h
```diff
@@ -500,6 +500,50 @@ namespace aspect
...
+           */
+          virtual
+          void
+          parse_parameters (ParameterHandler &prm);
+
+        private:
+          /**
+           * The number of integrator properties to store.
+           */
+          unsigned int n_integrator_properties;
```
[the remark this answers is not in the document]
bangerth: I see now why you can't do that. Can you add a comment that says that this variable is initialized in `parse_parameters()`?

bangerth (commented):
Getting close :-)

include/aspect/particle/property/interface.h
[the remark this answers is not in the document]
gassmoeller: Done

source/particle/interpolator/bilinear_least_squares.cc
gassmoeller: I had added a comment before the assert explaining it. Let me know in case the comment is not clear enough.

source/particle/property/interface.cc
gassmoeller: done

source/particle/property/interface.cc
gassmoeller: I find returning curly-braced lists hard to ready for non-c++ experts, but I changed it now. Initializer lists are weird for everyone coming from a c-background (not saying we should not use them, I just try to avoid using them when not necessary).

source/particle/property/interface.cc
gassmoeller: yup that works. I changed it.

include/aspect/particle/integrator/rk_2.h
gassmoeller: I feel that is longer than is useful, but what about `property_index`? It reads quite well in the new code I think.

gassmoeller:
Thanks for reminding me about some missed comments (my bad). I think I addressed all comments now.

bangerth (approved):

bangerth:
I'm often in favor of longer names, but don't need to impose my will every time :-) Thanks for your patience!

PR #4063, opened 2021-06-16
Make particle removal compatible with latest deal.II particle structure

gassmoeller:
Since dealii/dealii#11577 particle iterators are invalidated if another iterator in the same cell is removed. This broke the functionality to limit the number of particles in a cell, because ASPECT stores a list of iterators to remove before removing them one by one. Instead we should remove all of them in one go as is now possible with the `remove_particles` function.

### DIFF
--- a/source/particle/world.cc
+++ b/source/particle/world.cc
@@ -389,7 +389,14 @@ namespace aspect
                         particles_to_remove.push_back(particle_to_remove);
                       }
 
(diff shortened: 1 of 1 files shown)


tjhei (approved):
Did you test this works with 9.3 as well?

tjhei:
Well, the function is new on master...

PR #4071, opened 2021-06-19
fix bunge cookbook

tjhei:
fix:
1. missing slash (Ascii data file                                                                                       </home/heister/aspect-git/cookbooks/bunge_et_al_mantle_convectionvisc_depth_a.txt>
    not found!
2. add POINTS header to viscosity files, reported as
    Could not successfully read in the file header of the ascii data file
3. add column information to parse column names / count correctly


gassmoeller (approved):
Changes look good. It looks like this cookbook is not executed by check.mk otherwise our CI would have caught this? Do you see an easy way to fix this? Could be a follow-up PR, or an issue just to remind us to fix this.

tjhei:
It is part of check.mk but we only run --validate, which does not run the simulation. These are runtime errors...

Issue #4064, opened 2021-06-17
Averaging material properties for graphical output


bangerth:
On the call on Tuesday, we discovered that the graphical output shows material properties as they are computed pointwise, whereas the internals of ASPECT may use them after averaging. We concluded that we should show material properties as they are used, i.e., we should add the averaging there as well.

This is not too hard to add, except for the Q1 projection. There, we compute a projection matrix that uses the *quadrature formula* that is being used in the place where the averaging is requested. This is a nuisance because it means that we may be computing different projections in different places, which I think is undesirable. It is also annoying because it means that the Q1 averaging cannot easily be performed in places where we don't actually do quadrature -- like in the viz postprocessors.

What to do?

tjhei:
What stops us from always using a QGauss(veldegree+1) or whatever we use in the assembly? 

At first I thought that it is important to use exactly that quadrature because it is not the exact projection otherwise, but now I am thinking that it really doesn't matter what quadrature we choose (as long as we are consistent) because it is just an approximation of the continuous projection.

bangerth:
Yes, something like this is on my mind.

bangerth:
Actually, having read through the code more carefully (and improved the documentation), we are not free to choose the quadrature formula because @egpuckett and colleagues chose the definition of what it means to do a Q1 projection tied to a specific quadrature formula. In essence, the way it is implemented right now is that values at specific points on the real cell are given (conceptually, at least -- these locations are only provided in reference coordinates, but one should think of the projection to happen in real space), and we are looking for a function that matches these point values best where each point is weighted by the quadrature weight.

One *could* have defined things on the reference cell and omitted the weights -- i.e., we pull back the values to the reference cell and do the projection there, with equal weight given to all points. That's just not what's currently done, and I'm loath to change this -- so #4076 simply punts if Q1 projection of properties was selected and skips the averaging for graphical output.

PR #4076, opened 2021-06-22
Average before producing graphical output.

bangerth:
In reference to #4064. Gets us 90% of the way there. Will probably break a lot of tests -- let's see what happens :-)

/rebuild


bangerth:
Oh, that's a surprise: No failures?

bangerth:
I've confirmed that none of the tests that do averaging actually output graphical output as well.

gassmoeller (commented):
Looks good to me, but have you checked that the output looks reasonable? Maybe add the graphical output to one of the tests that use material averaging, that way we would also have a test for this.

bangerth:
I added a test that starts from the SolKz model, which has an exponentially growing viscosity in z-direction. When visualized without material averaging, one actually sees the exponential growth, but when using material averaging, it shows a stepwise behavior.

This test is replicated three times for arithmetic, geometric, and harmonic averaging. Actually, that's really only what I wanted -- for geometric averaging, the solver doesn't converge, so I didn't press the issue and just check the arithmetic and harmonic averaging, without investigating too deeply what is happening for the geometric case.  I verified that the outputs for the two tests I have look reasonable in comparison to each other, and to the original smooth viscosity.

source/postprocess/visualization/material_properties.cc
```diff
@@ -106,6 +112,36 @@ namespace aspect
...
+        // between input and output of the averaging operation is generally
+        // small and probably not visible anyway.
+        //
+        // The average() function checks whether any quadrature object it uses
+        // has the correct size. So passing an invalid object carries little
+        // risk if the list of averaging modes that require a quadrature
+        // object expands: Every time we generate graphical output for
+        // a model that uses this kind of averaging, we will trigger an
+        // exception.
+        Quadrature<dim> quadrature;
```
tjhei, line 131: is this variable used? It seems you construct an unnamed quadrature in the call below.

bangerth: Ah, that's a left-over from a previous iteration. Let me remove it.

bangerth:
Good spot, @tjhei ! I fixed it!

tjhei (approved):
Looks good to me!

gassmoeller:
Thanks for adding the tests. Ready to merge.

PR #4077, opened 2021-06-22
Better document what the projection actually does.

bangerth:
Just a write-up of what a function does.

### DIFF
--- a/source/material_model/interface.cc
+++ b/source/material_model/interface.cc
@@ -687,32 +687,60 @@ namespace aspect
        * Given a quadrature formula, compute a matrices $E, M^{-1}F$
        * representing a linear operator in the following way: Let
        * there be a vector $F$ with $N$ elements where the elements
-       * are data stored at each of the $N$ quadrature points. Then
-       * project this data into a Q1 space and evaluate this
-       * projection at the quadrature points. This operator can be
-       * expressed in the following way where $P=2^{dim}$ is the
-       * number of degrees of freedom of the $Q_1$ element:
+       * are data stored at each of the $N$ quadrature points of the
+       * given quadrature formula. (For this quadrature formula, we only
+       * care about the locations of the quadrature points on the
+       * reference cell, but not about quadrature weights.)
+       * Then project this data into a $Q_1$ space and evaluate this
+       * projection at the same quadrature points.
        *
-       * Let $y$ be the input vector with $N$ elements. Then let
-       * $F$ be the $P \times N$ matrix so that
-       * @f{align}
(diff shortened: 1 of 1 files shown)


bangerth:
Separately (and we don't have to wait for the answer to merge this):

@egpuckett Do you recall why you included the term `|J(x_q)| w_q` in the objective function that is being minimized here? I guess it makes sense if you think of `y` as a vector of values corresponding to some function, and that what you really want to minimize is `\| u_h(x) - y(x) \|^2_{L_2(K)}`.

Related to #4064.

source/material_model/interface.cc
```diff
@@ -687,32 +687,60 @@ namespace aspect
        * Given a quadrature formula, compute a matrices $E, M^{-1}F$
        * representing a linear operator in the following way: Let
        * there be a vector $F$ with $N$ elements where the elements
-       * are data stored at each of the $N$ quadrature points. Then
-       * project this data into a Q1 space and evaluate this
-       * projection at the quadrature points. This operator can be
-       * expressed in the following way where $P=2^{dim}$ is the
-       * number of degrees of freedom of the $Q_1$ element:
+       * are data stored at each of the $N$ quadrature points of the
+       * give quadrature formula. (For this quadrature formula, we only
```
tjhei, line 691:
```suggestion
       * given quadrature formula. (For this quadrature formula, we only
```

source/material_model/interface.cc
```diff
@@ -687,32 +687,60 @@ namespace aspect
...
+       *   min_{u_h \in Q_1(K)} \frac 12 \sum_q |u_h(x_q) - y_q|^2 |J(x_q)| w_q.
        * @f}
-       * where $\varphi_i$ are the $Q_1$ shape functions, $x_q$ are
-       * the quadrature points, $J$ is the Jacobian matrix of the
+       * where $x_q$ are the evaluation points, $J$ is the Jacobian matrix of the
        * mapping from reference to real cell, and $w_q$ are the
        * quadrature weights.
-       *
+       * Expanding $u_h = \sum_j U_j \varphi_j$, this leads to the following
+       * minimization problem for the vector $U$ of coefficients:
```
tjhei, line 710:
```suggestion
       * minimization problem for the vector of coefficients $U$:
```

bangerth:
Fixed as requested.

bangerth:
Aw, @tjhei, you're up late/early as well today... :-(

tjhei:
Yeah, I am (virtually) in Milano for SIAM GS21. 😉

bangerth:
Ah, then at least one of us has a good reason other than having to finish stuff that didn't get finished before traveling :-(

PR #4084, opened 2021-06-23
Rename a couple of functions.

bangerth:
This was the conclusion of #4051. These functions are not used in ASPECT itself, but in some (external?) plugins. I renamed it in the one test that used it.

/rebuild

### DIFF
--- a/include/aspect/structured_data.h
+++ b/include/aspect/structured_data.h
@@ -146,18 +146,19 @@ namespace aspect
         has_equidistant_coordinates() const;
 
         /**
-         * Returns the coordinates at which data is stored. This function
-         * can be used to determine the number of data points, or to query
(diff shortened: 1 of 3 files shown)


tjhei:
I am okay with this but I will let @gassmoeller decide. Should this go into 2.3?

gassmoeller (approved):
Yes, I am ok with this renaming, and we should merge it into 2.3 to avoid having to deprecate this function before removing it (it was added last November after the release of 2.2).

gassmoeller:
@alarshi: This PR will require you making a minor change to our custom plugin, a simple renaming from `get_coordinates` to `get_interpolation_point_coordinates`. But you can wait with that until we use a newer ASPECT version.

PR #4085, opened 2021-06-23
Minor optimization in read_and_distribute_file_content().

bangerth:
Just skip some MPI calls if we are have only one process.

/rebuild

### DIFF
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -1106,30 +1106,41 @@ namespace aspect
               filesize = data_string.size();
             }
 
-          // Distribute data_size and data across processes
-          int ierr = MPI_Bcast(&filesize,1,MPI_UNSIGNED,0,comm);
-          AssertThrowMPI(ierr);
-          ierr = MPI_Bcast(&data_string[0],filesize,MPI_CHAR,0,comm);
-          AssertThrowMPI(ierr);
+          // If there are other processes, distribute data_size and
+          // data across the MPI universe
+          if (Utilities::MPI::n_mpi_processes(comm) > 1)
+            {
+              int ierr = MPI_Bcast(&filesize,1,MPI_UNSIGNED,0,comm);
+              AssertThrowMPI(ierr);
+              ierr = MPI_Bcast(&data_string[0],filesize,MPI_CHAR,0,comm);
+              AssertThrowMPI(ierr);
+            }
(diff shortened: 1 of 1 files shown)


source/utilities.cc
```diff
@@ -1106,31 +1106,42 @@ namespace aspect
...
             }
 
-          // Distribute data_size and data across processes
-          int ierr = MPI_Bcast(&filesize,1,MPI_UNSIGNED,0,comm);
-          AssertThrowMPI(ierr);
-          ierr = MPI_Bcast(&data_string[0],filesize,MPI_CHAR,0,comm);
-          AssertThrowMPI(ierr);
+          // If there are other processes, distribute data_size and
+          // data across the MPI universe
+          if (Utilities::MPI::n_mpi_processes(comm) > 1)
```
tjhei, line 1111: I am not sure this is particularly useful as MPI_BCast is a noop in this situation. It doesn't make the code simpler or easier to understand, in my opinion.

bangerth: I'm not particularly adamant about this patch. It mainly resulted from my desire to be symmetric between the send and receive parts. I was wondering while reading through the code what happens if there is only one processor. In that case, we go into the big `if` case and encounter a statement marked as "Distribute data_size and data across processes", but nobody gets to the statement that's marked as "Receive and store data". This works precisely because `MPI_Bcast` doesn't do anything if `nprocs==1`, but it looks a bit funny to see a send but no receive statement.

Either way, I'm ok not merging this PR if you feel strongly about it.

bangerth:
What should we do here?

source/utilities.cc
gassmoeller: I would be fine adding this part of the PR (in the `if rank == 0` branch) to be clear that we only start an MPI operation if there are more than 1 MPI ranks. But I am wondering about the else branch. In the else-case we are not rank 0, in this case the condition `Utilities::MPI::n_mpi_processes(comm) > 1` should really always be fulfilled.
Merging only half of the branch will of course not help your goal of symmetry either. Maybe we just close it? I think all of this would be much clearer, if we could refactor all of the lengthy code above into a separate function so that we can actually see the `if` condition for all of this.

bangerth:
Let's not agonize over it :-)

PR #4086, opened 2021-06-23
Enable data sharing across processes

bangerth:
This builds on #4083, so the first patch is duplicated from there (but because it's identical, we can still merge this assuming #4083 were merged).

I *believe* this should work, but I don't know for sure and it's late and I'm leaving town tomorrow afternoon until Monday. Let us see what the testers say -- they will at least tell us whether it works with deal.II 9.2. Do any of the testers run 9.3?

The real test would be if (i) we have tests that use ASCII data tables in parallel, and (ii) we check with deal.II 9.3. I don't know whether we have any that satisfy (i). Anyone know for sure?

/rebuild

### DIFF
--- a/include/aspect/structured_data.h
+++ b/include/aspect/structured_data.h
@@ -90,7 +90,7 @@ namespace aspect
          * The `coordinate_values` and `data_table` arguments are rvalue references,
          * and the function will move the data so provided into another storage
          * location. In other words, after the call, the variables passed as the
-         * last two arguments may be empty or otherwise altered.
+         * second and third arguments may be empty or otherwise altered.
          *
         #if DEAL_II_VERSION_GTE(9,3,0)
          * If ASPECT is built on deal.II version 9.3 or higher, this class
--- a/include/aspect/structured_data.h
+++ b/include/aspect/structured_data.h
@@ -87,14 +87,41 @@ namespace aspect
          *
          * The data in @p data_table consists of a Table for each of the @p n_components components.
          *
-         * The last two arguments are rvalue references, and the function will
-         * move the data so provided into another storage location. In other
-         * words, after the call, the variables passed as the last two
-         * arguments may be empty or otherwise altered.
+         * The `coordinate_values` and `data_table` arguments are rvalue references,
+         * and the function will move the data so provided into another storage
+         * location. In other words, after the call, the variables passed as the
+         * last two arguments may be empty or otherwise altered.
+         *
+        #if DEAL_II_VERSION_GTE(9,3,0)
+         * If ASPECT is built on deal.II version 9.3 or higher, this class
+         * is able to share data between processes located on the same
+         * machine. In this case, if the last argument, `root_process` is
+         * set to anything other than `numbers::invalid_unsigned_int`, then only the
+         * indicated root process within the given MPI communicator needs to
+         * pass in valid arguments whereas on all other processes the values
+         * of `column_names`, `coordinate_values`, and `data_table` are
+         * ignored. Instead, the indicated root process will make sure that
+         * the other processes obtain valid data, for example by sharing
+         * data tables in shared memory spaces. This reduces both the
+         * computational effort in reading data from disk and parsing it on
+         * all processes, and can vastly reduce the amount of memory required
+         * on machines where many processor cores have access to shared memory
+         * resources.
+         *
+         * If `root_process` equals `numbers::invalid_unsigned_int`, then
+         * every process needs to pass data for all arguments and the
+         * `mpi_communicator` argument is ignored.
+        #else
+         * If ASPECT is built on a version of deal.II older than 9.3,
+         * then all processes need to pass in valid values for the first
+         * three arguments and the last two arguments are ignored.
+        #endif
          */
         void reinit(const std::vector<std::string> &column_names,
                     std::vector<std::vector<double>> &&coordinate_values,
-                    std::vector<Table<dim,double>> &&data_table);
+                    std::vector<Table<dim,double>> &&data_table,
+                    const MPI_Comm &mpi_communicator = MPI_COMM_SELF,
+                    const unsigned int root_process = numbers::invalid_unsigned_int);
 
         /**
          * Loads a data text file. Throws an exception if the file does not
--- a/include/aspect/structured_data.h
+++ b/include/aspect/structured_data.h
@@ -92,8 +92,8 @@ namespace aspect
          * location. In other words, after the call, the variables passed as the
          * second and third arguments may be empty or otherwise altered.
          *
-        #if DEAL_II_VERSION_GTE(9,3,0)
-         * If ASPECT is built on deal.II version 9.3 or higher, this class
+        #if DEAL_II_VERSION_GTE(9,4,0)
+         * If ASPECT is built on deal.II version 10.0 or higher, this class
          * is able to share data between processes located on the same
          * machine. In this case, if the last argument, `root_process` is
          * set to anything other than `numbers::invalid_unsigned_int`, then only the
@@ -112,7 +112,7 @@ namespace aspect
          * every process needs to pass data for all arguments and the
          * `mpi_communicator` argument is ignored.
         #else
-         * If ASPECT is built on a version of deal.II older than 9.3,
+         * If ASPECT is built on a version of deal.II 9.3 or older,
          * then all processes need to pass in valid values for the first
          * three arguments and the last two arguments are ignored.
         #endif
--- a/include/aspect/structured_data.h
+++ b/include/aspect/structured_data.h
@@ -93,7 +93,7 @@ namespace aspect
          * second and third arguments may be empty or otherwise altered.
          *
         #if DEAL_II_VERSION_GTE(9,4,0)
-         * If ASPECT is built on deal.II version 10.0 or higher, this class
+         * If ASPECT is built on deal.II version 9.4 or higher, this class
          * is able to share data between processes located on the same
          * machine. In this case, if the last argument, `root_process` is
          * set to anything other than `numbers::invalid_unsigned_int`, then only the
--- a/doc/modules/changes/20210728_bangerth
+++ b/doc/modules/changes/20210728_bangerth
@@ -1,0 +1,6 @@
+New: Where possible, when using large data tables as input (e.g., for
+initial conditions specified as tables), these data are now stored
+only once on each node in memory areas that is accessible by all MPI
+processes on that node.
+<br>
+(Wolfgang Bangerth, 2021/07/28)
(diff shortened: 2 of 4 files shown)


tjhei:
> Do any of the testers run 9.3?

No.

My vote would be to wait with merging this until after the 2.3 release as it is not tested at this point.

gassmoeller:
> My vote would be to wait with merging this until after the 2.3 release as it is not tested at this point.

I agree. This is pretty useful functionality, but I am afraid of subtle bugs. Also if we merge this when requiring 9.3 we could avoid the ifdefs.

bangerth:
OK. Do any of you have a setup running regularly where this could be tested? I'll think of a test in the next few days, but I can't test on more than one node (but multiple processes) since I don't have access to any such systems on a regular basis.

include/aspect/structured_data.h
```diff
@@ -87,14 +87,41 @@ namespace aspect
...
          * The data in @p data_table consists of a Table for each of the @p n_components components.
          *
-         * The last two arguments are rvalue references, and the function will
-         * move the data so provided into another storage location. In other
-         * words, after the call, the variables passed as the last two
-         * arguments may be empty or otherwise altered.
+         * The `coordinate_values` and `data_table` arguments are rvalue references,
+         * and the function will move the data so provided into another storage
+         * location. In other words, after the call, the variables passed as the
+         * last two arguments may be empty or otherwise altered.
```
gassmoeller, line 93: They are not the last arguments any longer.

include/aspect/structured_data.h
```diff
@@ -87,14 +87,41 @@ namespace aspect
...
+         * then all processes need to pass in valid values for the first
+         * three arguments and the last two arguments are ignored.
+        #endif
          */
         void reinit(const std::vector<std::string> &column_names,
                     std::vector<std::vector<double>> &&coordinate_values,
-                    std::vector<Table<dim,double> > &&data_table);
+                    std::vector<Table<dim,double> > &&data_table,
+                    const MPI_Comm &mpi_communicator,
+                    const unsigned int root_process);
```
gassmoeller, line 124: could you provide defaults for these parameters? That way we stay backward compatible with user plugins that use this class.

gassmoeller (commented):
The code looks good, I only have two smaller comments.

About the test: Maybe @alarshi has a setup that uses an ascii data file that is somewhat large? Otherwise we have lots of ascii_data tests, but all of them use very small tables.

include/aspect/structured_data.h
alarshi: @gassmoeller: Yes, I can test it using the global tomography Ascii data, we input in our models.

gassmoeller: You may want to talk to @bangerth first, I think the purpose is not only to find out if the code works, but also if it actually only stores a single copy per node (instead of a single copy per process).

tjhei: It might not be trivial to see if we actually need less memory (not sure how this shows up in the top, maybe under "shared"). But running it with 2 nodes to see if it works (including the check that the values are set on each rank!), would be good to know.

tjhei: Finally, we will need to check with Intelmpi, openmpi and mpich.

bangerth: We already established that it stores only one copy with the underlying deal.II patches. I would love for someone to just try a larger model without and with this patch just to make sure that we get the same results. @alarshi: If that is something that would be easy for you to try, that would be fantastic!

include/aspect/structured_data.h
bangerth: Hm, what should the default values be? I could give `MPI_COMM_SELF` to the former and the special value `numbers::invalid_unsigned_int` to the latter. If a user just calls the old version with both arguments defaulted, that works. If a user calls the new version with both arguments given, that also works. This scheme also seems to work if someone uses a provided communicator and defaults the last argument -- in that case, the communicator is simply ignored, which is also what is documented. 

So I think I just talked myself through doing just that! :-)

include/aspect/structured_data.h
alarshi: @bangerth: Absolutely! So, is the goal to run on 2 nodes and then compare the difference?

bangerth: Yes, I think that would be totally sufficient to convince ourselves that it works. Thanks for giving it a try!

alarshi: @bangerth: I am getting segmentation fault when running my models with the branch. The error message is: 
[output_screen.txt](https://github.com/geodynamics/aspect/files/6787460/output_screen.txt)
I checked that the model is running with the current master. 
The input prm file and the ascii data I used are here: https://www.dropbox.com/sh/kxr9p2cw1j9ysjo/AAAvCZJ8Jhull2zLpUES5HZEa?dl=0

bangerth:
@alarshi Thanks for providing me with a testcase! I've had to update a couple more places where we store duplicate information and where all processes needed to be given a copy of the data. This now works on my workstation. Would you be willing to run the same test case you gave me again on a larger machine?

alarshi:
@bangerth: Absolutely! I just submitted the job for the model on HPC, will keep you updated on the results :)

alarshi:
@bangerth: The test ran on the HPC, but I'm unable to render the model results in ParaView. It gives me an error as "byte index ... not well formed". I am running the model again to see if it is reproducible.

bangerth:
That sounds unrelated. How do the other statistics ASPECT outputs compare with/without this patch? Do you get the same number of iterations, the same average velocity, etc?

alarshi:
@bangerth: The number of iterations and the average velocity are the same between the two branches. Here are the two log files: 
[master-log.txt](https://github.com/geodynamics/aspect/files/6842269/master-log.txt)
[bangerth-log.txt](https://github.com/geodynamics/aspect/files/6842272/bangerth-log.txt)

bangerth:
Excellent, that's good news -- thanks very much for checking, @alarshi !

What does everyone else think?

gassmoeller:
Looks good to me. Can you rebase and add a changelog entry?

bangerth:
OK, this was frustrating because I had introduced a bug in deal.II three weeks ago that required a patch, and there was one other bug that was exposed by the unit tests (but in a code path that is, I believe, not actually used in ASPECT itself). In any case, the deal.II bug has a patch in https://github.com/dealii/dealii/pull/13361, and with this patch applied, the version of this patch works locally on my hard drive.

It will still fail the CI check with the deal.II dev version until that container uses a patched deal.II. But I'm reasonably sure that this is now correct and does what is expected.

tjhei:
Can you fix the conflicts, please? It should be good to go now.

gassmoeller:
This is likely the time to finish this PR :-)

bangerth:
It's going to take a bit to get that rebased :-(

bangerth:
Well, that actually turned out to be quite the pain. In the end, I first reverted two of @jdannberg 's patches, then cherry-picked all of my patches, and then re-applied the two patches. If you'd like to follow along (or at the very least to document this), here's what I did:

* I reverted some indentation changes in 888e90d56ccce1fc5f2c9b53e8df34d392d56bf7
* I reverted c017f06611f02a3e5739fa503fa7fbc6c177e51d in 8c3074d14dc6f3c59058e1f7354ff2c0eb3704d3
* I reverted a96196d3c13c8b33632ca12a8e04231f107d40b4 in 97d13a134d67f122f64c5f67445febd690c014dd
* I cherry-picked my patches 6a2e136e9a4e599cb42396aa3eacec702893f740...ed98a7fc068521c3733e6510c1154dba616c49a4
* I applied current indentation style in 286bd6bd4b236d09e40668314a55dcfeaeff42ad
* I reapplied @jdannberg 's two patches (a96196d3c13c8b33632ca12a8e04231f107d40b4 and c017f06611f02a3e5739fa503fa7fbc6c177e51d) in bd6dbaae507b8c7cb2e42fee645e9dc0842460f7

I am *pretty* sure that I did this right, but the tester will show for sure.

/rebuild

tjhei:
That proves the point that even small indentation changes can be disruptive...

bangerth:
The indentation change was the least of my concerns -- it was only one line in this file that was affected. @jdannberg's patches were a real hassle, though..

Anyway, it tests ok now.

include/aspect/structured_data.h
```diff
@@ -87,14 +87,41 @@ namespace aspect
...
-         * move the data so provided into another storage location. In other
-         * words, after the call, the variables passed as the last two
-         * arguments may be empty or otherwise altered.
+         * The `coordinate_values` and `data_table` arguments are rvalue references,
+         * and the function will move the data so provided into another storage
+         * location. In other words, after the call, the variables passed as the
+         * second and third arguments may be empty or otherwise altered.
+         *
+        #if DEAL_II_VERSION_GTE(9,4,0)
+         * If ASPECT is built on deal.II version 10.0 or higher, this class
```
gassmoeller, line 96:
```suggestion
         * If ASPECT is built on deal.II version 9.4 or higher, this class
```

gassmoeller (approved):
Looks good to go to me. Let's get this merged when the testers are done.

include/aspect/structured_data.h
```diff
@@ -87,14 +87,41 @@ namespace aspect
...
-         * The last two arguments are rvalue references, and the function will
-         * move the data so provided into another storage location. In other
-         * words, after the call, the variables passed as the last two
-         * arguments may be empty or otherwise altered.
+         * The `coordinate_values` and `data_table` arguments are rvalue references,
+         * and the function will move the data so provided into another storage
+         * location. In other words, after the call, the variables passed as the
+         * second and third arguments may be empty or otherwise altered.
+         *
+        #if DEAL_II_VERSION_GTE(9,4,0)
```
tjhei, line 95: Why do you put this ifdef inside the comment? I don't think this will work and it makes sense to have both parts in the documentation.

tjhei (approved):

gassmoeller:
Let's sort out the documentation later in order to avoid further conflicts with the other StructuredData PRs.

include/aspect/structured_data.h
bangerth: Doxygen runs (some variation of) the preprocessor before ingesting the sources. That's how our `#ifndef DOXYGEN` works.

tjhei: But macros don't work inside a comment! See https://aspect.geodynamics.org/doc/doxygen/classaspect_1_1Utilities_1_1StructuredDataLookup.html#a7a5625b138bc39bf4e53e9be4d9b3314

bangerth: Oh, interesting. On it -- see #4789.

PR #4087, opened 2021-06-23
Cherry pick the cleanups in #4086

bangerth:
If we're unsure about #4086, let me at least get the underlying cleanups out of the way. No functional changes here, just reorganization of code, individually packaged into three commits.

/rebuild


gassmoeller (approved):
This looks ok and uncontroversial to me.

PR #4090, opened 2021-06-25
Allow selecting individual compositions for world builder

gassmoeller:
We have a model setup where using the world builder to prescribe initial compositions is very expensive, however we only want to use it for 1 of our 5 compositions. At the moment it is impossible to call the world builder function only for selected compositions, but this is now added in this PR. 

@MFraters can you take a look if this makes sense?

### DIFF
--- a/tests/world_builder_select_composition.prm
+++ b/tests/world_builder_select_composition.prm
@@ -1,0 +1,67 @@
+# WORLD BUILDER
+# This test is a copy of world_builder_simple.prm and checks that we can select
+# a subset of compositions for which to evaluate the world builder.
+
+set World builder file                     = $ASPECT_SOURCE_DIR/tests/world_builder_select_composition.wb
+set Use years in output instead of seconds = false
+set Adiabatic surface temperature = 1613
+set End time = 0
+
+subsection Adiabatic conditions model
+  subsection Compute profile
+    set Function expression = if(x > 1500000,1.0,0.0);0.0;0.0
+    set Composition reference profile = function
+  end
+end
+
+subsection Boundary velocity model
+  set Zero velocity boundary indicators = 1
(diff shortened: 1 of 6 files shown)


tests/world_builder_select_composition.prm
```diff
@@ -0,0 +1,72 @@
...
+    set Density differential for compositional field 1 = 500
+  end
+end
+
+subsection Mesh refinement
+  set Initial global refinement = 5
+end
+
+subsection Postprocess
+  set List of postprocessors = visualization
```
jdannberg, line 66: It could be useful to add the composition statistics postptocessor here, which would make it super easy to see that only one composition is nonzero.

jdannberg (commented):
Nice! I think this will be really useful. Except for the few comments, I think this is great!

MFraters:
Yes, this pull request makes sense to me. I agree with you and Juliane that having a default of returning all the compositions is correct. I am on the fence about having the default value being empty or "all", but I am kinda leaning to the side of keeping it empty. There is no reason for the user to provide an empty list, because then they just shouldn't use the world builder for compositions to begin with. And theoretically a user could have a compositional field called "all".

tjhei:
2.3 or can this wait?

gassmoeller:
I addressed all comments, this should be ready to merge. I chose to use an empty list as default, because it keeps the code simpler. 

This does not need to go into 2.3, it is a significant performance improvement for some models, but only for very specific applications (models using expensive world builder setups with many compositions).

tjhei:
I assume you will be running master anyways...

gassmoeller:
Yes, for us it does not matter at all.

jdannberg (approved):
Okay, looks good to me!

PR #4091, opened 2021-06-26
Define NDEBUG in release mode

gassmoeller:
As I noticed in GeodynamicWorldBuilder/WorldBuilder#290 the world builder uses the NDEBUG definition to decide whether to activate or deactivate asserts, which seems a standard thing to do (at least it is also used for assertions in the standard library). Cmake automatically sets NDEBUG in anything but debug mode, however deal.II overwrites these defaults in the `DEAL_II_INITIALIZE_CACHED_VARIABLES` macro. This means at the moment world builder asserts are always executed even in release mode. We should avoid this, and the easiest thing to do would be to restore cmake's default behavior of defining NDEBUG whenever not in debug mode. @tjhei @bangerth Is there a reason we should not do this? 

I will open a separate PR for deal.II to ask if this is intended behavior, but I would like a fix for ASPECT before the release as this pretty heavily impacts performance with the world builder.


tjhei (approved):
I don't see a problem in doing this. Can you please add a comment above the reason for doing this?

MFraters (approved):
Looking back, although it was a continuous decision to use NDEBUG, I never realize that this would be an issue when integrating it into Aspect (which I did right from the start). Thanks for finding and fixing this issue @gassmoeller!

gassmoeller:
Great, I added a comment. 

> I never realize that this would be an issue when integrating it into Aspect

Yeah, it is also something that is invisible if running the world builder standalone. The only way it could have been discovered was if somebody triggered one of the WB asserts in aspect in release mode and wondered about it. I found it only because operations on world builder points appeared really slow in callgrind profiles :-).

tjhei (approved):

PR #4096, opened 2021-06-28
Initialize current linearization point after melt restart

### LINKED ISSUE #4088: Different advection system after restart
Looking into an issue where after a restart the number of iterations for a Stokes solve differs greatly, I noticed that the rhs and the matrix of the advection solves also differ between a restarted timestep and the corresponding timestep of the original run. 

For example, with the test `mass_reaction_term.prm` at higher resolution and tolerance, the norm of the rhs of the advection system differs 2 OOM after a restart in the first nonlinear iteration. I didn't expect this, and wonder if it can lead to the Stokes solver differences I see in more complex models. Should I look further into this difference?

This test also includes operator splitting, melting, iterated Advection and Stokes.

Image shows the fourth timestep for the original run on the right and the restarted run on the left with a mainline aspect apart from the edit to output the norm of the rhs and matrix. The rhs norm differs 2 OOM in the first nonlinear iteration. In the second iteration, the nr of iterations for the porosity solve also differs.

Screenshot 2021-06-23 at 14 52 19

gassmoeller:
This should fix #4088. The current constraints in computations with melt depend on the current linearization point (to determine which cells should be treated as melt cells), but we do not serialize the current linearization point and only initialize it after computing the current constraints. This could lead to different results in the first nonlinear iteration after a restart (presumably all cells were treated as non melt cells and only the normal Stokes equations were solved on them).

This PR makes sure the current linearization point is initialized before computing the current constraints after a restart. This means we now initialize the current linearization point twice after a resume, but that should not matter (it's reasonably fast and only happens occasionally).

@anne-glerum: Can you check if this fixes your issue?

### DIFF
--- a/source/simulator/checkpoint_restart.cc
+++ b/source/simulator/checkpoint_restart.cc
@@ -580,7 +580,10 @@ namespace aspect
     // We have to compute the constraints here because the vector that tells
     // us if a cell is a melt cell is not saved between restarts.
     if (parameters.include_melt_transport)
-      compute_current_constraints ();
+      {
+        initialize_current_linearization_point();
+        compute_current_constraints ();
+      }
   }
 
 }


tjhei (approved):

tjhei:
It's a step forward so we should merge...

PR #4033, opened 2021-05-12
Catch exceptions to show message before abort

gassmoeller:
This fixes an issue with weird MPI error messages we saw in our group, and should also fix #4028. Catching the exception and showing it's error message is sufficient to illustrate the real problem instead of the fake errors that result from MPI calls in destructors of objects further up in the callstack. I necessary we could also additionally call MPI_Abort, but at least for my problems this worked nicely already. @sac-bsa: Could you give this branch a try and see if your errors in #4028 look better now?

PR #4097, opened 2021-06-29
cherry-picks for 2.3

tjhei:
#4095 and #4096


tjhei:
@gassmoeller is this everything we need?

gassmoeller:
And #4033 please. Otherwise yes this is everything. It seems #4096 does not fully fix #4088, but the remaining differences after restart seem small and disappear after nonlinear iterations, so it should not block the release.

tjhei:
updated.

gassmoeller (approved):

PR #4099, opened 2021-06-29
Add dealii dev tester

gassmoeller:
This adds a github action that compiles ASPECT against the dealii/dealii:master-focal container with every PR. Currently does not run the test suite.

No more PRs that work on the tester but break on deal.II dev :tada:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -68,3 +68,27 @@ jobs:
       with:
         name: manual.log
         path: doc/manual/manual.log
+
+  dealii-dev:
+    # linux build with deal.II dev
+    name: test-dealii-dev
+    runs-on: [ubuntu-18.04]
+    container:
+      image: dealii/dealii:master-focal
+      options: --user 0
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: compile
+      run: |
+        mkdir build
+        cd build
+        cmake \
+        -G 'Ninja' \
+        -D CMAKE_CXX_FLAGS='-Werror' \
+        -D ASPECT_TEST_GENERATOR='Ninja' \
+        -D ASPECT_PRECOMPILE_HEADERS=ON \
+        -D ASPECT_UNITY_BUILD=ON \
+        ..
+        ninja
+        ./aspect --test


gassmoeller:
Rebased on #4100.

gassmoeller:
I rebased on #4100, this should be ready if we agree it is a good idea.

tjhei:
Is it useful to do this on each PR? I think we talked about it to be run on master at regular intervals instead. It is a bit weird that a change of deal.ii master will make the next user PR fail...

gassmoeller:
I am a bit undecided. Yes it is weird that deal.II dev would make our tester fail, on the other hand it ensures high priority. If we run it daily instead we can actually run the full test suite (without comparing results), but we would still merge PRs that break with deal.II dev. Not sure which (or both) is better, what do you think?

tjhei:
Not sure either. I would say it is more likely to have deal.ii master change, than a user adding something that works with the last release but not with master.

gassmoeller:
Alright, I will change it to run daily on master.

bangerth:
Whatever breaks with deal.II dev at any given moment will require a fix sooner rather than later anyway. We might as well catch this at the time of proposing a patch, and if necessary spend a few minutes explaining the situation to someone, rather than requiring someone else to go back to fix it.

bangerth:
In other words, I'm for it.

gassmoeller:
@bangerth: Do you mean you are for a daily run of the full testsuite (would require changes), or a smaller run for every PR that only compiles (the current state of this PR)? I am fine with either or both.

bangerth:
If we have the resources, I'm always for adding things to the CI :-)

gassmoeller:
In that case this PR is ready for merge (this is the quick 10 min test that aspect compiles with deal.II dev). I agree that it puts more pressure on us to fix problems quickly. Having a daily test that just sends an email, or changes a badge color is not as obvious. If it triggers too often because of unrelated things we can always disable it later.

bangerth:
I'm in favor. Should we merge?

gassmoeller:
Yes.

tjhei:
I am not convinced that this won't end up confusing PR creators, but we can try...

bangerth:
Looks like we need a bold leader willing to take a stand and press the button, consequences be damned! I will be that person for today!

ljhwang:
Was that Gru I saw on your desktop?

bangerth:
I don't know who Gru is, but I had the minion, the superhero, and a bunch of dinos from last year that have kept me company since then :-)

PR #4100, opened 2021-06-29
Fix unit tests for deal.II dev

gassmoeller:
Found because of #4099. It seems the formatting of assertion output has changed (or catch has become more strict about comparisons), anyway it does not ignore newlines in the output. Adjust the unit test requirements to not contain strings that are broken by a newline.


tjhei (approved):

gassmoeller:
Forgot two. Should be good to merge now, if the tests on #4099 pass.

tjhei (approved):

PR #4101, opened 2021-06-30
Set Github Actions concurrency to speed up feedback

gassmoeller:
This ensures that only a single github actions workflow per PR is run at the same time, and that new pushes will cancel existing runs. Will be useful to get quicker results from Github actions during the hackathon.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -2,6 +2,10 @@ name: linux
 
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   linux-nounity:
     #linux build with unity/precomiled headers disabled


tjhei:
Do you have a source for this? It seems github.ref is something like "refs/heads/feature-branch-1" and does not contain the repo itself?

gassmoeller:
For concurrency:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency

For github.ref:
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

github.ref : "The branch or tag ref that triggered the workflow run. For branches this in the format refs/heads/<branch_name>, and for tags it is refs/tags/<tag_name>."

I have tested this in #4052 and it seems to work as expected. Once I push a new commit the existing workflow run is cancelled and the runs for the new commits start.

I also put the same into Rayleigh, and it seems to work fine: geodynamics/Rayleigh#279

tjhei:
> github.ref : "The branch or tag ref that triggered the workflow run

I guess it could be the same ref if I use the same branch name... Anyways, not likely.

gassmoeller:
> I guess it could be the same ref if I use the same branch name... Anyways, not likely.

Ah, I see why you were concerned. See #4104.

PR #4102, opened 2021-06-30
Fix a test for certain versions of openmpi

gassmoeller:
This test runs with MPI and starts new MPI jobs inside the shared library it uses. This seems to work for the tester (which uses mpich), but my openmpi systems (and the new tester I am working on for #4052) crash with MPI error messages. I am not sure if that is because one MPI process returns early, or because of the duplicate use of MPI_COMM_WORLD, but it seems like a weird test anyway. There is no reason the outer job should run in parallel, and with the change the test still works as before, because the inner jobs (inside the .cc file) are still started with `mpirun -np 2`.


tjhei (approved):

gassmoeller:
Ah, please wait with merging this PR, it seems to not have fixed the problem in all cases. I will report back when I found the issue.

bangerth:
Can you show the error messages? The patch can't be quite right. The idea of the `if` statement you are removing is that the follow-up job (with two processes) is only started once, from the root process. If you remove the `if`, then each of the two processes will run another `mpirun -np 2` job, which is probably not what you want. I think the existing behavior is correct, so if it doesn't work, we need to find out why not.

gassmoeller:
This is the output of the test:
```
--     . version 2.4.0-pre (fix_test_for_openmpi, 515c85770)
--     . using deal.II 9.3.0-pre (master, d2841aca35)
--     .       with 64 bit indices and vectorization level 2 (256 bits)
--     . using p4est 2.0.0

Loading shared library <./libcheckpoint_05_mpi.so>
* starting from beginning:
Executing the following command:
cd output-checkpoint_05_mpi ; (cat /home/rene/software/aspect/tests/checkpoint_05_mpi.prm  ;  echo 'set Output directory = output1.tmp'  ;  rm -rf output1.tmp ; mkdir output1.tmp ) | mpirun -np 2 ../../aspect -- > /dev/null
system() returned error 256
Executing the following command:
cd output-checkpoint_05_mpi ;  rm -rf output2.tmp ; mkdir output2.tmp ;  cp output1.tmp/restart* output2.tmp/
cp: cannot stat 'output1.tmp/restart*': No such file or directory
system() returned error 256
* now resuming:
Executing the following command:
cd output-checkpoint_05_mpi ; (cat /home/rene/software/aspect/tests/checkpoint_05_mpi.prm  ;  echo 'set Output directory = output2.tmp'  ;  echo 'set Resume computation = true' ) | mpirun -np 2 ../../aspect -- > /dev/null
system() returned error 256
* now comparing:
cp: cannot stat 'output1.tmp/log.txt': No such file or directory
cp: cannot stat 'output2.tmp/log.txt': No such file or directory
cp: cannot stat 'output1.tmp/statistics': No such file or directory
cp: cannot stat 'output2.tmp/statistics': No such file or directory
system() returned error 256
```
I have tried a lot to get more descriptive output than `system() returned error 256`, but I cannot seem to get at the error that is produced inside the system command (I removed the forwarding to /dev/null for example and tried to reroute stderr and stdout to a file, but they stay empty). 

This I tried:
- As soon as I run the commands that are specified in the .cc file in my normal shell, everything works as expected. 
- As soon as I remove the `mpirun` part inside the .cc file, everything works as expected (this is the test checkpoint_01)
- If I replace `mpirun -np 2 ../../aspect ...` by `mpirun -np 2 echo hello` it still crashes with the same error. 

So I think it is an issue with running openmpi inside the system command (maybe only a system command that is called from another MPI progrem, even if the outer MPI program is not actually run in parallel? or an issue with not available environment variables?).

gassmoeller:
I was unable to figure out how to launch an mpirun command inside the `system()` command when using openmpi, but I think this new solution may be more elegant anyway. cmake allows to define execution orders for tests with the `DEPENDS` test property. Instead of having one test that abuses a shared library to run 2 tests (for a total of 3 aspect instances), we could have two normal tests that depend on each other (create checkpoint / resume from checkpoint). The only remaining trick is to copy the restart files from one directory to the other before resuming. Do you think this is an acceptable workaround?

This could be made even better in a future PR by:
 - using cmake test fixtures that guarantee a test is run before another (currently you can specify ctest -R checkpoint_05_mpi_resume and if you havent run the write test before it will crash). But this requires cmake 3.7.
 - rework all of our existing checkpoint tests to the same pattern. This is currently not necessary, but their current structure makes it very hard to understand what goes wrong if they fail.

bangerth:
I think the key to the original error was this: `cp: cannot stat 'output1.tmp/restart*': No such file or directory` It is either that there really is no restart file in this directory, or that whatever shell is being call doesn't want to do wildcard expansion. I don't think it's the latter, so the question to ask is why there is no restart file. One could execute an `ls -lR` in a `system()` call immediately before the call to `mpirun -np2` just to see what actually exists in these directories.

tests/CMakeLists.txt
```diff
@@ -71,6 +71,27 @@ FUNCTION(get_mpi_count _filename)
...
+#    '# DEPENDS-ON: testname'
+# The result is returned in a variable _depends_on in the
+# caller's scope.
+FUNCTION(get_depends_on _filename)
+  FILE(STRINGS ${_filename} _input_lines
+       REGEX "DEPENDS-ON:")
+  IF("${_input_lines}" STREQUAL "")
+    SET(_depends_on "" PARENT_SCOPE)
+  ELSE()
+    # go over the (possibly multiple) lines with MPI markers and choose the last
```
bangerth, line 85:
```suggestion
    # go over the (possibly multiple) lines with DEPENDS_ON markers and choose the last
```

tests/CMakeLists.txt
```diff
@@ -71,6 +71,27 @@ FUNCTION(get_mpi_count _filename)
   ENDIF()
 ENDFUNCTION()
 
+# A function that checks if a test depends on another test.
+# This is encoded in .prm files through lines of the form
+#    '# DEPENDS-ON: testname'
+# The result is returned in a variable _depends_on in the
+# caller's scope.
```
bangerth, line 78: maybe add a line about why one would want to have dependencies?

tests/CMakeLists.txt
```diff
@@ -394,6 +415,13 @@ FOREACH(_test ${_tests})
               -P ${CMAKE_SOURCE_DIR}/run_test.cmake
       WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
     SET_TESTS_PROPERTIES(${_testname} PROPERTIES TIMEOUT ${TEST_TIME_LIMIT})
+
+    # Check if this test depends on another test
+    GET_DEPENDS_ON(${CMAKE_CURRENT_SOURCE_DIR}/${_testname}.prm)
+    IF(NOT "${_depends_on}" STREQUAL "")
+      MESSAGE(STATUS "DEPENDENCY DETECTED. TEST:" ${_testname} "DEPENDENCY:" ${_depends_on})
```
bangerth, line 422: That's a lot of shouting in upper case letters :-)

bangerth (commented):
That's a really elegant solution! Do we document somewhere how we currently deal with the `MPI:` tag? If so, could we also document there how the `DEPENDS_ON:` tag works?

gassmoeller:
I addressed the comments and added some documentation about test properties to the manual.

doc/manual/manual.tex
```diff
@@ -11908,6 +11908,26 @@ \subsubsection{Writing tests}
...
+\subsubsection{Test properties}
+\label{sec:test_properties}
+
+\aspect{}'s test parameter files can contain certain special markers that
+indicate how this test should be executed. Placing such a marker at any
+point in the test parameter file will indicate to the test system what
+to do with this test. The following properties are currently supported:
+
+\begin{itemize}
+  \item "\texttt{\# MPI: x}": Indicates that this test should be executed with
```
bangerth, line 11920:
```suggestion
  \item ``\texttt{\# MPI: x}'': Indicates that this test should be executed with
```

Same below

bangerth (commented):
Thanks. OK if you fix up the latex markup in the three places.

gassmoeller:
Ok, done.

bangerth (approved):

Issue #3338, opened 2020-01-18
remove PETSc support


tjhei:
We currently do not have sufficient test coverage, no CI, and no developer using PETSc instead of Trilinos. In fact, we currently don't compile at all (even after #3337). Should we disable/remove it or is there a compelling reason/demand for improving things?

gassmoeller:
I am not opposed to removing it. As you said it is insufficiently tested and used, and since we are moving toward using the GMG preconditioner more widely it makes sense to remove an unused alternative to the AMG preconditioner (this was the only place where we used PETSc, right? we should confirm this).

tjhei:
> this was the only place where we used PETSc, right? we should confirm this

Well, all linear algebra would use PETSc instead of Trilinos. The most striking different is that the PETSc AMG is different than Trilinos ML, everything else is somewhat similar.

Note that we can not use GMG with PETSc right now (we copy vectors between Trilinos<->deal.II matrix free and that is only implemented for Trilinos).

tjhei:
> I am not opposed to removing it

How about I disable it for now and we decide later if we want to remove the code as well?

gassmoeller:
sounds good

tjhei:
Should we remove petsc support? I think that makes sense at this point.

tjhei:
@gassmoeller starter project?

gassmoeller:
Yes, good idea.

PR #4109, opened 2021-07-01
Removed a PETSC option

rfildes:
This removes an old PETSC option.

### DIFF
--- a/source/simulator/melt.cc
+++ b/source/simulator/melt.cc
@@ -1221,15 +1221,12 @@ namespace aspect
 
       LinearAlgebra::PreconditionAMG preconditioner;
       LinearAlgebra::PreconditionAMG::AdditionalData Amg_data;
-#ifdef ASPECT_USE_PETSC
-      Amg_data.symmetric_operator = false;
-#else
+
       // Amg_data.constant_modes = constant_modes;
       Amg_data.elliptic = true;
       Amg_data.higher_order_elements = false;
       Amg_data.smoother_sweeps = 2;
       Amg_data.aggregation_threshold = 0.02;
-#endif
       preconditioner.initialize(system_matrix.block(block_idx, block_idx));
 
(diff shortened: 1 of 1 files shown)


source/simulator/melt.cc
```diff
@@ -1221,15 +1221,7 @@ namespace aspect
 
       LinearAlgebra::PreconditionAMG preconditioner;
       LinearAlgebra::PreconditionAMG::AdditionalData Amg_data;
-#ifdef ASPECT_USE_PETSC
-      Amg_data.symmetric_operator = false;
-#else
-      // Amg_data.constant_modes = constant_modes;
-      Amg_data.elliptic = true;
-      Amg_data.higher_order_elements = false;
```
bangerth, line 1229: Actually, you want to keep these lines! You want to get rid of the ifdef branch because ASPECT no longer supports PETSc, but we still need the else branch.

bangerth (changes requested):

bangerth:
Other than that, nice job with your first pull request!

bangerth:
Related to #3338.

source/simulator/melt.cc
rfildes: Oh okay, so I should add back in lines 1226 to 1231 ?

rfildes:
I think fixed it, could you take another look?

tjhei:
This is better, but the indentation is slightly off, see https://jenkins.tjhei.info/blue/organizations/jenkins/aspect/detail/PR-4109/6/pipeline
Ask one of us to help you with this. If you have the tool astyle installed, you can run ``make indent`` in the build directory.

rfildes:
Menno helped me install astyle and use that which looks like it fixed the indent I think

bangerth (approved):

bangerth:
Nice job!

PR #4110, opened 2021-07-01
remove PETSC option

elodie-kendall:
This removes and old PETSC option

### DIFF
--- a/source/main.cc
+++ b/source/main.cc
@@ -678,12 +678,6 @@ int main (int argc, char *argv[])
         }
       else if (arg=="-j" || arg =="--threads")
         {
-#ifdef ASPECT_USE_PETSC
-          std::cerr << "Using multiple threads (using -j) is not supported when using PETSc for linear algebra. Exiting." << std::endl;
-          return -1;
-#else
-          use_threads = true;
-#endif
         }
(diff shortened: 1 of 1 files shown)


source/main.cc
```diff
@@ -678,12 +678,6 @@ int main (int argc, char *argv[])
         }
       else if (arg=="-j" || arg =="--threads")
         {
-#ifdef ASPECT_USE_PETSC
-          std::cerr << "Using multiple threads (using -j) is not supported when using PETSc for linear algebra. Exiting." << std::endl;
-          return -1;
-#else
-          use_threads = true;
```
bangerth, line 685: Actually, you want to keep this line! You want to get rid of the `ifdef` branch because ASPECT no longer supports PETSc, but we still need the `else` branch.

bangerth (changes requested):
Nice job with the first PR! Just a small request.

bangerth:
Related to #3338.

bangerth:
@elodie-kendall Any news? 

There is a conflict with this file now. You might want to check whether someone else has already had the change merged that you want to make here.

elodie-kendall:
Rene did this in the mean time so I'll pick another

PR #4112, opened 2021-07-01
Remove unused PETSC option

KiralyAgi:
This removes an old PETSC option

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -996,14 +996,11 @@ namespace aspect
     const Table<2,DoFTools::Coupling> coupling = setup_system_matrix_coupling();
     LinearAlgebra::BlockDynamicSparsityPattern sp;
 
-#ifdef ASPECT_USE_PETSC
-    sp.reinit (introspection.index_sets.system_relevant_partitioning);
-#else
     sp.reinit (system_partitioning,
                system_partitioning,
                introspection.index_sets.system_relevant_partitioning,
                mpi_communicator);
-#endif
+
 
     if ((parameters.use_discontinuous_temperature_discretization) ||
         (parameters.use_discontinuous_composition_discretization) ||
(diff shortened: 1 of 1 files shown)


bangerth (changes requested):
This looks ok, but you have a lot of merge commits in your branch. Do you know how to get rid of these (and, why they are there in the first place)?

bangerth:
Related to #3338.

KiralyAgi:
Hi Wolfgang,

I agree, it is strange that there are so many commits. As far as I see, they all originated from the times I updated ASPECT from the upstream. I usually do git pull upstream and then git push origin master. When I pull from upstream it always creates a commit for merging. But I don't know why is this visible on this branch. Is there a way to get the updates without having a commit for each time?

KiralyAgi:
Okay, solved it! I followed the advice from here: https://stackoverflow.com/questions/33502205/how-to-delete-merge-commits

source/simulator/core.cc
```diff
@@ -1047,17 +1044,9 @@ namespace aspect
...
-#ifdef ASPECT_USE_PETSC
-    SparsityTools::distribute_sparsity_pattern(sp,
-                                               dof_handler.locally_owned_dofs_per_processor(),
-                                               mpi_communicator, introspection.index_sets.system_relevant_set);
 
     sp.compress();
 
-    system_matrix.reinit (system_partitioning, system_partitioning, sp, mpi_communicator);
-#else
-    sp.compress();
```
tjhei, line 1059: Keep this line, please.

KiralyAgi:
I couldn't figure out how to change back 1 deleted line, so I did it all over again, and now it should be updated with the correct lines deleted/not-deleted

tjhei (approved):
Looks good!

PR #4118, opened 2021-07-02
Allow viscosity 0 if Stokes is not solved

### LINKED ISSUE #4116: advection/drop_ev.prm fails
``cd benchmarks; make -f check.mk BUILD=$BUILDDIR advection/`` fails with
```
An error occurred in line <444> of file </home/heister/aspect-git/source/simulator/helper_functions.cc> in function
    double aspect::Simulator<dim>::compute_pressure_scaling_factor() const [with int dim = 2]
The violated condition was: 
    out.viscosities[q] > 0
Additional information: 
    The viscosity needs to be a positive quantity.

Stacktrace:
-----------
#0  /home/heister/aspect-git/build64/aspect: aspect::Simulator<2>::compute_pressure_scaling_factor() const
#1  /home/heister/aspect-git/build64/aspect: aspect::Simulator<2>::start_timestep()
#2  /home/heister/aspect-git/build64/aspect: aspect::Simulator<2>::run()
#3  /home/heister/aspect-git/build64/aspect: void run_simulator<2>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, bool)
#4  /home/heister/aspect-git/build64/aspect: main

```

_Originally posted by @tjhei in https://github.com/geodynamics/aspect/issues/4069#issuecomment-872316216_

gassmoeller:
Should fix #4116. Restores the behavior from before #3681 for models that set viscosity to 0 and dont solve Stokes.

@tjhei: Do you think this is the correct way to fix it? I verified that the model now runs (there are only two models like this, the other one is benchmarks/advection/rotate_shape_*

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -402,13 +402,13 @@ namespace aspect
     // computing the *geometric* mean of the viscosities, but the
     // formula above is numerically stable.
 
-    // Do return 0 if we do not solve the Stokes equation. Technically, there is
+    // Do return NaN if we do not solve the Stokes equation. Technically, there is
     // no harm in computing the factor anyway, but the viscosity has to be positive
-    // for this function to work, and some models may set it to 0 if not solving
-    // the Stokes equation.
+    // for this function to work, and some models may set viscosity to 0 if not solving
+    // the Stokes equation. Returning NaN guarantees this value cannot be used.
     if (parameters.nonlinear_solver == NonlinearSolver::no_Advection_no_Stokes ||
         parameters.nonlinear_solver == NonlinearSolver::single_Advection_no_Stokes)
-      return 0.0;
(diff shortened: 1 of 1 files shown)


tjhei:
I assume pressure scaling is unused in these models? Then we could even return NaN.
Alternatively, 1 might also be a safe option on case it might be used for some other scaling.

gassmoeller:
Yes it is unused. I switched to NaN and tested that the model works now, but crashes if I try to access the pressure_scaling variable. Let's see if any tests use it (they should not, I searched for it).

tjhei (approved):

bangerth:
Nice! I like using the invalid value here, well done!

PR #4119, opened 2021-07-02
Fix crustal cookbook

### LINKED ISSUE #4117: crustal_deformation/crustal_model_2D.prm fails
``cd cookbooks; make -f check.mk BUILD=$BUILDDIR crustal_deformation/`` fails with
```
    The iterative Stokes solver did not converge. It reported the                                                   
    following error:  
                  Additional information:                                                                                         
    Iterative method reported convergence failure in step 5000. The                                                 
    residual in the last step was 1.69553e+15.                                                                      
                                                                                                                    
    The required residual for convergence is: 50735275.339261.
    See output-crustal_model_2D/solver_history.txt for convergence
    history.
                                                                                
```

_Originally posted by @tjhei in https://github.com/geodynamics/aspect/issues/4069#issuecomment-872315319_

gassmoeller:
Fix #4117. The solver restart length alone already fixes the issue of the non-converging linear solver (5000 iterations -> 195 iterations). However, the nonlinear tolerance in this cookbook is so small, that the nonlinear solver very rarely converges for this problem. With a more reasonable tolerance this cookbook is 5 times faster and the results are nearly the same (small visible difference in the shape of the free surface). Let me know if you want me to remove the change to the solver tolerance to keep the cookbook the same.

### DIFF
--- a/cookbooks/crustal_deformation/crustal_model_2D.prm
+++ b/cookbooks/crustal_deformation/crustal_model_2D.prm
@@ -132,5 +132,6 @@ subsection Solver parameters
     set Linear solver tolerance = 1e-9
     set Number of cheap Stokes solver steps = 0
     set Maximum number of expensive Stokes solver steps = 5000
+    set GMRES solver restart length = 200
   end
 end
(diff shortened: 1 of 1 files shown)


tjhei (approved):
Looks good to me. I will merge for 2.3 and @naliboff can do a followup if desired.

PR #4122, opened 2021-07-05
GMG option for additional details

tjhei:


tjhei:
related to #4008

include/aspect/stokes_matrix_free.h
```diff
@@ -530,6 +530,8 @@ namespace aspect
 
       Simulator<dim> &sim;
 
+      bool options_print_details;
```
bangerth, line 533: This needs a comment. What does the `options_` prefix indicate?

source/simulator/stokes_matrix_free.cc
```diff
@@ -1242,15 +1242,34 @@ namespace aspect
...
   {
+    prm.enter_subsection ("Solver parameters");
+    prm.enter_subsection ("Matrix Free");
+    {
+      prm.declare_entry ("Output details", "false",
+                         Patterns::Bool(),
+                         "Turns on extra information for the matrix free GMG solver to be printed.");
+
+
+    }
```
bangerth, line 1255: no need for the empty lines

source/simulator/stokes_matrix_free.cc
```diff
@@ -1766,6 +1787,12 @@ namespace aspect
 
         mg_smoother_A[level].estimate_eigenvalues(temp_velocity);
         mg_smoother_Schur[level].estimate_eigenvalues(temp_pressure);
+
+        if (level==0)
+          {
+            coarse_A_size = temp_velocity.size();
+            coarse_S_size = temp_pressure.size();
+          }
```
bangerth, line 1795: This is unrelated, right? What is its effect?

bangerth (changes requested):

source/simulator/stokes_matrix_free.cc
tjhei: This is how I compute the size of the coarse vectors to print below.

include/aspect/stokes_matrix_free.h
tjhei: Not sure, that it is an option? :-) I removed it.

tjhei:
updated.

include/aspect/stokes_matrix_free.h
```diff
@@ -530,6 +530,8 @@ namespace aspect
 
       Simulator<dim> &sim;
 
+      bool print_details;
```
bangerth, line 533: This variable has no documentation, but that is unfortunately the case also for everything below :-(

source/simulator/parameters.cc
```diff
@@ -2103,6 +2104,7 @@ namespace aspect
     Parameters<dim>::declare_parameters (prm);
     Melt::Parameters<dim>::declare_parameters (prm);
     Newton::Parameters::declare_parameters (prm);
+    StokesMatrixFreeHandler<dim>::declare_parameters (prm);
```
bangerth, line 2107: Ha, woops :-)

bangerth (approved):
I'm not super excited about the prospect of every class having their personal `print_details` variable, but I'm also not going to stand in the way of this patch :-)

tjhei:
> I'm not super excited about the prospect of every class having their personal `print_details` variable, but I'm also not going to stand in the way of this patch :-)

We could have a global section for a log verbosity or individual flags as proposed in #4008 if you prefer...

bangerth:
But then you have to annotate *all* places where you do output if you also want to support `quiet`, for example. You can only avoid annotating all places if you only have two levels: "normal" and "detailed", in which case you only have to annotate the "detailed" places. I could be talked into this. Making it finer-grained just looks too complicated.

PR #4124, opened 2021-07-06
Add option to force color output for ninja.

MFraters:
It has bothered me for a while that when using ninja you have to specifically add a flag to get color output. This pull request would automatically add that unless you turn it off manually. I don't see a downside of having the flag there, but if there is some, I will be happy to turn it off by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,8 @@ ENDIF()
 DEAL_II_INITIALIZE_CACHED_VARIABLES()
 PROJECT(${TARGET} CXX)
 
+SET(FORCE_COLORED_OUTPUT ON CACHE BOOL "Forces colored ouput when compiling with gcc and clang.")
+
 SET(ASPECT_WITH_WORLD_BUILDER ON CACHE BOOL "Whether to enable compiling aspect with the Geodynamic World Builder.")
 MESSAGE(STATUS "ASPECT_WITH_WORLD_BUILDER = '${ASPECT_WITH_WORLD_BUILDER}'")
 if(ASPECT_WITH_WORLD_BUILDER)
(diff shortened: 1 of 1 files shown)


tjhei:
Neat. What does it have to do with ninja? Does this not work with make?

MFraters:
I don't think make has the same issue. For me make outputs color without these flags.

tjhei:
Ok, can you test quickly that it also works with make still?

tjhei (approved):

MFraters:
> Ok, can you test quickly that it also works with make still?

Yep, tested it. No issues for me with make. I am also making the same pull request for the world builder at https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/307, and it seems to work fine on all the different systems and compilers I am testing for (mostly with make actually).

CMakeLists.txt
```diff
@@ -454,6 +456,16 @@ IF (ASPECT_HAVE_LINK_H)
   MESSAGE(STATUS "Enabling checking of compatible deal.II library when loading plugins")
 ENDIF()
 
+if (${FORCE_COLORED_OUTPUT})
+  if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER MATCHES "AppleClang")
+    STRING(APPEND DEAL_II_CXX_FLAGS_DEBUG " -fcolor-diagnostics")
+    STRING(APPEND DEAL_II_CXX_FLAGS_RELEASE " -fcolor-diagnostics")
+  elseif(CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+    STRING(APPEND DEAL_II_CXX_FLAGS_DEBUG " -fdiagnostics-color=always")
+    STRING(APPEND DEAL_II_CXX_FLAGS_RELEASE " -fdiagnostics-color=always")
```
bangerth, line 465: Do we know that these flags are understood by all compiler versions we support?

tjhei:
Alright, we will see what happens!

MFraters:
wow, a lot of daring decisions are being made today :P 

I can't find exact details, but I can find several references to it being part of gcc 4.9. I also can't really find it for clang, but it is at least supported in 3.8.0 (https://releases.llvm.org/3.8.0/tools/clang/docs/UsersManual.html)

PR #4126, opened 2021-07-07
Subduction cookbook case1

anne-glerum:
This PR contributes the first case of 4 of a subduction benchmark by Quinquis et al as a new cookbook. 

Question:
The benchmark requires some new statistics plugins and I'm wondering whether they are generally useful, or should just stay with the cookbook. It concerns plugins that track the deepest depth of a given compositional field (in this case this represents the slab tip depth), the RMS velocity of one or more compositional fields (which here make up the subducting plate) and the viscous dissipation of the whole domain and the slab only. I haven't added the plugins here yet. 

### For all pull requests:


MFraters:
Hey Anne,

Very cool! I can take a look at this today.

As for the question, both the first and second plugin actually seem very useful to me for general subduction models. The third one (viscous dissipation), I would personally not really know what to use for yet, but I would not be against adding it as a statistic plugin since it might be useful for other cases as well.

anne-glerum:
Ok, then I will contribute those as separate PRs after making them more generally applicable.

MFraters:
I don't know how much extra would it would be, but for the slab tip depth plugin, it could actually be useful to get also the min and maximum surface coordinates besides the depth.

bangerth:
A statistics plugin for the global viscous dissipation might be interesting in its own right. I've often wondered how much the heating of the entire mantle is due to viscous dissipation and how much due to radioactive decay. It would be nice to have a way to compare the two :-)

doc/manual/manual.bib
```diff
@@ -11939,3 +11939,10 @@ @software{fraters_menno_2020_3900603
   doi          = {10.5281/zenodo.3900603},
   url          = {https://doi.org/10.5281/zenodo.3900603}
 }
+
+@phdthesis{Quinquis2014,
+author = {Quinquis, M},
+school = {Charles University},
+title = {{A numerical study of subduction zone dynamics using linear viscous to thermo-mechanical model setups including (de)hydration processes}},
+year = {2014}
+}
```
bangerth, line 11948: I think you can drop this now, given that #4125 is merged.

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
 \end{itemize}
 
 
+\subsubsection{Kinematically-driven 2d oceanic subduction}
+\label{sec:cookbooks-kinematic-2d-oceanic-subduction}
 
+\textit{This section was contributed by Anne Glerum.}
+
+This subduction example is based on the benchmark effort of Quinquis et al., of which initial results were published in \cite{Quinquis2014}. In 4 increasingly complex setups we will go from isoviscous materials without any temperature effects and to a fully thermo-mechanical, nonlinear, strain-weakened visco-plastic, externally-driven model of oceanic subduction. The setup of the most complex case is outlined in Fig.~\ref{fig:QQ_setup}. The models are run for 15 My and slab tip depth, trench location, RMS velocity and temperature, and viscous dissipation are compared between different codes. In addition, we discuss the effects of the element size of the subduction interface and crustal layers, viscosity averaging and the solver tolerance.
```
bangerth, line 7081:
```suggestion
This subduction example is based on the benchmark effort of Quinquis et al., of which initial results were published in \cite{Quinquis2014}. In four increasingly complex setups we will go from isoviscous materials without any temperature effects to a fully thermo-mechanical, nonlinear, strain-weakened visco-plastic, externally-driven model of oceanic subduction. The setup of the most complex case is outlined in Fig.~\ref{fig:QQ_setup}. The models are run for 15 My and slab tip depth, trench location, RMS velocity and temperature, and viscous dissipation are compared between different codes. In addition, we discuss the effects of the element size of the subduction interface and crustal layers, viscosity averaging and the solver tolerance.
```

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
...
+
+\begin{figure}
+    \centering
+    \includegraphics[width=0.9\textwidth]{cookbooks/kinematically_driven_subduction_2d_case1/doc/setup_Quinquis2014.png}
+    \caption{Case 4 model setup. Copied from Quinquis (2014).}
```
bangerth, line 7086: We always mark up captions in italics:
```suggestion
    \caption{\it Case 4 model setup. Copied from Quinquis (2014).}
```

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
...
+    \label{fig:QQ_setup}
+\end{figure}
+
+\paragraph{Case 1 - Simple geometry and rheology}
```
bangerth, line 7090:
```suggestion
\paragraph{Case 1: Simple geometry and rheology}
```
Maybe similarly for the other cases.

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
...
+The Case 1 model setup considers 3 materials (compositional fields) apart from the background sublithospheric mantle (see Fig.~\ref{fig:QQ_case1_setup}):
```
bangerth, line 7091:
```suggestion
The Case 1 model setup considers three materials (compositional fields) apart from the background sublithospheric mantle (see Fig.~\ref{fig:QQ_case1_setup}): 
```

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
...
+\begin{enumerate}
+    \item the lithosphere of the overriding plate (combining the BOC, SHB and thermal layer of the overriding plate in Fig.~\ref{fig:QQ_setup}),
+    \item the crust of the subducting plate (weak seed and BOC combined), and
+    \item the mantle lithosphere of the subducting plate (SHB and thermal layer combined).
+\end{enumerate}{}
+The geometry of these compositions is implemented as follows:
+\lstinputlisting[language=prmfile]{cookbooks/kinematically_driven_subduction_2d_case1/Case1_compositions.prm}
+
+No differences in material properties exist, except for density and viscosity, so we use the multicomponent material model. To keep the boundaries between fields as sharp as possible in terms of viscosity, we use the maximum composition to determine the viscosity in each evaluation point (note that this can be harder on the solver):
+\lstinputlisting[language=prmfile]{cookbooks/kinematically_driven_subduction_2d_case1/Case1_materialmodel.prm}
```
bangerth, line 7101: I think you want the marked up version:
```suggestion
\lstinputlisting[language=prmfile]{cookbooks/kinematically_driven_subduction_2d_case1/Case1_materialmodel.prm.out}
```

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
...
+
+Temperature effects are ignored. Subduction is driven by prescribed in- and outflow through the right boundary (with a gradual transition of the flow direction), all other boundaries are free slip. The volume of material that flows in is balanced by the volume that is prescribed to flow out (this is important as the model is incompressible). A weak crust along the plate interface and the subducting lithosphere facilitates subduction. Through the function plugin, we prescribe the in- and outflow:
+\lstinputlisting[language=prmfile]{cookbooks/kinematically_driven_subduction_2d_case1/Case1_velocity.prm}
```
bangerth, line 7104: same here and below

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
...
+
+To follow the slab on its descent into the mantle, we use adaptive mesh refinement based on viscosity i.c.w. the minimum refinement strategy to ensure sufficient resolution in the crust at the surface and the weak zone that allow the slab to detach from the surface:
```
bangerth, line 7106: what does "i.c.w." mean? In conjunction with? This is not a common abbreviation in American English, I believe.

bangerth:
```suggestion
To follow the slab on its descent into the mantle, we use adaptive mesh refinement based on viscosity i.c.w. the minimum refinement strategy to ensure sufficient resolution in the crust at the surface and the weak zone that allows the slab to detach from the surface:
```

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
...
+
+To monitor the model evolution, several diagnostic quantities are tracked over time: the depth of the tip of the slab, the position of the trench, the RMS velocity of the slab and the whole model domain, and the work done (viscous dissipation) in the slab and total model domain. The computation of these quantities in done in several new plugins:
+\lstinputlisting[language=prmfile]{cookbooks/kinematically_driven_subduction_2d_case1/Case1_postprocessing.prm}
+
+\begin{figure}
+    \centering
+    \includegraphics[width=0.6\textwidth,trim=1cm 7cm 1cm 8cm, clip=true]{cookbooks/kinematically_driven_subduction_2d_case1/doc/Case1_dens_t0.png}
+    \includegraphics[width=0.6\textwidth,trim=1cm 7cm 1cm 8cm, clip=true]{cookbooks/kinematically_driven_subduction_2d_case1/doc/Case1_visc_t0.png}
+    \includegraphics[width=0.6\textwidth,trim=1cm 7cm 1cm 8cm, clip=true]{cookbooks/kinematically_driven_subduction_2d_case1/doc/Case1_vel_t0.png}
+    \caption{Case 1 density, viscosity and velocity at time zero.}
```
bangerth, line 7117:
```suggestion
    \caption{\it Case 1 density, viscosity and velocity at time zero.}
```

doc/manual/manual.tex
```diff
@@ -7073,7 +7073,68 @@ \subsubsection{Melt migration in a 2D mid-ocean ridge model}
...
+    \label{fig:QQ_case1_setup}
+\end{figure}
+
+We run the Case 1 setup for 15 My of model time. The diagnostic quantities in Fig.~\ref{fig:QQ_case1_diagnostics} show three stages of model evolution: first trench advance (top right plot), then free subduction (increasing slab RMS velocity), and after about 13 My interaction between the slab and bottom boundary at 660 km depth, which slows down the slab. The slab then curves inward along the bottom boundary. This can also be seen in Fig~\ref{fig:QQ_case1_results}.
+
+\begin{figure}
+    \centering
+    \includegraphics[width=.9\textwidth]{cookbooks/kinematically_driven_subduction_2d_case1/doc/Case1_diagnostics}
+    \caption{Case 1 diagnostic quantities of ASPECT, Sulec and Elefant results.}
```
bangerth, line 7126: same here and again a few lines down

bobmyhill:
+1 for all the plugins :) ... especially the viscous dissipation one.

A couple of thoughts:
For compressible models where materials can take fractions other than 0 or 1, it will be necessary to think about how the RMS velocity is calculated (i.e. RMS by mass or by volume?). For viscous dissipation, we'll need to think about what to include and what to exclude (grain size decrease and elasticity both store energy, for example).

doc/manual/manual.tex
bangerth: Are you really going to compare results between different codes? Or, if not, maybe change the wording to say that the comparison is in Quinquis2014

bangerth (commented):
Always fun to read through new cookbooks :)

doc/manual/manual.bib
anne-glerum: #4125 added the reference to the citation bib that is used to create the publication list on the website. In manual.tex, manual.bib is used instead of this citation bib. Is the reference nevertheless available?

bangerth: Ah no, I just didn't pay attention.

anne-glerum:
TODO: update the names of the postprocessors in the input file(s) once they are merged.

bangerth:
Do I understand correctly that this is ready to be merged and that the last comment with the TODO is a note for the future and not something left to do here?

anne-glerum:
There is one more postprocessor that I could contribute that was needed for the benchmark, one that computes the rms velocity per compositional field instead of over the whole domain. If it's not of general use, maybe I can add it as a shared library to the cookbook?

I will have to change the names of the postprocessors in the prm file in this PR in any case.

gassmoeller:
> one that computes the rms velocity per compositional field instead of over the whole domain

That sounds pretty specific to me. I am fine either way (included in cookbook or full postprocessor), but adding the postprocessor to this PR as shared library is probably the easier way.

anne-glerum:
Alright, I've added the postprocessor in the cookbook and fixed the postprocessor names in the prm files. Ready from my side.

gassmoeller:
Hi Anne, the cookbook looks good. Please add a changelog entry, and take a look at the changes I made in https://github.com/geodynamics/aspect/pull/4126/commits/444ef34a3c102a530bb98e91e81243a0ce566802 and correct anything I might have broken. Let me know when you are done, then we can merge this.

gassmoeller (approved):

PR #4130, opened 2021-07-07
Postprocessor to compute the maximum depth of each compositional field

anne-glerum:
One of the postprocessors for #4126. It just computes the maximum depth of each of the compositional fields. 

### For all pull requests:


source/postprocess/max_depth_field.cc
```diff
@@ -0,0 +1,131 @@
...
+                               this->get_fe(),
+                               quadrature_formula,
+                               update_values   |
+                               update_quadrature_points |
+                               update_JxW_values);
+
+      std::vector<double> compositional_values(n_q_points);
+      std::vector<Point<dim> > position_values(n_q_points);
+
+      std::vector<double> local_max_depth(this->n_compositional_fields(), std::numeric_limits<double>::min());
```
bangerth, line 58: It is confusingly named, but `std::numeric_limits<double>::min()` is actually `1e-16` or thereabouts -- the smallest *positive* number representable in floating point, see https://en.cppreference.com/w/cpp/types/numeric_limits/min You probably want to use 
```suggestion
      std::vector<double> local_max_depth(this->n_compositional_fields(), std::numeric_limits<double>::lowest());
```
See also #4135.

source/postprocess/max_depth_field.cc
```diff
@@ -0,0 +1,131 @@
...
+      const QGauss<dim> quadrature_formula(this->get_fe().base_element(this->introspection().base_elements.compositional_fields).degree + 1);
+
+      const unsigned int n_q_points = quadrature_formula.size();
+
+      FEValues<dim> fe_values (this->get_mapping(),
+                               this->get_fe(),
+                               quadrature_formula,
+                               update_values   |
+                               update_quadrature_points |
+                               update_JxW_values);
```
bangerth, line 53: You don't actually need the JxW values -- just drop them from the list

source/postprocess/max_depth_field.cc
```diff
@@ -0,0 +1,131 @@
...
+      std::vector<double> global_max_depth(this->n_compositional_fields());
+      Utilities::MPI::max (local_max_depth,
+                           this->get_mpi_communicator(),
+                           global_max_depth);
+
+      // finally produce something for the statistics file
+      for (unsigned int c = 0; c < this->n_compositional_fields(); ++c)
+        {
+          statistics.add_value("Max depth [km] for composition " + this->introspection().name_for_compositional_index(c),
+                               global_max_depth[c] / 1000.0);
```
bangerth, line 90: Why use km as units here? Do we do this elsewhere as well? I think most of the time we just use SI units (except when instructed we use years instead of seconds).

source/postprocess/max_depth_field.cc
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+      // also make sure that the other columns filled by the this object
```
bangerth, line 93:
```suggestion
      // also make sure that the other columns filled by this object
```
(I fixed the other places where you have this from in #4136.)

source/postprocess/max_depth_field.cc
```diff
@@ -0,0 +1,131 @@
...
+
+// explicit instantiations
+namespace aspect
+{
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(MaxDepthField,
+                                  "maximum depth of field",
+                                  "A postprocessor that outputs the biggest depth "
+                                  "of the given field or of several given fields.")
```
bangerth, line 129: Would you mind how this is defined? Namely, the largest depth at which one can find a quadrature point where that field has a value of 0.5 or larger.

I would say "compositional field" rather than "given field".

source/postprocess/max_depth_field.cc
```diff
@@ -0,0 +1,131 @@
...
+
+
+// explicit instantiations
+namespace aspect
+{
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(MaxDepthField,
+                                  "maximum depth of field",
+                                  "A postprocessor that outputs the biggest depth "
```
bangerth, line 128:
```suggestion
                                  "A postprocessor that outputs the largest depth "
```

include/aspect/postprocess/max_depth_field.h
```diff
@@ -0,0 +1,52 @@
...
+         * compositional field indicated by the user.
+         **/
+        std::pair<std::string,std::string>
+        execute (TableHandler &statistics) override;
+    };
+  }
+}
+
+
+#endif
```
tjhei, line 52: Please delete the .swp file you added

anne-glerum:
Thanks for the reviews, I've addressed the comments.

source/postprocess/max_depth_field.cc
```diff
@@ -0,0 +1,134 @@
...
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(MaxDepthField,
+                                  "maximum depth of field",
+                                  "A postprocessor that for each compositional field "
+                                  "outputs the largest depth at which "
+                                  "a quadrature point is found where the field "
+                                  "has a value of 0.5 or larger. For fields that do not "
+                                  "represent materials, but for example track a certain quantity "
+                                  "like strain, this value of 0.5 does not necessarily make sense. ")
```
bangerth, line 132: :-)

PR #4132, opened 2021-07-07
Add steady state heat flux termination criteria

jdannberg:
This pull request add a termination criterion based on the steady state heat flux. 


include/aspect/termination_criteria/steady_heat_flux.h
```diff
@@ -0,0 +1,98 @@
+/*
+  Copyright (C) 2019 by the authors of the ASPECT code.
```
bangerth, line 2:
```suggestion
  Copyright (C) 2021 by the authors of the ASPECT code.
```

naliboff, line 2:
```suggestion
  Copyright (C) 2021 by the authors of the ASPECT code.
```

include/aspect/termination_criteria/steady_heat_flux.h
```diff
@@ -0,0 +1,98 @@
...
+        /**
+         * Read the parameters this class declares from the parameter file.
+         */
+        void
+        parse_parameters (ParameterHandler &prm) override;
+
+      private:
+        double                                  necessary_time_in_steady_state;
+        double                                  allowed_relative_deviation;
+
```
naliboff, line 80: Add brief documentation for the above two parameters?

source/termination_criteria/steady_heat_flux.cc
```diff
@@ -0,0 +1,212 @@
+/*
+  Copyright (C) 2019 - 2020 by the authors of the ASPECT code.
```
naliboff, line 2:
```suggestion
  Copyright (C) 2021  by the authors of the ASPECT code.
```

source/termination_criteria/steady_heat_flux.cc
```diff
@@ -0,0 +1,212 @@
...
+  namespace TerminationCriteria
+  {
+    namespace internal
+    {
+      void trim_time_heat_flux_list (const double necessary_time_in_steady_state,
+                                     std::list<std::pair<double, double> > &time_heat_flux_list)
+      {
+        // Remove old times until we're at the correct time period
+        // but ensure at least two entries remain in the list (one old, one current timestep)
+        auto it = time_heat_flux_list.begin();
```
naliboff, line 35: Out of curiosity, does `it` refer to "iterator time"?

include/aspect/termination_criteria/steady_heat_flux.h
```diff
@@ -0,0 +1,98 @@
...
+         */
+        std::set<types::boundary_id> boundary_indicators;
+
+        /**
+         * A list of pairs (time, heat flux) that we have computed at
+         * previous time steps. This is used to determine when we have reached
+         * steady state.
+         */
+        std::list<std::pair<double, double> >   time_heat_flux;
+
```
bangerth, line 93:
```suggestion
```

include/aspect/termination_criteria/steady_heat_flux.h
```diff
@@ -0,0 +1,98 @@
...
+    namespace internal
+    {
+      /**
+       * A function that trims the handed over list and removes all entries from the front that are
+       * further back in time measured from the last entry than given by the first argument.
+       * Additionally it makes sure to always keep two entries in the list, if the list had
+       * two or more entries. Otherwise the function does not change the list.
+       */
+      void trim_time_heat_flux_list (const double necessary_time_in_steady_state,
+                                     std::list<std::pair<double, double> > &time_heat_flux_list);
```
bangerth, line 42: This function should just live in the `.cc` file. See #4133.

source/termination_criteria/steady_heat_flux.cc
bangerth, line 2:
```suggestion
  Copyright (C) 2021 by the authors of the ASPECT code.
```

source/termination_criteria/steady_heat_flux.cc
```diff
@@ -0,0 +1,212 @@
...
+#include <aspect/termination_criteria/steady_heat_flux.h>
+#include <aspect/postprocess/heat_flux_map.h>
+
+namespace aspect
+{
+  namespace TerminationCriteria
+  {
+    namespace internal
+    {
+      void trim_time_heat_flux_list (const double necessary_time_in_steady_state,
```
bangerth, line 30: just move the documentation of the function here and make the namespace anonymous

source/termination_criteria/steady_heat_flux.cc
```diff
@@ -0,0 +1,212 @@
...
+
+        time_heat_flux_list.erase(time_heat_flux_list.begin(), it);
+      }
+    }
+
+    template <int dim>
+    bool
+    SteadyHeatFlux<dim>::execute()
+    {
+      std::vector<std::vector<std::pair<double, double> > > heat_flux_and_area =
```
bangerth, line 48:
```suggestion
      const std::vector<std::vector<std::pair<double, double> > > heat_flux_and_area =
```

source/termination_criteria/steady_heat_flux.cc
```diff
@@ -0,0 +1,212 @@
...
+
+      // If the length of the simulation time covered in the list is shorter than the
+      // specified parameter, we must continue the simulation
+      if ((time_heat_flux.size() <= 2)
+          ||
+          (time_heat_flux.back().first - time_heat_flux.front().first < necessary_time_in_steady_state))
+        return false;
+
+      // Remove old entries outside of current time window
+      internal::trim_time_heat_flux_list(necessary_time_in_steady_state,time_heat_flux);
```
bangerth, line 78:
```suggestion
      trim_time_heat_flux_list(necessary_time_in_steady_state,time_heat_flux);
```

source/termination_criteria/steady_heat_flux.cc
```diff
@@ -0,0 +1,212 @@
...
+          flux_min = std::min(flux_min, (*it).second);
+          flux_max = std::max(flux_max, (*it).second);
+          flux_sum += (((*it).second + flux_prev)/2.0)*((*it).first-time_prev);
+          time_prev = (*it).first;
+          flux_prev = (*it).second;
+        }
+
+      flux_mean = flux_sum/(time_heat_flux.back().first-time_heat_flux.front().first);
+
+      // If the min and max are within the acceptable deviation of the mean,
```
naliboff, line 96: A general question about the approximations being made here: if `time_heat_flux` only contains two entries, that would mean the steady state heat flux is being computed from adjacent time steps. 

Is this reasonable? Or would this only happen on the second time step? 

I'm trying to remember how I used to determine when steady was reached and I recall using quite a few time steps. Again, likely a mute point.

bangerth (commented):
Just small stuff.

source/termination_criteria/steady_heat_flux.cc
```diff
@@ -0,0 +1,212 @@
...
+      // we are in steady state and return true, otherwise return false
+      deviation_max = std::max(flux_mean - flux_min, flux_max - flux_mean);
+
+      AssertThrow(std::abs(flux_mean) > std::numeric_limits<double>::min(),
+                  ExcMessage("The average heat flux is close to 0.0. The "
+                             "'steady state heat flux' plugin can not compute a "
+                             "relative deviation of heat flux in this case."));
+
+      if (deviation_max/flux_mean > allowed_relative_deviation)
+        return false;
```
naliboff, line 106: Just to be sure, `return true` will not be reached if the above condition is met (e.g., even though it is not within the conditional statement)?

naliboff (commented):
@jdannberg - Nice! A few comments and suggestions after a quick look, but I'm sure others will weigh in as well.

naliboff:
Ah, @bangerth beat me to it 😄 (and also provide a thorough review 😉 )

bangerth:
Two reviews is better than one. And you asked the more substantive questions -- I just checked what is different from the existing plugin :-)

source/termination_criteria/steady_heat_flux.cc
jdannberg: I think it just stands for iterator, but I only copied this from the steady state temperature plugin.

source/termination_criteria/steady_heat_flux.cc
jdannberg: Yes, that's right. But I would say, in that case, the user should pick a different "Time in steady state" in the input file, or set a minimum time step that is smaller than the "Time in steady state". If the time steps are substantially smaller than that, there should always be enough entries in the `time_heat_flux` vector.

source/termination_criteria/steady_heat_flux.cc
jdannberg: As soon as the `return false` is reached, the function will return the value false and no other code in the function will be executed. Was that the question?
If you want I can change this to 
```
      if (deviation_max/flux_mean > allowed_relative_deviation)
        return false;
      else
        return true;
```
I didn't think about what would be easier to read, I just copied it from an existing plugin.

jdannberg:
Okay, thank you! I think I've addressed all comments.

bangerth (commented):
OK from my side, but I'll let @naliboff see whether he's satisfied as well.

bangerth (approved):

source/termination_criteria/steady_heat_flux.cc
naliboff: No fine as is, just wanted to make sure I understood. Thanks for the explanation!

source/termination_criteria/steady_heat_flux.cc
naliboff: @jdannberg - Makes sense, thanks for the explanation! Could you add some of the above text either at the top of the test case prm file and /or in the documentation for the "Time in steady state" parameter? You could almost copy this verbatim into the prm file. Other than that, ready to be merged!

jdannberg: I have added a few lines to the documentation of the input parameter. I thought that would be the place where it is most useful for the user.

jdannberg:
@naliboff Can you have another look?

source/termination_criteria/steady_heat_flux.cc
```diff
@@ -135,6 +135,12 @@ namespace aspect
                              Patterns::Double (0.),
                              "The minimum length of simulation time that the system "
                              "should be in steady state before termination. "
+                             "Note that if the time step size is similar to or larger than "
+                             "this value, the termination criterion will only have very few "
+                             "(in the most extreme case, just two) heat flux values to check. "
+                             "To ensure that a larger number of time steps are included in "
+                             "the check for steady state, this value should be much larger "
+                             "than the time step size. "
```
naliboff, line 143: @jdannberg - Perfect, thank you!

naliboff (approved):
Looks great, thank you for the additional documentation and adding this feature!

PR #4133, opened 2021-07-07
Make a function purely internal to a .cc file.

bangerth:
There is no need to have this function declared in the .h file and exported from the object file.

/rebuild


MFraters (approved):
Looks good to me.

PR #4137, opened 2021-07-08
Add gmg assert throw for direction at parse parameters phase.

MFraters:
I came across this as an assert in the `setup_dofs`, but I think it would be better to have this as an assert throw at the beginning of the program.


tjhei:
Makes sense, thank you.

Would it be possible to phrase the error somehow like:
... only works if all velocity components are fixed, so no [XYZ] can be specified...

MFraters:
sure, I change it :)

bangerth (approved):

PR #4143, opened 2021-07-08
change lower limit quad increase

cedrict:
Although we nearly always want to have a higher quadrature in the elements in the gravity post-processor it would be nice to allow the point mass approach. The default number of quad pts in the pp is 2^3 so that using (2+(-1))^3=1 point would do just that. For large distances between measurement point and element this approximation is justified. I make use of this in the first gravity cookbook about to be submitted.


bobmyhill:
It looks like we don't currently have any tests that explicitly set `Quadrature degree increase`. Could you make a test where the value is set to -1?

tests/gravity_quadrature_degree_increase/screen-output
```diff
@@ -0,0 +1,42 @@
...
+| Postprocessing                  |         1 |   0.00818s |      0.66% |
+| Setup dof systems               |         1 |     0.116s |       9.4% |
+| Setup initial conditions        |         1 |     0.132s |        11% |
+| Setup matrices                  |         1 |    0.0134s |       1.1% |
++---------------------------------+-----------+------------+------------+
+
+-----------------------------------------------------------------------------
+-- For information on how to cite ASPECT, see:
+--   https://aspect.geodynamics.org/citing.html?ver=2.4.0-pre&sha=55a563b75&src=code
+-----------------------------------------------------------------------------
```
bobmyhill, line 42: Could you delete the version/wallclock/blurb lines (basically, any lines which start with `--`, `|` or `+`)?
Alternatively, if you ran your test with ctest, output files with the correct format are stored as 
`tests/output-<name-of-test>/<name-of-file>.notime`

tests/gravity_quadrature_degree_increase/Testing/Temporary/LastTest.log
```diff
@@ -0,0 +1,3 @@
+Start testing: Jul 09 16:46 CEST
+----------------------------------------------------------
+End testing: Jul 09 16:46 CEST
```
bobmyhill, line 3: Could you delete this file?

tests/gravity_quadrature_degree_increase/Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
bobmyhill, line 1: Could you delete this file?

tests/gravity_quadrature_degree_increase.prm
```diff
@@ -0,0 +1,73 @@
...
+
+# Material model
+subsection Material model
+  set Model name = simple
+  subsection Simple model
+    set Reference density                 = 1e6
+  end
+end
+
+# Model boundary temperature
```
bobmyhill, line 34: Can you remove the "Boundary temperature model" section? It is currently ignored by ASPECT anyway, and will soon throw an error (when #4162 is merged).

bobmyhill (changes requested):
Brilliant, thank you! Just a few changes and then this is good to go.

tests/gravity_quadrature_degree_increase.prm
cedrict: hopefully correctly addressed all points

bobmyhill:
Ok, this looks good to go after you patch the test output.

cedrict:
applied diff :)

bobmyhill (approved):

PR #4103, opened 2021-06-30
Switch to deal.II 9.3 tester

gassmoeller:
First stab at setting up a 9.3 tester.

In order to make this work for github actions we need a tester with OpenMPI (because MPICH seems to require compile-time configuration for oversubscription). I set up the dockerfile to build from a clean ubuntu, because this way all information about the test image is inside the repo (and this way I had control over the build variables), but we can discuss this.

Test results are very similar, the only large relative variations are small numbers (heatflux and net rotation mostly).

PR #3898, opened 2020-11-19
CI: add clang-tidy [WIP]

tjhei:
Add a clang-tidy check like we do for deal.II.

PR #4144, opened 2021-07-08
Add clang-tidy option to CMakeLists.txt.

MFraters:
This pull request would add an option to cmake to check the code with clang tidy, and even has the option to automatically fix a lot of the issues it finds which you can turn on and of with this script. Both options are off by default. I don't know if that is something we would want in ASPECT, but I found it to be useful for the world builder I would like to at least offer it for consideration. 

If we decide we would want something like this in aspect, we can take a look at the precise flags which would be relevant for aspect. These are just the flags I use for the world builder, but I don't think they would all fully fit with the desired style of aspect.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -121,6 +121,23 @@ PROJECT(${TARGET} CXX)
 
 SET(FORCE_COLORED_OUTPUT ON CACHE BOOL "Forces colored ouput when compiling with gcc and clang.")
 
+# determine whether to use clang-tidy
+set(ASPECT_CHECK_CLANG_TIDY OFF CACHE BOOL "Whether or not to check the code with clang-tidy while compiling the world builder.")
+
+if(ASPECT_CHECK_CLANG_TIDY)
+  set(ASPECT_CHECK_CLANG_TIDY_CHECKS "clang-analyzer-*,performance-*,modernize-a*,modernize-c*,modernize-c*,modernize-l*,modernize-m*,modernize-p*,modernize-r*\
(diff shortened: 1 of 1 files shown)


tjhei:
We already have .clang-tidy files in the repo that specify the check.
We also have aspect/contrib/utilities/run_clang_tidy.sh.
What advantage does your approach have?

bangerth:
Is our code base even clean with regard to these flags at the moment? If not, maybe we can start by creating patches to this end.

Other than that, we could also make that a CI step rather than hooking it into the cmake configuration?

gassmoeller:
There is still #3898, Timo do you plan to revive that at some point? I can certainly integrate clang/clang-tidy into the new tester in #4103. I am also voting for a CI step instead of (or in addition to) the cmake option.

MFraters:
Ah, sorry. I totally forgot, and did not check, that there was already a clang tidy. Sorry about that. Integrating it into the ci would be nice. 

The advantage of the cmake option is that you can set it to fix some stuff automatically (I didn't see that option in the script), and you can set it to automacally check every time you compile. The advantage of he script is that it is run outside the compilation step.  I don't have preference for the way it is done, and I am happy to close this if we decide to go for the script option.

tjhei:
> you can set it to fix some stuff automatically

I should add that to the script. I used to just change that locally...



> Timo do you plan to revive that at some point?

I can revive it, but we can talk about what the better option is...

bangerth:
Anyone want to revive this?

PR #4148, opened 2021-07-08
Created multicomponent EoS->Material averaging function 

bobmyhill:
This PR adds a function that does multicomponent EquationOfState property averaging. 
This function is stored in the equation of state interface, in the MaterialModel namespace.

### DIFF
--- a/include/aspect/material_model/equation_of_state/interface.h
+++ b/include/aspect/material_model/equation_of_state/interface.h
@@ -23,6 +23,7 @@
 
 #include <aspect/global.h>
 #include <aspect/material_model/utilities.h>
+#include <aspect/material_model/interface.h>
 
 
 
--- a/include/aspect/material_model/equation_of_state/interface.h
+++ b/include/aspect/material_model/equation_of_state/interface.h
@@ -99,6 +99,27 @@ namespace aspect
     };
 
     /**
+    * This function computes averages of multicomponent thermodynamic properties
+    * that are stored in a vector of EquationOfStateOutputs.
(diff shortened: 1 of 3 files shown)


naliboff:
@bobmyhill - Awesome! Let me know when you would like a review for this!

bobmyhill:
Hi @naliboff, this is ready for a review now :)

naliboff (commented):
@bobmyhill - looks good, thanks for moving this code to a utilities function! Before merging, is your plan to go through the other material models and use this function? I believe this could at least be used in `visco_plastic` that has nearly identical lines of code?

bobmyhill:
Hi @naliboff, yes, that's the intention (in another PR)!

naliboff (approved):
@bobmyhill - Sounds good, let me know when the other PR is ready for review!

jdannberg:
So @bobmyhill I thought we had discussed yesterday that the averaging is the responsibility of the material model, not of the equation of state, so this functionality should be part of the material model utilities, not of the interface of the equation of state. The equation of state only knows things about individual compositions, and the material model decides how the material properties are computed from that. Could you fix that in a follow-up pull request? Or did you have a different reason why you thought that should go into the equation of state rather than in the material model?

bobmyhill:
@jdannberg I did try to do it the way we discussed, but the include hierarchy seems to be structured in such a way as to make that difficult (the EoS interface knows about the material model utilities/classes, not the other way round). Do you think there's a robust workaround?

PR #4156, opened 2021-07-08
Enforce consistent constant temperature boundary indicators

bobmyhill:
When using the constant temperature boundary plugin, ASPECT now checks that the Fixed temperature boundary indicators match the indicators in the model subsection.

### DIFF
--- a/source/boundary_temperature/constant.cc
+++ b/source/boundary_temperature/constant.cc
@@ -145,12 +145,6 @@ namespace aspect
                 {
                   boundary_id
                     = this->get_geometry_model().translate_symbolic_boundary_name_to_id (parts[0]);
-
-                  AssertThrow((this->get_fixed_temperature_boundary_indicators().find(boundary_id) != this->get_fixed_temperature_boundary_indicators().end()),
-                              ExcMessage ("You have indicated a temperature mapping for "
-                                          "Boundary indicator " + parts[0] + ", but that "
-                                          "indicator isn't in the "
-                                          "list of Fixed temperature boundary indicators."));
                 }
               catch (const std::string &error)
                 {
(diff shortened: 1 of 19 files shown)


tjhei (approved):

source/boundary_temperature/constant.cc
```diff
@@ -145,6 +145,12 @@ namespace aspect
                 {
                   boundary_id
                     = this->get_geometry_model().translate_symbolic_boundary_name_to_id (parts[0]);
+
+                  AssertThrow((this->get_fixed_temperature_boundary_indicators().find(boundary_id) != this->get_fixed_temperature_boundary_indicators().end()),
+                              ExcMessage ("You have indicated a temperature mapping for "
+                                          "Boundary indicator " + parts[0] + ", but that "
```
bangerth, line 151:
```suggestion
                                          "boundary indicator " + parts[0] + ", but that "
```

source/boundary_temperature/constant.cc
```diff
@@ -145,6 +145,12 @@ namespace aspect
                 {
                   boundary_id
                     = this->get_geometry_model().translate_symbolic_boundary_name_to_id (parts[0]);
+
+                  AssertThrow((this->get_fixed_temperature_boundary_indicators().find(boundary_id) != this->get_fixed_temperature_boundary_indicators().end()),
```
bangerth, line 149: I had to look at this for a little bit because you throw an exception inside a `try`-`catch` block. Would you mind moving it after the block? The `try`-`catch` block simply tries to deal with potential failure of `translate_symbolic_boundary_name_to_id`, but the error you're producing here is logically unrelated.

tests/chunk_topography_postprocessor.prm
```diff
@@ -15,8 +15,9 @@ set Use years in output instead of seconds = true
 
 subsection Boundary temperature model
   set List of model names = constant
+  set Fixed temperature boundary indicators   = inner, outer
   subsection Constant
-    set Boundary indicator to temperature mappings = 0:0,1:0,2:0,3:0
+    set Boundary indicator to temperature mappings = inner:0,outer:0
```
bangerth, line 20: I guess what this means is that the input file previously specified temperatures for boundaries for which there is nothing to be specified, rigth?

bangerth (changes requested):
Nice work, and an impressive number of places you're finding with the check!

tests/chunk_topography_postprocessor.prm
bobmyhill: There's no functional change to this test; I've just changed the indicator numbers to their associated strings and removed the two boundaries that weren't listed as Fixed temperature boundary indicators.

source/boundary_temperature/constant.cc
bobmyhill: Done!

tests/chunk_topography_postprocessor.prm
bangerth: Right, that's what I meant: you removed indicators that the user (probably in error) specified.

bangerth (approved):

PR #4162, opened 2021-07-09
Add assert to avoid silently ignored boundary temperature constraints

bobmyhill:
The boundary temperature model plugin requires that the user indicate the boundaries on which the model should be applied using the "Fixed temperature boundary indicators" parameter. If this was not set manually, the plugin ignored whatever else was in that section of the parameter file, even if it was otherwise complete. 

In this PR:
- a new assert has been added that requires that if a model name has been chosen, the "Fixed temperature boundary indicators" parameter should have at least one boundary indicator.
- all of the ignored boundary temperature models have been removed from the tests, benchmarks and cookbooks.

### DIFF
--- a/source/boundary_temperature/interface.cc
+++ b/source/boundary_temperature/interface.cc
@@ -171,14 +171,13 @@ namespace aspect
             // If that is indeed the case, clear the model_operators vector. Otherwise, raise an exception.
             if (fixed_temperature_boundary_indicators.size() == 0)
               {
-                if (model_names.size() == 0)
-                  model_operators.clear();
-                else
-                  AssertThrow(false,
-                              ExcMessage ("You have indicated that you wish to apply a boundary temperature "
-                                          "model, but the <Fixed temperature boundary indicators> parameter "
-                                          "is empty. Please use this parameter to specify the boundaries "
-                                          "on which the model(s) should be applied."));
+                AssertThrow(model_names.size() == 0,
+                            ExcMessage ("You have indicated that you wish to apply a boundary temperature "
+                                        "model, but the <Fixed temperature boundary indicators> parameter "
+                                        "is empty. Please use this parameter to specify the boundaries "
+                                        "on which the model(s) should be applied."));
+
+                model_operators.clear();
               }
           }
         catch (const std::string &error)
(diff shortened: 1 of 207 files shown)


source/boundary_temperature/interface.cc
```diff
@@ -167,12 +167,18 @@ namespace aspect
...
-                model_names.clear();
-                model_operators.clear();
+                if (model_names.size() == 0)
+                  model_operators.clear();
+                else
+                  AssertThrow(false,
+                              ExcMessage ("You have indicated that you wish to apply a boundary temperature "
+                                          "model, but the <Fixed temperature boundary indicators> parameter "
+                                          "is empty. Please use this parameter to specify the boundaries "
+                                          "on which the model(s) should be applied."));
```
gassmoeller, line 181: If you write it like this the condition appears in the error message, which is helpful to understand the problem.

```suggestion
                  AssertThrow(model_names.size() == 0,
                              ExcMessage ("You have indicated that you wish to apply a boundary temperature "
                                          "model, but the <Fixed temperature boundary indicators> parameter "
                                          "is empty. Please use this parameter to specify the boundaries "
                                          "on which the model(s) should be applied."));
                                          
                 model_operators.clear();
```

gassmoeller (commented):
One suggestion for how to do the assert differently, apart from that this looks good. Thanks!

source/boundary_temperature/interface.cc
```diff
@@ -171,14 +171,13 @@ namespace aspect
...
               {
-                if (model_names.size() == 0)
-                  model_operators.clear();
-                else
-                  AssertThrow(false,
-                              ExcMessage ("You have indicated that you wish to apply a boundary temperature "
-                                          "model, but the <Fixed temperature boundary indicators> parameter "
-                                          "is empty. Please use this parameter to specify the boundaries "
-                                          "on which the model(s) should be applied."));
+                AssertThrow((model_names.size() == 0),
```
gassmoeller, line 174: Yes, except these parentheses are not necessary

```suggestion
                AssertThrow(model_names.size() == 0,
```

bobmyhill: Darn!

gassmoeller (approved):
Very nice. Looks good to go once the tester is ready.

PR #4163, opened 2021-07-09
Added contributions directory to indent

bobmyhill:
This PR adds contrib to the list of directories that are tidied by contrib/utilities/indent.


tjhei:
What is the reasoning for this? Most of the code is not developed Herr, but copied in.

bobmyhill:
I think it would be good to indent all the code/prm files that are developed (or at least maintained) within the ASPECT project. 

My logic was that as all of the changes at the moment are either for prms or for the worldbuilder, it was probably ok to have a blanket indenter. But this isn't future-proof, and worldbuilder is maintained elsewhere, so maybe it would be better just to do the check for prms?

bangerth:
Regardless of whether you want to do that automatically, the changes that just strip trailing space and/or replace tabs and/or add newlines at the end of the files are ok by themselves as long as they are to files that we actually maintain.

bobmyhill:
For now, I'll open up a new PR dealing with the prms manually. There are only four of them, so automating wouldn't be very useful anyway.

PR #4164, opened 2021-07-09
gravity benchmark thin shell

cedrict:
This is the first out of three gravity benchmarks/cookbooks I intend to contribute this hackathon.


benchmarks/gravity_thin_shell/gravity_thin_shell.prm
```diff
@@ -0,0 +1,82 @@
+# A simple setup for testing the gravity postprocessor using the 
+# sample scheme 'map'.
```
bobmyhill, line 2: Benchmarks prms ideally have descriptions at the beginning that describe what the prm is benchmarking, how it is doing that, what the numerical comparisons are and why it is useful. See, for example `benchmarks/viscoelastic_bending_beam/viscoelastic_bending_beam.prm` It would be good to extend this blurb by a few lines to cover these details.

benchmarks/gravity_thin_shell/gravity_thin_shell.prm
```diff
@@ -0,0 +1,82 @@
...
+
+# General parameters
+set Dimension                              = 3
+set End time                               = 0
+set Output directory                       = output_thin_shell_gravity
+set Nonlinear solver scheme                = no Advection, no Stokes
+
+# Model geometry
```
bobmyhill, line 10: Please add a sentence explaining why there are commented lines in the following section.

benchmarks/gravity_thin_shell/gravity_thin_shell.prm
```diff
@@ -0,0 +1,82 @@
...
+    # layer 3000km depth
+    #set Inner radius = 3366e3
+    #set Outer radius = 3376e3
+    set Custom mesh subdivision = number of slices
+    set Number of slices = 1
+    set Initial lateral refinement = 5
+  end
+end
+
+# Model boundary velocity
```
bobmyhill, line 35: All of these section-beginning comment lines are identical to the section names themselves, so they can be removed.

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
 
 The parameter file and initial temperature model for this benchmark can be found at \url{benchmarks/yamauchi_takei_2016_anelasticity/yamauchi_takei_2016_anelasticity.prm} and \url{benchmarks/yamauchi_takei_2016_anelasticity/anelasticity_temperature.cc}. Code to recreate  Fig.~\ref{fig:anelasticity}  is provided in \url{benchmarks/yamauchi_takei_2016_anelasticity/plot_output}.
 
+\subsubsection{Thin shell gravity benchmark}
+\label{sec:benchmark-thin-shell-gravity}
+
+\textit{This section was contributed by Cedric Thieulot, Bart Root and Paul Bremmer.}
+
+The domain is a spherical shell of 10~\si{\km} radius centered at a depth $D$, i.e. the inner radius is $R_i=6371-D-5~\si{\km}$ and the outer radius $R_o=6371-D+5~ \si{\km}$. It is filled with a fluid of constant density $\rho_0=3300~\si{\kg\per\cubic\metre}$ with total mass $M=\frac43 \pi (R_o^3-R_i^3)$.
```
bobmyhill, line 9482: In the manual, subscripts that aren't indices are typically in roman typeface. The easiest way to do this is, for example: `$R\textsubscript{inner}=6371-D-5~\si{\km}$`. Same for `outer`, rather than `$o$`

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+
+\noindent Both the Stokes and energy equations solve are bypassed:
+
+\lstinputlisting[language=prmfile]{cookbooks/benchmarks/gravity_thin_shell/doc/thinshell_b.prm.out}
+
+We make use of the Custom mesh subdivision option to generate a mesh where a single cell is used in the radial direction (parameterized with the number of 'slices') while 6 blocks (default value) of $2^5\times 2^5$ cells are used in the lateral direction. This gives a total of 6,144 mesh cells. For $D=100~\si{\km}$ the parameterization of the mesh is then as follows:
+
+\lstinputlisting[language=prmfile]{cookbooks/benchmarks/gravity_thin_shell/doc/thinshell_a.prm.out}
+
+We wish to compute the gravity potential $U$ and the gravity vector ${\mathbf g}$ at a height of $250~\si{\km}$ above the surface of the Earth (i.e. the satellite is flying at a radius of $R_s=6371+250=6621~\si{\km}$) on a regular longitude-latitude grid spanning the whole Earth, with a $2\si{\degree}\times 2\si{\degree}$ resolution:
```
bobmyhill, line 9492:
```suggestion
We wish to compute the gravity potential $U$ and the gravity vector ${\mathbf g}$ at a radius of $R\textsubscript{s}=6371+250=6621~\si{\km}$ (i.e. the gravity experienced by a satellite flying at a height of $250~\si{\km}$ above the surface of the Earth) on a regular longitude-latitude grid spanning the whole Earth, with a $2\si{\degree}\times 2\si{\degree}$ resolution:
```

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+
+\lstinputlisting[language=prmfile]{cookbooks/benchmarks/gravity_thin_shell/doc/thinshell_c.prm.out}
+
+The Gauss-Legendre Quadrature (GLQ) algorithm is central to the ASPECT code as it is used to compute the elemental integrals stemming from the discretisation of the weak form of the PDEs. Logically, the gravity postprocessor is also based on the GLQ since the potential $U$ at any location in space is given by the integral 
+\begin{equation}
+U({\mathbf r}) = \iiint_\Omega \frac{G \rho({\mathbf r}')}{|{\mathbf r}-{\mathbf r}'|} d{\mathbf r}'
+= \sum_K \iiint_{\Omega_K} \frac{G \rho({\mathbf r}')}{|{\mathbf r}-{\mathbf r}'|} d{\mathbf r}'
+\end{equation}
+where the sum runs over all cells of the mesh and ${G}=6.67430\times 10^{-11}~\si{\cubic\metre\per\kg\per\square\second}$ is the gravitaional constant. The gravity acceleration vector is obtained via ${\mathbf g}=-{\mathbf \nabla} U$, or
```
bobmyhill, line 9501:
```suggestion
where the sum runs over all cells of the mesh and ${G}=6.67430\times 10^{-11}~\si{\cubic\metre\per\kg\per\square\second}$ is the gravitational constant. The gravity acceleration vector is obtained via ${\mathbf g}=-{\mathbf \nabla} U$, or 
```

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+
+\lstinputlisting[language=prmfile]{cookbooks/benchmarks/gravity_thin_shell/doc/thinshell_a.prm.out}
+
+We wish to compute the gravity potential $U$ and the gravity vector ${\mathbf g}$ at a height of $250~\si{\km}$ above the surface of the Earth (i.e. the satellite is flying at a radius of $R_s=6371+250=6621~\si{\km}$) on a regular longitude-latitude grid spanning the whole Earth, with a $2\si{\degree}\times 2\si{\degree}$ resolution:
+
+\lstinputlisting[language=prmfile]{cookbooks/benchmarks/gravity_thin_shell/doc/thinshell_c.prm.out}
+
+The Gauss-Legendre Quadrature (GLQ) algorithm is central to the ASPECT code as it is used to compute the elemental integrals stemming from the discretisation of the weak form of the PDEs. Logically, the gravity postprocessor is also based on the GLQ since the potential $U$ at any location in space is given by the integral
```
bobmyhill, line 9496:
```suggestion
The Gauss-Legendre Quadrature (GLQ) algorithm is central to the ASPECT code as it is used to compute the elemental integrals stemming from the discretization of the weak form of the PDEs. Logically, the gravity postprocessor is also based on the GLQ since the potential $U$ at any location in space is given by the integral 
```

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+\begin{equation}
+{\mathbf g}({\mathbf r}) = 
+\iiint_\Omega  \frac{G \rho({\mathbf r}')}{|{\mathbf r}-{\mathbf r}'|^2} d{\mathbf r}'
+= 
+\sum_K \iiint_{\Omega_K} \frac{G \rho({\mathbf r}')}{|{\mathbf r}-{\mathbf r}'|^2} d{\mathbf r}'.
+\end{equation}
+
+The default number of quadrature points for each cell in the postprocessor is $2^{ndim}$, where $ndim$ is the number of dimensions. The $n$-point GLQ allows exact integration of $(2n-1)$-order polynomials. However, the integrand of the Newton integral is not a polynomial (it contains a term $\sim r^{-m}$), so there is not an optimal number of quadrature points to use. Therefore, the postprocessor allows the user to choose how many additional points per dimension are used with an expectation that an increase in the number of quadrature points inside the cells leads to a more accurate calculation. This increase number $I$ is set to 1 in the example above, although it can also be chosen to be 0, 1, 2 or even -1. The case $I=-1$ approximates the cell as a point mass since there is a single quadrature point in the middle of the cell.
```
bobmyhill, line 9509:
```suggestion
The default number of quadrature points for each cell in the postprocessor is $2\textsuperscript{ndim}$, where ndim is the number of dimensions. The $n$-point GLQ allows exact integration of $(2n-1)$-order polynomials. However, the integrand of the Newton integral is not a polynomial (it contains a term $\sim r^{-m}$), so there is not an optimal number of quadrature points to use. Therefore, the postprocessor allows the user to choose how many additional points per dimension are used with an expectation that an increase in the number of quadrature points inside the cells leads to a more accurate calculation. This increase number $I$ is set to 1 in the example above, although it can also be chosen to be 0, 1, 2 or even -1. The case $I=-1$ approximates the cell as a point mass since there is a single quadrature point in the middle of the cell.
```

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+
+Given the symmetry of the problem the value of the potential and acceleration depend solely on the distance $r$ from the origin, see Turcotte and Schubert \cite{TS14}. Their analytical values are given by $U(R_s) = - {\cal G}M/R_s$ and $|{\mathbf g}(R_s)|= g_r(R_s)={\cal G}M/R_s^2$.
```
bobmyhill, line 9511: more opportunity for \textsubscript{s}.

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+-47503.2981 &
+-47503.2981 \\
+&{\small a.v.} && 717.4641 &&& -47503.2981 && \\
+\hline
+\end{tabular}\\
+\caption{\it Thin shell gravity benchmark: $1\si{mGal}=10^{-5}\si{\metre\per\square\second}$. $n_q$ is the number of GLQ points per cell. 'a.v.' stands for analytical value.}
+\label{tab:thin_shell_gravity_benchmark}
+\end{table}
+
+The accuracy of the calculations increases with $I$ but so does the time spent in the postprocessor: for $I\in\{-1,0,1,2\}$  this time was about 18, 132, 440 and 1040~\si{s} respectively. This is easily explained when one realises that increasing $I$ from 0 to 1 means that the number of GLQ points per cell goes from $2^3=8$ to $3^3=27$, i.e. a $3.375$ increase in operations for the same number of cells and measurement points. Coincidentally, the time spent in the postprocessor increases by the same amount $440/132\sim 3.4$.
```
bobmyhill, line 9679:
```suggestion
The accuracy of the calculations increases with $I$ but so does the time spent in the postprocessor: for $I\in\{-1,0,1,2\}$  this time was about 18, 132, 440 and 1040~\si{s} respectively. This is easily explained when one realises that increasing $I$ from 0 to 1 means that the number of GLQ points per cell goes from $2^3=8$ to $3^3=27$, i.e. a $3.375$ increase in operations for the same number of cells and measurement points. The time spent in the postprocessor increases by a similar factor ($440/132\sim 3.4$).
```

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+The accuracy obtained with lower $I$ values increases with increasing anomaly depth (or increasing distance from the observation point). Note that this simplistic case has uniform density so that the projection of the density from the nodes onto the quadrature points is exact and it does not introduce any smoothing of data which might occur in the presence of density discontinuities (e.g. air-water, water-crust, moho, etc.) inside a cell.
```
bobmyhill, line 9680:
```suggestion
The accuracy obtained with lower $I$ values increases with increasing anomaly depth (or increasing distance from the observation point). Note that this benchmark has uniform density so that the projection of the density from the nodes onto the quadrature points is exact and it does not introduce any smoothing of data which might occur in the presence of density discontinuities (e.g. air-water, water-crust, moho, etc.) inside a cell.
```

doc/manual/manual.tex
bobmyhill: You could also quantify the relative accuracies when using the different values of $I$ here.

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+In conclusion, $I=1$ seems to be the best compromise between accuracy and compute time for this experiment.
```
bobmyhill, line 9681:
```suggestion
$I=1$ seems to be the best compromise between accuracy and compute time for this experiment.
```

bobmyhill: It might be worth another sentence explaining how you define "best". What kind of accuracy is desirable?

bobmyhill (changes requested):
Great! Just a few minor changes.

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
 
 The parameter file and initial temperature model for this benchmark can be found at \url{benchmarks/yamauchi_takei_2016_anelasticity/yamauchi_takei_2016_anelasticity.prm} and \url{benchmarks/yamauchi_takei_2016_anelasticity/anelasticity_temperature.cc}. Code to recreate  Fig.~\ref{fig:anelasticity}  is provided in \url{benchmarks/yamauchi_takei_2016_anelasticity/plot_output}.
 
+\subsubsection{Thin shell gravity benchmark}
+\label{sec:benchmark-thin-shell-gravity}
+
+\textit{This section was contributed by Cedric Thieulot, Bart Root and Paul Bremmer.}
```
bobmyhill, line 9480:
```suggestion
\textit{This section was contributed by Cedric Thieulot, Bart Root and Paul Bremner.}
```

doc/manual/manual.tex
bangerth: Separately, would you mind adding a short paragraph at the top of this section that starts it out by saying what the goal here is? You're jumping right into the weeds :-)

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+
+The domain is a spherical shell of 10~\si{\km} radius centered at a depth $D$, i.e. the inner radius is $R_i=6371-D-5~\si{\km}$ and the outer radius $R_o=6371-D+5~ \si{\km}$. It is filled with a fluid of constant density $\rho_0=3300~\si{\kg\per\cubic\metre}$ with total mass $M=\frac43 \pi (R_o^3-R_i^3)$.
+
+\noindent Both the Stokes and energy equations solve are bypassed:
```
bangerth, line 9484: Leave formatting to latex :-)
```suggestion
Both the Stokes and energy equations solve are bypassed:
```

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,211 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+
+\lstinputlisting[language=prmfile]{cookbooks/benchmarks/gravity_thin_shell/doc/thinshell_b.prm.out}
+
+We make use of the Custom mesh subdivision option to generate a mesh where a single cell is used in the radial direction (parameterized with the number of 'slices') while 6 blocks (default value) of $2^5\times 2^5$ cells are used in the lateral direction. This gives a total of 6,144 mesh cells. For $D=100~\si{\km}$ the parameterization of the mesh is then as follows:
```
bangerth, line 9488:
```suggestion
We make use of the Custom mesh subdivision option to generate a mesh where a single cell is used in the radial direction (parameterized with the number of `slices') while 6 blocks (default value) of $2^5\times 2^5$ cells are used in the lateral direction. This gives a total of 6,144 mesh cells. For $D=100~\si{\km}$ the parameterization of the mesh is then as follows:
```

doc/manual/manual.tex
bangerth:
```suggestion
The Gauss-Legendre Quadrature (GLQ) algorithm is central to the \aspect{} code as it is used to compute the elemental integrals stemming from the discretisation of the weak form of the PDEs. Logically, the gravity postprocessor is also based on the GLQ since the potential $U$ at any location in space is given by the integral 
```

bangerth: @bobmyhill That must have hurt to correct @cedrict to use AE spelling :-)

doc/manual/manual.tex
bangerth:
```suggestion
The accuracy of the calculations increases with $I$ but so does the time spent in the postprocessor: for $I\in\{-1,0,1,2\}$  this time was about 18, 132, 440 and 1040~\si{s} respectively. This is easily explained when one realizes that increasing $I$ from 0 to 1 means that the number of GLQ points per cell goes from $2^3=8$ to $3^3=27$, i.e. a $3.375$ increase in operations for the same number of cells and measurement points. Coincidentally, the time spent in the postprocessor increases by the same amount $440/132\sim 3.4$.
```

benchmarks/gravity_thin_shell/gravity_thin_shell.prm
cedrict: done

benchmarks/gravity_thin_shell/gravity_thin_shell.prm
cedrict: done

benchmarks/gravity_thin_shell/gravity_thin_shell.prm
cedrict: done

doc/manual/manual.tex
cedrict: done

doc/manual/manual.tex
cedrict: did that

doc/manual/manual.tex
cedrict: exactly my thought :)

doc/manual/manual.tex
cedrict: done

doc/manual/manual.tex
cedrict: indeed! exciting! :)

doc/manual/manual.tex
cedrict: this benchmark is part of a paper about to be submitted in which we discuss the errors and compare those with errors obtained with other codes/methods.

doc/manual/manual.tex
cedrict: done

doc/manual/manual.tex
cedrict: oups!!!

cedrict:
I hope I have adequately addressed all your comments.

doc/manual/manual.tex
cedrict: also: fun fact, M is wrong here :)

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,212 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
 
 The parameter file and initial temperature model for this benchmark can be found at \url{benchmarks/yamauchi_takei_2016_anelasticity/yamauchi_takei_2016_anelasticity.prm} and \url{benchmarks/yamauchi_takei_2016_anelasticity/anelasticity_temperature.cc}. Code to recreate  Fig.~\ref{fig:anelasticity}  is provided in \url{benchmarks/yamauchi_takei_2016_anelasticity/plot_output}.
 
+\subsubsection{Thin shell gravity benchmark}
+\label{sec:benchmark-thin-shell-gravity}
+
+\textit{This section was contributed by Cedric Thieulot, Bart Root and Paul Bremner.}
+
+The idea behind this benchmark is to test the accuracy of the gravity postprocessor: the domain is a thin shell of constant density somewhere in the Earth mantle and we wish to compute the resultiong gravity field and potential at satellite height.
```
bobmyhill, line 9482:
```suggestion
The idea behind this benchmark is to test the accuracy of the gravity postprocessor: the domain is a thin shell of constant density somewhere in the Earth mantle and we wish to compute the resulting gravity field and potential at satellite height.
```

doc/manual/manual.tex
```diff
@@ -9474,6 +9474,212 @@ \subsubsection{Yamauchi \& Takei anelastic shear wave velocity-temperature conve
...
+-47503.2981 &
+-47503.2981 \\
+&{\small a.v.} && 717.4641 &&& -47503.2981 && \\
+\hline
+\end{tabular}\\
+\caption{\it Thin shell gravity benchmark: $1\si{mGal}=10^{-5}\si{\metre\per\square\second}$. $n_q$ is the number of GLQ points per cell. 'a.v.' stands for analytical value.}
+\label{tab:thin_shell_gravity_benchmark}
+\end{table}
+
+The accuracy of the calculations increases with $I$ but so does the time spent in the postprocessor: for $I\in\{-1,0,1,2\}$  this time was about 18, 132, 440 and 1040~\si{s} respectively. This is easily explained when one realizes that increasing $I$ from 0 to 1 means that the number of GLQ points per cell goes from $2^3=8$ to $3^3=27$, i.e. a $3.375$ increase in operations for the same number of cells and measurement points. The time spent in the postprocessor increases by a similar factor $440/132\sim 3.4$.
```
bobmyhill, line 9681:
```suggestion
The accuracy of the calculations increases with $I$ but so does the time spent in the postprocessor: for $I\in\{-1,0,1,2\}$  this time was about 18, 132, 440 and 1040~\si{s} respectively. This is easily explained when one realizes that increasing $I$ from 0 to 1 means that the number of GLQ points per cell goes from $2^3=8$ to $3^3=27$, i.e. a $3.375$ increase in operations for the same number of cells and measurement points. The time spent in the postprocessor increases by a similar factor ($440/132\sim 3.4$).
```

benchmarks/gravity_thin_shell/gravity_thin_shell.prm
```diff
@@ -0,0 +1,83 @@
...
+# and potential so that values returned by the gravity postprocessor
+# can be benchmarked on a longitude-latitude grid at satellite orbit height.  
+
+set Dimension                              = 3
+set End time                               = 0
+set Output directory                       = output_thin_shell_gravity
+set Nonlinear solver scheme                = no Advection, no Stokes
+
+# Model geometry
+# The thin shell (5km thickness) can be centered around 5 different depths,
```
bobmyhill, line 13: Could you add a space between the number and "km" for this and the following lines (also in the Spherical shell subsection)?

bobmyhill (commented):
Ok, a couple of trivial typos, and then this is good to go.

bobmyhill:
(Also, you'll need to run contrib/utilities/indent)

benchmarks/gravity_thin_shell/gravity_thin_shell.prm
cedrict: done

doc/manual/manual.tex
cedrict: done

benchmarks/gravity_thin_shell/gravity_thin_shell.prm
cedrict: also indent

bobmyhill:
The manual doesn't compile at the moment because your code snippets need to go into a subdirectory named /doc/ (i.e. manual/cookbooks/benchmarks/gravity_thin_shell/doc/thinshell_*.prm)

After that, everything *should* work automagically.

cedrict:
ok, but then there is a pb: I copied the structure of the crameri benchmark and the prm file snippets are in doc/manual/cookbooks/benchmarks/crameri  , and the .out files are created in the doc folder

bobmyhill:
At the very least, the `doc` folder must be in the `manual/cookbooks/benchmarks/gravity_thin_shell/` directory. I guess you could leave it empty, but I think it's better to fill it with the relevant prms and figures.

cedrict:
so what happens is that when I do make manual.pdf, the prm files are automatically converted into .out files and placed in the doc folder. Since the manual is probably generated the same way on Timo's machine, why would I place these .out files by hand?
When I git clone aspect, the doc folder of crameri does not contain .out files

bangerth:
Correct. `doc/Makefile` has rules to generate `.prm.out` files from the `.prm` files.

bobmyhill:
I'm not arguing that you add the .out files to the PR. I'm just saying that when I try to compile the manual using this branch, I get the same error as the tester, and that error is fixed by adding an empty `doc` directory.

cedrict:
oh so I need to add an empty doc!

cedrict:
problem: I do git add doc/ (which is empty). and then git status tells me 
On branch gravity_benchmark_thin_shell
nothing to commit, working tree clean

bobmyhill:
Yes, that's one of the reasons I suggested that you could fill it with the prms and/or figures :) 
(you would have to change the paths in manual.tex)

cedrict:
done. seems to work too.

bobmyhill:
In fact, git won't allow empty directories to be added, so we have to fill it with something: https://git.wiki.kernel.org/index.php/GitFaq#Can_I_add_empty_directories.3F

bobmyhill:
Hurrah!

bobmyhill (approved):

PR #4171, opened 2021-07-09
more cleanup in steinberger model

jdannberg:
This addresses @gassmoeller's comments on #4160.

### DIFF
--- a/source/material_model/utilities.cc
+++ b/source/material_model/utilities.cc
@@ -723,8 +723,7 @@ namespace aspect
 
       std::vector<double>
       compute_composition_fractions(const std::vector<double> &compositional_fields,
-                                    const ComponentMask &field_mask,
-                                    const bool has_background_field)
+                                    const ComponentMask &field_mask)
       {
         std::vector<double> composition_fractions(compositional_fields.size()+1);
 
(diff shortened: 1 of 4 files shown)


source/material_model/utilities.cc
```diff
@@ -779,6 +775,13 @@ namespace aspect
       {
         const unsigned int n_fields = masses.size();
         std::vector<double> volumes(n_fields);
+
+        if (n_fields == 1)
+          {
+            volumes[0] = 1.0;
+            return volumes;
+          }
+
```
gassmoeller, line 784: Shouldnt the existing code below do the same thing for n_fields == 1? If so remove this to keep the function simpler (also: with your current implementation what would happen if n_fields == 1, but return_as_fraction == false? then this function would return the wrong value I think).

gassmoeller (commented):
Thanks, just two small comments.

source/material_model/utilities.cc
jdannberg: No, I actually implemented this because a test failed. The existing version of this assumed that there is always a background field, so it didn't run into this problem. But there were steinberger tests without a background field, that still had compositions of zero somewhere. In that case, sum_volumes is zero and you get a floating point exception. But I can fix the case with return_as_fraction == false.

gassmoeller (approved):

PR #4172, opened 2021-07-09
Use C++11-style double closing template brackets.

bangerth:
I wonder why none of us ever felt like doing this. It must count as one of the dumber patches for the hackathon, but hey, I'm saving us all ~1500 bytes on our disks! :-)

/rebuild


gassmoeller (approved):
Also one of the more boring patches to review :-) ... you will need to rebase though, there is a conflict.

source/boundary_composition/interface.cc
```diff
@@ -397,11 +397,11 @@ namespace aspect
...
       template <>
-      std::list<internal::Plugins::PluginList<BoundaryComposition::Interface<2> >::PluginInfo> *
-      internal::Plugins::PluginList<BoundaryComposition::Interface<2> >::plugins = nullptr;
+      std::list<internal::Plugins::PluginList<BoundaryComposition::Interface<2>>::PluginInfo> *
+                                                                              internal::Plugins::PluginList<BoundaryComposition::Interface<2>>::plugins = nullptr;
       template <>
-      std::list<internal::Plugins::PluginList<BoundaryComposition::Interface<3> >::PluginInfo> *
-      internal::Plugins::PluginList<BoundaryComposition::Interface<3> >::plugins = nullptr;
+      std::list<internal::Plugins::PluginList<BoundaryComposition::Interface<3>>::PluginInfo> *
+                                                                              internal::Plugins::PluginList<BoundaryComposition::Interface<3>>::plugins = nullptr;
```
MFraters, line 404: this looks kind of weird to me. Is this the indentation style we have? (there are a few more of these)

MFraters:
This is an amazing pull request to hide an easter egg. Just saying...

source/boundary_composition/interface.cc
bangerth: astyle doesn't understand C++11 is what it probably is...

bangerth:
Ha, yes, that would be the way to place an Easter egg :-)

I rebased.

bangerth:
@MFraters One way to see that there really are no untoward changes in this patch is to look at it with all whitespace-only changes hidden: https://github.com/geodynamics/aspect/pull/4172/files?w=1

I'm sure there are programming languages where whitespace changes can be exploited for functional changes, but C++ is not one that makes that easy. I think that the lexing into different tokens when writing `>>` vs `> >` is the only one that could conceivably be used to write programs in which both the form with and without space could represent a valid program. In all other cases I can come up with, say `+=` vs `+ =`, the program with the space would be ill-formed since these operators cannot follow each other when separate.

gassmoeller (approved):

MFraters:
@bangerth, sorry for the late reply, but thanks for showing this feature in github. I didn't know that existed :)

PR #4174, opened 2021-07-10
read in the dominant phase from a lookup table

jdannberg:
The pull request allows it to read in a table containing the dominant phase from a lookup table. At the moment, this information is only used in the visualization output, but in the future, material models could use this to assign other properties like, for example, the viscosity) based on the dominant phase. Note that ASPECT could already read in phase fractions before (implemented last year by @bobmyhill) but the HeFESTo table I got did not have the volume fractions of each individual phase, it only had one column with the name of the dominant phase. This also potentially makes the data table smaller, in case there are a lot of phases. 

Note that the reason using these names is a bit complicated is that we don't want to store them as strings (because string comparisons are slow), and we cannot output them as strings in the visualization output. So instead I am using an index for each phase. 

### For all pull requests:


source/material_model/equation_of_state/thermodynamic_table_lookup.cc
```diff
@@ -351,6 +424,9 @@ namespace aspect
         // fill phase volume outputs if they exist
         if (NamedAdditionalMaterialOutputs<dim> *phase_volume_fractions_out = out.template get_additional_output<NamedAdditionalMaterialOutputs<dim> >())
           fill_phase_volume_fractions(in, volume_fractions, phase_volume_fractions_out);
+
+        if (PhaseOutputs<dim> *dominant_phases_out = out.template get_additional_output<PhaseOutputs<dim> >())
+          fill_dominant_phases(in, volume_fractions, dominant_phases_out);
```
gassmoeller, line 429: Keep the line above, but hand over a reference instead of a pointer like follows:
```suggestion
          fill_dominant_phases(in, volume_fractions, *dominant_phases_out);
```

gassmoeller (commented):
The similarity to the existing functionality is somewhat unfortunate, but as long as you can explain the difference in the documentation it should be ok. I added some comment and would like to take another look after they have been addressed.

jdannberg:
I addressed you comments, @gassmoeller could you have another look?

As mentioned earlier today, this is only implemented for the perplex reader. I wanted to check what happens for the hefesto reader, but it turns out we don't have a single test or test data for it. It also looks like the columns are just hard-coded (so there is no header), there's also a derivative file needed, and when that isn't there the model just crashes with a segmentation fault (so there are no checks that the file is actually there, or about the correct number of files). So in summary, we should add more checks in the future, or remove this option altogether. I am not even sure that this corresponds to anything that is a current output format of HeFESTo. 

But I added a test, which produces exactly the same results as the same setup with the perplex reader, so at least now we know it's working; and the good new is: it just ignores all of the columns after the ones that are hard-coded, so it still works if there is a column with phase names, there's just no way to read in that column.

source/material_model/equation_of_state/thermodynamic_table_lookup.cc
```diff
@@ -91,6 +93,58 @@ namespace aspect
...
+                if (it == list_of_dominant_phases.end())
+                  list_of_dominant_phases.push_back(phase_name);
+              }
+
+            // Make sure that either all or none of the tables have a column with the dominant phase.
+            if (any_lookup_has_dominant_phase)
+              AssertThrow(material_lookup[i]->has_dominant_phase(),
+                          ExcMessage("Some of the lookup tables you read in contain outputs for the dominant phase, "
+                                     "as indicated by the column 'phase', but in at least of of the tables you use "
+                                     "this column is missing."));
```
gassmoeller, line 120: You will have to move this part into a separate loop over n_material_lookups after the current loop finishes. In the current form, you could find the first two tables without phase columns, the next three with phase columns, and this check would not notice, because it only checks all future materials for a phase column.

jdannberg: I think I found an even simpler way. Should be fixed now.

gassmoeller:
Nice, ready to go once the tester is finished.

PR #4180, opened 2021-07-12
Add check that creep prefactors are larger than zero

anne-glerum:
Avoid division by zero by checking the creep prefactors in parse_parameters. With the new phase transition functionality, the pattern of the prefactors can be anything, so there is no checking of the bounds. Values have to be given for all fields, even if they're masked. Often, users give masked fields a nonsensical value, like zero. This can trigger asserts in e.g. harmonic averaging. This PR just checks that the prefactors are larger than zero after reading them in. 

Related to #4182. 

### For all pull requests:


bangerth (approved):

bangerth:
Nice, thanks!

PR #4184, opened 2021-07-12
Add plate cooling to adiabatic initial temperature plugin

danieldouglas92:

* [ ] I have read the guidelines in our [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) document.

### For all pull requests:


source/initial_temperature/adiabatic.cc
```diff
@@ -33,210 +33,279 @@ namespace aspect
...
     Adiabatic<dim>::
     initial_temperature (const Point<dim> &position) const
     {
+      double age_top = 0;
+      double age_bottom = (this->convert_output_to_years() ? age_bottom_boundary_layer * year_in_seconds
+                           : age_bottom_boundary_layer);
+      if (read_from_ascii_file)
+        {
+          age_top = Utilities::AsciiDataBoundary<dim>::get_data_component(surface_boundary_id,
+                                                                                position,
```
naliboff, line 68: Remove white spaces to align this line and the following one to match the start of `surface_boundary_id`

naliboff (commented):
@danieldouglas92 - Thanks again for tackling this! A few initial comments at specific locations, more tomorrow. 

A few general questions:
1. Can you run make indent and update the PR? It looks like the tester is failing due to indentation issues and there is some formatting off here and there.
2. It looks like large section of code have been removed and added that it is separate from the new ASCII data functionality. Is this just moving code around to conform with the new logic?
3. Do you have a low resolution test that uses the new AsciiData functionality of the plugin? 

Thanks again!

source/initial_temperature/adiabatic.cc
```diff
@@ -316,6 +394,10 @@ namespace aspect
 
       prm.enter_subsection ("Initial temperature model");
       {
+        Utilities::AsciiDataBase<dim>::declare_parameters(prm,
+                                                          "$ASPECT_SOURCE_DIR/data/initial-temperature/adiabatic/",
+                                                          "adiabatic.txt",
+                                                          "Adiabatic");
```
jdannberg, line 400: I think this is why the tester does not work. This is the declare parameters function of the ascii data base class, but you're calling it in the parse_parameters function. Instead you need to move this up to line 315 and call the appropriate parse_parameters function here instead.

jdannberg, line 400:
```suggestion
        Utilities::AsciiDataBase<dim>::parse_parameters(prm, "Adiabatic");
```

jdannberg (commented):
Thank you for contributing this!
I have added a few suggestions for how to make the code shorter and easier to read, and I agree with John; it would be great if you could create a test case. 
Let us know if you need help!

It looks like at the moment the tester does not run all tests because the documentation can not be built, I hope that will be fixed by addressing my comment about the declare/parse parameter function.

danieldouglas92:
I addressed most of the comments, but the test case that I've written does not have a varying temperature. The temperature everywhere in the model is equal to the adiabatic surface temperature set in the input file (for both the half space and plate cooling model, and for ages read in from an ASCII file or set in the input file)

source/initial_temperature/adiabatic.cc
```diff
@@ -33,17 +33,47 @@ namespace aspect
...
+    void
+    Adiabatic<dim>::initialize ()
+    {
+      // Find the boundary indicator that represents the surface
+      surface_boundary_id = this->get_geometry_model().translate_symbolic_boundary_name_to_id("top");
+      std::set<types::boundary_id> surface_boundary_set;
+      surface_boundary_set.insert(surface_boundary_id);
+
+      // The input ascii table contains one data column (LAB depths(m)) in addition to the coordinate columns.
+      Utilities::AsciiDataBoundary<dim>::initialize(surface_boundary_set,
```
jdannberg, line 53: This is not a full review, but I just noticed this when I looked through the file: You only want to do this in case 
```
if (read_from_ascii_file)
```
otherwise ASPECT will try to read in a data file that may not be there.

danieldouglas92: I tried this and it doesn't allow it because of the following error:

```
cannot assign to non-static data member within const member function 'initial_temperature'
 surface_boundary_id = this->get_geometry_model().translate_symbolic_boundary_name_to_id("top");
```
Hopefully this error can be fixed and there's a way around this, but if not I don't think it would matter. If someone were using this plugin and didn't want to use an ASCII file, since the default directory is the file in data/initial-temperature/adiabatic they would automatically have this file when they clone ASPECT and it would never not find the file.

danieldouglas92: I should mention that the error I just sent was what I tried to do after I tried doing what you suggested. When I just added the two lines that you highlighted in your review I received this error:

```
error: use of undeclared identifier 'surface_boundary_set'; did you mean 'surface_boundary_id'?
          surface_boundary_set.insert(surface_boundary_id);
```

source/initial_temperature/adiabatic.cc
```diff
@@ -33,17 +33,48 @@ namespace aspect
...
+      double age_top = 0;
       const double age_bottom = (this->convert_output_to_years() ? age_bottom_boundary_layer * year_in_seconds
                                  : age_bottom_boundary_layer);
-
+      if (read_from_ascii_file)
+        {
+          // The input ascii contains the age of the seafloor. User must provide ages in seconds
+          age_top = Utilities::AsciiDataBoundary<dim>::get_data_component(surface_boundary_id,
+                                                                          position,
+                                                                          0);
```
jdannberg, line 70: I think it would be better to allow it to input the age in years. 
```suggestion
          const double age_in_seconds = Utilities::AsciiDataBoundary<dim>::get_data_component(surface_boundary_id,
                                                                          position,
                                                                          0);
          age_top = this->convert_output_to_years() ? age_in_seconds * year_in_seconds
                        : age_in_seconds);
```

source/initial_temperature/adiabatic.cc
```diff
@@ -99,18 +128,41 @@ namespace aspect
...
+                                       : 0.0;
+        }
+
+      if (cooling_model == "plate cooling")
+        {
+          if (depth > lithosphere_thickness)
+            {
+              surface_cooling_temperature = 0;
+            }
+          else
```
jdannberg, line 155: You can make this much shorter if you change it to 
```suggestion
          surface_cooling_temperature = 0;
          if (depth <= lithosphere_thickness)
```

source/initial_temperature/adiabatic.cc
```diff
@@ -99,18 +128,41 @@ namespace aspect
...
+                                        : 0.0;
+          bottom_heating_temperature = (age_bottom > 0.0 && this->get_adiabatic_conditions().is_initialized()) ?
+                                       (T_bottom - adiabatic_bottom_temperature + subadiabaticity)
+                                       * erfc((this->get_geometry_model().maximal_depth()
+                                               - this->get_geometry_model().depth(position)) /
+                                              (2 * sqrt(kappa * age_bottom)))
+                                       : 0.0;
+        }
+
+      if (cooling_model == "plate cooling")
```
jdannberg, line 149:
```suggestion
      else if (cooling_model == "plate cooling")
```

jdannberg (commented):
This looks good, there are only some small changes I pointed out (mostly the one with reading in the file), and then we have to figure out why the tester doesn't pass. 

You also mentioned you think it's not intuitive that the adiabatic surface temperature needs to be the same as the mantle temperature. The adiabatic surface temperature is simply the potential temperature of the mantle, so if it is set correctly, it should always be the mantle temperature. But I agree that this is not really intuitive or error-proof. So I thought we could add a check that makes sure the adiabatic surface temperature is set correctly (and otherwise the user gets an error message), but I didn't really have a good idea of what to check. If we checked if the adiabatic surface temperature equals the temperature at the bottom of the model, that would make it impossible to run models that also have a boundary layer at the bottom, or where the temperature varies on the bottom boundary. But if you have a good idea for a check like this, I'd be happy to help implement it.

source/initial_temperature/adiabatic.cc
danieldouglas92: This still gives an error regarding the surface_boundary_set not being declared

```
use of undeclared identifier 'surface_boundary_set'; did you mean 'surface_boundary_id'?
          Utilities::AsciiDataBoundary<dim>::initialize(surface_boundary_set, 1);
```
It's not recognizing it because it's defined in the `initialize` function instead of the `initial_temperature` I guess?

source/initial_temperature/adiabatic.cc
naliboff: @danieldouglas92 - note the suggested fix above from @jdannberg

source/initial_temperature/adiabatic.cc
naliboff: @danieldouglas92 - was this comment in reference to the discussion above?

If so, can you confirm that the code gives the error listed above even after implementing?:
```
      if (read_from_ascii_file)
        Utilities::AsciiDataBoundary<dim>::initialize(surface_boundary_set,
```

source/initial_temperature/adiabatic.cc
```diff
@@ -33,17 +33,48 @@ namespace aspect
...
+        {
+          // The input ascii contains the age of the seafloor. User must provide ages in seconds
+          age_top = Utilities::AsciiDataBoundary<dim>::get_data_component(surface_boundary_id,
+                                                                          position,
+                                                                          0);
+        }
+      else
+        {
+          // convert input ages to seconds
+          age_top =    (this->convert_output_to_years() ? age_top_boundary_layer * year_in_seconds
```
naliboff, line 75:
```suggestion
          age_top = (this->convert_output_to_years() ? age_top_boundary_layer * year_in_seconds
```

naliboff: Adjust line below to match ...

source/initial_temperature/adiabatic.cc
```diff
@@ -99,18 +128,41 @@ namespace aspect
...
-                                                (T_bottom - adiabatic_bottom_temperature + subadiabaticity)
-                                                * erfc((this->get_geometry_model().maximal_depth()
-                                                        - this->get_geometry_model().depth(position)) /
-                                                       (2 * sqrt(kappa * age_bottom)))
-                                                : 0.0;
+      double surface_cooling_temperature = 0;
+      double bottom_heating_temperature = 0;
+      if (cooling_model == "half-space cooling")
+        {
+          // analytical solution for the thermal boundary layer from half-space cooling model
```
naliboff, line 135:
```suggestion
          // analytical solution for the thermal boundary layer for a half-space cooling model
```

source/initial_temperature/adiabatic.cc
naliboff: Could you write out the equation for the plate cooling model above this line in a comment? I find a short/compact form of the equation in a comment helps quite a bit in reading/debugging. Could you also do the same for the half-space cooling model above?

source/initial_temperature/adiabatic.cc
```diff
@@ -289,6 +344,15 @@ namespace aspect
                              "This function is one-dimensional and depends only on depth. The format of this "
                              "functions follows the syntax understood by the "
                              "muparser library, see Section~\\ref{sec:muparser-format}.");
+          prm.declare_entry ("Use ASCII file for seafloor age", "false",
+                             Patterns::Bool (),
+                             "Whether to define seafloor ages with an ASCII data file.");
```
naliboff, line 349:
```suggestion
                             "Whether to define seafloor ages with an ASCII data file. "
                              "Note that if the top boundary is not included in the list "
                              "of fixed temperature boundary indicators, the values"
                              "from the ascii file will be ignored and the initial temperature "
                              "profile will be entirely derived from the adiabatic profile.");
```

naliboff: @danieldouglas92 @jdannberg - How does this sound? Alternatively, we could put in an Assert that in order to use the ascii data option one must have the top boundary in the list of fixed temperature boundary indicators?

naliboff (commented):
@danieldouglas92 - This PR looks nearly ready to go aside from the few remaining comments and technical issues, thanks! Can you follow-up in the comments about the details of the `surface_boundary_set` issue?

John

source/initial_temperature/adiabatic.cc
jdannberg: I think this could go in the general description of the plugin. After all, the "Age top boundary layer" parameter will also be ignored if the temperature is not fixed at the top boundary, so this is not specific to reading in the ascii data file.

source/initial_temperature/adiabatic.cc
danieldouglas92: Yes when implementing 

`      if (read_from_ascii_file)
        Utilities::AsciiDataBoundary<dim>::initialize(surface_boundary_set,` 

I still get an error message

source/initial_temperature/adiabatic.cc
danieldouglas92: I've addressed all the other comments, besides the issue with the error loading in the ASCII file. What do you think is the best course of action regarding this?

PR #4185, opened 2021-07-12
thick_shell_grav_bench

cedrict:
The second gravity benchmark. Results will be analyzed in detail in a forthcoming publication so I keep it short wrt accuracy.


doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
 
+\subsubsection{Thick shell gravity benchmark}
+\label{sec:benchmark-thick-shell-gravity}
+
+\textit{This section was contributed by Cedric Thieulot.}
+
+This benchmark tests the accuracy of the gravity field and gravitational potential computed by the gravity postprocessor inside and outside an Earth-sized planet (without its core) of constant density. The domain is a spherical shell with inner radius $R\textsubscript{inner}=3840~\si{\km}$ and outer radius $R\textsubscript{outer}=6371~\si{\km}$. The density is constant in the domain and set to $\rho_0=3300~\si{\kg\per\cubic\metre}$.
+
+First, let us calculate the exact profile which we expect the benchmark to reproduce. The gravitational potential $U$ of a spherically symmetric object satisfies the Poisson equation $\Delta U = 4\pi G \rho(\mathbf r)$. For a constant density shell, this equation can be solved analytically for the gravitational acceleration and potential inside and outside the planet.
```
bangerth, line 9691: Shouldn't it be
```suggestion
First, let us calculate the exact profile which we expect the benchmark to reproduce. The gravitational potential $U$ of a spherically symmetric object satisfies the Poisson equation $-\Delta U = 4\pi G \rho(\mathbf r)$. For a constant density shell, this equation can be solved analytically for the gravitational acceleration and potential inside and outside the planet.
```
?

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+Outside the spherical shell the Poisson equation simplifies to the Laplace equation $\Delta U=0$:
```
bangerth, line 9692: Same here. At least among mathematicians, the convention is to use `-\Delta` instead of `\Delta`, because the former is a *positive* definite operator.

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+\begin{equation}
+\frac{1}{r^2} \frac{\partial }{\partial r} \left(r^2 \frac{\partial U}{\partial r} \right) = 0
```
bangerth, line 9694:
```suggestion
-\frac{1}{r^2} \frac{\partial }{\partial r} \left(r^2 \frac{\partial U}{\partial r} \right) = 0.
```

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+First, let us calculate the exact profile which we expect the benchmark to reproduce. The gravitational potential $U$ of a spherically symmetric object satisfies the Poisson equation $\Delta U = 4\pi G \rho(\mathbf r)$. For a constant density shell, this equation can be solved analytically for the gravitational acceleration and potential inside and outside the planet.
+Outside the spherical shell the Poisson equation simplifies to the Laplace equation $\Delta U=0$:
+\begin{equation}
+\frac{1}{r^2} \frac{\partial }{\partial r} \left(r^2 \frac{\partial U}{\partial r} \right) = 0 
+\end{equation}
+The solution to this expression is:
+\begin{equation}
+g=\frac{\partial U}{\partial r} = \frac{C}{r^2} \label{eq:app1},
+\end{equation}
+where $C$ is a constant of integration.
```
bangerth, line 9700: You can also add a constant. You make the assumption that the gravity potential is zero at infinity, which is reasonable by maybe worth stating?

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+In order to avoid an infinite gravity field at $r=0$ (where the density is also zero in this particular setup of a shell), we need to impose $C=0$, i.e. the gravity is zero for $r\leq R\textsubscript{inner}$. Another way of arriving at the same conclusion is to realize that $g$ is zero at the centre of the body because the material around it exerts an equal force in every direction.
```
bangerth, line 9701: I don't understand this argument. You derived this equation for the *outside* of the sphere, but here you talk about what happens at r=0?

bangerth, line 9701:
```suggestion
In order to avoid an infinite gravity field at $r=0$ (where the density is also zero in this particular setup of a shell), we need to impose $C=0$, i.e. the gravity is zero for $r\leq R\textsubscript{inner}$. Another way of arriving at the same conclusion is to realize that $g$ is zero at the center of the body because the material around it exerts an equal force in every direction.
```

bangerth: Oh, do you mean that this solution holds both *inside* and *outside* the sphere? If so, maybe it would be clearer to say that, and say that you have $C/r^2$ on the outside and $D/r^2$ on the inside, because these constants can be different. And that you want $D=0$ on the inside to keep the potential finite.

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+Inside the shell, $\rho=\rho_0$, yielding
```
bangerth, line 9702: The maybe say "within the interior of the shell, i.e., the volume occupied by material, ..."

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+\end{equation}
+where $A$ is another integration constant. 
+At the inner boundary, $r=R\textsubscript{inner}$ and $g=0$, allowing $A$ to be computed. Substituting in the value of $A$,
+\begin{equation}
+g=\frac{\partial U}{\partial r} = \frac{4 \pi}{3} G \rho_0 
+\left(r - \frac{R\textsubscript{inner}^3}{r^2} \right). \label{eqgin}
+\end{equation}
+When $r\geq R\textsubscript{outer}$, the gravitational potential is given by Eq. (\ref{eq:app1}). Requiring the gravity field to be continuous at $r=R\textsubscript{outer}$:
+\begin{equation}
+g(r) = \frac{G M}{r^2} \label{eqgout},
```
bangerth, line 9714:
```suggestion
g(r) = \frac{G M}{r^2} \label{eq:gout},
```
and everywhere you use it

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+\end{equation}
+where $M=\frac{4 \pi}{3} \rho_0(R\textsubscript{outer}^3-R\textsubscript{inner}^3)$ is the mass contained in the shell.
+For $r\ge R\textsubscript{outer}$, the potential is obtained by integrating Eq.(\ref{eqgout}):
+\begin{equation}
+U(r)=-\frac{GM}{r} +D,
+\end{equation}
+where $D$ is an integration constant which has to be zero since we require the potential to vanish for $r\rightarrow \infty$.
```
bangerth, line 9721: and also because you need it to be that way for the potential to be continuous at R_{outer}

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+\begin{equation}
+U(r)= \frac{4 \pi}{3} G \rho_0 \left(\frac{r^2}{2} + \frac{R\textsubscript{inner}^3}{r} \right)  + E,
+\end{equation}
+where $E$ is a constant. Continuity of the potential at $r=R\textsubscript{outer}$ requires that $E=-2\pi\rho_0 G R\textsubscript{outer}^2$. 
+Gravitational acceleration is zero for $r\leq R\textsubscript{inner}$, so the potential there is constant and a continuity requirement yields
+\begin{equation}
+U(r)=2\pi G \rho_0 (R\textsubscript{inner}^2-R\textsubscript{outer}^2).
+\end{equation}
+
+The gravity postprocessor in \aspect{} can be used to calculate the radial components of gravity ($g_r$ and $U$) at arbitrary points using the Sampling scheme `\textit{list of points}'. For this benchmark we calculate points along a line from the center of the planet to a distant point, $r=0$ to $r=10,000~\si{\km}$ (Figure \ref{fig:grav-thick-shell}). Arbitrarily, the latitude and longitude are both set to $13\si{\degree}$ so as to avoid potential measurement artefacts due to symmetry. The list of radii is defined as follows:
```
bangerth, line 9732:
```suggestion
The gravity postprocessor in \aspect{} can be used to calculate the radial components of gravity ($g_r$ and $U$) at arbitrary points using the sampling scheme `\textit{list of points}'. For this benchmark we calculate points along a line from the center of the planet to a distant point, $r=0$ to $r=10,000~\si{\km}$ (Figure \ref{fig:grav-thick-shell}). Arbitrarily, the latitude and longitude are both set to $13\si{\degree}$ so as to avoid potential measurement artifacts due to symmetry. The list of radii is defined as follows:
```

doc/manual/manual.tex
```diff
@@ -9681,6 +9681,69 @@ \subsubsection{Thin shell gravity benchmark}
...
+
+\lstinputlisting[language=prmfile]{cookbooks/benchmarks/gravity_thick_shell/doc/thick_shell.prm.out}
+
+The resulting measurements obtained for a mesh composed of 12 caps of $32^3$ cells (i.e., 393,216 total mesh cells) are shown in Fig.~(\ref{fig:grav-thick-shell}) and are in good agreement with the analytical profiles.
```
bangerth, line 9736: Is 12 correct? How do you get a coarse mesh for that?

bangerth (changes requested):
Nice!

doc/manual/manual.tex
cedrict: it's all a matter of convention... no minus sign on wiki: https://en.wikipedia.org/wiki/Poisson%27s_equation  , but a minus sign in Schubert, Turcotte, Olson, and again no minus sign in Gerya's book.

doc/manual/manual.tex
cedrict: I disagree: if you take the equation above, get rid of the 1/r^2 , and integrate once there is only a single constant that appears.

doc/manual/manual.tex
cedrict: yes, inside the shell rho=0 so Eq.(app1) holds (as it does outside). I added "Inside and outside the spherical shell (i.e. where $\rho=0$) the Poisson equation simplifies to the Laplace equation $\Delta U=0$ " a few lines above to clarify.

doc/manual/manual.tex
cedrict: done

doc/manual/manual.tex
cedrict: sure, but we have not yest computed U inside the mantle at that stage

doc/manual/manual.tex
cedrict: well, I learned something. Apparently both spellings are ok: https://en.wikipedia.org/wiki/Artifact_(error)  :)

doc/manual/manual.tex
cedrict: 393216/12=32768=12*32^3. Also it checks out visually.

doc/manual/manual.tex
bangerth: You're right. I had misread and thought you were talking about U, which will have two unknown constant.

doc/manual/manual.tex
bangerth: Maybe "interior/exterior of the shell" would be a term that is clear?

doc/manual/manual.tex
bangerth: Yes, I think artefact is British English and artifact American English: https://en.wiktionary.org/wiki/artifact

doc/manual/manual.tex
bangerth: I had thought that we start with either 6 or 24 or 96 cells. But I had to look it up and you're right that the choices are 6, 12 or 96. The default is 96, so what you really seem to have is 16^3 * 96?

doc/manual/manual.tex
cedrict: added: Inside ($r<R\textsubscript{inner}$) and outside ($r>R\textsubscript{outer}$) the spherical shell

doc/manual/manual.tex
cedrict: Screenshot from 2021-07-13 21-45-30

then it is confusing because this is the mesh and it looks like the citcom mesh which is made of 12 surface 'diamonds' of 16x16 elements.

bangerth: Ah, nice investigative work :-) I must be wrong!

bangerth:
You'll have to run the indentation script. I think that there are some spaces at the ends of lines that trip up the script!

bangerth (approved):

doc/manual/manual.tex
bangerth: Oh, since you need to run the indentation anyway, do me a favor and also still fix centre->center here. Good to go otherwise!

cedrict:
@bangerth ping? :)

bangerth:
OK to merge if you squash the commits.

bangerth:
Ping?

gassmoeller:
I took the liberty of rebasing and squashing. Should be ready to go once the testers are finished.

cedrict:
thanks again & sorry I was not faster to respond :)

gassmoeller:
No worries, I was just browsing through PRs that were essentially ready.

PR #4186, opened 2021-07-12
Add rising velocity depth average postprocessor

jdannberg:
There has been a sinking velocity depth average postprocessor for a long time, and I needed the same thing for a rising velocity (to look at the velocity of plumes and if they stop at phase transitions). This is what this pull request adds (with everything being copied from the sinking velocity version). 

I haven't created a test yet because I first wanted to see if there are existing tests that just output all depth average variables. 


source/postprocess/depth_average.cc
```diff
@@ -321,14 +321,22 @@ namespace aspect
...
-            "velocity magnitude|sinking velocity|Vs|Vp|"
+            "velocity magnitude|sinking velocity|rising velocity|Vs|Vp|"
             "viscosity|vertical heat flux|vertical mass flux|composition mass";
           prm.declare_entry("List of output variables", "all",
                             Patterns::MultipleSelection(variables.c_str()),
                             "A comma separated list which specifies which quantities to "
                             "average in each depth slice. It defaults to averaging all "
                             "available quantities, but this can be an expensive operation, "
-                            "so you may want to select only a few.\n\n"
+                            "so you may want to select only a few.\n"
```
bangerth, line 331: You'll have to keep the second `\n` or it won't matter to latex :-)

source/simulator/lateral_averaging.cc
```diff
@@ -736,6 +793,13 @@ namespace aspect
                                 &this->get_gravity_model(),
                                 this->convert_output_to_years()));
           }
+        else if (property_names[property_index] == "rising_velocity")
+          {
+            functors.push_back(std_cxx14::make_unique<FunctorDepthAverageRisingVelocity<dim>>
```
bangerth, line 798:
```suggestion
            functors.push_back(std::make_unique<FunctorDepthAverageRisingVelocity<dim>>
```

bangerth (commented):
Looks easy enough. Is it true that for an incompressible model with no flux across the boundary, the sum of sinking and rising velocities must be zero? That would make for a nice way to test both at the same time :-)

bangerth:
Well, there are 59 failing tests -- who would have thunk! :-) I'd be curious to hear what both you and I missed!

jdannberg:
I updated the tests (apparently there were quite a lot of tests that just used the default of outputting all depth average variables, which now also includes the rising velocity) and addressed your comments. Checking the tests, it turns out that sinking and rising velocity are the same even in compressible models (I assume they're all quite close to the reference adiabat). So this is only useful in models with open boundaries.

source/simulator/lateral_averaging.cc
jdannberg: I updated it in the other places too, just to be consistent.

include/aspect/lateral_averaging.h
```diff
@@ -248,7 +248,7 @@ namespace aspect
       get_velocity_magnitude_averages(std::vector<double> &values) const;
 
       /**
-       * Compute a lateral average of the current sinking velocity.
+       * Compute a lateral average of the current rising velocity.
```
bangerth, line 251: You'll have to undo this change, and make a change sinking->rising in line 261 so that it matches the function.

source/simulator/lateral_averaging.cc
```diff
@@ -712,52 +769,59 @@ namespace aspect
       {
         if (property_names[property_index] == "temperature")
           {
-            functors.push_back(std_cxx14::make_unique<FunctorDepthAverageField<dim>>
+            functors.push_back(std::make_unique<FunctorDepthAverageField<dim>>
```
bangerth, line 772: @tjhei reminded me that we can't do that yet :-( (See #4188 .)

bangerth (commented):
Sorry, a couple more changes.

Is one of the many test changes a new test? I feel like the additional output in all of these tests is probably good enough.

As for the analogy: I'm thinking of a fat pipe in which hot light material rises up and a narrow pipe in which cold dense material falls down. At the top, the material is cooled and at the bottom it is heated. We can adjust the two pipe diameters independently and if the rising and sinking velocities are chosen accordingly. The *mass fluxes* will of course balance, but the *average* velocities will not be of the same magnitude. So forget about my idea :-)

source/simulator/lateral_averaging.cc
jdannberg: Okay, I removed that from the last commit.

jdannberg:
Okay, this should be fixed now. 

But your idea of the rising and sinking velocity being the same led me to the following conclusion: looking at the test results, I couldn't find a single test where the rising and sinking velocity were different. So I looked at the code and tested this locally, and I think the code for the sinking velocity is wrong (it also computes the rising velocity; if I have a pipe with downwards flow, then the sinking velocity is zero). 

I also know how to fix this, do you want me to add that here or open a separate pull request?

gassmoeller:
> I also know how to fix this, do you want me to add that here or open a separate pull request?

Let's merge this and have the fix separately. Also add a changelog entry for both changes with the follow-up PR (I am merging this now so that you can start on the follow-up without conflicts).

PR #4187, opened 2021-07-12
Use std::unique_ptr instead of std::shared_ptr in a number of places.

bangerth:
/rebuild

### DIFF
--- a/source/material_model/utilities.cc
+++ b/source/material_model/utilities.cc
@@ -1151,7 +1151,7 @@ namespace aspect
         // Retrieve the list of composition names
         const std::vector<std::string> list_of_composition_names = this->introspection().get_composition_names();
 
-        n_phase_transitions_per_composition.reset(new std::vector<unsigned int>());
+        n_phase_transitions_per_composition = std::make_unique<std::vector<unsigned int>>();
 
         use_depth_instead_of_pressure = prm.get_bool ("Define transition by depth instead of pressure");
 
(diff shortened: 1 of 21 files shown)


source/material_model/utilities.cc
```diff
@@ -1151,7 +1151,7 @@ namespace aspect
         // Retrieve the list of composition names
         const std::vector<std::string> list_of_composition_names = this->introspection().get_composition_names();
 
-        n_phase_transitions_per_composition.reset(new std::vector<unsigned int>());
+        n_phase_transitions_per_composition = std::make_unique<std::vector<unsigned int>>();
```
anne-glerum, line 1154: Just out of curiosity, is it better to do an assignment than to reset?

anne-glerum (approved):

source/material_model/utilities.cc
bangerth: It makes no practical difference, but the C++ community has really tried to avoid using bare `new` operator because it is easy to introduce memory leak bugs when you forget the corresponding `delete` operator. As a consequence, all style guides these days suggest to use `std::make_shared` or `std::make_unique` where possible, avoid `new` and `delete`.

There is no risk here because we pass the newly created object to a class that manages its lifetime (the `std::unique_ptr` object), but I chose to just follow the usual style.

anne-glerum: Thanks for the explanation!

PR #4188, opened 2021-07-12
[deal.II 9.3] Use std::make_unique instead of std_cxx14::make_unique.

bangerth:
We had to write the latter before because `make_unique` didn't make it into C++11. But we now require C++14, so this is no longer necessary.

/rebuild


tjhei:
We currently only require 9.2, which works with c++11. We should wait until we make the switch to 9.3.

bangerth:
Ah, right, I had forgotten. Yes, we'll wait till we use deal.II 9.3.

bangerth:
When we get to this, we should also revert #4261.

tjhei (approved):
We have a lot more unique pointers than I thought!

PR #4191, opened 2021-07-13
Update Material properties plugin description

anne-glerum:
There are no visu plugins anymore that individually evaluate the material model (although you can still use the names in the list of visu output), so I've updated the description of this plugin. 

### For all pull requests:


source/postprocess/visualization/material_properties.cc
```diff
@@ -275,16 +275,17 @@ namespace aspect
...
-                                                  "for the material properties given by the material model."
-                                                  "There are a number of other visualization postprocessors "
-                                                  "that offer to write individual material properties. However, "
-                                                  "they all individually have to evaluate the material model. "
-                                                  "This is inefficient if one wants to output more than just "
-                                                  "one or two of the fields provided by the material model. "
+                                                  "for the material properties given by the material model. "
+                                                  "It is inefficient to output more than just "
+                                                  "one or two of the fields provided by the material model "
+                                                  "when the material model is evaluated again for each field. "
```
gassmoeller, line 281: Maybe we can also remove this part? There is no point in mentioning the inefficiency, if we cannot go the inefficient route anyway.

source/postprocess/visualization/material_properties.cc
```diff
@@ -275,16 +275,17 @@ namespace aspect
...
                                                   "The current postprocessor allows to output a (potentially "
                                                   "large) subset of all of the information provided by "
                                                   "material models at once, with just a single material model "
-                                                  "evaluation per output point."
+                                                  "evaluation per output point. "
+                                                  "Although individual properties can still be listed in the "
+                                                  "``List of output variables'', this visualization plugin "
+                                                  "is called internally anyway. "
```
gassmoeller, line 288:
```suggestion
                                                  "is called internally to avoid duplicated evaluations of the material model. "
```

gassmoeller (commented):
Thanks for the cleanup! I think this can be shortened even more. Do you agree?

anne-glerum:
Agreed!

tjhei (approved):

PR #4200, opened 2021-07-13
Migrate support points function to utilities

bobmyhill:
The `get_unit_support_points_for_velocity` function used to prescribe velocities was duplicated in all three plugins. This PR migrates that function into utilities.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -203,13 +203,12 @@ namespace aspect
      * The DGP element used when 'set Use locally conservative discretization = true' does not
      * have support points. If these elements are in use, a fictitious support point at the cell
      * center is returned for each shape function that corresponds to the pressure variable,
-     * whereas the support points for the velocity are correct; the fictitious points don't matter
+     * whereas the support points for the velocity are correct. The fictitious points don't matter
      * because we only use this function when interpolating the velocity variable, and ignore the
      * evaluation at the pressure support points.
      */
     template <int dim>
-    std::vector<Point<dim>>
-                         get_unit_support_points_for_velocity(const SimulatorAccess<dim> &simulator_access);
+    std::vector<Point<dim>> get_unit_support_points(const SimulatorAccess<dim> &simulator_access);
 
 
 
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -198,6 +198,19 @@ namespace aspect
     IndexSet extract_locally_active_dofs_with_component(const DoFHandler<dim> &dof_handler,
                                                         const ComponentMask &component_mask);
 
+    /**
+     * This function retrieves the unit support points (in the unit cell) for the current element.
+     * The DGP element used when 'set Use locally conservative discretization = true' does not
+     * have support points. If these elements are in use, a fictitious support point at the cell
+     * center is returned for each shape function that corresponds to the pressure variable,
+     * whereas the support points for the velocity are correct; the fictitious points don't matter
+     * because we only use this function when interpolating the velocity variable, and ignore the
+     * evaluation at the pressure support points.
+     */
+    template <int dim>
+    std::vector<Point<dim>>
+                         get_unit_support_points_for_velocity(const SimulatorAccess<dim> &simulator_access);
+
 
 
     namespace Coordinates
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -430,6 +430,47 @@ namespace aspect
 
 
 
+    template <int dim>
+    std::vector<Point<dim>>
+                         get_unit_support_points_for_velocity(const SimulatorAccess<dim> &simulator_access)
+    {
+      std::vector<Point<dim>> unit_support_points;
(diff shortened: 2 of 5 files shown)


include/aspect/utilities.h
```diff
@@ -198,6 +198,19 @@ namespace aspect
...
+     * The DGP element used when 'set Use locally conservative discretization = true' does not
+     * have support points. If these elements are in use, a fictitious support point at the cell
+     * center is returned for each shape function that corresponds to the pressure variable,
+     * whereas the support points for the velocity are correct; the fictitious points don't matter
+     * because we only use this function when interpolating the velocity variable, and ignore the
+     * evaluation at the pressure support points.
+     */
+    template <int dim>
+    std::vector<Point<dim>>
+                         get_unit_support_points_for_velocity(const SimulatorAccess<dim> &simulator_access);
```
gassmoeller, line 212: This indentation is a weird quirk of astyle. Just work around it by putting it all on one line.

Also reading through the documentation and the function itself it really returns the unit support points for all elements, right? At least I dont see where it limits itself to the Stokes variables. If so please rename to `get_unit_support_points` then.

source/utilities.cc
```diff
@@ -430,6 +430,47 @@ namespace aspect
 
 
 
+    template <int dim>
+    std::vector<Point<dim>>
+                         get_unit_support_points_for_velocity(const SimulatorAccess<dim> &simulator_access)
```
gassmoeller, line 435: same here, put on one line to avoid the weird indent

source/utilities.cc
```diff
@@ -430,6 +430,47 @@ namespace aspect
...
+          std::vector<Point<dim>> unit_support_points;
+          const unsigned int dofs_per_cell = simulator_access.get_fe().dofs_per_cell;
+          for (unsigned int dof=0; dof < dofs_per_cell; ++dof)
+            {
+              // base will hold element, base_index holds node/shape function within that element
+              const unsigned int
+              base       = simulator_access.get_fe().system_to_base_index(dof).first.first,
+              base_index = simulator_access.get_fe().system_to_base_index(dof).second;
+              // get the unit support points for the relevant element
+              std::vector<Point<dim>> my_support_points = simulator_access.get_fe().base_element(base).get_unit_support_points();
```
gassmoeller, line 454: Can you make this the following? That would avoid copying a vector unnecessarily.
```suggestion
              std::vector<Point<dim>> &my_support_points = simulator_access.get_fe().base_element(base).get_unit_support_points();
```

source/utilities.cc
```diff
@@ -430,6 +430,47 @@ namespace aspect
...
+        {
+          //special case for discontinuous pressure elements, which lack unit support points
+          std::vector<Point<dim>> unit_support_points;
+          const unsigned int dofs_per_cell = simulator_access.get_fe().dofs_per_cell;
+          for (unsigned int dof=0; dof < dofs_per_cell; ++dof)
+            {
+              // base will hold element, base_index holds node/shape function within that element
+              const unsigned int
+              base       = simulator_access.get_fe().system_to_base_index(dof).first.first,
+              base_index = simulator_access.get_fe().system_to_base_index(dof).second;
```
gassmoeller, line 452: I know this is just moved code, but now I cant unsee it :-).
```suggestion
              const unsigned int base_element = simulator_access.get_fe().system_to_base_index(dof).first.first,
              const unsigned int base_index = simulator_access.get_fe().system_to_base_index(dof).second;
```

source/utilities.cc
```diff
@@ -430,6 +430,47 @@ namespace aspect
 
 
 
+    template <int dim>
+    std::vector<Point<dim>>
+                         get_unit_support_points_for_velocity(const SimulatorAccess<dim> &simulator_access)
+    {
+      std::vector<Point<dim>> unit_support_points;
```
gassmoeller, line 437: I think this is superfluous, it will be redeclared in line 445.

source/utilities.cc
```diff
@@ -430,6 +430,47 @@ namespace aspect
...
+      if ( !simulator_access.get_parameters().use_locally_conservative_discretization )
+        {
+          return simulator_access.get_fe().get_unit_support_points();
+        }
+      else
+        {
+          //special case for discontinuous pressure elements, which lack unit support points
+          std::vector<Point<dim>> unit_support_points;
+          const unsigned int dofs_per_cell = simulator_access.get_fe().dofs_per_cell;
```
gassmoeller, line 446: Some speedup:

```suggestion
          const unsigned int dofs_per_cell = simulator_access.get_fe().dofs_per_cell;
          std::vector<Point<dim>> unit_support_points;
          unit_support_points.reserve(dofs_per_cell);
```

gassmoeller (commented):
Thanks for moving this function! I think there are a number of things that should be improved and since the function is short, could you give my suggestions a try? I realize all of this was there before, but since we move the function anyway, we might as well improve it in the process.

gassmoeller (approved):
Thanks! Looks good to go if the tests succeed.

Issue #4199, opened 2021-07-13
ASPECT on M1 ARM


tjhei:
TODO list:
- [x] write wiki page
- [x] fix "unrecognized instruction int $3" in catch.hpp
- [x] fix "uniform_int_distribution: error non-constant condition"
- [x] run a few tests
- [x] make plugins work

tjhei:
```
~/aspect/build$ ./aspect --version
-----------------------------------------------------------------------------
-- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.
--     . version 2.4.0-pre (catch_on_m1, 21ce336f9)
--     . using deal.II 10.0.0-pre (master, dce1df1fd0)
--     .       with 32 bit indices and vectorization level 0 (64 bits)
--     . using Trilinos 12.18.1
--     . using p4est 2.2.0
--     . running in DEBUG mode
--     . running with 1 MPI process
-----------------------------------------------------------------------------
~/aspect/build$ uname -a
Darwin Timos-Mac-mini.attlocal.net 20.5.0 Darwin Kernel Version 20.5.0: Sat May  8 05:10:31 PDT 2021; root:xnu-7195.121.3~9/RELEASE_ARM64_T8101 arm64
```
:-)

gassmoeller:
GMG on vectorization level 0 may be interesting ;-)

tjhei:
> GMG on vectorization level 0 may be interesting ;-)

yes, vectorization is not implemented. Will get to that after the hack. Runs, though.

gassmoeller:
> Runs, though.

That's the main thing. Pretty cool!

tjhei:
Here are the instructions: https://github.com/geodynamics/aspect/wiki/Installation-on-ARM-OSX

tjhei:
Things are working correctly for me, so I think we can close this issue. Further optimizations in deal.II will come...

gassmoeller:
Awesome, thanks!

PR #4201, opened 2021-07-13
fix catch compilation on M1 arm

tjhei:
part of #4199


gassmoeller (approved):

gassmoeller:
If you say that works I'll merge it :-). It does not affect other platforms.

bangerth (commented):
Could we have imported this from a newer upstream catch.hpp?

PR #4203, opened 2021-07-13
compute correct sinking velocity

jdannberg:
This fixes the sinking velocity depth average postprocessor. See #4186. 

### For all pull requests:


doc/modules/changes/20210713_jdannberg
```diff
@@ -0,0 +1,6 @@
+New: There is now a rising velocity output variable in the depth 
+average postprocessor.
+Fixed: The sinking velocity depth average postprocessor now computes
```
tjhei, line 3: This will look weird in the generated documentation.aybe two separate files?

source/simulator/lateral_averaging.cc
```diff
@@ -149,9 +149,9 @@ namespace aspect
           for (unsigned int q=0; q<output.size(); ++q)
             {
               const Tensor<1,dim> g = gravity_->gravity_vector(in.position[q]);
-              const Tensor<1,dim> vertical = (g.norm() > 0 ? g/g.norm() : Tensor<1,dim>());
+              const Tensor<1,dim> vertically_down = (g.norm() > 0 ? g/g.norm() : Tensor<1,dim>());
 
-              output[q] = std::fabs(std::min(0.0, velocity_values[q] * vertical))
+              output[q] = std::fabs(std::max(0.0, velocity_values[q] * vertically_down))
```
tjhei, line 154: Why do you need fabs? Max(0,x) is always positive...

doc/modules/changes/20210713_jdannberg
```diff
@@ -0,0 +1,6 @@
+New: There is now a rising velocity output variable in the depth 
+average postprocessor.
+Fixed: The sinking velocity depth average postprocessor now computes 
+the actual sinking velocity (and not the rising velocity).  
+<br>
```
tjhei, line 5: I don't see this change...

jdannberg: I changed this in #4186 and @gassmoeller merged it and told me to add the changelog entry in this follow-up.

jdannberg: Or do you mean the Fixed part? That's changing the std::min to std::max in the lateral_averaging.cc.

tjhei: Nevermind, I didn't check the other pr

tjhei (approved):
Holy moly, that is a lot of test change.

jdannberg:
I had to rebase to update the three tests that were failing because they were added on master in the meantime. But that should be fixed now. 

I tried to look at what changed in the tests, and it looks like in many of them the differences between rising and sinking velocity are small (which is what we would expect if there are no big density deviations from the reference profile). Some tests even do not change at all (#4186 had more tests that needed to be updated). In some tests the differences are larger, but most of the ones I saw did not have Earth-like parameters (for example, a viscosity of 1) and so it was hard to tell what to expect.

source/simulator/lateral_averaging.cc
```diff
@@ -149,9 +149,9 @@ namespace aspect
           for (unsigned int q=0; q<output.size(); ++q)
             {
               const Tensor<1,dim> g = gravity_->gravity_vector(in.position[q]);
-              const Tensor<1,dim> vertical = (g.norm() > 0 ? g/g.norm() : Tensor<1,dim>());
+              const Tensor<1,dim> vertically_down = (g.norm() > 0 ? g/g.norm() : Tensor<1,dim>());
 
-              output[q] = std::fabs(std::min(0.0, velocity_values[q] * vertical))
+              output[q] = std::max(0.0, velocity_values[q] * vertically_down)
```
bangerth, line 154: Oh, had this always been wrong!?!?

jdannberg: I think so. The last changes to that file were 6 years ago...
I think this was just hard to notice, because sinking and rising velocities are so similar when all boundaries are closed. I just noticed it because I computed the actual rising velocity, and one of the test cases was a pipe flow model.

PR #4207, opened 2021-07-14
use top instead of outer as boundary id

jdannberg:
A while ago we renamed all boundary names that were "outer" to "top". This updates the use of this in the adiabatic boundary plugin (which should allow it to be used with Cartesian geometries in the future).


gassmoeller (approved):
Thanks, looks good. No test changed, because currently we still support both names, but as you write "top" is geometry-independent.

bangerth (approved):

PR #4209, opened 2021-07-14
fixed stress update in damped elasticity

bobmyhill:
The stress update for the elasticity model was not updated to the correct expression when we added the viscous damper yesterday. This PR fixes that problem.


source/material_model/rheology/elasticity.cc
```diff
@@ -314,11 +314,20 @@ namespace aspect
 
                 // Calculate the current (new) viscoelastic stress, which is a function of the material
                 // properties (viscoelastic viscosity, shear modulus), elastic time step size, strain rate,
-                // vorticity and prior (inherited) viscoelastic stresses (see equation 29 in Moresi et al.,
-                // 2003, J. Comp. Phys.)
-                SymmetricTensor<2,dim> stress_new = ( 2. * average_viscoelastic_viscosity * deviator(in.strain_rate[i]) ) +
-                                                    ( ( average_viscoelastic_viscosity / calculate_elastic_viscosity(average_elastic_shear_moduli[i]) ) *
-                                                      (stress_old + dte * ( symmetrize(rotation * Tensor<2,dim>(stress_old) ) - symmetrize(Tensor<2,dim>(stress_old) * rotation) ) ) );
+                // vorticity, prior (inherited) viscoelastic stresses and viscosity of the elastic damper.
+                // In the absence of the elastic damper, this expression is identical to the one found in
```
naliboff, line 318:
```suggestion
                // In the absence of the elastic damper, the expression for "stress_new" below is identical to the one found in
```

source/material_model/rheology/elasticity.cc
```diff
@@ -314,11 +314,20 @@ namespace aspect
...
+                // Moresi et al. (2003, J. Comp. Phys., equation 29).
+                const double damped_elastic_viscosity = calculate_elastic_viscosity(average_elastic_shear_moduli[i]);
+
+                // stress_0 is the combination of the elastic stress stored at the end of the last time step and the stress change generated by local rotation
+                const SymmetricTensor<2,dim> stress_0 = (stress_old + dte * ( symmetrize(rotation * Tensor<2,dim>(stress_old) ) - symmetrize(Tensor<2,dim>(stress_old) * rotation) ) );
+
+                // stress_creep is the stress experienced by the viscous and elastic components.
+                const SymmetricTensor<2,dim> stress_creep = 2. * average_viscoelastic_viscosity * ( deviator(in.strain_rate[i]) + stress_0 / (2. * damped_elastic_viscosity ) );
+
+                // stress new is the (new) viscoelastic stress
```
naliboff, line 328:
```suggestion
                // stress_new is the (new) viscoelastic stress
```

naliboff (commented):
@bobmyhill - Nice catch and thanks for the fix! In addition to the test that makes sure this works, is there a problem we can think of with an analytical solution for the elastic damper? A test for this would help ensure we are computing the right values in addition to the code actually working.

bobmyhill:
Thanks @naliboff, 
I'll have a think about analytical solutions for this model for a future PR. The effective inverse viscosity should decay exponentially, so there should be a nice test there.

bobmyhill:
(I guess this would be a benchmark, rather than a test?)

naliboff (approved):
@bobmyhill - sounds good, let's indeed think about some benchmarks to add soon for this and the other dampers + iterative composite formulations! Ready to go when the tester finishes.

PR #4210, opened 2021-07-14
update dynamic topography postprocessor to use translate_symbolic_bou…

jaustermann:
…ndary_name_to_id to access top and bottom cells

The dynamic topography postprocessor wasn't using the translate_symbolic_boundara_name_to_id to find the top and bottom boundary and instead compared the vertex distance to the cell depth. Both should give the same but the former seems less error-prone so I replaced it. 

### For all pull requests:

### DIFF
--- a/source/postprocess/dynamic_topography.cc
+++ b/source/postprocess/dynamic_topography.cc
@@ -118,24 +118,20 @@ namespace aspect
             unsigned int face_idx = numbers::invalid_unsigned_int;
             for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
               {
-                const double depth_face_center = this->get_geometry_model().depth (cell->face(f)->center());
-                const double upper_depth_cutoff = cell->face(f)->minimum_vertex_distance()/3.0;
-                const double lower_depth_cutoff = this->get_geometry_model().maximal_depth() - cell->face(f)->minimum_vertex_distance()/3.0;
-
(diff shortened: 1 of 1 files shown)


source/postprocess/dynamic_topography.cc
```diff
@@ -247,34 +243,24 @@ namespace aspect
             // is true and will be changed to false if it's at the lower boundary. If the
             // cell is at neither boundary the loop will continue to the next cell.
             bool at_upper_surface = true;
-            for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
+            for (unsigned int f=0; f<GeometryInfo<3>::faces_per_cell; ++f)
```
gassmoeller, line 246: Why 3? Isnt this class allowed in 2d?

gassmoeller (commented):
Except for my one question this looks good.

source/postprocess/dynamic_topography.cc
jaustermann: Ah, thanks for catching that! I brought this over from the geoid postprocessor, which only works in 3D but this postprocessor is of course not limited to 3D so I will update that. Does this need a changelog entry?

gassmoeller: No need for a changelog entry, it is simply a better way to do the same thing.

gassmoeller (approved):

PR #4212, opened 2021-07-14
replace boost random by std

tjhei:
We require std c++11, so we might as well switch to the std functions
for random number generation

part of #4199


gassmoeller (approved):
Thanks

tjhei:
I should have known that test output will change...

tjhei:
I updated the test results but I have no idea if these changes are reasonable.

source/particle/generator/interface.cc
```diff
@@ -108,7 +108,7 @@ namespace aspect
       {
         // Uniform distribution on the interval [0,1]. This
         // will be used to generate random particle locations.
-        boost::uniform_01<double> uniform_distribution_01;
+        std::uniform_real_distribution uniform_distribution_01(0.0, 1.0);
```
bangerth, line 111: Allowed in C++17 but not before. You need this:
```suggestion
        std::uniform_real_distribution<double> uniform_distribution_01(0.0, 1.0);
```

gassmoeller (approved):
Thats alright, the only difference can be that the standard library's implementation of the random number generator is different from Boost's.

bangerth (commented):
Hard to tell whether the changes are correct or not. I looked carefully at the code and the changes look correct to me, with the exception of the one missing template argument.

Someone else wants to try and see whether the changes are reasonable?

bangerth:
Oh, @gassmoeller beat me to it. Too much thinking about the correctness :-)

I'll follow up with a patch that changes the missing template argument.

PR #4214, opened 2021-07-14
Viscous dissipation postprocessors

anne-glerum:
This PR adds two postprocessors:
1. One that computes the viscous dissipation over the whole domain
2. One that computes the viscous dissipation over the area that each of the compositional fields occupy.
3. 
I've tested them with the test setup at higher resolution, for which I know what the values should more or less be. I'll run that setup over time as well to see if the temporal evolution of the dissipation also matches that of the other codes that ran the same setup. 

Related to #4126. @bobmyhill In #4126 you mentioned deciding what to do about for example elasticity. At the moment, the plugin does not consider elasticity (as I don't need it for the benchmark I made this for). I can add this in the description of the benchmark, include asserts or actually include it. What other processes that store energy were you thinking of?  

### For all pull requests:


gassmoeller:
@anne-glerum how is the viscous_dissipation_statistics postprocessor different from the heating_statistics postprocessor? That one already outputs an integral of all individual heating plugins into the statistics file (shear heating being one of them). Could you use heating_statistics instead?

 I understand the purpose of the per_composition postprocessor.

anne-glerum:
@gassmoeller That's a good point. But for the heating_statistics postprocessor, the shear heating needs to be an actual active heating model, while the benchmark I made this for does not include shear heating in the temperature equation.

gassmoeller:
Oh, I see. Ok then, I will take a look later (maybe tomorrow :-)).

source/postprocess/viscous_dissipation_per_composition.cc
```diff
@@ -0,0 +1,177 @@
...
+      std::ostringstream output;
+      output.precision(4);
+      for (unsigned int c = 0; c < this->n_compositional_fields(); ++c)
+        {
+          if (this->convert_output_to_years() == true)
+            output << viscous_dissipation[c] * year_in_seconds
+                   << " J/year/m";
+          else
+            output << viscous_dissipation[c]
+                   << " W/m";
```
gassmoeller, line 149: The correct unit should be `W` (`eta * epsilon * epsilon * dV = Pa * s * 1/s * 1/s * m^3 = N * 1/s * m = J/s = W`)

source/postprocess/viscous_dissipation_per_composition.cc
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+
+      std::ostringstream output;
+      output.precision(4);
+      for (unsigned int c = 0; c < this->n_compositional_fields(); ++c)
+        {
+          if (this->convert_output_to_years() == true)
+            output << viscous_dissipation[c] * year_in_seconds
+                   << " J/year/m";
```
gassmoeller, line 146: I would not bother converting the years here, it is confusing. Nobody uses `J/year`, everyone uses `W (J/s)`.

source/postprocess/viscous_dissipation_per_composition.cc
```diff
@@ -0,0 +1,177 @@
...
+                        const SymmetricTensor<2, dim> deviatoric_strain_rate =
+                          (this->get_material_model().is_compressible()
+                           ? in.strain_rate[q] - 1. / 3. * trace(in.strain_rate[q]) * unit_symmetric_tensor<dim>()
+                           : in.strain_rate[q]);
+                        local_dissipation_integral[c] += (out.viscosities[q] * deviatoric_strain_rate) *
+                                                         deviatoric_strain_rate * fe_values.JxW(q);
+                      }
+                  }
+              }
+          }
```
gassmoeller, line 99: If you write it like this you save a lot of time, and can combine both postprocessors into this one:

```suggestion
                     const unsigned int n_compositional_fields = this->n_compositional_fields();     
                for (unsigned int q = 0; q < n_q_points; ++q)
                  {
                        // Viscous dissipation D:
                        // D = 1/2 * volume_integral(deviatoric_stress*deviatoric_strain_rate)
                        //   = 1/2 * volume_integral(2*eta*deviatoric_strain_rate*deviatoric_strain_rate)
                        //   = volume_integral(eta*deviatoric_strain_rate*deviatoric_strain_rate)
                        const SymmetricTensor<2, dim> deviatoric_strain_rate =
                          (this->get_material_model().is_compressible()
                           ? in.strain_rate[q] - 1. / 3. * trace(in.strain_rate[q]) * unit_symmetric_tensor<dim>()
                           : in.strain_rate[q]);
                        const double local_dissipation = (out.viscosities[q] * deviatoric_strain_rate) *
                                                         deviatoric_strain_rate * fe_values.JxW(q);
                        

                         // for this to work make the vector one element longer, and store the overall integral in
                         // the last element.                                                       
                        local_dissipation_integral[n_compositional_fields] += local_dissipation;
                      
                                  for (unsigned int c = 0; c<n_compositional_fields; ++c)
                                                      if (in.composition[q][c] >= 0.5)
                                                        local_dissipation_integral[c] += local_dissipation;
              }
          }
```

gassmoeller (commented):
This undoes a part of #2054. I understand the reason for this postprocessor (models without shear heating that want to compute viscous dissipation, or models that want to get viscous dissipation by composition), but I would prefer if we could keep it as one postprocessor. For most models `heating statistics` will be the better postprocessor to call, except for the few mentioned above. I would suggest to rename viscous_dissipation_statistics_per_composition into viscous_dissipation_statistics and let it compute both quantities (vd per composition and integrated vd). That way models also have a check to see if the viscous dissipation per composition adds up to the total viscous dissipation (as long as any point in the domain only has a single compositional field > 0.5). Do you think that would work for you?

include/aspect/postprocess/viscous_dissipation_per_composition.h
```diff
@@ -0,0 +1,53 @@
+/*
+  Copyright (C) 2011, 2012 by the authors of the ASPECT code.
```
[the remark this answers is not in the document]
anne-glerum: yep!

source/postprocess/viscous_dissipation_per_composition.cc
anne-glerum: Ah yes, we looked the 2d case only, then it would be W/m.

anne-glerum:
I've merged the two postprocessors and addressed the other comments as well.

gassmoeller (approved):
Nice, thanks!

PR #4216, opened 2021-07-14
update catch library to 2.13.6

tjhei:
based on https://github.com/geodynamics/aspect/pull/4201#pullrequestreview-705728250


bangerth (commented):
That file has gotten a LOT larger!

PR #4222, opened 2021-07-14
Fix missing template argument.

bangerth:
Follow-up to #4212. Necessary to make things compile on my end.

/rebuild


gassmoeller (approved):
Thanks for fixing my reviews :-), I relied on the tester, but that seems to use C++17 already.

tjhei:
Ah, sorry. I meant to only delete the double and keep <>.

tjhei (approved):

PR #4225, opened 2021-07-14
Sea level 2

ahollyday:
Sea level post processor with ascii topography read-in used for ocean function construction. 

The topography files is located in 'aspect/data/geometry-model/initial-topography-model/ascii-data/test/topo_new.txt'

@mfmweerdesteijn @jaustermann


gassmoeller:
@ahollyday I added WIP in the title to show other reviewers that this is not meant for review at this point yet. Our plan is to merge this manually with #3773 and create a combined postprocessor.

bangerth:
That's ok, but for the record, it is typically easier to review two smaller patches than one large one :-)

bangerth:
Separately:
* You've got a file `data/geometry-model/initial-topography-model/ascii-data/test/.DS_Store`. Please remove that from the patch.
* Did you really mean to add a text file with 250,000 lines? What does it represent?

mfmweerdesteijn:
What is this statement for?
if (this->get_geometry_model().natural_coordinate_system() == Utilities::Coordinates::spherical)
The topography ascii input data should be in longitude colatitude nd then third column the topo value. Doesn't have anything to do with the coordinate system of the model, does it?

gassmoeller:
@mfmweerdesteijn The statement you mentioned checks if the geometry model is best described in spherical coordinates (i.e. a spherical shell or a piece of it, or an ellipsoid). This is used by the StructuredDataLookup class to determine if the first two coordinate columns are interpreted as spherical coodinates or cartesian coordinates. So yes it is needed. I am not sure if the postprocessor you two are working on has any meaning for box models (you can disable it for cartesian coordinates if you think not), but you need to at least convert the cartesian point `position` into spherical coordinates `spherical_position`.

gassmoeller:
The if-statement was copied from another place that works for both cartesian and spherical geometries. If we assert that this postprocessor is only run in spherical geometries the if-statement can be removed.

bangerth:
Now that @mfmweerdesteijn 's #3773 has been merged, what should we do with this one? How does the material in this PR differ from what's in #3773?

Issue #4205, opened 2021-07-14
newton_solver_benchmark_set issues


zjiaqi2018:
report issues on benchmarks/newton_solver_benchmark_set/nonlinear_channel_flow
input_v.prm doesn't converge:
![image](https://user-images.githubusercontent.com/36655731/125548706-e55f29d1-84f2-4ab1-8f4f-9be35042d85a.png)

input_t.prm:

--------------------------------------------------------
An error occurred in line <618> of file </home/jiaqi2/aspect/aspect/source/simulator/simulator_access.cc> in function
    const aspect::NewtonHandler<dim>& aspect::SimulatorAccess<dim>::get_newton_handler() const [with int dim = 2]
The violated condition was: 
    simulator->newton_handler.get() != nullptr
Additional information: 
    You can not call this function if the Newton solver is not enabled.

MFraters:
Thanks for making the issue. I have tired to reproduce the issue. I had some problems with the shared library, so I just put the file into the material model directory and compiled it. I don't think that matters for this issue. 

I can reproduce the issue for input_v.prm. This seems to be very specific to a grid level of 6 and 4 mpi processes. I also tried it with 1,2,3,5 and 6 mpi processes and they work fine. On grid level 5 and 4 mpi processes it it works and on grid level 7 with 4 mpi processes it fails, but with 3 and 5 mpi processes it works. I am not sure what is going on here.

The second errorI have been able to reproduce. It it caused by the spd post processor. For now you can turn it off, since it is not really needed. I can investigate why it is not working.

zjiaqi2018:
For the second case, are you suggesting removing `set Stabilization preconditioner = SPD`? It doesn't work for me.

MFraters:
Ah, sorry. I should have been more precise. In  `subsection Postprocess`->`subsection Visualization`, remove from `set List of output variables = ` the value `spd factor`.

MFraters:
I have found the issue. The input file uses `iterated Advection and Stokes` by default instead of `iterated Advection and Newton Stokes`. The reason why this has not been an issue for me is because I always used the metabash.sh and bash.sh scripts to replace it automatically. I think I should:
1.  add an assert in this plugin which checks that a newton type solver is used
2. fix the default input file to actually use the newton solver.

PR #4226, opened 2021-07-15
Assert SPD factor visualization against full Stokes solver.

MFraters:
Addressing one of the two issues in #4205. This adds an assert to prevent the SPD factor visualization plugin to be used in when we know that the  `simulator->newton_handler.get()` is a  `nullptr`, which is when a non-defect correction type nonlinear solver is used.

### DIFF
--- a/include/aspect/postprocess/visualization/spd_factor.h
+++ b/include/aspect/postprocess/visualization/spd_factor.h
@@ -56,6 +56,13 @@ namespace aspect
           void
           evaluate_vector_field(const DataPostprocessorInputs::Vector<dim> &input_data,
                                 std::vector<Vector<double>> &computed_quantities) const override;
+
+          /**
+           * Read the parameters this class declares from the parameter file.
+           */
+          void
+          parse_parameters (ParameterHandler &prm) override;
+
       };
     }
   }
--- a/source/postprocess/visualization/spd_factor.cc
+++ b/source/postprocess/visualization/spd_factor.cc
@@ -73,6 +73,15 @@ namespace aspect
                                                                            this->get_newton_handler().parameters.SPD_safety_factor);
           }
       }
+
+      template <int dim>
+      void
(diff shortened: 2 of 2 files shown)


bangerth (approved):

bangerth:
I had to think about whether it is safe to assert in that place or whether that depends on the assumption that this class gets to read its parameters *after* the solver scheme has been set correctly. Are we sure that this is the case? There is probably little harm here: It must clearly work at the moment if the tests pass, and if the order changes in the future, then we will get a spuriously triggered assertion. That will be obvious to debug.

For the record, I would have probably put this into the `evaluate_vector_field()` function, but I don't think it matters.

/rebuild

MFraters:
Thanks for the quick review. I have tested it, and the assert was triggered with a full Stokes solver. Also, if it somehow manages to pass this one (I don't see how), the other assert is still there.

> For the record, I would have probably put this into the evaluate_vector_field() function, but I don't think it matters.

Do you think the value could change during the simulation?

 I put in in parse parameters because in this case I can use an AssertThrow pretty much for free. If I would have put it the evaluate function (called at every location) it would either be just an Assert or there would be a performance impact in release mode. It would probably be tiny, but this way of thinking is the (unfortunate) results of weeks of trying to optimizing the world builder ;)

bangerth:
On 7/14/21 7:17 PM, Menno Fraters wrote:
> Do you think the value could change during the simulation?

I thought about that as well, but we don't currently support that.


> I put in in parse parameters because in this case I can use an AssertThrow 
> pretty much for free. If I would have put it the evaluate function (called at 
> every location) it would either be just an Assert or there would be a 
> performance impact in release mode. It would probably be tiny, but this way of 
> thinking is the (unfortunate) results of weeks of trying to optimizing the 
> world builder ;)

:-)

I think it's ok where it is.

Issue #4193, opened 2021-07-13
Error: Negative diffusion viscosity detected;  but only use dislocation in VP material model


sibiaoliu:
I set " Viscous flow law = dislocation" in the material model of Visco Plastic. The model crashes with an error message: 

"The violated condition was: 
    viscosity_diffusion > 0.0
Additional information: 
    Negative diffusion viscosity detected. This is unphysical and should
    not happen. Check for negative parameters."

I think that the diffusion calculation should not be implemented or it is calculated with default values since the 'diffusion viscous flow' option is not enabled. Can anyone give me some advice on the reason behind this error? Has anyone encountered the same problem before? Thanks!

<img width="1191" alt="iShot2021-07-13 13 32 26" src="https://user-images.githubusercontent.com/36626697/125445830-235f1e42-9393-4a80-a4ef-4ba2609f6e31.png">

anne-glerum:
Hi @sibiaoliu, looking at /material_model/rheology/visco_plastic.cc, the diffusion viscosity is still computed (line 150) even if it's not used later on (in the switch starting from line 161).

Did you set the diffusion creep parameters in the input file? The defaults seem ok.

sibiaoliu:
> Hi @sibiaoliu, looking at /material_model/rheology/visco_plastic.cc, the diffusion viscosity is still computed (line 150) even if it's not used later on (in the switch starting from line 161).
> 
> Did you set the diffusion creep parameters in the input file? The defaults seem ok.

Hi Anne, I did not set the diffusion parameters. I thought that the values of the pressure or temperature solutions may cause this error. However, When I check the output, the values look correctly. It is weird.

sibiaoliu:
Find the reason for this issue: the temperature in a few points is negative, which should be wrong.

PR #4229, opened 2021-07-15
Avoid some viscosity computations when not necessary.

bangerth:
Based on an observation by @naliboff in https://github.com/geodynamics/aspect/issues/4193#issuecomment-879061394 . I surrounded the whole block by curly braces to limit the scope of the variables I'm touching to convince myself that they are not actually used again further down below in places I had not considered.

/rebuild

### DIFF
--- a/source/material_model/rheology/visco_plastic.cc
+++ b/source/material_model/rheology/visco_plastic.cc
@@ -132,70 +132,84 @@ namespace aspect
         for (unsigned int j=0; j < volume_fractions.size(); ++j)
           {
             // Step 1: viscous behavior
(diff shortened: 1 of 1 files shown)


bangerth:
Best viewed via https://github.com/geodynamics/aspect/pull/4229/files?w=1

MFraters (approved):
I think you meant @anne-glerum's comment? Anyway, looks good to me. 

Just to note, you are creating extra branches, but I am quite sure the computation is more than expensive enough that this is worth it.

I am also fine with leaving the curly braces there if it means the compiler will do some more sanity checks :)

PR #4231, opened 2021-07-15
Avoid introducing alias variables.

bangerth:
Style guides typically recommend not to use alias variables, i.e. stuff like
```
  const Introspection<dim> &introspection = this->introspection();
```
because at the places where this alias variable is used one is left to wonder "which introspection" is this? It's just easier to read with the original variable -- no ambiguity.

This patch removes these alias variables from various assemblers. The downside is that lines get longer, so the patch is probably debatable. What do others think? In various places where I touch the code, there are also aliases for the finite element, which I could remove in a follow-up.

/rebuild

### DIFF
--- a/source/simulator/assemblers/newton_stokes.cc
+++ b/source/simulator/assemblers/newton_stokes.cc
@@ -48,7 +48,6 @@ namespace aspect
       internal::Assembly::Scratch::StokesPreconditioner<dim> &scratch = dynamic_cast<internal::Assembly::Scratch::StokesPreconditioner<dim>& > (scratch_base);
       internal::Assembly::CopyData::StokesPreconditioner<dim> &data = dynamic_cast<internal::Assembly::CopyData::StokesPreconditioner<dim>& > (data_base);
 
-      const Introspection<dim> &introspection = this->introspection();
       const FiniteElement<dim> &fe = this->get_fe();
       const unsigned int stokes_dofs_per_cell = data.local_dof_indices.size();
       const unsigned int n_q_points           = scratch.finite_element_values.n_quadrature_points;
@@ -59,7 +58,7 @@ namespace aspect
       // save the component (pressure and dim velocities) each belongs to.
       for (unsigned int i = 0, i_stokes = 0; i_stokes < stokes_dofs_per_cell; /*increment at end of loop*/)
         {
-          if (introspection.is_stokes_component(fe.system_to_component_index(i).first))
+          if (this->introspection().is_stokes_component(fe.system_to_component_index(i).first))
             {
               scratch.dof_component_indices[i_stokes] = fe.system_to_component_index(i).first;
               ++i_stokes;
@@ -73,12 +72,12 @@ namespace aspect
         {
           for (unsigned int i = 0, i_stokes = 0; i_stokes < stokes_dofs_per_cell; /*increment at end of loop*/)
             {
-              if (introspection.is_stokes_component(fe.system_to_component_index(i).first))
+              if (this->introspection().is_stokes_component(fe.system_to_component_index(i).first))
                 {
                   scratch.grads_phi_u[i_stokes] =
-                    scratch.finite_element_values[introspection.extractors
+                    scratch.finite_element_values[this->introspection().extractors
                                                   .velocities].symmetric_gradient(i, q);
-                  scratch.phi_p[i_stokes] = scratch.finite_element_values[introspection
+                  scratch.phi_p[i_stokes] = scratch.finite_element_values[this->introspection()
                                                                           .extractors.pressure].value(i, q);
 
 #if DEBUG
@@ -262,7 +261,6 @@ namespace aspect
       internal::Assembly::Scratch::StokesSystem<dim> &scratch = dynamic_cast<internal::Assembly::Scratch::StokesSystem<dim>& > (scratch_base);
       internal::Assembly::CopyData::StokesSystem<dim> &data = dynamic_cast<internal::Assembly::CopyData::StokesSystem<dim>& > (data_base);
 
-      const Introspection<dim> &introspection = this->introspection();
       const FiniteElement<dim> &fe = this->get_fe();
       const unsigned int stokes_dofs_per_cell = data.local_dof_indices.size();
       const unsigned int n_q_points    = scratch.finite_element_values.n_quadrature_points;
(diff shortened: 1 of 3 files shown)


tjhei:
This is in an inner loop. Are we sure the compiler can inline the call without causing any overhead?

bangerth:
The function isn't currently `inline`, but it looks like this:
```
  template <int dim>
  const Introspection<dim> &
  SimulatorAccess<dim>::introspection () const
  {
    return simulator->introspection;
  }
```
I am pretty certain that even calling this function out-of-line is far far cheaper than anything else that is happening in these loops :-) But if it would make you feel better, I'd be happy to move that function into the `.h` file!

bangerth:
See #4234.

tjhei:
I am wondering if @gassmoeller introduced this after profiling...

gassmoeller:
I am not sure if I introduced this, but if I did I am pretty sure it was not for performance reasons (more for making the code easier to read). I dont know how much the additional function call matters, but assembly at the moment is already a quite slow part of ASPECT, it would be good to make sure to not slow it down further. 

That being said I personally find the old code is also easier to read, but maybe that is just me. There is *exactly* one introspection in all of our models and the likelihood of that ever changing is really close to zero. But if you feel we should follow the guidelines I will not stand in the way.

bangerth:
I'll let others choose what to do here. I personally can't see how jumping to an address and back (the cost of a non-inlined function) would even prove measurable in terms of run time.

As far as inlining the function in #4234: I tried this for a while, but it introduces cyclic dependencies between `Introspection`, `Simulator`, and `SimulatorAccess`. I don't think this is feasible but, as mentioned above, I don't think it will matter either.

bangerth:
This is interesting -- here are instruction counts without (first) and with (second) this patch for a version of the `shell_simple_2d.prm` in which I'm only running the first 15 time steps (each run twice because, it seems, these numbers change from run to run):

BEFORE:
```
==49747== I   refs:      79,266,464,572
==49747== I1  misses:       216,050,043
==49747== LLi misses:            74,410
==49747== I1  miss rate:           0.27%
==49747== LLi miss rate:           0.00%
==49747== 
==49747== D   refs:      32,148,910,714  (26,269,693,358 rd   + 5,879,217,356 wr)
==49747== D1  misses:     1,487,914,915  ( 1,305,933,424 rd   +   181,981,491 wr)
==49747== LLd misses:         2,672,916  (     1,028,744 rd   +     1,644,172 wr)
==49747== D1  miss rate:            4.6% (           5.0%     +           3.1%  )
==49747== LLd miss rate:            0.0% (           0.0%     +           0.0%  )
==49747== 
==49747== LL refs:        1,703,964,958  ( 1,521,983,467 rd   +   181,981,491 wr)
==49747== LL misses:          2,747,326  (     1,103,154 rd   +     1,644,172 wr)
==49747== LL miss rate:             0.0% (           0.0%     +           0.0%  )

==50097== I   refs:      79,275,082,394
==50097== I1  misses:       216,227,937
==50097== LLi misses:            74,582
==50097== I1  miss rate:           0.27%
==50097== LLi miss rate:           0.00%
==50097== 
==50097== D   refs:      32,148,197,990  (26,269,841,915 rd   + 5,878,356,075 wr)
==50097== D1  misses:     1,487,277,841  ( 1,304,785,355 rd   +   182,492,486 wr)
==50097== LLd misses:         2,673,887  (     1,028,783 rd   +     1,645,104 wr)
==50097== D1  miss rate:            4.6% (           5.0%     +           3.1%  )
==50097== LLd miss rate:            0.0% (           0.0%     +           0.0%  )
==50097== 
==50097== LL refs:        1,703,505,778  ( 1,521,013,292 rd   +   182,492,486 wr)
==50097== LL misses:          2,748,469  (     1,103,365 rd   +     1,645,104 wr)
==50097== LL miss rate:             0.0% (           0.0%     +           0.0%  )
```

AFTER:
```
==48306== I   refs:      79,591,750,551
==48306== I1  misses:       219,162,928
==48306== LLi misses:            74,434
==48306== I1  miss rate:           0.28%
==48306== LLi miss rate:           0.00%
==48306== 
==48306== D   refs:      32,384,498,015  (26,434,489,350 rd   + 5,950,008,665 wr)
==48306== D1  misses:     1,491,235,047  ( 1,308,309,870 rd   +   182,925,177 wr)
==48306== LLd misses:         2,673,023  (     1,028,805 rd   +     1,644,218 wr)
==48306== D1  miss rate:            4.6% (           4.9%     +           3.1%  )
==48306== LLd miss rate:            0.0% (           0.0%     +           0.0%  )
==48306== 
==48306== LL refs:        1,710,397,975  ( 1,527,472,798 rd   +   182,925,177 wr)
==48306== LL misses:          2,747,457  (     1,103,239 rd   +     1,644,218 wr)
==48306== LL miss rate:             0.0% (           0.0%     +           0.0%  )

==50529== I   refs:      79,596,168,923
==50529== I1  misses:       219,166,267
==50529== LLi misses:            74,400
==50529== I1  miss rate:           0.28%
==50529== LLi miss rate:           0.00%
==50529== 
==50529== D   refs:      32,384,424,573  (26,434,281,421 rd   + 5,950,143,152 wr)
==50529== D1  misses:     1,486,832,978  ( 1,304,872,838 rd   +   181,960,140 wr)
==50529== LLd misses:         2,673,018  (     1,028,772 rd   +     1,644,246 wr)
==50529== D1  miss rate:            4.6% (           4.9%     +           3.1%  )
==50529== LLd miss rate:            0.0% (           0.0%     +           0.0%  )
==50529== 
==50529== LL refs:        1,705,999,245  ( 1,524,039,105 rd   +   181,960,140 wr)
==50529== LL misses:          2,747,418  (     1,103,172 rd   +     1,644,246 wr)
==50529== LL miss rate:             0.0% (           0.0%     +           0.0%  )
```

So apparently the patch increased the number of instructions executed by about 323 million, or 0.4%. That comes as a surprise to me. It's a shame we can't inline this function.

So, scrap the whole idea?

tjhei:
The difference is likely bigger for 3d models, so my vote would be against this patch.
Sorry, Wolfgang.

gassmoeller:
It's often surprising what a compiler cannot optimize and how much of a difference it makes. I am also for closing this PR.

bangerth:
So it is!

PR #4232, opened 2021-07-15
Move simple nonlinear to newton nonlinear channel

### LINKED ISSUE #2861: missing file or readme in benchmarks/nonlinear_channel_flow
I know that this plugin is used inside ``newton_solver_benchmark_set/nonlinear_channel_flow/`` but there is no readme to explain that. Ideally, we would put a simple test .prm file in here.

MFraters:
This resolves #2861 by removing the separate nonlinear channel flow directory and merging it with the one in the newton solver benchmark set.


tjhei:
I think we will need to adjust benchmarks/check.mk
Let me know if you need help with that.

MFraters:
> I think we will need to adjust benchmarks/check.mk
> Let me know if you need help with that.

I think that I have updated it correctly, but please check it. I have removed the line stating to make_lib in the `benchmarks/nonlinear_channel_flow`. I don't think I need to add it for the new location, because it already seems to do it for that folder. Let me know if I misinterpreted it :)

MFraters:
hmm, if I remove that line, it actually fail, while if I keep it there it works. So I removed it for now. @tjhei, let me know what you want to have changed in that file.

tjhei:
If testers pass, we can merge and I will follow up with a fix for that.

PR #4234, opened 2021-07-15
Make SimulatorAccess::introspection() 'inline'.

bangerth:
We've got more than 500 places where this function is called, many in inner loops. Might as well remove the function call overhead for this one :-)

/rebuild

### DIFF
--- a/include/aspect/simulator_access.h
+++ b/include/aspect/simulator_access.h
@@ -934,6 +934,18 @@ namespace aspect
       const Simulator<dim> *simulator;
   };
 
+
+
+  template <int dim>
+  inline
+  const Introspection<dim> &
+  SimulatorAccess<dim>::introspection () const
+  {
+    return simulator->introspection;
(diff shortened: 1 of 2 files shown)


MFraters (approved):
I have certainly noticed that this kind of in lining can significantly help with the performance. Looks good to me once it is indented. 

Just out of curiosity, did you notice any difference in run time in a small test case, or would you only expect this to matter for really large runs?

gassmoeller (approved):

gassmoeller (commented):
Nice idea, but Simulator is forward declared in simulator_access.h to avoid including simulator.h in nearly every plugin. Is there a way around this?

bangerth:
I tried this for a while, but it introduces cyclic dependencies between Introspection, Simulator, and SimulatorAccess. I don't think this is feasible without a lot more work or without putting things into the same file :-(

PR #4236, opened 2021-07-15
Updating viscoplastic_strain_invariants particles

erinheilman:
I updated the viscoplastic_strain_invariants particle property to follow the method in the elastic stress particle property, this allows strain invariants to evaluate the viscoplastic material model and use outputs to update the reaction terms so that code does not have to be duplicated to include strain healing in particles.

### For all pull requests:

* [X ] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/master/tests/) directory.

### DIFF
--- a/include/aspect/particle/property/viscoplastic_strain_invariants.h
+++ b/include/aspect/particle/property/viscoplastic_strain_invariants.h
@@ -101,6 +101,7 @@ namespace aspect
            * but it is currently not used in a threaded context.
            */
           mutable MaterialModel::MaterialModelInputs<dim> material_inputs;
+          mutable MaterialModel::MaterialModelOutputs<dim> material_outputs;
       };
     }
   }
--- a/source/particle/property/viscoplastic_strain_invariants.cc
+++ b/source/particle/property/viscoplastic_strain_invariants.cc
@@ -34,7 +34,8 @@ namespace aspect
       ViscoPlasticStrainInvariant<dim>::ViscoPlasticStrainInvariant ()
         :
         n_components(0),
-        material_inputs(1,0)
+        material_inputs(1,0),
+        material_outputs(1,0)
       {}
 
 
@@ -50,6 +51,8 @@ namespace aspect
 
         n_components = 0;
         material_inputs = MaterialModel::MaterialModelInputs<dim>(1,this->n_compositional_fields());
+        material_outputs = MaterialModel::MaterialModelOutputs<dim>(1, this->n_compositional_fields());
+
 
         // Find out which fields are used.
         if (this->introspection().compositional_name_exists("plastic_strain"))
@@ -96,8 +99,8 @@ namespace aspect
                                                                  const std::vector<Tensor<1,dim>> &gradients,
                                                                  typename ParticleHandler<dim>::particle_iterator &particle) const
       {
-        // Current timestep
-        const double dt = this->get_timestep();
+        // Current timestep shouldn't need this anymore
+        //const double dt = this->get_timestep();
 
         // Velocity gradients
         Tensor<2,dim> grad_u;
@@ -118,11 +121,16 @@ namespace aspect
           }
 
         // Find out plastic yielding by calling function in material model.
-        const MaterialModel::ViscoPlastic<dim> &viscoplastic
-          = Plugins::get_plugin_as_type<const MaterialModel::ViscoPlastic<dim>>(this->get_material_model());
+        //const MaterialModel::ViscoPlastic<dim> &viscoplastic
+        //  = Plugins::get_plugin_as_type<const MaterialModel::ViscoPlastic<dim>>(this->get_material_model());
 
-        const bool plastic_yielding = viscoplastic.is_yielding(material_inputs);
+        //const bool plastic_yielding = viscoplastic.is_yielding(material_inputs);
+        
+        // Evaluate directly in the viscoplastic material model and modify the reaction outputs
+        this->get_material_model().evaluate (material_inputs,material_outputs);
 
+        for (unsigned int i = 0; i < SymmetricTensor<2,dim>::n_independent_components ; ++i)
+          particle->get_properties()[data_position + i] += material_outputs.reaction_terms[0][i];
 
         /* Next take the integrated strain invariant from the prior time step. When
          * there are two fields (plastic and viscous), this assumes plastic
@@ -130,7 +138,7 @@ namespace aspect
          * In this case old_strain will first be given the plastic strain, and then,
          * if there is no plastic yielding it will update to the viscous strain instead.
          */
-        auto &data = particle->get_properties();
+       /* auto &data = particle->get_properties();
         double old_strain = data[data_position];
         if (n_components == 2 && plastic_yielding == false)
           old_strain = data[data_position+(n_components-1)];
@@ -142,13 +150,13 @@ namespace aspect
         // New strain is the old strain plus dt*edot_ii
         const double new_strain = old_strain + dt*edot_ii;
 
-
+      */
         /* Once we know whether the particle underwent plastic, viscous, or
          * total strain assign the new strain to the correct data position.
          * NOTE: This assumes that total strain cannot be used in combination with the
          * other fields. If this changes in the future, this will need to be updated.
          * */
-        if (this->introspection().compositional_name_exists("plastic_strain") && plastic_yielding == true)
+      /*  if (this->introspection().compositional_name_exists("plastic_strain") && plastic_yielding == true)
           data[data_position] = new_strain;
 
         if (this->introspection().compositional_name_exists("viscous_strain") && plastic_yielding == false)
@@ -156,6 +164,7 @@ namespace aspect
 
         if (this->introspection().compositional_name_exists("total_strain"))
           data[data_position] = new_strain;
+      */
       }
 
 
--- a/source/particle/property/viscoplastic_strain_invariants.cc
+++ b/source/particle/property/viscoplastic_strain_invariants.cc
@@ -120,51 +120,11 @@ namespace aspect
             material_inputs.composition[0][i] = solution[this->introspection().component_indices.compositional_fields[i]];
           }
 
-        // Find out plastic yielding by calling function in material model.
-        //const MaterialModel::ViscoPlastic<dim> &viscoplastic
-        //  = Plugins::get_plugin_as_type<const MaterialModel::ViscoPlastic<dim>>(this->get_material_model());
-
-        //const bool plastic_yielding = viscoplastic.is_yielding(material_inputs);
-        
         // Evaluate directly in the viscoplastic material model and modify the reaction outputs
         this->get_material_model().evaluate (material_inputs,material_outputs);
 
         for (unsigned int i = 0; i < SymmetricTensor<2,dim>::n_independent_components ; ++i)
           particle->get_properties()[data_position + i] += material_outputs.reaction_terms[0][i];
-
-        /* Next take the integrated strain invariant from the prior time step. When
-         * there are two fields (plastic and viscous), this assumes plastic
-         * strain is always in data position one and viscous strain in position two.
-         * In this case old_strain will first be given the plastic strain, and then,
-         * if there is no plastic yielding it will update to the viscous strain instead.
-         */
-       /* auto &data = particle->get_properties();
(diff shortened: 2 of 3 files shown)


source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -96,8 +99,8 @@ namespace aspect
                                                                  const std::vector<Tensor<1,dim>> &gradients,
                                                                  typename ParticleHandler<dim>::particle_iterator &particle) const
       {
-        // Current timestep
-        const double dt = this->get_timestep();
+        // Current timestep shouldn't need this anymore
+        //const double dt = this->get_timestep();
```
gassmoeller, line 103: Remove these lines

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -117,45 +120,11 @@ namespace aspect
...
-        auto &data = particle->get_properties();
-        double old_strain = data[data_position];
-        if (n_components == 2 && plastic_yielding == false)
-          old_strain = data[data_position+(n_components-1)];
-
-
-        // Calculate strain rate second invariant
-        const double edot_ii = std::sqrt(std::fabs(second_invariant(deviator(material_inputs.strain_rate[0]))));
+        // Evaluate directly in the viscoplastic material model and modify the reaction outputs
+        this->get_material_model().evaluate (material_inputs,material_outputs);
```
gassmoeller, line 124: I like that we now rely on the computation in the material model, which should reduce potential sources of error. However, I remember that this may be way more expensive than the old way to do things (I think the property was written in the current form, because it was much faster this way. 

Do you have a simple model that uses strain healing and that you can compare between the current master branch and your branch? Take a particular look at the "Particles: Update properties" timing section and if that has changed. Dont use old model timing results, because #4060 and #4062 also got merged in the meantime and will have changed particle performance.

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -117,45 +120,11 @@ namespace aspect
...
-        if (this->introspection().compositional_name_exists("plastic_strain") && plastic_yielding == true)
-          data[data_position] = new_strain;
-
-        if (this->introspection().compositional_name_exists("viscous_strain") && plastic_yielding == false)
-          data[data_position+(n_components-1)] = new_strain;
-
-        if (this->introspection().compositional_name_exists("total_strain"))
-          data[data_position] = new_strain;
+        for (unsigned int i = 0; i < SymmetricTensor<2,dim>::n_independent_components ; ++i)
+          particle->get_properties()[data_position + i] += material_outputs.reaction_terms[0][i];
```
gassmoeller, line 127: I dont understand this part. This particle property has between 0 and 3 components (take a look at lines 58 and 65), but not `SymmetricTensor<2,dim>::n_independent_components` (=3 for dim==2 and =6 for dim==3). You probably want to assign reaction terms of certain indices (the indices of the strain components) to the particle property components `0` to `n_components-1`. To find out the indices of the reaction terms, you can use the same approach as in lines 82-89. 

After that works please check our existing test cases and make sure the results with particles are similar to the results with compositional fields (they will likely not be completely identical, but should be close enough).

gassmoeller (commented):
I like that the code is shorter now, but I think you need to fix how to assign the reaction terms to the particle properties. The current approach can only work under very special circumstances. See my comment there. Let me or @naliboff know if you need help figuring out which reaction terms to assign to which particle properties.

source/particle/property/viscoplastic_strain_invariants.cc
erinheilman: Whoops, yes it should be n_components-1 to match, I had copied that line over from elastic stress and forgot to change it.

naliboff: @erinheilman - to follow-up on the comment from  @gassmoeller, you can get the compositional index for a certain strain field using the field names. 

For example, to get the index for plastic strain you could use:
`const int plastic_strain_index = this->introspection().compositional_index_for_name("plastic_strain")`

This could then be used as follows (I think):
```
if (this->introspection().compositional_name_exists("plastic_strain")
    particle->get_properties()[data_position] += material_outputs.reaction_terms[0][plastic_strain_index];
```

The logic in the old version of this file (lines 154-158) would be needed for the viscous and total strain.

erinheilman: So essentially do this,
        const int plastic_strain_index = this->introspection().compositional_index_for_name("plastic_strain") 
        const int viscous_strain_index = this->introspection().compositional_index_for_name("viscous_strain")
        const int total_strain_index = this->introspection().compositional_index_for_name("total_strain")
        
        if (this->introspection().compositional_name_exists("plastic_strain")
          particle->get_properties()[data_position] += material_outputs.reaction_terms[0][plastic_strain_index];
        
        if (this->introspection().compositional_name_exists("viscous_strain")
          particle->get_properties()[data_position] += material_outputs.reaction_terms[0][viscous_strain_index];

        if (this->introspection().compositional_name_exists("total_strain")
          particle->get_properties()[data_position] += material_outputs.reaction_terms[0][total_strain_index]; 
instead of the for loop to update the reaction terms?

include/aspect/particle/property/viscoplastic_strain_invariants.h
```diff
@@ -101,6 +101,7 @@ namespace aspect
            * but it is currently not used in a threaded context.
            */
           mutable MaterialModel::MaterialModelInputs<dim> material_inputs;
+          mutable MaterialModel::MaterialModelOutputs<dim> material_outputs;
```
gassmoeller, line 104: Add the following here for the speedup we discussed:

```suggestion
          mutable MaterialModel::MaterialModelOutputs<dim> material_outputs;
          
          /**
           * Store the indices of relevant compositional fields to avoid repeated lookups.
           */
          unsigned int plastic_strain_index;
          unsigned int viscous_strain_index;
          unsigned int total_strain_index;
```

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -34,7 +34,8 @@ namespace aspect
       ViscoPlasticStrainInvariant<dim>::ViscoPlasticStrainInvariant ()
         :
         n_components(0),
-        material_inputs(1,0)
+        material_inputs(1,0),
+        material_outputs(1,0)
```
gassmoeller, line 38:
```suggestion
        material_outputs(1,0),
        plastic_strain_index(numbers::invalid_unsigned_int),
        viscous_strain_index(numbers::invalid_unsigned_int),
        total_strain_index(numbers::invalid_unsigned_int)
```

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -50,6 +51,8 @@ namespace aspect
 
         n_components = 0;
         material_inputs = MaterialModel::MaterialModelInputs<dim>(1,this->n_compositional_fields());
+        material_outputs = MaterialModel::MaterialModelOutputs<dim>(1, this->n_compositional_fields());
+
 
         // Find out which fields are used.
         if (this->introspection().compositional_name_exists("plastic_strain"))
```
gassmoeller, line 58: Replace the part below with this to set the compositional field indices:
```suggestion
        if (this->introspection().compositional_name_exists("plastic_strain"))
          {
              n_components += 1;
            plastic_strain_index = this->introspection().compositional_index_for_name("plastic_strain");
          }
          
                  if (this->introspection().compositional_name_exists("viscous_strain"))
          {
              n_components += 1;
            plastic_strain_index = this->introspection().compositional_index_for_name("viscous_strain");
          }
          
                  if (this->introspection().compositional_name_exists("total_strain"))
          {
              n_components += 1;
            total_strain_index = this->introspection().compositional_index_for_name("plastic_strain");
          }
```

source/particle/property/viscoplastic_strain_invariants.cc
```diff
@@ -117,45 +117,24 @@ namespace aspect
...
-          data[data_position+(n_components-1)] = new_strain;
+        const int plastic_strain_index = this->introspection().compositional_index_for_name("plastic_strain"); 
+        const int viscous_strain_index = this->introspection().compositional_index_for_name("viscous_strain");
+        const int total_strain_index = this->introspection().compositional_index_for_name("total_strain");
+        
+        if (this->introspection().compositional_name_exists("plastic_strain"))
+          particle->get_properties()[data_position] += material_outputs.reaction_terms[0][plastic_strain_index];
+        
+        if (this->introspection().compositional_name_exists("viscous_strain"))
+          particle->get_properties()[data_position] += material_outputs.reaction_terms[0][viscous_strain_index];
```
gassmoeller, line 131: This can now be the following:

```suggestion
        unsigned int particle_property = 0;
        if (plastic_strain_index != numbers::invalid_unsigned_int)
        {
          particle->get_properties()[data_position+i] += material_outputs.reaction_terms[0][plastic_strain_index];
          ++i;
        }
        
        if (viscous_strain_index != numbers::invalid_unsigned_int)
        {
          particle->get_properties()[data_position+i] += material_outputs.reaction_terms[0][viscous_strain_index];
          ++i;
         }
          
        if (total_strain_index != numbers::invalid_unsigned_int)
        {
          particle->get_properties()[data_position+i] += material_outputs.reaction_terms[0][total_strain_index];
          ++i;
          }
```

gassmoeller (commented):
Try the following improvements before measuring the performance of your code.

source/particle/property/viscoplastic_strain_invariants.cc
naliboff: @erinheilman - In case you did not see it, @gassmoeller provided the right structure in his last review.

naliboff:
@gassmoeller - Can you summarize the discussion regarding a benchmark for this PR? To make sure the models run for a large number of time steps, we could simply run the strain weakening/healing tests for a large number (100s) of time steps. Or do you prefer a new benchmark is added to the repository?

gassmoeller:
I dont quite recall what we discussed about a benchmark. Do we not have a benchmark that makes sure that the current implementation of strain healing (with fields) works? Then we could simply compare that one to the particles.

If we do not have a benchmark at the moment, here are two ideas:
- Can we prescribe an initial constant strain state everywhere, prescribe the velocity to be zero everywhere (using prescribed stokes solutions), and just watch the strain heal over time (it should be possible to write down analytically how much strain should have healed after a fixed time). In idea this is similar to the test `radiogenic_heating` mixed with a prescribed stokes solution (see e.g. `particle_linear_accelerated_flow`).
-  Alternatively: We have models that prescribe a constant simple shear strain rate. Tracking the finite strain and using strain healing that should lead to a steady state between strain accumulation and strain healing, and it should be possible to write down analytically at which value this equilibrium should occur. This is similar to the test `shear_heating.prm` (just without heating, and with strain accumulation and healing)

naliboff:
@gassmoeller - thanks for the clarification!

There are indeed two tests in the repository that use strain healing:
- visco_plastic_yield_plastic_viscous_strain_healing.prm
- visco_plastic_yield_total_strain_healing.prm

These do not run for more than one time step, but they can easily be modified to do so.

@erinheilman - can you do the following:
1. Open a new PR that changes the two tests above to run for a few (say 5-10) time steps and also includes comments at the top that show analytically how much strain should be recovered over time?
2. Once that PR merged, rebase this PR and add a new versions of these tests using particles instead of fields.

Let me know if you have any questions about either of the above steps!

erinheilman:
Yeah I can get started on that!

naliboff:
thanks @erinheilman!

@gassmoeller - are there any additional changes you would like to this PR aside from the updated tests? I know you are going offline for a bit, but we will wait to merge if you would like to take another look after the test update.

gassmoeller:
Sorry for the late reply. This code looks ready, but I do not want to merge it at the moment, because we see a change in the test result (which you updated), but we have no way to check if this output is still correct. We need the simplest possible test case that proves that the strain healing on particles works as intended.

erinheilman:
Yes, I'm going to add tests to this pull request for the particles. John said he has some comments for this that he will add once he gets back from vacation next week as well.

PR #4247, opened 2021-07-16
Allow to use adiabatic P for creep viscosity calculation in visco-plastic rheology

sibiaoliu:
Determine whether to use adiabatic pressure instead of full pressure (default) when calculating creep viscosity. Because in some models (for example, the dike injection model), the full pressure may have an unusually large negative value due to a large negative dynamic pressure, resulting in a zero viscosity and an error being reported.

### For all pull requests:


source/material_model/rheology/visco_plastic.cc
```diff
@@ -148,12 +148,22 @@ namespace aspect
                             + Utilities::to_string(adiabatic_temperature_gradient_for_viscosity) + ") and pressure ("
                             + Utilities::to_string(in.pressure[i]) + ")."));
 
-              // Step 1a: compute viscosity from diffusion creep law, at
-              // least if it is going to be used
+              // Step 1a: compute viscosity from diffusion creep law, at least if it is going to be used
+
+              // Determine whether to use adiabatic pressure instead of full pressure (default) when
+              // calculating creep viscosity. Because in some models, the full pressure may have an
```
naliboff, line 154:
```suggestion
              // calculating creep viscosity. 
```

naliboff: Also remove the next two lines. This documentation is certainly helpful, but I think it should go into the parameter description rather than here in the code.

source/material_model/rheology/visco_plastic.cc
```diff
@@ -148,12 +148,22 @@ namespace aspect
...
+              // unusually large negative value due to a large negative dynamic pressure, resulting
+              // in a zero viscosity and an error being reported.
+
```
naliboff, line 157: Remove empty line

source/material_model/rheology/visco_plastic.cc
```diff
@@ -512,6 +522,10 @@ namespace aspect
                            "dynamic stresses are much higher than the lithostatic pressure. "
                            "If false, the minimum pressure in the plasticity formulation will "
                            "be set to zero.");
+        prm.declare_entry ("Use adiabatic pressure in creep viscosity", "false",
+                           Patterns::Bool (),
+                           "Whether to use adiabatic pressure instead of full pressure (default) "
```
naliboff, line 527: Can you expand the documentation here to explain why this option would be useful?

include/aspect/material_model/rheology/visco_plastic.h
```diff
@@ -249,6 +249,12 @@ namespace aspect
            */
           bool allow_negative_pressures_in_plasticity;
 
+          /**
+           * Whether to use adiabatic pressure instead of full pressure (default)
+           * when calculating creep (diffusion, dislocation, and peierls) viscosity.
```
naliboff, line 254: Expand the documentation here as to why this option may be useful (see related comments below).

naliboff (changes requested):
@sibiaoliu - Thank you for adding this option! Overall the code is in great shape and I only have a few comments about documentation.

As a next step, can you add a test that uses the option "use_adiabatic_pressure_in_creep == true"? This should be part of the "visco_plastic" test suite series.

Thanks!

sibiaoliu:
> @sibiaoliu - Thank you for adding this option! Overall the code is in great shape and I only have a few comments about documentation.
> 
> As a next step, can you add a test that uses the option "use_adiabatic_pressure_in_creep == true"? This should be part of the "visco_plastic" test suite series.
> 
> Thanks!

Added the test; addressed comments :)

include/aspect/material_model/rheology/visco_plastic.h
```diff
@@ -249,6 +249,15 @@ namespace aspect
...
+           * Because in some models, the full pressure may have an unusually large
```
naliboff, line 255:
```suggestion
           * This may be helpful in models where the full pressure has an unusually large
```

include/aspect/material_model/rheology/visco_plastic.h
```diff
@@ -249,6 +249,15 @@ namespace aspect
...
+           * negative value due to a large negative dynamic pressure, resulting in
```
naliboff, line 256:
```suggestion
           * negative value due arising from large negative dynamic pressure, resulting in
```

include/aspect/material_model/rheology/visco_plastic.h
```diff
@@ -249,6 +249,15 @@ namespace aspect
...
+           * a zero viscosity and an error being reported.
```
naliboff, line 257:
```suggestion
           * solver convergence issue and in some cases a viscosity of zero.
```

source/material_model/rheology/visco_plastic.cc
```diff
@@ -148,12 +148,19 @@ namespace aspect
                             + Utilities::to_string(adiabatic_temperature_gradient_for_viscosity) + ") and pressure ("
                             + Utilities::to_string(in.pressure[i]) + ")."));
 
-              // Step 1a: compute viscosity from diffusion creep law, at
-              // least if it is going to be used
+              // Step 1a: compute viscosity from diffusion creep law, at least if it is going to be used
+
+              // Determine whether to use adiabatic pressure instead of full pressure (default) when
```
naliboff, line 153:
```suggestion
              // Determine whether to use the adiabatic pressure instead of the full pressure (default) when
```

source/material_model/rheology/visco_plastic.cc
```diff
@@ -512,6 +519,13 @@ namespace aspect
                            "dynamic stresses are much higher than the lithostatic pressure. "
                            "If false, the minimum pressure in the plasticity formulation will "
                            "be set to zero.");
+        prm.declare_entry ("Use adiabatic pressure in creep viscosity", "false",
```
naliboff, line 522: Please update the parameter description based on my comments in the header file.

tests/visco_plastic_adiabatic_pressure.prm
```diff
@@ -0,0 +1,107 @@
+# A test for using adiabatic pressure in creep viscosity calculation for
+# the visco_plastic material model.
```
naliboff, line 2: I find it really helpful to have some comments at the top of the test that provide a prediction of the expected values. For example, here you could take the adiabatic pressure at the top/bottom of the model and calculate (include equation in comment) what the expected viscosity values should be in the test.

tests/visco_plastic_adiabatic_pressure.prm
```diff
@@ -0,0 +1,107 @@
...
+
+# Global parameters
+set Dimension                              = 2
+set Start time                             = 0
+set End time                               = 0
+set Use years in output instead of seconds = true
+set Nonlinear solver scheme                = single Advection, iterated Stokes
+set Max nonlinear iterations               = 1
+set Output directory                       = visco_plastic_adiabatic_pressure
```
naliboff, line 11: The name of this test is a bit ambiguous. I would suggest renaming to something like `visco_plastic_adiabatic_pressure_in_viscous_creep`?

You can change the name of files with `git mv original_file_name new_file_name`

tests/visco_plastic_adiabatic_pressure.prm
```diff
@@ -0,0 +1,107 @@
...
+  subsection Function
+    set Function expression = 1573
+  end
+end
+
+subsection Material model
+  set Model name = visco plastic
+  subsection Visco Plastic
+    set Reference strain rate = 1.e-16
+    set Viscous flow law = composite
```
naliboff, line 72: Can you also add Peierls creep to make sure it works there as well?

naliboff (changes requested):
@sibiaoliu - Thanks for the quick response! Another round of comments below.

tests/visco_plastic_adiabatic_pressure.prm
sibiaoliu: Addressed comments. Can you review it again? Thanks!

tests/visco_plastic_adiabatic_pressure_in_viscous_creep.prm
```diff
@@ -0,0 +1,149 @@
...
+#  T    = 1573       # temperature [K]
+#  E    = 375.e3     # activation energy for diffusion [J/mol]
+#  V    = 10.e-6     # activation volume for diffusion [m^3/mol]
+#  P_ad = 3.3e9      # adiabatic pressure at the bottom (Pa)
+
+# Similarly, the exact dislocation viscosity 'eta_disl' and peierls
+# viscosity 'eta_peir' can be obtained by the corresponding equations.
+# The following 'subsection Visco Plastic' provides the relevant parameter
+# values. The material model should return the same viscosity value as
+# the minimum value between these three viscosity values at the bottom.
```
naliboff, line 25: This is not exactly true as the output viscosity is the harmonic average of the three values. However, if the values differ significantly (orders of magnitude), then the harmonic average will effectively be equal to the lowest value. I would still reword as suggested below.

Also, in the comments above can you provide the predicted dislocation, Peierls, and composite (average) viscosity? If the results change for some reason, it will be helpful to have these values available for debugging!

tests/visco_plastic_adiabatic_pressure_in_viscous_creep.prm
```diff
@@ -0,0 +1,149 @@
...
+#  R    = 8.314      # gas constant [J/mol*K]
+#  T    = 1573       # temperature [K]
+#  E    = 375.e3     # activation energy for diffusion [J/mol]
+#  V    = 10.e-6     # activation volume for diffusion [m^3/mol]
+#  P_ad = 3.3e9      # adiabatic pressure at the bottom (Pa)
+
+# Similarly, the exact dislocation viscosity 'eta_disl' and peierls
+# viscosity 'eta_peir' can be obtained by the corresponding equations.
+# The following 'subsection Visco Plastic' provides the relevant parameter
+# values. The material model should return the same viscosity value as
```
naliboff, line 24:
```suggestion
# values. The material model returns the harmonic average of the three creep viscosities.
```

naliboff (commented):
@sibiaoliu - Thank you for addressing all the comments! I have one remaining minor item to address, but after that this will be ready to merge.

Also, you will need to update the test output for the new test.

tests/visco_plastic_adiabatic_pressure_in_viscous_creep.prm
sibiaoliu: Addressed these comments

naliboff (approved):
👍 , thanks!

bangerth:
Can I ask a question out of curiosity? In reality, of course, material properties depend on the real pressure, not the adiabatic pressure that has no physical reality. Why would you want to make material respond to the adiabatic pressure? This seems unphysical. Are you doing it only because there are places where the true pressure is negative? Wouldn't it be more correct in that case to change the material model so that it gives reasonable material properties even for negative pressures, rather than using the wrong pressure to begin with?

naliboff:
@bangerth - you are absolutely correct that in nearly all cases one should use the full pressure for the creep laws. Where this arose from is in models with two-phase flow (reports from @jdannberg) or prescribed dilation (see just opened #4263) that produce strongly negative pressures or large pressure gradients, which cause havoc with the flow laws and convergence. 

Long-term, we will need to come up with ways of stabilizing these terms as the tectonic pressures are actually quite important for many processes! For example, see the following recent paper on seismicity at the brittle-ductile transition zone, tectonic pressure, and fluid migration: [Tectonic pressure gradients during viscous creep drive fluid flow and brittle failure at the base of the seismogenic zone](https://pubs.geoscienceworld.org/gsa/geology/article/doi/10.1130/G49012.1/606040/Tectonic-pressure-gradients-during-viscous-creep)

bangerth:
I understand where you are coming from, but it looks like the wrong place to fix this. The pressures really are smaller than everywhere else (and negative if one chooses one particular normalization of the pressure), and it seems wrong to pretend that they are not. If that means that material models return nonsensical material values, then I think the better places to fix this are (i) to choose a better pressure normalization -- say, that the *minimal* surface pressure is zero, not the average surface pressure is zero, or (ii) fix the material models to deal with negative pressures. 

I think that if you just choose the adiabatic pressure, you will not see many interesting effects in tectonic models. For example, in that case the pressure is simply depth-dependent (but laterally constant) and I would suspect that that means that many plastic effects will simply not be visible because the pressure does not see, for example, an extensional fault.

naliboff:
I agree that this feature will in most cases produce undesirable effects in tectonic models (e.g., models are far oversimplified). I certainly have no plans to use it, but I think it is useful for testing purposes and in some production models. 

My recollection is that @jdannberg encountered the dynamic pressure issue (unrealistic gradients) deep in the mantle, so the pressure values were not negative. We also already cap the pressure supplied to the Drucker Prager plasticity at 0, which is a standard practice in the community (absent a different formulation). 

But again, your points are well taken.

I see a few options going forward:
1. Revert these changes after @sibiaoliu and @jdannberg have a chance to weigh in
2. Add a note to the parameter description that makes it clear this option should really only be used in exceptional cases, if it all, and list alternative methods for handling the issue.
3. Only use the adiabatic pressure where the pressure is negative (least desirable fix in my view and may make things worse)

naliboff:
@bangerth  while I am thinking about it another option to consider. We could disable use of this feature unless melt transport or prescribed dilation is included (i.e., cases that give rise to this issue). That would prevent the feature from being used unless it is really needed. Let me know if you would like a PR opened that implements this.

bangerth:
@naliboff  You know the material models better than me, so I'll leave the resolution to you. Both options 1 and 2 seem reasonable to me; I'm not a big fan of option 3.

Should we open a github issue that tracks this question, rather than continuing in a merged PR?

PR #4248, opened 2021-07-16
Specify requested properties for Material Model

anne-glerum:
There are many places where we evaluate the material model to get at one material property. However, many other properties are also computed, like the reaction_terms. This PR specifies in several places that only the viscosity is requested. This should avoid computing the reaction terms and force terms for elasticity and strain dependent rheologies, as both Elasticity and StrainDependent check for `in.requests_property(MaterialProperties::reaction_terms)`.

However, currently request_property() still does
```
      //TODO: Remove this once all callers set requested_properties correctly
      if ((property & MaterialProperties::Property::reaction_terms) != 0)
        return (strain_rate.size() != 0);
```
so if in.strain_rate is filled (like for viscosity) then reaction_terms are still computed. 

### For all pull requests:

### DIFF
--- a/include/aspect/material_model/interface.h
+++ b/include/aspect/material_model/interface.h
@@ -203,6 +203,9 @@ namespace aspect
         entropy_derivative_pressure    = 128,
         entropy_derivative_temperature = 256,
         reaction_terms                 = 512,
+        reaction_rates                 = 1024,
+        force_terms                    = 2048,
+        additional_outputs             = 4096,
 
         equation_of_state_properties   = density |
                                          thermal_expansion_coefficient |
@@ -213,7 +216,10 @@ namespace aspect
         all_properties                 = equation_of_state_properties |
                                          viscosity |
                                          thermal_conductivity |
-                                         reaction_terms
+                                         reaction_terms |
+                                         reaction_rates |
+                                         force_terms |
+                                         additional_outputs
       };
 
       /**
--- a/source/material_model/interface.cc
+++ b/source/material_model/interface.cc
@@ -283,7 +283,12 @@ namespace aspect
       requested_properties(MaterialProperties::all_properties)
     {
       if (compute_strain_rate == false)
-        this->strain_rate.resize(0);
+        {
+          this->strain_rate.resize(0);
+          requested_properties = MaterialProperties::Property(requested_properties & ~MaterialProperties::viscosity);
+        }
+      else
+        requested_properties = requested_properties | MaterialProperties::viscosity;
 
       for (unsigned int q=0; q<input_data.solution_values.size(); ++q)
         {
@@ -370,9 +375,15 @@ namespace aspect
       // length of the strain_rate vector to 0, we signal to evaluate()
       // that we do not need to access the viscosity.
       if (compute_strain_rate)
-        fe_values[introspection.extractors.velocities].get_function_symmetric_gradients (solution_vector,this->strain_rate);
+        {
+          fe_values[introspection.extractors.velocities].get_function_symmetric_gradients (solution_vector,this->strain_rate);
+          requested_properties = requested_properties | MaterialProperties::viscosity;
+        }
       else
-        this->strain_rate.resize(0);
+        {
+          this->strain_rate.resize(0);
+          requested_properties = MaterialProperties::Property(requested_properties & ~MaterialProperties::viscosity);
+        }
 
       // Vectors for evaluating the compositional field parts of the finite element solution
       std::vector<std::vector<double>> composition_values (introspection.n_compositional_fields,
@@ -409,14 +420,6 @@ namespace aspect
     bool
     MaterialModelInputs<dim>::requests_property(const MaterialProperties::Property &property) const
     {
-      //TODO: Remove this once all callers set requested_properties correctly
-      if ((property & MaterialProperties::Property::viscosity) != 0)
-        return (strain_rate.size() != 0);
-
-      //TODO: Remove this once all callers set requested_properties correctly
-      if ((property & MaterialProperties::Property::reaction_terms) != 0)
-        return (strain_rate.size() != 0);
-
       // Note that this means 'requested_properties' can include other properties than
       // just 'property', but in any case it at least requests 'property'.
       return (requested_properties & property) != 0;
--- a/source/postprocess/visualization/seismic_anomalies.cc
+++ b/source/postprocess/visualization/seismic_anomalies.cc
@@ -68,6 +68,7 @@ namespace aspect
 
         MaterialModel::MaterialModelInputs<dim> in(n_q_points, this->n_compositional_fields());
         MaterialModel::MaterialModelOutputs<dim> out(n_q_points, this->n_compositional_fields());
+        in.requested_properties = MaterialModel::MaterialProperties::additional_outputs;
 
         std::vector<std::vector<double>> composition_values (this->n_compositional_fields(),std::vector<double> (quadrature_formula.size()));
 
@@ -211,6 +212,7 @@ namespace aspect
 
         MaterialModel::MaterialModelInputs<dim> in(n_q_points, this->n_compositional_fields());
         MaterialModel::MaterialModelOutputs<dim> out(n_q_points, this->n_compositional_fields());
+        in.requested_properties = MaterialModel::MaterialProperties::additional_outputs;
 
         std::vector<std::vector<double>> composition_values (this->n_compositional_fields(),std::vector<double> (quadrature_formula.size()));
 
@@ -296,9 +298,6 @@ namespace aspect
                     fe_values[this->introspection().extractors.pressure].get_function_gradients (this->get_solution(),
                                                                                                  in.pressure_gradient);
                     in.position = fe_values.get_quadrature_points();
-
-                    // we do not need the strain rate
-                    in.strain_rate.resize(0);
 
                     // Loop over compositional fields to get composition values
                     for (unsigned int c=0; c<this->n_compositional_fields(); ++c)
--- a/source/postprocess/visualization/shear_stress.cc
+++ b/source/postprocess/visualization/shear_stress.cc
@@ -58,6 +58,9 @@ namespace aspect
         MaterialModel::MaterialModelOutputs<dim> out(n_quadrature_points,
                                                      this->n_compositional_fields());
 
+        // We do not need to compute anything but the viscosity
+        in.requested_properties = MaterialModel::MaterialProperties::viscosity;
+
         // Compute the viscosity...
         this->get_material_model().evaluate(in, out);
 
--- a/benchmarks/annulus/annulus.cc
+++ b/benchmarks/annulus/annulus.cc
@@ -527,6 +527,7 @@ namespace aspect
                   MaterialModel::MaterialModelInputs<dim> in_face(fe_face_values, cell, this->introspection(), this->get_solution());
                   MaterialModel::MaterialModelOutputs<dim> out_face(fe_face_values.n_quadrature_points, this->n_compositional_fields());
                   fe_face_values[this->introspection().extractors.temperature].get_function_values(topo_vector, topo_values);
+                  in_face.requested_properties = MaterialModel::MaterialProperties::density;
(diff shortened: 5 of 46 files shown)


source/postprocess/visualization/shear_stress.cc
```diff
@@ -57,6 +57,9 @@ namespace aspect
         MaterialModel::MaterialModelOutputs<dim> out(n_quadrature_points,
                                                      this->n_compositional_fields());
 
+        // We do not need to compute anything but the viscosity
+        in.requested_properties = MaterialModel::MaterialProperties::viscosity;
```
gassmoeller, line 61: It would be nice if MaterialModelInputs had a constructor that would take this as input, then you could write above:
```
        MaterialModel::MaterialModelInputs<dim> in(input_data,
                                                   this->introspection(),
                                                   MaterialModel::MaterialProperties::viscosity);
```

Want to give this a try? If not the current form works as well.

gassmoeller (commented):
As discussed this works and is a step in the right direction, but does not do anything at the moment because of the compatibility fixes in `MaterialModelInputs<dim>::requests_property`. Still a step forward to setting this correctly.

For the future: It would be nice if MaterialModelInputs had a constructor that would take this as input, then you could write:
```
        MaterialModel::MaterialModelInputs<dim> in(input_data,
                                                   this->introspection(),
                                                   MaterialModel::MaterialProperties::viscosity);
```

Do you want to give this a try? If not the current form works as well.

bangerth:
On 7/16/21 10:50 AM, Rene Gassmoeller wrote:
> For the future: It would be nice if MaterialModelInputs had a constructor that 
> would take this as input, then you could write:
> 
> |MaterialModel::MaterialModelInputs<dim> in(input_data, this->introspection(), 
> MaterialModel::MaterialProperties::viscosity); |
> 
> Do you want to give this a try? If not the current form works as well.

Actually, I would prefer if this flag was set on the *output* object 
(MaterialModelOutputs). The *input* object does not need to know what it is 
being used for, but the output object should specify which fields need to be 
filled.

anne-glerum:
@bangerth The requested_properties are already a part of MaterialModelInputs, are you saying that should be changed to MaterialModelOutputs? MaterialModels already check the input for the requested properties (although the check doesn't return the right answer for until it's implemented everywhere).

gassmoeller (approved):
Nice, thanks for the cleanup! I left a suggestion for how to resolve your question.

@bangerth: I see where you are coming from, but as @anne-glerum mentioned this is already established (#3457), this PR only uses the functionality in more places. I see why you say it could go into the outputs object, but I think when I implemented this I had the following two thoughts:
1. Telling the material model what to do is an input operation.
2. Having this in the output object would mean it is handed over as non-const, meaning the material model could modify what properties to compute internally (not likely, but it seems conceptually not ok).

If you want it to be moved I don't feel very strongly about it, but my preference would be an input. A fix would only need a bit of search and replace.

In any case I think we should not delay this PR for it.

bangerth:
On 7/20/21 10:02 AM, Rene Gassmoeller wrote:
> @bangerth <https://github.com/bangerth>: I see where you are coming from, but 
> as @anne-glerum <https://github.com/anne-glerum> mentioned this is already 
> established (#3457 <https://github.com/geodynamics/aspect/pull/3457>), this PR 
> only uses the functionality in more places. I see why you say it could go into 
> the outputs object, but I think when I implemented this I had the following 
> two thoughts:
> 
>  1. Telling the material model what to do is an input operation.
>  2. Having this in the output object would mean it is handed over as
>     non-const, meaning the material model could modify what properties to
>     compute internally (not likely, but it seems conceptually not ok).
> 
> If you want it to be moved I don't feel very strongly about it, but my 
> preference would be an input. A fix would only need a bit of search and replace.

No need. I had forgotten that we already had that and not read the patch to 
remind myself of it.

> In any case I think we should not delay this PR for it.

Correct!

source/postprocess/visualization/seismic_anomalies.cc
```diff
@@ -147,6 +147,11 @@ namespace aspect
                     in.position = fe_values.get_quadrature_points();
 
                     // we do not need the strain rate
+                    // TODO this statement should be changed to
+                    // in.requested_properties = ...
+                    // but I don't know what is actually needed here,
+                    // as it seems only additional output is requested.
+                    // There is no property for that.
                     in.strain_rate.resize(0);
```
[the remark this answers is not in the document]
anne-glerum: I've added the property `additional_outputs` and included it in `all_properties` as well.

source/material_model/interface.cc
```diff
@@ -370,9 +370,17 @@ namespace aspect
...
+      {
         fe_values[introspection.extractors.velocities].get_function_symmetric_gradients (solution_vector,this->strain_rate);
+        requested_properties = requested_properties | MaterialProperties::viscosity;
+      }
       else
+      {
         this->strain_rate.resize(0);
+        requested_properties = MaterialProperties::equation_of_state_properties |
+                                         MaterialProperties::reaction_terms |
+                                         MaterialProperties::additional_outputs;
```
gassmoeller, line 382: This way you overwrite the existing value of requested_properties, which is different from the behavior in the if-branch. I think you were looking for only disabling viscosity? If so that would probably work like this:
```suggestion
        requested_properties = requested_properties & ~MaterialProperties::viscosity;
```

gassmoeller (commented):
This looks nearly ready, but it acquired a few conflicts. If you make the one change I mentioned and rebase this should be good to go.

source/material_model/interface.cc
anne-glerum: Ah didn't know about `~`, cool!

anne-glerum:
Fixed the merge conflicts and addressed your comment. 
This PR has become more than I originally described. The next step would be to remove the `compute_strainrate` flag altogether, in favor of the requested_properties, either using the constructor or the current form.

anne-glerum:
Hmm, one test still fails. I'll look into it some more.

anne-glerum:
I cannot seem to figure out why this last test shear_heating_with_melt is failing. @jdannberg would you maybe have an idea?

bangerth:
@anne-glerum Would you mind rebasing this one more time to resolve the merge conflict?

jdannberg:
I had a look at the shear_heating_with_melt test. It looks like the changes in the statistics file are very small. Basically, only the last (8th) digit changes, which really could be due to anything. It looks like the biggest changes are in the screen output in the residual of the compositional fields in the last nonlinear iteration of a given time step (so when the residuals are 1e-16, they might be twice than that or half of that).

The test uses the melt simple material model, which is one of the ones you changed, but I honestly do not see how your change could have affected the results (at least the change makes sense to me). In addition, there are other tests that use the same material model and also use operator splitting that have not changed at all. So I would say it's probably fine; at least I do not have any other good ideas.

tjhei:
@zjiaqi2018 Do you have an idea what is going on, why the test output changed? Can you test if on current master tests/gmg_mesh_deform_ghost_entries.prm returns for ``RMS, max velocity`` with and without GMG?

tjhei:
I talked with @zjiaqi2018 and it seems this prm might be problematic because we have seen free surface + GMG + elasticity to give incorrect solutions (it is in fact currently disabled, but not with a prescribed surface like this test).
That doesn't explain why the output changes here...

tjhei:
@anne-glerum I asked @zjiaqi2018 to either find the problem or work around it by disabling/deleting the test temporarily so we can merge this PR.

@zjiaqi2018 It would still be good to know why the changes in this PR change the output. That doesn't make sense to me. Can you maybe confirm that RMS changes with this PR (try rebasing to current master to exclude the possibility that some unrelated change causes the issue).

zjiaqi2018:
I ran gmg_mesh_deform_ghost_entries.prm with the deal.ii master, but got nan in `in.strain_rate`, I don't know what caused the problem.

anne-glerum:
I've made some changes in stokes_matrix_free.cc, but they actually shouldn't change anything, because the defaults should be correct. But let's see what the tester says. 

@zjiaqi2018 Where/when did you get in.strain_rate = nan? Do you have a line nr or function name or something?

zjiaqi2018:
> I've made some changes in stokes_matrix_free.cc, but they actually shouldn't change anything, because the defaults should be correct. But let's see what the tester says.
> 
> @zjiaqi2018 Where/when did you get in.strain_rate = nan? Do you have a line nr or function name or something?

image

image

anne-glerum:
Thank you @zjiaqi2018 ! I suspect that in assembly.cc line 633 `need_viscosity` and line 578-579 `update_strain_rate` are not set correctly for the GMG case. Could you have a look and see whether they should be true in the case of failing test? I will set them to true in the code just to see if the test then succeeds.

anne-glerum:
So was this test wrong before this PR? The test fails, but this branch gives the same velocities as the AMG:
```
-   Solving Stokes system... 34+0 iterations.
+   Solving Stokes system... 43+0 iterations.
 
-     RMS, max velocity:         0.142 m/year, 0.448 m/year
+     RMS, max velocity:         0.129 m/year, 0.408 m/year
```

anne-glerum:
@tjhei Is the gmg_mesh_deform_ghost_entries.prm test broken? The GMG result isn't the same as for AMG on main (see my message from 20 days ago), but with this PR it is the same (and therefore different from the test results).

tjhei:
Could be. @zjiaqi2018 ?

tjhei:
Thank you, @anne-glerum this helped! Let me see if I can fix this correctly.

tjhei:
@anne-glerum I think you can drop the last commit after you rebase when #4713 is merged.

bangerth:
@tjhei suggested merging this before #4725. What is the status of this patch?

tjhei:
Needs a rebase and the last commit to be dropped. Then some code reviews as well. At least I focused on the GMG bugs so far.

anne-glerum:
I'm done presenting, so I can get back to my PR's. I'll rebase and drop the test commits today or tomorrow. And yes, then it needs a thorough review :)

tjhei (approved):
Great! Can you please squash your commits?

anne-glerum:
@tjhei Anything else I should do?

source/postprocess/visualization/shear_stress.cc
tjhei: Yes, that would be a great new variant of reinit() after deprecating the old.

tjhei (commented):
This all looks reasonable to me, but I think another pair of eyes would be good.

include/aspect/material_model/interface.h
```diff
@@ -203,6 +203,9 @@ namespace aspect
         entropy_derivative_pressure    = 128,
         entropy_derivative_temperature = 256,
         reaction_terms                 = 512,
+        reaction_rates                 = 1024,
+        force_terms                    = 2048,
```
gassmoeller, line 207: I am not so sure about making this one separate from `additional_outputs`. Structurally it is a subset of `additional_outputs`. I understand we may want to ask for it separately, but in the current form there is a problem: Someone could ask for all `additional_outputs` and then assume all additional outputs will be computed, while `force_terms` will not be computed if not requested. Is there a reason we need this as a separate entry?

source/material_model/interface.cc
```diff
@@ -283,7 +283,12 @@ namespace aspect
       requested_properties(MaterialProperties::all_properties)
     {
       if (compute_strain_rate == false)
-        this->strain_rate.resize(0);
+        {
+          this->strain_rate.resize(0);
+          requested_properties = MaterialProperties::Property(requested_properties & ~MaterialProperties::viscosity);
+        }
+      else
+        requested_properties = requested_properties | MaterialProperties::viscosity;
```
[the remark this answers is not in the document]
gassmoeller: Yes, I think this line is not needed.

gassmoeller (approved):
I would say let's get this merged and handle the few small comments in a follow-up. Because this PR touches so many files the risk of adding new conflicts soon is pretty large.

gassmoeller:
And thanks for all the work @anne-glerum!

PR #4255, opened 2021-07-16
Surface strain residual

alarshi:
This PR aims to add a visualization postprocessor, surface_strain_rate_residual, that computes the difference between the modeled strain rate components and an input ASCII data file. The residuals are only calculated at the top surface and statistics are currently not implemented. I added a test case that gives reasonable values at the top surface, but not on other surfaces. 
I need help to review the current implementation and also how to proceed to add the statistics of the strain rate.

### For all pull requests:


alarshi:
@jdannberg @gassmoeller Could you please take a look at it? : )

include/aspect/postprocess/visualization/boundary_strain_rate_residual.h
```diff
@@ -0,0 +1,81 @@
...
+namespace aspect
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      using namespace dealii;
+      /**
+       * A class derived from DataPostprocessor that takes an output vector
+       * and computes a variable that represents the velocity residual between
```
gassmoeller, line 41:
```suggestion
       * and computes a variable that represents the strain rate residual between
```

include/aspect/postprocess/visualization/boundary_strain_rate_residual.h
```diff
@@ -0,0 +1,81 @@
...
+       * the input velocity data and the velocity computed for the model domain.
```
gassmoeller, line 42: strain rate here and update description below

include/aspect/postprocess/visualization/boundary_strain_rate_residual.h
```diff
@@ -0,0 +1,81 @@
...
+        : public DataPostprocessorTensor<dim>,
+          public SurfaceOnlyVisualization<dim>,
+          public SimulatorAccess<dim>,
+          public Interface<dim>
+      {
+        public:
+          BoundaryStrainRateResidual ();
+
+          /**
+           * Evaluate the velocity residual for the current cell.
```
gassmoeller, line 62: strain ratge

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -42,6 +42,13 @@ namespace aspect
           data_lookup = std_cxx14::make_unique<Utilities::StructuredDataLookup<dim>>(dim, scale_factor);
           data_lookup->load_file(data_directory + data_file_name, this->get_mpi_communicator());
         }
+
+      else if (use_surface_strain_rate_data)
+        {
+          strain_rate_data_lookup = std_cxx14::make_unique<Utilities::StructuredDataLookup<dim> >((dim-1)*(dim-1), scale_factor);
+          strain_rate_data_lookup->load_file(data_directory + data_file_name, this->get_mpi_communicator());
+        }
```
gassmoeller, line 50: So at the moment you can only use this postprocessor for the velocity or the strain rate, but not both? This suggests to me we should instead have a new `boundary_strain_rate_residual_statistics` so that you could output both at the some time. Or is there a reason you would never want to do both at the same time?

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -55,6 +62,23 @@ namespace aspect
...
+    {
+      const std::array<double,dim> spherical_position = this->get_geometry_model().
+                                                        cartesian_to_natural_coordinates(p);
+      Point<dim> internal_position = p;
+
+      for (unsigned int d = 0; d < dim; ++d)
+        internal_position[d] = spherical_position[d];
+
+      return internal_position;
     }
```
gassmoeller, line 82: Unless you need this function in a few other places as well, I think it does not make things simpler here (it saves about 3-5 lines in line 128, but adds this whole function instead). Also at the moment this function is public, but it is not used outside of this class. I would remove it.

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -92,6 +116,26 @@ namespace aspect
 
 
 
+    template <int dim>
+    std::vector<Vector<double> >
```
gassmoeller, line 120: This happens to be the return type in the other function where you want to use the value this function returns, but in reality it is an awkward type for a strain rate that is only necessary because of the constraints of the other postprocessor. But you convert it anyway before returning it (subtracting it from the real strain rate). So I would suggest to return a `Tensor<2,dim-1>` instead. This will also make this function faster, because the size of a Tensor<2,dim-1> is known at compile-time and can be allocated on the stack instead of the heap memory (http://www.programmerinterview.com/data-structures/difference-between-stack-and-heap/).

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -92,6 +116,26 @@ namespace aspect
...
+      Vector <double> temp_vector (dim-1);
+      std::vector<Vector<double> > data_surface_strain_rate (dim-1, temp_vector);
+      Point<dim> position = p;
+
+      if (this->get_geometry_model().natural_coordinate_system() == Utilities::Coordinates::spherical)
+        position = get_spherical_position(p);
+
+      for (unsigned int d=0; d<dim-1; ++d)
+        for (unsigned int e=0; e<dim-1; ++e)
+          data_surface_strain_rate[d][e] = strain_rate_data_lookup->get_data(position, d+e);
```
gassmoeller, line 132: I think this should be:

```suggestion
          data_surface_strain_rate[d][e] = strain_rate_data_lookup->get_data(position, d*(dim-1)+e);
```

Otherwise d=1,e=0 and d=0,e=1 would access the same data column (which in theory would be fine if the tensor is symmetric and the data file only stores 3 columns, but that does not seem to be the case).

data/postprocess/boundary-velocity-residual/box_3d_boundary_strain_rate-3d.txt
```diff
@@ -0,0 +1,22 @@
+# Test data for ascii data boundary velocity.
```
gassmoeller, line 1: Where do you use this file? It looks like your postprocessor expects a strain rate tensor of dimension dim-1 (i.e. strain rate on the surface), not a volume strain rate.

source/postprocess/boundary_velocity_residual_statistics.cc
```diff
@@ -148,7 +192,15 @@ namespace aspect
                 for (unsigned int q=0; q<fe_face_values.n_quadrature_points; ++q)
                   {
                     const Point<dim> point_at_surface = fe_face_values.quadrature_point(q);
-                    // Extract data velocity.
+                    // Extract data velocity or strain rate.
+
+                    if (use_surface_strain_rate_data)
+                      {
+                        std::vector<Vector<double> > data_surface_strain_rate = get_data_surface_strain_rate(point_at_surface);
```
gassmoeller, line 199: this would be now
```suggestion
                        Tensor<2,dim-1> data_surface_strain_rate = get_data_surface_strain_rate(point_at_surface);
```

source/postprocess/visualization/boundary_strain_rate_residual.cc
```diff
@@ -0,0 +1,139 @@
...
+        for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
+          if (cell->at_boundary(f) &&
+              (this->get_geometry_model().translate_id_to_symbol_name (cell->face(f)->boundary_id()) == "top"))
+            cell_at_top_boundary = true;
+
+        if (cell_at_top_boundary)
+          for (unsigned int q=0; q<computed_quantities.size(); ++q)
+            {
+              Tensor<2,dim> grad_u;
+              const std::vector<Vector<double> > data_surface_strain_rate =
```
gassmoeller, line 79:
```suggestion
              const Tensor<2,dim-1> data_surface_strain_rate =
```

source/postprocess/visualization/boundary_strain_rate_residual.cc
```diff
@@ -0,0 +1,139 @@
...
+
+      }
+
+
+
+      template <int dim>
+      std::list<std::string>
+      BoundaryStrainRateResidual<dim>::required_other_postprocessors() const
+      {
+        return std::list<std::string> (1, "boundary velocity residual statistics");
```
gassmoeller, line 113:
```suggestion
        return std::list<std::string> (1, "boundary strain rate residual statistics");
```

source/postprocess/visualization/boundary_strain_rate_residual.cc
```diff
@@ -0,0 +1,139 @@
...
+      ASPECT_REGISTER_VISUALIZATION_POSTPROCESSOR(BoundaryStrainRateResidual,
+                                                  "boundary strain rate residual",
+                                                  "A visualization output object that generates output for the velocity "
+                                                  "residual at the top surface. The residual is computed at each point at the "
+                                                  "surface as the difference between the modeled velocities and the input "
+                                                  "data velocities for each vector component. The user has an option to choose "
+                                                  "the input data as ascii data files (e.g. GPS velocities) with columns "
+                                                  "in the same format as described for the 'ascii data' initial temperature plugin "
+                                                  "or a velocity field computed from the GPlates program as described in the gplates "
+                                                  "boundary velocity plugin. ")
```
gassmoeller, line 136: adapt to strain rate

tests/box_simple_surface_strain_rate_residual.prm
```diff
@@ -0,0 +1,122 @@
+# A testcase with manufactured solution that demonstrates that the mass
+# formulation "hydrostatic pressure" is correct and converges. The example is
+# constructed to have
+#
+# -\nabla \cdot \mathbf{u} = \left( \kappa \rho \textbf{g} - \alpha \nabla T \right) \cdot \textbf{u}
+#
+# using a specific gravity and a force vector is used to enforce the correct u
+# and p (by canceling out the gravity term). This means the test will only
+# work for this mass formulation.
```
gassmoeller, line 9: Update description

gassmoeller (commented):
Nice! As you said this is WIP and I left comments on the general direction. I think this warrants its own statistics postprocessor, because it is essentially disconnected from the functionality of boundary_velocity_residual_statistics (it is similar functionality but for a different property, and you may want to have both at the same time).

alarshi:
@gassmoeller: Thank you for your suggestions!
I will make another statistics file for strain rate, my initial goal was to have a boundary_residual_statistics file with several ```else if``` statements for velocity or strain_rate. But, I don't think it will be possible without making the ```execute``` too long.

source/postprocess/boundary_velocity_residual_statistics.cc
alarshi: Got it, thanks for explaining!

data/postprocess/boundary-velocity-residual/box_3d_boundary_strain_rate-3d.txt
alarshi: Currently, I am not using it. But, the visualization postprocessor outputs boundary_strain_rate_residual of dim dimension. 
In the current implementation, the boundary_strain_rate_residual[3] is actually equivalent to strain_rate[4], if the input strain rate is 0. 
Therefore, I also created the data file with all the components and was not sure if I needed the volume strain rate in the data or not. I will remove the file in the next commit.

tests/box_simple_surface_strain_rate_residual.prm
```diff
@@ -0,0 +1,121 @@
+# A testcase with manufactured solution that demonstrates that the mass
```
gassmoeller, line 1: Currently the test case is missing its output data.

gassmoeller (commented):
Here is the long delayed review. The code looks pretty good already. I left a few comments, and take a look at #4342 for how to improve the output in this PR as well.

Please add a changelog entry and test output.

Let me know when you want me to take another look.

alarshi:
@gassmoeller : Thank you for the detailed review! I am sorry I accidentally resolved some previous suggestions from you last time, but I addressed them in the latest commits.
 The output of the test looks like:
strain_rate_outputs
```boundary strain rate residual``` (left)  and ```strain rate``` (right) . The yellow is the nan values because the ```boundary strain rate residual``` is only computed at the top surface.

I also tried this post-processor for spherical geometry using Global Strain Rate Model for input strain in our global mantle flow models. The outputs looks like: 
strain_rate_outputs-1
The model is centered at the African plate.

Let me know your suggestions, whenever you get a chance to look at it again. :-)

**Comment by alarshi** [comment]
@gassmoeller : Thank you for the review!

I addressed your comments in the latest 3 commits. I also reduced the refinement for the test case from 4 to 3 to make it faster. Below are the results for this test. The residual is only calculated at the top surface and gives nan values everywhere else. Since the input data file has 0 values everywhere, we would expect the computed strain rate and strain rate residuals to be same and opposite in signs (because the residual is `data-modeled` values). 

image

**Comment by alarshi** [comment]
@gassmoeller : The difference in the visualization output was fixed by setting the `set Point-wise stress and strain = true`

image

File: source/postprocess/boundary_strain_rate_residual_statistics.cc [hunk 15]
```diff
@@ -0,0 +1,333 @@
...
+                             "interpreted as the path in which the ASPECT source files were "
+                             "located when ASPECT was compiled. This interpretation allows, "
+                             "for example, to reference files located in the `data/' subdirectory "
+                             "of ASPECT.");
+          prm.declare_entry ("Data file name", "box_3d_boundary_strain_rate.txt",
+                             Patterns::Anything (),
+                             "The file name of the input surface strain rate an ascii data. "
+                             "The file has (dim-1)*(dim-1) columns in addition to the coordinate columns. "
+                             "The strain rate data columns correspond to the different strain rate components "
+                             "at the surface, i.e. 2 components for 2d models and 4 components for 3d models. ");
```
gassmoeller, line 282: Is this true or is it only one column with the strain rate invariant?

source/postprocess/boundary_strain_rate_residual_statistics.cc
```diff
@@ -0,0 +1,333 @@
...
+// explicit instantiations
+namespace aspect
+{
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(BoundaryStrainRateResidualStatistics,
+                                  "boundary strain rate residual statistics",
+                                  "A postprocessor that computes some statistics about "
+                                  "the surface strain rate residual along the top boundary. The residual "
+                                  "is the difference between the second invariant of model solution gradients and "
```
gassmoeller, line 328: How the strain rate is calculated is not as important here, it is more important to stay consistent with the rest of the documentation.
```suggestion
                                  "is the difference between the second invariant of the model strain rate and "
```

source/postprocess/visualization/boundary_strain_rate_residual.cc
```diff
@@ -0,0 +1,127 @@
...
+              Tensor<2,dim> grad_u;
+              for (unsigned int d=0; d<dim; ++d)
+                grad_u[d] = input_data.solution_gradients[q][d];
+              const SymmetricTensor<2,dim> strain_rate = symmetrize(grad_u);
+
+              const double data_surface_strain_rate =
+                boundary_strain_rate_residual_statistics.get_data_surface_strain_rate(input_data.evaluation_points[q]);
+
+              // Only compute residual for doubles. This condition checks for nan values
+              if (data_surface_strain_rate < 1e300)
```
gassmoeller, line 83: this condition is different from the statistics postprocessor, but they should be the same.

gassmoeller (commented):
Looks generally good, thank you. I had some small comments, let me know when I should take another look.

source/postprocess/boundary_strain_rate_residual_statistics.cc
```diff
@@ -0,0 +1,333 @@
...
+                      // Extract data strain rate invartiant.
+                      double e_ii_data = get_data_surface_strain_rate(point_at_surface);
+
+                      if (this->convert_output_to_years() == true)
+                        e_ii_data = e_ii_data/year_in_seconds;
+
+                      // The strain rate invariant residual is calculated here. Ignore the computation if we
+                      // encounter nan or 1e300 in the input data.
+                      double strain_invariant_residual = 0;
+                      if (e_ii_data < 1e300 || e_ii_data == e_ii_data)
```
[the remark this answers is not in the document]
alarshi: Oh, I did not know about this function. Thanks!

source/postprocess/boundary_strain_rate_residual_statistics.cc
alarshi: This is not true anymore. Thanks for catching it!

alarshi:
Thank you for the review! I addressed your suggestions in the recent commit. I also added the parameter `set Point-wise stress and strain  = true` in the test file but now that I think about it maybe it is not relevant since we are not visualizing the output(?).

alarshi:
Oh I did not change the results for the parameter, `set Point-wise stress and strain = true` and that caused the tester to failed. I removed this option in the latest commit.

alarshi:
I modified the test case to include stress and strain at points in the latest commit. Hopefully, this works :)

gassmoeller (approved):
Thank you, ready to go.

PR #4256, opened 2021-07-16
GMG: support mesh deformation

tjhei:
well, this is more complicated than expected:
- [x] Generate a different MappingQ1Eulerian on each GMG level
- [x] interpolate_to_mg for the matrix-free object
- [x] use the mapping in StokesMatrixFreeHandler

reference: https://github.com/dealii/dealii/blob/master/tests/mappings/mapping_q_eulerian_08.cc


tjhei:
Well, the test seems to run now! 🥳

I need a more complicated test problem, @anne-glerum @naliboff any suggestions for a benchmark or something?

tjhei:
@zjiaqi2018 can you please review this?

source/simulator/stokes_matrix_free.cc
```diff
@@ -2449,7 +2462,11 @@ namespace aspect
             additional_data.mg_level = level;
             std::shared_ptr<MatrixFree<dim,GMGNumberType>>
                                                         mg_mf_storage_level(new MatrixFree<dim,GMGNumberType>());
-            mg_mf_storage_level->reinit(*sim.mapping, dof_handler_p, level_constraints,
+            const Mapping<dim> &mapping =
+              (sim.mesh_deformation) ?
+              sim.mesh_deformation->get_mapping_on_level(level)
```
zjiaqi2018, line 2467:
```suggestion
              sim.mesh_deformation->get_level_mapping(level)
```

source/simulator/stokes_matrix_free.cc
```diff
@@ -2418,7 +2425,13 @@ namespace aspect
             additional_data.mg_level = level;
             std::shared_ptr<MatrixFree<dim,GMGNumberType>>
                                                         mg_mf_storage_level(new MatrixFree<dim,GMGNumberType>());
-            mg_mf_storage_level->reinit(*sim.mapping, dof_handler_v, level_constraints,
+
+            const Mapping<dim> &mapping =
+              (sim.mesh_deformation) ?
+              sim.mesh_deformation->get_mapping_on_level(level)
```
zjiaqi2018, line 2431:
```suggestion
              sim.mesh_deformation->get_level_mapping(level)
```

zjiaqi2018 (commented):
It looks good to me. I would prefer `get_level_mapping()` to `get_mapping_on_level()`, since you also use `level_mappings`, but both are OK.

naliboff:
@tjhei - super cool! For a benchmark, are you looking for a problem that develops topography over time or something similar to the current one where there is only initial mesh deformation? 

Lots of benchmark options for the current first case (mesh deforms over time). 

For a problem where only the initial mesh is deformed (and stays static), I could add a more complicated rheology to the current model and extend it 3D. For a more realistic case, we could use the setup from a [paper](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2020GL090483) Tahiry, Sarah and I published recently that looks at instantaneous deformation in East Africa (includes initial topography).

Let me know how I can help with any of these options, really excited to see how the GMG performs with the deformed mesh!

tjhei:
Well, first, I need to know if I am solving the right thing. It could be static or changing in time. Do we have something with an exact solution?
A fancy example would also be good to see if things work when the problem gets large...

naliboff:
@tjhei - I can think of a few analytical solutions with a deformable free surface (ex: topography relaxation or diffusion, boundary traction loading). Would these be helpful at this stage?

I'm having a hard time thinking of analytical solutions for models where there is initial topography (mesh deformation), but the mesh does not deform with time. I suppose an exact solution for pressure would be straightforward, but I'm having trouble thinking of ones for velocity. 

@cedrict - thoughts for analytical benchmarks where there is initial topography, but the mesh does not deform over time?

tjhei:
If it is difficult, how about an example where I can compare velocities matrix-free vs matrix-based?

tjhei:
Free surface is not going to work anytime soon...

gassmoeller:
I think @anne-glerum and @Djneu were using some models with initial topography, maybe they have an example case that could be run with matrix-free vs matrix-based? You would need to disable the free surface, just use the initial deformation and keep the mesh constant.

include/aspect/mesh_deformation/interface.h
```diff
@@ -290,6 +293,15 @@ namespace aspect
...
         get_matching_mesh_deformation_object () const;
 
+
+        /**
+         * If multilevel solvers are used, we need a mapping on each multigrid level. These
+         * are automatically updated by this handler class and can be accessed with this
+         * method.
+         */
+        const Mapping<dim> &
+        get_level_mapping(int level) const;
```
gassmoeller, line 303:
```suggestion
        get_level_mapping(const unsigned int level) const;
```

include/aspect/mesh_deformation/interface.h
```diff
@@ -493,6 +510,22 @@ namespace aspect
 
         bool include_initial_topography;
 
+        /**
+         * If required, store a Mapping for each multigrid level.
```
gassmoeller, line 514:
```suggestion
         * If required, store a mapping for each multigrid level.
```

gassmoeller (commented):
Looks generally good, just two minor comments. What is the status of this PR in general? Can the WIP tag be removed?

gassmoeller:
Also this probably deserves a changelog entry.

tjhei:
Agreed, thank you.

I have two things left to do:
1. Validate that we get similar results to matrix-based for a more complicated example.
2. Hunt down a bug in a 3d test case that I see.

tjhei:
For reference, this is the problem in deal.II https://github.com/dealii/dealii/issues/12633 that I encountered.

tjhei:
The implementation was, unsurprisingly, incorrect as the mapping information was not updated correctly. This is now fixed.
While I still have this one case where we Assert in deal.II (to be fixed), I think we can proceed with merging this PR. Ready for review!

tjhei:
https://youtu.be/yz9CCYinN7Q

gassmoeller (commented):
Very nice, I have a few small comments.

On a philosophical note: We have a number of classes like MeshDeformationHandler that are both derived from SimulatorAccess and have direct simulator access through `sim`. I think it would be better if we would access the `const` functions of SimulatorAccess through `this->` whenever possible and only fall back to `*sim.` in the places where it is absolutely necessary to have write access. We are not very consistent about this at the moment. I could prepare a PR that updates the existing places, but this PR adds a number of new places where this happens. What do you think?

source/simulator/stokes_matrix_free.cc
```diff
@@ -1344,6 +1348,17 @@ namespace aspect
...
   {
+    if (sim.mesh_deformation)
+      {
+        // Update the geometry information stored in the MatrixFree
+        // objects from the mapping.  Grab the mapping stored in the
+        // object and do not replace with sim.mapping as we have
+        // different mappings per level.
+        for (auto &obj : matrix_free_objects)
+          obj->update_mapping(*obj->get_mapping_info().mapping);
+      }
```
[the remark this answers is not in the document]
tjhei: How about we rename ``evaluate_material_model`` to something like ``assemble``? This is how I see all the operations that have to happen. I don't think it is useful to expose more than one function if you only ever call all of them in the same order.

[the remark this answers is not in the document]
gassmoeller: Sounds reasonable to me.

tjhei:
updated.

> I think it would be better if we would access the `const` functions of SimulatorAccess through `this->` whenever possible and only fall back to `*sim.`

I agree that this makes sense. I changed a couple of places (but only those that I newly introduced) as it would be weird to include other changes in this PR.

How about you (or I) do a followup?

gassmoeller (approved):
Looks good, thanks!

Issue #4246, opened 2021-07-16
Improve field parsing


bobmyhill:
ASPECT has "compositional" fields not only for compositions, but also other properties like grain size and stress (and soon, a lot of other fields with the updates planned by @MFraters). 
Some material models require certain fields in specific (but different) orders, and different rheologies don't play nicely with the others. With the increasing modularisation of the material models, I think this may become a significant stumbling block.

Some potential improvements:
[ ] Fix Rheology::Elasticity so that it doesn't require material parameters for the stress fields.
[ ] Make a new field label parameter, to automatically identify the type of field to which each name corresponds
[ ] Make masks for each type of field
[ ] Update the material models so that they don't care about the order of different fields.

Some of these changes are likely to break backwards compatibility, so I'm opening the issue for discussion.

tjhei:
Agreed. I think we should have a conversation about this.

Related #4182

naliboff:
Agree this needs to be discussed and that many of the proposed changes are needed, regardless of whether backwards compatibility is broken. #3927 is also related.

PR #4258, opened 2021-07-18
add new field names parameter

bobmyhill:
This PR adds a new parameter called "Types of fields" that is used to specify the types of all the "compositional" fields. 
The parameter is used to fill a new FieldTypeIndices structure, that itself contains a series of std::vector<unsigned int> objects of the indices corresponding to each field type.

This structure can be called from within material models via `this->introspection().get_field_type_indices()`

* [x] I have modified a testcase for the new feature/benchmark in the [tests/](../blob/master/tests/) directory.


bobmyhill:
This PR starts to address #4246.

tjhei:
Bob, did a conversation about this happen last week that I missed?

bobmyhill:
Hi @tjhei, no, you didn't miss anything. 
This PR is to kick off that discussion - it's an initial idea for what the first step should be (and is one that allows a series of small changes), but it doesn't have to be the final decision.

naliboff:
@bobmyhill - Thank you for tackling this and getting the technical discussion started!

I have a high-level question about the current structure - Is there a need to define more than two compositional field types in `FieldTypeIndices`? 

If the primary motivation is to (1) not use "non-lithology" fields in the material averaging and (2) not have to assign "non-lithology" fields entries in various portions of the parameter file, could `FieldTypeIndices` simply have two compositional field types: lithology and non-lithology (or composition and non-composition)?

Is there another motivation to assign additional categories (grain size, stress, etc)? I am envisioning that people may want to add a large number of new types to this category as time goes in, but perhaps this is a non-issue?

bobmyhill:
Hi @naliboff, 
I'm not sure whether *every* field would need an indices vector, but there are material models that need to know where the grain size, stress and porosity fields are. And while most of these only need two types of fields at the moment (so you could find the non-compositional fields by negation), I think it'll be good to be able to have asserts like `AssertThrow(this->introspection().get_field_type_indices().grain_size.size() == 1, ...)`.

I don't think its a massive problem to add more vectors when needed.

bobmyhill:
Another thought while writing #4260: it might be possible to avoid making the new "Types of fields" parameter if we enforce naming for all non-compositional "compositional" fields. i.e. if stress fields always start with the characters "stress" etc etc, then we can use these names in parameters.cc to make the vectors of indices.

This might work quite well, as I *think* all field types are already parsed by name.

tjhei:
> it might be possible to avoid making the new "Types of fields" parameter if we enforce naming for all non-compositional "compositional" fields

I don't think we would be able to have all information be encoded in names, but maybe I am wrong.

Before we go this way, we should think about a few more use cases for these types. I can think of:
- This can be used to clamp field values automatically to [0,1] or to 0 or 1 depending on the type.
- Can we encode if a field is a mass fraction, volume fraction, or a binary yes/no?
- How to deal with the background field? should we maybe include the background field as an explicit field that is computed as 1-c1-c2... automatically?

naliboff:
> Another thought while writing #4260: it might be possible to avoid making the new "Types of fields" parameter if we enforce naming for all non-compositional "compositional" fields. i.e. if stress fields always start with the characters "stress" etc etc, then we can use these names in parameters.cc to make the vectors of indices.

In either scenario, I think we just need to have one generic field that can be assigned to all non-compositions/lithologies, but does not fit into an existing or logical category. This field would be called something like "non-lithology" or  "non-composition"

bobmyhill:
> I don't think we would be able to have all information be encoded in names, but maybe I am wrong.

It would certainly be more messy, but I can imagine an implementation where every name was just prefaced with its type. I don't think that's the best strategy, but it would be the change that would be most backwards-compatible.

> Before we go this way, we should think about a few more use cases for these types. I can think of:

>     * This can be used to clamp field values automatically to [0,1] or to 0 or 1 depending on the type.

Oh, interesting. I hadn't thought of this, but yes, that might be useful.

>     * Can we encode if a field is a mass fraction, volume fraction, or a binary yes/no?

maybe instead of "composition" we could call the fields "mass_fraction" and "volume_fraction". But we do then hit the issue that for incompressible models it doesn't make much sense to talk about mass or volume fractions.

>     * How to deal with the background field? should we maybe include the background field as an explicit field that is computed as 1-c1-c2... automatically?

You mean so that we can give it a name? I can't think of any occasion when we would want a background field that wasn't a mass/volume fraction.

bobmyhill:
> > Another thought while writing #4260: it might be possible to avoid making the new "Types of fields" parameter if we enforce naming for all non-compositional "compositional" fields. i.e. if stress fields always start with the characters "stress" etc etc, then we can use these names in parameters.cc to make the vectors of indices.
> 
> In either scenario, I think we just need to have one generic field that can be assigned to all non-compositions/lithologies, but does not fit into an existing or logical category. This field would be called something like "non-lithology" or "non-composition"

So one future case I can think of where that isn't true would be a model with melt and elasticity. Then you'd need fields for composition, porosity and stress.

naliboff:
> So one future case I can think of where that isn't true would be a model with melt and elasticity. Then you'd need fields for composition, porosity and stress.
@bobmyhill - Sorry, my explanation above did not come out as intended. In addition to all the named fields types (porosity, stress, etc), I think we would want an **additional** generic type that one could use. Agree including the specific names makes sense for sense for many field types.

bobmyhill:
> @bobmyhill - Sorry, my explanation above did not come out as intended. In addition to all the named fields types (porosity, stress, etc), I think we would want an **additional** generic type that one could use. Agree including the specific names makes sense for sense for many field types.

Ah, I see what you mean. That's why I've added the option "other" to the list. We could make that another vector too, if the generic field indices are needed for anything.

bangerth:
I agree with @tjhei that encoding information in *names* is not a scalable strategy. In essence, you want to attach semantic meaning to fields (like his idea of volume vs mass fraction, 0-1 fields, etc.) and that can be done by either encoding field properties via `enum`s if you can enumerate both the kinds of traits you want to encode and their possible values, or by using a `std::map<std::string,std::string>` if you can't enumerate either all possible traits, or a trait's values.

include/aspect/introspection.h
```diff
@@ -353,6 +353,13 @@ namespace aspect
       const std::vector<std::string> &
       get_composition_names () const;
 
+
+      /**
+       * A function that returns the full list of compositional field type indices.
+       */
+      const typename Parameters<dim>::FieldTypeIndices &
+      get_field_type_indices () const;
```
bangerth, line 361: Could you document what a "field type index" is?

include/aspect/introspection.h
```diff
@@ -381,6 +388,8 @@ namespace aspect
        */
       std::vector<std::string> composition_names;
 
+      typename Parameters<dim>::FieldTypeIndices field_type_indices;
```
bangerth, line 391: documentation missing

include/aspect/parameters.h
```diff
@@ -373,6 +373,18 @@ namespace aspect
       }
     };
 
+    /**
+     * A data structure containing vectors of the indices of each
+     * compositional field type.
+     */
+    struct FieldTypeIndices
+    {
+      std::vector<unsigned int> composition;
```
bangerth, line 382: Oh, I see. Are you saying that this vector contains the indices of those fields that correspond to compositions? If so, what is the basis of this index? Within all compositional fields? Or vector components? Or vector blocks?

bangerth (commented):
Out of curiosity, why do you store this as vectors of indices, rather than one vector of "the type of the i'th field"?

gassmoeller:
Can we discuss this PR at the next virtual user meeting (Aug 3rd)? I like the intended functionality, but the way we implement it will be a bit tricky. An `std::vector<FieldType>` (with n_compositional_fields entries) with `FieldType` being an enum of all types seems good, unless there are fields that could have multiple `FieldTypes` (`std::vector<std::vector<FieldType>>`?). As Wolfgang mentioned std::multimap<std::string,std::string> is the most general structure, but it could be extremely slow and we should think about alternatives first. Also the way we access this matters, do we want to loop over compositional fields and check for each its type? Or do we often want to find all compositional fields of a certain type? Or both?

gassmoeller:
Oh, I should mention I ask for the next user meeting, because I am away between tomorrow and then, but I would like to join the discussion.

bobmyhill:
Hi @gassmoeller, hi @bangerth,

The most common (?only) use of this new structure will be to loop over all fields of a certain type. This is why I chose to store the information as vectors of indices corresponding to each type. The alternative (one vector of "the type of the i'th field") would require always looping over the whole vector. I know that enum comparisons are not particularly expensive, but they aren't necessary at all, and I don't think they would add much in the way of readability (see #4260 for how the modifications would look with the current set up). Is there a benefit that I'm missing?

Yes, we can discuss this at the next user meeting.

bangerth:
I'm just thinking of what happens if we end up adding more and more semantic properties to fields. I'd much rather have a 
```
  struct FieldDescription
  {
    enum {...} property_1;
    enum {...} property_2;
  };
```
and then store a `std::vector<FieldDescription>` somewhere. If that turns out to be slow, we can still introduce the transpose table you suggest (indexing not by row=field index but by column=property) or do that only for the most commonly queried properties. I would be very much surprised if that is actually needed, and would like to avoid making this the first step.

bobmyhill:
Thanks for all of the suggestions everyone! I think this is ready for another round of reviews.

gassmoeller (commented):
As discussed in the meeting today, structure looks great, please try to unify names with existing functions to use `compositional field` if you talk about all compositional fields, and `chemical composition` if you talk about compositional fields representing chemical composition.
Just a few additional comments.

source/simulator/parameters.cc
```diff
@@ -1167,6 +1167,12 @@ namespace aspect
       prm.declare_entry ("Names of fields", "",
                          Patterns::List(Patterns::Anything()),
                          "A user-defined name for each of the compositional fields requested.");
+      prm.declare_entry ("Types of fields", "",
```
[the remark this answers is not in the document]
tjhei: I think this should either be empty (meaning all are generic) or have exactly n_compositions entries, so I think the default here is ok.

source/simulator/parameters.cc
```diff
@@ -1784,6 +1790,30 @@ namespace aspect
...
+          (prm.get ("Types of fields"));
+
+      AssertThrow ((x_compositional_field_types.size() == 0) ||
+                   (x_compositional_field_types.size() == 1) ||
+                   (x_compositional_field_types.size() == n_compositional_fields),
+                   ExcMessage ("The length of the list of names for the field types of compositional "
+                               "fields needs to be empty, or have one entry, or have a length equal to "
+                               "the number of compositional fields."));
+
+      // If only one method is specified apply this to all fields
```
tjhei, line 1804: Not sure how useful it is to support the case of extending one type to all fields

include/aspect/parameters.h
```diff
@@ -373,6 +373,50 @@ namespace aspect
...
+     * (i.e., whether the field corresponds to a composition, porosity, etc.).
+     */
+    struct FieldDescription
+    {
+      /**
+       * This enum lists available field types.
+       */
+      enum Kind
+      {
+        composition,
```
tjhei, line 388: this should be updated to ``chemical_compostion``, right?

source/simulator/parameters.cc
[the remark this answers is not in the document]
bobmyhill: I prefer @gassmoeller's suggestion, because there are several tests where all fields are chemical, and it would be easier to set that in one word. I agree that generic should be the default.

source/simulator/parameters.cc
```diff
@@ -1784,6 +1790,30 @@ namespace aspect
       AssertThrow (normalized_fields.size() <= n_compositional_fields,
                    ExcMessage("Invalid input parameter file: Too many entries in List of normalized fields"));
 
+      std::vector<std::string> x_compositional_field_types
+        = Utilities::split_string_list
+          (prm.get ("Types of fields"));
+
+      AssertThrow ((x_compositional_field_types.size() == 0) ||
```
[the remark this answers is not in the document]
bobmyhill: I agree.

include/aspect/parameters.h
bobmyhill: Or just "chemical"?

source/simulator/parameters.cc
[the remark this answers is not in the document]
gassmoeller: Also extending a parameter that has only a single entry into a list that is as long as number of compositional fields is something we do in a lot of material models. I do not recall another parameter that is empty, but interpreted as something else.

include/aspect/parameters.h
gassmoeller: I think I would prefer chemical_composition, mostly because I associate it more with major element composition. I am thinking of how to classify fields like water content or trace elements, which we would not want to fall into the category of fields that are typically used to average density etc., but would clearly be 'chemical'. Is that just my personal opinion?

bobmyhill: I was just choosing the shorter option. I'm happy to go with your suggestion!

source/simulator/parameters.cc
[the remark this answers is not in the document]
tjhei: Sure, sounds good.

tjhei:
Some places you call it Type, in others Description. We should probably agree on one or the other.

bobmyhill:
Hi @tjhei, the type/Description distinction is intentional. I tried to follow @bangerth's suggestion in making a structure (`CompositionalFieldDescription`) that can contain multiple enums. `type` is the first of those enums.

include/aspect/parameters.h
```diff
@@ -373,6 +373,50 @@ namespace aspect
...
+     * A data structure containing a description of each compositional field.
+     * At present, this structure only includes the field type
+     * (i.e., whether it is of type chemical_composition, porosity, etc.).
+     */
+    struct CompositionalFieldDescription
+    {
+      /**
+       * This enum lists available compositional field types.
+       */
+      enum Kind
```
tjhei, line 386: Would it make sense to call it Type?

source/simulator/introspection.cc
```diff
@@ -300,6 +301,18 @@ namespace aspect
     return composition_names;
   }
 
+
+
+  template <int dim>
+  const std::vector<typename Parameters<dim>::CompositionalFieldDescription> &
+  Introspection<dim>::get_composition_descriptions () const
+  {
+    // Return the full vector of CompositionalFieldDescription objects
```
tjhei, line 310: This comment is somewhat redundant, I think.

source/simulator/parameters.cc
```diff
@@ -1784,6 +1789,29 @@ namespace aspect
...
+                   ExcMessage ("The length of the list of names for the field types of compositional "
+                               "fields needs to have one entry or have a length equal to "
+                               "the number of compositional fields."));
+
+      // If only one method is specified apply this to all fields
+      if (x_compositional_field_types.size() == 1)
+        x_compositional_field_types = std::vector<std::string> (n_compositional_fields, x_compositional_field_types[0]);
+
+      // If field types have been defined, fill the corresponding index vectors
+      if (x_compositional_field_types.size() == n_compositional_fields)
```
tjhei, line 1807: Isn't this always true, because you extend it above?

bobmyhill:
@tjhei, @gassmoeller, I've restored the default option to have an empty vector of types. This provides the cleanest way to assert that prm files have the correct format, and to provide a useful error message to users if they need to update their param files. I hope that's ok with both of you. I've added a comment in the code describing the different use cases.

Hopefully this PR is nearly ready to go. Thanks for your patience.

source/simulator/parameters.cc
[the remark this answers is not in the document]
bobmyhill: See my new comment. It's easier to write good error messages if we can preserve whether the user has actively chosen to make all fields *something*, or just used a default option.

tjhei:
> Hi @tjhei, the type/Description distinction is intentional. I tried to follow @bangerth's suggestion in making a structure (`CompositionalFieldDescription`) that can contain multiple enums. `type` is the first of those enums.

I see. That makes sense. Should we move other information into this struct (maybe not in this PR)?

bobmyhill:
We can do. Definitely not in this PR, but I think the field names could easily be moved in to this struct, and probably the methods too.

tjhei (approved):
Looks good to me, but I think this needs a few more eyes.

Can you please squash the commits?

include/aspect/introspection.h
```diff
@@ -353,6 +353,14 @@ namespace aspect
       const std::vector<std::string> &
       get_composition_names () const;
 
+
```
gassmoeller, line 356: Remove this empty line.

include/aspect/parameters.h
```diff
@@ -373,6 +373,50 @@ namespace aspect
...
+
+      /**
+       * This function translates an input string into the
+       * available enum options for the type of compositional field.
+       */
+      static
+      Type
+      parse_type(const std::string &input)
+      {
+        if (input == "chemical_composition")
```
gassmoeller, line 403: I would prefer the input file to be free of underscores and use spaces instead. Internally `chemical_composition` is the right name, but in the input file `chemical composition` is more consistent with all our other parameters. Same goes for grain_size below. Sorry I didnt see that in my earlier review.

include/aspect/parameters.h
```diff
@@ -373,6 +373,50 @@ namespace aspect
...
+          return CompositionalFieldDescription::chemical_composition;
+        else if (input == "stress")
+          return CompositionalFieldDescription::stress;
+        else if (input == "grain_size")
```
gassmoeller, line 407: see above

source/simulator/parameters.cc
```diff
@@ -1172,11 +1172,16 @@ namespace aspect
       prm.declare_entry ("Names of fields", "",
                          Patterns::List(Patterns::Anything()),
                          "A user-defined name for each of the compositional fields requested.");
+      prm.declare_entry ("Types of fields", "",
+                         Patterns::List (Patterns::Selection("chemical_composition|stress|grain_size|porosity|generic")),
```
gassmoeller, line 1176: this needs to change then as well

source/simulator/parameters.cc
```diff
@@ -1172,11 +1172,16 @@ namespace aspect
...
+                         "A type for each of the compositional fields requested. "
+                         "Each entry of the list must be "
+                         "one of several recognised types.");
```
gassmoeller, line 1179: Sneaky :wink:
```suggestion
                         "one of several recognized types.");
```

source/simulator/parameters.cc
```diff
@@ -1172,11 +1172,16 @@ namespace aspect
...
       prm.declare_entry ("Compositional field methods", "",
                          Patterns::List (Patterns::Selection("field|particles|volume of fluid|static|melt field|prescribed field|prescribed field with diffusion")),
                          "A comma separated list denoting the solution method of each "
                          "compositional field. Each entry of the list must be "
-                         "one of the currently implemented field types."
+                         "one of the currently implemented field methods."
```
gassmoeller, line 1184: :+1:

source/simulator/parameters.cc
gassmoeller: Agreed, and also at the moment you do not specify what happens for x_compositional_field_types.size() == 0. Wouldnt we want to do the same thing as when x_compositional_field_types.size() == 1, except setting it to `generic` for all entries?

source/simulator/parameters.cc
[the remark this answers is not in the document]
gassmoeller: I am ok with that, but I dont see how you handle the case when the user does not specify anything at all. At the moment it will just not set any types, which I think is not what we want. I thought we always want to have this vector filled with n_compositional_fields entries, and if none are given they should all be `generic`.

gassmoeller (commented):
Very close, except I am not clear on what happens if the user does not specify anything at all. What would you want the behavior to be if nothing is set in the input file? I would think a vector with n_compositional_field entries that are all `generic` is the preferred behavior, but it seems you would prefer an empty vector? I like my suggestion, because then we would never have to check if this vector has the correct length (every field would always have a type, and if nothing else is specified this type would be `generic`). This is up for debate of course. Opinions?

bobmyhill:
Thanks @gassmoeller. I'm happy to remove the underscores and stray blank line.

I would like to keep the default behaviour as it is (so that if the user does not specify types, the vector is empty).
The reason for this is that in the next PR we are going to require types to be specified for some material models. This is going to break a lot of prms, and I want to add a simple assert to inform users of the changed behaviour. It's remarkably difficult to do this in a robust way if we have set all Types to generic (as in, I spent several hours trying to work out how to do it using #4260 and failed miserably).

Are you happy with this? If not, I can change it and then we can collectively try to solve the resulting problem with #4260.

gassmoeller:
I guess what I do not understand is which case does not work if all fields are by default generic fields. Would it not be the expected case that if a field is not specified it is a generic field? Any material model can require that the user specifies at least X fields of a specific type if it wants to make use of it.

My main opposition to the current approach is that every place that wants to use the new functionality has to do the following:
```
const auto &field_descriptions = this->introspection().get_field_descriptions();
// check if field descriptions have been set
if (field_descriptions.size() != 0)
{
  // only here we can use field_descriptions
}
```

Additionally think of the case when we move other information into FieldDescription. Once we move field names or advection methods into the same structure we have to have one entry per field, this information is not optional. So my question is mostly:
1. What case does currently not work if all fields are generic?
2. How would you circumvent the problems I listed above.

bobmyhill:
Ok, to answer your questions:
> What case does currently not work if all fields are generic?

None. Everything will currently work, because the types are not used.

> How would you circumvent the problems I listed above.

I prefer to see those "problems" as "features" :). I think that ASPECT throwing an error if the types vector is empty is a good thing - whereever we use this new vector, we should add asserts to make sure that the user *explicitly* says which fields they are using in the prm file.

To answer a question you didn't ask:
> What case will not work in future PRs if all fields are generic?

Basically everything. In #4260, I have started adding lines like
`std::vector<unsigned int> names_of_compositional_fields = find_names_of_type("chemical composition")`
Now, if the user hasn't declared that Field 1 is a chemical composition, it will be treated as generic. After this, parse_map_to_double_array takes a string (Densities, for example) that it has to parse, and the names of the compositional fields (empty). If the two have different sizes, there will be a somewhat cryptic error (which I improved a bit in #4337, but not enough to make it obvious what is going wrong).

What ASPECT *should* do (IMO) is check that "Types of fields" have been specified explicitly by the user for the material models that require Types of fields. With the PR as it currently stands, that is easy:
`AssertThrow(field_descriptions.size() != 0, ExcMessage("This material model requires that you set Types of fields in the Compositional fields subsection of the prm file."));`
Conversely, if we set all fields to generic, we can't know what the user intended. In that case, it would be vital to somehow check that the number of values processed by parse_map_to_double_array matches the number of fields of type X (usually X will be of type chemical composition). Even then, the error message will be more vague:
`AssertThrow(names_of_fields.size() == length_of_processed_string_vector, ExcMessage("The number of fields specified for the material property <XXXX> does not match the string provided. This could be because of a mistake in the property string, or because you have incorrectly specified the types of fields."));`

Actually, I'd prefer to have both types of AssertThrow. But we can only do that if we preserve what the user has included in their prm file.

bobmyhill:
Oh, wait, I think I understand why you dislike this approach.

In the future, we might have several things to put in field description, in which case we wuld want to have the correct size of vector, even if types have not been filled. So both strategies have drawbacks at the moment.

Here's an alternative: what if we include another Type, "unspecified", as the default option? Then the assert in the material models is still straightforward:
`AssertThrow(none(field_descriptions.type = Unspecified), ExcMessage("This material model requires that you set Types of fields in the Compositional fields subsection of the prm file to something other than 'unspecified'."));` 

Would that be ok with you?

gassmoeller:
Thanks for the explanation, and yes having an "unspecified" default is perfect. So we would have the options: "chemical composition, ... lots_of_other_things ..., generic, unspecified" and material models could distinguish between default values ("unspecified") and specified values that just dont have a category ("generic").

bobmyhill:
Brilliant :) ... thanks again; this has been a marathon, but hopefully it'll be worth it!

tjhei:
This looks good, but I think we need to explain to the user:
 1. choosing a type has influence on material models (and potentially other places) 
 2. what is the difference between generic and unspecified.

tjhei (approved):

PR #4260, opened 2021-07-19
[WIP] Field types for viscoelasticity and corresponding rheology and EoS modules

bobmyhill:
A continuation of #4258, to give a feeling for how extensive the changes to the material models will have to be. Opened for discussion.

This will break a lot of tests (61 from the first commit). This is because (a) EquationOfState::MulticomponentIncompressible, Rheology::Elasticity and MaterialModel::Viscoelastic now require the "Types of fields" parameter, and (b) because other rheology models that derive from EquationOfState::MulticomponentIncompressible and Rheology::Elasticity may not yet have been updated.

### DIFF
--- a/benchmarks/viscoelastic_bending_beam/viscoelastic_bending_beam.prm
+++ b/benchmarks/viscoelastic_bending_beam/viscoelastic_bending_beam.prm
@@ -90,10 +90,11 @@ subsection Boundary velocity model
   set Tangential velocity boundary indicators = bottom, top, right
 end
 
-# Number and name of compositional fields
+# Number, name and type of compositional fields
 subsection Compositional fields
   set Number of fields = 4
   set Names of fields  = ve_stress_xx, ve_stress_yy, ve_stress_xy, beam
+  set Types of fields  = stress,       stress,       stress,       composition
 end
 
 # Spatial domain of different compositional fields
@@ -138,9 +139,9 @@ subsection Material model
   set Model name = viscoelastic
 
   subsection Viscoelastic
-    set Densities            =  2800,  2800,  2800,  2800,  3300
-    set Viscosities          = 1.e18, 1.e18, 1.e18, 1.e18, 1.e24
-    set Elastic shear moduli = 1.e11, 1.e11, 1.e11, 1.e11, 1.e10
+    set Densities            =  2800,  3300
+    set Viscosities          = 1.e18, 1.e24
+    set Elastic shear moduli = 1.e11, 1.e10
     set Fixed elastic time step     = 1e3
     set Use fixed elastic time step = false
     set Use stress averaging        = false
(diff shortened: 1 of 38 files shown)


naliboff (commented):
I have only had a quick scan through, but overall this looks extremely promising!!

benchmarks/viscoelastic_bending_beam/viscoelastic_bending_beam.prm
```diff
@@ -138,9 +139,9 @@ subsection Material model
   set Model name = viscoelastic
 
   subsection Viscoelastic
-    set Densities            =  2800,  2800,  2800,  2800,  3300
-    set Viscosities          = 1.e18, 1.e18, 1.e18, 1.e18, 1.e24
-    set Elastic shear moduli = 1.e11, 1.e11, 1.e11, 1.e11, 1.e10
+    set Densities            =  2800,  3300
+    set Viscosities          = 1.e18, 1.e24
+    set Elastic shear moduli = 1.e11, 1.e10
```
naliboff, line 144:
```
  set Densities            =  2800,  3300
  set Viscosities          = 1.e18, 1.e24
  set Elastic shear moduli = 1.e11, 1.e10
```
The parameter files for 3D models with stress and multiple strain invariants are going to look quite a bit simpler 😄

Users will still need to list the values for "composition" type fields in the same order they are listed in Names of fields, but hopefully that does not lead to confusion. We would likely need to update the cookbooks, benchmarks, and tests to include comments to this effect.

source/material_model/rheology/elasticity.cc
```diff
@@ -119,11 +119,33 @@ namespace aspect
...
+
+        AssertThrow (stress_field_indices.size() == (dim == 2 ? 3 : 5),
+                     ExcMessage("The Compositional fields subsection must contain the parameter "
+                                "Types of fields. Additionally, there must be three or five fields "
+                                "corresponding to stress (for two- or three-dimensional problems respectively)."));
+
+        const std::vector<std::string> list_of_field_names = this->introspection().get_composition_names();
+
+        std::vector<std::string> list_of_composition_names(compositional_field_indices.size());
+        std::transform(compositional_field_indices.begin(),
```
anne-glerum, line 135: This function will probably be used a lot, maybe it should go into Utilities?

bobmyhill: Thanks @anne-glerum! I agree. The changes that @wolfgang and @gassmoeller proposed at our last meeting will result in a few new utilities functions. I'll try to shift as many multi-use functions out of the material models as possible.

anne-glerum: Good to know! Too bad I couldn't make it to the meeting.

MFraters (approved):
This looks good to me, and I think it is something we should be merging soon. It will allow to simplify parameter files quite a bit and prevent fields from being used incorrectly. ''

I just have a few small comments:
1. Please include in the changelog entry that includes that it will break current prm files which do use viscoelastic and do not include the field descriptions entry.
2. it contains duplicate code/commits with https://github.com/geodynamics/aspect/pull/4258 (thinking of the `CompositionalFieldDescription` code. Can you rebase?
3. Can you squash your commits in some logical blocks (e.g. on for the unit test, one or multiple for the test, one or multiple for code changes etc.)?

@naliboff and @anne-glerum do you have any further comments

PR #4261, opened 2021-07-19
Don't use std::make_unique just yet.

bangerth:
We jumped the gun on these places :-)

Related to #4188 .

/rebuild


tjhei:
I am not sure how useful this is, as we probably merge the version bump in the next few days. Anyways.

PR #4264, opened 2021-07-20
Fix RK2 particle advection across periodic boundaries in a spherical shell

gassmoeller:
Advecting particles across a periodic boundary in a spherical shell created some spurious patterns in the particle locations (see attached image, note the clumping of particles at the left boundary, and the streaks or lines of particles at the other side of the boundary). We figured out that the particle location was updated successfully across the boundary, but the intermediate storage vectors of the RK2 scheme need to be rotated as well, otherwise the intermediate storage and the current velocities are in different coordinate systems. In other words particles advected across the boundary in the first RK step were advected in different coordinate systems in the second step. Particles crossing the boundary in the second RK step are fine, because all their movement happened in the old coordinate system (this is what creates the stripes on one side of the boundary).

@jdannberg and @kchotalia can you give this branch a try and make sure it solves the issue? I tested it at our test case and it affects about 5 out of the 50 particles in the test.

This PR is an incomplete fix that only fixes the RK2 scheme (not RK4), and does also not address the fact that other tensor properties such as the integrated strain tensor would need to be rotated as well. I see no easy and clean way for a general fix inside the particle world class. Presumably we would need new interface functions in the geometry model (to provide the rotation matrix) and in the particle property interface (to let individual plugins decide if their property needs to be rotated when transported across a periodic boundary with orientation change). 

But I think this issue should definitely be fixed, and is also a candidate for cherry-picking into 2.3, as RK2 is the default integration scheme (although periodic boundaries in a spherical shell is kind of a niche application). Thoughts?

Screenshot 2021-06-21 at 18 38 48


source/particle/world.cc
```diff
@@ -553,28 +558,64 @@ namespace aspect
...
+                }
+
               typename ParticleHandler<dim>::particle_iterator particle = particle_handler->begin();
               for (; particle != particle_handler->end(); ++particle)
                 {
                   // modify the particle position if it crossed a periodic boundary
                   Point<dim> particle_position = particle->get_location();
 
+                  // define a rotation matrix for the new position depending on the boundary
+                  Tensor<2,dim> rotation_matrix;
```
bangerth, line 586: Can you explain this a bit? If I understand your code correctly, what you're really trying to do is move a particle that has moved out of one side of the domain back in from the other side, and because (i) the domain is a quarter shell and (ii) the origin is the center of the shell, this corresponds to rotating the point around the origin by a certain amount? I think this kind of explanation would be useful.

At the same time, I don't understand why this is correct. I can see that this works if a particle moves out of one side and back in on the other side, but not the other way around. What do you do there?

source/particle/world.cc
```diff
@@ -553,28 +558,64 @@ namespace aspect
...
+                      rotation_matrix[0][1] = 1.;
+                      rotation_matrix[1][0] = -1.;
                     }
                   else if (particle_position[1] < 0.)
                     {
-                      const double temp = particle_position[0];
-                      particle_position[0] = -particle_position[1];
-                      particle_position[1] = temp;
+                      rotation_matrix[0][1] = -1.;
+                      rotation_matrix[1][0] = 1.;
```
bangerth, line 596: Oh, I see. This must be the case. Maybe explain the construction of the two cases.

source/particle/world.cc
```diff
@@ -553,28 +558,64 @@ namespace aspect
...
                     }
                   else
                     continue;
 
-                  particle->set_location(particle_position);
+                  const Tensor<1,dim> rotated_tensor (rotation_matrix * particle_position);
```
bangerth, line 601: call this `periodic_position` to make it clearer what this actually is:
```suggestion
                  const Tensor<1,dim> periodic_position = rotation_matrix * particle_position;
```

source/particle/world.cc
```diff
@@ -553,28 +558,64 @@ namespace aspect
...
+              // Find out if we need to rotate the temporary properties
+              unsigned int property_index_to_rotate = numbers::invalid_unsigned_int;
+              if (Plugins::plugin_type_matches<Particle::Integrator::Euler<dim>>(*integrator) == true)
+                {
+                  // nothing to do
+                }
+              else if (Plugins::plugin_type_matches<Particle::Integrator::RK2<dim>>(*integrator) == true)
+                {
+                  // rotate original location
+                  property_index_to_rotate = get_property_manager().get_data_info().get_position_by_field_name("internal: integrator properties");
```
bangerth, line 570: This is ugly. First, you make the assumption that the RK2 method stores a tensor in this location, but this is really something only the RK2 method should know. Second, you make the assumption that there is only one field that needs to be rotated.

source/particle/world.cc
```diff
@@ -553,28 +558,64 @@ namespace aspect
...
 
-                  particle->set_location(particle_position);
+                  const Tensor<1,dim> rotated_tensor (rotation_matrix * particle_position);
+                  particle->set_location(Point<dim>(rotated_tensor));
+
+                  // do the same for tensor properties that need to be rotated
+                  if (property_index_to_rotate != numbers::invalid_unsigned_int)
+                    {
+                      // get the property
+                      Tensor<1,dim> property_tensor;
```
bangerth, line 608: Call this variable `intermediate_position`, because that is what it really is.

source/particle/world.cc
```diff
@@ -553,28 +558,64 @@ namespace aspect
...
+                      for (unsigned int i = 0; i < dim; ++i)
+                        property_tensor[i] = particle->get_properties()[property_index_to_rotate+i];
+
+                      // rotate the property
+                      const Tensor<1,dim> rotated_property_tensor = rotation_matrix * property_tensor;
```
bangerth, line 613: And then `periodic_intermediate_position`

bangerth (commented):
I've got to say that this is pretty ugly. It introduces a piece of code that needs specific knowledge about the geometry model, and moreover makes assumptions about what the integrators store in their internal storage fields. 

There is probably a "right" way to fix this in which the geometry model is given a position and, if it has periodic boundaries, returns the "periodic position" back on the other side. That would require a new `virtual` function. Then one would also have to introduce a `virtual` function into the integrators that lets these classes update their internal properties upon discovering that a particle has moved across a periodic boundary. The end result is probably a 300-lines patch. I'm not sure that's worth it, but the version you have here really looks quite brittle as well :-( Could you at least introduce a couple of TODOs that explain what one *should* be doing here?

gassmoeller:
> I've got to say that this is pretty ugly.

I agree on all points :-) (except I think the particle properties should be responsibly for deciding how to modify at periodic boundaries). I just needed a quick fix for @kchotalia's models before leaving this for a while (and it may be important for the release, because currently the functionality is broken).

On the plus side: We have not implemented new advection schemes since the particles were introduced, and I doubt that anyone uses something different than RK2 in production models. And RK2 will always only store a single quantity which is exactly the original location. Not sure what to do best (only cherry pick this into the release branch once I addressed your comments, but keep it out of master until we have a better fix?).

bangerth:
Cherry-picking seems reasonable.

I understand that my criticism of your current approach implies a lot of work, and I don't want to put that on you either. If you feel like you can't invest the time, then leave a fat TODO so we can get around to it at a later time.

tjhei:
My vote would be to leave the release as is. I was going to tag it yesterday, but I didn't get to it. If we pick anything, I probably need to retest compilers, etc..

gassmoeller:
> My vote would be to leave the release as is. I was going to tag it yesterday, but I didn't get to it. If we pick anything, I probably need to retest compilers, etc..

I understand that. Let me open a separate PR that is cherry-pickable, and that only adds a single assert that disables all integrators except for forward euler for periodic side bcs. Would you be ok with including that?

jdannberg:
I added a test case that makes it easier to see what happens to particles at periodic boundaries in a spherical shell (which I thought would be useful since we didn't spot this bug with the existing tests).

gassmoeller:
I came up with a solution that avoids any specific code inside world.cc and moves the responsibility as far as possible into the individual plugins. Geometry models would be now responsible for providing a function that can adjust positions according to their periodic boundaries and the particle integrators are responsible for adjusting the particle positions by calling this geometry model function. The only quirk is that when a particle needs to be adjusted, the temporary data like the RK2 intermediate values need to be adjusted even if they are not outside the domain, because all stored data needs to be in the same orientation so to speak. Otherwise the new location could be on the left side of the domain, but the old temporary location on the right side of the domain. I included that in the interface to the geometry model function, but it is a bit intuitive (I explain it in the documentation).

Let me know if you think this is in general the correct approach, then I can polish the rest of the PR and implement it for RK4 as well.

bangerth:
Can you rebase?

include/aspect/geometry_model/spherical_shell.h
```diff
@@ -136,6 +136,16 @@ namespace aspect
         std::set< std::pair< std::pair<types::boundary_id, types::boundary_id>, unsigned int>>
             get_periodic_boundary_pairs () const override;
 
+        /**
+         * @copydoc Interface::adjust_positions_for_periodicity
+         *
+         * Apply a rotation to all points to account for periodicity.
```
bangerth, line 142: same question as for the box geometry

include/aspect/geometry_model/two_merged_boxes.h
```diff
@@ -135,6 +135,16 @@ namespace aspect
         std::set< std::pair< std::pair<types::boundary_id, types::boundary_id>, unsigned int>>
             get_periodic_boundary_pairs () const override;
 
+        /**
+         * @copydoc Interface::adjust_positions_for_periodicity
+         *
+         * Apply a translation to all points to account for periodicity.
```
bangerth, line 141: and here

source/geometry_model/spherical_shell.cc
```diff
@@ -372,6 +372,47 @@ namespace aspect
...
+              rotation_matrix[0][1] = 1.;
+              rotation_matrix[1][0] = -1.;
+            }
+          else if (position[1] < 0.)
+            {
+              rotation_matrix[0][1] = -1.;
+              rotation_matrix[1][0] = 1.;
+            }
+          else
+            return;
```
bangerth, line 403: You can probably make this easier to understand if you used
```suggestion
          Tensor<2,dim> rotation_matrix;

          if (position[0] < 0.)
            {
              rotation_matrix = Physics::Transformations::Rotations::rotation_matrix_2d(...plus or minus pi/2);
            }
          else if (position[1] < 0.)
            {
              rotation_matrix = Physics::Transformations::Rotations::rotation_matrix_2d(...minus or plus pi/2);
            }
          else
            return;
```
But then I realized that the function only exists in 2d and you have dimension independent code here. So forget about it.

include/aspect/particle/integrator/euler.h
```diff
@@ -77,6 +86,14 @@ namespace aspect
...
             */
           static const unsigned int n_integrator_properties = 0;
+
+        private:
+          /**
+           * Cache whether the geometry model has periodic boundaries. If so,
+           * we need to adjust the particle position if it crosses a periodic
+           * boundary.
+           */
+          bool has_periodic_boundary;
```
bangerth, line 96: I think it would be useful to rename the variable as geometry_has_periodic_boundary because it's a property of the geometry, not of the Euler/RK2/RK4 integrator.

I also thought some about whether it's really useful to store this sort of information in the class at all. It's really not a property of the integrator. You're just caching this information because you think it's expensive to re-compute over and over. But is this actually true? My preference would be to just compute it in the one place (in each class) where you actually use it.

source/particle/integrator/euler.cc
```diff
@@ -49,7 +55,12 @@ namespace aspect
              it != end_particle; ++it, ++old_velocity)
           {
             const Point<dim> loc = it->get_location();
-            it->set_location(loc + dt * (*old_velocity));
+            Point<dim> new_location = loc + dt * (*old_velocity);
+
+            if (has_periodic_boundary)
+              this->get_geometry_model().adjust_positions_for_periodicity(new_location);
```
bangerth, line 61: This is the place where I mention that you use the `has_periodic_boundary` variable. I think you might as well inline the check for periodicity in the `if` condition.

source/particle/integrator/rk_2.cc
```diff
@@ -75,12 +78,18 @@ namespace aspect
             if (integrator_substep == 0)
               {
                 const Tensor<1,dim> k1 = dt * (*old_velocity);
-                const Point<dim> loc0 = it->get_location();
+                Point<dim> loc0 = it->get_location();
+                Point<dim> new_location = loc0 + 0.5 * k1;
+
+                // Check if we crossed a periodic boundary and if necessary adjust positions
+                if (has_periodic_boundary)
+                  this->get_geometry_model().adjust_positions_for_periodicity(new_location,
```
bangerth, line 86: same here

source/particle/integrator/rk_4.cc
```diff
@@ -71,64 +74,113 @@ namespace aspect
...
+                if (has_periodic_boundary)
+                  {
+                    for (unsigned int i=0; i<dim; ++i)
+                      {
+                        k[1][i] = properties[property_index_k[1] + i];
+                        k[2][i] = properties[property_index_k[2] + i];
+                      }
+
+                    this->get_geometry_model().adjust_positions_for_periodicity(new_location,
+                                                                                ArrayView<Point<dim>>(&k[0],4));
```
bangerth, line 148: Nice trick with building the array view with 2, 3, 4 elements :-)

source/particle/world.cc
```diff
@@ -456,140 +452,7 @@ namespace aspect
...
-            }
-
-        }
-      else
-        {
-          AssertThrow(this->get_geometry_model().get_periodic_boundary_pairs().size() == 0,
-                      ExcMessage("Periodic boundaries combined with particles currently "
-                                 "only work with box, two merged boxes, and spherical shell geometry models."));
-        }
-    }
```
bangerth, line 592: That's a nice chunk of code that's gone! :-)

bangerth (commented):
My apologies for taking so long to look at this. Feel free to send me another email if you want more feedback!

gassmoeller:
Thanks for coming back to this! I think I addressed all comments, ready for another look.

Caching the periodic boundaries is necessary, because our current geometry interface requires constructing the std::set of the boundary indicators (even if the geometry model would cache it -- which they do not at the moment -- they would need to return by value). I profiled a benchmark and if I wouldnt precompute the value 25% of the total cost of particle advection would be checking if there are periodic boundaries. But I reworked this to only check once per cell instead of once per particle, this makes it a local variable and reduces the cost to about 1% of the total advection cost, which I think is acceptable (the cost shrinks further the more particles per cell we have, I profiled with 20 particles per cell). Are you ok with the current state?

source/particle/world.cc
```diff
@@ -21,12 +21,8 @@
 #include <aspect/particle/world.h>
 #include <aspect/global.h>
 #include <aspect/utilities.h>
-#include <aspect/compat.h>
-#include <aspect/geometry_model/box.h>
-#include <aspect/geometry_model/two_merged_boxes.h>
-#include <aspect/geometry_model/spherical_shell.h>
-#include <aspect/particle/integrator/euler.h>
 #include <aspect/citation_info.h>
+#include <aspect/geometry_model/interface.h>
```
bangerth, line 25: I don't think you need this any more. If it compiled before, it should be compiling now even without this addition given that you are only removing code in this file.

include/aspect/particle/integrator/euler.h
```diff
@@ -38,7 +39,7 @@ namespace aspect
        * @ingroup ParticleIntegrators
        */
       template <int dim>
-      class Euler : public Interface<dim>
+      class Euler : public Interface<dim>, public SimulatorAccess<dim>
```
bangerth, line 42: I think that you don't actually need this any more now that you don't need the `initialize()` function any more.

bangerth: Same with the `#include` file.

bangerth (approved):

bangerth:
Yes, looks good. Thanks for going the extra mile and timing the cost!

I'm going to merge so you can make progress. There are a couple of comments I left above that you can, if you want, address in a follow-up. But it's not worth waiting for it to finally get this merged!

gassmoeller:
Thanks! The include in world.cc is indeed unnecessary, I removed it in #4421. 
But Integrator::Euler needs to be derived from SimulatorAccess now, because it requires access to the geometry model for the periodicity correction.

bangerth:
On 12/1/21 06:34, Rene Gassmoeller wrote:
> But Integrator::Euler needs to be derived from SimulatorAccess now, 
> because it requires access to the geometry model for the periodicity 
> correction.

Ah yes, that makes sense!

PR #4272, opened 2021-07-21
Disable failing particle integration schemes in periodic shells

gassmoeller:
This is a minimal version of #4264 that only disables the broken functionality. I have checked that the test still gives reasonable results with forward euler, and that it now crashes if used with rk2 or rk4.#4264 
@tjhei: Would you be ok cherry-picking this into the release?


tjhei (approved):
Thank you.

tjhei:
Do you want this change on master as well?

bangerth (approved):

PR #4276, opened 2021-07-24
remove PETSc cmake and misc

tjhei:
remove PETSc from cmake, manual, CI
part of #3338


MFraters (approved):
What about the PETSC tests? Or do you remove them in a separate pull request

tjhei:
I left that for another day. There are tests and test logic that needs to be cleaned up.

bangerth (approved):

PR #4278, opened 2021-07-24
update zenodo DOI in readme

tjhei:


MFraters:
I tested the link and the preview, and both look good :)

PR #4285, opened 2021-07-28
Use \[...\] instead of $$...$$.

bangerth:
See https://tex.stackexchange.com/questions/503/why-is-preferable-to for the reason.

/rebuild


source/particle/integrator/interface.cc
```diff
@@ -189,11 +189,11 @@ namespace aspect
                                "solves. If, for example, we call the numerical solution "
                                "of the ODE $\\tilde{\\mathbf x}_{k,h}(t)$, then the "
                                "error will typically satisfy a relationship like "
-                               "$$"
+                               "$["
```
tjhei, line 192: Shouldn't this be \\[ ?

bangerth:
Oh, doh. Fixed!

tjhei:
I think you need to do ``\\[`` (which is what I had typed in my first comment, but github doesn't display ?!)

bangerth:
Brain doesn't appear to work today. ... Fixed.

tjhei (approved):

PR #4290, opened 2021-07-28
Move a screen message.

bangerth:
I was confused by the presence of an message on screen that turns out to just be written in the wrong place. Move it to where the action described in the message is actually happening, not before the guarding `if`.

/rebuild


tjhei (approved):

tjhei:
Please update test results.

bangerth:
Should be fixed now!

tjhei (approved):

Issue #4190, opened 2021-07-13
Avoid shady use of probable '\0' at the end of a std::string


bangerth:
We have this kind of code in a number of places in the postprocessors:
```
     if (Utilities::MPI::this_mpi_process(this->get_mpi_communicator()) == 0)
        {
          std::ofstream file (filename.c_str());

          file << "# "
               << ((dim==2)? "x y" : "x y z")
               << " density  viscosity " << std::endl;

          // first write out the data we have created locally
          file << output.str();

          std::string tmp;
          tmp.resize (max_data_length, '\0');

          // then loop through all of the other processors and collect
          // data, then write it to the file
          for (unsigned int p=1; p<Utilities::MPI::n_mpi_processes(this->get_mpi_communicator()); ++p)
            {
              MPI_Status status;
              // get the data. note that MPI says that an MPI_Recv may receive
              // less data than the length specified here. since we have already
              // determined the maximal message length, we use this feature here
              // rather than trying to find out the exact message length with
              // a call to MPI_Probe.
              MPI_Recv (&tmp[0], max_data_length, MPI_CHAR, p, mpi_tag,
                        this->get_mpi_communicator(), &status);

              // output the string. note that 'tmp' has length max_data_length,
              // but we only wrote a certain piece of it in the MPI_Recv, ended
              // by a \0 character. write only this part by outputting it as a
              // C string object, rather than as a std::string
              file << tmp.c_str();
            }
        }
      else
        // on other processors, send the data to processor zero. include the \0
        // character at the end of the string
        {
          MPI_Send (&output.str()[0], output.str().size()+1, MPI_CHAR, 0, mpi_tag,
                    this->get_mpi_communicator());
        }
```
This makes the assumption that `std::string` actually stores a `\0` at the end of the string, that we can access it, and that that whole scheme makes sense. This is not useful -- we should only transport the text we know exists, not something we *believe* should be there.

Found while looking at #3713, which I suspect copies this from other places.

PR #4291, opened 2021-07-29
simplify MPI gather for file concatenation

tjhei:
proposed solution to address #4190 by using deal.II functionality.

MPI::gather uses boost serialization and receives all data before we can
process it. So this solution might not have great performance for large
runs. I don't think that matters here, though.


bangerth (approved):
Nice work, thanks!

bangerth:
The same check as always hangs. I'll merge anyway.

PR #4293, opened 2021-07-31
remove PETSc support from testsuite

### LINKED ISSUE #3338: remove PETSc support

tjhei:
with that, we are done with #3338.

closes #3338.


bangerth (approved):
Yay!

PR #3808, opened 2020-08-15
corrected geoid parameter typo

bobmyhill:
The PR corrects a geoid parameter typo:
Include the contributon from dynamic topography -> Include the contribution from dynamic topography

PR #4298, opened 2021-08-03
Fix a mis-spelled parameter.

bangerth:
Found while reading through the hackathon report where this parameter name was copy-pasted :-) @mfmweerdesteijn @jaustermann @gassmoeller -- FYI, I think this is going to affect your input files.

/rebuild

### DIFF
--- a/doc/manual/parameters.tex
+++ b/doc/manual/parameters.tex
@@ -17967,13 +17967,13 @@ This parameter is an alias for the parameter ``\texttt{Output surface topography
 
 
 {\it Possible values:} A boolean value (true or false)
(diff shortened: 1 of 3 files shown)


gassmoeller:
This parameter was replaced by two new parameters in #3776 ("Include surface topography contribution" and "Include CMB topography contribution") and the current code already checks that the misspelled parameter is only set to the default value (see also #3808). Rather than fixing it, let's remove it completely. If we break backward compatibility we can just get rid of it.

bangerth:
Ah, I totally forgot. I had dimly remembered that I had seen this before...

PR #4299, opened 2021-08-04
GMG Newton: correct Stokes rhs

zjiaqi2018:
Adding the Newton part to the function`StokesMatrixFreeHandlerImplementation::correct_stokes_rhs()`. @tjhei

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1774,6 +1774,10 @@ namespace aspect
   template <int dim, int velocity_degree>
   void StokesMatrixFreeHandlerImplementation<dim, velocity_degree>::correct_stokes_rhs()
   {
+
+    // We never include Newton terms in step 0 and after that we solve with zero boundary conditions.
+    // Therefore, we don't need to include Newton terms here.
+
     const bool is_compressible = sim.material_model->is_compressible();
 
     dealii::LinearAlgebra::distributed::BlockVector<double> rhs_correction(2);
@@ -1835,9 +1839,9 @@ namespace aspect
 
             SymmetricTensor<2,dim,VectorizedArray<double>> sym_grad_u =
                                                           velocity.get_symmetric_gradient (q);
-            VectorizedArray<double> pres = pressure.get_value(q);
-            VectorizedArray<double> div = trace(sym_grad_u);
-            pressure.submit_value   (sim.pressure_scaling*div, q);
+            const VectorizedArray<double> pres = pressure.get_value(q);
+            const VectorizedArray<double> div = trace(sym_grad_u);
+            pressure.submit_value(sim.pressure_scaling*div, q);
 
             sym_grad_u *= viscosity_x_2;
 


tjhei:
Looks reasonable. Does it make a difference for the Newton benchmarks? It could be that it is only a tiny difference in the updates close to the boundary, that doesn't matter...
/rebuild

tjhei:
No difference in test output...

source/simulator/stokes_matrix_free.cc
```diff
@@ -1837,7 +1837,18 @@ namespace aspect
                                                           velocity.get_symmetric_gradient (q);
             VectorizedArray<double> pres = pressure.get_value(q);
             VectorizedArray<double> div = trace(sym_grad_u);
-            pressure.submit_value   (sim.pressure_scaling*div, q);
+
+            if (active_cell_data.enable_newton_derivatives)
+              {
+                // Note that derivative_scaling_factor has already been multiplied to newton_factor_wrt_pressure_table.
+                VectorizedArray<double> newton_pressure_term =
```
bangerth, line 1844:
```suggestion
                const VectorizedArray<double> newton_pressure_term =
```

source/simulator/stokes_matrix_free.cc
```diff
@@ -1837,7 +1837,18 @@ namespace aspect
                                                           velocity.get_symmetric_gradient (q);
             VectorizedArray<double> pres = pressure.get_value(q);
             VectorizedArray<double> div = trace(sym_grad_u);
```
bangerth, line 1839: These two variables could be `const` as well, I believe.

As a general rule, mark everything as `const` that can be marked as `const`.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1848,6 +1859,20 @@ namespace aspect
               for (unsigned int d=0; d<dim; ++d)
                 sym_grad_u[d][d] -= viscosity_x_2/3.0*div;
 
+            if (active_cell_data.enable_newton_derivatives)
+              {
+                SymmetricTensor<2,dim,VectorizedArray<double>> grads_phi_u_i = velocity.get_symmetric_gradient(q);
+
+                sym_grad_u +=  (grads_phi_u_i * active_cell_data.strain_rate_table(cell,q))
+                               * active_cell_data.newton_factor_wrt_strain_rate_table(cell,q);
```
bangerth, line 1867: This is now no longer the symmetric gradient, i.e., the variable *name* no longer matches its *value*. I think it would be better to do
```
  if (enable_newton_derivatives)
  {
    const something = sym_grad_u + ...;
    velocity.submit_sym_grad(...);
  }
  else
    velocity.submit_sym_grad (existing arguments);
```

bangerth (changes requested):

source/simulator/stokes_matrix_free.cc
zjiaqi2018: > No difference in test output...

Exactly! Is it because in the Newton method, the rhs is already the residual, so the system already has homogeneous Dirichlet boundary conditions?

tjhei:
You are correct, see https://github.com/geodynamics/aspect/blob/9eaaf4fbbdb5e81e1d2855464b8de45a139b8e97/source/simulator/core.cc#L1286
We never include Newton terms in step 0 and after that we solve with zero boundary conditions.

This means you can replace this PR by a comment stating that we don't need to include the Newton terms because of that.
Sorry.

tjhei (approved):

tjhei:
@bangerth will be unavailable for a week or so, so I will go ahead and merge.

bangerth:
Thanks, looked good!

PR #4300, opened 2021-08-05
Fix spherical velocity postprocessor

### LINKED ISSUE #4253: spherical_velocity_components test fails with deal.II9.3.1 ?
when I run the "viz_spherical_velocity_components" test on my aspect+deal.II9.3.1 it fails since the v_r, v_theta and v_phi values returned are zero. When I run it on aspect+deal.II9.2.0 it works fine.
(thank you Menno for the detective work)

gassmoeller:
This should fix #4253 (at least it did for me). @cedrict can you give this a try?

### DIFF
--- a/source/postprocess/visualization/spherical_velocity_components.cc
+++ b/source/postprocess/visualization/spherical_velocity_components.cc
@@ -34,7 +34,7 @@ namespace aspect
       SphericalVelocityComponents ()
         :
         DataPostprocessorVector<dim> ("spherical_velocity_components",
-                                      update_quadrature_points)
+                                      update_values | update_quadrature_points)
       {}
 
 


tjhei:
Does the missing flag trigger an assert in deal.II?

tjhei (approved):

gassmoeller:
It does not trigger an assert, it fills the not computed elements of `DataPostprocessorInputs::Vector<dim> &input_data` with 0. Should they be filled with `signaling_nan` instead if they were not computed?

tjhei:
Ah yes, this would need to happen inside deal.II. One could either set entries to NaN or resize the std::vectors to size 0:
https://www.dealii.org/developer/doxygen/deal.II/structDataPostprocessorInputs_1_1Vector.html#a8e034ae97f0447fb541bdb8ad0a4f599

tjhei (approved):
This is certainly more correct than before. I will merge and let @cedric report later...

PR #4305, opened 2021-08-09
Adding recent publication, Gouiza and Naliboff

ljhwang:
Add a Nature Communications publication by Moh and John to the list of publications.


tjhei:
These were run with a free surface, right? Are you really running 400+ million DoFs matrix-based?

Issue #4308, opened 2021-08-11
Convergence problems with GMG and spherical shells and large viscosity contrasts


gassmoeller:
@alarshi, @jdannberg and myself are experiencing convergence issues with the GMG preconditioner in spherical shell models with sharp viscosity variations at the boundary (AMG converges for the same models). I will continue to investigate, but wanted to open this issue to remind myself of our current understanding and to give @tjhei an update.

The attached model (and similar models) will sometimes randomly crash inside the preconditioner for the top left A block of the Stokes matrix. I have used the branch https://github.com/gassmoeller/aspect/tree/entropy_merged_output to output A and S block iterations in real time and it looks like this (full output attached):
```
*** Timestep 0:  t=0 years, dt=0 years
   Solving temperature system... 0 iterations.
   Solving Stokes system... Solving Stokes inner A block... 8 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 8 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 16 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 15 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 11 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 42 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 9 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 9 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 31 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 10 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 13 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 35 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... 8 iterations.
Solving Stokes inner S block... 2 iterations.
Solving Stokes inner A block... ---------------------------------------------------------
TimerOutput objects finalize timed values printed to the
screen by communicating over MPI in their destructors.
Since an exception is currently uncaught, this
synchronization (and subsequent output) will be skipped
to avoid a possible deadlock.
---------------------------------------------------------
```

[...]

PR #4309, opened 2021-08-12
Remove unnecessary computation

gassmoeller:
This quantity is not used in this assembler.


tjhei (approved):

tjhei:
How do you end up discovering this?! 😉

gassmoeller:
I compared the AMG assembly with the GMG assembly when I looked into #4308  ;-).

PR #4310, opened 2021-08-12
Add missing mapping

gassmoeller:
I think there is a mapping argument missing here.  I have not seen a large influence of this (it only affects the main diagonal for curved meshes with tangential bcs), and it does not solve #4308, but it seem more correct this way anyway. Also fix a comment.


tjhei (approved):

bangerth:
I wished we had an automatic way to find all of these places :-(

PR #4311, opened 2021-08-13
Change the representative point to include initial topography in the box model geometry

Djneu:
This PR is a small fix to include the initial topography in the representative point for the box model geometry. When using initial topography with adiabatic conditions, a floating point exception happens while computing the adiabatic profile. This seems to occur because when the compute_profile::get_property function calls the geometry depth function the representative point does not include the initial topography while the depth calculation does. 

### For all pull requests:

* [ X] I have tested my new feature locally to ensure it is correct.


gassmoeller (commented):
I think this is an improvement, but you will have to update the test results.

Also: You compute the topography in the middle of the top surface, but the representative point that is returned is at the south-west most corner. Is there a reason you prefer the middle of the domain? There is no fundamental reason why we should not use the middle, but I think it would be more consistent to use x=0/y=0 as well if we return this as representative point.

Djneu:
Hi both,
Thank you for taking a look at it. I've made the suggested change to clarify the point location, and updated the tests.

gassmoeller (approved):
Thanks!

PR #4314, opened 2021-08-17
Use BoundingBox in the generation of particles.

bangerth:
While there, fix problems with an error message: It doesn't actually compute
what we think it computes, and then also outputs the inverse of it.

/rebuild


gassmoeller (approved):
Looks good, thanks for the fix :+1:

PR #4323, opened 2021-08-24
Added a limiter to the linear least squares particle interpolation algorithm

sac-bsa:
Worked with @egpuckett to add a limiter that keeps the interpolated values of particle properties within the bounds of particles within the cell. It has the ability to turn the limiter on and off for particle properties as desired. 
I am not sure if we need to move the string_to_bool function to another file (or maybe an equivalent already exists somewhere and I missed it). 
Pulling the pull request would make the interpolation space P1 instead of Q1. We were looking at having it be a separate file (leaving the Q1 alone) but @gassmoeller had suggested that we update the existing linear least squares limiter although I am not sure that we had made this change clear to him then. This change led to all of the current bilinear tests needing to be updated. 

### For all pull requests:


sac-bsa:
@gassmoeller, Regarding your broad comment 2 above `Is this limiter conservative? I.e. do you conserve the global mass when limiting?`
@egpuckett and I were wondering what do you mean by the global mass? For example, in a cell when there are n particles each of which has a value of the mass density on it, is it the average values of the mass density on the n particles?

egpuckett:
@sac-bsa @gassmoeller is all we are waiting for is Timo's tester before Rene can do a (hopefully) final review and hopefully merge?

sac-bsa:
@egpuckett, Yes now that it has passed the tests. It should be ready for a final review. The remaining tester is not expected to have run yet.

egpuckett:
"The remaining tester is not expected to have run yet."

Why is that?  Should I nudge Rene?

include/aspect/particle/interpolator/bilinear_least_squares.h
```diff
@@ -71,17 +71,11 @@ namespace aspect
...
            */
-          bool use_global_min_max_limiter;
 
-          /**
-           * For each interpolated particle property, a global max and global
-           * min are stored as elements of vectors.
-           */
-          std::vector<double> global_maximum_particle_properties;
-          std::vector<double> global_minimum_particle_properties;
+          ComponentMask use_linear_least_squares_limiter;
```
tjhei, line 78: Can you please add a comment for this variable? I assume the mask is set for each component where the limiter is enabled?

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -95,93 +101,177 @@ namespace aspect
...
         std::vector<Vector<double>> b(n_particle_properties, Vector<double>(n_particles));
-        std::vector<Vector<double>> QTb(n_particle_properties, Vector<double>(n_matrix_columns));
-        std::vector<Vector<double>> c(n_particle_properties, Vector<double>(n_matrix_columns));
-        for (unsigned int property_index = 0; property_index < n_particle_properties; ++property_index)
-          if (selected_properties[property_index])
-            b[property_index] = 0;
 
-        unsigned int particle_index = 0;
-        const double cell_diameter = found_cell->diameter();
+        AssertThrow(n_particles >= n_matrix_columns,
```
tjhei, line 109: Move this assert further up. Also, why do you assert instead of using the fallback interpolator?

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -95,93 +101,177 @@ namespace aspect
...
-                A[6][particle_index] = relative_particle_position[1] * relative_particle_position[2];
-                A[7][particle_index] = relative_particle_position[0] * relative_particle_position[1] * relative_particle_position[2];
+                relative_particle_position[i - 1] -= unit_offset;
+                A[i][positions_index] = relative_particle_position[i - 1];
               }
           }
 
-        for (unsigned int column_index = 0; column_index < n_matrix_columns; ++column_index)
-          qr.append_column(A[column_index]);
+        ImplicitQR<Vector<double>>qr;
```
tjhei, line 144:
```suggestion
        ImplicitQR<Vector<double>> qr;
```

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -95,93 +101,177 @@ namespace aspect
...
+                        double init = interpolated_value;
+                        interpolated_value = std::min(interpolated_value, property_bounds[property_index].second);
+                        interpolated_value = std::max(interpolated_value, property_bounds[property_index].first);
+                        // Due to floating point inaccuracies init and interpolated_value can differ resulting in
+                        // an overshoot or undershoot of around 1e-16. We resolve these small overshoot/undershoots by chopping
+                        // and ensuring that we chopped no more than a 1e-14th of the value.
+                        if (std::abs(init - interpolated_value) > 1e-14 * std::abs(init))
+                          {
+                            std::cout.precision(17);
+                            std::cout << std::fixed << "init: " << init << " inter: " << interpolated_value << " " << std::abs(init-interpolated_value) << std::endl;
```
tjhei, line 268: This seems to be some debug code. Can it be removed?

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -201,24 +291,14 @@ namespace aspect
...
-                                   "limiter for the bilinear least squares interpolation. The number of the input "
-                                   "'Global particle property minimum' values separated by ',' has to be "
-                                   "the same as the number of particle properties.");
-                prm.declare_entry("Use limiter", "false",
-                                  Patterns::Bool (),
-                                  "Whether to apply a global particle property limiting scheme to the interpolated "
-                                  "particle properties.");
+                prm.declare_entry("Use linear least squares limiter", "false",
+                                  Patterns::List(Patterns::Bool()),
+                                  "Limit the interpolation of all particle properties "
```
tjhei, line 296: The description should reflect that this is now a list.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -290,10 +368,13 @@ namespace aspect
     {
       ASPECT_REGISTER_PARTICLE_INTERPOLATOR(BilinearLeastSquares,
                                             "bilinear least squares",
-                                            "Interpolates particle properties onto a vector of points using a "
-                                            "bilinear least squares method. "
-                                            "Note that deal.II must be configured with BLAS and LAPACK to "
```
tjhei, line 295: Does this note still apply? If yes, then keep it please.

tjhei (commented):
I didn't read through the whole code, but I have a few comments already. I think it would be great to get @bangerth to have a look as well.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -95,93 +101,172 @@ namespace aspect
...
+                        current_total_slope = 0;
+                        for (unsigned int i = 1; i < n_matrix_columns; ++i)
+                          {
+                            c[property_index][i] = std::copysign(std::min(std::abs(c[property_index][i]), max_total_slope / half_h), c[property_index][i]);
+                            current_total_slope += std::abs(c[property_index][i]);
+                          }
+                        current_total_slope *= half_h;
+                        if (current_total_slope > max_total_slope)
+                          {
+                            double change_in_slope = (current_total_slope - max_total_slope)/(dim * half_h);
```
gassmoeller, line 197: My question here is: Why do you add a constant amount of slope to all components? This could push some of the components to having a different sign, which creates the problems you try to solve below  (lines 208 ff.). Would it not make more sense to compute the ratio between current_total_slope and max_total_slope, and then divide each `c` component by this ratio? This should also guarantee that you end up with at most `max_total_slope`, but it will apply the majority of the change to the largest components of c (i.e. the components that cause the overshoot in the first place), and it can never change the sign of the components.

I think the case that makes me uncomfortable with your current algorithm is the following:
Assume the properties you use as an example below ([.5, 1, .04, .01]). The large values of c[1] and c[2] cause the overshoot. However, your correction will be equally applied to c[1], c[2], *and* c[3]. In other words the linear plane is mostly tilted in x (most) and y (a little) direction, but after the correction the plane is also significantly tilted in z-direction. This seems to me like a bad property of the algorithm. If instead the amount of correction would scale with the size of c then we would not see properties developing large gradients in directions they had no gradient before.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -95,93 +101,172 @@ namespace aspect
...
+#endif
+                        // Due to floating point inaccuracies init and interpolated_value can differ resulting in
+                        // an overshoot or undershoot of around machine error. We resolve these small overshoot/undershoots by chopping.
+                        interpolated_value = std::min(interpolated_value, property_bounds[property_index].second);
+                        interpolated_value = std::max(interpolated_value, property_bounds[property_index].first);
+                        // If the limiter is working correctly, we should not be chopping anything more than machine error,
+                        // So the check ensures that we have chopped no more than a 1e-12th of the initial value.
+                        // To prevent division by zero, but still provide some amount of scale to our error checking,
+                        // we divide by the larger of 1 and the interpolated value
+                        Assert(std::abs(init - interpolated_value)/std::max(std::abs(interpolated_value), 1.0) <= 1e-12, ExcInternalError());
```
gassmoeller, line 265: I do not understand where the 1.0 comes from, does that assume the value is bounded by 1.0? I understand your concern about floating point precision, but if you write the assert in the following way you can save all the calculations above:
```suggestion
                        Assert(interpolated_value < property_bounds[property_index].second * (1.0 + std::numeric_limits<double>::epsilon()), ExcInternalError());
                        Assert(interpolated_value > property_bounds[property_index].first * (1.0 - std::numeric_limits<double>::epsilon()), ExcInternalError());
```

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -201,24 +286,15 @@ namespace aspect
...
-                                  Patterns::Bool (),
-                                  "Whether to apply a global particle property limiting scheme to the interpolated "
-                                  "particle properties.");
+                prm.declare_entry("Use linear least squares limiter", "false",
+                                  Patterns::List(Patterns::Bool()),
+                                  "Limit the interpolation of particle properties "
+                                  "onto the cell so the value of each property is no "
+                                  "smaller than its minimum and no larger than its "
+                                  "maximum on the particles in each cell. Currently "
+                                  "doesn't work on spherical grids. If more than one "
```
gassmoeller, line 295: You can remove the limitation of the spherical grids now. Since now all the operations you do are in the reference coordinate system instead of real space it should work independently of the geometry model.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -243,29 +319,26 @@ namespace aspect
...
-                                           "is equivalent to the number of particle properties."));
-
-                    // Check that if a global limiter is used, we were given the minimum and maximum value for each
-                    // particle property that is not an internal property.
-                    AssertThrow(global_maximum_particle_properties.size() == n_property_components - n_internal_components,
-                                ExcMessage("Make sure that the size of list 'Global maximum particle property' "
-                                           "is equivalent to the number of particle properties."));
+                    std::vector<bool> parsed(n_property_components, false);
+                    for (unsigned int i = 0; i < split.size(); i++)
+                      parsed[i] = string_to_bool(split[i]);
```
gassmoeller, line 336: You can use modern c++ feature to make this a bit easier:
```suggestion
                    std::vector<bool> parsed;
                    for (const auto &component: split)
                      parsed.push_back(string_to_bool(component));
```

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -290,9 +364,14 @@ namespace aspect
...
       ASPECT_REGISTER_PARTICLE_INTERPOLATOR(BilinearLeastSquares,
                                             "bilinear least squares",
-                                            "Interpolates particle properties onto a vector of points using a "
-                                            "bilinear least squares method. "
-                                            "Note that deal.II must be configured with BLAS and LAPACK to "
+                                            "Uses linear least squares to obtain the slopes and center of a 2D or "
+                                            "3D plane from the particle positions and a particular property value "
+                                            "on those particles. "
+                                            "Interpolate this property onto the support points or to initiate the "
+                                            "property value on a new particle. If the limiter is enabled then it "
```
gassmoeller, line 371: Avoid descriptions of how this is used at the moment, in the future this might be changed. This function only knows that it interpolates properties onto new locations, what these locations are (integration points, new particles, postprocessing locations) are not known to the plugin.
```suggestion
                                            "Interpolate this property onto a vector of points. "
                                            "If the limiter is enabled then it "
```

gassmoeller (commented):
The code is in much better shape now. I still have some comments about parts of the code that are hard to read and understand. Also I have on fundamental question about the workings of the algorithm that I could only formulate now that the code is easier to understand. Let me know what you think about it, maybe I misunderstood something.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -95,93 +101,172 @@ namespace aspect
...
+                              {
+                                c[property_index] = c_i;
+                              }
+                            else
+                              {
+                                // In two dimensions half the neccessary change can be safely removed from each slope component
+                                // however in three dimensions this additional check is neccessary to ensure that a third of the neccesary slope can be removed from each component.
+                                // An example where this could happen is if c is [.5, 1, .04, .01]' with boundaries of [0, 1].
+                                // Without these checks, c_3 would become negative, and would result in a different corner of the cell over(under)shooting.
+                                if (c[property_index][1] * c_i[1] <= 0)
```
gassmoeller, line 214: One question to also discuss when we meet to talk about this PR (just a note to myself to not forget about this thought): Is it possible that more than one component flip their sign? At the moment you only check if one of them changes.

gassmoeller (commented):
Not sure what is going on with your failing test, but the new code looks good at first sight. Please incorporate my one suggestion, otherwise you may have problems in cells in which a property is constant.

source/particle/interpolator/bilinear_least_squares.cc
```diff
@@ -95,93 +101,115 @@ namespace aspect
...
+                    for (unsigned int i = 1; i < n_matrix_columns; ++i)
+                      {
+                        relative_support_point_location[i - 1] -= unit_offset;
+                        interpolated_value += c[property_index][i] * relative_support_point_location[i - 1];
+                      }
+                    if (use_linear_least_squares_limiter[property_index] == true)
+                      {
+                        // If the limiter is working correctly, we should not be significantly more than machine error outside of the range of the property bounds
+                        Assert((property_minimums[property_index] - interpolated_value) <= (property_maximums[property_index] - property_minimums[property_index]) * 16 * std::numeric_limits<double>::epsilon(), ExcInternalError());
+                        Assert((interpolated_value - property_maximums[property_index]) <= (property_maximums[property_index] - property_minimums[property_index]) * 16 * std::numeric_limits<double>::epsilon(), ExcInternalError());
```
gassmoeller, line 208: This is hard to read. What was wrong with:
```suggestion
                        Assert(interpolated_value >= property_minimums[property_index] - std::abs(property_minimums[property_index]) * 10. * std::numeric_limits<double>::epsilon(), ExcInternalError());
                        Assert(interpolated_value <= property_maximums[property_index] + std::abs(property_maximums[property_index])) * 10. * std::numeric_limits<double>::epsilon(), ExcInternalError());
```

source/particle/interpolator/bilinear_least_squares.cc
gassmoeller: This comment is unresolved? Remove the sentence about spherical grids.

gassmoeller (commented):
Nearly ready except for my two comments. Please leave a comment when you want me to review again.

source/particle/interpolator/bilinear_least_squares.cc
sac-bsa: The problem with that is the error in interpolated value is going to be impacted by the largest value, not just the closest property bound. If for some reason we had a cell that had a range of [1, 1e15], we would expect that the error would be closer to epsilon * 1e15 than epsilon * 1. I have now implemented a slight modification on your suggestion though and I believe it is ready for review.

doc/modules/changes/20220208_sacbsa
```diff
@@ -0,0 +1,4 @@
+Added: A bounds perserving slope limiter for the bilinear least
```
anne-glerum, line 1:
```suggestion
Added: A bounds preserving slope limiter for the bilinear least
```

anne-glerum: Could you please fix the typo? Excited to see this merged!

doc/modules/changes/20220208_sacbsa
```diff
@@ -0,0 +1,4 @@
+Added: A bounds preserving slope limiter for the bilinear least
+squares particle interpolator.
+<br>
+(Mack Gregory, 2022/02/08)
```
gassmoeller, line 4: Please change to something like the following (properly broken into several lines, I cant do that easily here). The purpose of these message is to tell users about new features, but also about things that have changed so that they can adjust their parameter files if necessary.

```suggestion
Changed: The global limiter for the least squares particle interpolation plugin was replaced by a bounds preserving slope limiter that respects local bounds on each cell. The global limiter options have been removed from the parameter file options.
<br>
(Mack Gregory, Gerry Puckett, 2022/02/08)
```

After this this is ready to merge.

gassmoeller (approved):

PR #4329, opened 2021-08-27
Update the pressure scaling once per nonlinear iteration

gassmoeller:
Move the pressure scaling computation so that it is updated once per nonlinear iteration instead of the beginning of the timestep. We have a model that relies on the solution of the advection equations before it can reliably compute a viscosity. I timed the computation and it seems significantly faster than the Stokes solution even for simple models.

I expect this to potentially change some tests, although it should only improve convergence of the Stokes solver.

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -568,21 +568,6 @@ namespace aspect
     if (particle_world.get() != nullptr)
       particle_world->backup_particles();
 
-    // Re-compute the pressure scaling factor. In some sense, it would be nice
-    // if we did this not just once per time step, but once for each solve --
-    // i.e., multiple times per time step if we iterate out the nonlinearity
-    // during a Newton or Picard iteration. But that's more work,
-    // and the function would need to be called in more different places.
-    // Unless we have evidence that that's necessary, let's assume that
-    // the reference viscosity does not change too much between nonlinear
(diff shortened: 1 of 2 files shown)


source/simulator/core.cc
```diff
@@ -568,21 +568,6 @@ namespace aspect
...
-    // the reference viscosity does not change too much between nonlinear
-    // iterations and that it's ok to update it only once per time step.
-    //
-    // The call to this function must precede the one to the computation
-    // of constraints because some kinds of constraints require scaling
-    // pressure degrees of freedom to a size adjusted by the pressure_scaling
-    // factor.
-    pressure_scaling = compute_pressure_scaling_factor();
-
     // then interpolate the current boundary velocities. copy constraints
```
tjhei, line 571: I don't understand how this works, you need to update the factor before setting the constraints right below or not? Based on this comment it seems like you always have to redo the constraints after changing the pressure scaling, which you do not. Or am I missing something?

gassmoeller:
Ok, after discussing this offline we came to the conclusion this case is too rare to pose a problem. We fixed our particular model by returning a more reasonable viscosity in the first nonlinear iteration (even though we compute a different one in the later nonlinear iterations. I will close this PR for now.

PR #4331, opened 2021-08-27
Print cheap stokes iterations immediately

gassmoeller:
I was chasing a bug for quite some time until I realized it only happens in the expensive solver. It would be nice to get this indicated immediately, by printing the number of cheap Stokes solver steps as soon as we know it (to separate crashes in the cheap solver from crashes in the expensive solver). Related to #4330.

This should not change any test output.

### DIFF
--- a/source/simulator/solver.cc
+++ b/source/simulator/solver.cc
@@ -837,6 +837,13 @@ namespace aspect
                           distributed_stokes_rhs,
                           preconditioner_cheap);
 
+            // Success. Print all iterations to screen (0 expensive iterations).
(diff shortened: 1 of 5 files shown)


tjhei (approved):
This is very useful! 👍 
You need to delete the old printing in matrix-free...

I regret the code duplication I introduced with Conrad ... 😢

gassmoeller (commented):
Fixed, I had to adjust the additional output slightly to make it look pretty. Take a look if you think this is ok.

tjhei (approved):
looks good.

PR #4334, opened 2021-08-30
use SimulatorAccess in mesh deformation

tjhei:
This is the only place I could change to using SimulatorAccess. All
other occurrences require non-const access.


tjhei:
@gassmoeller FYI

gassmoeller:
Huh, too bad. Even this does not work, because in the constructor the pointer to simulator is not yet initialized (`initialize_simulator` has not yet been called). Maybe we need to postpone the issue.

tjhei:
Well, that is annoying.

gassmoeller:
Yeah, sorry for suggesting it without checking. I appreciate the try though :+1:

PR #4335, opened 2021-08-30
Unify failing solver output.

gassmoeller:
Further improvements to the output of failing solvers. This unifies some code, and in particular makes sure that information like the initial residual / final residual and target residual is also output by the GMG solver and the preconditioners. Currently the output was slightly different for advection solver, stokes solver, gmg, and preconditioners.


gassmoeller:
The one significant change of this is that if a preconditioner fails, it's message will be embedded in the message of the solver. Not sure if this is what we want, but I think it makes some sense.

tests/stokes_solver_fail_gmg/screen-output
```diff
@@ -0,0 +1,72 @@
...
+    The final residual is: 1.375952e-01
+    The required residual for convergence is: 1.169088e-21
+    See output-stokes_solver_fail_gmg/solver_history.txt for the full
+    convergence history.
+    
+    The solver reported the following error:
+    
+    --------------------------------------------------------
+    An error occurred in line <1285> of file
+    </opt/deal.II-v9.3.1/include/deal.II/lac/solver_gmres.h> in function
```
tjhei, line 36: Having the path to deal.II here is inconvenient. Do you know why the path is not filtered?

tjhei (approved):
Very nice cleanup! 👍

gassmoeller:
This should do it. /tests/cmake/default was only looking for `An error occured in` at the beginning of output lines, but not if they were indented. Should be fixed now.

tjhei (approved):

PR #4336, opened 2021-08-31
Update strain rheology

naliboff:
This PR involves changes to the internal logic and structure of the strain rheology module, and also add a new test. It is directly related to the discussion and changes in PRs #4236 (open) and #4271 (closed). This was originally intended to be just an update to #4271, after discussion with @erinheilman it turns out some code restructuring was needed.

In the first commit the following changes are applied to `strain_rheology`:
1. Checks for the type of active deformation (yielding or not) is removed from the reaction term updates
2. Specific variables for each type of strain (plastic, viscous, total) are calculated using the aforementioned checks for yielding.
3. Temperature-dependent strain healing is applied independent of the deformation state, which is the correct thing to do (e.g., follows previous studies). The new code structure will allow different strain healing mechanisms, when added, to use a different approach if warranted.

The new test provides a straightforward way to check wether strain healing is working as intended, with the output compared against an analytical solution. Given the number of changes and new name, I thought it would be easier to add a new test rather than rename the current ones (they can be deleted later).

At this point a review would be helpful to see if everyone agrees with the new code and test structure. If the group approves of these changes, I will add additional tests.


source/material_model/rheology/strain_dependent.cc
```diff
@@ -501,33 +501,52 @@ namespace aspect
...
+            if (plastic_yielding == true)
+              delta_e_ii_plastic = delta_e_ii;
+
+            // Viscous strain
+            double delta_e_ii_viscous = 0.;
+            if (plastic_yielding == false)
+              delta_e_ii_viscous = delta_e_ii;
+
+            // Total strain
+            double delta_e_ii_total = delta_e_ii;
```
gassmoeller, line 517: Just remove this line and use delta_e_ii for the total strain.

source/material_model/rheology/strain_dependent.cc
```diff
@@ -501,33 +501,52 @@ namespace aspect
...
+
+            // Now account for strain healing
+            if (healing_mechanism == temperature_dependent)
```
gassmoeller, line 520: Keep the if-condition, otherwise other healing mechanisms in the future will not work.
```suggestion
            if (healing_mechanism != no_healing)
```

gassmoeller (commented):
Thanks for the test :+1: this proves the strain healing itself is correct except for cases where the healing in one timestep is about the same size as the remaining strain. I left a note about what I think is the problem with the last timestep, and also a possible solution. The solution is not super pretty though, maybe you have another idea?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -501,33 +501,52 @@ namespace aspect
...
-            if (weakening_mechanism == viscous_weakening_with_viscous_strain_only && plastic_yielding == false)
+                std::max(delta_e_ii_plastic, -in.composition[i][this->introspection().compositional_index_for_name("plastic_strain")]);
+            if (weakening_mechanism == viscous_weakening_with_viscous_strain_only)
               out.reaction_terms[i][this->introspection().compositional_index_for_name("viscous_strain")] =
-                std::max(delta_e_ii, -in.composition[i][this->introspection().compositional_index_for_name("viscous_strain")]);
+                std::max(delta_e_ii_viscous, -in.composition[i][this->introspection().compositional_index_for_name("viscous_strain")]);
             if (weakening_mechanism == total_strain || weakening_mechanism == plastic_weakening_with_total_strain_only)
               out.reaction_terms[i][this->introspection().compositional_index_for_name("total_strain")] =
-                std::max(delta_e_ii, -in.composition[i][this->introspection().compositional_index_for_name("total_strain")]);
+                std::max(delta_e_ii_total, -in.composition[i][this->introspection().compositional_index_for_name("total_strain")]);
```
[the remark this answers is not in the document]
anne-glerum: This is related to #4198, but there I do store the extrapolated value at the beginning of the timestep. The comment in melt_global.cc says that the old solution is used because it needs a field that is not updated in nonlinear iterations, which is a different problem from what @gassmoeller  described here (although it has the same solution). Anyway, the Utilities function evaluate_advection_solution  that I introduce in #4198 could be used here as well.

naliboff:
@gassmoeller @anne-glerum - Thanks for the comment and suggestions. I've updated the PR to use the composition derived from the old solution vector, and now the predicted and analytical solutions match on all time steps. 

@gassmoeller - I couldn't think of any other solutions than the one you posed, and the amount of additional code is not cumbersome.

We'll see how many tests this breaks, potentially quite a few.

naliboff:
@anne-glerum @gassmoeller - the second to last commit updated all of the tests, and as expected a number of the results changed. However, most represent small changes in either the solution or compositional field values, which was expected with the update. Shall we discuss the tests results further before proceeding with a merge?

gassmoeller (approved):
This looks good to me. There is some potential for performance optimizations, but I will open another PR when this one is merged. Let's wait for the testers and then merge.

PR #4338, opened 2021-09-02
remove lines with code line number from screen-output

bobmyhill:
Several of the tests which purposefully fail report line numbers of errors in the screen-output. This PR replaces those lines so that the test outputs don't need to be changed so often.

### DIFF
--- a/tests/cmake/default
+++ b/tests/cmake/default
@@ -19,8 +19,6 @@ while ( <STDIN> )
   next if $_ =~ m/^-- /;
   next if $_ =~ m/^\|/;
 
-  $line = $_;
-
   if ($skip>0)
   {
     print "(line in output replaced by default.sh script)\n";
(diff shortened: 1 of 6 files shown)


tests/cmake/default
```diff
@@ -31,12 +33,17 @@ while ( <STDIN> )
 	$skip=1;
     }
 
+    if ($_ =~ m/^\s*An error occurred in line/)
+    {
+  $line = "(line in output replaced by default.sh script)\n";
```
tjhei, line 38: Indentation seems to be weird

bobmyhill: The indentation of the whole file was weird. Now fixed. I don't understand why the tests failed (the script works fine on my machine). Let's see if they pass now.

tjhei:
looks good, but you will need to update the test results from a couple other tests.

bobmyhill:
Ok, this should now work. I've restored some of the old deleted lines, as I think the original intention was to only replace lines that printed out relative file locations, so the stored files now retain more information.

tests/composition_names_duplicates/screen-output
```diff
@@ -15,7 +15,7 @@ Number of degrees of freedom: 49 (18+4+9+9+9)
 An exception happened on MPI process <0> while running the visualization postprocessor <N6aspect11Postprocess27VisualizationPostprocessors18MaterialPropertiesILi2EEE>: 
 
 An error occurred in file <visualization.cc> in function
-(line in output replaced by default.sh script)
+    void aspect::Postprocess::{anonymous}::add_data_names_to_set(const std::vector<std::__cxx11::basic_string<char> >&, std::set<std::__cxx11::basic_string<char> >&)
```
tjhei, line 18: the problem is that the arguments are formatted slightly differently if you run with a different compiler (for example ``__cxx11::basic_string here``).

bobmyhill: Oh darn. Ok, I've re-removed those lines and squashed the last two commits.

tjhei:
Can you rebase, please?

Issue #4341, opened 2021-09-03
Adding a free surface on a restart causes segmentation fault


erinheilman:
When adding a free surface to a 2D spherical shell model that has already computed timesteps, and then restarting I get a segmentation fault. I ran on debug to see why it was failing and got this error message. 
<img width="1773" alt="Screen Shot 2021-09-03 at 12 11 08 PM" src="https://user-images.githubusercontent.com/69014916/132043528-305127ef-1a5f-4386-865c-b7d20d26a342.png">
I've included the prm file that is updated with the free surface parameters. Is this an issue with my parameter file or something that shouldn't be done?
[2D_full_depth_dep_ys_nosw_nosh_free_surface.txt](https://github.com/geodynamics/aspect/files/7107583/2D_full_depth_dep_ys_nosw_nosh_free_surface.txt)

gassmoeller:
This is something we do not generally support. Changing model parameters between restarts is not guaranteed to work (although it does work for many parameters). However, for free surface models to be restarted the model requires the amount of accumulated deformation for each mesh vertex which of course was not saved during the restart before. With some effort there could be hack to support this (detecting if no deformation was stored and then initializing it with 0 deformation), but that will still come with some caveats such as that we use a different finite element mapping for models with mesh deformation, which will change the solution slightly and potentially cause problems for the solver to converge. 

Here is what you could try instead:

Start the model with mesh deformation already enabled, but set the boundary deformation model to `function` instead of `free surface` and the velocity boundary condition to tangential instead of empty. The default values for function are to not prescribe any deformation, so you will have the whole mesh deformation machinery activated, but the boundary should not actually deform and there should be no flow through the boundary. Then during the restart replace the function model by the free surface model and remove the boundary condition for velocity. This may still provide some convergence problems for the solver in the step after the restart, but it may just work (make sure to think of a test case, this solution is experimental).

erinheilman:
Thanks Rene, I'll try it on a simple model and see if it works.

elodie-kendall:
Hi Erin, Rene, I did already try this a few months ago- https://github.com/geodynamics/aspect/issues/4029. However, it didn't work because: "The pressure normalization method that was stored in the checkpoint file is not the same as the one you currently set in your input file. These need to be the same during restarting from a checkpoint."

erinheilman:
Yes, I got that message too. I got it to work by setting pressure normalization to no from the beginning of the model, then when switching to free surface it is satisfied

[...]

Issue #4029, opened 2021-05-11
seg fault on 2D shell restart with different velocity boundary condition


elodie-kendall:
Hi,

I'm struggling to get my models to converge with a free surface and it would take to long to go to the small time steps needed for convergence and so I would like to run my 2D viscoplastic shell models with free slip and then restart with a free surface when equilibrium is reached. I can restart a model with free slip or restart a model with a free surface however I cannot switch from free slip to the free surface. My error is not the same as Max's though. It is a segmentation fault: 
 29 [node031:140850] Signal: Segmentation fault (11)
 30 [node031:140850] Signal code: Address not mapped (1)

and in debug mode:
free(): invalid pointer: 0x000000000425f050 ***
munmap_chunk(): invalid pointer: 0x0000000003c0cf70 ***

Do you think you could have a look at this please? I attach a simple test that fails below. Thanks a lot.

``
set Additional shared libraries 		= /home/fregehrk/opt/asLib/reaction_depth_2_way_v2.1/build_release/libcomposition_reaction.so

####  Global parameters
set Dimension                              = 2
set Use years in output instead of seconds = true
set Start time                             = 0
set End time                               = 2.e6  
set Output directory                       = output
set Nonlinear solver scheme                = iterated Stokes  #single Advection, iterated Stokes
set Nonlinear solver tolerance             = 5e-4  # -4 -> -5
set Max nonlinear iterations               = 150
set Max nonlinear iterations in pre-refinement = 20
set Pressure normalization                 = no
set Maximum time step                      = 1e6
#set Number of cheap Stokes solver steps    = 200  # aspect 2.2
set Adiabatic surface temperature          = 1600  # need 4 initial adiabatic temp.
set CFL number                             = 0.5

set Resume computation          = auto 
subsection Checkpointing
  set Steps between checkpoint  = 100
  set Time between checkpoint   = 0
end
subsection Termination criteria
  set Checkpoint on termination           = true
  set End step                            = 1
  set Termination criteria                = end time #end step
end

subsection Solver parameters
  subsection Stokes solver parameters
    #set GMRES solver restart length = 50
    set Linear solver tolerance             = 1e-5
    set Maximum number of expensive Stokes solver steps = 1000
    set Number of cheap Stokes solver steps    = 200  # not using too much, only get a guess
  end
end

#### Model Geometry
subsection Geometry model
  set Model name = spherical shell

  subsection Spherical shell
    set Inner radius  = 3485e3
[...]

PR #4345, opened 2021-09-08
Reorder restart checks

gassmoeller:
This is an improvement related to #4341 and #4029. It reorders the operations upon resuming from a checkpoint so that:

- We correctly reach a useful assertion before trying to load the solution vector. This prevents the segmentation fault described in the linked issues and instead provides a useful error message (pressure normalization was changed incorrectly, or mesh deformation was not enabled for checkpoint)
- We now allow disabling pressure normalization during restart. We still do not allow to switch from surface to volume normalization or vice-versa since that makes no sense, however #4341 and #4029 showed that there is demand for switching on mesh deformation during a model for which we would like to be able to switch from surface pressure normalization to no normalization.

### DIFF
--- a/source/simulator/checkpoint_restart.cc
+++ b/source/simulator/checkpoint_restart.cc
@@ -209,15 +209,20 @@ namespace aspect
       // the change would then lead to a jump in pressure from one time step
       // to the next when we, for example, change from requiring the *surface*
       // average to be zero, to requiring the *domain* average to be zero.
-      // That's unlikely what the user really wanted.
+      // That's unlikely what the user really wanted. However, we do allow
+      // disabling the pressure normalization in case the user wants to enable
+      // a free surface boundary.
       std::string pressure_normalization;
       ia >> pressure_normalization;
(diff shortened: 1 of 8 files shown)


gassmoeller:
@erinheilman or @elodie-kendall: Could you take a look at my changes in this PR and let me know if this would help you?

tjhei:
> * We now allow disabling pressure normalization during restart. We still do not allow to switch from surface to volume normalization or vice-versa since that makes no sense,

Does it? We don't use the pressure from a previous time step, so it might be possible to change it freely. Did you try?

gassmoeller:
> Does it? We don't use the pressure from a previous time step, so it might be possible to change it freely. Did you try?

I did not try and it might work, but I guess what I was saying is I cannot imagine an application case where it would be useful to do that. Doesn't mean there is no case out there.

elodie-kendall:
Yes, @erinheilman says this has allowed a change in the P normalisation when switching the boundary condition. Thanks a lot!

PR #4346, opened 2021-09-08
reorganize nullspace

tjhei:
- No functional changes!
- introduce nullspace enums to select more than one option
- remove early return in setup_nullspace_constraints
- improve readability in remove_nullspace


gassmoeller (approved):
Looks good and most of the changes are only whitespace anyway.
:+1: once the tester is done

PR #3317, opened 2019-12-17
Add surface topography benchmark

anne-glerum:
This PR adds the benchmark from Ian Rose's paper. I used it to look at predicted surface topography for the different velocity projections onto the surface, but it of course also nicely demonstrates the drunken sailor effect.

* [x] I have followed the [instructions for indenting my code](../CONTRIBUTING.md#making-aspect-better).

### For new features/models or changes of existing features:

* [x] I have tested my new feature locally to ensure it is correct.
* [x] I have added a changelog entry in the [doc/modules/changes](../doc/modules/changes) directory that will inform other users of my change.

PR #4350, opened 2021-09-10
Apply free surface stabilization to other mesh deformation plugins

Djneu:
At the moment, the free surface stabilization can only be applied to the free surface plugin. However, since multiple plugins can deform the surface, it may help to be able to apply the stabilization to any one plugin. For example, when using FastScape to deform the surface in 2D the model gets stability issues at the same timesteps as the free surface in the benchmark from #3317. Applying the stabilization similarly fixes the issue.

Questions:
1. Does it make sense to apply the same stabilization parameter to different problems other than those with a free surface?

2. What would be the best way to implement this? In this pull request, I've transferred the functionality of the stabilization into the mesh deformation handler. This should apply it to all mesh deformation boundaries unless it is specifically set to 0.  @anne-glerum has suggested that we could also make it a separate mesh deformation plugin to be called in conjunction with the other plugins.

* [ x ] I have read the guidelines in our [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) document.

### For all pull requests:

* [ x ] I have followed the [instructions for indenting my code](../blob/master/CONTRIBUTING.md#making-aspect-better).

* [ x ] I have tested my new feature locally to ensure it is correct.

### DIFF
--- a/include/aspect/mesh_deformation/interface.h
+++ b/include/aspect/mesh_deformation/interface.h
@@ -34,11 +34,41 @@
 #include <deal.II/lac/la_parallel_vector.h>
 #include <deal.II/multigrid/mg_constrained_dofs.h>
 #include <deal.II/multigrid/mg_transfer_matrix_free.h>
+#include <aspect/simulator/assemblers/interface.h>
 
 
 namespace aspect
 {
   using namespace dealii;
+
+  namespace Assemblers
+  {
+    /**
+     * Apply stabilization to a cell of the system matrix. The
+     * stabilization is only added to cells on a free surface. The
+     * scheme is based on that of Kaus et. al., 2010. Called during
+     * assembly of the system matrix.
+     */
+    template <int dim>
+    class ApplyStabilization: public Assemblers::Interface<dim>,
+      public SimulatorAccess<dim>
+    {
+      public:
+        ApplyStabilization(const double stabilization_theta);
+
+        void
+        execute (internal::Assembly::Scratch::ScratchBase<dim>   &scratch,
+                 internal::Assembly::CopyData::CopyDataBase<dim> &data) const override;
+
+      private:
+        /**
+         * Stabilization parameter for the free surface. Should be between
+         * zero and one. A value of zero means no stabilization. See Kaus
+         * et. al. 2010 for more details.
+         */
+        const double free_surface_theta;
+    };
+  }
 
   template <int dim> class Simulator;
 
@@ -83,6 +113,13 @@ namespace aspect
          * The default implementation of this function does nothing.
          */
         virtual void update();
+
+
+        /**
+         * A function that will be called to check whether stabilization is needed.
+         */
+        virtual bool needs_surface_stabilization() const;
+
 
         /**
          * A function that returns the initial deformation of points on the
@@ -162,6 +199,13 @@ namespace aspect
         void initialize();
 
         /**
+         * Called by Simulator::set_assemblers() to allow the FreeSurface plugin
+         * to register its assembler.
+         */
+        void set_assemblers(const SimulatorAccess<dim> &simulator_access,
+                            aspect::Assemblers::Manager<dim> &assemblers) const;
+
+        /**
          * Update function of the MeshDeformationHandler. This function
          * allows the individual mesh deformation objects to update.
          */
@@ -238,6 +282,13 @@ namespace aspect
          */
         const std::set<types::boundary_id> &
         get_active_mesh_deformation_boundary_indicators () const;
+
+        /**
+         * Return a set of all the indicators of boundaries that
+         * require surface stabilization.
+         */
+        const std::set<types::boundary_id> &
+        get_boundary_indicators_requiring_stabilization () const;
 
         /**
          * Return the boundary id of the surface that has a free surface
@@ -506,7 +557,20 @@ namespace aspect
          */
         std::set<types::boundary_id> free_surface_boundary_indicators;
 
+        /**
+         * The set of boundary indicators for which the mesh deformation
+         * objects need surface stabilization.
+         */
+        std::set<types::boundary_id> boundary_indicators_requiring_stabilization;
+
         bool include_initial_topography;
+
+        /**
+         * Stabilization parameter for the free surface. Should be between
+         * zero and one. A value of zero means no stabilization.  See Kaus
+         * et. al. 2010 for more details.
+         */
+        double surface_theta;
 
         /**
          * If required, store a mapping for each multigrid level.
@@ -522,6 +586,7 @@ namespace aspect
         * Multigrid transfer operator for the displacements
         */
         MGTransferMatrixFree<dim, double> mg_transfer;
+
 
         /**
         * Multigrid level constraints for the displacements
--- a/include/aspect/mesh_deformation/interface.h
+++ b/include/aspect/mesh_deformation/interface.h
@@ -299,6 +299,11 @@ namespace aspect
         get_free_surface_boundary_indicators () const;
 
         /**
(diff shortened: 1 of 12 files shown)


gassmoeller (commented):
Interesting that you see similar oscillations even when only modifying the boundary according to surface evolution. I guess it makes sense that the solution reacts similarly to boundary deformations from inside (as result of our Stokes solve) or from outside (surface evolution). However, I dont think this should be part of every plugin, as for example the function plugin prescribes mesh motions that are exactly known, no possibility for oscillations there.

I would suggest instead of creating a new plugin (which everyone would have to remember to activate or deactivate) to move the function into the mesh deformation handler like you have done, but give the plugin interface a new function called `bool needs_surface_stabilization()`. This function would be implemented in every plugin and returns if the plugin requires stabilization (we could have a default implementation in the interface that returns false, and only free surface and your fastscape plugin would return true). If any active plugin would return true the mesh deformation handler would then activate the stabilization assembler. Does this sound reasonable to you?

Before moving forward: Please make sure that the derivation of the stabilization term is generic, and not specifically dependent on the free surface use case. I know too little about how the term was derived to know if it makes sense to apply it to other mesh movements as well.

Djneu:
Hi Rene,
Thank you for taking a look through this. I like your suggestion on the implementation, and will try adding that when I get a chance to look at this again. 


On whether the stabilization should apply to fastscape:

In the coupling, instead of using FastScape in conjuction with the free surface, we have Fastscape handle advection and uplift of the surface. If I understand the stabilization correctly, instabilities form because of the large density variations that occur at the surface when it moves, which would normally require small timesteps to solve without issues. The stabilization seems to work by using the stokes velocity to predict what the topography at the end of the timestep will be, it then uses this prediction and the surface density to add a pressure term. As in these cases we have FastScape handling the movement of the surface I think it makes sense we would still need to account for these variations to avoid instabilities. Based on tests we tried from #3317, this fixed the instabilities we were seeing in the 2D drip models. 


dripdepth_vstimestep

Using FastScape solely for surface uplift and advection with no surface processes (in 2D), the surface topography ended up comparable to the free surface with normal projection, though both were a bit different from the free surface with vertical projection (yellow and dashed-purple line in graph).

surface_comparison


One issue I see with this implementation is that the stabilization uses the stokes velocity at the beginning of timestep and not the mesh velocity. Because of this, while the overall movement of the mesh will be accounted for in the prediction, I don't think any changes to the surface due to erosion or deposition are accounted for. In the models we've ran since using the stabilization this has not been a major issue so it seems that the movement of the mesh is probably the greatest factor in developing these instabilities. We tried the drip models using FastScape with surface processes and they remained stable, but I didn't check how much they were actually changing the surface at the time. In recent models with fairly significant erosion and basin formation we've seen some drunken-sailor instability development at timesteps of 5 kyr, though I haven't tested whether the value of the  stabilization theta helps in these cases. 

Does this sound reasonable, or is there anything I'm missing? @anne-glerum @tjhei @naliboff

Djneu:
Hi all, I updated this according to Rene's suggestions. Now the interface has a function "needs_surface_stabilization", which returns a true or false flag and can be overwritten by functions within the plugins. During initialization, this flag is checked and if it is true then the assemblers are set up to allow free surface stabilization. Does this seem like it works for everyone?

source/mesh_deformation/interface.cc
```diff
@@ -38,10 +38,119 @@
...
+
+      // only apply on mesh deformation faces
+      if (cell->at_boundary() && cell->is_locally_owned())
+        for (const unsigned int face_no : cell->face_indices())
+          if (cell->face(face_no)->at_boundary())
+            {
+              const types::boundary_id boundary_indicator
+                = cell->face(face_no)->boundary_id();
+
+              if (tmp_mesh_deformation_boundary_indicators.find(boundary_indicator)
```
tjhei, line 97: I think this needs to be fixed: You now unconditionally apply stabilization to all mesh deformation boundaries even if only some of them needs stabilization. It might make sense to compute the "set of indicators that require stabilization" and use that here.

Djneu:
Thanks for catching that! I've updated it to account for this, and reordered how things are done a bit. Now, I created a new function "get_boundary_indicators_requiring_stabilization()", which will send a variable that holds all boundaries that need to be stabilized. Filling of the variable this returns is now done at the end of parse parameters, and then in initialize it only checks whether or not this is empty, and if it isn't sets up the assemblers. Within the assemblers, we now check the indicators sent from this function and not active_mesh_deformation indicators.

include/aspect/mesh_deformation/interface.h
```diff
@@ -508,7 +559,20 @@ namespace aspect
...
 
+        /**
+         * The set of boundary indicators for which the mesh deformation
+         * objects need surface stabilization.
+         */
+        std::set<types::boundary_id> boundary_indicators_requiring_stabilization;
+
         bool include_initial_topography;
+		
+		 /**
```
tjhei, line 570: please fix indentation here.

source/mesh_deformation/interface.cc
```diff
@@ -38,10 +38,119 @@
...
+      const typename DoFHandler<dim>::active_cell_iterator cell (&this->get_triangulation(),
+                                                                 scratch.finite_element_values.get_cell()->level(),
+                                                                 scratch.finite_element_values.get_cell()->index(),
+                                                                 &this->get_dof_handler());
+
+      const unsigned int n_face_q_points = scratch.face_finite_element_values.n_quadrature_points;
+      const unsigned int stokes_dofs_per_cell = data.local_dof_indices.size();
+
+      // Get the boundary indicators of those boundaries with mesh deformation
+      const std::set<types::boundary_id> tmp_mesh_deformation_boundary_indicators = this->get_mesh_deformation_handler().get_boundary_indicators_requiring_stabilization();
```
tjhei, line 87: I think the variable name is misleading as these are the "ids requiring stabilization".

source/mesh_deformation/interface.cc
```diff
@@ -210,17 +368,31 @@ namespace aspect
                            "\n\n"
                            "The format is id1: object1 \\& object2, id2: object3 \\& object2, where "
                            "objects are one of " + std::get<dim>(registered_plugins).get_description_string());
+        prm.declare_entry("Surface stabilization theta", "0.5",
```
tjhei, line 371: You are moving this parameter into a different subsection, which is not backwards compatible. I think we need to deprecate the old value or keep it in the old section.

Djneu: Okay, so if I understand correctly one option would be to move this back to being declared in free_surface.cc and then have the interface parse into that to get the value. Should I change the variable name back into "Free surface stabilization theta" in that case?

What would be the preferred way, keeping it in the old section or deprecating it? I am not sure the process for deprecating a value.

tjhei: No, you can leave it in the "free surface" subsection by entering the right subsection right here. No need to move this to the other file.

Djneu:
I think this should have fixed the issues you mentioned. I've moved the surface_theta declaration back into the free surface subsection, and updated the apply_stabilization_with_melt to also check using the correct boundary indicators.

tjhei:
you will need to rebase on the newest master before we can merge.

Djneu:
Alright, I've rebased to the main branch and moved the get_free_surface_theta() function into the interface as well.

tjhei:
Alright, looks good to me. /rebuild

Djneu:
I've rebased to the latest version to hopefully fix the merge conflict and updated a few lines to reflect the change from pull request #4725. @tjhei

tjhei (approved):

PR #4353, opened 2021-09-15
Apply strain weakening before elasticity

anne-glerum:
Currently, viscous strain weakening is applied to the visco-elastic viscosity when elasticity is switched on. It should only be applied to the viscous viscosity determined through diffusion and/or dislocation creep. This change will not matter for viscoplastic models, only for vep ones. 

### For all pull requests:


naliboff (commented):
👍 I agree that this is the right approach and the . I don't recall why we (likely I) set the viscous weakening to occur afterwards. 

Do you think we need a test for this? We already have a test for VP viscous strain weakening, and I'm not sure an additional one is needed. Thoughts from others on this?

bobmyhill:
I think if this PR changes/fixes functionality without breaking any tests, we should add a new test that would have been broken before the fix. Sorry for giving you more work, @anne-glerum. Thanks for catching these things!

anne-glerum:
I found two tests with elasticity enabled, and neither of them has viscous strain weakening. So a test with vep + viscous strain weakening is probably a good idea. Maybe I'll convert a vep benchmark to a test.

anne-glerum:
Ok, I've added a test, not sure it's very meaningful, but it should the ordering should matter for the results. Will update the tester when you agree.

naliboff (approved):
@anne-glerum - Thanks for adding the test and this fix!

PR #4354, opened 2021-09-15
Compute elastic force term at t0 as well

anne-glerum:
When the stored stresses at start-up are not zero (i.e. the compositional fields have some initial non-zero value), I believe the force term should be taken into account. The behavior for t0 with zero elastic stress should not change, as the force term is then a multiple of zero.

### For all pull requests:


bobmyhill (approved):
Yes!

bobmyhill:
(I assume that either there are some tests that will need fixing or a new test should be made)

naliboff:
Yep, good catch! 

I agree with @bobmyhill that we should add a test where the initial elastic stress is non-zero. Maybe the viscoelastic stress build-up example could be modified to have the initial stress correspond to the analytical solution at a given point in time? The value at the next step time step should then also match the analytical solution.

anne-glerum:
I'll add a test here as well. The stress build up one is a good idea.

anne-glerum:
Shall I update the new test's results?

naliboff (approved):
@anne-glerum - Thanks for adding the additional test!

naliboff:
@anne-glerum @bobmyhill  - Do you think it would be worth adding a change log entry for this PR? It is a minor change, but it will have a large effect on anyone who plans to run "pre-stressed" VE/VEP models. Otherwise, ready to merge.

bobmyhill:
A change log entry would be a good idea. I hope that the change will not affect many people, but if it affects anyone, then it would be good to highlight that change.

anne-glerum:
Done :)

naliboff (approved):
👍 Thanks @anne-glerum!

PR #4355, opened 2021-09-15
Remove stress averaging parameter (i.e. always average)

anne-glerum:
I believe there are three valid scenarios for the averaging of the old and new elastic stress:
1. use_fixed_elastic_timestep = false, have dte=dt, and use_stress_averaging=false
2. use_fixed_elastic_timestep=false, have stabilization_factor > 1 so that dte is not dt, use_stress_averaging=true
3. use_fixed_elastic_timestep=true,  use_stress_averaging=true

At the moment, the combination the above parameters as set by the user is not checked.
Also, there is only one case where stress averaging does not need to be applied, and when it is, dte=dt -> dte/dt=1 and averaging doesn't change anything.
However, when use_stress_averaging is incorrectly set to false (the default), scenario 2 and 3 are wrong. 

So I removed the parameter use_stress_averaging, and stress averaging is now always applied. 

### For all pull requests:

### DIFF
--- a/benchmarks/free_surface_tractions/viscoelastic/free_surface_VE_cylinder_2D_loading.prm
+++ b/benchmarks/free_surface_tractions/viscoelastic/free_surface_VE_cylinder_2D_loading.prm
@@ -128,7 +128,6 @@ subsection Material model
     set Elastic shear moduli        = 1.e10
     set Use fixed elastic time step = false
     set Fixed elastic time step     = 1e2
-    set Use stress averaging        = false
     set Viscosity averaging scheme  = harmonic
   end
 
--- a/benchmarks/free_surface_tractions/viscoelastic/free_surface_VE_cylinder_2D_loading_fixed_elastic_dt.prm
+++ b/benchmarks/free_surface_tractions/viscoelastic/free_surface_VE_cylinder_2D_loading_fixed_elastic_dt.prm
@@ -16,7 +16,6 @@ subsection Material model
   subsection Viscoelastic
     set Use fixed elastic time step = true
     set Fixed elastic time step     = 1e2
-    set Use stress averaging        = true
     set Viscosity averaging scheme  = harmonic
   end
 
--- a/benchmarks/free_surface_tractions/viscoelastic/free_surface_VE_cylinder_3D_loading.prm
+++ b/benchmarks/free_surface_tractions/viscoelastic/free_surface_VE_cylinder_3D_loading.prm
@@ -129,7 +129,6 @@ subsection Material model
     set Elastic shear moduli        = 1.e10
     set Use fixed elastic time step = false
     set Fixed elastic time step     = 1e2
-    set Use stress averaging        = false
     set Viscosity averaging scheme  = harmonic
   end
 
--- a/benchmarks/free_surface_tractions/viscoelastic/free_surface_VE_cylinder_3D_loading_fixed_elastic_dt.prm
+++ b/benchmarks/free_surface_tractions/viscoelastic/free_surface_VE_cylinder_3D_loading_fixed_elastic_dt.prm
@@ -16,7 +16,6 @@ subsection Material model
   subsection Viscoelastic
     set Use fixed elastic time step = true
     set Fixed elastic time step     = 1e2
-    set Use stress averaging        = true
     set Viscosity averaging scheme  = harmonic
   end
 
--- a/benchmarks/infill_density/infill_ascii.prm
+++ b/benchmarks/infill_density/infill_ascii.prm
@@ -165,7 +165,6 @@ subsection Material model
     set Elastic shear moduli        = 1.e10
     set Use fixed elastic time step = true
     set Fixed elastic time step     = 7500
-    set Use stress averaging        = true
     set Viscosity averaging scheme  = harmonic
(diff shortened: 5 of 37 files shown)


anne-glerum:
Three tests fail in this PR. The first, elastic_stabilization_time_scale_factor, builds on benchmarks/viscoelastic_plate_flexure/viscoelastic_plate_flexure.prm. The benchmark prm had the following settings:
```
set Maximum time step         = 5
      set Fixed elastic time step     = 5
      set Use fixed elastic time step = false
      set Use stress averaging        = false
```
I guess the Maximum timestep is very small, leading to always a timestep of 5. Since the elastic stabilization factor is 1 by default, the elastic timestep is also always 5. Stress averaging could therefore be turned off.
However, in the test elastic_stabilization_time_scale_factor:
```
subsection Material model
  subsection Viscoelastic
    set Stabilization time scale factor = 2 
  end 
end
```
Therefore the elastic timestep no longer equals the Stokes timestep and averaging should have been applied. But it wasn't, and therefore the test is now failing. @naliboff what do you think?

anne-glerum:
Similar issues for the other 2 failing tests (visco_plastic_vep_stress_build-up_fixed_elastic_time_step and viscoelastic_fixed_elastic_time_step), where I think averaging should have been applied.

anne-glerum:
ping @naliboff :)

naliboff:
@anne-glerum - Thanks for submitting this and I agree with the changes. Simplifying things and removing parameters when possible is always good 😄 

For the failing tests, I don't recall why the stabilization factor in the three failing tests was set to 2. This might have been part of the original PR @EstherHeck submitted for this? My preference would be to change these values back to 1 and a new test for the Stabilization time scale factor, if there is not already one. Does this sound like a reasonable approach?

John

EstherHeck:
Hi @naliboff and @anne-glerum , in my initial PR (#3770) I only created a new test: tests/elastic_stabilization_time_scale_factor.prm 
but didn't change any other tests, so I don't know why it is set to 2 there. As they are called *_fixed_elastic_time_step it also wouldnt make sense to apply the elastic time scale factor instead. Writing this I wonder if we have a check that you can only use either a fixed elastic time step OR the stabilization time scale factor?

anne-glerum:
@EstherHeck  thanks for having a look. The last two benchmarks have similar problems (i.e. because they depend on other prms), but not the same problem. Here what I think are the problems:
1. tests/elastic_stabilization_time_scale_factor.prm: uses a stabilization factor of 2
2. tests/visco_plastic_vep_stress_build-up_fixed_elastic_time_step.prm: uses an elastic timestep 2x the maximum computational timestep (2000 vs 1000) set in the base prm
3. tests/viscoelastic_fixed_elastic_time_step.prm: uses an elastic timestep 2x the maximum computational timestep (2000 vs 1000) set in the base prm

Do you agree with these solutions:
1. `set Use stress averaging  = true`
2. `set Use stress averaging  = true` or `set Maximum timestep = 2000` and check that the computational timestep is never below 2000
3. `set Use stress averaging  = true` or `set Maximum timestep = 2000` and check that the computational timestep is never below 2000
Using stress averaging in all 3 cases will be the easiest solution.

EstherHeck:
I would use stress averaging in let the time step sizes as they are. This is based on all those trial models I ran where it was never a good idea to have dt = dte.

anne-glerum:
Funnily, at the moment, we are only sure that dt = dte is correctly implemented 😆

anne-glerum:
@naliboff Shall I make a separate PR that sets the stress averaging to true for those 3 tests?

bangerth:
What should we do about this issue?

naliboff:
@anne-glerum @bangerth - sorry for holding up this PR. 

> Using stress averaging in all 3 cases will be the easiest solution.
@anne-glerum - Yes, let's just go with that approach for the three failing tests. If you don't have time over the next few days,  I'm happy to make a PR updating those tests.

anne-glerum:
@naliboff The averaging was set correctly for those three tests in #4471. I've rebased, let's see if the tester is happy now.

anne-glerum:
Actually, this PR is not backwards compatible as it removes an input parameter. What would be the way to go here? Leave the parameter in but ignore it and give a warning when it is false? @naliboff @gassmoeller

anne-glerum:
The cookbooks break because the parameter is still listed in the prm files. @gassmoeller what should I do about breaking backwards compatibility?

gassmoeller:
If we think the new version is correct and the old was not, let's break compatibility. It forces everyone to switch to the correct version. Make sure this is mentioned in a changelog entry to explain the background.

anne-glerum:
The old version wasn't wrong per se, but it could lead to wrong combinations of parameters (which were not checked). This is what happened to three tests before they were updated in #4471 . To avoid such situations, I suggest in this PR to always use the averaging, because it will have no effect when it's not necessary anyway.

gassmoeller:
Ok then let's move forward with this PR. I think you need to update tests though.

anne-glerum:
Okay then I'll remove the input parameter from all cookbook prms, that should fix the tester.

gassmoeller (approved):

PR #4357, opened 2021-09-17
Move dynamic friction into a friction rheology model

EstherHeck:
As part of the WIP to have rate-and-state friction in aspect, I moved the material model "dynamic friction" into a new rheology model "friction options", where I will add the rate-and-state implementation in a later PR as well.
This PR also includes changes to the output of calculate_isostrain_viscosities to make the strain-rate dependent current friction angle available to the plastic additional outputs. 
I furthermore introduced the "dynamic friction smoothness exponent" to the dynamic friction equation. 

### For all pull requests:


gassmoeller (commented):
Thanks for the cleanup. You will have to make this compile first though. Apart from this some early thoughts:

1. Removing a material model breaks backward compatibility. I actually think for this model we can do this, as it was only useful as a demonstration for dynamic friction, but you do need to at least:
  - Fix the test that uses this model. Can you recreate the behavior of the test by using visco plastic and setting the parameters appropriately?
  - Write a changelog entry that this material model has been removed

EstherHeck:
@gassmoeller, thanks for your response!
first part is done, it compiles :) 
For the rest, I had similar thoughts as well, that this will break compatibility, but that it might be reasonable in this particular case. 
And yes, I will rewrite the test and see if I can get the same behaviour. And write the changelog entry too.

EstherHeck:
I changed the test for dynamic_friction and use the visco plastic material model instead now. The resulting velocities in the statistics file are very similar (the first 6 digits being exactly the same) and also a visual comparison in paraview  looks good. According to statistics, the new model needs more iterations though.

EstherHeck:
The tester indent+documenation complains to have an error for manual.pdf. Unfortunately I cannot find what exactly goes wrong (e.g. in the raw log of the tester) and when I compile the manual locally I dont get any error... Does someone have an idea where I could find out?

source/material_model/rheology/friction_options.cc
```diff
@@ -0,0 +1,200 @@
...
+              // mu  = mu_d + (mu_s - mu_d) / ( (1 + strain_rate_dev_inv2/dynamic_characteristic_strain_rate)^X );
+              // Angles of friction are used in radians within ASPECT. The coefficient
+              // of friction (mu) is the tangent of the internal angle of friction, hence convergence is needed.
+              const double mu = (std::tan(dynamic_angles_of_internal_friction[j])
+                                 + (std::tan(current_friction) - std::tan(dynamic_angles_of_internal_friction[j]))
+                                 / (1. + std::pow((current_edot_ii / dynamic_characteristic_strain_rate),
+                                                  dynamic_friction_smoothness_exponent)));
+              current_friction = std::atan (mu);
+              Assert((mu < 1) && (0 < current_friction) && (current_friction <= 1.6), ExcMessage(
+                       "Something is wrong with the tan/atan conversion of friction coefficient to friction angle in RAD."));
```
anne-glerum, line 72:
```suggestion
                       "The friction coefficient should be larger than zero and smaller than 1. The friction angle should be smaller than 1.6 rad."));
```

source/material_model/rheology/friction_options.cc
```diff
@@ -0,0 +1,200 @@
...
+              // velocity divided by local element size. This reference strain rate is called
+              // the dynamic characteristic strain rate and is used to see what value between
+              // dynamic and static angle of internal friction should be used.
+              // Furthermore a smoothness exponent X is added, which determines whether the
+              // friction vs strain rate curve is rather step-like or more gradual.
+              // mu  = mu_d + (mu_s - mu_d) / ( (1 + strain_rate_dev_inv2/dynamic_characteristic_strain_rate)^X );
+              // Angles of friction are used in radians within ASPECT. The coefficient
+              // of friction (mu) is the tangent of the internal angle of friction, hence convergence is needed.
+              const double mu = (std::tan(dynamic_angles_of_internal_friction[j])
+                                 + (std::tan(current_friction) - std::tan(dynamic_angles_of_internal_friction[j]))
```
anne-glerum, line 67: Why is current_friction used to compute current_friction? Because it holds the static_friction_angles right? Please explain that in a comment. Also, instead of current_friction, could you call it current_friction_angles in this function? Just to make the distinction between friction angle and friction coefficient clearer.

source/material_model/rheology/friction_options.cc
```diff
@@ -0,0 +1,200 @@
...
+
+
+
+      template <int dim>
+      void
+      FrictionOptions<dim>::declare_parameters (ParameterHandler &prm)
+      {
+        prm.declare_entry ("Friction dependence mechanism", "none",
+                           Patterns::Selection("none|dynamic friction"),
+                           "Whether to make the friction angle dependent of strain rate. This rheology "
```
anne-glerum, line 97:
```suggestion
                           "Whether to make the friction angle dependent on strain rate or not. This rheology "
```

source/material_model/rheology/friction_options.cc
```diff
@@ -0,0 +1,200 @@
...
+                           "The equation is modified after Equation (13) in \\cite{van_dinther_seismic_2013}. "
+                           "$\\mu_s$ and $\\mu_d$ can be specified by setting 'Angles of internal friction' and "
+                           "'Dynamic angles of internal friction', respectively. "
+                           "This relationship is similar to rate-and-state friction constitutive relationships, which "
+                           "are applicable to the strength of rocks during earthquakes.");
+
+        // Dynamic friction paramters
+        prm.declare_entry ("Dynamic characteristic strain rate", "1e-12",
+                           Patterns::Double (0),
+                           "The characteristic strain rate value, where the angle of friction is "
```
anne-glerum, line 120:
```suggestion
                           "The characteristic strain rate value at which the angle of friction is "
```

source/material_model/rheology/friction_options.cc
```diff
@@ -0,0 +1,200 @@
...
+                           "equal to $\\mu = (\\mu_s+\\mu_d)/2$. When the effective strain rate "
+                           "is very high, the dynamic angle of friction is taken for computation, when it is very low, "
```
anne-glerum, line 122:
```suggestion
                           "is very high, the dynamic angle of friction is taken, when it is very low, "
```

source/material_model/rheology/visco_plastic.cc
```diff
@@ -256,12 +257,17 @@ namespace aspect
             // Step 3b: calculate current (viscous or viscous + elastic) stress magnitude
             double current_stress = 2. * viscosity_pre_yield * current_edot_ii;
 
-            // Step 4: calculate strain-weakened friction, cohesion
+            // Step 4a: calculate strain-weakened friction, cohesion
```
anne-glerum, line 260:
```suggestion
            // Step 4a: calculate strain-weakened friction and cohesion
```

source/material_model/rheology/visco_plastic.cc
```diff
@@ -256,12 +257,17 @@ namespace aspect
...
             const DruckerPragerParameters drucker_prager_parameters = drucker_prager_plasticity.compute_drucker_prager_parameters(j,
                                                                       phase_function_values,
                                                                       n_phases_per_composition);
             const double current_cohesion = drucker_prager_parameters.cohesion * weakening_factors[0];
-            const double current_friction = drucker_prager_parameters.angle_internal_friction * weakening_factors[1];
+            output_parameters.current_friction_angles[j] = drucker_prager_parameters.angle_internal_friction * weakening_factors[1];
```
anne-glerum, line 265:
```suggestion
            output_parameters.current_friction_angles[j] = current_friction;
```

source/material_model/rheology/visco_plastic.cc
```diff
@@ -256,12 +257,17 @@ namespace aspect
...
+
+            // Steb 4b: calculate friction angle dependent on strain rate if specified
+            if (friction_options.get_friction_dependence_mechanism() == dynamic_friction)
+              output_parameters.current_friction_angles[j] = friction_options.compute_dependent_friction_angle(current_edot_ii,
+                                                             j, output_parameters.current_friction_angles[j]);
```
anne-glerum, line 270: Is there a preferred order in which the changes to the friction angle should be applied. E.g. first compute the rate-dependent friction and then apply the strain weakening?

source/material_model/rheology/visco_plastic.cc
```diff
@@ -256,12 +257,17 @@ namespace aspect
...
             const DruckerPragerParameters drucker_prager_parameters = drucker_prager_plasticity.compute_drucker_prager_parameters(j,
                                                                       phase_function_values,
                                                                       n_phases_per_composition);
             const double current_cohesion = drucker_prager_parameters.cohesion * weakening_factors[0];
-            const double current_friction = drucker_prager_parameters.angle_internal_friction * weakening_factors[1];
+            output_parameters.current_friction_angles[j] = drucker_prager_parameters.angle_internal_friction * weakening_factors[1];
+
+            // Steb 4b: calculate friction angle dependent on strain rate if specified
+            if (friction_options.get_friction_dependence_mechanism() == dynamic_friction)
+              output_parameters.current_friction_angles[j] = friction_options.compute_dependent_friction_angle(current_edot_ii,
```
anne-glerum, line 269: I find it confusing that the current_friction is no longer the current friction when a friction_dependence_mechanism is active. Maybe you can first compute the right current_friction and then store the final product in the output_parameters. Then you also don't have to use output_parameters below.

source/material_model/rheology/visco_plastic.cc
```diff
@@ -256,12 +257,17 @@ namespace aspect
...
+            // Step 4a: calculate strain-weakened friction, cohesion
             const DruckerPragerParameters drucker_prager_parameters = drucker_prager_plasticity.compute_drucker_prager_parameters(j,
                                                                       phase_function_values,
                                                                       n_phases_per_composition);
             const double current_cohesion = drucker_prager_parameters.cohesion * weakening_factors[0];
-            const double current_friction = drucker_prager_parameters.angle_internal_friction * weakening_factors[1];
+            output_parameters.current_friction_angles[j] = drucker_prager_parameters.angle_internal_friction * weakening_factors[1];
+
+            // Steb 4b: calculate friction angle dependent on strain rate if specified
+            if (friction_options.get_friction_dependence_mechanism() == dynamic_friction)
```
anne-glerum, line 268:
```suggestion
            if (friction_options.get_friction_dependence_mechanism() != independent)
```
This already takes care of any future additional mechanisms.

source/material_model/rheology/visco_plastic.cc
```diff
@@ -729,7 +740,9 @@ namespace aspect
                                                                           n_phases_per_composition);
                 plastic_out->cohesions[i]   += volume_fractions[j] * (drucker_prager_parameters.cohesion * weakening_factors[0]);
                 // Also convert radians to degrees
-                plastic_out->friction_angles[i] += 180.0/numbers::PI * volume_fractions[j] * (drucker_prager_parameters.angle_internal_friction * weakening_factors[1]);
+                double current_friction = calculate_isostrain_viscosities(in, i, volume_fractions,
```
anne-glerum, line 743:
```suggestion
                const double current_friction = calculate_isostrain_viscosities(in, i, volume_fractions,
```

anne-glerum, line 743: It's a bit annoying we have to evaluate calculate_isostrain_viscosities again, but I don't see a simple solution at the moment.

tests/dynamic_friction.prm
```diff
@@ -47,20 +35,26 @@ subsection Mesh deformation
...
-       set Coefficients of static friction = 0.5547 , 0.5
-       set Coefficients of dynamic friction = 0.6 , 0.5
-       set Cohesions = 20e6
-    end
+    set Minimum viscosity = 1e19
+    set Maximum viscosity = 1e25
+    set Reference strain rate = 1e-15
+
+    # Dislocation creep parameters for
+    set Prefactors for dislocation creep          =      1e-20, 1e-25
```
anne-glerum, line 47: Although the background viscosities were not used in the old test, there is a factor 2 in the creep law that you should take into account in the prefactor to get the values of 1e19 and 1e25.

tests/dynamic_friction.prm
```diff
@@ -47,20 +35,26 @@ subsection Mesh deformation
...
+    set Stress exponents for dislocation creep    =      1.0
+    set Activation energies for dislocation creep =      0.0
+    set Activation volumes for dislocation creep  =      0.0
+
+  # Plasticity parameters
+  set Friction dependence mechanism       = dynamic friction
+  set Dynamic characteristic strain rate  = 1e-15
```
anne-glerum, line 54: Align

tests/dynamic_friction.prm
```diff
@@ -47,20 +35,26 @@ subsection Mesh deformation
...
+    set Angles of internal friction         = 29, 26.6
+    set Dynamic angles of internal friction = 31, 26.6
```
anne-glerum, line 56: Maybe the roundoff of the friction angles also plays a role in the differences you see with the new mm?

tests/dynamic_friction.prm
anne-glerum, line 54: What about the smoothing exponent? Is it by default not applied?

anne-glerum (changes requested):
Thanks for this update! I've left some comments to address.

tests/dynamic_friction.prm
EstherHeck: smoothing exponent is 1 by default, so it is not applied

source/material_model/rheology/visco_plastic.cc
EstherHeck: Can you explain why? there is no current_friction variable at this point in the code...

source/material_model/rheology/visco_plastic.cc
EstherHeck: that is a very good question, thanks for spotting this! I am actually not sure whether there is a preferred order or not. I tend to say that rate-and-state should come first and then strain weakening? Does someone else have a preference or even an explanation ready?

source/material_model/rheology/visco_plastic.cc
EstherHeck: yes sure, I can do this. I thought it would be good to remove unnecessary variables, but if you think that is more intuitive I am happy to change it back

source/material_model/rheology/visco_plastic.cc
EstherHeck: yes, I totally agree!

source/material_model/rheology/visco_plastic.cc
EstherHeck: Ah I guess this is explained by your other comment below

anne-glerum: I might have been confused. Originally current_friction was computed here, but I see now that you have removed that line and just fill output_parameters.current_friction_angles. If you don't keep current_friction, than you can leave it as it is, also below! Not necessary to have two parameters with the same value around.

source/material_model/rheology/visco_plastic.cc
EstherHeck: @anne-glerum  can you have a quick look at the new commit if this is better now?

tests/dynamic_friction.prm
EstherHeck: thought about that too. Will try it again

tests/dynamic_friction.prm
EstherHeck: ok. What exactly would I have to change here? 
And do I really have to? The viscosities agree with the old test result when I apply the changes I describe below

EstherHeck:
Hi @anne-glerum thanks a lot for all your comments! I adressed most of them. The remaining issue is now the failing test. 

With the visco_plastic material model plus dynamic_friction theology it shows slightly different results than with the old material model, though qualitatively the result is very similar. 
I already found that I can get the exact same paraview values for properties like viscosities, velocities, pressure when I change the following computations in rheology/visco_plastic.cc
- always assume yielding when using dynamic friction rheology
- use a different approach to cut viscosities with min and max. In visco_plastic it is done with
output_parameters.composition_viscosities[j] = std::min(std::max(viscosity_yield, min_visc), max_visc);
while in dynamic_friction it was done with
viscosities[i] = 1.0 / ( ( 1.0 / (plastic_viscosity + minimum_viscosity) ) + (1.0 / maximum_viscosity) );
- outcomment the assertThrow that temperatures cannot be zero and set it to zero 
However, the density in paraview is still off and so far I don't see why

source/material_model/rheology/visco_plastic.cc
EstherHeck: ok ;) I changed it back in the new commit. Should we have two parameters (current_friction and output_parameters.current_friction_angles) to make it more intuitive or just one parameter (output_parameters.current_friction_angles) to not have the same value stored twice?

tests/dynamic_friction.prm
EstherHeck: nope, setting the friction angles with less roundoff unfortunately does not make the density curves fit.

tests/dynamic_friction.prm
anne-glerum: * the old material model did not consider the background viscosities, except for the reference viscosity, which is set to the first background viscosity
* the old test specified background viscosities
* the new test specifies background viscosities
* the new material model does use background viscosities
* the reference viscosity in the new material model should be set from the input file (although this is no longer used in the scaling of the Stokes equations)
* the new material model uses eta_dislocation = 1/2 * 1/A * strainrate ^(1-n)/n * exp((E+VP)/(nRT)), which means with n=1, V=0, E=0 that 1e20 = 1/2 * 1/A -> A = 5e-21 and similar A = 5e-26
* So just change the prefactors and set the reference viscosity to 1e20 for completeness.

include/aspect/material_model/rheology/friction_options.h
```diff
@@ -0,0 +1,122 @@
...
+          declare_parameters (ParameterHandler &prm);
+
+          /**
+           * Read the parameters from the parameter file.
+           */
+          void
+          parse_parameters (ParameterHandler &prm);
+
+          /**
+           * A function that computes the new friction angle when rate and/or state
```
anne-glerum, line 72:
```suggestion
           * A function that computes the new friction angle when it is not independent.
```

include/aspect/material_model/rheology/friction_options.h
```diff
@@ -0,0 +1,122 @@
...
+           * dependence is taken into account. Given a volume fraction with
```
anne-glerum, line 73:
```suggestion
           * Given a volume fraction with
```

source/material_model/rheology/visco_plastic.cc
```diff
@@ -299,7 +307,7 @@ namespace aspect
                   if (current_stress >= yield_stress)
                     {
                       viscosity_yield = drucker_prager_plasticity.compute_viscosity(current_cohesion,
-                                                                                    current_friction,
+                                                                                    output_parameters.current_friction_angles[j],
```
anne-glerum, line 310: Now you can keep current_friction here right? current_friction and output_parameters.current_friction_angles[j] are the same.

tests/dynamic_friction.prm
```diff
@@ -47,20 +35,26 @@ subsection Mesh deformation
 end
 
 subsection Material model
-
-  set Model name = dynamic friction
-  subsection Dynamic Friction
-    set Densities         = 2800
+  set Model name = visco plastic
+  subsection Visco Plastic
+    set Densities                  = 2800w
```
anne-glerum, line 40:
```suggestion
    set Densities                  = 2800
```

tests/dynamic_friction.prm
anne-glerum: This might be the cause of the difference in density:
In line 142 of dynamic_friction.cc a default thermal expansivity is set to 4e-5. But for visco_plastic, the default is 0. So please try by setting the thermal expansivity from the input file to 4e-5.

EstherHeck: yes, density is fine now

EstherHeck:
here is an impression on how good / bad the new and old test results agree when making those changes I explained above. Green is the old model, pink dashed the new model. 

plot over line along the bottom boundary: 
image

plot over line along the bottom boundary, zoomed in at the places where it disagrees: 
image

plot over diagonal line:
image

plot over diagonal line after 100 iterations (not 1 as the test usually had)
image

source/material_model/rheology/visco_plastic.cc
EstherHeck: I put a comment into the code now, stating, that I dont know the right order....

EstherHeck:
Hi @gassmoeller, documentation builds now, and the visco_plastic_additional_output test is fixed too. Thanks for helping!

The only test failing now is the dynamic_friction test. You commented above that you "wouldnt be surprised if test results were slightly different due to my fix". I guess with fix you mean inserting the missing sqrt? Because even if I change it back to no sqrt, as I did in that test branch I mentioned above, the test results are not the same as before. I wrote about the differences in code and results somewhere above, before I inserted all the figures. Sorry to ask again, to be sure. Do you think the new test results are ok?

gassmoeller:
Hi @EstherHeck, sorry for ignoring this PR for so long, I got sidetracked and only your recent PR #4525 reminded me that this is still open. 

I think this is essentially ready, except I would like to take another look at the test failures. Could you rebase? This should trigger a rerun of the tests. I am mostly concerned, because in the last run some tests crashed (https://jenkins.tjhei.info/blue/organizations/jenkins/aspect/detail/PR-4357/81/pipeline) and I would like to see if that still happens in a rebased PR. Let me know when you got around to it and I will take another look as soon as possible.

EstherHeck:
Hi @gassmoeller I rebased this PR, hope the tests will look better now!

EstherHeck:
If I understand the system correctly it is only the dynamic friction test that is failing now after the rebase. If you are ok with the changes I wrote about above I could update the test results for that test. Please let me know if yes or no :)

gassmoeller:
That is correct, all tests pass except for dynamic friction, and even that one has the same results, and just a different number of iterations to reach the results. Please go ahead and update the test results with the official tester results, afterwards this is ready to merge.

EstherHeck:
great! I updated the test result and squashed the commits from the reviewing process. If you rather want me to squash all commits of this PR into one, let me know

gassmoeller (approved):
Only squashing the additions is fine. Thanks for the work! :+1:

Issue #4356, opened 2021-09-16
Compositions on particles incorrect when enabling free surface


MFraters:
I have a very simple test model based on the `particle_property_composition` test. There are trhee compositions, each with a different value (2, 5 and 8 respectively). I output the compositions both on the mesh in paraview and on the particles. When there is no free surface, the values are the same and correct. When I turn on the free surface, the values on the mesh are still correct, but the values on the particles are all set to the first composition (a value of 2 in this case). I also tested it for 2 compositions with the same result.

I have been able to establish with git bisect (I had not used this before, amazing tool!) that the bug is introduced in pull request https://github.com/geodynamics/aspect/pull/4060. This is where some deal.ii particle FE evaluation was replaced with a new class in aspect. 

I have traced it back to this line having the issue https://github.com/geodynamics/aspect/blob/70493434301737e8caa4f2c94d7959bedf952050/source/particle/world.cc#L1129, which means that either the compositions variable is not set correctly or the get_value does something incorrect. I printed the compositions variable in gdb and when I print it it doesn't contain any 5 value (which is the value of the second composition should be). I have put the gdb print of the compositions (the case of 3 compositions) below in case it is useful.

Some help with this would be much appreciated!

The input file:
```
set Adiabatic surface temperature          = 1623               # default: 0
set CFL number                             = 1.0
set Nonlinear solver scheme                = iterated Advection and Stokes
set Output directory                       = particle_property_composition
set Max nonlinear iterations               = 10
set Nonlinear solver tolerance             = 1e-5

set Dimension                              = 2
set End time                               = 0

set Pressure normalization                 = no #surface
set Surface pressure                       = 0
set Resume computation                     = false
set Start time                             = 0

set Use years in output instead of seconds = true

subsection Discretization
  subsection Stabilization parameters
    set beta  = 0.001 # default: 0.078
  end
end

subsection Compositional fields
  set Number of fields = 3
  set Names of fields = porosity,peridotite, 3
end

subsection Boundary temperature model
  set List of model names = initial temperature

  subsection Initial temperature
    set Maximal temperature = 3773 # default: 6000
    set Minimal temperature = 273  # default: 0
  end

end

subsection Boundary composition model
  set List of model names = initial composition
end

subsection Boundary velocity model
  subsection Function
    set Function expression = 0;-1.65673998e-4
  end
end

subsection Geometry model
  set Model name = box # default:

  subsection Box
    set X extent  = 10
    set Y extent  = 400
[...]

PR #4360, opened 2021-09-22
Evaluate compositions more efficiently for particles

gassmoeller:
This is an optimization I was working on for a while, but it should also be a fix for #4356. 

Currently the fast path for particle evaluation using the FEPointEvaluation class in deal.II used an `std::vector<FEPointEvaluation<1>> (n_compositions)` to evaluate the compositional fields. That was necessary because FEPointEvaluation is templated on the number of components, which is only known at runtime in ASPECT. Somehow this does not seem to work correctly however, because it looks like FEPointEvaluation always starts to evaluate the first component of a finite element with multiplicities (like our composition elements). I am not sure if this is a bug in deal.II or if I used FEPointEvaluation in a way it was not  intended for, but I wanted to optimize this anyway.

Instead with this PR I now create a base class of SolutionEvaluators that is not templated on the number of compositions (so that it can be created without that knowledge), and a derived class that is templated on the number of compositions. The derived class can then use an `FEPointEvaluation<n_compositions>` object, and can be created at runtime and stored as a pointer to the non-templated base class. Currently I instantiate the derived class for up to 10 compositional fields, is this sufficient?

@tjhei: Could you take a look if I did the instantiation trick correctly?
@MFraters: Could you give this a try and check if it resolves you issue? I tried with your example model and that seems to work now. Also do you know of many models with more than 10 compositions? If so, how many models and how many compositions?


MFraters:
Yes, this fixes the problem for me! I don't think I have every used more than 10 compositional fields (max is probably 6), but if I would start to use elasticity in 3D then I might get close or go over. Maybe @naliboff or @anne-glerum may do this?

source/particle/world.cc
```diff
@@ -1176,6 +1173,121 @@ namespace aspect
...
+                                                                                      const unsigned int n_compositional_fields,
+                                                                                      const UpdateFlags update_flags)
+      {
+        AssertThrow(n_compositional_fields <= 10, ExcInternalError());
+
+        std::unique_ptr<internal::SolutionEvaluators<dim>> evaluators;
+        switch (simulator_access.n_compositional_fields())
+          {
+            case 0:
+              evaluators = std::make_unique<internal::SolutionEvaluatorsImplementation<dim,0>>(simulator_access,  update_flags);
```
tjhei, line 1201: I would do return std::make_unique...

source/particle/world.cc
```diff
@@ -1176,6 +1173,121 @@ namespace aspect
...
+              evaluators = std::make_unique<internal::SolutionEvaluatorsImplementation<dim,8>>(simulator_access,  update_flags);
+              break;
+            case 9:
+              evaluators = std::make_unique<internal::SolutionEvaluatorsImplementation<dim,9>>(simulator_access,  update_flags);
+              break;
+            case 10:
+              evaluators = std::make_unique<internal::SolutionEvaluatorsImplementation<dim,10>>(simulator_access,  update_flags);
+              break;
+            default:
+              AssertThrow(false, ExcInternalError());
```
tjhei, line 1234: Maybe a better error message? Or can we make a slower fallback?

source/particle/world.cc
```diff
@@ -882,65 +882,69 @@ namespace aspect
...
-                                             particles_in_cell.end(),
-                                             evaluators);
-                    else
-                      local_update_particles(cell,
-                                             particles_in_cell.begin(),
-                                             particles_in_cell.end());
-                  }
+      // This class evaluates the solution vector at arbitrary positions inside a cell.
+      // This base class only provides the interface for SolutionEvaluatorsImplementation.
+      // See there for more details.
```
tjhei, line 889: Is there a reason for not using doxygen comments?

tjhei (approved):
I am not sure if 10 is enough (I doubt it), but the instantiation looks correct to me.

anne-glerum:
We do have use-cases with more than 10 fields.

naliboff:
I also have use cases with more than 10 fields. That will certainly be the case for many future 3D models that use elasticity.

gassmoeller:
Can you provide an estimate for a realistic upper limit? 15? 20? I need to manually instantiate them, and every instantiation will increase the time to compile and the size of the binary (slightly).

naliboff:
I think 20 is a reasonable upper limit. For example, in a recent paper with 3D models I had 9 different fields for lithologies (rock types), and 2 additional fields for strain invariants. Adding in elasticity would bring this to 17 fields.

tjhei:
Rene, is there a way to evaluate say 10 at a time for a large number of fields? Or even better with a couple powers of two after say 10?

gassmoeller:
> Rene, is there a way to evaluate say 10 at a time for a large number of fields? Or even better with a couple powers of two after say 10?

Not quite sure how you mean this. Do you suggest to e.g. use a `std::vector<FEPointEvaluation<5>>` with 3 vector entries instead of a `FEPointEvaluation<15>` because then we dont need the additional instantiations? This may be possible, but would be complicated by the fact that FEPointEvaluation falls back to the slow path whenever the last component exceeds the current base element of the FESystem (so we can not simply use more components than actually exist). Also as Menno pointed out in #4356 it looks like FEPointEvaluation currently does not support to start evaluating at a middle component of a base element with multiplicities.

One possibility would be to have a `std::vector<FEPointEvaluation<1>>` as fallback when the number of fields surpasses the number of components we instantiate for (this essentially means use the current implementation for many compositional fields, and the new one of this PR for less than X). If I find and fix the bug in FEPointEvaluation we could even always use the `FEPointEvaluation<X>` for the first X components and only use `std::vector<FEPointEvaluation<1>>` for the remaining compositions.

tjhei:
> Not quite sure how you mean this.

My suggestion was based on the fact that you can start evaluating "in the middle". Say you have instantiations for 1,2,3, ..., 8,and 16, and 24. With that, you could evaluate all fields with one call for N<=8 and two calls for N<=32, etc.

> One possibility would be to have a `std::vector<FEPointEvaluation<1>>` as fallback when the number of fields surpasses the number of components we instantiate for

I think this is not a bad idea if my suggestion above is too complicated.

gassmoeller:
> My suggestion was based on the fact that you can start evaluating "in the middle". Say you have instantiations for 1,2,3, ..., 8,and 16, and 24. With that, you could evaluate all fields with one call for N<=8 and two calls for N<=32, etc.

Ah, I see now, nice idea. I will give that a try.

gassmoeller:
I think I addressed all comments. The class is now instantiated to be fast up to 20 compositional fields, and afterwards allocates individual scalar fields, so any field above #20 increases the cost by as much as every field does before this PR. With deal.II 9.3 beyond 20 fields FEPointEvaluation is not used, because it contains the FEPointEvaluation bug, so #4356 will be fixed no matter if deal.II 10.0 or 9.3 is used. The approach to evaluate in multiples of 2 beyond 8 turned out to be too complicated in the code (too many if's inside the evaluation, confusing to read), so while the current approach has some many instantiations, at least the code inside the evaluator class is pretty compact.

Is this ready to merge? I would like to fix #4356 soon before more users stumble into it (I realize this PR was waiting for me to address the comments).

PR #4361, opened 2021-09-22
Apply update scripts to source and parameter files

gassmoeller:
Apply our parameter and source file update scripts to eliminate some outdated variables and input parameters.

### DIFF
--- a/cookbooks/plume_2D_chunk/plume2D.prm
+++ b/cookbooks/plume_2D_chunk/plume2D.prm
@@ -62,7 +62,7 @@ end
 # equator (i.e. x-axis) then the patch is defined by 7pi/16<phi<pi/2.
 subsection Boundary temperature model
   set Fixed temperature boundary indicators = top, bottom
-  set Model name = function
+  set List of model names = function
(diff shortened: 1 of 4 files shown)


tjhei (approved):
So the scripts apply cleanly? Can we do this in CI as part of the indent step?

gassmoeller:
Unfortunately they do not apply cleanly, some of the scripts to update from 1.5.0 to 2.0.0 collide with some comments and names of newer parameter and source files and we also reverted one of the changes of that script (we now again have a postprocessor viscous dissipation statistics that we removed before). The new ones for 2.2.0 apply cleanly though. We could decide to only apply a subset of scripts? But that would have to be maintained for future versions (changing which scripts to apply as part of the CI).

tjhei:
> The new ones for 2.2.0 apply cleanly though. We could decide to only apply a subset of scripts? But that would have to be maintained for future versions (changing which scripts to apply as part of the CI).

I think that would be reasonable. We could create a folder "current" with symlinks to the latest update scripts and just run those.

PR #4367, opened 2021-09-25
fix MeshDeformationHandler<dim>::*_matching_mesh_deformation_object

zjiaqi2018:
Fix a few places in `MeshDeformationHandler<dim>::*_matching_mesh_deformation_object` as suggested by @tjhei .

### DIFF
--- a/include/aspect/mesh_deformation/interface.h
+++ b/include/aspect/mesh_deformation/interface.h
@@ -538,7 +538,7 @@ namespace aspect
     bool
     MeshDeformationHandler<dim>::has_matching_mesh_deformation_object () const
     {
-      for (typename std::map<types::boundary_id, std::vector<std::unique_ptr<Interface<dim>>>>::iterator boundary_id
+      for (typename std::map<types::boundary_id, std::vector<std::unique_ptr<Interface<dim>>>>::const_iterator boundary_id
            = mesh_deformation_objects.begin();
            boundary_id != mesh_deformation_objects.end(); ++boundary_id)
         for (const auto &p : boundary_id->second)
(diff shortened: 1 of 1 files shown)


include/aspect/mesh_deformation/interface.h
```diff
@@ -538,7 +538,7 @@ namespace aspect
     bool
     MeshDeformationHandler<dim>::has_matching_mesh_deformation_object () const
     {
-      for (typename std::map<types::boundary_id, std::vector<std::unique_ptr<Interface<dim>>>>::iterator boundary_id
+      for (typename std::map<types::boundary_id, std::vector<std::unique_ptr<Interface<dim>>>>::const_iterator boundary_id
```
tjhei, line 541: Can you use a ranged for loop with const auto & here? Also, I would call the variable something else: it is not a boundary id but an iterator.

include/aspect/mesh_deformation/interface.h
```diff
@@ -562,20 +562,17 @@ namespace aspect
                              "that could not be found in the current model. Activate this "
                              "mesh deformation in the input file."));
 
-      for (typename std::map<types::boundary_id, std::vector<std::unique_ptr<Interface<dim>>>>::iterator boundary_id
+      for (typename std::map<types::boundary_id, std::vector<std::unique_ptr<Interface<dim>>>>::const_iterator boundary_id
```
tjhei, line 565: Same here.

tjhei (approved):
Thank you. It is worth mentioning that the old code did not compile...

tjhei (approved):
So much nicer than before! 👍

PR #4369, opened 2021-09-30
A postprocessor that outputs the second invariant of the deviatoric stress tensor

anne-glerum:
### For all pull requests:


bobmyhill:
Could the calculation of the stress tensor be moved into a utilities function? There's quite a lot of code overlap with the Stress visualization function (which is currently in flux; see #4375).

anne-glerum:
I agree that it would make sense not to repeat assembling the stress tensor in different postprocessors. Currently there are 4 pps that could benefit from a utilities function:
```
stress.cc
shear_stress.cc
maximum_horizontal_compressive_stress.cc
principal_stress.cc
```
However, I think it falls outside the scope of this current PR to create such a function. Happy to open a new PR of course.

anne-glerum:
I'll add the changelog, but should I undo the change in the way the stress is computed when elasticity is used? Seeing that we decided to not change the other stress postprocessors, undoing the last commit would at least make all the stress postprocessors consistent.

PR #4373, opened 2021-10-01
Print solver history on failure if only cheap Stokes solver is used

gassmoeller:
This fixes a bug reported by @alarshi. At the moment if we only use cheap Stokes solver steps and fail to converge, we only print the assertion message from the solver, but do not print the solver history. This was an oversight of me in #4335. I added a test that shows the correct behavior.


alarshi:
@gassmoeller : Hi Rene,
It took me some time to understand what was going on, but I get it now and the PR looks good to me. Thank you for the review opportunity, I learned more about exception handling! :)

gassmoeller:
> Thank you for the review opportunity, I learned more about exception handling! :)

Ha, I feel bad when you thank me to give you more work ;-), I mostly thought it is a good opportunity to learn something.

Per our merge policy we need one of the other principal developers to approve this PR. Then we can go ahead and merge.

PR #4374, opened 2021-10-01
Stokes: need grads_phi_u when enable_elasticity

zjiaqi2018:
For GMG, we need grads_phi_u when we have elasticity. 
@tjhei Does this look right to you?


tjhei (approved):
Yes, looks reasonable. This has also incorrect for matrix-based, but we likely never trigger it.

PR #4376, opened 2021-10-06
fix GMG mesh deformation missing ghost entries

tjhei:
Without adding additional ghost entries to the level displacement vectors used by the ``MappingQEulerian``, we run into crashes like
```
An error occurred in line <1627> of file </ssd/candi-v9.3.1-r1/tmp/unpack/deal.II-v9.3.1/include/deal.II/lac/la_parallel_vector.h> in function
    Number dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::operator()(dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::size_type) const [with Number = double; MemorySpace = dealii::MemorySpace::Host; dealii::LinearAlgebra::distributed::Vector<Number, MemorySpace>::size_type = unsigned int]
The violated condition was: 
    partitioner->in_local_range(global_index) || partitioner->ghost_indices().is_element(global_index)
```
when running in parallel.

### DIFF
--- a/source/mesh_deformation/interface.cc
+++ b/source/mesh_deformation/interface.cc
@@ -913,9 +913,23 @@ namespace aspect
           const unsigned int n_levels = this->get_triangulation().n_global_levels();
 
           level_displacements.resize(0, n_levels-1);
-          level_mappings.resize(0, n_levels-1);
+          // Important! Preallocate level vectors with all needed ghost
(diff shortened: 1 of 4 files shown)


tjhei:
@zjiaqi2018 this fixes the problem for me.

PR #4377, opened 2021-10-06
add failing 9.3 test

tjhei:
MGLevelGlobalTransfer in deal.II 9.3.1 contains a bug that has been fixed on master. This is triggered in ASPECT with GMG, mesh deformation, and adaptive refinement (in certain cases).
Add the test now (and disable it until we switch to 10.0).


tjhei:
@zjiaqi2018 also relevant for you. Use deal.II master for now.

tjhei:
fixed in https://github.com/dealii/dealii/pull/12634

tjhei:
> I assume this case is disabled in master for deal.II prior to the fix? If so, ready to merge.

The test added here is disabled (note the extension is not prm) as it crashes. We will need to remember to update the name and add the correct output.

PR #4379, opened 2021-10-20
higher order mesh deformation [WIP]

tjhei:
initial experiment to use a higher order polynomial degree (same as velocity) for the mesh deformation.

FYI @mfmweerdesteijn

### DIFF
--- a/source/mesh_deformation/interface.cc
+++ b/source/mesh_deformation/interface.cc
@@ -95,13 +95,10 @@ namespace aspect
     template <int dim>
     MeshDeformationHandler<dim>::MeshDeformationHandler (Simulator<dim> &simulator)
       : sim(simulator),  // reference to the simulator that owns the MeshDeformationHandler
-        mesh_deformation_fe (FE_Q<dim>(1),dim), // Q1 elements which describe the mesh geometry
+        mesh_deformation_fe (FE_Q<dim>(sim.parameters.stokes_velocity_degree),dim),
         mesh_deformation_dof_handler (sim.triangulation),
         include_initial_topography(false)
     {
-      // Now reset the mapping of the simulator to be something that captures mesh deformation in time.
-      sim.mapping.reset (new MappingQ1Eulerian<dim, LinearAlgebra::Vector> (mesh_deformation_dof_handler,
-                                                                            mesh_displacements));
     }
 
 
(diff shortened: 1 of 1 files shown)


tjhei:
This compiles and runs (free_surface cookbook), but I have not tested anything. We also need to decide if this should become an option or if we always go with higher order.

Tests are not updated yet either.

@mfmweerdesteijn can you please test this?

mfmweerdesteijn:
I did the same exercise as above but then for a chunk geometry (so based on the free surface cookbook). The box was 500x200 km. The chunk has a 6000 km outer radius and 5800 km inner radius (200 km depth), so then for a 500 km arc distance the longitude goes from 0 to 4.775 degree. And then using x=rcos(phi) and y=rsin(phi) (r is here 5900 km, and phi is 4.775/2 degree), I define the center of the circular temperature anomaly blob (in the box it is at x=250 and y=100 km), with a 25 km radius. This can be seen in this prm file: [free_surface_sph_cb.txt](https://github.com/geodynamics/aspect/files/7468212/free_surface_sph_cb.txt)

Then I did the same variations in parameters and on which branch executed as in the plot above. The only difference is that since it is a chunk geometry the free surface velocity projection has to be normal, so no green and cyan lines which correspond with a vertical projection in this plot.
free surface cookbook chunk

Again there is only a working solution on the mapping branch for a free surface stabilization beta of 1.0 (X's indicated that the run crashed because of the same error as before: "The image of the mapping applied to cell with center [xxxxxx xxxxxx] is distorted. The cell geometry or the mapping are invalid, giving a non-positive volume fraction of -xxxxxx in quadrature point x.").

However, the solution is not symmetric, not on the current aspect version (master) (blue line) and not on this branch (mapping) (grey line). In the box case, the blue and grey line are very similar (not exactly). In the chunk case they are bot skewed away from the center. Here are some animations which show the temperature and the total velocity through time for the box and the chunk model on the master branch (the blue cases) (similar to the mapping branch for the grey cases).
box, temperature: https://user-images.githubusercontent.com/57097877/140080872-8fe4d0c1-a7ee-43c2-ba5c-7e889c507395.mp4
box, velocity: https://user-images.githubusercontent.com/57097877/140080930-7b8dd889-cedd-4581-af4d-2302a05e959c.mp4
chunk, temperature: https://user-images.githubusercontent.com/57097877/140080982-a5ae1686-dc26-4f4d-92ec-00b9506b8dd0.mp4
chunk, velocity: https://user-images.githubusercontent.com/57097877/140081037-04134515-3791-4944-8bd6-f2e9bf9d6810.mp4

tjhei:
Thanks for looking into this. I spent some time going through the code but I couldn't see anything that is obviously incorrect.

Can you try testing if the "ascii data" and "function" mesh deformation plugins work correctly with this branch? You can probably find a test under tests/ that you can run and then check the output visually.

mfmweerdesteijn:
Sorry that it took so long. I had a week of Norwegian and Dutch royal family meetings 😄.

I used the test: box_initial_mesh_deformation_ascii_data.prm to test the ascii data on the master branch (current development version) and mapping branch (as this PR). The mesh is very coarse. Here are the top surface deformation results on both branches. The results are identical.

mesh_ascii_master
mesh_ascii_mapping

Couldn't find a test with the mesh deformation function, but then I used the same prm file and changed it to using Boundary function in the Mesh deformation to prescribe a boundary velocity (linearly along x from 0 to 20 km/yr, and let it run for 1 year, so from 0 to 20 km deformation along x). The results between the two branches are identical here too.

mesh_function_master
mesh_function_mapping

PR #4389, opened 2021-10-29
A postprocessor that determines Mobility of the lithosphere following Lourenco et al., 2020 G3

elodie-kendall:
This PR combines the velocity_statistics and velocity_boundary_statistics postprocessor to get the parameter Mobility following Lourenco et al., 2020 G3.

* [ X] I have read the guidelines in our [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md) document.

### For all pull requests:

* [ X] I have tested my new feature locally to ensure it is correct.
* [X ] I have added a changelog entry in the [doc/modules/changes](../blob/master/doc/modules/changes) directory that will inform other users of my change.


doc/modules/changes/20211029_ekendall
```diff
@@ -0,0 +1,3 @@
+New: This PR combines the velocity_statistics and velocity_boundary_statistics postprocessor to get the parameter Mobility following Lourenco et al., 2020 G3
```
gassmoeller, line 1: You will need to give this file a different name because another one with the same name already exists.

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,201 @@
...
+       double mobility =0.0;       
+
+       for (std::map<types::boundary_id, double>::const_iterator
+            svrms = global_rms_vel.begin();
+            svrms != global_rms_vel.end();
+            ++svrms, ++index)
+       
+        {
+           mobility  = svrms[top_boundary_id]/vrms; 
+        }
```
gassmoeller, line 166: This was the problem you mentioned in your message, correct? In a number of places you currently loop over all boundary indicators (because velocity boundary statistics considers all boundaries), while for your purpose you only want to consider a single boundary id (namely `top_boundary_id`). So what you want to do is to remove all the loops over boundary ids (I will mark them) and use `top_boundary_id` as the single boundary indicator.

So here:

```suggestion
           const double mobility  = global_rms_vel[top_boundary_id]/vrms; 
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,201 @@
...
+          {
+           global_rms_vel[boundary_id] = std::sqrt(global_velocity_square_integral_values[index] / global_boundary_area_values[index]);
+           ++index;
+          }
+       }
+       // now add the computed rms velocities to the statistics object
+       // and create a single string that can be output to the screen
+       //std::ostringstream screen_text;
+       unsigned int index = 0;
+       double mobility =0.0;
```
gassmoeller, line 157:
```suggestion
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,201 @@
...
+          {
+            fe_values.reinit (cell);
+            fe_values[this->introspection().extractors.velocities].get_function_values (this->get_solution(),
+                                                                                        velocity_values);
+            for (unsigned int q = 0; q < n_q_points; ++q)
+              {
+                local_velocity_square_integral += ((velocity_values[q] * velocity_values[q]) *
+                                                   fe_values.JxW(q));
+              }
+          }
```
gassmoeller, line 66: In the current state you copied together the code of the other two postprocessors, which means you loop through all cells twice (here and in line 98) and do a lot of the work twice. You can merge these two loops into a single one. First do all the work necessary for the velocity_rms on the cell (this code), then do all the work necessary for the faces (lines 100-123). Then after this loop over all cells do all communication that is necessary. This will make the code easier to read and also much faster.

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,201 @@
...
+                                        this->get_fe(),
+                                        quadrature_formula_for_surface,
+                                        update_values |
+                                        update_JxW_values |
+                                        update_quadrature_points);
+      
+      std::vector<Tensor<1,dim>> velocities (fe_face_values.n_quadrature_points);
+      
+      std::map<types::boundary_id, double> local_velocity_square_integral_surface;
+      std::map<types::boundary_id, double> local_boundary_area;
```
gassmoeller, line 88: Maps are only necessary for the place where you copied this from, because it computes these values for all boundaries. Here you want two single values (rms velocity and area of the top surface). Therefore this will make a lot of things simpler:

```suggestion
      double local_velocity_square_integral_top_boundary;
      double local_top_boundary_area;
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,201 @@
...
+                for (unsigned int q=0; q<fe_face_values.n_quadrature_points; ++q)
+                  {
+                    const double vel_mag = velocities[q].norm();
+                    local_sqvel += ((vel_mag * vel_mag) * fe_face_values.JxW(q));
+                    local_fe_face_area += fe_face_values.JxW(q);
+                  }
+                // then merge them with the squared velocities
+                // we found for other faces with the same boundary indicator
+                const types::boundary_id boundary_indicator
+                  = cell->face(f)->boundary_id();
```
gassmoeller, line 119: No need to recompute the boundary indicator. Look at line 101: You only get here if the boundary indicator is the top boundary. If you make the change from maps to doubles that I mentioned above the lines below will not need the boundary id at all.

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,201 @@
...
+       
+                local_velocity_square_integral_surface[boundary_indicator] += local_sqvel;
+                local_boundary_area[boundary_indicator] += local_fe_face_area;
```
gassmoeller, line 122:
```suggestion
                local_velocity_square_integral_top_boundary += local_sqvel;
                local_top_boundary_area += local_fe_face_area;
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,201 @@
...
+
+        // and now take them apart into the global map again
+        // At the same time, calculate the rms velocity for each boundary
+        unsigned int index = 0;
+        for (const auto boundary_id: boundary_indicators)
+          {
+           global_rms_vel[boundary_id] = std::sqrt(global_velocity_square_integral_values[index] / global_boundary_area_values[index]);
+           ++index;
+          }
+       }
```
gassmoeller, line 152: No need for all this complicated stuff if you do not use maps anymore. Check how the rms velocity is communicated across MPI processes in line 68/69, you can do the same here:

```suggestion
      const double global_top_velocity_square_integral = Utilities::MPI::sum (...
      const double global_top_boundary_area_integral = Utilities::MPI::sum (...
```

gassmoeller (commented):
Hi Elodie,
As discussed I think this will be a nice postprocessor addition, but I think the code needs to restructured quite a bit. In the current form it is a duplicate of the other two postprocessors and much more complicated than necessary. I left comments for how and where you can simplify things. Leave a comment if some of the comments are not clear.

Also we would need a test for this postprocessor. Can you think of a simple test case that proves the value is computed correctly? I think you can copy and repurpose the files of the simple shear test in tests/simple_shear*. Because in a simple shear flow the rms velocity will have a fixed ratio to the surface velocity (1/2?) we should get a well defined output for the mobility.

elodie-kendall:
Thanks so much for your comments, Rene! It's looking much much better. I have addressed all your suggestions and it compiles. I also added the test you recommended. I am about to patch the test now.

elodie-kendall:
I don't understand why the test is failing. It ran fine locally and then I did the patch from the diff file..

include/aspect/postprocess/mobility_statistics.h
```diff
@@ -0,0 +1,52 @@
+/*
+  Copyright (C) 2011 - 2019 by the authors of the ASPECT code.
```
bobmyhill, line 2:
```suggestion
  Copyright (C) 2011 - 2021 by the authors of the ASPECT code.
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,137 @@
+/*
+  Copyright (C) 2011 - 2019 by the authors of the ASPECT code.
```
bobmyhill, line 2:
```suggestion
  Copyright (C) 2011 - 2021 by the authors of the ASPECT code.
```

doc/modules/changes/20211111_ekendall
```diff
@@ -0,0 +1,3 @@
+New: Added a new postprocessor which comptes the parameter "Mobility" following Lourenco et al., 2020 G3
```
bobmyhill, line 1:
```suggestion
New: Added a new postprocessor which computes the parameter "Mobility" following Lourenco et al., 2020 G3.
```

tests/simple_shear_output_the_mobility.prm
```diff
@@ -0,0 +1,134 @@
...
+#
+# where u is the velocity, t is time, and a comma represents a derivative
+# in that particular direction.
+# Because u_x,y = 2.0 / 1.0 is constant in the whole domain (and over time) we expect
+# a constant F at any given time, and in particular:
+# F_xy (t=1.0) = 2.0
+#
+# This parameter file tracks the finite strain in two ways, once by using
+# compositional fields as illustrated in the finite strain cookbook, and
+# once by using a particle particle that sits and remains in the center of the domain.
```
bobmyhill, line 26:
```suggestion
# once by using a particle that sits and remains in the center of the domain.
```

tests/simple_shear_output_the_mobility.prm
```diff
@@ -0,0 +1,134 @@
...
+# The results should of course be the same.
```
bobmyhill, line 27:
```suggestion
# The results should be the same.
```

anne-glerum:
The test computes a NaN on the failing tester according to the diff:
-     Mobility:                  1.52
+     Mobility:                  -nan
Can you check again locally?
The second failing tester cannot build the commit because it needs to be told to do so by a developer:
/rebuild

tests/simple_shear_output_the_mobility.prm
```diff
@@ -0,0 +1,134 @@
...
+#
+# Numerical results for particle tracking and compositional fields
+# show equally excellent agreement:
```
bobmyhill, line 30:
```suggestion
# show equally excellent agreement.
```

tests/simple_shear_output_the_mobility.prm
```diff
@@ -0,0 +1,134 @@
...
+# F_xy (t=1.0) = 2.0
```
bobmyhill, line 31: repetition of Line 22.

bobmyhill (approved):
A few minor typos to go along with the test fix.

elodie-kendall:
Thanks Anne and Bob. I think the problem was that I was using the screen-shot etc ending in *.no.time. This is now passing all tests and should be ready to merge :)

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,137 @@
...
+              if (cell->face(f)->boundary_id() == top_boundary_id)
+                {
+                  // determine the squared velocity on the face
+                  double local_sqvel = 0.0;
+                  double local_fe_face_area = 0.0;
+                  for (unsigned int q = 0; q < n_q_points; ++q)
+                    {
+                      const double vel_mag = velocity_values[q].norm();
+                      local_sqvel += ((vel_mag * vel_mag) * fe_values.JxW(q));
+                      local_fe_face_area += fe_values.JxW(q);
```
gassmoeller, line 83: Minor improvements to avoid unnecessary computations (saves one square root in the norm, and halves the number of times JxW needs to be computed). I realize you copied this from velocity_boundary_statistics, but there we need the actual norm to store the max and min of the velocity, which we do not need here. This optimization is not strictly necessary for postprocessing, because it is usually fast, but I thought it is instructive.
```suggestion
                      const double JxW = fe_values.JxW(q);
                      local_sqvel += velocity_values[q].norm_square() * JxW;
                      local_fe_face_area += JxW;
```

tests/simple_shear_output_the_mobility.prm
```diff
@@ -0,0 +1,133 @@
+# This parameter file is a copy of simple_shear.prm but with the added postprocessor
+# "mobility statistics" following Lourenco et al., 2020.
+# Mobility = surface rms velocity / whole domain rms velocity.
+# This value is ~1.5 in this test.
```
gassmoeller, line 4: I dont really understand where the 1.5 comes from. I would expect the following: The velocity norm in the domain is (2y-1). I can either integrate the square of this directly, or (to make it simpler) shift the origin to 0.5 and integrate twice over 2y from 0 to 0.5 (both give the same result). I would then get: 
```
v_rms = std::sqrt (1/A^2 * int_A v_squared)
int_A v_squared = 2 * int_0^0.5 (2y)^2 dy = 2 * [4/3 y^3]_0^0.5 = 2 * [4/3*1/8 - 0] = 8/24 = 1/3
with A = 1
v_rms = std::sqrt(1/3)
v_top_rms = 1 (since v=1 for the top, and A_top = 1)
mobility = v_top_rms / v_rms = std::sqrt(3) = 1.73
```
In other words the model should give a mobility of 1.73 not 1.5.

gassmoeller: I think I found the error, will mark it above.

source/postprocess/mobility_statistics.cc
gassmoeller, line 83: You can not use an `fe_values` object to compute properties on faces of cells. `fe_values` is meant to evaluate or integrate the solution over the volume of cells, while `fe_face_values` is meant to integrate over the surface of cells. Keep everything in the one loop you have now, but above you need to keep an `fe_face_values` object and the corresponding quadrature rule.

gassmoeller (commented):
The code looks much better now, but you need to reintroduce the correct surface integral. I think at the moment you integrate over the volume of each surface cell (instead of the area of each top boundary face) to compute the top vrms which leads to the wrong test result.

source/postprocess/mobility_statistics.cc
elodie-kendall: Ok thanks!

elodie-kendall: Ah yes

tests/simple_shear_output_the_mobility.prm
elodie-kendall: yes!

elodie-kendall:
Thanks for reviewing again Rene. I've addressed your comments. The test works locally and fails of course on this tester but the changes-diff file is empty so I don't know how to patch it?

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,157 @@
...
+      const double global_top_rms_vel = std::sqrt(global_top_velocity_square_integral) /
+                                        std::sqrt(global_top_boundary_area_integral);
+
+      // now add the computed rms velocities to the statistics object
+      // and create a single string that can be output to the screen
+      const double mobility  = global_top_rms_vel/global_rms_vel;
+      const std::string name_mobility = "Mobility";
+
+      statistics.add_value (name_mobility,
+                            mobility );
```
gassmoeller, line 129:
```suggestion
                            mobility);
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,157 @@
...
+
+// explicit instantiations
+namespace aspect
+{
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(MobilityStatistics,
+                                  "mobility statistics",
+                                  "A postprocessor that computes some statistics about the "
+                                  "mobility of the lithosphere.")
```
gassmoeller, line 155: Try to keep this description in sync with what you write at the top of the header file. It is a bit  duplicative, but the text in the header file will end up in our online documentation (https://aspect.geodynamics.org/doc/doxygen/annotated.html), while the text here will end up in the manual and the parameter website (https://aspect.geodynamics.org/doc/parameter_view/parameters.xml). Ideally both should contain the same information, in particular the references to the papers.

tests/simple_shear_output_the_mobility.prm
```diff
@@ -0,0 +1,133 @@
+# This parameter file is a copy of simple_shear.prm but with the added postprocessor
+# "mobility statistics" following Lourenco et al., 2020.
+# Mobility = surface rms velocity / whole domain rms velocity.
+# This value is ~1.73 in this test.
```
gassmoeller, line 4: Don't be modest here, the test is incredibly close to the analytical result of sqrt(3) now (looking at the statistics file). I would write this as follows (also to emphasize that his was tested against an analytical result):
```suggestion
# This result of ~1.73205 in this test is very close to the analytical solution of sqrt(3).
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,157 @@
...
+                               this->get_fe(),
+                               quadrature_formula,
+                               update_values   |
+                               update_quadrature_points |
+                               update_JxW_values);
+
+      std::vector<Tensor<1,dim>> velocity_values(n_q_points);
+
+      // create a quadrature formula for the velocity for the surface of cells
+      const QGauss<dim-1> quadrature_formula_face (this->introspection().polynomial_degree.velocities+1);
```
gassmoeller, line 56: You probably want to introduce a similar variable like in the volume integration above:
```suggestion
      const QGauss<dim-1> quadrature_formula_face (this->introspection().polynomial_degree.velocities+1);
      const unsigned int n_q_points_face = quadrature_formula_face.size();
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,157 @@
...
+
+      FEFaceValues<dim> fe_face_values (this->get_mapping(),
+                                        this->get_fe(),
+                                        quadrature_formula_face,
+                                        update_values |
+                                        update_JxW_values |
+                                        update_quadrature_points);
+
+      std::vector<Tensor<1,dim>> surface_velocity_values (fe_face_values.n_quadrature_points);
```
gassmoeller, line 65:
```suggestion
      std::vector<Tensor<1,dim>> surface_velocity_values (n_q_points_face);
```

source/postprocess/mobility_statistics.cc
```diff
@@ -0,0 +1,157 @@
...
+                {
+                  fe_face_values.reinit (cell, f);
+                  // extract velocities
+                  fe_face_values[this->introspection().extractors.velocities].get_function_values (this->get_solution(),
+                      surface_velocity_values);
+
+                  // determine the squared velocity on the face
+                  double local_sqvel = 0.0;
+                  double local_fe_face_area = 0.0;
+                  for (unsigned int q = 0; q < n_q_points; ++q)
```
gassmoeller, line 99: This is what is causing the crash for the test. The volume integration has more integration points than the one over the surface, and calling fe_face_values.JxW(i) with an i larger than the maximum number of points crashes. That is why the test file was empty, but the error was posted here https://github.com/geodynamics/aspect/runs/4274792601?check_suite_focus=true (under `report test results`).
```suggestion
                  for (unsigned int q = 0; q < n_q_points_face; ++q)
```

gassmoeller (commented):
Excellent, this is nearly ready to go. Just some finishing touches left.

source/postprocess/mobility_statistics.cc
elodie-kendall: Ok, thanks

source/postprocess/mobility_statistics.cc
elodie-kendall: Ah yes! Thanks for the tip on finding the error

gassmoeller:
Thanks for making the updates. I was confused for a while because of the one failing tester, but I reran it and now it passes all tests (likely a one-time fail on the test machine). Anyway this is ready to go, congrats on the merged PR!

elodie-kendall:
Awesome, thanks everyone!

PR #4390, opened 2021-10-29
Make compile with deal.II-dev

gassmoeller:
This makes ASPECT compile with the recent deal.II dev. dealii/dealii#12874 made a parameter const that is set in ASPECT. ASPECT uses our own modified copy of the `ParticleOutput` class.

### DIFF
--- a/source/postprocess/particles.cc
+++ b/source/postprocess/particles.cc
@@ -121,7 +121,11 @@ namespace aspect
           {
             patches[i].vertices[0] = particle->get_location();
             patches[i].patch_index = i;
+
+#if !DEAL_II_VERSION_GTE(10,0,0)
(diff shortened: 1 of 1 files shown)


tjhei (approved):
@bangerth breaking ASPECT. 😉

tjhei:
How does this work @gassmoeller , shouldn't you set the value as long as dim>0?

bangerth (approved):
Ah, my bad. I think the line was always unnecessary, but this is the safe solution.

tjhei:
I don't understand. Why is setting it to one optional only now? I there a default value set somewhere? Wouldn't the correct solution be to delete the line completely or set it only for dim>0?

PR #4391, opened 2021-10-29
Explicitly find zlib in cmake

gassmoeller:
I was compiling ASPECT on a cluster where zlib is not available by default (possible by module, but even then in a non-standard location). I vaguely remember in the past simply using candi to install zlib besides deal.II. I tried this and zlib was correctly picked up and included when deal.II was compiled, however when I compiled ASPECT I got the following error:
```
In file included from /home/rene.gassmoeller/software/aspect-hpg3/build/CMakeFiles/aspect.dir/Unity/unity_35_cxx.cxx:7:
/home/rene.gassmoeller/software/aspect-hpg3/source/simulator/checkpoint_restart.cc:33:12: fatal error: zlib.h: No such file or directory
 #  include <zlib.h>
            ^~~~~~~~
```
Since candi exports the ZLIB locations as environment variables I could make the attached change to CMakeLists to make it compile. 

However, wouldnt it be nice if deal.II exports the necessary information so that derived packages automatically find the include directories and have the correct link flags already? Or is this unnecessary, because most derived packages would only include deal.II but not the other libraries like zlib directly?

Anyway, if this problem had happened to a user with less understanding of cmake it would probably get them completely stuck, so I guess my question is if the attached solution is reasonable or if there is a cleaner solution.


tjhei:
I don't think we export zlib from deal.II, which is a bit of a shame. So, this makes sense. I think we could do
```
SET(ZLIB_DIR "" CACHE PATH "An optional hint to a ZLIB installation")
SET_IF_EMPTY(ZLIB_DIR "$ENV{ZLIB_DIR}")
FIND_PACKAGE(ZLIB)
INCLUDE_DIRECTORIES(${ZLIB_INCLUDE_DIR})
TARGET_LINK_LIBRARIES(${TARGET} ${ZLIB_LIBRARY})
```
instead, see https://github.com/dealii/dealii/blob/de2d25b129e8a78a45365d0622e73ebe192bd758/cmake/modules/FindZLIB.cmake#L34

gassmoeller:
Is this better? I had to modify your suggestion a bit, because `SET_IF_EMPTY` is a deal.II macro that ASPECT doesnt seem to know (cmake doesnt complain about using it either, it just does nothing). At some point I really want to understand cmake's internal logic, every time I do something because I find it intuitive cmake expects it differently.

tjhei:
That is normal when dealing with cmake. 😉

tjhei (approved):

PR #4394, opened 2021-11-03
Fix pressure depth derivative in latent heat material model

gassmoeller:
I think this fixes a bug in the latent heat material model. If you compare this way to compute the pressure_depth_derivative with the one in line 114 in the same file you notice that we actually compute the inverse here (the depth_pressure_derivative). This PR unifies the way to compute this derivative. 

The fix likely leads to a displacement of the latent heat that is generated at the phase transition. In the old version the density change of the phase transition would take place at the correct location (one that is displaced from the standard phase transition location considering the current temperature and clapeyron slope), but the latent heat change would be added at a different depth, one that is computed with a wrong Clapeyron slope of nearly 0 (the pressure_depth_derivative is large, so the inverse is small). All of this would only happen if the phase transition is defined in terms of depth instead of pressure, because otherwise the pressure_depth_derivative is not needed to compute the offset of the phase transition.

I first want to check what this changes in the tests, I will add a second commit later that refactors the whole computation to avoid having to compute the same value twice.

@jdannberg can you take a look?

### DIFF
--- a/source/material_model/latent_heat.cc
+++ b/source/material_model/latent_heat.cc
@@ -38,7 +38,7 @@ namespace aspect
         {
           const double temperature = in.temperature[i];
           const double pressure = in.pressure[i];
-          const std::vector<double> composition = in.composition[i];
+          const std::vector<double> &composition = in.composition[i];
           const Point<dim> position = in.position[i];
 
           // Assign constant material properties
@@ -80,6 +80,20 @@ namespace aspect
               out.viscosities[i] = visc_composition_dependence * visc_temperature_dependence * eta;
           }
 
+          // Prepare some values for phase dependent properties
+          const double depth = this->get_geometry_model().depth(in.position[i]);
+          const double adiabatic_pressure = this->get_adiabatic_conditions().pressure(in.position[i]);
+          const double pressure_depth_derivative = (depth > 0.0 && adiabatic_pressure > 0.0)
+                                                   ?
+                                                   adiabatic_pressure / depth
+                                                   :
+                                                   this->get_gravity_model().gravity_vector(in.position[i]).norm() * reference_rho;
+          MaterialUtilities::PhaseFunctionInputs<dim> phase_in(temperature,
+                                                               pressure,
+                                                               depth,
+                                                               pressure_depth_derivative,
+                                                               0);
+
           // Calculate density
           // and phase dependence of viscosity
           {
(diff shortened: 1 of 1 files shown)


source/material_model/latent_heat.cc
```diff
@@ -177,11 +177,11 @@ namespace aspect
               for (unsigned int phase=0; phase<phase_function.n_phase_transitions(); ++phase)
                 {
                   const double depth = this->get_geometry_model().depth(in.position[i]);
-                  const double pressure_depth_derivative = (this->get_adiabatic_conditions().pressure(position) > 0)
+                  const double pressure_depth_derivative = (depth > 0.0)
```
jdannberg, line 180: I think technically, both the depth and the adiabatic pressure need to be larger than 0 (if one of them is negative, the derivative is negative and doesn't make sense anymore).

source/material_model/latent_heat.cc
```diff
@@ -177,11 +177,11 @@ namespace aspect
               for (unsigned int phase=0; phase<phase_function.n_phase_transitions(); ++phase)
                 {
                   const double depth = this->get_geometry_model().depth(in.position[i]);
```
jdannberg, line 179: Since you changed this to `position` below, you might as well change it here too.

jdannberg (commented):
This looks good to me, and is definitely a bugfix. 
In your description, you wrote that you wanted to add a new commit, that was why I haven't looked at it earlier. But if you don't have the time to do that right now, I am also fine to merge just the bugfix. 

But I do think we need a test, since apparently none of the tests we have is affected by this.

gassmoeller:
Ok, I added the second commit (it removes the duplicated computation of the phase inputs). I also figured out why no test is affected, and why it is impossible to write a test for this. The input variable `pressure_depth_derivative` that was computed wrongly in the one of the two phase inputs before is only used in the `compute_value` function of the phase function, but it was computed wrongly for the call to `compute_derivative`. In other words, the material model computed a wrong input, but that input was never used. The new code is still cleaner and a fix, but no models should be affected. Can you take another look?

jdannberg (approved):
Nice, this looks good to me! Thank you for fixing this. 
It looks like two of the tests have changed though (which I guess shouldn't be the case?). Can you figure out what the problem is with these?

bangerth:
@gassmoeller says "Don't merge yet"!

gassmoeller:
Superseded by #6218

PR #4397, opened 2021-11-07
GMG timings option

zjiaqi2018:
Add GMG timing options: 
```
subsection Solver parameters
  subsection Matrix Free
    set Output details = true
    set Execute solver timings = true
  end
end
```
image

@tjhei

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -2188,34 +2188,35 @@ namespace aspect
 
           dealii::LinearAlgebra::distributed::BlockVector<double> tmp_dst = solution_copy;
           dealii::LinearAlgebra::distributed::BlockVector<double> tmp_src = rhs_copy;
-          time_this("Stokes_solve_cheap", 1,
-                    [&]
+          time_this("Stokes_solve_cheap_idr", 1,
+          [&]
           {
-            if (sim.parameters.stokes_krylov_type == Parameters<dim>::StokesKrylovType::gmres)
-              {
-                SolverGMRES<dealii::LinearAlgebra::distributed::BlockVector<double>>
-                solver(solver_control_cheap, mem,
-                SolverGMRES<dealii::LinearAlgebra::distributed::BlockVector<double>>::
-                AdditionalData(sim.parameters.stokes_gmres_restart_length+2,
-                true));
+            SolverIDR<dealii::LinearAlgebra::distributed::BlockVector<double>>
+            solver(solver_control_cheap, mem,
+            SolverIDR<dealii::LinearAlgebra::distributed::BlockVector<double>>::
+            AdditionalData(sim.parameters.idr_s_parameter));
 
-                solver.solve (stokes_matrix,
-                tmp_dst,
-                tmp_src,
-                preconditioner_cheap);
-              }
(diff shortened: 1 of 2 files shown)


tjhei (approved):
Thank you. This is in part written by me, so I let others review.

MFraters:
This looks like very useful development information. I am just a bit confused by the implementation. My first thought was that it would just time specific parts of your run. Looking at the code, it looks like it an internal benchmarking, and not something you would ever want to turn on for a production run. Is that correct?

Could you add a bit more explanation in both the declare entry documentation and above do_timings?

tjhei:
Not useful in production runs, correct. This is only interesting to test the scaling of the solver. But I would think this is not only to be used by my group developing the solver as it can be used to check new clusters to see how they perform.
I prefer having this inside ASPECT rather than having to use a specific branch...

MFraters:
Yes, I definitely agree that it is useful to have this kind of tooling in aspect. I would just like to see a bit more explanation in the declare entry documentation and also a comment above above do_timings function explaining the purpose of this function. Otherwise it is good to merge.

tjhei:
Sure, I agree with that.

tjhei:
@zjiaqi2018 any update on the documentation?

zjiaqi2018:
> @zjiaqi2018 any update on the documentation?

updated.

tjhei (approved):
Thanks. Looks ok to me.

gassmoeller:
I think the first part of @MFraters comment is not yet addressed: `I would just like to see a bit more explanation in the declare entry documentation ... `.  I think the crucial information here is that this is for *internal* benchmarking purposes only and that it will run synthetic operations and benchmark them. In other words this is not only adding timing information to already existing operations, but actually adding computational cost and so no one who wants to run productive models will want to enable this setting (did I understand this correctly?).

MFraters:
Thanks for adding the extra documentation, but like @gassmoeller states, I think it is important to make clear to the user that these are internal benchmarks which add (I assume) significant extra cost and that you don't want to enable this (generally?) for production runs. I think the best way of doing this is in the declare entries part, but also adding it to other parts of the documentation could be useful.

zjiaqi2018:
Thanks, I added the information to the declare entries part.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1232,6 +1232,12 @@ namespace aspect
       prm.declare_entry ("Output details", "false",
                          Patterns::Bool(),
                          "Turns on extra information for the matrix free GMG solver to be printed.");
+      prm.declare_entry ("Execute solver timings", "false",
+                         Patterns::Bool(),
+                         "Executes different parts of the Stokes solver repeatedly and print timing information."
+                         "This is for internal benchmarking purposes, it's useful if you want to see how the solver"
+                         "performs. Otherwise, you don't want to enable this, since it adds additional computatinal costs"
+                         "to get the timing information.");
```
tjhei, line 1240:
```suggestion
                         "Executes different parts of the Stokes solver repeatedly and print timing information. "
                         "This is for internal benchmarking purposes: It is useful if you want to see how the solver "
                         "performs. Otherwise, you don't want to enable this, since it adds additional computational cost "
                         "to get the timing information.");
```

tjhei (approved):

MFraters (approved):
Thanks for the changes!

PR #4398, opened 2021-11-08
Do not build particle output if no output format is selected

gassmoeller:
Our current particle postprocessor incorrectly builds the particle output patches even if no output format or the output format "none" is selected. This is wasted time. Avoid this by returning earlier if no output is going to be written.

I want to test this a bit further before merging, but reviews are welcome.

### DIFF
--- a/source/postprocess/particles.cc
+++ b/source/postprocess/particles.cc
@@ -376,12 +376,10 @@ namespace aspect
 
       for (const auto &output_format : output_formats)
         {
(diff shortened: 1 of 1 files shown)


tjhei (approved):
Looks good to me.

gassmoeller:
I did some minor refactoring and my tests look fine (I confirmed the speedup). Ready from my side.

tjhei (approved):

PR #4399, opened 2021-11-09
Use fast path for particle advection in box

gassmoeller:
This PR optimizes the particle advection in box models without mesh deformation by using the FEPointEvaluation class with MappingCartesian mappings. This was made possible by #dealii/dealii#12773 and results in a speedup of around 3x for particle advection. Non-box models and box models with mesh deformation had this speedup already before.


jdannberg (approved):
Looks good to me!

PR #4400, opened 2021-11-10
Change output format of particle test

gassmoeller:
The output of this test would be more useful in a text format (related to something I am working on separately).


tjhei (approved):
Makes sense.

PR #4401, opened 2021-11-10
[WIP] Postprocessor for "plateness"

elodie-kendall:
This PR is inspired by the parameter in Lourenco et al., 2020 which computes how plate like the lid in spherical models is. This PR is in it's early stages but the bulk part is there and we can hopefully fix the few remaining issues.

### For all pull requests:

* [ X] I have added a changelog entry in the [doc/modules/changes](../blob/master/doc/modules/changes) directory that will inform other users of my change.

### DIFF
--- a/source/postprocess/plateness_statistics.cc
+++ b/source/postprocess/plateness_statistics.cc
@@ -54,8 +54,6 @@ namespace aspect
       double local_second_invariant_of_strain_rate_integral = 0.0;
       double local_surface_area_integral = 0.0;
 
-      std::vector<Tensor<1,dim>> local_second_invariant_of_strain_rate(fe_face_values.n_quadrature_points);
-      std::vector<Tensor<1,dim>> local_surface_area(fe_face_values.n_quadrature_points); 
       std::vector<std::pair<double,double>> local_second_invariant_of_strain_rate_and_corresponding_area; 
       std::vector<SymmetricTensor<2,dim>> strain_rate (quadrature_formula.size());
  
@@ -84,24 +82,19 @@ namespace aspect
               
                 for (unsigned int q = 0; q<fe_face_values.n_quadrature_points; ++q)
                   {
-                    local_second_invariant_of_strain_rate[q] = std::max(std::sqrt(std::fabs(second_invariant(deviator(strain_rate[q])))),min_strain_rate);
-                    local_surface_area[q] = fe_face_values.JxW(q);
-
+                    local_second_invariant_of_strain_rate_and_corresponding_area[q].first = std::max(std::sqrt(std::fabs(second_invariant(deviator(strain_rate[q])))),min_strain_rate);
+                    local_second_invariant_of_strain_rate_and_corresponding_area[q].second = fe_face_values.JxW(q);
+           
                     // second invariant of strain rate over the whole surface
-                    local_second_invariant_of_strain_rate_integral += local_second_invariant_of_strain_rate[q] * local_surface_area[q];
-                    local_surface_area_integral += local_surface_area[q];
+                    local_second_invariant_of_strain_rate_integral += local_second_invariant_of_strain_rate_and_corresponding_area[q].first * local_second_invariant_of_strain_rate_and_corresponding_area[q].second;
+                    // surface area over the whole surface
+                    local_surface_area_integral += local_second_invariant_of_strain_rate_and_corresponding_area[q].second;
                   }
               }
           }
-
       const double total_second_invariant_of_strain_rate = Utilities::MPI::sum (local_second_invariant_of_strain_rate_integral, this->get_mpi_communicator());
       const double total_surface_area = Utilities::MPI::sum (local_surface_area_integral, this->get_mpi_communicator());
 
-      const unsigned int n = local_second_invariant_of_strain_rate.size();
-      // Entering values in vector of pairs
-      // i.e., merge so column 1 is strain rate invariants and column 2 are the corresponding areas
-      for (int i=0; i<n; i++)
-        local_second_invariant_of_strain_rate_and_corresponding_area.push_back( std::make_pair(local_second_invariant_of_strain_rate[i],local_surface_area[i]));
       //sort local second invariants of strain rate in size order
       std::sort(local_second_invariant_of_strain_rate_and_corresponding_area.begin(), local_second_invariant_of_strain_rate_and_corresponding_area.end());
 


source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+
+namespace aspect
+{
+  namespace Postprocess
+  {
+    template <int dim>
+    std::pair<std::string,std::string>
+    PlatenessStatistics<dim>::execute (TableHandler &statistics)
+    {
+      // Create a quadrature formula based on the velocity element.
```
tjhei, line 40: Move this comment down two lines in front of the quadrature.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+      const types::boundary_id top_boundary_id = this->get_geometry_model().translate_symbolic_boundary_name_to_id("top");
+
+      const unsigned int quadrature_degree = this->introspection().polynomial_degree.temperature;
+      const QGauss<2> quadrature_formula_face(quadrature_degree);
+
+      FEFaceValues<3> fe_face_values (this->get_mapping(),
```
tjhei, line 46: Is this postprocessor only supposed to work in 3d? I am surprised this compiles. It would be better to use "dim" here.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+                    {
+                      // If the cell is at the top boundary, assign face_idx.
+                      face_idx = f;
+                      at_upper_surface = true;
+                      break;
+                    }
+                }
+
+            }
+            fe_face_values.reinit(cell, face_idx);
```
tjhei, line 87: Only do this inside the if below. This is not needed otherwise.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+
+            if (at_upper_surface)
+              {
+                in.reinit(fe_face_values,
+                          cell,
+                          this->introspection(),
+                          this->get_solution());
+                //in.requested_properties = MaterialModel::MaterialProperties::viscosity;
```
tjhei, line 95: What is this line about?

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+                                                                                 this->get_solution(),
+                                                                                 fe_face_values,
+                                                                                 this->introspection());
+
+                this->get_material_model().evaluate(in, out);
+
+                for (unsigned int q = 0; q<fe_face_values.n_quadrature_points; ++q)
+                  {
+                    // Calculate the square root of the second moment invariant for the deviatoric strain rate tensor.
+                    //edot_ii = std::max(std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[q])))),
```
tjhei, line 107: Remove this commented code.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+
+                    // second invariant of strain rate over the whole surface
+                    local_second_invariant_of_strain_rate_integral += local_second_invariant_of_strain_rate;
+                    local_surface_area_integral += local_surface_area;
+                  }
+              }
+          }
+
+      double total_second_invariant_of_strain_rate (local_second_invariant_of_strain_rate_integral.size());
+      Utilities::MPI::sum (local_second_invariant_of_strain_rate_integral, this->get_mpi_communicator(), total_second_invariant_of_strain_rate);
```
tjhei, line 121: These two lines look incorrect: no need to initialize the double with a value. Also, doesn't sum() return the value (you are ignoring the return value)?

gassmoeller:
@tjhei: I asked @elodie-kendall to open this PR in an incomplete state to be able to leave comments directly on the code. I will add some more comments later.

tjhei:
Alright, then let us know when we should take a look.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+      std::vector<double> sorted_local_second_invariant_of_strain_rate;
+
+      const unsigned int n_q_points = quadrature_formula_face.size();
+      const unsigned int n_compositional_fields = this->n_compositional_fields();
+      //std::vector<double> compositional_values(n_q_points);
+
+      typename MaterialModel::Interface<dim>::MaterialModelInputs in(n_q_points,
+                                                                     n_compositional_fields);
+      typename MaterialModel::Interface<dim>::MaterialModelOutputs out(n_q_points,
+                                                                       n_compositional_fields);
```
gassmoeller, line 67: You somewhat abuse the MaterialModelInputs object here to store the strain rate, but you dont need all of the other properties in the object (because the material properties are not actually needed for your postprocessor). Instead you can store the strain rate in a vector of symmetric tensors directly: `std::vector<SymmetricTensor<2,dim>> strain_rate;`. This is exactly how it is stored inside of MaterialModelInputs (see `include/aspect/material_model/interface.h:411`).

source/postprocess/plateness_statistics.cc
gassmoeller: not needed, since we do not need the MaterialModel at all.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+                          this->introspection(),
+                          this->get_solution());
+                //in.requested_properties = MaterialModel::MaterialProperties::viscosity;
+
+                this->get_material_model().fill_additional_material_model_inputs(in,
+                                                                                 this->get_solution(),
+                                                                                 fe_face_values,
+                                                                                 this->introspection());
+
+                this->get_material_model().evaluate(in, out);
```
gassmoeller, line 102: No need to evaluate the material model since you dont need material properties, just the strain rate which you can compute from the solution variable velocity.

Compute the strain rate directly from the fe_face_values object like it is done inside MaterialModelInputs (`source/material_model/interface.cc:368`):
```
fe_face_values[this->introspection().extractors.velocities].get_function_symmetric_gradients (this->get_solution(),strain_rate);
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+
+                for (unsigned int q = 0; q<fe_face_values.n_quadrature_points; ++q)
+                  {
+                    // Calculate the square root of the second moment invariant for the deviatoric strain rate tensor.
+                    //edot_ii = std::max(std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[q])))),
+                    //          min_strain_rate);
+
+                    const double local_second_invariant_of_strain_rate = (std::max(std::sqrt(std::fabs(second_invariant(deviator(in.strain_rate[q])))),min_strain_rate) * fe_face_values.JxW(q));
```
gassmoeller, line 110: Dont multiply with the volume weight of this quadrature point (JxW) here, you are not integrating yet.
```suggestion
                    const double local_second_invariant_of_strain_rate = std::max(std::sqrt(std::fabs(second_invariant(deviator(strain_rate[q])))),min_strain_rate);
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+                    const double local_surface_area = fe_face_values.JxW(q);
+
+                    // second invariant of strain rate over the whole surface
+                    local_second_invariant_of_strain_rate_integral += local_second_invariant_of_strain_rate;
```
gassmoeller, line 114:
```suggestion
                    local_second_invariant_of_strain_rate_integral += local_second_invariant_of_strain_rate * local_surface_area;
```

source/postprocess/plateness_statistics.cc
gassmoeller: So in essence this should be enough:
```suggestion
      const double total_second_invariant_of_strain_rate = Utilities::MPI::sum (local_second_invariant_of_strain_rate_integral, this->get_mpi_communicator(), total_second_invariant_of_strain_rate);
```

But also note that you use the name `local_second_invariant_of_strain_rate_integral` twice, once in line 55 (as a vector) once in line 114 (where it looks like you want to use it like a scalar double). What would you like to store in this variable?

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+                    local_surface_area_integral += local_surface_area;
+                  }
+              }
+          }
+
+      double total_second_invariant_of_strain_rate (local_second_invariant_of_strain_rate_integral.size());
+      Utilities::MPI::sum (local_second_invariant_of_strain_rate_integral, this->get_mpi_communicator(), total_second_invariant_of_strain_rate);
+
+      double total_surface_area (local_surface_area_integral.size());
+      Utilities::MPI::sum (local_surface_area_integral, this->get_mpi_communicator(), total_surface_area);
```
gassmoeller, line 124:
```suggestion
      const double total_surface_area =
      Utilities::MPI::sum (local_surface_area_integral, this->get_mpi_communicator(), total_surface_area);
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+                                      update_values |
+                                      update_quadrature_points |
+                                      update_JxW_values);
+
+      // Vector to store the integral of the second invariant of the strain rate
+      // over the surface.
+      std::vector<double> local_second_invariant_of_strain_rate_integral;
+      std::vector<double> local_surface_area_integral;
+      std::vector<double> local_second_invariant_of_strain_rate_and_area;
+      std::vector<double> sorted_local_second_invariant_of_strain_rate;
```
gassmoeller, line 58: You need to rethink which of these vectors you need, and which ones actually need to be vectors and which ones are only double. Also the name `local_second_invariant_of_strain_rate_and_area` is confusing, a single vector of numbers can not contain two different things (strain rate and area)

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+
+      //Merge so column 1 is strain rate invariants and column 2 are the corresponding areas
+      for (int i = 0; i < local_second_invariant_of_strain_rate_integral.size(); ++i)
+        {
+          local_second_invariant_of_strain_rate_and_area[i][0] = local_second_invariant_of_strain_rate_and_area[i];
+          local_second_invariant_of_strain_rate_and_area[i][1] = local_surface_area_integral[i];
+        }
+
+      //sort local second invariants of strain rate in size order
+      sorted_local_second_invariant_of_strain_rate = sort(local_second_invariant_of_strain_rate[:][0].begin(), local_second_invariant_of_strain_rate[:][0].end(), greater<int>());
```
gassmoeller, line 135: `vector[:]` is syntax from Python, but not valid C++. What do you try to do? If you want to sort `local_second_invariant_of_strain_rate_and_area` it should look like this: 
```
```suggestion
      std::sort(local_second_invariant_of_strain_rate.begin(), local_second_invariant_of_strain_rate.end());
```
However you want to keep the strain rate and area of each cell associated, correct? Then I would suggest you store them in a data structure of type `std::vector<std::pair<double,double>>`. std::sort should understand automatically that it should sort by the first of the two pair members (see https://www.geeksforgeeks.org/sorting-vector-of-pairs-in-c-set-1-sort-by-first-and-second/).

Also this is the part of the algorithm that will not work in parallel unless you communicate every vector entry from every MPI process with one MPI process (e.g. process 0). Your `MPI::sum` command above does not do that, you would need to use an `MPI::gather`. For now let's ignore this part and make the algorithm work in serial computations.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+                                      quadrature_formula_face,
+                                      update_values |
+                                      update_quadrature_points |
+                                      update_JxW_values);
+
+      // Vector to store the integral of the second invariant of the strain rate
+      // over the surface.
+      std::vector<double> local_second_invariant_of_strain_rate_integral;
+      std::vector<double> local_surface_area_integral;
+      std::vector<double> local_second_invariant_of_strain_rate_and_area;
```
gassmoeller, line 57: I think you want this to be `std::vector<std::pair<double,double>>`. Then you can access the first and second member of the i'th pair like this `local_second_invariant_of_strain_rate_and_area[i].first` and `local_second_invariant_of_strain_rate_and_area[i].second`.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+      sorted_local_second_invariant_of_strain_rate = sort(local_second_invariant_of_strain_rate[:][0].begin(), local_second_invariant_of_strain_rate[:][0].end(), greater<int>());
+
+      //find the fraction of the surface area where 80% of the deformation occurs
+      //double f_80 = 0.0;
+      double cumulative_second_invariant_of_strain_rate = 0.0;
+      double cumulative_surface_area = 0.0;
+      int row_number = 0;
+      do
+        {
+          cumulative_second_invariant_of_strain_rate = cumulative_second_invariant_of_strain_rate + local_second_invariant_of_strain_rate_and_area[row_number][0];
```
gassmoeller, line 144: This type of loop is somewhat dangerous. If the while condition is not well defined you may end up accessing memory beyond the end of local_second_invariant_of_strain_rate_and_area. Better add this, which will cause ASPECT to crash with an error message if we ever not meet the while condition.
```suggestion
      Assert(row_number < local_second_invariant_of_strain_rate_and_area.size(), ExcInternalError());
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+          cumulative_surface_area = cumulative_surface_area + local_second_invariant_of_strain_rate_and_area[row_number][1];
+          row_number = row_number + 1;
+        }
+      while (cumulative_second_invariant_of_strain_rate<0.8*total_second_invariant_of_strain_rate);
+      double f_80 = cumulative_surface_area/total_surface_area;
+
+      //Compute plateness
```
gassmoeller, line 151: Add a comment here that the specific values of 0.8 and 0.6 are taken from Lourenco et al 2020. We could in the future make that an input parameter (in case someone wants to compute with different values), but for now explaining where they come from is enough.

gassmoeller (commented):
Some first comments (in addition to Timo's comments that also apply). As I mentioned this algorithm can only work in serial models for now, but let's get this to work first before thinking about the parallel case. Leave comments if some of my points are not clear.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,196 @@
...
+  namespace Postprocess
+  {
+    template <int dim>
+    std::pair<std::string,std::string>
+    PlatenessStatistics<dim>::execute (TableHandler &statistics)
+    {
+      // Create a quadrature formula based on the velocity element.
+      const types::boundary_id top_boundary_id = this->get_geometry_model().translate_symbolic_boundary_name_to_id("top");
+
+      const unsigned int quadrature_degree = this->introspection().polynomial_degree.temperature;
```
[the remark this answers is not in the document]
elodie-kendall: Thanks for the explanation:)

source/postprocess/plateness_statistics.cc
elodie-kendall: Yes, great

source/postprocess/plateness_statistics.cc
elodie-kendall: Sorry this should be a double. it should be "local_second_invariant_of_strain_rate * local_surface_area" integrated over the whole surface.

source/postprocess/plateness_statistics.cc
elodie-kendall: Yes, thanks. I have now set "local_second_invariant_of_strain_rate_integral" and "local_surface_area_integral" to be purely doubles. They are the second invariant of strain rate over the whole surface and total area, respectively. But I now save each local value into the vectors: "local_second_invariant_of_strain_rate[q]" and "local_surface_area[q]" so prepare for the cumulative addition loop for reaching f_80. This "local_second_invariant_of_strain_rate_and_corresponding_area" should contain "local_second_invariant_of_strain_rate" in column 1 and "local_surface_area" in column 2 so that we can re-order the each row based on the size in column 1.

source/postprocess/plateness_statistics.cc
elodie-kendall: I see, thanks for the explanation.

source/postprocess/plateness_statistics.cc
elodie-kendall: Ah yes, thanks, that's much simpler!

source/postprocess/plateness_statistics.cc
elodie-kendall: Ok thanks. I assume you don't mean to remove line 144 also "cumulative_second_invariant_of_strain_rate = cumulative_second_invariant_of_strain_rate + local_second_invariant_of_strain_rate_and_area[row_number][0];". We still need that.

elodie-kendall:
Thanks a lot for your comments, @tjhei, @gassmoeller. I've addressed them all, I'm just stuck with these last two issues:

tjhei:
Elodie, try reading the error messages. The first one says paraphrased "You are trying to add a double and a Tensor. I don't know know how to do that." That of course doesn't make any sense. Agreed?

bangerth:
@elodie-kendall What help do you need to get this pushed over the finish line?

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+      const QGauss<dim-1> quadrature_formula (this->introspection().polynomial_degree.velocities+1);
+   
+      FEFaceValues<dim> fe_face_values (this->get_mapping(),
+                                        this->get_fe(),
+                                        quadrature_formula,
+                                        update_gradients |
+                                        update_quadrature_points |
+                                        update_JxW_values);
+         
+      std::vector<Tensor<1,dim>> velocities (fe_face_values.n_quadrature_points);
```
bangerth, line 52: I think you don't actually use this variable and can remove it.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+
+      // Vector to store the integral of the second invariant of the strain rate
+      // over the surface.
+      double local_second_invariant_of_strain_rate_integral = 0.0;
+      double local_surface_area_integral = 0.0;
+
+      std::vector<Tensor<1,dim>> local_second_invariant_of_strain_rate(fe_face_values.n_quadrature_points);
+      std::vector<Tensor<1,dim>> local_surface_area(fe_face_values.n_quadrature_points);
```
bangerth, line 60: Your error is due to the fact that this should probably be
```suggestion
      std::vector<double> local_surface_area(fe_face_values.n_quadrature_points); 
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+      std::vector<std::pair<double,double>> local_second_invariant_of_strain_rate_and_corresponding_area; 
+      std::vector<SymmetricTensor<2,dim>> strain_rate (quadrature_formula.size());
+ 
+      for (const auto &cell : this->get_dof_handler().active_cell_iterators())
+        if (cell->is_locally_owned() && cell->at_boundary())
+          {
+            unsigned int face_idx = numbers::invalid_unsigned_int;
+            bool at_upper_surface = false;
+            {
+              for (unsigned int f=0; f<GeometryInfo<3>::faces_per_cell; ++f)
```
bangerth, line 70:
```suggestion
              for (unsigned int f=0; f<cell->n_faces(); ++f)
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+                {
+                  if (cell->at_boundary(f) && cell->face(f)->boundary_id() == top_boundary_id)
```
bangerth, line 72:
```suggestion
                  if (cell->at_boundary(f) && (cell->face(f)->boundary_id() == top_boundary_id))
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+                fe_face_values[this->introspection().extractors.velocities].get_function_symmetric_gradients (this->get_solution(), strain_rate);
+              
+                for (unsigned int q = 0; q<fe_face_values.n_quadrature_points; ++q)
+                  {
+                    //Merge so column 1 is strain rate invariants and column 2 are the corresponding areas
+                    local_second_invariant_of_strain_rate[q] = std::max(std::sqrt(std::fabs(second_invariant(deviator(strain_rate[q])))),min_strain_rate);
+                    local_surface_area[q] = fe_face_values.JxW(q);
+                    
+                    // second invariant of strain rate over the whole surface
+                    local_second_invariant_of_strain_rate_integral = local_second_invariant_of_strain_rate[q] * local_surface_area[q];
```
bangerth, line 94: Did you mean this to be a `+=` here?

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+                    local_surface_area_integral += local_surface_area[q];
```
bangerth, line 95: This is not inherently wrong, but it's inefficient. You are in essence packing information into the `local_second_invariant_of_strain_rate` and `local_surface_area` arrays by writing into the `[q]`th entries in the first two lines. In the second set of two lines, you then use the information so stored, but after that, you never access the entries of these arrays again and simply overwrite these values on the next cell that happens to be on the boundary.

A more efficient way would be to just not put things into an array to begin with, but instead do something like the following:
```suggestion
                    const double local_second_invariant_of_strain_rate = std::max(std::sqrt(std::fabs(second_invariant(deviator(strain_rate[q])))),min_strain_rate);
                    const double local_surface_area[q] = fe_face_values.JxW(q);
                    
                    // second invariant of strain rate over the whole surface
                    local_second_invariant_of_strain_rate_integral += local_second_invariant_of_strain_rate * local_surface_area;
                    local_surface_area_integral += local_surface_area;
```
Then you can get rid of the array declarations above.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+              }
+          }
+
+      const double total_second_invariant_of_strain_rate = Utilities::MPI::sum (local_second_invariant_of_strain_rate_integral, this->get_mpi_communicator());
+      const double total_surface_area = Utilities::MPI::sum (local_surface_area_integral, this->get_mpi_communicator());
+
+      int n = sizeof(local_second_invariant_of_strain_rate)/sizeof(local_second_invariant_of_strain_rate[0]);
+      // Entering values in vector of pairs
+      for (int i=0; i<n; i++)
+        local_second_invariant_of_strain_rate_and_corresponding_area.push_back( std::make_pair(local_second_invariant_of_strain_rate[i],local_surface_area[i]));
```
bangerth, line 106: This cannot be a useful operation. The contents of `local_second_invariant_of_strain_rate[i]` and `local_surface_area[i]` are simply what you evaluated at the quadrature point of the last cell you visited. Which cell that is (or whether you visited any cell at all) is arbitrary, so I don't think that these lines do what you think they do. Can you explain in words what your goal is?

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+      int row_number = 0;
+      do
+        {
+          Assert(row_number < local_second_invariant_of_strain_rate_and_area.size(), ExcInternalError());
+          cumulative_second_invariant_of_strain_rate = cumulative_second_invariant_of_strain_rate + local_second_invariant_of_strain_rate_and_corresponding_area[row_number].first * local_second_invariant_of_strain_rate_and_corresponding_area[row_number].second;
+          cumulative_surface_area = cumulative_surface_area + local_second_invariant_of_strain_rate_and_corresponding_area[row_number].second;
+          row_number = row_number + 1;
+        }
+      while (cumulative_second_invariant_of_strain_rate<0.8*total_second_invariant_of_strain_rate);
+      double f_80 = cumulative_surface_area/total_surface_area;
```
bangerth, line 123:
```suggestion
      const double f_80 = cumulative_surface_area/total_surface_area;
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+
+      // Compute plateness
+      // As explained in Tackely (2000) and Lourenco et al., 2020 for an isoviscous, 
+      // internally heated calculation with Ra~10^6, 80% of the surface deformation
+      // occurs in 60% of the surface area. Therefore we express plateness as described below:
+      double plateness = 0.0;
+      plateness = 1 - f_80/0.6;
```
bangerth, line 130:
```suggestion
      const double plateness = 1 - f_80/0.6;
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+          Assert(row_number < local_second_invariant_of_strain_rate_and_area.size(), ExcInternalError());
+          cumulative_second_invariant_of_strain_rate = cumulative_second_invariant_of_strain_rate + local_second_invariant_of_strain_rate_and_corresponding_area[row_number].first * local_second_invariant_of_strain_rate_and_corresponding_area[row_number].second;
+          cumulative_surface_area = cumulative_surface_area + local_second_invariant_of_strain_rate_and_corresponding_area[row_number].second;
+          row_number = row_number + 1;
+        }
+      while (cumulative_second_invariant_of_strain_rate<0.8*total_second_invariant_of_strain_rate);
+      double f_80 = cumulative_surface_area/total_surface_area;
+
+      // Compute plateness
+      // As explained in Tackely (2000) and Lourenco et al., 2020 for an isoviscous,
```
bangerth, line 126:
```suggestion
      // As explained in Tackley (2000) and Lourenco et al., 2020 for an isoviscous, 
```

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+      // As explained in Tackely (2000) and Lourenco et al., 2020 for an isoviscous, 
+      // internally heated calculation with Ra~10^6, 80% of the surface deformation
+      // occurs in 60% of the surface area. Therefore we express plateness as described below:
+      double plateness = 0.0;
+      plateness = 1 - f_80/0.6;
+
+      statistics.add_value ("Plateness ",
+                            plateness);
+
+      const std::string column = "Plateness ";
```
bangerth, line 135: Let's remove the space at the end of the strings.

bangerth (commented):
Here are some more comments. I don't think the implementation is correct yet, but I can see how you will get there!

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+namespace aspect
+{
+  namespace Postprocess
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(PlatenessStatistics,
+                                  "plateness statistics",
+                                  "A postprocessor that outputs the plateness. "
+                                  "We define plateness following lourenco et al., 2020 and "
+                                  "based on the works by Weinstein and Olson (1992) "
+                                  "and Tackley (2000). The square root of the second invariant of "
```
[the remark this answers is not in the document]
elodie-kendall: Sure

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,182 @@
...
+                                  "strain rate, $\\dot{\\sigma_surf} = \\frac{\\dot{\\sigma_\\phi\\phi}}{\\sqrt(2)}$ "
```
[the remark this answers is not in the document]
elodie-kendall: Thanks, I removed the \phi's (which were from eq.7 Lourenco et al., 2020) and replaced it with notation for the sec inv of strain rate so \varepsilon_{ij}.

source/postprocess/plateness_statistics.cc
elodie-kendall: Yes, thank you:)

source/postprocess/plateness_statistics.cc
elodie-kendall: Yes, thanks for catching this!

source/postprocess/plateness_statistics.cc
elodie-kendall: In the "for (unsigned int q = 0; q<fe_face_values.n_quadrature_points; ++q)" statement above this "for" loop I wanted to save the right value for each quadrature point into the following vectors "local_second_invariant_of_strain_rate" and "local_surface_area". So for the first quadrature point has it's values in the first row of the vectors above etc. Then I assumed that the "for (int i=0; i<n; i++)" statement you describe here- would just put these two vectors (local_second_invariant_of_strain_rate and local_surface_area) together side by side so it is a 2 column vector. This can then be sorted into smallest strains first and accumulated until the f_80 is reached. Is that not ok?

elodie-kendall:
Thanks for your reminder @bangerth. I've addressed your comments and it now compiles. I was thinking to add an example with the burnman_test.prm, that's ok for you?

bangerth:
Yes, examples are always good, as they make sure that what works today also works tomorrow!

As for the merge conflict: You added an entry to the `.bib` file, but someone else did the same in a patch that was merged after you proposed yours. And because both add an entry to the end of the file, git says "these patches conflict". You'll have to resolve this "merge conflict" before this can be merged.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,172 @@
...
+                    // surface area over the whole surface
+                    local_surface_area_integral += local_second_invariant_of_strain_rate_and_corresponding_area[q].second;
+                  }
+              }
+          }
+      const double total_second_invariant_of_strain_rate = Utilities::MPI::sum (local_second_invariant_of_strain_rate_integral, this->get_mpi_communicator());
+      const double total_surface_area = Utilities::MPI::sum (local_surface_area_integral, this->get_mpi_communicator());
+
+      //sort local second invariants of strain rate in size order
+      std::sort(local_second_invariant_of_strain_rate_and_corresponding_area.begin(), local_second_invariant_of_strain_rate_and_corresponding_area.end());
```
bangerth, line 99: This can still not be right. The code above basically boils down to the following (some snippets omitted):
```
  for (cell = ...)
  {
    if (something interesting is the case with 'cell')
      fill array 'local_second_invariant_of_strain_rate_and_corresponding_area' based on the quadrature points on 'cell'
  }
  ...
  std::sort (local_second_invariant_of_strain_rate_and_corresponding_area.begin(),
                 local_second_invariant_of_strain_rate_and_corresponding_area.end());
```
When you get to the call to `std::sort`, what you are sorting are the values computed on the quadrature points *of the last cell you visited*. You are not sorting the values computed on all of the cells, which I believe is what you really wanted.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,172 @@
...
+
+            }
+            if (at_upper_surface)
+              {
+                fe_face_values.reinit(cell, face_idx);
+                fe_face_values[this->introspection().extractors.velocities].get_function_symmetric_gradients (this->get_solution(), strain_rate);
+              
+                for (unsigned int q = 0; q<fe_face_values.n_quadrature_points; ++q)
+                  {
+                    local_second_invariant_of_strain_rate_and_corresponding_area[q].first = std::max(std::sqrt(std::fabs(second_invariant(deviator(strain_rate[q])))),min_strain_rate);
```
bangerth, line 85: This will probably compile but not run. Note how you are not setting the size of  `local_second_invariant_of_strain_rate_and_corresponding_area` at the top, so it's an empty array. Indexing the `[q]`'th element is likely going to give you a segmentation fault.

source/postprocess/plateness_statistics.cc
```diff
@@ -0,0 +1,172 @@
...
+      double cumulative_surface_area = 0.0;
+      int row_number = 0;
+      do
+        {
+          Assert(row_number < local_second_invariant_of_strain_rate_and_area.size(), ExcInternalError());
+          cumulative_second_invariant_of_strain_rate = cumulative_second_invariant_of_strain_rate + local_second_invariant_of_strain_rate_and_corresponding_area[row_number].first * local_second_invariant_of_strain_rate_and_corresponding_area[row_number].second;
+          cumulative_surface_area = cumulative_surface_area + local_second_invariant_of_strain_rate_and_corresponding_area[row_number].second;
+          row_number = row_number + 1;
+        }
+      while (cumulative_second_invariant_of_strain_rate<0.8*total_second_invariant_of_strain_rate);
```
bangerth, line 113: As a consequence of the comment above, `local_second_invariant_of_strain_rate_and_corresponding_area` has length zero, and so the assertion in line 108 will fail in the first go-around already. Based on how you index the array via ` local_second_invariant_of_strain_rate_and_corresponding_area[q]`, one could think that you really wanted the array to have size `n_q_points`, but that's not going to work either: I *think* what you need is an array with as many rows as there are cells, so that you can select a number of cells that together make up a certain fraction of the total plateness.

bangerth (changes requested):
@elodie-kendall Here are a few more comments.

But I think that in truth, the algorithm you want isn't actually this easy to implement. If I understand right what this postprocessor is supposed to do is "choose those cells that together make up 80% of the plateness". For this to work, you will need an array that has as many entries as there are cells -- but you can't store such an array on a single processor, and instead need to let each processor store its own fraction of the cell. This means that you have to run the algorithm to select which cells you want to tag without being able to sort *all* of the cells.

Such an algorithm can be written. You need something along the lines of what's described in Section 5.1 of https://www.math.colostate.edu/~bangerth/publications/2010-distributed.pdf .

bangerth:
@elodie-kendall What should we do with this pull request?

PR #4402, opened 2021-11-10
Modernize particle generation interface

gassmoeller:
This PR rewrites our particle generation interface in a backward compatible way. The new interface and all the included plugins now need much less copies of particle objects and use the deal.II particle generation functions, which reduces code duplication.
Once dealii/dealii#12931 is merged we can do some more optimizations for some plugins. In some of my tests I have seen speedups of particle generation of up to 5x with the new interface (although generation is of course a small part of the overall cost).

All particle tests work as before, except for:
- tests that add or remove particles after initialization do so with their own random number generator that is now different from the one used to generate the particles initially. This leads to different locations, and also different number of particles (because the removed/inserted particles now cause different change later)
- There was a bug in the random uniform particle generator which would create incorrect particle distributions (see changelog entry). Fixing this bug changed the output of particle_periodic_quarter_shell.


bangerth:
Can you rebase this?

bangerth (commented):
Nice work. Just small stuff!

gassmoeller:
Thanks for the review. I think I addressed all comments. Sadly leaving `DEAL_II_DEPRECATED` at `using` directives does not work, but I left a `TODO` at all of the places that includes the information when we can remove the directive.

gassmoeller:
@bangerth could you take another look? Thanks

bangerth (approved):

bangerth:
Thank you!

PR #4403, opened 2021-11-10
Keep constrained dofs in composition block

gassmoeller:
@anne-glerum: First baby step towards being able to disable boundary conditions for some compositional fields by keeping constrained dofs in the sparsity pattern. @tjhei and @bangerth is this what we talked about on Tuesday?

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -1046,14 +1046,65 @@ namespace aspect
                                               face_coupling,
                                               Utilities::MPI::
                                               this_mpi_process(mpi_communicator));
+
+        if (solver_scheme_solves_advection_equations(parameters)
+            &&
+            compositional_fields_need_matrix_block(introspection))
+          {
+            // If we solve for more than one compositional field make sure we keep constrained entries
+            // to allow different boundary conditions for different fields.
+            Table<2,DoFTools::Coupling> composition_coupling(introspection.n_components,
+                                                             introspection.n_components);
+            composition_coupling.fill (DoFTools::none);
+
+            const unsigned int component = introspection.component_indices.compositional_fields[0];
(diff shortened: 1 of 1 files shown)


source/simulator/core.cc
```diff
@@ -1046,14 +1046,65 @@ namespace aspect
                                               face_coupling,
                                               Utilities::MPI::
                                               this_mpi_process(mpi_communicator));
+
+        if (solver_scheme_solves_advection_equations(parameters)
+            &&
+            compositional_fields_need_matrix_block(introspection))
+          {
+            // If we solve for more than one compositional field make sure we keep constrained entries
+            // to allow different boundary conditions for different fields.
```
bangerth, line 1055: It took me a good long while to understand why you don't just merge the call to `make_flux_sparsity_pattern()` into the one above. Can you add a comment here or at the actual call site below mentioning that the point is that in the call here, "keeping constrained entries" implies passing `true` as fourth argument?

source/simulator/core.cc
```diff
@@ -1046,14 +1046,65 @@ namespace aspect
...
-                                       this_mpi_process(mpi_communicator));
+      {
+        DoFTools::make_sparsity_pattern (dof_handler,
+                                         coupling, sp,
+                                         current_constraints, false,
+                                         Utilities::MPI::
+                                         this_mpi_process(mpi_communicator));
+
+        // If we solve for more than one compositional field make sure we keep constrained entries
+        // to allow different boundary conditions for different fields.
```
bangerth, line 1085: Same here

bangerth (commented):
OK with two small additions to the commentary.

gassmoeller:
I had to spend a few minutes to remind myself again. I extended the comment.

bangerth (approved):

PR #4405, opened 2021-11-11
Switch tester to 9.3.2

tjhei:
Update docker image for Jenkins to 9.3.2

TODO: update github actions once new ubuntu packages are released.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -17,7 +17,7 @@ jobs:
     - name: setup
       run: |
         sudo apt-get install -y software-properties-common
-        sudo add-apt-repository ppa:ginggs/deal.ii-9.3.0-backports
+        sudo add-apt-repository ppa:ginggs/deal.ii-9.3.2-backports
         sudo apt-get update
         sudo apt-get install -yq --no-install-recommends libdeal.ii-dev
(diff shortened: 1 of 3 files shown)


tjhei:
@gassmoeller could you please generate the new image? I would like to move the testers to 9.3.2 so that I can use the bugfixes relevant for mesh deformation.

gassmoeller:
I started the build, will let you know when it is done and pushed. Can you also update `contrib/docker/tester/build.sh` to tag the image with the new version?

gassmoeller:
The new image is build and pushed. Let's see what the tester says.

/rebuild

tjhei:
"Found deal.II version 9.3.2". :+1:

tjhei:
Maybe we should wait for the new ubuntu packages so that we can update the github testers as well. Should only take a couple of days, hopefully.

bangerth:
What should we do here? It looks good to me.

tjhei:
Ubuntu packages are not updated to 9.3.2 so we can not update the github actions without installing manually.

@gassmoeller Should we consider switching github actions to use our own image?

tjhei:
Updated to 9.3.2. If everything passes, we can merge this.

PR #4406, opened 2021-11-12
Respect manifold for heat flux map postprocessor output

gassmoeller:
A minor fix that makes sure that the points written by the heat_flux_map postprocessor actually lie on the surface of the domain (instead of being only linearly interpolated between the vertex locations of the boundary face). This makes no difference for box models, but is important for curved geometries like the spherical shell. The same fix was necessary for the dynamic topography postprocessor in #3370, but I did not check other postprocessors for the same problem at the time.

### DIFF
--- a/source/postprocess/heat_flux_map.cc
+++ b/source/postprocess/heat_flux_map.cc
@@ -449,15 +449,19 @@ namespace aspect
       std::vector<std::pair<Point<dim>,double>> stored_values;
 
       // loop over all of the surface cells and evaluate the heat flux
+      const types::boundary_id top_boundary_id = this->get_geometry_model().translate_symbolic_boundary_name_to_id("top");
+      const types::boundary_id bottom_boundary_id = this->get_geometry_model().translate_symbolic_boundary_name_to_id("bottom");
+
       for (const auto &cell : this->get_dof_handler().active_cell_iterators())
         if (cell->is_locally_owned())
           for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
             if (cell->at_boundary(f) &&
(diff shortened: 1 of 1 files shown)


tjhei:
Makes sense. To me it looks like there are a few other places as well:
```
$ git grep -e '->center()'
doc/manual/manual.tex:        if (cell->face(f)->center()[0] == -1)
source/particle/generator/probability_density_function.cc:        // using cell->center() and cell->measure() would be equivalent. This
source/particle/interpolator/nearest_neighbor.cc:                    const double dist = (positions[pos_idx] - neighbors[i]->center()).norm_square();
source/postprocess/boundary_densities.cc:              if (cell->at_boundary(f) && this->get_geometry_model().depth (cell->face(f)->center())
source/postprocess/boundary_densities.cc:              if (cell->at_boundary(f) && this->get_geometry_model().depth (cell->face(f)->center())
source/postprocess/boundary_pressures.cc:              if (cell->at_boundary(f) && this->get_geometry_model().depth (cell->face(f)->center())
source/postprocess/boundary_pressures.cc:              if (cell->at_boundary(f) && this->get_geometry_model().depth (cell->face(f)->center())
source/postprocess/heat_flux_map.cc:                const Point<dim> midpoint_at_surface = cell->face(f)->center();
source/simulator/helper_functions.cc:                      (geometry_model->depth (face->center()) <
source/simulator/stokes_matrix_free.cc:                  face->get_manifold().normal_vector(face, face->center());
tests/global_refine_amr.cc:        if (cell->is_locally_owned() && cell->center()[0]<0.5)
```

tjhei (approved):

gassmoeller:
I checked the other places, but all of them use it in a different way where accuracy does not matter so they are technically correct. I agree many of them should be replaced with a better implementation, but it is a slightly different issue.

PR #4407, opened 2021-11-12
Improve boundary checks in a number of places

gassmoeller:
Follow up to #4406 to improve the check for top/bottom boundary faces and more appropriate use of the manifold to compute centers of faces and cells. I did not explain the parameter more in the manual, because this is way too specific for the section it is used in, and it is just an example code sample, not something the user is expected to copy and use.


tjhei (approved):

tjhei:
can you check and adjust test results? Ready to go otherwise.

gassmoeller:
I have updated the tests. The changes were surprising to me, but they are caused by the improvements to the normalize_pressure function when normalizing the surface pressure. It seems the old check did miss some surface faces if the cells are coarse and of a bad aspect ratio. E.g. the average surface pressure in the merged_chunks_3d test went down from -2.841e+08 Pa to 0.01564 Pa in this PR.

tjhei:
Fascinating. Would it make sense to add a changelog entry about this fact?

tjhei (approved):

gassmoeller:
Yes it would. Changelog entry added.

tjhei:
great, thank you.

PR #4409, opened 2021-11-13
Use the new stabilization interface in entropy benchmark

jdannberg:
This is a follow-up to #3497. I noticed that the stabilization interface was not actually being used in the benchmark that solves the entropy equation, which is the test case that needs different terms in the stabilization. 
Specifically, the entropy equation has the prefactors density * temperature in the advcetion term, and conductivity * temperature / specific_heat in the conduction term. This pull request implements that for the benchmark. 

### For all pull requests:


gassmoeller (commented):
The code changes look good, but you have to update the test results. Thanks for fixing my oversight.

jdannberg:
Should be fixed now. Is the change in test results about what you would expect?

gassmoeller (approved):
Yes the change in test results look reasonable. Thanks for the fix.

PR #4410, opened 2021-11-17
Fixed an error message for pre-compiled headers.

alarshi:
This PR fixes for an an error message in pre-compiled headers.

### For all pull requests:

### DIFF
--- a/source/particle/world.cc
+++ b/source/particle/world.cc
@@ -33,6 +33,7 @@
 #include <deal.II/grid/grid_tools.h>
 
 #include <deal.II/matrix_free/fe_point_evaluation.h>
+#include <deal.II/fe/mapping_cartesian.h>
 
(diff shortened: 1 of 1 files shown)


gassmoeller (approved):
Looks good thanks for the fix. 

To provide a bit more context: Since #2714 we use MappingCartesian inside world.cc without specfically including the correct header. This is not a problem if precompiling headers (because all headers are merged into the precompiled file), or if using deal.II 9.3 (because it is removed by the precompiler) or a recent deal.II dev (because then it is implicitly included in fe_point_evaluation.h). However for deal.II dev versions between 9.3 and https://github.com/dealii/dealii/pull/12773 if not precompiling headers MappingCartesian is not known. And we should include it explicitly anyway since we cannot rely on deal.II including it for us.

PR #4411, opened 2021-11-18
Mesh deformation: get the free surface stabilization parameter

zjiaqi2018:
Add a function to get the stabilization parameter for the free surface, which is required by the matrix-free implementation. @tjhei


tjhei (approved):
Good.

tjhei:
This probably need to change with #4350, but we will get there...

PR #4413, opened 2021-11-21
nullspace rotation constraints

tjhei:
Add constraints for systems with a rotational nullspace similar to what
we do for translation. While it is not strictly required for AMG (we
have done without), it is required for GMG to work well.


source/simulator/nullspace.cc
```diff
@@ -19,22 +19,134 @@
...
+                                                 quadrature,
+                                                 update_quadrature_points |
+                                                 update_normal_vectors);
+
+      for (const auto &cell : dof_handler.active_cell_iterators())
+        if (cell->is_ghost() || cell->is_locally_owned())
+          for (unsigned int face_no = 0;
+               face_no < GeometryInfo<dim>::faces_per_cell; ++face_no)
+            {
+              typename DoFHandler<dim, spacedim>::face_iterator face = cell->face(face_no);
```
gassmoeller, line 73: Somewhere here you need to check if you are at a boundary.

source/simulator/nullspace.cc
```diff
@@ -19,22 +19,134 @@
...
+              face->get_dof_indices(face_dofs);
+              fe_face_values.reinit(cell, face_no);
+
+              for (unsigned int i = 0; i < face_dofs.size(); ++i)
+                {
+                  const Point<dim> position = fe_face_values.quadrature_point(i);
+                  if (position.distance(location) < 1e-6*cell->diameter())
```
gassmoeller, line 80: This check assumes the position can be found. Better check after the loop that it indeed was found.

source/simulator/nullspace.cc
```diff
@@ -108,6 +220,56 @@ namespace aspect
...
   {
+
+    if (parameters.nullspace_removal & Parameters<dim>::NullspaceRemoval::any_rotation)
+      {
+        const bool is_spherical = Plugins::plugin_type_matches<const GeometryModel::SphericalShell<dim>>(*geometry_model)
+                                  || Plugins::plugin_type_matches<const GeometryModel::Sphere<dim>>(*geometry_model);
+
+        Assert(is_spherical,
+               ExcNotImplemented("nullspace constraints for rotation are currently only implemented Sphere and "
+                                 "SphericalShell geometries."));
```
gassmoeller, line 231: This will probably trigger warnings in release model that `is_spherical` is not used.

source/simulator/nullspace.cc
```diff
@@ -108,6 +220,56 @@ namespace aspect
...
   void Simulator<dim>::setup_nullspace_constraints(AffineConstraints<double> &constraints)
   {
+
+    if (parameters.nullspace_removal & Parameters<dim>::NullspaceRemoval::any_rotation)
+      {
+        const bool is_spherical = Plugins::plugin_type_matches<const GeometryModel::SphericalShell<dim>>(*geometry_model)
+                                  || Plugins::plugin_type_matches<const GeometryModel::Sphere<dim>>(*geometry_model);
+
+        Assert(is_spherical,
+               ExcNotImplemented("nullspace constraints for rotation are currently only implemented Sphere and "
```
gassmoeller, line 230:
```suggestion
               ExcNotImplemented("Nullspace constraints for rotation are currently only implemented for the Sphere and "
```

source/simulator/nullspace.cc
```diff
@@ -108,6 +220,56 @@ namespace aspect
...
+            mask.set(1, true);
+            internal::constrain_point_on_boundary(dof_handler,
+                                                  *mapping,
+                                                  location1,
+                                                  mask,
+                                                  constraints);
+
+            // point 2 is in positive x direction, constrain y
+            Point<dim> location2;
+            location2[0] = location1[dim-1];
```
gassmoeller, line 263: This will lead to problems in deformed meshes., does it not? Maybe we need to disable this when mesh deformation is active. Or we need to rethink the approach of using known points to calculate the rotational directions.

source/simulator/nullspace.cc
```diff
@@ -19,22 +19,134 @@
...
+      const std::vector<Point<dim - 1>> &unit_support_points = fe.get_unit_face_support_points();
+      const Quadrature<dim - 1> quadrature(unit_support_points);
+      const unsigned int dofs_per_face = fe.dofs_per_face;
+      std::vector<types::global_dof_index> face_dofs(dofs_per_face);
+
+      FEFaceValues<dim, spacedim> fe_face_values(mapping,
+                                                 fe,
+                                                 quadrature,
+                                                 update_quadrature_points |
+                                                 update_normal_vectors);
```
gassmoeller, line 66: I dont see where you need the normal vectors here or in the function below.

source/simulator/nullspace.cc
```diff
@@ -206,6 +368,65 @@ namespace aspect
   }
 
 
+  template <int dim>
+  void Simulator<dim>::setup_nullspace_constraints(AffineConstraints<double> &constraints,
```
gassmoeller, line 372: Same general comments for this function as for the one above

gassmoeller (commented):
You were lucky I was setting up a PR myself right now. Express turnaround time on the review ;-). Really nice to handle the nullspace constraints uniformly for AMG and GMG.

I have three concerns:
1. I dont see where you limit your search to the boundary cells and faces. In the current form it looks like you loop over all faces which is probably slow. Can you check for cell->at_boundary and/or face->boundary_id?
2. Currently the internal function will silently exit without constraining anything if the given location could not be found, I think we should crash in that case. It is not currently an issue, but just to be safe in case someone messes around with the geometry models and you assumed points not longer are at the surface.
3. Does this work if the mesh is deformed?

tjhei:
@gassmoeller, all good points. Thank you for the quick review!

I addressed everything except:
> Does this work if the mesh is deformed?

Good question, probably not. It looks like ``representative_point`` for shells/spheres does not take deformation into account, so the point will be in the interior.
But does it make sense to do mesh deformation and remove the rotational nullspace? We won't have a perfect sphere, so the solution is likely already unique.

Option 1: Assert that rotational nullspace + mesh deformation is ExcNotImplemented()
Option 2: Disable this new feature if mesh deformation is enabled and wait until someone has convergence problems. :-)
Option 3: Do not use the Mapping when finding the points and hope that we still find them and that everything works. Maybe actually test it...

What do you think? I would like to avoid having to set this problem up to test (Option 3.). I would probably pick 2.

tjhei:
> But does it make sense to do mesh deformation and remove the rotational nullspace?

The answer is yes: tests/geoid_freesurface.prm

gassmoeller:
I am pretty sure a spherical shell with a free surface still has a rotational nullspace because material can move freely through the surface. However our stabilization term might save us, because it is applied normal to the surface. Yes, lets go with option 2.

gassmoeller:
I cannot spot a problem in the code, but do you know why the convergence behavior and the remaining rotation is worse in some of the tests compared to the earlier results? That does not look right to me.

tjhei:
> I cannot spot a problem in the code, but do you know why the convergence behavior and the remaining rotation is worse in some of the tests compared to the earlier results? That does not look right to me.

I am not sure, honestly. Fixing a DoF might cause a slight change in condition number. I am not confident that fixing the nullspace is always the best choice for matrix-based. I now for sure that it is the right choice for GMG as it doesn't converge in many situations without this patch.

tjhei:
This is now updated. I fixed various smaller bugs and cleaned up the implementation.

There is now a new parameter that, by default, causes constraints only be added when using GMG as AMG does not always profits from the constraints.

But interestingly, I have a new example in tests/ that only converges with constraints with AMG/GMG (free surface + free slip). This example is comes from @elodie-kendall .

tjhei:
Well, I celebrated too early. If I run the new example for longer with more refinements, GMG will stop converging eventually. Sigh.

bangerth:
The last comment makes it sound like this isn't ready. @tjhei -- what's the state, do you want another look at the patch, or are you still searching for issues?

tjhei:
Thanks for the ping. We could merge the patch "as is" because it is a strict improvement over the current situation. 

Sadly, I haven't figured out what is happening in the situations where this patch doesn't help.

source/simulator/parameters.cc
```diff
@@ -767,6 +767,17 @@ namespace aspect
                          "\n\n"
                          "Note that while more than one operation can be selected it only makes sense to "
                          "pick one rotational and one translational operation.");
+
+      prm.declare_entry("Remove rotational nullspace in linear solver", "false",
```
gassmoeller, line 771: Shouldnt the default be `auto` to avoid users having to set this parameter for all gmg models?

```suggestion
      prm.declare_entry("Remove rotational nullspace in linear solver", "auto",
```

gassmoeller (commented):
I am happy to merge this in the current state to get the improvements for gmg in, but I have two small comments left.

tjhei:
Switching to "auto" increases the number of iterations in a couple of tests slightly, see the last commit.

gassmoeller:
I wouldnt worry about that, still better than everyone running GMG having to manually include this parameter if their models crash. But it seems there are two tests for which the auto detection does not give the correct result, they still crash. Can you take a look?

tjhei:
need a rebase after #4610

source/simulator/nullspace.cc
```diff
@@ -103,11 +238,124 @@ namespace aspect
...
+                                                             level,
+                                                             constraints);
+        bool global_success = (1==Utilities::MPI::max(success?1:0, mpi_communicator));
+        AssertThrow(level != -1 || global_success,
+                    ExcInternalError("ERROR: constrain_point_on_boundary() could not find the specified support point."));
+
+        // point 2 is in positive x direction, constrain y
+        Point<dim> location2;
+        location2[0] = location1[dim-1];
+        mask.set(0, false);
```
bangerth, line 335: Just create a different mask variable that says `y_velocity`.

source/simulator/nullspace.cc
```diff
@@ -103,11 +238,124 @@ namespace aspect
...
+        success = internal::constrain_point_on_boundary(dof_handler,
+                                                        mapping,
+                                                        location2,
+                                                        mask,
+                                                        level,
+                                                        constraints);
+        global_success = (1==Utilities::MPI::max(success?1:0, mpi_communicator));
+        AssertThrow(level != -1 || global_success,
+                    ExcInternalError("ERROR: constrain_point_on_boundary() could not find the specified support point."));
```
bangerth, line 344: If you're already using different mask variables, just put this whole block into `{...}` and then you can make the two success variables `const`.

source/simulator/nullspace.cc
```diff
@@ -103,11 +238,124 @@ namespace aspect
...
+      }
+
+
+  }
 
   template <int dim>
   void Simulator<dim>::setup_nullspace_constraints(AffineConstraints<double> &constraints)
   {
+
+    if (parameters.nullspace_removal & Parameters<dim>::NullspaceRemoval::any_rotation && parameters.constrain_rotational_nullspace)
```
bangerth, line 354:
```suggestion
    if ((parameters.nullspace_removal & Parameters<dim>::NullspaceRemoval::any_rotation)
        && 
        parameters.constrain_rotational_nullspace)
```

source/simulator/nullspace.cc
```diff
@@ -103,11 +238,124 @@ namespace aspect
...
+      {
+        setup_rotational_constraints(*geometry_model, mesh_deformation.get(), dof_handler, *mapping, constraints, /*level = */ -1, mpi_communicator);
```
bangerth, line 356: Oh, so you do actually use these constraints for the solution as well, not just the solver?

source/simulator/nullspace.cc
```diff
@@ -206,6 +454,22 @@ namespace aspect
   }
 
 
+  template <int dim>
+  void Simulator<dim>::setup_nullspace_constraints(AffineConstraints<double> &constraints,
+                                                   const DoFHandler<dim> &dof_handler,
+                                                   const int level)
```
bangerth, line 460: same thing here with the level variable, assuming you are again treating `-1` as special.

source/simulator/nullspace.cc
```diff
@@ -206,6 +454,22 @@ namespace aspect
...
+  {
+    if (parameters.nullspace_removal & Parameters<dim>::NullspaceRemoval::any_rotation && parameters.constrain_rotational_nullspace)
```
bangerth, line 462:
```suggestion
    if ((parameters.nullspace_removal & Parameters<dim>::NullspaceRemoval::any_rotation)
       && 
       parameters.constrain_rotational_nullspace)
```

source/simulator/stokes_matrix_free.cc
```diff
@@ -2963,10 +2966,13 @@ namespace aspect
...
             const Quadrature<dim> &quadrature_formula = sim.introspection.quadratures.velocities;
-            FEValues<dim> fe_values (*(sim.mapping), fe_v, quadrature_formula,
+            const Mapping<dim> &mapping =
+              (sim.mesh_deformation) ? sim.mesh_deformation->get_level_mapping(level) : *sim.mapping;
+
+            FEValues<dim> fe_values (mapping, fe_v, quadrature_formula,
                                      update_values   | update_gradients |
                                      update_quadrature_points | update_JxW_values);
-            FEValues<dim> fe_values_projection (*(sim.mapping),
+            FEValues<dim> fe_values_projection (mapping,
```
bangerth, line 2975: Is this unrelated? If so, happy to take that as a separate patch :-)

bangerth (changes requested):
Apologies for all of the comments...

source/simulator/nullspace.cc
```diff
@@ -19,22 +19,157 @@
...
+                                                 fe,
+                                                 quadrature,
+                                                 update_quadrature_points);
+
+      for (const auto &cell : dof_handler.active_cell_iterators())
+        if (cell->at_boundary()
+            &&
+            (cell->is_locally_owned() || cell->is_ghost()))
+          for (unsigned int face_no = 0;
+               face_no < GeometryInfo<dim>::faces_per_cell; ++face_no)
```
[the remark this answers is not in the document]
tjhei: good catch, thank you!

source/simulator/nullspace.cc
```diff
@@ -103,11 +238,124 @@ namespace aspect
...
+  void setup_rotational_constraints(const GeometryModel::Interface<dim> &geometry_model,
+                                    const MeshDeformation::MeshDeformationHandler<dim> *mesh_deformation,
+                                    const DoFHandler<dim> &dof_handler,
+                                    const Mapping<dim> &mapping,
+                                    AffineConstraints<double> &constraints,
+                                    int level,
+                                    const MPI_Comm &mpi_communicator)
+  {
+    // A spherical object has 1 rotational mode in 2d and 3 in
+    // 3d. We will fix an appropriate number of velocity
```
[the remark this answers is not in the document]
tjhei: Yes, rotations around x, y, and z (I added a comment). Or am I wrong?

source/simulator/nullspace.cc
```diff
@@ -103,11 +238,124 @@ namespace aspect
...
+    bool mesh_deformation_bottom = false;
+    if (mesh_deformation_enabled)
+      {
+        const std::set<types::boundary_id> ids = mesh_deformation->get_active_mesh_deformation_boundary_indicators();
+
+        if (ids.find(geometry_model.get_symbolic_boundary_names_map()["top"]) != ids.end())
+          mesh_deformation_top = true;
+        if (is_spherical_shell && ids.find(geometry_model.get_symbolic_boundary_names_map()["bottom"]) != ids.end())
+          mesh_deformation_bottom = true;
+      }
```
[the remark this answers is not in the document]
tjhei: Yes, renamed the variables and added a comment.

include/aspect/parameters.h
```diff
@@ -625,6 +625,13 @@ namespace aspect
      * field.
      */
     typename NullspaceRemoval::Kind nullspace_removal;
+
+    /**
+     * If true, we add constraints for the linear solver for rotational
+     * modes.
+     */
+    bool constrain_rotational_nullspace;
```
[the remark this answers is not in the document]
tjhei: Comment updated: Yes, correct. It does modify the solution, but the removal happening afterwards will hopefully undo this.

source/simulator/nullspace.cc
```diff
@@ -19,22 +19,157 @@
...
+
+      FEFaceValues<dim, spacedim> fe_face_values(mapping,
+                                                 fe,
+                                                 quadrature,
+                                                 update_quadrature_points);
+
+      for (const auto &cell : dof_handler.active_cell_iterators())
+        if (cell->at_boundary()
+            &&
+            (cell->is_locally_owned() || cell->is_ghost()))
```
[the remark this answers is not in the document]
tjhei: Yes, I am pretty sure I need it for locally_relevant.

source/simulator/nullspace.cc
bangerth: Forget about this. I now realize that this happens in a separate step (setting up the linear system that needs to be solved) but that this does not affect the later step where we're actually removing the net rotation.

zjiaqi2018:
Still don't understand why `add_user_constraints()` causes the divergence in solving mesh displacement. Without it, it converges in 200 steps.

After removing `add_user_constraints()`, I got the following results, is the difference acceptable?
left: amg

image

tjhei:
I will need to do some more research here, thank you @zjiaqi2018

source/simulator/nullspace.cc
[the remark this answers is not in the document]
bangerth: You're right of course.

include/aspect/simulator.h
```diff
@@ -1428,21 +1428,29 @@ namespace aspect
...
+       * respectively.
        */
       void setup_nullspace_constraints(AffineConstraints<double> &constraints);
 
+      /**
+       * Like above, but on a given multigrid level @p level.
+       */
+      void setup_nullspace_constraints(AffineConstraints<double> &constraints,
+                                       const DoFHandler<dim> &dof_handler,
+                                       const int level);
```
bangerth, line 1453: Out of curiosity, why do you need to pass a `DoFHandler` argument to the second function but not the first?

source/simulator/nullspace.cc
```diff
@@ -19,22 +19,164 @@
...
+    constrain_point_on_boundary_to_zero(const DoFHandler<dim,spacedim> &dof_handler,
+                                        const Mapping<dim> &mapping,
+                                        const Point<dim> &location,
+                                        const ComponentMask &mask,
+                                        const int level,
+                                        AffineConstraints<double> &constraints)
+    {
+      // Call the active one instead if level==-1. This indirection simplifies the code using this function.
+      if (level==-1)
+        return constrain_point_on_boundary_to_zero(dof_handler, mapping, location, mask, constraints);
```
bangerth, line 123: Well, I'm still not fond of the -1 thing :-( Do you insist on it?

source/simulator/nullspace.cc
```diff
@@ -19,22 +19,164 @@
...
+          cell->is_locally_owned_on_level() ||  cell->is_ghost_on_level()
+#else
+          cell->level_subdomain_id() != numbers::artificial_subdomain_id
+#endif
+        )
+          for (unsigned int face_no = 0;
+               face_no < GeometryInfo<dim>::faces_per_cell; ++face_no)
+            if (cell->at_boundary(face_no))
+              {
+                typename DoFHandler<dim, spacedim>::level_face_iterator face = cell->face(face_no);
```
bangerth, line 150:
```suggestion
                const typename DoFHandler<dim, spacedim>::level_face_iterator face = cell->face(face_no);
```

source/simulator/nullspace.cc
```diff
@@ -19,22 +19,164 @@
...
+
+      for (const auto &cell : dof_handler.active_cell_iterators())
+        if (cell->at_boundary()
+            &&
+            (cell->is_locally_owned() || cell->is_ghost()))
+          for (unsigned int face_no = 0;
+               face_no < GeometryInfo<dim>::faces_per_cell; ++face_no)
+            if (cell->at_boundary(face_no))
+              {
+                typename DoFHandler<dim, spacedim>::face_iterator face = cell->face(face_no);
```
bangerth, line 80:
```suggestion
                const typename DoFHandler<dim, spacedim>::face_iterator face = cell->face(face_no);
```

source/simulator/nullspace.cc
```diff
@@ -103,11 +238,124 @@ namespace aspect
...
   }
 
+  template <int dim>
+  void setup_rotational_constraints(const GeometryModel::Interface<dim> &geometry_model,
+                                    const MeshDeformation::MeshDeformationHandler<dim> *mesh_deformation,
+                                    const DoFHandler<dim> &dof_handler,
+                                    const Mapping<dim> &mapping,
+                                    AffineConstraints<double> &constraints,
+                                    int level,
+                                    const MPI_Comm &mpi_communicator)
```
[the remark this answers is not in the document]
bangerth: Still open.

source/simulator/nullspace.cc
```diff
@@ -103,11 +238,124 @@ namespace aspect
...
+        return;
+      }
 
+    // Use the bottom surface unless we have mesh deformation at the bottom:
+    const double point_depth = ((is_sphere || (is_spherical_shell && mesh_deformation_bottom)) ? 0.0 : geometry_model.maximal_depth());
+
+    if (dim==2)
+      {
+        // pick point (in positive y direction) and fix x component of velocity:
+        const Point<dim> location = geometry_model.representative_point(point_depth);
```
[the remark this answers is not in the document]
bangerth: also still open

source/simulator/nullspace.cc
```diff
@@ -103,11 +245,126 @@ namespace aspect
...
+        const bool success = internal::constrain_point_on_boundary_to_zero(dof_handler,
+                                                                           mapping,
+                                                                           location,
+                                                                           x_velocity_mask,
+                                                                           level,
+                                                                           constraints);
+
+        const bool global_success = (Utilities::MPI::max(success?1:0, mpi_communicator) == 1);
+        AssertThrow(level != -1 || global_success,
+                    ExcInternalError("ERROR: constrain_point_on_boundary() could not find the specified support point."));
```
bangerth, line 321: No need to say "Error:" The whole exception says so already.

source/simulator/nullspace.cc
```diff
@@ -103,11 +245,126 @@ namespace aspect
...
+        y_velocity_mask.set(1, true);
+        success = internal::constrain_point_on_boundary_to_zero(dof_handler,
+                                                                mapping,
+                                                                location2,
+                                                                y_velocity_mask,
+                                                                level,
+                                                                constraints);
+        global_success = (Utilities::MPI::max(success?1:0, mpi_communicator)==1);
+        AssertThrow(level != -1 || global_success,
+                    ExcInternalError("ERROR: constrain_point_on_boundary() could not find the specified support point."));
```
bangerth, line 353:
```suggestion
                    ExcInternalError("constrain_point_on_boundary() could not find the specified support point."));
```

bangerth: On second thought, the function name here is also no longer correct. Just drop it, the error message can be followed to the place where the call happens.

bangerth (commented):
There are still a number of unaddressed comments.

source/simulator/nullspace.cc
tjhei: No, still working on it...

tjhei:
I am still working on addressing some of the issues and let you know once I think I am done.

> I would much rather there be a flag bool global_instead_of_level_cells and if true, then level==invalid_unsigned_int.

I rewrote the somewhat awkward indirection by introduction another function ``constrain_point_on_boundary_to_zero()`` that only does the active or not decision. This should make things cleaner.
I also implemented your suggestion using an additional bool. I am not sure this is superior to using ``level == invalid_unsigned_int`` with a default value. I let you decide (I prefer the one without a bool, because bool arguments always need to be awkwardly annotated at the call site).

bangerth:
I share your observation about `bool` variables, but passing `invalid_unsigned_int` or `-1` is not much more descriptive than `false` :-) The "right" solution would be an enum, but that's overkill for a single use.

If you want to be extra explicit, write it as
```
  /* global_instead_of_level_constraint = */ false,
```
or some such where you call the function. Use your discretion.

tjhei:
> but passing `invalid_unsigned_int` or `-1` is not much more descriptive than `false`

Yes, but now you have two arguments that have to be consistent.

> If you want to be extra explicit, write it as

That is what I did (and what I meant by "need to be awkwardly annotated at the call site")

bangerth:
On 5/23/22 08:24, Timo Heister wrote:
> That is what I did (and what I meant by "need to be awkwardly annotated at
> the call site")

Would be nice if we could have python style
   foo(bar=13)
syntax :-)

anne-glerum:
@tjhei From the description of the new parameters, I gather this PR can also be useful for setups that use AMG but fail in the linear solve for a strict tolerance with null space removal, right?

tjhei:
Yes, while not useful in most situations, I have at least one example where it helps:
https://github.com/geodynamics/aspect/blob/d4f2d64b806809085cb1ae602f9b153a6ec1202f/tests/nullspace-freesurface-amg.prm#L1-L6

anne-glerum:
Ah great, that might help @elodie-kendall's models. A recent setup runs without nullspace removal for ~30 My, but fails with nullspace removal around 25 My. (As it's a quarter shell with free slip top and bottom and periodic side boundaries, the net rotation would have to be removed.) 

Would the PR be ready for another review by @bangerth?

bangerth:
What's the status here?

gassmoeller:
@tjhei: ping? Happy to help if it is useful.

PR #4414, opened 2021-11-21
Reduce memory allocations

gassmoeller:
After optimizing the particle update routines in previous PRs the memory allocation now shows up as a major cost when profiling the particles (15% of the total update cost in 2D spherical models, likely more in 2d box models, less in 3d spherical). This PR reorganizes the SolutionEvaluators class to work with preallocated vectors. It also avoids the calls to `std::distance` which with the new particle organization inside deal.II are more expensive then simply calling `n_particles_in_cell`. No change in functionality, just refactoring.


tjhei (approved):
Looks good.

PR #4415, opened 2021-11-22
Use precomputed mapping information during particle updates

gassmoeller:
This PR makes use of the improvements in dealii/dealii#12983 and should only be merged after that PR is merged. It reduces the duplicated computation of the mapping data during particle updates.


tjhei (approved):

gassmoeller:
Now that dealii/dealii#12983 is merged I rebased this PR and if the tests pass this should be ready to merge. We may need to update this call if the deal.II interface is modified in dealii/dealii#13245

anne-glerum:
Thanks Rene!

PR #4416, opened 2021-11-22
Optimize elasticity evaluation in material model

gassmoeller:
This is another seemingly small change that drastically improves the speed of computing the reaction terms for elasticity, in particular for particles. By using the deal.II class `FEPointEvaluation` instead of `FEValues` we no longer need to create a new `FEValues` object for every evaluation (which in case of particles was for every particles). The new implementation is fully equivalent and gives the following speedups for the bending beam benchmark with particles. I benchmarked this in reduced resolution on a 25x10 grid and with 1e3 particles for 2 timesteps, I used aspect from PR #4415 and dealii from PR #12983, all in optimized mode:

- Particle Update -> from 0.131s (19%) to 0.00818s (1.7%) runtime, 16X speedup
- Stokes assembly -> from 0.038s (5.5%) to 0.0225s (4.7%), 1.7X speedup
- Temperature assembly ->  from 0.0542s (7.9%) to 0.0285s (6%), 1.9X speedup

@naliboff and @anne-glerum FYI.

### DIFF
--- a/include/aspect/material_model/rheology/elasticity.h
+++ b/include/aspect/material_model/rheology/elasticity.h
@@ -192,7 +192,7 @@ namespace aspect
            * By caching the evaluator, we can avoid recreating it every time we
            * need it.
            */
-          std::unique_ptr<FEPointEvaluation<dim, dim>> evaluator;
+          mutable std::unique_ptr<FEPointEvaluation<dim, dim>> evaluator;
       };
     }
   }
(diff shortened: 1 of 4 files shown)


tjhei (approved):
Very nice. I am not surprised that creating a new FEValues every time is much slower.

anne-glerum:
Wonderful, thanks Rene!

gassmoeller:
I had to change the initialization of FEPointEvaluation, because in models with mesh deformation the mapping is changed after the material model is initialized. But now everything works, ready from my side.

naliboff:
Indeed, thanks Rene!

PR #4424, opened 2021-12-07
Remove uses of GeometryInfo.

bangerth:
GeometryInfo is quad/hex-specific, and so will have to be removed from deal.II at some point
in the near future. This patch removes a whole bunch of uses of that class and replaces
them by querying this information from the cell we are working on.

/rebuild

### DIFF
--- a/source/termination_criteria/steady_heat_flux.cc
+++ b/source/termination_criteria/steady_heat_flux.cc
@@ -60,7 +60,7 @@ namespace aspect
 
       for (const auto &cell : this->get_dof_handler().active_cell_iterators())
         if (cell->is_locally_owned())
-          for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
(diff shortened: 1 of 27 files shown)


source/termination_criteria/steady_heat_flux.cc
```diff
@@ -60,7 +60,7 @@ namespace aspect
 
       for (const auto &cell : this->get_dof_handler().active_cell_iterators())
         if (cell->is_locally_owned())
-          for (unsigned int f=0; f<GeometryInfo<dim>::faces_per_cell; ++f)
+          for (const unsigned int f : cell->face_indices())
```
tjhei, line 63: I really like that these can be const now. :+1:

tjhei (approved):

source/termination_criteria/steady_heat_flux.cc
bangerth: Yes :-)

PR #4425, opened 2021-12-08
Remove a few more uses of GeometryInfo.

bangerth:
Follow-up to #4424 .

/rebuild

### DIFF
--- a/source/mesh_refinement/maximum_refinement_function.cc
+++ b/source/mesh_refinement/maximum_refinement_function.cc
@@ -57,7 +57,7 @@ namespace aspect
               bool coarsen = false;
               bool clear_refine = false;
 
-              for ( unsigned int v = 0; v < GeometryInfo<dim>::vertices_per_cell;  ++v)
+              for (const unsigned int v : cell->vertex_indices())
                 {
                   const Point<dim> vertex = cell->vertex(v);
                   Utilities::NaturalCoordinate<dim> point =
--- a/source/mesh_refinement/minimum_refinement_function.cc
+++ b/source/mesh_refinement/minimum_refinement_function.cc
@@ -57,7 +57,7 @@ namespace aspect
               bool refine = false;
               bool clear_coarsen = false;
 
-              for ( unsigned int v = 0; v < GeometryInfo<dim>::vertices_per_cell;  ++v)
+              for (const unsigned int v : cell->vertex_indices())
                 {
                   const Point<dim> vertex = cell->vertex(v);
                   Utilities::NaturalCoordinate<dim> point =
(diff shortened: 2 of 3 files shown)


source/geometry_model/chunk.cc
```diff
@@ -266,21 +266,22 @@ namespace aspect
         // center is less than 1.e-5 of the distance between vertices (as
         // measured by the minimum distance from any of the other vertices
         // to the first vertex), then we call this a horizontal face.
-        constexpr unsigned int n_vertices = GeometryInfo<dim>::vertices_per_face;
-        std::array<double, n_vertices>     distances_to_center;
-        std::array<double, n_vertices - 1> distances_to_first_vertex;
+        constexpr unsigned int max_n_vertices_per_face = (1 << (dim-1));
```
peterrum, line 269: This does not look optimal. We should make this part of `ReferenceCell`!

bangerth: Didn't expect to get a review from you here :-)

To the point: True. It's the max over the number of vertices per cell for the `dim-1` dimensional reference cells. But I'm not sure where such a function would go. It isn't information that is part of any specific reference cell. I'm also not sure whether there ought to be a member function that gives us this information.

I did think of writing it as
```
  std::max(ReferenceCells::Hexahedron.face(0).vertices_per_cell, 
                ReferenceCells::Tetrahedron.face(0).vertices_per_cell);
```
But this only works for 3d and would need more work to also make work for 2d, and lastly it is not `constexpr`. So I decided that this must be good enough :-)

peterrum: > To the point: True. It's the max over the number of vertices per cell for the dim-1 dimensional reference cells. But I'm not sure where such a function would go. It isn't information that is part of any specific reference cell. I'm also not sure whether there ought to be a member function that gives us this information.

We will need this info in `RefernceCell` else we won't be able to remove `GeometryInfo`: the `GeometryInfo` functions are used to initialize `std::arrays` and similar data structures at compile time.  Maybe we can think about `static constexpr` functions. But we don't need to spam aspect, we can continue the discussion at the deal.II GitHub page.

gassmoeller: I agree that this is kind of good enough in terms of functionality, but I am pretty sure 95% of the readers of this code will not understand what is happening in this line. Wouldnt `(dim == 3) ? 4 : 2` be good enough? Or is that not possible for `constexpr`? We only support 2D and 3D cases in ASPECT anyway.

gassmoeller (commented):
Thanks for the improvements. I have one question (see above) but otherwise I think this is good to go. But I will also wait to give @peterrum time to reply.

gassmoeller:
It seems the testers think there is a functional difference in the chunk geometry model between the old version and the new one.

source/geometry_model/chunk.cc
bangerth: That would work just fine :-)

bangerth:
I don't understand the differences in output. Let me remove the geometry part of the patch for the moment and see that the rest of the patch tests correctly. This should make getting it merged easier.

gassmoeller (approved):

PR #4426, opened 2021-12-08
Provide a 'const' version of NaturalCoordinate::get_coordinates().

bangerth:
Previously, there was only a non-const version that returned a non-const reference to
a member variable. This isn't usable when the object it is called on is 'const'. So
provide a 'const' version that returns a 'const' reference.

/rebuild


gassmoeller (approved):
Yes this is very reasonable.

tjhei (approved):

PR #4427, opened 2021-12-08
Replace FEValues by FEPointEvaluation where useful

gassmoeller:
This is essentially the same PR as #4416 but for a few additional uses of FEValues.

@naliboff: Can you check if this accelerates the Kaus2010 benchmark case?

### DIFF
--- a/include/aspect/material_model/grain_size.h
+++ b/include/aspect/material_model/grain_size.h
@@ -24,6 +24,9 @@
 
 #include <aspect/material_model/interface.h>
 #include <aspect/simulator_access.h>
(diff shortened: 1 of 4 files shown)


anne-glerum (approved):

anne-glerum:
Thanks for these updates @gassmoeller !
The Kaus extension benchmark currently in the repo does not include strain weakening, so I don't think it would speed up from this PR. Or did you try with strain weakening @naliboff ?

anne-glerum:
Oeh that's wonderful!

PR #4428, opened 2021-12-09
More GeometryInfo stuff.

bangerth:
I broke these out of #4425. These should be uncontroversial, but let's see what we find.

Follow-up to #4424.

/rebuild

### DIFF
--- a/source/geometry_model/spherical_shell.cc
+++ b/source/geometry_model/spherical_shell.cc
@@ -153,13 +153,11 @@ namespace aspect
                   for (unsigned int cell_layer = 0; cell_layer < R_values.size() - 1; ++cell_layer)
                     {
                       CellData<dim> this_cell;
-                      for (unsigned int vertex_n = 0;
-                           vertex_n < GeometryInfo<dim-1>::vertices_per_cell;
-                           ++vertex_n)
+                      for (const unsigned int vertex_n : cell->vertex_indices())
                         {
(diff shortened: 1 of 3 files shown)


source/geometry_model/spherical_shell.cc
```diff
@@ -153,13 +153,11 @@ namespace aspect
...
-                      for (unsigned int vertex_n = 0;
-                           vertex_n < GeometryInfo<dim-1>::vertices_per_cell;
-                           ++vertex_n)
+                      for (const unsigned int vertex_n : cell->vertex_indices())
                         {
                           this_cell.vertices[vertex_n] =
                             cell->vertex_index(vertex_n) + cell_layer * sphere_mesh.n_vertices();
-                          this_cell.vertices[vertex_n + GeometryInfo<dim-1>::vertices_per_cell] =
+                          this_cell.vertices[vertex_n + cell->n_vertices()] =
                             cell->vertex_index(vertex_n) +
```
bangerth, line 161: This change looks funny (the first line has `dim-1`, but the second one queries `cell->n_vertices()`, but that's because if you look higher up, `cell` is a `dim-1` dimensional cell.

That the conversion is correct is also clear by looking at the last line: `cell->vertex_index(vertex_n)` clearly indicates that `vertex_n` needs to run over the vertex indices of *cell*.

gassmoeller (approved):

PR #4430, opened 2021-12-09
More GeometryInfo work, the questionable part.

bangerth:
This is the part of #4425 that didn't work yesterday. I checked that at least on my system, there is no difference between the before/after outputs of the `merged_chunks_2D` test that failed yesterday, though the output is (expectedly) slightly different than what's stored.

/rebuild

### DIFF
--- a/source/geometry_model/chunk.cc
+++ b/source/geometry_model/chunk.cc
@@ -266,21 +266,25 @@ namespace aspect
         // center is less than 1.e-5 of the distance between vertices (as
         // measured by the minimum distance from any of the other vertices
         // to the first vertex), then we call this a horizontal face.
-        constexpr unsigned int n_vertices = GeometryInfo<dim>::vertices_per_face;
-        std::array<double, n_vertices>     distances_to_center;
-        std::array<double, n_vertices - 1> distances_to_first_vertex;
+        constexpr unsigned int max_n_vertices_per_face = (dim==2 ? 2 : 4);
+        std::array<double, max_n_vertices_per_face>     distances_to_center;
+        std::array<double, max_n_vertices_per_face - 1> distances_to_first_vertex;
         distances_to_center[0] = face->vertex(0).norm_square();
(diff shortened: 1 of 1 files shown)


gassmoeller (commented):
Maybe the tester was a bit overzealous when comparing the numbers in the other PR. I remember that the difference was something like 1e-8 and I thought our numdiff tolerance was coarser anyway. But could you replace the `1 << x` line with what I suggested earlier: `(dim == 3) ? 4 : 2;`? I think it is just easier to understand.

bangerth:
Bizarre that it works now. I've made the change you requested (just the other way around :-) )

source/geometry_model/chunk.cc
```diff
@@ -266,21 +266,25 @@ namespace aspect
...
               (face->vertex(i) - face->vertex(0)).norm_square();
           }
         const auto minmax_distance =
-          std::minmax_element(distances_to_center.begin(), distances_to_center.end());
+          std::minmax_element(distances_to_center.begin(),
+                              distances_to_center.begin()+face->n_vertices());
         const auto min_distance_to_first_vertex =
           std::min_element(distances_to_first_vertex.begin(),
-                           distances_to_first_vertex.end());
+                           distances_to_first_vertex.begin()+face->n_vertices());
```
gassmoeller, line 287: Oh, I think I see what is happening, another random test failure due to access to non-initialized memory. Does this need to be the following? (see the size of the array above)
```suggestion
                           distances_to_first_vertex.begin()+face->n_vertices()-1);
```

bangerth: Oh, you're totally right! How insidious! How did this work before?

bangerth (commented):
OK, let's try this again!

gassmoeller (approved):
Looks good, feel free to merge if the tests pass.

Before the call was to `distances_to_first_vertex.end()` which worked correctly.

PR #4431, opened 2021-12-09
More GeometryInfo work.

bangerth:
This seems uncontroversial.

/rebuild

### DIFF
--- a/source/geometry_model/two_merged_boxes.cc
+++ b/source/geometry_model/two_merged_boxes.cc
@@ -41,32 +41,32 @@ namespace aspect
       // iterate over all active cells and (re)set the boundary indicators
       for (const auto &cell : triangulation.active_cell_iterators())
         {
-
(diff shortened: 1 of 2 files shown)


bangerth:
Ah, uncontroversial doesn't mean that it works apparently. But I can't make sense of the CI error message. Did something just happen to die independent of what my patch contains?

gassmoeller:
We have one test that occasionally provides different results (simple_shear_output_the_mobility) at the moment. I think #4433 should fix the issue, and the test is completely unrelated to the change in this PR (it uses the box geometry). Could you take a look at #4433 and rebase this PR after the other one has been merged? I am pretty sure that should fix the issue.

bangerth:
Rebased now on top of #4433.

gassmoeller (approved):
Ready to merge if the tests pass.

PR #4432, opened 2021-12-10
Modernize code a bit in places.

bangerth:
Specifically, use a better way to deal with unspecified array sizes than
  sizeof(X)/sizeof(X[0])

/rebuild


gassmoeller (approved):
Looks good. :+1:

PR #4433, opened 2021-12-10
Initialize variables in mobility statistics postprocessor

gassmoeller:
Avoid uninitialized variables in the mobility_statistics postprocessor. This would work most of the times, but occasionally lead to test failures like in #4431 (at least I think this is the reason). Also improve the code style a bit and avoid unnecessary computations. @elodie-kendall FYI, you need to initialize variables of type `double` otherwise their value is not determined and can be random (or 0 if you are lucky). I missed this during my review.

### DIFF
--- a/source/postprocess/mobility_statistics.cc
+++ b/source/postprocess/mobility_statistics.cc
@@ -66,8 +66,9 @@ namespace aspect
       std::vector<Tensor<1,dim>> surface_velocity_values (n_q_points_face);
 
       double local_velocity_square_integral = 0;
-      double local_velocity_square_integral_top_boundary;
-      double local_top_boundary_area;
+      double local_velocity_square_integral_top_boundary = 0;
+      double local_top_boundary_area = 0;
+
       const std::set<types::boundary_id>
       boundary_indicators
         = this->get_geometry_model().get_used_boundary_indicators ();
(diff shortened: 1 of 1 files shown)


elodie-kendall:
Great, thanks for the follow-up!

source/postprocess/mobility_statistics.cc
```diff
@@ -66,8 +66,9 @@ namespace aspect
       std::vector<Tensor<1,dim>> surface_velocity_values (n_q_points_face);
 
       double local_velocity_square_integral = 0;
-      double local_velocity_square_integral_top_boundary;
-      double local_top_boundary_area;
+      double local_velocity_square_integral_top_boundary = 0;
+      double local_top_boundary_area = 0;
```
bangerth, line 70: Ah, great catch. Well done.

Out of curiosity, how did you find this?

bangerth (approved):

bangerth:
You'll have to run the indenter, but other than that this is ok. Please merge yourself when ready, and then I will rebase #4431.

source/postprocess/mobility_statistics.cc
gassmoeller: I saw that it was the test `simple_shear_output_the_mobility` that would occasionally fail in PRs, but pass when rerun (or run locally). The test was added recently as part of #4389 so I knew to look into that postprocessor, and that it was something that is essentially non-deterministic. I am actually surprised we did not get a compiler warning about uninitialized use.

PR #4436, opened 2021-12-11
Remove a few more occurrences of GeometryInfo.

bangerth:
Follow-up to #4424.

/rebuild

### DIFF
--- a/source/simulator/assemblers/interface.cc
+++ b/source/simulator/assemblers/interface.cc
@@ -450,21 +450,21 @@ namespace aspect
                         finite_element.dofs_per_cell),
           local_matrices_int_ext ((field_is_discontinuous
                                    ?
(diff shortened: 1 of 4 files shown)


source/simulator/assemblers/interface.cc
```diff
@@ -450,21 +450,21 @@ namespace aspect
                         finite_element.dofs_per_cell),
           local_matrices_int_ext ((field_is_discontinuous
                                    ?
-                                   GeometryInfo<dim>::max_children_per_face * GeometryInfo<dim>::faces_per_cell
+                                   finite_element.reference_cell().n_faces() * GeometryInfo<dim>::max_children_per_face
```
bangerth, line 453: The remaining use of `max_children_per_face` is something I will have to fix in a follow-up. But the replacement I make is easy enough.

tjhei (approved):

PR #4441, opened 2021-12-18
add spherical nsinker benchmark

tjhei:


tjhei:
I am not sure if it makes more sense to combine the two benchmarks in a single directory as there is some code duplication this way.
Thoughts?

tjhei:
@gassmoeller What do you think, separate or integrate into existing benchmark?

gassmoeller:
I think the overlap is sufficiently small that having them in separate directories is fine. From a practical perspective I would suggest renaming the directory to nsinker_spherical_shell, then it shows up in the list of directories right next to the existing nsinker benchmark.
Do you want me to do a review at the current state?

tjhei:
> Do you want me to do a review at the current state?

Sure, I think this is good to go. Thank you.

tjhei:
Interestingly, jenkins and ``linux / tests`` disagree on the correct test output?!

tjhei:
> Interestingly, jenkins and `linux / tests` disagree on the correct test output?!

This was just VMPeak changing with each run. Fixed.

benchmarks/nsinker_spherical_shell/nsinker_gmg.prm
```diff
@@ -0,0 +1,95 @@
+# This is the "Spherical Shell NSinker" benchmark
```
jdannberg, line 1: Would it make sense to rename the file to `nsinker_spherical_shell.prm` (since there is no other version that does not use the GMG)?

tests/spherical_nsinker.prm
```diff
@@ -0,0 +1,15 @@
+set Dimension = 3
```
jdannberg, line 1: You don't need that line, right?

benchmarks/nsinker_spherical_shell/nsinker_gmg.prm
```diff
@@ -0,0 +1,95 @@
...
+set Nonlinear solver scheme                = no Advection, single Stokes
+set Use years in output instead of seconds = false
+
+subsection Solver parameters
+  subsection Matrix Free
+    set Output details = true
+  end
+  subsection Stokes solver parameters
+    set Stokes solver type = block GMG
+    set Number of cheap Stokes solver steps             = 500
```
jdannberg, line 17: This is different compared to the box case. Does it make sense to explain somewhere why for spherical models, it's useful to pick a larger value?

benchmarks/nsinker_spherical_shell/nsinker.cc
```diff
@@ -0,0 +1,338 @@
+#include <aspect/material_model/simple.h>
```
jdannberg, line 1: Do you need this?

benchmarks/nsinker_spherical_shell/nsinker.cc
```diff
@@ -0,0 +1,338 @@
...
+                                 "Viscosity in the sinkers.");
+              prm.declare_entry ("Sinker density", "10",
+                                 Patterns::Double (0),
+                                 "Density in the sinkers.");
+              prm.declare_entry ("Number of sinkers", "4",
+                                 Patterns::Integer (0,31),
+                                 "Number of sinking spheres.");
+              prm.declare_entry ("Transition zone viscosity jump", "1.0",
+                                 Patterns::Double (0),
+                                 "The factor by which the viscosity increases below the 660km transition zone.");
```
jdannberg, line 103:
```suggestion
                                 "The factor by which the viscosity increases below the base of the transition zone at a depth of 660 km.");
```

benchmarks/nsinker_spherical_shell/nsinker.cc
```diff
@@ -0,0 +1,338 @@
...
+      centers[72] = Point<3>(8.2344259034e-01, 5.9961585635e-01, 7.4369772512e-01);
+      centers[73] = Point<3>(3.2766604253e-01, 8.3176720460e-02, 9.5114077951e-01);
+      centers[74] = Point<3>(8.2308128282e-01, 5.2712029523e-01, 3.1080186614e-01);
+
+      // First rescale their location from [0,1] to [-1,1]:
+      for (auto &p: centers)
+        for (int d=0; d<3; ++d)
+          p[d] = p[d]*2.0-1.0;
+
+      // Now erase all points with center outside our shell:
```
jdannberg, line 283: How do you make sure you don't have any sinkers that are only halfway in the model domain? I assume the initial locations were picked that way? Or is this not important?

benchmarks/nsinker_spherical_shell/README.md
```diff
@@ -0,0 +1,21 @@
...
+Heterogeneous Viscosity
+Johann Rudi, Georg Stadler, Omar Ghattas
+https://doi.org/10.1137/16M108450X
+
+but adapted to a spherical shell geometry. See the "nsinker"
+benchmark in this repository for the version in a unit box.
+
+We consider a spherical shell with inner radius 0.54 and outer radius
+1.0 with N spherical, high-viscosity, high-density sinking spheres at
+random locations in the domain. Additionally, a transition zone with
```
jdannberg, line 17: This sounds like they would be different every time, or at least like ASPECT would create some pseudo-random numbers, but these are pre-generated locations from a paper. It would be good to make that clear.

jdannberg (commented):
This looks good to me, thank you for contributing this!

The only bigger comment I have is this: It would be great if you could provide some more information on what this benchmark is for and what it does, provided that this does not reproduce an existing paper, and that there's no run.sh script. Along the lines of: What parameter ranges have you tried/are reasonable? What output should I be looking at? What values of the output should I expect/what counts as good/bad? Is there any postprocessing that needs to be done?
Or is there a paper where you have used this that you could refer to?

tests/spherical_nsinker.prm
tjhei: Yes, it is necessary as the dimension can not be read from an include statement. I don't think we document this anywhere, but this is sadly the case.

benchmarks/nsinker_spherical_shell/nsinker_gmg.prm
tjhei: For nsinker we followed parameters from the literature, while I don't have to do that here. GMG typically does not need expensive iterations. (I have not tested this here, yet)

benchmarks/nsinker_spherical_shell/nsinker.cc
tjhei: This means we keep some sinkers that are at least halfway inside the domain, yes. I added a comment.

benchmarks/nsinker_spherical_shell/README.md
tjhei: fixed.

tjhei:
> Or is there a paper where you have used this that you could refer to?

I am planning to use this in a paper, but I don't have good answers to any of your questions at this point. We can wait with merging this PR if you prefer (but I don't think this is necessary).

benchmarks/nsinker_spherical_shell/README.md
```diff
@@ -9,13 +9,17 @@ Heterogeneous Viscosity
 Johann Rudi, Georg Stadler, Omar Ghattas
 https://doi.org/10.1137/16M108450X
 
-but adapted to a spherical shell geometry. See the "nsinker"
-benchmark in this repository for the version in a unit box.
+but adapted to a spherical shell geometry. See the "nsinker" benchmark
+in this repository for the version in a unit box. The motivation for
+creating this benchmark was de desire to test linear solver
```
jdannberg, line 14:
```suggestion
creating this benchmark was the desire to test linear solver
```

benchmarks/nsinker_spherical_shell/README.md
```diff
@@ -9,13 +9,17 @@ Heterogeneous Viscosity
...
+performance with geometry, which is often used in practice and that
```
jdannberg, line 15:
```suggestion
performance with spherical geometry, which is often used in practice and that
```
I assume this is what you meant?

benchmarks/nsinker_spherical_shell/amg.prm
```diff
@@ -0,0 +1,92 @@
+# This is the "Spherical Shell NSinker" benchmark
```
jdannberg, line 1: Oh, so there is an AMG case. It looks like this is the same as the GMG version, other than the solver section? In this case, I think it would make sense to include one file in the other, rather than copy the whole file. This is also what we do for other benchmarks.

tests/spherical_nsinker.prm
jdannberg: Oh wow, that's good to know!

benchmarks/nsinker_spherical_shell/nsinker_gmg.prm
jdannberg: We've had some problems with the expensive iterations too, for several projects. I think everyone in my group who's using the GMG uses something like a few thousand cheap solver steps to make sure we never use the expensive solver. It's of course a completely separate issue from this PR, but maybe we can document somewhere that it makes sense to use a lot of cheap steps with the GMG, or change the default.

benchmarks/nsinker_spherical_shell/nsinker.cc
```diff
@@ -100,7 +100,8 @@ namespace aspect
                                  "Number of sinking spheres.");
               prm.declare_entry ("Transition zone viscosity jump", "1.0",
                                  Patterns::Double (0),
-                                 "The factor by which the viscosity increases below the 660km transition zone.");
+                                 "The factor by which the viscosity increases below the base of the "
+                                 "transition zone at a dept of 660km.");
```
jdannberg, line 104:
```suggestion
                                 "transition zone at a depth of 660km.");
```

jdannberg (approved):
Okay, thank you! Looks good to me!
I have a few small comments (mostly typos) but feel free to merge yourself once you've addressed them.

tjhei:
Updated. And thank you for the thorough review!

jdannberg (approved):
Great, thank you!

And I am glad I finally have the time to review pull requests again now that I don't have to teach for this semester :-)

PR #4444, opened 2021-12-22
Don't name lambda functions 'lambda'.

bangerth:
It's not a descriptive variable name.

/rebuild

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -666,15 +666,13 @@ namespace aspect
         // there
         for (const auto p : boundary_temperature_manager.get_fixed_temperature_boundary_indicators())
           {
-            auto lambda = [&] (const dealii::Point<dim> &x) -> double
+            VectorFunctionFromScalarFunctionObject<dim> vector_function_object(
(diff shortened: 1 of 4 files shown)


source/simulator/core.cc
```diff
@@ -707,15 +705,13 @@ namespace aspect
...
               {
                 return boundary_composition_manager.boundary_composition(p, x, c);
-              };
-
-              VectorFunctionFromScalarFunctionObject<dim> vector_function_object(
-                lambda,
-                introspection.component_masks.compositional_fields[c].first_selected_component(),
-                introspection.n_components);
+              },
+              introspection.component_masks.compositional_fields[c].first_selected_component(),
```
tjhei, line 713: I don't mind passing the lambda directly to the constructor, but this confuses ``astyle``, sadly. Do you mind that the indentation is bad? (I think this is why we originally named the lambda)

bangerth: I can live with the indentation. How do others feel?

gassmoeller (approved):
I think its good enough.

PR #4449, opened 2022-01-07
Do not assign tangential boundaries as having outflow

jdannberg:
This is a bugfix. We have the feature that allows temperature and/or composition boundary conditions to be prescribed only at inflow parts of the boundary (which is useful at open boundaries). However, I noticed that especially in models with a spherical geometry, tangential boundaries are often in part assumed to have outflow, because the normal velocity at the boundary is not exactly zero. That means that at those boundaries, the boundary conditions is not set correctly. This pull request fixes this: Tangential boundary indicators are no longer considered when checking for which boundaries have outflow. 


tjhei:
Makes sense. Can you add a changelog entry please? And, is there a way to test this?

tjhei (approved):

jdannberg:
Yes, of course. I added a test. Below is an image that shows the old vs. the new behavior. 
The test is a copy of some spherical shell model where all boundaries are tangential, and temperature is supposed to be fixed at top and bottom. I changed "set Allow fixed temperature on outflow boundaries" to false. Because the normal velocity near the center of the top boundary is not exactly zero, but points slightly outwards, the temperature is not prescribed on that part of the boundary. 

old_new_BC

tests/shell_quarter_BC/log.txt
```diff
@@ -0,0 +1,165 @@
+-----------------------------------------------------------------------------
```
tjhei, line 1: you don't need this file (as screen-output is good enough)

tests/shell_quarter_BC/screen-output
```diff
@@ -0,0 +1,136 @@
+-----------------------------------------------------------------------------
```
tjhei, line 1: do you also want to add ``statistics`` file?

tjhei (approved):

tests/shell_quarter_BC/screen-output
jdannberg: Yes, I just gave the file the wrong name :-( Should be fixed now.

jdannberg:
Alright, I fixed the statistics file in the test and all tests are passing now.

tjhei (approved):

tjhei:
:+1:

PR #4450, opened 2022-01-07
Do not apply boundary conditions to projected density field

jdannberg:
In general, we apply boundary conditions to compositional fields that use the method "prescribed field". However, in the case of the density_field used in the projected density approximation, this is not correct: The boundary conditions come from the temperature, but the density itself should not have boundary conditions. However, if other compositional fields are used in the model, and they need boundary conditions, we can not just switch off boundary conditions for the projected density field in the input file. This pull request disables boundary conditions for the density_field used in the projected density approximation. 

This is something that may also be relevant for other fields with the method "prescribed field", so at some point we can discuss finding a more general solution. This PR just makes sure the densities are computed correctly in the projected density approximation. 

Below are images of the test case, with the current master (top) and the new version (bottom).

BC
 Old version: Boundary conditions are applied to the density field, causing it to be different from the actual density in the model. 

no_BC_new
New version: density and projected density are the same.


source/simulator/helper_functions.cc
```diff
@@ -1922,7 +1922,11 @@ namespace aspect
     // updating the ghost elements of the 'solution' vector.
     const unsigned int advection_block = adv_field.block_index(introspection);
     distributed_vector.block(advection_block).compress(VectorOperation::insert);
-    current_constraints.distribute (distributed_vector);
+
+    if (adv_field.is_temperature() ||
+        introspection.name_for_compositional_index(adv_field.compositional_variable) != "density_field")
```
tjhei, line 1927: same here, we should probably introduce a type for the density...

tjhei (approved):

gassmoeller (approved):
This can probably be updated in the follow-up to #4451 to use the density field type, but for now let's merge this in the current form.

PR #4451, opened 2022-01-07
Allow steinberger model to be used with projected density

jdannberg:
The Steinberger material model assumes that all compositional fields used in a simulation are chemical compositions and have a corresponding look-up table. This limits the use of compositional fields (to track the stress, the grain size, the projected density, etc.)

This pull request uses the new types of compositional fields to check if a field is really a composition, which allows it to use additional, other fields in a simulation. It also implements the additional outputs required to use the projected density approximation with the Steinberger material model. 

### For all pull requests:


source/material_model/steinberger.cc
```diff
@@ -270,6 +280,27 @@ namespace aspect
...
+    Steinberger<dim>::
+    fill_prescribed_outputs(const unsigned int q,
+                            const std::vector<double> &,
+                            const MaterialModel::MaterialModelInputs<dim> &,
+                            MaterialModel::MaterialModelOutputs<dim> &out) const
+    {
+      // set up variable to interpolate prescribed field outputs onto compositional field
+      PrescribedFieldOutputs<dim> *prescribed_field_out = out.template get_additional_output<PrescribedFieldOutputs<dim> >();
+
+      if (this->introspection().compositional_name_exists("density_field")
```
tjhei, line 294: Would it make sense to introduce ``CompositionalFieldDescription::density_field`` ? Identification by name is no longer necessary now that we have this system...

source/material_model/steinberger.cc
```diff
@@ -375,24 +406,39 @@ namespace aspect
...
+          // If the field type is unspecified (has not been set in the input file),
+          // we have to assume it also represents a chemical composition for reasons of
+          // backwards compatibility.
+          composition_mask = std::make_unique<ComponentMask> (this->n_compositional_fields(), false);
+          for (unsigned int c=0; c<this->n_compositional_fields(); ++c)
+            if (composition_descriptions[c].type == Parameters<dim>::CompositionalFieldDescription::chemical_composition
+                || composition_descriptions[c].type == Parameters<dim>::CompositionalFieldDescription::unspecified)
+              {
+                composition_mask->set(c, true);
+                n_chemical_fields++;
```
tjhei, line 423:
```suggestion
                ++n_chemical_fields;
```
but I think you can use ``n_chemical_fields = composition_mask->n_selected_components();`` at the end.

tests/steinberger_projected_density.prm
```diff
@@ -0,0 +1,132 @@
+# This is a copy of the steinberger compressible test.
+# It is intended to test the ability of the steinberger
+# material model to be used with the projected density
+# approximations, which requires using compositional
```
tjhei, line 4:
```suggestion
# approximation, which requires using compositional
```

tests/steinberger_projected_density.prm
```diff
@@ -0,0 +1,132 @@
...
+
+  set Strategy                           = velocity
+
+  set Time steps between mesh refinement = 0
+end
+
+
+# The parameters below this comment were created by the update script
+# as replacement for the old 'Model settings' subsection. They can be
+# safely merged with any existing subsections with the same name.
```
tjhei, line 92: remove

source/material_model/steinberger.cc
jdannberg: Thanks! I was wondering how to do that.

jdannberg:
I addressed your comments, and I also tried to introduce and use a `CompositionalFieldDescription::density`. 
But this is a bit awkward. The problem is that CompositionalFieldDescription is not a vector of field types (like the compositional field names), but it's a vector of structs, and the type of field is an enum in that struct. This means I can't use things like find/count etc. and always need several lines of code to check if a field exists/what index a field of a given type has etc. 

Any suggestions?

bobmyhill:
I think this is the kind of problem I wanted to avoid in the first commit of the FieldDescriptions PR (https://github.com/geodynamics/aspect/pull/4258). Would reinstating the transpose table I initially implemented work? I think @bangerth was open to the possibility if it was needed.

jdannberg:
Maybe we do not need to change the CompositionalFieldDescription. @gassmoeller made the point that it doesn't make the performance worse, it's just more lines of code; basically, we have to write our own find/count etc. functions. I only needed one of these, so I added that, and maybe it's fine as it is. 

@tjhei Can you have another look?

source/material_model/steinberger.cc
```diff
@@ -247,7 +247,16 @@ namespace aspect
             mass_fractions.push_back(1.0);
           else
             {
-              mass_fractions = MaterialUtilities::compute_composition_fractions(in.composition[i], *composition_mask);
+              // We only want to compute mass/volume fractions for fields that are chemical compositions.
+              std::vector<double> chemical_compositions;
+              std::vector<typename Parameters<dim>::CompositionalFieldDescription> composition_descriptions = this->introspection().get_composition_descriptions();
```
tjhei, line 252: can you make this ``const``?

source/material_model/steinberger.cc
```diff
@@ -375,24 +406,37 @@ namespace aspect
           equation_of_state.initialize_simulator (this->get_simulator());
           equation_of_state.parse_parameters(prm);
 
+          // Check if compositional fields represent a composition
+          std::vector<typename Parameters<dim>::CompositionalFieldDescription> composition_descriptions = this->introspection().get_composition_descriptions();
```
tjhei, line 410: same here

source/simulator/introspection.cc
```diff
@@ -312,6 +312,30 @@ namespace aspect
...
+
+
+  template <int dim>
+  unsigned int
+  Introspection<dim>::find_composition_type (const typename Parameters<dim>::CompositionalFieldDescription::Type &type) const
+  {
+    for (unsigned int c=0; c<composition_descriptions.size(); ++c)
+      if (composition_descriptions[c].type == type)
+        return c;
+    return composition_descriptions.size();
```
tjhei, line 334: do we want to assert here instead?

tjhei (approved):

tjhei:
This is good and adding the helper functions is a good idea.

I think we should adjust the "projected density" method to also use this field type instead of the name "density_field". I also looked at the source and could not find a check in that method that a field like this exists. We should add that. We also want a changelog for density_field. These can be a separate PRs, though.

source/simulator/introspection.cc
jdannberg: I implemented it like this because that's equivalent to the `std::find` function one would normally use for a vector, and it has the advantage that it doesn't crash if the field does not exist. 

But I am not feeling strongly about this; if you think it's better with an assert then I can change it.

jdannberg:
I have addressed your comments and also included the new field type in the changelog entry. 

I am happy to do a follow-up PR where I change this everywhere where we ask for the density field in the projected density approximation. My only question there would be: This is backwards incompatible in the sense that existing input files only have the correct name of the density field and not the type, and will then crash. I don’t know if anyone outside my group has been using the approximation so far, but there’s not really a good way to make this backwards compatible other than also to allow ‘unspecified’ as field type, which defies the whole purpose of the check.

tjhei (approved):

tjhei:
> I have addressed your comments and also included the new field type in the changelog entry.

thanks, looks good.

> I am happy to do a follow-up PR where I change this everywhere where we ask for the density field in the projected density approximation. My only question there would be: This is backwards incompatible in the sense that existing input files only have the correct name of the density field and not the type, and will then crash.

That is true. I agree that this is difficult to do. I don't mind making this change regardless, @gassmoeller ?

jdannberg:
I talked to @gassmoeller yesterday, and he suggested that for now, if a field is named "density_field" and its type is "unspecified", we just assign it the "density" type when reading it in. This means we can check for the type everywhere in the code, and if someone wants to call their field density_field but give it a different type, they can still do that. 

But this PR can be merged now, right?

gassmoeller:
I agree this looks good to go and a follow-up for the PDA using the field_type would be appreciated.

PR #4454, opened 2022-01-19
Matrix-free free surface stabilization

zjiaqi2018:
Matrix-free implementation of the free surface stabilization.
Currently not working for free surface boundaries + GMG + elasticity.
@tjhei  @naliboff


include/aspect/stokes_matrix_free.h
```diff
@@ -123,6 +128,17 @@ namespace aspect
       Table<2, SymmetricTensor<2, dim, VectorizedArray<number>>>
       newton_factor_wrt_strain_rate_table;
 
+      /**
+       * Table which stores the product of the pressure perturbation
+       * and the normalized gravity.
```
tjhei, line 133: This is a tensor on each boundary face, right (even the ones without free surface)? Mention this in the comment here.

tests/crameri_benchmark_1_gmg.prm
```diff
@@ -0,0 +1,59 @@
+# Can check this benchmark in gnuplot by opening it and running:
```
tjhei, line 1: For all prms added, change the description at the top to indicate that this is to test gmg+free surface.

doc/modules/changes/20220119_jiaqi
```diff
@@ -0,0 +1,5 @@
+New: The matrix-free GMG Stokes preconditioner is now 
+implemented for the free surface stabilization, but
+multilevel is not implemented in GMG yet.
```
tjhei, line 3:
```suggestion
New: The matrix-free GMG Stokes preconditioner is now 
implemented for the free surface stabilization. 
```

Can you explain the limitation about missing multilevel surface terms in the source code instead? (with a TODO)

tjhei (approved):
Looks good to me, thank you @zjiaqi2018 . I would like somebody else to review as well of course.

source/simulator/stokes_matrix_free.cc
```diff
@@ -2544,6 +2766,14 @@ namespace aspect
       additional_data.mapping_update_flags = (update_values | update_gradients |
                                               update_JxW_values | update_quadrature_points);
 
+      if (sim.mesh_deformation
+          &&!sim.mesh_deformation->get_free_surface_boundary_indicators().empty())
```
MFraters, line 2770: Space between && and !

MFraters (commented):
Thanks for adding this! I have a few small style comments. 

I do have a worry about the difference between the results of the tests at timestep 10. For example for `free_surface_blob` the topograhpy is `Topography min/max: -389.2 m, 740.4 m`, while for you it is `Topography min/max: -470.9 m, 868.8 m`, which seems a large difference to me. Is that something you would expect for this test because of the accuracy of the test?

tjhei:
> Is that something you would expect for this test because of the accuracy of the test?

I assume this is because the test is very coarse. Can you confirm @zjiaqi2018 that the solutions are similar on a finer mesh?

zjiaqi2018:
> > Is that something you would expect for this test because of the accuracy of the test?
> 
> I assume this is because the test is very coarse. Can you confirm @zjiaqi2018 that the solutions are similar on a finer mesh?

I think so. Number of degrees of freedom: 134,692 (82,818+10,465+41,409):
image

MFraters (approved):
Thanks for the changes and extra test which shows that both converge to the same solution.

naliboff (approved):
@zjiaqi2018 - This looks good to me as well, but perhaps good for someone else with a bit deeper knowledge of the solvers (in addition to @MFraters) to have a look as well. I will open a new PR with the elasticity tests.

tjhei:
This PR currently adds a couple of .prms in cookbooks/ and benchmarks/. I am not sure which ones are useful. Somebody want to voice an opinion?

MFraters:
> This PR currently adds a couple of .prms in cookbooks/ and benchmarks/. I am not sure which ones are useful. Somebody want to voice an opinion?

I do not have a strong opinion on this, and I am not sure if we have a policy on it. In general I had the impression that cookbooks should be accompanied with a section in the manual, but I am not sure if that is actually always the case. Benchmarks for aspect in my mind are models that are too big or complex to be run in the tester, but which a user can use to verify themselves. There is also a section for this in the manual, and some of them also have a section in the manual (but not all of them). Are they run before each release? I might be completely wrong about these, but this has mostly been my impression and made sense in my mind :) 

I think that is correct that the benchmark files are added as a gmg version within the Crameri benchmark folder. I am not sure how much the file in the cookbook folder adds without a section in the manual. I think that for the functionality which is added, just adding to the benchmark folder is makes the most sense to me, but if you think there is a good section to be written for the manual, that could justify a cookbook.

tjhei:
> Are they run before each release?

We run simplified versions (0 timesteps, sometimes coarser) of all benchmark and cookbooks prms before the release by using the check.mk files:
https://github.com/geodynamics/aspect/blob/main/benchmarks/check.mk

source/simulator/stokes_matrix_free.cc
```diff
@@ -1703,6 +1779,111 @@ namespace aspect
...
+                          = pressure_perturbation * g_hat[d];
+
+                    }
+
+                }
+            }
+
+
+
+        }
```
gassmoeller, line 1884: Remove the empty lines between closing braces here.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1703,6 +1779,111 @@ namespace aspect
...
+                  if (active_cell_data.free_surface_boundary_indicators.find(boundary_indicator)
+                      == active_cell_data.free_surface_boundary_indicators.end())
+                    continue;
+
+                  fe_face_values.reinit(simulator_cell, cell_face_pair.second);
+
+                  face_material_inputs.reinit(fe_face_values, simulator_cell, sim.introspection, sim.solution);
+
+                  sim.compute_material_model_input_values(sim.solution, fe_face_values,
+                                                          simulator_cell, false, face_material_inputs);
```
gassmoeller, line 1858: Just a style comment, but I would prefer you would break long lines like these into one parameter per line like you did above in line 1840.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1703,6 +1779,111 @@ namespace aspect
...
+
+          const QGauss<dim-1> face_quadrature_formula (sim.parameters.stokes_velocity_degree+1);
+
+          const unsigned int n_face_q_points = face_quadrature_formula.size();
+
+          FEFaceValues<dim> fe_face_values (*sim.mapping,
+                                            sim.finite_element,
+                                            face_quadrature_formula,
+                                            update_values   |
+                                            update_gradients |
```
gassmoeller, line 1813: I guess you need the gradients for the material model inputs? Because for everything else you don't seem to need them.

source/simulator/stokes_matrix_free.cc
```diff
@@ -1703,6 +1779,111 @@ namespace aspect
...
+      // Create active mesh tables to store the product of the pressure perturbation and
+      // the normalized gravity used in the free surface stabilization.
+      // Currently, mutilevel is not implemented yet, it may slow down the convergence.
+
+      // TODO: implement multilevel surface terms for the free surface stabilization.
+
+      active_cell_data.apply_stabilization_free_surface_faces = false;
+
+      if (sim.mesh_deformation
+          && !sim.mesh_deformation->get_free_surface_boundary_indicators().empty())
```
gassmoeller, line 1793: This will save you line 1803 below:
```suggestion
      active_cell_data.apply_stabilization_free_surface_faces = sim.mesh_deformation
          && !sim.mesh_deformation->get_free_surface_boundary_indicators().empty();

      if (active_cell_data.apply_stabilization_free_surface_faces == true)
```

source/simulator/stokes_matrix_free.cc
```diff
@@ -1703,6 +1779,111 @@ namespace aspect
...
+        {
+          const double free_surface_theta =
+            sim.mesh_deformation->template get_matching_mesh_deformation_object<MeshDeformation::FreeSurface<dim>>()
+          .get_free_surface_theta();
+
+          // GMG doesn't support melt transport
+          AssertThrow(!sim.parameters.include_melt_transport, ExcNotImplemented());
```
gassmoeller, line 1800: Flip these two lines. Checking for impossible conditions should go before anything you do.

Also shouldn't this go into the constructor near line 1368? Or is this parameter not yet set at that time?

source/simulator/stokes_matrix_free.cc
```diff
@@ -2544,6 +2759,14 @@ namespace aspect
       additional_data.mapping_update_flags = (update_values | update_gradients |
                                               update_JxW_values | update_quadrature_points);
 
+      if (sim.mesh_deformation
+          && !sim.mesh_deformation->get_free_surface_boundary_indicators().empty())
+        additional_data.mapping_update_flags_boundary_faces =
+          (update_values  |
```
gassmoeller, line 2765: Why do you not need the gradients here, but you do above?

cookbooks/S20RTS_gmg.prm
```diff
@@ -0,0 +1,34 @@
...
+  end
+end
+
+subsection Material model
+  set Material averaging =  harmonic average
+end
+
+subsection Postprocess
+  subsection Visualization
+    set Output mesh velocity = true
```
gassmoeller, line 19: I do not understand this output, shouldnt this always be 0 because this cookbook does not use a free surface?

gassmoeller (commented):
Very nice functionality! I have some comments about the code style that should be simple to address. I did not see anything wrong, but I have to admit the matrix free code in general is just hard to read (lots of implicit knowledge about the order of face contributions), but that is something you cannot change.

Concerning the parameter files: I would prefer we create as few duplicates as possible, they only confuse users that go into those directories. So what I would suggest is the following: 
- Replace the existing S20RTS cookbook with one that uses GMG (why is this part of this PR anyway? the current S20RTS cookbook doesnt use mesh deformation, is this a leftover from the work on nullspace constraints?).
- Only add a single parameter file to the crameri benchmark named crameri_benchmark_1_gmg.prm which is as similar as possible to the existing parameter file, i.e. it is the same benchmark, just with gmg. The current parameter files are neither tests (too expensive), nor real benchmarks (too short and different parameters than the original file). They were probably useful during development, but I don't think they would be useful once this is merged.

Does that make sense?

zjiaqi2018:
> (why is this part of this PR anyway? the current S20RTS cookbook doesnt use mesh deformation, is this a leftover from the work on nullspace constraints?).

This is one of your suggestions when Timo asked for tests for GMG+free surface :)
If it's not useful, I will remove it.

gassmoeller:
I would guess that my suggestion was to add the free surface to that cookbook ;-). But even without the free surface I think it makes sense to replace the current form of that cookbook with the GMG, so feel free to replace the existing AMG version with a GMG version (or remove the gmg parameter file).

In any case the Crameri benchmark is a good case for the free surface.

tjhei:
@zjiaqi2018 did you do the modifications Rene is asking about? I think it is easiest to delete the S20RTS files you added for example.

source/simulator/stokes_matrix_free.cc
zjiaqi2018: Forgot to remove it.

zjiaqi2018:
I removed S20RTS files.

source/simulator/stokes_matrix_free.cc
zjiaqi2018: Because we don't need it in the matrix-free operator, material values are precomputed and stored.

source/simulator/stokes_matrix_free.cc
tjhei: Jiaqi, when you get a question like this, it is better to address it by adding a comment in the code. This is something the next person looking at the code might ask as well.

tjhei:
@gassmoeller are you ok with the current state? If not, what needs changing?

gassmoeller (approved):
Looks good, thanks!

tjhei:
Awesome, thank you @gassmoeller and @zjiaqi2018 !

PR #4456, opened 2022-01-20
Rework the gravity postprocessor

bangerth:
@cedrict complained that the gravity postprocessor is slow. This patch reworks a good bit of it to make it a bit easier to read and (slightly) more efficient. I have more work to do regarding efficiency, but thought I'd get this out of the way already. It's probably easiest to read side-by-side.

I've broken the work into self-contained commits that I've individually tested. I think that also makes it easier to follow the flow of the transformations I've applied.

/rebuild

### DIFF
--- a/source/postprocess/gravity_point_values.cc
+++ b/source/postprocess/gravity_point_values.cc
@@ -326,17 +326,17 @@ namespace aspect
       std::vector<Tensor<1,dim>>          local_g_anomaly (n_satellites);
       std::vector<SymmetricTensor<2,dim>> local_g_gradient (n_satellites);
 
-      for (unsigned int p=0; p < n_satellites; ++p)
-        {
-          const Point<dim> satellite_position = satellite_positions_cartesian[p];
-
-          // For each point (i.e. satellite), the fourth integral goes over cells and
-          // quadrature points to get the unique distance between those, to calculate
-          // gravity vector components x,y,z (in tensor), potential and gradients.
-          local_cell_number = 0;
-          for (const auto &cell : this->get_dof_handler().active_cell_iterators())
-            if (cell->is_locally_owned())
+      // For each point (i.e. satellite), the fourth integral goes over cells and
+      // quadrature points to get the unique distance between those, to calculate
+      // gravity vector components x,y,z (in tensor), potential and gradients.
+      local_cell_number = 0;
+      for (const auto &cell : this->get_dof_handler().active_cell_iterators())
+        if (cell->is_locally_owned())
+          {
+            for (unsigned int p=0; p < n_satellites; ++p)
               {
+                const Point<dim> satellite_position = satellite_positions_cartesian[p];
+
                 for (unsigned int q = 0; q < n_quadrature_points_per_cell; ++q)
                   {
                     const unsigned int array_index = local_cell_number * n_quadrature_points_per_cell + q;
@@ -365,9 +365,10 @@ namespace aspect
                                                                 * r_vector[e] * r_vector[f]
                                                                 - (e==f ? r_squared : 0));
                   }
-                ++local_cell_number;
               }
-        }
+
+            ++local_cell_number;
+          }
 
       for (unsigned int p=0; p < n_satellites; ++p)
         {
--- a/source/postprocess/gravity_point_values.cc
+++ b/source/postprocess/gravity_point_values.cc
@@ -98,7 +98,7 @@ namespace aspect
       else n_satellites = 1;
 
       // *** Second assign the coordinates of all satellites:
-      satellites_coordinate.resize(n_satellites);
+      satellite_positions_spherical.resize(n_satellites);
       if (sampling_scheme == map)
         {
           unsigned int p = 0;
@@ -109,17 +109,17 @@ namespace aspect
                   for (unsigned int j=0; j < n_points_latitude; ++j)
                     {
                       if (n_points_radius > 1)
-                        satellites_coordinate[p][0] = minimum_radius + ((maximum_radius - minimum_radius) / (n_points_radius - 1)) * h;
+                        satellite_positions_spherical[p][0] = minimum_radius + ((maximum_radius - minimum_radius) / (n_points_radius - 1)) * h;
                       else
-                        satellites_coordinate[p][0] = minimum_radius;
+                        satellite_positions_spherical[p][0] = minimum_radius;
                       if (n_points_longitude > 1)
-                        satellites_coordinate[p][1] = (minimum_colongitude + ((maximum_colongitude - minimum_colongitude) / (n_points_longitude - 1)) * i) * numbers::PI / 180.;
+                        satellite_positions_spherical[p][1] = (minimum_colongitude + ((maximum_colongitude - minimum_colongitude) / (n_points_longitude - 1)) * i) * numbers::PI / 180.;
                       else
-                        satellites_coordinate[p][1] = minimum_colongitude * numbers::PI / 180.;
+                        satellite_positions_spherical[p][1] = minimum_colongitude * numbers::PI / 180.;
                       if (n_points_latitude > 1)
-                        satellites_coordinate[p][2] = (minimum_colatitude + ((maximum_colatitude - minimum_colatitude) / (n_points_latitude - 1)) * j) * numbers::PI / 180.;
+                        satellite_positions_spherical[p][2] = (minimum_colatitude + ((maximum_colatitude - minimum_colatitude) / (n_points_latitude - 1)) * j) * numbers::PI / 180.;
                       else
-                        satellites_coordinate[p][2] = minimum_colatitude * numbers::PI / 180.;
+                        satellite_positions_spherical[p][2] = minimum_colatitude * numbers::PI / 180.;
                       ++p;
                     }
                 }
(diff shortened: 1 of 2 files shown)


bangerth:
I've finished what I meant to do by pushing a whole bunch of additional commits. I imagine that it is impossible to read the final result, so left it as individual, self-contained commits.

The main changes in the last few commits are:
* In the old code, we first loop over all cells and create a few rather large data structures that store all of the information from all cells. Then we loop over evaluation points, within which we loop over cells, and compute something from these large data structure. The new code just turns these loops around: loop over all cells, evaluate stuff and store it in a single small data structure (the `MaterialModelOutputs`), loop over evaluation points and compute what we need to compute. This should be far more efficient, and for sure more cache friendly.
* The existing code did a bunch of MPI sums for each evaluation point. These can be blocked into a few MPI sums over arrays of size `n_evaluation_points` (`==n_satellites`). One just has to be a bit creative. In any case, this should reduce communication by a rather large factor when we have a lot of evaluation points.

Each commit individually clears the gravity tests in the test suite.

source/postprocess/gravity_point_values.cc
```diff
@@ -240,17 +150,210 @@ namespace aspect
...
+                const Point<dim> satellite_position = satellite_positions_cartesian[p];
+
+                for (unsigned int q = 0; q < n_quadrature_points_per_cell; ++q)
+                  {
+                    const Tensor<1,dim> r_vector = satellite_position - fe_values.quadrature_point(q);
+
+                    const double r_squared = r_vector.norm_square();
+                    const double r = std::sqrt(r_squared);
+
+                    const double density_JxW = out.densities[q] * fe_values.JxW(q);
```
gassmoeller, line 272: Did you measure if the new implementation is faster for a reasonably large number of satellites (say 1000 or so)? I think part of why the old implementation had this complicated split loop was in order to compute `density_JxW` exactly once per quadrature point. With your current implementation you compute it `n_satellites * n_quadrature_points` times. Is it possible to reorder the loops over satellites and quadrature points to avoid this?
I can do a complete review later, I have some ideas for smaller optimizations, but wanted to ask if you benchmarked it first.

bangerth: I could move this out, but the real cost is *computing* `JxW` and the densities. For `JxW`, for example, you have to evaluate the gradients of the shape functions of a Q4 mapping at the current quadrature point, multiply these gradients by the locations of the support points, and compute the determinant of the resulting matrix. All of this is going to cost hundreds of FP operations. Is your question whether the multiplications here are too expensive if one has a lot of satellites? I guess that could be -- I could just move this outside the loop over the satellites.

bangerth: I should probably also move `r_squared` and `r` out of the loop.

bangerth: Ah no, I can't, it depends on `q` :-(

gassmoeller (commented):
I think with the small modification I suggested this looks good. I am somewhat suspicious about using `std::pow(double, int)` for small integers a lot, probably just because I dont trust it to be properly optimized, but unless we have evidence it is slow no need to worry for now. Did you measure that your new implementation is actually faster?

source/postprocess/gravity_point_values.cc
gassmoeller: >  Is your question whether the multiplications here are too expensive if one has a lot of satellites? I guess that could be -- I could just move this outside the loop over the satellites.

Yes that was my question, please do move it out of the loop.

bangerth:
I pulled two computations out of the inner loop.

bangerth:
I took the `gravity_point_values_spiral.prm` test, increased the number of "points per spiral" to 10,000, and then let valgrind's cachegrind tool count instructions and cache misses. In release mode, this runs for some 20 seconds, and the postprocessors take a combined 65% or so of the overall run time. Here is what I get for before this series of patches:
```
==45404== I   refs:      4,931,419,113
==45404== I1  misses:        7,346,435
==45404== LLi misses:           58,608
==45404== I1  miss rate:          0.15%
==45404== LLi miss rate:          0.00%
==45404== 
==45404== D   refs:      1,783,092,496  (1,200,750,400 rd   + 582,342,096 wr)
==45404== D1  misses:       46,716,070  (   43,437,345 rd   +   3,278,725 wr)
==45404== LLd misses:        1,635,945  (    1,316,164 rd   +     319,781 wr)
==45404== D1  miss rate:           2.6% (          3.6%     +         0.6%  )
==45404== LLd miss rate:           0.1% (          0.1%     +         0.1%  )
==45404== 
==45404== LL refs:          54,062,505  (   50,783,780 rd   +   3,278,725 wr)
==45404== LL misses:         1,694,553  (    1,374,772 rd   +     319,781 wr)
==45404== LL miss rate:            0.0% (          0.0%     +         0.1%  )
```
And after the series (that is, the current state of the patch):
```
==44875== I   refs:      5,048,254,350
==44875== I1  misses:        4,720,751
==44875== LLi misses:           58,667
==44875== I1  miss rate:          0.09%
==44875== LLi miss rate:          0.00%
==44875== 
==44875== D   refs:      1,769,385,416  (1,182,858,683 rd   + 586,526,733 wr)
==44875== D1  misses:       44,168,063  (   41,157,388 rd   +   3,010,675 wr)
==44875== LLd misses:        1,644,094  (    1,316,288 rd   +     327,806 wr)
==44875== D1  miss rate:           2.5% (          3.5%     +         0.5%  )
==44875== LLd miss rate:           0.1% (          0.1%     +         0.1%  )
==44875== 
==44875== LL refs:          48,888,814  (   45,878,139 rd   +   3,010,675 wr)
==44875== LL misses:         1,702,761  (    1,374,955 rd   +     327,806 wr)
==44875== LL miss rate:            0.0% (          0.0%     +         0.1%  )
```
That is, the number of instructions executed has gone up by 2.4% but both the number of instruction and data cache misses have gone down. I've tried to measure the time difference in run time, and it is within the margin of the number of instructions: the best of 5 times before the patch is 0.550, after the patch it is 0.560 seconds. That's a difference of 2%.

So for the moment, the patch only improves code structure and readability, but not speed -- I think that will only come from fundamentally rethinking the algorithm. I would love to hear anyone pinpoint where the difference of 2% comes from!

gassmoeller:
I ran the same test in release model with your branch and modifications through valgrind and visualized the result with kcachegrind (I also get around 63% of the time for the postprocessor). Around 65% of the time of the postprocessor is spent in std::ostream. I think this is a bit pathological, because the test has very few cells and many satellites, so it spends most of the time writing the result into the output stream. A further 19% (so about half of the rest) is spent inside `std::pow`. I don't know why this is not optimized for the types of calls we have here (`std::pow(double,int)`), but it further supports my suspicion that `r*r` or `r*r*r` is simply vastly faster than `std::pow(r,2)`. On average every call to `std::pow` needed 125 instructions for me. So I wouldn't worry about the 2%, but replacing `std::pow` especially in the innermost loops should give you around 50% speedup in the computation part.


screenshot

bangerth:
Wow, unbelievable. Replacing `std::pow` in the last commit here reduces the overall number of instructions by 20%:
```
==139913== I   refs:      4,125,925,371
==139913== I1  misses:        4,070,264
==139913== LLi misses:           58,533
==139913== I1  miss rate:          0.10%
==139913== LLi miss rate:          0.00%
==139913== 
==139913== D   refs:      1,487,320,486  (986,988,721 rd   + 500,331,765 wr)
==139913== D1  misses:       43,625,298  ( 40,588,346 rd   +   3,036,952 wr)
==139913== LLd misses:        1,643,346  (  1,315,543 rd   +     327,803 wr)
==139913== D1  miss rate:           2.9% (        4.1%     +         0.6%  )
==139913== LLd miss rate:           0.1% (        0.1%     +         0.1%  )
==139913== 
==139913== LL refs:          47,695,562  ( 44,658,610 rd   +   3,036,952 wr)
==139913== LL misses:         1,701,879  (  1,374,076 rd   +     327,803 wr)
==139913== LL miss rate:            0.0% (        0.0%     +         0.1%  )
```

bangerth:
I got it down another insignificant amount by using a temporary `ostringstream` for output:
```
==22061== I   refs:      4,103,757,265
==22061== I1  misses:        4,113,690
==22061== LLi misses:           58,600
==22061== I1  miss rate:          0.10%
==22061== LLi miss rate:          0.00%
==22061== 
==22061== D   refs:      1,478,373,709  (980,161,656 rd   + 498,212,053 wr)
==22061== D1  misses:       44,089,349  ( 40,785,873 rd   +   3,303,476 wr)
==22061== LLd misses:        1,885,272  (  1,315,902 rd   +     569,370 wr)
==22061== D1  miss rate:           3.0% (        4.2%     +         0.7%  )
==22061== LLd miss rate:           0.1% (        0.1%     +         0.1%  )
==22061== 
==22061== LL refs:          48,203,039  ( 44,899,563 rd   +   3,303,476 wr)
==22061== LL misses:         1,943,872  (  1,374,502 rd   +     569,370 wr)
==22061== LL miss rate:            0.0% (        0.0%     +         0.1%  )
```

Nice lessons learned here. We should go hunt for more places where we use `std::pow` with integer exponents -- nice hunch to be suspicious of that!

gassmoeller (approved):
Thanks, looks good!

PR #4459, opened 2022-01-22
add pressure- and temperature-dependent thermal conductivity

jdannberg:
This pull request adds the option to use a pressure- and temperature-dependent thermal conductivity in the steinberger material model. As far as I know, this is the first material model that will have this option. As the thermal conductivity can change from around 2-4 W/m/K near the surface up to 10 W/m/K near the core-mantle boundary, this could make a big difference for the heat flux. 

There are two published formulations/parameter sets that I have tested that can be reproduced with this implementation:
(1) from Tosi et al. (2013), which is a general formulation for the whole mantle that has been used in geodynamic models already, 
(2) from Stackhouse et al. (2015), which is a more recent study for the lower mantle based on ab-initio simulations. 
If the parameter values are changed, it would be possible to fit other data as well. 

* [x] I have [created two test cases](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/main/tests/) directory.


source/material_model/steinberger.cc
```diff
@@ -220,6 +220,46 @@ namespace aspect
...
+
+      else if (conductivity_formulation == p_T_dependent)
+        {
+          // Find the conductivity layer that corresponds to the depth of the evaluation point.
+          const double depth = this->get_geometry_model().depth(position);
+          unsigned int layer_index = std::distance(conductivity_transition_depths.begin(),
+                                                   std::lower_bound(conductivity_transition_depths.begin(),conductivity_transition_depths.end(), depth));
+
+          // Make sure the temperature is positive.
+          const double T = std::max(temperature, 1.e-3 * conductivity_reference_temperatures[layer_index]);
```
gassmoeller, line 241: Shouldnt this then be:
```suggestion
          const double T = std::max(temperature, 0.0);
```
Why did you choose 1e-3 * ref_temp?

source/material_model/steinberger.cc
```diff
@@ -220,6 +220,46 @@ namespace aspect
...
+          const double T_dependence = std::pow(conductivity_reference_temperatures[layer_index] / T, conductivity_exponents[layer_index]);
+          const double p_dependence = reference_thermal_conductivities[layer_index] + conductivity_pressure_dependencies[layer_index] * pressure;
+
+          // Function based on the theory of Roufosse and Klemens (1974) that accounts for saturation.
+          // For the Tosi formulation, the scaling should be zero so that this term is 1.
+          double saturation_function = 1.0;
+          if (1./T_dependence > 1)
```
gassmoeller, line 248:
```suggestion
          if (1./T_dependence > 1.)
```

gassmoeller (commented):
Except for my two comments this looks good to go. Thanks for the improvements!

source/material_model/steinberger.cc
jdannberg: We can choose a different number, but the temperature needs to be positive, it can't be zero, because we divide by this temperature in the lines below. So if I choose a really small number, then the conductivity will get some crazy high value, and I thought that might become a problem for the model (for example, in our 3D models the temperature is sometimes below zero due to over- and undershoots, and we don't want the conductivity to be 1e200 is these cases). So I picked something that would give us a conductivity around 3 orders of magnitude larger than at the reference temperature (which is probably already more than what anyone would want in terms of variations). It's not great, but I don;t have a better idea. Instead, we could also put a maximum conductivity in the input file (and then set it to that value if T is small). 

But let me know if you have a better idea!

jdannberg:
I've made the maximum thermal conductivity an input parameter so that I don't need a hard-coded arbitrary number to limit the conductivity. I've also changed one of the test cases to have really small temperature values. I would have liked to go all the way to zero, but that triggers a floating point exception for the viscosity (which might be a separate issue to look into). 

Does this address your comments?

source/material_model/steinberger.cc
```diff
@@ -237,19 +237,23 @@ namespace aspect
...
-          // Make sure the temperature is positive.
-          const double T = std::max(temperature, 1.e-3 * conductivity_reference_temperatures[layer_index]);
-          const double T_dependence = std::pow(conductivity_reference_temperatures[layer_index] / T, conductivity_exponents[layer_index]);
           const double p_dependence = reference_thermal_conductivities[layer_index] + conductivity_pressure_dependencies[layer_index] * pressure;
 
+          // Make reasonably sure we will not compute any invalid values due to the temperature-dependence.
+          // Since both the temperature-dependence and the saturation term scale with (Tref/T), we have to
+          // make sure we can compute the square of this number. If the temperature is small enough to
+          // be close to yielding NaN values, the conductivity will be set to the maximum value anyway.
+          const double T = std::max(temperature, sqrt(std::numeric_limits<double>::min() * conductivity_reference_temperatures[layer_index]));
```
gassmoeller, line 246: Yes, this looks better, but I think you meant this:
```suggestion
          const double T = std::max(temperature, std::sqrt(std::numeric_limits<double>::min()) * conductivity_reference_temperatures[layer_index]);
```
In the current form if conductivity_reference_temperatures < 1.0 the square root would be 0.

Ready to go otherwise.

gassmoeller (approved):

PR #4460, opened 2022-01-22
CMake: let's not forget tests set up

### LINKED ISSUE #4458: Running cmake lets us forget about all tests currently set up
When you make a commit, the next build invokes `cmake`, which in turns somehow makes us forget about all of the tests we had previously set up. As a consequence, command lines such as
```
  ninja && ctest -j16 -R gravity
```
fail because `ctest` reports that "there are no tests". One has to run
```
  ninja setup_tests
```
again, which is time consuming and, furthermore, breaks the nicety of the first command line above to which one can simply come back a minute or two later to see whether the tests in question succeeded.

@tjhei FYI

tjhei:
When rerunning cmake we forgot to tell ctest about the tests inside the
tests/ project. Fix this.

Also, make sure we remember ASPECT_RUN_ALL_TESTS in the test project
(just in case the test project reconfigures for some reason).

fixes #4458


bangerth (approved):

bangerth:
Thanks!

PR #4461, opened 2022-01-25
fix radial visc simple data file

jdannberg:
This pull request fixes the "simple" version of the steinberger viscosity profile data file. It now correctly excludes both the top and bottom boundary layer. These need to be removed from the original profile in case the model uses a temperature-dependent viscosity based on the adiabatic profile (rather than the laterally averaged temperature). Otherwise the temperature effect will be included twice (in case the laterally averaged temperature is used, there is still the other version of the file: [radial-visc.txt](https://github.com/geodynamics/aspect/blob/main/data/material-model/steinberger/radial-visc.txt)). The old version of the radial-visc-simple.txt file had excluded the top, but not the bottom boundary layer:

radial-visc
For comparison: radial-visc.txt (with both boundary layers). This file remains unchanged. 

&nbsp;

radial-visc-2 | radial-visc-2
:-------------------------:|:-------------------------:
Original radial-visc-simple.txt file. The top boundary layer is removed, but the bottom boundary layer is still there. | New radial-visc-simple.txt file in this pull request. Both the top and bottom boundary layers are removed. 

------------------------------------------------------
As another slightly related point: I haven't found any cookbook where this is used, or any place where it would be documented how to use these files. There is a [description.txt](https://github.com/geodynamics/aspect/blob/main/data/material-model/steinberger/description.txt) file with a very short overview in the same directory. If we don't find another good place for where to put that explanation, I am happy to add it to that file, or we could add that explanation to the description of the material model, or even to the description of the parameter itself. 

------------------------------------------------------


gassmoeller (approved):
Yes this makes sense, thanks for updating the data. Improving the description.txt in a follow up PR would also be welcome. There is an unfinished cookbook for this material model under cookbooks/future/steinberger.prm. I never finished that, which is why there is no better description of how to use these files.

PR #4462, opened 2022-01-25
use new "density" field type for density fields

jdannberg:
This is the promised follow-up to #4450 and #4451. Everywhere in the code where we were asking for a field named "density_field", we now check for the compositional field type "density" instead. I hope I found everything. 

This is also backwards compatible, since a field that is called "density_field" but has the type unspecified will be converted to type density upon reading in the parameters. 


gassmoeller (approved):
Very nice, thank you!

PR #4464, opened 2022-01-26
fix periodic boundary conditions in compressible models

jdannberg:
Compressible models with periodic boundaries did not work before. The ones I tried always crashed in the Stokes solver or had very large iteration counts. The problem was that we did not do the pressure right-hand side compatibility modification in case a model had periodic boundaries. This PR fixes this. 
I added a very simple test case, but I also tested this with a more complex model (spherical geometry, nullspace removal, strong viscosity contrast, material properties read in from a look-up table) and it worked as well. 

### DIFF
--- a/doc/modules/changes/20220126_jdannberg
+++ b/doc/modules/changes/20220126_jdannberg
@@ -1,3 +1,9 @@
-Fixed: Periodic boundaries now work correctly in compressible models.
+Fixed: Compressible models with periodic boundaries that did not also
+have one or more open boundaries did not correctly apply the necessary
+pressure right-hand side compatibility modification when solving the
+Stokes equations. This generally caused the linear solver to fail or to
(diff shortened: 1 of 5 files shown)


doc/modules/changes/20220126_jdannberg
```diff
@@ -0,0 +1,3 @@
+Fixed: Periodic boundaries now work correctly in compressible models.
```
gassmoeller, line 1: Please add a bit more precise description of the problem, e.g.
```suggestion
Fixed: Compressible models with periodic boundaries, but without open boundaries would not correctly apply the necessary Stokes right-hand side correction. This often caused the Stokes solver to not converge. The right hand side correction is now applied correctly, fixing the convergence problems.
```

jdannberg:
Fixed.

gassmoeller (approved):

PR #4465, opened 2022-01-27
Do not create unnecessary failing assembler for compressible Newton models with full A block preconditioner

gassmoeller:
This was a bug in debug mode I noticed in one of @elodie-kendall's models. If we set up a compressible model with a Newton solver and use the full A block of the Stokes matrix for the preconditioner, then the NewtonHandler still adds an assembler for the compressible terms in the preconditioner, although the preconditioner matrix is not used (the full matrix is used instead). Additionally the added assembler will throw an assertion. This change unifies the code paths between the Newton solver and the regular solver (which does not add the assembler for these models).


jdannberg (approved):
This looks good to me. Thanks for the fix! 
Since it's apparently a bugfix, do you want to add a changelog entry?

gassmoeller:
Changelog added.

PR #4467, opened 2022-01-28
Adds first publication of the year to the publication list

ljhwang:
First publication of 2022 !


gassmoeller (approved):
Thanks for the update!

PR #4468, opened 2022-01-28
add steinberger cookbook

jdannberg:
This pull request adds a new cookbook using the Steinberger material model explaining

1. How to use the projected density approximation, 
2. How to read in a material look-up table file, and how this file needs to be structured (as far as I know, this format isn't documented anywhere else), 
3. How to read in the files that determine the viscosity profile (which was what @gassmoeller suggested as a follow-up to #4461), 
4. How to use periodic boundary conditions in a spherical shell with nullspace removal. 

There is one change in the steinberger material model that was required to get this to work (using the adiabatic pressure rather than the real pressure to compute material properties), and I am happy to pull that out into a separate pull request if you want me to do that. 

To make reviewing this easier, I am attaching the relevant pages of the manual as pdf.  
[manual.pdf](https://github.com/geodynamics/aspect/files/7955409/manual.pdf)


jdannberg:
Now that #4469 is merged, I've rebased this branch and it should be ready for someone to review.

cookbooks/steinberger/steinberger.prm
```diff
@@ -0,0 +1,224 @@
+# This is a cookbook for a convection model with realistic material properties.
+# Specifically, it uses material properties read in from a look-up table
+# computed with mineral physics software and the projected density approximation.
+# Since it is a quarter of a apherical shell, it also uses periodic boundary
+# conditions.
+
+set Dimension                              = 2
+set End time                               = 3e8
+set Adiabatic surface temperature          = 1613.0
+set Output directory                       = steinberger
```
gassmoeller, line 10: to follow the other cookbooks this should be `output-steinberger`

cookbooks/steinberger/steinberger.prm
```diff
@@ -0,0 +1,224 @@
...
+
+# Because the model is compressible, we use a nonlinear solver.
+# We here choose a tolerance of 1e-3 to limit model runtime, but in a
+# research application one would test what tolerance would be required
+# to obtain a sufficiently accurate solution, and 1e-3 is quite a high
+# value.
+set Nonlinear solver scheme                = iterated Advection and Stokes
+set Nonlinear solver tolerance             = 1e-3
+
+# We run a 2d convection model in a quarterof a spherical shell.
```
gassmoeller, line 20:
```suggestion
# We run a 2d convection model in a quarter of a spherical shell.
```

cookbooks/steinberger/steinberger.prm
```diff
@@ -0,0 +1,224 @@
...
+    # software providing density, thermal expansivity and specific heat.
+    set Data directory                                = $ASPECT_SOURCE_DIR/data/material-model/steinberger/
+    set Lateral viscosity file name                   = temp-viscosity-prefactor.txt
+    set Material file names                           = pyr-ringwood88.txt
+    set Radial viscosity file name                    = radial-visc-simple.txt
+
+    # Because of the resolution in this cookbook we limit the overall viscosity
+    # between 1e20 Pa s and 5e23 Pa s and the lateral viscosity variations to
+    # three orders of magnitude in both directions (for a total of six orders of
+    # magnitude. This allows the features of the flow field to be resolved.
```
gassmoeller, line 99: I dont think this is how it works. Should `Maximum viscosity` and `Minimum viscosity` put a hard limit on viscosity independent of temperature variations?

cookbooks/steinberger/steinberger.prm
```diff
@@ -0,0 +1,224 @@
...
+    # In the Earth, we would expect higher viscosities in the lithosphere and
+    # lower viscosities in plumes and near the core-mantle boundaries.
+    # In addition, this model only accounts for diffusion creep, so the
+    # lithosphere has a high viscosity and forms a stagnant lid on top of the
+    # model. In order to have more realistic subduction in a model like this,
+    # one would have to either prescribe plate velocities at the surface
+    # (forcing plate to subduct) or take into account plastic yielding (so that
```
gassmoeller, line 106:
```suggestion
    # (forcing plates to subduct) or take into account plastic yielding (so that
```

cookbooks/steinberger/steinberger.prm
```diff
@@ -0,0 +1,224 @@
...
+
+    # We base our viscosity profile on the adiabat computed based on the
+    # material porperties used in the model rather than the laterally
+    # averaged temperature.
+    set Use lateral average temperature for viscosity = false
+
+    # We want to interpolate between data points in the look-up table rather
+    # than simply taking the value of the closest point, since this is more
+    # accurate.
+    set Bilinear interpolation                        = true
```
gassmoeller, line 121: This is true by default anyway, can we remove these lines? Ideally I would like to remove the whole parameter and just always interpolate, I dont think there is a reason to not do it (I think I implemented this at the beginning, because I wasnt sure I implemented the interpolation correctly, yes it is that old :-)).

gassmoeller (approved):
First batch of comment on the parameter file. I will look at the .tex file later. Please put a test in the tests directory that just includes the cookbook .prm at a low resolution for a timestep or two. That way we will always be sure the cookbook still runs.

gassmoeller (changes requested):
Sorry, meant to request changes, not approve.

bangerth:
Can you rebase?

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+
+In this cookbook we will go one step further from the last one and set up a fully compressible mantle convection model using the projected density approximation (where the density is interpolated onto the finite element grid to compute the density gradients in the mass conservtaion equation rather than approximating these gradients using a reference profile or tempertaure/pressure derivatives of the density, see \cite{gassmoller2020formulations}). To compute the material properties, we read in a look-up table of material properties in dependence of temperature and pressure computed using a mineral physics software (in this case, Perple\_X, \cite{connolly2005computation}). The table is based on the thermodynamic database by \cite{stixrude2011thermodynamics} and a pyrolitic composition \cite{ringwood1988nature}. Compared to a 1D profile, a temperature--pressure look-up table has the advantage that material properties are accurate not only around one reference adiabat, but also for strongly deviating pressures and temperatures. This is particularly important at phase transitions, because their depth depends on the temperature and pressure.
```
bangerth, line 6:
```suggestion
In this cookbook we will go one step further from the last one and set up a fully compressible mantle convection model using the projected density approximation (where the density is interpolated onto the finite element grid to compute the density gradients in the mass conservation equation rather than approximating these gradients using a reference profile or temperature/pressure derivatives of the density, see \cite{gassmoller2020formulations}). To compute the material properties, we read in a look-up table of material properties in dependence of temperature and pressure, originally computed using a mineral physics software (in this case, Perple\_X, \cite{connolly2005computation}). The table is based on the thermodynamic database of \cite{stixrude2011thermodynamics} and a pyrolitic composition \cite{ringwood1988nature}. Compared to a 1D profile, a temperature-pressure look-up table has the advantage that material properties are accurate not only around one reference adiabat, but also for strongly deviating pressures and temperatures. This is particularly important at phase transitions, because their depth depends on the temperature and pressure. 
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+
+This cookbook also demonstrates how to read in a viscosity profile from a data file. Specifically, we use the profile and lateral viscosity variations due to temperature from \cite{stca06}, which are based on mineral physics constraints and surface observations. 
+
+In addition, this cookbook shows the use of periodic boundary conditions. 
+
+\paragraph{Geometry and periodic boundaries.}
+The model setup is a quarter spherical shell with periodic side boundaries. The inner and outer radius are 3481~km and 6371~km, respectively, so that the mantle is 2900~km deep. In the same section of the input file, we also need to specify that the model should have periodic boundaries in angular ($\phi$) direction:
```
bangerth, line 13:
```suggestion
The model setup is a quarter spherical shell with periodic side boundaries. The inner and outer radius are \SI{3481}{km} and \SI{6371}{km}, respectively, so that the mantle is \SI{2900}{km} deep. In the same section of the input file, we also need to specify that the model should have periodic boundaries in angular ($\phi$) direction:  
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+
+In addition, this cookbook shows the use of periodic boundary conditions. 
+
+\paragraph{Geometry and periodic boundaries.}
+The model setup is a quarter spherical shell with periodic side boundaries. The inner and outer radius are 3481~km and 6371~km, respectively, so that the mantle is 2900~km deep. In the same section of the input file, we also need to specify that the model should have periodic boundaries in angular ($\phi$) direction:  
+\lstinputlisting[language=prmfile]{cookbooks/steinberger/doc/geometry.part.prm.out}
+Both the top and bottom boundaries allow for free slip. Because the model has periodic side boundary conditions and free slip boundaries at top and bottom, the amount of rigid-body rotation in $\phi$ direction is not constrained. In other words: There is no unique solution. \aspect{} allows it to remove this nullspace from the model (see Section~\ref{sec:nullspace}). Here, we do this by setting the net rotation to zero.
```
bangerth, line 15:
```suggestion
Both the top and bottom boundaries allow for free slip. Because the model has periodic side boundary conditions and free slip boundaries at top and bottom, the amount of rigid-body rotation in $\phi$ direction is not constrained. In other words: There is no unique solution. \aspect{} can remove this nullspace from the model (see Section~\ref{sec:nullspace}). Here, we do this by setting the net rotation to zero:
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+
+In addition, this cookbook shows the use of periodic boundary conditions. 
+
+\paragraph{Geometry and periodic boundaries.}
+The model setup is a quarter spherical shell with periodic side boundaries. The inner and outer radius are 3481~km and 6371~km, respectively, so that the mantle is 2900~km deep. In the same section of the input file, we also need to specify that the model should have periodic boundaries in angular ($\phi$) direction:  
+\lstinputlisting[language=prmfile]{cookbooks/steinberger/doc/geometry.part.prm.out}
+Both the top and bottom boundaries allow for free slip. Because the model has periodic side boundary conditions and free slip boundaries at top and bottom, the amount of rigid-body rotation in $\phi$ direction is not constrained. In other words: There is no unique solution. \aspect{} allows it to remove this nullspace from the model (see Section~\ref{sec:nullspace}). Here, we do this by setting the net rotation to zero. 
+\lstinputlisting[language=prmfile]{cookbooks/steinberger/doc/nullspace.part.prm.out}
+The temperature is fixed to 273~K at the top and 3773~K at the bottom boundary.
```
bangerth, line 17:
```suggestion
The temperature is fixed to \SI{273}{K} at the top and \SI{3773}{K} at the bottom boundary. 
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+\lstinputlisting[language=prmfile]{cookbooks/steinberger/doc/nullspace.part.prm.out}
+The temperature is fixed to 273~K at the top and 3773~K at the bottom boundary. 
+The initial temperature model consists of an adiabatic profile,
+thermal boundary layers at the surface and the core-mantle boundary, 
+and a small harmonic perturbation to initiate convection.
+The gravity profile in the model is based on PREM. 
+
+\paragraph{The equation of state.}
+To use material properties from a temperature--pressure look-up table, we use the Steinberger material model. We have to specify the path to the directory where all the data files we want to use for this model are stored. This includes the files for the viscosity profile, the lateral viscosity variations due to temperature, and all material files containing look-up tables computed by mineral physics software.
```
bangerth, line 24:
```suggestion
To use material properties from a temperature-pressure look-up table, we use the Steinberger material model. We have to specify the path to the directory where all the data files we want to use for this model are stored. This includes the files for the viscosity profile, the lateral viscosity variations due to temperature, and all material files containing look-up tables computed by mineral physics software.
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+For intermediate composition values, material properties will then be averaged based on the mass/volume fractions of the individual compositions. 
+
+In addition, there are a few options we can select about how these look-up tables should be used:
+We can decide between interpolating between data points in the lookup table based on the pressure and temperature at the point we need the material properties for, or we can simply take the value from the table that is closest. In our case, we choose the bilinear interpolation because it is more accurate. 
+Second, we can decide how latent heat should be computed: from the thermal expansivity and specific heat, 
+or from the enthalpy (all three properties should be columns in the look-up table). 
+In some cases the look-up table contains the effective thermal expansivity and specific heat, which includes the effect of phase transformations. In a case like that, we simply want to use these values without performing additional latent heat computations because latent heat is already included automatically when using the properties from the look-up table.
```
bangerth, line 34: I don't understand this last half-sentence. The look-up table provides *static* data: for a given p-T pair, what is the density/thermal expansivity/... On the other hand, latent heat is a *dynamic* effect: What happens (i.e., how much heat is released) if you are *moving* through p-T space.

It is ok to ignore this effect, but I think it is *unrelated* to the tabulated properties.

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+We can decide between interpolating between data points in the lookup table based on the pressure and temperature at the point we need the material properties for, or we can simply take the value from the table that is closest. In our case, we choose the bilinear interpolation because it is more accurate. 
+Second, we can decide how latent heat should be computed: from the thermal expansivity and specific heat, 
+or from the enthalpy (all three properties should be columns in the look-up table). 
+In some cases the look-up table contains the effective thermal expansivity and specific heat, which includes the effect of phase transformations. In a case like that, we simply want to use these values without performing additional latent heat computations because latent heat is already included automatically when using the properties from the look-up table. 
+If the look-up table contains thermal expansivity and specific heat without the effect of phase transitions, then \aspect{} can compute latent heat effects based on the pressure and 
+temperature derivatives of the specific enthalpy (using the approach of \cite{nakagawa2009incorporating}).
```
bangerth, line 36: Do we do this in one of the other cookbooks that one could reference here?

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+In our case, we do not include latent heat in our model, mainly because it can lead to numerical problems.
```
bangerth, line 37: Let's be less ominous :-)
```suggestion
In our case, we simply do not include latent heat at all in our model.
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+So the look-up table is computed without latent heat effects, and we set the ``Latent heat'' parameter to false.
```
bangerth, line 38:
```suggestion
So the look-up table is computed without latent heat effects, and we set the ``Latent heat'' parameter to \texttt{false}. 
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+If the look-up table contains thermal expansivity and specific heat without the effect of phase transitions, then \aspect{} can compute latent heat effects based on the pressure and 
+temperature derivatives of the specific enthalpy (using the approach of \cite{nakagawa2009incorporating}).
+In our case, we do not include latent heat in our model, mainly because it can lead to numerical problems.
+So the look-up table is computed without latent heat effects, and we set the ``Latent heat'' parameter to false. 
+\lstinputlisting[language=prmfile]{cookbooks/steinberger/doc/lookup.part.prm.out}
+In a research application, it would be consistent with the projected density approximation (or any other compressible approximation) to compute latent heat. To avoid the numerical instabilities, either the resolution needs to be fine enough so that each phase transitions is resolved by several mesh cells,
```
bangerth, line 40:
```suggestion
In an actual research application, it would be appropriate (and consistent with the projected density approximation, or any other compressible approximation) to compute latent heat instead of neglecting it as we do. This often leads to numerical instabilities that one typically addresses by ensuring that either the resolution is fine enough so that each phase transitions is resolved by several mesh cells, 
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+In our case, we do not include latent heat in our model, mainly because it can lead to numerical problems.
+So the look-up table is computed without latent heat effects, and we set the ``Latent heat'' parameter to false. 
+\lstinputlisting[language=prmfile]{cookbooks/steinberger/doc/lookup.part.prm.out}
+In a research application, it would be consistent with the projected density approximation (or any other compressible approximation) to compute latent heat. To avoid the numerical instabilities, either the resolution needs to be fine enough so that each phase transitions is resolved by several mesh cells, 
+or the energy equations needs to be solved for entropy instead of pressure (which is an option available in \aspect{}; in this case, the look-up table needs to be given in terms of entropy and pressure). 
+
+\paragraph{The look-up table format.}
+The format of these look-up table is implemented in the \href{https://aspect.geodynamics.org/doc/doxygen/classaspect_1_1MaterialModel_1_1MaterialUtilities_1_1Lookup_1_1MaterialLookup.html}{aspect::MaterialModel::MaterialUtilities::Lookup::MaterialLookup} class.
```
bangerth, line 44:
```suggestion
The format of these look-up table is described in the documentation of the \href{https://aspect.geodynamics.org/doc/doxygen/classaspect_1_1MaterialModel_1_1MaterialUtilities_1_1Lookup_1_1MaterialLookup.html}{aspect::MaterialModel::MaterialUtilities::Lookup::MaterialLookup} class. 
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+This file format is the default, it is also the more flexible format and it is what is used in this cookbook. Since the only requirements for the format are the header and the order of some of the columns, 
+files created with other mineral physics software can also be converted to this format.  
+
+The Perple\_X header contains the following in the first 13 lines: 
+The Perple\_X version, the name of the data table, the dimensions of the data table (for example, for a table with one dimension being pressure, the other temperature, this would be 2), the variable in the first dimension (this either needs to be `T(K)' for temperature, or `P(bar)' for pressure), the minimum value of this variable, the increments this variable will be increased with in the table, and the number of different values of this variable the table contains, the second variable (again with minimum, increment and number of values), the number of material properties in the table, and finally, the names of the columns. The first two columns need to be the pressure and temperature (in any order). The other required column names are:
```
bangerth, line 51: Can you just convert this into a numbered list? I think this would be easier to read

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+`rho,kg/m3' (for the density), `alpha,1/K' (for the thermal expansivity), `cp,J/K/kg' (for the specific heat), `vp,km/s' (for the P-wave velocity), `vs,km/s' (for the S-wave velocity)), `h,J/kg' (for the specific enthalpy).
```
bangerth, line 52: You start with `T(K) for the temperature` and `P(bar)`, but later have `rho,kg/m3 (for the density)`. Pick one format.

Also, mark up units with `\si{kg/m^3}` etc.

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+T(K)    
+   400.00360000000001     
+   19.999960000020000     
+         181
+P(bar)  
+   15001.334999999999     
+   5114.9322988556905     
+         262
+           8
+T(K)           P(bar)         rho,kg/m3      alpha,1/K      cp,J/K/kg      vp,km/s        vs,km/s        h,J/kg
```
bangerth, line 69: Oh, I see -- is this really how Perple_X does it? How inconsistent :-(

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+         181
+P(bar)  
+   15001.334999999999     
+   5114.9322988556905     
+         262
+           8
+T(K)           P(bar)         rho,kg/m3      alpha,1/K      cp,J/K/kg      vp,km/s        vs,km/s        h,J/kg 
+\end{lstlisting}
+
+It is also useful to know that \aspect{} does not actually read in the values of the pressures and temperatures in the first two columns, but instead uses the minimum, increment, and number of values parameters given in the header, assuming a uniform step size. The first column is always assumed to be the inner loop (i.e., it needs to increase first while the second column stays constant).
```
bangerth, line 72: I'm confused by the word "column" since the file above doesn't actually have rows/columns. Can you explain what "column" refers to here?

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+
+
+
+
+\paragraph{The rheology.}
+The rheology of this model consists of two parts: The viscosity profile, and the lateral variations due to temperature. For each of these, we need to read in a data file. In this example, we use files that are based on \cite{stca06} for both. The viscosity profile is based on mineral physics and surface constrains, and the lateral viscosity variations use an Arrhenius law with a depth-dependent activation enthalpy. For more details and a derivation, see \cite{stca06}. 
+
+Other rheology models can be used by reading in different files. The formatting of these files is the following: The radial viscosity file contains two columns, where the first is the viscosity in Pa~s, and the second is the depth in km (note that this is an exception to the usual \aspect{} convention of using SI units). The lateral viscosity file also contains two columns, the first being the activation enthaly divided by the gas constant and the nondimensional stress exponent (which is 1 for diffusion creep/in the lower mantle, and 3.5 for dislocation creep/in the upper mantle and transition zone in the model of \cite{stca06}). The second column is depth, again in km.
```
bangerth, line 80:
```suggestion
Other rheology models can be used by reading in different files. The formatting of these files is the following: The radial viscosity file contains two columns, where the first is the viscosity in \si{\pascal\second}, and the second is the depth in \si{km} (note that this is an exception to the usual \aspect{} convention of using SI units). The lateral viscosity file also contains two columns, the first being the activation enthalpy divided by the gas constant and the nondimensional stress exponent (which is 1 for diffusion creep/in the lower mantle, and 3.5 for dislocation creep/in the upper mantle and transition zone in the model of \cite{stca06}). The second column is depth, again in \si{km}. 
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+  \eta = \eta_\text{rad} \exp{ \left( -\frac{V_\text{lat} \Delta T}{T T_\text{ref}} \right)} , 
+\end{equation}
+where $\eta_\text{rad}$ is the value from the radial viscosity file, 
+$V_\text{lat}$ is the value from the lateral viscosity file, 
+$T$ is temperature, $T_\text{ref}$ is the reference temperature profile, 
+and $\Delta T$ is the deviation from the reference temperature profile. 
+
+This reference profile can be chosen in two different ways: On the one hand, it can be chosen as the laterally averaged temperature (and in this case, a number of depth slices for this lateral averaging can be specified as  well). This is the original formulation of \cite{stca06}, and the default of the material model. 
+On the other hand, the adiabatic temperature profile can be chosen as the reference. 
+However, the radial profile needs to be adapted based on how this referene temperature is chosen.
```
bangerth, line 92:
```suggestion
However, the radial profile needs to be adapted based on how this reference temperature is chosen. 
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+\begin{figure}
+  \includegraphics[width=0.48\textwidth]{cookbooks/steinberger/doc/radial-visc.pdf}
+  \includegraphics[width=0.48\textwidth]{cookbooks/steinberger/doc/radial-visc-simple.pdf}
+  \caption{\it Left: Viscosity profile based on the original \cite{stca06} formulation, intended for use with a temperature-dependence of viscosity based on the laterally averaged temperature. Right: Modified viscosity profile without boundary layers, intended for use with a temperature-dependence of viscosity based on an adiabatic temperature profile.}
+  \label{fig:steinberger-viscosity}
+\end{figure}
+
+In order to improve solver convergence, the material model has additional parameters that allow it to limit the viscosity variations. 
+Because of the resolution in this cookbook we limit the overall viscosity
+between $10^{20}$~Pa~s and $5 \times 10^{23}$~Pa~s and the lateral viscosity variations to
```
bangerth, line 107: Use SI markup

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+  \includegraphics[width=0.48\textwidth]{cookbooks/steinberger/doc/radial-visc-simple.pdf}
+  \caption{\it Left: Viscosity profile based on the original \cite{stca06} formulation, intended for use with a temperature-dependence of viscosity based on the laterally averaged temperature. Right: Modified viscosity profile without boundary layers, intended for use with a temperature-dependence of viscosity based on an adiabatic temperature profile.}
+  \label{fig:steinberger-viscosity}
+\end{figure}
+
+In order to improve solver convergence, the material model has additional parameters that allow it to limit the viscosity variations. 
+Because of the resolution in this cookbook we limit the overall viscosity
+between $10^{20}$~Pa~s and $5 \times 10^{23}$~Pa~s and the lateral viscosity variations to 
+three orders of magnitude in both directions (for a total of six orders of
+magnitude. This allows the features of the flow field to be resolved.
```
bangerth, line 109: I think I don't understand this. If you limit the overall viscosity to 1e20...1e23, how can the lateral variation be in a larger range? Or maybe I don't understand the word "overall" :-)

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+\lstinputlisting[language=prmfile]{cookbooks/steinberger/doc/rheology.part.prm.out}
+In the Earth, we would expect higher viscosities in the lithosphere and 
+lower viscosities in plumes and near the core-mantle boundaries. 
+This type of viscosity formulation is appropriate for global convection models. 
+However, it does not approximate lithospheric deformation well. 
+The model only accounts for diffusion creep, so the 
+lithosphere has a high viscosity and forms a stagnant lid on top of the 
+sublithospheric mantle. In order to achieve more realistic subduction in a model like this, 
+one would have to either prescribe plate velocities at the surface 
+(forcing plate to subduct) or take into account plastic yielding (so that
```
bangerth, line 119:
```suggestion
(forcing plates to subduct) or take into account plastic yielding (so that
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+finite element grid) rather than approximating it with a reference profile 
+or temperature/pressure derivatives of the density. 
+
+To use the projected density approximation, we need to specify it as the form of the equations we want to use, and we need to provide a field that the density values can be interpolated on. 
+The first part is handled in the `Formulation' section of the input file. This is where we can select 
+the projected density approximation as the formulation we want to use for the mass conservation equation. 
+The temperature equation uses the real density (rather than a reference profile) as well. 
+
+To allow for the interpolation, we create a compositional field that we call `density\_field'. 
+We assign the field the type `density', so that ASPECT knows that this is the field it should use to compute the density gradient required to solve the equations. ASPECT does not need to solve an
```
bangerth, line 138:
```suggestion
We assign the field the type `density', so that \aspect{} knows that this is the field it should use to compute the density gradient required to solve the equations. \aspect{} does not need to solve an 
```

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
...
+The complete input file can be found in \url{cookbooks/steinberger/doc/steinberger.prm}. 
+
+\paragraph{Results.} We run the model for 300 million years. Over the time of the model evolution, some plumes rise and spread beneath the base of the lithosphere, and some cold downwellings detach from the base of the lithosphere. The temperature at the end of the model run and some of the material properties are shown in Figure~\ref{fig:steinberger-end-state}. 
+
+\begin{figure}
+  \includegraphics[width=0.48\textwidth]{cookbooks/steinberger/doc/temperature.png}
+  \includegraphics[width=0.48\textwidth]{cookbooks/steinberger/doc/viscosity.png}
+  \includegraphics[width=0.48\textwidth]{cookbooks/steinberger/doc/density.png}
+  \includegraphics[width=0.48\textwidth]{cookbooks/steinberger/doc/specific_heat.png}
+  \caption{\it End state of the model. From left to right and top to bottom: Temperature, viscosity, density and specific heat capacity.}
```
bangerth, line 154:
```suggestion
  \caption{\it End state of the model. From left to right and top to bottom: Temperature, viscosity, density, and specific heat capacity.}
```

cookbooks/steinberger/steinberger.prm
```diff
@@ -0,0 +1,219 @@
...
+# gradient required to solve the equations. ASPECT does not need to solve an
+# equation for this field, it only needs to interpolate the density values onto
+# it. This is covered by the compositional field method 'prescribed field'.
+# For fields of this type, the material model provides the values that should be
+# interpolated onto the field.
+subsection Compositional fields
+  set Number of fields = 1
+  set Names of fields  = density_field
+  set Types of fields = density
+  set Compositional field methods = prescribed field
```
bangerth, line 167: Align the equals signs here as you do elsewhere.

data/material-model/steinberger/description.txt
```diff
@@ -17,4 +17,6 @@ thermal conductivity: constant
 
 The provided lookup table was constructed with the software PERPLEX (Connolly, 2005) using the thermodynamic database by Stixrude (2011) and assuming a pyrolitic composition (Ringwood 1988).
 
+For more explanation, see the cookbook on "Convection using a pressure--temperature look-up table and the rheology of Steinberger and Calderwood (2006)".
```
bangerth, line 20:
```suggestion
For more explanation, see the cookbook on "Convection using a pressure-temperature look-up table and the rheology of Steinberger and Calderwood (2006)". 
```

bangerth (commented):
Nice cookbook! I left a bunch of comments, but in hindsight not all of the SI markup of physical units I suggested is correct. You might want to do `grep -r '\\si' source/` to see our usual style.

cookbooks/steinberger/doc/steinberger.tex
```diff
@@ -0,0 +1,157 @@
+\subsubsection{Convection using a pressure--temperature look-up table and the rheology of Steinberger and Calderwood (2006)}
```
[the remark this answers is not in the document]
jdannberg: This is correct as it is: https://www.thepunctuationguide.com/en-dash.html (and I actually learned about when to use an en dash rather than a hyphen while visiting you and reading "How to write a lot" with your working group).

cookbooks/steinberger/doc/steinberger.tex
jdannberg: See above.

cookbooks/steinberger/doc/steinberger.tex
jdannberg: Unfortunately, there is no other cookbook that uses look-up tables (this was one of the reasons I added this one).

cookbooks/steinberger/doc/steinberger.tex
jdannberg: I can not change this just in the cookbook. This is the implementation in the MaterialModel::MaterialUtilities::Lookup::MaterialLookup class. It looks like this:
```
              if (column_name == "rho,kg/m3")
                prp_indices[0] = n;
              else if (column_name == "alpha,1/K")
                prp_indices[1] = n;
```
So if the table uses different column names, it will not be read in correctly. That is also why I am not using marking up the units. I am just documenting this here, and I now changed it to use \textttt to make this clear. 

Are you suggesting to change that in the code? That would mean that all existing tables would not work anymore.

jdannberg:
@bangerth Thank you for your review! I addressed all of your comments, and I also rebased the branch. Could you have another look?

gassmoeller (approved):
This looks good to me. All comments seem to be adressed. Should be ready to merge when the testers are done.

cookbooks/steinberger/steinberger.prm
```diff
@@ -0,0 +1,219 @@
+# This is a cookbook for a convection model with realistic material properties.
+# Specifically, it uses material properties read in from a look-up table
+# computed with mineral physics software and the projected density approximation.
+# Since it is a quarter of a apherical shell, it also uses periodic boundary
```
tjhei, line 4:
```suggestion
# Since it is a quarter of a spherical shell, it also uses periodic boundary
```

tjhei: I also don't like the reasoning, one can do many different other boundary conditions (free slip, no slip, etc.)

cookbooks/steinberger/steinberger.prm
```diff
@@ -0,0 +1,219 @@
...
+
+set Dimension                              = 2
+set End time                               = 3e8
+set Adiabatic surface temperature          = 1613.0
+set Output directory                       = output-steinberger
+
+# Because the model is compressible, we use a nonlinear solver.
+# We here choose a tolerance of 1e-3 to limit model runtime, but in a
+# research application one would test what tolerance would be required
+# to obtain a sufficiently accurate solution, and 1e-3 is quite a high
```
tjhei, line 15:
```suggestion
# to obtain a sufficiently accurate solution, and 1e-3 is a quite large
```

tjhei (approved):

gassmoeller (approved):
Looks ready, Timo's comments seem to be addressed.

cookbooks/steinberger/doc/steinberger.tex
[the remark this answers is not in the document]
bangerth: :-) Out of curiosity, on that page, you consider "pressure-temperature" as a "connection or conflict"?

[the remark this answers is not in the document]
jdannberg: Yes, it's a connection of two different things (maybe even in the sense of different directions) as opposed to a compound. It's not one thing (a pressure-temperature in the sense that it would be a temperature where the word pressure defines more specifically what kind of temperature it is), but instead a connection of two separate things (pressure and temperature are two different axes in the coordinate system of the table).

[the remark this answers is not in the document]
bangerth: I learn something new every day :-)

PR #4469, opened 2022-01-28
use the hydrostatic pressure for thermodynamic lookup

jdannberg:
This PR changes the steinberger material model to use the hydrostatic pressure rather than the full pressure to compute the material properties from the thermodynamic lookup equation of state (density, etc.). This is consistent with the projected density approximation and avoids convergence issues of the nonlinear solver related to jumps of material properties at phase transitions, which are common in thermodynamic lookup tables. At the same time, using the full pressure does not really make the results more accurate since the dynamic pressure is so small that it changes material properties very little otherwise (not more than 0.1% compared to the other variables even in the most extreme cases). 

This is the change that is needed for the steinberger cookbook in #4468 to work properly (otherwise the nonlinear solver stagnates and does not converge). 

### For all pull requests:


jdannberg:
This obviously changes some tests that use the steinberger model. I've looked through the changes when I updated the test results, and they are small, and in addition, the overall iteration count decreases.

gassmoeller (approved):
Yes in retrospective this would have been the reasonable thing to do from the start. In my defense, I think the adiabatic conditions plugin did not even exist when I wrote this material model :-).

PR #4470, opened 2022-01-28
Add geometric average only viscosity operation

jdannberg:
We have a harmonic average only viscosity operation to use with the GMG, but I found that this causes problems in some models: Harmonic averaging favors low viscosity values, and so it can happen that if high-viscosity features are not well resolved, they are deformed/destroyed by the surrounding material. Geometric averaging provides the option to average to more intermediate viscosity values. 

My specific problem is a global-scale model that can't afford a high resolution (because it's 3d and time-dependent). Because of that, the lithosphere is not well resolved, and whenever a plume arrives at the base of the lithosphere it can very easily decrease the viscosity of the cells around it, and pushes the lithosphere to both sides (rather than spreading below the lithosphere). This then causes extreme temperature over- and undershoots when the hot plume material reaches the surface. Using geometric averaging improves the issue. 


tjhei (approved):
Great, thank you! Looks good to me.

gassmoeller (approved):

PR #4471, opened 2022-02-01
Use stress averaging in tests where required

anne-glerum:
As I noted in #4355, there are three tests where the combination of fixed elastic timestep, stress averaging, and elastic timestep stabilization is not correct. Here I apply stress averaging in all three cases, as the elastic timestep and numerical timestep are not equal. I'll have to update the tests with the results from the tester. 

### For all pull requests:


gassmoeller (approved):
Good to go if you update the tests. Thanks for the fixes!

Issue #4466, opened 2022-01-27
GMG solver crashes in models with Material averaging = project to Q1 only viscosity and strong viscosity contrasts


jdannberg:
I have been trying to set up models with the GMG and the option `set Material averaging = project to Q1 only viscosity`. In many of my Earth-like application cases (with moderately strong viscosity contrasts), I get the following error message:

```
--------------------------------------------------------
An error occurred in line <1207> of file </home/juliane/software/aspect-6/source/simulator/stokes_matrix_free.cc> in function
    void aspect::MatrixFreeStokesOperators::ABlockOperator<dim, degree_v, number>::set_diagonal(const dealii::LinearAlgebra::distributed::Vector<number>&) [with int dim = 2; int degree_v = 2; number = double]
The violated condition was: 
    local_element > 0.
Additional information: 
    No diagonal entry in a positive definite operator should be zero or
    negative.

Stacktrace:
-----------
#0  /home/juliane/software/aspect-6/build/aspect: aspect::MatrixFreeStokesOperators::ABlockOperator<2, 2, double>::set_diagonal(dealii::LinearAlgebra::distributed::Vector<double, dealii::MemorySpace::Host> const&)
#1  /home/juliane/software/aspect-6/build/aspect: aspect::StokesMatrixFreeHandlerImplementation<2, 2>::build_preconditioner()
#2  /home/juliane/software/aspect-6/build/aspect: aspect::Simulator<2>::assemble_and_solve_stokes(bool, double*)
#3  /home/juliane/software/aspect-6/build/aspect: aspect::Simulator<2>::solve_single_advection_single_stokes()
#4  /home/juliane/software/aspect-6/build/aspect: aspect::Simulator<2>::solve_timestep()
#5  /home/juliane/software/aspect-6/build/aspect: aspect::Simulator<2>::run()
#6  /home/juliane/software/aspect-6/build/aspect: void run_simulator<2>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool, bool, bool)
#7  /home/juliane/software/aspect-6/build/aspect: main
--------------------------------------------------------

```

This is on the current main branch, and with deal.II 9.3.0. I’ve checked in the debugger, and the `local_element` value that triggers the error is around -1e21 (so it’s on the order of magnitude of a viscosity, just negative). 

[...]

PR #4472, opened 2022-02-02
GMG: fix max/min of the evaluated viscosities

zjiaqi2018:
Fix max/min of the evaluated viscosities in `evaluate_material_model()`. @tjhei

### DIFF
--- a/include/aspect/stokes_matrix_free.h
+++ b/include/aspect/stokes_matrix_free.h
@@ -576,7 +576,8 @@ namespace aspect
       /**
        * The max/min of the evaluated viscosities.
        */
-      double viscosity_max, viscosity_min;
+      double minimum_viscosity;
+      double maximum_viscosity;
 
       DoFHandler<dim> dof_handler_v;
       DoFHandler<dim> dof_handler_p;
--- a/include/aspect/stokes_matrix_free.h
+++ b/include/aspect/stokes_matrix_free.h
@@ -573,6 +573,11 @@ namespace aspect
        */
       bool do_timings;
 
+      /**
+       * The max/min of the evaluated viscosities.
+       */
+      double viscosity_max, viscosity_min;
+
       DoFHandler<dim> dof_handler_v;
       DoFHandler<dim> dof_handler_p;
       DoFHandler<dim> dof_handler_projection;
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1370,8 +1370,8 @@ namespace aspect
 
     const QGauss<dim> quadrature_formula (sim.parameters.stokes_velocity_degree+1);
 
-    double min_el = std::numeric_limits<double>::max();
-    double max_el = std::numeric_limits<double>::lowest();
+    double min_el_local = std::numeric_limits<double>::max();
+    double max_el_local = std::numeric_limits<double>::lowest();
 
     // Fill the DGQ0 or DGQ1 vector of viscosity values on the active mesh
     {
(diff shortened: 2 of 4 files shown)


tjhei (approved):
Perfect, thank you.

tjhei:
Please apply the test changes...

source/simulator/stokes_matrix_free.cc
```diff
@@ -1908,6 +1911,8 @@ namespace aspect
         const double imbalance = MGTools::workload_imbalance(sim.triangulation);
         sim.pcout << "    GMG workload imbalance: " << imbalance << std::endl
                   << "    Stokes solver: " << std::flush;
+
+        sim.pcout << "Viscosity range: " << viscosity_min << " - " << viscosity_max << std::endl;
```
tjhei, line 1915: Move this new code above the line 1913. Right now you are printing in the middle of the line of the Stokes iterations.

tjhei (approved):

tjhei (approved):
awesome, thank you.

include/aspect/stokes_matrix_free.h
```diff
@@ -573,6 +573,11 @@ namespace aspect
        */
       bool do_timings;
 
+      /**
+       * The max/min of the evaluated viscosities.
+       */
+      double viscosity_max, viscosity_min;
```
gassmoeller, line 579:
```suggestion
      double minimum_viscosity;
      double maximum_viscosity;
```

source/simulator/stokes_matrix_free.cc
```diff
@@ -1370,8 +1370,8 @@ namespace aspect
 
     const QGauss<dim> quadrature_formula (sim.parameters.stokes_velocity_degree+1);
 
-    double min_el = std::numeric_limits<double>::max();
-    double max_el = std::numeric_limits<double>::lowest();
+    double min_el_local = std::numeric_limits<double>::max();
+    double max_el_local = std::numeric_limits<double>::lowest();
```
gassmoeller, line 1374:
```suggestion
    double minimum_viscosity_local = std::numeric_limits<double>::max();
    double maximum_viscosity_local = std::numeric_limits<double>::lowest();
```

source/simulator/stokes_matrix_free.cc
```diff
@@ -1492,7 +1495,7 @@ namespace aspect
                   // of the evaluated viscosity on the active level.
                   for (unsigned int q=0; q<n_q_points; ++q)
                     active_cell_data.viscosity(cell, q)[i]
-                      = std::min(std::max(values_on_quad[q], min_el), max_el);
+                      = std::min(std::max(values_on_quad[q], viscosity_min), viscosity_max);
```
gassmoeller, line 1498: Was this the bug before? Did we limit the viscosity only to the local minimum and maximum? But I do not understand why this had caused the issue, limiting to the local minimum and maximum should only be stricter than limiting to the global minimum/maximum. Can you explain what the problem was?

gassmoeller (commented):
I have a suggestion for renaming variables to better conform to other places in ASPECT (not your fault, the naming was a bit weird before already). Also can you explain what you fixed? Just so that if someone (e.g. I) read over this PR again in the future know what happened. Also please add a changelog entry if it was a fixed bug.

tjhei:
> Was this the bug before? Did we limit the viscosity only to the local minimum and maximum? But I do not understand why this had caused the issue, limiting to the local minimum and maximum should only be stricter than limiting to the global minimum/maximum. Can you explain what the problem was?

Without the MPI communication we could have slightly wrong viscosities on coarser GMG levels maybe resulting in higher iteration numbers (if at all). This is not a fix for #4466 but one step along the way. That will come as a followup.

tjhei:
Not sure if we need a changelog entry at this point.

gassmoeller:
Thanks for the explanation. No if this is not a fix for https://github.com/geodynamics/aspect/issues/4466 we dont need a changelog yet. Do you want to do the renaming as part of this PR or the next? I am also fine pushing that to the next Pr and just merge this one.

tjhei:
I would prefer if Jiaqi would do the rename here. @zjiaqi2018 ?

PR #4476, opened 2022-02-03
add some additional error checks and information

jdannberg:
This pull request adds some additional information to the error messages when reading in ascii data files. 
Specifically, it checks that the data file has the correct number of columns before reading in the whole file (which generally would trigger another error message that would tell the user that their coordinates are not correct, while the actual problem is the wrong number of data columns). 

There is one line of code that is an actual change here: The change in line 330, changing `components` to `components+dim`. This assert is only executed in cases where the header of the file is read in (so if it's not commented out with a #). The original version of the file checked that the header had as many entries as data columns, but excluding the coordinates. This means that in a file where my columns are
` r phi theta composition_1 composition_2`
the valid header would be
`composition_1 composition_2`
This didn't make much sense to me. If this was actually intended by someone, let me know. If this does not trigger any changing tests, I will also add a new test (or modify an existing one) that actually reads in the header. 


jdannberg:
Since it looks like none of the tests was actually reading in a header with coordinates, I modified one of the test data files (which is used in the `ascii_data_initial_composition_3d_shell` test). This new version of the header would have triggered the Assert that I talk about above.

source/structured_data.cc
```diff
@@ -380,6 +381,29 @@ namespace aspect
             column_names.push_back("column " + Utilities::int_to_string(c,2));
         }
 
+      // Make sure the data file actually has as many columns as we think it has
+      // (either based on the header, or based on what was passed to the reinit function).
+      const unsigned int position = in.tellg();
```
gassmoeller, line 386: Your current code works because streampos is implicitly convertable to integers, but the correct type should be this:
```suggestion
      // (either based on the header, or based on what was passed to the constructor).
      const std::streampos position = in.tellg();
```

gassmoeller (commented):
Looks good, except for my one comment.

jdannberg:
Thank you for your review! I simply committed your suggestion.

gassmoeller (approved):

PR #4479, opened 2022-02-05
GMG: fix negative viscosity

zjiaqi2018:
Fix issue https://github.com/geodynamics/aspect/issues/4466


tjhei:
Great, this makes sense. Do both examples in the issue work now?
Can you try to make a test out of the simpler one? Please confirm that the test fails without this patch.

zjiaqi2018:
Yes, they both work. I will add a test later.

zjiaqi2018:
it doesn't seem right, it takes much more iterations than using harmonic average only viscosity :
image

and the following is for harmonic average only viscosity 
image

tjhei:
Can you test the examples where you use zero instead of tangential velocity at the "bottom" boundary? This might be a case where we need the nullspace PR that is not merged yet.

tjhei:
I think this PR is correct regardless. Please update test results and add a changeling entry. Something like:

Fixed: The GMG solver could fail in problems with large viscosity variations and Q1 viscosity averaging, because viscosities would become negative on coarser GMG levels. This is now fixed.

source/simulator/stokes_matrix_free.cc
```diff
@@ -2799,7 +2799,8 @@ namespace aspect
                                                 ?
                                                 level_viscosity_vector[level](dg_dof_indices[0])
                                                 :
-                                                visc_on_quad[q]);
+                                                std::min(std::max(visc_on_quad[q], static_cast<GMGNumberType>(minimum_viscosity)),
+                                                         static_cast<GMGNumberType>(minimum_viscosity)));
```
tjhei, line 2803: Shouldn't this be the maximum viscosity?

zjiaqi2018:
after fixing the bug, now it makes sense:
image

tjhei (approved):
Looks good to me! Thank you.

jdannberg:
Great to see this is working now! Thank you for the fix!

gassmoeller (approved):

PR #4481, opened 2022-02-08
Ascii data slice initial composition

jdannberg:
This pull request adds a new initial composition plugin called ascii data slice. It can read in 3d ascii data sets to use them in a 2d model. This has the advantage that one can run high-resolution 2d models of problems with actual data that are usually run in 3d (such as models using seismic tomography as an input). Doing this is particularly useful for testing features of complicated 3d models in the same resolution one would use for the final 3d model, but without having to use up so much computational time. 

You should probably ignore the second commit (this was my first try for the rotation and didn't work very well), but I still left the commits separate because I moved some functions from the GPlates plugin into utilities. So my third commit is just moving these functions, and the 4th commit shows how I updated these functions (the general functionality should still be the same, but I added some checks that caused models to fail for certain values of the input parameters). So I thought it would make reviewing easier to keep those commits separate for now. 

I've also kept it simple for now and only allow Cartesian coordinates as input, but I think we do have other functions that would allow it to use other coordinate systems/use the natural coordinate system of a geometry model. This is something that could also be added. 

I am also attaching the graphical output of the test: Left is the output of the ascii data slice model, right is the same model in a 3d geometry, reading in the same data set, with the "Slice" filter applied in Paraview (using the normal that is defined by the cross product of the two points in the input file). 

 slice_comparison

### For all pull requests:


jdannberg:
There are some failing tests, but I believe these results were actually wrong in the original version of the implementation. All the failing tests have coordinates of theta = pi/2 (and two different values for phi). So they already are in the x-y-plane. The old algorithm only checked if the norm of the original normal vector of the plane (here, in z direction) is different from the new normal vector (which is also is in z direction, since we're projecting onto the x-y-plane). If they're the same, the whole rotation is skipped. But this means that Point 1 is not actually rotated onto the (0,1,0) direction, which is what the function does for all other points (this is the second rotation). For other coordinate directions, the gplates plugin still works as before, for example, the gplates_rotation test, which does not use a rotation plane that is normal to any of the axes of the coordinate system, still works as before. @gassmoeller can you check if that make sense? 

In the meanwhile, I will update the test results.

include/aspect/utilities.h
```diff
@@ -327,6 +327,24 @@ namespace aspect
     std::array<Tensor<1,dim>,dim-1>
     orthogonal_vectors (const Tensor<1,dim> &v);
 
+    /**
+      * A function that returns the corresponding euler angles for a
```
gassmoeller, line 331:
```suggestion
      * A function that returns the corresponding rotation matrix for a
```

gassmoeller (approved):
This looks good, thanks for the work! I think I have an idea for how to integrate this functionality into the main AsciiDataInitial class so that it is is available for all ascii data plugins (hopefully backwards compatible), but let's merge this first and try to simplify afterwards since I am not sure it will work.

PR #4484, opened 2022-02-10
Cleanup and fix large file issues

bangerth:
The first commit is just a collection of small cleanups. The second one fixes the issue that we broadcast file sizes as `unsigned int`, but that may be too small and is in fact sometimes as evidenced on the mailing list.

/rebuild

### DIFF
--- a/source/utilities.cc
+++ b/source/utilities.cc
@@ -993,15 +993,15 @@ namespace aspect
 
       if (Utilities::MPI::this_mpi_process(comm) == 0)
         {
-          // set file size to an invalid size (signaling an error if we can not read it)
-          unsigned int filesize = numbers::invalid_unsigned_int;
+          unsigned int filesize;
 
           // Check to see if the prm file will be reading data from disk or
           // from a provided URL
           if (filename_is_url(filename))
             {
 #ifdef ASPECT_WITH_LIBDAP
-              libdap::Connect *url = new libdap::Connect(filename);
+              std::unique_ptr<libdap::Connect> url
+                = std::make_unique<libdap::Connect>(filename);
               libdap::BaseTypeFactory factory;
               libdap::DataDDS dds(&factory);
               libdap::DAS das;
@@ -1057,9 +1057,7 @@ namespace aspect
               std::vector<std::string> points;
               for (libdap::AttrTable::Attr_iter i = das.var_begin(); i != das.var_end(); i++)
                 {
-                  libdap::AttrTable *table;
-
-                  table = das.get_table(i);
+                  libdap::AttrTable *table = das.get_table(i);
                   if (table->get_attr("POINTS") != "")
                     points.push_back(table->get_attr("POINTS"));
                   if (table->get_attr("points") != "")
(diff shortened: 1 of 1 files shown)


source/utilities.cc
```diff
@@ -60,6 +60,148 @@ namespace aspect
    */
   namespace Utilities
   {
+    namespace internal
+    {
+      namespace MPI
+      {
+#ifdef DEAL_II_WITH_MPI
```
tjhei, line 67: isn't this already available in deal.II in mpi.templates.h?

source/utilities.cc
```diff
@@ -1093,12 +1234,16 @@ namespace aspect
...
+
 #else // ASPECT_WITH_LIBDAP
 
-              // broadcast failure state, then throw
-              const int ierr = MPI_Bcast(&filesize, 1, MPI_UNSIGNED, 0, comm);
-              AssertThrowMPI(ierr);
+              // Broadcast failure state, then throw. We signal the failure by
+              // setting the file size to an invalid size, then trigger an assert.
+              {
+                std::size_t invalid_filesize = numbers::invalid_unsigned_int;
```
tjhei, line 1243: shouldn't this be an invalid size_t? You can use -1 directly.

source/utilities.cc
```diff
@@ -1113,11 +1258,11 @@ namespace aspect
               if (!filestream)
                 {
                   // broadcast failure state, then throw
-                  const int ierr = MPI_Bcast(&filesize,1,MPI_UNSIGNED,0,comm);
+                  std::size_t invalid_filesize = numbers::invalid_unsigned_int;
```
tjhei, line 1261: same here

tjhei (commented):
This makes sense, but you will still fail with the Bcast of the data if it is more than 2^31 bytes.

source/utilities.cc
bangerth: Ah, I had not realized that we actually have `invalid_size_type`. So fixed.

source/utilities.cc
tjhei: Not sure as it is an internal detail right now. Maybe we should promote this from internal to external in deal.II. Maybe leave a comment about this fact here?

source/utilities.cc
```diff
@@ -1128,44 +1273,46 @@ namespace aspect
...
+          int ierr = MPI_Bcast(&filesize, 1, Utilities::internal::MPI::mpi_type_id(&filesize), 0, comm);
           AssertThrowMPI(ierr);
           if (filesize == numbers::invalid_unsigned_int)
             throw QuietException();
 
           data_string.resize(filesize);
 
           // Receive and store data
-          ierr = MPI_Bcast(&data_string[0],filesize,MPI_CHAR,0,comm);
+          ierr = MPI_Bcast(&data_string[0], filesize, MPI_CHAR, 0, comm);
```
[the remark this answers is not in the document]
bangerth: Are you saying that `MPI_Bcast` can't deal with larger message sizes? What is the maximum size it can do?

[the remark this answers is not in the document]
bangerth: (I'm asking because deal.II's `MPI::broadcast` has the same problem, and I'd like to fix that as well.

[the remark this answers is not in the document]
bangerth: See https://github.com/dealii/dealii/issues/13357

tjhei (approved):

tjhei:
```
/home/runner/work/aspect/aspect/source/utilities.cc:1299:33: error: comparison of integer expressions of different signedness: ‘std::size_t’ {aka ‘long unsigned int’} and ‘int’ [-Werror=sign-compare]
 1299 |           AssertThrow (filesize < std::numeric_limits<signed int>::max(),
```

tjhei (approved):

bangerth:
I wished I could have just used the deal.II version of `broadcast()`. It's a shame that that is broken in exactly the same way :-(

bangerth:
OK, seems to work at least as far as producing an error for cases that currently just generate hard-to-read MPI error messages.

PR #4483, opened 2022-02-10
add sheared torsion BM after Farrington et al. 2014

EstherHeck:
This adds the input file for a 3D torsion plus shearing benchmark after Farrington et al. 2014.  

*Describe what you did in this PR and why you did it.*

### Before your first pull request:

* [X ] I have read the guidelines in our [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) document.

### For all pull requests:

* [X ] I have followed the [instructions for indenting my code](../blob/main/CONTRIBUTING.md#making-aspect-better).

### For new features/models or changes of existing features:

* [X ] I have tested my new feature locally to ensure it is correct.
* [ ] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/main/tests/) directory.
* [ ] I have added a changelog entry in the [doc/modules/changes](../blob/main/doc/modules/changes) directory that will inform other users of my change.

PR #4487, opened 2022-02-11
add the comments recommended for #4483 also to the simple shear BM

EstherHeck:
This is kind of a follow up PR for #4483 where I used the prm file for the simple shear BM as the underlying model to create the new torsion BM. As some typos and many comments that were requested during reviewing of #4483 actually also concern this underlying prm file, I thought it would be useful to change them in this file too. 

* [X] I have read the guidelines in our [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) document.

### For all pull requests:


anne-glerum (commented):
Hi @EstherHeck Thank you for making these updates. The removal of the unused compositional field will probably lead to failing tests. Can you check that once the tester is done? The tester should be triggered to run again (it failed initially while checking permissions).

bangerth:
The `visco_plastic_vep_simple_shear` test currently fails.

bangerth:
You'll either have to also update the output file, or only make those changes that leave the behavior of the test unchanged. The latter for sure includes all of the commentary changes.

EstherHeck:
Thanks for reminding me! I updated the test results now.
The failing of the test was indeed due to one composition less. I checked that the rest (like the nonlinear residuals) is still the same. There are few minor variations, e.g. nonlinear residual of the last dt is not exactly the same, but could be seen as rounded to the third digit after the comma. Some variations are also in the statistics file for the global mass of stress_yy. I guess these variations are acceptable?

bangerth (approved):
Excellent, thank you!

gassmoeller (commented):
edit: sorry my previous comment was meant for #4483

PR #4490, opened 2022-02-14
Integrate dataset slicing into normal ascii data plugin

gassmoeller:
@jdannberg this is my suggestion for improving #4481. I like it more than duplicating plugins, because it is far less code (even though we still have to duplicate a bit inside the plugin). It should be reasonably simple to move this code up in the inheritance order (e.g. into AsciiDataInitial or even AsciiDataBase) if we want to make it available to other plugins. What do you think?


include/aspect/initial_composition/ascii_data.h
```diff
@@ -79,6 +76,23 @@ namespace aspect
...
+        bool slice_data;
+
+        // The two points that define the location of the slice.
+        Tensor<1,3> first_point_on_slice;
+        Tensor<1,3> second_point_on_slice;
+
+        /**
+          * The matrix that describes the rotation by which a 2D model
+          * needs to be transformed to a plane that contains the origin and
+          * the prescribed point given in the input.
```
jdannberg, line 93: This needs to be updated. 
```suggestion
          * the two prescribed points given in the input.
```

source/initial_composition/ascii_data.cc
```diff
@@ -59,6 +76,31 @@ namespace aspect
...
                                                           "box_2d.txt");
+
+        prm.enter_subsection("Ascii data model");
+        {
+          prm.declare_entry("Slice dataset in 2D plane", "false",
+                            Patterns::Bool (),
+                            "Whether to use a 2D data slice from a 3D data file "
+                            "or the entire data file. Slicing a 3D dataset is "
+                            "only supported for 2D models.");
+          prm.declare_entry ("First point on slice", "0.0,1.0,0.0",
```
jdannberg, line 87: Can you add a comment that these parameters will be ignored if "Slice dataset in 2D plane" is false.

tests/ascii_data_slice.prm
```diff
@@ -50,14 +50,12 @@ end
...
-  set List of model names = ascii data slice
+  set List of model names = ascii data
 
   subsection Ascii data model
     set Data directory = $ASPECT_SOURCE_DIR/data/initial-composition/ascii-data/test/
     set Data file name = shell_3d_2.txt
-  end
-
-  subsection Ascii data slice
+    set Slice data in 2D plane = true
```
jdannberg, line 58:
```suggestion
    set Slice dataset in 2D plane = true
```
This is why the test is failing.

jdannberg (approved):
I like this much better than my version! Thank you!

What do you think about the coordinate system? Should we add that as an option in a follow-up PR (especially if we want to make versions of this for other ascii data plugins)?

gassmoeller:
Thanks for the comments, I think I have addressed them in the last commit. 

> This is why the test is failing.

Yes, I was experimenting with different parameter names. Do you think the current one is ok?

jdannberg:
Looks good, thank you!

For the name: I was confused at first, since 'slice' can also be used as a noun, so I was wondering what's a 'slice dataset'? But I don't have a better idea, so I think this is fine.

PR #4494, opened 2022-02-16
Do not rely on uninitialized state in AsciiDataBoundary

gassmoeller:
This fixes a problem reported by @jdannberg. The AsciiDataBoundary class relied on the current model time to determine if it should apply its boundary condition, but this variable is NaN during initialization. As a consequence anything that could call the boundary condition early (like the material model, adiabatic conditions etc) had to have checks against initialization time to prevent floating point exceptions.

Thinking about this some more, it seems weird that the AsciiDataBoundary class arbitrarily returns 0 before the time of the first data file (this may or may not make sense for velocity boundary condition, but for temperature and composition it seems wrong). Should I change the behavior to be symmetric to times past the last data file, i.e. at times before the first data file the class will return the values of the first data file? I think this would make the whole if condition here redundant. This would be a change in behavior though.


tjhei:
> , it seems weird that the AsciiDataBoundary class arbitrarily returns 0 before the time of the first data file

Yes, this is weird. I would find it more sensible if we fail if simulation time < first_file_time instead of returning 0 or using the first data file.
Regarding NaN at initialization time: Maybe we should return data at time 0 instead. Or, we can look into making a decision to set the current time earlier in the setup.

jdannberg:
I think what @gassmoeller suggests is consistent with how all of the other data file processing happens. When we use gplates files as velocity boundary conditions, the last file keeps being used when we're beyond the time we have data files for. If we read in any material data with the ascii data/lookup plugins, material properties are extrapolated using the last available value if we're beyond the depth/pressure/temperature whatever conditions we have data for. 

That being said, this kind of behavior causes a lot of bugs that are hard to find, and in most cases is likely not intended, so I think what @tjhei suggests makes sense. We can do both: we can return data at time 0 before initialization, and also not allow simulation time < first_file_time once we have the time initialized (we have some asserts that say something along the lines of "If you REALLY want to do this, then remove this assert").

bangerth:
My suggestion would be to throw an error if the current model time is still NaN. It just means that we can't know what to return (as the discussion above shows), and if we don't know what to return, we shouldn't return anything.

Where is this called when the time is still NaN? I had looked into `core.cc` a while ago and realized that we call the `initialize()` functions of the various plugin systems pretty much right after `parse_parameters()`. I think that's wrong: We should first set up all of the plugins, and then call `initialize()` on all of them, but that was a bigger change than I was willing to make at the time. I don't know, of course, if that has anything to do with what you are saying here.

tjhei:
> suggests is consistent with how all of the other data file processing happens.

I disagree: extrapolation into the future makes sense, but going backwards in time does not make sense to me. I think this is being proposed here, or not?

gassmoeller:
I had to think about this for a while, but I am pretty sure I found a solution. Here are my thoughts: AsciiDataBoundary can be asked for the boundary values in 3 different scenarios:

1. During the setup phase of the simulator. Here we do not know the time (it is NaN), in particular because we do not know if we are restarting a model or not. This option is currently not allowed, but is required for some applications (like if the BC is needed for the initial condition or reference profile). These models currently require awkward workarounds (two difference code paths for setup and actual model). However, thinking about this further, we do not need to know the actual time for this phase. We want the setup phase to always behave the same way as if we are just starting a model, so we should pretend the time during this phase is `parameters.start_time`. If we would do anything different, e.g. the reference profile of a restarted computation may be different from one that is not restarted, something we would want to avoid. Since `parameters.start_time` is available right from the start we can avoid calling `this->get_time()`.

2. At a time before `first_data_file_model_time`. Currently we return 0 in this case, which had made sense for the one specific application I used this for at the time (velocity boundary conditions at the surface that should only kick in later than the model start). However as I pointed out in my first post this makes absolutely no sense for most other scenarios (temperature/composition/stress/pressure). Crashing also does not make sense in this case, because then we may as well remove the option to read in data files that only start later than the model start. I agree with @jdannberg that the most reasonable behavior in this case is to be symmetric to the times after the model files and we should just return the values of the first data files. To restore our previous behavior users could always create an additional data file that only contains zeroes and put that in as the first data file.

3. After `first_data_file_model_time`. Everything is fine here.

I am happy to modify the behavior of AsciiDataBoundary this way, but would want to know first if you agree with this approach.

tjhei:
> 1. During the setup phase of the simulator.

Agreed. One could think about populating ``this->get_time()`` earlier than we do right now. Is there a good reason not to set it earlier?


> A compromise might be to output a warning (the same way we do if we're extrapolating velocities forward in time after the last file is read).

I would be happy with that solution.

gassmoeller:
> Agreed. One could think about populating this->get_time() earlier than we do right now. Is there a good reason not to set it earlier?

There is a reason, we cannot really tell which time we are at until we have read the checkpoint file (in case we are restarting). We could of course initialize the variable with the start time in the constructor, but then it may be wrong during initialization until we overwrite it from the restart file.

> A compromise might be to output a warning (the same way we do if we're extrapolating velocities forward in time after the last file is read).

That is a bit awkward because we cannot do it in the `initialize` function when we do not know the current time yet. If we do not want to output this warning for every model we would need to check for it in the `update` function and store the state if we have already written the warning. I would suggest we just remove the whole parameter instead, then we avoid a lot of the complications. It is a very specific parameter that is rarely used. Are you ok with change in behavior?

source/structured_data.cc
```diff
@@ -783,13 +782,20 @@ namespace aspect
...
-      if (time_dependent && (this->get_time() - first_data_file_model_time >= 0.0))
+      // always initialize with the start time during model setup, even
+      // when restarting (to have identical setup in both cases)
+      double model_time = this->get_parameters().start_time;
+
+      // if we are past initialization use the current time instead
+      if (this->simulator_is_past_initialization())
+        model_time = this->get_time();
+
+      if (time_dependent == true)
```
jdannberg, line 793: I don't understand why you removed this. Wouldn't this now try to load files even after the last available file is loaded and crash?

jdannberg (approved):
Looks good to me, other than the one small comment.

source/structured_data.cc
gassmoeller: I had another look at this PR, but @jdannberg I dont understand your comment (maybe it was placed on a different line by github in the meantime). Specifically, I am not sure what you mean by `why you removed this`, I dont see anything I removed in this code that would cause it to load another file after the last available file. After the last available file is loaded and the code can not find the next file it will set `time_dependent` to `false`, and that should prevent it from entering the if-condition again in the future.

Let me know if I misunderstood you, otherwise I think this is ready.

PR #4495, opened 2022-02-17
GMG: compute mesh displacement

zjiaqi2018:
Use GMG to compute mesh displacements. I am keeping both AMG and GMG right now, because I want to 
compare the two, and I am not sure if we want to always use GMG. 

The setup of GMG follows step-37. I am not sure whether the GMG preconditioner is working correctly, I tested it using
free_surface_viscous_cylinder_3D_loading_gmg.prm, and it takes more than 60 iterations,
see the following:

```
free_surface_viscous_cylinder_3D_loading_gmg.prm
running with 168 MPI processes

mesh displacements dofs: 1,594,323

 Timestep 150:  t=1500 years, dt=10 years
   Solving mesh displacement system... 10 iterations. 
 initial residual: 1.6558e-05 final residual: 1.14323e-12
   Relative error in l2 norm :  3.21233e-08
   Relative error in inf norm : 8.54385e-08
   Solving mesh displacement system using gmg... 67 iterations.
 inital residual: 1.6558e-05 final residual: 1.21086e-12 


| amg displacement                 |       152 |       273s |       4.2% |
| gmg displacement                 |       152 |       169s |       2.6% |

```


source/mesh_deformation/interface.cc
```diff
@@ -603,100 +603,295 @@ namespace aspect
...
-      Amg_data.aggregation_threshold = 0.02;
-      preconditioner_stiffness.initialize(mesh_matrix);
-
-      // we solve with higher accuracy in the initial timestep:
-      const double tolerance
-        = sim.parameters.linear_stokes_solver_tolerance
-          * ((this->simulator_is_past_initialization()) ? 1.0 : 1e-5);
+        // setup GMG, follow step-37:
+        MGLevelObject<SystemMatrixType> mg_matrices;
+        mesh_deformation_dof_handler.distribute_mg_dofs();
```
tjhei, line 773: add a TODO here: this only needs to happen once and not every time step.

source/mesh_deformation/interface.cc
```diff
@@ -603,100 +603,295 @@ namespace aspect
...
+        MGConstrainedDoFs mg_constrained_dofs;
+        mg_constrained_dofs.initialize(mesh_deformation_dof_handler);
+
+        // Not sure if we need to consider other boundary ids here.
+        mg_constrained_dofs.make_zero_boundary_constraints(mesh_deformation_dof_handler,
+                                                           zero_mesh_deformation_boundary_indicators);
+        mg_constrained_dofs.make_zero_boundary_constraints(mesh_deformation_dof_handler,
```
tjhei, line 780: I don't think you are handling the constraints correctly.  See make_constraints():
- periodic constraints (ignore for now, but add an ``Assert(false, ExcNotImeplemented);``
- 0 for zero_mesh_deformation_boundary_indicators
- no normal flux for tangential_mesh_deformation_boundary_indicators
- all other ids get their values from the plugins using compute_velocity_constraints_on_boundary

Note that the last kind has inhomogenous constraints.

source/mesh_deformation/interface.cc
```diff
@@ -603,100 +603,295 @@ namespace aspect
...
+                                                           prescribed_mesh_deformation_boundary_indicators);
```
tjhei, line 781: I think this set is not to be used here.

tjhei (commented):
I think you are not handling constraints correctly, especially no normal flux (see the GMG stokes solver for how we implement these). We might need to move some of the implementation into deal.II. Also see https://github.com/geodynamics/aspect/issues/3894

Can you report what tests you ran and if they have adaptive refinement or not? (try both) Finally, can you run GMRES instead of CG (this is to test if preconditioner happens to be not SPD).

zjiaqi2018:
Tests :
1. free_surface_blob_gmg.prm with 
  set Initial adaptive refinement        = 2
  set Initial global refinement          = 4
CG number of iterations: 4-5
GMRES: same

2. free_surface_viscous_cylinder_3D_loading_gmg.prm with 
  set Initial adaptive refinement        = 0
  set Initial global refinement          = 3
CG number of iterations: 4
GMRES number of iterations: 2

I need tests with curved boundaries, @tjhei any suggestions?

tjhei:
You can try this one. It is a relatively simple 2d spherical shell:

[shell_2d_free_surface.prm.txt](https://github.com/geodynamics/aspect/files/8142864/shell_2d_free_surface.prm.txt)

tjhei:
> From the three cases tested, GMRES converges faster than CG.

GMRES might have smaller number of iterations, but GMRES is more expensive. I asked you to try GMRES just in case the number is drastically lower (5 vs 60). If that is the case, the system might not be symmetric positive definite. Anyways, it makes sense to go back to CG.

zjiaqi2018:
> > From the three cases tested, GMRES converges faster than CG.
> 
> GMRES might have smaller number of iterations, but GMRES is more expensive. I asked you to try GMRES just in case the number is drastically lower (5 vs 60). If that is the case, the system might not be symmetric positive definite. Anyways, it makes sense to go back to CG.

zjiaqi2018:
closed accidentally!

zjiaqi2018:
Since GMG doesn't support periodic boundary conditions, do you think it's better to keep both AMG and GMG? @tjhei

tjhei:
> Since GMG doesn't support periodic boundary conditions, do you think it's better to keep both AMG and GMG?

Yes, that is true. We will need to keep them for now (but we should be able to support periodic GMG soonish). Can you put matrix-based and matrix-free into separate functions?

zjiaqi2018:
>Can you put matrix-based and matrix-free into separate functions?

Done.

include/aspect/mesh_deformation/interface.h
```diff
@@ -24,7 +24,7 @@
 
 #include <aspect/plugins.h>
 #include <aspect/simulator_access.h>
-
+#include <aspect/stokes_matrix_free.h>
```
tjhei, line 27: Do you need this include? Can you try adding it only to the .cc file?

include/aspect/mesh_deformation/interface.h
```diff
@@ -34,6 +34,11 @@
 #include <deal.II/base/mg_level_object.h>
 #include <deal.II/lac/la_parallel_vector.h>
 #include <deal.II/multigrid/mg_transfer_matrix_free.h>
+#include <deal.II/matrix_free/operators.h>
```
tjhei, line 37: Same with all of these: move to the .cc. You should only need MGConstrainedDofs here.

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +739,253 @@ namespace aspect
...
+        const periodic_boundary_pairs pbp = this->get_geometry_model().get_periodic_boundary_pairs();
+        AssertThrow(pbp.size() == 0,
+                    ExcMessage("Periodic boundary constraints are not supported in computing mesh displacements using GMG."));
+      }
+
+      mg_constrained_dofs.make_zero_boundary_constraints(mesh_deformation_dof_handler,
+                                                         zero_mesh_deformation_boundary_indicators);
+
+      {
+        // Handle no normal flux, copyied from GMG Stokes solver.
```
tjhei, line 825: copied.

tjhei (commented):
This looks quite good now. I have a few minor comments.

tjhei:
Can you add a changelog please?

Also, you mentioned that it is much faster: can you post some numbers here?

zjiaqi2018:
> Also, you mentioned that it is much faster: can you post some numbers here?
I only tested free_surface_viscous_cylinder_3D_loading_gmg.prm on Frontera with 3*56 mpi ranks,
and it's around 4 times faster, would you call it "much" faster?
image
image

doc/modules/changes/20220311_jiaqi
```diff
@@ -0,0 +1,4 @@
+New: The mesh displacements are computed by the 
+GMG solver when the GMG Stokes solver is used.
+<br>
+(Timo Heister, Jiaqi Zhang, 2022/03/01)
```
tjhei, line 4: please add a newline at the end

tjhei (approved):
Very nice, nearly good to go from my point of view.

tjhei:
can you update the test output?

tjhei (approved):
Thank you.

tjhei:
This is ready for review, @geodynamics/aspect-developers

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+      // We use this gmg solver only when the gmg stokes solver is used
+      // for the following reasons (TODO):
+      // 1. this gmg solver does not support periodic boundary conditions
+      // 2. To use this solver even when gmg stokes solver is not used, we need to
+      //    initialize the triangulation with Triangulation<dim>::limit_level_difference_at_vertices
+      //    and parallel::distributed::Triangulation<dim>::construct_multigrid_hierarchy
+      // 3. Although this gmg solver is much faster than the amg solver, it's only tested for
+      //    limited free surface cases.
+
+      TimerOutput::Scope timer (this->sim.computing_timer, "gmg displacement");
```
gassmoeller, line 757: The other `compute_mesh_displacements` function does not have such a timer section, and both function are embedded in another timer section (`Mesh deformation initialize`), which makes the output confusing (because the total does add up to more than 100% because of the duplication). I understand this was useful for benchmarking, but I would prefer you remove this timer section for the merge.

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+
+      Assert(mesh_deformation_fe.degree == 1, ExcNotImplemented());
+      const unsigned int mesh_deformation_fe_degree = 1;
```
gassmoeller, line 760: Wouldnt this be more consistent?
```suggestion
      const unsigned int mesh_deformation_fe_degree = mesh_deformation_fe.degree;
```

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+                               LaplaceOperator<dim, mesh_deformation_fe_degree, mesh_deformation_fe_degree + 1, dim>;
+
+      SystemMatrixType mf_mesh_matrix;
+
+      MGLevelObject<SystemMatrixType> mg_matrices;
+
+      typename MatrixFree<dim, double>::AdditionalData additional_data;
+      additional_data.tasks_parallel_scheme =
+        MatrixFree<dim, double>::AdditionalData::none;
+      UpdateFlags update_flags = UpdateFlags(update_values | update_JxW_values | update_gradients);
```
gassmoeller, line 772:
```suggestion
      const UpdateFlags update_flags (update_values | update_JxW_values | update_gradients);
```

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+            }
+          phi.integrate(EvaluationFlags::gradients);
+          phi.distribute_local_to_global(corrected_rhs);
+        }
+      corrected_rhs.compress(VectorOperation::add);
+
+      // setup GMG, follow step-37:
+      const unsigned int n_levels = sim.triangulation.n_global_levels();
+
+      // Currently not support periodic boundary constraints
```
gassmoeller, line 814:
```suggestion
      // Currently does not support periodic boundary constraints
```

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
 
 
 
+    template <int dim>
+    void MeshDeformationHandler<dim>::compute_mesh_displacements_gmg()
+    {
+      // Same as compute_mesh_displacements, but using GMG instead of AMG.
```
gassmoeller, line 746:
```suggestion
      // Same as compute_mesh_displacements, but using matrix-free GMG instead of matrix-based AMG.
```

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+      //    and parallel::distributed::Triangulation<dim>::construct_multigrid_hierarchy
+      // 3. Although this gmg solver is much faster than the amg solver, it's only tested for
+      //    limited free surface cases.
+
+      TimerOutput::Scope timer (this->sim.computing_timer, "gmg displacement");
+
+      Assert(mesh_deformation_fe.degree == 1, ExcNotImplemented());
+      const unsigned int mesh_deformation_fe_degree = 1;
+
+      using SystemMatrixType = dealii::MatrixFreeOperators::
```
gassmoeller, line 762: Maybe like this? In general I find references to matrices when no matrices are stored confusing.
```suggestion
      using SystemOperatorType = dealii::MatrixFreeOperators::
```

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+      mf_mesh_matrix.initialize_dof_vector(u0);
+      mf_mesh_matrix.initialize_dof_vector(corrected_rhs);
+      mf_mesh_matrix.initialize_dof_vector(mf_solution);
+      u0 = 0.;
+      mesh_velocity_constraints.distribute(u0);
+      u0.update_ghost_values();
+
+      corrected_rhs = 0.;
+
+      FEEvaluation<dim, mesh_deformation_fe_degree, mesh_deformation_fe_degree + 1, dim, double> phi(*mf_mesh_matrix.get_matrix_free());
```
gassmoeller, line 794: I know that this is syntax in line with examples in deal.II, but in the ASPECT matrix free code the FEEvaluation objects are named after their solution variable (compare stokes_matrix_free.cc:733). I would suggest to follow this here too, i.e. `phi` -> `mesh_deformation`.

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+          for (unsigned int q = 0; q < phi.n_q_points; ++q)
+            {
+              phi.submit_gradient(-1.0 * phi.get_gradient(q), q);
+            }
+          phi.integrate(EvaluationFlags::gradients);
+          phi.distribute_local_to_global(corrected_rhs);
+        }
+      corrected_rhs.compress(VectorOperation::add);
+
+      // setup GMG, follow step-37:
```
gassmoeller, line 811:
```suggestion
      // setup GMG, following deal.II step-37:
```

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+
+      mg::SmootherRelaxation<SmootherType, dealii::LinearAlgebra::distributed::Vector<double>> mg_smoother;
+
+      MGLevelObject<typename SmootherType::AdditionalData> smoother_data;
+      smoother_data.resize(0, n_levels - 1);
+
+      for (unsigned int level = 0; level < n_levels;
+           ++level)
+        {
+          if (level > 0)
```
gassmoeller, line 909: Please add some comments here on the choice of smoother_data. Why is it different depending on level (I guess coarse grid solver vs smoother on other levels, but we need this documented).

I guess you copied this from stokes_matrix_free.cc. Copy the comment describing the parameters as well.

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+      const double tolerance
+        = sim.parameters.linear_stokes_solver_tolerance
+          * ((this->simulator_is_past_initialization()) ? 1.0 : 1e-5);
+
+      SolverControl solver_control_mf(5 * corrected_rhs.size(),
+                                      tolerance * corrected_rhs.l2_norm());
+      SolverCG<dealii::LinearAlgebra::distributed::Vector<double>> cg(solver_control_mf);
+
+      mesh_velocity_constraints.set_zero(mf_solution);
+      cg.solve(mf_mesh_matrix, mf_solution, corrected_rhs, preconditioner);
```
gassmoeller, line 954: As stated above, I think this would be easier to read the following way:
```suggestion
      cg.solve(laplace_operator/mesh_deformation_operator, solution, rhs, preconditioner);
```

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+      this->get_pcout() << "   Solving mesh displacement system using GMG... " << solver_control_mf.last_step() <<" iterations."<< std::endl;
```
gassmoeller, line 955: Do you think we need the distinction in the output between the two functions (GMG vs non GMG)? I expect we can eventually replace the matrix-based function with this new one, and then it would be easiest if they already produce the same output.

source/mesh_deformation/interface.cc
```diff
@@ -722,6 +740,253 @@ namespace aspect
...
+
+      mesh_velocity_constraints.distribute(mf_solution);
+      mf_solution.update_ghost_values();
+
+      // copy solution:
+      LinearAlgebra::Vector solution_tmp;
+      solution_tmp.reinit(mesh_locally_owned, sim.mpi_communicator);
+      dealii::LinearAlgebra::ReadWriteVector<double> rwv(solution_tmp.locally_owned_elements());
+      rwv.import(mf_solution, VectorOperation::insert);
+      solution_tmp.import(rwv, VectorOperation::insert);
```
gassmoeller, line 965: Can you extend the comment above to why this is so complicated? I know it is because of the conversion ghosted/non-ghosted vectors, but it seems surprising we need to make that many copies just make the assignment `mf_solution` -> `fs_mesh_velocity`.

source/simulator/stokes_matrix_free.cc
```diff
@@ -61,7 +61,7 @@ namespace aspect
       add_constraint(const std::array<types::global_dof_index,dim> &dof_indices,
                      const Tensor<1, dim> &constraining_vector,
                      AffineConstraints<double> &constraints,
-                     const double inhomogeneity = 0)
+                     const double inhomogeneity)
```
gassmoeller, line 64: Why did you remove the default value here, but kept it above for mesh_deformation? There may be a reason, I just want to make sure it was not an accident.

gassmoeller (commented):
Very nice, I think the algorithm is ready, but I have a number of question/comments on naming things. Please take a look and change what you think is reasonable. Let me know when I can take another look.

source/mesh_deformation/interface.cc
tjhei: Yes, it needs to be known at compile time. I don't think it is a big issue because we don't support higher order displacements anyways at this point.

tjhei:
Let us know once things are ready, @zjiaqi2018

source/mesh_deformation/interface.cc
zjiaqi2018: Is there an easier way to do this? @tjhei

gassmoeller:
@zjiaqi2018 I think you made the changes, but your last force push overwrote them again with an earlier version. Not sure what happened, but the current version on github does not contain any of the changes I requested. The last force push (from commit  5ead3df to 56e0fd2) seems to revert all of the changes. Take a look at what happened, and let me or @tjhei know in case you need help sorting things out.

zjiaqi2018:
> @zjiaqi2018 I think you made the changes, but your last force push overwrote them again with an earlier version. Not sure what happened, but the current version on github does not contain any of the changes I requested. The last force push (from commit [5ead3df](https://github.com/geodynamics/aspect/commit/5ead3df9e5fa744ac3ac5849c0231a0e414bf4ee) to [56e0fd2](https://github.com/geodynamics/aspect/commit/56e0fd2579c1c263523066b3c85143765744e0aa)) seems to revert all of the changes. Take a look at what happened, and let me or @tjhei know in case you need help sorting things out.

Forgot to commit those changes...

PR #4503, opened 2022-02-22
fix gplates scaling factor

jdannberg:
This PR fixes the issue described in this forum post: https://community.geodynamics.org/t/gplates-scale-factor-parameter/2270

The velocities in the gplates model should now be scales correctly. I added a new test case, which is the same as the existing gplates_1_4 test, but with a scaling factor of 2. Comparing the screen output of both tests shows that the maximum velocity is now twice as high as in the original, and I am attaching a comparison of the graphical output below (Left is with scaling factor 1, right is with scaling factor 2, and then the velocities in ParaView are divided by 2, so in the end the values shown in the two panels are the same). 
gplates


source/boundary_velocity/gplates.cc
```diff
@@ -727,12 +727,12 @@ namespace aspect
 
       if ((this->get_time() - first_data_file_model_time >= 0.0) && (this->get_geometry_model().depth(position) <= lithosphere_thickness + magic_number))
         {
-          const Tensor<1,dim> data = lookup->surface_velocity(position);
+          const Tensor<1,dim> data = scale_factor * lookup->surface_velocity(position);
```
bangerth, line 730: Could we rename this `velocity_scaling_factor`? I think that would make it clearer what this factor should be applied to.

bangerth (commented):
OK other than the one issue (which could be done as a follow-up, separate from this PR).

But there are some failing tests.

source/boundary_velocity/gplates.cc
jdannberg: I changed it but unfortunately it's still called "Scale factor" in the input file. Given that the parameter did not do anything in the past, do you want me to change that too?

jdannberg:
The test failing was just the one I added (so it was the difference between my local version and the tester). That should be fixed now.

bangerth (approved):
This is ok. The documentation does say that it's scaling the velocity, so I'm ok with leaving it as is.

PR #4511, opened 2022-03-02
Add a necessary header file.

bangerth:
Probably slipped through under the radar because everyone but me uses unity builds :-)

/rebuild


tjhei:
> Probably slipped through under the radar because everyone but me uses unity builds :-)

That is surprising because we have a "no-unity-build" tester. Is this maybe a change on deal.II master?

tjhei (approved):

bangerth:
Possible. I have not investigated why I needed the include file and others did not.

Issue #4507, opened 2022-02-28
Why is the mesh refined for tagging strategies even if the initial adaptive refinement is set to 0?


anne-glerum:
When ` Initial adaptive refinement = 0` and `Time steps between mesh refinement = 0`, the mesh will still be adaptively refined at t0 when the `Strategy` is one that tags cells, like `minimum refinement function`, given that the `Minimum refinement level` allows it. See examples below. This is not what I would expect, and I wonder whether it is intentional. 

Found when looking into #4088.

For these settings:
```
subsection Mesh refinement
  set Initial global refinement = 2
  set Initial adaptive refinement = 2 
  set Time steps between mesh refinement = 0 
  set Strategy = minimum refinement function, maximum refinement function
  set Minimum refinement level = 0
end
```
the mesh looks like this at t0:
![IGR2_IAR2](https://user-images.githubusercontent.com/7631624/155988988-5ce729b9-fac0-4afb-964b-e4b5773afdeb.png)

And this is what the log file says about the DOFs:
```
Number of degrees of freedom: 663,706 (93,206+46,128+93,206+11,739+46,603+46,603+46,603+46,603+46,603+46,603+46,603+46,603+46,603)
Number of mesh deformation degrees of freedom: 23478
Number of degrees of freedom: 1,721,180 (241,692+119,866+241,692+30,316+120,846+120,846+120,846+120,846+120,846+120,846+120,846+120,846+120,846)
Number of mesh deformation degrees of freedom: 60632
Number of degrees of freedom: 4,563,217 (640,636+318,859+640,636+80,224+320,318+320,318+320,318+320,318+320,318+320,318+320,318+320,318+320,318)
Number of mesh deformation degrees of freedom: 160448
```

For these settings:
```
subsection Mesh refinement
  set Initial global refinement = 4
  set Initial adaptive refinement  = 0 
  set Time steps between mesh refinement = 0 
  set Strategy = minimum refinement function, maximum refinement function
  set Minimum refinement level = 0
end
```
the t0 mesh looks like:
![IGR4_IAR0_MRL0_Stratminmax](https://user-images.githubusercontent.com/7631624/155989063-f776224e-8599-49a6-9aed-e3df889c1d3f.png)

and the log file states:
```
Number of degrees of freedom: 4,765,459 (669,078+332,660+669,078+83,792+334,539+334,539+334,539+334,539+334,539+334,539+334,539+334,539+334,539)
Number of mesh deformation degrees of freedom: 167584
[...]

PR #4513, opened 2022-03-03
Do initial global refinement correctly in parallel

jdannberg:
### For all pull requests:

### DIFF
--- a/source/simulator/core.cc
+++ b/source/simulator/core.cc
@@ -1894,7 +1894,8 @@ namespace aspect
         for (unsigned int n=0; n<parameters.initial_global_refinement; ++n)
           {
             for (const auto &cell : triangulation.active_cell_iterators())
-              cell->set_refine_flag ();
+              if (cell->is_locally_owned())
+                cell->set_refine_flag ();
 
             mesh_refinement_manager.tag_additional_cells ();
             triangulation.execute_coarsening_and_refinement();
--- a/tests/initial_global_refinement.prm
+++ b/tests/initial_global_refinement.prm
@@ -1,0 +1,101 @@
+# This model tests if the maximum refinement function is
+# applied correctly in time step zero during global refinement.
+# The mesh should have the highest refinement level of 4 in the
+# upper quarter of the domain, the lowest refinement level of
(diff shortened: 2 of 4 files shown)


jdannberg:
See #4507.

bangerth (commented):
This is probably not wrong, but it shouldn't make a difference. Or does it for you?

At least in principle, deal.II shouldn't care about what flags are or aren't set on anything other than locally owned cells. The only place I can think of is the mesh smoothing algorithm, but I don't recall what it does in parallel computations -- I'm pretty sure it doesn't import neighboring cells's information, so it's possible that whether or not you set something on a ghost cell makes a difference there.

tjhei:
It does make a difference because of mesh smoothing.
I think we need a different solution, though. We should do the same communication step we do in the normal refinement.

bangerth:
I agree that we should probably do that, but it doesn't change the fact that mesh refinement might depend on the number of processes because the `prepare_coarsening_and_refinement()` function doesn't do this every time it does a sweep over all cells to do the smoothing. We should do it there as well.

But I would accept a patch along the lines you mention. I would change the return type of the `pack()` function to `unsigned char`, though. (Or `std::uint8_t` if you like that better.)

bangerth:
What do we do with the current patch? I think it's useful if only because it reduces the amount of work by a marginal amount. I'm tempted to merge, but it might be incompatible (despite the tests succeeding) so I thought I'd give others the chance to object.

bangerth (approved):

tjhei:
This patch might change the refinement in some situations, but I would argue that the old behavior is broken. I agree with merging and i will follow up with the other suggestion.

tjhei:
Actually, we should add a test as the current change did not change any of the existing ones. @jdannberg do you have something that you could use?

jdannberg:
I have added a slightly modified version of the test case that Anne sent. The image below shows the difference between the new version (left) and the current master (right) run with 8 processors on my laptop. The minimum/maximum refinement function only depend on the vertical coordinate direction, so the image on the right shows the problem that Anne mentioned. 

(I was on vacation last week and just came back to this today.)

mesh

tjhei (approved):
awesome, thank you.

PR #4516, opened 2022-03-06
Get rid of a bunch of GeometryInfo related things.

bangerth:
This continues my quest to get rid of places that use `GeometryInfo`. `GeometryInfo` is used widely in all places where we do assembly for DG elements because we need to allocate and then access vectors and matrices for all faces and subfaces. Counting how many of these there are, and then what index to use was previously done via `GeometryInfo::faces_per_cell` and `GeometryInfo::subfaces_per_face`. 

This patch introduces two new functions, `n_interface_matrices(...)` and `nth_interface_matrix(...)` that tell us the number of matrices we need to provision, and the index within such arrays that we want to access for a given face (and subface, if so relevant). The places where this is relevant are in the advection assemblers (where we sometimes use DG elements) and in the volume-of-fluid assemblers.

I'd be surprised if this patch doesn't break something, so let's see what happens.

/rebuild

### DIFF
--- a/include/aspect/simulator/assemblers/interface.h
+++ b/include/aspect/simulator/assemblers/interface.h
@@ -419,15 +419,16 @@ namespace aspect
            */
           FullMatrix<double>          local_matrix;
 
-          /** Local contributions to the global matrix from the face terms in the
-           * discontinuous Galerkin method. The vectors are of length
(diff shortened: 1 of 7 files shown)


bangerth:
Ha, whaddaya know -- it actually works!

bangerth:
Ping?

gassmoeller (approved):
Looks all good. Thanks for the update :+1:.

PR #4520, opened 2022-03-07
Test for deal.II >= 9.4 instead of 10.0.

bangerth:
People keep mentioning that the next deal.II version won't be 10.0, but 9.4. We should conservatively check for 9.4 then, which is the correct check even if the release ultimately ends up being 10.0.

/rebuild


tjhei (approved):
Yes, that makes sense.

tjhei:
Note that we already have 4 occurrences of >=10.0 in the ASPECT 2.3 release done in July 2021.

Issue #4521, opened 2022-03-08
Evaluate min/max refinement level function at the center of a cell.


bangerth:
In #4507, we figured out that one awkward spot in the min/max refinement level scheme is that we look at the min/max function at the *vertices* of cells, but that that is of course not unique across a cell: the function might have different values for different vertices, in particular in view of round-off deciding whether a particular vertex is on one side of a line or another.

We should instead be looking at the center of the cell.

anne-glerum:
Fixed by #4523.

PR #4523, opened 2022-03-16
Evaluate cell center instead of vertices for min/max mesh refinement function

anne-glerum:
This addresses #4521: instead of evaluating all vertices of a cell to see whether they should refine/coarsen according to the min/max refinement functions, only the cell center is evaluated. This should avoid differences in refinement due to round-off e.g. along a specified depth level that coincides with cell boundaries.

### For all pull requests:


anne-glerum:
PS I didn't add a test, because there are several tests for the min/max refinement functions already, but I think a change log might be useful?

tjhei:
Yes, please add a changelog as it is a change in behavior.

tjhei (approved):

anne-glerum:
Alright, I've added a change log, and I looked at the failing tests. For three tests I modified the refinement settings so that they show the same levels of refinement as before. For these tests, the center of the coarse elements did not meet the minimum refinement criterion, so they weren't refined any more. I'll update the test results once the tester is done.

tjhei (approved):
Looks great!👍

anne-glerum:
Okay thanks, test results are updated.

tjhei (approved):

bangerth:
Thanks!

PR #4525, opened 2022-03-18
Move dynamic friction into a friction rheology add function for friction

EstherHeck:
This is a follow-up on PR #4357. It adds the option to specify friction angles as functions of space and time per composition in the input file. 
I put a little test case into the tests folder to show @elodie-kendall how to use this functionality. If this could be useful as an aspect test, let me know, otherwise I will remove it. 

### For all pull requests:

### DIFF
--- a/include/aspect/material_model/rheology/friction_models.h
+++ b/include/aspect/material_model/rheology/friction_models.h
@@ -48,7 +48,8 @@ namespace aspect
       enum FrictionMechanism
       {
         static_friction,
-        dynamic_friction
+        dynamic_friction,
+        function
       };
 
       template <int dim>
@@ -76,7 +77,8 @@ namespace aspect
           double
           compute_friction_angle(const double current_edot_ii,
                                  const unsigned int volume_fraction_index,
-                                 const double static_friction_angle) const;
+                                 const double static_friction_angle,
+                                 const Point<dim> &position) const;
 
           /**
            * A function that returns the selected type of friction dependence.
@@ -87,7 +89,7 @@ namespace aspect
         private:
           /**
            * Select the mechanism to be used for the friction dependence.
-           * Possible options: static friction | dynamic friction
+           * Possible options: static friction | dynamic friction | function
            */
           FrictionMechanism friction_mechanism;
 
@@ -114,6 +116,18 @@ namespace aspect
            * more step-like.
            */
           double dynamic_friction_smoothness_exponent;
+
+          /**
+           * Parsed functions that specify the friction angle which must be
+           * given in the input file using the function method.
+           */
+          std::unique_ptr<Functions::ParsedFunction<dim> > friction_function;
+
+          /**
+           * The coordinate representation to evaluate the function for the friction angle.
+           * Possible choices are depth, cartesian and spherical.
+           */
+          Utilities::Coordinates::CoordinateSystem coordinate_system_friction_function;
       };
     }
   }
--- a/source/material_model/rheology/visco_plastic.cc
+++ b/source/material_model/rheology/visco_plastic.cc
@@ -269,9 +269,10 @@ namespace aspect
             // Note: Maybe this should also be turned around to first apply strain rate dependence and then
             // the strain weakening to the dynamic friction angle. Didn't come up with a clear argument for
             // one order or the other.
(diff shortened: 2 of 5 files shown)


EstherHeck:
Hi, as  #4357 is merged now, I rebase this PR on current main.

include/aspect/material_model/rheology/friction_models.h
```diff
@@ -76,7 +77,8 @@ namespace aspect
           double
           compute_friction_angle(const double current_edot_ii,
                                  const unsigned int volume_fraction_index,
-                                 const double static_friction_angle) const;
+                                 const double static_friction_angle,
+                                 const Point<dim> &position) const;
```
gassmoeller, line 81: If we want to add more options in the future that need other input parameters we need to find a different solution than to keep adding parameters here, but for now it's fine. Do you already have plans to add other options?

source/material_model/rheology/visco_plastic.cc
```diff
@@ -271,7 +271,7 @@ namespace aspect
             // one order or the other.
             if (friction_models.get_friction_mechanism() != static_friction)
               current_friction = friction_models.compute_friction_angle(current_edot_ii,
-                                                                        j, current_friction);
+                                                                        j, current_friction, in.position[i]);
```
gassmoeller, line 274: Please wrap this line to put each parameter on its own line. Also do we need the if-condition in line 272? The `compute_friction_angle` function correctly handles the case of static friction as far as I can tell.

tests/visco_plastic_friction_function.prm
```diff
@@ -0,0 +1,111 @@
+# Global parameters
```
gassmoeller, line 1: Add a line that described what the test is for.

tests/visco_plastic_friction_function.prm
```diff
@@ -0,0 +1,111 @@
...
+    set Activation energies for dislocation creep = 0.
+    set Activation volumes for dislocation creep = 0.
+    set Yield mechanism = drucker
+    set Angles of internal friction = 0.
+    set Cohesions = 1.e6
+
+    # these are the lines you need to use the friction function
+    set Friction mechanism = function
+    subsection Friction function
+      set Coordinate system   = cartesian
```
gassmoeller, line 87: Have you tested manually that the other coordinate systems work as well?

gassmoeller (commented):
Code looks good, I just have a few minor comments. Let me know when I should take another look.

include/aspect/material_model/rheology/friction_models.h
EstherHeck: Yes, I see. To include rate and state eventually, I will need the following inputs as well: 

```
                                       const std::vector<double> &composition, 
                                       typename DoFHandler<dim>::active_cell_iterator current_cell,
```
Should I then add these already?

source/material_model/rheology/visco_plastic.cc
EstherHeck: I removed the if now, as you are right, that this should not change the behaviour. I wasn't sure if it would make a difference in walltime and also which one would be the faster option.

tests/visco_plastic_friction_function.prm
EstherHeck: Yes, I tested that cartesian, depth and spherical are all working :)

EstherHeck:
Hi @gassmoeller thanks for your comments, I went to all of them and updated the PR, let me know what you think

gassmoeller:
You will need to update the test case, but otherwise this seems ready to go.

> I removed the if now, as you are right, that this should not change the behaviour. I wasn't sure if it would make a difference in walltime and also which one would be the faster option.

The difference will be very small. The new version has one virtual function call more, but that is a pretty fast operation. Guessing how fast or slow certain parts of the code are is nearly impossible without a lot of experience, or by profiling the code, so a general rule of thumb is to keep the code as simple as possible, unless you notice some part is slow. The only operation I typically try to avoid because of speed is to allocate dynamic memory (e.g by creating/destroying std::vector) or doing string comparisons inside a loop that is executed extremely often (e.g. over all quadrature points in every time step).

EstherHeck:
Hi @gassmoeller thanks for your explanations! 
I now updated the test results

PR #4527, opened 2022-03-22
use round() and convert to int

tjhei:
I don't think using rint() makes a lot of sense as its rounding depends on some floating point setting, instead use round()
followed by a cast.

### DIFF
--- a/source/mesh_refinement/maximum_refinement_function.cc
+++ b/source/mesh_refinement/maximum_refinement_function.cc
@@ -60,9 +60,10 @@ namespace aspect
 
               const double maximum_refinement_level = max_refinement_level.value(Utilities::convert_array_to_point<dim>(point.get_coordinates()));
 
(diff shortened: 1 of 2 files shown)


bangerth (commented):
I would probably have applied the rounding to where we assign to the variable in line 61 of both files and made that variable integral. But it doesn't matter.

But, separately, why do we need to round at all? Couldn't we just do the whole comparison in floating point arithmetic?

tjhei:
> But, separately, why do we need to round at all? Couldn't we just do the whole comparison in floating point arithmetic?

``level`` is an integer. You could convert it to a double but we need to differentiate between ``>`` and ``>=``, which seems a bit dangerous due to floating point arithmetic or not?

bangerth (approved):

tjhei:
I agree that it is not great, but it works, so...

PR #4530, opened 2022-03-28
Prescribed temperature field with diffusion

jdannberg:
The pull request adds a new advection method for the temperature field called 'prescribed field with diffusion', analogous to the same method that already exists for compositional fields. This is useful for models with prescribed temperature outputs (such as when solving the energy equation for entropy instead of temperature as in the entropy_adiabat benchmark) because it allows for diffusion of the temperature field without applying any of the other terms in the equation. 

I have a follow-up pull request that applies this method using the entropy_adiabat benchmark. 

### DIFF
--- a/doc/modules/changes/20220328_jdannberg
+++ b/doc/modules/changes/20220328_jdannberg
@@ -1,0 +1,6 @@
+New: There is now a new advection method for the temperature
+called 'prescribed field with diffusion', which works in the
+same way as the corresponding advection field method for
+compositional fields. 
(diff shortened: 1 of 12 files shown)


gassmoeller (approved):
Looks great, I found nothing to comment on. Someone may wonder why anyone would ever do this, but as you said that will be explained in a follow-up PR. Also it is nice for symmetry reasons to have the same methods available as for the compositional fields. Ready to merge.

PR #4531, opened 2022-03-29
Entropy with diffusion

jdannberg:
This pull request updates the entropy adiabat benchmark, an example case for solving the energy equation using entropy rather than temperature. It turned out that we forgot a term in the equation, which covered thermal conduction along the adiabat. This pull request fixes this and also adds three more example cases to show that conduction now works correctly. 

Specifically, the fix is the following: We first solve only the diffusion equation to get an update for the temperature, using the new 'prescribed field with diffusion' method for the temperature. We then use this update when we solve the entropy (composition) equation. Since we need to use a nonlinear solver anyway, this should still be accurate. 

The reason we need to decouple the equations like this is that the term we forgot, nabla (k nabla (alpha T g / c_p)), contains both the thermal expansivity and specific heat, which both jump at phase transitions. If the phase transitions are too thin for those jumps to be resolved, then the temperature jump at phase transitions associated with latent heat release/consumption will not diffuse over time. 

(This is the promised follow-up to #4530.)


benchmarks/entropy_adiabat/plugins/entropy_advection.cc
```diff
@@ -75,75 +79,94 @@ namespace aspect
...
-          // do the actual assembly. note that we only need to loop over the advection
-          // shape functions because these are the only contributions we compute here
-          for (unsigned int i=0; i<advection_dofs_per_cell; ++i)
+          // solve the diffusion equation for the temperature
+          if (advection_field.is_temperature())
+            {
+              const double density_c_P = scratch.material_model_outputs.densities[q] *
+                                         scratch.material_model_outputs.specific_heat[q];
+
+              const double field_term_for_rhs = scratch.old_field_values[q] * density_c_P;
```
gassmoeller, line 90: Could we not use the same BDF2 time-stepping scheme that we use down in line 125? That would ensure the two separate solver schemes have the same convergence rate in time. It would probably not matter for practical results, but I would feel less concerned about the accuracy of the splitting we do.

benchmarks/entropy_adiabat/plugins/entropy_advection.cc
```diff
@@ -75,75 +79,94 @@ namespace aspect
...
+                  data.local_rhs(i)
+                  += field_term_for_rhs * scratch.phi_field[i] * JxW;
+
+                  for (unsigned int j=0; j<advection_dofs_per_cell; ++j)
+                    {
+                      data.local_matrix(i,j)
+                      += (
+                           (time_step * scratch.material_model_outputs.thermal_conductivities[q]
+                            * (scratch.grad_phi_field[i] * scratch.grad_phi_field[j]))
+                           + (scratch.phi_field[i] * scratch.phi_field[j]) * density_c_P
```
gassmoeller, line 105: if BDF2 is possible, we would need to multiply this line with `bdf2_factor` like in the assembly below (line 165).

benchmarks/entropy_adiabat/plugins/entropy_advection.cc
```diff
@@ -169,6 +192,11 @@ namespace aspect
                                                  scratch.material_model_outputs,
                                                  scratch.heating_model_outputs);
 
+      // We compute the amount of diffusion based on the solution of the temperature equation.
+      std::vector<double> old_temperatures (n_q_points);
+      scratch.finite_element_values[this->introspection().extractors.temperature].get_function_values (this->get_old_solution(),
+          old_temperatures);
```
gassmoeller, line 198: here `scratch.old_temperature_values` should even be filled already with the correct values, no need to recompute (see entropy_viscosity.cc:467).

benchmarks/entropy_adiabat/plugins/entropy_advection.cc
```diff
@@ -181,17 +209,19 @@ namespace aspect
...
-          const double k_Delta_field = entropy_conductivity * (scratch.old_field_laplacians[q] +
-                                                               scratch.old_old_field_laplacians[q]) / 2;
+          const double density      = scratch.material_model_outputs.densities[q];
+          const double gamma        = scratch.heating_model_outputs.heating_source_terms[q];
+
+          // Because we solve the diffusion equation for the temperature before the advection equation, we can use
+          // the current and old temperature here, together with the current time step.
+          const double diffusion_term = (this->get_timestep() == 0.0) ? 0.0
+                                        :
+                                        (scratch.material_model_inputs.temperature[q] - old_temperatures[q]) / this->get_timestep()
```
gassmoeller, line 219: I am slightly concerned about `scratch.material_model_inputs.temperature[q]` here. all other properties here are computed at time `old_... + old_old_... /2` and I think we would need the same for the diffusion term? I.e. `scratch.old_temperature_values[q] - scratch.old_old_temperature_values[q]`. That would also align it with how `dField_dt` is computed.

benchmarks/entropy_adiabat/plugins/prescribed_temperature.h
```diff
@@ -0,0 +1,55 @@
...
+    template <int dim>
+    class PrescribedTemperature : public Interface<dim>, public ::aspect::SimulatorAccess<dim>
+    {
+      public:
+        /**
+         * Return the initial temperature as a function of position.
+         */
+        double initial_temperature (const Point<dim> &position) const override;
+
+      private:
```
gassmoeller, line 49: No need to keep the `private:` if there are no private members.

gassmoeller (commented):
Looks generally good, thanks for figuring out this solution! I have some minor comments, the only big one is: Can we use BDF2 for the temperature diffusion equation to stay consistent with the rest of our time-stepping schemes? I am not quite sure what the splitting scheme in this PR does in general to our convergence rate, but using the same time-stepping scheme in both equations seems like a good idea, unless there is a reason not to.

benchmarks/entropy_adiabat/plugins/entropy_advection.cc
jdannberg: As we discussed, this does not work because we don't have anything we could use for the old_old_solution to make this second order.

benchmarks/entropy_adiabat/plugins/entropy_advection.cc
jdannberg: My reasoning for this was that the diffusion term is only first order, so the current minus the old temperature is what is actually being used in the diffusion equation. This is also exactly what is being used in the actual assembly.

One way to see it is that now this term is explicit instead of implicit, so the values we use here can be the exact same values being used in the assembly. So this now works the same way as a term like gamma.

jdannberg:
I think I addressed all your comments. Can you have another look?

gassmoeller (approved):
Looks good, thanks for making the changes. Ready to merge if the tester approves.

PR #4533, opened 2022-03-31
Make sure environment variable is respected when searching for deal.II

gassmoeller:
This PR fixes annoying behavior that happens if you have multiple deal.II versions installed and at least one of them is one or two directory levels above your build folder. In such configurations it is impossible to specify the used deal.II version via an environment variable, because the first deal.II found in the level above will be used instead. deal.II paths directly specified as cmake parameters are respected, but environment variables are not. This patch unifies the behavior for cmake variables and environment variables. deal.II versions in the levels above are still found, but only if there are no cmake or environment variables set.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ LIST(APPEND CMAKE_MODULE_PATH
   )
 
 FIND_PACKAGE(deal.II 9.3.0 QUIET
-  HINTS ${deal.II_DIR} ${DEAL_II_DIR} ../ ../../ $ENV{DEAL_II_DIR}
+  HINTS ${deal.II_DIR} ${DEAL_II_DIR} $ENV{DEAL_II_DIR} ../ ../../
   )
 IF(NOT ${deal.II_FOUND})
   MESSAGE(FATAL_ERROR "\n*** Could not find a suitably recent version of deal.II. ***\n"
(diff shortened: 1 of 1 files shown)


tjhei:
My vote would be to get rid of the .. and ../...
I think these are leftovers from the deal.ii tutorials.

gassmoeller:
I am fine with that too. We are telling everyone in the manual to use the variables anyway. Will update. Can you take a look why the tester fails for `gmg_mesh_deform_adaptive_bug`? Must be related to #4528 (maybe because of the overlap with #4495), but in that PR the tester succeeded. My PR is based on the latest master.

CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ LIST(APPEND CMAKE_MODULE_PATH
   )
 
 FIND_PACKAGE(deal.II 9.3.0 QUIET
-  HINTS ${deal.II_DIR} ${DEAL_II_DIR} ../ ../../ $ENV{DEAL_II_DIR}
+  HINTS ${deal.II_DIR} ${DEAL_II_DIR} $ENV{DEAL_II_DIR}
```
bangerth, line 80: Yes, I think this is the right choice.

tjhei (approved):

gassmoeller:
I rebased, tests should be good now.

PR #4534, opened 2022-03-31
Fix tester results

gassmoeller:
https://github.com/geodynamics/aspect/pull/4528 was merged with test results that were correct before https://github.com/geodynamics/aspect/pull/4495 was merged, but wrong afterwards. Due to the overlap the results were not correctly updated (I didnt notice that during reviewing) so the tester currently fails on `main`. This applies the fix and is identical to the changes that #4495 caused in the test `gmg_mesh_deform_adaptive`.


jdannberg (approved):

tjhei:
Thank you!

PR #4537, opened 2022-04-08
2 citation updates

ljhwang:
Note that Heyn and Conrad have yet to be assigned a volume, page or year. This will need to be updated at a later date.

Hopefully everything else is ok this time around ;).


tjhei (approved):

bangerth:
The patch had a merge conflict, which I resolved by hand to expedite the process.

PR #4541, opened 2022-04-12
Avoid compiler warnings for a test.

bangerth:
The `.cc` file that goes with this test has numerous unused arguments and variables that show up when you do `ctest -V`.

/rebuild


tjhei (approved):

tjhei:
Interesting. I never use -V but I use ``--output-on-failure`` which you can set automatically if you have ``export CTEST_OUTPUT_ON_FAILURE=1``in your environment.

bangerth:
The test doesn't fail because, I assume, we don't compile with `-Werror`.

tjhei:
That is right. There should be a ton of warnings in other tests...

bangerth:
Yes, I"m sure there are many more warnings in tests.

PR #4542, opened 2022-04-12
Simplify sheared torsion test

gassmoeller:
Follow up to #4483. This is a simpler version of the same input file that has the additional advantage that any future change to the benchmark will also be tested and we avoid duplicated information. @EstherHeck FYI.


anne-glerum:
Thank you for the clean up! Can you fix the test prm by changing `input` to `include`?
```
This is the error message:
    Line <4> of file <input string>: The line
        <input /jenkins/workspace/aspect_PR-4542/benchmarks/viscoelastic_sheared_torsion/viscoelastic_sheared_torsion.prm>
could not be parsed: please check to make sure that the file is not missing a 'set', 'include', 'subsection', or 'end' statement.
```

anne-glerum (approved):

PR #4549, opened 2022-04-19
Use centralized quadrature formulas.

bangerth:
Follow-up to #4536 and #4544.

/rebuild

### DIFF
--- a/source/simulator/helper_functions.cc
+++ b/source/simulator/helper_functions.cc
@@ -802,7 +802,7 @@ namespace aspect
       {
         const types::boundary_id top_boundary_id = geometry_model->translate_symbolic_boundary_name_to_id("top");
 
-        const QGauss<dim-1> quadrature (parameters.stokes_velocity_degree + 1);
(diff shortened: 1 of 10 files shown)


source/simulator/helper_functions.cc
```diff
@@ -1402,8 +1402,10 @@ namespace aspect
     Quadrature<dim> quadrature_formula(quadrature_points);
 
     // Quadrature rules only used for the numerical integration for better accuracy
-    const QGauss<dim> quadrature_formula_0 (advection_field.polynomial_degree(introspection)+1);
-
+    const Quadrature<dim> &quadrature_formula
+      = (advection_field.is_temperature() ?
+         introspection.quadratures.temperature :
+         introspection.quadratures.compositional_fields);
```
bangerth, line 1408: It took me quite a while to understand that here we are not always dealing with compositional fields, and that the tests that fail are exactly those where the temperature and compositional fields use different polynomial degrees -- which seems to be the case only for melt tests, getting me completely onto the wrong track :-)

Anyway, with a bit of luck this will now test correctly.

tjhei (approved):

PR #4552, opened 2022-04-21
More using of centralized quadrature objects.

bangerth:
Another follow-up to #4536 and #4544. This converts most of the rest of the uses of `QGauss`, but not all. 

I needed to also create a quadrature object for the pressure, which is what the first commit does. I somehow forgot that when I created the other quadrature objects.

/rebuild


gassmoeller (approved):
Looks good, thanks!

PR #4557, opened 2022-04-26
Correct the computation of e_dot_ii.

bangerth:
#3755 made me think about the computation of e_dot_ii. In quite a number of places, we compute it as
```
  sqrt( abs( second_invariant( deviator( eps ) ) ) )
```
but that seems wrong. I don't think we should be taking the absolute value -- the second invariant of a deviator should probably already be non-negative (I haven't tried to prove this, though).

This patch fixes this, but I imagine that it won't work because there will likely be cases where numerical round-off means that the second invariant is `-1e-16` or some such, and the square root will not be happy. I just want to try this out.

The "correct" patch -- should the current patch fail because of the anticipated issue with round off -- would be to treat everything negative as zero. I'll do that instead if the patch fails.

/rebuild

### DIFF
--- a/source/material_model/diffusion_dislocation.cc
+++ b/source/material_model/diffusion_dislocation.cc
@@ -41,7 +41,7 @@ namespace aspect
       // to prevent a division-by-zero, and a floating point exception.
       // Otherwise, calculate the square-root of the norm of the second invariant of the deviatoric-
       // strain rate (often simplified as epsilondot_ii)
-      const double edot_ii = std::max(std::sqrt(std::fabs(second_invariant(deviator(strain_rate)))),
+      const double edot_ii = std::max(std::sqrt(std::max(-second_invariant(deviator(strain_rate)), 0.)),
                                       min_strain_rate);
 
 
--- a/source/material_model/grain_size.cc
+++ b/source/material_model/grain_size.cc
@@ -184,7 +184,7 @@ namespace aspect
       // get grain size and limit it to a global minimum
       const unsigned int grain_size_index = this->introspection().compositional_index_for_name("grain_size");
       double grain_size = composition[grain_size_index];
-      grain_size = std::max(std::exp(-grain_size),min_grain_size);
+      grain_size = std::max(std::exp(-grain_size), min_grain_size);
 
       composition[grain_size_index] = grain_size;
     }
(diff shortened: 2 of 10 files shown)


bangerth:
Ah, as expected, this patch falls in the "Puckett bucket" category :-) 

So is the right approach to use `std::sqrt( std::max(..., 0))` instead? What do you material model people say?

anne-glerum:
From what I remember, the `abs` is there because for deviatoric tensors, the second moment invariant equals the -1x the second principal invariant (Eq. A18 of Zienkewicz & Taylor, The finite element method). 
The deal.ii function second_invariant computes the principle invariant (`1/2(Tr(A)^2-Tr(A^2)`), but we want the moment invariant (`1/2 EijEji`). E.g. Ismael-Zadeh & Tackley and Gerya define the second invariant of the strain rate as `sqrt(0.5 EijEji)`. (The sqrt makes it the _effective_  strain rate I think though.)
But abs(B) is not necessarily the same as -1xB.

bangerth:
So should we just replace `abs(...)` by `-...` then?

gassmoeller:
If @anne-glerum is correct (I remember the same reasoning she mentioned, but have not checked myself), then replacing `abs` with `-...` should be fine, and should give us the additional advantage of crashing (with FPE?) if the second invariant is ever not negative. Want to give it a try? There could be a funny interaction with #4559 if our deviatoric tensor (with -1/3 in 2d compressible cases instead of -1/2) has a second invariant that is not always negative, but that would be good to know too.

bobmyhill:
I always assumed that the `abs` was there to avoid small -ve numbers. 
I'd be in favour of `std::sqrt( std::max(X, 0))`, where `X` is either `second_invariant( deviator( eps ) )` or `second_invariant( deviator( eps ) )`, whichever is positive.

bangerth:
I've implemented this now, as suggested: Where we previously say
```
  std::sqrt( std::abs( second_invariant ) )
```
I'm now using
```
  std::sqrt( std::max( -second_invariant, 0. ) )
```
instead.

Let's see whether that actually works or not :-)

bangerth:
Oh, surprise, this actually works! Hurray!

bangerth:
Any other thoughts?

gassmoeller (approved):
Looks good to me, let's make this the new standard.

PR #4562, opened 2022-04-26
Use C++11-style closing double template angle brackets.

bangerth:
/rebuild

### DIFF
--- a/include/aspect/particle/property/interface.h
+++ b/include/aspect/particle/property/interface.h
@@ -529,8 +529,8 @@ namespace aspect
            * @return A vector that contains pairs of the property names and the
            * number of components this property plugin defines.
            */
-          std::vector<std::pair<std::string, unsigned int> >
-          get_property_information() const override;
+          std::vector<std::pair<std::string, unsigned int>>
+                                                         get_property_information() const override;
 
           /**
            * Read the parameters this class needs to determine which integrator is used,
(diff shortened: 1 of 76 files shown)


include/aspect/particle/property/interface.h
```diff
@@ -529,8 +529,8 @@ namespace aspect
            * @return A vector that contains pairs of the property names and the
            * number of components this property plugin defines.
            */
-          std::vector<std::pair<std::string, unsigned int> >
-          get_property_information() const override;
+          std::vector<std::pair<std::string, unsigned int>>
+                                                         get_property_information() const override;
```
bangerth, line 533: astyle gives us a bit of a bum indentation for these kinds of changes, but this isn't new -- we have plenty of these places already.

My main motivation for this patch is that I'd like to experiment with ways of tricking astyle into doing better, but for that I need all of these places to uniformly use `>>` instead of `> >`.

tjhei (approved):
I am okay with that change.

include/aspect/particle/property/interface.h
bangerth: I now have a patch that makes this look good again (along with many other places).

gassmoeller: Do you want us to merge this PR first and apply the patch in a separate PR, or do you want to include it here?

bangerth: I'm an incrementalist. Let's do things in separate PRs -- if you merge this, then I'll post the other patch as a follow-up once I get home.

PR #4566, opened 2022-04-27
Fix deprecation warnings and incompatibility with deal.II

gassmoeller:
This PR supersedes #4565. It fixes a number of deprecation warnings and uses a new version of a deal.II signal. In the process it also fixes a crash in current master with a current deal.II dev. Because this PR has a different implementation of the `cell_weight` function than #4565 for deal.II 9.4 the fix in the earlier PR is no longer necessary.


tjhei (approved):
Looks good!

PR #4575, opened 2022-04-29
Improve astyle indentation

bangerth:
This is the promised follow-up to #4562. It improves indentation via the hack described in the `indent` script where I temporarily pretend that everything that contains `>>` is `> >`, then run astyle, then change it back. The result looks much nicer.

@MFraters -- FYI

/rebuild

### DIFF
--- a/benchmarks/tosi_et_al_2015_gcubed/tosi.cc
+++ b/benchmarks/tosi_et_al_2015_gcubed/tosi.cc
@@ -578,7 +578,7 @@ namespace aspect
       // the values of the compositional fields are stored as blockvectors for each field
       // we have to extract them in this structure
       std::vector<std::vector<double>> prelim_composition_values (this->n_compositional_fields(),
-                                                                  std::vector<double> (n_q_points));
+                                                                   std::vector<double> (n_q_points));
 
       typename MaterialModel::Interface<dim>::MaterialModelInputs in(n_q_points,
                                                                      this->n_compositional_fields());
--- a/contrib/utilities/indent
+++ b/contrib/utilities/indent
@@ -149,8 +149,11 @@ dos_to_unix()
 {
     f=$1
     tr -d '\015' <$f >$f.tmp
-    diff -q $f $f.tmp >/dev/null || mv $f.tmp $f
-    rm -f $f.tmp
+    if ! diff -q $f $f.tmp >/dev/null ; then
+      mv $f.tmp $f
+    else
+      rm $f.tmp
+    fi
 }
 export -f dos_to_unix
 find tests include source unit_tests cookbooks benchmarks \( -name '*.cc' -o -name '*.h' -o -name '*.prm' -o -name '*.h.in' \) -print | xargs -n 1 -P 10 -I {} bash -c 'dos_to_unix "$@"' _ {}
--- a/contrib/utilities/indent
+++ b/contrib/utilities/indent
@@ -103,6 +103,11 @@ echo "--- Found astyle 2.04 at ${astyle}"
 # look-behind expression is not part of the match itself (which is only ">"
 # in the expression above) and is only used to restrict which occurrences
 # of ">" are actually matches (and consequently substituted).
+#
+# For unclear reasons, astyle insists on appending a separate empty line
+# at the end of the file. Fix this up using the last two 'sed' commands,
+# using the same logic as in the 'ensure_single_trailing_newline()'
+# function below.
 indent_file() {
   file="${1}"
   tmpfile="$(mktemp "${TMPDIR}/$(basename "$1").tmp.XXXXXXXX")"
@@ -110,7 +115,10 @@ indent_file() {
   cat ${file} \
       | perl -p -e 's/(?<=>)>/ >/g;' \
       | "${astyle}" --options=contrib/utilities/astyle.rc \
-      | perl -p -e 's/(?<=>) >/>/g;' > ${tmpfile}
+      | perl -p -e 's/(?<=>) >/>/g;' \
+      | sed -e :a -e '/^\n*$/{$d;N;};/\n$/ba' \
+      | sed -e '$a\' \
+      > ${tmpfile}
 
   if ! diff -q "${file}" "${tmpfile}" >/dev/null; then
     mv "${tmpfile}" "${file}"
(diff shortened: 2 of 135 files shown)


benchmarks/tosi_et_al_2015_gcubed/tosi.cc
```diff
@@ -578,7 +578,7 @@ namespace aspect
       // the values of the compositional fields are stored as blockvectors for each field
       // we have to extract them in this structure
       std::vector<std::vector<double>> prelim_composition_values (this->n_compositional_fields(),
-                                                                  std::vector<double> (n_q_points));
+                                                                   std::vector<double> (n_q_points));
```
bangerth, line 581: This is an unfortunate side effect. Because astyle indents with the `>>` in the first line treated as `> >` , it then moves the last line by one space to the right. Fortunately, these occurrences are quite rare, as can be seen from the rest of the patch.

tjhei:
Did you experiment with the option "--close-templates" yet? It is supposed to remove spacing around < >. Maybe that improves the result (or makes this PR superfluous).

tjhei:
I just tried astyle 3.1 (newest release), but it does not make a big difference: https://github.com/geodynamics/aspect/compare/main...tjhei:astyle-3.1?expand=1

tjhei:
A few questions for you:
- Is it acceptable to require perl for indentation? If I understand it correctly, we did not require it before.
- Does this work on OSX?
- I am worried that a ``make indent`` will now always modifies all files, which leads to recompilation of most of ASPECT while this was not true before. Can you confirm? If that is the case, we might need to create a temporary file, replace, indent, and then compare against the original.

bangerth:
I did not know about that option. Let's get that merged in #4614 anyway, and then I'll rebase this PR on top of #4614 to see what changes there remain. From a cursory look, I believe that this patch still improves the indentation of follow-up lines.

bangerth:
I've updated this following #4614. I think it is still an improvement.

include/aspect/geometry_model/chunk.h
```diff
@@ -142,7 +142,7 @@ namespace aspect
            * Return a copy of this manifold.
            */
           std::unique_ptr<Manifold<dim,dim>>
-                                          clone() const override;
+          clone() const override;
```
bangerth, line 145: It's these kinds of declarations that now look nicer.

benchmarks/tosi_et_al_2015_gcubed/tosi.cc
gassmoeller: Can you rebase again after #4635 is merged? That should get rid of these changes.

contrib/utilities/indent
```diff
@@ -61,11 +71,74 @@ fi
...
+#
+# We use astyle 2.04 for indentation, which is a tool that predates C++11
+# and that consequently has trouble with things such as
+#   std::vector<Point<dim>>
+# where it interprets the closing `>>` as a right shift operator. It then
+# gets its template closing angle bracket counter all out of whack and
+# produces awkward indentations. To avoid this, we first convert all closing
+# `>>` into `> >`, run astyle, and then change it back. This leads to reasonable
+# results.
+#
```
gassmoeller, line 84: Is this documentation still up-to-date after #4614 and #4635?

gassmoeller (commented):
Looks reasonable to me except for my two questions. We should let someone with a Mac test if this still works for them, we had trouble with MacOS and these things in the past.

contrib/utilities/indent
bangerth: Yes. The way I interpret the flag used for #4614 and made the default in #4635 is that it "keeps the template brackets close to the text", not that it deals with the closing brackets as ">>".

bangerth:
Anyone with a Mac who can test this?

contrib/utilities/indent
```diff
@@ -61,11 +71,74 @@ fi
...
+      | "${astyle}" --options=contrib/utilities/astyle.rc \
+      | perl -p -e 's/(?<=>) >/>/g;' \
+      | sed -e :a -e '/^\n*$/{$d;N;};/\n$/ba' \
+      | sed -e '$a\' \
+      > ${tmpfile}
+
+  if ! diff -q "${file}" "${tmpfile}" >/dev/null; then
+    mv "${tmpfile}" "${file}"
+  else
+    rm ${tmpfile}
```
tjhei, line 126:
```suggestion
    rm "${tmpfile}"
```

tjhei (approved):

tjhei:
let's wait for @naliboff to report if it works for him

naliboff:
@bangerth @tjhei - I compiled this branch on my Mac in debug mode (using deal.II 9.3.0) and was able to run make indent with no issue:
```
[100%] Built target aspect
bash-3.2$ make indent
[100%] Indenting all ASPECT header and source files...
--- Found astyle 2.04 at /Users/naliboff/software/astyle/build/mac/bin/astyle
--- Indenting all ASPECT header and source files
[100%] Built target indent
bash-3.2$ 
```

tjhei (approved):

bangerth:
If I see this right, then we are checking the indentation with the `indent` script from the `main` branch, right? Is that why the testers fail? The indent check wants to revert all of my indentation changes.

tjhei:
That is weird, because it should run the script after looking at the merged state (take master, merge this branch into it, then test).

Can you temporarily modify the indent script to print something to the screen to confirm?

bangerth:
I'm confused then, also because before I merged your one-line quotation change, the testers were green. Anyway, checking this out now on my own system, it seems broken. Will get to it asap.

bangerth:
Ah, bizarre. It's the interaction with `--close-templates` introduced in #4635. In the current patch, I first expand every occurrence of `>>` into `> >` (with a space). This allows astyle to see that it's two closing template brackets, not the right-shift-operator.

When you use `--close-templates`, astyle apparently first removes the space between, turning it back to `>>`, *and then it interprets the result as a right-shift-operator*. As a consequence, we get back the original indentation!

The solution is to undo #4635. That feature was nice as a one-time fix of a bunch of code, but it's broken...

I pushed the reversion here. This should work again now.

tjhei:
That is just bizarre indeed.

tjhei:
Let's see how it goes!

PR #4584, opened 2022-05-04
Moved several benchmark's documentation

sac-bsa:
Moved the documentation for the Rayleigh-Taylor instability, sinking block, SolCx, SolKz, and inclusion benchmarks. I also updated the Makefile so that the .prm files in compositional_field subdirectories would not cause errors in building the manual.

### For all pull requests:


gassmoeller (commented):
All the moved benchmarks look fine, but can you explain more why the parameter files in the `compositional_fields` directories cause a problem? I can see that they are now reachable from doc/ which they were not before, but why is that a problem?

sac-bsa:
It became a problem as make finds the parameter files in the `compositional_fields` directories to be required to build the manual as part of `aspect-cookbook-prms` and `aspect-cookbook-prms-out`. The rules of the Makefile as currently in main tried to generate `cookbooks/benchmarks/solkz/compositional_fields/doc/solkz_compositional_fields.prm.out` while no `cookbooks/benchmarks/solkz/compositional_fields/doc` directory exists and additionally such a file would not be used for documentation. If I understand the Makefile correctly, once all the cookbooks (and benchmarks) are moved, all of the parameter files that we actually need `.prm.out`s for will be in `aspect-doc-prms`.

gassmoeller:
Ok, thanks for the explanation. I was just worried that we will need to add lots of extra cases to the Makefile whenever we have benchmarks with different subdirectories, but let's face that when it actually happens.

PR #3768, opened 2020-08-12
Add teaching cookbooks and materials

jdannberg:
I'd like to contribute my teaching materials to the ASPECT repository, so that others can use them, and so that the input files keep working with the development version. Here, I've only added a very first input file, so that we can discuss if the structure I am using works, or if we want to change something.

PR #4586, opened 2022-05-06
set velocity in computation of initial lithostatsic pressure

jdannberg:
In the computation of the initial lithostatic pressure, the velocity, which is a material model input, was not set to any value, which means it was NaN. Material models that use the velocity in their evaluate function would then cause a floating point exception. That is fixed by this PR. 

This is needed so that #3768 works. 

### DIFF
--- a/source/boundary_traction/initial_lithostatic_pressure.cc
+++ b/source/boundary_traction/initial_lithostatic_pressure.cc
@@ -142,8 +142,8 @@ namespace aspect
       // We do not need the viscosity.
       in0.strain_rate.resize(0);
 
-      // The velocity is zero for the lithostatic case.
-      in0.velocity[0]= Tensor<1,dim> ();
+      // We set all entries of the velocity vector to zero since this is the lithostatic case.
+      in0.velocity[0] = Tensor<1,dim> ();
 
       // Evaluate the material model to get the density.
       this->get_material_model().evaluate(in0, out0);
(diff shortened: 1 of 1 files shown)


source/boundary_traction/initial_lithostatic_pressure.cc
```diff
@@ -142,6 +142,9 @@ namespace aspect
       // We do not need the viscosity.
       in0.strain_rate.resize(0);
 
+      // The velocity is zero for the lithostatic case.
```
naliboff, line 145:
```suggestion
      // The velocity is zero for the lithostatic case. Note that Tensor<1,dim> () initializes a 1D array of length "dim" with zeros.
```

naliboff (commented):
@jdannberg - Thanks for the fix and interesting this error had not appeared in any previous tests, cookbooks or benchmarks. 

I did not know that the deal.II Tensor function is initializing the arrays with zeros, but this is indeed clearly stated in the deal.II documentation (and the fact that the code here works). Would it be worth adding additional text to the comments in this PR, so it is clear to others?  I added an example suggestion.

Otherwise good to go.

source/boundary_traction/initial_lithostatic_pressure.cc
jdannberg: Is this better? I didn't want to make the comment too long, since that kind of defeats the purpose of using this shorter version of the constructor that does not include the values of the tensor entries (and I just copied this line from a different place in the code).

naliboff: Yep, this works, thanks!

gassmoeller (approved):

Issue #4177, opened 2021-07-12
Two stress visualization tests are the same


anne-glerum:
The tests
`visualization_stress.prm`
and 
`visualization_stress_pointwise.prm`
are actually the same as
`set Point-wise stress and strain = true` is set in both.
Output (only gnuplot) is the same. 
The first test should not output point-wise stress and strain.

bangerth:
Did the default for that parameter change at some point?

anne-glerum:
I don't know, but both tests literally set 
`set Point-wise stress and strain = true` in the prm

bangerth:
Ha, yes, this makes no sense :-)

PR #4592, opened 2022-05-12
Disabled pointwise visualization

Wang-yijun:
Fixed issue #4177 by `set Point-wise stress and strain = false` in `visualization_stress.prm` so that it is different from `visualization_stress_pointwise.prm.`

### For all pull requests:


MFraters:
Hey Yijun,

Thanks for making this pull request! 

Looking at #4177 it also states: 
> The first test should not output point-wise stress and strain.

Looking at the file, I am not sure what is mend with that. @anne-glerum, do you want `set List of output variables` to be empty for `visualization_stress.prm`

/rebuild

anne-glerum:
@Wang-yijun 
Thanks for making this change!

@MFraters 
These are two identical tests, apart from their name. One has the extension `_pointwise` to indicate that the stress and strain output is point-wise instead of element-wise (as well as the comment `# Like visualization_stress.prm but with point-wise output`). But both tests use the point-wise output. So the visualization_stress.prm should not have `set Point-wise stress and strain = true` but `set Point-wise stress and strain = false`.

MFraters:
@anne-glerum Thanks for the explanation, it makes sense now :)

@timo, it seems like the rebuild command didn't work for me. Did I do it incorrectly?

bangerth:
@MFraters It worked now. Can you check again?

MFraters (approved):
This looks good to me. @anne-glerum Since it was your issue, does this look good to you?

anne-glerum:
Yes, great! @Wang-yijun Once it's merged, you could try to close the issue (I'm not you have the rights).

gassmoeller:
I was not part of this discussion, but it looks like everyone agreed this is resolved. I will merge this.

PR #4600, opened 2022-05-16
QLS particle interpolation limiting scheme

sac-bsa:
@egpuckett and I created a limiter for the quadratic least squares particle interpolator that preserves the bounds it finds in the model. The limiting methodology is to take a convex combination of the QLS function and the arithmetic average of the particles in the cell (taken from the cell average interpolation) so that the maximum and minimum values of the QLS function within the cell are within the bounds. The bounds differ slightly from those implemented in `source/particle/interpolator/bilinear_least_squares.cc` including the cell averages of neighboring cells and an optional boundary extrapolation. Both of these changes to the bounds help improve the accuracy of the limited QLS on smooth problems. They should have no effect on characteristic functions (those in the form `if(condition, 1, 0)` such as the beam in `benchmarks/viscoelastic_bending_beam/viscoelastic_bending_beam_particles.prm`.

### For all pull requests:


source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -155,39 +250,398 @@ namespace aspect
...
+
+                        point[0] = 0.5;
+                        point[1] = 0.5;
+                        value_at_corner_point =  evaluate(c[property_index], point);
+                        interpolation_min = std::min(interpolation_min, value_at_corner_point);
+                        interpolation_max = std::max(interpolation_max, value_at_corner_point);
+                      }
+                    else
+                      {
+                        // If finding the critical point of the function (or along a cell edge) would
```
gassmoeller, line 352: I will take another look at the 3d case after the changes I suggested above.

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -155,39 +250,398 @@ namespace aspect
...
+                    if ((interpolation_max - cell_average_values[property_index]) > std::numeric_limits<double>::epsilon() &&
+                        (cell_average_values[property_index] - interpolation_min) > std::numeric_limits<double>::epsilon())
+                      {
+                        const double alpha = std::max(std::min((cell_average_values[property_index] - property_minimums[property_index])/(cell_average_values[property_index] - interpolation_min),
+                                                               (property_maximums[property_index]-cell_average_values[property_index])/(interpolation_max - cell_average_values[property_index])), 0.0);
+                        // If alpha >= 1, then using it would make the function grow from the cell average to meet the bounds.
+                        if (alpha < 1.0)
+                          {
+                            c[property_index] *= alpha;
+                            c[property_index][0] += (1-alpha) * cell_average_values[property_index];
```
gassmoeller, line 624: I understand the first line, but why do you have to modify the cell average (coefficient[0]) at all? Should'nt the average be conserved?

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -155,39 +250,398 @@ namespace aspect
...
+                          }
+                      }
+                  }
               }
           }
         unsigned int index_positions = 0;
+        const auto &mapping = this->get_mapping();
```
gassmoeller, line 631: I do not think there is much saved with this renaming. Just use `this->get_mapping()` below.

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -155,39 +250,398 @@ namespace aspect
...
-
+                double interpolated_value = evaluate(c[property_index], relative_support_point_location);
                 // Overshoot and undershoot correction of interpolated particle property.
-                if (use_global_values_limiter)
+                if (use_quadratic_least_squares_limiter[property_index])
                   {
-                    interpolated_value = std::min(interpolated_value, global_maximum_particle_properties[property_index]);
-                    interpolated_value = std::max(interpolated_value, global_minimum_particle_properties[property_index]);
+                    interpolated_value = std::min(interpolated_value, property_maximums[property_index]);
+                    interpolated_value = std::max(interpolated_value, property_minimums[property_index]);
```
gassmoeller, line 644: This should not happen anymore after line 639, correct? For the bilinear least squares interpolator we added Asserts here that ensure we never exceed the bounds of the particle values after interpolation, can we do the same here?

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -210,25 +664,20 @@ namespace aspect
...
-                                   "'Global particle property minimum' values separated by ',' has to be "
-                                   "the same as the number of particle properties.");
-                prm.declare_entry("Use limiter", "false",
-                                  Patterns::Bool (),
-                                  "Whether to apply a global particle property limiting scheme to the interpolated "
-                                  "particle properties.");
-
+                prm.declare_entry("Use quadratic least squares limiter", "true",
+                                  Patterns::List(Patterns::Bool()),
+                                  "Limit the interpolation of particle properties onto the cell, so "
```
gassmoeller, line 669:
```suggestion
                                  "Limit the interpolation of particle properties onto the cell, so that"
```

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -210,25 +664,20 @@ namespace aspect
...
+                                  "the value of each property is no smaller than its minimum and no "
+                                  "larger than its maximum on the particles of each cell, and the "
+                                  "average of neighboring cells. If more than one value is given, "
+                                  "it will be treated as a list. Cannot be used in combination with 'Use limiter'");
```
gassmoeller, line 673:
```suggestion
                                  "it will be treated as a list with one component per particle property.");
```

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -210,25 +664,20 @@ namespace aspect
...
+                prm.declare_entry("Use boundary extrapolation", "false",
+                                  Patterns::List(Patterns::Bool()),
+                                  "Extends the range used by 'Use quadratic limiter' by linearly "
+                                  "interpolating values at cell boundaries from neighboring cells. "
+                                  "If more than one value is given, it will be treated as a list. "
+                                  "Enabling 'Use boundary extrapolation' without also enabling "
+                                  "'Use quadratic least squares limiter' does nothing.");
```
gassmoeller, line 680: Please change this behavior to throw an assert instead. We try to avoid having parameters that do not do anything unless another parameter is set.

gassmoeller (commented):
Looks generally good. I have a question about the boundary values that we should discuss in person, and I need to take another look at the 3D case after you have addressed the current comments.

It looks like you are computing the cell average for every neighbor of every cell. This may be expensive, can you post the timing output of a model with a large number of particles in optimized mode? How much time does the particle interpolation take, e.g. compared to the particle advection?

source/particle/interpolator/quadratic_least_squares.cc
sac-bsa: Unlike bilinear least squares, in quadratic least squares c[0] cannot be expected to approach the cell average as the number of particles in a cell increases (keep this thought in mind as one of the changes @egpuckett and I are looking to make to bilinear will involve this idea). This same change applied later in the code could instead read 
```
interpolated_value = evalutate_...
interpolated_value = alpha * interpolated_value + (1-alpha) * cell_average
```
As alpha decreases, the output of the limiter scheme approaches cell average.

sac-bsa:
I ran SolKz at refinement 5 with 100ppcpd to measure the time it should take. I ran each interpolator ten times in a row.
| Cell Average | Bilinear | Quadratic w/o lim | Quadratic w/ lim |
|------------------|------------|--------------------------|------------------------|
| 10.627s       | 10.852s | 12.174s                  | 12.425s                |

This is the time output from the last quadratic least squares with limiting.
```
| Total wallclock time elapsed since start     |      10.2s |            |
| Assemble Stokes system           |         1 |    0.0142s |      0.14% |
| Assemble temperature system      |         1 |    0.0215s |      0.21% |
| Build Stokes preconditioner      |         1 |    0.0206s |       0.2% |
| Initialization                   |         1 |     0.115s |       1.1% |
| Particles: Advect                |         2 |       1.9s |        19% |
| Particles: Copy                  |         1 |      1.65s |        16% |
| Particles: Exchange ghosts       |         2 |     0.583s |       5.7% |
| Particles: Generate              |         1 |      2.03s |        20% |
| Particles: Initialization        |         1 |  0.000113s |         0% |
| Particles: Initialize properties |         1 |      1.32s |        13% |
| Particles: Interpolate           |         2 |      1.81s |        18% |
| Particles: Sort                  |         2 |     0.578s |       5.7% |
| Postprocessing                   |         1 |    0.0187s |      0.18% |
| Setup dof systems                |         1 |   0.00863s |         0% |
| Setup initial conditions         |         1 |      3.59s |        35% |
| Setup matrices                   |         1 |    0.0553s |      0.54% |
| Solve Stokes system              |         1 |    0.0645s |      0.63% |
```

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -136,17 +187,61 @@ namespace aspect
...
+            for (unsigned int face_id = 0; face_id < dim * 2; ++face_id)
+              {
+                if (found_cell->at_boundary(face_id))
+                  {
+                    // Faces 0 and 1, 2 and 3, and 4 and 5 oppose each other. Note by XOR with 1 we get the opposing face. In 2D faces 4 and 5 do not exist.
+                    const unsigned int opposing_face_id = face_id ^ 1;
+                    const auto &opposing_cell = found_cell->neighbor(opposing_face_id);
+                    if (!opposing_cell->is_locally_owned())
+                      {
+                        continue;
```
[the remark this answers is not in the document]
sac-bsa: At the moment yes, if we are not able to access the values in opposing_cell then we ignore it. This is definitely not the desired behavior, so if you have suggestions on this I am all ears. If I am not mistaken though if we neighbor a cell that is not locally owned, we are already in a ghost cell?
Since this limiter uses particles in neighboring cells, should I add this?
```AssertThrow(prm.get_bool("Update ghost particles") == true, ExcMessage("The quadratic least squares limiting scheme requires that 'Update ghost particles' is enabled."));```

File: source/particle/interpolator/quadratic_least_squares.cc [hunk 9]
```diff
@@ -155,42 +474,50 @@ namespace aspect
...
-                                          c[property_index][9] * relative_support_point_location[2] * relative_support_point_location[2];
+                    double interpolated_value = evaluate_interpolation_function(c[property_index], relative_support_point_location);
+                    // Overshoot and undershoot correction of interpolated particle property.
+                    if (use_quadratic_least_squares_limiter[property_index])
+                      {
+                        Assert(interpolated_value >= property_minimums[property_index] - std::max(std::abs(property_minimums[property_index]), std::abs(property_maximums[property_index])) * 10. * std::numeric_limits<double>::epsilon(), ExcInternalError());
+                        Assert(interpolated_value <= property_maximums[property_index] + std::max(std::abs(property_minimums[property_index]), std::abs(property_maximums[property_index])) * 10. * std::numeric_limits<double>::epsilon(), ExcInternalError());
+
+                        interpolated_value = std::min(interpolated_value, property_maximums[property_index]);
```
gassmoeller, line 515: Add a comment here that you do that to avoid values that are just barely outside the interpolation bounds.

source/particle/interpolator/quadratic_least_squares.cc
```diff
@@ -251,34 +575,63 @@ namespace aspect
...
+                  }
+                else
+                  {
+                    AssertThrow(false, ExcMessage("The size of 'Use quadratic least squares limiter' should either be 1 or the number of particle properties"));
+                  }
+                for (unsigned int i = 0; i < n_internal_components; ++i)
+                  quadratic_least_squares_limiter_parsed.push_back(false);
+                use_quadratic_least_squares_limiter = ComponentMask(quadratic_least_squares_limiter_parsed);
+
+                limiter_enabled_for_at_least_one_property = (use_quadratic_least_squares_limiter.n_selected_components(n_property_components) != 0);
```
gassmoeller, line 603: Oh, I see now why you stored this, you dont want to access the postprocessor to compute `n_components` again above. Ok, then. Keep the variable.

tests/particle_interpolator_quadratic_least_squares_2d_add_particles/screen-output
```diff
@@ -1,6 +1,10 @@
+-----------------------------------------------------------------------------
```
gassmoeller, line 1: Undo the changes to this file.

tests/particle_interpolator_quadratic_least_squares_2d_limiter.prm
```diff
@@ -0,0 +1,122 @@
+# A test based on the particle version of the SolKz benchmark that (1) tests to
```
gassmoeller, line 1: Update this text. Also it looks like this is nearly identical to the other test above. If so, you can use `include $ASPECt-SOURCE_DIR/tests/particle_interpolator_quadratic_least_squares_2d_add_particles.prm` to avoid most of the lines in the .prm file (except for Dimension, and Additional shared libraries).

gassmoeller (commented):
Very close, thanks for the changes. Let me know when I should take another look.

gassmoeller (approved):
Ready to go, thank you! Do not forget to add a section to the hackathon log document about it, I think it is a very useful feature.

Issue #4597, opened 2022-05-16
deal.ii version required for using the GMG solver


jdannberg:
I've been trying to run a model with the GMG, and I get the error message that "Mesh deformation with the GMG solver requires deal.II 9.3.2 or newer." But my model does not have mesh deformation (only GMG).

The error comes from these lines of code: https://github.com/geodynamics/aspect/blob/325d09ae4d3ea0634feb767828b31d305c785686/source/simulator/stokes_matrix_free.cc#L1375-L1377

I am not sure the newer version of deal.ii is required in that case, or if the error message should only trigger in cases with free surface.

tjhei:
@zjiaqi2018 can you try to take a look? The error should only trigger if we have mesh deformation.

jdannberg:
I can reproduce this with the ascii data slice test: https://github.com/geodynamics/aspect/blob/main/tests/ascii_data_slice.prm

PR #4602, opened 2022-05-17
Disable mesh deformation+GMG for dealii versions older than 9.3.2

zjiaqi2018:
Fix https://github.com/geodynamics/aspect/issues/4597


jdannberg:
This fixes the problem for me.

naliboff (approved):

PR #4610, opened 2022-05-17
fix test crameri_benchmark_1_gmg.prm

tjhei:
a box should not have a rotational nullspace


gassmoeller (approved):
Looks good, but you will have to update the test results.

gassmoeller:
Also do we not have an assert that catches this? Should we add one?

bangerth (approved):

tjhei:
How would you do that: list all geometry models and disallow unless Sphere, She'll, or spherical chunk? Not sure.

bangerth:
We do similar things for some of the boundary and initial conditions, which check via `dynamic_cast` that the geometry model is one of the ones the condition supports.

gassmoeller:
I think it is reasonable to only support the nullspace removal for geometry models that we know have a rotational nullspace (so spherical shell, sphere, chunk). 
We can leave the assert for another branch (like your nullspace branch), but you need to update the test.

tjhei:
I will do the assert separately.

gassmoeller (approved):

PR #4617, opened 2022-05-17
Documentation: Basic Equations - adiabatic heating

ljhwang:
correct a simple grammatical error


bangerth (commented):
I think we convinced ourselves that it was grammatically correct as is :-)

PR #4620, opened 2022-05-17
Be explicit by marking functions in a derived class as 'virtual'.

bangerth:
/rebuild

### DIFF
--- a/source/postprocess/visualization.cc
+++ b/source/postprocess/visualization.cc
@@ -59,6 +59,7 @@ namespace aspect
       {
         public:
 
+          virtual
           void
           evaluate_vector_field(const DataPostprocessorInputs::Vector<dim> &input_data,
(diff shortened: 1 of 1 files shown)


gassmoeller (changes requested):
See our discussion in https://github.com/geodynamics/aspect/pull/3249. According to c++ core guidelines a function should be either marked `virtual` or `override` but not both (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rh-override). Marking it as `override` includes that it is a virtual function. Do you feel strongly about this?

bangerth:
Oh, I see. Are we actually consistent in that?

bangerth:
Oh! (Insert terrified emoticon here.) I accidentally pushed the merged button here like on @ljhwang 's pull requests that I had all open. This was not intentional. I'll revert the addition of the `virtual` keyword instantaneously!

PR #4626, opened 2022-05-17
Documentation: Init cond adiabatic

ljhwang:
Some more minor punctuation clean up **AND** should the last word really be 'mantle' vs. manual ??? **PLEASE VERIFY.**


bangerth (approved):
Yes, should have been "mantle" all along :-)

PR #4632, opened 2022-05-17
Documentation: ICA punctuation

ljhwang:
I do not think we capitalize equation everywhere. Certainly not in the first paragraph.


bangerth:
Same as the other one.

PR #4633, opened 2022-05-17
Documentation: Mass Conservation punctuation

ljhwang:
Enforce lower case 'equation'.


tjhei:
Isn't it common to use Figure, Section, Equation X?

ljhwang:
I concede the point which means we need to change all lower case to upper case ;).

bangerth:
I think the rule is that if you are referencing a specific section, equation, figure with a name/number, then you use uppercase. So that would call for uppercase here. We are probably reasonably consistent for `Figure X`, but not sure about equations.

bangerth:
So this was correct to begin with because it references a concrete equation.

ljhwang:
Idk. But it should be consistent.  

Best,
-Lorraine

*****************************
Lorraine Hwang, Ph.D. - Director
UC Davis
geodynamics.org

PR #4244, opened 2021-07-15
add units for visualization postprocessors

ahollyday:
@bangerth 

Interface and CellDataVectorCreator clauses implemented in visualization postprocessors that store unit information.

PR #4637, opened 2022-05-17
Output physical units in viz output

bangerth:
This patch implements a scheme whereby viz postprocessors declare the physical units of what they are outputting, and this information is then delivered to deal.II's VTU output file writers. (Other formats could in principle make use of this, but don't currently have a way to obtain it nor use it.) 

This is work I originally started with @ahollyday last year in July (see #4244, which this PR supersedes), and I'm crediting him with the second commit. I went through all of the places in the second commit and corrected/augmented the units they report where necessary. I've also added the blurb at the end of the doc string for each of these plugins.

There is probably a substantial amount of breakage in all of the VTU files we write. Let's see what the testers say -- you can already place the Puckett Bucket by my feet, for later placement over my head.

/rebuild


include/aspect/postprocess/visualization.h
```diff
@@ -124,6 +124,20 @@ namespace aspect
...
+           * derived classes. The empty string, "", refers to an unknown
+           * or nonexistent unit.
+           *
+           * If a visualization postprocessor generates more than one
+           * output component, and if the different components have different
+           * physical units, then they should be separated by commas. If
+           * the different components have the same physical units, these units
+           * need to be specified only once and will apply to all components.
+           */
+          Interface (const std::string &physical_units = "");
```
tjhei, line 139:
```suggestion
          explicit Interface (const std::string &physical_units = "");
```

include/aspect/postprocess/visualization.h
```diff
@@ -140,6 +154,12 @@ namespace aspect
            */
           virtual void update();
 
+          /**
+           * Return the string representation of the physical units that a
+           * derived class has provided to the constructor of this class.
+           */
+          std::string get_physical_units () const;
```
tjhei, line 161: Is it on purpose that you return a copy?

source/postprocess/visualization.cc
```diff
@@ -144,10 +166,12 @@ namespace aspect
       {
         public:
           MeshDeformationPostprocessor ()
-            : DataPostprocessorVector<dim>( "mesh_velocity", UpdateFlags(update_values) )
+            : DataPostprocessorVector<dim>("mesh_velocity",
+                                           update_values)
           {}
 
 
+          virtual
```
tjhei, line 174: Remove this.

source/postprocess/visualization.cc
```diff
@@ -547,28 +620,84 @@ namespace aspect
...
+       * of entries in the two input arguments must be the same.
+       *
+       * This function also checks that output names are unique.
        */
-      void track_output_field_names(const std::vector<std::string> &output_data_names,
-                                    std::set<std::string> &output_data_names_set)
+      void
+      track_output_field_names_and_units(const std::vector<std::string> &output_data_names,
+                                         const std::vector<std::string> &output_units,
+                                         std::map<std::string,std::string> &output_data_names_and_units)
```
tjhei, line 631: Why not a const ref?

include/aspect/postprocess/visualization.h
tjhei: That is true, but I flag any one-argument constructors in reviews without thinking about these details. :-)

source/postprocess/visualization.cc
tjhei: Well, that makes sense. :-)

tjhei:
So, all tests continue to work because we don't use master for the tester. Does that mean that many tests will now break with incorrect output when somebody uses deal.II master? Not sure what to do about that but it is a bit unfortunate.

bangerth:
That is possible. But I think that might happen anyway -- we don't guarantee identical output between versions.

Fortunately, there are only relatively few `.vtu` files in the `tests/` directory. Almost all of them are for particle tests. In fact, there is apparently only a single test that stores `.vtu` files for the solution:
```{sh}
> find tests | egrep '\.vtu' | grep -v particles/
tests/parallel_output_group_2/solution/solution-00000.0001.vtu
tests/parallel_output_group_2/solution/solution-00001.0001.vtu
tests/parallel_output_group_2/solution/solution-00001.0000.vtu
tests/parallel_output_group_2/solution/solution-00000.0000.vtu
tests/parallel_output_group_0/solution/solution-00000.0002.vtu
tests/parallel_output_group_0/solution/solution-00001.0002.vtu
tests/parallel_output_group_0/solution/solution-00000.0001.vtu
tests/parallel_output_group_0/solution/solution-00001.0001.vtu
tests/parallel_output_group_0/solution/solution-00001.0000.vtu
tests/parallel_output_group_0/solution/solution-00000.0000.vtu
tests/parallel_output_group_1/solution/solution-00001.0000.vtu
tests/parallel_output_group_1/solution/solution-00000.0000.vtu
```
That seems acceptable. But these tests are *already* broken since your patch in https://github.com/dealii/dealii/pull/13336 .

bangerth:
For the record, this is how it's going to look like with current deal.II:
```
    <DataArray type="Float32" Name="velocity" NumberOfComponents="3" format="binary" units="m/s">
AQAAAOABAADgAQAAlQAAAA==eNqLVf1pa3Cjfy8DEDzISLZduad4D4hdwD1zN3ukoS2ILRlyxMadgdEOXQ0yQFYDAoHuorbo5rie2mBzveEumL1Vev/eoNtqYPaOaZ17QxqqMdQgm4msxjb0w55STl+85uACuMzBBZDdkNMkYbN85zIMvXWOfbs/aBbYoKtBDgdkNSDA4emyB585uAAuc3ABAEraYuc=
    </DataArray>
    <DataArray type="Float32" Name="p" format="binary" units="kg/m/s/s">
AQAAAKAAAACgAAAAbAAAAA==eNo7w8B+5u3nvT4+AiqB2/8pB4LY02X1fEDs3p/KgSDxRnaVwBNfCgJrVxeA2SA5EPt9YwFc/LT6/EDTx/Pg4iB258F5YLVKf5TBYno1BWA2yFwQ27qpAC4OUlt2Zh5cHMTecm9eIACnoUtH
    </DataArray>
    <DataArray type="Float32" Name="T" format="binary" units="K">
AQAAAKAAAACgAAAAUAAAAA==eNpjcCh2YQDhD8qujA56rgxQPpgNFAPhI1bargwN4a5CP2LAbJAciM1gEw8XZ+hoBWO4OJQPUgvWDxSD6QeZCVYH1AsTR9EPEofyAebeMZ0=
    </DataArray>
```

bangerth:
Tester now works.

source/postprocess/visualization.cc
```diff
@@ -548,28 +620,84 @@ namespace aspect
...
+            output_data_names_and_units[output_data_names[i]] = output_units[i];
+          }
+      }
+
 
-        for (const auto &name: set_of_names)
+
+      /**
+       * Keep track of the names and physical units of output quantities. The
+       * of entries in the two input arguments must be the same.
```
gassmoeller, line 674: This sentence is missing some
:slightly_smiling_face:

source/postprocess/visualization.cc
```diff
@@ -548,28 +620,84 @@ namespace aspect
...
+       *
+       * This variant of the function expand the single 'output_units' variable into as
```
gassmoeller, line 676:
```suggestion
       * This variant of the function expands the single 'output_units' variable into as
```

gassmoeller, line 676:
```suggestion
       * This variant of the function expands the single 'output_units' variable into as
```

source/postprocess/visualization/adiabat.cc
```diff
@@ -33,9 +33,12 @@ namespace aspect
       Adiabat<dim>::
       Adiabat ()
         :
-        DataPostprocessor<dim> ()
+        DataPostprocessor<dim> (),
+        Interface<dim>("K,kg/s/s/m,kg/m/m/m,kg/m/m/m/m")
```
gassmoeller, line 37: I appreciate that you stick very close to the basic SI units, but could you still use `Pa` for units of pressure and stress and `Pa*s` for viscosity? Not many users will recognize `kg/s/s/m` as `Pa`

gassmoeller, line 37: I appreciate that you stick very close to the basic SI units, but could you still use `Pa` for units of pressure and stress and `Pa*s` for viscosity? Not many users will recognize `kg/s/s/m` as `Pa`

source/postprocess/visualization/artificial_viscosity.cc
```diff
@@ -28,6 +28,15 @@ namespace aspect
   {
     namespace VisualizationPostprocessors
     {
+      template <int dim>
+      ArtificialViscosity<dim>::
+      ArtificialViscosity ()
+        :
+        CellDataVectorCreator<dim>("m/s/s")
```
gassmoeller, line 35: Shouldnt this be the unit of thermal conductivity (`W / (m*K)`, or in your expression `kg * m / ( s^3 * K)`?

gassmoeller, line 35: Shouldnt this be the unit of thermal conductivity (`W / (m*K)`, or in your expression `kg * m / ( s^3 * K)`?

gassmoeller (commented):
I can come back later to this, but for now I just want to request that we at least for very common units use the derived units (like pressure = Pa, viscosity Pa*s).
One other questions are properties that depend on `use_years_instead_of_seconds` (e.g. strain rate as `1/yr` instead of `1/s`). They are currently specified in seconds, and it is not easy to change them depending on this input parameter, because they are stored in a `const` class member. What do we do about this ?

bangerth:
I think I have addressed all of your comments, @gassmoeller !

source/postprocess/visualization.cc
gassmoeller: I should have been more clear and less witty in my comment, the second sentence is not clear and is likely missing a word. Which entries in which input arguments need to be the same (there are three)?

gassmoeller (commented):
Except for the one issue with the documentation of a function this is ready to go.

gassmoeller:
Can you check why the latex manual is not building anymore? Is it related to this PR?

gassmoeller:
I think you need to rebase this. The branch includes the broken test results on master that are unrelated to this PR.

bangerth:
OK, rebased. Let's see whether this is finally good enough :-)

gassmoeller (approved):

bangerth:
Alright, it works now!

bangerth:
Hallelujah!

PR #4639, opened 2022-05-17
Use standard idiom to write data in the background.

bangerth:
We're doing things right here with allocating a pointer and letting the consumer function (which may or may not run on a separate thread) delete it, but it can be done automatically if we use `std::unique_ptr` instead. If we run on the same thread, `std::unique_ptr` deletes the object. If we run on a separate thread, we *move* the pointer into the lambda function and the lambda function object deletes the pointed to object once it's done running (on the other thread).

/rebuild

### DIFF
--- a/include/aspect/postprocess/visualization.h
+++ b/include/aspect/postprocess/visualization.h
@@ -555,13 +555,12 @@ namespace aspect
          * A function that writes the text in the second argument to a file
          * with the name given in the first argument. The function is run on a
          * separate thread to allow computations to continue even though
-         * writing data is still continuing. The function takes over ownership
-         * of these arguments and deletes them at the end of its work.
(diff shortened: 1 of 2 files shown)


bangerth:
I'll point out that the other two local variables being referenced here, `filename` and `temporary_output_location` are not actually copied into the other thread's space (via the lambda function's captured variables). That's a bug. I'm surprised we never got bitten by that. I'll fix that in a follow-up.

gassmoeller (approved):
This makes sense and is safer. The reason we never noticed the other bug may be that there is no test using this feature. By default `write_in_background_thread` is false, and in a quick search I didnt see a test using it (I may have missed it).

bangerth:
The fix for the latter is ready and standing by.

PR #4640, opened 2022-05-18
Create indentation artifact by Github tester

gassmoeller:
Currently the Github actions workflow only checks the indentation, but does not provide a diff file that fixes it. This PR adds that by uploading the diff as an artifact. The artifact will be uploaded for every build, but if the indentation is fine the artifact will be empty.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -50,9 +50,17 @@ jobs:
         cd
         rm -rf astyle*
         astyle --version
-    - name: indent
+    - name: make indent
       run: |
         ./contrib/utilities/indent
+        git diff > changes-astyle.diff
+    - name: archive indent results
+      uses: actions/upload-artifact@v2
+      with:
+        name: changes-astyle.diff
+        path: changes-astyle.diff
+    - name: check indentation
+      run: |
         git diff --exit-code
     - name: compile
       run: |


gassmoeller:
This seems to work. I have on purpose broken the indentation, and after the testers are done you can download an indentation artifact from the workflow page (under "Summary"). I have not yet figured out how to post the artifact while some jobs are still running, it will only show up after the tests have finished.
I will leave the indentation broken until someone has reviewed this and fix it afterwards.

.github/workflows/linux.yml
```diff
@@ -50,9 +50,17 @@ jobs:
...
         ./contrib/utilities/indent
+        git diff > changes-astyle.diff
+    - name: archive indent results
+      uses: actions/upload-artifact@v2
+      with:
+        name: changes-astyle.diff
+        path: changes-astyle.diff
+    - name: check indentation
+      run: |
         git diff --exit-code
```
bangerth, line 64: You could presumably make this a bit cheaper by just saying "cat changes-astyle.diff". Or do you need the return code of `git diff` here?

tjhei:
Works, nice. Is there a way to download the file without having it inside a .zip file?

.github/workflows/linux.yml
gassmoeller: I need the return code to let the test crash if indentation changes are necessary, I could probably check if `changes-astyle.diff` contains anything, but I do not think it will save more than a few seconds.

gassmoeller:
> Is there a way to download the file without having it inside a .zip file?

No I do not think there is currently a way (at least there wasnt last year when I checked for it), it is a limitation by Github.

gassmoeller:
I removed the broken indentation, if you are ok with my replies this should be ready.

gassmoeller:
I was testing a few different versions based on Wolfgang's comment above. This is the final version and should be ready.

bangerth (approved):
Thank you :-)

gassmoeller:
Oh no :-(. Ok, can we go back to my initial version? I have timed a call to `git diff` and it literally takes 0.1 s on my laptop. Even if it 1s on the tester, I do not think this matters much (with the time I have spend on this I could have saved more than 1s on another PR).

bangerth:
Do it!

bangerth:
Go back, that is!

gassmoeller:
Done, this should be ready when the testers are done.

PR #4642, opened 2022-05-18
Add two merged chunks to initial lithostatic pressure plugin.

MFraters:
This adds the option for the two merged chunks to the  initial lithostatic pressure plugin.


gassmoeller (approved):
Have you checked that it actually works?

MFraters:
Yes, this is what I have been using in my models. I can make a new test if you want?

gassmoeller:
I think it is very unlikely to find a situation where this works for one chunk and two merged boxes, but not for two merged chunks. Let's merge it like this.

PR #4645, opened 2022-05-18
Fix which variables are captured by the lambda function for background output.

bangerth:
Follow-up to #4639. This patch actually fixes the issue that the variables we use in the lambda function (on another thread) are captured by reference and so may go stale.

/rebuild


tjhei (approved):

bangerth:
Can I do that in a separate PR? I've got a follow-up for this one that I'd like to get out of my system.

tjhei:
Alright. I don't trust that it works at this point,though. ;-)

bangerth:
For the test, see #4685.

PR #4646, opened 2022-05-18
Use unix line endings where possible.

bangerth:
The sphinx files were apparently checked in with DOS line endings. This is unfortunate, and git should choose line endings based on platform preferences. This patch changes things as far as I can see, and in the process also fixes a couple of unicode characters that have snuck their way into things.

/rebuild


bangerth:
This patch is best viewed via https://github.com/geodynamics/aspect/pull/4646/files?w=1#

tjhei (approved):

tjhei:
Awesome, I had noticed the DOS ending earlier as well.

Issue #4595, opened 2022-05-16
sphinx online documentation


tjhei:
Meta issue for anything related to the new documentation.

- [x] initial conversion #4593
- [x] parse version number from ~/VERSION
- [ ] move output into ASPECT build directory
- [x] setup readthedocs
- [x] update links on website, README, etc.
- [ ] make a helpful index page
- [x] run doc in CI?

gassmoeller:
We have a very early version of the documentation online here: https://aspect-doc.readthedocs.io/en/latest/user/index.html

cmills1095:
- [X] Introduction
  - [X] Referencing ASPECT
  - [X] Acknowledgements
- [ ] Geodynamic modeling assumptions and numerical methods in ASPECT
  - [X] Basic equations
    - [X] A comment on adiabatic heating
    - [X] Boundary conditions
    - [X] Two-dimensional models
    - [X] Comments on the final set of equations
  - [X] Coefficients
    - [X] Coefficient self-consistency
    - [X] Coefficient averaging
  - [X] Dimensional or non-dimensionalized equations?
    - [X] Years or seconds??
  - [X] Static or dynamic pressure?
  - [X] Pressure normalization
  - [X] Initial conditions and the adiabatic pressure/temperature
  - [X] Compositional fields
  - [X] Constitutive laws
  - [X] Numerical methods
  - [X] Approximate equations
    - [X] ALA
    - [X] TALA
    - [X] BA
    - [X] ICA
  - [X] Choosing a formulation in ASPECT
    - [X] Mass conservation approximation
    - [X] Temperature equation approximation
    - [X] Approximation of the buoyancy term
    - [X] Reference state: The adiabatic profile
    - [X] Combined formulations
  - [ ] Advection Stabilization
    - [ ] SUPG Stabilization
    - [ ] Entropy viscosity 
  - [ ] Free surface calculation
    - [ ] Arbitrary Lagrangian-Eulerian implementation
    - [ ] Free surface stabilization
  - [ ] Calculations with melt transport
  - [ ] Nullspace removal
  - [ ] Particles
  - [ ] Geometric Multigrid
- [ ] Installation
  - [ ] Docker Container
    - [ ] Installing Docker and downloading the ASPECT image 
    - [ ] Running ASPECT models
    - [ ] Developing ASPECT within a container
  - [ ] Virtual Machine
    - [ ] Installing VM software and setting up the virtual machine
    - [ ] Running ASPECT models 
[...]

PR #4649, opened 2022-05-18
Fix .md markup problems with <span.

bangerth:
We had all of these occurrences of `<span` at the end of the line when there is ASPECT first on the next line. I don't know where that came from (to mark up ASPECT in a specific way?) but either way it's wrong. This fixes it.

/rebuild

### DIFF
--- a/doc/sphinx/user/extending/compatibility.md
+++ b/doc/sphinx/user/extending/compatibility.md
@@ -2,14 +2,14 @@
 
 We strive to maintain compatibility for user written plugins with new versions
 of ASPECT for as long as possible. However,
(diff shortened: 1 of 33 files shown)


tjhei (approved):

tjhei:
> (to mark up ASPECT in a specific way?

Yes, please see https://github.com/geodynamics/aspect/issues/4595#issuecomment-1130182371
and https://github.com/geodynamics/aspect/issues/4595#issuecomment-1131809646
We should discuss how we want to print it.

bangerth:
Whatever anyone tells me regarding markup, I'll script it.

Issue #4485, opened 2022-02-10
Further optimize the gravity postprocessor


bangerth:
#4456 reshuffles much of the code in the gravity postprocessor. Another relatively low-hanging fruit pertains to the fact that process 0 currently computes a good bit of reference values (the `_theory` variables) that are then output along with the data computed on the mesh. It would not be very difficult to split this work across the various MPI processes instead, later to be collected on process 0 for output. This doesn't accelerate single-process jobs, but at least makes parallel jobs faster.

@cedrict -- FYI

bangerth:
As pointed out in #4652, there are also plenty of places in the file where we take something to the second or third power using `std::pow`. We know that that's slow. We should fix that as well.

bangerth:
The issue with `std::pow` is still there, so let me reopen this PR.

PR #4652, opened 2022-05-18
Separate computation and output in the gravity postprocessor.

bangerth:
This is my first step towards addressing #4485. We currently have a loop of the form
```
  if (processor zero)
  {
    do some output;
    for (i=...)
      {
        compute something;
        output the computed thing;
      }
  }
```

This patch restructures things to do the following instead:
```
  std::vector<...> results;
  if (processor zero)
    for (i=...)
      compute something;

  if (processor zero)
  {
    do some output;
    for (i=...)
        output the computed thing;
  }
```

In the second step, I will make computing the thing distributed across all processors, rather than let everyone other than process zero wait.

/rebuild

### DIFF
--- a/source/postprocess/gravity_point_values.cc
+++ b/source/postprocess/gravity_point_values.cc
@@ -372,6 +372,67 @@ namespace aspect
         }
 
 
+      // Now also compute theoretical values. These are output only
+      // on process zero and, for now, also only computed on process zero.
+      std::vector<double> g_theory(n_satellites);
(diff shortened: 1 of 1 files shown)


gassmoeller (commented):
Looks good except for my one comment (replacing std::pow where possible).

bangerth:
There are 11 places in this file (not just in the part I change) where we raise something to the second or third power. Let me address that in a separate PR throughout the file if you allow me to.

gassmoeller (approved):

Issue #3652, opened 2020-08-03
Move documentation generation to cmake


bangerth:
It has bothered me for many years now (I'd say for (n-1) of the n years that we've used cmake) that the documentation generation is driven by `doc/Makefile` and not part of the cmake machinery. What do people think about fixing this wart?

Upsides:
* More systematic
* Keeping the source tree free of the artifacts of running latex, doxygen, etc
* We can actually install the documentation

Downsides:
* When adding new files that need to be processed (specifically, `.prm` fragments), one has to run cmake to pick them up
* Output ends up in the build directory where people may not expect it (yet -- that's an educational question)
* Need to update website generation scripts
* Might require fiddling with the graphics path if compiling `manual.tex` from a different directory (see https://texblog.org/2017/12/05/the-path-to-your-figures/)

Opinions?

tjhei:
That makes sense to me. Would it make sense to do the markdown conversion first?

bangerth:
No longer relevant.

PR #4656, opened 2022-05-19
Create the .prm.tex files from all .prm files.

bangerth:
We need to eventually process all of the `.prm` snippets as input to sphinx. Right now, for the latex manual, this is done with a separate `Makefile` in the `doc/` directory, but that needs to eventually be replaced by something we do in cmake. This is a first step towards it.

The patch right now really only gathers the `.prm` files and runs them through a couple of perl files. This is essentially the same as the current `Makefile`, except that I'm using perl for both scripts where the `Makefile` uses one perl and one sed script. The result of the current steps also ends up in a sub-directory of the build directory, rather than the source directory.

At the moment, no target depends on doing all of this, so unless you type `make manual` in your build directory, nothing happens. And if you do say `make manual`, all that happens is that the `.prm` snippets are converted into `.prm.tex` snippets which are not actually used anywhere. All of that will happen in follow-up patches, but I thought it easier to propose these as separate and individually reviewable pull requests.

This is a first step towards #3652. 

/rebuild

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -673,6 +673,16 @@ ELSE()
 ENDIF()
 
 
+###########################################################
+# Having configured most of the compilation phase, now also
+# deal with other parts of the system.
+###########################################################
+
+# Start with the documentation
+
+ADD_SUBDIRECTORY(doc)
+
+
 # Find the deal.II parameter GUI and install helper script
 FIND_PROGRAM(PARAMETER_GUI_EXECUTABLE
         parameter_gui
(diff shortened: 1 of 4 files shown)


bangerth:
I should add, while this patch is of course pretty pointless by itself, it's building the infrastructure for what we need to do for the sphinx doc generation. I just thought I'd start by converting the latex generator to cmake because I know what the outcome should be.

tjhei:
> I just thought I'd start by converting the latex generator to cmake because I know what the outcome should be.

Is it useful to do this step considering we will throw it away anyways? How about you process them from prm to md (not doing anything except a code environment for now)? We can then try to embed those.

bangerth:
That's the idea: If I know how to set this up for latex, I have 90% of the work already done for md. I just don't yet know what the right format for md is with index etc, and we don't have the setup yet to generate the sphinx documentation in the build directory. So I decided to do it first for the latex version, with minimal changes necessary later on.

There is likely going to be much more setup necessary for building documentation in the build directory. I thought I'd create all of this in steps based on latex first because we know that that's already there and working. Call it a practice run.

gassmoeller (approved):
Looks good to me. Even if we have to change things a lot for the Sphinx documentation, including building the manual step-by-step into the cmake system is worth it.

Issue #4653, opened 2022-05-18
Documentation: Problems with section on Arbitrary Lagrangian-Eulerian implementation


ljhwang:
We noticed multiple issues with the page in the documentation for Arbitrary Lagrangian-Eulerian implementation:

https://aspect-documentation.readthedocs.io/en/latest/user/methods/freesurface/arbitrary-le-implementation.html

1. The header repeats. This has brought up the issue of what the proper header levels should be throughout the document as level 1 etc headers look like they are used inconsistently and the next header used is level 4.
2. The page is incomplete. Text and equations are missing from the bottom of the page (in comparison to the pdf).
3.

MFraters:
https://github.com/geodynamics/aspect/pull/4663 stated that that pull request fixed this issue, so I will close it.

PR #4663, opened 2022-05-19
Fix a number of sphinx markdown problems.

bangerth:
This addresses all occurrences of ` ` `{math} not starting at the top of the line.

The patch also contains a number of fixes where the `{math}` tag was missing from ` ` `. I suspect that there are many places, and I'm afraid they all need to be fixed by hand.

/rebuild


tjhei (approved):

bangerth:
Rebased.

gassmoeller:
This fixes #4653, see https://aspect-documentation--4663.org.readthedocs.build/en/4663/user/methods/freesurface/arbitrary-le-implementation.html.

PR #4661, opened 2022-05-19
urgent: fix indentation

tjhei:

PR #4665, opened 2022-05-19
Fix code block markup.

bangerth:
See the correct syntax in quickref.md, and confirmed by @cmills1095 .

/rebuild

### DIFF
--- a/doc/sphinx/user/extending/idea-of-plugins.md
+++ b/doc/sphinx/user/extending/idea-of-plugins.md
@@ -35,7 +35,7 @@ The way this is achieved is through the following two steps:
     the [aspect::GravityModel::Interface][] class (documentation comments have
     been removed):
 
-    ``` c++
+    ```{code-block} c++
     class Interface
         {
           public:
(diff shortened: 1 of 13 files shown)


tjhei (approved):
OK!

tjhei:
let's wait for #4661

bangerth:
Rebased.

bangerth:
I've also pushed a second commit that does the same markup change for ksh/gnuplot/prmfile. The latter two may not be configured so far, but we can get to that at some point.

bangerth:
Ugh, this triggers an internal error in sphinx. What do we do about that?

gassmoeller:
My guess would be Sphinx just doesnt know some of these other file formats. Can we merge the c++ separately and figure out which of the other formats work? If I followed this thread correctly (https://stackoverflow.com/questions/11315504/sphinx-list-of-supported-languages-for-highlighting) this is the list of supported languages: https://pygments.org/languages/

bangerth:
See also #4669.

bangerth:
But yes, sure. I dropped the second commit, let's see whether that helps.

gassmoeller (approved):

PR #4668, opened 2022-05-19
Remove .pdf files; use .svg instead.

### LINKED ISSUE #4666: Remove .pdf versions of figures.
Following #4659, we now have both pdf and svg files for figures. We should get rid of the pdf files and instead use `\includegraphics` with the svg versions.

bangerth:
This fixes #4666, but we should debate whether we want this. It builds on #4659 which adds `.svg` versions of all `.pdf` figures that go into the manual.

pdflatex *can*, after all, include `.svg` files. But it can only do that if we call `pdflatex --shell-escape` (which I believe is fine, and I've switched that on in the `Makefile`) and, more importantly, if the inkscape package is installed. If we can do that, then this patch will save 8MB of disk space. What do you all think?

/rebuild


bobmyhill:
I don't see saving 8 MB as a massive benefit, but I do think that duplicating images in different formats is a recipe for disaster.

IMO, svg -> pdflatex is marginally preferable to repeated pdf -> svg conversion followed by deletion. It's a shame we would require inkscape as a new dependency too, but I guess this gets an unenthusiastic thumbs up from me!

tjhei:
CI fails because inkscape is not installed. :-)
I don't like having that dependency. Should we just wait and delete the pdf files once we get rid of the latex manual?

bangerth:
Let's give it a whirl with inkscape installed :-)

tjhei:
This won't work for jenkins CI. We would need an updated docker image...

bobmyhill:
With some more digging, rsvg-convert might be a bit more lightweight than inkscape:
https://github.com/missinglinkelectronics/sphinxcontrib-svg2pdfconverter

There are a few comments about making rsvg-convert available in RTD too. Not sure whether they've done this yet, but it would be nice to let RTD do the PDF creation.

bangerth:
The easier way is to add the package to the main branch first, no new docker image required :-) See #4707.

gassmoeller (approved):
This passed all the tests and we have the .svg files already, so let's merge this.

PR #4685, opened 2022-05-19
Add a test for writing data in the background.

### LINKED ISSUE #4670: Add a test for writing output in the background.
As requested in #4645.

bangerth:
See the comment in #4645 by @tjhei that requested it. I chose to clone this test from a previous one that runs for four time steps on two processes to really make sure that doing things in the background has a chance of happening. This is in particular because the output for the current time step can only be queued if the previous time step's output has already finished, so it matters that we run more than one time step.

Fixes #4670.

/rebuild


bangerth:
Follow-up to https://github.com/geodynamics/aspect/pull/4639.

tjhei (approved):

tjhei:
Why does the test produce different output, can you take a look?

bangerth:
The test I cloned from only stored a subset of 3 of the generated output files, whereas I wanted to store all that are generated by running the test. So I used the ones created on my system, and the diffs are all in the ones that came from my system. I've updated them to the canonical ones from the tester.

bangerth:
Works now.

PR #4687, opened 2022-05-19
Add a test for parallel computation of gravity point values.

bangerth:
In preparation for #4485.

/rebuild


MFraters (approved):
Looks like it works to me. The resulting values are the same as the non-parallel test.

PR #4678, opened 2022-05-19
Drop duplicate markdown headers.

bangerth:
@cmills1095 -- FYI, the command line was this:
```
for i in `find . | grep .md` ; do cat $i | perl -e "local $/; my \$x = <>; \$x =~ s/\# (.*)\s+\#+ +\1/\# \1/g; print \$x;" > $i.tmp ; mv $i.tmp $i ; done
```
This probably took me ten minutes to figure out, though :-)

/rebuild

PR #4698, opened 2022-05-19
Documentation: large data

ljhwang:
Removes a level 4 header (why are these still showing up?) and adds the notebox.  
Still needs to be fixed for references.


MFraters:
Let's wait with merging this until #4678 is merged, which removes the duplicate headers. Otherwise looks good to me.

bangerth (approved):

PR #4704, opened 2022-05-20
Parallelize the computation of theoretical gravity values.

### LINKED ISSUE #4485: Further optimize the gravity postprocessor

bangerth:
Finally fixes #4485. Passes the tests, including the one that runs the gravity postprocessor in parallel (a test that was quite useful because I broke it with several intermediate versions of this patch).

@cedrict FYI

/rebuild

### DIFF
--- a/source/postprocess/gravity_point_values.cc
+++ b/source/postprocess/gravity_point_values.cc
@@ -378,60 +378,78 @@ namespace aspect
       std::vector<double> g_potential_theory(n_satellites);
       std::vector<SymmetricTensor<2,dim>> g_gradient_theory(n_satellites);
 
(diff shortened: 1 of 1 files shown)


gassmoeller (approved):
Looks good to go me.

PR #4707, opened 2022-05-20
Also require inkscape for the CI.

bangerth:
Necessary for #4668. @gassmoeller -- FYI.

Not sure whether rebuilding anything will actually do anything useful here.


gassmoeller (approved):
:+1: I will wait with merging until I see that this phase of the workflow succeeded and the required testers are done, but no need to wait for all of them.

PR #4711, opened 2022-05-20
Add Lorraine to list of trusted testers

gassmoeller:
With @ljhwang's new role as chief documentation officer (CDO) it would be useful if would not have to mark all of her PRs as ready to test before they run through the testers. This PR does this.


MFraters (approved):

ljhwang:
I'm not sure I should be trusted. ;)

PR #4622, opened 2022-05-17
add tests for GMG with elasticity

naliboff:
This PR adds two tests that use elasticity in combination with the GMG solver. While this combination is currently permitted, there are no tests that explicitly test to make sure it works. However, @zjiaqi2018 @tjhei and I did test the GMG + elasticity combination a few months ago and confirmed it worked for models with mesh deformation and/or surface tractions. This PR does not include tests for either of the latter two features (still disabled with GMG), but does ensure GMG is able to run with elasticity.
 
* [X] I have followed the [instructions for indenting my code](../blob/main/CONTRIBUTING.md#making-aspect-better).
* [X] I have tested my new feature locally to ensure it is correct.
* [X] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/main/tests/) directory.
* [ ] I have added a changelog entry in the [doc/modules/changes](../blob/main/doc/modules/changes) directory that will inform other users of my change.

PR #4713, opened 2022-05-20
Elasticity: unconditionally fill elastic outputs

tjhei:
We used to only fill elastic outputs when reaction terms are requested
(meaning strain_rate is provided). This seems to be a bug, because we
don't need the strain rate for this.

This probably fixes a bug in elasticity with GMG, where we do not
request the strain rate in all cases.

related to #4248


tjhei:
FYI @naliboff @zjiaqi2018 @anne-glerum

tjhei:
This indeed fixes the test that is found to be #4248. Ready for review/merge.

gassmoeller (approved):
Looks good to me.

tjhei:
@naliboff Can you check if this changes any of the test output of your GMG tests?

naliboff:
@tjhei - I just compared the tests results for the viscoelastic_stress_build-up problem, and either running the low-resolution test case (1 tilmestep) or regular benchmark to 20 time steps gives identical values for the stress_xx and stress_yy terms, but slightly different values for the much smaller stress_xy terms. Given they were run on the same machine, I'm inclined to merge #4622 so we can check how much the test changes in this PR.

naliboff:
@tjhei - I just ran the bending beam test in #4622 and those results changes in non trivial ways (RMS velocity, etc).

naliboff (approved):
This looks good to me as well. 

As noted, I think we should  merge first #4622 to have additional elasticity tests to compare against. The bending beam test case results from that PR does change with the fix here. 

Hopefully this resolves the elasticity + mesh deformation issue entirely!

gassmoeller:
Let's get this merged so it does not block the other PRs.

PR #4716, opened 2022-05-20
Add infill density benchmark

danieldouglas92:

### For all pull requests:


jdannberg (commented):
These are the comments I made last night (so they are for an older version). I am submitting that review for now and then I'll have a look at the new version.

benchmarks/infill_density/infill_ascii_data.cc
```diff
@@ -0,0 +1,253 @@
...
+{
+  namespace BoundaryTraction
+  {
+    ASPECT_REGISTER_BOUNDARY_TRACTION_MODEL(InfillAsciiData,
+                                            "infill ascii data",
+                                            "Implementation of a model in which the boundary "
+                                            "traction is derived from files containing pressure data "
+                                            "in ascii format. The pressure given in the data file is "
+                                            "applied as traction normal to the surface of a given boundary, "
+                                            "pointing inwards. Note the required format of the "
```
jdannberg, line 227: This looks very much like a description of a general ascii data boundary traction model. Can you add more details on what this specific mode does?

benchmarks/infill_density/infill_ascii_data.cc
```diff
@@ -0,0 +1,253 @@
...
+      const double gravity_norm = this->get_gravity_model().gravity_vector(position).norm();
+      \
+      const double elevation = this->get_geometry_model().height_above_reference_surface(position);
+
+      // Check if the load at the current point has a height which exceeds the
+      // 'infill height'. This is useful for more complicated topographic loads
+      // when its harder to define the precise location of a volcanic edifice,
+      // for example. If the load does exceed the rock_infill_height, the infill
+      // material has the density of rock, otherwise it has sediment density.
+      if (load >= rock_infill_height*gravity_norm*rock_density)
```
jdannberg, line 146: I am trying to understand this but I am not quite sure I do. Can you explain a little bit more the reasoning behind how the load from the ascii file is modified and what the assumptions are? Here is what I think you do: the idea is that you know the actual height of the volcano (but not the actual load), so you have to modify the load based on what the actual topography in the model is? If it's negative, you increase the load because you would have infill in the depression the load creates. And if it's positive, you decrease the load because some of it is already accounted for by the positive surface topography?

But I don't really understand the infill_height part. Why does this depend on the load, and not on the elevation?
Why does it use the `rock_density` rather than `sediment_density`? In the current file, it is zero, so for the benchmark this does not make a difference, but I imagine for an application model is does. So it would be good to explain this (and the bast place might be the class description, or the parameter descriptions).

jdannberg (commented):
This is great, thank you for contributing this! If you have time, it would also be great if you could add a short description in the manual, but if not, that's fine as well. I mostly have a few suggestions for documentation, but otherwise this looks great!

benchmarks/infill_density/infill_ascii.prm
```diff
@@ -0,0 +1,199 @@
...
+# Prescribe a fixed vertical traction on the top boundary
+subsection Boundary traction model
+  set Prescribed traction boundary indicators = top y: infill ascii data, right: initial lithostatic pressure
+end
+
+subsection Boundary traction model
+  subsection Ascii data model
+    set Data directory = /Users/danieldouglas/software/aspect/cookbooks/infill_density/data/
+    set First data file model time = 0
+    set Data file name = box_2d_%s.%d.txt
```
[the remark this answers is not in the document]
danieldouglas92: Yes I did!

benchmarks/infill_density/infill_ascii_data.cc
danieldouglas92: I updated the documentation but maybe it would be best if I also included another .png showing how this works? I'm finding it hard to explain in just text alone.

benchmarks/infill_density/infill_ascii.prm
```diff
@@ -0,0 +1,199 @@
...
+
+subsection Mesh deformation
+  set Mesh deformation boundary indicators = top: free surface
+  set Additional tangential mesh velocity boundary indicators = left, right
+  subsection Free surface
+    set Surface velocity projection = vertical
+  end
+end
+
+# Velocity boundary conditions
```
naliboff, line 87: Can you add a comment here (or in the section below) about why the left boundary is free-slip and the right is open? I assume this helps remove artifacts from return flow induced by the surface loading?

benchmarks/infill_density/infill_ascii.prm
```diff
@@ -0,0 +1,199 @@
...
+  subsection Function
+    set Function expression = 293
+  end
+end
+
+# Material model
+subsection Material model
+  set Model name = viscoelastic
+  subsection Viscoelastic
+  # First 3 Parameters Dictate ve_stresses, 4th the mantle, 5th the Lithosphere
```
naliboff, line 154: Move this documentation down one line (suggestion below)

benchmarks/infill_density/infill_ascii.prm
```diff
@@ -0,0 +1,199 @@
...
+    set Densities                   = 3300
+    set Viscosities                 = 1.e21, 1.e21, 1.e21, 1.e21, 1.e26
```
naliboff, line 156:
```suggestion
    # The first three values are assigned to the compositional fields representing viscoelastic 
    # stresses, but significantly these values are not used in the viscosity calculation (i.e., only
    # values representing lithologies are used). Respectively, the fourth and fifth values are 
    # assigned to the compositional fields representing the mantle and lithosphere.
    set Viscosities                 = 1.e21, 1.e21, 1.e21, 1.e21, 1.e26
```

benchmarks/infill_density/infill_ascii.prm
```diff
@@ -0,0 +1,199 @@
...
+subsection Boundary traction model
+  subsection Ascii data model
+    set Data directory = $ASPECT_SOURCE_DIR/benchmarks/infill_density/data/
+    set First data file model time = 0
+    set Data file name = box_2d_%s.%d.txt
+    set Data file time step = 750
+  end
+  subsection Infill ascii
+    set Rock density = 2000
+    set Sediment density = 2000
```
naliboff, line 106: Out of curiosity, why does the GMT benchmark assume the rock and sediment density to be equal?

benchmarks/infill_density/infill_ascii_data.cc
```diff
@@ -0,0 +1,262 @@
+/*
+  Copyright (C) 2011 - 2019 by the authors of the ASPECT code.
```
naliboff, line 2:
```suggestion
  Copyright (C) 2022 by the authors of the ASPECT code.
```

benchmarks/infill_density/infill_ascii_data.cc
naliboff: The figure definitely helps. I don't have any suggestions to further improve the documentation above, and think it is in good shape.

benchmarks/infill_density/infill_ascii_data.cc
```diff
@@ -0,0 +1,262 @@
...
+    {
+      prm.enter_subsection("Boundary traction model");
+      {
+        Utilities::AsciiDataBoundary<dim>::declare_parameters(prm,
+                                                              "$ASPECT_SOURCE_DIR/tests/infill_density/",
+                                                              "triangle_load.txt");
+        prm.enter_subsection("Infill ascii");
+        {
+          prm.declare_entry ("Rock density", "2800",
+                             Patterns::List(Patterns::Double(0.)),
```
naliboff, line 186: Could this and the patterns below just be `Patterns::Double (0.)`? In other words, the plugin is currently designed to only use one value for each and they are declared as doubles (rather than vectors of doubles).

naliboff (changes requested):
@danieldouglas92 - Nice work, this is a really nice contribution! I only have minor questions and comments, and overall the PR is in great shape.

benchmarks/infill_density/infill_ascii_data.cc
danieldouglas92: Ah yes good catch!

benchmarks/infill_density/infill_ascii.prm
```diff
@@ -0,0 +1,199 @@
...
+# Formulation classification
+subsection Formulation
+  set Enable elasticity = true
+end
+
+subsection Mesh deformation
+  set Mesh deformation boundary indicators = top: free surface
+  set Additional tangential mesh velocity boundary indicators = left, right
+  subsection Free surface
+    set Surface velocity projection = vertical
```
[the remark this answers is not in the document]
danieldouglas92: This actually was supposed to be set to normal but it does not seem to influence the results in any meaningful way

benchmarks/infill_density/infill_ascii.prm
danieldouglas92: I honestly dont 100% understand why this is the case, but apparently the fourier solution is only exact with a constant density contrast between the load and the infill. If the infill varies then the fourier solution is still good but it is no longer 'exact' and now has to be approximated with a scaling prefactor.

benchmarks/infill_density/infill_ascii.prm
```diff
@@ -0,0 +1,199 @@
...
+    set Function expression                    = if(y>=1050e3, 3, 0)
+  end
+end
+
+# Element types
+subsection Discretization
+  set Composition polynomial degree           = 2
+  set Stokes velocity polynomial degree       = 2
+  set Temperature polynomial degree           = 1
+  set Use locally conservative discretization = true
```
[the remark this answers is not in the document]
danieldouglas92: I remember it increasing the accuracy yea, and my assumptions for why that was the case is what you added as a comment further down about better tracking for the lithosphere (since the analytic solution depends on lithosphere thickness).

jdannberg (commented):
I think there were some comments from my first review you did not address (maybe because they were listed as outdated), I copied them here again. Otherwise this looks good!

I haven't looked at the manual text yet (since it does not build yet) and I can do this next after you rebase. 

The only other point I remembered: We usually have a test case for each benchmark. You don't need to copy the .prm file, but you can instead include your benchmark prm file in a test and then set the end time to zero (you can see an example of that in sol_kz_2_project_q1_only_visc.prm). The point is to make sure that even if we make future changes in ASPECT, the tester will still execute the files you added and we will notice if it breaks.

benchmarks/infill_density/data/box_2d_top.0.txt
```diff
@@ -0,0 +1,1002 @@
+# POINTS: 1001
```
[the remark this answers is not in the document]
danieldouglas92: They are different! The load linearly increases with time, I found this helps stabilize the elastic response instead of dumping the entire load on all at once.

jdannberg (approved):
Great, looks good! Thank you for contributing this!
This is ready to merge once the tester is finished.

jdannberg:
@naliboff You still have to approve before we can merge this.

benchmarks/infill_density/infill_ascii.prm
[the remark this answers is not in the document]
naliboff: This parameter is actually independent from the DG composition formulation, but let's check it's affect later (probably has little to no influence).

naliboff (approved):
@danieldouglas92 - Nice work, ready to merge!

PR #4717, opened 2022-05-20
Add functionality to read compressed gzip files.

gassmoeller:
As the changelog says: 
The `read_and_distribute_file_content` utility function can now also read
binary files compressed with gzip, which allows for example the StructuredData
class (and therefore all ascii data plugins) to read gzip compressed data.
Compressed files are detected based on their file ending '.gz'. I.e. if you want
to read a compressed file, make sure it's name ends in '.gz', and if you want to
read a plain text file, make sure its name does not end in '.gz'. Compressed
files can save around 75% to 80% of disk space for typical ascii data files.


source/utilities.cc
```diff
@@ -1363,21 +1371,26 @@ namespace aspect
...
-
               // Read data from disk
               std::stringstream datastream;
-              filestream >> datastream.rdbuf();
 
-              if (!filestream.eof())
+              try
+                {
+                  boost::iostreams::filtering_istreambuf in;
+                  if (filename.find(".gz") == filename.size()-3)
```
tjhei, line 1380: What if my file is called ``test.gzinga.txt.gz``?

doc/modules/changes/20220520_gassmoeller
```diff
@@ -0,0 +1,9 @@
+New: The `read_and_distribute_file_content` utility function can now also read
+binary files compressed with gzip, which allows for example the StructuredData
+class (and therefore all ascii data plugins) to read gzip compressed data.
+Compressed files are detected based on their file ending '.gz'. I.e. if you want
+to read a compressed file, make sure it's name ends in '.gz', and if you want to
```
bangerth, line 5:
```suggestion
to read a compressed file, make sure its name ends in '.gz', and if you want to
```

source/utilities.cc
```diff
@@ -1351,7 +1355,11 @@ namespace aspect
             }
           else
             {
-              std::ifstream filestream(filename.c_str());
+              std::ifstream filestream;
+              if (filename.find(".gz") == filename.size()-3)
```
bangerth, line 1359: Maybe easier to read, at least for those of us with magic regex skills :-)
```suggestion
              if (std::regex_match (filename, std::regex("\\.gz$"))
```

source/utilities.cc
bangerth, line 1380: same here. or maybe introduce a local variable that stores whether you're dealing with a compressed file.

bangerth (commented):
Very nice!

bangerth (approved):

tjhei (approved):

Issue #4339, opened 2021-09-02
Remove reference viscosity parameter from material models?


naliboff:
A follow-up question regarding PR #3681, which added a function to automatically compute a reference viscosity at each time step. 

After discussion with @anne-glerum, we are wondering if the reference viscosity parameter should be removed from each material model. @gassmoeller - is there a motivation to keep these parameters in each material for the time being?

tjhei:
I don't think it is used anymore, so we should remove it.

gassmoeller:
Agreed this can be removed. When doing this we will have to check if there are individual plugins that still use it, so the process would be:
1. Only remove it from the interface, but keep the implementation in individual plugins. This will fail for all places that still use the reference viscosity from outside the functions. I am think of composite material models or postprocessors that may still call the function. During this step we need to test benchmarks and cookbooks as well.
2. Remove it from all plugins. This step is optional, if there are some plugins that make use of the reference viscosity function for internal purposes, we can also keep the function in those plugins.

jdannberg:
This was addressed by #4722 and #4733.

PR #4722, opened 2022-05-20
remove reference viscosity from the material model interface

jdannberg:
This PR removes the reference viscosity from the material model interface since we do not require it any more for the pressure scaling (or anywhere else in ASPECT). 

### For all pull requests:

### DIFF
--- a/doc/modules/changes/20220520_jdannberg
+++ b/doc/modules/changes/20220520_jdannberg
@@ -1,0 +1,8 @@
+Changed: The reference viscosity function is no longer part
+of the material model interface (because we do not require 
+it anymore for the pressure scaling or anywhere else). The
+One material model affected by this is the depth-dependent
+model, which now has its own reference viscosity parameter
+rather than use the reference viscosity of the base model.
+<br>
+(Juliane Dannberg, 2022/05/20)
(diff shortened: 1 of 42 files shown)


doc/modules/changes/20220520_jdannberg
```diff
@@ -0,0 +1,8 @@
+Changed: The reference viscosity function is no longer part
+of the material model interface (because we do not require 
+it anymore for the pressure scaling or anywhere else). The
+One material model affected by this is the depth-dependent
```
bangerth, line 4:
```suggestion
one material model affected by this is the depth-dependent
```

include/aspect/material_model/composition_reaction.h
```diff
@@ -74,7 +74,7 @@ namespace aspect
          * @name Reference quantities
          * @{
          */
-        double reference_viscosity () const override;
+        double reference_viscosity () const;
```
bangerth, line 77: Is this still called somewhere?

bangerth (commented):
Yay. Long time to get rid of this.

In all of these material models, you leave these functions in place, just no longer `virtual`. But who calls them? Are they still really needed?

jdannberg:
Most of these are probably no longer needed. I asked @gassmoeller and he suggested to remove the ones that are not needed in a follow-up PR so that this patch is shorter.

jdannberg:
It looks like there already is an issue (#4339) that says to do this in two steps. So this PR addresses part one of the issue.

jdannberg:
I have also addressed your one comment.

bangerth:
Open an issue so we don't forget, fix the typo, and let's get this merged.

gassmoeller (approved):

PR #4727, opened 2022-05-20
Material model based on tomography data or table properties

alarshi:
This PR adds the material model we (@gassmoeller , @jdannberg ) have been using for our global mantle flow models. All the relevant input files and the prm file we want to use for a cookbook are in relevant folders. 

### For all pull requests:


bangerth:
Will review after dinner. If you want to make sure that the indentation is correct before I get to it, that'll be fine too.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+        /**
+         * Return whether the model is compressible or not.
+         */
+        virtual bool is_compressible () const;
+
+        /**
+         * This function is not defined because we use get_reference_viscosity() function
+         * instead but kept here according to material model interface.
+         */
+        virtual double reference_viscosity () const;
```
bangerth, line 112: I think this is actually the function that can be removed following #4733.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+         */
+        Utilities::AsciiDataProfile<dim> thermal_expansivity_profile;
+        unsigned int thermal_expansivity_column_index;
+
+        /**
+         * A reference profile for temperatura scaling. Values are from Steinberger and
+         * Calderwood, 2006.
+         */
+        Utilities::AsciiDataProfile<dim> dT_vs_depth_profile;
+        /**
```
bangerth, line 227:
```suggestion
        Utilities::AsciiDataProfile<dim> dT_vs_depth_profile;
        
        /**
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+        double max_specific_heat;
+        double min_thermal_expansivity;
+        double max_thermal_expansivity;
+        unsigned int max_latent_heat_substeps;
+        double min_grain_size;
+
+        double diffusion_viscosity (const double      temperature,
+                                    const double      pressure,
+                                    const std::vector<double>    &compositional_fields,
+                                    const SymmetricTensor<2,dim> &,
```
bangerth, line 269: Can you give this argument a name? Or do you not actually use it in the implementation? In the latter case, just remove the argument altogether.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+                                    const Point<dim> &position) const;
+
+        /**
+         * Compute the equilibrium grain size for a given temperature and
+         * pressure.
+         * This computation is based on the rate of grain size growth
+         * (Ostwald ripening) or reduction(due to dynamic recrystallization
```
bangerth, line 276:
```suggestion
         * (Ostwald ripening) or reduction (due to dynamic recrystallization
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+         * with a dislocation viscosity calculated for the whole strain rate
+         * unless a guess for the viscosity is provided, which can reduce the
+         * number of iterations significantly.
+         */
+        double dislocation_viscosity (const double      temperature,
+                                      const double      pressure,
+                                      const std::vector<double>    &compositional_fields,
+                                      const SymmetricTensor<2,dim> &strain_rate,
+                                      const Point<dim> &position,
+                                      const double      viscosity_guess = 0) const;
```
bangerth, line 308: Maybe align the names of arguments like for other functions.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+         * Returns the s-wave velocity as calculated by HeFESTo.
+         */
+        double seismic_Vs (const double      temperature,
+                           const double      pressure,
+                           const std::vector<double> &compositional_fields,
+                           const Point<dim> &position) const;
+
+        /**
+         * Function that defines the phase transition interface
+         * (0 above, 1 below the phase transition).This is done
```
bangerth, line 352:
```suggestion
         * (0 above, 1 below the phase transition). This is done
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+
+        /**
+         * Function that returns the phase for a given
+         * position, temperature, pressure and compositional
+         * field index.
+         */
+        unsigned int
+        get_phase_index (const Point<dim> &position,
+                         const double temperature,
+                         const double pressure) const;
```
bangerth, line 369: Is the documentation for these two functions switched? The documentation for the first function describes a sum over all phases but the function takes the `phase` as an argument. The second function does not take such an argument, so one may assume that it sums over all phases?

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+
+
+        /**
+         * list of depth, width and Clapeyron slopes for the different phase
```
bangerth, line 373:
```suggestion
         * Lists of depth, width, and Clapeyron slopes for the different phase
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+         * transitions and in which phase they occur
```
bangerth, line 374:
```suggestion
         * transitions and in which phase they occur.
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+        std::vector<double> transition_temperatures;
+        std::vector<double> transition_slopes;
+        std::vector<double> transition_widths;
+
+
+        /**
+         * The following variables are properties of the material files
+         * we read in.
+         */
+        std::string datadirectory;
```
bangerth, line 386: maybe call it `data_directory` to be consistent with our usual spelling for variable names

cookbooks/equilibrium_grain_size/plugins/reference_profile.h
```diff
@@ -0,0 +1,207 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_adiabatic_conditions_reference_profile_h
```
bangerth, line 22: You probably want to name this preprocessor symbol so that it uses the correct directory name -- maybe `_aspect_cookbooks_equilibrium_grain_size_reference_profile_h`?

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file doc/COPYING.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_model_equilibrium_grain_size_h
+#define _aspect_model_equilibrium_grain_size_h
```
bangerth, line 23: You probably want to name this preprocessor symbol so that it uses the correct directory name -- maybe `_aspect_cookbooks_equilibrium_grain_size_h`?

cookbooks/equilibrium_grain_size/plugins/reference_profile.h
```diff
@@ -0,0 +1,207 @@
...
+         */
+        bool initialized;
+
+        /**
+         * Number of points at which we compute the adiabatic values.
+         */
+        unsigned int n_points;
+
+        /**
+         * Vectors of values of temperatures and pressures on a transect into
```
bangerth, line 136:
```suggestion
         * Vectors of values of temperatures, pressures, and densities on a transect into
```

cookbooks/equilibrium_grain_size/plugins/reference_profile.h
```diff
@@ -0,0 +1,207 @@
...
+     * and the composition is either given by the initial
+     * composition at reference points or computed
+     * as a reference depth-function.
+     * All material parameters are computed by the
+     * material model plugin. The surface conditions are
+     * either constant or changing over time as prescribed
+     * by an user-provided function.
+     */
+    template <int dim>
+    class ReferenceProfile : public Interface<dim>
```
bangerth, line 50: Is this a copy of a similar class from elsewhere? If so, can you describe what is different? I wonder whether we could find a way to achieve the same goal with less code than results from copying a class and its implementation.

bangerth (commented):
Here a first review:
* I've taken a look at the data files and the `.h` files. The `.cc` files have to wait for tomorrow.

* Is `cookbooks/equilibrium_grain_size/input_data/LLNL_model_cropped_cratons_faults.txt` read by the ASCII-Data reader? Would it make sense to store it in the repository as a .gz file, to be read by @gassmoeller 's recent patch #4717? (If so, one may consider using compression for all files, for consistency.)

cookbooks/equilibrium_grain_size/input_data/1D_reference_profiles/gypsum_ref.txt
```diff
@@ -0,0 +1,59 @@
+# Header for the PREM reference profile data
```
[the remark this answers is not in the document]
alarshi: Yes, it is. Since we are not using this file, I will remove it. Thank you.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+
+        /**
+         * Compute the reference viscosity against which the viscosities in the model
+         * are scaled. The reference viscosity is from a depth-dependent ascii profile and
+         * the function returns as the second parameter in the pair the support point in
+         * that profile that has the next smaller depth value than the input parameter
+         * @p depth.
+         */
+        std::pair<double, unsigned int>
+        get_reference_viscosity (const double depth) const;
```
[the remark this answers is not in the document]
alarshi: We are using this function, but I removed `reference_viscosity()` based on your suggestion below :)

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+
+        /**
+         * Return the depth of the base of the uppermost mantle. Below that depth,
+         * material properties are based on seismic tomography. Above that depth, material
+         * properties are computed based on the model of Tutu et al., 2018.
+         */
+        double
+        get_uppermost_mantle_thickness () const;
```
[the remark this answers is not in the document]
alarshi: It extends to the surface, I will modify the name. Thank you!

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+     */
+    template <int dim>
+    class MaterialTypeAdditionalOutputs : public NamedAdditionalMaterialOutputs<dim>
+    {
+      public:
+        MaterialTypeAdditionalOutputs(const unsigned int n_points)
+          : NamedAdditionalMaterialOutputs<dim>(std::vector<std::string>(1, "material_type"),
+                                                n_points)
+        {}
+    };
```
bangerth, line 69: These two classes, along with the `FunctorDepthAverageUnscaledViscosity` one below, are purely local to this file. It might be nice to surround their definition with `namespace { ... };`, i.e., put them into an "anonymous namespace".

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
```diff
@@ -0,0 +1,497 @@
...
+
+        /**
+         * Compute the scaling factors for each depth layer such that the laterally
+         * averaged viscosiy in that layer is the same as the reference vicosity.
+         */
+        double
+        compute_viscosity_scaling (const double depth) const;
```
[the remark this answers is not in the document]
alarshi: We are using a reference viscosity profile, based on the fit to the observed Geoid, against which our laterally averaged viscosities are scaled to. Do you suggest another name for this function?

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+        this->update();
+      });
+
+      this->get_signals().post_advection_solver.connect([&](const SimulatorAccess<dim> &,
+                                                            const unsigned int ,
+                                                            const unsigned int ,
+                                                            const SolverControl &)
+      {
+        this->update();
+      });
```
bangerth, line 163: Can you explain in a comment why you want the `update()` function to run at this point (i.e. post-stokes- and post-advection-solve)?

Maybe more generally, we've generally tried to follow a workflow where every time step is computed based on the previous timesteps' results, rather than relying on the order in which the different parts of the current time step are solved. As a consequence, I think it would be nice *not* to update the viscosity between the parts of the current time step, unless there is a good reason for it.

bangerth: Separately, you should probably capture the return object of the `connect()` call and store it in a `boost::signals2::connection` variable so that the connection can be disconnected once the material model object is destroyed.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+
+
+    template <int dim>
+    void
+    EquilibriumGrainSize<dim>::update()
+    {
+      if (use_depth_dependent_viscosity)
+        {
+          std::vector<std::unique_ptr<internal::FunctorBase<dim>>> lateral_averaging_properties;
+          lateral_averaging_properties.push_back(std::make_unique<internal::FunctorDepthAverageUnscaledViscosity<dim>>());
```
bangerth, line 193:
```suggestion
          lateral_averaging_properties.emplace_back(std::make_unique<internal::FunctorDepthAverageUnscaledViscosity<dim>>());
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+
+          std::vector<std::vector<double>> averages =
+            this->get_lateral_averaging().compute_lateral_averages(reference_viscosity_coordinates,
+                                                                   lateral_averaging_properties);
+
+          average_viscosity_profile.swap(averages[0]);
```
bangerth, line 199:
```suggestion
          average_viscosity_profile.swap = std::move(averages[0]);
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
bangerth, line 193: Or maybe even easier, just create a one-element array:
```suggestion
          const std::vector<std::unique_ptr<internal::FunctorBase<dim>>> lateral_averaging_properties
          = { std::make_unique<internal::FunctorDepthAverageUnscaledViscosity<dim>>() };
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+          for (const auto &lateral_viscosity_average: average_viscosity_profile)
+            AssertThrow(numbers::is_finite(lateral_viscosity_average),
+                        ExcMessage("In computing depth averages, there is at"
+                                   " least one depth band that does not have"
+                                   " any quadrature points in it."
+                                   " Consider reducing number of depth layers"
+                                   " for averaging."));
+        }
+
+      initialized = true;
```
bangerth, line 210: Shouldn't this be part of the `initialize()` function?

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+    template <int dim>
+    std::pair<double, unsigned int>
+    EquilibriumGrainSize<dim>::get_reference_viscosity (const double depth) const
+    {
+      // Make maximal depth slightly larger to ensure depth < maximal_depth
+      const double maximal_depth = this->get_geometry_model().maximal_depth() *
+                                   (1.0+std::numeric_limits<double>::epsilon());
+      (void) maximal_depth;
+
+      Assert(depth < maximal_depth, ExcInternalError());
```
bangerth, line 224: Would you also want the corresponding minimum check?
```suggestion
      Assert(depth < maximal_depth, ExcInternalError());
      Assert(depth > -maximal_depth*std::numeric_limits<double>::epsilon(), ExcInternalError());
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+        }
+      else
+        {
+          depth_index = std::distance(reference_viscosity_coordinates.begin(),
+                                      std::lower_bound(reference_viscosity_coordinates.begin(),
+                                                       reference_viscosity_coordinates.end(),
+                                                       depth));
+        }
+      if (depth_index > 0)
+        depth_index -= 1;
```
bangerth, line 243:
```suggestion
        --depth_index;
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+      // When evaluating reference viscosity, evaluate at the next lower depth that is stored
+      // in the reference profile instead of the actual depth. This makes the profile piecewise
+      // constant. This will be specific to the viscosity profile used (and ignore the entry with
+      // the largest depth in the profile).
+      double reference_viscosity = reference_viscosity_profile->compute_viscosity(reference_viscosity_coordinates.at(depth_index));
+
+      // This parameter is only because we change the asthenosphere viscosity in our models.
+      // By default, it is set to the value in the reference profile.
+      if (depth_index == 1)
+        reference_viscosity = asthenosphere_viscosity;
```
bangerth, line 254: What does "this parameter" refer to?

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+
+      return std::make_pair (reference_viscosity, depth_index);
+
+    }
```
bangerth, line 258:
```suggestion
      return std::make_pair (reference_viscosity, depth_index);
    }
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+
+
+
+    template <int dim>
+    double
+    EquilibriumGrainSize<dim>:: reference_viscosity () const
+    {
+      // We do not use this function so just return a reasonable value.
+      return 2*min_eta*max_eta/(min_eta + max_eta);
+    }
```
bangerth, line 268: I think you can remove this function post-#4733.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+              // because the diffusion viscosity depends on the grain size itself, and we need it to compute the dislocation strain rate,
+              // we have to iterate in the computation of the equilibrium grain size
+              while ((std::abs((grain_size-old_grain_size) / grain_size) > dislocation_viscosity_iteration_threshold)
+                     && (j < dislocation_viscosity_iteration_number))
+                {
+                  composition[grain_size_index] = std::max(min_grain_size, grain_size);
+                  const double diff_viscosity = diffusion_viscosity(in.temperature[i], pressure, composition, in.strain_rate[i], in.position[i]);
+                  const double disl_viscosity = dislocation_viscosity(in.temperature[i], pressure, composition, in.strain_rate[i], in.position[i], diff_viscosity);
+
+                  if (disl_viscosities_out != NULL)
```
bangerth, line 394:
```suggestion
                  if (disl_viscosities_out != nullptr)
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+                else if (c == fault_index)
+                  prescribed_field_out->prescribed_field_outputs[i][c] = in.composition[i][fault_index];
+                else
+                  prescribed_field_out->prescribed_field_outputs[i][c] = 0.;
+              }
+
+          if (this->get_nonlinear_iteration() == 0 && use_depth_dependent_viscosity)
+            out.viscosities[i] = get_reference_viscosity(depth).first;
+        }
+      return;
```
bangerth, line 507:
```suggestion
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+          dis_viscosity_old = dis_viscosity;
+          dis_viscosity = dislocation_viscosity_fixed_strain_rate(temperature,
+                                                                  pressure,
+                                                                  dislocation_strain_rate,
+                                                                  position);
+          ++i;
+
+          Assert(dis_viscosity > 0.0,
+                 ExcMessage("Encountered negative dislocation viscosity in iteration " + std::to_string(i) +
+                            ". Dislocation viscosity is: " + std::to_string(dis_viscosity)));
```
bangerth, line 693: You first increment `i` and then use it in the generation of the error message. Is that on purpose? Wouldn't you want the error message to reflect the `i` that was used to compute the quantity you are complaining about in the error message?

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+
+      double vp = 0.0;
+      if (n_material_data == 1)
+        vp = material_lookup[0]->seismic_Vp(temperature,pressure);
+      else
+        {
+          for (unsigned i = 0; i < n_material_data; i++)
+            vp += compositional_fields[i] * material_lookup[i]->seismic_Vp(temperature,pressure);
+        }
+      return vp;
```
bangerth, line 776: It might be easier to read this way:
```suggestion
      if (n_material_data == 1)
        return material_lookup[0]->seismic_Vp(temperature,pressure);
      else
        {
          double vp = 0.0;
          for (unsigned i = 0; i < n_material_data; i++)
            vp += compositional_fields[i] * material_lookup[i]->seismic_Vp(temperature,pressure);
          return vp;
        }
```

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+
+      double vs = 0.0;
+      if (n_material_data == 1)
+        vs = material_lookup[0]->seismic_Vs(temperature,pressure);
+      else
+        {
+          for (unsigned i = 0; i < n_material_data; i++)
+            vs += compositional_fields[i] * material_lookup[i]->seismic_Vs(temperature,pressure);
+        }
+      return vs;
```
bangerth, line 800: Same here

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+        {
+          rho = material_lookup[0]->density(temperature,pressure);
+        }
+      else
+        {
+          for (unsigned i = 0; i < n_material_data; i++)
+            rho += compositional_fields[i] * material_lookup[i]->density(temperature,pressure);
+        }
+
+      return rho;
```
bangerth, line 824: and here

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+    {
+      double alpha = 0.0;
+      if (n_material_data == 1)
+        alpha = material_lookup[0]->thermal_expansivity(temperature,pressure);
+      else
+        {
+          for (unsigned i = 0; i < n_material_data; i++)
+            alpha += compositional_fields[i] * material_lookup[i]->thermal_expansivity(temperature,pressure);
+        }
+      return alpha;
```
bangerth, line 878: and here

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+    void
+    EquilibriumGrainSize<dim>::declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Material model");
+      {
+        prm.enter_subsection("Equilibrium grain size model");
+        {
+          prm.declare_entry ("Reference density", "3300",
+                             Patterns::Double (0),
+                             "The reference density $\\rho_0$. Units: $kg/m^3$.");
```
bangerth, line 1081: Here and for all of the parameters below: We try to use semantic mark-up for units. For example, this could look like this:
```
Units: \\si{\\meter}.
```
Take a look at some of the other files

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+          Utilities::AsciiDataBase<dim>::declare_parameters(prm, "../../input_data/", "rho_vs_scaling.txt", "Density velocity scaling");
+
+          // Depth-dependent density scaling parameters
+          Utilities::AsciiDataBase<dim>::declare_parameters(prm, "../../input_data/", "dT_vs_scaling.txt", "Temperature velocity scaling");
+
+          // Depth-dependent thermal expansivity parameters
+          Utilities::AsciiDataBase<dim>::declare_parameters(prm, "../../input_data/", "thermal_expansivity_steinberger_calderwood.txt", "Thermal expansivity profile");
+
+          // Crustal boundary depths parameters
+          Utilities::AsciiDataBoundary<dim>::declare_parameters(prm,  "../../input_data/", "crustal_structure.txt", "Crustal depths");
```
bangerth, line 1390: Break these long lines into shorter ones

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+                                   "size evolution in terms of the strain rate, "
+                                   "temperature, phase transitions, and the creep regime. "
+                                   "This material model only works if a compositional field "
+                                   "named 'grain_size' is present. "
+                                   "In the diffusion creep regime, the viscosity depends "
+                                   "on this grain size field. "
+                                   "We use the grain size evolution laws described in Behn "
+                                   "et al., 2009. Implications of grain size evolution on the "
+                                   "seismic structure of the oceanic upper mantle, "
+                                   "Earth Planet. Sci. Letters, 282, 178–189. "
```
bangerth, line 1725: Is this a reference that is already in `references.bib`? (Of which there currently are two copies, and both should get it if not in there yet.) If so, just use `\\cite{...}` to reference it.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
```diff
@@ -0,0 +1,1750 @@
...
+                                   "Other material parameters are either prescribed similar "
+                                   "to the 'simple' material model, or read from data files "
+                                   "that were generated by the Perplex or Hefesto software. "
+                                   "This material model "
+                                   "is described in more detail in Dannberg, J., Z. Eilon, "
+                                   "U. Faul, R. Gassmoeller, P. Moulik, and R. Myhill (2017), "
```
bangerth, line 1731: same here

bangerth (commented):
Here's the next batch:

* This being a cookbook, there should be a corresponding `.md` or `.tex` file. Did I miss that?
* I don't know enough about the material model you use, so haven't checked the implementation of it.

cookbooks/equilibrium_grain_size/plugins/reference_profile.h
alarshi: Yes, this is very similar to the `adiabatic_conditions/compute_profile.h`. The only difference is that we use a reference density from prem model below a certain depth defined by `uppermost_mantle_thickness` parameter. I tried to remove the code and use `compute_profile` as a base model, but I was not very successful given the required functions from the interface class and the fact that computation of density required parameters that we needed to define. It might be easier to add the changes from our adiabatic profile in the current `compute_profile` file, but I am not sure if it will be useful for others.
Maybe I can ask you more about this later today and find a solution.

alarshi: @gassmoeller suggested a way to remove this class and it works with the existing implementation of `compute_profile` and the modifications to the material model. Thank you for pointing that out!

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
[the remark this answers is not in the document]
alarshi: It seems like with the new changes, we would not need this function. I will remove it. Thank you.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
alarshi: It is the `asthenosphere_viscosity`. I will remove the comment, it is not relevant anymore.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
alarshi: I guess it is because our value of i is 0 for the first iteration and we would want to say "iteration 1" in our message.

cookbooks/equilibrium_grain_size/models/gypsum_model/grain-size-test.prm
```diff
@@ -0,0 +1,185 @@
+# This is a copy of the grain size strain test. The test imposes simple shear
+# across a 100 km square domain with a rheology derived from experiments on
+# olivine. The model domain is isobaric and initially isothermal.
```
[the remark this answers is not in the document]
jdannberg: I think this would be a great file to make into a test (it's definitely not an application model).

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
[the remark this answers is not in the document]
jdannberg: I think that name makes sense. In the depth-average material model, which is the only one that actually still has a reference viscosity, it is used for a similar purpose.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.h
jdannberg: So these functions seem to exist in identical form in the grain_size material model (with the same documentation). 
I think the documentation is not switched, just outdates. But if we change it here, it will still be wrong in the grain size model. Maybe it would make sense to open and issue to unify these functions later?

The `get_phase_index` computes the index of the stable phase for a given position, temperature and pressure. 

The `phase_function` computes the value of the phase function (which is the fraction of material that has already undergone a certain phase transition) for an individual transition. 
(So there's no summing up as part of the function.)

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
jdannberg: The update() function updates the profile that stores the laterally averaged viscosity. This is needed to compute the viscosity in the material model (since it is rescaled so that the lateral average matches the reference profile). Since the viscosity depends on both temperature and velocity, we want to update the lateral average profile after each temperature and Stokes solve. 

(@alarshi you can add this as a comment)

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
jdannberg: I didn't see this used anywhere in the code (we instead use `average_viscosity_profile.size() != 0` to check if the lateral average profile has been computed yet. So I think you can just remove this.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
jdannberg: I don't remember why we are doing this check, but I think that suggestion makes sense.

cookbooks/equilibrium_grain_size/plugins/equilibrium_grain_size.cc
jdannberg: I think that makes sense, but I would also be fine with switching it.

jdannberg:
@alarshi I commented on all the points that you did not resolve and that seemed complicated. I hope that is enough information for you to finish this PR. If not, let me know!

alarshi:
Something went wrong during the rebase. Closing this pull request to see if reopening fixes the github changes page.

PR #4728, opened 2022-05-21
Add instructions for simple figure in documentation quickref

gassmoeller:


tjhei (commented):
Why do we use a html version of an image above? Do you know if both options work for the pdf version?

gassmoeller:
I am not sure. I found this https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#images. It says both markdown and html versions work (for both plain images and figures with captions), but html version may not be copied into non-html documentations. I will work on getting the pdf version of the sphinx documentation to compile tomorrow, then we will see.

bangerth (commented):
This should be relatively uncontroversial, right?

bangerth:
I think what @gassmoeller is saying here is "Here is a different way to embed a figure". It doesn't say that that's what *should* be used. I think we should just merge this, except of course that it's got a merge conflict right now.

@gassmoeller Please rebase.

gassmoeller:
Rebase is done and testers are happy. Can I merge this?

PR #4740, opened 2022-05-21
[ci] disable building pdf manual

tjhei:
The pdf manual doesn't build anymore and maybe we don't need to run the
tester then.


gassmoeller (approved):
I guess that works too. Maybe then #4742 is not needed.

PR #4745, opened 2022-05-21
Parameters documentation in Sphinx

sac-bsa:
I created a Python script to read the parameters.json file output by ASPECT, and convert them to markdown for Sphinx documentation. It keeps the same reference name as the LaTeX used. The script can leave whitespace at the end of lines, so the indent script should also be run. I also put examples into `quickref.md` for reference as we convert.
### For all pull requests:


gassmoeller (approved):
I think this is awesome and exactly what we need. I will let other take a look at it too before merging. If you want to see a quick example how it works check this: https://aspect-documentation--4745.org.readthedocs.build/en/4745/quickref.html#parameter-file-referencing

tjhei:
This is awesome and saves me the effort to do this in deal.II! :+1:

I see that some of the latex quotes don't work, see https://aspect-documentation--4745.org.readthedocs.build/en/4745/parameters.html#parameter-name-allow-fixed-composition-on-outflow-boundaries

Finally, the page ends up being very big (chrome on mobile is not happy). How can we tell sphinx to split it up into multiple pages?

sac-bsa:
> Finally, the page ends up being very big (chrome on mobile is not happy). How can we tell sphinx to split it up into multiple pages?
I just split the `parameters.md` into 29 smaller markdown files (One made for every top-level parameter subsection e.g. Postprocess). It loads a lot faster on my laptop and phone this way.

contrib/utilities/jsontomarkdown.py
```diff
@@ -0,0 +1,115 @@
...
+                             ":::\n"
+                             "\n"
+                             ":::{toctree}\n"
+                             "---\n"
+                             "maxdepth: 2\n"
+                             "---\n"
+                             "global.md\n")
+    global_parameters = open("doc/sphinx/parameters/global.md", "w")
+    print("# Global \n\n", file=global_parameters)
+    print("## **Parameters in section** global \n\n", file=global_parameters)
```
bangerth, line 62: This is going to look funny. Why not this:
```suggestion
    print("## **Global parameters**\n\n", file=global_parameters)
```

Separately, is it necessary to mark up by hand that this should be printed as boldface? Should that be controlled by some style file?

contrib/utilities/jsontomarkdown.py
```diff
@@ -0,0 +1,115 @@
...
+            print("**Pattern:** " +  data[key]["pattern_description"] + " \n", file=output_file)
+            print("**Documentation:** " + data[key]["documentation"] + " \n", file=output_file)
+        elif "alias" in data[key]:
+            # This is an alias for a parameter
+            print("(parameters:" + path_str + ")=", file=output_file)
+            aliased_name = data[key]["alias"]
+            alias_path_str = "/".join(cur_path) + "/" + aliased_name.replace(" ", "_20") 
+            print("### __Parameter name__: " +  true_name, file=output_file)
+            print("**Alias:** [" + aliased_name + "](parameters:" + alias_path_str + ")\n", file=output_file)
+            print("**Deprecation Status:** " + data[key]["deprecation_status"] + "\n", file=output_file)
```
bangerth, line 31: This is the same code as for function below, right? Could you refactor this into a function of its own so we don't duplicate things?

bangerth (approved):
I've got a couple of comments that I would appreciate if you could take care of. But let's get this merged first, and if you don't mind, address the comments in a follow-up patch.

contrib/utilities/jsontomarkdown.py
tjhei: fixed in https://github.com/geodynamics/aspect/pull/4758

PR #4749, opened 2022-05-22
shorten an assert

jdannberg:

### DIFF
--- a/source/simulator/parameters.cc
+++ b/source/simulator/parameters.cc
@@ -1964,18 +1964,14 @@ namespace aspect
 
       if (std::find(compositional_field_methods.begin(), compositional_field_methods.end(), AdvectionFieldMethod::fem_darcy_field)
           != compositional_field_methods.end())
-        AssertThrow (std::find(names_of_compositional_fields.begin(), names_of_compositional_fields.end(), "porosity")
-                     != names_of_compositional_fields.end(),
-                     ExcMessage ("The Darcy advection field method only works if there is a compositional field named 'porosity'"));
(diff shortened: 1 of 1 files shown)


source/simulator/parameters.cc
```diff
@@ -1964,18 +1964,13 @@ namespace aspect
...
-           != compositional_field_methods.end()) && (std::find(names_of_compositional_fields.begin(), names_of_compositional_fields.end(), "porosity") != names_of_compositional_fields.end()) )
-        for (unsigned int i=0; i<n_compositional_fields; ++i)
-
-          {
-            AssertThrow ((((compositional_field_methods[i] != AdvectionFieldMethod::fem_darcy_field) &&
-                           (names_of_compositional_fields[i] == "porosity")) == false),
-                         ExcMessage ("When using the darcy field method, the porosity field must be advected with the darcy method."));
-          }
+        {
+          const unsigned int porosity_idx = std::find(names_of_compositional_fields.begin(), names_of_compositional_fields.end(), "porosity");
```
tjhei: find () returns an iterator not an int, so I don't see how this can work.

jdannberg: Yes. I think maybe I should have taken the whole day off rather than trying to work after coming back from the hike...

bangerth (approved):

bangerth:
So many times already I've wondered why it took the C++ standards committee till 2020 to introduce functions like `std::set::contains()` (https://en.cppreference.com/w/cpp/container/set/contains). I wished they had also done that for vectors...

PR #4755, opened 2022-05-22
Avoid latex markup in author list.

bangerth:
I'm assuming that we use HTML markup for non-ASCII characters?

/rebuild


gassmoeller (approved):
Looks good.

PR #4757, opened 2022-05-22
Docs: Update parameter references and TODOs

cmills1095:
A couple of bugs within the parameter documentation itself was preventing some references from being recognized, so I fixed those as well: though that issue will probably have to be addressed on a wider scale (the issue is the use of backticks instead of apostrophes or double quotes).


doc/sphinx/parameters/Formulation.md
```diff
@@ -41,7 +41,7 @@
   \item `custom': A custom selection of `Mass conservation' and `Temperature equation'.
 \end{enumerate}
 
-\note{Warning: The `custom' option is implemented for advanced users that want full control over the equations solved. It is possible to choose inconsistent formulations and no error checking is performed on the consistency of the resulting equations.}
+\note{Warning: The 'custom' option is implemented for advanced users that want full control over the equations solved. It is possible to choose inconsistent formulations and no error checking is performed on the consistency of the resulting equations.}
```
bangerth, line 44: That's the wrong place to fix these things -- this file is autogenerated from the source files. Whenever we update the `parameters.json` file, this change will be overwritten.

The solution is to replace these ticks by proper `&lsquo;` and `&rsquo;`.

bangerth: @sac-bsa Might be able to help you with that in the script that created these files. It should be a relatively straightforward regex replace thing.

bangerth: Oh, on second thought, this isn't generated from parameters.jason, is it?

cmills1095: I know its something that will have to be addressed in a more formal capacity in another pull request (Timo brought it up in another discussion) but for the purposes of this pull request I changed that one instance so that everything works correctly in the test build.

gassmoeller (approved):
Let's get this PR merged and address the problematic quotes properly in another PR.

bangerth:
Good to go, but apparently needs to pass the testers first :-(

PR #4760, opened 2022-05-22
update readme with doc links

tjhei:


gassmoeller (approved):
Not sure the `/en/latest` is needed, it is added automatically anyway. But doesnt matter too much, merge if you are ok with the current form.

tjhei:
doesn't matter, I would think.

PR #4762, opened 2022-05-22
Docs: move images to images directories

cmills1095:
Put images in separate image directories within the same directory as the markdown files that reference them, as per our meeting this morning


gassmoeller (approved):
:+1:

PR #4766, opened 2022-05-22
StructuredData: support large indices

tjhei:
Correctly support StructuredData with more than 4 billion data fields. This might not be something easy to encounter as ASCII files would need to be ~20 GB big, but it is not impossible to trigger.


gassmoeller (approved):
Good to go if you resolve the conflicts.

bangerth (commented):
I agree that this is good to go if you address the merge conflict.

tjhei:
rebased.

bangerth (approved):

PR #4767, opened 2022-05-22
(WIP) load netcdf in StructuredData

tjhei:
This is an early version that is confirmed to work for 1 dimensional data at least. It is not hooked up to be used in any of the AsciiData* plugins yet.

Sits on top of #4766


tjhei:
@bangerth FYI

include/aspect/structured_data.h
```diff
@@ -132,6 +132,13 @@ namespace aspect
         load_file(const std::string &filename,
                   const MPI_Comm &communicator);
 
+        /**
+         * Fill the current object with data read from a NetCDF file
+         * with filename @p filename. This call will fail if ASPECT is not
+         * configured with NetCDF support.
+         */
+        void load_netcdf(const std::string &filename);
```
bangerth, line 140: Wouldn't you want to do the same here as with the previous function where you read only once and then distribute?

include/aspect/structured_data.h
```diff
@@ -261,10 +268,10 @@ namespace aspect
 
         /**
          * Computes the table indices given the size @p sizes of the
-         * i-th entry.
+         * entry with index @p idx.
          */
         TableIndices<dim>
-        compute_table_indices(const TableIndices<dim> &sizes, const unsigned int i) const;
+        compute_table_indices(const TableIndices<dim> &sizes, const std::size_t idx) const;
```
bangerth, line 274: This will go away by rebasing, right?

source/structured_data.cc
```diff
@@ -604,6 +607,174 @@ namespace aspect
...
 
 
+
+    template <int dim>
+    void
+    StructuredDataLookup<dim>::load_netcdf(const std::string &filename)
+    {
+#ifndef ASPECT_WITH_NETCDF
+      (void)filename;
+      AssertThrow(false, ExcMessage("Loading NetCDF files is only supported if ASPECT is configured with the NetCDF library!"));
```
bangerth, line 617: Is it ASPECT or deal.II? (I forgot which project provides the netCDF bindings...)

source/structured_data.cc
```diff
@@ -604,6 +607,174 @@ namespace aspect
...
+      std::vector<std::string> coordinate_column_names(dim);
+      std::vector<std::string> data_column_names;
+      std::vector<Table<dim,double>> data_tables;
+      std::vector<std::vector<double>> coordinate_values(dim);
+
+      int ncid;
+      int status;
+
+      status = nc_open(filename.c_str(), NC_NOWRITE, &ncid);
+      AssertThrow(status == NC_NOERR, ExcMessage("NetCDF Error occured!"));
```
bangerth, line 629: Might be useful at some point to add an AssertNetCDF like we have AssertMPI and friends.

source/structured_data.cc
```diff
@@ -604,6 +607,174 @@ namespace aspect
...
+
+      int ndims, nvars, ngatts, unlimdimid;
+      status = nc_inq(ncid, &ndims, &nvars, &ngatts, &unlimdimid);
+      AssertThrow(status == NC_NOERR, ExcMessage("NetCDF Error occured!"));
+
+
+      for (int dimid=0; dimid<ndims; ++dimid)
+        {
+          char  name[NC_MAX_NAME];
```
bangerth, line 638: Ugh, this is solidly stuck in C land, isn't it? ;-)

source/structured_data.cc
```diff
@@ -604,6 +607,174 @@ namespace aspect
...
+          size_t length;
+          status = nc_inq_dim(ncid, dimid, name, &length);
+          AssertThrow(status == NC_NOERR, ExcMessage("NetCDF Error occured!"));
+          std::cout << "dim " << dimid << "'" << name << "' has length " << length << std::endl;
+          // TODO: parse/store coordinate name?
```
bangerth, line 643: What do we do with the table headings for ASCII data?

source/structured_data.cc
```diff
@@ -604,6 +607,174 @@ namespace aspect
...
+            const std::size_t n_elements = data_tables[var].n_elements();
+            raw_data.resize(n_elements);
+
+            // Load the data
+            status = nc_get_var_double(ncid, varids_to_use[var], raw_data.data());
+            AssertThrow(status == NC_NOERR, ExcMessage("NetCDF Error occured!"));
+
+            // .. and copy it over:
+            for (std::size_t n = 0; n < n_elements; ++n)
+              data_tables[var](compute_table_indices(new_table_points, n)) = raw_data[n];
```
bangerth, line 762: I guess this is the place you mentioned where you have to check that coordinates run in the same order as for our `Table<*,dim>`?

bangerth (commented):
Yes. right track for sure. Let me know when you want me to look at it again!

source/structured_data.cc
tjhei: ASPECT.

source/structured_data.cc
tjhei: for the data columns: we use them, for the coordinates: discard at the moment

include/aspect/structured_data.h
tjhei: Yes, but I was planning to do this in a separate PR.

source/structured_data.cc
tjhei: Fun fact: we are linking to the netcdf-c library in ASPECT, so this is in fact a C API. There used to be a netcdf-cxx library but that has been deprecated without a release since 2019. We used to have deal.II bindings to that, that we removed in 2018: https://github.com/dealii/dealii/pull/7392

source/structured_data.cc
tjhei: done.

tjhei:
While it is still not usable from user-code (TODO) and contains a bunch of debug messages, it is now at a point where it is working correctly.

I am not too happy about the reversed handling of 2d/3d datasets, but this seems to be specific to NetCDF. The example file loads correctly into ParaView and gives the expected result in ASPECT (x coordinate is the first argument to Point<3>()).
Am I missing something here? @pmbremner , what do you think?
See:
https://github.com/geodynamics/aspect/blob/9b5a9ad5edd8e8246018c54301a730cfa9539b1a/data/test/netcdf/example-3d.py#L46-L52

https://github.com/geodynamics/aspect/blob/9b5a9ad5edd8e8246018c54301a730cfa9539b1a/source/structured_data.cc#L733-L734

tjhei:
Ping @bangerth

data/test/netcdf/example-3d.py
```diff
@@ -0,0 +1,80 @@
...
+    data_2.long_name = 'the depth of the point'
+    
+    print(f'\n\nVariables:\n{ds}'
+      f'{data_1}\n'
+      f'{data_2}\n')
+
+    # coordinate data
+    data_x[:] = coord_x
+    data_y[:] = coord_y
+    data_z[:] = coord_z
```
bangerth, line 65: Is that not identical to 
```suggestion
    data_x = coord_x
    data_y = coord_y
    data_z = coord_z
```
?

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+      AssertThrowNetCDF(status);
+
+
+      for (int dimid=0; dimid<ndims; ++dimid)
+        {
+          char  name[NC_MAX_NAME];
+          size_t length;
+          status = nc_inq_dim(ncid, dimid, name, &length);
+          AssertThrowNetCDF(status);
+          std::cout << "dim " << dimid << "'" << name << "' has length " << length << std::endl;
```
bangerth, line 654: I think you want to remove this debug output here.

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+          // TODO: We could store coordinate name here, but the StructuredData class does not store them at the     moment.
```
bangerth, line 655: maybe wrap to avoid the awkward spacing before the last word?

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+          char  name[NC_MAX_NAME];
+          int natts;
+
+          status = nc_inq_var (ncid, varid, name, &xtype, &ndims, dimids,
+                               &natts);
+          AssertThrowNetCDF(status);
+          std::cout << "var " << varid << "'" << name << "' has ndims = " << ndims << ": ";
+          for (int d=0; d<ndims; ++d)
+            std::cout << " " << dimids[d];
+          std::cout << std::endl;
```
bangerth, line 683: Same here

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+
+          if (varid < dim)
+            {
+              std::cout << "\tignoring variable of dimension." << std::endl;
+            }
```
bangerth, line 688: And here. Unless you want to promote this to an error.

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+                }
+
+              if (use)
+                {
+                  varids_to_use.push_back(varid);
+                  data_column_names.push_back(name);
+                  std::cout << "\tusing variable." << std::endl;
+                }
+              else
+                std::cout << "\tignoring variable with different dims." << std::endl;
```
bangerth, line 720: and here

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+
+            }
+          else
+            std::cout << "\tignoring variable with the wrong number of dims." << std::endl;
```
bangerth, line 724: and here

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+        }
+
+
+      {
+        // Now load coordinate data
+
+        for (int d=0; d<dim; ++d)
```
bangerth, line 731:
```suggestion
        for (unsigned int d=0; d<dim; ++d)
```

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+                                 &natts);
+            AssertThrow(ndims == 1, ExcMessage("A variable of a dimension should have only one dimension."));
+
+            AssertThrow(xtype == NC_DOUBLE || xtype == NC_FLOAT, ExcMessage("We only support float or double data."));
+
+            coordinate_values[d].resize(new_table_points[d]);
+            status = nc_get_var_double(ncid, varid, coordinate_values[d].data());
+            AssertThrowNetCDF(status);
+
+            std::cout << "coordinate " << d << " length= " << new_table_points[d] << " data=  " << coordinate_values[d][0] << ' ' << coordinate_values[d][1] << " ..." << std::endl;
```
bangerth, line 752: and here

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+            // Allocate space
+            data_tables[var].TableBase<dim,double>::reinit(new_table_points);
+            const std::size_t n_elements = data_tables[var].n_elements();
+            raw_data.resize(n_elements);
+
+            // Load the data
+            status = nc_get_var_double(ncid, varids_to_use[var], raw_data.data());
+            AssertThrowNetCDF(status);
+
+            // .. and copy it over:
```
bangerth, line 772:
```suggestion
            // ... and copy it over:
```

source/structured_data.cc
```diff
@@ -604,6 +619,181 @@ namespace aspect
...
+      }
+
+
+      status = nc_close(ncid);
+      AssertThrowNetCDF(status);
+
+      // ready to go:
+      this->reinit(data_column_names, std::move(coordinate_values),std::move(data_tables));
+
+      std::cout << "load_netcdf: OK" << std::endl;
```
bangerth, line 793: and here

data/test/netcdf/example-3d.py
```diff
@@ -0,0 +1,80 @@
+''' Create example NetCDF files
+
+see https://unidata.github.io/netcdf4-python/ for more information
+
+'''
```
bangerth, line 5: Can you add a sentence or two what the resulting files are used for?

tjhei:
I addressed all comments and added support for specifying a list of columns to load.

This is ready for review again. Please note: there is still no way to call this from usercode yet.

gassmoeller (commented):
I don't particularly like all the C-style arrays, but I guess they are required by netcdf? Otherwise I have only one comment, we can merge this afterwards.

source/structured_data.cc
```diff
@@ -607,6 +623,251 @@ namespace aspect
...
+
+          for (int varid=0; varid<nvars; ++varid)
+            {
+              // Each netCDF dimension also has an associated variable that stores the
+              // coordinate data. We are looking for data columns, so skip them:
+              if (varid < ndims)
+                continue;
+
+              nc_type xtype;
+              int ndims;
```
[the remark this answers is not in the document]
tjhei: good, point. I will rename all these local variables.

tjhei:
> I don't particularly like all the C-style arrays,

You mean the ``int dim_ids[]`` and similar where we query information? One could replace them with ``std::array`` but I am not sure the code would be much better (as the number of returned items is less than the maximum).

gassmoeller (approved):

PR #4751, opened 2022-05-22
Docs: Convert Extending and Answers chapters

cmills1095:
This pull request, together with the others I've submitted, should wrap up everything in the manual that isn't a cookbook, benchmark, or parameter file.

PR #4773, opened 2022-05-22
Documentation: plugins

ljhwang:
Very minor update to clean up rendering.  File paths still need to be added.


gassmoeller:
Already fixed as part of #4751.

PR #4779, opened 2022-05-22
Add label to global parameters. Improve formatting.

gassmoeller:
A few improvements to the formatting of the parameter index.
1. Add a label for global parameters (this is referenced already in a few places).
2. Remove the misleading `parameters in section global` heading
3. Rename "global" to "Global parameters"
4. Shorten `Parameters in section ...` to `Subsection: ...`, which is also closer to https://aspect.geodynamics.org/doc/parameter_view/parameters.xml. (It is different from the current manual, but I think it is better).
5. Added spaces around "/" separating subsections to make subsection names more readable.

Let me know in case you want any of them reverted, I think the new format is much easier to read.

### DIFF
--- a/contrib/utilities/jsontomarkdown.py
+++ b/contrib/utilities/jsontomarkdown.py
@@ -38,7 +38,8 @@ def handle_subsection(data, cur_path, output_file):
             print("(parameters:" + path_str + ")=", file=output_file)
             new_path = cur_path + [key]
             section_name = path_str.replace("_20", " ")
-            print("## **Parameters in section** " + section_name, file=output_file)
+            section_name = section_name.replace("/"," / ")
+            print("## **Subsection:** " + section_name, file=output_file)
             handle_subsection(data[key], new_path, output_file)
 
 def handle_parameters(data):
@@ -58,8 +59,8 @@ def handle_parameters(data):
                              "---\n"
                              "global.md\n")
     global_parameters = open("doc/sphinx/parameters/global.md", "w")
-    print("# Global \n\n", file=global_parameters)
-    print("## **Parameters in section** global \n\n", file=global_parameters)
+    print("(parameters:global)=", file=global_parameters)
+    print("# Global parameters\n\n", file=global_parameters)
 
     cur_path = []
     keys = list(data.keys())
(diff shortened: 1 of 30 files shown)


contrib/utilities/jsontomarkdown.py
```diff
@@ -38,7 +38,8 @@ def handle_subsection(data, cur_path, output_file):
             print("(parameters:" + path_str + ")=", file=output_file)
             new_path = cur_path + [key]
             section_name = path_str.replace("_20", " ")
-            print("## **Parameters in section** " + section_name, file=output_file)
+            section_name = section_name.replace("/"," / ")
+            print("## **Subsection:** " + section_name, file=output_file)
```
bangerth, line 42: I think @tjhei removed the `** ... **` from the parameters already. I would favor it if we left the markup to sphinx's style sheets, rather than do it ourselves.

contrib/utilities/jsontomarkdown.py
```diff
@@ -100,7 +101,7 @@ def handle_parameters(data):
             new_path = cur_path + [key]
             section_name = path_str.replace("_20", " ").replace(":", "/")
             print("# " + section_name + "\n\n", file=subfile)
-            print("## **Parameters in section** " + section_name + "\n\n", file=subfile)
+            print("## **Subsection:** " + section_name + "\n\n", file=subfile)
```
bangerth, line 104: same hre

doc/sphinx/parameters/Adiabatic_20conditions_20model.md
```diff
@@ -2,7 +2,7 @@
 # Adiabatic conditions model
 
 
-## **Parameters in section** Adiabatic conditions model
+## **Subsection:** Adiabatic conditions model
```
bangerth, line 5: How is this shown in the latex manual? I think a better way to mark this up may be 
```suggestion
## Subsection &ldquo;Adiabatic conditions model&rdquo;
```

gassmoeller:
Hmm, I do understand that you do not want to format manually, but either way we want to emphasize one part of the heading differently from the other part, so we need to do some formatting (either bold or quotation marks or something else).
Have a look at the two version:

1. The version in this PR:


2. The version you suggest:


I personally like the bold version more, because it is more similar to the formatting of the **Parameter name** line, and makes more clear that it is a heading. What do you think?

PR #4780, opened 2022-05-22
Use 'include' instead of 'input' in latex

bangerth:
This also allows us to use paths relative to the local .tex files when further including .prm files and pictures. This makes sense because right now, when you're in `cookbooks/ABC/doc/something.tex` and you're trying to include a picture or a .prm file, you have to provide a path *relative to doc/manual* because, well, that's how the latex `\input` command works: It doesn't reset the directory you're in -- it works like the C preprocessor that just pastes text into the main document.

`\include` does set the path to where the included file is, but it has the downside that it also issues a `\clearpage` command, which nobody on the internet seems to think was ever a sensible default. So I use the `\include*` command from the `newclude` package, which does what `\include` does without the `\clearpage`.

All of this makes a bunch of things possible that I'd like to do next when converting the latex manual creation to cmake. In the end, this may not matter much any more, but it's a good testing ground for the rest of the infrastructure as already mentioned in #4656. 

Part of #3652.

/rebuild

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -11,6 +11,7 @@
 \usepackage{siunitx}
 \usepackage{cite}
 \usepackage{svg}
+\usepackage{newclude}
 
 % This adds space between the Appendix number (e.g., A.101) and the title.
 
@@ -143,7 +144,7 @@ COMPUTATIONAL INFRASTRUCTURE FOR GEODYNAMICS (CIG)
   \begin{textblock*}{0in}(0.0in,0.8in)
     \begin{center}
       \vspace{1em}
-      \includegraphics[width=4.5in]{../logo/unlabeled_logo.pdf}
+      \includesvg[width=4.5in]{../logo/unlabeled_logo.svg}
       \hspace{5em}
     \end{center}
   \end{textblock*}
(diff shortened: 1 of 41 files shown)


doc/manual/manual.tex
```diff
@@ -4602,7 +4603,7 @@ \subsubsection{A note on the syntax of formulas in input files}
 case is this snippet of code discussed in
 Section~\ref{sec:cookbooks-simple-box-3d}:
 %
-\lstinputlisting[language=prmfile]{cookbooks/convection_box_3d/doc/initial.part.prm.out}
+\lstinputlisting[language=prmfile]{../../cookbooks/convection_box_3d/doc/initial.part.prm.out}
```
gassmoeller, line 4606: Why is this necessary? For the other includes you do not need it.

doc/manual/manual.tex
```diff
@@ -11,6 +11,7 @@
 \usepackage{siunitx}
 \usepackage{cite}
 \usepackage{svg}
+\usepackage{newclude}
```
gassmoeller, line 14: Is this package part of our tester? I guess we will find out :-)

doc/manual/manual.tex
```diff
@@ -143,7 +144,7 @@
   \begin{textblock*}{0in}(0.0in,0.8in)
     \begin{center}
       \vspace{1em}
-      \includegraphics[width=4.5in]{../logo/unlabeled_logo.pdf}
+      \includesvg[width=4.5in]{../logo/unlabeled_logo.svg}
```
gassmoeller, line 147: Did you forget to add this file to the PR?

gassmoeller (commented):
I am ok with the changes, but I have a few questions.

doc/manual/manual.tex
bangerth: This is the main `manual.tex` file, and we're referencing a `.prm` file that is in one of the cookbooks. Previously, we went through the symlinks created in a previous PR, but those don't work in the cmake setup I'm building. So I used the true relative path within the repository.

doc/manual/manual.tex
bangerth: Oh, I had forgotten that I use that here. It's part of #4781. I can rebase this one on top of #4781 once that is merged.

gassmoeller (approved):
Good to go.

PR #4781, opened 2022-05-22
Add an svg version of the logo.

bangerth:
Like for all the other image files we previously converted. In this case, I propose to leave the .pdf file around simply because it's a file people will occasionally put into their (beamer) presentations.

/rebuild


gassmoeller (approved):

bangerth:
It looks like @tjhei's tester had a hick-up starting parallel jobs -- certainly unrelated to the patch itself. I think the patch is pretty safe if someone wanted to merge it without the tester succeeding. Or I can try to rebase.

gassmoeller:
No need, I thought the same when I saw the error.

PR #4782, opened 2022-05-22
Bilinear Least Squares particle interpolation updates

sac-bsa:
1. Use the cell averages of neighbors of the current cell to extend the property bounds used by the limiter. (Mirroring the same feature in #4600)
2. Added the option for boundary extrapolation for property bounds used by the limiter.  (Mirroring the same feature in #4600)
3. Use cell average in place of a constant coefficient for the least squares. This has the benefit of conserving mass between the cell average interpolator, bilinear least squares with  limiting, and bilinear least squares without limiting.
### For all pull requests:

I didn't add another test case(there are 9(edit actually 11) that use bilinear least squares already), but I looked at the existing ones, so that there are some that have no limiter, some with limiting but not boundary extrapolation, and some with limiting and boundary extrapolation.


gassmoeller (commented):
Just two minor comments. Better check if one of them changes the results (I think at the moment you never do boundary extrapolation.

gassmoeller (approved):
Good to go when the testers are done.

PR #4786, opened 2022-05-23
Create svg version of the plugin graph.

bangerth:
Also use it in manual.tex.

/rebuild


gassmoeller (commented):
@cmills1095 already added a .svg plugin graph in `./user/extending/images/plugin_graph.svg` can you use that? The rest of the PR makes sense.

bangerth:
If it's ok with you, let's go with this patch instead -- because this is the location where the `.dot` input file is already located. Then when I switch the sphinx documentation to run via cmake, I will remove the copy @cmills1095 created and generate the file on the fly (or take it from where I put my copy).

gassmoeller (approved):

PR #4790, opened 2022-05-23
Simplify code structure in python script slightly.

bangerth:
Just introducing shortcuts so we don't have to write the same expression over and over.

This script is not run automatically anywhere, so the testers will not do much good. But I have verified by hand that the output is the same.

/rebuild

### DIFF
--- a/contrib/utilities/jsontomarkdown.py
+++ b/contrib/utilities/jsontomarkdown.py
@@ -4,6 +4,24 @@ import json
 from sys import argv, exit, stderr
 
 
+def print_if_parameter_or_alias(entry, path_str, true_name, cur_path, output_file) :
+    if "value" in entry:
+        # this is a parameter
+        print("(parameters:" + path_str + ")=", file=output_file)
+        print("### __Parameter name:__ " + true_name, file=output_file)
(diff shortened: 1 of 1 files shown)


contrib/utilities/jsontomarkdown.py
```diff
@@ -3,6 +3,25 @@
 import json
 from sys import argv, exit, stderr
 
+
+def print_if_parameter_or_alias(entry, path_str, true_name, cur_path, output_file) :
+    if "value" in entry:
+        # this is a parameter
+        print("(parameters:" + path_str + ")=", file=output_file)
```
tjhei, line 10: This line is still duplicated. Do you want to move it outside the ``if``?

tjhei (approved):

contrib/utilities/jsontomarkdown.py
bangerth: I was about to do it, but then realized that it's really one part of a two-line statement whereby the first writes the label and the second the section heading. I think that conceptually they should stay together.

tjhei: OK.

tjhei (approved):

PR #4791, opened 2022-05-23
Convert some latex to md in doc strings.

bangerth:
Currently builds on #4790, only the last commit is new.

This PR introduces a function that we will use to escape some of the latex markup we have in our doc strings so that it renders reasonably in markdown. There is probably quite a lot to do eventually, but I'm starting with just quotes here as a proof of concept because the backticks we use in many places trip up markdown. I'm replacing back- and forward ticks by `&lsquo;` (left single quote) and `&rsquo;` (right single quote) respectively.

/rebuild


tjhei (approved):
Great, this was quite annoying to look at.

bangerth:
Now also contains the two commits from #4792.

gassmoeller:
Good to go if you fix the conflicts.

bangerth:
So done.

bangerth:
Seems to work now!

PR #4792, opened 2022-05-23
Escape some quotes in .md files

bangerth:
The first commit of this PR just updates the current set of parameters to what the main branch provides.

The second commit is the interesting one: It applies the transformation in #4791 to escape quotes. I made this a separate PR because if we merge this, the rendering of parameters will already be much nicer. 

/rebuild


tjhei:
I would rather merge  #4791 with the changes here included there as well to be consistent.

bangerth:
So you just want me to to throw these two commits into #4791 instead?

bangerth:
OK, so done.

gassmoeller:
This was part of #4791, right? Can this be closed?

PR #4793, opened 2022-05-23
(WIP) GMG: use deal.II compute_diagonal()

tjhei:
Use compute_diagonal() inside deal.II instead of our own version. This needs to be tested more, but we would be able to get rid of a lot of code this way.

I am opening this to be able to run the tests.

FYI: @zjiaqi2018


peterrum:
Why are the test outputs changing? Is it due to different values on the constrained entries or is this just a numerical artifact?

tjhei:
I am not sure, but this could be a bug. It seems like it only makes a difference if the Newton solver is used.

tjhei:
@zjiaqi2018 Can you take a look and review this PR?

zjiaqi2018:
> @zjiaqi2018 Can you take a look and review this PR?

Sure

zjiaqi2018:
> I am not sure, but this could be a bug. It seems like it only makes a difference if the Newton solver is used.

Strange, ABlock doesn't even have Newton derivatives.

source/simulator/stokes_matrix_free.cc
```diff
@@ -2928,139 +2904,14 @@ namespace aspect
...
-    const bool is_compressible = sim.material_model->is_compressible();
-
-    // Assemble and store the diagonal of the GMG level matrices derived from:
-    // 2*eta*(symgrad u, symgrad v) - (if compressible) 2*eta/3*(div u, div v)
     for (unsigned int level=0; level < sim.triangulation.n_global_levels(); ++level)
       {
         mg_matrices_Schur_complement[level].compute_diagonal();
 
-        // If we have a tangential boundary we must compute the A block
-        // diagonal outside of the matrix-free object
```
zjiaqi2018, line 2940: Just curious, why do we need to do this before?

source/simulator/stokes_matrix_free.cc
```diff
@@ -2928,139 +2904,14 @@ namespace aspect
...
-          }
-        else
-          {
-            mg_matrices_A_block[level].compute_diagonal();
-          }
+        mg_matrices_A_block[level].compute_diagonal();
 
         // This vector is no longer needed. Resize to 0.
-        level_viscosity_vector[level].reinit(0);
+        //level_viscosity_vector[level].reinit(0);
```
zjiaqi2018, line 2914: I guess we can remove these two lines, and move `level_viscosity_vector[level].reinit(0);`  to `evaluate_material_model()` after `mg_matrices_A_block[level].set_cell_data (level_cell_data[level]);`

tjhei:
superseded by #5060 #5061 #5062

PR #4794, opened 2022-05-23
Build the latex manual via cmake

bangerth:
The first commit is part of #4780. The second commit does everything necessary to let `pdflatex` run in the build directory to build the latex version of the manual.

/rebuild

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -28,27 +28,57 @@ FIND_PACKAGE(Perl)
 ####################################################
 
 # Step 1: Find all of the .prm files that will be used in
-#         the manual
-FILE(GLOB _dot_prm_directories_cookbooks
+#         the manual. Do the same for the .tex and .md
+#         files
+FILE(GLOB _doc_directories_cookbooks
      ${CMAKE_SOURCE_DIR}/cookbooks/*/doc)
-FILE(GLOB _dot_prm_directories_benchmarks
+FILE(GLOB _doc_directories_benchmarks
      ${CMAKE_SOURCE_DIR}/benchmarks/*/doc)
-SET(_dot_prm_directories ${CMAKE_SOURCE_DIR}/doc/manual)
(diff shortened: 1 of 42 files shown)


doc/CMakeLists.txt
```diff
@@ -93,7 +123,187 @@ FOREACH(_dot_prm ${_dot_prm_files})
...
+
+# Step 8: Run pdflatex to generate the pdf version of the manual. This
+#         obviously depends on all of the previous targets.
+ADD_CUSTOM_COMMAND(
+    OUTPUT
+      ${CMAKE_CURRENT_BINARY_DIR}/doc/manual/manual.pdf
+    WORKING_DIRECTORY
+      doc/manual
+    COMMAND
+      rm -r svg-inkscape
```
bangerth, line 270: For reasons not entirely clear to me, when building by hand in the build directory, it occasionally trips up on some of the svg-converted-to-pdf files. Nuking the directory makes it work reliably, so I put that into the recipe here as well. I'm not sure it's strictly necessary, but it probably also doesn't hurt: We run latex four times, and the conversions from svg to pdf happen only on the first occasion. I don't observe a noticeable slowdown.

doc/CMakeLists.txt
```diff
@@ -93,7 +123,187 @@ FOREACH(_dot_prm ${_dot_prm_files})
...
+    COMMAND
+      bibtex manual
+    COMMAND
+      makeindex prmindex
+    COMMAND
+      pdflatex --shell-escape --interaction=batchmode manual.tex
+    COMMAND
+      pdflatex --shell-escape --interaction=batchmode manual.tex
+    COMMAND
+      pdflatex --shell-escape --interaction=batchmode manual.tex
```
bangerth, line 282: This is what is currently in `doc/Makefile`.

gassmoeller (approved):
I am ok to merge this if you address the typos in a follow-up.

PR #4801, opened 2022-05-23
include cc file from benchmarks folder

jdannberg:
This PR changes one of the tests so that it includes the cc file in the benchmarks folder rather than making a copy of it.


tjhei (approved):

bangerth (approved):
If this works, it's ok. I would have chosen
```
  #include "..."
```
with quotes instead of angle brackets since for quotes it's clear that the path is interpreted as relative to the current file.

jdannberg:
I just copied it from some other test cc file that includes `#include <../benchmarks/tosi_et_al_2015_gcubed/tosi.cc>`. Maybe that would be something to fix in all of these places as part of a separate PR?

bangerth:
Sure would. See #4820.

PR #4814, opened 2022-05-23
fix a section reference

jdannberg:
This should fix the reference in #4810.


bangerth (approved):
My bad. I'm not going to double-check the correctness of the new link, though ;-)

PR #4821, opened 2022-05-23
Replace html tags in documentation

gassmoeller:
This will create a lot of conflicts, maybe we need to wait until things cool down after the hackathon (we should have done this beforehand).

@bangerth made a very valid point that we are loosing information here, and it is not correct to replace ldquo and rdquo with " (similar for lsquo and rsquo). I understand that point, but I am of the opinion it is more important to let the source files of our documentation appear less intimidating, to encourage users to write them. The main idea of markdown after all is to have files that are easily readable both in source form and when rendered.

Commands to create this PR (in the `doc/sphinx` directory):
```
perl -pi -e "s/\[\]//gm;" `find -L user -name '*.md'`
perl -pi -e "s/\&nbsp\;/ /gm;" `find -L user -name '*.md'`

perl -pi -e "s/\&lsquo\;/'/gm;" `find -L user -name '*.md'`
perl -pi -e "s/\&rsquo\;/'/gm;" `find -L user -name '*.md'`

perl -pi -e 's/\&ldquo\;/"/gm;' `find -L user -name '*.md'`
perl -pi -e 's/\&rdquo\;/"/gm;' `find -L user -name '*.md'`
```


bangerth:
.md are unicode, right? Or at least can be? If so, what if you replaced `&ldquo;` and friends by their respective unicode characters, rather than the ASCII names of these characters? People with European keyboards enter them all of the time by accident anyway, as do folks who write their texts in text processors like MS Word, and then copy-paste.

gassmoeller:
Good idea. I will try once the hackathon PRs have slowed down a bit.

tjhei:
I like the simplicity of using normal double-quotes in the text.

Did anyone of you check https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-smartquotes yet? Is this supposed to make normal double-quotes look pretty?

gassmoeller:
@tjhei Good point. https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-smartquotes automatically converts quotes into the correct opening and closing quotes, and also -- and --- into the correct dashes, so this is much less controversial than we thought. That option is on by default, which is also why  we did not see any difference between the rendered html in this PR and main (I have confirmed that the quotes are used literally, if I disable that setting). I have incorporated the commands into #4832, and will close this (it has acquired lots of conflicts already anyway). I have left `&nbsp;` out of the other PR, since it actually represents a loss of functionality.

PR #4824, opened 2022-05-23
Remove the latex manual parts of doc/Makefile.

bangerth:
Now that we can build the latex manual via cmake, we no longer need that logic in `doc/Makefile`.

/rebuild


bangerth:
Oh, the tester still builds the latex manual. What do we want to do about that? Switch the tester to using cmake for the manual? Not build the latex manual at all any more?

gassmoeller:
I can easily switch the tester to build the pdf manual using cmake for now (when #4827 is resolved), then we are at least sure it still builds until we remove it.

bangerth:
This is superseded now by #5200 .

PR #4828, opened 2022-05-24
Fix typo in viscoelastic_beam_modified benchmark

sac-bsa:
The beam described is supposed to be 9x as long as wide. It is also being described as 20km. Neither of which were true. Switching these numbers fixes that.
## For all pull requests:


danieldouglas92 (approved):
This is great!

gassmoeller (approved):
Thanks!

PR #4829, opened 2022-05-24
Fill additional material model inputs before evaluating material model

anne-glerum:
It seems logical to first fill the material model inputs and then evaluate the material model, as is done in all other places in the code.

### For all pull requests:

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1723,8 +1723,8 @@ namespace aspect
                   fe_values.reinit(simulator_cell);
                   in.reinit(fe_values, simulator_cell, sim.introspection, sim.current_linearization_point);
 
-                  sim.material_model->evaluate(in, out);
                   sim.material_model->fill_additional_material_model_inputs(in, sim.current_linearization_point, fe_values, sim.introspection);
+                  sim.material_model->evaluate(in, out);
 
                   const MaterialModel::MaterialModelDerivatives<dim> *derivatives
                     = out.template get_additional_output<MaterialModel::MaterialModelDerivatives<dim>>();


tjhei (approved):
Wow, I am surprised that this worked before! Thank you.

tjhei:
@zjiaqi2018 This could mean the Newton derivatives were incorrect...

tjhei:
... and I am surprised no test output changed.

gassmoeller (approved):
Nice find! We can merge this as is, but I am also noticing that we are not averaging here (like we do here:  https://github.com/geodynamics/aspect/blob/main/source/simulator/assembly.cc#L298). Do you think that is also missing? Maybe it does not matter here, because the Newton solver averages separately anyway? (if you dont know the reply no worry, we can merge this, and ask @tjhei).

zjiaqi2018:
> @zjiaqi2018 This could mean the Newton derivatives were incorrect...

I will double-check.

zjiaqi2018:
> ... and I am surprised no test output changed.

We only have two gmg-newton tests: 
tests/nonlinear_channel_flow_tractions_Newton_Stokes_GMG.prm
tests/tosi_benchmark_gmg_newton_solver.prm

I guess it's because in both material models, the additional output `MaterialModelDerivatives` is computed in `evaluate()`, so the order doesn't matter in these two tests.  For example: https://github.com/geodynamics/aspect/blob/6b31dc0ac900d80347022ed3bee2980109b7be12/benchmarks/newton_solver_benchmark_set/nonlinear_channel_flow/simple_nonlinear.cc#L128

https://github.com/geodynamics/aspect/blob/ed42380fb29307253c4d093bd9b3f29b6dd6c37f/benchmarks/tosi_et_al_2015_gcubed/tosi.cc#L89

PR #4830, opened 2022-05-25
Improve automatic manual conversion

gassmoeller:
These are some improvements to the automatically converted manual that we should have done at the start of the hack. I added the script commands I used to improve the conversion into the detailed description of the git commits. There is still some manual fix-up to do for all remaining cookbooks and benchmarks, but this should make the remaining proof-reading simpler.

@cmills1095: We talked about regular expressions at the hack, this might be interesting for you.


gassmoeller:
I will merge this with #4831 to a single script. That way we avoid conflicts and have it in a single place in case we need it again.

bangerth (approved):

bangerth:
OK to merge. Why don't you want to merge this by itself?

gassmoeller:
I thought I add the commands to the script I am writing for #4831. But of course we can merge this anyway and still add the commands to the script for future reference.

PR #4831, opened 2022-05-25
[WIP] Doc fix figures

gassmoeller:
Another step towards improving the automatic manual conversion, this time doing most of the fixes for figures. This is WIP, e.g. I did not catch all occurences of `embed` as well as `img` and I want to extract the labels as well. I will continue this soon. Currently I record the perl expressions in the commit messages, should I keep them in a script in case we need them again later?

PR #4832, opened 2022-05-25
Automatically fix documentation

gassmoeller:
Supersedes #4831. I had to exclude some files from the automatic improvements (those that were manually converted during the hackathon), but those would not have benefited anyway. Almost all figures render now (except for those without captions), and most figures convert perfectly (simple figures) to reasonably well (subfigures, they are now duplicated in multiple figures with the same caption).
I can read through the whole changes again tomorrow if needed. But I think this is pretty good so far.

@cmills1095 FYI


contrib/utilities/update_doc_files.pl
```diff
@@ -0,0 +1,77 @@
+#!/bin/perl
+
+# This script fixes some artefacts in the automatically converted documentation
```
tjhei, line 3:
```suggestion
# This script fixes some artifacts in the automatically converted documentation
```

gassmoeller:
I should add the table environment is a bit unfortunate. The table caption has to be on the same line as the directive (` ```{table}`) and can only be one line long, otherwise the table does not render. So I had to squeeze all the captions into one line.

Anyway your comments should be addressed.

gassmoeller:
Are you ok with these changes? I am a bit concerned that this PR will acquire a lot of conflicts, so I would like to get it merged before we continue with manual changes to the sphinx documentation.

gassmoeller:
Are you ok with these changes? I am a bit concerned that this PR will acquire a lot of conflicts, so I would like to get it merged before we continue with manual changes to the sphinx documentation.

tjhei (approved):

tjhei:
Let's go with it!

gassmoeller:
Thanks!

PR #4834, opened 2022-05-27
Follow-up fixes to 4248

gassmoeller:
Follow-up to #4248. @anne-glerum what do you think about removing the `force_terms`?


tjhei (approved):

anne-glerum:
Thank you for the follow-up @gassmoeller , I agree with keeping `force_terms` in `additional_outputs` for clarity.

PR #4836, opened 2022-05-27
[WIP] Include .prm file parts directly in documentation

gassmoeller:
I saw an interesting Sphinx feature while working on #4832: `literalinclude` can not only include parts of other files by line numbers, but also by searching for text strings and starting at or after them. The include can end at or before another string. This could allow us to get rid of a lot of `part.prm` files in the cookbooks and benchmarks. It will have some implication of the pre-processing of the .prm files though, so this is WIP for now, but I wanted to let you know about the features and how it looks like.


bangerth:
Doxygen has a similar feature, whereby you select based on line numbers. We used to use that at some point, but it was incredibly fragile because changing the source code made doxygen include nonsensical lines because line numbers no longer matched what was expected. Of course, in return you ended up with something that stayed relatively consistent with whatever syntax we preferred at the time, and with changing function names. In the end, though, it was not a workable solution and we decided that it is better to include code examples directly into the documentation.

I have a suspicion that we will find the same to be true here. When we rename parameters, at least we know that everything that has .prm in its name needs to be changed, which is already a substantial step forward beyond having these parameter snippets actually inside the .md files (where it would be much harder to recognize if something needed to be updated).

gassmoeller:
I see your point and agree. I will close this for now, I was mostly positively surprised that this is possible at all.

PR #4838, opened 2022-05-30
Documentation: Fix missing command?

bangerth:
I *think* that the list of commands in this section is missing the call to `make`. 

/rebuild


tjhei (approved):
Makes sense.

gassmoeller (approved):

PR #4840, opened 2022-06-03
Relax particle limiter tolerance a bit

gassmoeller:
At the moment we assert a very strict tolerance for the new particle limiters (10 epsilon relative to the property boundaries). This works for mappings that are analytically invertible like all of our box geometries and the 2D MappingQ for spherical shells, but at least in 3D MappingQ spherical shell geometries I have observed cases for which we trigger this assert. The MappingQ inversion internally uses a Newton iteration with a tolerance of 1e-12 so we can not expect an accuracy as good as we currently enforce. So far  1e-9 seems to be a good tolerance, at least I have never triggered this so far in my models, which proves the algorithm works as intended, it just cannot be as accurate as we thought due inaccuracies outside of the algorithms control.

@sac-bsa FYI.

### DIFF
--- a/source/particle/interpolator/bilinear_least_squares.cc
+++ b/source/particle/interpolator/bilinear_least_squares.cc
@@ -250,9 +250,10 @@ namespace aspect
                       }
                     if (use_linear_least_squares_limiter[property_index] == true)
                       {
-                        // If the limiter is working correctly, we should not be significantly more than machine error outside of the range of the property bounds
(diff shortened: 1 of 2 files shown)


tjhei (approved):
Looks good. I assume a test that runs fast is difficult to add, right? I am okay without.

Note that deal.II master is failing...

tjhei:
:+1:

PR #4847, opened 2022-06-18
fix: GMG free surface with adaptivity

tjhei:
The mesh deformation for the GMG solver on each multigrid level is
interpolated from the active level, but we renumbered the DoFs after
creating the transfer operator. This would mess up the data inside it
and causes the level data to be nonsense.


tjhei:
FYI @zjiaqi2018 @naliboff

tjhei:
It looks like it improves iteration numbers slightly for several tests! :-)

gassmoeller (approved):
Nice find.

PR #4851, opened 2022-06-22
Add 9.4 quick tester.

gassmoeller:
This PR adds an additional tester that compiles ASPECT against deal.II v9.4.0-rc2 (in a docker image I created) and runs the quick tests (we already compile against deal.II master in a separate tester). 

The nice thing about this design is that when we want to transition to 9.4 reference results, we can simply update the test results, and switch the 'ON' and 'OFF' settings below and continue to test if 9.3 still compiles.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -110,7 +110,14 @@ jobs:
     #linux build including tests
     name: tests
     runs-on: [ubuntu-latest]
-    container: geodynamics/aspect-tester:focal-dealii-9.3-v2
+    strategy:
+      matrix:
+        include:
+          - image: "geodynamics/aspect-tester:focal-dealii-9.3-v2"
+            tests: "ON"
+          - image: "geodynamics/aspect-tester:focal-dealii-9.4-v1"
+            tests: "OFF"
+    container: ${{ matrix.image }}
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


tjhei (approved):
Great!

tjhei:
Do you know what these errors are?
```
[6bc4f59955de:00754] Read -1, expected 4800, errno = 1
```
see https://github.com/geodynamics/aspect/runs/7031767323?check_suite_focus=true#step:5:12

MFraters:
I have no idea, but they also occur with the normal tester: https://github.com/geodynamics/aspect/runs/7032300798?check_suite_focus=true#step:5:3066 

Also before this pull request was merged: https://github.com/geodynamics/aspect/runs/7031715282?check_suite_focus=true#step:5:3061

MFraters:
seem to be a docker thing: https://github.com/horovod/horovod/issues/503#issuecomment-422461863

gassmoeller:
I think geodynamics/Calypso and geodynamics/Rayleigh already have a fix for this (see https://github.com/geodynamics/Rayleigh/pull/321), I just didnt  notice it appears in ASPECT as well. It would be an easy fix, if someone wants to open an issue I will get to it eventually (or if someone opens a PR I am happy to merge that). It does not affect the results or return value of the tests so it is not super urgent, just slightly annoying.

tjhei:
@gassmoeller Can you update to 9.4.0 as well? Thank you.

PR #4854, opened 2022-06-23
Add pp that computes the total volume of the domain

anne-glerum:
This PR adds a statistics postprocessor that computes the volume of the computational domain. This can be useful when you apply certain in- and outflow boundary conditions in combination with a free surface or other type of mesh deformation and you want to check that volume is conserved.

@Djneu I noticed that you wrote a plugin that also outputs the domain volume, but as this a one-liner, I thought this implementation would be the preferred one.

### For all pull requests:


include/aspect/postprocess/domain_volume_statistics.h
```diff
@@ -0,0 +1,54 @@
...
+
+
+namespace aspect
+{
+  namespace Postprocess
+  {
+
+    /**
+     * A postprocessor that computes some statistics about the material_flux
+     * in and out of boundaries.
```
gassmoeller, line 36: This needs an update

source/postprocess/domain_volume_statistics.cc
```diff
@@ -0,0 +1,72 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#include <aspect/postprocess/domain_volume_statistics.h>
+#include <aspect/utilities.h>
+#include <aspect/geometry_model/interface.h>
```
gassmoeller, line 24: These are not needed I think.

source/postprocess/domain_volume_statistics.cc
```diff
@@ -0,0 +1,72 @@
...
+
+#include <deal.II/base/quadrature_lib.h>
+#include <deal.II/fe/fe_values.h>
```
gassmoeller, line 27: These can also be removed.

source/postprocess/domain_volume_statistics.cc
```diff
@@ -0,0 +1,72 @@
...
+    template <int dim>
+    std::pair<std::string,std::string>
+    DomainVolume<dim>::execute (TableHandler &statistics)
+    {
+      const double global_volume = GridTools::volume (this->get_triangulation(), this->get_mapping());
+
+      // add the volume to the statistics object
+      // and create a single string that can be output to the screen
+      std::ostringstream screen_text;
+      const std::string name = "Model domain volume (m3) ";
```
gassmoeller, line 43: Can you make the unit dependent on dimension? Here and in line 53.

gassmoeller (approved):
Very useful! Just a few comments, and please add a changelog entry.

gassmoeller:
After checking: Actually, the Simulator class already has a variable `global_volume`, which is updated after ever mesh deformation. Can you instead of computing it again use the SimulatorAccess function `get_volume` to access the existing variable? Would save a bit of time.

source/postprocess/domain_volume_statistics.cc
```diff
@@ -0,0 +1,70 @@
...
+    template <int dim>
+    std::pair<std::string,std::string>
+    DomainVolume<dim>::execute (TableHandler &statistics)
+    {
+      // const double global_volume = this->get_volume();
+      const double global_volume = GridTools::volume(this->get_triangulation(), this->get_mapping());
+
+      // add the volume to the statistics object
+      // and create a single string that can be output to the screen
+      std::ostringstream screen_text;
```
tjhei, line 39: Move the declaration down to where it is used first.

tjhei:
> Hence the global volume in the simulator is not correct.

Well, I think we should look at that as well and fix it.

bangerth:
@anne-glerum Can you rebase this patch and try again? With @gassmoeller 's patch in, this might now actually work!

anne-glerum:
With #4881 the `simulator_access` function `get_volume` now returns the correct value of the domain volume over time. I'll update the test results when the tester is done.

source/postprocess/domain_volume_statistics.cc
```diff
@@ -0,0 +1,69 @@
...
+  {
+    template <int dim>
+    std::pair<std::string,std::string>
+    DomainVolume<dim>::execute (TableHandler &statistics)
+    {
+      // Retrieve the current domain volume
+      const double global_volume = this->get_volume();
+
+      // add the volume to the statistics object
+      const std::string unit = (dim == 2) ? "m2" : "m3";
```
bangerth, line 38:
```suggestion
      const std::string unit = (dim == 2) ? "m^2" : "m^3";
```

bangerth (commented):
OK if you fix the one string!

tjhei:
You need to update the output of ``free_surface_tangential_mesh_velocity``, @anne-glerum

anne-glerum:
Done!

gassmoeller (approved):
Thanks!

PR #4855, opened 2022-06-23
Add extra deal.ii hint to tester cmake file.

MFraters:
in my setup cmake worked fine as long as you didn't setup the tester. The path to my deal.ii dir is `/usr/lib64/cmake/deal.II` and it found it, but for some reason @DEAL_II_PATH@ only provided `/usr`. I was able to resolve the issue by adding @deal.II_DIR@. I don't know what the cause of this issue is, but to me is seem like an easy fix without any downsides (since it is just an extra hint).

Edit: I also had a change in here which made sure it is search for the version I you are actually using for your aspect, but that doesn't seem to work with the 9.4.0-pre version (`FIND_PACKAGE called with invalid argument "9.4.0-pre"`), so I just removed it from this pull request.


tjhei:
> I don't know what the cause of this issue is,

This is weird, because it should find deal.II without any hints as it is in a system location.



> I also had a change in here which made sure it is search for the version I you are actually using for your aspect,

You might need ``"9.4.0-pre" EXACT`` to get this to work. A better option might be to compare DEAL_II_PATH with @DEAL_II_PATH@, though.

MFraters:
> > I don't know what the cause of this issue is,
> 
> This is weird, because it should find deal.II without any hints as it is in a system location.

Yes, I don't know why it can't find it. It is strange.


> > I also had a change in here which made sure it is search for the version I you are actually using for your aspect,
> 
> You might need `"9.4.0-pre" EXACT` to get this to work. A better option might be to compare DEAL_II_PATH with @DEAL_II_PATH@, though.

I am fine with either option. I am happy to make a separate pull request with the DEAL_II_PATH @DEAL_II_PATH@ comparison if that is your preferred option.

tjhei:
> I am happy to make a separate pull request with the DEAL_II_PATH @DEAL_II_PATH@ comparison if that is your preferred option.

If you want to tackle #4866, that would be great. We need to test this a bit more, I would think (and maybe not include in 2.4).

PR #4856, opened 2022-06-23
Add projected density field to newton solver

MFraters:
@elodie-kendall reported to me that using the Steinberger cookbook didn't work with DC Picard, because it gave an error: `Unknown mass conservation equation approximation. There is no assembler`. I traced it back to that the Newton solver doesn't have that implemented. 

I have implemented the function and copied the Steinberger test, and only change the directory names in the output, which worked. I am not familiar enough with the projected density method to know if this is good enough or if further testing is needed.


MFraters:
I forgot to actually update the test prm file. Now that I did it, there are some small differences. I am guessing the difference mostly from switching from the default (`single Advection, single Stokes`) to `iterated Advection and defect correction Stokes`.

tests/steinberger_projected_density_dcPicard.prm
```diff
@@ -0,0 +1,128 @@
+# This is a copy of the steinberger compressible test.
```
tjhei, line 1: update this comment to mention Newton.

MFraters:
Thanks for the review. I have fixed it and also added a changelog entry

PR #4857, opened 2022-06-23
json2markdown: mangle/demangle for special characters

tjhei:
Correctly support all special characters in parameter names. Before we
were only handling space (" ") correctly.


tjhei:
This fixes https://aspect-documentation.readthedocs.io/en/latest/parameters/global.html#parameter-name-max-nonlinear-iterations-in-pre-2drefinement for example.

FYI @gassmoeller

gassmoeller (approved):
:+1: Thanks

PR #4858, opened 2022-06-23
update .gitattributes

tjhei:


MFraters (approved):
Looks good to me. Also nice that you added a comment explaining what that list of items is doing :)

P.S. I didn't know about the .gitattributes until you and Rene talked about it during the last aspect meeting. Seem also interesting for the world builder.

Issue #4859, opened 2022-06-23
SIGFPE in cookbooks/continental_extension/


tjhei:
```
-----------------------------------------------------------------------------
-- This is ASPECT, the Advanced Solver for Problems in Earth's ConvecTion.
--     . version 2.4.0-pre
--     . using deal.II 9.3.2
--     .       with 32 bit indices and vectorization level 1 (128 bits)
--     . using Trilinos 12.18.1
--     . using p4est 2.2.0
--     . running in DEBUG mode
--     . running with 16 MPI processes
-----------------------------------------------------------------------------

-----------------------------------------------------------------------------
-- For information on how to cite ASPECT, see:
--   https://aspect.geodynamics.org/citing.html?ver=2.4.0-pre&NewtonSolver=1&sha=&src=code
-----------------------------------------------------------------------------
Number of active cells: 3,056 (on 4 levels)
Number of degrees of freedom: 103,689 (25,122+3,201+12,561+12,561+12,561+12,561+12,561+12,561)

[...]

*** Timestep 1:  t=20000 years, dt=20000 years
   Solving mesh surface diffusion
   Solving mesh displacement system... 7 iterations.
   Solving temperature system... 11 iterations.
   Solving noninitial_plastic_strain system ... 17 iterations.
   Solving plastic_strain system ... 15 iterations.
   Solving crust_upper system ... 15 iterations.
   Solving crust_lower system ... 15 iterations.
   Solving mantle_lithosphere system ... 13 iterations.
SIGFPE received
```

bangerth:
FWIW, I cannot reproduce this:
```
Number of active cells: 3,200 (on 4 levels)
Number of degrees of freedom: 107,649 (26,082+3,321+13,041+13,041+13,041+13,041+13,041+13,041)

Number of mesh deformation degrees of freedom: 6642
   Solving mesh displacement system... 0 iterations.
*** Timestep 0:  t=0 years, dt=0 years
   Solving mesh displacement system... 0 iterations.
   Solving temperature system... 0 iterations.
   Skipping noninitial_plastic_strain composition solve because RHS is zero.
   Solving plastic_strain system ... 0 iterations.
   Solving crust_upper system ... 0 iterations.
   Solving crust_lower system ... 0 iterations.
   Solving mantle_lithosphere system ... 0 iterations.
   Initial Newton Stokes residual = 2.13503e+15, v = 2.13385e+15, p = 7.11214e+13

   Rebuilding Stokes preconditioner...
[...]

PR #4862, opened 2022-06-25
stokes assembler: do not request reaction_terms

tjhei:
part of #4859

The visco-plastic material model uses the strain_rate when computing reaction terms, but they might not be available leading to floating point exceptions. Work around this for the 2.4 release.

Most notably visible in the test gmg_mesh_deform_adaptive_bug2.

### DIFF
--- a/source/simulator/assembly.cc
+++ b/source/simulator/assembly.cc
@@ -524,14 +524,20 @@ namespace aspect
       data.local_pressure_shape_function_integrals = 0;
 
     // initialize the material model data on the cell
-    const bool update_strain_rate =
+    const bool need_viscosity =
       assemble_newton_stokes_system || this->parameters.enable_prescribed_dilation || rebuild_stokes_matrix;
 
     scratch.material_model_inputs.reinit  (scratch.finite_element_values,
                                            cell,
                                            this->introspection,
                                            current_linearization_point,
-                                           update_strain_rate);
+                                           need_viscosity);
+    scratch.material_model_inputs.requested_properties
+      =
+        MaterialModel::MaterialProperties::equation_of_state_properties |
+        MaterialModel::MaterialProperties::additional_outputs
+        |
+        (need_viscosity ? MaterialModel::MaterialProperties::viscosity : MaterialModel::MaterialProperties::uninitialized);
 
     for (unsigned int i=0; i<assemblers->stokes_system.size(); ++i)
       assemblers->stokes_system[i]->create_additional_material_model_outputs(scratch.material_model_outputs);


tjhei:
So, indeed, we illegally access in.strain_rate in several places even though it is empty. All the cases here should be fixed by #4861.

gassmoeller:
#4861 is merged. Do you want to rebase this so we can merge?

tjhei:
So the visco-plastic material model needs the strain-rate to compute reaction outputs, which are requested by default. Here we crash in the stokes rhs assembly, which does not need viscosity and therefore does not fill the strain rate.

tjhei:
This is kind of a hack for now (I just don't request reaction rates in the stokes rhs assembly). This should fix the test, but a real fix would be #4865


Should this go into 2.4?!

source/material_model/rheology/strain_dependent.cc
```diff
@@ -502,6 +502,7 @@ namespace aspect
         // Calculate changes in strain and update the reaction terms
         if  (this->simulator_is_past_initialization() && this->get_timestep_number() > 0 && in.requests_property(MaterialProperties::reaction_terms))
           {
+            Assert(in.strain_rate.size()>0, ExcInternalError());
```
naliboff, line 505: Is this assert just to provide an additional error message in the event that the conditions are met on line 503 but the strain rate is still empty? If so, would it be worth adding a more detailed error message or a comment above?

source/material_model/visco_plastic.cc
```diff
@@ -252,7 +252,8 @@ namespace aspect
             out.reaction_terms[i][c] = 0.0;
 
           // Calculate changes in strain invariants and update the reaction terms
-          rheology->strain_rheology.fill_reaction_outputs(in, i, rheology->min_strain_rate, plastic_yielding, out);
+          if (in.requests_property(MaterialProperties::reaction_rates))
```
naliboff, line 255: Would this condition still be needed if we added @anne-glerum suggestion in #4865? I suppose this check should save a little bit of computational time regardless, so perhaps it is worth adding both?

naliboff (approved):
@tjhei - Thanks for the fix and sorry for not getting to this sooner. The fixes seem reasonable, particularly in view of getting the 2.4 release finished. 

@anne-glerum suggestion in #4865 also seems like a good fix for when reaction rates are computed, and may catch more similar cases in other places. For example, I'm surprised this issue did not affect any of the elastic rheology, which currently use the reaction rates. But, up to you if that should be added before or after this PR/release.

source/material_model/rheology/strain_dependent.cc
tjhei: This assert is the only reliable way to trigger an error when you illegally access in.strain_rate[0] when it is empty. We should probably put it in a few more places inside material models just in case.
I can make the message better of course.

tjhei: Do you think this is the only place in ASPECT where we use the strain rate for something unrelated to the viscosity?

naliboff, line 505: How about?
```suggestion
            Assert(in.strain_rate.size()>0, "An illegal attempt was made to access in.strain_rate. It is currently empty and must be filled by the MaterialModelInputs in order for it to be accessed");
```

source/material_model/visco_plastic.cc
gassmoeller, line 255: As far as I understand this needs to be `MaterialProperties::reaction_terms`, and not `MaterialProperties::reaction_rates`. (see what happens inside StrainDependent<dim>::fill_reaction_outputs, it sets out.reaction_terms).

gassmoeller (commented):
If you include my comment, I think we should merge this independent on what we do in the other PRs, it is an improvement in any case.

tjhei:
Turns out StrainDependent already contains ``if (in.requests_property(MaterialProperties::reaction_terms))``, so this can be removed. I also removed the assert because we will go with #4873 instead.

This is hopefully good to merge.

PR #4876, opened 2022-07-04
Adds several theses and a recent publication

ljhwang:
Adds a master and phd theses and a new publication.

### DIFF
--- a/doc/manual/citing_aspect.bib
+++ b/doc/manual/citing_aspect.bib
@@ -1205,6 +1205,7 @@ DOI = {10.5194/se-13-849-2022}
 }
 
 @mastersthesis { quiroga2022numericalm,
+  author = {Quiroga, David },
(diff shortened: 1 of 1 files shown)


doc/manual/citing_aspect.bib
```diff
@@ -1203,3 +1203,32 @@ @Article{se-13-849-2022
 URL = {https://se.copernicus.org/articles/13/849/2022/},
 DOI = {10.5194/se-13-849-2022}
 }
+
+@mastersthesis { quiroga2022numericalm,
+	title = {Numerical Models of Lithosphere Removal in the Sierra Nevada de Santa Marta, Colombia},
```
tjhei, line 1208: there seems to be a missing ``author`` here and below.

ljhwang: Wow, that should not have got by me but did point out that the export feature on the website does not work so well.

bangerth (approved):
Thanks!

PR #4872, opened 2022-07-01
Use deal.II v9.4 release

gassmoeller:
Update the 9.4 tester to the official release.

I saw a number of tests crashing with 9.4 tester (when I manually run all tests, not just the quick tests), but that is independent of this PR. 

Here is a list for my own later reference, I will try to look though them in the next days (the gmg tests are likely fixed by #4871 or #4862):

	289 - free_surface_VE_cylinder_2D_loading_fixed_elastic_dt_gmg (Failed)
	310 - global_latent_heat (Failed)
	313 - global_refine_amr (Failed)
	320 - gmg_mesh_deform_ghost_entries (Failed)
	411 - latent_heat_melt_transport (Failed)
	421 - mass_reaction_term (Failed)
	483 - melting_rate_operator_splitting2 (Failed)
	502 - multiple_named_additional_outputs (Failed)
	664 - prescribed_field_temperature_max_isotherm (Failed)
	665 - prescribed_field_with_diffusion (Failed)
	670 - prescribed_temperature_with_diffusion (Failed)
	862 - traction_ascii_data (Failed)
	936 - viscoelastic_bending_beam_gmg (Failed)
	946 - viscoelastic_stress_build-up_gmg (Failed)

PR #4878, opened 2022-07-04
Add more testers

gassmoeller:
Some update to our CI workflow:

1. Only run on pull requests (opened, pushed, reopened). No longer run when you push a branch to your personal repo without opening a PR. (to do some good and save energy that is 99% wasted).
2. set environment variables in the workflow (for the deal.II dev container that does not have them in the container).
3. Integrate the deal.II dev container into the test matrix run (saves some lines, makes it easier to read)
4. Run all tests for deal.II 9.4, and deal.II-dev (just ignore the results, except for crashes). This will prevent situations like #4872 in the future.

It is possible that with this setting we need to switch off `run_tests` during hackathons for the not-reference deal.II version, because we essentially double the test time, by running 4 instead of 2 long jobs now. However, I think this is still preferable and will show us problems with recent deal.II version much earlier.

I do not think this needs to be in the 2.4 release, but having the PR open will already tell us if I caught all tests with the fixes to MappingCartesian.


gassmoeller:
I would like to keep this contained in one PR, because I always need the tester changes to check if my other changes work (so I would need to rebase the tester PR on all the other changes). It just turned out we need way more changes than expected to make ASPECT work correctly with deal.II 9.4 and dev. I think I got them all now.

If the tests succeed, let me know if you are ok merging this as is, or if I should break commits into multiple PRs.

source/simulator/core.cc
```diff
@@ -108,8 +108,13 @@ namespace aspect
     {
       if (geometry_model.has_curved_elements())
         return std::make_unique<MappingQCache<dim>>(4);
+
+#if !DEAL_II_VERSION_GTE(9,4,0) || DEAL_II_VERSION_GTE(9,4,1)
       if (Plugins::plugin_type_matches<const InitialTopographyModel::ZeroTopography<dim>>(initial_topography_model))
         return std::make_unique<MappingCartesian<dim>>();
+#else
+      (void) initial_topography_model;
```
tjhei, line 116: do we need a warning/error here?

tjhei (approved):
You can leave everything in one PR if you prefer.

tjhei:
Perfect, thank you!

PR #4880, opened 2022-07-05
Add MappingCartesian compatibility implementation

gassmoeller:
This PR moves a copy of the fixed MappingCartesian from dealii/dealii#14092 into our compatibility system. I have checked that this works with deal.II 9.4.0 and the version of dealii/dealii#14092. The tester will tell us if it works with 9.3. I think we need something like this in our release, otherwise a lot of models will break with 9.4.0.

I have put the compat .cc file into the simulator subfolder to hide it, but I am open to instead make an aspect/source/compat folder. Alternatively, I could also add everything to the header, it is just a lot of lines.

### DIFF
--- a/include/aspect/compat.h
+++ b/include/aspect/compat.h
@@ -29,7 +29,7 @@
 #include <memory>
 
 #if DEAL_II_VERSION_GTE(9,4,0)
-#if !DEAL_II_VERSION_GTE(9,5,0)
+#if !DEAL_II_VERSION_GTE(9,4,1)
 #include <aspect/compat/mapping_cartesian.h>
 #endif
 #endif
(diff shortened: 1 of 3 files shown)


include/aspect/compat.h
```diff
@@ -28,7 +28,11 @@
 #include <functional>
 #include <memory>
 
-
+#if DEAL_II_VERSION_GTE(9,4,0)
+#if !DEAL_II_VERSION_GTE(9,5,0)
```
tjhei, line 32: I would do 9.4.1 here.

tjhei (approved):
I am ok with this, but would limit to 9.4.0 only. Also, let's wait until the fix is merged in deal.II.

tjhei:
Good from my end. Feel free to merge if you think it is ready.

PR #4881, opened 2022-07-05
Recompute global volume after each mesh deformation

gassmoeller:
@anne-glerum I thought I implement this right away for #4854, then we already know tomorrow if any tests change. I saw that global_volume is only used in the right hand side correction of the pressure, which is disabled for models with open boundaries like free surface. So hopefully nothing should change, except we now have a more accurate global volume stored.


tjhei:
Do you want this in the release? I don't think we need #4854 or do we?

tjhei (approved):
Interesting. RMS velocity depends on the volume, which explains the large changes.

anne-glerum:
Thanks @gassmoeller . I'd suggest to also change the comment in line 1792 of core.cc ( `     // calculate global volume after displacing mesh (if we have, in fact, displaced it)`), because this global_volume computation does not happen after displacing the mesh, but after refining it. 

I'm glad the global volume was only used in 7 statistics postprocessors, apart from the rhs pressure correction.

I'll update #4854 after this is merged. @tjhei #4854 is not relevant for the release.

bangerth (approved):

bangerth:
Nice catch. Ok if you address @anne-glerum 's comment. 

Do you want to add a changelog entry?

tjhei (approved):

gassmoeller:
I updated the comments and added a changelog entry. Should be ready to go when the testers are done.

PR #4885, opened 2022-07-07
fix copying of prescribed temperature

jdannberg:
The function that writes the prescribed temperature outputs into the actual temperature vector used to only do this for the solution vector and not for the current linearization point. However, it seems like the Stokes solver uses the current linearization point to compute the material properties. Consequently, the solver would use the initial temperature rather than the prescribed temperature output to compute material properties. This was not obvious, since after solving everything, the solution is copied into the current linearization point, and all postprocessors use the correct temperature. It just produces the wrong solution. 

This pull request fixes this bug. 

I think this is an important bugfix and should also go into the release. 

### For all pull requests:


gassmoeller (commented):
This is one of the sneakiest bugs I have ever seen. The output in models with prescribed temperature/composition looks completely correct (after solving Stokes), but the assembly of the Stokes system is wrong, because it uses the current_linearization_point, which was not updated with the correct  temperature/composition. Thanks for all the debugging @jdannberg! The change looks good if you add a test and changelog entry and the testers succeed.

jdannberg:
I have added a test case: This is before (left) and after (right). Even though the visocsity is two orders of magnitude different, strain rate is constant in the image on the left.
test_prescribed

tjhei (approved):
Thank you, great job. 👍

jdannberg:
Don't merge this yet. This apparently breaks the nonlinear convergence of the entropy_half_space test and benchmark (specifically, the temperature). This can be fixed by making the time step smaller, but we should still look into this.

PR #4886, opened 2022-07-07
GMG: evaluate level data of Newton and free surface (WIP)

zjiaqi2018:
Added the level data of the Newton derivatives and free surface stabilization to ABlock for the GMG preconditioner, 
but it seems to increases the number of iterations based of the test results. Maybe there is a bug in the code. 
@tjhei can you take a look?

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -1175,11 +1175,72 @@ namespace aspect
   template <int dim, int degree_v, typename number>
   void
   MatrixFreeStokesOperators::ABlockOperator<dim,degree_v,number>
+  ::local_apply_face(const dealii::MatrixFree<dim, number> &,
+                     dealii::LinearAlgebra::distributed::Vector<number> &,
+                     const dealii::LinearAlgebra::distributed::Vector<number> &,
+                     const std::pair<unsigned int, unsigned int> &) const
+  {
+  }
+
+
+
+  template <int dim, int degree_v, typename number>
+  void
+  MatrixFreeStokesOperators::ABlockOperator<dim,degree_v,number>
+  ::local_apply_boundary_face(const dealii::MatrixFree<dim, number> &data,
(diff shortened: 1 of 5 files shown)


source/simulator/stokes_matrix_free.cc
```diff
@@ -1156,14 +1172,75 @@ namespace aspect
...
+                              const dealii::LinearAlgebra::distributed::Vector<number> &src,
+                              const std::pair<unsigned int, unsigned int> &face_range) const
+  {
+    // Assemble the fictive stabilization stress (phi_u[i].g)*(phi_u[j].n)
+    // g=pressure_perturbation * g_hat is stored in free_surface_stabilization_term_table
+    //  n is the normal vector
+    FEFaceEvaluation<dim, degree_v, degree_v + 1, dim, number> velocity(data);
+    const unsigned int n_faces_interior = data.n_inner_face_batches();
+
+    for (unsigned int face = face_range.first; face < face_range.second; ++face)
```
tjhei, line 1201: this goes over internal and boundary faces? But the check below skips interior faces?

tjhei:
The resulting iteration numbers do not differ by much (+/- a few iterations). Did you test this on a more difficult problem that requires more iterations?

tjhei:
And it makes sense to do that separately for free surface and newton terms. It might be that only one of them is correct to use on the levels.

tjhei:
I did not find a mistake in the code...

zjiaqi2018:
> The resulting iteration numbers do not differ by much (+/- a few iterations). Did you test this on a more difficult problem that requires more iterations?

Not yet. I am trying to find a more complicated test, any suggestions?

tjhei:
John's regional models with free surface. What did you use to make the Newton solver work with GMG? The Newton benchmarks/?
I would not draw any conclusions from tiny ``tests/`` but would look at ``cookbooks/`` or ``benchmarks/`` with higher resolution.

bangerth:
What should we do here?

PR #4887, opened 2022-07-07
Prescribed temperature fix

gassmoeller:
This PR supersedes and improves on #4885: We cannot put the update of the linearization point into `interpolate_material_output_into_advection_field`, because for prescribed fields with diffusion that means the linearization point is overwritten before computing the residual with the current best guess. Therefore the nonlinear solver does not converge. What I did here instead is to make the update of the current linearization point for the temperature block identical to how we handle it for the composition blocks, i.e. we update the linearization point at the end of the `assemble_and_solve_temperature` function for all possible advection methods.

### DIFF
--- a/doc/modules/changes/20220706_jdannberg
+++ b/doc/modules/changes/20220706_jdannberg
@@ -1,0 +1,8 @@
+Fixed: The prescribed temperature is now being used in the
+computation of material properties when assembling the
+Stokes equations. Before, the Stokes equations used the
+initial temperature rather than the prescribed temperature
+because it was only copied into the solution and not the
+current linearization point. 
+<br>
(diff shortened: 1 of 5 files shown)


doc/modules/changes/20220706_jdannberg
```diff
@@ -0,0 +1,8 @@
+Fixed: The prescribed temperature is now being used in the
+computation of material properties when assembling the
+Stokes equations. Before, the Stokes equations used the
+initial temperature rather than the prescribed temperature
+because it was only copied into the solution and not the
+current linearization point. 
+<br>
+(Juliane Dannberg, 2022/07/06)
```
tjhei, line 8: And Rene?

tjhei (approved):
Looks good!

PR #4889, opened 2022-07-07
Release initial conditions objects when possible.

bangerth:
I've tried to break this into individual commits that might be easier to review separately. The first n-1 commits provide the infrastructure and use it, and the last one actually releases the objects when possible.

It is not easy to identify when one can release these objects because we sometimes call `set_initial_temperature_and_compositional_fields ()` multiple times, depending on whether we are still in the initial refinement phase. As a consequence, these objects have to live at least until the last `goto start_time_iteration`, even though we do solve full time steps during that time. That means that I can't release them until after the first full solve, but that might still be worth it.

Let's see what happens with the testers.

/rebuild

### DIFF
--- a/benchmarks/solitary_wave/solitary_wave.cc
+++ b/benchmarks/solitary_wave/solitary_wave.cc
@@ -381,10 +381,17 @@ namespace aspect
 
         virtual double reference_darcy_coefficient () const
         {
+          // Make sure we keep track of the initial composition manager and
+          // that it continues to live beyond the time when the simulator
+          // class releases its pointer to it.
+          if (initial_composition_manager == nullptr)
+            const_cast<std::shared_ptr<const aspect::InitialComposition::Manager<dim>>&>(initial_composition_manager)
+              = this->get_initial_composition_manager_pointer();
+
           // Note that this number is based on the background porosity in the
           // solitary wave initial condition.
           const SolitaryWaveInitialCondition<dim> &initial_composition =
-            this->get_initial_composition_manager().template
+            initial_composition_manager->template
             get_matching_initial_composition_model<SolitaryWaveInitialCondition<dim>>();
 
           return reference_permeability * pow(initial_composition.get_background_porosity(), 3.0) / eta_f;
@@ -460,6 +467,14 @@ namespace aspect
         double xi_0;
         double eta_f;
         double reference_permeability;
+
+        /**
+         * A shared pointer to the initial composition object
+         * that ensures that the current object can continue
+         * to access the initial composition object beyond the
+         * first time step.
+         */
+        std::shared_ptr<const aspect::InitialComposition::Manager<dim>> initial_composition_manager;
     };
 
     template <int dim>
--- a/include/aspect/simulator_access.h
+++ b/include/aspect/simulator_access.h
@@ -637,6 +637,43 @@ namespace aspect
        * This can then, for example, be used to get the names of the initial temperature
        * models used in a computation, or to compute the initial temperature
        * for a given position.
+       *
+       * While the Simulator class creates a shared pointer to an initial
+       * temperature manager before the first time step, it releases
+       * the pointer once it no longer needs access to the initial
+       * compositions. As a consequence, you can only call this function
+       * during the first time step.
+       *
+       * If the Simulator's shared pointer were the only
+       * one that points to the initial temperature manager object, that
+       * would also destroy the object pointed to. However, plugin classes
+       * can have member variables that are *also* shared pointers to
+       * these manager objects, and if you initialize such a shared
+       * pointer from the result of this function -- typically in the
+       * `initialize()` function of a plugin class -- then the Simulator
+       * giving up its shared pointer does not actually destroy the
+       * manager object but extends its lifetime until the last plugin
+       * that has a pointer to it is destroyed itself. As a consequence,
+       * if you need access to the initial temperature in a plugin, you
+       * will need to keep a shared pointer to it around for as long
+       * as you need it.
+       */
+      std::shared_ptr<const InitialTemperature::Manager<dim>>
+      get_initial_temperature_manager_pointer () const;
+
+      /**
+       * Return a reference to the manager of the initial temperature model.
+       * This can then, for example, be used to get the names of the initial temperature
+       * models used in a computation.
+       *
+       * While the Simulator class creates a shared pointer to an initial
+       * temperature manager before the first time step, it releases
+       * the pointer once it no longer needs access to the initial
+       * temperature. As a consequence, you can only call this function
+       * during the first time step. If a plugin needs access to the initial
+       * temperature at a later time, it has to store its own shared
+       * pointer to that object, and that is what can be achieved using
+       * the get_initial_temperature_manager_pointer() function above.
        */
       const InitialTemperature::Manager<dim> &
       get_initial_temperature_manager () const;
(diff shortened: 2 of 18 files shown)


bangerth:
Absolutely!

include/aspect/simulator_access.h
```diff
@@ -637,6 +637,43 @@ namespace aspect
...
+       * one that points to the initial temperature manager object, that
+       * would also destroy the object pointed to. However, plugin classes
+       * can have member variables that are *also* shared pointers to
+       * these manager objects, and if you initialize such a shared
+       * pointer from the result of this function -- typically in the
+       * `initialize()` function of a plugin class -- then the Simulator
+       * giving up its shared pointer does not actually destroy the
+       * manager object but extends its lifetime until the last plugin
+       * that has a pointer to it is destroyed itself. As a consequence,
+       * if you need access to the initial compositions in a plugin, you
```
tjhei, line 657: Temperature

include/aspect/simulator_access.h
```diff
@@ -637,6 +637,43 @@ namespace aspect
...
+       * models used in a computation.
+       *
+       * While the Simulator class creates a shared pointer to an initial
+       * temperature manager before the first time step, it releases
+       * the pointer once it no longer needs access to the initial
+       * temperature. As a consequence, you can only call this function
+       * during the first time step. If a plugin needs access to the initial
+       * temperature at a later time, it has to store its own shared
+       * pointer to that object, and that is what can be achieved using
+       * the get_initial_composition_manager_pointer() function above.
```
tjhei, line 676: temperature

source/simulator/simulator_access.cc
```diff
@@ -579,14 +579,36 @@ namespace aspect
...
+  }
+
+
+
   template <int dim>
   const InitialTemperature::Manager<dim> &
   SimulatorAccess<dim>::get_initial_temperature_manager () const
   {
-    return simulator->initial_temperature_manager;
+    return *get_initial_temperature_manager_pointer();
```
tjhei, line 607: This will always copy the shared pointer, modify the ref count, etc..
Wouldn't it make sense to access it directly from the simulator here?

source/simulator/simulator_access.cc
```diff
@@ -597,14 +619,36 @@ namespace aspect
...
+  }
+
+
+
   template <int dim>
   const InitialComposition::Manager<dim> &
   SimulatorAccess<dim>::get_initial_composition_manager () const
   {
-    return simulator->initial_composition_manager;
+    return *get_initial_composition_manager_pointer();
```
tjhei, line 647: Same here

tjhei (approved):
Great! I have a few minor comments.

tjhei:
Can you also add a changelog entry?

source/simulator/simulator_access.cc
bangerth: Yes. I didn't see another way: Right now we have
```
      std::shared_ptr<InitialTemperature::Manager<dim>>                      initial_temperature_manager;
```
and
```
      std::shared_ptr<const InitialTemperature::Manager<dim>>
      get_initial_temperature_manager_pointer () const;
```
The pointed to object is non-`const` in `class Simulator`, but we want to return a pointer to `const`. The current declaration of `get_initial_temperature_manager_pointer()` does that. The only other option I see is to let the function return a `const` reference to the non-`const` object.

...

...

Well, of course I could also just let the function that returns the reference to the object do the access itself and not go through the `_pointer()` function. On second thought, that's exactly what you are proposing -- I guess I just needed to think that through myself :-)

So implemented now!

bangerth:
OK, finally seems to work!

bangerth:
Ping?

benchmarks/solitary_wave/solitary_wave.cc
```diff
@@ -381,10 +381,17 @@ namespace aspect
 
         virtual double reference_darcy_coefficient () const
         {
+          // Make sure we keep track of the initial composition manager and
+          // that it continues to live beyond the time when the simulator
+          // class releases its pointer to it.
+          if (initial_composition_manager == nullptr)
```
tjhei, line 387: Shouldn't this be in a initialize () function?

benchmarks/solitary_wave/solitary_wave.cc
```diff
@@ -460,6 +467,14 @@ namespace aspect
         double xi_0;
         double eta_f;
         double reference_permeability;
+
+        /**
+         * A shared pointer to the initial temperature object
```
tjhei, line 472: Composition?

benchmarks/solitary_wave/solitary_wave.cc
bangerth: This ended up not working because it is in the material model for which we call `initialize()` *before* the initial conditions objects are created (line 298 of `core.cc` for the `initialize()` call of the material model vs lines 308 and following for the initial condition creation). 

This was part of the reason for #4906 . I can address this as a follow-up to #4906 if you'd like me to, but would like to leave it as-is in this patch for now.

benchmarks/solitary_wave/solitary_wave.cc
bangerth: Yes, good catch. Fixed.

benchmarks/solitary_wave/solitary_wave.cc
tjhei: Got it.

tjhei (approved):

source/simulator/simulator_access.cc
```diff
@@ -579,14 +579,47 @@ namespace aspect
...
   template <int dim>
   const InitialTemperature::Manager<dim> &
   SimulatorAccess<dim>::get_initial_temperature_manager () const
   {
-    return simulator->initial_temperature_manager;
+    Assert (simulator->initial_temperature_manager,
+            ExcMessage ("You are trying to access the initial temperature manager "
+                        "object, but the Simulator object is no longer keeping "
+                        "track of it because the initial time has passed. If "
+                        "you need to access to this object after the first time "
```
gassmoeller, line 611:
```suggestion
                        "you need to access this object after the first time "
```

source/simulator/simulator_access.cc
```diff
@@ -579,14 +579,47 @@ namespace aspect
...
+
+  template <int dim>
+  std::shared_ptr<const InitialTemperature::Manager<dim>>
+  SimulatorAccess<dim>::get_initial_temperature_manager_pointer () const
+  {
+    Assert (simulator->initial_temperature_manager,
+            ExcMessage ("You are trying to access the initial temperature manager "
+                        "object, but the Simulator object is no longer keeping "
+                        "track of it because the initial time has passed. If "
+                        "you need to access to this object after the first time "
```
gassmoeller, line 591:
```suggestion
                        "you need to access this object after the first time "
```

source/simulator/simulator_access.cc
```diff
@@ -597,14 +630,47 @@ namespace aspect
...
+
+  template <int dim>
+  std::shared_ptr<const InitialComposition::Manager<dim>>
+  SimulatorAccess<dim>::get_initial_composition_manager_pointer () const
+  {
+    Assert (simulator->initial_composition_manager,
+            ExcMessage ("You are trying to access the initial composition manager "
+                        "object, but the Simulator object is no longer keeping "
+                        "track of it because the initial time has passed. If "
+                        "you need to access to this object after the first time "
```
gassmoeller, line 642:
```suggestion
                        "you need to access this object after the first time "
```

source/simulator/simulator_access.cc
```diff
@@ -597,14 +630,47 @@ namespace aspect
...
   template <int dim>
   const InitialComposition::Manager<dim> &
   SimulatorAccess<dim>::get_initial_composition_manager () const
   {
-    return simulator->initial_composition_manager;
+    Assert (simulator->initial_composition_manager,
+            ExcMessage ("You are trying to access the initial composition manager "
+                        "object, but the Simulator object is no longer keeping "
+                        "track of it because the initial time has passed. If "
+                        "you need to access to this object after the first time "
```
gassmoeller, line 662:
```suggestion
                        "you need to access this object after the first time "
```

gassmoeller (changes requested):
Just one small typo.

While the code is correct, I think the reason I have been putting this review off for so long is that I am not quite comfortable with this change. I am concerned we will get a number of forum messages that will use the initial condition in custom plugins and run into the new asserts, and do not know how to keep a shared ptr around. On the other hand the number of models that will actually benefit from the released memory is quite small. Would you be ok with introducing a new input parameter `Release initial condition after setup` or similar that is off by default and needs to be enabled to make use of the new feature? It would make the feature less visible and useful, but we would not break user models. I should have brought this up earlier, but did not have a good idea for how to work around it (weirdly I only thought about the input parameter today).

tjhei:
>  I am concerned we will get a number of forum messages that will use the initial condition in custom plugins and run into the new asserts, and do not know how to keep a shared ptr around. 

I don't think it would be a big issue pointing them at one of the many existing plugins that grab a pointer to the initial conditions in initialize().

gassmoeller:
> I don't think it would be a big issue pointing them at one of the many existing plugins that grab a pointer to the initial conditions in initialize().

True, but it still feels surprising that accessing the initial condition after the first timestep now always requires additional work on the side of the plugin. Anyway, maybe it is just my intuition, I am not totally opposed to doing it the way this PR is currently set up. I can also see that adding the parameter as in my suggestion will mean many plugins will not be made compatible with turning the parameter on, so maybe requiring it for every plugin is actually the better solution. I am ok with merging it the current way if @bangerth also thinks this is better than the alternative.

bangerth:
I do think that I would really like to have the optimization on by default (in which case there is no point in having a parameter: users have to do *something* anyway, and we can point them at exactly what they need to do).

I suspect that the number of models that are using initial conditions is not actually very large. I commit to answering all questions about this on the forum :-)

(I fixed the grammar issue pointed out by @gassmoeller .)

gassmoeller (approved):
Ok, let's push the button and see what happens :-). It is an improvement in any case. Thanks @bangerth.

PR #4891, opened 2022-07-07
Use std::unique_ptr for plugin creation, rather than raw pointers.

bangerth:
A rabbit hole opened up right in front of me looking at just one small place. I fell right into it.

Long story short, it turns out that in our plugin systems, we had all of these factory functions that create objects. They all returned raw pointers; in many but not all places, these returned raw pointers were then converted to `std::unique_ptr`s. This is just not very modern any more. This patch addresses this by (hopefully) consistently letting factory functions return `std::unique_ptr`s right away, and then dealing with them at the receiving end.

Also not for the upcoming release.

/rebuild


bangerth:
Alright, seems to work now. That took a bit of investigating but I got it to work and we're now using a substantially more modern way of doing things.

tjhei (approved):
Looks good!

tjhei (approved):

PR #4895, opened 2022-07-09
Minor clean-up in the plugin infrastructure.

bangerth:
A companion to #4891.

/rebuild

### DIFF
--- a/include/aspect/plugins.h
+++ b/include/aspect/plugins.h
@@ -157,10 +157,12 @@ namespace aspect
         /**
          * A type describing everything we need to know about a plugin.
          *
-         * The entries in the tuple are: - The name by which it can be
-         * selected. - A description of this plugin that will show up in the
-         * documentation in the parameter file. - A function that can declare
-         * the run-time parameters this plugin takes from the parameter file.
+         * The entries in the tuple are:
+         * - The name by which it can be selected.
+         * - A description of this plugin that will show up in the
+         *   documentation in the parameter file.
+         * - A function that can declare the run-time parameters this
+         *   plugin takes from the parameter file.
          * - A function that can produce objects of this plugin type.
          */
         using PluginInfo
@@ -175,7 +177,7 @@ namespace aspect
          * The object is a pointer rather than an object for the following
          * reason: objects with static initializers (such as =0) are
          * initialized before any objects for which one needs to run
-         * constructors. consequently, we can be sure that this pointer is set
+         * constructors. Consequently, we can be sure that this pointer is set
          * to zero before we ever try to register a postprocessor, and
          * consequently whenever we run Manager::register_postprocessor, we
          * need not worry whether we try to add something to this list before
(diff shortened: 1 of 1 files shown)


tjhei:
There is also a compile error. 😉

include/aspect/plugins.h
```diff
@@ -181,7 +181,7 @@ namespace aspect
          * need not worry whether we try to add something to this list before
          * the lists's constructor has successfully run
          */
-        static std::list<PluginInfo> *plugins;
+        static std::unique_ptr<std::list<PluginInfo>> plugins;
```
tjhei, line 184: Does the comment above still apply regarding initialization order? unique_ptr does have a constructor.

bangerth:
Oh, I had totally forgotten about the reason why we used pointers. Thanks for digging this out of the comments!

This invalidates the whole idea and probably explains why the original idea resulted in all sorts of segfaults on the testers (though interestingly not on my laptop). I've stripped the patch back to the minimal pieces that keep things as a pointer.

tjhei:
Didn't have to do much digging, it was right there above the declaration. I actually looked because I was wondering why we have a pointer instead of a std::list directly.

Another way to work around the awkward pointer would be to have a getter that returns a reference to a static std::list.

include/aspect/plugins.h
```diff
@@ -323,18 +325,16 @@ namespace aspect
...
+          Assert (std::get<0>(p) != name,
                   ExcMessage ("A plugin with name <" + name + "> has "
                               "already been registered!"));
 
         // now add one record to the list
-        plugins->push_back (PluginInfo(name,
-                                       description,
-                                       declare_parameters_function,
-                                       factory_function));
+        plugins->emplace_back (PluginInfo(name,
```
tjhei, line 334:
```suggestion
        plugins->emplace_back (name,
```
Otherwise emplace_back will do the same as push_back.

bangerth: Yes, good eye! So fixed!

tjhei (approved):

bangerth:
Ah, good grief. That's what happens when you work on your laptop and are working on the one file that is literally included into every single .h file and for which compiling is going to fry your knees. Not even `simulator.h` is that bad...

tjhei:
I think ASPECT is s pleasure to develop because of the fast compilation times. Always annoyed when I have to recompile deal.II...
Are you using unity and precompiled headers?

tjhei (approved):

bangerth:
Neither on my laptop. But there really also isn't all that much you can do when you're mucking around with the most central pieces. You're entirely right when working on plugins or individual `.cc` files -- they are not very connected and pleasantly fast to compile.

(Why not unity: Because not using unity is fastest when you're working on a single `.cc` file. I don't know what impact PCH would have.)

tjhei:
Try both some day. Recompiling one of the unity files when a single .cc changes is also surprisingly fast.

PR #4902, opened 2022-07-11
update copyright

tjhei:


bangerth (approved):
It fails one test, but I cannot see how that is related to this patch. Merging.

PR #4907, opened 2022-07-11
Add copyright header to files currently lacking them.

### LINKED ISSUE #4901: tests/ .cc files are missing ASPECT header
see for example ./tests/prescribed_temperature.cc
Somebody want to run a script?

bangerth:
Fixes #4901. I have not tried to figure out which years these files were added and simply used 2022 as the (single) copyright year.

/rebuild


tjhei (approved):

tjhei:
We could figure it out from the git history, but maybe just choosing 2022 is enough.

PR #4909, opened 2022-07-12
Fix cmake with newer version of world builder.

MFraters:
The current dev branch of the world builder moved the source files of the library to a sub folder of source (see https://github.com/GeodynamicWorldBuilder/WorldBuilder/pull/443). This means that aspect will need to look in a different location if linked to the most resent 0.5.0-pre version of the world builder. This pull request add a check to make sure that it find the correct path. Even though we won't update the world builder for release 2.4.0 of aspect, it might be good to still get this in there before the release.


tjhei (approved):

tjhei:
I cherry-picked on to aspect-2.4.

PR #4911, opened 2022-07-13
Prevent overwriting artifacts in CI pipeline

### LINKED ISSUE #4897: [CI] Github runner artifacts are overwritten
It looks like we no longer have the .diff files for updating test results since we switched to the new matrix testing in #4878. I assume this is because the files get overwritten.

Example: https://github.com/geodynamics/aspect/actions/runs/2647098191

@gassmoeller We should make the artifact upload conditional on test comparison. Do you have time to look into that?

gassmoeller:
@tjhei: What about this solution to #4897?

Fixes #4897.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -100,14 +100,17 @@ jobs:
           - image: "geodynamics/aspect-tester:focal-dealii-9.3-v2"
             run-tests: "ON"
             compare-tests: "ON"
+            result-file: "changes-test-results-9.3.diff"
             container-options: '--name container'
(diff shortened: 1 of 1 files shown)


tjhei (approved):
I would have disabled the step for the other runners, but this works as well. The only thing is that you have to tell people to pick the non-empty file.

Anyways, I am okay with this. Do you want to test this by breaking a test?

gassmoeller:
I think if we get three separate artifact files in the end that should be test enough, no?

tjhei:
cherry-picked, just in case.

PR #4918, opened 2022-07-21
update enum name for deal.II master

tjhei:
see https://github.com/dealii/dealii/pull/14123

### DIFF
--- a/source/postprocess/visualization.cc
+++ b/source/postprocess/visualization.cc
@@ -610,7 +610,7 @@ namespace aspect
 
           data_out.write_deal_II_intermediate_in_parallel(filename,
                                                           this->get_mpi_communicator(),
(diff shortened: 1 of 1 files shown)


bangerth (approved):

tjhei:
> One tester failed during compilation due to "error 137".

That was the deal.II master image. I don't know about the error code, but this is indeed a problem. Other PRs are now failing the same way.

gassmoeller:
"error 137" may be related to running out of memory on the testers (7 GB available). I do not know if there was a significant change in the memory requirements for compiling something against deal.II recently. I will do a bit more testing. In case we need to, we could reduce the number of simultaneous make jobs from 2 to 1 , but we should look into options to avoid doubling the compile time (e.g. larger unity batch sizes to reduce the number of compile targets).

Issue #4905, opened 2022-07-11
tests/gmg_mesh_deform fails with deal.II master


tjhei:
with a floating point exception, see https://github.com/geodynamics/aspect/pull/4902#event-6973011711 for example

bangerth:
This is outside my area of expertise, so I can not really help with what is going on. But this is what I find:
```
(gdb) list
3336      DEAL_II_ALWAYS_INLINE
3337      VectorizedArray &
3338      operator+=(const VectorizedArray &vec)
3339      {
3340    #    ifdef DEAL_II_COMPILER_USE_VECTOR_ARITHMETICS
3341        data += vec.data;                      *********************************************** error happens here
3342    #    else
3343        data = _mm_add_pd(data, vec.data);
3344    #    endif
3345        return *this;
(gdb) p vec.data
value has been optimized out
(gdb) p vec
$1 = <optimized out>
(gdb) up
#1  dealii::internal::EvaluatorTensorProduct<(dealii::internal::EvaluatorVariant)2, 3, 2, 2, dealii::VectorizedArray<double, 2ul>, dealii::VectorizedArray<double, 2ul> >::apply<1, false, true, 1, false> (shapes=0x55555ac62580, in=0x55555af7a640, out=0x55555af7a340)
    at /home/bangerth/p/deal.II/1/install/include/deal.II/matrix_free/tensor_product_kernels.h:1876
1876                        out[stride * col] += r0 + r1;
(gdb) p r0
$2 = {
  <dealii::VectorizedArrayBase<dealii::VectorizedArray<double, 2>, 2>> = {<No data fields>}, 
  members of dealii::VectorizedArray<double, 2>:
  data = {-3.9405883313765773e+264, -inf}
}
(gdb) p r1
$3 = {
  <dealii::VectorizedArrayBase<dealii::VectorizedArray<double, 2>, 2>> = {<No data fields>}, 
  members of dealii::VectorizedArray<double, 2>:
  data = {0, 0}
}
```

I don't know why `r0` has the infinity in it.

tjhei:
Thanks. I will take a look. I could not reproduce on my system sorry this is helpful.

tjhei:
Can you share the rest of the backtrace?

bangerth:
Here is the whole backtrace:
```
[...]

PR #4922, opened 2022-07-22
Reduce gcc memory usage on deal.II development tester

gassmoeller:
This PR reduces the required memory for the tester that uses the deal.II development version. With the previous official deal.II image we often saw the github action run out of memory. We now build our own deal.II image once a day and use the same settings as for the testers that use deal.II releases.


.github/workflows/linux.yml
```diff
@@ -132,8 +132,9 @@ jobs:
         -D ASPECT_UNITY_BUILD=ON \
         -D ASPECT_RUN_ALL_TESTS='${{ matrix.run-tests }}' \
         -D ASPECT_COMPARE_TEST_RESULTS='${{ matrix.compare-tests }}' \
+        -D CMAKE_UNITY_BUILD_BATCH_SIZE=16
```
tjhei, line 135: \

tjhei:
I would try -j2 first because ninja uses n+1=3 by default.

gassmoeller:
Interesting, I didnt know that. `man ninja` says the default is #CPUS, but it does not specify how it computes #CPUS. I can try 2 if 1 works.

gassmoeller:
Looks like 2 processes is too much and leads to the crash. Let's see if the tester runs if I use a custom build docker image (the same way we use a custom image for the releases). I imagine that might use less memory, because it only includes the deal.II dependencies that ASPECT needs. We could build the custom image automatically, just as we build the main ASPECT docker image. (maybe once daily or every few days).

gassmoeller:
With the custom docker image the tester succeeds and compiling is also more than twice as fast as the official deal.II docker image (likely because less dependencies are included). I think this is something we should pursue, but I want to set up a github action first that automatically rebuilds the deal.II image in a reasonable time. I will create a follow-up PR some time next week. Do we need a quick fix for the dev tester?

tjhei:
> Interesting, I didnt know that. `man ninja` says the default is #CPUS

If you do ``ninja -h`` it will tell you the default. My machine has 8 cores, 16 threads, and ninja uses 18.

tjhei:
> (likely because less dependencies are included).

How can fewer deal.II dependencies result in faster ASPECT compilation with less memory?
Is the compiler the same?

gassmoeller:
> How can fewer deal.II dependencies result in faster ASPECT compilation with less memory?
> Is the compiler the same?

Well, I do not know the details, but I would assume more included headers, less removed lines by the preprocessor, and a larger deal.II library means more work for the compiler and linker. This is consistent with the difference in size of the precompiled header file between geodynamics/aspect-tester (635MB) vs dealii/dealii:master-focal (851MB). One additional influence may be that for geodynamics/aspect-tester we remove debug symbols, and compile deal.II debug with -O3 instead of -O0. 

Anyway to compare the results here are two outputs of /usr/bin/time on the same machine with exactly the same cmake options and use of ninja -j2 in the two different docker containers:

geodynamics/aspect-tester:latest:
```
[117/117] Linking CXX executable aspect
773.03user 43.34system 6:58.67elapsed 194%CPU (0avgtext+0avgdata 3766996maxresident)k
7968inputs+1994408outputs (26major+16075361minor)pagefaults 0swaps
```

dealii/dealii:master-focal:
```
[117/117] Linking CXX executable aspect
1458.11user 79.70system 13:04.87elapsed 195%CPU (0avgtext+0avgdata 5970244maxresident)k
8inputs+11733408outputs (158067major+27066299minor)pagefaults 0swaps
```

So not only does it take twice the time for dealii/dealii it also require nearly 6GB RAM instead of 3.76GB.

tjhei:
That difference is impressive. How do you want to go about this now? How do we update deal.II?

bangerth:
How do we want to proceed here?

tjhei:
Discussed this with @gassmoeller: I will prepare an alternative running on my jenkins instead.

gassmoeller:
Actually I tested this a bit more on my fork. Building a new dealii dev docker container for this tester just takes about 2 hours (https://github.com/gassmoeller/aspect/runs/7657669991?check_suite_focus=true) so we could easily do it as a github action nightly. The benefit would be that we have the three github actions build on containers that have consistent build parameters for deal.II (all with minimal dependencies and -O3 for debug mode).

I agree that using the official deal.II image for your Jenkins tester would add value (because then we also test the official image), but we could also do that in addition to the changes in this PR. In the current state this PR should work and fix the build problems for the deal.II-dev github action.

tjhei:
If you want to set it up this way, please go ahead.

Did you see that the most recent dealii master is working again (but we have failing tests)?

gassmoeller:
> If you want to set it up this way, please go ahead.

Ok, I interpret that as a ready to merge?

> Did you see that the most recent dealii master is working again (but we have failing tests)?

Interesting, it seems gmg mesh deformation tests are failing with floating point exceptions. But this is currently using a deal.II dev from a few weeks ago when I set up this PR first. Let me merge this and tomorrow we will have a fresh deal.II image.

.github/workflows/docker_tester.yml
```diff
@@ -0,0 +1,41 @@
+name: github-docker-tester
```
tjhei, line 1: I think the naming of this job and filename can be improved.

tjhei:
No, the Deal-master is working on main without this PR merged. See any other new PR. We might not need to merge this PR after all.

We are working on the FPEs in #4905.

gassmoeller:
Oh interesting. Well I am ok with closing this then, we can always come back to it if it becomes a problem in the future again.

gassmoeller:
Let me reopen this. @tjhei this should exactly fix #4961 as we discussed this morning. I renamed the file, do you think the new name is good enough? I am open for suggestions.

.github/workflows/build_tester_base_image.yml
```diff
@@ -0,0 +1,41 @@
...
+          password: ${{ secrets.DOCKER_HUB_LOGIN }}
+
+      - name: Build and push Docker image for main
+        uses: docker/build-push-action@v3
+        with:
+          context: ./contrib/docker/tester/
+          cache-from: type=registry,ref=ubuntu:20.04
+          cache-to: type=inline
+          push: true
+          tags: geodynamics/aspect-tester:latest
```
tjhei, line 40: Should we add deal.II or "master" or something to the name? It shows up in the description of the runs on each pull request.

contrib/docker/tester/Dockerfile
```diff
@@ -22,13 +22,13 @@ RUN cd /opt && \
...
     mv /opt/local.cfg . && \
-    ./candi.sh -p /opt -j64 && \
+    ./candi.sh -p /opt -j2 && \
     rm -rf /opt/tmp && \
     ln -s /opt/astyle-2.04/astyle /usr/bin/astyle
 
 # Set environment variables for this image to be used
 # by Github Actions
-ENV DEAL_II_DIR /opt/deal.II-v9.4.0
+ENV DEAL_II_DIR /opt/deal.II-master
```
tjhei, line 31: Does it make sense to keep this folder intact or maybe introduce subfolders for the different versions?

tjhei (approved):

tjhei:
Can you also please update the PR description?

contrib/docker/tester/Dockerfile
gassmoeller: This is the name candi chooses inside the install directory. I do not know if I can change that.

gassmoeller (commented):
Updated the PR description and container name. I do not know if I can do anything about the deal.II install directory inside the container though.

gassmoeller:
If you agree with the container name I can build it once manually to make the tester succeed (it currently fails because the container will only be built once this PR is merged).

contrib/docker/tester/Dockerfile
tjhei: I am sorry, I meant "does it make sense to keep ``contrib/docker/tester/Dockerfile`` as is or introduce a 9.4 and a master version"?

contrib/docker/tester/build.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
 # This script generates a docker image for the ASPECT tester.
-# It requires a docker installation on the local machine, and the ability to
-# communicate with the docker daemon without root user privileges (see the docker
-# webpage for an explanation).
 
-docker build -t geodynamics/aspect-tester:focal-dealii-9.4-v2 .
+docker build -t geodynamics/aspect-tester:latest .
```
tjhei, line 5: why is this ``latest`` and not ``focal-dealii-master``?

contrib/docker/tester/Dockerfile
gassmoeller: Ah I see. Well they are pretty identical and easy to change. I would prefer not creating another copy that can get out of date.

contrib/docker/tester/build.sh
gassmoeller: My bad I did not check that file. Fixed now.

tjhei:
Do you want to manually push ``geodynamics/aspect-tester:focal-dealii-master`` so that the master tester succeeds? Otherwise, feel free to merge.

tjhei (approved):

tjhei:
Looks like #4973 did the trick!

PR #4923, opened 2022-07-22
Enable use of Iterative Advection schemes in combination with strain-dependent rheologies

naliboff:
This PR enables the use of iterative Advection schemes in the strain rheology module, after testing and discussions with @anne-glerum, @jdannberg, and others confirmed that nonlinear iterative Advection schemes should not produce unrealistic accumulation of strain within compositional fields updated with reactions terms.


tjhei (approved):
So everything was already working correctly? Good!

naliboff:
> So everything was already working correctly? Good!

Yep, it appears so. @jdannberg write out a nice explanation of why it should work elsewhere, which may be worth adding to this PR for documentation. Short answer - during each nonlinear iteration the reaction term is being added to the old solution and not the current linearization point.

tests/visco_plastic_yield_strain_weakening_iterated_advection.prm
```diff
@@ -0,0 +1,44 @@
+# A test to check whether the calculated finite strain invariant
+# (total strain in this case) does not unrealistically accumulate during
+# nonlinear iterations when an iterative Advection scheme is used. We
+# check this by running the postprocessors on every nonlinear iteration.
+# The test output from the composition statistics postprocessor,
+# which prints out min/max/mass of each compositional field, confirms
+# that the
```
anne-glerum, line 7: sentence unfinished

tests/visco_plastic_yield_strain_weakening_iterated_advection.prm
```diff
@@ -0,0 +1,44 @@
...
+set Dimension                              = 2
+set Start time                             = 0
+set End time                               = 2e5
+set Maximum time step                      = 1e5
+set Use years in output instead of seconds = true
+
+set Nonlinear solver scheme                = iterated Advection and Stokes
+
+# Set a very strict nonlinear tolerance such that we perform more than
+# one nonlinear iterations after the first time step.
```
anne-glerum, line 20:
```suggestion
# one nonlinear iteration after the first time step.
```

tests/visco_plastic_yield_strain_weakening_iterated_advection.prm
```diff
@@ -0,0 +1,44 @@
...
+set Nonlinear solver tolerance             = 1e-14
+
+# Cap the number of nonlinear iterations at 5, as we
+# only need to a few iterations to how the strain
```
anne-glerum, line 24:
```suggestion
# only need to a few iterations to see how the strain
```
?

tests/visco_plastic_yield_strain_weakening_iterated_advection.prm
```diff
@@ -0,0 +1,44 @@
...
+# field evolves within each timestep.
+set Max nonlinear iterations               = 5
+
+set Output directory                       = visco_plastic_yield_strain_weakening_iterated_advection
+
+# Start with no initial (total) strain such that the
+# accumualated strain should be constant throughout
```
anne-glerum, line 31:
```suggestion
# accumulated strain should be constant throughout
```

tests/visco_plastic_yield_strain_weakening_iterated_advection.prm
```diff
@@ -0,0 +1,44 @@
...
+# the domain over the course of this model run.
```
anne-glerum, line 32: This sentence sounds like the strain doesn't change over time. Do you mean it's uniform throughout the domain at each timestep?

source/material_model/rheology/strain_dependent.cc
```diff
@@ -271,21 +271,32 @@ namespace aspect
                          Parameters<dim>::NonlinearSolver::single_Advection_iterated_Stokes
                          ||
                          this->get_parameters().nonlinear_solver ==
+                         Parameters<dim>::NonlinearSolver::iterated_Advection_and_Stokes
```
anne-glerum, line 274: Does this mean we're now checking against all the available solver schemes? Can the whole Assert be removed? Or switched to checking what solvers should _not_ be used?
Or is it better to keep the Assert for future cases where new solvers are introduced?

anne-glerum (changes requested):

anne-glerum:
@naliboff For completeness, did you also check the Newton solver?

jdannberg:
@naliboff It looks like Anne already did a review, so here just for completeness is my explanation for why it should already work as intended:

If you use reaction terms, then they’re assembled in the equation. So if we simplify everything to a first-order scheme, it would look something like
`(C_new – C_old)/Delta t + u grad C = R/Delta t`
where `R` is the reaction term. If we make this super simple and ignore advection for now, we can get the new `C` as
`C_new = C_old + R`
`C_old` is the old solution, which is not updated within nonlinear iterations (unless you use advection methods like prescribed fields with diffusion, for them it’s different). So even if `R` depends on the solution (here, I believe, it is something like strain_rate * Delta t), you get a reasonable `C_new`, because `R` is added to `C_old`, which does not change. The only problem would be if you had something like `R = -C` (using the current linearization point for `R`). In this case, your first iteration would set `C_new` to `C_old – C_old = 0`, but then in the next one, `R` would be zero because `C_new` is zero, so `C_new = C_old`, so then in the next iteration `R = - C_old`, and so on. You basically jump between the values. That is the problem I had in the melt models, but I do not see this as a problem here.
And I think the idea of using reaction_terms = max(update, - C) is that you can never make the composition smaller than zero with your reaction terms (you can never subtract anything more than the current value of the composition). I think that makes sense too.

source/material_model/rheology/strain_dependent.cc
naliboff: Good point. My preference for now would be to keep the current assert in case something does break with the current allowed solvers. However, if everything seems to be working smoothly after some period of time (months, one year) then maybe we can remove it altogether?

naliboff:
@anne-glerum - thanks for the review, and I've addressed all of the comments you made above (one more note below). @jdannberg - Thanks for investigating and posting the notes above.

@anne-glerum - I tried to run the same test with the Newton Stokes solver, but always converged within one iteration after the first time step. At least during the first step the strain values stayed at 0 during the nonlinear iterations. Next, I ran the continental_extension cookbook test with the iterative Advection scheme and there the change in the max/min strain values during each nonlinear iteration was much smaller than the magnitude of the strain values. Would you like me to add this test to the PR?

tests/visco_plastic_yield_strain_weakening_iterated_advection.prm
```diff
@@ -0,0 +1,45 @@
...
+set Use years in output instead of seconds = true
+
+set Nonlinear solver scheme                = iterated Advection and Stokes
+
+# Set a very strict nonlinear tolerance such that we perform more than
+# one nonlinear iteration after the first time step.
+set Nonlinear solver tolerance             = 1e-14
+
+# Cap the number of nonlinear iterations at 5, as we
+# only need to a few iterations to see how the strain
```
anne-glerum, line 25:
```suggestion
# only need a few iterations to see how the strain
```

anne-glerum (approved):

anne-glerum:
During t0 the reaction terms are not computed, so no changes in the strain would be accumulated anyway. If the Picard Stokes solver does several nonlinear iterations, is there a way to make the Newton Stokes solver do several as well? By lowering the switch tolerance for example so that it does more Picard solves? What about the defect correction solver, does that do more iterations?

PR #4930, opened 2022-07-28
fix zenodo.json

tjhei:
This is the reason zenodo gave an 'internal server error' when we created the 2.4 release. Sigh.


gassmoeller (approved):
You would think with all the talk about big data and AI our systems would not stumble over something like this :-).

tjhei:
Yeah. 😉
It turns out that it was invalid json to have the comma there.

PR #4931, opened 2022-07-31
Fix a few bibtex entries.

bangerth:
/rebuild

### DIFF
--- a/doc/manual/citing_aspect.bib
+++ b/doc/manual/citing_aspect.bib
@@ -1084,7 +1084,7 @@ abstract = {Orogeny results from crustal thickening at active margins, and much 
 
 @Article{geosciences12020050,
 AUTHOR = {Palmiotto, Camilla and Ficini, Eleonora and Loreto, Maria Filomena and Muccini, Filippo and Cuffaro, Marco},
-TITLE = {Back-Arc Spreading Centers and Superfast Subduction: The Case of the Northern Lau Basin (SW Pacific Ocean)},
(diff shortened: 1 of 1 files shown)


doc/manual/citing_aspect.bib
```diff
@@ -1138,7 +1138,7 @@ @article{STEIN2022229276
 
 @article{https://doi.org/10.1029/2021JB023244,
 author = {Zha, Caicai and Lin, Jian and Zhou, Zhiyuan and Xu, Min and Zhang, Xubo},
-title = {Effects of Hotspot-Induced Long-Wavelength Mantle Melting Variations on Magmatic Segmentation at the Reykjanes Ridge: Insights From 3D Geodynamic Modeling},
+title = {Effects of Hotspot-Induced Long-Wavelength Mantle Melting Variations on Magmatic Segmentation at the {R}eykjanes Ridge: Insights From 3D Geodynamic Modeling},
```
tjhei, line 1141: How do you decide what letters to force uppercase? Ridge is part of the name, shouldn't it also be uppercase? What about 3D and Insights?

ljhwang: I do not decide. I take whatever is the "official" export for the publisher is and fix any obvious translation issues of special characters.

bangerth: I can't say that I had a philosophy, I just wanted to improve things :-) If you tell me what you want, I'll fix it.

It probably doesn't matter much here because we don't use bibtex for these entries, but jabref. I don't believe that jabref actually translates to lower-case, though I may be wrong.

tjhei: yeah, jabref keeps the capitalization as is it seems. All the same to me. :-)

tjhei (approved):

Issue #4919, opened 2022-07-21
Convergence issue: 3D convection with an Earth-like initial condition (S40RTS)


fadhli-atarita:
Hello,

I've been trying to run this cookbook for a while but I've always encountered convergence issues. I've been using the exact same parameters as in the S40RTS.prm and have tried to increase the refinements and maximum stokes solver steps and decrease the linear solver tolerance. However, these changes also produced convergence issues.

Here are the output, error, and solver history files for your reference

[solver_history.txt](https://github.com/geodynamics/aspect/files/9162061/solver_history.txt)
[error_5214159.log](https://github.com/geodynamics/aspect/files/9162062/error_5214159.log)
[output_5214159.log](https://github.com/geodynamics/aspect/files/9162063/output_5214159.log)

Thank you,

Fadhli

tjhei:
Your ASPECT version seems to be quite old. Can you retry with the current development version, please? Is this the output without any changes to the .prm file?

fadhli-atarita:
There was a segmentation fault issue on the clusters with v2.3.0 so the admin suggested that I use v2.2.0 for now. This output was produced with changes on the 3 parameters that I mentioned earlier. Without any changes it would still have similar issues but at a different timestep.

tjhei:
@zjiaqi2018 Can you try cookbooks/initial-condition-S20RTS/S20RTS.prm with the current ASPECT master and run it for a while to check if it crashes for you as well?

zjiaqi2018:
Seems to be fine:
```

*** Timestep 121:  t=2.36743e+07 years, dt=155599 years
   Solving temperature system... 23 iterations.
   Rebuilding Stokes preconditioner...
   Solving Stokes system... 92+0 iterations.

   Postprocessing:
     RMS, max velocity:                  0.261 m/year, 1.15 m/year
     Writing heat flux map:              output-S20RTS/heat_flux.00121
     Heat fluxes through boundary parts: -1.219e+14 W, 7.079e+14 W
     Density at top/bottom of domain:    3359 kg/m^3, 3260 kg/m^3
     Pressure at top/bottom of domain:   8.67e-07 Pa, 9.576e+10 Pa
     Computing dynamic topography        
     Writing geoid anomaly:              output-S20RTS/geoid_anomaly.00121
     Writing graphical output:           output-S20RTS/solution/solution-00121
```

zjiaqi2018:
update: It does fail if it's long enough.

```

[...]

PR #4934, opened 2022-08-03
cookbooks/initial-condition-S20RTS

zjiaqi2018:
Turn off mesh refinement to avoid divergence in a time dependent problem as discussed in https://github.com/geodynamics/aspect/issues/4919


tjhei (approved):
Thank you.

tjhei:
Can you rebase to current master? Otherwise this is good to go but I want to see what is going on with the tests.

gassmoeller:
For some reason the jenkins tester does not seem to run this PR even though it has the correct label. All other testers perform as expected so I would say we go ahead and merge anyway.

PR #4936, opened 2022-08-03
Improve output of copying prescribed outputs to fields.

gassmoeller:
A small improvement to the output when prescribing advection fields based on material model output. So far all the output happened at the start of the function, making it hard to determine if the function had finished if the model would crash or get stuck before the next screen output. This new output pattern makes that immediately clear by splitting the output into an intro and a final part, it also makes the structure identical to how we do it for the direct Stokes solver.


tjhei (approved):

tjhei:
Ready if you update test results!

bangerth (approved):

bangerth:
Two groups of tests fail on one of the tester, but for reasons I don't believe are related.

tjhei:
Unrelated, dealii-master is currently broken.

PR #4937, opened 2022-08-04
Assert density is positive in release mode

gassmoeller:
As discussed in our meeting on Tuesday it would be helpful if this check would be performed in release mode as well as debug mode. I checked that the assert now triggers in models that have negative density. The change slows down the assembly of the temperature system by a bit less than 0.1%, and this assembly typically makes up less than 15% of the total run time, so I think this is completely acceptable (given how often we had reports about crashing solvers caused by this issue).

### DIFF
--- a/source/simulator/assemblers/advection.cc
+++ b/source/simulator/assemblers/advection.cc
@@ -111,9 +111,9 @@ namespace aspect
              :
              1.0);
 
-          Assert (density_c_P >= 0,
-                  ExcMessage ("The product of density and c_P needs to be a "
-                              "non-negative quantity."));
+          AssertThrow (density_c_P >= 0,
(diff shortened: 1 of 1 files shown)


tjhei (approved):

tjhei:
Should we run all benchmarks/cookbooks first to see what breaks? I can start these later...

gassmoeller:
Do you usually run the benchmarks and cookbooks in release mode? If you run them in debug mode we already know everything works. Otherwise, yes it might be a good idea to run them first.

tjhei:
You are right, I run them in debug mode using check.mk, which only does a single timestep, though. Maybe we need to run some of them for a longer time after this is merged.

PR #4938, opened 2022-08-04
Release the WorldBuilder object early.

bangerth:
This follows the same approach as taken for the initial conditions objects in #4889.

/rebuild

### DIFF
--- a/source/simulator/simulator_access.cc
+++ b/source/simulator/simulator_access.cc
@@ -588,12 +588,12 @@ namespace aspect
             ExcMessage ("You are trying to access the initial temperature manager "
                         "object, but the Simulator object is no longer keeping "
                         "track of it because the initial time has passed. If "
-                        "you need to access to this object after the first time "
+                        "you need to access this object after the first time "
                         "step, you need to copy the object returned by "
                         "this function before or during the first time step "
                         "into a std::shared_ptr that lives long enough to "
                         "extend the lifetime of the object pointed to "
-                        "beyond the timeframe that the Simulator object "
+                        "beyond the time frame that the Simulator object "
                         "keeps track of it."));
     return simulator->initial_temperature_manager;
   }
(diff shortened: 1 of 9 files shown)


source/simulator/simulator_access.cc
```diff
@@ -719,9 +719,44 @@ namespace aspect
...
+  template <int dim>
+  std::shared_ptr<const WorldBuilder::World>
+  SimulatorAccess<dim>::get_world_builder_pointer () const
+  {
+    Assert (simulator->world_builder.get() != nullptr,
+            ExcMessage ("You are trying to access the WorldBuilder "
+                        "object, but the Simulator object is not currently storing "
+                        "a pointer to such an object. This may be because you have "
+                        "not enabled WorldBuilder support during configuration of "
+                        "ASPECT. Alternatively, the Simulator object may no "
```
tjhei, line 749: This implementation is inside an ``#ifdef ASPECT_WITH_WORLDBUILDER``, so I don't think this situation can happen. Would it make sense to remove this sentence?

tjhei (approved):
Good to go otherwise.

MFraters (approved):
I agree with Timo's comment that this should not happen since it is in a ifdef block. On the other hand, the get_world_builder function also had this message before, even though it is in the ifdef statement. So this would only be the reason if for some reason the ifdef is removed or not configured correctly. I am personally fine with the text staying there as it it, but also if it is removed or changed. If it it changed I think the emphasis should be more on the second part (instead of `Alternatively`, something like, `More likely`) in both texts.

bangerth:
Yes, I copied that text. The idea was to have the documentation be independent of the `#ifdef`. I can go either way -- @tjhei your call!

tjhei:
I don't see the danger of the ifdef "not working" or somebody moving the function out. I would remove the sentence.

bangerth:
So updated!

tjhei (approved):

PR #4943, opened 2022-08-10
Update parameters.cc

### LINKED ISSUE #4496: Limit the growth of the time step size
@tjhei dug up a masters thesis at https://dc.uwm.edu/cgi/viewcontent.cgi?article=1405&context=etd that is about time step choices for BDF2. Among the conclusions there is that one should not let the time step size grow by too much from one time step to another. They call this grow the "amplification factor", and reference a number of papers that apparently show that it should not grow by more than either a factor of 1+sqrt(2) or 1.86 or 1.91 (pages 15 and 16), depending on who you trust. We should probably incorporate a limit in our code base to that end as well.

laila-16:

This pull request is meant to resolve #4496 by imposing a default upper bound on the rate of increase of the time step size.

* [X] I have read the guidelines in our [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) document.

### For all pull requests:


laila-16:
here we're changing the default value of the maximum relative increase to 86 in order to limit the growth of the step size

anne-glerum:
Hi @laila-16 , thank you for your PR. Could you explain why you think the default value should be changed? Is it not enough to be able to change the maximum timestep increase from the input file? By default, we do not want to restrict the timestep size increase unnecessarily.

laila-16:
> Hi @laila-16 , thank you for your PR. Could you explain why you think the default value should be changed? Is it not enough to be able to change the maximum timestep increase from the input file? By default, we do not want to restrict the timestep size increase unnecessarily.

according to the master thesis that was mentioned here https://github.com/geodynamics/aspect/issues/4496 , the growth of the time step should be limited by  a factor of 1.86, changing the default value will ensure this, unless if the user change it from the input file.

tjhei:
Hi @laila-16, that would have been useful to mention when you open up the pull request. :-) We typically describe what we are doing and more importantly why we are doing it in the pull request description at the very top.

The change makes sense, but we will have to do more work for this to work:
- This number needs to be documented in the parameter description
- We should do some tests by running a few of the problems in cookbooks/benchmarks to see how big the influence is
- Tests will need updating

anne-glerum:
> > Hi @laila-16 , thank you for your PR. Could you explain why you think the default value should be changed? Is it not enough to be able to change the maximum timestep increase from the input file? By default, we do not want to restrict the timestep size increase unnecessarily.
> 
> according to the master thesis that was mentioned here #4496 , the growth of the time step should be limited by a factor of 1.86, changing the default value will ensure this, unless if the user change it from the input file.

Thank you for the explanation and the link to the issue this PR addresses!

laila-16:
> > according to the master thesis that was mentioned here #4496 , the growth of the time step should be limited by a factor of 1.86, changing the default value will ensure this, unless if the user change it from the input file.
> 
> Thank you for the explanation and the link to the issue this PR addresses!
> 
> According to the issue, a factor of 1+sqrt(2) or 1.86 or 1.91 should be chosen. Why did you decide on 1.86?

Hi @anne-glerum ,
sorry for the late reply. Here is the reason why I decided to choose 1.91 instead of 1.86 as I mentioned previously .
The factor 1+sqrt(2) used for ODEs, while 1.86 and 1.91 used for PDEs. I looked into this a bit more, and the best available step-size ratio bound guaranteeing stability in the PDE context seems to be 1.91, from E. Emmrich (2006).  Although that bound was proved in the context of semilinear parabolic problems, and doesn't strictly apply here, it seems like the best available point of reference and a reasonable default.  I also ran a few tests comparing results with different values for this parameter and there was not a clear winner in terms of accuracy and efficiency.

source/simulator/parameters.cc
```diff
@@ -167,7 +167,7 @@ namespace aspect
                        "Units: Years or seconds, depending on the ``Use years "
                        "in output instead of seconds'' parameter.");
 
-    prm.declare_entry ("Maximum relative increase in time step", boost::lexical_cast<std::string>(std::numeric_limits<int>::max()),
+    prm.declare_entry ("Maximum relative increase in time step", "86.0",
                        Patterns::Double (0.),
                        "Set a percentage with which the time step is limited to increase. Generally the "
                        "time step based on the CFL number should be sufficient, but for complicated models "
```
bangerth, line 173: Would you mind adding a sentence to the end of the description of the parameter to explain where that 86% choice comes from? It will probably not be obvious to anyone why that is a useful choice, and we developers will likely forget about it as well.

More specifically, I think it would be useful if you could add the reasoning you provided in the comments in the pull request to the description of the parameter. People will read the manual where this information is replicated, but will not go back to the pull request where the functionality was added.

tests/convection_box_2D/screen-output.txt
```diff
@@ -0,0 +1,189 @@
+
+*** Timestep 1054:  t=0.492509 seconds, dt=0.000450269 seconds
+   Solving temperature system... 0 iterations.
+   Solving Stokes system... 1+0 iterations.
```
bangerth, line 4: This and the other files you are adding are for existing tests. What is the reason why you want to add these output files?

doc/modules/changes/change in Maximum relative increase.rtf
```diff
@@ -0,0 +1,42 @@
...
+\deftab720
+
+\itap1\trowd \taflags1 \trgaph108\trleft-108 \trbrdrt\brdrnil \trbrdrl\brdrnil \trbrdrr\brdrnil 
+\clmgf \clvertalt \clshdrawnil \clwWidth23260\clftsWidth3 \clbrdrt\brdrnil \clbrdrl\brdrnil \clbrdrb\brdrnil \clbrdrr\brdrnil \clpadl200 \clpadr200 \gaph\cellx4320
+\clmrg \clvertalt \clshdrawnil \clwWidth23260\clftsWidth3 \clbrdrt\brdrnil \clbrdrl\brdrnil \clbrdrb\brdrnil \clbrdrr\brdrnil \clpadl200 \clpadr200 \gaph\cellx8640
+\pard\intbl\itap1\pardeftab720\partightenfactor0
+
+\f0\fs24 \cf2 \cb3 \expnd0\expndtw0\kerning0
+The maximum relative increase is bound now by a factor of 1.91. Where before it was unbounded.\cb1 \cell 
+\pard\intbl\itap1\cell \row
```
bangerth, line 17: Instead of this .rtf file, would you mind using a text-only file? You can create the file with a regular text editor, rather than a word processor.

bangerth (changes requested):
Hi @laila-16 -- first, welcome to the ASPECT developer community! I left a couple of comments below about the specifics of your patch. Let us know if you have questions!

tests/convection_box_2D/screen-output.txt
laila-16: For the test case I understood that I have to add the output files to the tests directory. Should I do other things?

gassmoeller:
I think this was superseded by #5199. I will close it for now, but feel free to reopen the PR if it offers something in addition to #5199.

PR #4948, opened 2022-08-20
Add a new strain healing mechanism used for fracture (or fault) deactivation

sibiaoliu:
Since the strain healing function of deactivated fractures (faults) during shallow crustal deformation is missing in the code, I added a new strain healing mechanism following the approach of Poliakov & Buck (1998). This mechanism is widely used for crustal deformation at divergent margins, such as the formation and evolution of faults at mid-ocean ridges.

So far, I have only tested it in the visco_plastic model with turning on the "plastic weakening with plastic strain only".

### For all pull requests:


sibiaoliu:
Need to check for correct functionality before reopening.

PR #4954, opened 2022-09-08
Also output the yield stress

anne-glerum:
This PR add the current yield stress to the PlasticAdditionalOutputs.

### For all pull requests:


tjhei (approved):
Looks reasonable to me

anne-glerum:
Alright, all the failed tests output the `named additional outputs`, so they now include the yield stress. Therefore I've updated all these tests.

anne-glerum:
The two tests that are failing now, pass on my local machine. They use a shared lib material model that also has PlasticAdditionalOutputs, but which I haven't adapted. Should I?

tjhei:
> They use a shared lib material model that also has PlasticAdditionalOutputs

I think these are indeed clashing and causing the crash.

What is the point of having them in the tests, do you know? Can you try removing them? Otherwise, we should rename the class.

anne-glerum:
I think it's nice to have the PlasticAdditionalOutputs in the test, especially the `is_yielding` and the new `yield_strength` because the test's material model is a changed plasticity formulation. 

I've added the `yield_strength` so that the PlasticAdditionalOutputs in the tests are the same as in the source/material_model/visco_plastic. In this case, the tester runs (and fails because of the extra output field). 

I'll revert those changes and change the name of PlasticAdditionalOutputs to see if that also works.

gassmoeller (approved):
Looks like everything is working with the tests. Is there anything else you wanted to do in this PR before this can be merged?

anne-glerum:
Great! Ready then, unless you want me to squash.

PR #4965, opened 2022-09-23
Add water fugacity to diffusion creep

danieldouglas92:

* [ ] I have read the guidelines in our [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) document.

### For all pull requests:

### DIFF
--- a/source/material_model/rheology/diffusion_creep.cc
+++ b/source/material_model/rheology/diffusion_creep.cc
@@ -54,6 +54,11 @@ namespace aspect
             creep_parameters.activation_volume = activation_volumes_diffusion[composition];
             creep_parameters.stress_exponent = stress_exponents_diffusion[composition];
             creep_parameters.grain_size_exponent = grain_size_exponents_diffusion[composition];
+
+	    creep_parameters.wet_prefactor = wet_prefactors_diffusion[composition];
+	    creep_parameters.wet_activation_energy = wet_activation_energies_diffusion[composition];
+	    creep_parameters.wet_activation_volume = wet_activation_volumes_diffusion[composition];
+	    creep_parameters.water_fugacity_exponent = water_fugacity_exponents_diffusion[composition];
           }
         else
           {
@@ -68,6 +73,15 @@ namespace aspect
                                                stress_exponents_diffusion, composition);
             creep_parameters.grain_size_exponent = MaterialModel::MaterialUtilities::phase_average_value(phase_function_values, n_phases_per_composition,
                                                    grain_size_exponents_diffusion, composition);
+
+	    creep_parameters.wet_prefactor = MaterialModel::MaterialUtilities::phase_average_value(phase_function_values, n_phases_per_composition,
+                                         wet_prefactors_diffusion, composition,  MaterialModel::MaterialUtilities::PhaseUtilities::logarithmic);
+	    creep_parameters.wet_activation_energy = MaterialModel::MaterialUtilities::phase_average_value(phase_function_values, n_phases_per_composition,
+                                                 wet_activation_energies_diffusion, composition);
+	    creep_parameters.wet_activation_volume = MaterialModel::MaterialUtilities::phase_average_value(phase_function_values, n_phases_per_composition,
+                                                 wet_activation_volumes_diffusion, composition);
+	    creep_parameters.water_fugacity_exponent = MaterialModel::MaterialUtilities::phase_average_value(phase_function_values, n_phases_per_composition,
+                                               water_fugacity_exponents_diffusion, composition);
           }
         return creep_parameters;
       }
@@ -91,11 +105,34 @@ namespace aspect
         // A: prefactor,
         // d: grain size, m: grain size exponent, E: activation energy, P: pressure,
         // V; activation volume, R: gas constant, T: temperature.
-        double viscosity_diffusion = 0.5 / p.prefactor *
-                                     std::exp((p.activation_energy +
-                                               pressure*p.activation_volume)/
-                                              (constants::gas_constant*temperature)) *
-                                     std::pow(grain_size, p.grain_size_exponent);
+
+	double viscosity_diffusion = 0;
+
+	if (use_water_fugacity_diffusion)
+	  {
+	     AssertThrow(this->introspection().compositional_name_exists("water"),
+                         ExcMessage("The water fugacity rheology only works if "
+                                    "there is a compositional field called water."));
+            const unsigned int water_idx = this->introspection().compositional_index_for_name("water");
+            double water_content = this->introspection().component_indices.compositional_fields[water_idx];
+
+            viscosity_diffusion = 0.5 / (p.prefactor / std::pow(p.wet_prefactor, p.water_fugacity_exponent)) *
+		                         std::pow(water_content, -p.water_fugacity_exponent) *
+                                         std::exp(((p.activation_energy - p.wet_activation_energy*p.water_fugacity_exponent) +
+                                                   pressure*(p.activation_volume - p.wet_activation_volume*p.water_fugacity_exponent))/
+                                                  (constants::gas_constant*temperature)) *
+                                         std::pow(grain_size, p.grain_size_exponent);
+	  }
+
+	else
+	  {
+            viscosity_diffusion = 0.5 / p.prefactor *
+                                         std::exp((p.activation_energy +
+                                                   pressure*p.activation_volume)/
+                                                  (constants::gas_constant*temperature)) *
+                                         std::pow(grain_size, p.grain_size_exponent);
+	  }
+
 
         Assert (viscosity_diffusion > 0.0,
                 ExcMessage ("Negative diffusion viscosity detected. This is unphysical and should not happen. "
@@ -176,6 +213,35 @@ namespace aspect
                            "Units: \\si{\\meter\\cubed\\per\\mole}.");
         prm.declare_entry ("Grain size", "1e-3", Patterns::Double (0.),
                            "Units: \\si{\\meter}.");
+	prm.declare_entry ("Use water fugacity for diffusion creep", "false",
+			   Patterns::Bool(),
+			   "Whether or not to include water fugacity in the diffusion creep rheology. Setting this to false "
+			   "gives the default diffusion creep rheology. If set to true, a compositional field named 'water' "
+			   "must exist, and the material model will query the water content at each mesh point and include "
+			   "that in the calculation for the viscosty. Wet rheological parameters must also be specified.");
+	prm.declare_entry ("Wet prefactors for diffusion creep", "3e-24",
+			   Patterns::Anything(),
+			   "List of wet viscosity prefactors, $A^{wet}$, for background material and compositional fields, "
+			   "for a total of N+1 values, where N is the number of compositional fields. "
+			   "If only one value is given, then all use the same value. "
+			   "Units: \\si{\\per\\pascal\\meter}$^{m_{\\text{diffusion}}}$\\si{\\per\\second}.");
+	prm.declare_entry ("Wet activation energies for diffusion creep", "375e3",
+			   Patterns::Anything(),
+			   "List of wet activation energies, $E_a^{wet}, for background material and compositional fields, "
+			   "for a total of N+1 values, where N is the number of compositional fields. "
+			   "If only one value is given, then all use the same value. "
+			   "Units: \\si{\\joule\\per\\mole}.");
+	prm.declare_entry ("Wet activation volumes for diffusion creep", "23e-6",
+                           Patterns::Anything(),
+                           "List of wet activation volumes, $V_a^{wet}$, for background material and compositional fields, "
+                           "for a total of N+1 values, where N is the number of compositional fields. "
+                           "If only one value is given, then all use the same value. "
+                           "Units: \\si{\\meter\\cubed\\per\\mole}.");
+	prm.declare_entry ("Water fugacity exponents for diffusion creep", "1.0",
+			   Patterns::Anything(),
+			   "List of water fugacity exponents, $r$, for background material and compositional fields, "
+			   "for a total of N+1 values, where N is the number of compositional fields. "
+			   "If only one value is given, then all use the same value. Units: None.");
       }
 
 
@@ -222,7 +288,33 @@ namespace aspect
                                                                             "Activation volumes for diffusion creep",
                                                                             true,
                                                                             expected_n_phases_per_composition);
+	wet_prefactors_diffusion = Utilities::parse_map_to_double_array(prm.get("Wet prefactors for diffusion creep"),
+                                                                       list_of_composition_names,
+                                                                       has_background_field,
+                                                                       "Wet prefactors for diffusion creep",
+                                                                       true,
+                                                                       expected_n_phases_per_composition);
+	wet_activation_energies_diffusion = Utilities::parse_map_to_double_array(prm.get("Wet activation energies for diffusion creep"),
+                                                                                list_of_composition_names,
+                                                                                has_background_field,
+                                                                                "Wet activation energies for diffusion creep",
+                                                                                true,
+                                                                                expected_n_phases_per_composition);
+	wet_activation_volumes_diffusion = Utilities::parse_map_to_double_array(prm.get("Wet activation volumes for diffusion creep"),
+                                                                               list_of_composition_names,
+                                                                               has_background_field,
+                                                                              "Wet activation volumes for diffusion creep",
+                                                                               true,
+                                                                               expected_n_phases_per_composition);
+	water_fugacity_exponents_diffusion = Utilities::parse_map_to_double_array(prm.get("Water fugacity exponents for diffusion creep"),
+									          list_of_composition_names,
+										  has_background_field,
+										  "Water fugacity exponents for diffusion creep",
+										  true,
+										  expected_n_phases_per_composition);
         grain_size = prm.get_double("Grain size");
+
+	use_water_fugacity_diffusion = prm.get_bool("Use water fugacity for diffusion creep");
 
         // Check that there are no entries set to zero,
         // for example because the entry is for a field
--- a/include/aspect/material_model/rheology/diffusion_creep.h
+++ b/include/aspect/material_model/rheology/diffusion_creep.h
@@ -48,6 +48,11 @@ namespace aspect
         double activation_volume;
         double stress_exponent;
         double grain_size_exponent;
+
+	double wet_prefactor;
+	double wet_activation_energy;
+	double wet_activation_volume;
+	double water_fugacity_exponent;
       };
 
       template <int dim>
@@ -121,7 +126,7 @@ namespace aspect
         private:
 
           /**
-           * List of diffusion creep prefactors A.
+           * List of dry diffusion creep prefactors A_dry.
            */
           std::vector<double> prefactors_diffusion;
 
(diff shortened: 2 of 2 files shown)


source/material_model/rheology/diffusion_creep.cc
```diff
@@ -91,11 +105,34 @@ namespace aspect
...
+
+	else
+	  {
+            viscosity_diffusion = 0.5 / p.prefactor *
+                                         std::exp((p.activation_energy +
+                                                   pressure*p.activation_volume)/
+                                                  (constants::gas_constant*temperature)) *
+                                         std::pow(grain_size, p.grain_size_exponent);
+	  }
+
```
bobmyhill, line 135: The `compute_strain_rate_and_derivative` function also needs to be modified.

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -222,8 +288,34 @@ namespace aspect
...
+                                                                               expected_n_phases_per_composition);
+	water_fugacity_exponents_diffusion = Utilities::parse_map_to_double_array(prm.get("Water fugacity exponents for diffusion creep"),
+									          list_of_composition_names,
+										  has_background_field,
+										  "Water fugacity exponents for diffusion creep",
+										  true,
+										  expected_n_phases_per_composition);
         grain_size = prm.get_double("Grain size");
 
+	use_water_fugacity_diffusion = prm.get_bool("Use water fugacity for diffusion creep");
```
bobmyhill, line 317: This change to the rheology needs to be documented somewhere.

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -91,11 +105,34 @@ namespace aspect
...
-                                     std::exp((p.activation_energy +
-                                               pressure*p.activation_volume)/
-                                              (constants::gas_constant*temperature)) *
-                                     std::pow(grain_size, p.grain_size_exponent);
+
+	double viscosity_diffusion = 0;
+
+	if (use_water_fugacity_diffusion)
+	  {
+	     AssertThrow(this->introspection().compositional_name_exists("water"),
```
bobmyhill, line 113: I'm not sure "water" is a good name for the compositional field. Water is a free phase, not a chemical component in a solid. Perhaps "H2O concentration" would be more appropriate?

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -91,11 +105,34 @@ namespace aspect
...
+                         ExcMessage("The water fugacity rheology only works if "
+                                    "there is a compositional field called water."));
+            const unsigned int water_idx = this->introspection().compositional_index_for_name("water");
+            double water_content = this->introspection().component_indices.compositional_fields[water_idx];
+
+            viscosity_diffusion = 0.5 / (p.prefactor / std::pow(p.wet_prefactor, p.water_fugacity_exponent)) *
+		                         std::pow(water_content, -p.water_fugacity_exponent) *
```
bobmyhill, line 120: I'd suggest changing the sign of `p.water_fugacity_exponent`. It makes little sense to raise water concentration to a negative of a fugacity exponent.

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -91,11 +105,34 @@ namespace aspect
...
+
+	double viscosity_diffusion = 0;
+
+	if (use_water_fugacity_diffusion)
+	  {
+	     AssertThrow(this->introspection().compositional_name_exists("water"),
+                         ExcMessage("The water fugacity rheology only works if "
+                                    "there is a compositional field called water."));
+            const unsigned int water_idx = this->introspection().compositional_index_for_name("water");
+            double water_content = this->introspection().component_indices.compositional_fields[water_idx];
```
bobmyhill, line 117:
```suggestion
            double water_concentration = this->introspection().component_indices.compositional_fields[water_idx];
```

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -176,6 +213,35 @@ namespace aspect
...
+			   "for a total of N+1 values, where N is the number of compositional fields. "
+			   "If only one value is given, then all use the same value. "
+			   "Units: \\si{\\joule\\per\\mole}.");
+	prm.declare_entry ("Wet activation volumes for diffusion creep", "23e-6",
+                           Patterns::Anything(),
+                           "List of wet activation volumes, $V_a^{wet}$, for background material and compositional fields, "
+                           "for a total of N+1 values, where N is the number of compositional fields. "
+                           "If only one value is given, then all use the same value. "
+                           "Units: \\si{\\meter\\cubed\\per\\mole}.");
+	prm.declare_entry ("Water fugacity exponents for diffusion creep", "1.0",
```
bobmyhill, line 240: If you do change the sign of this exponent in the equations above, the default value should be -1, rather than 1.

bobmyhill (changes requested):
Looks pretty good to me.

A few minor comments. 

 A good place to document the rheology would be in the comment block in `compute_strain_rate_and_derivative` (the code in that block also needs updating). You should also provide a literature reference for the rheology.
 
An important note: this (common) formulation implies that H2O is a vital chemical component in allowing minerals to deform. Using this formulation, as H2O concentration -> 0, viscosity -> infinity. This is not a reasonable assertion. However, the formulation is empirically justified within a limited range of H2O concentrations. You should probably add a note to say that the `H2O concentration` compositional field should always have positive values to avoid infinite viscosities.

bobmyhill:
You will also need to:
- run the indent script after making the changes
- provide a new testcase and
- add a changelog entry.

naliboff:
> An important note: this (common) formulation implies that H2O is a vital chemical component in allowing minerals to deform. Using this formulation, as H2O concentration -> 0, viscosity -> infinity. This is not a reasonable assertion. However, the formulation is empirically justified within a limited range of H2O concentrations. You should probably add a note to say that the H2O concentration compositional field should always have positive values to avoid infinite viscosities.

source/material_model/rheology/diffusion_creep.cc
danieldouglas92: Wouldn't this require inputting a negative water_fugacity_exponent in the parameter file though? It seems like mathematically it'll do the same thing but it'll be more confusing when inputting values for models

danieldouglas92:
I updated the code and addressed your comments @bobmyhill but at the moment the method John and I talked about required adding an argument to compute_viscosity that I hardcoded to be 1 inside rheology/composite_visco_plastic.cc and diffusion_dislocation.cc, definitely not ideal but thought it would be useful to push anyway to see if you had any thoughts on an approach

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -91,11 +106,41 @@ namespace aspect
...
+	// V' = V - r*V_H2O
+	// A_H2O: prefactor for water component
+	// E_H2O: activation energy for water component, V_H2O: activation volume for water component
+	// C_H2O: water concentration in ppm H/Si, r: water fugacity exponent
+
+	double viscosity_diffusion = 0;
+
+	if (use_water_fugacity_diffusion)
+	  {
+	     AssertThrow(this->introspection().compositional_name_exists("water_concentration"),
```
bobmyhill, line 123: This assert should go into parse_parameters.

source/material_model/rheology/diffusion_creep.cc
bobmyhill: Mea culpa, I misread the equation. It's common to group the fugacity term with the prefactor (which is already part of the denominator of the first term), and if you did that, the negative sign would go away.

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -91,11 +106,41 @@ namespace aspect
...
-                                     std::exp((p.activation_energy +
-                                               pressure*p.activation_volume)/
-                                              (constants::gas_constant*temperature)) *
-                                     std::pow(grain_size, p.grain_size_exponent);
+	//
+	// If use_water_fugacity_diffusion = true this equation is modified:
+	//    viscosity = 0.5 * A'^(-1) * d^(m) * C_H2O^(-r) * exp((E' + P*V')/(RT))
+	// A' = A / A_H2O^(r)
+	// E' = E - r*E_H2O
+	// V' = V - r*V_H2O
```
bobmyhill, line 114: Now that I see this equation written down, most of the terms are pointless. A', E' and V' are all constants, so why not just require the user to put those in the prm file rather than make ASPECT calculate them billions of times?

I think the only term that needs to be in here is C_H2O^(-r), which is the standard term found in Mei and Kohlstedt (https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2000JB900179), among others.

bobmyhill: That would mean that you could just add a single new creep parameter (creep_parameters.water_fugacity_exponent), rather than four.

source/material_model/rheology/diffusion_creep.cc
bobmyhill: > I updated the code and addressed your comments @bobmyhill but at the moment the method John and I talked about required adding an argument to compute_viscosity that I hardcoded to be 1 inside rheology/composite_visco_plastic.cc and diffusion_dislocation.cc, definitely not ideal but thought it would be useful to push anyway to see if you had any thoughts on an approach

I agree that it isn't a beautiful solution, but I don't think it's a bad choice. The alternative would be to make a separate "wet diffusion" rheology, but that doesn't seem any better.

source/material_model/rheology/diffusion_creep.cc
bobmyhill: Sorry, in the cold light of day my comment from last night seems overly harsh. I could potentially see modifications to A, E and V being necessary, but perhaps via a relationship with `C_H2O`, rather than `r`? If you have a reference for this formulation, that would be great :)

danieldouglas92: No need to apologize I think you make a great point! The equations could only be modified to include the water concentration and if you just calculate the reduction in the creep parameters  before hand then only the water_fugacity_exponent would need to be defined and that would save a lot of lines of code. I don't have a formulation for modifying A, E or V with variable C_H2O but I'll see what I can dig up!

danieldouglas92: I didn't find a formulation for how to vary the creep parameters with varying C_H2O so I just simplified the existing code and included the same thing for dislocation creep, if you do know or come across something where A, E and V can be modified with C_H2O let me know! That would be very cool to try to implement

source/material_model/diffusion_dislocation.cc
```diff
@@ -84,8 +84,8 @@ namespace aspect
                  && stress_iteration < stress_max_iteration_number)
             {
 
-              const std::pair<double, double> diff_edot_and_deriv = diffusion_creep.compute_strain_rate_and_derivative(stress_ii, pressure, temperature, diffusion_creep_parameters);
-              const std::pair<double, double> disl_edot_and_deriv = dislocation_creep.compute_strain_rate_and_derivative(stress_ii, pressure, temperature, dislocation_creep_parameters);
+              const std::pair<double, double> diff_edot_and_deriv = diffusion_creep.compute_strain_rate_and_derivative(stress_ii, pressure, temperature, 1., diffusion_creep_parameters);
```
naliboff, line 87:
```suggestion
              const std::pair<double, double> diff_edot_and_deriv = diffusion_creep.compute_strain_rate_and_derivative(stress_ii, pressure, temperature, water_concentration, diffusion_creep_parameters);
```

Would this work since `1.` is the default value for the parameter?

naliboff: Same for places where `1.` is defined instead of `water_concentration`.

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -68,17 +69,20 @@ namespace aspect
...
+                                               water_fugacity_exponents_diffusion, composition);
           }
         return creep_parameters;
       }
 
 
 
       template <int dim>
-      double
+      double // Add position to args of this function
```
naliboff, line 81: Does this function still need the position argument for this PR? If yes, can you add in the next commit? If not, remove comment?

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -91,11 +95,19 @@ namespace aspect
...
+	// If use_water_fugacity_diffusion = true this equation is modified:
+	//    viscosity = 0.5 * A^(-1) * d^(m) * C_H2O^(-r) * exp((E + P*V)/(RT))
+	// C_H2O: water concentration in ppm H/Si, r: water fugacity exponent.
+
         double viscosity_diffusion = 0.5 / p.prefactor *
-                                     std::exp((p.activation_energy +
-                                               pressure*p.activation_volume)/
-                                              (constants::gas_constant*temperature)) *
-                                     std::pow(grain_size, p.grain_size_exponent);
+                                           std::exp((p.activation_energy +
```
naliboff, line 104: Was this indentation manual, or done via astyle?

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -176,6 +198,17 @@ namespace aspect
                            "Units: \\si{\\meter\\cubed\\per\\mole}.");
         prm.declare_entry ("Grain size", "1e-3", Patterns::Double (0.),
                            "Units: \\si{\\meter}.");
+	prm.declare_entry ("Use water fugacity in diffusion creep", "false",
+			   Patterns::Bool(),
+			   "Whether or not to include water fugacity in the diffusion creep rheology. Setting this to false "
+			   "gives the default diffusion creep rheology. If set to true, a compositional field named 'water' "
+			   "must exist, and the material model will query the water content at each mesh point and include "
+			   "that in the calculation for the viscosty. Wet rheological parameters must also be specified.");
```
naliboff, line 206:
```suggestion
			   "that in the calculation for the viscosity. Flow law parameters modified for the effects "
			   "hydration state must be specified in the parameter file.");
```

source/material_model/rheology/diffusion_creep.cc
```diff
@@ -176,6 +198,17 @@ namespace aspect
                            "Units: \\si{\\meter\\cubed\\per\\mole}.");
         prm.declare_entry ("Grain size", "1e-3", Patterns::Double (0.),
                            "Units: \\si{\\meter}.");
+	prm.declare_entry ("Use water fugacity in diffusion creep", "false",
+			   Patterns::Bool(),
+			   "Whether or not to include water fugacity in the diffusion creep rheology. Setting this to false "
+			   "gives the default diffusion creep rheology. If set to true, a compositional field named 'water' "
```
naliboff, line 204:
```suggestion
			   "gives the default diffusion creep rheology. If set to true, a compositional field named 'water_concentration' "
```

source/material_model/rheology/dislocation_creep.cc
```diff
@@ -90,10 +95,17 @@ namespace aspect
...
+	//
+	// If use_water_fugacity_dislocation = true, this equation is modified:
+	//    viscosity = 0.5 * A^(-1/n) * C_H2O^(r) *  edot_ii^((1-n)/n) * exp((E + P*V)/(nRT))
+	// C_H2O: water concentration in ppm H/Si, r: water fugacity exponent.
+
         double viscosity_dislocation = 0.5 * std::pow(p.prefactor,-1/p.stress_exponent) *
-                                       std::exp((p.activation_energy + pressure*p.activation_volume)/
-                                                (constants::gas_constant*temperature*p.stress_exponent)) *
-                                       std::pow(strain_rate,((1. - p.stress_exponent)/p.stress_exponent));
+                                      std::exp((p.activation_energy + pressure*p.activation_volume)/
```
naliboff, line 104: Indentation update from astyle?

source/material_model/rheology/dislocation_creep.cc
```diff
@@ -169,6 +194,17 @@ namespace aspect
                            "for a total of N+1 values, where N is the number of compositional fields. "
                            "If only one value is given, then all use the same value. "
                            "Units: \\si{\\meter\\cubed\\per\\mole}.");
+        prm.declare_entry ("Use water fugacity in dislocation creep", "false",
+                           Patterns::Bool(),
+                           "Whether or not to include water fugacity in the dislocation creep rheology. Setting this to false "
+                           "gives the default dislocation creep rheology. If set to true, a compositional field named 'water' "
```
naliboff, line 200:
```suggestion
                           "gives the default dislocation creep rheology. If set to true, a compositional field named 'water_concentration' "
```

source/material_model/rheology/dislocation_creep.cc
```diff
@@ -169,6 +194,17 @@ namespace aspect
...
+                           "must exist, and the material model will query the water content at each mesh point and include "
```
naliboff, line 201:
```suggestion
                           "must exist, which is used to modify the dislocation creep viscosity."
```

tests/visco_plastic_diffusion_dislocation_composite_water_fugacity.prm
```diff
@@ -0,0 +1,137 @@
+# A test for composite creep in the visco_plastic material model which includes effects from
```
naliboff, line 1:
```suggestion
# A test for composite creep in the visco_plastic material model,  which includes the effects of
```

tests/visco_plastic_diffusion_dislocation_composite_water_fugacity.prm
```diff
@@ -0,0 +1,137 @@
...
+# water fugacity. The composite creep viscosity (vis_composite) is defined as:
+#   (vis_diffusion * vis_dislocation)/(vis_diffusion + vis_dislocation)
+# Where vis_diffusion and vis_dislocation are the normal diffusion viscosity and 
+# dislocation viscosity multiplied by the concentration of water given as H ppm Si.
+# In this model, half of the box is hydrated with a water concentratin of 100 H ppm Si,
+# and the other half of the box has 0 H ppm Si. The viscosity is therefore reduced in
+# the region where there is water relative to the region that has no water.
```
naliboff, line 8: Can you include a section of comments here (or in the Material Model section), on how the flow law parameters were modified to account from water concentration.

naliboff: Can you also include in the comments the calculation of what water viscosity should be expected.

tests/visco_plastic_diffusion_dislocation_composite_water_fugacity.prm
```diff
@@ -0,0 +1,137 @@
...
+  end
+end
+
+# Post processing
+subsection Postprocess
+  set List of postprocessors = velocity statistics, mass flux statistics, visualization
+  subsection Visualization
+    set Interpolate output = false
+    set List of output variables = viscosity, strain rate
+    set Output format            = gnuplot
```
naliboff, line 129:
```suggestion
    set Output format            = gnuplot
    set Number of grouped files   = 1
```

source/material_model/rheology/diffusion_creep.cc
naliboff: @bobmyhill - Are you ok with the current approach (user modifies values in PRM file), and then modify the flow law parameters automatically later on if we find a robust relationship with C_H20?

naliboff (changes requested):
@danieldouglas92 - thanks for the contribution and this looks to be in pretty good shape. Only few minor comments here and there, largely about formatting and documentation.

bobmyhill:
@danieldouglas92 did you want to take another look at this, or shall we close it?

danieldouglas92:
This functionality has been merged in #5678.

PR #4966, opened 2022-09-25
GMG:clear out mg constraints

zjiaqi2018:
1. Remove `user_level_constraints` and `level_constraints` and use `mg_level_constraints` to store boundary constraints on all levels. 
2. Use `level_boundary_indices` to store zero boundary constraints, and then add to `mg_level_constraints`.
3. A new function to import  `mg_level_constraints` to `mg_constrained_dofs`.


zjiaqi2018:
deal.ii 9.3 doesn't have the following function
`mg_constrained_dofs.add_boundary_indices()`

tjhei:
Can you try https://github.com/geodynamics/aspect/pull/4413/commits/fa3b677fe03f5e42029690ba16bd36e2dce9471c or something similar for <9.5? I am curious if any tests are affected.

zjiaqi2018:
> Can you try [fa3b677](https://github.com/geodynamics/aspect/commit/fa3b677fe03f5e42029690ba16bd36e2dce9471c) or something similar for <9.5? I am curious if any tests are affected.

So you want to clear the level constraints before adding new ones?

tjhei:
> So you want to clear the level constraints before adding new ones?

Ideally, we would only clear out the user constraints, but the next best thing (as this is not available unless we use master) would be to clear and redo all constraints. Does that makes sense?

bangerth:
@tjhei What do we do about this PR from last year?

tjhei:
Will look at it again this hackathon

gassmoeller:
@tjhei: What about this hackathon? ;-)

PR #4968, opened 2022-09-28
Update the default value of the Newton maximum Stokes sovler tolerance to a more conservative value.

MFraters:
Update the default value of the Newton maximum Stokes sovler tolerance to a more conservative value, as discussed during the user meeting. The current default value is too loose for most real world applications.


tjhei (approved):

tjhei:
You might need to update ``checkpoint_05_mpi_resume``

MFraters:
It looks like the testers now all pass, except for the deal.ii master tester

PR #4973, opened 2022-10-02
always initialize dst vector in preconditioner

### LINKED ISSUE #4960: deal.II master floating point exception
This currently happens:
```
solver_scheme_no_advection_single_stokes ...........................***Failed    1.48 sec
*** "Test solver_scheme_no_advection_single_stokes failed": ***
[1/1] Generating output-solver_scheme_no_advection_single_stokes/screen-output
FAILED: output-solver_scheme_no_advection_single_stokes/screen-output 
cd /__w/aspect/aspect/build/tests && /__w/aspect/aspect/tests/cmake/timeout.sh 600s /__w/aspect/aspect/tests/cmake/run_test.sh 0 /__w/aspect/aspect/build/aspect /__w/aspect/aspect/build/tests/solver_scheme_no_advection_single_stokes.x.prm /__w/aspect/aspect/build/tests/output-solver_scheme_no_advection_single_stokes/screen-output.tmp && /__w/aspect/aspect/tests/cmake/default screen-output </__w/aspect/aspect/build/tests/output-solver_scheme_no_advection_single_stokes/screen-output.tmp >/__w/aspect/aspect/build/tests/output-solver_scheme_no_advection_single_stokes/screen-output
/__w/aspect/aspect/tests/cmake/run_test.sh: line 16: 29001 Floating point exception(core dumped) $BINARY $PRM > ${OUTPUT} 2>&1
--     . version 2.5.0-pre
--     . using deal.II 9.5.0-pre (master, 3e88b26172)
--     . using Trilinos 12.14.1
--     . using p4est 2.2.0

Loading shared library <./libsolver_scheme_no_advection_single_stokes.so>

Vectorization over 2 doubles = 128 bits (SSE2), VECTORIZATION_LEVEL=1
-- For information on how to cite ASPECT, see:
--   https://aspect.geodynamics.org/citing.html?ver=2.5.0-pre&mf=1&sha=&src=code
Number of active cells: 256 (on 5 levels)
Number of degrees of freedom: 3,556 (2,178+289+1,089)

*** Timestep 0:  t=0 seconds, dt=0 seconds
   Solving Stokes system... 0+
```

see here for example: https://github.com/geodynamics/aspect/actions/runs/3090660726/jobs/4999751630#step:6:1627

tjhei:
Current deal.II master does not initialize the vector in all cases and we might up using invalid/random data.

fixes #4960

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -249,6 +249,11 @@ namespace aspect
           ptmp.reinit(src);
         }
 
+      // This needs to be done explicitly, as GMRES does not
+      // initialize the data of the vector dst before calling
+      // us. Otherwise we might use random data as our initial guess.
+      dst = 0.0;
+
       // either solve with the Schur complement matrix (if do_solve_Schur_complement==true)
       // or just apply one preconditioner sweep (for the first few
       // iterations of our two-stage outer GMRES iteration)
@@ -269,7 +274,6 @@ namespace aspect
             {
               try
                 {
-                  dst.block(1) = 0.0;
                   solver.solve(Schur_complement_block,
                                dst.block(1), src.block(1),
                                Schur_complement_preconditioner);
@@ -313,7 +317,6 @@ namespace aspect
           SolverCG<dealii::LinearAlgebra::distributed::Vector<double>> solver(solver_control);
           try
             {
-              dst.block(0) = 0.0;
               solver.solve(A_block, dst.block(0), utmp.block(0),
                            A_block_preconditioner);
               n_iterations_A_ += solver_control.last_step();


gassmoeller (approved):
What a hidden bug. Thanks for fixing.

source/simulator/stokes_matrix_free.cc
```diff
@@ -249,6 +249,11 @@ namespace aspect
           ptmp.reinit(src);
         }
 
+      // This needs to be done explicitly, as GMRES does not
+      // initialize the data of the vector dst before calling
+      // us. Otherwise we might use random data as our initial guess.
+      dst = 0.0;
```
bangerth, line 255: This is of course a feature: If you start with the solution vector, GMRES should be done after one (or zero) iterations.

bangerth:
Out of curiosity, how does this actually change anything? Both components of the vector are initialized at some point; the new code line introduced in this patch just moves this into a single place.

tjhei:
> This is of course a feature: If you start with the solution vector, GMRES should be done after one (or zero) iterations.

I don't think what you write has anything to do with the situation here:
The function in question is ``Precondioner::vmult(dst, src)`` where ``dst`` is a pure "out" parameter and contains no useful information. As the ``Solver::solve`` calls _do_ use the current value in ``dst`` as a starting value, we need to make sure to initialize them to zero before calling solve (this was already the case before this PR).

The problem this PR fixes is that we missed one case:
If the ``if``
https://github.com/tjhei/aspect/blob/b5cd64df7f85c74c56fdc83fb228cdacb13fd31c/source/simulator/stokes_matrix_free.cc#L273

fails, we never initialize ``dst.block(1)`` and then continue computing ``*= -1`` on random data:
https://github.com/tjhei/aspect/blob/b5cd64df7f85c74c56fdc83fb228cdacb13fd31c/source/simulator/stokes_matrix_free.cc#L301
This caused a floating point exception (or would succeed and fail later).

Now, instead of putting the initialization in an ``else`` branch, I decided to unconditionally initialize the vector at the beginning. I don't think we are wasting any cycles this way.

PR #4974, opened 2022-10-05
Add rk2 particle integrator first order accuracy option

gassmoeller:
An additional option for the RK2 particle integrator that I would like to have included for some benchmarks and a paper. I could have made another integrator, but that would be literally copy-paste of the existing one except for the one line, so I thought this is easier. I will open PRs for the benchmarks that use this option later.


tjhei (approved):

tjhei:
Is there any reason to use the new option? Speed perhaps?

gassmoeller:
> Speed perhaps?

In the current implementation this is actually not much faster than the normal RK2 because both velocities are evaluated, one is just ignored inside the integrator. Anyway if speed is the most important factor the forward euler integrator is the way to go.

tjhei:
Do you know what is going on with the github action testers? (good to go otherwise)

gassmoeller:
I dont know why they didnt trigger, but updating the PR triggered it and succeeded.

PR #4975, opened 2022-10-06
Add CPO DRex computation

MFraters:
Add initial D-Rex code as close as possible to the ASPECT CPO/ D-Rex paper, but with a different random number generator. This also already include the fabric selector based on Karato 2008.

I had to think a bit about the structure to both keep the main code and unit test code reasonable. I also thought about generalizing the current 4 slip planes to n slip planes (it is an actual use case), but I think it is better to leave that for now as it should not be hard to do later.

I also added a CPO cite tag to the initialization and the references to the ASPECT CPO/D-Rex paper. Since D-Rex is a crucial part of it, I am considering also adding the 2 D-Rex papers to the list of requested citations. Is there any policy on that?

(point 2 in #3885)

### DIFF
--- a/include/aspect/particle/property/crystal_preferred_orientation.h
+++ b/include/aspect/particle/property/crystal_preferred_orientation.h
@@ -39,12 +39,9 @@ namespace aspect
        * @brief The type of deformation used by the CPO code.
        *
        * passive: Only to be used with the spin tensor CPO Derivative algorithm.
-       * olivine_a_fabric: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a Olivine A Fabric, which influences the relative strength of the slip planes.
-       * olivine_b_fabric: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a Olivine B Fabric, which influences the relative strenght of the slip planes.
-       * olivine_c_fabric: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a Olivine C Fabric, which influences the relative strenght of the slip planes.
-       * olivine_d_fabric: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a Olivine D Fabric, which influences the relative strenght of the slip planes.
-       * olivine_e_fabric: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a Olivine E Fabric, which influences the relative strenght of the slip planes.
-       * enstatite: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a enstatite Fabric, which influences the relative strenght of the slip planes.
+       * olivine_a_fabric to olivine_e_fabric: Only to be used with the D-Rex CPO Derivative algorithm.
+       *  Sets the deformation type of the mineral to a Olivine A-E Fabric, which influences the relative strength of the slip planes. See table 1 in Fraters and Billen (2021).
+       * enstatite: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a enstatite Fabric, which influences the relative strength of the slip planes.
        */
       enum class DeformationType
       {
@@ -62,7 +59,7 @@ namespace aspect
        * passive: Only to be used with the spin tensor CPO Derivative algorithm.
        * olivine_a_fabric to olivine_e_fabric: Only to be used with the D-Rex CPO Derivative algorithm.
        *  Sets the deformation type of the mineral to a Olivine A-E Fabric, which influences the relative strength of the slip planes. See table 1 in Fraters and Billen (2021).
-       * enstatite: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a enstatite Fabric, which influences the relative strenght of the slip planes.
+       * enstatite: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a enstatite Fabric, which influences the relative strength of the slip planes.
        * olivine_karato_2008: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a olivine fabric based on the table in Karato 2008.
        */
       enum class DeformationTypeSelector
@@ -215,14 +212,19 @@ namespace aspect
           /**
            * @brief Computes the volume fraction and grain orientation derivatives of all the grains of a mineral.
            *
-           * @param cpo_index The index in the particle data array where the cpo data starts
-           * @param data The particle data array.
-           * @param mineral_i The mineral index for which to compute the derivatives.
-           * @param strain_rate is the strain-rate at the location of the particle.
-           * @param velocity_gradient_tensor is the velocity gradient tensor at the location of the particle.
-           * @param ref_resolved_shear_stress is the reference resolved shear stress of the mineral.
-           * @return A pair containing the derivatives for
-           * the change is size in the first part and the derivatives for the change in rotation in the second part.
+           * @param cpo_index The location where the CPO data starts in the data array.
+           * @param data The data array containing the CPO data.
+           * @param mineral_i The mineral for which to compute the derivatives for.
+           * @param strain_rate_3d The 3D strain rate at the location where the derivative is requested.
+           * @param velocity_gradient_tensor The velocity gradient tensor at the location where the derivative is requested.
+           * @param position the location for which the derivative is requested.
+           * @param temperature The temperature at the location where the derivative is requested.
+           * @param pressure The pressure at the location where the derivative is requested.
+           * @param velocity The veloicty at the location where the derivative is requested.
+           * @param compositions The compositios at the location where the derivative is requested.
+           * @param strain_rate The strain-rate at the location where the derivative is requested.
+           * @param deviatoric_strain_rate The deviatoric strain-rate at the location where the derivative is requested.
+           * @param water_content The water content at the location where the derivative is requested.
            */
           std::pair<std::vector<double>, std::vector<Tensor<2,3>>>
           compute_derivatives(const unsigned int cpo_index,
@@ -236,10 +238,24 @@ namespace aspect
                               const Tensor<1,dim> &velocity,
                               const std::vector<double> &compositions,
                               const SymmetricTensor<2,dim> &strain_rate,
-                              const SymmetricTensor<2,dim> &compressible_strain_rate,
+                              const SymmetricTensor<2,dim> &deviatoric_strain_rate,
                               const double water_content) const;
 
-
+          /**
+           * @brief Computes the CPO derivatives with the D-Rex 2004 algorithm.
+           *
+           * @param cpo_index The location where the CPO data starts in the data array.
+           * @param data The data array containing the CPO data.
+           * @param mineral_i The mineral for which to compute the derivatives for.
+           * @param strain_rate_3d The 3D strain rate
+           * @param velocity_gradient_tensor The velocity gradient tensor
+           * @param ref_resolved_shear_stress Represent one value per slip plane.
+           * The planes are ordered from weakest to strongest with relative values,
+           * where the inactive plane is infinity strong. So it is a measure of strength
+           * on each slip plane.
+           * @param prevent_nondimensionalization Prevent nondimensializing values internally.
+           * Only for unit testing purposes.
+           */
           std::pair<std::vector<double>, std::vector<Tensor<2,3>>>
           compute_derivatives_drex_2004(const unsigned int cpo_index,
                                         const ArrayView<double> &data,
@@ -277,6 +293,12 @@ namespace aspect
 
           /**
            * @brief Determines the deformation type from the deformation type selector.
+           * If the provided @p deformation_type_selector is a specific deformation type,
+           * the function will simply return the corresponding deformation type. However,
+           * if the @p deformation_type_selector is an algorithm to determine the current
+           * deformation type (e.g. based on measured lab data or analytical models), then
+           * the function computes the appropriate deformation type at the given conditions
+           * and returns the compute deformation type.
            */
           DeformationType
           determine_deformation_type(const DeformationTypeSelector deformation_type_selector,
@@ -286,7 +308,7 @@ namespace aspect
                                      const Tensor<1,dim> &velocity,
                                      const std::vector<double> &compositions,
                                      const SymmetricTensor<2,dim> &strain_rate,
-                                     const SymmetricTensor<2,dim> &compressible_strain_rate,
+                                     const SymmetricTensor<2,dim> &deviatoric_strain_rate,
                                      const double water_content) const;
 
           /**
@@ -302,6 +324,9 @@ namespace aspect
            *
            * The inactive plane should theoretically be infinitly strong, but this is nummerically not desirable,
            * so an optional max_value can be set to indicate an inactive plane.
+           *
+           * It is currently designed to return the relative strength of the slip planes for olivine, which are are 4,
+           * but this could be generalized.
            */
           std::array<double,4>
           reference_resolved_shear_stress_from_deformation_type(DeformationType deformation_type,
@@ -527,6 +552,11 @@ namespace aspect
           unsigned int n_minerals;
 
           /**
+           * The index of the water composition.
+           */
+          unsigned int water_index;
+
+          /**
            * A vector containing the deformation type selectors provided by the user.
            * Should be one of the following: "Olivine: Karato 2008", "Olivine: A-fabric",
            * "Olivine: B-fabric", "Olivine: C-fabric", "Olivine: D-fabric", "Olivine: E-fabric",
@@ -579,7 +609,7 @@ namespace aspect
            * efficiency of nucleation parameter.
            * lambda_m in equation 8 of Kaminski et al. (2004, Geophys. J. Int)
            */
-          double nucleation_efficientcy;
+          double nucleation_efficiency;
 
           /**
            * An exponent described in equation 10 of Kaminski and Ribe (2001, EPSL)
@@ -587,12 +617,15 @@ namespace aspect
           double exponent_p;
 
           /**
-           * The Grain Boundary Sliding threshold
+           * The Dimensionless Grain Boundary Sliding (GBS) threshold.
+           * This is a grain size threshold below which grain deform by GBS and
+           * become strain-free grains.
            */
           double threshold_GBS;
 
           /**
-           * grain boundary mobility
+           * Dimensionless grain boundary mobility as described by equation 14
+           * in Kaminski and Ribe (2001, EPSL).
            */
           double mobility;
 
--- a/source/particle/property/crystal_preferred_orientation.cc
+++ b/source/particle/property/crystal_preferred_orientation.cc
@@ -53,6 +53,15 @@ namespace aspect
         CitationInfo::add("CPO");
         const unsigned int my_rank = Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
         this->random_number_generator.seed(random_number_seed+my_rank);
+
+        // Don't assert when called by the unit tester.
+        if (this->simulator_is_past_initialization())
+          {
+            AssertThrow(this->introspection().compositional_name_exists("water"),
+                        ExcMessage("Particle property CPO only works if"
+                                   "there is a compositional field called water."));
+            water_index = this->introspection().compositional_index_for_name("water");
+          }
       }
 
 
@@ -229,7 +238,7 @@ namespace aspect
 
         // Calculate strain rate from velocity gradients
         const SymmetricTensor<2,dim> strain_rate = symmetrize (velocity_gradient);
-        const SymmetricTensor<2,dim> compressible_strain_rate
+        const SymmetricTensor<2,dim> deviatoric_strain_rate
           = (this->get_material_model().is_compressible()
              ?
              strain_rate - 1./3 * trace(strain_rate) * unit_symmetric_tensor<dim>()
@@ -238,16 +247,11 @@ namespace aspect
 
         const double pressure = solution[this->introspection().component_indices.pressure];
         const double temperature = solution[this->introspection().component_indices.temperature];
-        // Only assert in debug mode, because it should already be checked during initialization.
-        AssertThrow(this->introspection().compositional_name_exists("water"),
-                    ExcMessage("Particle property CPO only works if"
-                               "there is a compositional field called water."));
-        const unsigned int water_idx = this->introspection().compositional_index_for_name("water");
-        const double water_content = solution[this->introspection().component_indices.compositional_fields[water_idx]];
+        const double water_content = solution[this->introspection().component_indices.compositional_fields[water_index]];
 
         // get the composition of the particle
         std::vector<double> compositions;
-        for (unsigned int i = 0; i < this->n_compositional_fields(); i++)
+        for (unsigned int i = 0; i < this->n_compositional_fields(); ++i)
           {
             const unsigned int solution_component = this->introspection().component_indices.compositional_fields[i];
             compositions.push_back(solution[solution_component]);
@@ -267,8 +271,8 @@ namespace aspect
           {
             strain_rate_3d[0][2] = strain_rate[0][2];
             strain_rate_3d[1][2] = strain_rate[1][2];
-            //sym: strain_rate_3d[2][0] = strain_rate[2][0];
-            //sym: strain_rate_3d[2][1] = strain_rate[2][1];
+            //sym: strain_rate_3d[2][0] = strain_rate[0][2];
+            //sym: strain_rate_3d[2][1] = strain_rate[1][2];
             strain_rate_3d[2][2] = strain_rate[2][2];
           }
         Tensor<2,3> velocity_gradient_3d;
@@ -307,7 +311,7 @@ namespace aspect
                                                            velocity,
                                                            compositions,
                                                            strain_rate,
-                                                           compressible_strain_rate,
+                                                           deviatoric_strain_rate,
                                                            water_content);
 
             switch (advection_method)
@@ -349,7 +353,7 @@ namespace aspect
                                   + std::to_string(inv_sum_volume_mineral) + "."));
 
                 /**
-                 * Correct direction cosine matrices numerical error (orthnormality) after integration
+                 * Correct direction rotation matrices numerical error (orthnormality) after integration
                  * Follows same method as in matlab version from Thissen (see https://github.com/cthissen/Drex-MATLAB/)
                  * of finding the nearest orthonormal matrix using the SVD
                  */
@@ -556,7 +560,7 @@ namespace aspect
                                                             const Tensor<1,dim> &velocity,
                                                             const std::vector<double> &compositions,
                                                             const SymmetricTensor<2,dim> &strain_rate,
-                                                            const SymmetricTensor<2,dim> &compressible_strain_rate,
+                                                            const SymmetricTensor<2,dim> &deviatoric_strain_rate,
                                                             const double water_content) const
       {
         std::pair<std::vector<double>, std::vector<Tensor<2,3>>> derivatives;
@@ -577,7 +581,7 @@ namespace aspect
                                                                                   velocity,
                                                                                   compositions,
                                                                                   strain_rate,
-                                                                                  compressible_strain_rate,
+                                                                                  deviatoric_strain_rate,
                                                                                   water_content);
 
               set_deformation_type(cpo_index,data,mineral_i,static_cast<unsigned int>(deformation_type));
@@ -660,13 +664,13 @@ namespace aspect
             Tensor<2,3> G;
             Tensor<1,3> w;
             Tensor<1,4> beta({1.0, 1.0, 1.0, 1.0});
-            std::vector<Tensor<1,3>> slip_normal_reference {Tensor<1,3>({0,1,0}),Tensor<1,3>({0,0,1}),Tensor<1,3>({0,1,0}),Tensor<1,3>({1,0,0})};
-            std::vector<Tensor<1,3>> slip_direction_reference {Tensor<1,3>({1,0,0}),Tensor<1,3>({1,0,0}),Tensor<1,3>({0,0,1}),Tensor<1,3>({0,0,1})};
+            std::array<Tensor<1,3>,4> slip_normal_reference {{Tensor<1,3>({0,1,0}),Tensor<1,3>({0,0,1}),Tensor<1,3>({0,1,0}),Tensor<1,3>({1,0,0})}};
+            std::array<Tensor<1,3>,4> slip_direction_reference {{Tensor<1,3>({1,0,0}),Tensor<1,3>({1,0,0}),Tensor<1,3>({0,0,1}),Tensor<1,3>({0,0,1})}};
 
             // these are variables we only need for olivine, but we need them for both
             // within this if block and the next ones
             // Ordered vector where the first entry is the max/weakest and the last entry is the inactive slip system.
-            std::vector<unsigned int> indices(4,0);
+            std::array<unsigned int,4> indices;
 
             // compute G and beta
             Tensor<1,4> bigI;
@@ -689,7 +693,7 @@ namespace aspect
               {
                 // compute the element wise absolute value of the element wise
                 // division of BigI by tau (tau = ref_resolved_shear_stress).
-                std::vector<double> q_abs(4);
+                std::array<double,4> q_abs;
                 for (unsigned int i = 0; i < 4; ++i)
                   {
                     q_abs[i] = std::abs(bigI[i] / tau[i]);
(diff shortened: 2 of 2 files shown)


doc/make_cite_html.py
```diff
@@ -72,6 +72,7 @@
                "fraters_menno_2020_3900603" : "GWB",
                "Fraters2019c" : "GWB",
                "FBTGS19" : "NewtonSolver",
+               "fraters_billen_2021_cpo", "CPO"
```
tjhei, line 75:
```suggestion
               "fraters_billen_2021_cpo": "CPO"
```

doc/make_cite_html.py
```diff
@@ -103,6 +104,7 @@
...
         "heister_aspect_methods2" : "Timo Heister, Juliane Dannberg, Rene Gassmöller, and Wolfgang Bangerth. 2017. “High Accuracy Mantle Convection Simulation through Modern Numerical Methods – II: Realistic Models and Problems.” Geophysical Journal International 210 (2) (May 9): 833–851. doi:10.1093/gji/ggx195. http://dx.doi.org/10.1093/gji/ggx195.",
         "rose_freesurface" : "Ian Rose, Bruce Buffett, and Timo Heister. 2017. “Stability and Accuracy of Free Surface Time Integration in Viscous Flows.” Physics of the Earth and Planetary Interiors 262 (January): 90–100. doi:10.1016/j.pepi.2016.11.007. http://dx.doi.org/10.1016/j.pepi.2016.11.007."
+        "fraters_billen_2021_cpo" : "Fraters, M. R. T. and Billen, M. I. 2021. “On the Implementation and Usability of Crystal Preferred Orientation Evolution in Geodynamic Modeling” Geochemistry, Geophysics, Geosystems 22 (10): e2021GC009846. doi:10.1029/2021GC009846. https://doi.org/10.1029/2021GC009846."
```
tjhei, line 107: missing comma before

doc/sphinx/references.bib
```diff
@@ -11590,6 +11590,22 @@ @Article{fraters:etal:2019
     eprint = {http://oup.prod.sis.lan/gji/article-pdf/218/2/873/28693654/ggz183.pdf},
 }
 
+@article{fraters:billen:2021,
```
tjhei, line 11593: The keys in the dictionary in the make_cite_html.py file do not match this bibtex key. Use either ":" or "_" consistently.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -37,12 +37,12 @@ namespace aspect
     {
       enum class DeformationType
       {
-        passive
+        passive, olivine_a_fabric, olivine_b_fabric, olivine_c_fabric, olivine_d_fabric, olivine_e_fabric, enstatite
```
tjhei, line 40: I know that we don't have documentation in the other places here either, but it would be nice to have short doxygen documentation what each of the enums means.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -222,6 +240,39 @@ namespace aspect
           unsigned int
           get_number_of_minerals();
 
+          /**
+           * @brief Deterimens the deformation type from the deformation type selector.
+           *
+           * @param stress
```
tjhei, line 246: This documentation is incomplete and empty. Either complete it or remove it, otherwise doxygen will warn.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -222,6 +240,39 @@ namespace aspect
...
+                                     const std::vector<double> &compositions,
+                                     const SymmetricTensor<2,dim> &strain_rate,
+                                     const SymmetricTensor<2,dim> &compressible_strain_rate,
+                                     const double water_content) const;
+
+          /**
+           * @brief Computes the deformation type given the stress and water content according to the
+           * table in Karato 2008.
+           *
+           * @param stress
```
tjhei, line 265: same here

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -222,6 +240,39 @@ namespace aspect
...
+           * @param water_content
+           */
+          DeformationType
+          determine_deformation_type_karato_2008(const double stress,
+                                                 const double water_content) const;
+
+          std::array<double,4>
+          reference_resolved_shear_stress_from_deformation_type(DeformationType deformation_type,
```
tjhei, line 273: maybe document ``maxvalue``? units?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+      CrystalPreferredOrientation<dim>::compute_derivatives_drex_2004(const unsigned int cpo_index,
+                                                                      const ArrayView<double> &data,
+                                                                      const unsigned int mineral_i,
+                                                                      const SymmetricTensor<2,3> &strain_rate_3d,
+                                                                      const Tensor<2,3> &velocity_gradient_tensor,
+                                                                      const std::array<double,4> ref_resolved_shear_stress,
+                                                                      const bool prevent_nondimensionalization) const
+      {
+        SymmetricTensor<2,3> strain_rate_nondimensional = strain_rate_3d;
+        Tensor<2,3> velocity_gradient_tensor_nondimensional = velocity_gradient_tensor;
```
tjhei, line 628: Can you move those two linear below the if statement and always do ``= strain_rate_3d / nondimensionalization_value;``?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+        std::vector<double> deriv_volume_fractions(n_grains);
+        std::vector<Tensor<2,3>> deriv_a_cosine_matrices(n_grains);
+
+        // create shorcuts
+        const std::array<double, 4> &tau = ref_resolved_shear_stress;
+
+        std::vector<double> strain_energy(n_grains);
+        double mean_strain_energy = 0;
+
+        // loop over grains
```
tjhei, line 659: remove this comment (doesn't give additional information)

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+            // todo: initialize to dealii uninitialized value
+            unsigned int index_max_q = 0;
+            unsigned int index_intermediate_q = 0;
+            unsigned int index_min_q = 0;
+            unsigned int index_inactive_q = 0;
+
+            // compute G and beta
+            Tensor<1,4> bigI;
+            // this should be equal to a_cosine_matrices[grain_i]*a_cosine_matrices[grain_i]?
+            // todo: check and maybe replace?
```
tjhei, line 680: todo?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+            const Tensor<2,3> rotation_matrix = get_rotation_matrix_grains(cpo_index,data,mineral_i,grain_i);
+            for (unsigned int i = 0; i < 3; ++i)
+              {
+                for (unsigned int j = 0; j < 3; ++j)
+                  {
+                    bigI[0] = bigI[0] + strain_rate_nondimensional[i][j] * rotation_matrix[0][i] * rotation_matrix[1][j];
```
tjhei, line 686: Is there maybe a tensor operator that does this multiplication?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+
+                q_abs[index_intermediate_q] = -1;
+
+                index_min_q = std::distance(q_abs.begin(),max_element(q_abs.begin(), q_abs.end()));
+
+                q_abs[index_min_q] = -1;
+
+                index_inactive_q = std::distance(q_abs.begin(),max_element(q_abs.begin(), q_abs.end()));
+
+                // todo: explain
```
tjhei, line 729: yes please

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+                Assert(bigI[index_max_q] != 0.0, ExcMessage("Internal error: bigI is zero."));
+                double ratio = tau[index_max_q]/bigI[index_max_q];
+
+                double q_intermediate = ratio * (bigI[index_intermediate_q]/tau[index_intermediate_q]);
```
tjhei, line 733: add const where ever you can.

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+            // todo: expand
+            double top = 0;
+            double bottom = 0;
+            for (unsigned int i = 0; i < 3; ++i)
+              {
+                // Following the Drex code, which differs from EPSL paper,
+                // which says gamma_nu depends on i+1: actually uses i+2
+                unsigned int ip2 = i + 2;
+                if (ip2 > 2)
+                  ip2 = ip2-3;
```
tjhei, line 765: ``const unsigned int ip2 = (i<=2) ? (i+2) : (i-1);`` ?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+                bottom = bottom - (G[i][ip2]-G[ip2][i])*(G[i][ip2]-G[ip2][i]);
+
+                for (unsigned int j = 0; j < 3; ++j)
+                  {
+                    top = top + 2.0 * G[i][j]*velocity_gradient_tensor_nondimensional[i][j];
+                    bottom = bottom + 2.0* G[i][j] * G[i][j];
+                  }
+              }
+            // see comment on if all BigI are zero. In that case gamma should be zero.
+            double gamma = bottom != 0.0 ? top/bottom : 0;
```
tjhei, line 777:
```suggestion
            const double gamma = (bottom != 0.0) ? (top/bottom) : 0.0;
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+
+            // compute w (equation 8, Kaminiski & Ribe, 2001)
+            // todo: explain what w is
+            // todo: there was a loop around this in the phyton code, discuss/check
```
tjhei, line 781: todo

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,400 @@ namespace aspect
...
+            // the code. So maybe remove it in the future.
+            case DeformationType::enstatite :
+              ref_resolved_shear_stress[0] = max_value;
+              ref_resolved_shear_stress[1] = max_value;
+              ref_resolved_shear_stress[2] = max_value;
+              ref_resolved_shear_stress[3] = 1;
+              break;
+
+            default:
+              break;
```
tjhei, line 1007: right default?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -642,6 +1096,47 @@ namespace aspect
...
+              {
+                prm.declare_entry ("Minerals", "Olivine: Karato 2008, Enstatite",
+                                   Patterns::List(Patterns::Anything()),
+                                   "This determines what minerals and fabrics or fabric selectors are used used for the LPO calculation. "
+                                   "The options are Olivine: A-fabric, Olivine: B-fabric, Olivine: C-fabric, Olivine: D-fabric, "
+                                   "Olivine: E-fabric, Olivine: Karato 2008 or Enstatite. The "
+                                   "Karato 2008 selector selects a fabric based on stress and water content as defined in "
+                                   "figure 4 of the Karato 2008 review paper (doi: 10.1146/annurev.earth.36.031207.124120).");
+
+                prm.declare_entry ("Mobility", "50",
```
tjhei, line 1110: can you add units to all parameters?

unit_tests/crystal_preferred_orientation.cc
```diff
@@ -511,3 +525,563 @@ TEST_CASE("CPO core: Spin tensor")
       }
   }
 }
+
+
+TEST_CASE("Fabric determination function")
```
tjhei, line 530: very nice use of the unit tests!

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -263,24 +288,27 @@ namespace aspect
...
             */
             double sum_volume_mineral = 0;
-            std::pair<std::vector<double>, std::vector<Tensor<2,3>>> derivatives_grains = this->compute_derivatives(data_position,
-                                                                                            data,
-                                                                                            mineral_i,
-                                                                                            strain_rate_3d,
-                                                                                            velocity_gradient_3d,
-                                                                                            ref_resolved_shear_stress);
+            std::pair<std::vector<double>, std::vector<Tensor<2,3>>>
+            derivatives_grains = this->compute_derivatives(data_position,
```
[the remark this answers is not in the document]
MFraters: Yes, just below in the advection method block :)

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -532,6 +567,31 @@ namespace aspect
...
+                                                                            position,
+                                                                            temperature,
+                                                                            pressure,
+                                                                            velocity,
+                                                                            compositions,
+                                                                            strain_rate,
+                                                                            compressible_strain_rate,
+                                                                            water_content);
+
+              set_deformation_type(cpo_index,data,mineral_i,static_cast<unsigned int>(DeformationType::passive));
```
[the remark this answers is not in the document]
MFraters: ah, that was a mistake, thanks for catching it!

source/particle/property/crystal_preferred_orientation.cc
MFraters: Yes, it is taking the outer product of the slip system normal and slip direction, and then taking the scalar product of that with the non dimensional strain-rate. I have replaced the code and the test still works :)

MFraters:
Hey @tjhei, Thanks for the review! I think I have addressed all your comments. Based on some of your comments I realized that it would probably be better to do some more groundwork for the generalization from 4 slip systems to n slip systems.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -35,21 +35,56 @@ namespace aspect
...
+       * olivine_c_fabric: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a Olivine C Fabric, which influences the relative strenght of the slip planes.
+       * olivine_d_fabric: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a Olivine D Fabric, which influences the relative strenght of the slip planes.
+       * olivine_e_fabric: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a Olivine E Fabric, which influences the relative strenght of the slip planes.
+       * enstatite: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a enstatite Fabric, which influences the relative strenght of the slip planes.
+       * olivine_karato_2008: Only to be used with the D-Rex CPO Derivative algorithm. Sets the deformation type of the mineral to a olivine fabric based on the table in Karato 2008.
+       */
       enum class DeformationTypeSelector
```
tjhei, line 69: What is the difference between DeformationType and DeformationTypeSelector?

MFraters:
@tjhei I have addressed your comments. I removed some of the repetition in the description of the enum and added some text to explain the difference between the deformation type and the deformation type selector. Can you take a look to see whether the text is more clear now?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,369 @@ namespace aspect
...
+            // from Kaminski and Ribe, GJI 2004 and
+            // Becker et al., 2007 (http://www-udc.ig.utexas.edu/external/becker/preprints/bke07.pdf)
+            case DeformationType::olivine_c_fabric :
+              ref_resolved_shear_stress[0] = 3;
+              ref_resolved_shear_stress[1] = max_value;
+              ref_resolved_shear_stress[2] = 2;
+              ref_resolved_shear_stress[3] = 1;
+              break;
+
+            // from Kaminski and Ribe, GRL 2002and
```
gassmoeller, line 944:
```suggestion
            // from Kaminski and Ribe, GRL 2002 and
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,369 @@ namespace aspect
...
+            // Becker et al., 2007 (http://www-udc.ig.utexas.edu/external/becker/preprints/bke07.pdf)
+            case DeformationType::olivine_d_fabric :
+              ref_resolved_shear_stress[0] = 1;
+              ref_resolved_shear_stress[1] = 1;
+              ref_resolved_shear_stress[2] = 3;
+              ref_resolved_shear_stress[3] = max_value;
+              break;
+
+            // Kaminski, Ribe and Browaeys, JGI, 2004 (same as in the matlab code)and
```
gassmoeller, line 953:
```suggestion
            // Kaminski, Ribe and Browaeys, GJI, 2004 (same as in the matlab code) and
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,369 @@ namespace aspect
...
+            // Becker et al., 2007 (http://www-udc.ig.utexas.edu/external/becker/preprints/bke07.pdf)
+            case DeformationType::olivine_e_fabric :
+              ref_resolved_shear_stress[0] = 2;
+              ref_resolved_shear_stress[1] = 1;
+              ref_resolved_shear_stress[2] = max_value;
+              ref_resolved_shear_stress[3] = 3;
+              break;
+
+            // from Kaminski and Ribe, GJI 2004.
+            // Todo: this one is not used in practice, since there is an optimalisation in
```
gassmoeller, line 963:
```suggestion
            // Todo: this one is not used in practice, since there is an optimization in
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -642,6 +1065,45 @@ namespace aspect
...
+                                   "The options are Olivine: A-fabric, Olivine: B-fabric, Olivine: C-fabric, Olivine: D-fabric, "
+                                   "Olivine: E-fabric, Olivine: Karato 2008 or Enstatite. The "
+                                   "Karato 2008 selector selects a fabric based on stress and water content as defined in "
+                                   "figure 4 of the Karato 2008 review paper (doi: 10.1146/annurev.earth.36.031207.124120).");
+
+                prm.declare_entry ("Mobility", "50",
+                                   Patterns::Double(0),
+                                   "The dimensionless intrinsic grain boundary mobility for both olivine and enstatite.");
+
+                prm.declare_entry ("Volume fractions minerals", "0.5, 0.5",
```
gassmoeller, line 1083: Is 0.5,0.5 a good default value? What is the proportion in the mantle?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -642,6 +1065,45 @@ namespace aspect
...
+                prm.declare_entry ("Stress exponents", "3.5",
+                                   Patterns::Double(0),
+                                   "This is the power law exponent that characterizes the rheology of the "
+                                   "slip systems. It is used in equation 11 of Kaminski et al., 2004.");
+
+                prm.declare_entry ("Exponents p", "1.5",
+                                   Patterns::Double(0),
+                                   "This is exponent p as defined in equation 11 of Kaminski et al., 2004. ");
+
+                prm.declare_entry ("Nucleation efficientcy", "5",
```
gassmoeller, line 1097:
```suggestion
                prm.declare_entry ("Nucleation efficiency", "5",
```

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -642,6 +1065,45 @@ namespace aspect
...
+                                   Patterns::Double(0),
+                                   "This is the dimensionless nucleation rate as defined in equation 8 of "
+                                   "Kaminski et al., 2004. ");
+
+                prm.declare_entry ("Threshold GBS", "0.3",
+                                   Patterns::Double(0),
+                                   "This is the dimensionless grain-boundary sliding threshold. ");
```
gassmoeller, line 1104: Can you elaborate what the threshold is for? I.e. is it a stress threshold? slip rate threshold?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -744,6 +1234,17 @@ namespace aspect
...
               }
               prm.leave_subsection();
+
+              prm.enter_subsection("D-Rex 2004");
+              {
+                mobility = prm.get_double("Mobility");
+                volume_fractions_minerals = Utilities::string_to_double(dealii::Utilities::split_string_list(prm.get("Volume fractions minerals")));
+                stress_exponent = prm.get_double("Stress exponents");
+                exponent_p = prm.get_double("Exponents p");
+                nucleation_efficientcy = prm.get_double("Nucleation efficientcy");
```
gassmoeller, line 1244:
```suggestion
                nucleation_efficiency = prm.get_double("Nucleation efficiency");
```

gassmoeller (commented):
A bunch of comments :-). But mostly requests for more documentation, or suggestions to improve efficiency. Can you please at least add one test model that uses the new functionality. In particular there were typos in input parameters that were not caught, because no test actually uses the whole D-REX calculation together (the existing tests use the passive case without fabric evolution).

MFraters:
> Can you please at least add one test model that uses the new functionality. In particular there were typos in input parameters that were not caught, because no test actually uses the whole D-REX calculation together (the existing tests use the passive case without fabric evolution).

This is a bit of a problem of cutting up a pull request this much. I have tests for this which depend on other parts (need at least the postprocessor). That is why I put it as a seperate point (point 9) in the tracking issue #3885. What do you think is the best way to move forward on this?

And about the typos in the input parameters, don't worry, they are also definitely in the input files of the tests ;)

include/aspect/particle/property/crystal_preferred_orientation.h
MFraters: It returns the relative strength of the slip planes, which for olivine are 4. This is (more or less) how it is done in the fortran code. I have been thinking to generalize it (some minerals have way more slip planes), but I would rather to do that in a follow up pull request, because this is probably already complicated enough as it is. I will add better documentation.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -35,21 +35,55 @@ namespace aspect
...
-        passive
+        passive, olivine_a_fabric, olivine_b_fabric, olivine_c_fabric, olivine_d_fabric, olivine_e_fabric, enstatite
       };
 
+
+      /**
+       * @brief The type of deformation selector used by the CPO code.
+       *
+       * The selector is a input parameter and it can either set a deformation type directly or deterimine the deformation type through an algorithm.
+       * The deformation type selector is used to determine/select the the deformation type. It can be a fixed deformation type, for example,
```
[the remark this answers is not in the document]
MFraters: For example for the `DeformationTypeSelector` `olivine_karato_2008`, the actual `DeformationType` can be `olivine_a_fabric`, or `olivine_a_fabric`, etc.

MFraters:
> (the existing tests use the passive case without fabric evolution).

There is a unit test which tests 2 different computations of the derivatives with an actual fabric (one of which I computed a analytic solution of the derivatives by hand). The adding computation of these derivatives is the core part of this pull request, since it is what the D-Rex algorithm essentially is.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -222,6 +274,38 @@ namespace aspect
           unsigned int
           get_number_of_minerals() const;
 
+          /**
+           * @brief Deterimens the deformation type from the deformation type selector.
```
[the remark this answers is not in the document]
MFraters: see my comment above on the difference

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -227,6 +229,29 @@ namespace aspect
...
+        const double temperature = solution[this->introspection().component_indices.temperature];
+        // Only assert in debug mode, because it should already be checked during initialization.
+        AssertThrow(this->introspection().compositional_name_exists("water"),
+                    ExcMessage("Particle property CPO only works if"
+                               "there is a compositional field called water."));
+        const unsigned int water_idx = this->introspection().compositional_index_for_name("water");
+        const double water_content = solution[this->introspection().component_indices.compositional_fields[water_idx]];
+
+        // get the composition of the particle
+        std::vector<double> compositions;
```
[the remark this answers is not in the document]
MFraters: The reason I did that was because it would then need to be extracted within the loop over the minerals. Although my case are currently are only 2 minerals, but it might become inefficient for cases with many minerals. It would simplify the interface considerably though...

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,369 @@ namespace aspect
...
+                    bottom = bottom + 2.0* G[i][j] * G[i][j];
+                  }
+              }
+            // see comment on if all BigI are zero. In that case gamma should be zero.
+            const double gamma = bottom != 0.0 ? top/bottom : 0;
+
+            // compute w (equation 8, Kaminiski & Ribe, 2001)
+            // w is the Rotation rate vector of the crystallographic axes of grain
+            w[0] = 0.5*(velocity_gradient_tensor_nondimensional[2][1]-velocity_gradient_tensor_nondimensional[1][2]) - 0.5*(G[2][1]-G[1][2])*gamma;
+            w[1] = 0.5*(velocity_gradient_tensor_nondimensional[0][2]-velocity_gradient_tensor_nondimensional[2][0]) - 0.5*(G[0][2]-G[2][0])*gamma;
```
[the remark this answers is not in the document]
MFraters: The equation on which this is based (equation 8, Kaminiski & Ribe, 2001) states: 

$$
\begin{align}
\omega^v_i = \frac{L_{i+2 i+1}-L_{i+1 i+2}}{2} -  \frac{G^v_{i+2 i+1}-G^v_{i+1 i+2}}{2} \gamma_v
\end{align}
$$

where i = 1,2,3, but with c++ i = 0,1,2

With the symmetry I can see where you are coming from, and I had to look into it to understand again why it is done this way. The goal here is to get a rotation rate tensor $\bf \omega$. We can get a rotation rate tensor with the following: $W = \frac{L-L^T}{2}$. The next step is to convert this second order tensor to a vector. Looking at https://www.continuummechanics.org/velocitygradient.html, the rotation rate tensor $W$ can be summarized as the following: 

$$
W = \pmatrix{0 &-\omega_z &\omega_y \\
\omega_z &0 &-\omega_x \\
-\omega_y &\omega_x &0} = 
\left[ \matrix{
0 &
{1 \over 2} \left( {\partial v_x \over \partial y} - {\partial v_y \over \partial x} \right) & 
{1 \over 2} \left( {\partial v_x \over \partial z} - {\partial v_z \over \partial x} \right) \\
\\
{1 \over 2} \left( {\partial v_y \over \partial x} - {\partial v_x \over \partial y} \right) & 
0 &
{1 \over 2} \left( {\partial v_y \over \partial z} - {\partial v_z \over \partial y} \right) \\
\\
{1 \over 2} \left( {\partial v_z \over \partial x} - {\partial v_x \over \partial z} \right) & 
{1 \over 2} \left( {\partial v_z \over \partial y} - {\partial v_y \over \partial z} \right) &
0 
} \right]
$$

The equation for the elements of the vector is 

$$
\omega_x = {1 \over 2} \left( {\partial v_z \over \partial y} - {\partial v_y \over \partial z} \right) 
$$

$$
\omega_y  = {1 \over 2} \left( {\partial v_x \over \partial z} - {\partial v_z \over \partial x} \right)
$$

$$
\omega_z = {1 \over 2} \left( {\partial v_y \over \partial x} - {\partial v_x \over \partial y} \right)
$$

So due to the "asymmetry" in $W$ in terms of defining $\bf \omega$, you get an asymmetry in the construction of vector $\bf \omega$

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,369 @@ namespace aspect
...
+
+            // Now calculate the analytic solution to the deformation minimization problem
+            // compute gamma (equation 7, Kaminiski & Ribe, 2001)
+            double top = 0;
+            double bottom = 0;
+            for (unsigned int i = 0; i < 3; ++i)
+              {
+                // Following the Drex code, which differs from EPSL paper,
+                // which says gamma_nu depends on i+1: actually uses i+2
+                const unsigned int ip2 = (i==0) ? (i+2) : (i-1);
```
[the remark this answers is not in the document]
MFraters: it does indeed stand for "i plus 2", but since it wraps around from i=1 -> ip2 = 0 and i=2 -> ip2 = 1. I guess the best name would be `i_plus_2_wrapped_around`, but that is a bit long. I have changed it to `i_offset`. It is a bit generic, but maybe more clear. What do you think?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -554,6 +614,369 @@ namespace aspect
...
+                                         + beta[2] * rotation_matrix[2][i] * rotation_matrix[1][j]
+                                         + beta[3] * rotation_matrix[2][i] * rotation_matrix[0][j]);
+                      }
+                  }
+              }
+
+            // Now calculate the analytic solution to the deformation minimization problem
+            // compute gamma (equation 7, Kaminiski & Ribe, 2001)
+            double top = 0;
+            double bottom = 0;
```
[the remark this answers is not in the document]
MFraters: They are the numerator and the denominator in equation equation 7 in Kaminiski & Ribe (2001), but I can never tell apart which is which by the names, so I just named them top and bottom. I will document it a bit better.

source/particle/property/crystal_preferred_orientation.cc
MFraters: You are right, given the other defaults, it have now set it to 0.7, 0.3.

MFraters:
@gassmoeller Thanks for the review! I have addressed all your comments. I have left the comment I thought would be good for you to read if my reply is satisfactory unresolved.

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -222,6 +274,38 @@ namespace aspect
           unsigned int
           get_number_of_minerals() const;
 
+          /**
+           * @brief Determines the deformation type from the deformation type selector.
```
gassmoeller, line 278: Would this correctly explain this function? If so, please include. If not, please include a correct explanation ;-).

```suggestion
           * @brief Determines the deformation type from the deformation type selector.
           * If the provided @p deformation_type_selector is a specific deformation type,
           * the function will simply return the corresponding deformation type. However,
           * if the @p deformation_type_selector is an algorithm to determine the current
           * deformation type (e.g. based on measured lab data or analytical models), then
           * the function computes the appropriate deformation type at the given conditions
           * and returns the compute deformation type.
```

source/particle/property/crystal_preferred_orientation.cc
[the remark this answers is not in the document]
gassmoeller: Well, the only thing that is faster in the current implementation than directly accessing the data from solution is that `this->introspection` is called less often. In other words the following two algorithms are really nearly identical in speed (but the second is much easier to read if you have a lot of indices and values like you do):
```
Vector<double> solution (some content);
const double pressure = solution[this->introspection().component_indices.pressure];

for (unsigned int i=0; i<50; ++i)
  sum += pressure;
```

and:

```
Vector<double> solution (some content);
const auto &introspection = this->introspection();

for (unsigned int i=0; i<50; ++i)
  sum += solution[introspection.component_indices.pressure];
```

In fact after compiler optimizations they may be the same anyway.

Anyway, I guess this is a long winded way of saying: If you like your implementation because you find it easier to read that is ok, we can keep it. But if you like it, because you guess that it is faster, I would say it probably is not. Also if you are concerned about speed: If I would need to guess the slowest part of your algorithm, I would suspect line 720 and 774, where you call std::pow for every mineral and grain and slip system. But even with this guess, I would not start any optimizations before I had measured its impact.

So bring this into whichever form you think is best to read for now, and we can worry about optimizations afterwards.

source/particle/property/crystal_preferred_orientation.cc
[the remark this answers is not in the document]
gassmoeller: Ok, thanks for checking!

gassmoeller (commented):
Thank you, and sorry for the delay, looks much better now. A few comments left.

source/particle/property/crystal_preferred_orientation.cc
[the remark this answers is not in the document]
MFraters: I can see readability advantages for both approaches, so I don't really have a strong preference necessarily. 

>  If I would need to guess the slowest part of your algorithm, I would suspect line 720 and 774, where you call std::pow for every mineral and grain and slip system.

haha, yes, everytime I profile stuff I keep being supersized how slow std::pow can be :)

I gave it a quick try to see if I would like it more the other way, and another problem quickly showed up. It introduces introspection calls into the function, which make it much harder to unit test the function. I am sure we could find a solution to make the unit tests work with these introspection calls, but I kinda like that the function is fully self contained. Let me know what you think.

MFraters (commented):
Thanks for the review. I addressed your comments. The only outstanding issue is about using `solution` and `gradients` in `compute_derivates`. Let me know what you think about that.

gassmoeller (approved):
> I gave it a quick try to see if I would like it more the other way, and another problem quickly showed up. It introduces introspection calls into the function, which make it much harder to unit test the function.

Ok, that is a good point. Let's not worry about it then for now.

PR #4977, opened 2022-10-11
Interpolate all particle properties in one operation

gassmoeller:
Our particle interpolators support interpolating more than one property per call, but the calling side in the Simulator class does not make use of this. Instead we call one interpolation at a time, because it is embedded in a loop over all compositional fields. This is not efficient, but often does not matter much, since interpolation is usually quick. However, @KiralyAgi recently sent me a model with 37(!) particle properties, and there interpolation made up a significant part of the runtime. This change interpolates all properties in one call, and should decrease that cost significantly. @KiralyAgi can you try to take these changes, incorporate them into your branch and maybe report back if it helped your runtime? The change itself should be pretty uncontroversial, the only extra condition it applies to the interpolation is that all interpolated fields need to have the same base element. We do not support different elements for different fields anyway at the moment, but this is something to keep in mind for the future.

I should mention that this does not affect the results, because the current_linearization_point is only updated after all compositional fields (particles or otherwise) have been solved. This is done to avoid solutions that depend on the order in which compositional fields are specified.

### DIFF
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -840,7 +840,7 @@ namespace aspect
        * Interpolate the corresponding particle properties into the given
        * @p advection_fields solution fields.
        */
-      void interpolate_particle_property_vector (const std::vector<AdvectionField> &advection_fields);
+      void interpolate_particle_properties (const std::vector<AdvectionField> &advection_fields);
 
       /**
        * Solve the Stokes linear system.
--- a/include/aspect/simulator.h
+++ b/include/aspect/simulator.h
@@ -831,8 +831,16 @@ namespace aspect
 
       /**
        * Interpolate a particular particle property to the solution field.
+       *
+       * @deprecated: Use interpolate_particle_property_vector() instead.
        */
-      void interpolate_particle_properties (const AdvectionField &advection_field);
+      void interpolate_particle_properties (const AdvectionField &advection_field) DEAL_II_DEPRECATED;
+
+      /**
+       * Interpolate the corresponding particle properties into the given
+       * @p advection_fields solution fields.
+       */
+      void interpolate_particle_property_vector (const std::vector<AdvectionField> &advection_fields);
 
       /**
        * Solve the Stokes linear system.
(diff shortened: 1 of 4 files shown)


gassmoeller:
I got timing information from @KiralyAgi (I left out everything not particle related as those lines are near identical). A model before this PR:
```
| Total wallclock time elapsed since start     |  2.21e+03s |            |
| Particles: Advect                |       758 |     0.399s |         0% |
| Particles: Copy                  |       379 |      2.25s |       0.1% |
| Particles: Generate              |         1 |      2.02s |         0% |
| Particles: Initialization        |         1 |    0.0186s |         0% |
| Particles: Initialize properties |         1 |     0.338s |         0% |
| Particles: Interpolate           |     13644 |       342s |        15% |
| Particles: Sort                  |       758 |      1.18s |         0% |
| Particles: Update properties     |       379 |       992s |        45% |
```
And with this PR:
 ```
| Total wallclock time elapsed since start     |  1.94e+03s |            |
| Particles: Advect                |       758 |     0.306s |         0% |
| Particles: Copy                  |       379 |       2.2s |      0.11% |
| Particles: Generate              |         1 |      4.75s |      0.24% |
| Particles: Initialization        |         1 |    0.0187s |         0% |
| Particles: Initialize properties |         1 |     0.337s |         0% |
| Particles: Interpolate           |       379 |      15.5s |       0.8% |
| Particles: Sort                  |       758 |      1.15s |         0% |
| Particles: Update properties     |       379 |     1e+03s |        52% |
```

While the most expensive part is still the particle update this PR reduces the time to interpolate from 342 to 15.5 seconds. I think this is worth pursuing :-). Ready for a review. All results should be identical and we have tests with more than one composition that are tracked on particles (e.g. particle_interpolator_harmonic_average_solkz.prm) that should show a difference if this PR had changed behavior.

File: include/aspect/simulator.h [hunk 1]
```diff
@@ -831,8 +831,16 @@ namespace aspect
...
+       * @deprecated: Use interpolate_particle_property_vector() instead.
+       */
+      void interpolate_particle_properties (const AdvectionField &advection_field) DEAL_II_DEPRECATED;
+
+      /**
+       * Interpolate the corresponding particle properties into the given
+       * @p advection_fields solution fields.
        */
-      void interpolate_particle_properties (const AdvectionField &advection_field);
+      void interpolate_particle_property_vector (const std::vector<AdvectionField> &advection_fields);
```
tjhei, line 843: How about

```suggestion
      void interpolate_particle_properties (const std::vector<AdvectionField> &advection_fields);
```

tjhei (approved):

tjhei:
Feel free to merge if you think it is ready.

gassmoeller:
I used your suggestion and avoided a compiler warning I got with newer gcc compilers. Ready.

PR #4978, opened 2022-10-11
add particle get_plugin_pointer_by_name() function.

MFraters:
Working on point 3 of issue #3885, I have been struggling to make the interface between the particle property and the cpo post-processor working cleanly. I previously did it through providing static functions in the cpo property plugin. This worked fine, but it lead to also requiring a static variable in the class. 

I think the cleanest way of implementing it is through accessing the actual particle property plugin, but that requires a way to access it. That is what this pull request proposes to add.


gassmoeller (commented):
This is one way to achieve this, but I would prefer we stay consistent with the other places where we do something similar already. I.e. to have a function `has_particle_property` and another one `get_matching_particle_property`. You can for example check the postprocessor interface. Header file here:
https://github.com/geodynamics/aspect/blob/3eef1314235db317da36e76bff00921a404d93d1/include/aspect/postprocess/interface.h#L246

And implementation here:
https://github.com/geodynamics/aspect/blob/3eef1314235db317da36e76bff00921a404d93d1/include/aspect/postprocess/interface.h#L406

An usage example is here:
https://github.com/geodynamics/aspect/blob/3eef1314235db317da36e76bff00921a404d93d1/source/postprocess/geoid.cc#L319

This would have the advantage that you do not need to use shared pointers, instead the functions return const references to the plugins. Shared pointers symbolize shared ownership, but here you do not want the postprocessor to take ownership of the particle property, just temporary access. Also this way you avoid expensive string comparisons.

MFraters:
Thanks for the quick review, sounds good! I have made the changes and they work equally well in the branches where this is used.

tjhei (approved):

source/particle/property/interface.cc
```diff
@@ -643,6 +643,44 @@ namespace aspect
 
 
 
+      template <int dim>
+      template <typename ParticlePropertyType>
```
gassmoeller, line 647: I think these functions need to be in the header file for the compiler to know which types to instantiate for.

gassmoeller (commented):
Some comments to fix, and I think the function implementations need to move into the header.

source/particle/property/interface.cc
MFraters: It seemed to work without it, but I have no objections to put it in the header file :)

MFraters:
@gassmoeller Thanks for the quick follow up, I have addressed your comments :)

source/particle/property/interface.cc
MFraters: I had an error in the plugins I was using this in which made that I actually didn't use the function, so it may not have worked. Anyway, it does seem to work now.

gassmoeller (approved):

PR #4979, opened 2022-10-11
Add CPO main postprocessor

MFraters:
This pull request adds a post-processor to output relevant CPO information.

This implementation is based on #4978, so this should not be merged until that one is. It is ready for review though. 

(point 3 in #3885)

### DIFF
--- a/source/postprocess/crystal_preferred_orientation.cc
+++ b/source/postprocess/crystal_preferred_orientation.cc
@@ -235,6 +235,21 @@ namespace aspect
 
           Point<dim> position = it->get_location();
 
+          // always make a vector of rotation matrices
+          std::vector<std::vector<Tensor<2,3>>> rotation_matrices;
+          rotation_matrices.resize(n_minerals);
+          for (unsigned int mineral_i = 0; mineral_i < n_minerals; ++mineral_i)
+            {
+              rotation_matrices[mineral_i].resize(n_grains);
+              for (unsigned int i_grain = 0; i_grain < n_grains; i_grain++)
+                {
+                  rotation_matrices[mineral_i][i_grain] = cpo_particle_property.get_rotation_matrix_grains(
+                                                            cpo_data_position,
+                                                            properties,
+                                                            mineral_i,
+                                                            i_grain);
+                }
+            }
           // write master file
           string_stream_master << id << " " << position << " " << properties[cpo_data_position] <<  std::endl;
 
@@ -249,11 +264,7 @@ namespace aspect
                   for (unsigned int i_grain = 0; i_grain < n_grains; i_grain++)
                     {
                       euler_angles[mineral_i][i_grain] = Utilities::zxz_euler_angles_from_rotation_matrix(
-                                                           cpo_particle_property.get_rotation_matrix_grains(
-                                                             cpo_data_position,
-                                                             properties,
-                                                             mineral_i,
-                                                             i_grain));
+                                                           rotation_matrices[mineral_i][i_grain]);
                     }
                 }
             }
@@ -315,11 +326,7 @@ namespace aspect
                             break;
 
                           case Output::RotationMatrix:
-                            string_stream_content_raw << cpo_particle_property.get_rotation_matrix_grains(
-                                                        cpo_data_position,
-                                                        properties,
-                                                        write_raw_cpo[property_i].first,
-                                                        grain_i) << " ";
+                            string_stream_content_raw << rotation_matrices[write_raw_cpo[property_i].first][grain_i]<< " ";
                             break;
 
                           case Output::EulerAngles:
@@ -340,6 +347,7 @@ namespace aspect
             }
           if (write_draw_volume_weighted_cpo.size() != 0)
             {
+              std::vector<std::vector<dealii::Tensor<2,3>>> weighted_rotation_matrices;
               std::vector<std::vector<std::vector<double>>> weighted_euler_angles;
 
               weighted_euler_angles.resize(n_minerals);
@@ -354,30 +362,16 @@ namespace aspect
                                                                       mineral_i,
                                                                       i_grain);
                     }
-                  weighted_euler_angles[mineral_i] = random_draw_volume_weighting(volume_fractions_grains[mineral_i], euler_angles[mineral_i]);
+                  weighted_rotation_matrices[mineral_i] = Utilities::rotation_matrices_random_draw_volume_weighting(volume_fractions_grains[mineral_i], rotation_matrices[mineral_i], n_grains, this->random_number_generator);
 
-                  Assert(weighted_euler_angles[mineral_i].size() == euler_angles[mineral_i].size(),
-                         ExcMessage("Weighted angles vector (size = " + std::to_string(weighted_euler_angles[mineral_i].size()) +
+                  Assert(weighted_rotation_matrices[mineral_i].size() == euler_angles[mineral_i].size(),
+                         ExcMessage("Weighted rotation matrices vector (size = " + std::to_string(weighted_rotation_matrices[mineral_i].size()) +
                                     ") has different size from input angles (size = " + std::to_string(euler_angles[mineral_i].size()) + ")."));
 
-                }
-
-
-              std::vector<std::vector<Tensor<2,3>>> weighted_a_cosine_matrices;
-
-              if (compute_weighted_A_matrix == true)
-                {
-                  weighted_a_cosine_matrices.resize(n_minerals);
-                  for (unsigned int mineral_i = 0; mineral_i < n_minerals; ++mineral_i)
+                  for (unsigned int i_grain = 0; i_grain < n_grains; i_grain++)
                     {
-                      weighted_a_cosine_matrices[mineral_i].resize(weighted_euler_angles[mineral_i].size());
-
-                      for (unsigned int i = 0; i < weighted_euler_angles.size(); ++i)
-                        {
-                          weighted_a_cosine_matrices[mineral_i][i] = Utilities::zxz_euler_angles_to_rotation_matrix(weighted_euler_angles[mineral_i][i][0],
-                                                                     weighted_euler_angles[mineral_i][i][1],
-                                                                     weighted_euler_angles[mineral_i][i][2]);
-                        }
+                      weighted_euler_angles[mineral_i][i_grain] = Utilities::zxz_euler_angles_from_rotation_matrix(
+                                                                    weighted_rotation_matrices[mineral_i][i_grain]);
                     }
                 }
 
@@ -432,7 +426,7 @@ namespace aspect
                             break;
 
                           case Output::RotationMatrix:
-                            string_stream_content_draw_volume_weighting << weighted_a_cosine_matrices[write_draw_volume_weighted_cpo[property_i].first][grain_i] << " ";
+                            string_stream_content_draw_volume_weighting << weighted_rotation_matrices[write_draw_volume_weighted_cpo[property_i].first][grain_i] << " ";
                             break;
 
                           case Output::EulerAngles:
@@ -525,69 +519,6 @@ namespace aspect
                             particle_cpo_output);
       return std::make_pair("Writing particle cpo output:", particle_cpo_output);
     }
-
-    template<int dim>
-    std::vector<std::vector<double>>
-    CrystalPreferredOrientation<dim>::random_draw_volume_weighting(const std::vector<double> &fv,
-                                                                   const std::vector<std::vector<double>> &angles) const
-    {
-      // Get volume weighted euler angles, using random draws to convert odf
-      // to a discrete number of orientations, weighted by volume
-      // 1a. Get index that would sort volume fractions AND
-      //ix = np.argsort(fv[q,:]);
-      // 1b. Get the sorted volume and angle arrays
-      std::vector<double> fv_to_sort = fv;
-      std::vector<double> fv_sorted = fv;
-      std::vector<std::vector<double>> angles_sorted = angles;
-
-      unsigned int n_grains = fv_to_sort.size();
-
-      for (int i = n_grains-1; i >= 0; --i)
-        {
-          unsigned int index_max_fv = std::distance(fv_to_sort.begin(),max_element(fv_to_sort.begin(), fv_to_sort.end()));
-
-          fv_sorted[i] = fv_to_sort[index_max_fv];
-          angles_sorted[i] = angles[index_max_fv];
-          Assert(angles[index_max_fv].size() == 3, ExcMessage("angles vector (size = " + std::to_string(angles[index_max_fv].size()) +
-                                                              ") should have size 3."));
-          Assert(angles_sorted[i].size() == 3, ExcMessage("angles_sorted vector (size = " + std::to_string(angles_sorted[i].size()) +
-                                                          ") should have size 3."));
-          fv_to_sort[index_max_fv] = -1;
-        }
-
-
-      // 2. Get cumulative weight for volume fraction
-      std::vector<double> cum_weight(n_grains);
-      std::partial_sum(fv_sorted.begin(),fv_sorted.end(),cum_weight.begin());
-      // 3. Generate random indices
-      boost::random::uniform_real_distribution<> dist(0, 1);
-      std::vector<double> idxgrain(n_grains);
-      for (unsigned int grain_i = 0; grain_i < n_grains; ++grain_i)
-        idxgrain[grain_i] = dist(this->random_number_generator); //random.rand(ngrains,1);
-
-      // 4. Find the maximum cum_weight that is less than the random value.
-      // the euler angle index is +1. For example, if the idxGrain(g) < cumWeight(1),
-      // the index should be 1 not zero)
-      std::vector<std::vector<double>> angles_out(n_grains,std::vector<double>(3));
-      for (unsigned int grain_i = 0; grain_i < n_grains; ++grain_i)
-        {
-          unsigned int counter = 0;
-          for (unsigned int grain_j = 0; grain_j < n_grains; ++grain_j)
-            {
-              if (cum_weight[grain_j] < idxgrain[grain_i])
-                {
-                  counter++;
-                }
-              else
-                {
-                  break;
-                }
-            }
-          angles_out[grain_i] = angles_sorted[counter];
-        }
-      return angles_out;
-    }
-
 
 
     template <int dim>
--- a/source/postprocess/crystal_preferred_orientation.cc
+++ b/source/postprocess/crystal_preferred_orientation.cc
@@ -464,11 +464,13 @@ namespace aspect
             background_thread_master.join ();
 
           // then continue with writing the master file
-          background_thread_master = std::thread (&writer,
-                                                  filename_master,
-                                                  temporary_output_location,
-                                                  *file_contents_master,
-                                                  false);
+          background_thread_master
+            = std::thread([ my_filename = std::move(filename_master),
+                            my_temporary_output_location = temporary_output_location,
+                            my_file_contents = std::move(file_contents_master)]()
+          {
+            writer (my_filename, my_temporary_output_location, *my_file_contents, false);
+          });
 
           if (write_raw_cpo.size() != 0)
             {
@@ -478,11 +480,14 @@ namespace aspect
                 background_thread_content_raw.join ();
 
               // then continue with writing our own data.
-              background_thread_content_raw = std::thread (&writer,
-                                                           filename_raw,
-                                                           temporary_output_location,
-                                                           *file_contents_raw,
-                                                           compress_cpo_data_files);
+              background_thread_content_raw
+                = std::thread([ my_filename = std::move(filename_raw),
+                                my_temporary_output_location = temporary_output_location,
+                                my_file_contents = std::move(file_contents_raw),
+                                my_compress_cpo_data_files = compress_cpo_data_files]()
+              {
+                writer (my_filename, my_temporary_output_location, *my_file_contents, my_compress_cpo_data_files);
+              });
             }
 
           if (write_draw_volume_weighted_cpo.size() != 0)
@@ -493,11 +498,14 @@ namespace aspect
                 background_thread_content_draw_volume_weighting.join ();
 
               // then continue with writing our own data.
-              background_thread_content_draw_volume_weighting = std::thread (&writer,
-                                                                             filename_draw_volume_weighting,
-                                                                             temporary_output_location,
-                                                                             *file_contents_draw_volume_weighting,
-                                                                             compress_cpo_data_files);
+              background_thread_content_draw_volume_weighting
+                = std::thread([ my_filename = std::move(filename_draw_volume_weighting),
+                                my_temporary_output_location = temporary_output_location,
+                                my_file_contents = std::move(file_contents_draw_volume_weighting),
+                                my_compress_cpo_data_files = compress_cpo_data_files]()
+              {
+                writer (my_filename, my_temporary_output_location, *my_file_contents, my_compress_cpo_data_files);
+              });
             }
         }
       else
--- a/source/postprocess/crystal_preferred_orientation.cc
+++ b/source/postprocess/crystal_preferred_orientation.cc
@@ -1,0 +1,885 @@
+/*
+  Copyright (C) 2022 by the authors of the ASPECT code.
+ This file is part of ASPECT.
+ ASPECT is free software; you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation; either version 2, or (at your option)
+ any later version.
+ ASPECT is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+ You should have received a copy of the GNU General Public License
+ along with ASPECT; see the file LICENSE.  If not see
+ <http://www.gnu.org/licenses/>.
+ */
+
+#include <aspect/global.h>
+#include <aspect/postprocess/crystal_preferred_orientation.h>
+#include <aspect/particle/property/crystal_preferred_orientation.h>
+#include <aspect/utilities.h>
+
+#include <boost/archive/text_oarchive.hpp>
+#include <boost/archive/text_iarchive.hpp>
+
+#include <boost/iostreams/filtering_streambuf.hpp>
+#include <boost/iostreams/copy.hpp>
+#include <boost/iostreams/filter/gzip.hpp>
+
+#include <stdio.h>
+#include <unistd.h>
+
+namespace aspect
+{
+  namespace Postprocess
+  {
+    template <int dim>
+    CrystalPreferredOrientation<dim>::CrystalPreferredOrientation ()
+      :
+      // the following value is later read from the input file
+      output_interval (0),
+      // initialize this to a nonsensical value; set it to the actual time
+      // the first time around we get to check it
+      last_output_time (std::numeric_limits<double>::quiet_NaN()),
+      output_file_number (numbers::invalid_unsigned_int),
+      group_files(0),
+      write_in_background_thread(false)
+    {}
+
+    template <int dim>
+    CrystalPreferredOrientation<dim>::~CrystalPreferredOrientation ()
+    {
+      // make sure a thread that may still be running in the background,
+      // writing data, finishes
+      if (background_thread_master.joinable())
+        background_thread_master.join ();
+
+      if (background_thread_content_raw.joinable())
+        background_thread_content_raw.join ();
+
+      if (background_thread_content_draw_volume_weighting.joinable())
+        background_thread_content_draw_volume_weighting.join ();
+    }
+
+    template <int dim>
+    void
+    CrystalPreferredOrientation<dim>::initialize ()
+    {
+      const unsigned int my_rank = Utilities::MPI::this_mpi_process(MPI_COMM_WORLD);
+      this->random_number_generator.seed(random_number_seed+my_rank);
+    }
+
+    template <int dim>
+    std::list<std::string>
+    CrystalPreferredOrientation<dim>::required_other_postprocessors () const
+    {
+      return {"particles"};
+    }
+
+
+    template <int dim>
+    // We need to pass the arguments by value, as this function can be called on a separate thread:
+    void CrystalPreferredOrientation<dim>::writer (const std::string filename, //NOLINT(performance-unnecessary-value-param)
+                                                   const std::string temporary_output_location, //NOLINT(performance-unnecessary-value-param)
+                                                   const std::string *file_contents,
+                                                   const bool compress_contents)
+    {
+      std::string tmp_filename = filename;
+      if (temporary_output_location != "")
+        {
+          tmp_filename = temporary_output_location + "/aspect.tmp.XXXXXX";
+
+          // Create the temporary file; get at the actual filename
+          // by using a C-style string that mkstemp will then overwrite
+          std::vector<char> tmp_filename_x (tmp_filename.size()+1);
+          std::strcpy(tmp_filename_x.data(), tmp_filename.c_str());
+          const int tmp_file_desc = mkstemp(tmp_filename_x.data());
+          tmp_filename = tmp_filename_x.data();
+
+          // If we failed to create the temp file, just write directly to the target file.
+          // We also provide a warning about this fact. There are places where
+          // this fails *on every node*, so we will get a lot of warning messages
+          // into the output; in these cases, just writing multiple pieces to
+          // std::cerr will produce an unreadable mass of text; rather, first
+          // assemble the error message completely, and then output it atomically
+          if (tmp_file_desc == -1)
+            {
+              const std::string x = ("***** WARNING: could not create temporary file <"
+                                     +
+                                     tmp_filename
+                                     +
+                                     ">, will output directly to final location. This may negatively "
+                                     "affect performance. (On processor "
+                                     + Utilities::int_to_string(Utilities::MPI::this_mpi_process (MPI_COMM_WORLD))
+                                     + ".)\n");
+
+              std::cerr << x << std::flush;
+
+              tmp_filename = filename;
+            }
+          else
+            close(tmp_file_desc);
+        }
+
+      std::ofstream out(tmp_filename.c_str(), std::ofstream::binary);
+
+      AssertThrow (out, ExcMessage(std::string("Trying to write to file <") +
+                                   filename +
+                                   ">, but the file can't be opened!"))
+
+      // now write and then move the tmp file to its final destination
+      // if necessary
+      if (compress_contents)
+        {
+          namespace bio = boost::iostreams;
+
+          std::stringstream compressed;
+          std::stringstream origin(*file_contents);
+
+          bio::filtering_streambuf<bio::input> compress_out;
+          compress_out.push(bio::zlib_compressor());
+          compress_out.push(origin);
+          bio::copy(compress_out, out);
+        }
+      else
+        {
+          out << *file_contents;
+        }
+
+      out.close ();
+
+      if (tmp_filename != filename)
+        {
+          std::string command = std::string("mv ") + tmp_filename + " " + filename;
+          int error = system(command.c_str());
+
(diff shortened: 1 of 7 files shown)


MFraters:
Now that #4978 is merged, I have rebased this pull request to the current main branch. It should now be fully ready.

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+  namespace Postprocess
+  {
+    /**
+     * A Postprocessor that creates particles, which follow the
+     * velocity field of the simulation. The particles can be generated
+     * and propagated in various ways and they can carry a number of
+     * constant or time-varying properties. The postprocessor can write
+     * output positions and properties of all particles at chosen intervals,
+     * although this is not mandatory. It also allows other parts of the
+     * code to query the particles for information.
```
gassmoeller, line 44: Update

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with ASPECT; see the file LICENSE.  If not see
+ <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _aspect_postprocess_particle_crystal_preferred_orientation_h
```
gassmoeller, line 21: Shouldnt "crystal_preferred_orientation" be enough as a name for the postprocessor? "particle" just happens to be the method we are using at the moment.

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+         * are collected throughout the simulation and that will be written to
+         * an output file at the end of each time step. Postprocessors may
+         * deposit data in these tables for later visualization or further
+         * processing.
+         *
+         * @return A pair of strings that will be printed to the screen after
+         * running the postprocessor in two columns; typically the first
+         * column contains a description of what the data is and the second
+         * contains a numerical value of this data. If there is nothing to
+         * print, simply return two empty strings.
```
gassmoeller, line 81: Update this description to be specific for this postprocessor, not the general one from the interface.

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+        parse_parameters (ParameterHandler &prm);
+
+      private:
+
+
+        /**
+         * Get volume weighted euler angles, using random draws to convert
+         * to a discrete number of orientations, weighted by volume.
+         */
+        std::vector<std::vector<double>> random_draw_volume_weighting(std::vector<double> fv,
```
gassmoeller, line 115: hand over the two vectors as references

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+                                                                       std::vector<std::vector<double>> angles) const;
+
+        /**
+        * Wraps angle between 0 and 360 degrees.
+        */
+        double wrap_angle(const double angle) const;
+
+        /**
+         * Compute Z-X-Z euler angles from rotation matrix
```
gassmoeller, line 124: what is Z-X-Z?

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+         */
+        std::vector<double> euler_angles_from_rotation_matrix(const Tensor<2,3> &rotation_matrix) const;
```
gassmoeller, line 126: I think this function and the one below would be good to have in utilities.cc. Maybe move there.

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+
+        /**
+         * Compute rotation matrix from Z-X-Z euler angles
+         */
+        Tensor<2,3> euler_angles_to_rotation_matrix(double phi1, double theta, double phi2) const;
+
+        double end_time;
```
gassmoeller, line 133: documentation

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+
+        /**
+         * Enums specifying what information write
```
gassmoeller, line 136:
```suggestion
         * Enums specifying what information to write:
```

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+         * RotationMatrix: Write the whole rotation matrix of a grain (in Deal.ii order)
+         * EulerAngles: Write out the Z-X-Z Euler angle of a grain
+         * not_found: Error enum, the requested output was not found
+         */
+        enum class Output
+        {
+          VolumeFraction, RotationMatrix, EulerAngles, not_found
+        };
+
+        Output string_to_output_enum(std::string string);
```
gassmoeller, line 148:
```suggestion
        Output string_to_output_enum(const std::string &string);
```

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+
+        const double rad_to_degree = 180.0/M_PI;
+        const double degree_to_rad = M_PI/180.0;
```
gassmoeller, line 151: use `numbers::PI` instead to be consistent with rest of code

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+
+        mutable std::mt19937            random_number_generator;
+
+        unsigned int random_number_seed;
```
gassmoeller, line 155: documentation

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+
+        /**
+         * number of minerals in each particle
+         */
+        unsigned int n_minerals;
+
+        /**
+         * number of minerals in each grain
+         */
+        unsigned int n_grains;
```
gassmoeller, line 165: shouldnt this be information in the particle property?

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+        bool compute_raw_euler_angles;
+
+        /**
+         * Handle to a thread that is used to write content file data in the
+         * background. The writer() function runs on this background thread.
+         */
+        std::thread background_thread_content_raw;
+
+        /**
+         * What draw volume weighted cpo data to write out
```
gassmoeller, line 280: I do not understand what you want to say with this comment.

include/aspect/postprocess/particle_crystal_preferred_orientation.h
```diff
@@ -0,0 +1,322 @@
...
+        std::thread background_thread_content_draw_volume_weighting;
+
+        /**
+         * Whether to compress the raw and weighed cpo data output files with zlib.
+         */
+        bool compress_cpo_data_files;
+
+        /**
+         * Stores the particle property fields which are ouptut to the
+         * visualization file.
```
gassmoeller, line 302: this comment seems wrong given the name of the variable

source/postprocess/particle_crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1036 @@
...
+    }
+
+    template <int dim>
+    std::pair<std::string,std::string>
+    ParticleCrystalPreferredOrientation<dim>::execute (TableHandler &statistics)
+    {
+
+      const Particle::Property::Manager<dim> &manager = this->get_particle_world().get_property_manager();
+
+      bool hexagonal_plugin_exists = manager.plugin_name_exists("decompose elastic matrix");
```
gassmoeller, line 173: Is this plugin part of another open PR?
```suggestion
      const bool hexagonal_plugin_exists = manager.plugin_name_exists("decompose elastic matrix");
```

source/postprocess/particle_crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1036 @@
...
+      prm.enter_subsection("Postprocess");
+      {
+        prm.enter_subsection("Particles");
+        {
+          prm.enter_subsection("Crystal Preferred Orientation");
+          {
+            prm.declare_entry ("Number of grains per praticle", "50",
+                               Patterns::Integer (0),
+                               "The number of grains of olivine and the number of grain of enstatite "
+                               "each particle contains.");
```
gassmoeller, line 765: You dont need to declare this again, if it is already declared in the particle property. Same below with the other parameters.

source/postprocess/particle_crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,1036 @@
...
+  {
+    ASPECT_REGISTER_POSTPROCESSOR(ParticleCrystalPreferredOrientation,
+                                  "crystal preferred orientation",
+                                  "A Postprocessor that creates particles that follow the "
+                                  "velocity field of the simulation. The particles can be generated "
+                                  "and propagated in various ways and they can carry a number of "
+                                  "constant or time-varying properties. The postprocessor can write "
+                                  "output positions and properties of all particles at chosen intervals, "
+                                  "although this is not mandatory. It also allows other parts of the "
+                                  "code to query the particles for information.")
```
gassmoeller, line 1034: Update.

gassmoeller (commented):
A bunch of first comments on the header and parameters, but not on the implementation.

Just for my understanding: Is this essentially a copy of the particle postprocessor, except it does only process the CPO particle properties in a special way before outputting these postprocessed quantities into files that are separate from the normal particle output files?

include/aspect/postprocess/particle_crystal_preferred_orientation.h
MFraters: Because that is what D-Rex uses. I will add that in the description.

source/postprocess/particle_crystal_preferred_orientation.cc
MFraters: That pull request is not open yet since it depends on #4987, which depends on this pull request. I will remove it for now and add it (and the related code) back in that pull request.

include/aspect/postprocess/particle_crystal_preferred_orientation.h
MFraters: I did that to indicate that it build on the particle plugin. I will change it.

include/aspect/postprocess/particle_crystal_preferred_orientation.h
MFraters: sorry, I read why, not what. With Euler angles you have to three rotations about axis. This form of Euler angles first does a rotation around the Z axis, then around the X axis and then around the Z axis again. I will clarify it a bit more in the documentation.

MFraters:
Thanks for the review! I have addressed your initial comments :)

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+      std::string particle_file_prefix_content_raw = this->get_output_directory() +  "particle_CPO/CPO-" + Utilities::int_to_string (output_file_number, 5);
+      std::string particle_file_prefix_content_draw_volume_weighting = this->get_output_directory() +  "particle_CPO/weighted_CPO-" + Utilities::int_to_string (output_file_number, 5);
+
+      const auto &particle_handler = this->get_particle_world().get_particle_handler();
+
+      std::stringstream string_stream_master;
+      std::stringstream string_stream_content_raw;
+      std::stringstream string_stream_content_draw_volume_weighting;
+
+      string_stream_master << "id x y" << (dim == 3 ? " z" : " ") << " olivine_deformation_type" << std::endl;
```
gassmoeller, line 207: currently you create a double space in 2D.

```suggestion
      string_stream_master << "id x y" << (dim == 3 ? " z" : "") << " olivine_deformation_type" << std::endl;
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+          AssertThrow(property_information.fieldname_exists("cpo mineral 0 type") ,
+                      ExcMessage("No CPO particle properties found. Make sure that the CPO particle property plugin is selected."));
+
+
+
+          const unsigned int cpo_data_position = property_information.n_fields() == 0
+                                                 ?
+                                                 0
+                                                 :
+                                                 property_information.get_position_by_field_name("cpo mineral 0 type");
```
gassmoeller, line 234: Move lines 223-234 outside the loop over all particles, they are the same for all particles.

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+
+          Point<dim> position = it->get_location();
```
gassmoeller, line 236:
```suggestion
          const Point<dim> position = it->get_location();
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+
+          // always make a vector of rotation matrices
+          std::vector<std::vector<Tensor<2,3>>> rotation_matrices;
+          rotation_matrices.resize(n_minerals);
```
gassmoeller, line 240: I would move this outside the loop over all particles as well. Also you can immediately create the correct vector size like this:
```suggestion
          std::vector<std::vector<Tensor<2,3>>> rotation_matrices (n_minerals, {n_grains});
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+          for (unsigned int mineral_i = 0; mineral_i < n_minerals; ++mineral_i)
+            {
+              rotation_matrices[mineral_i].resize(n_grains);
+              for (unsigned int i_grain = 0; i_grain < n_grains; i_grain++)
```
gassmoeller, line 244:
```suggestion
              for (unsigned int i_grain = 0; i_grain < n_grains; ++i_grain)
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+                                                 0
+                                                 :
+                                                 property_information.get_position_by_field_name("cpo mineral 0 type");
+
+          Point<dim> position = it->get_location();
+
+          // always make a vector of rotation matrices
+          std::vector<std::vector<Tensor<2,3>>> rotation_matrices;
+          rotation_matrices.resize(n_minerals);
+          for (unsigned int mineral_i = 0; mineral_i < n_minerals; ++mineral_i)
```
gassmoeller, line 241: This naming scheme is inconsistent with grain below. Also we usually to not append _i to counter variables. Either use `i,j,...` or use `mineral, grain, ...`.

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+                {
+                  rotation_matrices[mineral_i][i_grain] = cpo_particle_property.get_rotation_matrix_grains(
+                                                            cpo_data_position,
+                                                            properties,
+                                                            mineral_i,
+                                                            i_grain);
+                }
+            }
+          // write master file
+          string_stream_master << id << " " << position << " " << properties[cpo_data_position] <<  std::endl;
```
gassmoeller, line 254:
```suggestion
          string_stream_master << id << " " << position << " " << properties[cpo_data_position] << std::endl;
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+
+          // write content file
+          std::vector<std::vector<std::vector<double>>> euler_angles;
```
gassmoeller, line 257: same as above, move outside loop, and:
```suggestion
          std::vector<std::vector<std::vector<double>>> euler_angles(n_minerals,{n_grains});
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+          if (compute_raw_euler_angles == true)
+            {
+              euler_angles.resize(n_minerals);
+              for (unsigned int mineral_i = 0; mineral_i < n_minerals; ++mineral_i)
```
gassmoeller, line 261: same as above for the names of the variables here and in the loop below.

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+      std::stringstream string_stream_master;
+      std::stringstream string_stream_content_raw;
+      std::stringstream string_stream_content_draw_volume_weighting;
+
+      string_stream_master << "id x y" << (dim == 3 ? " z" : " ") << " olivine_deformation_type" << std::endl;
+
+      // get particle data
+      bool wrote_weighted_header = false;
+      bool wrote_unweighted_header = false;
+      for (typename dealii::Particles::ParticleHandler<dim>::particle_iterator it = particle_handler.begin(); it != particle_handler.end(); ++it)
```
gassmoeller, line 212: there is a better way to iterate over particles now using range-based for loops:
```suggestion
      for (const auto &particle: particle_handler)
```
You can then access particles by reference (e.g. `particle.get_location()`).

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+
+                  for (unsigned int i_grain = 0; i_grain < n_grains; i_grain++)
+                    {
+                      weighted_euler_angles[mineral_i][i_grain] = Utilities::zxz_euler_angles_from_rotation_matrix(
+                                                                    weighted_rotation_matrices[mineral_i][i_grain]);
+                    }
+                }
+
+              // write weighted header
+              if (wrote_weighted_header == false)
```
gassmoeller, line 379: why not move writing the headers outside of the loop as well (to line 207 where you write the other header). That saves you the awkward check if you have already written the header. You could even create a new function `write_headers(std::stringstream &, std::stringstream &, std::stringstream &)` to shorten this function.

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+                            Assert(false, ExcMessage("Internal error: raw CPO postprocess case not found."));
+                            break;
+                        }
+                    }
+                  string_stream_content_draw_volume_weighting << std::endl;
+                  wrote_weighted_header = true;
+                }
+
+              // write data
+              for (unsigned int grain_i = 0; grain_i < n_grains; ++grain_i)
```
gassmoeller, line 417: same comment as above about the names of counting variables

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+                                   filename +
+                                   ">, but the file can't be opened!"))
+
+      // now write and then move the tmp file to its final destination
+      // if necessary
+      if (compress_contents)
+        {
+          namespace bio = boost::iostreams;
+
+          std::stringstream compressed;
```
gassmoeller, line 136: compressed is not used as far as I can see

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+            }
+        }
+
+      std::string filename_master = particle_file_prefix_master + "." + Utilities::int_to_string(dealii::Utilities::MPI::this_mpi_process (MPI_COMM_WORLD),4) + ".dat";
+      std::string filename_raw = particle_file_prefix_content_raw + "." + Utilities::int_to_string(dealii::Utilities::MPI::this_mpi_process (MPI_COMM_WORLD),4) + ".dat";
+      std::string filename_draw_volume_weighting = particle_file_prefix_content_draw_volume_weighting + "." + Utilities::int_to_string(dealii::Utilities::MPI::this_mpi_process (MPI_COMM_WORLD),4) + ".dat";
+
+      std::string *file_contents_master = new std::string (string_stream_master.str());
+      std::string *file_contents_raw = new std::string (string_stream_content_raw.str());
+      std::string *file_contents_draw_volume_weighting = new std::string (string_stream_content_draw_volume_weighting.str());
```
gassmoeller, line 456: This works, but looks weird. I realize we do the same inside particles.cc. But at least use the better way of doing it from visualization.cc:570-575 where we use a std::unique_ptr instead of a raw pointer.

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+          const double magic = 1.0+2.0*std::numeric_limits<double>::epsilon();
+          last_output_time = last_output_time + std::floor((current_time-last_output_time)/output_interval*magic) * output_interval/magic;
+        }
+    }
+
+    template <int dim>
+    typename CrystalPreferredOrientation<dim>::Output
+    CrystalPreferredOrientation<dim>::string_to_output_enum(const std::string &string)
+    {
+      //olivine volume fraction, olivine A matrix, olivine Euler angles, enstatite volume fraction, enstatite A matrix, enstatite Euler angles
```
gassmoeller, line 547:
```suggestion
      // olivine volume fraction, olivine A matrix, olivine Euler angles, enstatite volume fraction, enstatite A matrix, enstatite Euler angles
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+                             "Units: years if the "
+                             "'Use years in output instead of seconds' parameter is set; "
+                             "seconds otherwise.");
+
+          prm.declare_entry ("Random number seed", "1",
+                             Patterns::Integer (0),
+                             "The seed used to generate random numbers. This will make sure that "
+                             "results are reproducable as long as the problem is run with the "
+                             "same amount of MPI processes. It is implemented as final seed = "
+                             "user seed + MPI Rank. ");
```
gassmoeller, line 580:
```suggestion
                             "random number seed + MPI Rank. ");
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+
+          prm.declare_entry ("Write in background thread", "false",
+                             Patterns::Bool(),
+                             "File operations can potentially take a long time, blocking the "
+                             "progress of the rest of the model run. Setting this variable to "
+                             "`true' moves this process into a background threads, while the "
```
gassmoeller, line 586:
```suggestion
                             "`true' moves this process into background threads, while the "
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+      {
+        prm.enter_subsection("Particles");
+        {
+          prm.enter_subsection("Crystal Preferred Orientation");
+          {
+            prm.enter_subsection("Initial grains");
+            {
+              // Static variable of CPO has not been initialize yet, so we need to get it directly.
+              const std::vector<std::string> temp_deformation_type_selector = dealii::Utilities::split_string_list(prm.get("Minerals"));
+              n_minerals = temp_deformation_type_selector.size();
```
gassmoeller, line 661: this variable has the wrong name

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,816 @@
...
+          if (this->convert_output_to_years())
+            output_interval *= year_in_seconds;
+
+          random_number_seed = prm.get_integer ("Random number seed");
+
+          //AssertThrow(this->get_parameters().run_postprocessors_on_nonlinear_iterations == false,
+          //            ExcMessage("Postprocessing nonlinear iterations in models with "
+          //                       "particles is currently not supported."));
+
+          aspect::Utilities::create_directory (this->get_output_directory() + "particle_CPO/",
```
gassmoeller, line 680: stay consistent with the existing directory name

```suggestion
          aspect::Utilities::create_directory (this->get_output_directory() + "particles_CPO/",
```

But ideally I would prefer an even more expressive name. Something like `particles_cpo_properties`?

gassmoeller (commented):
All the comments for today. I mostly made it through the implementation. I kind of dislike how much code we have to duplicate from the particles postprocessor, but I guess there is nothing we can do about it. Can you please add to the plugin documentation a few sentences of the form: `The work of this postprocessor should better be done by the main particles postprocessor, however we need XYZ which does not work with that postprocessor. XYZ is necessary, because ... . If XYZ is added to the other postprocessor in the future this one becomes obsolete.`?

source/postprocess/crystal_preferred_orientation.cc
MFraters: I am usually not that picky on naming schemes (and I do find it hard to stay consistent), but I usually actually try to go for something like `grain_i` over `i` or `grain` in loop which last more than one or two lines of code. The reason for that is that single letter variables are impossible to search for, and I use the plain one (`grain`) when using a iterator based loop when it is actually pointing to the grain. The `grain_i` indicates for me that it is looping over grains with a unsigned int which should be used as an index for a vector or array. Is there a reason this should be avoided?

MFraters: sure, sounds good. I will remove the `_i`'s from this code :)

source/postprocess/crystal_preferred_orientation.cc
MFraters: I don't think this is correct. The z also has a space in front of it.

source/utilities.cc
```diff
@@ -2783,6 +2783,99 @@ namespace aspect
     }
 
 
+    double
+    wrap_angle(const double angle)
+    {
+      return angle - 360.0*std::floor(angle/360.0);
```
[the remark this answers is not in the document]
MFraters: I have added unit tests for this

source/postprocess/crystal_preferred_orientation.cc
MFraters: never mind, you are right

MFraters:
Thanks for the review. I have adressed all your comments and it i ready for another review :)

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+                        }
+                    }
+                  string_stream_content_draw_volume_weighting << std::endl;
+                }
+            }
+        }
+
+      std::string filename_master = particle_file_prefix_master + "." + Utilities::int_to_string(dealii::Utilities::MPI::this_mpi_process (MPI_COMM_WORLD),4) + ".dat";
+      std::string filename_raw = particle_file_prefix_content_raw + "." + Utilities::int_to_string(dealii::Utilities::MPI::this_mpi_process (MPI_COMM_WORLD),4) + ".dat";
+      std::string filename_draw_volume_weighting = particle_file_prefix_content_draw_volume_weighting + "." + Utilities::int_to_string(dealii::Utilities::MPI::this_mpi_process (MPI_COMM_WORLD),4) + ".dat";
```
gassmoeller, line 453: Dont assume MPI_COMM_WORLD is the correct communicator, we could have been started with a different one (e.g. as subprocess called from a different MPI program). This is currently not possible, but we want to limit the use of MPI_COMM_WORLD as much as possible to main.cc so that it is easy to extend ASPECT in the future. Use `this->get_mpi_communicator()` instead.

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+          if (background_thread_master.joinable())
+            background_thread_master.join ();
+
+          // then continue with writing the master file
+          background_thread_master
+            = std::thread([ my_filename = std::move(filename_master),
+                            my_temporary_output_location = temporary_output_location,
+                            my_file_contents = std::move(file_contents_master)]()
+          {
+            writer (my_filename, my_temporary_output_location, *my_file_contents, false);
```
gassmoeller, line 472: Any reason compressing this file is never allowed?

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+                                my_file_contents = std::move(file_contents_draw_volume_weighting),
+                                my_compress_cpo_data_files = compress_cpo_data_files]()
+              {
+                writer (my_filename, my_temporary_output_location, *my_file_contents, my_compress_cpo_data_files);
+              });
+            }
+        }
+      else
+        {
+          writer(filename_master,temporary_output_location,*file_contents_master, false);
```
gassmoeller, line 513: same question as above

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+                             "On large clusters it can be advantageous to first write the "
+                             "output to a temporary file on a local file system and later "
+                             "move this file to a network file system. If this variable is "
+                             "set to a non-empty string it will be interpreted as a "
+                             "temporary storage location.");
+
+          prm.declare_entry ("Write out raw cpo data",
+                             "olivine volume fraction,olivine Euler angles,enstatite volume fraction,enstatite Euler angles",
+                             Patterns::List(Patterns::Anything()),
+                             "A list containing the what part of the particle cpo data needs "
```
gassmoeller, line 613:
```suggestion
                             "A list containing what particle cpo data needs "
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+                             "to be written out after the particle id. This writes out the raw "
+                             "cpo data files for each MPI process. It can write out the following data: "
+                             "olivine volume fraction, olivine A matrix, olivine Euler angles, "
```
gassmoeller, line 616: here you list three options for both enstatite and olivine, but in line 611 there are only 2 options per mineral (the A matrix does not exist as option). Why is that?

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+            {
+              std::vector<std::string> split_raw_cpo_instructions = Utilities::split_string_list(write_raw_cpo_list[i],':');
+
+              AssertThrow(split_raw_cpo_instructions.size() == 2,
+                          ExcMessage("Value \""+ write_raw_cpo_list[i] +"\", set in \"Write out raw cpo data\", is not a correct option "
+                                     + "because it should contain a mineral identification and a output specifier seprated by a colon (:). This entry "
+                                     + "does not follow those rules."));
+
+              // get mineral number
+              std::vector<std::string> mineralnstructions = Utilities::split_string_list(split_raw_cpo_instructions[0],' ');
```
gassmoeller, line 725:
```suggestion
              std::vector<std::string> mineral_instructions = Utilities::split_string_list(split_raw_cpo_instructions[0],' ');
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+              AssertThrow(mineralnstructions.size() == 2,
+                          ExcMessage("Value \""+ write_raw_cpo_list[i] +"\", set in \"Write out raw cpo data\", is not a correct option "
+                                     + "because the mineral identification part should contain two elements, the word mineral and a number."));
+
+              AssertThrow(mineralnstructions[0] == "mineral",
+                          ExcMessage("Value \""+ write_raw_cpo_list[i] +"\", set in \"Write out raw cpo data\", is not a correct option "
+                                     + "because the mineral identification part should start with the word mineral and it starts with \""
+                                     + mineralnstructions[0] + "\"."));
+
+              int mineral_number = Utilities::string_to_int(mineralnstructions[1]);
```
gassmoeller, line 735: why not `unsigned int`?

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+            {
+              std::vector<std::string> split_draw_volume_weighted_cpo_instructions = Utilities::split_string_list(write_draw_volume_weighted_cpo_list[i],':');
+
+              AssertThrow(split_draw_volume_weighted_cpo_instructions.size() == 2,
+                          ExcMessage("Value \""+ write_draw_volume_weighted_cpo_list[i] +"\", set in \"Write out draw volume weighted cpo data\", is not a correct option "
+                                     + "because it should contain a mineral identification and a output specifier seprated by a colon (:). This entry "
+                                     + "does not follow those rules."));
+
+              // get mineral number
+              std::vector<std::string> mineralnstructions = Utilities::split_string_list(split_draw_volume_weighted_cpo_instructions[0],' ');
```
gassmoeller, line 767:
```suggestion
              std::vector<std::string> mineral_instructions = Utilities::split_string_list(split_draw_volume_weighted_cpo_instructions[0],' ');
```

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+              AssertThrow(mineralnstructions.size() == 2,
+                          ExcMessage("Value \""+ write_draw_volume_weighted_cpo_list[i] +"\", set in \"Write out draw volume weighted cpo data\", is not a correct option "
+                                     + "because the mineral identification part should contain two elements, the word mineral and a number."));
+
+              AssertThrow(mineralnstructions[0] == "mineral",
+                          ExcMessage("Value \""+ write_draw_volume_weighted_cpo_list[i] +"\", set in \"Write out draw volume weighted cpo data\", is not a correct option "
+                                     + "because the mineral identification part should start with the word mineral and it starts with \""
+                                     + mineralnstructions[0] + "\"."));
+
+              int mineral_number = Utilities::string_to_int(mineralnstructions[1]);
```
gassmoeller, line 777: same question as above

gassmoeller (commented):
Getting closer. A number of smaller comments, but the rest looks good.

source/postprocess/crystal_preferred_orientation.cc
MFraters: This is an index with not that much data relative to the other files. Compressing this will make reading it out a bit more annoying. It can be solved, but I didn't think it was worth it.

source/postprocess/crystal_preferred_orientation.cc
MFraters: There are 3 options, but the A matrix was changed to rotation matrix. I have updated the names in the documents now.

source/postprocess/crystal_preferred_orientation.cc
MFraters: There is no string_to_unsigned_int in dealii utilities. I could make a pull request to add it, but it would take a year until we can start using it.

MFraters:
Thanks for the review. I have addressed your comments.

source/postprocess/crystal_preferred_orientation.cc
```diff
@@ -0,0 +1,833 @@
...
+                                     + "because the mineral identification part should contain two elements, the word mineral and a number."));
+
+              AssertThrow(mineral_instructions[0] == "mineral",
+                          ExcMessage("Value \""+ write_raw_cpo_list[i] +"\", set in \"Write out raw cpo data\", is not a correct option "
+                                     + "because the mineral identification part should start with the word mineral and it starts with \""
+                                     + mineral_instructions[0] + "\"."));
+
+              unsigned int mineral_number = Utilities::string_to_int(mineral_instructions[1]);
+
+              AssertThrow((unsigned int) mineral_number < n_minerals,
```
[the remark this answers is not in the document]
gassmoeller: And unify whatever you do here with lline 777

gassmoeller (commented):
A few comments. And A_matrix still appears in the code.

MFraters:
Thanks, I have addressed your comments.

MFraters:
Fixed the changlog entry

gassmoeller (approved):

PR #4986, opened 2022-10-14
Remove outdated and wrong option in signal_handler for program termination

### LINKED ISSUE #4984: Possible MPI Deadlock in GMG preconditioner
As discussed yesterday with @tjhei we (@jdannberg, @RanpengLi and myself) are investigating a problem that looks like a MPI dead lock inside the GMG preconditioner.  We are still working to break it down to a simple model (we currently need 64-128 processes and more than 10 hours of runtime to reproduce). I attach a log.txt and two stack traces of two different processes below.

Observations so far:
1. The problem seems to be reproducible (running the same model stops in the same time step).
2. The model seems to happen consistently for all models with similar parameters. We have not checked significantly different models so far.
3. Restarting a stopped model allows us to continue running further than the previous stop point.
4. Switching from GMG preconditioner to AMG preconditioner resolves the issue (we can run without deadlock models that would always stop for GMG).
5. We can successfully run the same model on different hardware, compiler, and MPI versions, while using the same ASPECT and deal.II. Environment that crashes: Intel 19.1.0.166 / openmpi 4.1.1. Environment that works: GCC 9.4.0 / openmpi 4.0.3. 

Analyzing the stacktrace shows the following:
1. Both processes are stuck within AffineConstraints::distribute, when creating a new distributed vector and calling `Partitioner::set_ghost_indices`, which calls `ConsensusAlgorithms::NBX::run` (just a simplified summary of the stacktrace).
2. However the two processes are stuck in different places that should not be simultaneously reachable:
  - One is stuck in the MPI_Barrier in the destructor of the ScopedLock created [here](https://github.com/dealii/dealii/blob/b2edb79603b8f8089c4508e0dd4ecec9ddd4e81b/include/deal.II/base/mpi_consensus_algorithms.h#L1602). This suggests to me this process is done with the function and in the process of jumping back to the calling function.
  - The other is stuck in MPI_Test and the only MPI_Test I found in the algorithm is inside `all_remotely_originated_receives_are_completed()` [here](https://github.com/dealii/dealii/blob/b2edb79603b8f8089c4508e0dd4ecec9ddd4e81b/include/deal.II/base/mpi_consensus_algorithms.h#L1900).
  - However, the MPI_Test in question checks if all processes have completed the MPI_IBarrier that is placed further up in the function. So the first process must have passed this test already, while the second one is stuck. In other words, it looks like the return value of the MPI_Test makes some processes believe everyone has passed the MPI_IBarrier, while some others are not notified that the MPI_IBarrier has been passed by every process (and therefore wait endlessly for completion). I am not sure how this can happen (is it possible to get here is one process throws an exception inside the function?).

Things we test at the moment:
- We are in the process of running the model again on the failing hardware with Intel 19.1.0.166 / openmpi 4.0.3 to see if something in openmpi 4.1.1 is causing the issue. We will afterwards look into using gcc instead of Intel compiler.
- We will try running the failing model with the latest deal.II instead of v9.4.0.

Other ideas to test are appreciated. I will update this issue when we find out more.

[test.48589940.log.txt](https://github.com/geodynamics/aspect/files/9779274/test.48589940.log.txt)
[126267.txt](https://github.com/geodynamics/aspect/files/9779261/126267.txt)
[126279.txt](https://github.com/geodynamics/aspect/files/9779262/126279.txt)

gassmoeller:
While looking more into a solution for #4984 I found this wrong code in ASPECT's signal handler. `DEAL_II_USE_CXX11` is no longer set in deal.II, meaning we consistently used the pre C++11 option in ASPECT since deal.II 9.3.0

I have not yet tested if this resolves #4984, but  so far I was unsuccessful to create a test purely in deal.II to create the same deadlock that seems to happen in the ASPECT model (so maybe us messing with the signal_handler is involved).

### DIFF
--- a/source/main.cc
+++ b/source/main.cc
@@ -511,16 +511,10 @@ void signal_handler(int signal)
     {
       std::cerr << "Unexpected signal " << signal << " received\n";
     }
-#if DEAL_II_USE_CXX11
-  // Kill the program without performing any other cleanup, which is likely to
-  // lead to a deadlock
+
+  // Kill the program without performing any other cleanup, which would likely
+  // lead to a deadlock.
   std::_Exit(EXIT_FAILURE);
-#else
-  // Kill the program, or at least try to. The problem when we get here is
-  // that calling std::exit invokes at_exit() functions that may still hang
-  // the MPI system
-  std::exit(1);
-#endif
 }
 
 


tjhei (approved):
Curious if this makes a difference in your case.

gassmoeller:
Doesnt seem to make a difference, but it is right change anyway.

PR #4988, opened 2022-10-20
Call MPI_Abort after handling exception

gassmoeller:
If a subset of processes throw an exception during communication, we can end up in a deadlock when the destructor of `MPI_InitFinalize` is called. `MPI_InitFinalize` checks for uncaught exceptions, but the exception was already correctly caught by the catch block we have in ASPECT. We should call `MPI_Abort`, just like deal.II's step-50 already does (https://github.com/dealii/dealii/blob/aa053f40abf916e6b3acdb843b636b7cd577f0e7/examples/step-50/step-50.cc#L1588).

### DIFF
--- a/tests/advection_solver_fail/screen-output
+++ b/tests/advection_solver_fail/screen-output
@@ -60,3 +60,4 @@ Additional information:
     
 
 Aborting!
+(rest of the output replaced by default.sh script)
--- a/source/main.cc
+++ b/source/main.cc
@@ -803,6 +803,8 @@ int main (int argc, char *argv[])
                 << "Aborting!" << std::endl
                 << "----------------------------------------------------"
                 << std::endl;
+
+      MPI_Abort(MPI_COMM_WORLD, 1);
       return 1;
     }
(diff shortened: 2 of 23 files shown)


tjhei (approved):
Looks good but you will need to update test results.
I have no recollection of adding the MPI_Abort in step-50, but it must have been me while debugging... 🤷‍♂️

tests/advection_solver_fail/screen-output
```diff
@@ -60,3 +60,9 @@ Additional information:
     
 
 Aborting!
+MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
```
tjhei, line 63: Would it make sense to filter these out in the default script? It certainly won't work with mpich if we leave it in...

tjhei (approved):

gassmoeller:
I added the lines to the replacement script. Turns out we already had one output committed. Ready to go from my side if the tester is happy.

tjhei (approved):

Issue #4995, opened 2022-11-01
CPO doxygen warnings


tjhei:
```
/ssd/aspect-git-clean/include/aspect/particle/property/crystal_preferred_orientation.h:181: warning: argument 'volume_fractions' of command @param is not found in the argument list of aspect::Particle::Property::CrystalPreferredOrientation< dim >::compute_derivatives(const unsigned int cpo_index, const ArrayView< double > &data, const unsigned int mineral_i, const SymmetricTensor< 2, 3 > &strain_rate, const Tensor< 2, 3 > &velocity_gradient_tensor, const std::array< double, 4 > &ref_resolved_shear_stress) const
/ssd/aspect-git-clean/include/aspect/particle/property/crystal_preferred_orientation.h:181: warning: argument 'rotation_matrices' of command @param is not found in the argument list of aspect::Particle::Property::CrystalPreferredOrientation< dim >::compute_derivatives(const unsigned int cpo_index, const ArrayView< double > &data, const unsigned int mineral_i, const SymmetricTensor< 2, 3 > &strain_rate, const Tensor< 2, 3 > &velocity_gradient_tensor, const std::array< double, 4 > &ref_resolved_shear_stress) const
/ssd/aspect-git-clean/include/aspect/particle/property/crystal_preferred_orientation.h:181: warning: argument 'volume_fraction_mineral' of command @param is not found in the argument list of aspect::Particle::Property::CrystalPreferredOrientation< dim >::compute_derivatives(const unsigned int cpo_index, const ArrayView< double > &data, const unsigned int mineral_i, const SymmetricTensor< 2, 3 > &strain_rate, const Tensor< 2, 3 > &velocity_gradient_tensor, const std::array< double, 4 > &ref_resolved_shear_stress) const
/ssd/aspect-git-clean/include/aspect/particle/property/crystal_preferred_orientation.h:193: warning: The following parameters of aspect::Particle::Property::CrystalPreferredOrientation::compute_derivatives(const unsigned int cpo_index, const ArrayView< double > &data, const unsigned int mineral_i, const SymmetricTensor< 2, 3 > &strain_rate, const Tensor< 2, 3 > &velocity_gradient_tensor, const std::array< double, 4 > &ref_resolved_shear_stress) const are not documented:
  parameter 'cpo_index'
  parameter 'data'
  parameter 'mineral_i'
[...]

PR #4997, opened 2022-11-01
fix cpo doxygen warning.

MFraters:
resolves issue #4995. I missed updating the documentation of this function when it was changed in the pull request.

### DIFF
--- a/include/aspect/particle/property/crystal_preferred_orientation.h
+++ b/include/aspect/particle/property/crystal_preferred_orientation.h
@@ -187,7 +187,7 @@ namespace aspect
            * @param strain_rate is the strain-rate at the location of the particle.
            * @param velocity_gradient_tensor is the velocity gradient tensor at the location of the particle.
            * @param ref_resolved_shear_stress is the reference resolved shear stress of the mineral.
-           * @return std::pair<std::vector<double>, std::vector<Tensor<2,3>>> A pair containing the derivatives for
+           * @return A pair containing the derivatives for
            * the change is size in the first part and the derivatives for the change in rotation in the second part.
            */
           std::pair<std::vector<double>, std::vector<Tensor<2,3>>>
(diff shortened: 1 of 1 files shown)


include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -179,15 +179,16 @@ namespace aspect
...
-           * @param rotation_matrices are the current rotation matrices of the grains in a mineral.
+           * @param cpo_index The index in the particle data array where the cpo data starts
+           * @param data The particle data array.
+           * @param mineral_i The mineral index for which to compute the derivatives.
            * @param strain_rate is the strain-rate at the location of the particle.
            * @param velocity_gradient_tensor is the velocity gradient tensor at the location of the particle.
-           * @param volume_fraction_mineral is the volume fraction of the current mineral with respect to
-           * the other minerals in the particle.
            * @param ref_resolved_shear_stress is the reference resolved shear stress of the mineral.
+           * @return std::pair<std::vector<double>, std::vector<Tensor<2,3>>> A pair containing the derivatives for
```
tjhei, line 190: I think one typically does not write out the return type
```suggestion
           * @return A pair containing the derivatives for
```

include/aspect/particle/property/crystal_preferred_orientation.h
```diff
@@ -353,6 +354,7 @@ namespace aspect
            * @param data The particle data vector.
            * @param mineral_i The mineral to set the value of the rotation matrix of a grain for.
            * @param grain_i The grain to get the value of the rotation matrix of.
+           * @param rotation_matrix The rotation matrix to set for the grain in the mineral.
```
tjhei, line 357: unrelated, but rotation_matrix should be passed by const reference. Can you open a PR please?

tjhei (approved):

MFraters:
Thanks for the review. I have committed your suggestion to remove the return type. It was there because the vs code plugin which auto generates doxygen documentation made it, but I am fine with removing it since it doesn't really add much.

I will make the pull request to make the rotation matrix in that function a const ref

PR #4998, opened 2022-11-01
Adding recent publications

ljhwang:
Getting caught up by adding 4 recent publications to the citations list.


ljhwang:
Please note that the Holt citation is recent and will need to be updated with proper year, volume, and number. Currently the latter 2 are n/a's. My apologies.

naliboff (approved):
Thanks!

PR #5004, opened 2022-11-28
Description updates

RanpengLi:
I updated the description of some input parameters in grain size model to specify their number of entries.


jdannberg (commented):
The changes you made look good to me, but I think it would make sense to also add this line to the "Grain growth activation energy" (line 1033) and the "Reciprocal required strain" (line 1061), since these are also parameters that we check the size of later on in the file. 

In addition, there is the "Recrystallized grain size" (line 1065), which must have the same number of entries as Phase transition depths.

Thank you for improving this!

RanpengLi:
Thank you! I modified it as you suggested.

jdannberg (approved):
Great, thank you! That looks good to me now!

PR #5007, opened 2022-12-02
Update annulus benchmark and include transient version of the benchmark

gassmoeller:
This is a rework of the 'Annulus' benchmark to improve code, documentation and add a new transient version of the benchmark.

Significant changes:
- Rewrote the plugin code to be easier to understand and change
- Added a transient version of the benchmark
- Added plotting scripts and replaced the manual figure with a version generated by the script

PR #5008, opened 2022-12-02
Update rigid shear benchmark with transient version.

gassmoeller:
Similar to #5007, but for the rigid shear benchmark. Contains an improvement for the accuracy of the instantaneous rigid shear benchmark, a new transient version, and a rewrite of some of the plugin code. I also renamed the unfortunately named `time-dependent` version into `steady-state`, which is a more accurate description.


benchmarks/rigid_shear/plugin/rigid_shear.cc
```diff
@@ -64,25 +65,56 @@ namespace aspect
      */
     namespace AnalyticSolutions
     {
+      double
+      phase_function(const double t)
+      {
+        return std::exp(t)-1;
```
tjhei, line 71:
```suggestion
        return std::exp(t)-1.;
```

benchmarks/rigid_shear/plugin/rigid_shear.cc
```diff
@@ -64,25 +65,56 @@ namespace aspect
...
+              values[0] += std::exp(t);
+
+            values[1] = -std::cos(pi*(p[0]-phase)) * std::sin(pi*p[1]);
+            values[2] = 2.0 * pi * std::cos(pi*(p[0]-phase)) * std::cos(pi*p[1]);
+            values[4] = density(p,t,transient_);
             return;
           }
+
+        private:
+          bool transient_;
```
tjhei, line 117: Not a big fan of name_. It should also work if you use 
```suggestion
          bool transient;
```
(even if the constructor argument has the same name)

benchmarks/rigid_shear/plugin/rigid_shear.cc
```diff
@@ -93,155 +125,128 @@ namespace aspect
...
      */
     template<int dim>
-    class RigidShearMaterial : public MaterialModel::Interface<dim>
+    class RigidShearMaterial : public MaterialModel::Interface<dim>, public SimulatorAccess<dim>
     {
       public:
         virtual void evaluate(const MaterialModel::MaterialModelInputs<dim> &in,
                               MaterialModel::MaterialModelOutputs<dim> &out) const
         {
+          double t = 0.0;
```
tjhei, line 134:
```suggestion
          const double t = (this->simulator_is_past_initialization() ? this->get_time() : 0.0;
```

benchmarks/rigid_shear/plugin/rigid_shear.cc
```diff
@@ -93,155 +125,128 @@ namespace aspect
...
-                                         get_particle_world().
-                                         get_property_manager();
-          const unsigned density_index = property_manager.get_data_info().get_field_index_by_name("function");
-
-          const Point<dim> position = particle->get_location();
-          const double g_y = -4.0 * numbers::PI * numbers::PI * std::cos(numbers::PI * position[0]) / std::sin(numbers::PI * position[0]);
-          particle->get_properties()[data_position] = particle->get_properties()[density_index]*g_y;
-        }
+          const double pi = numbers::PI;
+          double t = 0.0;
```
tjhei, line 230: same here.

tjhei (approved):

tjhei (approved):

tjhei:
Thanks!

PR #5010, opened 2022-12-06
Make grain size work on particles

gassmoeller:
This PR adds a particle property that tracks the grain size on particles. It can be used as a replacement for the field based implementation that was used so far in the grain size material model. It always requires a compositional field named 'grain_size' to be present and the 'grain size' material model to be used (i.e. it is not independent of these two implementations).

I am working on a cookbook that will use this property, for now I just added a test in this PR that shows that the grain size on particles develops as the corresponding grain size in fields.


source/particle/property/grain_size.cc
```diff
@@ -0,0 +1,165 @@
...
+
+
+      template <int dim>
+      void
+      GrainSize<dim>::initialize ()
+      {
+        material_inputs  = MaterialModel::MaterialModelInputs<dim>(1, this->n_compositional_fields());
+        material_outputs = MaterialModel::MaterialModelOutputs<dim>(1, this->n_compositional_fields());
+
+        AssertThrow(Plugins::plugin_type_matches<const MaterialModel::GrainSize<dim>>(this->get_material_model()),
```
tjhei, line 47: Why this limitation? I don't think you access anything directly from this material model, or do you? What if I want to implement a slight variation of that class by copying it?

tjhei (approved):

source/particle/property/grain_size.cc
```diff
@@ -0,0 +1,163 @@
...
+namespace aspect
+{
+  namespace Particle
+  {
+    namespace Property
+    {
+      ASPECT_REGISTER_PARTICLE_PROPERTY(GrainSize,
+                                        "grain size",
+                                        "A plugin in which the particle property is "
+                                        "defined as the evolving grain size of a particle."
```
tjhei, line 157:
```suggestion
                                        "defined as the evolving grain size of a particle. "
```

tjhei (approved):

gassmoeller:
Done

tjhei (approved):

tjhei:
:+1:

PR #5012, opened 2022-12-11
Update strain rheology

sibiaoliu:
This PR involves the same changes to update the strain rheology as #4336, but further addresses the issue of benchmark deviation by following the suggestions of Rene and Anne. I adopted Anne's method (#4198), which creates a copy of the solution for the current linearization point. This may be helpful for applications that will use previous solutions in composition reaction terms at the beginning of the timestep, such as melt_global or elasticity material models.

@naliboff @anne-glerum @gassmoeller Will you have time for a review? Thank you very much!

Next, I will add another test that will combine the free surface or any other test you are interested in.

### For all pull requests:


naliboff:
@sibiaoliu - Thanks for this PR and sorry it took so long to get to it! I will do a full review sometime in the next few days, but before then here is a general question:
-  Is storing the solution for the current linearization point necessary here? 

In #4923 I proposed enabling the iterative advection schemes with the strain rheology as the tests do now show unrealistic accumulation of the strain values through the reaction terms through multiple advection solves. Is there another reason to use the stored linearization point? Let's decide on this point, and then move forward with the full review.

sibiaoliu:
> @sibiaoliu - Thanks for this PR and sorry it took so long to get to it! I will do a full review sometime in the next few days, but before then here is a general question:
> 
> * Is storing the solution for the current linearization point necessary here?
> 
> In #4923 I proposed enabling the iterative advection schemes with the strain rheology as the tests do now show unrealistic accumulation of the strain values through the reaction terms through multiple advection solves. Is there another reason to use the stored linearization point? Let's decide on this point, and then move forward with the full review.

Hi John, after discussion with Anne,  I don't think it is necessary to use the stored linearization point here. What I would have needed was to get the previous solution (in.composition (plastic_strain) to calculate the reaction term for plastic strain.  I will close this PR.

PR #5016, opened 2023-01-10
Fix build of deal.II dev tester image

gassmoeller:
The build of our deal.II development tester docker image was silently failing, because candi now ships Trilinos 13.2, which requires cmake 3.17, which is higher than the cmake package packaged in Ubuntu 20.04 (3.16). This PR fixes the build by using candi to install cmake instead. It also adds badges to the README that will notify us in the future if the build is failing again (currently I only noticed by checking the actions tab at https://github.com/geodynamics/aspect/actions/workflows/build_tester_base_image.yml). I am open to where these badges should go (do not need to be at the top), but I think the number of badges is still acceptable, let me know in case they should go somewhere else.


tjhei (approved):
Sorry for breaking this.

tjhei:
> I am open to where these badges should go (do not need to be at the top), but I think the number of badges is still acceptable, let me know in case they should go somewhere else.

That's fine.

PR #5018, opened 2023-01-11
Always update apt package list before install in tester

gassmoeller:
I have seen a few failures of the no-unity-build tester lately with errors related to installing apt packages (e.g. https://github.com/geodynamics/aspect/actions/runs/3887199332/jobs/6633184425). This may help.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -24,6 +24,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: setup
       run: |
+        sudo apt-get update
         sudo apt-get install -y software-properties-common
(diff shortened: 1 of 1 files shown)


gassmoeller:
Could we merge this PR soonish? We had another test failure in #5021 that I think may have been related to this fix. I mostly want to see if this actually fixes the issue or if we need to dig deeper.

jdannberg (approved):
Well let's hope that fixes it! Thanks for working on a fix for this problem!

tjhei:
We see similar failures in the deal.II repo. I think this is not due to a missing update.

PR #5019, opened 2023-01-11
Some optimizations for the grain size material model

gassmoeller:
The grain size material model always has been a very expensive material model due to the iterations for composite viscosity and the operator splitting for the grain size evolution (requiring many evaluations of the viscosity per time step of the operator splitting). The model is also pretty old and could use some improvements in code quality. These changes are only a first step in improving the quality of the code (I did not want to start a full rewrite), but they significantly speed up the evaluation of the model.

In a simple test model (with cheap solvers) this PR speeds up the evaluate function by about 3x, which results in an about 2x speedup for the total assembly. The material model can still make up several percent of total model runtime (in my worst case test the material model still makes up 25% of the total runtime), but at this point  further optimizations likely require larger changes than they are worth.

@jdannberg FYI

### DIFF
--- a/source/material_model/grain_size.cc
+++ b/source/material_model/grain_size.cc
@@ -244,8 +244,8 @@ namespace aspect
 
           // grain size growth due to Ostwald ripening
           const double m = grain_growth_exponent[phase_index];
-          const double grain_size_growth_rate = grain_growth_rate_constant[phase_index] / (m * pow(grain_size,m-1))
-                                                * exp(- (grain_growth_activation_energy[phase_index] + pressure * grain_growth_activation_volume[phase_index])
+          const double grain_size_growth_rate = grain_growth_rate_constant[phase_index] / (m * std::pow(grain_size,m-1))
+                                                * std::exp(- (grain_growth_activation_energy[phase_index] + pressure * grain_growth_activation_volume[phase_index])
                                                       / (constants::gas_constant * temperature));
           const double grain_size_growth = grain_size_growth_rate * grain_growth_timestep;
 
@@ -271,7 +271,7 @@ namespace aspect
             {
               // paleowattmeter: Austin and Evans (2007): Paleowattmeters: A scaling relation for dynamically recrystallized grain size. Geology 35, 343-346
               const double stress = 2.0 * second_strain_rate_invariant * current_viscosity;
-              const double grain_size_reduction_rate = 2.0 * stress * boundary_area_change_work_fraction[phase_index] * dislocation_strain_rate * pow(grain_size,2)
+              const double grain_size_reduction_rate = 2.0 * stress * boundary_area_change_work_fraction[phase_index] * dislocation_strain_rate * std::pow(grain_size,2)
                                                        / (geometric_constant[phase_index] * grain_boundary_energy[phase_index]);
               grain_size_reduction = grain_size_reduction_rate * grain_growth_timestep;
             }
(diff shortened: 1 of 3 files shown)


tjhei (approved):
Looks reasonable, thanks.

source/material_model/grain_size.cc
```diff
@@ -486,12 +486,33 @@ namespace aspect
...
 
       double effective_viscosity;
       if (std::abs(second_strain_rate_invariant) > 1e-30)
         {
-          const double disl_viscosity = dislocation_viscosity(temperature, pressure, composition, strain_rate, position);
+          const double disl_viscosity = dislocation_viscosity(temperature,
+                                                              pressure,
+                                                              composition,
+                                                              strain_rate,
+                                                              position);
```
jdannberg, line 515: I liked the original version better.

source/material_model/grain_size.cc
```diff
@@ -393,70 +398,38 @@ namespace aspect
...
+      const double adiabatic_temperature = this->get_adiabatic_conditions().is_initialized()
+                                           ?
+                                           this->get_adiabatic_conditions().temperature(position)
+                                           :
+                                           temperature;
       const double adiabatic_pressure = this->get_adiabatic_conditions().is_initialized()
                                         ?
                                         this->get_adiabatic_conditions().pressure(position)
                                         :
                                         pressure;
```
jdannberg, line 410: Just a question: For the diffusion viscosity, you compute this outside the function call. Here, you compute it inside. 
Wouldn't it make more sense to do it the same way for both functions? (Both in terms of speed and readability?)

source/material_model/grain_size.cc
```diff
@@ -486,12 +486,33 @@ namespace aspect
       const SymmetricTensor<2,dim> shear_strain_rate = strain_rate - 1./dim * trace(strain_rate) * unit_symmetric_tensor<dim>();
```
jdannberg, line 486: Do we even need this whole function? It doesn't seem like this is what's being used to fill out.viscosities. 

(And in case this function is being used somewhere, I would also use it in evaluate(), otherwise this is duplicating code).

jdannberg (commented):
I have some minor comments, but otherwise this looks good to me!

gassmoeller:
All good points. I have addressed all of them, which should also de-duplicate some more computations, except for the comment on breaking lines. I personally prefer breaking the arguments onto separate lines once they become unwieldy (it avoids scrolling to the right to check arguments). Do you feel strongly about keeping them on one line?

tests/grain_size_latent_heat.cc
```diff
@@ -173,11 +173,44 @@ namespace aspect
...
+                                                    ?
+                                                    this->get_adiabatic_conditions().pressure(in.position[i])
+                                                    :
+                                                    in.pressure[i];
+
+                  const unsigned int grain_size_index = this->introspection().compositional_index_for_name("grain_size");
+
+                  const double diff_viscosity = this->diffusion_viscosity(in.temperature[i],
+                                                                          adiabatic_temperature,
+                                                                          adiabatic_pressure,                                                                composition[grain_size_index],
```
jdannberg, line 200: I think you forgot to add a new line here.

jdannberg (approved):
Looks good to me! Can you fix the one line change, and then feel free to merge.

gassmoeller:
Testers are happy now. Thanks for the review.

PR #5021, opened 2023-01-13
New shear heating output

jdannberg:
This PR creates a new additional material model output that contains the fraction of work that goes into shear heating. This is useful because the same term is needed both in the material model and in the heating model, and with the new implementation, the shear heating does not have to know about the material model, but each material model could still implement its own model for this term. I have checked that the `DislocationViscosityOutputs`, which were used for that purpose before, are only used in the grain size material model. 

I want to make this change because there are different models for how this work fraction is being computed (some of them including dependencies on solution variables like the temperature), and as a next step I want to implement one of these more complex models. 

### For all pull requests:

### DIFF
--- a/cookbooks/anisotropic_viscosity/av_material.cc
+++ b/cookbooks/anisotropic_viscosity/av_material.cc
@@ -50,11 +50,10 @@
 #include <deal.II/fe/fe_values.h>
 #include <deal.II/dofs/dof_tools.h>
 #include <deal.II/base/geometry_info.h>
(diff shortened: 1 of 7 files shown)


gassmoeller:
The linux / tests (geodynamics/aspect-tester:focal-dealii-master, tester is currently broken, you can ignore that tester for now. I am working on a fix.

jdannberg:
Looks like all the tests are working now except for the tester that is currently broken.

PR #5024, opened 2023-01-17
add a new mesh refinement plugin for the nonadiabatic temperature

jdannberg:
There is now a new mesh refinement plugin called "Nonadiabatic temperature threshold" that uses the temperature
anomaly (compared to the adiabat) to mark additional cells for refinement.

Below is an image that shows the output of the test case I created if run for a longer time (end time of 0.1). The output shows that

1. The mesh is finer where the absolute value of the nonadiabatic temperature exceeds 0.25, as intended;
2. The minimum and maximum refinement levels as given by the minimum and maximum refinement function plugins are still respected; they are finer (4 and 5) in the top half compared to the bottom half (3 and 4). 

Screenshot from 2023-01-17 16-19-20


source/mesh_refinement/nonadiabatic_temperature_threshold.cc
```diff
@@ -0,0 +1,156 @@
...
+        if (cell->is_locally_owned())
+          {
+            bool refine = false;
+
+            fe_values.reinit(cell);
+            fe_values[this->introspection().extractors.temperature].get_function_values (this->get_solution(),
+                                                                                         temperature_values);
+
+            // if the nonadiabatic temperature exceeds the threshold, cell is marked for refinement
+            for (unsigned int j=0; j<this->get_fe().base_element(this->introspection().base_elements.temperature).dofs_per_cell; ++j)
```
tjhei, line 60: Can you compute the loop bound once outside the loop?

tjhei (approved):

source/mesh_refinement/nonadiabatic_temperature_threshold.cc
jdannberg: Like this?

The previous version is how it's done in a couple of other plugins (where I copied this from). Should I go through those and change it as well? I think it would make sense to do this the same way everywhere.

tjhei: I think you can move it outside the cell loop as well. If you have time to do it, feel free to look at other places as well.

jdannberg: Alright, I've moved it out of the cell loop as well, and I'll make a follow-up PR for the other plugins.

source/mesh_refinement/nonadiabatic_temperature_threshold.cc
```diff
@@ -0,0 +1,157 @@
...
+    void
+    NonadiabaticTemperatureThreshold<dim>::
+    declare_parameters (ParameterHandler &prm)
+    {
+      prm.enter_subsection("Mesh refinement");
+      {
+        prm.enter_subsection("Nonadiabatic temperature threshold");
+        {
+          prm.declare_entry ("Threshold",
+                             "0",
```
gassmoeller, line 100: The code looks ready to merge, but I think we want to think about this threshold. '0' in the current setting means everything is always refined as soon as this criterion is selected. Would we prefer a setting where nothing is ever refined if this plugin is selected without changing this parameter? I think that would be easier to understand for new users, but I am open for discussion.
That would then mean:
```suggestion
                             boost::lexical_cast<std::string>(std::numeric_limits<double>::max()),
```

tjhei: I would argue refinishing nothing or everything are not good defaults. We could also leave the default empty so the user is forced to specify it?

jdannberg: I like that idea. I've changed the code to make the default empty.

jdannberg: Update: Leaving it empty does not work (see failed tests below; there might be a way to make it work, but I think it would be more complicated). So I instead changed it to 100 K, which seems like a reasonable default to me.

jdannberg:
Does this look reasonable now or is there anything else I should change?

bobmyhill (approved):
Looks ready to merge to me. Happy to do it if you don't want to check in with @tjhei or @gassmoeller first.

I agree, 100 K is a good default.

PR #5025, opened 2023-01-18
Slab2 model

alarshi:
This PR adds an initial composition model that outputs 1 at the locations of the slab2 model (https://www.science.org/doi/10.1126/science.aat4723) and 0 otherwise. 
The slab2 model is input as an ascii boundary datafile with columns representing depths and thickness at the lat/long points on the Earth spaced at 0.05 degrees. This results in an ascii file that is too big to be added with aspect, and therefore a jupyter notebook is added to format the input slab2 model for an aspect input file.
Below is the paraview output from aspect using higher resolution on the test file.  

slab2

### DIFF
--- a/data/initial-composition/slab-model/create_slab2_database.py
+++ b/data/initial-composition/slab-model/create_slab2_database.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# This script uses the slab2 model data, available to download from
+# https://www.sciencebase.gov/catalog/item/5aa1b00ee4b0b1c392e86467
+# (Choose the file "Slab2Distribute_Mar2018.tar.gz" under the
+# "Attached files" heading and unpack it in this directory).
+# 
+# This script then formats the data in a way that it can be used by
+# the slab model plugin in ASPECT.
+# 
+# Input files: Each input file contains attributes representing either depth
+# "dep", dip ("dip"), strike ("str"), and thickness ("thk") information for
+# each of the 27 slabs in the slab2 model in a structured ascii grid. The files
+# contains columns corresponding to longitude, latitude, and attribute ("dep,
+# "dip" etc). More information on the model can be found in the paper: 
+# Hayes, G. P., Moore, G. L., Portner, D. E., Hearne, M., Flamme, H., Furtney,
+# M., & Smoczyk, G. M.  (2018). Slab2,a comprehensive subduction zone geometry
+# model. Science, 362(6410), 58-61.
+# 
+# Output: A structured file containing all the slabs for the Earth with columns
+# representing longitude, colatitude, slab depths, and slab thickness
+
+import numpy as np
+import os
+from glob import glob
+from scipy.interpolate import griddata
+
+# Save the current working directory
+cwd = os.getcwd()
+
+# Set the path to the downloaded slab2 model here. This path is correct if you
+# download the model from the address above and extract in the current folder.
+os.chdir('./Slab2Distribute_Mar2018/Slab2_TXT')
+
+
+def interpolate_data (name_key, type_key):
+    """
+    From initial data inspection, we found that most slabs have spacing of 0.05 degrees except for 5 slabs,   
+    which have 0.02 degree spacing. Therefore, this function interpolates those slabs to 0.05 spacing for 
+    consistency.
+    """
+
+    file_name = glob(name_key + '*' + type_key +'*.xyz')
+    data_in   = np.loadtxt (file_name[0], delimiter=",")
+    
+    # We convert latitudes to co-latitudes for all the slab files.
+    points_x  = data_in[:, 0] 
+    points_y  = 90 - data_in[:, 1]
+    field     = data_in[:, 2]
+    
+    x0 = np.min(points_x)
+    x1 = np.max(points_x)
+    y0 = np.min(points_y)
+    y1 = np.max(points_y)
+    
+    # create the desired mesh
+    x = np.arange (x0, x1 + 0.05, 0.05)
+    y = np.arange (y0, y1 + 0.05, 0.05)
+    
+    x_grid, y_grid = np.meshgrid (x, y)
+    
+    depth_interp = griddata ((points_x, points_y), field, (x_grid, y_grid), method = 'linear')
+    data_save    = np.column_stack ((x_grid.flatten(), 90 - y_grid.flatten(), depth_interp.flatten()))
+    
+    np.savetxt (file_name[0], data_save, fmt='%1.2f', delimiter=',')
+
+
+# Slabs with 0.02 degree spacing
+file_name_key = ['hin', 'man', 'mue', 'pam', 'puy']
+file_type_key = ['dep', 'thk', 'dip']
+
+for i in range(len(file_name_key)):
+    for j in range(len(file_type_key)):
+        interpolate_data (file_name_key[i], file_type_key[j])
+
+# Create the grid for aspect, with global longitudes and colatitudes and
+# resolution same as all the slabs (0.05 degrees)
+longitudes_target = np.arange (0 , 360.05, 0.05)
+latitudes_target  = np.arange (0,  180.05, 0.05)
+longitude_grid, latitude_grid = np.meshgrid (longitudes_target, latitudes_target)
+
+# We initialize the depth array with a very large number and the
+# thickness value with 0 to make it easy to determine regions
+# without slabs.
+depths_target       = np.finfo(np.float64).max / 1e4 * np.ones (len(longitude_grid.flatten()),)
+thickness_target    = np.zeros (len(longitude_grid.flatten()),)
+dips_target         = np.zeros (len(longitude_grid.flatten()),)
+
+nan_indx = 0 
+def slab2_into_asciiboundary (slab_filename):
+    """
+    This function uses the above target mesh and fills it with the input slab depths, thicknesses
+    and dip angles.
+    Since the input data grid for each slab is a subset of this target grid, we can simply replace 
+    the target grid at those locations with the input slab grid. 
+    We have checked that in the input files longitude increases first and then latitude, suggesting that the 
+    elements are ordered identically to the target mesh.
+    """
+    
+    # load the input slab depth files
+    slab_data = np.loadtxt (slab_filename, delimiter=",")
+    points_x  = slab_data[:, 0] 
+    points_y  = 90 - slab_data[:, 1]
+    depths    = slab_data[:, 2]    
+    
+    x0 = np.min(points_x)
+    x1 = np.max(points_x)
+    y0 = np.min(points_y)
+    y1 = np.max(points_y)
+    
+    # need this to compare two floats
+    numerical_error = 1e-6
+   
+    indx = np.where ( (longitude_grid.flatten() >= x0 - numerical_error) & 
+                      (longitude_grid.flatten() <= x1 + numerical_error) &
+                      (latitude_grid.flatten()  >= y0 - numerical_error) & 
+                      (latitude_grid.flatten()  <= y1 + numerical_error) )
+    
+    # We do not simply replace all the points with the slab grid. This is because there 
+    # are some slabs in the western US with overlapping grid points. However, locations 
+    # in the slab grid where depths and thickness are not defined are nan values, and we
+    # want to avoid replacing existing depth and thickness values with nan values 
+    # from another slab grid.
+    # Therefore, we create a dummy variable that stores all the grid points in the slab 
+    # and then copies only the non-nan values into the target array.
+    depths_temp       = np.empty (len(longitude_grid.flatten()),)  
+    depths_temp[:]    = np.nan
+    depths_temp[indx] = slab_data[:, 2]
+    
+    # replace the target mesh with the slab data
+    indx_not_a_nan                = np.where(~np.isnan (depths_temp))
+    depths_target[indx_not_a_nan] = depths_temp[indx_not_a_nan]
+    
+    # do the same for the thickness file
+    thk_file  = slab_filename.replace ('dep', 'thk')
+    thk_data  = np.loadtxt (thk_file, delimiter=",", usecols=2)
+    
+    # do the same for the dip file
+    dip_file  = slab_filename.replace ('dep', 'dip')
+    dip_data  = np.loadtxt (dip_file, delimiter=",", usecols=2)
+
+    # We checked that slab thickness and dips are defined at points where
+    # slab depths are defined.
+    target_indx_not_a_nan            = np.where(~np.isnan (depths))
+    thickness_target[indx_not_a_nan] = thk_data[target_indx_not_a_nan]
+    dips_target[indx_not_a_nan]      = dip_data[target_indx_not_a_nan]
+
+    return (depths_target, thickness_target, dips_target)
+
+
+input_filenames = glob('*_slab2*' + file_type_key[0]  + '*.xyz')
+
+for i in range(len(input_filenames)):
+    print (input_filenames[i])
+    slab_depths, slab_thickness, slab_dips = slab2_into_asciiboundary (input_filenames[i])
+
+# The thickness in the slab2 model is perpendicular to the slab surface, therefore, we need
+# to convert it into equivalent distance vertically from the slab surface.
+thickness_vertical = slab_thickness/np.cos(np.deg2rad(slab_dips))
+
+# Save the arrays in the format for aspect ascii boundary :
+# 1. use radians in longitudes and co-latitudes, 2. first longitudes increase then co-latitudes
+# 3. convert depth and thickness from km to meters 4. depth is negative in the downward
+# direction of the slab2 database, convert to positive values. 5. Add necessary header
+# information.
+
+km_to_m          = 1e3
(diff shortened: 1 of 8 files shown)


source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,123 @@
...
+      // The input ascii file is structure and as a result has many data points where slabs are absent. We give a very high
+      // number at those locations and therefore the first two conditions check if we are within the slabs.
+      // In the input file, slab depths are to the top of the slab surface.
+      // The hyperbolic tangent function smooths the slabs in the depth direction.
+      if ( (compositional_index == slab_index) && (slab_depths < 1e10) && (slab_thickness < 1e10) &&
+           (depth >= slab_depths) && (depth <= slab_depths + slab_thickness) )
+        {
+          const double slab_center    = slab_depths + slab_thickness/2.;
+          const double half_thickness = slab_thickness/2;
+          slab_composition            = ( 1 - std::tanh( 10 * (std::abs (slab_center - depth) - half_thickness)/half_thickness ) )/2;
```
gassmoeller, line 69: At the moment you have hard-coded the thickness of the transition region and let it scale with the thickness of the slab. Is that  intended? I would have thought it is more useful to either have a constant transition region that is an input parameter (with the disadvantage like we discussed that it is only diffused in depth, not laterally). The alternative would be to simply not have a transition region at all, which may be simpler. E.g. something like this:
```suggestion
          slab_composition = 1.0;
```

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,123 @@
...
+    double
+    SlabModel<dim>::
+    initial_composition (const Point<dim> &position,
+                         const unsigned int compositional_index) const
+    {
+      const unsigned int surface_boundary_id = this->get_geometry_model().translate_symbolic_boundary_name_to_id("top");
+      const unsigned int slab_index          = this->introspection().compositional_index_for_name("slabs");
+
+      const double depth                     = this->get_geometry_model().depth(position);
+      double slab_composition                = 0;
```
gassmoeller, line 54:
```suggestion
      double slab_composition                = 0.0;
```

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// explicit instantiations
+namespace aspect
+{
+  namespace InitialComposition
+  {
+    ASPECT_REGISTER_INITIAL_COMPOSITION_MODEL(SlabModel,
+                                              "slab model",
+                                              "An initial composition model computed to define slabs "
```
gassmoeller, line 109:
```suggestion
                                              "An initial composition model that defines slabs "
```

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,123 @@
...
+                                              "using the slab2 model (Hayes et al., 2018) input in "
+                                              "an ascii format. The file describes the depths to the "
+                                              "top of the slabs and their thickness."
+                                              "The computed compositional value is 1 within the slabs "
+                                              "and zero elsewhere. In order to prevent sharp jumps, the "
+                                              "slabs are smoothened in the depth direction. "
```
gassmoeller, line 115: Either adjust or remove.

```suggestion
                                              "and zero elsewhere. "
```

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,123 @@
...
+                                              "More details on the slab2 model can be found in "
+                                              "Hayes, G. P., Moore, G. L., Portner, D. E., Hearne, M., "
+                                              "Flamme, H., Furtney, M., & Smoczyk, G. M. (2018). Slab2, "
+                                              "a comprehensive subduction zone geometry model. Science, "
+                                              "362(6410), 58-61. The script to convert the slab2 model "
+                                              "into an aspect input data file is available.")
```
gassmoeller, line 121:
```suggestion
                                              "into an aspect input data file is available "
                                              "in the directory data/initial_composition/slab/.")
```

tests/slab2_initial_composition_3d_shell.prm
```diff
@@ -0,0 +1,110 @@
...
+
+subsection Initial composition model
+  set Model name = slab model
+  # The following lines needs to be modified such that
+  # they point to the location of the output slab2 model
+  # datafile, generated using the jupyter notebook in
+  # (contrib/python) folder in the aspect base directory.
+  subsection Slab model
+    set Data directory       = /home/postdoc/Documents/
+    set Data file name       = slab2_depth_thickness.txt
```
gassmoeller, line 72: We will have to find a way to make this test run on our tester machine. Can you include a very low resolution version of the slab files in `data/initial_composition/slab`? If that does not work, you could also make up a file with a new dataset (e.g. a single slab with a very large thickness). The important part is just that it can be included in the repository and can be used in this test.

contrib/python/create_slab2_database.ipynb
```diff
@@ -0,0 +1,267 @@
+{
```
gassmoeller, line 1: Can you set this up so that the script automatically places the output data file in `data/initial_composition/slab`? You could also move the script over there.

gassmoeller (commented):
Thank you, this is great. 
I had a first look at the plugin and left a number of small comments, I did not look at the python notebook yet.
I think the main point to improve is to turn the test into something that can be run automatically, I left comments, but we can talk about it tomorrow as well.

include/aspect/initial_composition/slab_model.h
```diff
@@ -0,0 +1,92 @@
...
+
+  You should have received a copy of the GNU General Public License
+  along with ASPECT; see the file LICENSE.  If not see
+  <http://www.gnu.org/licenses/>.
+*/
+
+
+#ifndef _aspect_initial_composition_slab2_model_h
+#define _aspect_initial_composition_slab2_model_h
+
```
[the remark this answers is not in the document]
alarshi: Will do, thanks.

source/initial_composition/slab_model.cc
gassmoeller: I think it should go into the data folder. We have some other python scripts already and they are mostly in the folders where they are needed. I see `contrib/python` more as a general folder for python scripts we do not know where else to put.

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,123 @@
...
+    template <int dim>
+    void
+    SlabModel<dim>::initialize ()
+    {
+      // The input slabs are defined from the surface of the model
+      std::set<types::boundary_id> surface_boundary_set;
+      surface_boundary_set.insert(this->get_geometry_model().translate_symbolic_boundary_name_to_id("top"));
+
+      // The two columns correspond to slabs depth and thickness
+      slab_boundary.initialize(surface_boundary_set, 2);
```
gassmoeller, line 40: Please make the same change here that I did in #5028 for the other places where this operation happens.

alarshi:
Thank you, Rene, for the reviews! I added a low resolution data file in `initial-composition/slab` folder, that is now used for the test case in the recent commits. I also modified the notebook so that it outputs the data file in `initial-composition/slab` folder.

gassmoeller:
I think this is ready now. @jdannberg or someone else: could you do a review?

data/initial-composition/slab-model/create_slab2_database.py
```diff
@@ -0,0 +1,164 @@
...
+
+for i in range(len(input_filenames)):
+    print (input_filenames[i])
+    slab_depths, slab_thickness = slab2_into_asciiboundary (input_filenames[i])
+
+# Save the arrays in the format for aspect ascii boundary :
+# 1. use radians in longitudes and latitudes, 2. first longitudes increase then latitudes
+# 3. convert depth and thickness from km to meters 4. depth is negative in the downward
+# direction of the slab2 database, convert to positive values. 5. Add necessary header
+# information.
```
jdannberg, line 151: Above it says the output is in colatitude, here it says latitude. Can you fix whichever is not correct?

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,139 @@
...
+namespace aspect
+{
+  namespace InitialComposition
+  {
+    template <int dim>
+    void
+    SlabModel<dim>::initialize ()
+    {
+      AssertThrow(this->introspection().compositional_name_exists("slabs"),
+                  ExcMessage("The initial composition plugin `slabs' did not find a "
```
jdannberg, line 37:
```suggestion
                  ExcMessage("The initial composition plugin `slab model' did not find a "
```

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,139 @@
...
+    template <int dim>
+    double
+    SlabModel<dim>::
+    initial_composition (const Point<dim> &position,
+                         const unsigned int compositional_index) const
+    {
+      if (compositional_index != slab_index)
+        return 0.0;
+
+      // The first data column correspond to the slab depth and the second to slab thickness
```
jdannberg, line 60:
```suggestion
      // The first data column corresponds to the slab depth and the second column to the slab thickness.
```

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,139 @@
...
+      // 'Slab depth' stands for the depth of the upper surface of the slab, 'Slab thickness'
+      // for the vertical distance between upper and lower surface. In lateral positions that
+      // do not have a slab thickness should be set to 0.0 and depth to a very large value
+      // that is outside the model domain.
```
jdannberg, line 64: I don't really understand that comment (also, I first read it as "positions that do not have a slab thickness"). 
Does it mean: In locations without slabs, we set the slab thickness to zero and the depth to a very large value? Or that in locations without slabs, the data file contains a zero slab thickness and a large depth? I guess the second?

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,139 @@
...
+                                              "and zero elsewhere. "
+                                              "An example model that is included is Slab2 described in "
+                                              "Hayes, G. P., Moore, G. L., Portner, D. E., Hearne, M., "
+                                              "Flamme, H., Furtney, M., \\& Smoczyk, G. M. (2018). Slab2, "
+                                              "a comprehensive subduction zone geometry model. Science, "
+                                              "362(6410), 58-61. The script to convert the Slab2 model "
+                                              "into an aspect input data file is available in the directory "
+                                              "data/initial-composition/slab-model/. Please note that "
+                                              "Slab2 and the example data file assume spherical geometry "
+                                              "(latitude, longitude coordinates), however that is not "
```
jdannberg, line 135:
```suggestion
                                              "(latitude, longitude coordinates), however, that is not "
```

tests/slab2_initial_composition_3d_shell.prm
```diff
@@ -0,0 +1,91 @@
...
+  set Number of fields = 1
+  set Names of fields  = slabs
+  set Compositional field methods = static
+end
+
+
+subsection Initial composition model
+  set Model name = slab model
+  # The following lines needs to be modified such that
+  # they point to the location of the output slab2 model
```
jdannberg, line 51:
```suggestion
  # The following lines point to the location of the output slab2 model
```

tests/slab2_initial_composition_3d_shell.prm
```diff
@@ -0,0 +1,91 @@
...
+  # datafile, generated using the jupyter notebook in
+  # (contrib/python) folder in the aspect base directory.
```
jdannberg, line 53:
```suggestion
  # (/data/initial-composition/slab-model/) folder in the aspect base directory.
```

tests/slab2_initial_composition_3d_shell.prm
```diff
@@ -0,0 +1,91 @@
...
+
+
+subsection Mesh refinement
+  set Initial global refinement                = 2
+  set Initial adaptive refinement              = 0
+  set Time steps between mesh refinement       = 0
+  set Strategy                                 = composition threshold, temperature
+  subsection Composition threshold
+    set Compositional field thresholds = 0.2
+  end
```
jdannberg, line 85: If there's no refinement, you don't really need that. 
```suggestion
```

jdannberg (approved):
I have some minor comments about the documentation, otherwise this looks good to me. Please also add a changelog entry.

data/initial-composition/slab-model/create_slab2_database.py
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# This script uses the slab2 model data, available to download from
+# https://www.sciencebase.gov/catalog/item/5aa1b00ee4b0b1c392e86467, and
```
jdannberg, line 5: I went to that website and there are different things that can be downloaded. It would be good to either provide the exact download link, or specify what needs to be downloaded.

gassmoeller:
> I went to that website and there are different things that can be downloaded. It would be good to either provide the exact download link, or specify what needs to be downloaded.

Unfortunately the website does not allow to copy link locations, but I specified the exact name and heading from where to download the data.

I addressed all comments.

alarshi:
Thank you, Juliane, for the reviews! I pushed the changes, i.e., thickness is perpendicular to the slab surface instead of in the vertical direction, in the last commit .

gassmoeller:
@alarshi can you check the files in the data directory? I renamed the directory (to the same name as the plugin) and replaced the python notebook with a simpler python script. Can you integrate your changes into mine?

alarshi:
@gassmoeller : Sorry, I didn't mean to add the `slab` folder, must have happened during the merge. Does it look okay now? Also, do you want me to squash some commits?

gassmoeller:
Yes, can you please squash the commits? I usually do not care about that, but here we have several merge commits that can look quite confusing in a timeline of all commits.

alarshi:
@gassmoeller : I squashed the commits in the latest push.

gassmoeller (approved):
Looks all good to me. We should let @jdannberg have another look, but from my side this is ready.

data/initial-composition/slab-model/create_slab2_database.py
```diff
@@ -0,0 +1,178 @@
...
+# thickness value with 0 to make it easy to determine regions
+# without slabs.
+depths_target       = np.finfo(np.float64).max / 1e4 * np.ones (len(longitude_grid.flatten()),)
+thickness_target    = np.zeros (len(longitude_grid.flatten()),)
+dips_target         = np.zeros (len(longitude_grid.flatten()),)
+
+nan_indx = 0 
+def slab2_into_asciiboundary (slab_filename):
+    """
+    This function uses the above target mesh and fills it with the input slab depths and thicknesses.
```
jdannberg, line 93: You also need to update the documentation here: 
```suggestion
    This function uses the above target mesh and fills it with the input slab depths, thicknesses and dip angles. 
```

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,137 @@
...
+                                              "Hayes, G. P., Moore, G. L., Portner, D. E., Hearne, M., "
+                                              "Flamme, H., Furtney, M., \\& Smoczyk, G. M. (2018). Slab2, "
+                                              "a comprehensive subduction zone geometry model. Science, "
+                                              "362(6410), 58-61. The script to convert the Slab2 model "
+                                              "into an aspect input data file is available in the directory "
+                                              "data/initial-composition/slab-model/. Please note that "
+                                              "Slab2 and the example data file assume spherical geometry "
+                                              "(latitude, longitude coordinates), however, that is not "
+                                              "necessary for this plugin, data files in "
+                                              "cartesian coordinates will work with box geometries.")
```
jdannberg, line 135: But in the initialization, we have an assert and the model will crash in cartesian coordinates. So at the moment, this does not work in box geometries.

source/initial_composition/slab_model.cc
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+
+    template <int dim>
+    double
+    SlabModel<dim>::
+    initial_composition (const Point<dim> &position,
+                         const unsigned int compositional_index) const
+    {
+      if (compositional_index != slab_index)
```
jdannberg, line 55: Sorry if this is somehow obvious, but: does this actually work? I don't see where the slab index is being set. 
In a previous version, there was a line
`const unsigned int slab_index          = this->introspection().compositional_index_for_name("slabs");`
and there was also an assert to make sure the slab index exists.

jdannberg (commented):
I just have two comments about documentation, but feel free to address them in a follow-up PR. 

But I am confused about the slab index. If that's just a misunderstanding on my part, feel free to merge.

source/initial_composition/slab_model.cc
alarshi: I don't think the assert on the geometry model was intended to be there, I will remove it. Thanks!

alarshi:
@jdannberg : Thank you for reviewing it! I addressed your comments in the latest push. I hope I did not break anything from the previous changes :(

gassmoeller (approved):
Thank you, this looks ready to merge now. I added a changelog entry and pushed the commit, because for some reason the testers did not trigger yesterday (looks like a problem on github's side).

Ready to merge if the testers are happy.

PR #5027, opened 2023-01-19
Fix compiler warnings gcc11

gassmoeller:
We have accumulated a few compiler warnings on the main branch (using gcc11), mostly over possibly non-initialized array entries and unused variables. This PR fixed all of them for me and should not change behavior.

### DIFF
--- a/source/particle/property/crystal_preferred_orientation.cc
+++ b/source/particle/property/crystal_preferred_orientation.cc
@@ -670,7 +670,7 @@ namespace aspect
             // these are variables we only need for olivine, but we need them for both
             // within this if block and the next ones
             // Ordered vector where the first entry is the max/weakest and the last entry is the inactive slip system.
-            std::array<unsigned int,4> indices;
+            std::array<unsigned int,4> indices {};
 
             // compute G and beta
             Tensor<1,4> bigI;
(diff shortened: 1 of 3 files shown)


MFraters (commented):
I tried to reproduce the issue with gcc 12, but it gave me no warnings with the default setup on main (both in debug and release). Do you use any extra flags, or is it really specific to gcc 11? 

I think the changes are mostly fine, but I am wondering whether the boost static vector is really needed.

gassmoeller:
Oh I should mention some warnings only appeared in release mode (e.g. the unused `idx` variable), can you check with gcc12?

The static vector is not really necessary (I could have instead initialized the array), I just find it nicer to read this way, because it combines the stack allocation of std::array with runtime determined sizes. Let me know if you prefer a different option.

MFraters (commented):
ah, I did the release test before updating the main and the debug after. Now that I tried release again I see the `idx` warning. I still don't see the others, but I think it is good to initialize variables anyway :)

I can see what you mean, but wouldn't `std::array<double, max_n_vertices_per_face>     distances_to_center{0};` also do the job and be equally readable?

source/particle/property/crystal_preferred_orientation.cc
```diff
@@ -671,6 +671,8 @@ namespace aspect
             // within this if block and the next ones
             // Ordered vector where the first entry is the max/weakest and the last entry is the inactive slip system.
             std::array<unsigned int,4> indices;
+            for (auto &index: indices)
```
tjhei, line 674: You can zero initialize without a loop by using ``indices{};`` in the declaration.

tjhei (approved):

gassmoeller:
All good points. Turns out I didnt know about array initialization :-). Does the current version look ok?

MFraters (approved):
Yes, looks great, thanks for fixing the warnings!

tjhei (approved):

PR #5028, opened 2023-01-19
Remove unnecessary lines of code

gassmoeller:
Something I saw while reviewing #5025. In a number of places we unnecessarily create a temporary `std::set` that is not really needed. This new version does not really change anything in the compiled code (it still needs the std::set internally), but I think the code is shorter and just as easy to read.


MFraters (approved):
yes, I agree this look better. Thanks!

PR #5023, opened 2023-01-17
[WIP] Field types for viscoelasticity and corresponding rheology and EoS modules

bobmyhill:
A rebased version of #4260.
Still under development...

* [ ] I have followed the [instructions for indenting my code](../blob/main/CONTRIBUTING.md#making-aspect-better).
* [ ] I have tested my new feature locally to ensure it is correct.
* [ ] I have [created a testcase](http://www.math.clemson.edu/~heister/manual.pdf#sec%3Awriting_tests) for the new feature/benchmark in the [tests/](../blob/main/tests/) directory.
* [ ] I have added a changelog entry in the [doc/modules/changes](../blob/main/doc/modules/changes) directory that will inform other users of my change.

PR #5029, opened 2023-01-19
Added field type functions

bobmyhill:
This PR pulls out the new field type introspection functions from #5023.
It also changes the Type "chemical composition" to "chemical component", to avoid confusion with the "Compositional field". 

@tjhei, @MFraters, @naliboff: at the moment the new functions aren't tested. If I incorporate these into Steinberger, is it good to go, or would you like more tests (and if so, where?).

### DIFF
--- a/doc/modules/changes/20230119_bobmyhill
+++ b/doc/modules/changes/20230119_bobmyhill
@@ -1,0 +1,13 @@
+New: ASPECT has new introspection functions 
+get_indices_for_fields_of_type and get_names_for_fields_of_type.
+There is also a new utilities function called
+compute_only_composition_fractions that takes a vector of
+compositional field values and a list of indices that designates
+which of those fields correspond to chemical compositions.
+It then calculates the missing fraction corresponding to the
+background field, and returns the full vector of fractions.
+
+New: The "Compositional fields / Types of fields" parameter in ASPECT 
+has a new type called "strain".
+<br>
+(Bob Myhill, 2023/01/19)
--- a/include/aspect/introspection.h
+++ b/include/aspect/introspection.h
@@ -464,14 +464,14 @@ namespace aspect
        * particular type (chemical composition, porosity, etc.).
        */
       const std::vector<unsigned int>
-      get_indices_for_fields_of_type (const std::string &type_name) const;
+      get_indices_for_fields_of_type (const typename Parameters<dim>::CompositionalFieldDescription::Type &type) const;
 
       /**
       * Get the names of the compositional fields which are of a
       * particular type (chemical composition, porosity, etc.).
        */
       const std::vector<std::string>
-      get_names_for_fields_of_type (const std::string &type_name) const;
+      get_names_for_fields_of_type (const typename Parameters<dim>::CompositionalFieldDescription::Type &type) const;
 
       /**
        * A function that gets a component index as an input
--- a/include/aspect/introspection.h
+++ b/include/aspect/introspection.h
@@ -460,6 +460,20 @@ namespace aspect
       compositional_name_exists (const std::string &name) const;
 
       /**
+       * Get the indices of the compositional fields which are of a
+       * particular type (chemical composition, porosity, etc.).
+       */
+      const std::vector<unsigned int>
+      get_indices_for_fields_of_type (const std::string &type_name) const;
+
+      /**
+      * Get the names of the compositional fields which are of a
+      * particular type (chemical composition, porosity, etc.).
+       */
+      const std::vector<std::string>
+      get_names_for_fields_of_type (const std::string &type_name) const;
+
+      /**
        * A function that gets a component index as an input
        * parameter and returns if the component is one of the stokes system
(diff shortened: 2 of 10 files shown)


MFraters (commented):
So this pull request has two changes in it, first the name change, which is tested by the tests which require a name change and adding the function `compute_only_composition_fractions`, which is currently not tested, but also not used.

I think a simple unit test of the function would be great. I don't think that should be too much work to add and it ensures that the function does what it is supposed to do.

jdannberg:
I know I am late to this and may have missed some previous discussion. 
But I have an additional comment: Your PR changes existing functionality (i.e. it renames an option in the input file). So it would be good to make this change backwards compatible for now, i.e. everyone who uses the name "chemical composition" as a field type should still be able to use that at least for a while. 

I also think you still need to update your changelog entry.

bobmyhill:
Thanks @jdannberg . Yes, I still need to change a few things. Thanks for the thoughts on the chemical composition option. In that case I will revert and not change this

jdannberg:
@bobmyhill If you wanted to make the name change, it's not that hard to keep backwards compatibility. You can have a look at how this is done in parameters.cc for the solver schemes. We changed them from i.e. `IMPES` to `single Advection, single Stokes`, but both options are still allowed in the input file, with a description in the original name that it's deprecated. At some point, we will then remove the old option. 

This of course clutters the code with all those options, so generally one would only make a name change if there's a good reason, for example that the old name is really confusing. So if you think it's worth it to make the change here, that's totally fine as well, as long as the old option still works.

MFraters:
hmm, I was under the impression that it wasn't actually used anywhere yet, in which case the name change wouldn't have been a significant problem, but I now see that it is used in the Steinberger material model already, which was part of the last release. So I agree that if you want to make the name change it will have to be backwards compatible.

bobmyhill:
Thanks everyone.
I think this is ready to merge now. I've removed all the breaking changes inherited from #5023, and added a new testcase for the new introspection and utility functions. @MFraters, is this enough for what you wanted to do?

I'll build on this in #5023 when I have time, but it's going to be a messy job.

include/aspect/introspection.h
```diff
@@ -459,6 +459,20 @@ namespace aspect
       bool
       compositional_name_exists (const std::string &name) const;
 
+      /**
+       * Get the indices of the compositional fields which are of a
+       * particular type (chemical composition, porosity, etc.).
+       */
+      const std::vector<unsigned int>
+      get_indices_for_fields_of_type (const std::string &type_name) const;
```
tjhei, line 467: why is the input a string and not a value of the type enum?

include/aspect/material_model/utilities.h
```diff
@@ -261,7 +261,24 @@ namespace aspect
...
+       * If the sum of the compositional_fields is greater than
+       * one, we assume that there is no background field (i.e., that field value
+       * is zero). Otherwise, the difference between the sum of the compositional
+       * fields and 1.0 is assumed to be the amount of the background field.
+       * This function makes no assumptions about the units of the
+       * compositional field values; for example, they could correspond to
+       * mass or volume fractions.
+       */
+      std::vector<double>
+      compute_only_composition_fractions(const std::vector<double> &compositional_fields,
```
tjhei, line 280: so these are all compositional fields and ``indices`` are M indices into that first vector? Maybe call the second parameter ``indices_to_use``?

bobmyhill: Good idea.

include/aspect/introspection.h
bobmyhill: Another good point. Both changed.

MFraters (approved):
Looks good to me. I actually mend a unit test like this: https://github.com/geodynamics/aspect/blob/main/unit_tests/mesh_refinement.cc, but I guess this also works :)

@tjhei  and @jdannberg is this ready to merge?

bobmyhill:
@MFraters thanks! The reason I didn't add a unit_test was because the test requires parameter parsing and introspection. I remember discussing this with Rene before #4258 was merged (probably on zoom in September 2021), but only remember the conclusion :/

doc/modules/changes/20230119_bobmyhill
```diff
@@ -0,0 +1,13 @@
+New: ASPECT has new introspection functions 
+get_indices_for_fields_of_type and get_names_for_fields_of_type.
+There is also a new utilities function called
+compute_only_composition_fractions that takes a vector of
+compositional field values and a list of indices that designates
+which of those fields correspond to chemical compositions.
+It then calculates the missing fraction corresponding to the
+background field, and returns the full vector of fractions.
+
+New: The "Compositional fields / Types of fields" parameter in ASPECT
```
tjhei, line 10: You need two separate files for the formatting of the changelog to look ok.

tjhei (approved):

doc/modules/changes/20230119_bobmyhill
tjhei: see #5037

PR #5031, opened 2023-01-21
automate citations/bibfile check

tjhei:
It turns out that it is easy to add invalid bibtex to
- doc/sphinx/references.bib
- manual/citing_aspect.bib

I am adding a doc/citations-list.tex that generates a pdf with all of the citations above. Because we will soon delete the pdf manual, I am not adding and fixing doc/manual/manual.bib

Add a CI step that fails if invalid characters or any other problem is detected.


bobmyhill (approved):
Looks good to me!

PR #5032, opened 2023-01-21
Increase use of parse_map in material models

bobmyhill:
This PR replaces use of the "possibly_extend_from_1_to_N" function with the more flexible "parse_map_to_double_array" in all material models apart from steinberger.

This is a small step towards unifying code functionality across material models.

### DIFF
--- a/source/material_model/diffusion_dislocation.cc
+++ b/source/material_model/diffusion_dislocation.cc
@@ -382,9 +382,6 @@ namespace aspect
     void
     DiffusionDislocation<dim>::parse_parameters (ParameterHandler &prm)
     {
-      // increment by one for background:
-      const unsigned int n_fields = this->n_compositional_fields() + 1;
-
       prm.enter_subsection("Material model");
       {
         prm.enter_subsection ("Diffusion dislocation");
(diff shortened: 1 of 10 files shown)


MFraters (approved):
The changes look reasonable to me. Is there any reason you explicitly left the Steinberger material model out? Is there an issue there?

bobmyhill:
@MFraters: steinberger (and melt_boukare) use compositional fields in a slightly different way to the other material models. In these models, the number of important `equation of state parameter fields n_p` does not correspond with the number of tracked `chemical component fields n_c`. 

In the case of melt_boukare the modification was easy to make (because n_p = the number of chemical endmembers, whose names are defined in the Material Model). For steinberger the EoS properties are provided in a lookup table, so `n_p = 0`, and the Material model parameter fields are instead reserved for different thermal conductivity domains, defined according to depth. The different depth regions are never given names, so `parse_map_to_double_array` is not a natural function to use. I also don't see anyone using phases in this material model (much better to add thermal conductivities to the lookup).

TL;DR: There's no issue with steinberger, it's just more natural to keep the current function calls for that particular model.

MFraters (approved):
Thanks for the explanation. Looks good to me and I see no other comments, so I am going to merge it. You are really on a roll with pull requests!

PR #5035, opened 2023-01-22
Clarified phase / phase transition arguments and docstrings

bobmyhill:
The utilities function `phase_average_value` takes a vector of [numbers of *phase transitions* per compositional field] as an input argument, but this argument is currently called `n_phases_per_composition`.

The current PR renames the argument to `n_phase_transitions_per_composition`, and also renames all the arguments of functions that make use this function. As far as I can tell, there's only one place in the code where a mistake is currently made based on the name (in `tests/composite_viscous_outputs.cc`). 

This PR also improves the docstrings of several `parse_parameters` functions, which *do* require the number of phases per compositional field rather than the number of phase transitions.

The PR shouldn't change any tests or functionality.

### DIFF
--- a/source/material_model/utilities.cc
+++ b/source/material_model/utilities.cc
@@ -1016,34 +1016,36 @@ namespace aspect
 
 
       double phase_average_value (const std::vector<double> &phase_function_values,
-                                  const std::vector<unsigned int> &n_phases_per_composition,
+                                  const std::vector<unsigned int> &n_phase_transitions_per_composition,
                                   const std::vector<double> &parameter_values,
-                                  const unsigned int composition,
+                                  const unsigned int composition_index,
                                   const PhaseUtilities::PhaseAveragingOperation operation)
       {
         // Calculate base index and assign base value
-        unsigned int base = 0;
-        for (unsigned int i=0; i<composition; ++i)
-          base += n_phases_per_composition[i] + 1;
+        unsigned int start_phase_index = 0;
+        for (unsigned int i=0; i<composition_index; ++i)
+          start_phase_index += n_phase_transitions_per_composition[i] + 1;
 
-        double averaged_parameter = parameter_values[base];
-        if (n_phases_per_composition[composition] > 0)
+        double averaged_parameter = parameter_values[start_phase_index];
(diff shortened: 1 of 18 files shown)


bobmyhill:
I realise that it's only really possible to review this one by pulling the branch. Hopefully once it's done it will make related reviews easier to do

source/material_model/utilities.cc
```diff
@@ -979,34 +979,36 @@ namespace aspect
...
           {
             // Do averaging when there are multiple phases
             if (operation == PhaseUtilities::logarithmic)
               averaged_parameter = log(averaged_parameter);
 
-            for (unsigned int i=0; i<n_phases_per_composition[composition]; ++i)
+            for (unsigned int i=0; i<n_phase_transitions_per_composition[composition_index]; ++i)
               {
-                Assert(base+i+1<parameter_values.size(), ExcInternalError());
+                unsigned int phase_index = start_phase_index + i;
```
tjhei, line 1001: Make it const, please.

tests/composite_viscous_outputs.cc
```diff
@@ -39,7 +39,7 @@ void f(const aspect::SimulatorAccess<dim> &simulator_access,
   const unsigned int composition = 0;
   const std::vector<double> volume_fractions = {1.};
   const std::vector<double> phase_function_values = std::vector<double>();
-  const std::vector<unsigned int> n_phases_per_composition = std::vector<unsigned int>(1);
+  const std::vector<unsigned int> n_phase_transitions_per_composition = std::vector<unsigned int>(0);
```
tjhei, line 42: Why 0 instead of 1?

source/material_model/utilities.cc
bobmyhill: Thanks, done

tests/composite_viscous_outputs.cc
bobmyhill: Ah, thank you. I initially wrote (1, 0) and deleted the wrong number. In this case it shouldn't make a difference, but I've changed it anyway.

tjhei (approved):

PR #5041, opened 2023-01-26
Fix ascii data initial for two merged chunks

gassmoeller:
This PR improves upon #4490 by moving the slice operation into the AsciiDataInitial class. This specifically improves three aspects:
- Slicing a 3d dataset for a 2D model is now also supported for temperature (previously only composition)
- The AsciiDataInitial class previously had to be adjusted for new geometry models, because it couldnt use the `cartesian_to_natural_coordinates` of the geometry model (because if a slice of a dataset was used the number of dimensions of a point was not equal to the number of dimensions of the model). This limitation is now gone, which in particular means `TwoMergedBoxes` and `EllipsoidalChunk` should now work with `AsciiDataInitial`. I will add a test for them to this PR tomorrow.
- We can now remove the awkward `spacedim` template that is not used in the rest of ASPECT and makes `AsciiDataInitial` harder to understand.


bobmyhill:
Looks great. I have no suggestions regarding coding style or comments.

Is specifying points in Cartesian coordinates the most natural choice for your use case? And is specifying two points (plus the origin) better than specifying the unit normal? Just thoughts; not suggestions :) 

I'll have another look after the test(s) is(are) added.

tjhei:
```
--- a/tests/ascii_data_slice_temperature.prm
+++ b/tests/ascii_data_slice_temperature.prm
@@ -29,7 +29,7 @@ end
 
 
 subsection Boundary velocity model
-  set Tangential velocity boundary indicators = bottom, top, 
+  set Tangential velocity boundary indicators = bottom, top,
 end
```

include/aspect/structured_data.h
```diff
@@ -508,15 +508,52 @@ namespace aspect
...
+         * the model or not. If true only a 2D slice of a 3D dataset is used.
+         */
+        bool slice_data;
+
+        /**
+         * The matrix that describes the rotation by which a 2D model
+         * needs to be transformed to a plane that contains the origin and
+         * the two prescribed points given in the input.
+         */
+        Tensor<2,3> rotation_matrix;
```
bangerth, line 556: This means that the slice has to go through the origin, right? I think this might be worth documenting. (But it shouldn't hold up the patch.)

source/structured_data.cc
```diff
@@ -1409,33 +1401,156 @@ namespace aspect
...
+      if (dim == 3)
+        {
+          // Unfortunately, the chunk model uses latitude as natural coordinate. We need to convert this
+          // to colatitude to be consistent with the input files for spherical shells and spheres.
+          if (Plugins::plugin_type_matches<const GeometryModel::Chunk<dim>> (this->get_geometry_model()) ||
+              Plugins::plugin_type_matches<const GeometryModel::TwoMergedChunks<dim>> (this->get_geometry_model()) ||
+              Plugins::plugin_type_matches<const GeometryModel::EllipsoidalChunk<dim>> (this->get_geometry_model()))
+            {
+              internal_position[2] = numbers::PI/2. - internal_position[2];
+            }
```
bangerth, line 1459: That's awkward :-(

bangerth (approved):

Issue #5042, opened 2023-01-27
"Allow fixed temperature on outflow boundaries = false" option does not work in time steps where mesh is refined


jdannberg:
The option that allows it to prescribe the temperature only on those parts of the boundary where material flows in does not seem to work in time steps where the mesh is refined. In those time steps, the temperature is simply prescribed everywhere at the boundary. Below is an example input that reproduces the problem, where flow is prescribed to be vertically upwards and the boundary temperature in the left half of the model is slightly hotter than the initial condition. 

The mesh is refined in every second step, and the boundary condition at the top boundary is also enforced every second step (even though the mesh does not actually change). The resulting problem is visible in the screenshots below:

![step0](https://user-images.githubusercontent.com/7517347/214999462-36c71e08-3b60-4588-b80d-36d99c4bbc00.png)
Step 0 is wrong. 

![step1](https://user-images.githubusercontent.com/7517347/214999476-2df90a33-09c4-41fe-b0c7-499565de79ff.png)
Step 1 is fine. 

![step2](https://user-images.githubusercontent.com/7517347/214999488-3c2e527b-a89e-44af-aaeb-c452fa2a4809.png)
Step 2 is fine. 

![step3](https://user-images.githubusercontent.com/7517347/214999493-6667c23a-7d9c-46c3-9bcf-56ff6f5328bf.png)
Step 3, when the mesh is refined, is wrong again. 

```
set Dimension                              = 2

set Use years in output instead of seconds = true
set End time                               = 2e6
set Output directory                       = test

# Then there are variables that describe how the pressure should
# be normalized. Here, we choose a zero average pressure
# at the surface of the domain (for the current geometry, the
# surface is defined as the top boundary).
set Pressure normalization                 = surface
set Surface pressure                       = 0
set Adiabatic surface temperature          = 1600

subsection Geometry model
  set Model name = box
  subsection Box
    set X extent = 800e3
    set Y extent = 800e3
  end
end

subsection Initial temperature model
  set Model name = function

  subsection Function
    set Variable names      = x,y
    set Function expression = 1500 + 250 * 0.5 * (1-tanh((x-450000)/40000))
  end
end

# This is 10 K hotter than the initial condition within the plume
subsection Boundary temperature model
  set Fixed temperature boundary indicators = bottom, top
  set Allow fixed temperature on outflow boundaries = false
  set Model name = function

  subsection Function
    set Variable names      = x,y
[...]

PR #5043, opened 2023-01-30
initialize current linearization point when we refine the mesh

jdannberg:
Possible fix for #5042. See discussion there. 

If this seems reasonable, I can add the test from #5042.


gassmoeller:
I think this is almost the correct solution. I had to dig a bit to find out why this is causing a problem, but here is what I gathered:
- When `setup_dofs()` is called after mesh refinement all solution vectors and current_initialization_point are reinitialized to the new dof structure and set to 0.0 (has to be done, because the vector entries no longer correspond to the current grid)
- `solution` and `old_solution` are immediately set to the new interpolated values, but `current_linearization_point` is only set to the new values the next time `solve_timestep` is called (which calls `initialize_current_linearization_point`), which is at the start of the next timestep.
- This is a problem, because `replace_outflow_boundary_ids()` which is part of `compute_current_constraints()` evaluates the current linearization point to determine outflow boundaries before `solve_timestep` reinitializes the vector to the correct values, therefore it always sees a velocity of 0 at the boundary after mesh refinement.

I think a better solution than the current one would be to do this:
```
current_linearization_point = distributed_system;
```
which would set `current_linearization_point` to exactly the value of `solution` (instead of an extrapolation from `old_solution` and `old_old_solution`). This is the natural state of `current_linearization_point` after a timestep anyway (i.e. the state it would have in a timestep without mesh refinement). It will then be reset to a new extrapolation for the next timestep at the usual time after the boundary ids have been reset.

Can you give this a try? If it works, feel free to add a test.

jdannberg:
I have changed the code as you suggested (which works), and as discussed, also modified the `replace_outflow_boundary_ids()` function so that zero flow counts as outflow, which fixes the problem with time step 0. 

I have also added a test and changelog entry.

jdannberg:
I have updated the test results. 
I have looked through the list of tests, and all of the ones that change have a fixed boundary composition at a model boundary where the velocity is prescribed. In some of these, the prescribed velocity is zero (so that the boundary composition was prescribed before over the whole model run and now isn't any more), in others, the prescribed composition is different than the initial composition (so that the composition is different in time step zero because we now no longer apply the composition boundary conditions before the velocity solution is calculated). 
So I think it makes sense that all these tests change.

source/simulator/core.cc
```diff
@@ -1746,6 +1746,8 @@ namespace aspect
       constraints.distribute (old_distributed_system);
       old_solution = old_distributed_system;
 
+      current_linearization_point = distributed_system;
```
tjhei, line 1749: Can you leave a comment here that summarizes the discussion (that we need some value for the inflow/outflow determination and that it will be updated later)

source/simulator/helper_functions.cc
```diff
@@ -2115,7 +2115,8 @@ namespace aspect
                   }
 
                 // ... and change the boundary id of any outflow boundary faces.
-                if (integrated_flow > 0)
+                // If there is no flow, we do not want to apply dirichlet boundary conditions either.
+                if (integrated_flow >= 0)
```
tjhei, line 2119: Yes, note that this is now the same as in DG:
https://github.com/geodynamics/aspect/blob/cbc7833440cd05c7370bc28f4119199f052dbcfa/source/simulator/assemblers/advection.cc#L780

tjhei (approved):
Very nice work!

source/simulator/helper_functions.cc
jdannberg: Oh, it's great that it's now consistent between the two methods.

jdannberg:
Thank you! I addressed your one comment.

jdannberg:
...and I have no idea why a test is failing, since all I changed was a comment.

tjhei (approved):

PR #5045, opened 2023-01-31
switch website to html doc

tjhei:


gassmoeller:
I guess we are ok to move forward? This would also increase the pressure to fix any remaining small issues like #5046.

ljhwang:
There are quite a few other places where this is linked that would need to be updated if so or wait until the next release?

PR #5050, opened 2023-02-04
Standardize 2D->2d and 3D->3d.

bangerth:
Seen while looking at #5041. We have a funny mix between 2d/3d and 2D/3D:
```
projects/aspect> egrep -r '\".*[23]d' include source/ | wc -l
123
projects/aspect> egrep -r '\".*[23]D' include source/ | wc -l 
82
```
So the lower-case `d` wins statistically speaking, and this patch fixes this uniformly in strings, almost all of which appear to be in the descriptions of parameters and models. Let's see whether this compiles and tests alright.

/rebuild


source/structured_data.cc
```diff
@@ -1478,24 +1478,24 @@ namespace aspect
 
       prm.enter_subsection (subsection_name);
       {
-        prm.declare_entry("Slice dataset in 2D plane", "false",
+        prm.declare_entry("Slice dataset in 2d plane", "false",
```
tjhei, line 1481: This will break things.

source/structured_data.cc
```diff
@@ -1516,7 +1516,7 @@ namespace aspect
 
       prm.enter_subsection(subsection_name);
       {
-        slice_data = prm.get_bool ("Slice dataset in 2D plane");
+        slice_data = prm.get_bool ("Slice dataset in 2d plane");
```
tjhei, line 1519: Same here

tjhei (changes requested):

bangerth:
Oh, doh -- it hadn't occurred to me that there could be places where that appears in the names of parameters. I reverted the two places, thanks for pointing it out!

tjhei:
> , thanks for pointing it out

The testers did. :-)

tjhei (approved):

PR #5052, opened 2023-02-05
Update installing-docker.md

tjhei:
- update file sizes
- add windows notes

### DIFF
--- a/doc/sphinx/user/install/docker-container/installing-docker.md
+++ b/doc/sphinx/user/install/docker-container/installing-docker.md
@@ -17,5 +17,20 @@ docker pull geodynamics/aspect
 ```
 
 Note that the transfer size of the compressed image containing
-ASPECT and all its dependencies is about 900&nbsp;MB.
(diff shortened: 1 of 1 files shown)


gassmoeller (approved):
Looks good, but why do we recommend running through the command line?

tjhei:
> why do we recommend running through the command line?

I was not able to figure out how to run through the GUI, because there is no way to pass ``-it`` there, which means the container exits immediately. It used to work for an older Docker version (see https://community.geodynamics.org/t/a-problem-occurred-when-installing-aspect-in-docker-windows11/2793/9)

gassmoeller:
The Jenkins tester seems unhappy, but that cannot be related to this PR. Do you have an idea for what is going on? 
Feel free to merge this PR.

tjhei:
I have seen the same test failure before, but I am not sure why.

gassmoeller:
Let's merge this anyway, it is definitely not related to that test.

PR #5055, opened 2023-02-06
Refactor parse_map_to_double_array and add option to ignore keys

gassmoeller:
@bobmyhill what do you think about this modification of parse_map_to_double_array to make #5039 work?
These are the changes I made:
- created a new function inside its own namespace that only takes the input string and a new `Options` struct that contains all the available options. This makes it simpler to extend this in the future if necessary.
- modified the existing `parse_map_to_double_array` to call the new function with all settings set so that the behavior should not change (at least none of the tests does)
- refactored `parse_map_to_double_array` into separate helper functions that do smaller pieces of the work
- added a parameter `list_of_allowed_keys` that can be longer than the existing `list_of_required_keys`.  Every key inside list_of_allowed_keys that is not inside list_of_required_keys will be ignored and not parsed into the map (technically it will be parsed, but then deleted). This should allow you to do what you want in #5039, just hand over the names of all compositional fields as list_of_allowed_keys, and the list of chemical compositions as list_of_required_keys. We can start using the new function for all the functions you change in that PR and replace the remaining calls to the old function at a later time.

Let me know what you think.

### DIFF
--- a/include/aspect/utilities.h
+++ b/include/aspect/utilities.h
@@ -74,6 +74,145 @@ namespace aspect
                                  const unsigned int N,
                                  const std::string &id_text);
 
+    namespace MapParsing
+    {
+      /**
+       * A struct that bundles all the available options for
+       * parse_map_to_double_array().
+       */
+      struct Options
+      {
+        /* A list of N valid key names that are allowed
+         * to appear in the map. If this list is empty
+         * it is assumed to be equal to the list of
+         * required keys. If this list is longer than
+         * list_of_required_keys, every key that is
+         * allowed but not required will be ignored when
+         * parsing the map.
+         */
+        std::vector<std::string> list_of_allowed_keys;
+
+        /* A list of N valid key names that are required
+         * to appear in the map. Only these keys will be
+         * parsed into the map structure and the order of
+         * these keys determines the order of entries
+         * in the output vector.
+         */
+        std::vector<std::string> list_of_required_keys;
+
+        /*
+         * A name that identifies the type of property
+         * that is being parsed by this function and that is used in generating
+         * error messages if the map does not conform to the expected format.
+         */
+        std::string property_name;
+
+        /*
+         * If true allow having multiple values
+         * for each key. If false only allow a single value per key. In either
+         * case each key is only allowed to appear once. Multiple values
+         * for a key are delimited by a "|" character, as in
+         * "key1: value1|value2|value3, key2: value1|value2".
+         */
+        bool allow_multiple_values_per_key;
+
+        /*
+         * Whether to allow that some keys in list_of_required_keys are
+         * not set to any values, i.e. they do not appear at all.
+         * This also allows a completely empty map.
+         */
+        bool allow_missing_keys;
+
+        /*
+         * Whether to store the number of values
+         * per key in n_values_per_key while creating
+         * the map. This allows to query the input
+         * structure and check that subsequent calls to
+         * parse other input parameters have the same
+         * structure.
+         */
+        bool store_values_per_key;
+
+        /*
+         * Whether to check the number of values
+         * per key in the map against values stored
+         * in n_values_per_key. This allows to
+         * check that subsequent calls to
+         * parse other input parameters have the same
+         * structure.
+         */
+        bool check_values_per_key;
+
+        /*
+         * A vector of unsigned
+         * integers that is used by store_values_per_key and
+         * check_values_per_key to either store the current map
+         * structure or check the map structure against an existing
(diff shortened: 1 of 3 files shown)


include/aspect/utilities.h
```diff
@@ -74,6 +74,145 @@ namespace aspect
...
 
+    namespace MapParsing
+    {
+      /**
+       * A struct that bundles all the available options for
+       * parse_map_to_double_array().
+       */
+      struct Options
+      {
+        /* A list of N valid key names that are allowed
```
bobmyhill, line 85:
```suggestion
        /* A list of key names that are allowed
```

include/aspect/utilities.h
```diff
@@ -74,6 +74,145 @@ namespace aspect
...
+         * to appear in the map. If this list is empty
+         * it is assumed to be equal to the list of
+         * required keys. If this list is longer than
+         * list_of_required_keys, every key that is
+         * allowed but not required will be ignored when
+         * parsing the map.
+         */
+        std::vector<std::string> list_of_allowed_keys;
+
+        /* A list of N valid key names that are required
```
bobmyhill, line 95:
```suggestion
        /* A list of key names that are required
```

include/aspect/utilities.h
```diff
@@ -74,6 +74,145 @@ namespace aspect
...
+         * to appear in the map. Only these keys will be
+         * parsed into the map structure and the order of
+         * these keys determines the order of entries
+         * in the output vector.
+         */
+        std::vector<std::string> list_of_required_keys;
+
+        /*
+         * A name that identifies the type of property
```
bobmyhill, line 104: Could the description be more specific about what "type of property" means? Now that we have Field Types, there's scope for confusion.

include/aspect/utilities.h
```diff
@@ -74,6 +74,145 @@ namespace aspect
...
+         * that is being parsed by this function and that is used in generating
+         * error messages if the map does not conform to the expected format.
+         */
+        std::string property_name;
+
+        /*
+         * If true allow having multiple values
```
bobmyhill, line 111:
```suggestion
         * If true, allow multiple values
```

include/aspect/utilities.h
```diff
@@ -74,6 +74,145 @@ namespace aspect
...
+         * for each key. If false only allow a single value per key. In either
+         * case each key is only allowed to appear once. Multiple values
+         * for a key are delimited by a "|" character, as in
+         * "key1: value1|value2|value3, key2: value1|value2".
+         */
+        bool allow_multiple_values_per_key;
+
+        /*
+         * Whether to allow that some keys in list_of_required_keys are
```
bobmyhill, line 120:
```suggestion
         * Whether to allow for some keys in list_of_required_keys to be
```

include/aspect/utilities.h
```diff
@@ -74,6 +74,145 @@ namespace aspect
...
+         * not set to any values, i.e. they do not appear at all.
+         * This also allows a completely empty map.
+         */
+        bool allow_missing_keys;
+
+        /*
+         * Whether to store the number of values
+         * per key in n_values_per_key while creating
+         * the map. This allows to query the input
```
bobmyhill, line 129:
```suggestion
         * the map. This vector can be subsequently accessed
```

include/aspect/utilities.h
```diff
@@ -74,6 +74,145 @@ namespace aspect
...
+         * structure and check that subsequent calls to
```
bobmyhill, line 130:
```suggestion
         * and used (for example) to check that subsequent calls to
```

source/utilities.cc
```diff
@@ -247,117 +247,291 @@ namespace aspect
...
 
-        if (Patterns::Map(Patterns::Anything(),
-                          Patterns::List(Patterns::Double(),
-                                         0,
-                                         std::numeric_limits<unsigned int>::max(),
-                                         "|")).match(input_string))
-          {
-            // Split the list by comma delimited components.
-            const std::vector<std::string> field_entries = dealii::Utilities::split_string_list(input_string, ',');
+          if (Patterns::Map(Patterns::Anything(),
```
bobmyhill, line 265: I think it might be worth adding a comment block up here describing the logic and differences between the different cases (Anything, list of allowed, list of required)

bobmyhill (changes requested):
This is great! Just a few suggestions for docstrings. 
Could you provide a unit test for the new functionality?

gassmoeller:
Thanks for the suggestions, I think I addressed them all.

bobmyhill (approved):
Brilliant, thank you!

PR #5057, opened 2023-02-07
Update description 'Allow fixed composition on outflow boundaries'

anne-glerum:
As discussed with @gassmoeller and @jdannberg.
Setting the parameter `Allow fixed composition on outflow boundaries` to `true` does not have any effect when discontinuous Galerkin elements are used for fields. Composition cannot be fixed on outflow boundaries for DG. This PR simply adds this to the description of the parameter. 

### For all pull requests:


tjhei:
I am confused. We do support boundary conditions for DG compositional fields, it just happens to be that this flag is being ignored:
https://github.com/geodynamics/aspect/blob/cbc7833440cd05c7370bc28f4119199f052dbcfa/source/simulator/assemblers/advection.cc#L693-L698

anne-glerum:
Yes, this flag is ignored when using DG fields. We support boundary conditions for DG, but not on those parts of fixed boundaries that experience **outflow**, even if this flag is set to true. I think that should be made clear to the user, hence I've included it in the parameter description.

tjhei:
Ok, I might have misunderstood you because you used the wording "not possible", whereas it is just not implemented right now.

tjhei:
Maybe we should make it an AssertThrow instead?

anne-glerum:
Sorry for the confusion. Discussing it with @jdannberg and @egpuckett, we also suspected that it is not possible to prescribe boundary conditions on outflow boundaries for DG, because there is no natural diffusion and an upwind scheme is used. But we meant to discuss this with you and Wolfgang to be sure. So you think it _is_ possible to prescribe values on outflow parts?

tjhei:
> But we meant to discuss this with you and Wolfgang to be sure. So you think it _is_ possible to prescribe values on outflow parts?

I am not 100% sure what would happen if you just ignore the computation of 
https://github.com/geodynamics/aspect/blob/cbc7833440cd05c7370bc28f4119199f052dbcfa/source/simulator/assemblers/advection.cc#L780
and set it to ``true``. One might need a different formulation for the boundary penalty term (or it might not work at all).

bangerth:
Why *would* you want to impose boundary conditions on outflow boundaries?

It is possible to do that in DG methods if you define what the flux should be. A strict upwind flux corresponds to not evaluating the boundary conditions at these parts of the boundary, but if you had a centered flux, or something that depends on the velocity, then you are asking for the boundary values. But I don't think you would *want* to do that.

So I think that the first sentence of the text @anne-glerum suggests makes sense: The parameter is simply not taken into account. I would drop the second sentence about it not being possible.

anne-glerum:
Okay, thanks for your input. I've dropped the second sentence.

tjhei (approved):

PR #5058, opened 2023-02-07
doc: nonlinear solver improvements

tjhei:
- fix citations
- fix formula formatting
- typos


MFraters (approved):
Thanks for the fixes, it look a lot better! I found one more * which should be a \cdot to stay consistent in the paragraph starting with `Note that the closer the iteration gets to the solution, the slower the iteration converges. As shown in` Could you also fix that?

tjhei:
Sure, done.

MFraters:
Thanks!

PR #5059, opened 2023-02-08
Add option for time constant ascii data boundary

gassmoeller:
The last step to make #5025 work. Add an option to AsciiDataBoundary to not support time dependent boundary conditions. This removes the irrelevant input parameters when a time-dependence of the dataset is not supported or allowed. Also make the class a bit safer to use without time dependence by initializing all time-dependent members to invalid values and reorder a few lines to only be executed when time-dependence is actually switched on.

This PR also fixes a long-standing property of AsciiDataBoundary, which is that even if we specify exactly one file without placeholder for timestep so far the class has issued a warning that it couldnt find the next file and therefore continues with a constant boundary condition. This has annoyed me for a long time, because if one specific file is given in the input file, it is obviously implied by the user that the boundary condition is not changing over time, and this warning was more confusing than helpful. In this new version we only issue the warning if we had found several files, and then reach the end of the file series.

### DIFF
--- a/source/structured_data.cc
+++ b/source/structured_data.cc
@@ -703,12 +703,12 @@ namespace aspect
     template <int dim>
     AsciiDataBoundary<dim>::AsciiDataBoundary ()
       :
-      current_file_number(0),
-      first_data_file_number(0),
+      current_file_number(numbers::invalid_unsigned_int),
+      first_data_file_number(numbers::invalid_unsigned_int),
       decreasing_file_order(false),
-      data_file_time_step(0.0),
-      time_weight(0.0),
-      time_dependent(true),
+      data_file_time_step(numbers::signaling_nan<double>()),
+      time_weight(numbers::signaling_nan<double>()),
+      time_dependent(false),
       lookups(),
       old_lookups()
     {}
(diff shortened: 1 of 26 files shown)


source/structured_data.cc
```diff
@@ -1016,28 +1011,32 @@ namespace aspect
...
 
     template <int dim>
     void
-    AsciiDataBoundary<dim>::end_time_dependence ()
+    AsciiDataBoundary<dim>::end_time_dependence (const bool issue_warning)
     {
       // no longer consider the problem time dependent from here on out
       // this cancels all attempts to read files at the next time steps
       time_dependent = false;
-      // Give warning if first processor
```
tjhei, line 1033: would it make sense to set other parameters to invalid here?

source/structured_data.cc
```diff
@@ -1178,29 +1191,47 @@ namespace aspect
...
+            AssertThrow (create_filename (0, 0) == create_filename (1, 0),
+                         ExcMessage("A boundary data file name was used that contained a placeholder for "
+                                    "the current file number, but this AsciiDataBoundary object does not support "
+                                    "time dependent information. Please remove the timestep placeholder."));
           }
+
+        // if filename does not contain a placeholder for timestep, no time dependence
+        // do not issue a warning, the parameter file is specifying exactly one file.
+        if (create_filename (0, 0) == create_filename (1, 0))
+          end_time_dependence (false);
```
tjhei, line 1232: if the function only changes the bool, I would prefer to get rid of the bool argument and just set ``time_dependent=false;`` here.

tjhei (approved):

tjhei:
that gets rid of a lot of these annoying warnings from the tests. 👍

tjhei (approved):
Very nice, thank you for the update! Looks much cleaner this way, in my opinion.

gassmoeller:
Yes, I actually like this form more as well. To your comment:
```
would it make sense to set other parameters to invalid here?
```
I would prefer not to. They are not used anymore, but I could see some applications for why we want to keep their values. E.g. one could ask the model: how many data files did you find? Or: please when you reach the last data file reverse the direction and go backward again to model an inversion of a rift basin. None of that is implemented right now, but I think it would be useful to have the possibility.

tjhei:
> I would prefer not to.

I agree. :+1:

PR #5060, opened 2023-02-08
GMG: use deal.II compute_diagonal() part I

tjhei:
This is part 1 of #4793

While some test results change very slightly, they seem harmless to me.


gassmoeller (approved):
I agree the test changes look like floating point rounding errors.

gassmoeller:
Looks like the Jenkins tester is offline (504 Gateway Time-out). Do you want to check that first? Otherwise feel free to merge.

tjhei:
It is also happy, so...

PR #5061, opened 2023-02-09
GMG use compute diagonal part II

tjhei:
Remove the handwritten work-around for computing the GMG diagonals. We used to get completely wrong results for tangential boundary conditions because the constraints were not handled correctly. This workaround is no longer necessary (only minor test output differences)

I also ran some longer computations with adaptivity for various 2d/3d problems and they only differ by tiny amounts.

second part of #4793


gassmoeller (approved):
Very nice.

PR #5064, opened 2023-02-10
Remove unnecessary asserts in ellipsoidal chunk

gassmoeller:
Small follow-up to #5050. A few more instances of 2D and 3D.
Also I moved some asserts. It is not necessary to check that we are in 3d every time the geometry model / manifold is called for a point. It is sufficient to check once when they are initialized.

### DIFF
--- a/source/geometry_model/ellipsoidal_chunk.cc
+++ b/source/geometry_model/ellipsoidal_chunk.cc
@@ -94,6 +94,8 @@ namespace aspect
                                                                              const double para_bottom_depth,
                                                                              const std::vector<Point<2>> &para_corners)
     {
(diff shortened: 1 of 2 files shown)


source/geometry_model/ellipsoidal_chunk.cc
```diff
@@ -653,7 +648,7 @@ namespace aspect
       AssertThrow(false, ExcMessage("Function height_above_reference_surface is not yet implemented "
                                     "for the ellipsoidal chunk geometry model. "
                                     "Consider using a box, spherical shell, or chunk.") );
-      return -999;
+      return numbers::signaling_nan<double>();
```
bangerth, line 651: :-)

bangerth (approved):

Issue #4955, opened 2022-09-12
nsinker_spherical_shell convergence issues


tjhei:
But nsinker_spherical_shell cannot converge using gmg,
and amg doesn't converge either. 

```
Additional information: 
    The iterative Stokes solver in
    StokesMatrixFreeHandlerImplementation::solve did not converge.
    
    The initial residual was: 1.000758e-02
    The final residual is: 1.745499e-08
    The required residual for convergence is: 1.000758e-08
    See output-gmg/solver_history.txt for the full convergence history.

```    

```
    void dealii::SolverFGMRES<VectorType>::solve(const MatrixType&,
    VectorType&, const VectorType&, const PreconditionerType&) [with
    MatrixType = aspect::internal::StokesBlock; PreconditionerType =
    aspect::internal::BlockSchurPreconditioner<dealii::TrilinosWrappers::PreconditionAMG,
    dealii::TrilinosWrappers::PreconditionBase>; VectorType =
    dealii::TrilinosWrappers::MPI::BlockVector]
    The violated condition was:
    false
    Additional information:
    Iterative method reported convergence failure in step 2000. The
    residual in the last step was 1.90377e-08.
```

_Originally posted by @zjiaqi2018 in https://github.com/geodynamics/aspect/pull/4852#issuecomment-1243046122_

tjhei:
Let's check if this is still an issue.

gassmoeller:
To be clear, the starter project part here is just to run the benchmark and check if there are still issues. Fixing the issues will likely be complicated.

quangx:
amg  (4 levels): 

 

> The iterative Stokes solver in Simulator::solve_stokes did not
> 
>     converge.
>     
>     The initial residual was: 1.000757e-02
>     The final residual is: 1.921676e-08
>     The required residual for convergence is: 1.000757e-08
>     See output-amg/solver_history.txt for the full convergence history.

gmg 

> *** Timestep 0:  t=0 seconds, dt=0 seconds
>    Solving Stokes system (GMG)... 
>     GMG coarse size A: 2910, coarse size S: 150
>     GMG n_levels: 3
>     Viscosity range: 0.1 - 1000
>     GMG workload imbalance: 1
>     Stokes solver: 116+0 iterations.
>     Schur complement preconditioner: 117+0 iterations.
>     A block preconditioner: 117+0 iterations.
> 
>    Postprocessing:
>      System matrix memory consumption:  0.35 MB
[...]

PR #5066, opened 2023-02-10
GMG: shared MatrixFree objects

tjhei:
Share the MatrixFree objects between StokesOperator, ABlockOperator, and SchurcomplementOperator for GMG.

This forces batching and iteration of cells in the same order. We assumed beforehand that they would happen in the same order, but this is not always the case! This fixes a problem that in certain situations wrong CellData (viscosity etc.) was being read leading to slow solver convergence (or no convergence).

This also reduces setup cost and memory consumption and requires less code. Win, win, win!?

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -2428,8 +2428,7 @@ namespace aspect
       typename MatrixFree<dim,double>::AdditionalData additional_data;
       additional_data.tasks_parallel_scheme =
         MatrixFree<dim,double>::AdditionalData::none;
-      additional_data.mapping_update_flags = (update_values | update_gradients |
-                                              update_JxW_values | update_quadrature_points);
+      additional_data.mapping_update_flags = (update_gradients | update_JxW_values);
 
       if (sim.mesh_deformation
           && !sim.mesh_deformation->get_free_surface_boundary_indicators().empty())
@@ -2524,7 +2523,6 @@ namespace aspect
             IndexSet relevant_dofs;
             DoFTools::extract_locally_relevant_level_dofs(dof_handler_p, level, relevant_dofs);
             level_constraints_p.reinit(relevant_dofs);
-            // TODO: do we need level_constraints_p.add_lines(mg_constrained_dofs_Schur_complement.get_boundary_indices(level));
             level_constraints_p.close();
           }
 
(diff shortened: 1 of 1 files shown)


tjhei:
possibly related to #4955 , #4563, #4248 and more.

MFraters:
This sound great! I will test out the changes :)

source/simulator/stokes_matrix_free.cc
```diff
@@ -2540,51 +2538,22 @@ namespace aspect
             additional_data.mapping_update_flags = (update_gradients | update_JxW_values |
```
gassmoeller, line 2538: why this change from value to gradients? is it because we now need gradients and values? should we then specify both?

source/simulator/stokes_matrix_free.cc
```diff
@@ -2444,94 +2446,90 @@ namespace aspect
...
+                // let Dirichlet values win over no normal flux:
+                level_constraints_v.merge(user_level_constraints, AffineConstraints<double>::left_object_wins);
+                level_constraints_v.close();
+              }
+          }
+          {
+            IndexSet relevant_dofs;
+            DoFTools::extract_locally_relevant_level_dofs(dof_handler_p, level, relevant_dofs);
+            level_constraints_p.reinit(relevant_dofs);
+            // TODO: do we need level_constraints_p.add_lines(mg_constrained_dofs_Schur_complement.get_boundary_indices(level));
```
gassmoeller, line 2527: is this TODO a question? is it necessary for finishing this PR or something for the future?

gassmoeller (approved):
Looks good. I have two questions, likely just things I didnt understand fully. Feel free to merge if my questions are irrelevant.

source/simulator/stokes_matrix_free.cc
tjhei: Good point. It was a question for myself and the answer is: no, not needed.

source/simulator/stokes_matrix_free.cc
tjhei: Good catch. Let me clean this up.

tjhei: it turns out that the flags were incorrect anyways (it turns out _values doens't matter).

tjhei:
updated.

PR #5068, opened 2023-02-13
GMG: code cleanup

tjhei:
- use make_shared
- rename variables
- formatting

### DIFF
--- a/source/simulator/stokes_matrix_free.cc
+++ b/source/simulator/stokes_matrix_free.cc
@@ -256,7 +256,7 @@ namespace aspect
       // iterations of our two-stage outer GMRES iteration)
       if (do_solve_Schur_complement)
         {
-          // first solve with the bottom left block, which we have built
+          // first solve with the bottom right block, which we have built
           // as a mass matrix with the inverse of the viscosity
           SolverControl solver_control(100, src.block(1).l2_norm() * Schur_complement_tolerance,true);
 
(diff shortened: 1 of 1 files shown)


source/simulator/stokes_matrix_free.cc
```diff
@@ -256,7 +256,7 @@ namespace aspect
       // iterations of our two-stage outer GMRES iteration)
       if (do_solve_Schur_complement)
         {
-          // first solve with the bottom left block, which we have built
+          // first solve with the bottom right block, which we have built
```
MFraters, line 259: hmm, that was probably just copied from here: https://github.com/geodynamics/aspect/blob/edcb0fa34bfdf4a5a49d58a90112837dfb95501e/source/simulator/solver.cc#L274, which also still has that. I guess that should be a different pull request.

MFraters (approved):
nice cleanup! Does it make sense to just write out mf as `matrix free` or `matrix_free_object`? It would make the code easier to read for me. Just a suggestion, fine either way.

source/simulator/stokes_matrix_free.cc
tjhei: true, see #5069

tjhei:
> Does it make sense to just write out mf as `matrix free`

sure, done.

source/simulator/stokes_matrix_free.cc
```diff
@@ -2396,14 +2396,13 @@ namespace aspect
     }
 
     // Setup the matrix-free operators
-    std::shared_ptr<MatrixFree<dim,double>> stokes_mf_storage(new MatrixFree<dim,double>());
-    matrix_free_objects.push_back(stokes_mf_storage);
+    std::shared_ptr<MatrixFree<dim,double>> matrx_free = std::make_shared<MatrixFree<dim,double>>();
```
MFraters, line 2399: Can you fix `matrx_free` to `matrix_free`? Otherwise good to merge.

tjhei: That would help, right? Haha.

PR #5070, opened 2023-02-13
skip refinement if mesh did not change

tjhei:
ASPECT now detects when refinement does not actually change the mesh and skips the redundant computation.


tjhei:
@mfmweerdesteijn FYI. This likely speeds up your "long" simulations with a fixed mesh by a significant amount.

tests/skip_refinement.prm
```diff
@@ -0,0 +1,162 @@
+# A simulation with adaptive refinement only based on minimum refinement
+# function. As a consequence, refinement will not change the mesh
+# after a few steps. Confirm that ASPECT detects this and skips the
```
jdannberg, line 3: When you talk about steps, I first thought of time steps (which of course does not make sense). 

```suggestion
# after this minimum refinement level is reached. Confirm that ASPECT detects this and skips the
```

tests/skip_refinement.prm
```diff
@@ -0,0 +1,162 @@
...
+# Mesh refinement specifications
+subsection Mesh refinement
+  set Initial adaptive refinement        = 3
+  set Initial global refinement          = 2
+  set Time steps between mesh refinement = 1
+  set Strategy                           = minimum refinement function
+  subsection Minimum refinement function
+    set Coordinate system   = cartesian
+  set Variable names      = x,y
+  set Function expression = y>=2791.e3 ? 4 : 1
```
jdannberg, line 47: Can you align this properly?

source/simulator/core.cc
```diff
@@ -1687,12 +1687,28 @@ namespace aspect
...
+              {
+                any_flags_set = true;
+                break;
+              }
+          }
+      }
+      const bool mesh_changed = Utilities::MPI::max(any_flags_set?1:0,mpi_communicator) == 1 ? true : false;
+      if (!mesh_changed)
+        {
+          pcout << "\tSkipping mesh refinement, because the mesh did not change." << std::endl;
```
jdannberg, line 1708: This needs less indentation (the regular output when cells are refined has no indentation at all), and another new line at the end (to be consistent with what we do in case of mesh refinement).

jdannberg (approved):
I like this! I only have a few comments on formatting.

source/simulator/core.cc
tjhei: I guess you are voting for 0 indentation? Ok.

jdannberg:
It looks like there are some tests where the actual solution changes (`edit_parameters`, `remove_angular_momentum`, `remove_net_rotation`). I wouldn't have expected this, but it seems like they all use material models where the viscosity does not depend on the solution (`simple` with specific options/`simpler`) and so they do not rebuild the Stokes preconditioner in these steps where the mesh does not change. But some of these changes aren't small. 

I think the `edit_parameters` test shows another problem: the change in parameters (reducing the CFL number) happens much later now than it was supposed to.

tjhei:
> t looks like there are some tests where the actual solution changes. I wouldn't have expected this,

Same here, but your explanation makes sense: We do not reassemble the matrix and preconditioner because the mesh did not change. If a problem is set up incorrectly (we claim the matrix won't change but it changes in practice), this would lead to different results.

I haven't checked each individual test case yet...

tests/remove_net_rotation/screen-output
```diff
@@ -26,344 +26,287 @@ Number of degrees of freedom: 2,832 (1,728+240+864)
...
 
 *** Timestep 1:  t=636369 years, dt=636369 years
    Solving temperature system... 15 iterations.
-   Rebuilding Stokes preconditioner...
    Solving Stokes system... 18+0 iterations.
 
    Postprocessing:
      RMS, max velocity:                                                               0.777 m/year, 1.16 m/year
-     Angular momentum, Moment of inertia, Angular velocity, Surface angular velocity: 2.52e+05 kg*m^2/year, 2.3e+27 kg*m^2, 1.1e-22 1/year, 4.71e-19 1/year
+     Angular momentum, Moment of inertia, Angular velocity, Surface angular velocity: 2.48e+05 kg*m^2/year, 2.3e+27 kg*m^2, 1.08e-22 1/year, 4.71e-19 1/year
```
tjhei, line 37: Same here.

tests/remove_net_surface_rotation/statistics
```diff
@@ -15,31 +15,31 @@
 # 15: Angular velocity (1/year)
 # 16: Surface angular velocity (1/year)
  0 0.000000000000e+00 0.000000000000e+00 192 1968 864  0 21 23 88 2.90543191e-01 4.30422599e-01 -1.90539675e+11 2.30087456e+27 -8.28118485e-17 7.50664936e-23 
- 1 6.363688260799e+05 6.363688260799e+05 192 1968 864 15 17 19 76 7.76991641e-01 1.16177514e+00 -1.03701572e+09 2.30087456e+27 -4.50705024e-19 7.77243503e-23
```
tjhei, line 18: same here

tests/shear_heating_with_melt/screen-output
```diff
@@ -97,17 +96,17 @@ Number of degrees of freedom: 1,663 (486+202+486+82+243+82+82)
    Rebuilding Stokes preconditioner...
    Solving Stokes system... 0+0 iterations.
    Solving fluid velocity system... 1 iterations.
-      Relative nonlinear residuals (temperature, compositional fields, Stokes system): 4.91022e-07, 9.24119e-05, 3.1096e-07, 1.60801e-08
-      Relative nonlinear residual (total system) after nonlinear iteration 3: 9.24119e-05
+      Relative nonlinear residuals (temperature, compositional fields, Stokes system): 4.91022e-07, 9.2412e-05, 3.1096e-07, 1.60806e-08
```
tjhei, line 99: tiny differences can be ignored, I think. (I will try to remove the changes and see if the test still passes)

tjhei: it does.

gassmoeller:
I agree with this. It probably needs a rebase after #5074 is merged?

gassmoeller:
Looks good. I will merge for now and make an issue about the edit_parameters test that it needs to be connected to a different signal.

tjhei:
Thanks for merging! I didn't get to modify the test yet but will adjust it now.

PR #5073, opened 2023-02-16
tests: skip refinement in remove_*

tjhei:
The tests remove_angular_momentum and remove_net_rotation do an adaptive refinement step in every timestep, which never causes any refinement. Disable this.

This will slightly change the output, as the preconditioner will no longer be rebuilt.


tjhei:
with this merged, most of the test changes in #5070 go away.

gassmoeller (approved):
Makes sense.

PR #5074, opened 2023-02-16
free surface rebuild preconditioner

### LINKED ISSUE #5071: Update Stokes preconditioner with free surface?
This might be a bug: if we have a free surface, shouldn't we always recompute the preconditioner? We currently have no logic to recompute the preconditioner even though the mesh gets deformed in every step.

While the preconditioner is likely similar (if deformations are small), there is no guarantee that it is still good. Note that tests like tests/remove_angular_momentum only update the preconditioner if the mesh changes (and would never update with a fixed mesh).

While it seems efficient to maybe reuse the preconditioner for a while, I would still suggest to always rebuild it to be consistent. Thoughts?

_Originally posted by @tjhei in https://github.com/geodynamics/aspect/pull/5070#discussion_r1106108216_

tjhei:
Force rebuilding the preconditioner for problems with constant viscosity that have a free surface.

Fixes #5071

### DIFF
--- a/source/simulator/solver_schemes.cc
+++ b/source/simulator/solver_schemes.cc
@@ -374,7 +374,9 @@ namespace aspect
     // don't need to force assembly of the matrix.
     if (stokes_matrix_depends_on_solution()
         ||
-        (boundary_velocity_manager.get_active_boundary_velocity_conditions().size() > 0))
+        (boundary_velocity_manager.get_active_boundary_velocity_conditions().size() > 0)
+        || parameters.mesh_deformation_enabled
+       )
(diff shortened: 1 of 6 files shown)


tjhei:
(I am opening this to see how many tests are affected)

gassmoeller (approved):
Yes, makes sense.

gassmoeller:
Is this still work in progress?

tjhei:
> Is this still work in progress?

I think this is a reasonable change and it is ready to go from my side unless somebody is unhappy with the additional computational cost (in this specific situation).

gassmoeller:
Ok, I am fine with it as well. Just wondering, because it still had the WIP label.

PR #5075, opened 2023-02-16
Add use merged grid option

MFraters:
Add an option to use one grid instead of two merged grids to get lithospheric boundary indicators. This is useful when you have a lithosphere of a 100km and a model which is very deep. Without this, you would need to use the repetitions function to subdivide the whole domain into 100km blocks (there is a minimum of repetition of 1, so 50km blocks?), before the normal refinement can work. To make it square you would need to do that in all dim directions. In my case my boundary elements are fixed in size the whole way around for the transition between the lithosphere and mantle, so it works well for me. 

This is off by default (use merged grid = true) and I added a warning about potential issues when turning this on in the parameter description.


include/aspect/geometry_model/two_merged_boxes.h
```diff
@@ -195,6 +195,15 @@ namespace aspect
         parse_parameters (ParameterHandler &prm) override;
 
       private:
+        /**
+         * Wether to make the grid by gluing together two boxes, or just
+         * use one chunk to make the grid. Using two grids glued together
+         * is a safer option, since it forces the the boundary conditions
+         * to be always applied to the same depth, but using one grid allows
+         * for a more flexable usage of the adaptive refinement.
```
gassmoeller, line 203:
```suggestion
         * Whether to make the grid by gluing together two boxes, or just
         * use one chunk to make the grid. Using two grids glued together
         * is a safer option, since it forces the boundary conditions
         * to be always applied to the same depth, but one unified grid allows
         * for a more flexible usage of the adaptive refinement.
```

source/geometry_model/two_merged_boxes.cc
```diff
@@ -443,6 +454,20 @@ namespace aspect
...
+
+          // grid creation paramters
+          prm.declare_entry ("Use merged grids", "true",
+                             Patterns::Bool (),
+                             "Wether to make the grid by gluing together two boxes, or just "
+                             "use one chunk to make the grid. Using two grids glued together "
+                             "is a safer option, since it forces the the boundary conditions "
+                             "to be always applied to the same depth, but using one grid allows "
+                             "for a more flexable usage of the adaptive refinement. Note that if "
+                             "there is no cell boudary exactly on the boundary between the lithosphere "
```
gassmoeller, line 466:
```suggestion
                             "there is no cell boundary exactly on the boundary between the lithosphere "
```

gassmoeller (commented):
I think this makes sense to have as an option. A few small comments on the wording.

MFraters:
Thanks for the review. I have applied your suggestions to fix the text.

include/aspect/geometry_model/two_merged_chunks.h
```diff
@@ -272,6 +272,15 @@ namespace aspect
         parse_parameters (ParameterHandler &prm) override;
 
       private:
+        /**
+         * Wether to make the grid by gluing together two chunks, or just
```
gassmoeller, line 276: This comment is missing the updates.

source/geometry_model/two_merged_boxes.cc
```diff
@@ -443,6 +454,20 @@ namespace aspect
                              Patterns::Bool (),
                              "Whether the box should be periodic in Y direction in the lithosphere. "
                              "This value is ignored if the simulation is in 2d. ");
+
+          // grid creation paramters
+          prm.declare_entry ("Use merged grids", "true",
+                             Patterns::Bool (),
+                             "Wether to make the grid by gluing together two boxes, or just "
```
gassmoeller, line 461:
```suggestion
                             "Whether to make the grid by gluing together two boxes, or just "
```

include/aspect/geometry_model/two_merged_chunks.h
```diff
@@ -272,6 +272,15 @@ namespace aspect
...
+         * use one chunk to make the grid. Using two grids glued together
+         * is a safer option, since it forces the the boundary conditions
```
gassmoeller, line 278:
```suggestion
         * is a safer option, since it forces the boundary conditions
```

gassmoeller (commented):
I think you missed some places (I only commented once, but most comments apply to both the box and the chunk, and to the place in the header, and in the parameter description, and in the plugin description at the bottom).

MFraters:
ah, sorry, I should have thought about that. I will make the fixes for both plugins

MFraters:
Oke, I think I now addressed all your comments. I did a one-by-one search and replace for all the spelling mistakes you found in all the header and source files of this pull request ;)

gassmoeller (approved):
Thanks :+1:

PR #5076, opened 2023-02-17
allow chunk and two merged chunks for spherical velocity statistics.

MFraters:
Since there are spherical geometries as well, it makes sense to me to allow them.

### DIFF
--- a/source/postprocess/spherical_velocity_statistics.cc
+++ b/source/postprocess/spherical_velocity_statistics.cc
@@ -21,6 +21,8 @@
 
 
 #include <aspect/postprocess/spherical_velocity_statistics.h>
+#include <aspect/geometry_model/two_merged_chunks.h>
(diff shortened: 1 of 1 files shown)


tjhei:
Did you check that it works correctly?

tjhei (approved):

MFraters:
I have checked that the values look reasonable with the two merged chunks:
```
RMS, max velocity:                              0.014 m/year, 0.133 m/year
Radial RMS, tangential RMS, total RMS velocity: 0.00272 m/yr, 0.0138 m/yr, 0.014 m/yr
```
This is subduction model, with a small slab in a large box, with a large part of the lithosphere moving at about 2 to 5 cm/yr horizontally. The total RMS is the same, the tangantial RMS (horizontal direction I assume) is nearly the same as the RMS, which I think makes sense in this mode and the Radial RMS (depth) is much lower.

PR #5077, opened 2023-02-21
Fix adiabatic properties in depth average postprocessor for some special cases

gassmoeller:
See the changelog entry. This started as the observation that I could not use the depth average postprocessor for a single slice, but it also uncovered that the adiabatic condition properties for the depth average output were computed in the wrong places for custom depth bands. The solution I propose is to treat the adiabatic conditions like all other output quantities and average them spatially (even if laterally they are always the same), which unifies the code and makes the buggy (and ill-defined / bad interface) functions obsolete. It also produces a more accurate depth averaging of the adiabatic properties.


tests/adiabatic_conditions_composition_function/depth_average.gnuplot
```diff
@@ -5,73 +5,73 @@
 # For a description of the GNUPLOT format see the GNUPLOT manual.
 #
 # <depth> <time> <temperature> <C_1> <adiabatic_temperature> <adiabatic_pressure> <adiabatic_density> <adiabatic_density_derivative> <velocity_magnitude> <sinking_velocity> <rising_velocity> <viscosity> <log_viscosity> <vertical_heat_flux> <vertical_mass_flux> <C_1_mass> 
-0 0 1613 0 1613 0 3300 0 7.06667e-16 7.80424e-17 7.80424e-17 1e+21 21 3.92899e-08 4.95593e-13 0 
-289000 0 1613 0 1613 0 3300 0 7.06667e-16 7.80424e-17 7.80424e-17 1e+21 21 3.92899e-08 4.95593e-13 0 
+0 0 1613 0 1613 4.64175e+09 3300 0 7.06667e-16 7.21373e-17 7.80424e-17 1e+21 21 3.92907e-08 4.95593e-13 0 
+289000 0 1613 0 1613 4.64175e+09 3300 0 7.06667e-16 7.21373e-17 7.80424e-17 1e+21 21 3.92907e-08 4.95593e-13 0
```
jdannberg, line 9: This file has changes that are not in the adiabatic properties, but instead in the sinking velocity and the vertical heat flux. They are very small, but I don't see why anything should be changed at all. Can you check what happened here?

tests/ascii_data_adiabatic_conditions/depth_average.gnuplot
```diff
@@ -5,73 +5,73 @@
 # For a description of the GNUPLOT format see the GNUPLOT manual.
 #
 # <depth> <time> <temperature> <adiabatic_temperature> <adiabatic_pressure> <adiabatic_density> <adiabatic_density_derivative> <velocity_magnitude> <sinking_velocity> <rising_velocity> <viscosity> <log_viscosity> <vertical_heat_flux> <vertical_mass_flux> 
-0 0 3.05 3 1 5 3.46021e-07 4.10651e-13 5.50849e-14 5.50849e-14 1e+21 21 1.6263e-06 7.13147e-18 
-289000 0 3.05 3 1 5 3.46021e-07 4.10651e-13 5.50849e-14 5.50849e-14 1e+21 21 1.6263e-06 7.13147e-18 
+0 0 3.05 3.05 1.05 5.05 3.46021e-07 4.10651e-13 1.27182e-14 5.50849e-14 1e+21 21 1.6263e-06 7.13147e-18 
+289000 0 3.05 3.05 1.05 5.05 3.46021e-07 4.10651e-13 1.27182e-14 5.50849e-14 1e+21 21 1.6263e-06 7.13147e-18
```
jdannberg, line 9: This is another one of these tests where the sinking velocity changes...

jdannberg (approved):
This looks good to me. Thank you for fixing it!
Can you check why some of the tests change properties other than the adiabatic ones? If you think there's a good explanation for that, this can be merged from my side.

gassmoeller:
> Can you check why some of the tests change properties other than the adiabatic ones? If you think there's a good explanation for that, this can be merged from my side.

I verified that those are small changes that got introduced in an earlier PR, but are too small to lead to a failing tester by themselves. I ran the test with the main branch (without the changes in this PR) and it produced exactly the same difference in sinking velocity and vertical heat flux. So I would assume all changes beside the adiabatic property diffs are caused by that.

So ready to go?

source/simulator/lateral_averaging.cc
```diff
@@ -62,6 +63,53 @@ namespace aspect
...
+            {
+              if (property == Property::temperature)
+                output[i] = adiabat.temperature(fe_values.quadrature_point(i));
+              else if (property == Property::pressure)
+                output[i] = adiabat.pressure(fe_values.quadrature_point(i));
+              else if (property == Property::density)
+                output[i] = adiabat.density(fe_values.quadrature_point(i));
+              else if (property == Property::density_derivative)
+                output[i] = adiabat.density_derivative(fe_values.quadrature_point(i));
+            }
```
tjhei, line 105: ``else AssertThrow(false, ExcNotImplemented());``

tjhei (approved):

gassmoeller:
>  `else AssertThrow(false, ExcNotImplemented());?`

Good suggestion, done.

PR #5079, opened 2023-02-22
Install cookbooks and benchmarks

gassmoeller:
I am working on setting up an executable ASPECT tool on geodynamics.org. In order to avoid duplicating our cookbooks and examples, I would like to make sure we can install them into the installation directory. I do not have access to the source and build folder after the installation, and I do not want to clone the repository again, just to get the cookbooks and benchmarks. Currently, this PR will just copy all the cookbooks and benchmarks into CMAKE_INSTALL_PREFIX/cookbooks and CMAKE_INSTALL_PREFIX/benchmarks. This would already work for me, but I wanted to ask:

1. Are these the right folder to use?
2. Would it be easy to set it up like deal.II where we can build all the examples during the build (i.e. find all CMakeLists and compile all shared libraries)? That would make running the examples even easier for users after the installation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -748,6 +748,16 @@ INSTALL(DIRECTORY include/ ${CMAKE_BINARY_DIR}/include/
   COMPONENT includes
   FILES_MATCHING PATTERN "*.h")
 
+# examples:
+INSTALL(DIRECTORY cookbooks/
+  DESTINATION cookbooks
+  COMPONENT examples
+  FILES_MATCHING PATTERN "*")
+INSTALL(DIRECTORY benchmarks/
+  DESTINATION benchmarks
+  COMPONENT examples
+  FILES_MATCHING PATTERN "*")
+
 # cmake stuff:
 INSTALL(FILES ${CMAKE_BINARY_DIR}/forinstall/AspectConfig.cmake ${CMAKE_BINARY_DIR}/AspectConfigVersion.cmake
         DESTINATION "lib/cmake/Aspect/")


CMakeLists.txt
```diff
@@ -748,6 +748,16 @@ INSTALL(DIRECTORY include/ ${CMAKE_BINARY_DIR}/include/
   COMPONENT includes
   FILES_MATCHING PATTERN "*.h")
 
+# examples:
+INSTALL(DIRECTORY cookbooks/
+  DESTINATION cookbooks
+  COMPONENT examples
+  FILES_MATCHING PATTERN "*")
```
tjhei, line 755: Do we want to be more specific about what files we want to install? Do you want everything (including documentation sources, readmes, etc.)? deal.II uses patterns here.

tjhei (approved):

tjhei:
> 1. Are these the right folder to use?

I think that makes sense. I wouldn't worry too much, because ASPECT, in contrast to a library like deal.II, is unlikely to be packaged for inclusion in something like debian or ubuntu, where these kind of questions matter more.

What is the path mapped to if you install to the default location (/usr/local)?

> 2\. Would it be easy to set it up like deal.II where we can build all the examples during the build (i.e. find all CMakeLists and compile all shared libraries)? That would make running the examples even easier for users after the installation.

I think this would be neat. In deal.II you can toggle this behavior on and off. I am not sure what the default option for us should be, as it would increase compilation time quite a bit...

gassmoeller:
> What is the path mapped to if you install to the default location (/usr/local)?

It would be /usr/local/cookbooks and /usr/local/benchmarks, which is obviously not ideal. I have a version that always installs into $CMAKE_INSTALL_PREFIX/share/aspect/cookbooks (or benchmarks), that would be closer to what is expected for software on linux, but would be harder to find for users. It looks like deal.II also always installs in $CMAKE_INSTALL_PREFIX/examples, so I guess we are ok with the current version? I am fine with either.

> I think this would be neat.

I think so too, but I will look into that as a next step in a follow up PR.

tjhei:
Great! Let's merge then.

PR #5085, opened 2023-02-24
Fix grain size additional outputs visualization

gassmoeller:
There was no test for the visualization of the additional material model outputs that the grain size model produces, therefore of course there were bugs :-). Some of these were introduced recently (#5021), but others have been in there for years without being discovered. The bugs only affected the visualization output of named additional outputs (and caused it to crash in debug mode), not the internal workings of the material model (there were tests for that). This PR fixes the asserts and floating point exceptions that were triggered by named additional outputs and adds a test that ensures it works.

@alarshi: This should fix the error you told me about.


anne-glerum (approved):
Great finds! Thanks for the fixes.

anne-glerum:
Indentation needs to be fixed though :)

gassmoeller:
Thanks, indentation is fixed.

>I see that in lines 890-896 seismic_out is only filled if use_table_properties == true, so this change is consistent, but out of curiosity: is there a reason that if density, thermal expansivity and specific heat are not read from a table, no seismic velocities can be outputted? Because it wouldn't be self-consistent?

It is more that the material model has an alternative way to compute density, thermal expansivity and so on if the tables are not used, but it has no way to compute seismic velocities without the tables. So instead of making something up, we just left the values at the default values. But the default values are signaling NaNs, which crashes the postprocessor when it tries to read them and write them into the Paraview output. That is why I decided the best way to fix that would be to only create these additional outputs if the material model can actually fill them.

PR #5086, opened 2023-02-24
Fix documentation for VM

### LINKED ISSUE #5084: Issue on page /user/install/virtual-machine/running-models.html
The run command is incorrect it should be:

~/aspect/build/aspect-release

The commang given in the manual does not work

gassmoeller:
Fixes #5084. @mibillen: do you think this will work?


tjhei (approved):
I can confirm that this works. And sorry for breaking the symlink. :-(

gassmoeller:
No worries, stuff happens. 

I added a note to mention debug mode. I prefer linking to release mode by default, but I agree there are arguments for both sides.

tjhei:
@gassmoeller I just reread the section and bullet point 5. talks about changing debug and release mode, which is not necessary as I compiled both versions.
Are you opposed to changing the instructions to always use ``~/aspect/build/aspect-debug`` or ``~/aspect/build/aspect-release``?

gassmoeller:
Not at all, I am fine either way. Just to clarify: You mean we remove the symlink and require the user to always specify aspect-debug or aspect-release in the build folder? I think that is actually better, because it makes the choice more visible for the user.

tjhei:
Precisely.

PR #5088, opened 2023-02-24
Include surface topography from mesh deformation in depth function

Djneu:
Hi all,
I noticed that the current depth function does not include changes to the surface related to mesh deformation. I made a new update function for the geometry model that will combine the surface from all processors and broadcast this. A new depth function uses this surface to find the two nearest surface points to the given position, and uses linear interpolation to find the surface topography and depth. Attached is an image to show the depth tracked using this new function vs. the current function when using the boundary function for mesh deformation.

depth_comp

For now the code is a bit messy, and only works for 2D box geometries (without any asserts to check for this). Also, at the moment the surface isn't sorted after combining them, as this wasn't necessary on the few processors I tested it on, but may need to be done later on.  I'm also not entirely sure when is the best time to update the surface in relation to the mesh deformation.

Before I spend more time on this, does this seem like a good way to implement it?


gassmoeller:
Hi Derek,
sorry for not reacting earlier to your PR. I think you have a valid point that this functionality would be useful, and we have discussed this a few times among the maintainers in the past. You also already realized that this is difficult to implement for a general model. Let me just mention the difficulties for completeness when others read this PR:
- the definition of 'depth' in an arbitrarily deformed model is not unique. it could mean: 
  - distance normal to surface, 
  - distance along an imaginary 'vertical' axis to surface which depends on the geometry (this is the definition you chose), 
  - distance along the gradient of the gravity potential field (this would be most natural to compute pressure profiles and is the same to the second option as long as gravity vectors are aligned with the 'vertical' axis, but that may not always be the case).

I think your chosen definition does make sense, but like you noticed the implementation adds further complications:
- how to track this in 3d models (possibly with different mesh refinement levels at different locations of the surface)?
- how to track this in arbitrary geometries?
- how to accurately retrieve this information for points at depth?

That being said, like I mentioned I think having something that only works for 2D boxes is better than having nothing, so I would be ok with moving forward with this PR. However, please make the following changes:

- rename the `update_surface` function for the geometry_model to `update` like all other plugin systems. Who knows what other update tasks may be necessary for geometry models in the future?
- instead of making a new `depth_including_mesh_deformation` function, overload the existing `depth` function with an additional argument called `include_mesh_deformation`. The new function declaration should look like this:
```
double depth(const Point<dim> &position,
                        const bool include_surface_deformation) const override;
```
- Instead of your own implementation of an interpolation function using surface_x and surface_y you can use the deal.II class `InterpolatedTensorProductGridData` (https://dealii.org/developer/doxygen/deal.II/classFunctions_1_1InterpolatedTensorProductGridData.html). You can see how this class is used in aspect/structured_data.h and source/structured_data.cc (admittedly in a very complicated example). Your case would be much simpler, more like in this deal.II test:  https://github.com/dealii/dealii/blob/9154fa4d2194ca38ae10214a0d5b302fc1b1d3bd/tests/base/functions_10.cc#L73.
- Only compute the surface deformation topography if mesh deformation is enabled and the model supports the retrieval of that information.
- Make sure you add asserts to the new depth function to check that we support the topography inclusion for the current model (i.e. only 2D boxes)


Let me know when you have made those changes then I will do a proper review of the code.

include/aspect/geometry_model/box.h
```diff
@@ -64,6 +64,11 @@ namespace aspect
          */
         void create_coarse_mesh (parallel::distributed::Triangulation<dim> &coarse_grid) const override;
 
+        /**
+         * Function to store data on current surface topography.
+         */
+        void update_surface () override;
```
bangerth, line 70: Like @gassmoeller already said, name it `update()`. There may in fact already be an `update()` function declared in the base class.

include/aspect/geometry_model/box.h
```diff
@@ -105,6 +110,15 @@ namespace aspect
          */
         double depth(const Point<dim> &position) const override;
 
+        /**
+         * Function that uses the stored surface topography to calculate the depth
+         * including changes from mesh deformation. This will linearly interpolate
+         * between the two nearest surface points to get a depth for the given
+         * position.
+         */
+        double depth_including_mesh_deformation(const Point<dim> &position) const override;
```
bangerth, line 119: @gassmoeller suggests to just extend the existing `depth()` function by another parameter. I'm not sure that's the right approach because changing the interface of a `virtual` function requires changing it in all derived classes (some of which may just be out there in user land). I'm ok with introducing a new function in the base class that by default just throws an exception, and that individual geometry plugins can overload.

include/aspect/geometry_model/box.h
```diff
@@ -239,6 +253,12 @@ namespace aspect
          * A pointer to the initial topography model.
          */
         InitialTopographyModel::Interface<dim> *topo_model;
+
+        /**
+         * Variables to hold surface topography when checking the depth including mesh deformation.
+         */
+        std::vector<double> surface_x;
+        std::vector<double> surface_y;
```
bangerth, line 261: The documentation here suggests that you only need these variables during specific phases of the algorithm. In that case, you could make them local to the function that needs them.

If you need these variables to communicate between functions (e.g., `update()` computes the values, `depth_including_mesh_deformation()` uses them), then member variables are the right way to go. But you should find a dimension-independent way to deal with this. As already mentioned, you could use one of the deal.II tables if the data is structured on an x-y grid.

source/geometry_model/box.cc
```diff
@@ -260,6 +260,183 @@ namespace aspect
...
+    {
+        this->get_pcout() << "   Updating surface values... " <<std::endl;
+        // loop over all of the surface cells and save the elevation to a stored value.
+        // This needs to be sent to 1 processor, sorted, and broadcast so that every processor knows the entire surface.
+        // TODO: Is there a better time to call this in regards to mesh deformation?
+        // TODO: Right now this is messy saving two variables, this can be cleaned up by
+        // combining surface_x and surface_y into one variable and broadcasting them in a loop.
+        std::vector<double> temp_surface_x;
+        std::vector<double> temp_surface_y;
+        std::vector<std::vector<double>> local_surface_height(2, std::vector<double>());
```
bangerth, line 276: as you already noted yourself, this needs a bit of work

source/geometry_model/box.cc
```diff
@@ -260,6 +260,183 @@ namespace aspect
...
+
+            double vector_size = temp_surface_x.size();
+            MPI_Bcast(&vector_size, 1, MPI_DOUBLE, 0, this->get_mpi_communicator());
+            MPI_Bcast(&temp_surface_x[0], vector_size, MPI_DOUBLE, 0, this->get_mpi_communicator());
+            MPI_Bcast(&temp_surface_y[0], vector_size, MPI_DOUBLE, 0, this->get_mpi_communicator());
+        }
+        else
+        {
+          for (unsigned int i=0; i<local_surface_height.size(); i++)
+              MPI_Ssend(&local_surface_height[i][0], local_surface_height[1].size(), MPI_DOUBLE, 0, 42, this->get_mpi_communicator());
```
bangerth, line 382: And here you want to use the corresponding isend function: https://dealii.org/developer/doxygen/deal.II/namespaceUtilities_1_1MPI.html#a82095f4e891152bab36ebfea1423620b

source/geometry_model/box.cc
```diff
@@ -260,6 +260,183 @@ namespace aspect
...
+            MPI_Get_count(&status, MPI_DOUBLE, &incoming_size);
+
+            // Resize the array so it fits whatever the process sends.
+            std::vector<std::vector<double>> temporary_surface(2, std::vector<double>());
+
+            for (unsigned int i=0; i<temporary_surface.size(); ++i)
+                temporary_surface[i].resize(incoming_size);
+
+            for (unsigned int i=0; i<temporary_surface.size(); ++i)
+              MPI_Recv(&temporary_surface[i][0], incoming_size, MPI_DOUBLE, p, 42, this->get_mpi_communicator(), &status);
```
bangerth, line 343: There are easier ways to do this. Here, you want to use this function: https://dealii.org/developer/doxygen/deal.II/namespaceUtilities_1_1MPI.html#ad45956802191bd6765cdc5291a605acf

source/geometry_model/box.cc
```diff
@@ -260,6 +260,183 @@ namespace aspect
...
+                    temp_surface_y.push_back(temporary_surface[1][i]);
+                  }
+                }
+              }
+          }
+
+            double vector_size = temp_surface_x.size();
+            MPI_Bcast(&vector_size, 1, MPI_DOUBLE, 0, this->get_mpi_communicator());
+            MPI_Bcast(&temp_surface_x[0], vector_size, MPI_DOUBLE, 0, this->get_mpi_communicator());
+            MPI_Bcast(&temp_surface_y[0], vector_size, MPI_DOUBLE, 0, this->get_mpi_communicator());
```
bangerth, line 377: Take a look at deal.II's broadcast function: https://dealii.org/developer/doxygen/deal.II/namespaceUtilities_1_1MPI.html#a1dc323423b78cd47ff56f1dde5f61488

source/geometry_model/box.cc
bangerth: The point being that deal.II's isend/irecv/broadcast functions can deal with arbitrary object types. You won't have to worry about it being a vector of doubles, for example. It can be more or less any data structure you choose, however complicated -- may be a `std::map`, for example.

source/geometry_model/box.cc
```diff
@@ -260,6 +260,183 @@ namespace aspect
...
+              x1 = x2; y1 = y2;
+            }
+
+            // If we are at the edge of the model domain, use that y value. Otherwise interpolate.
+            if(position[0] >= surface_x[surface_x.size() - 1])
+              height = surface_y[surface_y.size() - 1];
+            else if (position[0] <= surface_x[0])
+              height = surface_y[0];
+            else
+              height = ((y2 - y1)/(x2-x1))*(position[0] - x1) + y1;
```
bangerth, line 430: This will do (and probably do it better :-) ): https://dealii.org/developer/doxygen/deal.II/classFunctions_1_1InterpolatedTensorProductGridData.html

Or if the data are actually equally spaced, then https://dealii.org/developer/doxygen/deal.II/classFunctions_1_1InterpolatedUniformGridData.html

source/geometry_model/interface.cc
```diff
@@ -36,6 +36,19 @@ namespace aspect
...
+    Interface<dim>::update_surface ()
+    {}
+
+
+    template <int dim>
+    double
+    Interface<dim>::depth_including_mesh_deformation(const Point<dim> &position) const
+    {
+      return depth(position);
+    }
```
bangerth, line 50: You want to throw an exception here. Models that don't implement this function shouldn't return wrong values, but just abort.

source/postprocess/visualization/depth.cc
```diff
@@ -54,7 +54,7 @@ namespace aspect
 
         for (unsigned int q=0; q<n_quadrature_points; ++q)
           {
-            computed_quantities[q](0) = this->get_geometry_model().depth (input_data.evaluation_points[q]);
+            computed_quantities[q](0) = this->get_geometry_model().depth_including_mesh_deformation (input_data.evaluation_points[q]);
```
bangerth, line 57: And for this, I'd suggest that you duplicate the plugin and create a new one that calls the function you have here. Then the documentation would explain what each does, and how they differ.

bangerth (changes requested):

include/aspect/geometry_model/box.h
Djneu: Renamed to update.

Djneu:
Thank you for the comments and suggestions @gassmoeller and @bangerth! Sorry it took me awhile to get back to this, I've tried to clean up the code a bit and put in your suggestions. I've created a new visualization plugin that specifically uses the depth_with_deformation function, and renamed the main function to only update. 

I've also put the x and y coordinates of the  surface into a local vector of vectors that is then sorted/put into a data table on processor zero before being broadcast back to the others to create the InterpolatedTensorProductGridData function. Right now I have this function as a member variable so that it can then be called from the depth function. Does this make sense or would it be better to save the actual surface coordinate values as a member variable and create the function later?

I tried a bit to get this working in a 3D box but was running into ordering issues for InterpolatedTensorProductGridData. Right now in 3D the points will be sorted such that they ascend in Y before going up in X. I've tried this the other way with going up in Y before X, but it also didn't like this either. It seems to want to both Y and X always ascending, but I'm not sure how to make this work with the way I've organized the values. Is there something I'm missing on how to sort the data for a 2D grid?

source/geometry_model/box.cc
Djneu: I've kept it as is for now, but I can update it to isend in the future when dealii 9.5 is required.

source/geometry_model/box.cc
Djneu: Updated.

include/aspect/geometry_model/box.h
Djneu: Renamed.

include/aspect/geometry_model/box.h
Djneu: I've kept the function as is, though for now it is only called in the visualization plugins.

include/aspect/geometry_model/box.h
Djneu: I've rolled all the x,y and z values that may be at the surface into a local vector of vectors, and now have the Interpolated function as the member variable. Does it make sense to do it this way?

source/geometry_model/interface.cc
Djneu: I've added an assertthrow.

source/postprocess/visualization/depth.cc
Djneu: Done.

anne-glerum:
@Djneu, @gassmoeller and @bangerth This functionality would be great to have! From Derek's message I gather this is ready for another review, although there are some functions that could be changed now that deal.II 9.5 is required. Would you be willing to have another look?

source/geometry_model/box.cc
```diff
@@ -257,6 +259,163 @@ namespace aspect
...
+                  if ( cell->face(face_no)->boundary_id() != relevant_boundary)
+                    continue;
+
+                  fe_face_values.reinit(cell, face_no);
+
+                  for (unsigned int corner = 0; corner < face_corners.size(); ++corner)
+                    {
+                      const Point<dim> vertex = fe_face_values.quadrature_point(corner);
+
+                      // We can't push back a point so we convert it into a vector.
```
gassmoeller, line 304: I dont understand this comment. The following should work perfectly fine:
```
std::vector<Point<dim>> points;
Point<dim> point;
points.push_back(point);
```

Can you explain the problems you had in a bit more detail? I am sure we can figure out how to make this work with a vector of points.

source/geometry_model/interface.cc
```diff
@@ -36,6 +36,22 @@ namespace aspect
     {}
 
 
+    template <int dim>
+    void
+    Interface<dim>::update ()
+    {}
```
gassmoeller, line 42: You can delete this implementation, we do not need it anymore, because there is a default implementation in the base class of the interface (since the hackathon 2024, so quite recently).

source/geometry_model/interface.cc
```diff
@@ -36,6 +36,22 @@ namespace aspect
...
+
+
+    template <int dim>
+    double
+    Interface<dim>::depth_including_mesh_deformation(const Point<dim> &) const
+    {
+      AssertThrow(false,
+                  ExcMessage("The depth including mesh deformation function only works with a 2D box geometry model."));
```
gassmoeller, line 50: Better to write this message generically. Once we implement this for a few more models the current form will become out of date.
```suggestion
                  ExcMessage("The function depth_including_mesh_deformation () was called, but it is not implemented for the current geometry model."));
```

source/postprocess/visualization/depth.cc
```diff
@@ -73,7 +73,7 @@ namespace aspect
                                                   "depth",
                                                   "A visualization output postprocessor that outputs "
                                                   "the depth for all points inside the domain, as "
-                                                  "determined by the geometry model."
+                                                  "determined by the initial surface of the geometry model. "
```
gassmoeller, line 76: I think this is a bit more precise, as the initial surface could have already been deformed (mesh deformation can create initial topography).
```suggestion
                                                  "determined by the undeformed surface of the geometry model. "
```

source/postprocess/visualization/depth_including_mesh_deformation.cc
```diff
@@ -0,0 +1,82 @@
+/*
+  Copyright (C) 2011 - 2022 by the authors of the ASPECT code.
```
gassmoeller, line 2:
```suggestion
  Copyright (C) 2024 -  by the authors of the ASPECT code.
```

source/postprocess/visualization/depth_including_mesh_deformation.cc
```diff
@@ -0,0 +1,82 @@
...
+{
+  namespace Postprocess
+  {
+    namespace VisualizationPostprocessors
+    {
+      template <int dim>
+      DepthIncMeshDef<dim>::
+      DepthIncMeshDef ()
+        :
+        DataPostprocessorScalar<dim> ("depth",
```
gassmoeller, line 37: This is the name of the variable as it will appear in Paraview. Need to make sure it is not duplicated with the other postprocessor.
```suggestion
        DataPostprocessorScalar<dim> ("depth_including_mesh_deformation",
```

source/simulator/core.cc
```diff
@@ -1839,6 +1839,14 @@ namespace aspect
         // calculate global volume after deforming mesh
         global_volume = GridTools::volume (triangulation, *mapping);
         signals.post_mesh_deformation(*this);
+
+        // We update this after mesh deformation so the most recent surface is kept.
```
gassmoeller, line 1843: Just call the `update` function unconditionally here. The geometry model interface or the box geometry model is responsible to make sure to behave correctly for each call of `update`. Your current implementation makes assumptions about the implementation of specific plugins in the core part of ASPECT, which means we would have to change the core if the implementation of the plugin changes, which we want to avoid. 
There is still one small problem with the current implementation, which is that now the geometry model `update` function is called twice, once in `start_timestep` and another time here. This contradicts our documentation of the update function, namely that it is called exactly once per time step at the beginning of a new time step. I have to think about this a bit more, and if there is a way to prevent it.

source/geometry_model/box.cc
```diff
@@ -257,6 +259,163 @@ namespace aspect
...
+                      // This is needed later to keep the vertexes together when sorting.
+                      std::vector<double> vertex_row;
+                      for(unsigned int i=0; i<dim; ++i)
+                        vertex_row.push_back(vertex[i]);
+
+                      local_surface_height.push_back(vertex_row);
+                    }
+                }
+
+        // Combine all local_surfaces and broadcast back.
```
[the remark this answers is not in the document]
gassmoeller: Actually I accidentally discovered the following function in deal.II which probably does exactly what you need [set_uni](https://dealii.org/current/doxygen/deal.II/namespaceUtilities_1_1MPI.html#a76fee268a754b37289d7f335c23c1957). In fact this even sorts and eliminated duplicates entries for you. The implementation of the function is:
```
    template <typename T>
    std::vector<T>
    compute_set_union(const std::vector<T> &vec, const MPI_Comm comm)
    {
      return Utilities::MPI::all_reduce<std::vector<T>>(
        vec, comm, [](const auto &set_1, const auto &set_2) {
          // merge vectors, sort, and eliminate duplicates -> mimic std::set<T>
          auto result = set_1;
          result.insert(result.end(), set_2.begin(), set_2.end());
          std::sort(result.begin(), result.end());
          result.erase(std::unique(result.begin(), result.end()), result.end());
          return result;
        });
    }
```

source/geometry_model/box.cc
```diff
@@ -257,6 +259,163 @@ namespace aspect
       return std::min (std::max (d, 0.), maximal_depth());
     }
 
+    template <int dim>
+    void
+    Box<dim>::
+    update ()
+    {
+        AssertThrow(Plugins::plugin_type_matches<const GeometryModel::Box<dim>>(this->get_geometry_model()),
```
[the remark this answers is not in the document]
Djneu: I've removed both asserts.

source/geometry_model/box.cc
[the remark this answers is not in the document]
Djneu: This worked great, thank you!

source/geometry_model/box.cc
Djneu: You're right this does work I don't what the comment was for. I did run into an issue later when trying to use compute_set_union with "std::vector<Point<dim>> points;" should I leave it as a vector of vectors for this?

Here is an example of the error message:

> required from 'std::vector<ValueType> dealii::Utilities::MPI::compute_set_union(const >std::vector<ValueType>&, MPI_Comm) [with T = dealii::Point<2, double>; MPI_Comm = ompi_communicator_t*]'
>/home/bbpdneu1/software/depth_mesh/aspect/source/geometry_model/box.cc:327:45:   required from 'void aspect::GeometryModel::Box<dim>::update() [with int dim = 2]'
/home/bbpdneu1/software/depth_mesh/aspect/source/geometry_model/box.cc:609:5:   required from here
/sw/spack/24.05/linux-rocky8-x86_64/opt/spack/linux-rocky8-x86_64/gcc-8.5.0/gcc-13.2.0-ywo3vxag72klu2wt4of6ot7kw7mftyig/lib/gcc/x86_64-pc-linux-gnu/13.2.0/../../../../include/c++/13.2.0/bits/predefined_ops.h:69:22: error: no match for 'operator<' (operand types are 'dealii::Point<2, double>' and 'dealii::Point<2, double>')
   69 |       { return *__it < __val; }

source/geometry_model/box.cc
```diff
@@ -268,6 +269,115 @@ namespace aspect
...
+      if (dim==2)
+        {
+          for (unsigned int x=0; x<(data_table.size()[0]); ++x)
+            {
+              idx[0] = x;
+              data_table(idx) = temp_surface[x][1];
+            }
+        }
+
+      //Fill the coordinates with the x-values used for the data table.
```
gassmoeller, line 352:
```suggestion
      // Fill the coordinates with the x-values used for the data table.
```

source/geometry_model/box.cc
```diff
@@ -268,6 +269,115 @@ namespace aspect
...
+
+      // Remove non-unique rows from the sorted 2D vector
+      auto last = std::unique(temp_surface.begin(), temp_surface.end(), compareRows);
+      temp_surface.erase(last, temp_surface.end());
+
+
+      // Resize data table.
+      TableIndices<dim-1> size_idx;
+      for (unsigned int d=0; d<dim-1; ++d)
+        size_idx[d] = temp_surface.size();
```
gassmoeller, line 337: this part will be the problem to generalize to 3D, correct? We do not know how to create the dimensions of the table in when the surface points are only one long one-dimensional list.

source/geometry_model/box.cc
```diff
@@ -268,6 +269,115 @@ namespace aspect
       return std::min (std::max (d, 0.), maximal_depth());
     }
 
+    template <int dim>
+    void
+    Box<dim>::
+    update ()
+    {
+      this->get_pcout() << "   Updating surface values... " << std::endl;
+      TimerOutput::Scope timer_section(this->get_computing_timer(), "Geometry model surface update");
```
gassmoeller, line 278: Only do all of this if `dim == 2` for now, until we have an implementation for 3D as well.

include/aspect/geometry_model/box.h
```diff
@@ -247,6 +261,11 @@ namespace aspect
          * The number of cells in each coordinate direction.
          */
         std::array<unsigned int, dim> repetitions;
+
+        /**
+         * Function to interpolate surface topography.
+         */
+        Functions::InterpolatedTensorProductGridData<dim-1> *surface_function;
```
gassmoeller, line 268: please only use smart pointers like `std::unique_ptr`, not raw pointers like this one. Smart pointer avoid problems of variable ownership and guarantee the memory is correctly freed when the variable is destroyed. See e.g. [here](https://github.com/geodynamics/aspect/blob/e9cd1c8fc83f0d5c6e52c9d775a71382a1597049/include/aspect/material_model/visco_plastic.h#L233) and [here](https://github.com/geodynamics/aspect/blob/e9cd1c8fc83f0d5c6e52c9d775a71382a1597049/source/material_model/visco_plastic.cc#L376) for how to use them.

source/geometry_model/box.cc
```diff
@@ -268,6 +269,115 @@ namespace aspect
...
+
+      // loop over all of the surface cells and save the elevation to a stored value.
```
gassmoeller, line 280: Currently you do this computation unconditionally whenever the `update` function is called. However, that is expensive and has a number of problems:
- mesh deformation may not actually be active, then all of this is wasted
- in your current implementation `update` is called twice, once from [here](https://github.com/geodynamics/aspect/blob/e9cd1c8fc83f0d5c6e52c9d775a71382a1597049/source/simulator/core.cc#L645) and once from your new place [here](https://github.com/geodynamics/aspect/pull/5088/files#diff-72347c8b280f31adfdf4a3c112dbc352e9de7677d261e62dba25bef499fe674aR1897), only the second call is a useful update to the surface function.

As a first step you could only do this whole calculation if `this->get_parameters().mesh_deformation_enabled` is true. But then it would still be done twice per timestep. Additionally you could introduce a class member variable `bool need_surface_function_update`, which you set to false at the end of the update function. You then need to set this variable to true whenever the mesh changes, which you can do by connecting a small lambda function to the signal `this->simulator_signals().post_mesh_deformation`. This could look similar to the following (put this into the `initialize` function):
```
          this->get_signals().post_mesh_deformation.connect(
            [&](const SimulatorAccess<dim> &/*simulator_access*/)
          {
            need_surface_function_update = true;
          }
```

source/geometry_model/box.cc
```diff
@@ -268,6 +269,115 @@ namespace aspect
...
+              data_table(idx) = temp_surface[x][1];
+            }
+        }
+
+      //Fill the coordinates with the x-values used for the data table.
+      std::array<std::vector<double>, dim-1> coordinates;
+      for (unsigned int i=0; i<temp_surface.size(); ++i)
+        coordinates[0].push_back(temp_surface[i][0]);
+
+      surface_function = new Functions::InterpolatedTensorProductGridData<dim-1> (coordinates, data_table);
```
gassmoeller, line 357: Your use of a raw pointer has already created a memory leak here. Every time this function is called you will create a new object of type `InterpolatedTensorProductGridData`, but you do not delete the old one. If you switch to a `unique_ptr` as I suggested above this will be automatically fixed (because when the old unique_ptr is destroyed it will automatically delete the object it owned.

gassmoeller (commented):
Getting closer. I have some more comments, mostly dealing with how to efficiently integrate this into ASPECT so that this computation is not done unnecessarily. Let me know in case you have questions.

source/geometry_model/box.cc
Djneu: Done.

source/geometry_model/box.cc
Djneu: I think I've updated this correctly but let me know if it is off.

source/geometry_model/box.cc
```diff
@@ -268,6 +269,115 @@ namespace aspect
...
+    void
+    Box<dim>::
+    update ()
+    {
+      this->get_pcout() << "   Updating surface values... " << std::endl;
+      TimerOutput::Scope timer_section(this->get_computing_timer(), "Geometry model surface update");
+
+      // loop over all of the surface cells and save the elevation to a stored value.
+      // This needs to be sent to 1 processor, sorted, and broadcast so that every processor knows the entire surface.
+      // (Does bcast create memory of the entire variable on every processor or only a pointer to the one stored on 0?)
```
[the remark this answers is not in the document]
Djneu: I think for now I'll leave this as is, and if it becomes a problem later on for 3D I can revisit it.

include/aspect/geometry_model/box.h
Djneu: Done.

source/geometry_model/box.cc
Djneu: Yes, this seems to be the problem. I tried to get 3D working again and for uniform meshes it works fine. But once there is a specified region of higher resolution, the number of actual points in the temporary surface is less than what is expected in the data table that is made from the unique x and y points.

source/simulator/core.cc
Djneu: Done.

doc/modules/changes/20241104_neuharth
```diff
@@ -0,0 +1,6 @@
+New: In 2D, the 'box' geometry model can now account for changes
+to depth due to mesh deformation. This can be visualized using
+the new 'depth with mesh deformation' postprocessor. 
+Note: This change would need to be added to material models if needed.
```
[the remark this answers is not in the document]
Djneu: Done.

bangerth:
I'll put that on my list for this afternoon.

bangerth:
But you'll need to fix the indentation and test failures first.

bangerth:
Can you make sure the indentation and tests succeed?

Djneu:
Took awhile for the tests, but should be ready for a review now!

include/aspect/geometry_model/box.h
```diff
@@ -245,6 +259,13 @@ namespace aspect
          * The number of cells in each coordinate direction.
          */
         std::array<unsigned int, dim> repetitions;
+
+        bool need_surface_function_update = false;
```
bangerth, line 263: Can you document this variable? Separately, we tend not to initialize member variables in-class, but in the constructor of the class.

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+    void
+    Box<dim>::
+    update ()
+    {
+      if (dim==2 && this->get_parameters().mesh_deformation_enabled && need_surface_function_update)
+        {
+          TimerOutput::Scope timer_section(this->get_computing_timer(), "Geometry model surface update");
+
+          // loop over all of the surface cells and save the elevation to a stored value.
+          // This needs to be sent to 1 processor, sorted, and broadcast so that every processor knows the entire surface.
```
bangerth, line 290: That's both a lot of work, and a lot of memory. I think that pertains to my top-level comment about doing all of this for uncertain uses.

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+          std::vector<std::vector<double>> local_surface_height;
+          Table<dim-1, double> data_table;
+          const types::boundary_id relevant_boundary = this->get_geometry_model().translate_symbolic_boundary_name_to_id ("top");
```
bangerth, line 293: `this->get_geometry_model()` actually equals `this` :-)

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+                                           this->get_fe(),
+                                           face_corners,
+                                           update_quadrature_points);
+
+          // Loop over all corners at the surface and save their position.
+          // TODO: Update this to work in 3D. Spherical?
+          for (const auto &cell : this->get_dof_handler().active_cell_iterators())
+            if (cell->is_locally_owned() && cell->at_boundary())
+              for (const unsigned int face_no : cell->face_indices())
+                if (cell->face(face_no)->at_boundary())
```
bangerth, line 305: Internally, `cell->at_boundary()` loops over all faces and checks whether any of those are at the boundary. You don't save anything here with the second part of the top `if` condition because you're doing the same work again in the second and third line.

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+                  {
+                    if ( cell->face(face_no)->boundary_id() != relevant_boundary)
+                      continue;
```
bangerth, line 308: This is the same as
```suggestion
                if (cell->face(face_no)->at_boundary())
                    if ( cell->face(face_no)->boundary_id() == relevant_boundary)
                       {
```
which I think is easier to read.

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+                      {
+                        const Point<dim> vertex = fe_face_values.quadrature_point(corner);
+
+                        // We can't push back a point so we convert it into a vector.
+                        // This is needed later to keep the vertices together when sorting.
+                        std::vector<double> vertex_row;
+                        for (unsigned int i=0; i<dim; ++i)
+                          vertex_row.push_back(vertex[i]);
+
+                        local_surface_height.push_back(vertex_row);
```
bangerth, line 322: You can't push back because you define `local_surface_height` as a `vector<vector<double>>`. But you presumably *could* have just declared it as a `vector<Point<dim>>`. Then you don't need the intermediate `vertex_row`.

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+          std::sort(temp_surface.begin(), temp_surface.end(), [](const std::vector<double> &a, const std::vector<double> &b)
+          {
+            return a[0] < b[0];
+          });
+
+          // Define a comparison to remove duplicate surface points.
+          bool (*compareRows)(const std::vector<double> &, const std::vector<double> &) = [](const std::vector<double> &row1, const std::vector<double> &row2)
+          {
+            return row1 == row2;
+          };
```
bangerth, line 340: Not wrong. But this is the idiomatic way:
```suggestion
          const auto compareRows = [](const std::vector<double> &row1, const std::vector<double> &row2)
          {
            return row1 == row2;
          };
```

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+
+          // Remove non-unique rows from the sorted 2D vector
+          auto last = std::unique(temp_surface.begin(), temp_surface.end(), compareRows);
```
bangerth, line 343:
```suggestion
          const auto last = std::unique(temp_surface.begin(), temp_surface.end(), compareRows);
```

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+                        for (unsigned int i=0; i<dim; ++i)
+                          vertex_row.push_back(vertex[i]);
+
+                        local_surface_height.push_back(vertex_row);
+                      }
+                  }
+
+          // Combine all local_surfaces and broadcast back.
+          std::vector<std::vector<double>> temp_surface =
+            Utilities::MPI::compute_set_union(local_surface_height, this->get_mpi_communicator());
```
bangerth, line 328: This is of course going to bring a *lot* of data to all processes.

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+
+          // Fill the data table with the y-values that correspond to the surface.
+          if (dim==2)
+            {
+              for (unsigned int x=0; x<(data_table.size()[0]); ++x)
+                {
+                  idx[0] = x;
+                  data_table(idx) = temp_surface[x][1];
+                }
+            }
```
bangerth, line 362: You can use the same trick to avoid the `idx` object if you use `if constexpr`, as here: https://github.com/geodynamics/aspect/pull/6358/files

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+        }
+    }
+
+    template <int dim>
+    double
+    Box<dim>::depth_including_mesh_deformation(const Point<dim> &position) const
+    {
+
+      AssertThrow(Plugins::plugin_type_matches<const GeometryModel::Box<dim>>(this->get_geometry_model()),
+                  ExcMessage("Depth with mesh deformation currently only works with a 2D box geometry model."));
```
bangerth, line 384: But you know this is true because you're in the Box class!

source/geometry_model/box.cc
```diff
@@ -268,6 +277,128 @@ namespace aspect
...
+      AssertThrow (this->get_parameters().mesh_deformation_enabled,
+                   ExcMessage("Depth with mesh deformation must be used with mesh deformation activated."));
+
+      // Convert the point to dim-1, as we aren't interested in the vertical component
+      // for the function.
+      Point<dim-1> p;
+      for (unsigned int d=0; d<dim-1; ++d)
+        p[d] = position[d];
+
+      double depth_from_surface = surface_function->value(p) - position[dim-1];
```
bangerth, line 398:
```suggestion
      const double depth_from_surface = surface_function->value(p) - position[dim-1];
```

PR #5089, opened 2023-02-24
Update initial lithostatic pressure plugin to account for initial topography and mesh deformation.

anne-glerum:
I've updated the plugin to hopefully deal more self-consistently with initial topography and mesh deformation. The lithostatic pressure profile that is computed at t0 now uses the actual initial topography at the reference point instead of the maximum initial topography. Also, when traction is prescribed on a bottom boundary, the last pressure in the profile is prescribed, instead of using what the geometry model returns as depth to interpolate the pressure profile.

### For all pull requests:

### DIFF
--- a/source/boundary_traction/initial_lithostatic_pressure.cc
+++ b/source/boundary_traction/initial_lithostatic_pressure.cc
@@ -22,6 +22,8 @@
 #include <aspect/boundary_traction/initial_lithostatic_pressure.h>
 #include <aspect/initial_temperature/interface.h>
 #include <aspect/initial_composition/interface.h>
+#include <aspect/geometry_model/initial_topography_model/zero_topography.h>
+#include <aspect/geometry_model/initial_topography_model/interface.h>
 #include <aspect/gravity_model/interface.h>
 #include <aspect/global.h>
 #include <aspect/utilities.h>
@@ -59,9 +61,6 @@ namespace aspect
       // but we use the initial temperature and composition and only calculate
       // a pressure profile with depth.
 
-      // The spacing of the depth profile
-      delta_z = this->get_geometry_model().maximal_depth() / (n_points-1);
-
       // The number of compositional fields
       const unsigned int n_compositional_fields = this->n_compositional_fields();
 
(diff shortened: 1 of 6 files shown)


anne-glerum:
I'll simplify the changes based on #5090 .

gassmoeller:
Thanks! I think this makes sense. I have an idea for how to simplify this further and I will open a separate PR for that. But we can go ahead with this PR anyway. But we would need a test case, something with traction BC and initial topography.

anne-glerum:
Alright, I'll add an test!

anne-glerum:
@gassmoeller I've added a test that has initial topography + bottom traction boundary conditions. Let me know if something else needs to be done.

source/boundary_traction/initial_lithostatic_pressure.cc
```diff
@@ -348,7 +416,19 @@ namespace aspect
                                             "the number of integration points. "
                                             "The lateral coordinates of the point are used to calculate "
                                             "the lithostatic pressure profile with depth. This means that "
-                                            "the depth coordinate is not used."
+                                            "the depth coordinate is not used. "
+                                            "Note that when initial topography is included, the initial "
+                                            "topography at the user-provided representative point is used "
+                                            "to compute the profile. If at other points the (initial) topography is "
+                                            "is higher, the behavior of this plugin at later timesteps depends "
```
gassmoeller, line 423: Double `is`

source/boundary_traction/initial_lithostatic_pressure.cc
```diff
@@ -219,20 +270,29 @@ namespace aspect
...
+      // case we return the deepest pressure computed, as this is the
+      // normal traction component we want to prescribe even though the
+      // depth might not equal the depth of the initial profile because the
+      // surface has been deformed.
       Tensor<1,dim> traction;
-      traction = -interpolate_pressure(position) * normal_vector;
+      if (boundary_indicator == this->get_geometry_model().translate_symbolic_boundary_name_to_id("bottom"))
+        traction = -pressure.back() * normal_vector;
+      else
+        traction = -interpolate_pressure(position) * normal_vector;
```
gassmoeller, line 290: Does this mean if traction is prescribed at side and bottom boundaries then there will be a pressure jump between the lowest point on the side boundary and the bottom boundary? That seems a bit unintuitive. What is the reasoning for always using the last point of the profile for the bottom boundary instead of computing it using the position?

gassmoeller (commented):
The code looks fine except for a small comment. Also please add a changelog entry. 
When reading this again I had one question that I marked. If you have a good answer for that we can merge.

source/boundary_traction/initial_lithostatic_pressure.cc
anne-glerum: That's a good point; I'm biased by the particular boundary conditions we use for our simulations, which only have tractions on the bottom boundary.

The reasons I think I had for taking the deepest pressure in the pressure profile for the bottom boundary:
* We always want to prescribe the same pressure on the bottom boundary, independent of the location on the bottom boundary. 
* When mesh deformation occurs, the depth returned by the geometry model for a point on the bottom boundary might not actually correspond with the deepest depth in the pressure profile. Take the box geometry model for example: 
   * The depth returned is computed as depth(x,y) = extents[z] + initial_topography(x,y) - position[z], so with respect to the initial topography.
   * If the point on the bottom boundary where we initially compute the reference profile is at t1 higher than at t0, the depth returned by the geometry model is 0 (because the actual negative depth is cut off at zero).
   * If the same point is lower than its initial topography, the depth returned will be smaller than the depth of the bottom boundary in the pressure profile.
   * So for the bottom boundary, only taking the last pressure in the profile leads to getting the pressure we want to prescribe. 

This is making me think that instead we could also improve on the definition of depth with mesh deformation enabled, like Derek worked on in #5088 .

anne-glerum:
After a discussion with @gassmoeller , I'll remove the bottom boundary specific changes, and leave only the initial surface topography at the reference location changes. A separate PR will set some constraints on which combinations of initial topography, mesh deformation and traction boundaries are allowed.

gassmoeller (approved):

gassmoeller:
Tests seem to fail :-(

anne-glerum:
Should be fixed now :)

